PR #220, opened 2022-06-15
Rpc/pending blocks

joshklop:
## Types of changes

## Testing
**Requires testing**

- [X] Yes
- [ ] No

- [X] Yes
- [ ] No


pkg/rpc/api.go
```diff
@@ -6,12 +6,17 @@ import (
...
+	dbAbi "github.com/NethermindEth/juno/internal/db/abi"
 	"github.com/NethermindEth/juno/internal/services"
+	"github.com/NethermindEth/juno/pkg/feeder"
 	"github.com/NethermindEth/juno/pkg/types"
 
 	"github.com/NethermindEth/juno/internal/log"
 )
 
+// Global feederClient that we use to request pending blocks
+var feederClient = feeder.NewClient("https://alpha-mainnet.starknet.io", "/feeder_gateway", nil)
```
stdevMac, line 18: You shouldn't have hard-coded the feederClient should be part of the config. initialized on the run of the rpc or something like that

stdevMac (changes requested):

stdevMac (approved):

PR #221, opened 2022-06-15
feat: `starknet_call` and alternative trie implementation

tshakalekholoane:
## Changes:

- Implements a different trie that will allow Cairo to query the storage successfully.
- Adds bindings to perform the `starknet_call` RPC endpoint.

## Types of changes

- [x] Build related changes

## Testing
**Requires testing**

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No

## Comments

The branch `trie/reimpl` probably renders [`vm/grpc`](https://github.com/NethermindEth/juno/tree/vm/grpc) and [`core/rpc-call`](https://github.com/NethermindEth/juno/tree/core/rpc-call) obsolete.


internal/db/state/contract.go
```diff
@@ -0,0 +1,78 @@
...
+func (m *Manager) GetContractState(hash *felt.Felt) (*state.ContractState, error) {
+	raw, err := m.stateDatabase.Get(hash.ByteSlice())
+	if err != nil {
+		// Database error
+		return nil, err
+	}
+	if raw == nil {
+		// Not found
+		return nil, nil
+	}
```
tshakalekholoane, line 21: As far as I am aware, the db package has an error for queries that result in not found errors, `ErrNotFound`. Shouldn't we be checking for that instead of examining whether `raw == nil`?

PR #245, opened 2022-06-24
Update estimate_fee in gateway and juno-cli

### LINKED ISSUE #185: feeder.go missing some feeder_gateway methods
Commands get_transaction_trace and estimate_fee from starknet (see [here](https://www.cairo-lang.org/docs/hello_starknet/cli.html)) are missing from juno/pkg/feeder/feeder.go as the Starknet Feeder Gateway has been updated to include them.

We should add them along the other avaiable functions for completeness, and to aid #128 and any other integrations which require these functions.

D-DePablos:
Fixes | Closes | Resolves #185 

## Changes:
- Adding estimate_fee with multiple function inputs

## Types of changes

- [x] Bugfix
- [x] New Feature

 
## Testing
**Requires testing**

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No -> Planning to!
 
**Comments about testing , should you have some** (optional)
In the original implementation I assumed 1 input to the function, which means view functions or those with calldata with more than 1 element don't work. 

I am also using this chance to implement the call_contract functionality.


D-DePablos:
I have updated the linked issue

stdevMac (changes requested):

stdevMac (approved):

smishy05 (approved):
Looks good to me! Thanks for correcting the comments. That helps a lot!

cmd/juno-cli/cli/estimateFee.go
```diff
@@ -13,8 +13,25 @@ import (
...
-	Args:  cobra.MinimumNArgs(2),
+	Long: `See https://www.cairo-lang.org/docs/hello_starknet/cli.html#estimate-fee.
+
+	Please note that to estimate fee for functions with multiple inputs, you should use multiple -i flags with values, 
+	in the order of the necessary inputs. The network response will tell you that the number of inputs is wrong otherwise.
+	
+	Example with multiple inputs and expected response being returned:
+	
+	Input:
+	  juno-cli cx estimate_fee 0x0003a4d1be6ae6cccc7b6cc3cc16bbfd092c2a724ffe6be86c7b5b5fe6ec11d0 increase_decrease --network goerli -p -i 1 -i 2
```
adriantpaez, line 24: I think `cx` shouldn't be here.`juno-cli cx estimate_fee ...` doesn't work for me, `juno-cli estimate_fee ...` yes

adriantpaez (changes requested):

stdevMac (approved):

PR #252, opened 2022-06-28
Feat/health check229

### LINKED ISSUE #229: Health Check for Node
We should be able to provide a RPC endpoint, and API endpoint that basically knows when the node is running properly, the db connections are ok, and the feeder gateway is responding well.

D-DePablos:
Resolves #229 

## Changes:
- Add Health check for user so they can confirm whether node is synced, syncing, or running at all.

## Types of changes
- [ ] Documentation Update
- [ ] Build related changes

## Testing
**Requires testing**

- [x] Yes
- [ ] No

- [ ] Yes
- [x] No (but plan to!)


D-DePablos:
Putting this on hold for a bit, if somebody else would like to have a go, please feel free. My main trouble is consulting the block manager service from the health check service

PR #253, opened 2022-06-28
Improve RPC

### LINKED ISSUE #248: Improve RPC
Currently, we only support a subset of the specification of JSONRPC 2.0 and we don't support subscriptions.

To consider close this issue, we should:
- [x] Support JSONRPC 2.0 [spec](https://www.jsonrpc.org/specification)
- [x] Notifications
- [ ] Subscriptions

adriantpaez:
Resolves #248

## Changes:
- Add JSON-RPC 2.0 implementation under the `pkg/jsonrpc` package
- Add HTTP provider to use the RPC over HTTP under the `pkg/jsonrpc/providers` package
- Add the new StarkNet RPC implementation (support the same functionalities) to the `internal/rpc` package

## Types of changes

- [ ] Documentation Update
- [ ] Build related changes

## Testing
**Requires testing**

- [x] Yes
- [ ] No

- [x] Yes
- [ ] No


pkg/jsonrpc/error.go
```diff
@@ -0,0 +1,81 @@
...
+	Data    interface{} `json:"data,omitempty"`
+}
+
+// Error implements the error interface.
+func (err *RpcError) Error() string {
+	return err.Message
+}
+
+func newErrParseError(data interface{}) *RpcError {
+	return &RpcError{
```
tshakalekholoane, line 36: Might prefer the newer `any` instead of the older empty interface here.

tshakalekholoane (approved):
LGTM. See suggestions.

PR #254, opened 2022-06-30
Added functionality for closing the app on absence of primary services

### LINKED ISSUE #205: Proccess Handler Bug
**Describe the bug**
When we initialize the database services on the `root.go` file, like this:

```golang
handler.Add("TransactionService", services.TransactionService.Run, services.TransactionService.Close)
handler.Add("BlockService", services.BlockService.Run, services.BlockService.Close)
```
the application doesn't stop because it can't ever reach the `handler.Close()` line because they are stuck on the`select{}` line on the `handler.Run()`.

**To Reproduce**
1. Go to `cmd/juno/cli/root.go`
2. Paste this code on line 59-60 (after RPC initialization and before `log.Default.Info("Starting all processes...")` line)

```golang
handler.Add("TransactionService", services.TransactionService.Run, services.TransactionService.Close)
handler.Add("BlockService", services.BlockService.Run, services.BlockService.Close)
```
3. Make sure the RPC is disabled on the configuration file (we only want the database services active).
4. Run Juno with `make all` command
5. You may see something like this:
```
2022-06-08T09:22:32.988-0400    INFO    cli/root.go:109 Using config file:      {"File": "/path/to/juno/juno.yaml"}
2022-06-08T09:22:32.988-0400    INFO    cli/root.go:127 Config values.  {"Database Path": "/path/to/juno", "Rpc Port": 8080, "Rpc Enabled": false}
2022-06-08T09:22:32.988-0400    INFO    cli/root.go:64  Starting all processes...
2022-06-08T09:22:32.988-0400    INFO    Block Service   services/service.go:34  Service started
2022-06-08T09:22:32.989-0400    INFO    Transaction Service     services/service.go:34  Service started
```
6. After that the only way to close the app is with a SIGTERM or SIGKILL

**Expected behavior**
I expect the application to exit immediately because the database services are "internal" services, so they cannot block the application if any other service, other than a primary one (such as RPC), is not running.

smishy05:
Resolves #205 

## Changes:
- Added a `PrimaryServiceCheck` to `root.go` which checks if primary services are running or not
- Added the condition for checking running of the service while closing it. If the app is not running and it is closed, then it leads to `nil reference` error.

## Types of changes

- [ ] Documentation Update
- [ ] Build related changes

## Testing
**Requires testing**

- [ ] Yes
- [x] No

### DIFF
--- a/internal/process/process.go
+++ b/internal/process/process.go
@@ -89,3 +89,16 @@ func (h *Handler) Run() {
 	// Create an infinite loop.
 	select {}
 }
+
+func (h *Handler) PrimaryServiceChecker() bool {
+	primarypkg := false
+	for _, proc := range h.subprocs {
+		if proc.id == "Starknet Synchronizer" {
+			primarypkg = true
+		}
+	}
+	if !primarypkg {
+		log.Default.Info("No primary processes running. Closing the app.")
+	}
+	return primarypkg
+}
--- a/internal/process/process.go
+++ b/internal/process/process.go
@@ -17,10 +17,11 @@ type stopFunc func(ctx context.Context)
 
 // process contains information and required functions of each process
 type process struct {
-	id   string
-	err  error
-	run  runFunc
-	stop stopFunc
+	id        string
+	important bool
+	err       error
+	run       runFunc
+	stop      stopFunc
 }
 
 // Processer defines common methods for a context processor.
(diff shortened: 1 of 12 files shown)


D-DePablos:
Could you remove the pb file changes?

internal/process/process.go
```diff
@@ -89,3 +91,17 @@ func (h *Handler) Run() {
...
+func (h *Handler) PrimaryServiceChecker() int {
+	// notest
+	primarypkg := 0
+	for _, proc := range h.subprocs {
+		if proc.important {
+			primarypkg = primarypkg + 1
+		}
+	}
+	if primarypkg == 0 {
+		log.Default.Info("No primary processes running. Closing the app.")
```
D-DePablos, line 104: good explanation msg

D-DePablos (changes requested):
Only thing is to remove changes to protobuf files (state.pb, block.pb, abi.pb, transaction.pb)

smishy05:
> Only thing is to remove changes to protobuf files (state.pb, block.pb, abi.pb, transaction.pb)

Done in the new commit. Please check

PR #257, opened 2022-07-02
refactor: felt package

joshklop:
## Changes:
- Refactor felt package. 
- Refactor types package to use felt package.

## Types of changes

## Testing
**Requires testing**

- [X] Yes
- [ ] No

- [ ] Yes
- [X] No (WIP)


internal/db/abi/abi.pb.go
```diff
@@ -1,7 +1,7 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
 // 	protoc-gen-go v1.28.0
-// 	protoc        v3.17.3
```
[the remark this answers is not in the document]
joshklop: I ran `make generate` locally.

internal/db/transaction/transaction.pb.go
```diff
@@ -1,7 +1,7 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
 // 	protoc-gen-go v1.28.0
-// 	protoc        v3.17.3
```
jmederosalvarado, line 4: protoc version changed?

internal/db/block/block.pb.go
```diff
@@ -1,7 +1,7 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
 // 	protoc-gen-go v1.28.0
-// 	protoc        v3.17.3
```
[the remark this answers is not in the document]
joshklop: See above.

internal/db/state/state.pb.go
```diff
@@ -1,7 +1,7 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
 // 	protoc-gen-go v1.28.0
-// 	protoc        v3.17.3
```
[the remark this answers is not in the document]
joshklop: See above

internal/db/transaction/transaction.pb.go
joshklop: See above

joshklop:
Below is some additional context for this PR to help expedite reviews.

The PR provides an efficient implementation of the felt type. Much of the felt package is automatically generated by the `pkg/felt/generator/main.go` file, which uses the [gnark-crypto](https://github.com/ConsenSys/gnark-crypto) library.

The only non-generated files in the felt package are: 

* `felt_additional_test.go` - provides coverage for the auto-generated `felt.go` file, beyond the coverage that the generated `felt_test.go` file provides.
* `other.go` and `other_test.go` - additional helper functions useful in other areas of Juno's codebase.

Pretty much all of the changes outside of the `felt` directory simply modify code to use the new felt type rather than `types.Felt` and `big.Int`.

pkg/trie/trie_test.go
```diff
@@ -3,24 +3,25 @@ package trie
 import (
 	"encoding/json"
 	"fmt"
-	"math/big"
+	//"math/big"
```
adriantpaez, line 6: Remove this comment too 😅

adriantpaez (changes requested):
Looks good!

adriantpaez (approved):

pkg/felt/other.go
```diff
@@ -0,0 +1,150 @@
...
+	"github.com/NethermindEth/juno/pkg/common"
+)
+
+var (
+	Felt0 = new(Felt).SetZero()
+	Felt1 = new(Felt).SetOne()
+	Felt2 = new(Felt).SetUint64(2)
+	Felt3 = new(Felt).SetUint64(3)
+)
+
```
tshakalekholoane, line 13: I think the burden of these should be on the client. It doesn't make sense to impose this cost on everyone that wants to use the package especially when we have convenience methods defined that archive this without the significant memory overhead.

pkg/felt/other.go
```diff
@@ -0,0 +1,150 @@
...
+// ByteSlice() is exactly the same as Marshal() except that it returns
+// null when z is null. This mimics the behavior of big.Int.
+func (z *Felt) ByteSlice() []byte {
```
tshakalekholoane, line 16: Let's rename this to [`Bytes()`](https://pkg.go.dev/math/big#Int.Bytes) for consistency not only with `big.Int` but with the rest of the standard library i.e. [`Buffer.Bytes()`](https://pkg.go.dev/bytes@go1.18.4#Buffer.Bytes).

tshakalekholoane (changes requested):
Additionally, if other.go contains helpers, let's name the file to a more descriptive name i.e. helpers.go.

In terms of the placement of the generator, if we are going to have it, then to keep in line with the [existing project structure](https://github.com/golang-standards/project-layout#cmd), I would recommend pkg/felt/internal/cmd/generator/.

Otherwise, it looks good to me.

pkg/felt/other.go
joshklop: I agree this is preferable, but `Bytes()` is [already defined](https://github.com/NethermindEth/juno/blob/fe2c5e5b5eac5dafab577922a62e5a34b8f2bd8b/pkg/felt/felt.go#L848-L856). It has some inconvenient properties: it returns an array and crashes when the caller is nil (unlike `big.Int`). `felt.Marshal` returns a byte slice, but crashes when caller is nil.

stdevMac (approved):

tshakalekholoane (changes requested):
Let's nest the binary inside another folder name generator as I suggested i.e. pkg/felt/internal/cmd/generator/ so that it has the name generator instead of cmd.

pkg/felt/other.go
tshakalekholoane: Ah. I see.

stdevMac (approved):

tshakalekholoane (approved):

PR #263, opened 2022-07-04
Unexport point functions as not used elsewhere

### LINKED ISSUE #262: Unexport point methods
[This](https://github.com/NethermindEth/juno/blob/cbd368b222ac19a69e75e8ab82f01443ad21f590/pkg/crypto/pedersen/points.go#L2062) and any other method that is defined on a point do not have to be exported.

D-DePablos:
Resolves #262


tshakalekholoane (changes requested):
[Missed one bit](https://github.com/NethermindEth/juno/blob/5387a015dc1a70061f34a82c999a3052e262004e/pkg/crypto/pedersen/pedersen.go#L37)

D-DePablos (commented):
Done

tshakalekholoane (changes requested):
Doc strings as well. It would be easier if you just used the rename feature of your LSP in VSCode it's usually mapped to F2 😅.

tshakalekholoane (approved):

PR #267, opened 2022-07-04
Speed up fail rest tests

### LINKED ISSUE #256: Reduce overall testing time of juno through changes in REST tests
Tests currently take a long time because the rest api tests which are meant to fail (within /pkg/rest/tests/rest_test.go) take a long time to timeout. 

This means that, while the rest of the test suite finishes in less than 10 seconds, all the Test(...)Fail take ~ 10s each, 13 times, incrementing this substantially and wasting time.

It would be good to have 1 timeout test with the normal rate, and the other failing tests to have a much smaller timeout or use a different method to check for failure than to wait out the full 10s.

v4lproik:
Resolves #256 

## Changes:
- Extracted the hardcoded retry mechanism implemented in the http Do function.
- Allows the http client to take a retry mechanism as parameter which will then be used in the Do function.
- Test time before this PR
```
ok  	github.com/NethermindEth/juno/pkg/rest/tests	123.929s
```
- Test time after this PR
```
ok  	github.com/NethermindEth/juno/pkg/rest/tests	5.299s
```

## Types of changes

- [ ] Documentation Update
- [ ] Build related changes

## Testing
**Requires testing**

- [X] Yes
- [] No

- [X] Yes
- [ ] No

### DIFF
--- a/pkg/rest/tests/rest_test.go
+++ b/pkg/rest/tests/rest_test.go
@@ -110,7 +110,7 @@ func TestRestClientRetryFunction(t *testing.T) {
 	restHandler.GetBlock(rr, req)
 
 	// Assert error message
-	assert.DeepEqual(t, rr.Body.String(), "Invalid request body error:"+errorMessage)
+	assert.DeepEqual(t, rr.Body.String(), "Invalid request body error: "+errorMessage)
 }
 
 //---------------------------------------------
@@ -1028,7 +1028,7 @@ func TestGetBlockHandlerFeederFail(t *testing.T) {
 	failRestHandler.GetBlock(rr, req)
 
 	// Assert error message
-	assert.DeepEqual(t, rr.Body.String(), "Invalid request body error:feeder gateway failed")
+	assert.DeepEqual(t, rr.Body.String(), "Invalid request body error: feeder gateway failed")
 }
 
 // TestGetCodeHandlerFeederFail
@@ -1055,7 +1055,7 @@ func TestGetCodeHandlerFeederFail(t *testing.T) {
 	failRestHandler.GetCode(rr, req)
 
 	// Assert error message
-	assert.DeepEqual(t, rr.Body.String(), "Invalid request body error:feeder gateway failed")
+	assert.DeepEqual(t, rr.Body.String(), "Invalid request body error: feeder gateway failed")
 }
 
 // TestGetStorageAtHandlerFeederFail
@@ -1083,7 +1083,7 @@ func TestGetStorageAtHandlerFeederFail(t *testing.T) {
 	failRestHandler.GetStorageAt(rr, req)
 
 	// Assert error message
-	assert.DeepEqual(t, rr.Body.String(), "Invalid request body error:feeder gateway failed")
+	assert.DeepEqual(t, rr.Body.String(), "Invalid request body error: feeder gateway failed")
 }
 
 // TestGetTransactionStatusHandlerFeederFail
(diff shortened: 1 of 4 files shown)


stdevMac (approved):

stdevMac:
Just some coverage issues, you can take a look deeply if you look at [here](https://gojuno.xyz/docs/testing/coverage). Very well done work! Thanks for your contributions!

v4lproik:
I added a test that fails when using the default client implementation and so by extension the default retry mechanism. Coverage should be ok now.

Test time is now at `ok  	github.com/NethermindEth/juno/pkg/rest/tests	18.872s`

pkg/rest/tests/rest_test.go
```diff
@@ -76,6 +81,35 @@ func TestRestClient(t *testing.T) {
...
+	rr := httptest.NewRecorder()
+
+	errorMessage := "feeder gateway failed"
+	httpClient.DoReturns(nil, fmt.Errorf("%s", errorMessage))
+
+	// Send Request expecting an error
+	restHandler.GetBlock(rr, req)
+
+	// Assert error message
+	assert.DeepEqual(t, rr.Body.String(), fmt.Sprintf("Invalid request body error:%s", errorMessage))
```
tshakalekholoane, line 110: There's no string formating going on here so let's do a simple string concat.

tshakalekholoane (changes requested):
See comments. Otherwise great work! We appreciate your contribution.

pkg/rest/tests/rest_test.go
```diff
@@ -20,22 +21,28 @@ import (
...
-	failRestHandler = rest.RestHandler{}
+	httpClient         = &feederfakes.FakeHttpClient{}
+	failHttpClient     = &feederfakes.FailHttpClient{}
+	client             *feeder.Client
+	failClient         *feeder.Client
+	restHandler        = rest.RestHandler{}
+	failRestHandler    = rest.RestHandler{}
+	failRequestTimeout = time.Millisecond * 400
+	baseUrl            = "https://localhost:8100"
+	baseApi            = "/feeder_gateway/"
```
tshakalekholoane, line 32: Use consistent case for acronyms i.e. `baseURL` instead of `baseUrl`. See [guide](https://github.com/golang/go/wiki/CodeReviewComments#initialisms).

pkg/rest/tests/rest_test.go
```diff
@@ -76,6 +83,35 @@ func TestRestClient(t *testing.T) {
...
+	rq := req.URL.Query()
+	rq.Add("blockNumber", "1")
+	rq.Add("blockHash", "hash")
+	req.URL.RawQuery = rq.Encode()
+
+	// Build Response Object
+	rr := httptest.NewRecorder()
+
+	errorMessage := "feeder gateway failed"
+	httpClient.DoReturns(nil, fmt.Errorf("%s", errorMessage))
```
tshakalekholoane, line 106: Again, no formatting done so we could probably use `errors.New` here.

pkg/rest/tests/rest_test.go
```diff
@@ -76,6 +83,35 @@ func TestRestClient(t *testing.T) {
 	cancel()
 }
 
+// TestRestClientRetryFunction
+func TestRestClientRetryFunction(t *testing.T) {
+	queryStr := "http://localhost:8100/feeder_gateway/get_block"
```
tshakalekholoane, line 88: Use string concat with newly created variables.

pkg/rest/tests/rest_test.go
```diff
@@ -76,6 +83,35 @@ func TestRestClient(t *testing.T) {
...
+	rr := httptest.NewRecorder()
+
+	errorMessage := "feeder gateway failed"
+	httpClient.DoReturns(nil, fmt.Errorf("%s", errorMessage))
+
+	// Send Request expecting an error
+	restHandler.GetBlock(rr, req)
+
+	// Assert error message
+	assert.DeepEqual(t, rr.Body.String(), "Invalid request body error:"+errorMessage)
```
tshakalekholoane, line 112: Insert space after colon.

tshakalekholoane (changes requested):

v4lproik:
Thanks for the review @tshakalekholoane. I also ended up extracting query url and fixing the error formatting in the handlers.

tshakalekholoane (approved):

joshklop (approved):

PR #268, opened 2022-07-04
Improve database error management

adriantpaez:
## Changes:
- Change database panics for return errors

## Types of changes

- [ ] Documentation Update
- [x] Refactoring (no functional changes, no api changes)
- [ ] Build related changes

## Testing
**Requires testing**

- [ ] Yes
- [x] No

- [ ] Yes
- [ ] No

### DIFF
--- a/internal/db/abi/manager.go
+++ b/internal/db/abi/manager.go
@@ -41,8 +41,7 @@ func (m *Manager) PutABI(contractAddress string, abi *Abi) error {
 	if err != nil {
 		return err
 	}
-	err = m.database.Put(key, value)
-	if err != nil {
+	if err := m.database.Put(key, value); err != nil {
 		return err
 	}
 	return nil
--- a/internal/db/abi/manager.go
+++ b/internal/db/abi/manager.go
@@ -1,17 +1,8 @@
 package abi
 
 import (
-	"errors"
-	"fmt"
-
 	"github.com/NethermindEth/juno/internal/db"
 	"google.golang.org/protobuf/proto"
-)
-
-var (
-	DbError        = errors.New("database error")
-	UnmarshalError = errors.New("unmarshal error")
-	MarshalError   = errors.New("marshal error")
 )
 
 // Manager is a database to store and get the contracts ABI.
@@ -26,45 +17,35 @@ func NewABIManager(database db.Database) *Manager {
 
 // GetABI gets the ABI associated with the contract address. If the ABI does
 // not exist, then returns nil without error.
-func (m *Manager) GetABI(contractAddress string) *Abi {
+func (m *Manager) GetABI(contractAddress string) (*Abi, error) {
 	// Build the key from contract address
 	key := []byte(contractAddress)
 	// Query to database
 	data, err := m.database.Get(key)
 	if err != nil {
-		if db.ErrNotFound == err {
-			return nil
-		}
-		// notest
-		panic(any(fmt.Errorf("%w: %s", DbError, err)))
-	}
-	if data == nil {
-		// notest
-		return nil
+		return nil, err
 	}
 	// Unmarshal the data from database
 	abi := new(Abi)
 	if err := proto.Unmarshal(data, abi); err != nil {
-		// notest
-		panic(any(fmt.Errorf("%w: %s", UnmarshalError, err.Error())))
+		return nil, err
 	}
-	return abi
+	return abi, nil
 }
 
 // PutABI puts the ABI to the contract address.
-func (m *Manager) PutABI(contractAddress string, abi *Abi) {
+func (m *Manager) PutABI(contractAddress string, abi *Abi) error {
 	// Build the key from contract address
 	key := []byte(contractAddress)
 	value, err := proto.Marshal(abi)
 	if err != nil {
-		// notest
-		panic(any(fmt.Errorf("%w: %s", MarshalError, err)))
+		return err
 	}
 	err = m.database.Put(key, value)
 	if err != nil {
-		// notest
-		panic(any(fmt.Errorf("%w: %s", DbError, err.Error())))
+		return err
 	}
+	return nil
 }
 
 // Close closes the associated database
--- a/internal/db/abi/manager_test.go
+++ b/internal/db/abi/manager_test.go
@@ -18,10 +18,16 @@ func TestManager(t *testing.T) {
 	manager := NewABIManager(database)
 
 	for address, abi := range abis {
-		manager.PutABI(address, abi)
-		abi2 := manager.GetABI(address)
-		if abi2 == nil {
-			t.Errorf("ABI with key %s not found after insertion with the same key", address)
+		if err := manager.PutABI(address, abi); err != nil {
+			t.Error(err)
+		}
+		abi2, err := manager.GetABI(address)
+		if err != nil {
+			if db.IsNotFound(err) {
+				t.Errorf("ABI not found for address %s", address)
+			} else {
+				t.Error(err)
+			}
 		}
 		if !abi.Equal(abi2) {
 			t.Errorf("ABI are not equal after Put-Get operations, address: %s", address)
--- a/internal/db/abi/manager_test.go
+++ b/internal/db/abi/manager_test.go
@@ -1,6 +1,7 @@
 package abi
 
 import (
+	"errors"
 	"testing"
 
 	"github.com/NethermindEth/juno/internal/db"
@@ -23,11 +24,10 @@ func TestManager(t *testing.T) {
 		}
 		abi2, err := manager.GetABI(address)
 		if err != nil {
-			if db.IsNotFound(err) {
+			if errors.Is(err, db.ErrNotFound) {
 				t.Errorf("ABI not found for address %s", address)
-			} else {
-				t.Error(err)
 			}
+			t.Error(err)
 		}
 		if !abi.Equal(abi2) {
 			t.Errorf("ABI are not equal after Put-Get operations, address: %s", address)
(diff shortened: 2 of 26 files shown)


internal/db/abi/manager.go
```diff
@@ -26,45 +17,35 @@ func NewABIManager(database db.Database) *Manager {
...
 	// Build the key from contract address
 	key := []byte(contractAddress)
 	value, err := proto.Marshal(abi)
 	if err != nil {
-		// notest
-		panic(any(fmt.Errorf("%w: %s", MarshalError, err)))
+		return err
 	}
 	err = m.database.Put(key, value)
 	if err != nil {
```
tshakalekholoane, line 45: I suggest using the `if err = ... ; err != nil {}` idiom here.

internal/db/block/block_test.go
```diff
@@ -41,18 +41,29 @@ func TestManager(t *testing.T) {
...
-		returnedBlock = manager.GetBlockByNumber(block.BlockNumber)
-		if returnedBlock == nil {
+		returnedBlock, err = manager.GetBlockByNumber(block.BlockNumber)
+		if err != nil {
+			if db.IsNotFound(err) {
+				t.Errorf("unexpected nil after search for block with number %d", block.BlockNumber)
+			} else {
+				t.Error(err)
+			}
 			t.Errorf("unexpected nil after search for block with number %d", block.BlockNumber)
```
tshakalekholoane, line 67: This line is never reached. Typo? Also, the `else` clause can be omitted.

internal/db/block/manager.go
```diff
@@ -20,82 +20,68 @@ func NewManager(database db.Database) *Manager {
...
 	numberKey := buildNumberKey(block.BlockNumber)
 	// Encode the block as []byte
 	rawValue, err := marshalBlock(block)
 	if err != nil {
-		panic(any(err))
+		return err
 	}
 	// TODO: Use transaction?
 	// Save (hashKey, block)
 	err = manager.database.Put(hashKey, rawValue)
```
tshakalekholoane, line 75: The only return value is an error so again I suggest the `if err = ...` idiom here 😅.

internal/db/state/storage.go
```diff
@@ -20,33 +18,29 @@ func (s *Storage) Update(other *Storage) {
...
-	}
-	// Check not found
-	if rawData == nil {
-		// notest
-		return nil
+		return nil, err
 	}
 	value := new(Storage)
 	err = proto.Unmarshal(rawData, value)
 	if err != nil {
```
tshakalekholoane, line 28: Again, I suggest using the if err = ... ; err != nil {} idiom here.

internal/db/transaction/manager.go
```diff
@@ -25,75 +24,59 @@ func NewManager(txDb, receiptDb db.Database) *Manager {
...
-func (m *Manager) PutTransaction(txHash types.TransactionHash, tx types.IsTransaction) {
+func (m *Manager) PutTransaction(txHash types.TransactionHash, tx types.IsTransaction) error {
 	rawData, err := marshalTransaction(tx)
 	if err != nil {
-		// notest
-		log.Default.With("error", err).Panic("error marshalling Transaction")
+		return err
 	}
 	err = m.txDb.Put(txHash.Bytes(), rawData)
 	if err != nil {
```
tshakalekholoane, line 33: Here too, I suggest using the if err = ... ; err != nil {} idiom.

internal/db/abi/manager_test.go
```diff
@@ -18,10 +18,16 @@ func TestManager(t *testing.T) {
...
+		if err := manager.PutABI(address, abi); err != nil {
+			t.Error(err)
+		}
+		abi2, err := manager.GetABI(address)
+		if err != nil {
+			if db.IsNotFound(err) {
+				t.Errorf("ABI not found for address %s", address)
+			} else {
+				t.Error(err)
+			}
```
tshakalekholoane, line 30: `else` clause can be omitted e.g.

```go
  ...
  t.Errorf("ABI not found for address %s", address)
}
t.Error(err)
```

internal/db/block/block_test.go
```diff
@@ -41,18 +41,29 @@ func TestManager(t *testing.T) {
...
-		returnedBlock := manager.GetBlockByHash(key)
-		if returnedBlock == nil {
-			t.Errorf("unexpected nil after search for block with hash %s", block.BlockHash)
+		returnedBlock, err := manager.GetBlockByHash(key)
+		if err != nil {
+			if db.IsNotFound(err) {
+				t.Errorf("unexpected nil after search for block with hash %s", block.BlockHash)
+			} else {
+				t.Error(err)
+			}
```
tshakalekholoane, line 54: `else` clause can be omitted. See comment above for an example.

internal/db/state/code_test.go
```diff
@@ -43,8 +43,17 @@ func TestManager_Code(t *testing.T) {
...
+		if err := manager.PutCode(code.Address, code.Code); err != nil {
+			t.Error(err)
+		}
+		obtainedCode, err := manager.GetCode(code.Address)
+		if err != nil {
+			if db.IsNotFound(err) {
+				t.Errorf("Code not found for address %s", code.Address)
+			} else {
+				t.Error(err)
+			}
```
tshakalekholoane, line 55: `else` clause can be omitted. See comment above for an example.

internal/db/transaction/manager_test.go
```diff
@@ -96,11 +98,20 @@ func TestManager_GetTransaction(t *testing.T) {
...
 	// Get all the transactions and compare
 	for _, tx := range txs {
-		outTx := manager.GetTransaction(tx.GetHash())
+		outTx, err := manager.GetTransaction(tx.GetHash())
+		if err != nil {
+			if db.IsNotFound(err) {
+				t.Errorf("Transaction %s not found", tx.GetHash())
+			} else {
+				t.Error(err)
+			}
```
tshakalekholoane, line 113: `else` clause can be omitted. See comment above for an example.

pkg/rpc/api.go
```diff
@@ -339,13 +368,25 @@ func (HandlerRPC) StarknetGetTransactionByBlockHashAndIndex(c context.Context, b
...
 		}
 		txHash := block.TxHashes[index]
-		txn := services.TransactionService.GetTransaction(txHash)
+		txn, err := services.TransactionService.GetTransaction(txHash)
+		if db.IsNotFound(err) {
+			// notest
+			return nil, fmt.Errorf("transaction not found")
+		} else if err != nil {
+			return nil, fmt.Errorf("database error: %v", err)
+		}
```
tshakalekholoane, line 389: From the comment above.

> It seems like we might want to check first whether an error occured before asserting its kind and not the other way around just as you did above. I also don't see why the assertion has to be made if there is not way for the client to know which kind of error occured outside of say parsing the first few characters of the error string.

pkg/rpc/api.go
```diff
@@ -414,11 +455,16 @@ func (HandlerRPC) StarknetGetCode(
...
-	code := services.StateService.GetCode(contractAddress.Bytes())
-	if code == nil {
+	code, err := services.StateService.GetCode(contractAddress.Bytes())
+	if db.IsNotFound(err) {
 		// notest
 		return nil, fmt.Errorf("code not found")
+	} else if err != nil {
+		// notest
+		return nil, fmt.Errorf("database error: %v", err)
 	}
```
tshakalekholoane, line 468: From the comment above.

> It seems like we might want to check first whether an error occured before asserting its kind and not the other way around just as you did above. I also don't see why the assertion has to be made if there is not way for the client to know which kind of error occured outside of say parsing the first few characters of the error string.

internal/services/transactions_test.go
```diff
@@ -108,11 +110,20 @@ func TestManager_GetTransaction(t *testing.T) {
...
 	// Get all the transactions and compare
 	for _, tx := range txs {
-		outTx := TransactionService.GetTransaction(tx.GetHash())
+		outTx, err := TransactionService.GetTransaction(tx.GetHash())
+		if err != nil {
+			if db.IsNotFound(err) {
+				t.Errorf("transaction not found: %s", tx.GetHash().String())
+			} else {
+				t.Error(err)
+			}
```
tshakalekholoane, line 125: `else` clause can be omitted. See comment above for an example.

tshakalekholoane (changes requested):

internal/db/abi/manager_test.go
```diff
@@ -18,10 +18,16 @@ func TestManager(t *testing.T) {
...
-		manager.PutABI(address, abi)
-		abi2 := manager.GetABI(address)
-		if abi2 == nil {
-			t.Errorf("ABI with key %s not found after insertion with the same key", address)
+		if err := manager.PutABI(address, abi); err != nil {
+			t.Error(err)
+		}
+		abi2, err := manager.GetABI(address)
+		if err != nil {
+			if db.IsNotFound(err) {
```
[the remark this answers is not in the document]
stdevAdrianPaez: There is no difference, I can remove the `db.IsNotFound` and use `errors.Is` directly, which can improve the code readability

tshakalekholoane (changes requested):
See [unresolved conversation](https://github.com/NethermindEth/juno/pull/268#discussion_r913839812).

internal/db/block/manager.go
adriantpaez: In this case I need the `rawValue` below and I don't want to use an else statement there

joshklop (approved):

tshakalekholoane (approved):

PR #270, opened 2022-07-05
Improve panics

### LINKED ISSUE #227: Remove Panics of the app
**Describe the bug**
Each time the app encounters an error that we should not recover, instead of Panic, and printing all the stack traces, we should handle that error, print an Error log, and close properly the app.

There should not exist `panic`,

v4lproik:
Resolves #227 

## Changes:
- [X] Introduce env variable in config file
- [X] Initiate the logger according the env variable value so the devs can use Dpanic instead
- [ ] Remove panic instructions in the code where it should not panic
- [ ] Enforce rule in CI/CD that only allows panic() at bootstrap via global logger
- [ ] Add goroutine catching global panic (Useful when memory leak occurs and so on...)

## Types of changes

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation Update
- [ ] Build related changes

## Testing
**Requires testing**

- [X] Yes
- [ ] No

- [ ] Yes
- [ ] No

### DIFF
--- a/cmd/juno-cli/cli/root.go
+++ b/cmd/juno-cli/cli/root.go
@@ -11,7 +11,7 @@ import (
 
 	"github.com/NethermindEth/juno/internal/config"
 	"github.com/NethermindEth/juno/internal/errpkg"
-	"github.com/NethermindEth/juno/internal/log"
+	. "github.com/NethermindEth/juno/internal/log"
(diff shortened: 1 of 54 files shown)


v4lproik:
I am closing this PR as it's a duplicate of #268.

PR #278, opened 2022-07-09
Add cache

adriantpaez:
## Changes:
- Add the `internal/cache` package with the `Cache` interface and the `LRUCache` implementation

## Types of changes

## Testing
**Requires testing**

- Yes

- Yes

### DIFF
--- a/internal/cache/key.go
+++ b/internal/cache/key.go
@@ -1,0 +1,9 @@
+package cache
+
+import "crypto/sha1"
+
+type chacheKey [sha1.Size]byte
+
+func newCacheKey(v []byte) chacheKey {
+	return sha1.Sum(v)
+}
--- a/internal/cache/key.go
+++ b/internal/cache/key.go
@@ -1,9 +1,0 @@
-package cache
-
-import "crypto/sha1"
-
-type chacheKey [sha1.Size]byte
-
-func newCacheKey(v []byte) chacheKey {
-	return sha1.Sum(v)
-}
--- a/internal/cache/lru.go
+++ b/internal/cache/lru.go
@@ -6,10 +6,10 @@ import (
 )
 
 type cacheNode struct {
-	Key   uint64
-	Value []byte
-	Next  *cacheNode
-	Prev  *cacheNode
+	key   uint64
+	value []byte
+	next  *cacheNode
+	prev  *cacheNode
 }
 
 // LRUCache is a cache with the least-recently-used policy.
@@ -47,7 +47,7 @@ func (c *LRUCache) Put(k []byte, v []byte) {
 
 	// Build the cache key and node
 	key := c.key(k)
-	node := &cacheNode{Key: key, Value: v}
+	node := &cacheNode{key: key, value: v}
 
 	// Cache is empty
 	if c.count == 0 {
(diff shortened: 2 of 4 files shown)


stdevMac (approved):

internal/cache/key.go
```diff
@@ -0,0 +1,9 @@
+package cache
+
+import "crypto/sha1"
+
+type chacheKey [sha1.Size]byte
+
+func newCacheKey(v []byte) chacheKey {
+	return sha1.Sum(v)
```
tshakalekholoane, line 8: Is it important that we use a cryptographic hash? It seems to me the only property we care about here is that the key is unique so we might be better off with something less costly such as the [maphash](https://pkg.go.dev/hash/maphash) or [fnv](https://pkg.go.dev/hash/fnv) from the standard library.

tshakalekholoane (commented):
To keep consistent with Go conventions maybe instead of `Count` we could do `Len` (see [`bytes.Buffer.Len`](https://pkg.go.dev/bytes#Buffer.Len)) and similarly with the `Capacity` (see [`bytes.Buffer.Cap`](https://pkg.go.dev/bytes#Buffer.Cap)). Otherwise this looks great!

internal/cache/key.go
adriantpaez: I will change it to maphash

IronGauntlets:
Since this is a new package can we please move away from the convention of using internal packages? I would suggest adding this package under `juno/cache`.

internal/cache/lru.go
```diff
@@ -0,0 +1,119 @@
...
+}
+
+// LRUCache is a cache with the least-recently-used policy.
+type LRUCache struct {
+	hashMap  map[chacheKey]*cacheNode
+	start    *cacheNode
+	end      *cacheNode
+	count    int
+	capacity int
+	lock     sync.Mutex
```
IronGauntlets, line 21: Is the cache being used somewhere that is accessed by multiple go routine? 

If not, I would recommend not adding a mutex at this stage.

internal/cache/lru.go
```diff
@@ -0,0 +1,119 @@
...
+// the least-recently-used key-value pair is removed.
+func (c *LRUCache) Put(k []byte, v []byte) {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+
+	// Build the cache key and node
+	key := newCacheKey(k)
+	node := &cacheNode{Key: key, Value: v}
+
+	// Cache is empty
```
IronGauntlets, line 42: Probably unnecessary. Otherwise, please expand with what happens when cache is empty

adriantpaez:
> Since this is a new package can we please move away from the convention of using internal packages? I would suggest adding this package under `juno/cache`. Where the cache interface is stored under file `cache.go` and the lru implementation is stored under `juno/cache/lru` directory.

@IronGauntlets Sounds good to me, but because now the project structure is under the `internal` and `pkg` directories, we need to agree first if we will continue with that structure or move to the project root because I can't break the current structure

internal/cache/lru.go
adriantpaez: I think cache will be used for multiple RPC concurrently threads with certain managers

jmederosalvarado: > Is the cache being used somewhere that is accessed by multiple go routine?
> If not, I would recommend not adding a mutex at this stage.

The main purpose of the cache is to have it as a layer between the database and initially the trie, but probably all database access will use a cache, having it not thread-safe would be the same as having the db not thread-safe.

jmederosalvarado (changes requested):

PR #279, opened 2022-07-11
Change config easily from Terminal (TUI)

### LINKED ISSUE #271: Generate config dinamically
**Is your feature request related to a problem? Please describe.**
Generate a command like `juno init` that gives the user the ability to iterate over the options and create a new config file

**Additional context**
Currently, we are providing a config file as it is and then the user has to modify that file to make it work as expected or use env vars

A solution that will be good to add is using https://github.com/rivo/tview to generate a small intuitive interface to generate the config-file

D-DePablos:
Fixes | Closes | Resolves #271 

## Changes:
- Add a tview-based Go Terminal User Interface to make changes to the Juno config file

## Types of changes
- [ ] Documentation Update
- [ ] Build related changes

## Testing
- Requires user behaviour testing, multiOS check


cmd/juno/cli/createDynamicConfig.go
```diff
@@ -0,0 +1,186 @@
+package cli
```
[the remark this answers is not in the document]
stdevMac: Something like this https://medium.com/@kdnotes/golang-naming-rules-and-conventions-8efeecd23b68 might be useful

cmd/juno/cli/root.go
```diff
@@ -204,15 +204,6 @@ func initConfig() {
 		"Verbosity Level", verbosityLevel,
 		"Json Output", enableJsonOutput,
 	).Info("Logger values")
-
```
stdevMac, line 207: Why delete this log from the app? I think is something important for the users to know about the config options they are using

stdevMac (changes requested):

cmd/juno/cli/root.go
D-DePablos: @smishy05 if I remember right, this was to prevent the old config from being shown after update, could you confirm?

smishy05: Yes, @D-DePablos is correct. @stdevMac this log was giving us the `config` before it was updated. I am not sure if it makes sense to keep this. Please check the `create_dynamic_config.go` file ([link](https://github.com/NethermindEth/juno/blob/b9f0686c1090f74729a97c2c3e01905955269fac/cmd/juno/cli/create_dynamic_config.go#L29-L36))

cmd/juno/cli/createDynamicConfig.go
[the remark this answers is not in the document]
stdevMac: yes please

cmd/juno/cli/root.go
stdevMac: At some point, before node execution, the config that we are using, we need to print it, because the user needs to know the config they are using, please, keep the log, or move it to a place where prints properly the config

D-DePablos: OK - Now it should only show config values on the `juno` command to run the node

cmd/juno/cli/createDynamicConfig.go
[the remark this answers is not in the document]
D-DePablos: Solved

stdevMac (approved):

cmd/juno/cli/create_dynamic_config.go
```diff
@@ -0,0 +1,188 @@
...
+	verbosityList := []string{"DEBUG", "INFO", "WARN", "ERROR", "DPANIC", "PANIC", "FATAL"}
+	selectedVerb := runtime.Logger.VerbosityLevel
+	// Find index of selected verbosity level inside the verbosityList
+	selectedVerbIndex := -1
+	for i, verb := range verbosityList {
+		if verb == selectedVerb {
+			selectedVerbIndex = i
+			break
+		}
+	}
```
adriantpaez, line 78: I don't see the current log that I have set, I think you need to do all these checks with string lowercase, and that will fix the problem

D-DePablos: Solved with strings.EqualFold - thanks!

joshklop (approved):
LGTM. The interface is very nice! 

The app panics for me when I have the `rest.enable` set to `false`, but I don't think that's caused by PR.

Issue #285, opened 2022-07-15
Make staging environment for docs


D-DePablos:
RIght now, you can push changes to docs and break them in main.

joshklop:
It looks like [something like this](https://github.com/NethermindEth/juno/blob/a5adfc3b7e08ce1cd63702d73fe623925289046c/.github/workflows/docs-staging-tests.yml#L1) might have existed in the past?

joshklop:
[act](https://github.com/nektos/act#known-issues) looks like a good solution (thanks to @D-DePablos for pointing this out).

PR #287, opened 2022-07-16
Ci/gh pages

joshklop:
## Changes:
- Makes the deployment branch explicit in the docusaurus config and removes the corresponding variable from the gh-pages configuration.
- Configures [trailingSlash](https://docusaurus.io/docs/deployment#docusaurusconfigjs-settings) (probably not directly causing issues at this time, but it is recommended by docusarus for GitHub pages). 

## Types of changes

- [X] Documentation Update
- [X] Build related changes

## Testing
**Requires testing**

- [X] Yes
- [ ] No

- [X] Yes ...? I ran `yarn start` on machine and the website seems to build fine.
- [ ] No

What is the best way to test gh-actions (and specifically the documentation site deployment)? I think issue #285 is relevant here.

### DIFF
--- a/.github/workflows/docs-prod-deploy.yml
+++ b/.github/workflows/docs-prod-deploy.yml
@@ -44,7 +44,6 @@ jobs:
           USE_SSH: true
           GIT_USER: git
           CURRENT_BRANCH: ${{ secrets.CURRENT_BRANCH }}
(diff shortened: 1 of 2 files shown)


tshakalekholoane (approved):
LGTM.

PR #288, opened 2022-07-16
Delete processes and services

IronGauntlets:
## Changes:
- Delete processes and services to make it easier to run and maintain juno. Services were implemented as singletons which lead to side effects when they are used. The process struct provided a concurrency wrapper around the HTTP server and service calls which lead to concurrent bugs, such as panic on a negative wait group call. 

## Types of changes

- [ ] Documentation Update
- [x] Code style update (formatting, renaming)
- [x] Refactoring (no functional changes, no api changes)
- [ ] Build related changes

## Testing
**Requires testing**

- [ ] Yes
- [x] No


cmd/juno/cli/root.go
```diff
@@ -8,25 +8,41 @@ import (
...
+
+	"github.com/NethermindEth/juno/internal/db/state"
+
+	"github.com/NethermindEth/juno/internal/db/abi"
+
+	"github.com/NethermindEth/juno/internal/db/contracthash"
+	"github.com/torquem-ch/mdbx-go/mdbx"
+
+	"github.com/NethermindEth/juno/pkg/rest"
+	"github.com/NethermindEth/juno/pkg/starknet"
```
tshakalekholoane, line 27: Why is there space between imports?

pkg/starknet/state.go
```diff
@@ -10,11 +10,18 @@ import (
 	"strconv"
 	"time"
 
+	"github.com/NethermindEth/juno/internal/db/block"
+
+	"github.com/NethermindEth/juno/internal/db/transaction"
+
+	"github.com/NethermindEth/juno/internal/db/state"
+
```
tshakalekholoane, line 18: See comment above.

> Why is there space between imports?

tshakalekholoane (changes requested):
See comments. Also, coverage issues can be fixed by adding a `// notest` comment in the required block.

cmd/juno/cli/root.go
IronGauntlets: Yeah, its to do with the file watcher I used for goland. Goland support for gofumpt adds these spaces when it rearranges the imports.

IronGauntlets: Please view the latest commit and please resolve the conversation

pkg/starknet/state.go
IronGauntlets: same as above

tshakalekholoane (dismissed):
Brilliant work!

tshakalekholoane:
Sorry about that 😵 the `// notest` has to go right under the closure that Codecov is flagging i.e. just right above where you see banner in the details section.

tshakalekholoane (approved):

adriantpaez (approved):

Issue #230, opened 2022-06-19
Bind ENV vars and Flags to config


stdevMac:
We are using [Cobra](https://cobra.dev/) and [Viper](https://github.com/spf13/viper) for CLI and Config handles.

The idea is to allow the users to provide the configuration of the node using also the CLI commands and ENV vars

stdevMac:
Priority:
1. CLI params
2. ENV vars
3. config file (default)

PR #295, opened 2022-07-22
refactor config

joshklop:
Fixes: #230 #300 

## Changes
- Process configurations from various sources (in order of precedence): CLI parameters, environment variables, config file entries, and default values.
- Use consistent naming conventions regardless of the configuration source
- Never modify the user's environment except for creating database files in locations explicitly provided by the user or in clearly defined default locations.

## Types of changes

- [X] New feature (more CLI flags and corresponding environment variables)
- [X] Breaking change (previous configurations will break)
- [X] Refactoring

## Testing
**Requires testing**

- [ ] Yes
- [X] No (currently we do not test anything in `cmd/juno`)

- [ ] Yes
- [X] No

**Note**: now that `main.go` is getting more modular, I think we can add tests for most of it in the near future.


adriantpaez (approved):
LGTM

internal/config/config.go
```diff
@@ -4,175 +4,100 @@ package config
...
+// empty string and error otherwise.
+// It is analagous to os.UserConfigDir.
+func UserDataDir() (string, error) {
 	// notest
-	d, err := os.UserConfigDir()
-	if err != nil {
-		Logger.Fatal("Could not get config directory.")
-	}
-	Dir = filepath.Join(d, "juno")
+	junoDir := "juno"
```
tshakalekholoane, line 68: This doesn't look to ever be mutated afterwards to let's initialise it as a `const` for clarity.

tshakalekholoane (changes requested):

D-DePablos:
If we expect the user to use the --starknet-enable flag, it is not routed with make all `starknet enable`. 

Since our [docs](https://gojuno.xyz/docs/intro) indicate the user to `make run`, or `make all`, this would not run correctly as the docs stand. 

image

tshakalekholoane (approved):
LGTM.

stdevMac:
Can we add more details around Environment vars and params on the docs? @joshklop Thanks in advance

stdevMac (approved):

PR #296, opened 2022-07-22
Remove algolia (for now)

joshklop:
## Changes:
- Attempts to revert algolia docsearch changes in attempt to fix deployment on main. Not a long-term solution.

## Types of changes

- [X] Bugfix (non-breaking change which fixes an issue)
- [X] Build related changes

## Testing
**Requires testing**

- [X] Yes
- [ ] No

- [ ] Yes
- [X] No (somewhat irrelevant. We need to deploy to see if the deployment will work. While we have discussed tools like act that could solve this, no one on the team has the bandwidth to implement a long-term solution at the moment.)

### DIFF
--- a/.github/workflows/docs-prod-deploy.yml
+++ b/.github/workflows/docs-prod-deploy.yml
@@ -44,6 +44,7 @@ jobs:
           USE_SSH: true
           GIT_USER: git
           CURRENT_BRANCH: ${{ secrets.CURRENT_BRANCH }}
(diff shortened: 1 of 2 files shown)


D-DePablos (approved):

D-DePablos:
@joshklop Thinking about this, it's not just Algolia search, but also the Docusaurus version. Could you undo that as well?

adriantpaez (approved):

PR #298, opened 2022-07-22
Local Github ACTions 

D-DePablos:
## Changes:
- Added ACT environment variables
- Investigating Docs deployment error

## Types of changes

- [ ] Documentation Update
- [ ] Build related changes

Testing: Local runs are done with this .actrc from an M1 macbook:

-P ubuntu-latest=ghcr.io/catthehacker/ubuntu:act-latest
-P ubuntu-18.04=ghcr.io/catthehacker/ubuntu:act-18.04
-b
--container-architecture linux/amd64


D-DePablos:
TODO: I think this branch is bugged from changes to one or multiple actions.

D-DePablos:
This is giving me some problems due to how ACT creates an ACT.env parameter that is not available on GitHub. Actions seem to update when updating the PR, so they were not working for a while. They should be changed with care.

stdevMac (approved):

Makefile
```diff
@@ -44,11 +43,9 @@ codecov-test:
 	courtney/courtney -v -o coverage/coverage.out ./...
 	@cd internal/db && $(MAKE) rm-notest
 
-tidy: ## add missing and remove unused modules
-	 go mod tidy
-
 format: ## run go formatter
 	gofumpt -l -w .
+	go mod tidy
```
joshklop, line 48: I think implicitly running `go mod tidy` on the `format` command is unnecessary, and obfuscates what the actual command is doing. I personally prefer having `tidy` and `format` separated.

joshklop (changes requested):
Are there more changes with the GitHub actions? I only am seeing changes to the Makefile, go.mod, and go.sum.

D-DePablos:
This is quite a bit of a larger headache than I had thought. Additionally, if there is a correct docs staging environment, and smoke tests for Juno, it is not as useful.

The main pain points are related to how in ACT you must use a Docker container, while in GH actions you have a full system. ACT proves to be slower, less reliable, and to force some if conditionals on the workflow files.

PR #320, opened 2022-08-04
Upgrade docusaurus

joshklop:
## Changes:
- Upgrade docusaurus
- Clarify documentation for installing the docs dependencies (primarily `nvm` and `yarn`)

## Types of changes

What types of changes does your code introduce?

- [X] Documentation Update

## Testing
**Requires testing**

- [ ] Yes
- [X] No

PR #321, opened 2022-08-04
Revert "Upgrade docusaurus"

stdevMac:
Reverts NethermindEth/juno#320


joshklop:
We decided to fix the problem with a series of other PRs, concluding with #329.

PR #323, opened 2022-08-04
Fix docs deployment workflow

joshklop:
## Changes:
- Install yarn symlink to the current directory and call it. This way github actions can't insert its own version of yarn.

## Types of changes

- [X] Build related changes

**Note**: this is probably the fifth or sixth time we have tried to fix something related to github actions and node dependencies. If you could test it out on your end so we can avoid another PR that would be awesome (works on my end):

```shell
cd docs
rm -r node_modules
nvm use # nvm install if prompted
corepack enable yarn
yarn install
yarn start
```

If it builds and runs successfully, we should be good to go.


stdevMac (approved):

adriantpaez (commented):
I think we need to add the `yarn` and `yarnpkg` executables to the `.gitignore` file

joshklop:
Thanks for pointing that out @adriantpaez. 

I just [read about](https://github.com/yarnpkg/yarn/issues/7741) the [`yarnPath` param](https://yarnpkg.com/configuration/yarnrc#yarnPath) in the `.yarnrc.yml` config file and the `.yarn/releases` directory. With these, yarn will automatically switch to use the correct version without you needing to change anything.

This way you don't need to mess with the version on your machine. It is stored in the repository.

adriantpaez (approved):

IronGauntlets (approved):

PR #330, opened 2022-08-06
Fix missing constructor_calldata field

### LINKED ISSUE #312: Deploy transactions is not saved with 'constructor_calldata'
**Describe the bug**

The RPC endpoint `starknet_getTransactionByHash` doesn't return the `constructor_calldata` property for Deploy Transactions. I think the reason is that we aren't saving that during the sync process because I don't see this field on our feeder gateway types.

**To Reproduce**
1. Start the node synchronization on mainnet and with the RPC enabled
2. Wait after the first block was synchronized
2. Do this request
```
{
	"jsonrpc": "2.0",
	"method": "starknet_getTransactionByHash",
	"params": [
		"0x00e0a2e45a80bb827967e096bcf58874f6c01c191e0a0530624cba66a508ae75"
	],
	"id": "25"
}
```
The response is:
```
{
	"jsonrpc": "2.0",
	"result": {
		"txn_hash": "0x00e0a2e45a80bb827967e096bcf58874f6c01c191e0a0530624cba66a508ae75",
		"class_hash": "0x0000000000000000000000000000000000000000000000000000000000000000",
		"contract_address": "0x020cfa74ee3564b4cd5435cdace0f9c4d43b939620e4a0bb5076105df0a626c6",
		"constructor_calldata": []
	},
	"id": "25"
}
```

**Expected behavior**
I expect the `constructor_calldata` was not empty

adriantpaez:
Resolves #312

The sync process doesn't set the `constructor_calldata` field to deploy transactions

## Changes:
- Add the `ConstructorCalldata` field to our feeder gateway types
- Init the `ConstructorCalldata` field on the `feederTransactionToDBTransaction` function

## Types of changes

## Testing
**Requires testing**

- No


pkg/feeder/types.go
```diff
@@ -59,18 +59,19 @@ type InvokeFunction struct {
...
-	Type               string   `json:"type"`
+	ContractAddress     string   `json:"contract_address"`
+	ClassHash           string   `json:"class_hash"`
+	Nonce               string   `json:"nonce"`
+	SenderAddress       string   `json:"sender_address"`
+	Version             string   `json:"version"`
+	EntryPointSelector  string   `json:"entry_point_selector"`
+	EntryPointType      string   `json:"entry_point_type"`
+	Calldata            []string `json:"calldata"`
+	ConstructorCalldata []string `json:"constructor_calldata"`
```
joshklop, line 70: What is the difference between `Calldata` and `ConstructorCalldata`?

joshklop (commented):
LGTM. Nice find.

pkg/feeder/types.go
adriantpaez: Is the constructor arguments in DEPLOY transactions

joshklop (approved):

tshakalekholoane (approved):

stdevMac (approved):

Issue #119, opened 2022-04-04
Add colour coded logs


tshakalekholoane:
![Geth screenshot.](https://user-images.githubusercontent.com/23554017/161486550-23a76b45-4efb-4611-b0b1-c93b71d9f5c6.png)

Using [ANSI escape sequences](https://en.m.wikipedia.org/wiki/ANSI_escape_code) or a library such as [this one](https://github.com/fatih/color).

Allow disabling of this feature when:

- `NO_COLOR` environment variable is set.
- `TERM` environment variable has the value `dumb`.
- The user passes the option `--no-color`.
- `JUNO_NO_COLOR` environment variable is set for users that want to disable it specifically for Juno.
- `stdout` or `stderr` is not an interactive terminal ([a TTY](https://pubs.opengroup.org/onlinepubs/009604599/functions/isatty.html)). See the following [Stackoverflow answer](https://stackoverflow.com/questions/68889637/is-it-possible-to-detect-if-a-writer-is-tty-or-not) for guidance.

PR #335, opened 2022-08-09
add colour coded logs

jelilat:
Fixes https://github.com/NethermindEth/juno/issues/119

## Changes:
- Added colour coded logs

## Types of changes

- [ ] Documentation Update
- [ ] Build related changes


cmd/juno/main.go
```diff
@@ -99,6 +99,7 @@ func newRootCmd() *cobra.Command {
 	// Log
 	rootCmd.PersistentFlags().StringVar(&cfg.Log.Level, "log-level", "info", "verbosity of the logs. Options: debug, info, warn, error, dpanic, panic, fatal.")
 	rootCmd.PersistentFlags().BoolVar(&cfg.Log.Json, "log-json", false, "print logs in json format. Useful for automated processing. Typically omitted if logs are only viewed from console.")
+	rootCmd.PersistentFlags().BoolVar(&cfg.Log.Color, "no-color", false, "disable colour coded logs. Colour coded logs are enabled by default.")
```
joshklop, line 102: To maintain the naming convention for CLI flags, let's change to the name from `no-color` to `log-color` and set the default value to `true` (and modify the usage statement to reflect that). This way the UI is consistent for the CLI params, environment variables, and config file.

internal/log/log.go
```diff
@@ -24,9 +25,14 @@ func init() {
 }
 
 // ReplaceGlobalLogger replace the logger and inject it globally
-func ReplaceGlobalLogger(enableJsonOutput bool, verbosityLevel string) error {
+func ReplaceGlobalLogger(enableJsonOutput bool, verbosityLevel string, disableColorEncoder bool) error {
```
joshklop, line 28: I think it would be a bit clearer to say `enableColorEncoder` rather than `disableColorEncoder`. This prevents the "double negative" (`!disableColorEncoder`) in the conditional on line 32.

joshklop (changes requested):
Great work! It is very exciting to finally have this feature. I requested a few changes, but otherwise looks excellent.

cmd/juno/main.go
jelilat: Thanks! Got it.

internal/log/log.go
jelilat: I understand not wanting to have the "double negative". But my concern is that changing it to `enableColorEncoder` will change the meaning.

`!enableColorEncoder` will be made to mean `DO NOT enableColorEncoder` when what it's actually doing is `DO enableColorEncoder`.

joshklop: Correct, if `disableColorEncoder` is renamed to `enableColorEncoder`, then the conditional needs to be changed from `if !disableColorEncoder { ...` to `if enableColorEncoder { ...`. Does that make sense? Please let me know if I'm missing something.

jelilat: We only want to `enableColorEncoder` if `cfg.NoColor == false`.  Since the function input is `cfg.NoColor`, then we can't have `if enableColorEncoder { ...`.

It still needs to be negated to `if !enableColorEncoder { ...`

Can we just opt for a neutral name? E.g. `colorEncoder`. I don't think this will be subject to "double negative" or misinterpretation.

tshakalekholoane: I am also a bit confused. Why can't the input be passed in as `!cfg.NoColor`?

jelilat: Thought of that. But won't it also be "double negatives"?

tshakalekholoane: I suppose. But since all the other flags are written in the affirmative, it seems to me that we can make an exception here to maintain consistency?

I mean, most UNIX commands i.e. `ls`, require the user to explictly state that they want coloured output and if we went that route, we wouldn't have this issue but I am more in favour of having in on by default and so if it means we have the double negative where we pass in the variable, that seems fine to me.

jelilat: True that. I agree.

Any objections @joshklop?

tshakalekholoane: Just to be clear, I am in favour of passing in the variable as `!cfg.NoColor` [here](https://github.com/NethermindEth/juno/blob/c33c4175a033dca4bd55284628e738593be96229/cmd/juno/main.go#L233). So that the signature can read as

```go
func ReplaceGlobalLogger(enableJsonOutput bool, verbosityLevel string, enableColorEncoder bool) error
```

and then we can avoid the double negative inside the function.

joshklop: Thank you for clarifying @tshakalekholoane. That is what I was getting at in my original review, but didn't communicate it very clearly.

jelilat: Okay. Fixed it here https://github.com/NethermindEth/juno/pull/335/commits/045d368edf52218c65e681867ffbf1b737367f02

Thanks, guys.

joshklop (approved):

tshakalekholoane (approved):

joshklop:
I just remembered: we do have [a page](https://gojuno.xyz/docs/running/config) in the docs that displays all available configuration parameters. Since another param is added here (`log-nocolor`), it should probably be added to the documentation. This should be a one-line change to [this](https://github.com/NethermindEth/juno/blob/main/docs/docs/running/config.mdx) file.

stdevMac (approved):

internal/log/log.go
```diff
@@ -24,9 +25,14 @@ func init() {
...
 
 // ReplaceGlobalLogger replace the logger and inject it globally
-func ReplaceGlobalLogger(enableJsonOutput bool, verbosityLevel string) error {
+func ReplaceGlobalLogger(enableJsonOutput bool, verbosityLevel string, enableColorEncoder bool) error {
 	config := zap.NewProductionConfig()
 
+	// Colour coding
+	if enableColorEncoder {
+		config.EncoderConfig.EncodeLevel = zapcore.CapitalColorLevelEncoder
+	}
```
adriantpaez, line 34: I think we can add here an else statement setting the config.EncoderConfig.EncodeLevel to zapcore.CapitalLevelEncoder because right now we are using a lower case by default

adriantpaez (commented):
LGTM. This is a small change I think we can add

internal/log/log.go
jelilat: Fixed here https://github.com/NethermindEth/juno/pull/335/commits/148bfa04421758373996fe001dbe16941512c4b4. Thank you!

adriantpaez (approved):

Issue #301, opened 2022-07-26
Add unit tests for the main package


joshklop:
Right now, we don't test anything in the main package (`cmd/juno`). With [recent](https://github.com/NethermindEth/juno/pull/288) [PRs](https://github.com/NethermindEth/juno/pull/295), the main package has become much more modular and easier to test.

This is also be a good opportunity to consider moving configuration-specific logic such as [`loadConfig`](https://github.com/NethermindEth/juno/blob/2cff0703cbc73b9248be62dd80faae1303c4154b/cmd/juno/main.go#L146) to the `config` package.

IronGauntlets:
#359

PR #337, opened 2022-08-09
Remove --sync-trustless config param

joshklop:
## Description

Juno panics if only `--sync-enable` is provided. We can fix this by removing the need for the `--sync-trustless` param.

## Changes

- Remove the `--sync-trustless` config param. Instead, try to connect to an Ethereum node. If the connection fails for any reason, default to syncing with the feeder gateway.
- Update documentation

## Types of changes

- Documentation Update

## Testing

**Requires testing?** No (no unit tests written for the main package at this time, see #301). This was manually tested.

## Documentation


adriantpaez (approved):

internal/sync/sync.go
```diff
@@ -71,11 +71,13 @@ func NewSynchronizer(cfg *config.Sync, feederClient *feeder.Client, syncManager
...
-		ethereumClient, err := ethclient.Dial(cfg.EthNode)
-		if err != nil {
-			synchro.logger.Fatal("Unable to connect to Ethereum Client", err)
-		}
+	ethereumClient, err := ethclient.Dial(cfg.EthNode)
+	trusted := err != nil
+	if trusted {
+		// If we can't connect to an ethereum node, log an error
+		// and sync from the (trusted) feeder gateway.
+		synchro.logger.With("error", err).Error("Cannot connect to Ethereum Client: syncing state from the centralized feeder gateway")
```
tshakalekholoane, line 79: I don't think this should be an error. If the user did not provide an Ethereum node address, we should probably assume that was their intention and log that at `INFO` i.e. `defaulting to syncing from gateway`.

tshakalekholoane (changes requested):
See requested change.

internal/sync/sync.go
joshklop: Thanks! That would be much better. I've updated the code to incorporate your suggestions.

internal/sync/sync.go
```diff
@@ -71,12 +71,17 @@ func NewSynchronizer(cfg *config.Sync, feederClient *feeder.Client, syncManager
 	synchro := new(Synchronizer)
 	synchro.logger = Logger.Named("Sync Service")
 	synchro.feeder = feederClient
-	if !cfg.Trusted {
+	trusted := cfg.EthNode == ""
```
tshakalekholoane, line 74: Let's add spaces here for clarity because `trusted` is also used on line 96.

```suggestion

trusted := cfg.EthNode == ""

```

internal/sync/sync.go
```diff
@@ -71,12 +71,17 @@ func NewSynchronizer(cfg *config.Sync, feederClient *feeder.Client, syncManager
 	synchro := new(Synchronizer)
 	synchro.logger = Logger.Named("Sync Service")
 	synchro.feeder = feederClient
```
tshakalekholoane, line 73: Let's use a composite literal here instead. See [section in Effective Go](https://go.dev/doc/effective_go#composite_literals).

```suggestion
	synchro := &Synchronizer{
		logger: Logger.Named("Sync Service"),
		feeder: feederClient,
	}
```

internal/sync/sync.go
```diff
@@ -71,12 +71,17 @@ func NewSynchronizer(cfg *config.Sync, feederClient *feeder.Client, syncManager
...
+	} else {
 		ethereumClient, err := ethclient.Dial(cfg.EthNode)
-		if err != nil {
-			synchro.logger.Fatal("Unable to connect to Ethereum Client", err)
+		if err == nil {
+			synchro.l1Client = ethereumClient
+		} else {
+			synchro.logger.With("error", err).Error("Cannot connect to Ethereum Client: defaulting to syncing from gateway")
+			trusted = true
 		}
```
tshakalekholoane, line 84: I think we then want it to be fatal here because it means the user provided an Ethereum node address but that did not work.

tshakalekholoane (changes requested):

tshakalekholoane (approved):

Issue #336, opened 2022-08-09
Call shutdown if error is caught


joshklop:
**Problem**
If an error is caught in [`checkErrChs`](https://github.com/NethermindEth/juno/blob/7735bff86b9e243266dbe6070a329a949409bda0/cmd/juno/main.go#L397), we immediately exit with a log message.

**Solution**
Call [`shutdown`](https://github.com/NethermindEth/juno/blob/7735bff86b9e243266dbe6070a329a949409bda0/cmd/juno/main.go#L362), attempting to exit cleanly without a stack trace.

stdevMac:
closing because of https://github.com/NethermindEth/juno/pull/359

PR #338, opened 2022-08-10
shutdown on error

jelilat:
Resolves https://github.com/NethermindEth/juno/issues/336

## Description

If an error is caught in [checkErrChs](https://github.com/NethermindEth/juno/blob/7735bff86b9e243266dbe6070a329a949409bda0/cmd/juno/main.go#L397), we immediately exit with a log message. This PR is meant to allow it to exit cleanly without a trace

## Types of changes

### DIFF
--- a/cmd/juno/main.go
+++ b/cmd/juno/main.go
@@ -221,7 +221,7 @@ func juno(cfg *config.Juno) {
 	}
 
 	// Wait until error
-	checkErrChs(errChs)
+	checkErrChs(errChs, cfg)
 }
 
 func setupVirtualMachine() {
@@ -390,11 +390,11 @@ func shutdown(cfg *config.Juno) {
 	}
 }
 
-func checkErrChs(errChs []chan error) {
+func checkErrChs(errChs []chan error, cfg *config.Juno) {
 	for _, errCh := range errChs {
 		for {
 			if err := <-errCh; err != nil {
-				Logger.Fatal(err)
+				shutdown(cfg)
 			}
 		}
 	}


cmd/juno/main.go
```diff
@@ -390,11 +390,11 @@ func shutdown(cfg *config.Juno) {
 	}
 }
 
-func checkErrChs(errChs []chan error) {
+func checkErrChs(errChs []chan error, cfg *config.Juno) {
 	for _, errCh := range errChs {
 		for {
 			if err := <-errCh; err != nil {
-				Logger.Fatal(err)
+				shutdown(cfg)
```
IronGauntlets, line 397: I am not sure of the implications here:

The reason we were logging Fatal here was that something drastically went wrong. That means we cannot continue running the program and we need the trace for that. If we don't then call shutdown we can have unexpected behaviour.

IronGauntlets: Then what about other situations in the code where we call `Fatal()`?

In most cases, deeper parts of the code will not have access to `shutdown()`.

So if we were to mandate that then we need to have call shutdown() when Fatal() is called how are we going to ensure that? And if we can't then we will have unpredictable behaviour which can lead to errors which are difficult to debug.

joshklop: I personally think that the error channels should be used for non-fatal errors only. If a portion of the application (RPC server, synchronizer, etc.) incurs a catastrophic error, then that component should call `Logger.Fatal`. Otherwise, the error can be passed up the stack through the error channel, and the application can shut down gracefully in `main.go`. 

Right now we assume all errors received on the error channels are fatal, which seems incorrect to me.

Aside: I feel we are having these small design issues because of the global variables in `main.go`. I think it would be preferable to only have local variables in the `juno` function and then use `defer` for their `Close` methods. This way, Juno can clean up after itself even when it panics. This also avoids problems with repeating `if service.Enable { ... }` in multiple places. I haven't tried to implement this, but using `defer` instead of `shutdown` seems like idiomatic go to me.

IronGauntlets: It is very difficult to say whether an error is fatal or not. Especially in the case when you are calling packages which call other packages. 

In general, you would determine whether an error is fatal or not in the context the function is being called in and the lower the error occurs in the call stack the harder it is to determine.

The reason for using the error channel is that we have several calls that are blocking. We could use defer statements to call the close functions however, then the architecture of how go routines are stopped is completely different and then you have to take into consideration how the main function would return and work around that. It will be even more difficult to incorporate interrupt signals cleanly. Also, defer statements that don't allow us to handle errors. So if there was an error while closing it would simply be ignored.

joshklop: Thanks for the detailed response @IronGauntlets. Fair points. Let's go with your suggestion.

**Re: `defer`**

I'm not sure I'm following. My point is that if we `defer` all of the cleanup functions in `func juno(cfg)`, then we shouldn't need a separate `shutdown` function. Right now I think we are unnecessarily splitting the set-up and shutdown logic. We could still use the error channels same as we are now.

IronGauntlets: I misunderstood how you wanted to use the defer function. 

Yes, it is possible as you are suggesting but again incorporating it with interrupt signals would be difficult.

IronGauntlets:
closing in favour of #359

PR #356, opened 2022-08-15
Refactor

IronGauntlets:
## Description

This PR refactors the following:
- Removes rest server
- Removes juno cli

## Types of changes

## Testing

**Requires testing**: No

## Documentation


tshakalekholoane:
Why are we moving public packages up a level?

IronGauntlets:
There is no need for having a `pkg` package as explained [here](https://dave.cheney.net/tag/pkg).

tshakalekholoane:
> the boilerplate of an empty pkg/ directory distracts from the more useful idiom of an internal/ directory.

As far as I am aware Git does not track empty directories. Am I missing something?

> a directory that exists only to hold other packages is a potential design smell.

I don't agree and I think the [current structure](https://github.com/golang-standards/project-layout) serves us better because we have a lot of non-Go components like the documentation website and having a pkg directory makes it easier for developers to make the distinction. In addition to that, I do not see any material benefit from adopting this change.

IronGauntlets:
It is a matter of personal preference, if you feel strongly about keeping the current structure I am ok with that and I will drop the commit.

tshakalekholoane:
I am in favour of keeping the current structure.

I am also wondering whether it would make sense to put `internal/` under `cmd/juno/` given we now have a single binary and those internal packages aren't shared anymore. What do you think about that?

IronGauntlets:
Well based on the project layout you shared it should be kept separate. So I think let's stick with that.

tshakalekholoane (approved):
LGTM.

stdevMac (approved):

joshklop (changes requested):
I think we need to delete the [`docs/docs/features/cli.mdx`](https://gojuno.xyz/docs/features/cli) file as well.

IronGauntlets:
Done. Also deleted `docs/docs/features/rest.mdx` and updated the example in `config.mdx`.

joshklop (approved):

PR #360, opened 2022-08-18
Fix RPC bugs

### LINKED ISSUE #350: Some RPC methods panic when called without arguments
The following RPC `starknet_` methods panic when called without supplying the appropriate arguments. This includes methods that do not take any arguments.

- `getBlockWithTxHashes`
- `getBlockWithTxs`
- `getTransactionByBlockAndIndex`
- `getClass`
- `getClassHashAt`
- `call`
- `estimateFee`
- `blockNumber`
- `blockHashAndNumber`
- `chainId`
- `pendingTransactions`
- `syncing`

### LINKED ISSUE #352: Some methods return invalid JSON RPC 2.0 errors
The following error message is shown to the user sometimes which is is not valid according to the JSON RPC 2.0 specification.

```json
{
  "jsonrpc": "2.0",
  "error": {},
  "id": 1
}
```

- `getBlockWithTxHashes` and `getBlockWithTxs`.
  -  When `-1` is passed in as a parameter.
  - When a valid block is queried but one that is not yet synced.
- `getStorageAt`.
  - When no arguments are passed in.
- `getTransactionByBlockAndIndex`.
  - When the following parameters are passed in `[-1, 0]`.
  - When the field elements are not padded with enough zeroes according to the following regular expression `^0x0[a-fA-F0-9]{1,63}$`.  
- `getBlockTransactionCount`.
  - For the following arguments `[-1]` and a valid but block that does not exist e.g. `[99999]`.

adriantpaez:
resolves #350, resolves #352 

## Description

Fix bugs on the RPC related to errors and params management.

## Changes:

- Fix `starknet_getStorageAt` params order
- Add custom error types: `NewUnexpectedError`, `NewNotImplementedError`,  `NewInvalidStorageKey`
- Improve the error management on `getBlockById` function
- Check params on `starknet_call`
- Check `BlockId` is non-negative if is a number

## Types of changes

## Testing

**Requires testing**: Yes

**Did you write tests??**: No (Now we are testing these bugs by hand)

## Documentation


stdevMac (approved):

internal/rpc/starknet/errors.go
```diff
@@ -9,6 +9,30 @@ func (s *StarkNetError) Error() string {
 	return s.Message
 }
 
+func NewUnexpectedError() *StarkNetError {
```
joshklop, line 12: Should the errors be variables instead of functions? I'm not familiar with Go's idiom in this case.

joshklop (commented):
Looks good to me overall. I didn't try to re-run @tshakalekholoane  tests locally, though. 

Is it possible to automate those tests here or in a future PR? Something like [what was done](https://github.com/NethermindEth/juno/blob/cb1f2c5d5867179b26d6c458a52bece6900b5192/pkg/rpc/rpc_test.go#L24) in the previous implementation?

internal/rpc/starknet/errors.go
adriantpaez: Yep, definitely ✅ https://github.com/NethermindEth/juno/pull/360/commits/e2c4e4d26b68cdd451351f8fd5137477de3d32ae

joshklop (approved):

internal/rpc/starknet/starknet.go
```diff
@@ -65,161 +69,162 @@ func (s *StarkNetRpc) GetStateUpdate(blockId *BlockId) (any, error) {
...
-		// TODO: manage unexpected error
+		s.logger.Errorw(err.Error(), "function", "GetTransactionByHash")
+		return nil, UnexpectedError
 	}
 	return NewTxn(tx)
 }
 
-func (s *StarkNetRpc) GetTransactionByBlockIdAndIndex(blockId *BlockId, index uint64) (any, error) {
-	b, err := getBlockById(blockId, s.blockManager)
+func (s *StarkNetRpc) GetTransactionByBlockIdAndIndex(blockId *BlockId, index *uint64) (any, error) {
```
tshakalekholoane, line 117: Why are we using a pointer to a `uint64`?

tshakalekholoane (changes requested):
The following methods still panic when called without arguments so issue #350 should be unliked from this pull request and perhaps updated for methods now work.

- `blockNumber`
- `blockHashAndNumber`
- `chainId`
- `pendingTransactions`
- `syncing`

internal/rpc/starknet/starknet.go
adriantpaez: If the user doesn't pass the `index` param, then the 0 value is set by default (which is valid). But `index` is a required param, so if the user doesn't pass it we need to return the `InvalidTxnIndex` error. Basically is to manage the edge case for 0 value.

internal/rpc/starknet/errors.go
```diff
@@ -5,62 +5,53 @@ type StarkNetError struct {
...
+		Message: "Unexpected error",
+	}
+	NotImplementedError = StarkNetError{
+		Code:    -2,
+		Message: "Not implemented",
+	}
+	InvalidStorageKey = StarkNetError{
+		Code:    -3,
+		Message: "Invalid storage key",
+	}
```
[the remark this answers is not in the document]
adriantpaez: Done -> https://github.com/NethermindEth/juno/pull/360/commits/c8c650477f308449b324605ea510ccedae344a01

internal/rpc/starknet/errors.go
```diff
@@ -5,62 +5,51 @@ type StarkNetError struct {
...
 
-func NewContractNotFound() *StarkNetError {
-	return &StarkNetError{
+var (
+	// NOTE: This error isn't part of the StarWere specification
+	UnexpectedError = StarkNetError{
+		Code:    -1,
+		Message: "Unexpected error",
+	}
+	// NOTE: This error isn't part of the StarWere specification
```
tshakalekholoane, line 18: Typo.

```suggestion
	// NOTE: This error isn't part of the StarkNet RPC specification.
```

internal/rpc/starknet/starknet.go
tshakalekholoane: I see 😵.

internal/rpc/starknet/errors.go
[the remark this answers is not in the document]
tshakalekholoane: What about the `-1 Unexpected error`? I think that should also fall under `-32603 Internal error`.

internal/rpc/starknet/starknet.go
```diff
@@ -65,161 +69,138 @@ func (s *StarkNetRpc) GetStateUpdate(blockId *BlockId) (any, error) {
...
-	if !isFelt(address) {
-		return nil, NewContractNotFound()
-	}
-	addressF := new(felt.Felt).SetHex(address)
-	b, err := getBlockById(blockId, s.blockManager)
+func (s *StarkNetRpc) GetClassHashAt(blockId *BlockId, address *RpcFelt) (any, error) {
+	b, err := getBlockById(blockId, s.blockManager, s.logger)
 	if err != nil {
-		return nil, NewInvalidBlockId()
+		return nil, err
```
tshakalekholoane, line 152: Shouldn't this return an `InvalidBlockId` error?

tshakalekholoane (changes requested):

internal/rpc/starknet/starknet.go
adriantpaez: The `getBlockById` function returns the proper error

joshklop (approved):

internal/rpc/starknet/errors.go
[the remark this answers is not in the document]
adriantpaez: `Internal error` is returned when a panic happens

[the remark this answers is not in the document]
tshakalekholoane: Yes, but not exclusively. 

That's just a byproduct of Go serving each request within its own Go routine which allows the application to detect when a panic has occurred and thus send an internal error to the user. If it were any other language a panic would occur and nothing would be sent back right?

So one would assume an internal error is meant for exactly this purpose. In fact here's what the HTTP specification for a `500 Internal Server Error` says (because the JSON-RPC 2.0 specification doesn't go into much detail with regards to that):

> A generic error message, given when an **unexpected** condition was encountered and no more specific message is suitable.

_(Emphasis mine)._

Wouldn't the `-32603 Internal error` be the equivalent here?

tshakalekholoane (approved):

PR #362, opened 2022-08-18
[sync] solve error around starting block Number

### LINKED ISSUE #361: Sync returning 0 on starknet_blockNumber before first iteration
**Describe the bug**
When the server starts, the `starknet_blockNumber` RPC returns 0 while the first block isn’t synched, including the case when I’m restarting the sync from a block > 0. We need to update this value before starting the sync.

**Expected behavior**
Return the real value that we have already synced.

stdevMac:
Resolves #361 

## Description

Solve the problem that exists around the initial block fetched.

## Changes:

- Set as default that the minimum block synced and the current block synced, both are the minimum between the block and the state processed.
- Fix Syncing response on Sync

## Types of changes

## Testing

**Requires testing**: No

**Did you write tests??**: No

### DIFF
--- a/internal/sync/sync.go
+++ b/internal/sync/sync.go
@@ -33,14 +33,11 @@ type Synchronizer struct {
 	logger *zap.SugaredLogger
 
 	// startingBlockNumber is the block number of the first block that we will sync.
-	startingBlockNumber int64
-	// startingBlockHash is the hash of the first block that we will sync.
-	startingBlockHash string
+	startingBlockNumber      int64
+	startingBlockNumberSaved int64
 
 	// latestBlockNumberSynced is the last block that was synced.
 	latestBlockNumberSynced int64
-	// latestBlockHashSynced is the last block that was synced.
-	latestBlockHashSynced *felt.Felt
 
 	// stateDIffCollector
 	stateDiffCollector StateDiffCollector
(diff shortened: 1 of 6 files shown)


internal/sync/utils.go
```diff
@@ -64,6 +64,7 @@ func loadAbiOfContract(abiVal string) (abi.ABI, error) {
 func fetchContractCode(stateDiff *types.StateDiff, client *feeder.Client, logger *zap.SugaredLogger) *CollectorDiff {
 	collectedDiff := &CollectorDiff{
 		stateDiff: stateDiff,
+		Code:      map[string]*types.Contract{},
```
adriantpaez, line 67: Is it possible to use `make` here and set the map size properly with `len` of `stateDiff.DeployedContracts`?

adriantpaez (changes requested):

IronGauntlets (approved):

internal/sync/utils.go
stdevMac: Changed to make, good catch!

internal/sync/sync.go
```diff
@@ -220,8 +231,12 @@ func (s *Synchronizer) GetStateDiff(blockNumber int64) *types.StateDiff {
 	return s.syncManager.GetStateDiff(blockNumber)
 }
 
-func (s *Synchronizer) GetStateDiffFromHash(blockHash string) *types.StateDiff {
-	return s.syncManager.GetStateDiffFromHash(blockHash)
+func (s *Synchronizer) GetStateDiffFromFelt(blockHash *felt.Felt) *types.StateDiff {
```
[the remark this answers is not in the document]
stdevMac: Back to `FromHash`

joshklop (approved):

PR #365, opened 2022-08-22
[RPC] getStateUpdate

### LINKED ISSUE #353: `starknet_getStateUpdate` does not work
`starknet_getStateUpdate` panics when called. It also returns an empty response when called without arguments.

### LINKED ISSUE #364: `starknet_getStateUpdate` returns decimal representations of field elements
The JSON RPC method `starknet_getStateUpdate` returns decimal formatted strings in its output. For example the following request:

```json
{
  "jsonrpc": "2.0",
  "method": "starknet_getStateUpdate",
  "params": [
    0
  ],
  "id": 1
}
```

Returns the following response:

```jsonc
{
  "jsonrpc": "2.0",
  "result": {
    "storage_diffs": {
      "1407352284783671026127653444689025405903785508685418552906528356004194896722": [
        {
          "key": "394289168167265445374632363466357133981865040396795598116421516872827652841",
          "Value": "570920402159267707065786439757248226918653058110349152410418921644075880512"
        },
        {
          "key": "2698533540911089821226508289542548328632057663796456292913373387133967093855",
          "Value": 1991
        }
      ],
// ...
```

Note that some values are also formatted as JSON numbers instead of strings.

**Heads up**: This method will also panic when executed without enabling the synchronisation process. See #353 for details.

adriantpaez:
Resolves #364
Resolves #353 

## Description

The `starknet_getStateUpdate` RPC endpoint returns `Felt` values as decimal representation when we need it to return them as hexadecimal. Also, I introduce some improvements to the StateUpdate flow (get-save-serve).

## Changes:

- Add `sync.proto` file to manage all database marshal and unmarshal processes for `StateUpdate`
- Rename `StateDiff` to `StateUpdate`
- Use `Felt` as a key in `StorageDiff` map instead of `string`
- Parse `BlockHash` in `stateUpdateResponseToStateDiff` function
- Add the `Value()` method to `Felt` to get the value referenced by a pointer to `Felt`

## Types of changes

## Testing

**Requires testing**: Yes

**Did you write tests??**: No. I test the output of `starkent_getStateUpdate` by hand and all the errors are fixed. In the future, we need to add unit tests to this.

## Documentation

**If this requires a documentation update, did you add one?** No. The documentation is already on `main`


stdevMac (approved):

internal/rpc/starknet/types.go
```diff
@@ -490,3 +490,65 @@ func (r *RpcFelt) UnmarshalJSON(data []byte) error {
...
+	DeployedContracts []*DeployedContractItem `json:"deployed_contracts"`
+	DeclaredContracts []*DeclatedContractItem `json:"declared_contracts"`
+}
+
+type StateUpdate struct {
+	BlockHash string     `json:"block_hash"`
+	NewRoot   string     `json:"new_root"`
+	OldRoot   string     `json:"old_root"`
+	StateDiff *StateDiff `json:"state_diff"`
+}
```
joshklop, line 520: Is it possible to move these types to the `types` package and delete the types that are currently there? Having them both appears redundant, but these seem more well-written.

joshklop (commented):
LGTM

internal/db/sync/manager_test.go
```diff
@@ -1,9 +1,10 @@
 package sync_test
 
 import (
-	"reflect"
 	"testing"
 
+	"gotest.tools/assert"
```
IronGauntlets, line 6: Please use `github.com/stretchr/testify/assert`

internal/rpc/starknet/types.go
adriantpaez: I agree and it's something that I noticed too, but that means making some changes to the sync process and isn't related to the RPC. I think will be better if we can do that on other PR later

internal/db/sync/manager_test.go
adriantpaez: Done -> https://github.com/NethermindEth/juno/pull/365/commits/dc0cddb3b7e629ac3fc1feb2dccb30c63d5aca9d

joshklop: I have some reservations about this. Testify does not allow us to instrument comparisons in `DeepEqual`, which means asserts that try to compare `nil` felts will panic (gotest uses `go-cmp`, which fixes this problem). This has been a [long-standing issue](https://github.com/stretchr/testify/issues/535) in testify that has yet to be fixed. Other projects [have signaled](https://github.com/cosmos/cosmos-sdk/issues/12332) their desire to move away from testify to gotest.

I am personally in favor of exclusively using gotest where necessary, which is what we are already doing in a [few](https://github.com/NethermindEth/juno/blob/674508537411b0ad4a4dc100e7af0705c80963bb/internal/db/transaction/manager_test.go#L15) [places](https://github.com/NethermindEth/juno/blob/674508537411b0ad4a4dc100e7af0705c80963bb/internal/db/block/block_test.go#L58-L68).

stdevMac: I am with @joshklop on this, :+1:  on `gotest`

adriantpaez: I've never needed to use custom compare functions, but I agree with @joshklop on this because we're using it on other tests on the code

pkg/felt/helpers.go
```diff
@@ -34,6 +34,10 @@ func (z *Felt) Hex0x() string {
 	return fmt.Sprintf("0x0%063s", z.Hex())
 }
 
+func (z *Felt) Value() Felt {
```
tshakalekholoane, line 37: I find something like `Deref` a bit more descriptive.

tshakalekholoane (approved):

pkg/felt/helpers.go
adriantpaez: Solved here https://github.com/NethermindEth/juno/pull/365/commits/57701aee589b932031301d08cfcff28c9a1490a6

joshklop (approved):

PR #377, opened 2022-08-29
Test RPC input types

adriantpaez:
## Description

Add tests for UnmarshalJSON of `BlockId`, `StorageKey`, and `RpcFelt`.

## Types of changes

- Other (please describe):  Add tests


stdevMac (approved):

internal/rpc/starknet/types_test.go
```diff
@@ -0,0 +1,276 @@
...
+			data: []byte(`"0x01d328a71faf48c5c3857e99f20a77b18522480956d1cd5bff1ff2df3c8b427b"`),
+			want: "0x01d328a71faf48c5c3857e99f20a77b18522480956d1cd5bff1ff2df3c8b427b",
+			err:  false,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			var got StorageKey
+			if err := got.UnmarshalJSON(tt.data); (err != nil) != tt.err {
+				t.Errorf("StorageKey.UnmarshalJSON(), unexpected error: %v", err)
```
tshakalekholoane, line 170: In keeping with the how the rest of the other failed tests are reported.

```suggestion
				t.Errorf("StorageKey.UnmarshalJSON() = %v, want nil", err)
```

internal/rpc/starknet/types_test.go
```diff
@@ -0,0 +1,276 @@
...
+			data: []byte(`{}`),
+			want: BlockId{},
+			err:  ErrInvalidBlockId,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			var got BlockId
+			if err := got.UnmarshalJSON(tt.data); !errors.Is(err, tt.err) {
+				t.Errorf("BlockId.UnmarshalJSON() error = %v, wantErr %v", err, tt.err)
```
tshakalekholoane, line 114: For the sake of consistency with how all the other failed tests are reported. 

```suggestion
				t.Errorf("BlockId.UnmarshalJSON() = %v, want %v", err, tt.err)
```

internal/rpc/starknet/types_test.go
```diff
@@ -0,0 +1,276 @@
...
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got := tt.storageKey.Felt()
+			if !got.Equal(tt.want) {
+				t.Fatalf("StorageKey.Felt() = %v, want %v", got, tt.want)
+			}
+		})
+	}
+}
```
tshakalekholoane, line 207: It's not clear what is being tested here. Serialisation of a storage key to a `felt.Felt`? The test names could also be more descriptive.

tshakalekholoane (changes requested):

internal/rpc/starknet/types_test.go
adriantpaez: "want nil" isn't valid for all test cases, sometimes it is valid to return an error. Perhaps the best approach is:
```go
t.Errorf("StorageKey.UnmarshalJSON() = %v, unexpected error response", err)
```

internal/rpc/starknet/types_test.go
adriantpaez: Yes, tests the serialization of a storage key to a `felt.Felt`. I'll change the name to something more descriptive.

adriantpaez: Done here https://github.com/NethermindEth/juno/commit/3dfc627866e7bbace0bb0b2c06f049ecc011e30f

internal/rpc/starknet/types_test.go
tshakalekholoane: Ah. I see. We should also document that the key on line 155 is not valid because it's larger than `p`. That's the reason right?

tshakalekholoane (changes requested):

internal/rpc/starknet/types.go
```diff
@@ -52,10 +52,7 @@ func (id *BlockId) UnmarshalJSON(data []byte) error {
 	switch t := token.(type) {
 	case json.Number:
 		value, err := t.Int64()
-		if err != nil {
-			return err
-		}
-		if value < 0 {
+		if err != nil || value < 0 {
```
IronGauntlets, line 55: The actual error is getting suppressed here. We need to look into details regarding what type of error could be returned. Also if the value cannot be less than 0 why is `uint64` not being used?

internal/rpc/starknet/types_test.go
```diff
@@ -0,0 +1,276 @@
...
+import (
+	"errors"
+	"io"
+	"reflect"
+	"testing"
+
+	"github.com/NethermindEth/juno/pkg/felt"
+	gocmp "github.com/google/go-cmp/cmp"
+
+	"gotest.tools/assert"
```
IronGauntlets, line 12: Please use `"github.com/stretchr/testify/assert"`. It is more widely used.

internal/rpc/starknet/types_test.go
```diff
@@ -0,0 +1,276 @@
...
+					return x.value.(string) == y.value.(string)
+				case blockIdNumber:
+					return x.value.(int64) == y.value.(int64)
+				default:
+					return false
+				}
+			}))
+		})
+	}
+}
```
IronGauntlets, line 138: Good test cases!

Can you please explain why `assert.DeepEqaul` is not sufficient?

internal/rpc/starknet/types_test.go
```diff
@@ -0,0 +1,276 @@
...
+		{
+			name:       "test 2",
+			storageKey: "0x00898cca7dbf84c2213f3a00e84775013bc991bc104d3a00952ce4bc166a4a1a",
+			want:       new(felt.Felt).SetHex("0x00898cca7dbf84c2213f3a00e84775013bc991bc104d3a00952ce4bc166a4a1a"),
+		},
+		{
+			name:       "test 2",
+			storageKey: "0x0000000000000000000000000000000000000000000000000000000000000005",
+			want:       new(felt.Felt).SetHex("0x0000000000000000000000000000000000000000000000000000000000000005"),
+		},
```
IronGauntlets, line 197: Please use better names to describe the difference between the test cases.

internal/rpc/starknet/types_test.go
```diff
@@ -0,0 +1,276 @@
...
+		{
+			name:    "test 2",
+			rpcFelt: "0x0320e37cf7c972458a3edf08ab51f2ab7596857706af174a3d5be4e46f16c63e",
+			want:    new(felt.Felt).SetHex("0x0320e37cf7c972458a3edf08ab51f2ab7596857706af174a3d5be4e46f16c63e"),
+		},
+		{
+			name:    "test 2",
+			rpcFelt: "0x048636a22c6c74f5631b00c66ac3c8ab4714aa308ddd214af4089ccdfcee0f81",
+			want:    new(felt.Felt).SetHex("0x048636a22c6c74f5631b00c66ac3c8ab4714aa308ddd214af4089ccdfcee0f81"),
+		},
```
IronGauntlets, line 266: Same as above. If you just want to enumerate multiple felts, then change your test struct to the following 

```go
tests []struct{
  rpcFelt RpcFelt
  want *felt.Felt
}
```

Then iterate over the tests and assert on each case, similar to [this](https://github.com/NethermindEth/juno/blob/62f5d5d26001e1b60e8f2957f21e30282d22ce4e/internal/utils/datadir_test.go#L78).

internal/rpc/starknet/types_test.go
joshklop: `assert.DeepEqual` does not know how to compare `BlockId`s. Regardless, we could add an `Equal` method to `BlockId`, which is what `assert.DeepEqual` looks for.
```go
func (id BlockId) Equal(other BlockId) bool {
	xv := reflect.ValueOf(id)
	yv := reflect.ValueOf(other)
	if xv.IsZero() || yv.IsZero() {
		return xv.IsZero() && yv.IsZero()
	}
	if id.idType != other.idType {
		return false
	}
	switch id.idType {
	case blockIdHash:
		return id.value.(*felt.Felt).Equal(other.value.(*felt.Felt))
	case blockIdTag:
		return id.value.(string) == other.value.(string)
	case blockIdNumber:
		return id.value.(int64) == other.value.(int64)
	default:
		return false
	}
}
```
With this change, `assert.DeepEqual` alone should be sufficient. This is nice because we don't need to copy the custom comparison function every time we call `DeepEqual` on structs with `BlockId`s.

internal/rpc/starknet/types_test.go
adriantpaez: The reason is that doesn't match with the regex, but yes, isn't matching because is larger than `p`. I will add that doc to the test

internal/rpc/starknet/types.go
adriantpaez: `uint64` isn't used because the [`json.Number`](https://pkg.go.dev/encoding/json#Number) only supports `Int64`, `String`, or `Float`. But yes, I could convert the `int64` to `uint64` when stored in `id.value`

internal/rpc/starknet/types_test.go
adriantpaez: Because it broke on unexported fields
image

internal/rpc/starknet/types_test.go
adriantpaez: I'll just remove the name of the test, it looks like it's not necessary

internal/rpc/starknet/types_test.go
adriantpaez: Done -> https://github.com/NethermindEth/juno/pull/377/commits/89b6a2a09b2d9609ac60b2904cf961e3fdd5ebd1

internal/rpc/starknet/types.go
adriantpaez: Done -> https://github.com/NethermindEth/juno/pull/377/commits/fcfaf802764d63ce2cdc6ee870c9ef2ed8c6d8ec

internal/rpc/starknet/types_test.go
adriantpaez: With `github.com/stretchr/testify/assert` we can't define custom comparisons and with `gotest.tools/assert` yes

internal/rpc/starknet/types_test.go
adriantpaez: Done -> https://github.com/NethermindEth/juno/pull/377/commits/58b626cb47457c9890f4765a4a8288ea82bea185

internal/rpc/starknet/types_test.go
adriantpaez: Done -> https://github.com/NethermindEth/juno/pull/377/commits/58b626cb47457c9890f4765a4a8288ea82bea185

internal/rpc/starknet/types_test.go
adriantpaez: Done -> https://github.com/NethermindEth/juno/pull/377/commits/e7a67c1953fc6118db7cfbbf14b6bbb3ccb9fcc9

tshakalekholoane (approved):

internal/rpc/starknet/types_test.go
IronGauntlets: Ok sounds good. In future, then we should move away from using `testify/assert` and `testify/require` packages.

IronGauntlets (changes requested):

internal/rpc/starknet/types.go
```diff
@@ -78,6 +76,27 @@ func (id *BlockId) UnmarshalJSON(data []byte) error {
 	return nil
 }
 
+func (id BlockId) Equal(other BlockId) bool {
+	xv := reflect.ValueOf(id)
+	yv := reflect.ValueOf(other)
+	if xv.IsZero() || yv.IsZero() {
+		return xv.IsZero() && yv.IsZero()
+	}
```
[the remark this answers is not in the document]
adriantpaez: Please review the new commits

internal/rpc/starknet/types.go
```diff
@@ -78,6 +76,27 @@ func (id *BlockId) UnmarshalJSON(data []byte) error {
 	return nil
 }
 
+func (id BlockId) Equal(other BlockId) bool {
```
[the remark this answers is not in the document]
adriantpaez: Check the new commits, the `Equal` method is removed

internal/rpc/starknet/types.go
```diff
@@ -78,6 +76,27 @@ func (id *BlockId) UnmarshalJSON(data []byte) error {
...
+	case blockIdHash:
+		return id.value.(*felt.Felt).Equal(other.value.(*felt.Felt))
+	case blockIdTag:
+		return id.value.(string) == other.value.(string)
+	case blockIdNumber:
+		return id.value.(uint64) == other.value.(uint64)
+	default:
+		return false
+	}
+}
```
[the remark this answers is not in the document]
adriantpaez: Check the new commits

joshklop (approved):

PR #384, opened 2022-08-30
Shutdown multiple sync go-routines gracefully

IronGauntlets:
## Description

In the sync package, multiple go routines are started, however, they were not closed properly. Hence, when `synchroniser.Close()` was called those go-routines continued to run, which led to panic in the database as they would try to store/retrieve information.

A quit channel and select statement are used to ensure the go-routine exit when `synchroniser.Close()` is called.

## Types of changes

## Testing

**Requires testing**: Yes

**Did you write tests??**: No (manual tested)

### DIFF
--- a/internal/sync/sync.go
+++ b/internal/sync/sync.go
@@ -7,6 +7,7 @@ import (
 	"math"
 	"math/big"
 	"strconv"
+	sync2 "sync"
 	"time"
 
 	blockDB "github.com/NethermindEth/juno/internal/db/block"
@@ -57,6 +58,9 @@ type Synchronizer struct {
 
 	// Running is true when the service is active and running
 	Running bool
+
+	wg   *sync2.WaitGroup
+	quit chan struct{}
 }
 
 // NewSynchronizer creates a new Synchronizer.
@@ -68,6 +72,8 @@ func NewSynchronizer(n utils.Network, ethNode string, feederClient *feeder.Clien
 	synchro := &Synchronizer{
 		logger: Logger.Named("Sync Service"),
 		feeder: feederClient,
+		quit:   make(chan struct{}),
+		wg:     new(sync2.WaitGroup),
 	}
 
 	trusted := ethNode == ""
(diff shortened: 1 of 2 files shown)


joshklop (approved):

internal/sync/sync.go
```diff
@@ -114,10 +117,16 @@ func (s *Synchronizer) Run() {
...
+			return
+		default:
+			err := s.sync()
+			s.logger.With("Error", err).Info("Sync Failed, restarting iterator in 10 seconds")
+			time.Sleep(10 * time.Second)
+			s.stateDiffCollector.Close()
+
+		}
 	}
 }
```
adriantpaez, line 131: I think this loop always gets stuck on the first iteration because line 124 (`s.sync()`) starts the sync process which always waits for a new state update [here](https://github.com/NethermindEth/juno/blob/f2b29bd0fc9212093f95e5e4122aeb4c127fe158/internal/sync/sync.go#L140) unless an error occurs. So, that means for me the race condition is still happening inside the `sync` function

IronGauntlets: [This line](https://github.com/NethermindEth/juno/blob/f2b29bd0fc9212093f95e5e4122aeb4c127fe158/internal/sync/sync.go#L140) will also exit when the channel returned  `s.stateDiffCollector.GetChannel()` is closed. The for loop over a channel block forever unless it is closed. Once a channel is closed the for loop exits. Which then in turn causes `sync()` to exit.

The channel returned from `s.stateDiffCollector.GetChannel()` is closed when `s.stateDiffCollector.Close()` is called in `synchroniser.Close()` . Therefore, when the `sync()` returns without an error it will go over the loop again and then it will see that `s.quit` is closed and hence `handleSync()` would exit.

IronGauntlets: Please look at the latest commit

internal/sync/sync.go
```diff
@@ -372,27 +390,34 @@ func (s *Synchronizer) updateTransactionReceipts(receipt feeder.TransactionExecu
...
+				time.Sleep(time.Minute)
+			}
+			err := s.updateBlocks(currentBlock)
+			if err != nil {
+				time.Sleep(time.Minute)
+				continue
+			}
+			s.logger.With("Block Number", currentBlock).Info("Updated block info")
+			s.syncManager.StoreLatestBlockSaved(currentBlock)
+			currentBlock++
```
tshakalekholoane, line 419: Delete extraneous space.

```suggestion
```

internal/sync/sync.go
```diff
@@ -7,6 +7,7 @@ import (
 	"math"
 	"math/big"
 	"strconv"
+	sync2 "sync"
```
tshakalekholoane, line 10: We could probably do better and allow this package from the standard library to take precedence and then rename the other `dbsync` but this pattern seems to be used throughout so I have no strong opinions about it.

tshakalekholoane (approved):
LGTM.

PR #386, opened 2022-08-30
Add health check endpoint

### LINKED ISSUE #229: Health Check for Node

### LINKED ISSUE #324: Implement RPC health_check endpoint
**Is your feature request related to a problem? Please describe.**
We should implement an health_check endpoint like this one on the [Nethermind client](https://docs.nethermind.io/nethermind/ethereum-client/monitoring-node-health#enabling-and-configuring-health-checks)

**Describe the solution you'd like**
Create a new RPC endpoint, that can be fixed, query the sync status, and any other info from the node. 
For more details, reach @stdevMac.

stdevMac:
Fixes | Closes | Resolves #324 

## Description

Is needed to check that the node is running without errors

## Changes:

- Add `health` package to check that node is healthy
- Add documentation for it

## Types of changes

## Testing

**Requires testing**: Yes

**Did you write tests??**: No, only manual testing

## Documentation


docs/docs/features/rpc.mdx
```diff
@@ -42,6 +42,112 @@ Or with named params:
...
+
+```json
+{
+    "jsonrpc": "2.0",
+    "result": {
+        "status": "Healthy",
+        "message": "Node running",
+        "syncingStatus": {
+            "starting_block_hash": "0x0336912d969aaffcd93f6e236e76ebcffbb1a73a7ae9f5c230ef010b29d3c014",
+            "starting_block_number": "b",
```
IronGauntlets, line 78: A number should be user instead of a letter

docs/docs/features/rpc.mdx
```diff
@@ -42,6 +42,112 @@ Or with named params:
...
+            "current_block_hash": "0x0336912d969aaffcd93f6e236e76ebcffbb1a73a7ae9f5c230ef010b29d3c014",
+            "current_block_number": "b",
```
IronGauntlets, line 80: Same as above

internal/juno/juno.go
```diff
@@ -207,6 +209,8 @@ func starkNetJsonRPC(stateManager *state.Manager, blockManager *block.Manager,
 ) (*jsonrpc.JsonRpc, error) {
 	starkNetApi := starknet.New(stateManager, blockManager, transactionManager, synchronizer,
 		virtualMachine)
+	healthApi := health.New(synchronizer, virtualMachine)
+
```
IronGauntlets, line 213: Why are `synchronizer` and `virtualMachine` the only ones determining the node's health?

internal/rpc/health/errors.go
```diff
@@ -0,0 +1,25 @@
...
+	}
+	VMNotRunning = Error{
+		Code:    20,
+		Message: "Cairo VM is not running",
+	}
+	UnHealthy = Error{
+		Code:    30,
+		Message: "Node is not healthy",
+	}
+)
```
IronGauntlets, line 25: Const should be used with iota. Each error should be prefixed with `Err...` which is idiomatic golang

internal/rpc/health/health.go
```diff
@@ -0,0 +1,45 @@
...
+		}, nil
+	}
+	if !r.synchronizer.Running {
+		return nil, NodeNotSyncing
+	}
+	if !r.vm.Running() {
+		return nil, VMNotRunning
+	}
+	return nil, UnHealthy
+}
```
IronGauntlets, line 45: Since this is a new package I feel like it is relatively self-contained. Therefore, writing tests should be easy. The logger should have been passed as a parameter to start moving away from the log singleton. This only creates further technical debt.

internal/rpc/health/types.go
```diff
@@ -0,0 +1,9 @@
+package health
+
+import "github.com/NethermindEth/juno/pkg/types"
+
+type Status struct {
+	Status        string            `json:"status"`
+	Message       string            `json:"message"`
+	SyncingStatus *types.SyncStatus `json:"syncingStatus"`
+}
```
IronGauntlets, line 9: This struct alone doesn't warrant a separate file. Also, the file name should be a little more specific about what is contained within. `types.go` is too generic a name.

IronGauntlets (changes requested):

docs/docs/features/rpc.mdx
```diff
@@ -42,6 +42,112 @@ Or with named params:
 
 The RPC server is configured using the `rpc` section on the configuration file, see the [configuration section](/running/config.mdx).
 
+## Health Check
+
+Health checks are available at the endpoint `health_nodeStatus` and retrieves the status of the node.
```
joshklop, line 47:
```suggestion
```
To me, this seems redundant with the next paragraph.

joshklop (changes requested):

docs/docs/features/rpc.mdx
stdevMac: is a hex value, direct response from the api

docs/docs/features/rpc.mdx
stdevMac: is a hex value, direct response from the api

tshakalekholoane (changes requested):
I think there is some inconsistency with having an elaborate status page and still returning errors when one of the services is not running. 

According to my understanding, health checks are primarily designed for headless clients (otherwise the user could just have a peek at the logs right?) so it might be better to return the status page and nothing otherwise because the errors still return `200 OK` status messages which doesn't seem right.

internal/juno/juno.go
stdevMac: We have a couple of things to test:
- Blocks are stored properly
- The state is updating without problem
- We can call starknet_call and all the Python-related stuff
- Database is working properly

If we are syncing properly, that means that the blocks are stored properly and that the database is working. The only pending is the python-related stuff, which now is only the starknet_call, but soon we will add at also estimateFee.

internal/rpc/health/errors.go
stdevMac: I did the fix with the Err, but not the iota, we are handling the errors in that way on the starknet rpc handler and I prefer to keep it in that way for this PR, and maybe later change both places

stdevMac:
I changed the struct to something more simple on the health request:
```json
{
    "jsonrpc": "2.0",
    "result": {
        "status": "Healthy",
        "message": "Node running",
        "syncing_status": {
            "current_block_number": 32,
            "highest_block_number": 318586
        }
    },
    "id": 1
}
```
How it looks for you @tshakalekholoane @IronGauntlets

tshakalekholoane (changes requested):
I think there's still room to simplify this further. We can return something like:

```json
{
    "jsonrpc": "2.0",
    "result": {
        "status": "available"
    },
    "id": 1
}
```

and nothing else (not even error messages).

The block information is already contained in the logs and so I am of the opinion that we should not include it here. But if you insist on keeping it, then we should probably update the struct and its constructors as well that still have block hash fields after the last change.

Another thing with this more complex structure is that [these checks](https://github.com/NethermindEth/juno/blob/0b7bc3a7b8eb102bd00952fc21fcaeaae8334b6a/internal/rpc/health/health.go#L53-L59) are not exhaustive. That is, because we exit early, the condition where both the node is not syncing and the virtual machine is running is not explored. Should we define another error for that?

It might add some unnecessary complexity, which is why I am in favour of the above.

stdevMac:
Closed in favor of #390

stdevMac:
[](url)

PR #388, opened 2022-09-01
Implement exponential backoff

### LINKED ISSUE #380: Implement an exponential backoff algorithm for feeder gateway requests
Implement an exponential [backoff algorithm](https://cloud.google.com/iot/docs/how-tos/exponential-backoff#example_algorithm) for [feeder gateway requests](https://github.com/NethermindEth/juno/blob/b467e2a58636ea7cf35679615fc59540a09d3477/pkg/feeder/feeder.go#L54-L72) to avoid slamming the server.

ayushm2003:
Closes #380

## Description

Implements better waiting mechanism for failing requests.

## Changes:

- Changes [retryFuncForDoReq](https://github.com/NethermindEth/juno/blob/ayush/exp_backoff/pkg/feeder/feeder.go#L58) logic
## Types of changes

## Testing

**Requires testing**: Yes

**Did you write tests??**: Tested manually

## Documentation


pkg/feeder/feeder.go
```diff
@@ -52,15 +54,18 @@ func NewClient(baseURL, baseAPI string, client *HttpClient) *Client {
 	}
 
 	// retry mechanism for do requests
+	// https://cloud.google.com/iot/docs/how-tos/exponential-backoff
 	retryFuncForDoReq := func(req *http.Request, httpClient HttpClient) (*http.Response, error) {
 		var res *http.Response
-		wait := 2 * time.Second
-		for {
+		wait := 1*time.Second + time.Duration(rand.Intn(1000))*time.Millisecond
+		maxWait := 32 * time.Second
```
tshakalekholoane, line 61: Define this as a constant outside the loop and change it to 10 minutes.

pkg/feeder/feeder.go
```diff
@@ -52,15 +54,18 @@ func NewClient(baseURL, baseAPI string, client *HttpClient) *Client {
...
+		for i := 1; i <= 10; i++ {
 			res, err = httpClient.Do(req)
 			if err != nil || res == nil || res.StatusCode != http.StatusOK {
-				wait *= 2
 				Logger.With("Waiting:", wait.Seconds(), "Error", err).Info("Waiting to do again a request")
 				time.Sleep(wait)
+				x := time.Duration(math.Pow(2, float64(i)))*time.Second + time.Duration(rand.Intn(1000))*time.Millisecond
+				wait = minDuration(x, maxWait)
```
tshakalekholoane, line 68: You are slightly complicating this.

- The line `wait *= 2` that you just deleted describes `2^n` and it doesn't require maintaining an iteration variable or calling `math.Pow`.
- You do not need to define your own minimum function. Use `math.Min` with a cast. The values we are dealing with are guaranteed to stay within the appropriate range.

tshakalekholoane (changes requested):

pkg/feeder/feeder.go
ayushm2003: Changing `math.Pow` to `wait *= 2` does slightly change the logic, as we are now also doubling the random part every time, but this shouldn't be a problem for our use case.

tshakalekholoane: Remove the random element. It adds unnecessary complexity.

pkg/feeder/feeder.go
```diff
@@ -52,15 +54,18 @@ func NewClient(baseURL, baseAPI string, client *HttpClient) *Client {
 	}
 
 	// retry mechanism for do requests
+	// https://cloud.google.com/iot/docs/how-tos/exponential-backoff
 	retryFuncForDoReq := func(req *http.Request, httpClient HttpClient) (*http.Response, error) {
 		var res *http.Response
-		wait := 2 * time.Second
+		wait := 1*time.Second + time.Duration(rand.Intn(1000))*time.Millisecond
+		maxWait := float64(10 * time.Minute)
```
tshakalekholoane, line 61: Do the cast is the function that requires it, not here.

ayushm2003: Not sure I understand. Do you mean, I can let the `wait` variable be float and cast it to Duration when calling `time.Sleep()`?

tshakalekholoane: Try something like this:

```go

	retryFuncForDoReq := func(req *http.Request, httpClient HttpClient) (*http.Response, error) {
		var res *http.Response

		const maxWait = 10 * time.Minute
		wait := 1 * time.Second

		for {
			res, err = httpClient.Do(req)
			if err != nil || res == nil || res.StatusCode != http.StatusOK {
				Logger.With("wait", wait.Seconds(), "error", err).Info("Retrying request")
				time.Sleep(wait)
				wait = time.Duration(math.Min((wait * 2).Seconds(), maxWait.Minutes()))
				continue
			}
			if res.StatusCode == http.StatusOK {
				break
			}
		}
		
		return res, err
	}
```

pkg/feeder/feeder.go
```diff
@@ -52,15 +53,17 @@ func NewClient(baseURL, baseAPI string, client *HttpClient) *Client {
...
-		wait := 2 * time.Second
+		wait := 1 * time.Second
+		const maxWait = 10 * time.Minute
 		for {
 			res, err = httpClient.Do(req)
 			if err != nil || res == nil || res.StatusCode != http.StatusOK {
-				wait *= 2
 				Logger.With("Waiting:", wait.Seconds(), "Error", err).Info("Waiting to do again a request")
 				time.Sleep(wait)
+				wait = time.Duration(math.Min((wait * 2).Seconds(), maxWait.Minutes()))
```
adriantpaez, line 66:
```suggestion
				if newWait := wait * 2; newWait < maxWait {
					wait = newWait
				}
```
The `time.Duration` type is an `int64`, which means you can do the arithmetic operations directly. Also, I think an `if` statement is more clear than using `math.Min` and cast `time.Duration` to `float64`. Another detail is that you don't need to use the `Seconds()` or `Minutes()` functions because then you are overwriting the value of the current `wait` with a new `int64` that represents seconds or minutes and is used as nanoseconds by `time.Sleep()`.

pkg/feeder/feeder.go
```diff
@@ -52,15 +53,17 @@ func NewClient(baseURL, baseAPI string, client *HttpClient) *Client {
 	}
 
 	// retry mechanism for do requests
+	// https://cloud.google.com/iot/docs/how-tos/exponential-backoff
 	retryFuncForDoReq := func(req *http.Request, httpClient HttpClient) (*http.Response, error) {
 		var res *http.Response
-		wait := 2 * time.Second
+		wait := 1 * time.Second
```
adriantpaez, line 59:
```suggestion
		wait := time.Second
```

adriantpaez (changes requested):

adriantpaez (approved):

stdevMac (approved):

PR #391, opened 2022-09-04
Terraform script to run juno node

omarelsayed23:
Fixes | Closes | Resolves #

## Description

This is a terraform script that will run after a Github Actions script is triggered. The objective of the TF script is to run Juno node with different configurations.

## Types of changes

Adding three new files:
 • main.tf (contains the main TF script)
 • user-data.yaml (configuration file that has cloud-init instructions that will execute when the aws instance boots up)
 • variables.tf (variables definition)

## Testing

**Requires testing**: Yes

## Documentation

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -1,0 +1,132 @@
+terraform {
+
+
+  required_providers {
+    aws = {
+      source = "hashicorp/aws"
+    }
+  }
+
+ backend "s3" {
+   bucket = "custom-terraform-state"
+   key = "dev/s3/terraform.tfstate"
+   region= "us-west-2"
+   encrypt = true
+ }
+ 
+
+}
+
+provider "aws" {
+  region = var.selected_region
+}
+
+
+resource "aws_key_pair" "key-module" {
+  key_name   = "key-module"
+  public_key = var.pub_key
+}
+/*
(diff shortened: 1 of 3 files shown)


main.tf
```diff
@@ -0,0 +1,132 @@
...
+
+  server_side_encryption_configuration {
+  rule {
+  apply_server_side_encryption_by_default {
+  sse_algorithm = "AES256"
+  } 
+  }
+  }
+}
+*/
```
joshklop, line 48:
```suggestion
```
I'm not familiar with terraform at all, but would it be possible to remove these lines?

userdata.yaml
```diff
@@ -0,0 +1,15 @@
...
+
+packages:
+ - golang-go
+
+runcmd:
+ - export HOME=/home/ubuntu
+ - export GOPATH=$HOME/go
+ - export GOCACHE=$HOME/.cache/go-build
+ - go install github.com/NethermindEth/juno/cmd/juno@latest
+ - go install github.com/NethermindEth/juno/cmd/juno-cli@latest
```
tshakalekholoane, line 14: This binary does not exist anymore.

```suggestion
```

tshakalekholoane (changes requested):

PR #392, opened 2022-09-05
Solve error at block 0 on syncing status

### LINKED ISSUE #383: starknet_syncing returning null if first block not synced
**Describe the bug**
When the node starts to run for the first time, when called the method `starknet_syncing`, it will try to ask for block information that isn't yet on the database, so, it will return nil. 

**To Reproduce**
1. Run the node
2. Before syncing the first block information, call method `starknet_syncing`
3. Receive nil

**Expected behavior**
Return an empty response on those fields

stdevMac:
Resolves #383 

## Description

When the node starts to run for the first time when called the method `starknet_syncing`, it will try to ask for block information that isn't yet on the database, so, it will return nil. 

With this approach, we are returning 0 in case of those blocks are not found yet on the database.

## Changes:

- Catch edge case on `starknet_syncing` at block 0
- Refactor struct of SyncingStatus on RPC

## Types of changes

## Testing

**Requires testing**: Yes

**Did you write tests??**: No

## Documentation


internal/rpc/starknet/types.go
```diff
@@ -552,3 +567,7 @@ func NewStateUpdate(s *types.StateUpdate) *StateUpdate {
 		StateDiff: stateDiff,
 	}
 }
+
+type Status struct {
+	Status string `json:"status"`
+}
```
adriantpaez, line 573: Is this type used?

adriantpaez (changes requested):

internal/sync/sync.go
```diff
@@ -113,15 +112,16 @@ func (s *Synchronizer) setStateDiffCollector(apiSync bool) {
...
 
 func (s *Synchronizer) handleSync() {
 	s.wg.Add(1)
 	for {
+		s.Running = true
 		if err := s.sync(); err != nil {
+			s.Running = false
 			s.logger.With("Error", err).Info("Sync Failed, restarting iterator in 10 seconds")
 			time.Sleep(10 * time.Second)
 			s.stateDiffCollector.Close()
```
adriantpaez, line 127: Can the logic on lines 122, 124, and 127 be moved inside the `s.sync()` method? The reason is that looks like is the responsibility of that method to manage that and not of the caller, and then the `s.handleSync()` method only just cares about the retry

internal/sync/sync.go
```diff
@@ -179,32 +179,45 @@ func (s *Synchronizer) sync() error {
...
+	highestBlockHash := new(felt.Felt)
+	highestBlockNumber := uint64(0)
+
+	startingBlock, err := s.blockManager.GetBlockByNumber(startingBlockNumber)
 	if err != nil {
-		return nil
+		if startingBlockNumber != 0 {
+			return nil
+		}
+		startingBlockHash = new(felt.Felt).SetHex("0x0")
```
adriantpaez, line 196: I think will be more clear if the error is checked using `errors.Is(...)` with the [ErrNotFound](https://github.com/NethermindEth/juno/blob/stdevMac/fix_starknet_syncing/internal/db/mdbx.go#L15)

internal/sync/sync.go
```diff
@@ -179,32 +179,45 @@ func (s *Synchronizer) sync() error {
...
-	startingBlockNumber := uint64(math.Min(float64(s.startingBlockNumber), latestBlockSaved))
-
-	startingBlock, err := s.blockManager.GetBlockByNumber(startingBlockNumber)
+	block, err := s.blockManager.GetBlockByNumber(latestBlockNumber)
 	if err != nil {
-		return nil
+		if latestBlockNumber != 0 {
+			return nil
+		}
+		currentBlockHash = new(felt.Felt).SetHex("0x0")
```
adriantpaez, line 206: I think will be more clear if the error is checked using `errors.Is(...)` with the [ErrNotFound](https://github.com/NethermindEth/juno/blob/stdevMac/fix_starknet_syncing/internal/db/mdbx.go#L15)

internal/sync/sync.go
stdevMac: I am now checking with there Error, and also checking that is on the block 0, that is the only edge case, any other condition, is wrong

internal/sync/sync.go
stdevMac: I am now checking with there Error, and also checking that is on the block 0, that is the only edge case, any other condition, is wrong

internal/sync/sync.go
stdevMac: Done using defer, let me know wdyt

adriantpaez (approved):

internal/sync/sync.go
```diff
@@ -179,32 +183,47 @@ func (s *Synchronizer) sync() error {
 func (s *Synchronizer) Status() *types.SyncStatus {
 	latestBlockSaved := float64(s.syncManager.GetLatestBlockSaved())
 
+	startingBlockHash := new(felt.Felt)
```
tshakalekholoane, line 186: This value is never used.

```suggestion
	var startingBlockHash *felt.Felt
```

internal/sync/sync.go
```diff
@@ -179,32 +183,47 @@ func (s *Synchronizer) sync() error {
...
+	startingBlockNumber := uint64(math.Min(float64(s.startingBlockNumber), latestBlockSaved))
+
+	currentBlockHash := new(felt.Felt)
```
tshakalekholoane, line 189: Similarly here.

```suggestion
	var currentBlockHash *felt.Felt
```

tshakalekholoane (changes requested):

PR #395, opened 2022-09-05
Add getNonce endpoint

### LINKED ISSUE #347: Address nonce
We need to support and maintain the `nonce` value for the addresses. This feature is required by the `starknet_getNonce` RPC endpoint.

`nonce` properties:
- Counts all the transactions made by an address
- Its value is relative to a block (or state root)
- `nonce` never decreases. The `nonce` for address `X` at block `i+1` is always greater than or equal to the `nonce` at block `i`

## Solution proposal

We can store the `nonce` together with the storage root of the address, which means: in the Trie leaf resulting from traversing (following the address) the Trie from the block state root.

ayushm2003:
Fixes | Closes | Resolves #347 

## Description

Adds [`starknet_getNonce`](https://github.com/starkware-libs/starknet-specs/blob/master/api/starknet_api_openrpc.json#L625) endpoint.

## Changes:

- Stores `nonce` in the `ContractState` table
- Provides the `nonce` using the RPC

## Types of changes

## Testing

**Requires testing**: Yes

**Did you write tests??**: No

## Documentation


internal/rpc/starknet/starknet.go
```diff
@@ -174,6 +174,19 @@ func (s *StarkNetRpc) GetBlockTransactionCount(blockId *BlockId) (any, error) {
...
+	b, err := getBlockById(blockId, s.blockManager, s.logger)
+	if err != nil {
+		return nil, err
+	}
+	_state := state.New(s.stateManager, b.NewRoot)
+	nonce, err := _state.GetNonce(address.Felt())
+	if err != nil {
+		return nil, err
+	}
+	return nonce, nil
```
adriantpaez, line 187: Use the `felt.Hex0x()` to return felts in hex format, otherwise, the decimal representation is sent

PR #405, opened 2022-09-14
Improve StarkNet class management

adriantpaez:
## Description

Improve StarkNet class management

## Changes:

- A new database manager is added to be able to store StarkNet classes. With this new manager, all information related to the contract definition is stored in one place and under one type definition: `ContractClass`.
- The ABI manager is removed in favor of the new class manager which already contains the ABI definition of the class.
- The reason of the state is to maintain the correctness of the state and not to store all the program/class definitions. That's why the `GetContract` and `PutContract` methods are removed from the state manager and `GetContract` and `SetContract` from the state. Now the state has the `InitNewContract` method which will be used when a new contract is declared, and its value should be initialized as a new leaf of the trie.
- The CONTRACT_DEF database is removed in favor of the new CLASS database.
- The `starknet_getClass` and `starknet_getClassAt` RPC endpoints and the virtual machine are updated to use the new class manager.
- Before these changes, the feeder package fetches and stores class definitions as JSON format directly, to avoid that and to be able to compress program definitions as required by the RPC specification, the `FullContract` type is added to the feeder package types to parse correctly the information from `get_full_contract` endpoint.

## Types of changes

## Testing

**Requires testing**: Yes

## Documentation


jelilat (approved):

jelilat:
@adriantpaez Could you please edit the description to indicate whether you wrote tests or not?

Indicate one of `Yes/No`

PR #420, opened 2022-11-07
regenesis: reset repository for reimplementation

omerfirmak:
Kept the `cmd` dir and minimal node implementation only.


omerfirmak:
@IronGauntlets , If we are to delete `docs/`, we should disable related CI workflows as well, right?

`docs-prod-deploy.yml` and `docs-test.yml`

IronGauntlets:
@omerfirmak yes that sounds good. I will create another commit.

IronGauntlets (approved):

joshklop (approved):

IronGauntlets (approved):

PR #421, opened 2022-11-07
Use standard test coverage collector

omerfirmak:
## Description

one less thing to depend on

## Changes:

Use `go test` instead of `courtney`

## Types of changes

- Build related changes

## Testing

**Requires testing**: No

**Did you write tests??**: No

### DIFF
--- a/.github/workflows/juno-test.yml
+++ b/.github/workflows/juno-test.yml
@@ -22,14 +22,12 @@ jobs:
       - name: Install deps
         run: make install-deps
       - name: Tests
-        run: make codecov-test
+        run: make test-cover
       - name: Benchmark
         run: make benchmarks
       - name: Upload coverage to Codecov
-        uses: codecov/codecov-action@v2
+        uses: codecov/codecov-action@v3
         with:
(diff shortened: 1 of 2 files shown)


joshklop (approved):
Will we be using go [1.18](https://github.com/NethermindEth/juno/blob/832f2145fb255670b00df982c7f2da35f5acbd16/.github/workflows/juno-test.yml#L15) or 1.19?

omerfirmak:
> Will we be using go [1.18](https://github.com/NethermindEth/juno/blob/832f2145fb255670b00df982c7f2da35f5acbd16/.github/workflows/juno-test.yml#L15) or 1.19?

I would say, we can wait until it is available in a Debian `stable` (or a Ubuntu LTS) release. AFAICS, it is in `testing` already so should be in the next Debian `stable` release.

We can start testing in CI with 1.19 as well, to ease the transition when we really want to make the switch.

omerfirmak:
Added 1.19 as a non-required step and it failed immediately :)

IronGauntlets (approved):

PR #423, opened 2022-11-08
Add Block and Transaction Structs

IronGauntlets:
## Description

Add Block and Transaction types as defined in [StrakNet's Technical Documentation](https://docs.starknet.io/documentation/).

## Changes:

- Add Block Struct
- Add Deploy Transaction struct
- Add Invoke Transaction struct
- Add Declare Transaction struct

## Types of changes

## Testing

**Requires testing**: Yes

**Did you write tests??**: No


omerfirmak:
@IronGauntlets

jelilat:
@IronGauntlets According to the docs [here](https://docs.starknet.io/documentation/develop/Blocks/transactions/), `deploy transaction` and `invoke transaction 0` is depreciated. We should be using `invoke transaction 1` instead.

core/Transaction.go
```diff
@@ -0,0 +1,98 @@
...
+	// https://docs.starknet.io/documentation/develop/Blocks/transactions/#calculating_the_hash_of_a_deploy_transaction
+	return nil
+}
+
+type InvokeTransaction struct {
+	// Version 0 fields
+	// The address of the contract invoked by this transaction.
+	ContractAddress *felt.Felt
+	// The encoding of the selector for the function invoked (the entry point in the contract)
+	EntryPointSelector *felt.Felt
```
jelilat, line 39: Use `invoke transaction 1` instead as explained [here](https://docs.starknet.io/documentation/develop/Blocks/transactions/). It does not include `EntryPointSelector`

core/Transaction.go
```diff
@@ -0,0 +1,98 @@
...
+	ConstructorCalldata []*felt.Felt
+	// Who invoked the deployment. Set to 0 (in future: the deploying account contract).
+	CallerAddress *felt.Felt
+	// The transaction’s version. Possible values are 1 or 0.
+	//
+	// When the fields that comprise a transaction change,
+	// either with the addition of a new field or the removal of an existing field,
+	// then the transaction version increases.
+	// Transaction version 0 is deprecated and will be removed in a future version of StarkNet.
+	version *felt.Felt
```
joshklop, line 25: Maybe the version should be a `uint`?
```suggestion
	version uint
```

core/Transaction.go
```diff
@@ -0,0 +1,98 @@
+package core
+
+import (
+	"github.com/NethermindEth/juno/core/felt"
+)
+
+// Todo: define contract class
+type ContractClass struct{}
+
+type DeployTransaction struct {
```
jelilat, line 10: `deploy` is depreciated. We should use `deploy syscall` that's explained [here](https://docs.starknet.io/documentation/develop/Contracts/contract-classes/) instead.

joshklop (changes requested):

core/Transaction.go
joshklop: Thanks to @omerfirmak for pointing out that version is a felt in the specs.

core/Transaction.go
```diff
@@ -0,0 +1,98 @@
...
+	// The address of the contract invoked by this transaction.
+	ContractAddress *felt.Felt
+	// The encoding of the selector for the function invoked (the entry point in the contract)
+	EntryPointSelector *felt.Felt
+
+	// Version 1 fields
+	// The address of the sender of this transaction.
+	SenderAddress *felt.Felt
+	// The transaction nonce.
+	nonce *felt.Felt
```
joshklop, line 45: Should the nonce be a public field?
```suggestion
	Nonce *felt.Felt
```

joshklop (changes requested):
Tiny tiny detail: should the `Transaction.go` file be capitalized? `block.go` is not.

IronGauntlets:
> should the Transaction.go file be capitalized?

No, it should not. I have updated the file name.

joshklop (approved):

Issue #431, opened 2022-11-10
Define Contract Related Structs


IronGauntlets:
- [ ] [Contract ABI](https://docs.starknet.io/documentation/develop/Contracts/contract-abi/)
- [ ] [Contract address](https://docs.starknet.io/documentation/develop/Contracts/contract-address/)
- [ ] [Contract classes](https://docs.starknet.io/documentation/develop/Contracts/contract-classes/)
- [ ] [Class hash](https://docs.starknet.io/documentation/develop/Contracts/contract-hash/)
- [ ] [Contract storage](https://docs.starknet.io/documentation/develop/Contracts/contract-storage/)

PR #426, opened 2022-11-09
Database Proof of Concept

joshklop:
## Description

Adds core types and interfaces necessary for the database implementation. Still a heavy WIP.

## Changes:

TODO

## Types of changes

- New feature (non-breaking change which adds functionality)
- Documentation Update - TODO

## Testing

**Requires testing**: Yes

**Did you write tests??**: Yes

## Documentation

**If this requires a documentation update, did you add one?** Not yet

PR #432, opened 2022-11-10
Add contract types

joshklop:
## Description

See #431

Adds contract-related types:
- Contract
- Class
- EntryPoint

## Types of changes

## Testing

**Requires testing**: No

## Documentation

**If this requires a documentation update, did you add one?** Added comments where necessary.

### DIFF
--- a/core/contract.go
+++ b/core/contract.go
@@ -17,7 +17,7 @@ type Class struct {
 	Constructors []EntryPoint
 	// An ascii-encoded array of builtin names imported by the class.
 	Builtins []string
-	// The starknet_keccak hash of the JSON compiler output.
+	// The starknet_keccak hash of the ".json" file compiler output.
 	ProgramHash []byte
 	Bytecode    []*felt.Felt
 }
--- a/core/contract.go
+++ b/core/contract.go
@@ -18,7 +18,7 @@ type Class struct {
 	// An ascii-encoded array of builtin names imported by the class.
 	Builtins []string
 	// The starknet_keccak hash of the ".json" file compiler output.
-	ProgramHash []byte
+	ProgramHash *felt.Felt
 	Bytecode    []*felt.Felt
 }
 
--- a/core/contract.go
+++ b/core/contract.go
@@ -4,7 +4,7 @@ import (
 	"github.com/NethermindEth/juno/core/felt"
 )
 
-// Class unambiguously defines a Contract's semantics.
+// Class unambiguously defines a [Contract]'s semantics.
 type Class struct {
 	// The version of the class, currently always 0.
 	APIVersion uint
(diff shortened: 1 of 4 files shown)


core/contract.go
```diff
@@ -0,0 +1,49 @@
+package core
+
+import (
+	"github.com/NethermindEth/juno/core/felt"
+)
+
+// Class unambiguously defines a [Contract]'s semantics.
```
IronGauntlets, line 7: What do the `[]` square brackets mean?

core/contract.go
```diff
@@ -0,0 +1,49 @@
...
+
+// Contract is an instance of a [Class].
+type Contract struct {
+	// The number of transactions sent from this contract.
+	// Only account contracts can have a non-zero nonce.
+	Nonce uint
+	// Hash of the class that this contract instantiates.
+	ClassHash *felt.Felt
+	// Root of the contract's storage trie.
+	StorageRoot *felt.Felt // TODO: is this field necessary?
```
jelilat, line 40: Added a comment about this [here](https://github.com/NethermindEth/juno/pull/426).

But yes, I believe this is necessary for smart contracts. The contract storage contains the smart contract data that should be accessible to users.

jelilat:
```suggestion
	StorageRoot *felt.Felt 
```

core/contract.go
```diff
@@ -0,0 +1,49 @@
...
+	// External functions defined in the class.
+	Externals []EntryPoint
+	// Functions that receive L1 messages. See
+	// https://www.cairo-lang.org/docs/hello_starknet/l1l2.html#receiving-a-message-from-l1
+	L1Handlers []EntryPoint
+	// Constructors for the class. Currently, only one is allowed.
+	Constructors []EntryPoint
+	// An ascii-encoded array of builtin names imported by the class.
+	Builtins []string
+	// The starknet_keccak hash of the JSON compiler output.
```
IronGauntlets, line 20:
```suggestion
	// The starknet_keccak hash of the ".json" file compiler output.
```

core/contract.go
```diff
@@ -0,0 +1,49 @@
...
+	ProgramHash []byte
```
IronGauntlets, line 21: Do you think it would be good to use `*felt.Felt`?

omerfirmak: `sn_keccak` is a modified Keccak256 which outputs a value in the same field Pedersen operates on. So It probably should be a `*felt.Felt`

https://docs.starknet.io/documentation/develop/Hashing/hash-functions/#starknet_keccak

core/contract.go
joshklop: When the documentation is rendered as HTML it will link to the `Contract` struct.

core/contract.go
joshklop: I agree that the storage should be stored somewhere, but I don't yet see a point with storing the storage root in the `Contract` struct. I can't think of a place it would be used.

Right now I am thinking we will only need to store the class hash and nonce with the contract instance. The contract storage will be stored separately (for performance). Both the hash+nonce and storage will be keyed on the contract address. But none of this is set in stone so feel free to suggest enhancements.

core/contract.go
joshklop: Nice catch, thanks!

core/contract.go
```diff
@@ -0,0 +1,49 @@
...
+	// Hash of the class that this contract instantiates.
+	ClassHash *felt.Felt
+	// Root of the contract's storage trie.
+	StorageRoot *felt.Felt // TODO: is this field necessary?
+}
+
+// EntryPoint uniquely identifies a Cairo function to execute.
+type EntryPoint struct {
+	// Keccak hash of the function signature.
+	Selector []byte
```
omerfirmak, line 46: should be a field element, similar to `ProgramHash`

jelilat (approved):

IronGauntlets (approved):

PR #436, opened 2022-11-11
Implement Pedersen

omerfirmak:
uses NethermindEth/crypto-cpp

## Description

Currently fails to link because `crypto-cpp` needs extra build steps

## Types of changes

## Testing

**Requires testing**: Yes

## Documentation


omerfirmak:
All green @IronGauntlets

joshklop (changes requested):

omerfirmak:
false alarm, I compiled `crypto-cpp` in debug mode :| 

in release mode

BenchmarkArrayDigest/Benchmark_core.PedersenDigest_over_20_felt.Felts-16         	     135	   9109949 ns/op

IronGauntlets:
Could you please explain how the `make-crypto-cpp.sh` is working? Also, I am not sure why we need to clone the crypto repo. Is it not possible to use `go get` to get the dependencies? I have never had the chance to use `cgo` before.

omerfirmak:
> Could you please explain how the `make-crypto-cpp.sh` is working? Also, I am not sure why we need to clone the crypto repo.

It doesnt do much, clones the repo, builds it and puts the generated library in an env variable that `go build` will use when searching for it.

>  Is it not possible to use `go get` to get the dependencies? I have never had the chance to use `cgo` before.

it is possible for simple projects, but crypto-cpp is a `cmake` project with a complex build setup. AFAICS, there is no way to make `go build` to build a `cmake` project. Only option is to build the library separately and only link it in the `go build` stage.

IronGauntlets:
Once the dependency is downloaded it exists in the `$GOPATH/pkg/mod/github.com/...`, can we not just call cmake from there? Rather than cloning the repo again?

omerfirmak:
> Once the dependency is downloaded it exists in the `$GOPATH/pkg/mod/github.com/...`, 

Yeah, but  I didnt want to get in the way of `go`, since it manages that directory. I will give it a try and see how it does.

IronGauntlets:
You should be able to produce the output in the local directory without changing anything the gopath, right?

omerfirmak:
> You should be able to produce the output in the local directory without changing anything the gopath, right?

Not without changing the build scripts, which I dont want to. They could be really cumbersome to maintain if changed upstream

IronGauntlets:
I see

core/hash.go
```diff
@@ -0,0 +1,37 @@
+package core
+
+import (
+	lib "github.com/NethermindEth/crypto-cpp"
+	"github.com/NethermindEth/juno/core/felt"
+)
+
+// Implements the [Pedersen hash]
```
joshklop, line 8: I think the comments on the two Pedersen functions are swapped. Please feel free to ignore and resolve these comments if I'm missing something.

```suggestion
// PedersenDigest implements [Pedersen array hashing].
```

core/hash.go
```diff
@@ -0,0 +1,37 @@
...
+//
+// [Pedersen hash]: https://docs.starknet.io/documentation/develop/Hashing/hash-functions/#pedersen_hash
```
joshklop, line 10: Aside: would "PedersenArrayHash" be a more accurate and consistent name for the PedersenDigest function? I thought digest = hash, so PedersenDigest and PedersenHash are synonymous, but currently denote two different functions.

I think the comments on the two Pedersen functions are swapped. If not, please feel free to ignore and resolve these suggestions.
```suggestion
// PedersenDigest implements [Pedersen array hashing].
//
// [Pedersen array hashing]: https://docs.starknet.io/documentation/develop/Hashing/hash-functions/#pedersen_hash
```

core/hash.go
```diff
@@ -0,0 +1,37 @@
...
+	if err := PedersenHash(acc, len, acc); err != nil {
+		return nil, err
+	}
+
+	return acc, nil
+}
+
+// Implements [Pedersen array hashing]
+//
+// [Pedersen array hashing]: https://docs.starknet.io/documentation/develop/Hashing/hash-functions/#pedersen_hash
```
joshklop, line 29:
```suggestion
// PedersenHash implements the [Pedersen hash function].
//
// [Pedersen hash function]: https://docs.starknet.io/documentation/develop/Hashing/hash-functions/#pedersen_hash
```

core/hash.go
jelilat: Noticed that `ArrayDigest` was used in the test function [here](https://github.com/NethermindEth/juno/pull/436/files#diff-ea14ff3f44475668d0756a9a69bfc7f9ac63deef09948915436e3fbc903feaeeR40). We can either go with that or change both to `PedersenArrayHash`. But I also think `PedersenArrayHash` is more appropriate.

core/hash_test.go
```diff
@@ -0,0 +1,115 @@
...
+			want, _ := new(felt.Felt).SetString(tt.want)
+			err := PedersenHash(a, b, ans)
+			if err != nil || !ans.Equal(want) {
+				t.Errorf("TestHash got %s, want %s", ans.Text(16), want.Text(16))
+			}
+		})
+	}
+}
+
+func TestArrayDigest(t *testing.T) {
```
jelilat, line 40: I think this should be changed to match the naming of the [actual function](https://github.com/NethermindEth/juno/pull/436/files#diff-44445e1dae498e96f85d121a65c447ceece32d6bf8df1ecc1f6d427b7262f4afR11).
```suggestion
func TestPedersenArrayHash(t *testing.T) {
```

Suggesting `TestPedersenArrayHash` should the function name is changed to `PedersenArrayHash`

core/hash_test.go
```diff
@@ -0,0 +1,115 @@
...
+	tests := [...]struct {
+		input []string
+		want  string
+	}{
+		// Contract address calculation. See the following links for how the
+		// it is carried out and the result referenced.
+		//
+		//	- https://docs.starknet.io/docs/Contracts/contract-address/
```
jelilat, line 48: This is a dead link. Returns `Page not found`

```suggestion
		//	- https://docs.starknet.io/documentation/develop/Contracts/contract-address/
```

core/hash_test.go
```diff
@@ -0,0 +1,115 @@
...
+		//	- https://alpha-goerli.starknet.io/feeder_gateway/get_transaction?transactionHash=0x1b50380d45ebd70876518203f131a12428b2ac1a3a75f1a74241a4abdd614e8
```
jelilat, line 49: This is also a dead link
```suggestion
		//	- https://alpha4.starknet.io/feeder_gateway/get_transaction?transactionHash=0x1b50380d45ebd70876518203f131a12428b2ac1a3a75f1a74241a4abdd614e8
```

core/hash_test.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+}
+
+func TestArrayDigest(t *testing.T) {
+	tests := [...]struct {
+		input []string
+		want  string
+	}{
+		// Contract address calculation. See the following links for how the
+		// it is carried out and the result referenced.
```
jelilat, line 46:
```suggestion
		// Contract address calculation. See the following links for how the
		// calculation is carried out and the result referenced.
```

core/hash_test.go
```diff
@@ -0,0 +1,115 @@
...
+		}
+		want, _ := new(felt.Felt).SetString(test.want)
+		got, err := PedersenDigest(data...)
+		if err != nil || got.Cmp(want) != 0 {
+			t.Errorf("PedersenDigest(%x) = %x, want %x", data, got, want)
+		}
+	}
+}
+
+func BenchmarkArrayDigest(b *testing.B) {
```
jelilat, line 100: This should also be changed accordingly
```suggestion
func BenchmarkPedersenArrayHash(b *testing.B) {
```

jelilat (changes requested):

core/hash.go
```diff
@@ -0,0 +1,37 @@
...
+
+import (
+	lib "github.com/NethermindEth/crypto-cpp"
+	"github.com/NethermindEth/juno/core/felt"
+)
+
+// Implements the [Pedersen hash]
+//
+// [Pedersen hash]: https://docs.starknet.io/documentation/develop/Hashing/hash-functions/#pedersen_hash
+func PedersenDigest(elements ...*felt.Felt) (*felt.Felt, error) {
```
jelilat, line 11:
```suggestion
func PedersenArrayHash(elements ...*felt.Felt) (*felt.Felt, error) {
```

jelilat (changes requested):

omerfirmak:
Fixed MacOS

Moved CPP sources to juno repo.

Addressed all reviews

joshklop (approved):

jelilat (approved):

core/pedersen_hash.go
```diff
@@ -0,0 +1,45 @@
...
+
+// Pedersen implements the [Pedersen hash]
+//
+// [Pedersen hash]: https://docs.starknet.io/documentation/develop/Hashing/hash-functions/#pedersen_hash
+func Pedersen(a *felt.Felt, b *felt.Felt) (*felt.Felt, error) {
+	out := make([]byte, crypto.BufferSize)
+	if err := crypto.Hash(a.Marshal(), b.Marshal(), out); err != nil {
+		return nil, err
+	}
+	return new(felt.Felt).SetBytes(out[:crypto.FeltSize]), nil
```
omerfirmak, line 44: My use of a third pointer argument as the output was actually intentional. I was expecting it to let us avoid alot GC allocations when done that way. 

When done this way, a call to `PedersenArray` will allocate a `Felt` on heap for each `Pedersen` call. With a third output argument, it only needs to allocate once.

core/pedersen_hash.go
```diff
@@ -0,0 +1,45 @@
...
+// PedersenArray implements [Pedersen array hashing]
+//
+// [Pedersen array hashing]: https://docs.starknet.io/documentation/develop/Hashing/hash-functions/#pedersen_hash
+func PedersenArray(elems ...*felt.Felt) (*felt.Felt, error) {
+	if len(elems) < 3 {
+		return nil, errors.New("number of elems must be more than 2")
+	}
+
+	var err error
+	d := new(felt.Felt).SetZero()
```
omerfirmak, line 19: Does `new` always allocate on heap? Or does it do escape analysis?

`d` (same goes for `l`)  never escapes this function scope, so there is no need to allocate in on the heap. We should make sure that any ephemeral variable is allocated on stack.

core/pedersen_hash.go
```diff
@@ -0,0 +1,45 @@
...
+	}
+
+	return Pedersen(d, l)
+}
+
+// Pedersen implements the [Pedersen hash]
+//
+// [Pedersen hash]: https://docs.starknet.io/documentation/develop/Hashing/hash-functions/#pedersen_hash
+func Pedersen(a *felt.Felt, b *felt.Felt) (*felt.Felt, error) {
+	out := make([]byte, crypto.BufferSize)
```
omerfirmak, line 40: same here, should be on stack.

core/pedersen_hash.go
IronGauntlets: [Looks like](https://stackoverflow.com/questions/67895905/go-memory-allocation-new-objects-pointers-and-escape-analysis) that the `new` keyword doesn't necessarily allocate on the stack.

IronGauntlets: [This](https://blog.devgenius.io/in-depth-analysis-of-golang-memory-escape-edfbfb856913) article mentions the allocation happens after escape analysis.

Issue #433, opened 2022-11-10
Design Efficient State Storage


joshklop:
A flat storage is preferred for efficiency.

There are three parts to this task:
- [ ] State storage (class storage and contract instance storage)
- [ ] The Trie
- [ ] StateDiff/Changeset (i.e., storing historical state diffs)

omerfirmak:
@joshklop Do you mind if I take a shot at implementing the Trie?

joshklop:
I don't mind at all :)

PR #437, opened 2022-11-15
State Trie

omerfirmak:
Towards #433 

Depends on #436 

## Description

Implements the trie defined in https://docs.starknet.io/documentation/develop/State/starknet-state/

## Types of changes

## Testing

**Requires testing**: Yes

### DIFF
--- a/core/trie.go
+++ b/core/trie.go
@@ -3,6 +3,8 @@ package core
 import (
 	"bytes"
 	"errors"
+	"fmt"
+	"strings"
 
 	"github.com/NethermindEth/juno/core/felt"
 	"github.com/bits-and-blooms/bitset"
@@ -93,7 +95,7 @@ func (n *TrieNode) UnmarshalBinary(data []byte) error {
 }
 
 type Trie struct {
-	root    StoragePath
+	root    *StoragePath
 	storage TrieStorage
 }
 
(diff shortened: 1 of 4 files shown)


omerfirmak:
Now successfully builds state for the diff here https://alpha-mainnet.starknet.io/feeder_gateway/get_state_update?blockNumber=0

core/trie.go
```diff
@@ -0,0 +1,351 @@
...
+		}
+
+		var pathP **StoragePath
+		if head == 'l' {
+			pathP = &(n.left)
+		} else if head == 'r' {
+			pathP = &(n.right)
+		} else {
+			return errors.New("Malformed TrieNode bytedata")
+		}
```
joshklop, line 117:
```suggestion
		switch head {
		case 'l':
			pathP = &(n.left)
		case 'r':
			pathP = &(n.right)
		default:
			return errors.New("Malformed TrieNode bytedata")
		}
```

core/trie.go
```diff
@@ -0,0 +1,351 @@
...
+		_, err = (*pathP).ReadFrom(stream)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// State [Trie]
```
joshklop, line 129:
```suggestion
// Trie is a dense Merkle Patricia Trie (i.e., all internal nodes have
// two children).
//
// This implementation allows for a "flat" storage by keying nodes on
// their path rather than their hash, resulting in O(1) accesses and
// O(lg n) insertions.
//
// The state trie [specification] describes a sparse Merkle Trie. Note
// that this dense implementation results in an equivalent commitment.
//
// [specification]: https://docs.starknet.io/documentation/develop/State/starknet-state/
```

core/trie.go
```diff
@@ -0,0 +1,351 @@
...
+type Trie struct {
+	root    *StoragePath
+	storage TrieStorage
+}
+
```
joshklop, line 134: I think we need a `NewTrie` function here so we can create Tries outside this package, since `storage` is private.
```suggestion

func NewTrie(storage TrieStorage) *Trie {
	return &Trie{
		storage: storage,
	}
}

```

core/trie.go
```diff
@@ -0,0 +1,351 @@
...
+	commonPath := shorterPath.Clone()
+	for i := uint(0); i < shorterPath.Len()-divergentBit+1; i++ {
+		commonPath.DeleteAt(0)
+	}
+	return commonPath, divergentBit == shorterPath.Len()+1
+}
+
+// calculates node paths according to the [docs]
+//
+// [docs]: https://docs.starknet.io/documentation/develop/State/starknet-state/
```
joshklop, line 159:
```suggestion
// GetSpecPath returns the suffix of path that diverges from
// parentPath. For example, for a path 0b1011 and parentPath 0b11,
// this function would return the StoragePath object for 0b10.
//
// This is the canonical representation for paths used in the
// [specification]. Since this trie implementation stores nodes nodes by
// path, we need this function to convert paths to their canonical
// representation.
//
// [specification]: https://docs.starknet.io/documentation/develop/State/starknet-state/
```

core/trie.go
```diff
@@ -0,0 +1,351 @@
...
+func GetSpecPath(path, parentPath *StoragePath) *StoragePath {
+	specPath := path.Clone()
+	// drop parent path, and one more MSB since left/right relation already encodes that information
+	if parentPath != nil {
+		specPath.Shrink(specPath.Len() - parentPath.Len() - 1)
+		specPath.DeleteAt(specPath.Len() - 1)
+	}
+	return specPath
+}
+
```
joshklop, line 169:
```suggestion

// step is the on-disk representation of a [TrieNode], where path is the
// key and node is the value.
```

core/trie.go
```diff
@@ -0,0 +1,351 @@
...
+type step = struct {
+	path *StoragePath
+	node *StorageValue
+}
+
+// Enumerates the set of [TrieNode] objects that is on a given [StoragePath]
```
joshklop, line 175:
```suggestion
// stepsToRoot enumerates the set of [TrieNode] objects that are on a 
// given [StoragePath].
//
// The [step]s are returned in descending order beginning with the root.
```

core/trie.go
```diff
@@ -0,0 +1,351 @@
...
+func (t *Trie) stepsToRoot(path *StoragePath) (steps []step) {
+	cur := t.root
+	for cur != nil {
+		node, err := t.storage.Get(cur)
+		if err != nil {
+			return
```
joshklop, line 181: Should we return the error here?

core/trie.go
```diff
@@ -0,0 +1,351 @@
...
+
+type (
+	StoragePath  = bitset.BitSet
+	StorageValue = TrieNode
+)
+
+type (
+	TrieKey   = felt.Felt
+	TrieValue = felt.Felt
+)
```
joshklop, line 22: I personally prefer using the underlying types directly. IMHO, type aliasing usually creates an unnecessary layer of indirection. Using type names to describe "structure" and variable names to describe "purpose" is my preferred convention.

In the end this probably comes down to taste and I'm open to keeping things as they are.

joshklop (changes requested):
This looks great! I think I understand what's going on now. I haven't had a chance to look at the tests yet, but the implementation itself looks nice to me.

core/trie.go
```diff
@@ -0,0 +1,351 @@
...
+	specPath := path.Clone()
+	// drop parent path, and one more MSB since left/right relation already encodes that information
+	if parentPath != nil {
+		specPath.Shrink(specPath.Len() - parentPath.Len() - 1)
+		specPath.DeleteAt(specPath.Len() - 1)
+	}
+	return specPath
+}
+
+type step = struct {
```
joshklop, line 170:
```suggestion
type step struct {
```

omerfirmak:
Rebased, updated according to comments

IronGauntlets:
Just a small note on the comments: all types of comments should begin with the name of the type or with "A TypeName"/ "The TypeName". For example for `TrieStorage` it should be:

```go
// TrieStorage represents the interface for persistent storage for [Trie]
type TrieStorage interface {
	Put(key *bitset.BitSet, value *TrieNode) error
	Get(key *bitset.BitSet) (*TrieNode, error)
}
```
If you are using Golang, this should remove those annoying warnings.

IronGauntlets:
Do you think it is a good idea to move it to a separate package?

omerfirmak:
> Do you think it is a good idea to move it to a separate package ie core/trie?

My understanding of what a `package` should represent is still pretty vague. If you think it makes sense, we definitely can.

omerfirmak:
Made Trie height configurable as well.

IronGauntlets (approved):

Issue #440, opened 2022-11-22
Implement Feeder Gateway client


omerfirmak:
Feeder Gateway, for now, is one of the sources of truth in StarkNet. We need to be able to query the gateway to fetch data from it.

PR #441, opened 2022-11-22
Add a basic GatewayClient

omerfirmak:
Currently only supports "get_state_update" endpoint because this was started to be able to query the gateway for Trie test data.

Could be used a template for future endpoints

The change in felt.go is warranted because gateway inconsistently omits "0x" prefix for some hex values. So if we fail to parse a hex value as field element, we try again hoping that it is a hexadecimal with a missing prefix

Towards #440

### DIFF
--- a/clients/block_11817.json
+++ b/clients/block_11817.json
@@ -1,0 +1,4829 @@
+{
+  "block_hash": "0x24c692acaed3b486990bd9d2b2fbbee802b37b3bd79c59f295bad3277200a83",
+  "parent_block_hash": "0x3873ccb937f14429b169c654dda28886d2cc2d6ea17b3cff9748fe5cfdb67e0",
+  "block_number": 11817,
+  "state_root": "03df24be7b5fed6b41de08d38686b6142944119ca2a345c38793590d6804bba4",
+  "status": "ACCEPTED_ON_L2",
+  "gas_price": "0x27ad16775",
+  "transactions": [
+    {
+      "transaction_hash": "0xd38080575a9133cecdd5c80a79a513ff49d7dd478d8fd6ffe75fdac532e810",
+      "version": "0x0",
+      "max_fee": "0x135832dacd5000",
+      "signature": [
+        "0x25b889e9c7625eb35b647085ea149445dfaa893f95d0346f49bb80ab9c237b1",
+        "0x8fff8b2a44b5b89ec655492cd7a92fe5ef3986cd8b7d5237da52cc528b2d61"
+      ],
+      "contract_address": "0x1b706b18846667c44fdf2dc302ed90b98c0ae3aa6e041fdbdea1289d0284b03",
+      "entry_point_selector": "0x15d40a3d6ca2ac30f4031e42be28da9b056fef9bb7357ac5e85627ee876e5ad",
+      "calldata": [
+        "0x1",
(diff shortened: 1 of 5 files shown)


internal/clients/gateway.go
```diff
@@ -0,0 +1,238 @@
...
+
+type ClassDefinition struct {
+	Abi         interface{} `json:"abi"`
+	EntryPoints struct {
+		Constructor []EntryPoint `json:"CONSTRUCTOR"`
+		External    []EntryPoint `json:"EXTERNAL"`
+		L1Handler   []EntryPoint `json:"L1_HANDLER"`
+	} `json:"entry_points_by_type"`
+	Program struct {
+		Builtins         []string     `json:"builtins"`
```
jelilat, line 214: Compiler version was added in Starknet 0.10, so we need to include it here. All we just have to do is handle cases where they are not present.
```suggestion
		Builtins         []string     `json:"builtins"`
		CompilerVersion string `json:"compiler_version"`
```

internal/clients/gateway.go
```diff
@@ -0,0 +1,238 @@
...
+		Prime            string       `json:"prime"`
+		ReferenceManager interface{}  `json:"reference_manager"`
+		Identifiers      interface{}  `json:"identifiers"`
+		Attributes       interface{}  `json:"attributes"`
+		Data             []*felt.Felt `json:"data"`
```
jelilat, line 219: I think the data values returned from the gateway are strings, not Felts

internal/clients/gateway_test.go
```diff
@@ -0,0 +1,267 @@
...
+	assert.Equal(t, "3873ccb937f14429b169c654dda28886d2cc2d6ea17b3cff9748fe5cfdb67e0", block.ParentHash.Text(16))
+	assert.Equal(t, uint64(11817), block.Number)
+	assert.Equal(t, "3df24be7b5fed6b41de08d38686b6142944119ca2a345c38793590d6804bba4", block.StateRoot.Text(16))
+	assert.Equal(t, "ACCEPTED_ON_L2", block.Status)
+	assert.Equal(t, "27ad16775", block.GasPrice.Text(16))
+	assert.Equal(t, 52, len(block.Transactions))
+	assert.Equal(t, 52, len(block.Receipts))
+	assert.Equal(t, uint64(1669465009), block.Timestamp)
+	assert.Equal(t, "0.10.1", block.Version)
+}
```
jelilat, line 267: I think we need tests for `GetClassDefinition` as well

internal/clients/gateway.go
jelilat:
```suggestion
		Data             []string `json:"data"`
```

jelilat (changes requested):

internal/clients/gateway.go
omerfirmak: Unmarshaller converts them to felts.

internal/clients/gateway.go
omerfirmak: Nice catch, thank you!

internal/clients/gateway_test.go
omerfirmak: I must have forgotten to push the latest changes I amended, fixed

jelilat (changes requested):

IronGauntlets (approved):

jelilat (approved):

PR #443, opened 2022-11-22
Add StarkNet Keccak

IronGauntlets:
## Description

Add StarkNet Keccak

## Types of changes

## Testing

**Requires testing**: Yes

## Documentation

### DIFF
--- a/core/crypto/keccak.go
+++ b/core/crypto/keccak.go
@@ -1,0 +1,23 @@
+package crypto
+
+import (
+	"github.com/NethermindEth/juno/core/felt"
+	"golang.org/x/crypto/sha3"
+)
+
+var h = sha3.NewLegacyKeccak256()
+
+// StarkNetKeccak implements [StarkNet keccak]
+//
+// [StarkNet keccak]: https://docs.starknet.io/documentation/develop/Hashing/hash-functions/#starknet_keccak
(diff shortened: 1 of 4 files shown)


core/crypto/keccak.go
```diff
@@ -0,0 +1,23 @@
+package crypto
+
+import (
+	"github.com/NethermindEth/juno/core/felt"
+	"golang.org/x/crypto/sha3"
+)
+
+var h = sha3.NewLegacyKeccak256()
```
omerfirmak, line 8: Why are we using the legacy version?

Docs say;

```
// Only use this function if you require compatibility with an existing cryptosystem
// that uses non-standard padding. All other users should use New256 instead.
```

IronGauntlets: Yes, because legacy keccak and sha3 use different padding, so it results in a different hash. Otherwise, they would have said to use sha3 instead of Keccak.

IronGauntlets: Please have a look [here](https://ethereum.stackexchange.com/questions/30369/difference-between-keccak256-and-sha3). `LegacyKeccak256` is not officially supported by NIST.

omerfirmak: Got it, thanks

omerfirmak (approved):
LGTM, feel free to merge after rebasing on #446

PR #456, opened 2022-11-25
mdbx implementation

### LINKED ISSUE #455: MDBX implementation
Implementation of MDBX wrapper using the [mdbx-go](https://github.com/torquem-ch/mdbx-go) bindings.

joshklop:
Closes #455 

## Description

Add the database interfaces and implementation.
- `Database`: contains all of Juno's tables (or "buckets"). This is represented by mdbx's "environment" struct, which manages the on-disk database file. 
- `Transaction`: creates buckets and opens cursors on buckets.
- `Cursor`: modifies the database, moves to keys in the database

We will probably need to add more functionality to the `Cursor` interface (e.g, `ForEach`, `PutMulti`, etc.) in the future, but it may make sense to add those incrementally rather than making this PR larger.

## Types of changes

- Documentation Update

## Testing

**Requires testing**: Yes

## Documentation

**If this requires a documentation update, did you add one?** Yes: comments added where necessary

### DIFF
--- a/core/felt/doc.go
+++ b/core/felt/doc.go
@@ -21,30 +21,33 @@
 // The modulus is hardcoded in all the operations.
 //
 // Field elements are represented as an array, and assumed to be in Montgomery form in all methods:
-// 	type Felt [4]uint64
 //
-// Usage
+//	type Felt [4]uint64
+//
+// # Usage
 //
 // Example API signature:
-// 	// Mul z = x * y (mod q)
-// 	func (z *Element) Mul(x, y *Element) *Element
+//
+//	// Mul z = x * y (mod q)
+//	func (z *Element) Mul(x, y *Element) *Element
 //
 // and can be used like so:
-// 	var a, b Element
-// 	a.SetUint64(2)
-// 	b.SetString("984896738")
-// 	a.Mul(a, b)
-// 	a.Sub(a, a)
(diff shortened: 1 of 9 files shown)


joshklop:
Remaining to do:
* [ ] Add error types/functions
* [ ] Investigate mdbx flags and options to ensure we're using a robust configuration
* [ ] Add additional functionality to the cursor interface, if possible.

omerfirmak (approved):
Some descriptions to the interface members could be nice. It would help to guide people without much KV database experience. I had to go in to tests to understand what was happening.

omerfirmak (commented):
and references for the future;

Unless you are building a feature incrementally, commits that change/remove/fix code that were introduced in the same PR should be squashed in to the original commit. This makes it so much easier to review the PR commit by commit. In this case "remove ..." commits fall in to this category.

unrelated commits should be in a separate PR. ie "format" here

joshklop:
Thanks for the feedback and tips @omerfirmak! Much appreciated. I've updated the comments on the db interface and will aim for your suggested git conventions in the future.

omerfirmak (approved):
Feel free to merge at will.

Just a question, can we delete a key from a bucket? Or `Put`ing a `nil` to a key is essentially that?

PR #460, opened 2022-11-28
update pedersen

jelilat:
## Description

Pedersen Hash is used a lot in the hash calculations and sometimes, we can't predict the input number (whether the elements are > 2 or less than 2). For example, [this](https://github.com/NethermindEth/juno/blob/jelilat/implement_class_hash/core/contract/contract.go#L165-L178) in the class hash computation.

The current solution is checking this before computing the Pedersen hash. But given the fact that we may have to do this multiple times, it's neater to check in the Pedersen Hash function

## Changes:

- Change 1
- Change 2

## Types of changes

- Improve existing functionality

## Testing

**Requires testing**: Yes

### DIFF
--- a/core/crypto/pedersen_hash.go
+++ b/core/crypto/pedersen_hash.go
@@ -43,3 +43,15 @@ func Pedersen(a *felt.Felt, b *felt.Felt) (*felt.Felt, error) {
 	}
 	return new(felt.Felt).SetBytes(out[:starkware.FeltSize]), nil
 }
+
+func PedersenHash(elems ...*felt.Felt) (*felt.Felt, error) {
+	if len(elems) < 2 {
+		return nil, errors.New("number of elems must be 2 and above")
+	}
+
+	if len(elems) < 3 {
+		return Pedersen(elems[0], elems[1])
+	} else {
+		return PedersenArray(elems...)
+	}
+}
--- a/core/crypto/pedersen_hash.go
+++ b/core/crypto/pedersen_hash.go
@@ -45,13 +45,14 @@ func Pedersen(a *felt.Felt, b *felt.Felt) (*felt.Felt, error) {
 }
 
 func PedersenHash(elems ...*felt.Felt) (*felt.Felt, error) {
-	if len(elems) < 2 {
-		return nil, errors.New("number of elems must be 2 and above")
-	}
-
-	if len(elems) < 3 {
+	switch len(elems) {
+	case 0:
+		return nil, errors.New("number of elems must be more than 0")
+	case 1:
+		return Pedersen(new(felt.Felt).SetZero(), elems[0])
+	case 2:
 		return Pedersen(elems[0], elems[1])
-	} else {
+	default:
 		return PedersenArray(elems...)
 	}
 }
--- a/core/crypto/pedersen_hash_test.go
+++ b/core/crypto/pedersen_hash_test.go
@@ -131,6 +131,43 @@ func TestPedersenArray(t *testing.T) {
 	}
 }
 
+func TestPedersenHash(t *testing.T) {
+	tests := []struct {
+		elems   []*felt.Felt
+		wantErr bool
+	}{
+
+		{
+			elems:   []*felt.Felt{},
+			wantErr: true,
+		},
+		{
+			elems:   []*felt.Felt{new(felt.Felt).SetOne()},
+			wantErr: true,
+		},
+		{
+			elems:   []*felt.Felt{new(felt.Felt).SetOne(), new(felt.Felt).SetZero()},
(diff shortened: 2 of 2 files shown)


omerfirmak:
I am not sure ,f we need this, arrays are hashed with `starknet_keccak` not `pedersen`.

```
For every line in the above, excluding the first, compute h(h(previous_line),new_line), where the hash of an array is defined as [starknet_keccak](https://docs.starknet.io/documentation/architecture_and_concepts/Hashing/hash-functions/#starknet-keccak).
```

am I missing something?

jelilat:
> I am not sure if we need this, arrays are hashed with `starknet_keccak` not `pedersen`.
> 
> ```
> For every line in the above, excluding the first, compute h(h(previous_line),new_line), where the hash of an array is defined as [starknet_keccak](https://docs.starknet.io/documentation/architecture_and_concepts/Hashing/hash-functions/#starknet-keccak).
> ```
> 
> am I missing something?

No, you aren't. You are right.

I was using the Pedersen hash instead of the starknet hash for the array computation

core/crypto/pedersen_hash.go
```diff
@@ -43,3 +43,16 @@ func Pedersen(a *felt.Felt, b *felt.Felt) (*felt.Felt, error) {
 	}
 	return new(felt.Felt).SetBytes(out[:starkware.FeltSize]), nil
 }
+
+func PedersenHash(elems ...*felt.Felt) (*felt.Felt, error) {
+	switch len(elems) {
+	case 0:
+		return nil, errors.New("number of elems must be more than 0")
+	case 1:
+		return Pedersen(new(felt.Felt).SetZero(), elems[0])
```
IronGauntlets, line 52: Are you sure this is always the case? As far as I understand `Pedersen(a,b) != Pedersen(b,a)`. So shouldn't it be left to the caller to decide how they want to call Pedersen when they only want to hash a since value?

omerfirmak:
I still am not sure if we need this. You should be able to just use `PedersenArray` [here](https://github.com/NethermindEth/juno/blob/jelilat/implement_class_hash/core/contract/contract.go#L165-L178). HashChain in pathfinder does that, if I am not mistaken.

IronGauntlets:
I have looked at the description of the [class hash](https://docs.starknet.io/documentation/architecture_and_concepts/Contracts/contract-hash/#how_the_class_hash_is_computed), and it is very confusing. So it might work by asking pathfinder how it is done.

Also, you should make this change in the other branch and if the change works then include it as a separate commit in the pull request. Just as a standalone pr it doesn't add much context.

jelilat:
> I still am not sure if we need this. You should be able to just use `PedersenArray` [here](https://github.com/NethermindEth/juno/blob/jelilat/implement_class_hash/core/contract/contract.go#L165-L178). HashChain in pathfinder does that, if I am not mistaken.

Our `PedersenArray` only accepts 3 or more inputs. Pathfinder's `HashChain`, on the other hand, accepts 1 or more inputs.  

In the snippet below, for example, the `CONSTRUCTOR` has only 2 values  (selector and offset) while `EXTERNAL` has 4 values in total. If we are to hash this, we'll have to check if the inputs are 2 or more to decide which function to call (`PedersenHash` or `PedersenArray`).

```
"entry_points_by_type": {
    "CONSTRUCTOR": [
      {
        "selector": "0x28ffe4ff0f226a9107253e17a904099aa4f63a02a5621de0576e5aa71bc5194",
        "offset": "0x71"
      }
    ],
    "EXTERNAL": [
      {
        "selector": "0x362398bec32bc0ebb411203221a35a0301193a96f317ebe5e40be9f60d15320",
        "offset": "0x3a"
      },
      {
        "selector": "0x39e11d48192e4333233c7eb19d10ad67c362bb28580c604d67884c85da39695",
        "offset": "0x5b"
      }
    ],
    "L1_HANDLER": []
  },
```

jelilat:
> I have looked at the description of the [class hash](https://docs.starknet.io/documentation/architecture_and_concepts/Contracts/contract-hash/#how_the_class_hash_is_computed), and it is very confusing. So it might be worth asking pathfinder how it is done.
> 
> Also, you should make this change in the other branch and if the change works then include it as a separate commit in the pull request. Just as a standalone pr it doesn't add much context.

Closing this again due to this. Will be making the changes (if necessary) in the `implement_class_hash` branch

omerfirmak:
> Our PedersenArray only accepts 3 or more inputs

then let's just remove this constraint, it looks like an oversight.

jelilat:
> > Our PedersenArray only accepts 3 or more inputs
> 
> then let's just remove this constraint, it looks like an oversight.

Alright

PR #462, opened 2022-11-29
Refactor

IronGauntlets:
## Description

Move packages and remove the internal directory since it is not required for now.

## Changes:

- Rename `internal/juno` -> `internal/node`
- Move `internal/utils` and `internal/node` -> `utils` and `node` respectively

## Types of changes

## Testing

**Requires testing**: No


IronGauntlets:
I have looked at the description of the [class hash](https://docs.starknet.io/documentation/architecture_and_concepts/Contracts/contract-hash/#how_the_class_hash_is_computed), and it is very confusing. So it might work by asking pathfinder how it is done.

Also, you should make this change in the other branch and if the change works then include it as a separate commit in the pull request. Just as standalone pr it doesn't add much context.

Issue #459, opened 2022-11-25
Block implementation


jelilat:
The Block module should compute the Block hash for a given json blob

PR #467, opened 2022-12-13
Implement Block Hash

jelilat:
Closes https://github.com/NethermindEth/juno/issues/459

## Types of changes

## Testing

**Requires testing**: Yes


core/block.go
```diff
@@ -14,23 +18,157 @@ type Block struct {
...
+		if b.SequencerAddress == nil {
+			b.SequencerAddress = blockHashMetaInfo.FallBackSequencerAddress
+		}
+		blockHash, err := b.post07Hash()
+		if err != nil {
+			return nil, err
+		}
+
+		return blockHash, nil
+	}
```
joshklop, line 110:
```suggestion
	if b.Number < blockHashMetaInfo.First07Block {
		return b.pre07Hash(chain)
	} else if b.SequencerAddress == nil {
	        b.SequencerAddress = blockHashMetaInfo.FallBackSequencerAddress
        }
        return b.post07Hash()
```

core/block.go
```diff
@@ -14,23 +18,157 @@ type Block struct {
...
-	// Todo: implement pedersen hash as defined here
-	// https://docs.starknet.io/documentation/develop/Blocks/header/#block_hash
-	return nil
+type BlockHashMetaInfo struct {
+	First07Block             uint64     // First block that uses the post-0.7.0 block hash algorithm
+	UnverifiableRange        []uint64   // Range of blocks that are not verifiable
+	FallBackSequencerAddress *felt.Felt // The sequencer address to use for blocks that do not have one
+}
+
+func getBlockHashMetaInfo(chain string) (BlockHashMetaInfo, error) {
```
joshklop, line 42: Let's return a `*BlockHashMetaInfo` here so we can return `nil` on error.

core/crypto/pedersen_hash.go
```diff
@@ -11,17 +9,15 @@ import (
...
-	}
-
 	var err error
 	d := new(felt.Felt).SetZero()
 
-	for _, e := range elems {
-		d, err = Pedersen(d, e)
-		if err != nil {
-			return nil, err
+	if len(elems) > 0 {
```
joshklop, line 15: Do we need this check?

joshklop (changes requested):
Great work!

core/crypto/pedersen_hash.go
jelilat: I think I added it cos I was getting some error on arrays of length 0. Will cross-check again

core/block.go
```diff
@@ -14,23 +18,145 @@ type Block struct {
...
+	FallBackSequencerAddress *felt.Felt // The sequencer address to use for blocks that do not have one
+}
+
+func getBlockHashMetaInfo(chain string) (*BlockHashMetaInfo, error) {
+	switch chain {
+	case "SN_MAIN":
+		fallBackSequencerAddress, _ := new(felt.Felt).SetString("0x021f4b90b0377c82bf330b7b5295820769e72d79d8acd0effa0ebde6e9988bc5")
+		return &BlockHashMetaInfo{
+			First07Block:             883,
+			UnverifiableRange:        []uint64{},
```
joshklop, line 48: If there is no `UnverifiableRange` for a block, we don't need to initialize it. Then on line 85, we can check `if unverifiableRange != nil`.

core/block.go
```diff
@@ -14,23 +18,145 @@ type Block struct {
...
+	if err != nil {
+		return nil, err
+	}
+
+	unverifiableRange := blockHashMetaInfo.UnverifiableRange
+	if len(unverifiableRange) > 0 {
+		// Check if the block number is in the unverifiable range
+		if b.Number >= unverifiableRange[0] && b.Number <= unverifiableRange[len(unverifiableRange)-1] {
+			// If so, return unverifiable block error
+			return nil, errors.New("block " + strconv.FormatUint(b.Number, 10) + " is unverifiable")
```
joshklop, line 89: Slightly more idiomatic error message.
```suggestion
			return nil, errors.New("block is unverifiable: " + strconv.FormatUint(b.Number, 10))
```

core/block.go
```diff
@@ -14,23 +18,145 @@ type Block struct {
...
+		zeroFelt,                // reserved: protocol version
+		zeroFelt,                // reserved: extra data
+		chainId,                 // extra data: chain id
+		b.ParentHash,            // parent hash
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	return blockHash, nil
```
joshklop, line 136:
```suggestion
	return crypto.PedersenArray(
		blockNumber,             // block number
		b.GlobalStateRoot,       // global state root
		zeroFelt,                // reserved: sequencer address
		zeroFelt,                // reserved: block timestamp
		b.TransactionCount,      // number of transactions
		b.TransactionCommitment, // transaction commitment
		zeroFelt,                // reserved: number of events
		zeroFelt,                // reserved: event commitment
		zeroFelt,                // reserved: protocol version
		zeroFelt,                // reserved: extra data
		chainId,                 // extra data: chain id
		b.ParentHash,            // parent hash
	)
```

core/block.go
```diff
@@ -14,23 +18,145 @@ type Block struct {
...
+//
+// Also, for these blocks, we use zeroes for:
+// - sequencer address
+// - block timestamp
+// - number of events
+// - event commitment
+func (b *Block) pre07Hash(chain string) (*felt.Felt, error) {
+	blockNumber := new(felt.Felt).SetUint64(b.Number)
+
+	zeroFelt := new(felt.Felt).SetUint64(0)
```
joshklop, line 114:
```suggestion
	zeroFelt := new(felt.Felt)
```

core/block.go
```diff
@@ -14,23 +18,145 @@ type Block struct {
...
+		return nil, err
+	}
+
+	return blockHash, nil
+}
+
+func (b *Block) post07Hash() (*felt.Felt, error) {
+	blockNumber := new(felt.Felt).SetUint64(b.Number)
+
+	zeroFelt := new(felt.Felt).SetUint64(0)
```
joshklop, line 142:
```suggestion
	zeroFelt := new(felt.Felt)
```

core/block.go
```diff
@@ -14,23 +18,145 @@ type Block struct {
...
+		b.EventCommitment,       // event commitment
+		zeroFelt,                // reserved: protocol version
+		zeroFelt,                // reserved: extra data
+		b.ParentHash,            // parent block hash
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	return blockHash, nil
```
joshklop, line 161:
```suggestion
	return crypto.PedersenArray(
		blockNumber,             // block number
		b.GlobalStateRoot,       // global state root
		b.SequencerAddress,      // sequencer address
		b.Timestamp,             // block timestamp
		b.TransactionCount,      // number of transactions
		b.TransactionCommitment, // transaction commitment
		b.EventCount,            // number of events
		b.EventCommitment,       // event commitment
		zeroFelt,                // reserved: protocol version
		zeroFelt,                // reserved: extra data
		b.ParentHash,            // parent block hash
	)
```

joshklop (changes requested):

core/block.go
```diff
@@ -14,23 +19,130 @@ type Block struct {
...
+	FallBackSequencerAddress *felt.Felt // The sequencer address to use for blocks that do not have one
+}
+
+func getBlockHashMetaInfo(network utils.Network) *blockHashMetaInfo {
+	switch network {
+	case utils.MAINNET:
+		fallBackSequencerAddress, _ := new(felt.Felt).SetString("0x021f4b90b0377c82bf330b7b5295820769e72d79d8acd0effa0ebde6e9988bc5")
+		return &blockHashMetaInfo{
+			First07Block:             883,
+			UnverifiableRange:        nil,
```
joshklop, line 49: `UnverifiableRange` will automatically be set to `nil`.
```suggestion
```

core/block.go
```diff
@@ -14,23 +19,130 @@ type Block struct {
...
+		return &blockHashMetaInfo{
+			First07Block:             47028,
+			UnverifiableRange:        []uint64{119802, 148428},
+			FallBackSequencerAddress: fallBackSequencerAddress,
+		}
+	case utils.GOERLI2:
+		fallBackSequencerAddress, _ := new(felt.Felt).SetString("0x046a89ae102987331d369645031b49c27738ed096f2789c24449966da4c6de6b")
+		return &blockHashMetaInfo{
+			First07Block:             0,
+			UnverifiableRange:        nil,
```
joshklop, line 63: Same as above.
```suggestion
```

core/block.go
```diff
@@ -14,23 +19,130 @@ type Block struct {
...
+// Block hash computation according to https://docs.starknet.io/documentation/develop/Blocks/header/#block_hash
+func (b *Block) Hash(network utils.Network) (*felt.Felt, error) {
+	blockHashMetaInfo := getBlockHashMetaInfo(network)
+
+	unverifiableRange := blockHashMetaInfo.UnverifiableRange
+	if unverifiableRange != nil {
+		// Check if the block number is in the unverifiable range
+		if b.Number >= unverifiableRange[0] && b.Number <= unverifiableRange[1] {
+			// If so, return unverifiable block error
+			return nil, errors.New("block is unverifiable" + strconv.FormatUint(b.Number, 10))
```
joshklop, line 87:
```suggestion
			return nil, errors.New("block is unverifiable: " + strconv.FormatUint(b.Number, 10))
```

core/block.go
```diff
@@ -14,23 +19,130 @@ type Block struct {
...
+// - sequencer address
+// - block timestamp
+// - number of events
+// - event commitment
+func (b *Block) pre07Hash(chain *felt.Felt) (*felt.Felt, error) {
+	blockNumber := new(felt.Felt).SetUint64(b.Number)
+
+	zeroFelt := new(felt.Felt)
+
+	return crypto.PedersenArray(
```
joshklop, line 114:
```suggestion
	zeroFelt := new(felt.Felt)
	return crypto.PedersenArray(
```

core/block.go
```diff
@@ -14,23 +19,130 @@ type Block struct {
...
+		b.ParentHash,            // parent hash
+	)
+}
+
+func (b *Block) post07Hash() (*felt.Felt, error) {
+	blockNumber := new(felt.Felt).SetUint64(b.Number)
+
+	zeroFelt := new(felt.Felt)
+
+	return crypto.PedersenArray(
```
joshklop, line 135:
```suggestion
	zeroFelt := new(felt.Felt)
	return crypto.PedersenArray(
```

core/block.go
joshklop: On second thought, we need to export this error so clients can check against it. See [here](https://gosamples.dev/check-error-type/) for an example (the second example is most like what we need to do; the block number should be a field in the error struct).

joshklop (changes requested):
This is looking good! Can we add tests for the mainnet, goerli2, and integration networks? We should also test a block that's in the "unverifiable" range.

core/block.go
```diff
@@ -14,23 +19,134 @@ type Block struct {
...
-	// https://docs.starknet.io/documentation/develop/Blocks/header/#block_hash
-	return nil
+type blockHashMetaInfo struct {
+	First07Block             uint64     // First block that uses the post-0.7.0 block hash algorithm
+	UnverifiableRange        []uint64   // Range of blocks that are not verifiable
+	FallBackSequencerAddress *felt.Felt // The sequencer address to use for blocks that do not have one
+}
+
+type UnverifiableBlockError struct {
+	input string
```
joshklop, line 44: Let's make the field name a bit more descriptive like `blockNumber`. I think we can also use a `uint64` here, allowing us to skip the string conversion on line 93.

joshklop (changes requested):

joshklop (changes requested):

core/block.go
```diff
@@ -14,23 +18,134 @@ type Block struct {
...
+// Computes the block hash for blocks generated before Cairo 0.7.0
+//
+// The major difference between the pre-0.7.0 and post-0.7.0 block hashes is that
+// the pre-0.7.0 block hash includes the chain id in the hash computation.
+//
+// Also, for these blocks, we use zeroes for:
+// - sequencer address
+// - block timestamp
+// - number of events
+// - event commitment
```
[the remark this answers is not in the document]
jelilat: Why do you think we should remove the preceding comments?

joshklop (approved):

Issue #466, opened 2022-12-08
Use `gnark-crypto` to implement StarkNet Curve


omerfirmak:
See: https://github.com/ConsenSys/gnark-crypto/issues/273

For Pedersen Hash; we need scalar point multiplication and point addition on the StarkNet curve.
Down the line, we will need full ECDSA features as well.

PR #468, opened 2022-12-15
Use gnark-crypto for felt and Pedersen hash

joshklop:
Towards #466 

## Description

Use `gnark-crypto` for the felt and Pedersen hash primitives.

## Changes:

- Replace `felt` package with `starkcurve/fp` package
- Reimplement pedersen hash with `starkcurve` package, resulting in 3x performance improvement

## Types of changes

## Testing

**Requires testing**: No, only modifying existing tests in a few places (since `Pedersen` and `PedersenArray` functions no longer return errors in any situations).

## Documentation

**If this requires a documentation update, did you add one?** Comments where appropriate.


joshklop:
Some things we need to figure out:

- [ ] The feeder gateway has an inconsistent representation for felts (e.g., sometimes uses the `0x` prefix, sometimes not). The strict unmarshaler in the `fp` sub-package only accepts strings with the `0x` prefix. Previously, [we had modified](https://github.com/NethermindEth/juno/blob/0fce0f1e9fe53ea1de56660464f5f046800d4f63/core/felt/felt.go#L996-L1000) the `felt` package to fix this. Since we are now importing it as a third-party library, we need to find a different solution. 
- [ ] The implementation for `PedersenArray` previously returned an error if the user tried to hash an array with less than three elements. The spec doesn't say that the function is only defined for arrays with three or more elements, so I removed this behavior. Even though it probably won't have an impact (the `PedersenArray` is only used in a few places, every time on arrays of more than three elements), it would be best to double-check before proceeding.

joshklop:
Note to self: although the pull request is ready for review, we should probably wait to merge it until the stark-curve feature branch is in the latest `gnark-crypto` release.

joshklop:
Split this into two separate PRs #476 #475

PR #469, opened 2022-12-16
Implement class hash

joshklop:
## Description

Now that we have a functioning crypto llibrary, we can implement the class hash.

## Changes:

- Implement the class hash.

## Types of changes

## Testing

**Requires testing**: Yes

## Documentation

**If this requires a documentation update, did you add one?** Comments where appropriate.


jelilat (approved):

core/contract_test.go
```diff
@@ -0,0 +1,242 @@
...
+	//go:embed testdata/bytecode.json
+	bytecodeBytes []byte
+	//go:embed testdata/bytecode_genesis.json
+	bytecodeGenesisBytes []byte
+	//go:embed testdata/bytecode_0_8.json
+	bytecodeCairo08Bytes []byte
+)
+
+func TestClassHash(t *testing.T) {
+	var bytecode []*felt.Felt
```
jelilat, line 21: Can you confirm what Cairo version this is as well? We can also rename the `json` file to reflect the cairo version. 

Given that Starknet is known to sometimes change their hash implementations with newer Cairo versions, documenting the Cairo version could be useful in the future.

jelilat (changes requested):

core/contract_test.go
joshklop: Nice catch! Fixed now, thanks!

jelilat (approved):

PR #471, opened 2022-12-19
update networks

jelilat:
## Description

Starknet has 2 other networks that Juno doesn't support at the moment. Supporting these networks is required for us to correctly compute all block hashes

## Changes:

- Added support for the `GOERLI2` and `INTEGRATION` networks
- Added Chain id

## Types of changes

## Testing

**Requires testing**: Yes

### DIFF
--- a/utils/network.go
+++ b/utils/network.go
@@ -5,6 +5,8 @@ type Network uint8
 const (
 	GOERLI Network = iota
 	MAINNET
+	GOERLI2
+	INTEGRATION
 )
 
 func (n Network) String() string {
@@ -13,6 +15,10 @@ func (n Network) String() string {
 		return "goerli"
 	case MAINNET:
 		return "mainnet"
+	case GOERLI2:
+		return "goerli2"
+	case INTEGRATION:
+		return "integration"
 	default:
 		return ""
 	}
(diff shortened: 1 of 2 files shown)


utils/network.go
```diff
@@ -24,6 +30,34 @@ func (n Network) URL() string {
...
+	case GOERLI2:
+		return "https://alpha4.starknet.io"
+	case INTEGRATION:
+		return "https://external.integration.starknet.io"
+	default:
+		return ""
+	}
+}
+
+type Chain string
```
joshklop, line 42: Is this type necessary?

jelilat: The chains could as well just be a string but I felt it might be cool to different ChainIds from normal `strings`

joshklop: Is it possible for us to use the `Network` type exclusively? It seems to serve the same purpose as this `Chain` type.

jelilat: Okay. Made changes

Please take a look

utils/network.go
```diff
@@ -24,6 +30,25 @@ func (n Network) URL() string {
...
+	case GOERLI2:
+		return "https://alpha4.starknet.io"
+	case INTEGRATION:
+		return "https://external.integration.starknet.io"
+	default:
+		return ""
+	}
+}
+
+func (n Network) ChainId() string {
```
joshklop, line 42: Nice! I just noticed the chain ID is [defined](https://docs.starknet.io/documentation/architecture_and_concepts/Blocks/transactions/#chain-id) to be the `uint` value of the byte-encoded string (`[]byte("SN_MAIN")`). Since this only seems to be used for the transaction hash, we can cheat a little and return a `felt` here.

joshklop (changes requested):

utils/network.go
jelilat: Got you

joshklop (approved):

PR #474, opened 2022-12-20
Implement Transaction and Event Commitments

jelilat:
## Description

Transaction and event commitments are required to verify the block hash.

## Changes:

- Added transaction and event commitment functions.

## Types of changes

## Testing

**Requires testing**: Yes

### DIFF
--- a/core/block.go
+++ b/core/block.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/NethermindEth/juno/core/crypto"
 	"github.com/NethermindEth/juno/core/felt"
+	"github.com/NethermindEth/juno/core/trie"
 	"github.com/NethermindEth/juno/utils"
 )
 
@@ -148,4 +149,64 @@ func (b *Block) post07Hash() *felt.Felt {
 		zeroFelt,                // reserved: extra data
 		b.ParentHash,            // parent block hash
 	)
+}
+
+// TransactionCommitment is the root of a height 64 binary Merkle Patricia tree of the
+// transaction hashes and signatures in a block.
+func TransactionCommitment(receipts []*TransactionReceipt) (*felt.Felt, error) {
+	var transactionCommitment *felt.Felt
+	trie.RunOnTempTrie(64, func(trie *trie.Trie) error {
+		zeroFelt := new(felt.Felt)
+		emptySignatureHash := crypto.Pedersen(zeroFelt, zeroFelt)
+		for i, receipt := range receipts {
+			var signaturesHash *felt.Felt
+			if receipt.Type == Invoke {
+				signaturesHash = crypto.PedersenArray(receipt.Signatures...)
+			} else {
+				signaturesHash = emptySignatureHash
+			}
+			transactionAndSignatureHash := crypto.Pedersen(receipt.TransactionHash, signaturesHash)
+			if err := trie.Put(new(felt.Felt).SetUint64(uint64(i)), transactionAndSignatureHash); err != nil {
+				return err
+			}
+		}
+		root, err := trie.Root()
+		if err != nil {
+			return err
+		}
+		transactionCommitment = root
+		return nil
+	})
+
+	return transactionCommitment, nil
+}
+
+// EventData computes the event commitment and event count for a block.
+func EventData(receipts []*TransactionReceipt) (*felt.Felt, uint64, error) {
+	var eventCommitment *felt.Felt // root of a height 64 binary Merkle Patricia tree of the events in a block.
+	var eventCount uint64          // number of events in a block.
+	trie.RunOnTempTrie(64, func(trie *trie.Trie) error {
+		for _, receipt := range receipts {
+			for _, event := range receipt.Events {
+				eventHash := crypto.PedersenArray(
+					event.From,
+					crypto.PedersenArray(event.Keys...),
+					crypto.PedersenArray(event.Data...),
+				)
+
+				if err := trie.Put(new(felt.Felt).SetUint64(eventCount), eventHash); err != nil {
+					return err
+				}
+				eventCount++
+			}
+		}
+		root, err := trie.Root()
+		if err != nil {
+			return err
+		}
+		eventCommitment = root
+		return nil
+	})
+
+	return eventCommitment, eventCount, nil
 }
--- a/core/block.go
+++ b/core/block.go
@@ -155,15 +155,12 @@ func (b *Block) post07Hash() *felt.Felt {
 // transaction hashes and signatures in a block.
 func TransactionCommitment(receipts []*TransactionReceipt) (*felt.Felt, error) {
 	var transactionCommitment *felt.Felt
-	trie.RunOnTempTrie(64, func(trie *trie.Trie) error {
+	return transactionCommitment, trie.RunOnTempTrie(64, func(trie *trie.Trie) error {
 		zeroFelt := new(felt.Felt)
-		emptySignatureHash := crypto.Pedersen(zeroFelt, zeroFelt)
 		for i, receipt := range receipts {
-			var signaturesHash *felt.Felt
+			signaturesHash := crypto.Pedersen(zeroFelt, zeroFelt)
 			if receipt.Type == Invoke {
 				signaturesHash = crypto.PedersenArray(receipt.Signatures...)
-			} else {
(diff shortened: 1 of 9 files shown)


core/block.go
```diff
@@ -34,3 +39,119 @@ func (b *Block) Hash() *felt.Felt {
...
+
+func (s *commitmentTrieStorage) Delete(key *bitset.BitSet) error {
+	keyEnc, _ := key.MarshalBinary()
+	delete(s.storage, hex.EncodeToString(keyEnc))
+	return nil
+}
+
+// TransactionCommitment is the root of a height 64 binary Merkle Patricia tree of the
+// transaction hashes and signatures in a block.
+func TransactionCommitment(transactions []*clients.Transaction) (*felt.Felt, error) {
```
joshklop, line 84: We should use core types here. If the core types are missing information, we can modify them.

joshklop (changes requested):
This looks fantastic overall! I don't have much time for reviews today, so just leaving the single comment for now.

core/block.go
jelilat: Or maybe we should move the commitment implementation to `clients`? What do you think?

joshklop: I think it would be best to have the commitment calculation performed on the core types. This way, when the feeder gateway types are deprecated, we don't need to change nearly as much code.

jelilat: Yeah, you're right. I didn't consider the depreciation of the feeder gateway types

core/block.go
```diff
@@ -149,3 +150,77 @@ func (b *Block) post07Hash() *felt.Felt {
...
+			for _, event := range receipt.Events {
+				keys := crypto.PedersenArray(event.Keys...)
+
+				data := crypto.PedersenArray(event.Data...)
+
+				eventHash := crypto.PedersenArray(
+					event.From,
+					keys,
+					data,
+				)
```
joshklop, line 205:
```suggestion
				eventHash := crypto.PedersenArray(
					event.From,
					crypto.PedersenArray(event.Keys...),
					crypto.PedersenArray(event.Data...),
				)
```

core/block.go
```diff
@@ -149,3 +150,77 @@ func (b *Block) post07Hash() *felt.Felt {
...
+
+				err := trie.Put(new(felt.Felt).SetUint64(eventCount), eventHash)
+				if err != nil {
+					return err
+				}
```
joshklop, line 210:
```suggestion
				if err := trie.Put(new(felt.Felt).SetUint64(eventCount), eventHash); err != nil {
				        return err
				}
```

core/block.go
```diff
@@ -149,3 +150,77 @@ func (b *Block) post07Hash() *felt.Felt {
...
+// TransactionCommitment is the root of a height 64 binary Merkle Patricia tree of the
+// transaction hashes and signatures in a block.
+func TransactionCommitment(transactions []*TransactionReceipt) (*felt.Felt, error) {
+	var transactionCommitment *felt.Felt
+	trie.RunOnTempTrie(64, func(trie *trie.Trie) error {
+		zeroFelt := new(felt.Felt)
+		emptySignatureHash := crypto.Pedersen(zeroFelt, zeroFelt)
+		for i, transaction := range transactions {
+			var signaturesHash *felt.Felt
+			if transaction.Type == "INVOKE_FUNCTION" {
```
joshklop, line 163: Can we make `Transaction.Type` an enum? Something like:
```go
type TransactionType int

const (
        Declare TransactionType = iota
        Deploy
        Invoke
)
```

joshklop (changes requested):
Looks great.

core/transaction.go
```diff
@@ -1,9 +1,66 @@
...
 )
 
+type Event struct {
+	Data []*felt.Felt
+	From *felt.Felt
+	Keys []*felt.Felt
+}
+
+type L1ToL2Message struct {
+	From     string
```
joshklop, line 16: Is the `From` field an Ethereum address? If so, I assume the `To` field in `L2ToL1Message` is also an Ethereum address. Is that correct?

core/block.go
```diff
@@ -149,3 +150,72 @@ func (b *Block) post07Hash() *felt.Felt {
 		b.ParentHash,            // parent block hash
 	)
 }
+
+// TransactionCommitment is the root of a height 64 binary Merkle Patricia tree of the
+// transaction hashes and signatures in a block.
+func TransactionCommitment(transactions []*TransactionReceipt) (*felt.Felt, error) {
```
joshklop, line 156: Can we rename the parameter to `receipts`?

core/block.go
```diff
@@ -149,3 +150,72 @@ func (b *Block) post07Hash() *felt.Felt {
...
+			if transaction.Type == Invoke {
+				signaturesHash = crypto.PedersenArray(transaction.Signatures...)
+			} else {
+				signaturesHash = emptySignatureHash
+			}
+			transactionAndSignatureHash := crypto.Pedersen(transaction.TransactionHash, signaturesHash)
+			err := trie.Put(new(felt.Felt).SetUint64(uint64(i)), transactionAndSignatureHash)
+			if err != nil {
+				return err
+			}
```
joshklop, line 172:
```suggestion
			err := trie.Put(new(felt.Felt).SetUint64(uint64(i)), transactionAndSignatureHash); err != nil {
			        return err
			}
```

core/block.go
```diff
@@ -149,3 +150,72 @@ func (b *Block) post07Hash() *felt.Felt {
...
+	return transactionCommitment, nil
+}
+
+type EventData struct {
+	EventCommitment *felt.Felt // root of a height 64 binary Merkle Patricia tree of the events in a block.
+	EventCount      *felt.Felt // number of events in a block.
+}
+
+// ComputeEventData computes the event commitment and event count for a block.
+func ComputeEventData(receipts []*TransactionReceipt) (*EventData, error) {
```
joshklop, line 191: Rather than creating a new `EventData` type solely for the purpose of combining return values from this function, it may be better to return `(*felt.Felt, uint64, error)`.

core/block_test.go
```diff
@@ -177,3 +179,190 @@ func TestBlockHash(t *testing.T) {
...
+			"0x6f499789aabb31935810ce89d6ea9e9d37c5921c0d7fae2bd68f2fff5b7b93f",
+		},
+	}
+
+	for _, test := range tests {
+		eventData, _ := ComputeEventData(test.receipts)
+		if "0x"+(eventData.EventCommitment).Text(16) != test.want {
+			t.Errorf("got %s, want %s", "0x"+(eventData.EventCommitment).Text(16), test.want)
+		}
+	}
```
joshklop, line 367: We can have a subtest for each case, giving us more informative messages like "receipt1Goerli".

joshklop (changes requested):

core/transaction.go
jelilat: Yes, that is correct

joshklop: Great! In this case, we could use a more precise type than `string`. Maybe `common.Address` from geth or even `[20]byte`?

core/block_test.go
```diff
@@ -177,3 +179,213 @@ func TestBlockHash(t *testing.T) {
...
+}
+
+func assertCorrectCommitment(t *testing.T, got *felt.Felt, want string) {
+	t.Helper()
+	if "0x"+got.Text(16) != want {
+		t.Errorf("got %s, want %s", "0x"+got.Text(16), want)
+	}
+}
+
+func TestTransactionCommitment(t *testing.T) {
```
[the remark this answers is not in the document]
jelilat: I guess I misunderstood you when you said "receipt1Goerli". I thought you meant further splitting the subtests into the different networks.

core/transaction.go
jelilat: Alright. I think `common.Address` is better. That's what is used to represent Ethereum addresses.

core/block_test.go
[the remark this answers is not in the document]
joshklop: Ah ok no worries. I meant calling `t.Run` for each case inside the loop and adding the `description` field to the test table so we can have a unique subtitle for each case.

omerfirmak (approved):

IronGauntlets (approved):

joshklop (approved):

PR #476, opened 2022-12-22
Implement pedersen hash using gnark-crypto

joshklop:
See: https://github.com/ConsenSys/gnark-crypto/issues/273

Towards #466 

## Description

See linked issues above.

## Changes:

- Implement pedersen hash using gnark-crypto

## Types of changes

## Testing

**Requires testing**: Yes

**Did you write tests??**: Yes (modified tests where necessary)


core/crypto/pedersen_hash.go
```diff
@@ -1,38 +1,84 @@
 package crypto
 
 import (
-	"github.com/NethermindEth/juno/core/crypto/starkware"
+	"math/big"
+
 	"github.com/NethermindEth/juno/core/felt"
+	"github.com/consensys/gnark-crypto/ecc/stark-curve"
```
jelilat, line 7: Probably nothing
```suggestion
	starkcurve "github.com/consensys/gnark-crypto/ecc/stark-curve"
```

core/crypto/pedersen_hash.go
```diff
@@ -1,38 +1,84 @@
...
+)
+
+var (
+	shiftPoint  *starkcurve.G1Jac
+	p0          *starkcurve.G1Jac
+	p1          *starkcurve.G1Jac
+	p2          *starkcurve.G1Jac
+	p3          *starkcurve.G1Jac
+	lowPartMask [4]uint64
 )
```
jelilat, line 18: `lowPartMask` doesn't seem to be used. I think it should be removed if we aren't using it. Or we could do this
```suggestion
)

const (
	lowPartMask = 0xffffffffffffff
	lowPartBits = 56
)
```

then update then on lines [78](https://github.com/NethermindEth/juno/pull/476/files#diff-9ac8648882322077a92f77befbdec5aeee0a34a247faaea30e6a79a281fd1a55R78) and [81](https://github.com/NethermindEth/juno/pull/476/files#diff-9ac8648882322077a92f77befbdec5aeee0a34a247faaea30e6a79a281fd1a55R81)

jelilat (changes requested):

core/crypto/pedersen_hash.go
joshklop: The package name is `starkcurve`, so we don't need to set it manually. Package names [are distinct](https://go.dev/blog/package-names) from package paths (though they are often the same).

jelilat: Yeah. Got you.

The suggestion is more about readability for me

joshklop: I looked into this a bit more and I now think you're right (Uber's Go style guide [agrees](https://github.com/uber-go/guide/blob/master/style.md#import-aliasing) with your suggestion). I'll update the PR.

core/crypto/pedersen_hash.go
```diff
@@ -1,38 +1,88 @@
...
+	resultBytes := new(starkcurve.G1Affine).FromJacobian(result).X.Bytes()
+	return new(felt.Felt).SetBytes(resultBytes[:])
+}
+
+func processElement(a *fp.Element, p1 *starkcurve.G1Jac, p2 *starkcurve.G1Jac) *starkcurve.G1Jac {
+	lowPart := a.ToRegular()
+	lowPart[3] = lowPart[3] & lowPartMask // Zero-out the top nibble (bits 249-252)
+	m := new(starkcurve.G1Jac).ScalarMultiplication(p1, lowPart.ToBigInt(new(big.Int)))
+
+	highPart := new(big.Int).SetUint64(a.ToRegular()[3] >> lowPartBits & 0xf) // The top nibble (bits 249-252)
```
jelilat, line 85: `& 0xf` seem redundant here since `a` is a `felt(fp.Element)` which means that `a.ToRegular()[3] >> lowPartBits` can never be bigger than a `felt`
```suggestion
	highPart := new(big.Int).SetUint64(a.ToRegular()[3] >> lowPartBits) // The top nibble (bits 249-252)
```

jelilat (changes requested):

core/crypto/pedersen_hash.go
joshklop: Nice catch, thanks! Fixed.

jelilat (approved):

omerfirmak (approved):
I've done a very minor cleanup and added the migration to `fp.Element`.

IronGauntlets (approved):

PR #478, opened 2022-12-24
add trie storage

jelilat:
## Description

The Trie Storage is used by many parts of the application. So, it was moved to `trie.go` for accessibility

## Types of changes

## Testing

**Requires testing**: Yes

**Did you write tests??**: No. I used existing tests because they are sufficient.

### DIFF
--- a/core/trie.go
+++ b/core/trie.go
@@ -3,6 +3,7 @@ package core
 import (
 	"bytes"
 	"encoding/binary"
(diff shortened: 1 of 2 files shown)


jelilat:
@omerfirmak recommended we get rid of the trie storage completely and use`TrieTxn` with an in-memory badger database instead.

PR #480, opened 2022-12-26
Implement contract address

jelilat:
## Description

Contract address computation is required to compute the `deploy transaction` hash. Hence, the implementation.

## Types of changes

## Testing

**Requires testing**: Yes

### DIFF
--- a/core/contract.go
+++ b/core/contract.go
@@ -66,12 +66,9 @@ type Contract struct {
 }
 
 // ContractAddress computes the address of a StarkNet contract.
-func ContractAddress(callerAddress, classHash, salt *felt.Felt, constructorCalldata []*felt.Felt) (*felt.Felt, error) {
+func ContractAddress(callerAddress, classHash, salt *felt.Felt, constructorCalldata []*felt.Felt) *felt.Felt {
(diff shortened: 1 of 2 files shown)


omerfirmak:
Needs a rebase

core/contract.go
```diff
@@ -64,3 +64,18 @@ type Contract struct {
 	// Root of the contract's storage trie.
 	StorageRoot *felt.Felt // TODO: is this field necessary?
 }
+
+// ContractAddress computes the address of a StarkNet contract.
+func ContractAddress(callerAddress, classHash, salt *felt.Felt, constructorCalldata []*felt.Felt) *felt.Felt {
```
joshklop, line 69: Drive-by comment: intuitively, I feel this should be a method on `Contract` so we can do something like `contract.Address()`. It feels weird to me that we need information from outside the `Contract` to calculate the address, especially since a contract is stored [by its address](https://docs.starknet.io/documentation/architecture_and_concepts/State/starknet-state/) in the trie.

Happy to hear different opinions on this.

IronGauntlets: I have to agree. However, for now I will add it as a todo, rebase and merge the branch.

IronGauntlets (approved):

PR #482, opened 2022-12-28
Benchmark for different types of serializations

chee-chyuan:
## Description

Adding Marshalling and UnMarshalling functions for a few different types of serliazations. included benchmarking code

Local benchmarking results for speed:
image

Local benchmarking results for size after encoding:
image


IronGauntlets:
Closing since `cbor` was chosen as the preferred serialization.

PR #484, opened 2023-01-04
Add custom Unmarshal to StateDiff struct

chee-chyuan:
Fixes | Closes | Resolves #

## Description

Current we use encoding/json to property unmarshal json to type `map[string][]interface{}` in StateUpdate. This is fine but we will require an extra step to convert the string key to a Felt key. 

This pr aims to unmarshal to type `map[Felt][]interface{}` by having a custom UnmarshalJson logic

## Changes:

- add `func (s *StateDiff) UnmarshalJSON(data []byte) error {`

## Types of changes

- Other (please describe): Use custom `UnmarshalJSON` to unmarshal `StateUpdate`

## Testing

**Requires testing**: Yes/No

**Did you write tests??**: Yes/No

## Documentation


omerfirmak:
I don't think this needs to be done in `clients` package.

`GatewayClient` should be wrapped in another struct that exposes `DataSource` API. That struct will be responsible of serving requests from all parts of our codebase. Essentially it will let `GatewayClient` do the networking and convert the response it gets (eg. `clients.StateUpdate`) to our `core` types (eg. `core.StateUpdate`)

chee-chyuan:
> I don't think this needs to be done in `clients` package.
> 
> `GatewayClient` should be wrapped in another struct that exposes `DataSource` API. That struct will be responsible of serving requests from all parts of our codebase. Essentially it will let `GatewayClient` do the networking and convert the response it gets (eg. `clients.StateUpdate`) to our `core` types (eg. `core.StateUpdate`)

sounds like it is related to this issue https://github.com/NethermindEth/juno/issues/495 . happy to take that on too seeing that they are related

omerfirmak:
Done in https://github.com/NethermindEth/juno/pull/501

PR #487, opened 2023-01-06
Implement Program Hash

jelilat:
## Description

This PR adds a function to compute the program hash computation

## Changes:

- Adds the program hash computation

## Types of changes

## Testing

**Requires testing**: Yes

### DIFF
--- a/clients/programhash.go
+++ b/clients/programhash.go
@@ -1,0 +1,130 @@
+package clients
+
+import (
+	"bytes"
+	"encoding/json"
+	"sort"
+	"strconv"
+	"strings"
+
+	"github.com/NethermindEth/juno/core/crypto"
+	"github.com/NethermindEth/juno/core/felt"
+)
+
+type ContractCode struct {
+	Abi     interface{} `json:"abi"`
+	Program *Program    `json:"program"`
+}
+
+func ProgramHash(contractDefinition *ClassDefinition) (*felt.Felt, error) {
+	program := contractDefinition.Program
+
+	// make debug info None
+	program.DebugInfo = nil
+
+	// Cairo 0.8 added "accessible_scopes" and "flow_tracking_data" attribute fields, which were
+	// not present in older contracts. They present as null/empty for older contracts deployed
+	// prior to adding this feature and should not be included in the hash calculation in these cases.
+	//
+	// We, therefore, check and remove them from the definition before calculating the hash.
+	if program.Attributes != nil {
+		attributes := program.Attributes.([]interface{})
+		if len(attributes) == 0 {
+			program.Attributes = nil
+		} else {
+			for key, attribute := range attributes {
+				attributeInterface := attribute.(map[string]interface{})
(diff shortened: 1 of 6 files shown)


clients/programhash_test.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/NethermindEth/juno/core/felt"
+)
+
+var (
+	//go:embed testdata/class_genesis.json
+	classGenesisBytes []byte
+	//go:embed testdata/class_0_8.json
+	classCairo08Bytes []byte
+	//go:embed testdata/class_0_10.json
+	classCairo10Bytes []byte
```
IronGauntlets, line 18: Please add links to where this data was retrieved from

omerfirmak:
I think the way we generate the JSON is so brittle that it would be so cumbersome to maintain. I gave a shot at this one as well and made some progress but it is not complete.

check out `MarshalJSONWithSeparators` at https://github.com/NethermindEth/juno/commit/385474075d64c29d871418e7c438b3528e406645

jelilat:
> I think the way we generate the JSON is so brittle that it would be really cumbersome to maintain. I gave a shot at this one as well and made some progress but it is not complete.
> 
> check out `MarshalJSONWithSeparators` at [3854740](https://github.com/NethermindEth/juno/commit/385474075d64c29d871418e7c438b3528e406645)

I absolutely agree and have been experimenting with different things. Thanks for your update. Trying something new that could further simplify the implementation as well. Hopefully, it works.

jelilat:
@omerfirmak I just made some improvements to `MarshalsJSON`. Please take a look and check for further improvements.

clients/programhash.go
```diff
@@ -0,0 +1,130 @@
...
+	contractCodeString := string(contractCodeBytes)
+	hintsIndex := strings.Index(contractCodeString, `"hints":`)            // find location of "hints" in contractCodeString
+	identifierIndex := strings.Index(contractCodeString, `"identifiers":`) // find location of "identifiers" in contractCodeString
+	// replace json "hints" with reformatted "hints"
+	contractCodeString = contractCodeString[:(hintsIndex+8)] + newHint + contractCodeString[identifierIndex-1:]
+	// reformat contractCodeString with
+	contractCodeString = strings.ReplaceAll(contractCodeString, `,`, `, `)
+	contractCodeString = strings.ReplaceAll(contractCodeString, `,  `, `, `)
+	contractCodeString = strings.ReplaceAll(contractCodeString, `:`, `: `)
+	contractCodeString = strings.ReplaceAll(contractCodeString, `:  `, `: `)
```
omerfirmak, line 89: These could easily cause problems as those characters could be used anywhere, not just as delimiters. Lets merge and see if this ends up being a problem.

also we should use escaped whitespace characters (`\r`, `\n`, `\t` etc), it is not easily recognizable what the whitespaces are in string constants.

omerfirmak (approved):
We should push for standardization of the program JSON. This is just unnecessary amount of work for something so simple.

PR #483, opened 2023-01-01
Implement state

joshklop:
## Description

Starts implementing the state. Implements a few changes (mostly minor refactoring) on top of PR #465. See that PR for discussion/description.

PR #492, opened 2023-01-09
Badger DB storage backend for Trie

omerfirmak:
This provides us a both persistent and non-persistent storage for our Trie. Non-persistent Tries are particularly useful when calculating transaction and event commitments.

Extracted from #483


core/trie.go
```diff
@@ -144,6 +146,21 @@ func NewTrie(storage TrieStorage, height uint) *Trie {
...
+	}
+	defer db.Close()
+
+	txn := db.NewTransaction(true)
+	defer txn.Discard()
+
+	trieTxn := NewTrieTxn(txn, nil)
+	return do(NewTrie(trieTxn, height))
+}
+
```
IronGauntlets, line 163: It should be moved to the test file, right?

core/trietxn.go
```diff
@@ -0,0 +1,68 @@
...
+func NewTrieTxn(badgerTxn *badger.Txn, prefix []byte) *TrieTxn {
+	return &TrieTxn{
+		badgerTxn: badgerTxn,
+		prefix:    prefix,
+	}
+}
+
+// dbKey creates a byte array to be used as a key to our KV store
+// it simply appends the given key to the configured prefix
+func (txn *TrieTxn) dbKey(key *bitset.BitSet) ([]byte, error) {
```
IronGauntlets, line 23: Please change the `func (txn *TrieTxn)` -> `func (t *TrieTxn)` to follow go conventions for all the functions.

core/trietxn_test.go
```diff
@@ -0,0 +1,85 @@
...
+	if err := db.Update(func(txn *badger.Txn) error {
+		tTxn := &TrieTxn{txn, prefix}
+
+		value, _ := new(felt.Felt).SetRandom()
+		node.UnmarshalBinary(value.Marshal())
+
+		return tTxn.Put(key, node)
+	}); err != nil {
+		t.Error(err)
+	}
```
IronGauntlets, line 32: These should be separated into separate subsets so that if one of them fails we can see immediately which one failed. So please wrap all the cases in `t.Run(caseName, func(t *t.testing.T)`

core/trietxn.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	if item, err := txn.badgerTxn.Get(dbKey); err == nil {
+		node := new(TrieNode)
+		return node, item.Value(func(val []byte) error {
+			return node.UnmarshalBinary(val)
+		})
+	} else {
+		return nil, err
+	}
```
IronGauntlets, line 59:
```suggestion
	if item, err := txn.badgerTxn.Get(dbKey); err != nil {
		return nil, err
	} else {
		node := new(TrieNode)
		return node, item.Value(func(val []byte) error {
			return node.UnmarshalBinary(val)
		})
	}
```

core/trie_test.go
```diff
@@ -5,6 +5,9 @@ import (
 	"fmt"
 	"testing"
 
+	"github.com/NethermindEth/juno/db"
+	"github.com/dgraph-io/badger/v3"
+
```
IronGauntlets, line 10:
```suggestion
```

core/trietxn_test.go
```diff
@@ -0,0 +1,85 @@
...
+
+	key := bitset.New(44)
+	node := new(TrieNode)
+
+	// put a node
+	if err := db.Update(func(txn *badger.Txn) error {
+		tTxn := &TrieTxn{txn, prefix}
+
+		value, _ := new(felt.Felt).SetRandom()
+		node.UnmarshalBinary(value.Marshal())
```
IronGauntlets, line 27: Please don't ignore the error

core/trietxn_test.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/NethermindEth/juno/db"
+
+	"github.com/NethermindEth/juno/core/felt"
+	"github.com/bits-and-blooms/bitset"
+	"github.com/dgraph-io/badger/v3"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestTrieTxn(t *testing.T) {
+	db := db.NewTestDb()
```
IronGauntlets, line 16: Please rename db to something esle since it conflicts with import

core/trietxn_test.go
```diff
@@ -0,0 +1,85 @@
+package core
+
+import (
+	"errors"
+	"testing"
+
+	"github.com/NethermindEth/juno/db"
+
```
IronGauntlets, line 8:
```suggestion
```

core/trie.go
```diff
@@ -144,6 +146,21 @@ func NewTrie(storage TrieStorage, height uint) *Trie {
...
+// DoOnTempTrie creates an in-memory Trie of height `height` and runs `do` on that Trie
+func DoOnTempTrie(height uint, do func(*Trie) error) error {
+	db, err := db.NewInMemoryDb()
+	if err != nil {
+		return err
+	}
+	defer db.Close()
+
+	txn := db.NewTransaction(true)
+	defer txn.Discard()
```
IronGauntlets, line 158: Shouldn't this be `txn.Commit()`?

core/trie.go
```diff
@@ -144,6 +146,21 @@ func NewTrie(storage TrieStorage, height uint) *Trie {
 	}
 }
 
+// DoOnTempTrie creates an in-memory Trie of height `height` and runs `do` on that Trie
+func DoOnTempTrie(height uint, do func(*Trie) error) error {
```
IronGauntlets, line 150: Also can we please rename this to `RunOnTempTrie`?

core/trie.go
IronGauntlets: And we should follow the guide for manually creating transactions as mentioned [here](https://dgraph.io/docs/badger/get-started/#read-write-transactions). Especially. if the transaction become too big?

core/trie.go
IronGauntlets: And should be private?

core/trie.go
```diff
@@ -144,6 +146,21 @@ func NewTrie(storage TrieStorage, height uint) *Trie {
...
+	db, err := db.NewInMemoryDb()
```
IronGauntlets, line 151:
```suggestion
	db, err := db.NewTestDb()
```

core/trie_test.go
```diff
@@ -207,17 +166,21 @@ func TestTriePut(t *testing.T) {
...
+				t.Errorf("TestTriePut: Get() failed at test #%d", idx)
+			}
+			if test.root != nil && !test.root.Equal(trie.root) {
+				t.Errorf("TestTriePut: Unexpected root at test #%d", idx)
+			}
 		}
-	}
+
+		return nil
+	})
```
IronGauntlets, line 183: Can your please explain why the `trie.Get` works when `txn.Commit()` is yet to be called? Is it because the Get calls are being made on the same database Transactions on which Set was called?

core/trie_test.go
```diff
@@ -207,17 +166,21 @@ func TestTriePut(t *testing.T) {
...
-		if err := trie.Put(test.key, test.value); err != nil {
-			t.Errorf("TestTriePut: Put() failed at test #%d", idx)
-		}
-		if value, err := trie.Get(test.key); err != nil || !value.Equal(test.value) {
-			t.Errorf("TestTriePut: Get() failed at test #%d", idx)
-		}
-		if test.root != nil && !test.root.Equal(trie.root) {
-			t.Errorf("TestTriePut: Unexpected root at test #%d", idx)
+	DoOnTempTrie(251, func(trie *Trie) error {
+		for idx, test := range tests {
```
IronGauntlets, line 170: For each test we should create a subtest, to make it easier to identify which cases passed/failed. We can do this in a separate PR.

core/trie.go
omerfirmak: Not really, this is useful for transaction and event commitments as well. We will need temporary tries to build the commitment.

core/trie.go
omerfirmak: The DB we create here is completely temporary and wont persist beyond the scope of this function. So there is no need to go the extra mile and call `Commit()`. Using a helper like `tx.Update()` will automatically `Commit()` and we dont really need that.

`Commit()` is only necessary if you want the changes to be applied to the DB and visible to new transactions created after the `Commit()` call.

core/trie.go
omerfirmak: I actually chose `DoOnTempTrie` over `RunOnTempTrie`. But it was so close, will change:)

>  And should be private?

nope it will be used in commitments calculation.

core/trie.go
omerfirmak: same, DoOnTempTrie is not a test only function.

core/trie_test.go
omerfirmak: A transaction can see the changes it made to the DB without calling `Commit()`. After `Commit()`, only newly created transactions can see those changes.

So;
```

db := newDb

tx1 := db.newTransaction
tx2 := db.newTransaction

tx1.Put(key, value)
assert(value == tx1.Get(key)) // pass

tx2.Get(key) // not found

tx1.Commit()

tx2.Get(key) // not found

tx3 := db.newTransaction
assert(value == tx3.Get(key)) // pass
```

core/trie.go
omerfirmak: >Since discard should only be used for read transactions

It is perfectly normal to call Discard on RW transaction. db.Update() does that as well.https://github.com/dgraph-io/badger/blob/v1.6.2/txn.go#L689

omerfirmak:
@IronGauntlets , updated according to your comments.

core/trietxn.go
```diff
@@ -0,0 +1,68 @@
+package core
+
+import (
+	"github.com/bits-and-blooms/bitset"
+	"github.com/dgraph-io/badger/v3"
+)
+
+// TrieTxn is a database transaction on a trie.
+type TrieTxn struct {
```
IronGauntlets, line 9: Final comment, can we please rename it to `TrieBadgerTxn`

core/trie_test.go
IronGauntlets: Thank you for making that clear!

core/trietxn.go
omerfirmak: Done

IronGauntlets (approved):

jelilat (approved):

Issue #495, opened 2023-01-10
Implement `DataSource` interface using `GatewayClient`


omerfirmak:
Our business logic should use `DataSource`s to query different sources of data and not directly use any clients. We should implement a `Gateway` struct that uses `GatewayClient` as its networking component and implements the `DataSource` interface.

This may require updates to our core types, so feel free to change them when necessary.

PR #501, opened 2023-01-10
Add Gateway data source

omerfirmak:
Towards #495 

## Description

to start writing a state sync loop

## Changes:

- implements `GetStateUpdate` only, rest of the functions will panic at runtime.

## Types of changes

## Testing

**Requires testing**: Yes

### DIFF
--- a/clients/gateway.go
+++ b/clients/gateway.go
@@ -64,12 +64,12 @@ type StateUpdate struct {
 			Value *felt.Felt `json:"value"`
 		} `json:"storage_diffs"`
 
-		Nonces            interface{} `json:"nonces"` // todo: define
+		Nonces            map[string]*felt.Felt `json:"nonces"`
 		DeployedContracts []struct {
 			Address   *felt.Felt `json:"address"`
 			ClassHash *felt.Felt `json:"class_hash"`
 		} `json:"deployed_contracts"`
-		DeclaredContracts interface{} `json:"declared_contracts"` // todo: define
+		DeclaredContracts []*felt.Felt `json:"declared_contracts"`
 	} `json:"state_diff"`
 }
 
--- a/clients/gateway.go
+++ b/clients/gateway.go
@@ -11,13 +11,15 @@ import (
 	"github.com/NethermindEth/juno/core/felt"
 )
 
+const feederGatewayPath = "/feeder_gateway/"
+
 type GatewayClient struct {
 	baseUrl string
 }
 
 func NewGatewayClient(baseUrl string) *GatewayClient {
 	return &GatewayClient{
-		baseUrl: baseUrl,
+		baseUrl: baseUrl + feederGatewayPath,
 	}
 }
 
--- a/clients/gateway_test.go
+++ b/clients/gateway_test.go
@@ -294,7 +294,7 @@ func TestClassUnmarshal(t *testing.T) {
 func TestNewGatewayClient(t *testing.T) {
 	baseUrl := "https://mock_gateway.io"
 	gatewayClient := NewGatewayClient(baseUrl)
-	assert.Equal(t, baseUrl, gatewayClient.baseUrl)
+	assert.Equal(t, baseUrl+feederGatewayPath, gatewayClient.baseUrl)
 }
 
 func TestBuildQueryString(t *testing.T) {
@@ -304,8 +304,9 @@ func TestBuildQueryString(t *testing.T) {
 	args := make(map[string]string)
 	args["a"] = "b"
 	query := gatewayClient.buildQueryString(endpoint, args)
-	assert.Equal(t, baseUrl, gatewayClient.baseUrl)
-	assert.Equal(t, baseUrl+"?a=b", query)
+	feederGatewayUrl := baseUrl + feederGatewayPath
+	assert.Equal(t, feederGatewayUrl, gatewayClient.baseUrl)
+	assert.Equal(t, feederGatewayUrl+"?a=b", query)
 }
 
 func TestBuildQueryString_WithErrorUrl(t *testing.T) {
@@ -350,7 +351,7 @@ func TestGetStateUpdate(t *testing.T) {
 
 	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		switch r.URL.Path {
-		case "/get_state_update":
+		case feederGatewayPath + "get_state_update":
 			{
 				queryMap, err := url.ParseQuery(r.URL.RawQuery)
 				assert.Equal(t, nil, err, "No Query value")
@@ -442,7 +443,7 @@ func TestGetTransaction(t *testing.T) {
 
 	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		switch r.URL.Path {
-		case "/get_transaction":
+		case feederGatewayPath + "get_transaction":
 			{
 				queryMap, err := url.ParseQuery(r.URL.RawQuery)
 				assert.Equal(t, nil, err, "No Query value")
@@ -490,7 +491,7 @@ func TestGetBlock(t *testing.T) {
 
 	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		switch r.URL.Path {
-		case "/get_block":
+		case feederGatewayPath + "get_block":
 			{
 				queryMap, err := url.ParseQuery(r.URL.RawQuery)
 				assert.Equal(t, nil, err, "No Query value")
@@ -538,7 +539,7 @@ func TestGetClassDefinition(t *testing.T) {
 
 	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		switch r.URL.Path {
-		case "/get_class_by_hash":
+		case feederGatewayPath + "get_class_by_hash":
 			{
 				queryMap, err := url.ParseQuery(r.URL.RawQuery)
 				assert.Equal(t, nil, err, "No Query value")
--- a/clients/gateway_test.go
+++ b/clients/gateway_test.go
@@ -26,14 +26,18 @@ func TestStateUpdateUnmarshal(t *testing.T) {
         }
       ]
     },
-    "nonces": {},
+    "nonces": { 
+		"0x37" : "0x44"
+	},
     "deployed_contracts": [
       {
         "address": "0x20cfa74ee3564b4cd5435cdace0f9c4d43b939620e4a0bb5076105df0a626c6",
         "class_hash": "0x10455c752b86932ce552f2b0fe81a880746649b9aee7e0d842bf3f52378f9f8"
       }
 	],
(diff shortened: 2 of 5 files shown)


IronGauntlets:
Can we please rename the `DataSource` interface to `StarkNetData`? Also, Go's convention on [getters and setter](https://softwareengineering.stackexchange.com/questions/278807/does-go-implicitly-discourage-getters-and-setters-and-encourage-direct-member-ac). Can we please remove the `Get` for the function names.

data_source/gateway.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+func NewGateway(n utils.Network) *Gateway {
+	return &Gateway{
+		client: clients.NewGatewayClient(n.URL()),
+	}
+}
+
+func (g *Gateway) GetBlockByNumber(blockNumber uint64) (*core.Block, error) {
+	panic("not implemented")
```
IronGauntlets, line 21:
```suggestion
	fmt.Println("not implemented")
```

data_source/gateway.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+func (g *Gateway) GetTransaction(transactionHash *felt.Felt) (*core.Transaction, error) {
+	panic("not implemented")
```
IronGauntlets, line 25:
```suggestion
	fmt.Println("not implemented")
```

data_source/gateway.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+func (g *Gateway) GetClass(classHash *felt.Felt) (*core.Class, error) {
+	panic("not implemented")
```
IronGauntlets, line 29:
```suggestion
	fmt.Println("not implemented")
```

IronGauntlets (commented):
I would not panic on the unimplemented functions

omerfirmak:
> I would not panic on the unimplemented functions

Let's return `Error("not implemented")`?

IronGauntlets:
> Let's return Error("not implemented")?

Sounds good!

omerfirmak:
@IronGauntlets , done.

added another quick fix as well.

clients/gateway.go
```diff
@@ -17,7 +17,7 @@ type GatewayClient struct {
 
 func NewGatewayClient(baseUrl string) *GatewayClient {
 	return &GatewayClient{
-		baseUrl: baseUrl,
+		baseUrl: baseUrl + "/feeder_gateway/",
```
IronGauntlets, line 20: It would be good to extract it into `const`

omerfirmak: done

IronGauntlets (approved):
LGTM!

omerfirmak:
Rebased

clients/gateway_test.go
```diff
@@ -26,14 +26,18 @@ func TestStateUpdateUnmarshal(t *testing.T) {
...
         "address": "0x20cfa74ee3564b4cd5435cdace0f9c4d43b939620e4a0bb5076105df0a626c6",
         "class_hash": "0x10455c752b86932ce552f2b0fe81a880746649b9aee7e0d842bf3f52378f9f8"
       }
 	],
-    "declared_contracts": []
+    "declared_contracts": [
+		"0x3744"
+	]
   }
 }`)
```
jelilat, line 42: Can we have a reference for where the json data is gotten from? Adding the feeder gateway url, for instance, so that it'll be clear why these values were updated.

clients/gateway_test.go
```diff
@@ -26,14 +26,18 @@ func TestStateUpdateUnmarshal(t *testing.T) {
         }
       ]
     },
-    "nonces": {},
+    "nonces": {
```
jelilat, line 29: The `jsonData` used here and in `TestGetStateUpdate` are the same thing so I think this update should be made there as well. 

I understand that we don't need `nonces` and `declared_contracts` for the test. But I think the data should still be updated to tally with the data source.

jelilat: Alternatively, we can have a global variable for the state update's `jsonData` so we don't have duplicate code.

clients/gateway_test.go
jelilat: I think this should be done for all the `jsonData`.

data_source/gateway.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+func (g *Gateway) GetTransaction(transactionHash *felt.Felt) (*core.Transaction, error) {
+	return nil, errors.New("not implemented")
+}
+
+func (g *Gateway) GetClass(classHash *felt.Felt) (*core.Class, error) {
+	return nil, errors.New("not implemented")
+}
+
```
jelilat, line 33:
```suggestion

// GetStateUpdate gets the state update for a given block number from the feeder gateway,
// then adapts it to the core.StateUpdate type.
```

data_source/gateway.go
```diff
@@ -0,0 +1,84 @@
...
+type Gateway struct {
+	client *clients.GatewayClient
+}
+
+func NewGateway(n utils.Network) *Gateway {
+	return &Gateway{
+		client: clients.NewGatewayClient(n.URL()),
+	}
+}
+
```
jelilat, line 21:
```suggestion

// GetBlockByNumber gets the block for a given block number from the feeder gateway,
// then adapts it to the core.Block type.
```

data_source/gateway.go
```diff
@@ -0,0 +1,84 @@
...
+func (g *Gateway) GetBlockByNumber(blockNumber uint64) (*core.Block, error) {
+	return nil, errors.New("not implemented")
+}
+
```
jelilat, line 25:
```suggestion

// GetTransaction gets the transaction for a given transaction hash from the feeder gateway,
// then adapts it to the appropriate core.Transaction types.
```

data_source/gateway.go
```diff
@@ -0,0 +1,84 @@
...
+func (g *Gateway) GetTransaction(transactionHash *felt.Felt) (*core.Transaction, error) {
+	return nil, errors.New("not implemented")
+}
+
```
jelilat, line 29:
```suggestion

// GetClass gets the class for a given class hash from the feeder gateway,
// then adapts it to the core.Class type.
```

data_source/gateway.go
```diff
@@ -0,0 +1,84 @@
+package dataSource
```
jelilat, line 1: Since this file is an interface between the gateway (and probably P2P) and core types, the file name should be something like `adapter.go`. This is not specific to the feeder gateway and the naming shouldn't have to change when we switch to say P2P in the future.

data_source/gateway.go
```diff
@@ -0,0 +1,84 @@
...
+	return nil, errors.New("not implemented")
+}
+
+func (g *Gateway) GetStateUpdate(blockNumber uint64) (*core.StateUpdate, error) {
+	response, err := g.client.GetStateUpdate(blockNumber)
+	if err != nil {
+		return nil, err
+	}
+
+	return adaptStateUpdate(response)
```
jelilat, line 40: Since what we are getting from `g.client.GetStateUpdate(blockNumber)` is actually the state update, I think we should rename this from `response` to `stateUpdate`. 

If this had been a direct call to the feeder gateway, then the naming `response` would have been appropriate.
```suggestion
	stateUpdate, err := g.client.GetStateUpdate(blockNumber)
	if err != nil {
		return nil, err
	}

	return adaptStateUpdate(stateUpdate)
```

data_source/gateway.go
```diff
@@ -0,0 +1,84 @@
...
+			})
+		}
+	}
+
+	return &core.StateUpdate{
+		BlockHash: response.BlockHash,
+		NewRoot:   response.NewRoot,
+		OldRoot:   response.OldRoot,
+		StateDiff: stateDiff,
+	}, nil
```
jelilat, line 83: Same as above
```suggestion
func adaptStateUpdate(stateUpdate *clients.StateUpdate) (*core.StateUpdate, error) {
	stateDiff := new(core.StateDiff)
	stateDiff.DeclaredContracts = stateUpdate.StateDiff.DeclaredContracts
	for _, deployedContract := range stateUpdate.StateDiff.DeployedContracts {
		stateDiff.DeployedContracts = append(stateDiff.DeployedContracts, core.DeployedContract{
			Address:   deployedContract.Address,
			ClassHash: deployedContract.ClassHash,
		})
	}

	stateDiff.Nonces = make(map[felt.Felt]*felt.Felt)
	for addrStr, nonce := range stateUpdate.StateDiff.Nonces {
		addr, err := new(felt.Felt).SetString(addrStr)
		if err != nil {
			return nil, err
		}
		stateDiff.Nonces[*addr] = nonce
	}

	stateDiff.StorageDiffs = make(map[felt.Felt][]core.StorageDiff)
	for addrStr, diffs := range stateUpdate.StateDiff.StorageDiffs {
		addr, err := new(felt.Felt).SetString(addrStr)
		if err != nil {
			return nil, err
		}

		stateDiff.StorageDiffs[*addr] = []core.StorageDiff{}
		for _, diff := range diffs {
			stateDiff.StorageDiffs[*addr] = append(stateDiff.StorageDiffs[*addr], core.StorageDiff{
				Key:   diff.Key,
				Value: diff.Value,
			})
		}
	}

	return &core.StateUpdate{
		BlockHash: stateUpdate.BlockHash,
		NewRoot:   stateUpdate.NewRoot,
		OldRoot:   stateUpdate.OldRoot,
		StateDiff: stateDiff,
	}, nil
```

data_source/gateway_test.go
```diff
@@ -0,0 +1,103 @@
...
+	"encoding/json"
+	"testing"
+
+	"github.com/NethermindEth/juno/clients"
+	"github.com/NethermindEth/juno/core/felt"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestAdaptStateUpdate(t *testing.T) {
+	jsonData := []byte(`{
```
jelilat, line 13: As mentioned above, we should add a reference to where this data is gotten from

jelilat (changes requested):

data_source/gateway_test.go
omerfirmak: Some are created by hand and rest is either taken from gateway with some manual changes or verbatim. Either way I dont have access to the URLs anymore.

data_source/gateway.go
omerfirmak: IMO, `stateUpdate` would create ambiguity since we have multiple of those in different packages.

data_source/gateway.go
omerfirmak: We cant name everything by their types. In this context what is more important is that it is the response of gateway.

omerfirmak: If Go supported overloading, I would've called the function `adaptResponse`

data_source/gateway.go
omerfirmak: P2P happens over JSON RPC, so that is a different protocol. If `adapt*` family of functions turn out to be reusable, we can extract them in to another package when it comes to that.

jelilat (approved):

PR #505, opened 2023-01-15
Add guard for multiple node unmarshal

liusanchuan:
Fixes unmarshalling a node with multiple left and right children

## Description

Add a check for unmarshalling a node when the input byte array with multiple left or right childs.
[Mentioned in Here](https://github.com/NethermindEth/juno/blob/develop/core/trie/node.go#:~:text=//%20TODO%3A%20Implement%20and,will%20still%20succeed.)

## Changes:

- Add left and right flag to protect there only at most 1 left or right child when marshalling.
- Add unit test for this part

## Types of changes

## Testing

**Requires testing**: Yes

## Documentation

**If this requires a documentation update, did you add one?** No,No

### DIFF
--- a/core/trie/node.go
+++ b/core/trie/node.go
@@ -100,11 +100,7 @@ func (n *Node) UnmarshalBinary(data []byte) error {
 	data = data[felt.Bytes:]
 
 	stream := bytes.NewReader(data)
-	type Slot struct {
-		leftSlot  bool
-		rightSlot bool
-	}
-	slot := Slot{false, false}
+
 	for stream.Len() > 0 {
 		head, err := stream.ReadByte()
 		if err != nil {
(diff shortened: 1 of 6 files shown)


core/trie/node.go
```diff
@@ -109,9 +114,19 @@ func (n *Node) UnmarshalBinary(data []byte) error {
 		var pathP **bitset.BitSet
 		switch head {
 		case 'l':
-			pathP = &(n.left)
+			if !slot.leftSlot {
```
omerfirmak, line 117: you dont need to do any bookkeeping. checking if `n.left` or `n.right`, respectively, is not `nil` should be enough

omerfirmak (changes requested):

core/trie/node.go
liusanchuan: Ohh, your solution is  more clean. Let me update it!

liusanchuan:
@omerfirmak  Sorry, this branh is mismerge by me, I create a new PR at #508 . Can you review again?

PR #514, opened 2023-01-22
```NewTrie``` checks max height 251

### PR DESCRIPTION
**Pull request by MarcusWentz**
## Description

Checks for max new trie height to be 251. 
Prints message requesting the height to be 251 or lower, and returns no new trie.

## Changes:

- Branch if statement to check for max height

## Types of changes

## Testing

**Requires testing**: 

Yes

**Did you write tests??**: 

I tested different uint values for the branch statement.
I can help with formal tests as well if needed.

## Documentation

**If this requires a documentation update, did you add one?** 

```
maxHeight = 251 
```

specification stated here: 

### DIFF
--- a/core/trie/trie.go
+++ b/core/trie/trie.go
@@ -43,7 +43,13 @@ type Trie struct {
 }
 
 func NewTrie(storage Storage, height uint, rootKey *bitset.BitSet) *Trie {
-	// Todo: set max height to 251 and set max key value accordingly
+	if height > 251 {
+		fmt.Println("Height is over 251. Select a height under 251 to create a new trie.")
+		return
+	}
+	
+	// Todo: Set max key value for height 251	
+	
 	return &Trie{
 		storage: storage,
 		height:  height,


joshklop:
@IronGauntlets This is meant to resolve the TODO comment from the trie refactor. Instead of adding a check for the height here, could we make the trie support any height by changing the `*felt.Felt`s to `[]byte`s (in e.g., the [Put](https://github.com/MarcusWentz/juno/blob/028baf9c4859c7ad397891b6d4e4f55a97669053/core/trie/trie.go#L162) function)?

omerfirmak:
> could we make the trie support any height by changing the *felt.Felts to []bytes (in e.g., the [Put](https://github.com/MarcusWentz/juno/blob/028baf9c4859c7ad397891b6d4e4f55a97669053/core/trie/trie.go#L162) function)?

We could but then we would have to check length of the byte array against the trie height, on every Put/Get call.

IronGauntlets:
So there are 2 issues here:
- One to decide on the max height
- Check whether the put/get calls exceed the max key value

The reason we need the first one is that we don't want to have big ints in our trie library, since any height more than 251 bits would require the big ints library and we don't want to constantly switch between felts and big ints.

Using the height we can derive the max key value so that there is no panic if someone tries to insert a key of 5 in a trie height of 2. So we still need to check the length of the key in put/get calls.

omerfirmak:
Closing in favor of #518

PR #515, opened 2023-01-23
GetClass GetTransaction in data_source gateway

chee-chyuan:
Fixes | Closes | Resolves #

## Description

Adding implementation to `GetClass` & `GetTransaction` in `data_source/gateway.go`

Pending programHash implementation here https://github.com/NethermindEth/juno/pull/487

## Changes:

- Change 1
- Change 2

## Types of changes

_Leave only items that describe your changes, remove the rest. Remove this line too._

- New feature (non-breaking change which adds functionality)

## Testing

**Requires testing**: Yes

**Did you write tests??**: Yes

PR #516, opened 2023-01-24
 Implement Transaction Class adaptation layers 

chee-chyuan:
A replicate of this pr https://github.com/NethermindEth/juno/pull/515

## Description
Adding implementation to `GetClass` & `GetTransaction` in `data_source/gateway.go`

Pending programHash implementation here https://github.com/NethermindEth/juno/pull/487

## Changes:

- Change 1
- Change 2

## Types of changes

## Testing

**Requires testing**: Yes


data_source/gateway_test.go
```diff
@@ -101,3 +115,139 @@ func TestAdaptStateUpdate(t *testing.T) {
...
+
+func getMockClass() (*clients.ClassDefinition, *core.Class) {
+	classJson, _ := os.ReadFile("testdata/class.json")
+
+	response := new(clients.ClassDefinition)
+	json.Unmarshal(classJson, response)
+	class, _ := adaptClass(response)
+
+	return response, class
+}
```
IronGauntlets, line 186: I am a little confused by this function. Is this still required if you are using go embed?

chee-chyuan: oh you are right. i totally forgot to remove os.ReadFile. 
we stil need this function because we want `client.ClassDefinition` and `core.Class`

data_source/gateway.go
```diff
@@ -27,14 +27,152 @@ func (g *Gateway) GetBlockByNumber(blockNumber uint64) (*core.Block, error) {
...
+	}
+
+	if deployTx != nil {
+		return deployTx, nil
+	}
+
+	return invokeTx, nil
+}
+
+func adaptTransaction(response *clients.TransactionStatus, g *Gateway) (*core.DeclareTransaction, *core.DeployTransaction, *core.InvokeTransaction, error) {
```
omerfirmak, line 52: Do you really need to return one for all TX types here? Returning a single `core.Transaction` should be enough, AFAICS.

data_source/gateway.go
```diff
@@ -27,14 +27,144 @@ func (g *Gateway) GetBlockByNumber(blockNumber uint64) (*core.Block, error) {
...
+
+	tx, err := adaptTransaction(response, g)
+	if err != nil {
+		return nil, err
+	}
+
+	return tx, nil
+}
+
+func adaptTransaction(response *clients.TransactionStatus, g *Gateway) (core.Transaction, error) {
```
jelilat, line 44: Instead of passing in the transaction status, I think it's better to pass in the transaction instead. 
```suggestion
func adaptTransaction(transaction *clients.Transaction, g *Gateway) (core.Transaction, error) {
```

data_source/gateway.go
```diff
@@ -27,14 +27,144 @@ func (g *Gateway) GetBlockByNumber(blockNumber uint64) (*core.Block, error) {
...
+		if err != nil {
+			return nil, err
+		}
+		declareTx, err := adaptDeclareTransaction(response, class)
+		if err != nil {
+			return nil, err
+		}
+		return declareTx, nil
+	case "DEPLOY":
+		class, err := g.GetClass(response.Transaction.ClassHash)
```
jelilat, line 58: It'll help avoid a lot of repetition as well.

```suggestion
	txType := transaction.Type
	switch txType {
	case "DECLARE":
		class, err := g.GetClass(transaction.ClassHash)
		if err != nil {
			return nil, err
		}
		declareTx, err := adaptDeclareTransaction(transaction, class)
		if err != nil {
			return nil, err
		}
		return declareTx, nil
	case "DEPLOY":
		class, err := g.GetClass(transaction.ClassHash)
```

jelilat (changes requested):
Great work @chee-chyuan 

This looks good. I think it should be ready to merge once we remove the code repetition

jelilat (approved):

omerfirmak (approved):

IronGauntlets (approved):
LGTM, nice work!

If you are going to rebase and merge please clean up the history and make sure the starting word is a verb and the first letter is capitalised. Otherwise, you can squash and merge but please make sure to edit the commit message and body to follow our commit [guidelines](https://cbea.ms/git-commit/).

PR #517, opened 2023-01-24
Prepare Trie and DB for historical state storage

omerfirmak:
## Description

To be able to query historical storage data. This is required to be able to implement `getStorageAt` RPC endpoint.

## Changes:

See each commit message

## Types of changes

## Testing

**Requires testing**: Yes

### DIFF
--- a/db/badger.go
+++ b/db/badger.go
@@ -81,6 +81,26 @@ func (t *badgerTxn) Get(key []byte) ([]byte, error) {
 	})
 }
 
+// Seek : see db.Transaction.Seek
+func (t *badgerTxn) Seek(key []byte) (*Entry, error) {
+	it := t.badger.NewIterator(badger.DefaultIteratorOptions)
+	defer it.Close()
+
+	it.Seek(key)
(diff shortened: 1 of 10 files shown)


db/badger.go
```diff
@@ -81,6 +81,28 @@ func (t *badgerTxn) Get(key []byte) ([]byte, error) {
...
+		item := it.Item()
+		next := &Entry{
+			Key: item.Key(),
+		}
+		item.Value(func(val []byte) error {
+			next.Value = append([]byte{}, val...)
+			return nil
+		})
+
+		return next, nil
```
jelilat, line 100:
```suggestion
		return next, item.Value(func(val []byte) error {
			next.Value = append([]byte{}, val...)
			return nil
		})
```

jelilat (changes requested):

db/badger.go
omerfirmak: Thanks, done!

jelilat (approved):

PR #518, opened 2023-01-24
Trie maxheight fix

anshulforyou:
#514 

## Description
Checks that the trie height is not more than the max height supported.
Checks for the height of trie before putting the value in the trie

## Changes:

- Added condition for trie max height as already done in #514 
- Added check for trie height in the put function of trie

## Types of changes

## Testing

**Requires testing**: Yes

## Documentation

**If this requires a documentation update, did you add one?**

maxHeight = 251 
specification stated here:


core/trie/trie.go
```diff
@@ -154,7 +166,17 @@ func (t *Trie) Get(key *felt.Felt) (*felt.Felt, error) {
...
-	// Todo: check key is not bigger than max key value for a trie height.
+	var x big.Int
+	x.SetInt64(1).Lsh(&x, t.height)
+	fmt.Println(x.String())
+	trieHeightFelt, err := new(felt.Felt).SetInterface(x.String())
+	if err != nil {
+		return err
+	}
+
+	if key.Impl().Cmp(trieHeightFelt.Impl()) == 1 {
```
omerfirmak, line 177: Do not use Impl here, add `Cmp()` to our felt instead.

core/trie/trie.go
```diff
@@ -154,7 +166,17 @@ func (t *Trie) Get(key *felt.Felt) (*felt.Felt, error) {
 
 // Put updates the corresponding `value` for a `key`
 func (t *Trie) Put(key *felt.Felt, value *felt.Felt) error {
-	// Todo: check key is not bigger than max key value for a trie height.
+	var x big.Int
+	x.SetInt64(1).Lsh(&x, t.height)
```
omerfirmak, line 170: You should do this only once, not on every `Put` call. Try moving it to `NewTrie`.

omerfirmak (changes requested):

core/trie/trie_test.go
```diff
@@ -12,6 +12,12 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+func hexToFelt(hex string) *felt.Felt {
```
omerfirmak, line 15: There are multiple of this in our codebase already. Can you move the existing one to a `testutils` package or something and use that everywhere please.

core/trie/trie.go
```diff
@@ -154,7 +173,9 @@ func (t *Trie) Get(key *felt.Felt) (*felt.Felt, error) {
 
 // Put updates the corresponding `value` for a `key`
 func (t *Trie) Put(key *felt.Felt, value *felt.Felt) error {
-	// Todo: check key is not bigger than max key value for a trie height.
+	if key.Cmp(t.maxKeyValue) == 1 {
```
omerfirmak, line 176: should be `!= -1`, no?

core/trie/trie.go
```diff
@@ -2,16 +2,22 @@
...
 import (
+	"errors"
 	"fmt"
+	"math/big"
 	"strings"
 
 	"github.com/NethermindEth/juno/core/crypto"
 	"github.com/NethermindEth/juno/core/felt"
 	"github.com/NethermindEth/juno/db"
+
```
IronGauntlets, line 13:
```suggestion
```

core/trie/trie.go
```diff
@@ -2,16 +2,22 @@
...
 	// Todo: Go.19 introduced math/bits library. Replace bits-and-blooms/bitset with the math/bits.
 	"github.com/bits-and-blooms/bitset"
 )
 
+// Max height of a [Trie] is 251.
+var maxTrieHeight uint = 251
```
IronGauntlets, line 19:
```suggestion
const maxTrieHeight uint = 251
```

core/trie/trie.go
```diff
@@ -37,18 +43,28 @@ type Storage interface {
...
 
-func NewTrie(storage Storage, height uint, rootKey *bitset.BitSet) *Trie {
-	// Todo: set max height to 251 and set max key value accordingly
-	return &Trie{
-		storage: storage,
-		height:  height,
-		rootKey: rootKey,
+func NewTrie(storage Storage, height uint, rootKey *bitset.BitSet) (*Trie, error) {
+	if height > maxTrieHeight {
+		return nil, errors.New("trie height must be less than 251")
```
IronGauntlets, line 54:
```suggestion
		return nil, errors.New("trie height must be less than 252")
```

core/trie/trie.go
```diff
@@ -37,18 +43,28 @@ type Storage interface {
...
+	}
+	var x big.Int
+	x.SetInt64(1).Lsh(&x, height)
+	maxKeyValue, err := new(felt.Felt).SetInterface(&x)
+	if err != nil {
+		return nil, err
```
IronGauntlets, line 60: is there no shift operation on felt?

core/trie/trie.go
```diff
@@ -154,7 +173,9 @@ func (t *Trie) Get(key *felt.Felt) (*felt.Felt, error) {
 
 // Put updates the corresponding `value` for a `key`
 func (t *Trie) Put(key *felt.Felt, value *felt.Felt) error {
-	// Todo: check key is not bigger than max key value for a trie height.
+	if key.Cmp(t.maxKeyValue) != -1 {
+		return errors.New("key is bigger than the trie height")
+	}
```
IronGauntlets, line 178: Shouldn't this check be added to `Get` as well? 

Also, the error message doesn't represent the error and is misleading.

```suggestion
	if key.Cmp(t.maxKeyValue) != -1 {
		return errors.New("key is bigger than the maxKeyValue")
	}
```

utils/testutils.go
```diff
@@ -0,0 +1,9 @@
+package utils
+
+import "github.com/NethermindEth/juno/core/felt"
+
+func HexToFelt(hex string) *felt.Felt {
+	// We know our test hex values are valid, so we'll ignore the potential error
+	f, _ := new(felt.Felt).SetString(hex)
+	return f
+}
```
IronGauntlets, line 9: I am not sure that this should be in the utils package. It will encourage people to just use this over creating a new felt and will not be limited to tests.  This can lead to us just ignoring errors. I think it is ok to have a private function in each package where required as a private function. You can consider declaring the function as a variable in the test function. That way no one in the rest of the package will have access to that but the test function where it is needed.

This will open doors to ignoring errors.

IronGauntlets (changes requested):

core/trie/trie.go
anshulforyou: Nope, not able to find any.

core/trie/trie.go
anshulforyou: In this scenario, the Get function would return "Key not found" so I am not sure if the check is required.

utils/testutils.go
anshulforyou: Done, removed the public function HexToFelt

anshulforyou:
Please close this PR for #545

PR #521, opened 2023-01-25
Make state operate on a DB transaction

omerfirmak:
## Description

This will allow us to orchestrate any database operation related to a new block height in a single TX. So that it is trivial to revert an update attempt when we encounter a problem.

## Types of changes

## Testing

**Requires testing**: No

**Did you write tests??**: No


omerfirmak:
similar to how `Contract` works right now

IronGauntlets (approved):

PR #525, opened 2023-01-26
Implement Block Sync

IronGauntlets:
## Description

This PR implements the mechanism for fetching blocks and state updates and storing them locally. The blockchain struct does the verification, storing of blocks and application of state diff.

## Changes:

See commit history

## Types of changes

## Testing

**Requires testing**: No


blockchain/blockchain.go
```diff
@@ -0,0 +1,198 @@
...
+	database db.DB
+}
+
+func NewBlockchain(database db.DB, network utils.Network) (*Blockchain, error) {
+	chain := &Blockchain{
+		l:        &sync.RWMutex{},
+		database: database,
+		network:  network,
+	}
+	txn := chain.database.NewTransaction(false)
```
omerfirmak, line 76: missing `txn.Discard()`. We can just use db.View() so that we dont have to deal with txns explicitly.

blockchain/blockchain.go
```diff
@@ -0,0 +1,198 @@
...
+	headBlock := new(core.Block)
+	if err = gob.NewDecoder(bytes.NewReader(headBlockBin)).Decode(headBlock); err != nil {
+		return nil, err
+	}
+	chain.head = headBlock
+	return chain, nil
+}
+
+// Height returns the latest block height. If blockchain is empty nil is returned.
+func (b *Blockchain) Height() *uint64 {
```
omerfirmak, line 91: we are allocating a uint64 on the heap for no added benefit AFAICS. I think it is better to just return the value

blockchain/blockchain.go
```diff
@@ -0,0 +1,198 @@
...
+			- When Block is changed to include a list of Transaction and TransactionReceipts
+			- Further checks would need to be added to ensure Transaction Hash has been computed
+				properly.
+			- Sanity check would need to include checks which ensure there is same number of
+				Transactions and TransactionReceipts.
+	*/
+	var head *core.Block
+	b.l.RLock()
+	head = b.head
+	b.l.RUnlock()
```
omerfirmak, line 149: This is open to races. Our local reference to `HEAD` stays the same, but `b.head` could change anytime while we are verifying a block. If it does, we just verified a block for the previous `HEAD`, which then we will try to append on top of the new `HEAD`.

blockchain/blockchain.go
```diff
@@ -0,0 +1,198 @@
...
+		return txn.Set(bKey, blockBinary)
+	}); err != nil {
+		return err
+	}
+
+	b.l.Lock()
+	b.head = block
+	b.l.Unlock()
+	return nil
+}
```
omerfirmak, line 135: `Store()` as a whole should be atomic and reentrant. 

It is not atomic because, there are multiple points that we could error out of this function when we have already altered some state and we don't rollback any of those changes. For example, if gob.Encode() fails, we have already called the state.Update() and modified the DB. We should be creating and using only a single DB txn thru out the entire `Store()` call.

It is not reentrant because, for example, we could verify a block but before we can apply the state or store the block, another call to `Store()` could change the `HEAD`.

blockchain/blockchain.go
```diff
@@ -0,0 +1,198 @@
...
+	k.Number = binary.BigEndian.Uint64(data[1:9])
+	k.Hash = new(felt.Felt).SetBytes(data[9:41])
+	return nil
+}
+
+// Blockchain is responsible for keeping track of all things related to the StarkNet blockchain
+type Blockchain struct {
+	l *sync.RWMutex
+
+	head     *core.Block
```
omerfirmak, line 65: For simplicity, this should be in the database. So that we can easily prove consistency of our database by using DB transactions. If not, our DB and memory could be out of sync.

core/felt/felt.go
```diff
@@ -9,12 +9,12 @@ import (
 )
 
 type Felt struct {
-	val fp.Element
+	Val fp.Element
```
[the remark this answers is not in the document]
omerfirmak: And `cbor` [looked](https://github.com/NethermindEth/juno/pull/482) like a better alternative to me due to the much smaller size of the data it creates, are we sure we want to use `gob`?

node/node.go
```diff
@@ -60,23 +60,28 @@ func New(cfg *Config) (StarkNetNode, error) {
...
+	n.db, err = db.NewDb(n.cfg.DatabasePath)
+	if err != nil {
+		return err
+	}
+	defer n.db.Close()
+	n.blockchain, err = blockchain.NewBlockchain(n.db, n.cfg.Network)
+	if err != nil {
+		return err
+	}
+	n.synchronizer = sync.NewSynchronizer(n.blockchain, gateway.NewGateway(n.cfg.Network))
```
omerfirmak, line 79: IMO, all instantiations should happen in `New` where we create the `Node`

blockchain/blockchain.go
```diff
@@ -0,0 +1,198 @@
...
+	}
+
+	k.Number = binary.BigEndian.Uint64(data[1:9])
+	k.Hash = new(felt.Felt).SetBytes(data[9:41])
+	return nil
+}
+
+// Blockchain is responsible for keeping track of all things related to the StarkNet blockchain
+type Blockchain struct {
+	l *sync.RWMutex
```
omerfirmak, line 63: I think we should rely on our DB to provide the synchronization and not have any explicit mutexes.

DB Transactions are safe to use concurrently. So if we leverage that fact, all our concurrency issues will go away in an instant.

omerfirmak (changes requested):

jelilat (changes requested):

blockchain/blockchain.go
```diff
@@ -0,0 +1,198 @@
...
+
+type ErrIncompatibleBlock struct {
+	reason string
+}
+
+func (e ErrIncompatibleBlock) Error() string {
+	return fmt.Sprintf("incompatible block: %v", e.reason)
+}
+
+// BlockDbKey creates appends block number and hash to create a db key.
```
jelilat, line 34:
```suggestion
// BlockDbKey appends hash to block number to create a db key.
```

jelilat (changes requested):

core/felt/felt.go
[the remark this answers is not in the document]
IronGauntlets: This is removed and now we use cbor for encodiing

blockchain/blockchain.go
IronGauntlets: I have added it. I didn't use `db.View()` because get was already doing the copy so I thought I might as well use that instead of doing a copy explicitly in `db.View()`

blockchain/blockchain.go
IronGauntlets: AFAIK escape analysis determines whether allocation will happen on the heap or stack. Since the variable will be referenced from outside the function I believe both the pointer and the value will be allocated on the heap. Also, the reason for this is to determine whether the blockchain is empty or not. If I were to return the value the caller will not be able to tell the difference between an empty blockchain or a blockchain with a current height of 0. 

Also, It is idiomatic in go to use a pointer to differentiate between zero and nil values.

blockchain/blockchain.go
IronGauntlets: Yeah, you are right. I was supposed to create a deep copy but I forgot. I will use the DB to synchronise instead of having mutexes

blockchain/blockchain.go

blockchain/blockchain.go
IronGauntlets: it already is: `HeadBlock`. I will just remove this reference.

node/node.go
IronGauntlets: I did have all the instantiation in the New function but creating a db opens the db and we can't close it in the New function. This meant some of the tests were failing when we only wanted to check whether the config was set correctly or not.

I moved the db part to the Run function but then I had a mixture of instantiation across New and Run. So I decided that New will set the config while Run will handle all the instantiation and start the node.

blockchain/blockchain.go
IronGauntlets: Sure I will remove the mutex

core/transaction.go
```diff
@@ -57,14 +61,16 @@ type TransactionReceipt struct {
...
 	ExecutionResources *ExecutionResources
 	L1ToL2Message      *L1ToL2Message
-	L2ToL1Message      *[]L2ToL1Message
+	L2ToL1Message      []*L2ToL1Message
 	Signatures         []*felt.Felt
 	TransactionHash    *felt.Felt
 	TransactionIndex   *big.Int
-	Type               TransactionType
+	// Todo: Do we need to have this?
+	Type TransactionType
```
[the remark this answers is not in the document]
IronGauntlets: I think it is a question of whether we want to have an explicit type for Transactions or we want to have a single Transaction struct with a type field. We certainly don't want both. My personal preference is the former and it is out of the scope of this pull request to decide. I just added as a reminder.

starknetdata/gateway/gateway.go
```diff
@@ -0,0 +1,223 @@
...
+	}
+
+	return &core.TransactionReceipt{
+		ActualFee:          response.ActualFee,
+		TransactionHash:    response.TransactionHash,
+		TransactionIndex:   response.TransactionIndex,
+		Events:             events,
+		ExecutionResources: adaptExecutionResources(response.ExecutionResources),
+		L1ToL2Message:      adaptL1ToL2Message(response.L1ToL2Message),
+		L2ToL1Message:      l2ToL1Messages,
```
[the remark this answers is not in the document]
IronGauntlets: Good spot! I think this needs to be discussed in conjunction with the above comment and it is out of the scope of this pr. If it were to be fixed here it would require too many changes.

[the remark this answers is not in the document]
jelilat: I think we should implement the adapter in a separate PR. That way, we can also update the Block to include transaction and transaction receipt. 

We also won't need to add the transaction type to transaction receipts

blockchain/blockchain.go
IronGauntlets: Due to the removal of the mutex. This piece of code has changed

blockchain/blockchain.go

blockchain/blockchain.go

blockchain/blockchain.go

blockchain/blockchain.go
omerfirmak: a `uint64` fits nicely in a CPU register. So it will be returned as a value rather than a reference. But if using `nil` makes sense, we can keep it as it is.

blockchain/blockchain.go
```diff
@@ -59,48 +58,48 @@ func (k *BlockDbKey) UnmarshalBinary(data []byte) error {
...
 		network:  network,
 	}
-	txn := chain.database.NewTransaction(false)
+}
+
+// Height returns the latest block height. If blockchain is empty nil is returned.
+func (b *Blockchain) Height() *uint64 {
+	headBlock, err := b.Head()
+	if err != nil && !errors.Is(err, db.ErrKeyNotFound) {
+		panic(fmt.Sprintf("failed to retrieved head block: %v", err))
```
omerfirmak, line 76: was there a specific reason you didnt want to return an error instead of panicing here?

omerfirmak (approved):

jelilat (approved):

PR #526, opened 2023-01-27
Feat/add felt tests

0xkeivin:
## Description
Additional test cases ensure functionality of the package methods and operations without being too dependent on tests cases upstream.

## Changes:

- Added tests cases for felt.go package

## Types of changes

- Other (please describe):  Test Cases

## Testing

**Requires testing**: Yes

## Documentation

### DIFF
--- a/core/felt/felt_test.go
+++ b/core/felt/felt_test.go
@@ -1,7 +1,7 @@
 package felt
 
 import (
-	"fmt"
+	"reflect"
 	"testing"
 
 	"github.com/consensys/gnark-crypto/ecc/stark-curve/fp"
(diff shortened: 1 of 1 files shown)


IronGauntlets:
@0xkeivin Thank you very much for creating these tests. However, we have wrapped `fp.Element` in a `Felt` struct and since the implementation is trivial there is no need for more tests (we are assuming the `fp.Element` package is well tested). The only non-trivial part of Felt that we modified already has a test. Therefore, I would recommend that you create a pr on [`gnark-crypto`](https://github.com/consensys/gnark-crypto)'s repo if you feel there is a lack of tests.

PR #527, opened 2023-01-27
Implement Fetcher for concurrent gateway requests

omerfirmak:
## Description

Fetcher spawns multiple go routines and prefetches block&state update for a later height. A consumer can call GetNext to get the prefetched values in order.

Gateway is quite slow when queried in sequence, so this will speed up our syncing process.

## Changes:

- Implement fetcher

## Types of changes

## Testing

**Requires testing**: Yes


IronGauntlets:
I don't think the fetching is slow but rather computing transaction commitment and event commitment in adapt block is causing it it be slow. I don't think we should introduce this until we can process blocks and state updates faster than fetching. I am not completely sure about this btw.

omerfirmak:
> I don't think the fetching is slow but rather computing transaction commitment and event commitment in adapt block is causing it it be slow. I don't think we should introduce this until we can process blocks and state updates faster than fetching. I am not completely sure about this btw.

When I measured gateway latency last time, it was pretty considerable compared to what state update took. This sped up the process significantly.

omerfirmak:
@IronGauntlets moved this to `datasource`. it is source agnostic,  so it will work for both P2P and gateway. Will rebase after #525 is merged

omerfirmak:
Closing in favor of #556

PR #531, opened 2023-01-30
Explicitly marshal Felt to CBOR

omerfirmak:
## Description

Since `val` is private, we should explicitly marshal that.

## Changes:

- add MarshalCBOR, UnmarshalCBOR to Felt type

## Types of changes

## Testing

**Requires testing**: Yes

### DIFF
--- a/core/felt/felt.go
+++ b/core/felt/felt.go
@@ -5,6 +5,8 @@ import (
 	"math/big"
 	"sync"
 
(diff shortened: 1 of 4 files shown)


omerfirmak:
@IronGauntlets , apparently there was a v2 of CBOR package and it does not need that `EncOptions` when marshaling. Updated to V2.

IronGauntlets (approved):

PR #533, opened 2023-01-31
Fail  Path suffix not match

liusanchuan:
Bug report #

### DIFF
--- a/core/trie/trie_test.go
+++ b/core/trie/trie_test.go
@@ -139,20 +139,25 @@ func TestPath(t *testing.T) {
 		child  *bitset.BitSet
 		want   *bitset.BitSet
 	}{
+		// {
+		// 	parent: bitset.New(0),
+		// 	child:  bitset.New(251).Set(250).Set(249),
+		// 	want:   bitset.New(250).Set(249),
+		// },
+		// {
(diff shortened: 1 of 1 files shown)


omerfirmak:
You are not building your bitsets right. They should be;

```
{ 
  parent: bitset.New(4).Set(0).Set(2),
  child: bitset.New(8).Set(2).Set(3).Set(4).Set(6), 
  want: bitset.New(3).Set(2), 
}
```

//0b 0101

right most bit here is called LSB, and is at index 0
left most bit here is called MSB, and is at index 3

liusanchuan:
Hi @omerfirmak , Is the bit setting index is from tail to head? 
 `bitset.New(8).Set(2).Set(3).Set(4).Set(6), // is 0x01011100`
And If I want get the first index in path, I need to check path.Test(path.Len()-1), am I right?

omerfirmak:
path.Test(path.Len()-1) tests MSB
path.Test(0) tests LSB

>  bitset.New(8).Set(2).Set(3).Set(4).Set(6), // is 0x01011100

correct

liusanchuan:
Ohh, I got it, thanks a lot, 👍

PR #534, opened 2023-01-31
Implement Transaction and Receipt Storage

jelilat:
## Description

Allows the storage of the block number and index for a given transaction hash. It'll come in handy when implementing the `getTransactionByHash` RPC endpoint.

## Types of changes

## Testing

**Requires testing**: Yes


blockchain/blockchain.go
```diff
@@ -253,13 +253,17 @@ func (b *Blockchain) VerifyTransaction(transaction core.Transaction) (*felt.Felt
...
 	case *core.DeployTransaction:
-		hash = v.Hash
+		return v.Hash, nil
 	case *core.DeployAccountTransaction:
 		hash = v.Hash
 	case *core.DeclareTransaction:
 		hash = v.Hash
 	case *core.InvokeTransaction:
-		hash = v.Hash
+		if v.Version == new(felt.Felt) {
```
omerfirmak, line 262: use `felt.Zero.Equal(v.Version)`

blockchain/blockchain.go
```diff
@@ -208,3 +208,134 @@ func getBlockByHash(txn db.Transaction, hash *felt.Felt) (*core.Block, error) {
...
+// GetTransactionByHash gets the transaction for a given hash.
+func (b *Blockchain) GetTransactionByHash(hash *felt.Felt) (transaction core.Transaction, err error) {
+	return transaction, b.database.View(func(txn db.Transaction) error {
+		transaction, err = getTransactionByHash(txn, hash)
+		return err
+	})
+}
+
+// StoreTransaction takes a transaction and verifies its hash before storing it in the database.
+func (b *Blockchain) StoreTransaction(blockNumber, txnIndex uint64, transaction core.Transaction) error {
```
omerfirmak, line 230: `StoreTransaction` and `StoreReceipt` should not be public. They should be folded in to `StoreBlock`

omerfirmak (changes requested):

IronGauntlets:
We should merge this with #560. What do you guys think?

blockchain/blockchain.go
jelilat: What about tests?

IronGauntlets: There should be a public function called GetTransaction/GetTransactionReceipt. So after storing you call that to see that you get the expected output

jelilat: [Here](https://github.com/NethermindEth/juno/pull/534/files#diff-37b6e7ed4f299b87b357b54376123393b34ee0699fc035f25bc0621627dcc6baR221-R235)

omerfirmak:
Since we store transactions separately, putBlock should store headers only now, instead of the entire block.

jelilat:
> Since we store transactions separately, putBlock should store headers only now, instead of the entire block.

Was thinking we wanted to add that in a separate PR

blockchain/blockchain.go
```diff
@@ -90,8 +98,8 @@ func (b *Blockchain) GetBlockByHash(hash *felt.Felt) (block *core.Block, err err
 	})
 }
 
-// Store takes a block and state update and performs sanity checks before putting in the database.
-func (b *Blockchain) Store(block *core.Block, stateUpdate *core.StateUpdate) error {
+// StoreBlock takes a block and state update and performs sanity checks before putting in the database.
+func (b *Blockchain) StoreBlock(block *core.Block, stateUpdate *core.StateUpdate) error {
```
omerfirmak, line 102: I dont think we need to call this StoreBlock anymore right? There are no store functions anymore.

jelilat: We may still need to Store Class during `StateUpdate` though. But yeah, calling it `StoreBlock` is not needed at the moment

blockchain/blockchain_test.go
```diff
@@ -83,13 +83,13 @@ func TestGetBlockByNumberAndHash(t *testing.T) {
 		require.NoError(t, err)
 		require.NoError(t, putBlock(txn, block))
 
-		storedByNumber, err := getBlockByNumber(txn, block.Number)
+		storedBlockByNumber, err := getBlockByNumber(txn, block.Number)
```
omerfirmak, line 86: can we revert these as well? so we dont introduce unrelated changes.

blockchain/blockchain.go
```diff
@@ -102,6 +110,14 @@ func (b *Blockchain) Store(block *core.Block, stateUpdate *core.StateUpdate) err
 		if err := putBlock(txn, block); err != nil {
 			return err
 		}
+		for i, tx := range block.Transactions {
+			if err := storeTransaction(txn, block.Number, uint64(i), tx); err != nil {
+				return err
+			}
+			if err := storeReceipt(txn, block.Number, uint64(i), *block.Receipts[i]); err != nil {
```
omerfirmak, line 117: do we need to de-reference the pointer here?

blockchain/blockchain.go
```diff
@@ -212,3 +228,120 @@ func (b *Blockchain) SanityCheckNewHeight(block *core.Block, stateUpdate *core.S
...
+// getReceiptByHash gets the transaction receipt for a given hash.
+func getReceiptByHash(txn db.Transaction, hash *felt.Felt) (*core.TransactionReceipt, error) {
+	if bnIndexBytes, err := getBlockNumberAndIndex(txn, hash); err != nil {
+		return nil, err
+	} else {
+		return getReceipt(txn, bnIndexBytes)
+	}
+}
+
+func joinBlockNumberAndIndex(blockNumber, index uint64) []byte {
```
omerfirmak, line 341: let's make this a type like;

type TxKey struct {
   blockNum uint64
   index uint64
}

and use Marshal & Unmarshal, so that we can easily recover blockNum and indexes when iterating over keys just by unmarshaling a TxKey

IronGauntlets:
I am going to take over this branch and make the relevant changes requested by @omerfirmak. Please have a look #560 in the meantime. Once I have cleaned up the history and addressed the changes I will add these commit/s on top of #560.

PR #537, opened 2023-02-01
Update Transaction Chain Id

jelilat:
## Description

Our transaction hash implementation uses hard-coded chain ids. I updated that to use our standard `utils.Network` type.

## Changes:

- Updated transaction chain id

## Types of changes

## Testing

**Requires testing**: Yes

**Did you write tests??**: No. Updated existing tests.

### DIFF
--- a/core/transaction.go
+++ b/core/transaction.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/NethermindEth/juno/core/crypto"
 	"github.com/NethermindEth/juno/core/felt"
+	"github.com/NethermindEth/juno/utils"
 	"github.com/ethereum/go-ethereum/common"
 )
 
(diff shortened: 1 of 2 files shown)


core/transaction.go
```diff
@@ -71,7 +72,7 @@ type TransactionReceipt struct {
 
 type Transaction interface {
 	// Todo: Add Hash as a field to all the Transaction Objects
-	Hash() *felt.Felt
+	Hash(*utils.Network) *felt.Felt
```
omerfirmak, line 75: Network is just a `uint8` use the value instead of a pointer please.

omerfirmak (changes requested):

core/transaction.go
jelilat: Done

omerfirmak (approved):

PR #540, opened 2023-02-01
Introduce Transaction type constraint

IronGauntlets:
## Description

Introduce Transaction type constraints to represent different types of Transactions for our core package. Unfortunately, we cannot pass type parameters to interface functions or struct methods. This is why the `StarkNetData` interface's Transaction function's return type needed to be changed from `*core.Transaction` -> `any`.

## Types of changes

## Testing

**Requires testing**: No

### DIFF
--- a/core/transaction.go
+++ b/core/transaction.go
@@ -71,8 +71,7 @@ type TransactionReceipt struct {
 }
 
 type Transaction interface {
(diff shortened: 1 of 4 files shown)


IronGauntlets:
closing in favour of #541

PR #541, opened 2023-02-02
Configure cbor to allow unmarshaling specified types to interfaces they implement

omerfirmak:
For example;

``` golang

type A struct {
	B string
	C uint
}

func TestA(t *testing.T) {
	inst := A{"hello", 42}

//	encoder.RegisterType(reflect.TypeOf(inst))

	data, _ := encoder.Marshal(inst)
	var recovered any
	encoder.Unmarshal(data, &recovered)

	switch v := recovered.(type) {
	case A:
		t.Error("success:", v)
	default:
		panic(v)
	}
}
```

if `RegisterType` is not called, test will panic at the type switch. If type `A` is registered, we will be able to cast `recovered` to `A` in runtime.

This helps us to have a `[]Transaction` in our block body and allows us to cast them to our individual transaction types in runtime.

### DIFF
--- a/core/transaction_test.go
+++ b/core/transaction_test.go
@@ -6,7 +6,9 @@ import (
 	"testing"
 
 	"github.com/NethermindEth/juno/core/felt"
+	"github.com/NethermindEth/juno/encoder"
 	"github.com/NethermindEth/juno/utils"
+	"github.com/stretchr/testify/assert"
 )
 
 var (
@@ -52,6 +54,8 @@ func TestDeployTransactions(t *testing.T) {
 			if !transactionHash.Equal(test.want) {
 				t.Errorf("wrong hash: got %s, want %s", transactionHash.Text(16), test.want.Text(16))
 			}
+
+			checkTransactionSymmetry(t, &test.input)
 		})
 	}
 }
@@ -125,6 +129,8 @@ func TestInvokeTransactions(t *testing.T) {
 			if !transactionHash.Equal(test.want) {
 				t.Errorf("wrong hash: got %s, want %s", transactionHash.Text(16), test.want.Text(16))
 			}
+
+			checkTransactionSymmetry(t, &test.input)
 		})
 	}
 }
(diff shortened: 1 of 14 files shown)


encoder/encoder.go
```diff
@@ -0,0 +1,70 @@
...
+
+func RegisterType(rType reflect.Type) error {
+	err := ts.Add(
+		cbor.TagOptions{EncTag: cbor.EncTagRequired, DecTag: cbor.DecTagRequired},
+		rType,
+		tagNum,
+	)
+	if err != nil {
+		return err
+	}
```
jelilat, line 41:
```suggestion
func RegisterType(rType reflect.Type) error {
	if err := ts.Add(
		cbor.TagOptions{EncTag: cbor.EncTagRequired, DecTag: cbor.DecTagRequired},
		rType,
		tagNum,
	); err != nil {
		return err
	}
```

core/transaction_test.go
```diff
@@ -251,6 +257,27 @@ func TestDeclareTransaction(t *testing.T) {
 			if !transactionHash.Equal(test.want) {
 				t.Errorf("wrong hash: got %s, want %s", transactionHash.Text(16), test.want.Text(16))
 			}
+
+			CheckTransactionSymmetry(t, test.input)
 		})
 	}
 }
+
+func CheckTransactionSymmetry(t *testing.T, in Transaction) {
```
jelilat, line 266: I think we should make the parameter name more descriptive. Maybe `input` or `value`? 
```suggestion
func CheckTransactionSymmetry(t *testing.T, input Transaction) {
```

jelilat (changes requested):
Looks

core/trie/node_test.go
```diff
@@ -5,20 +5,14 @@ import (
 	"errors"
 	"testing"
 
+	"github.com/NethermindEth/juno/encoder"
+
```
joshklop, line 9: Spurious diff?
```suggestion
```

joshklop (changes requested):

core/transaction_test.go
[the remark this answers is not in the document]
omerfirmak: done

encoder/encoder.go
omerfirmak: done

core/trie/node_test.go
omerfirmak: done

core/transaction_test.go
omerfirmak: Done

omerfirmak:
When requesting cosmetic or nitpick changes, please approve the PR. So that we don't have to wait for a second round of reviews to merge a PR.

jelilat (approved):

joshklop (approved):

PR #542, opened 2023-02-02
Add retry mechanism to gateway client

joshklop:
## Description

We want to retry a few times in the gateway client in case of spurious errors.

## Changes:

See commits.

## Types of changes

## Testing

**Requires testing**: Yes

## Documentation

**If this requires a documentation update, did you add one?** No, No

### DIFF
--- a/clients/gateway.go
+++ b/clients/gateway.go
@@ -12,15 +12,13 @@ import (
 	"github.com/NethermindEth/juno/core/felt"
 )
 
-const feederGatewayPath = "/feeder_gateway/"
-
 type GatewayClient struct {
 	baseUrl string
 }
 
 func NewGatewayClient(baseUrl string) *GatewayClient {
 	return &GatewayClient{
-		baseUrl: baseUrl + feederGatewayPath,
+		baseUrl: baseUrl,
 	}
 }
 
--- a/clients/gateway.go
+++ b/clients/gateway.go
@@ -8,23 +8,57 @@ import (
 	"net/http"
 	"net/url"
 	"strconv"
+	"time"
 
 	"github.com/NethermindEth/juno/core/felt"
 )
 
+type Backoff func(wait time.Duration) time.Duration
+
 type GatewayClient struct {
-	baseUrl string
+	url        string
+	backoff    Backoff
+	maxRetries int
+	minWait    time.Duration
 }
 
-func NewGatewayClient(baseUrl string) *GatewayClient {
+func (c *GatewayClient) WithBackoff(b Backoff) *GatewayClient {
+	c.backoff = b
+	return c
+}
+
+func (c *GatewayClient) WithMaxRetries(num int) *GatewayClient {
+	c.maxRetries = num
+	return c
+}
+
+func (c *GatewayClient) WithMinWait(d time.Duration) *GatewayClient {
+	c.minWait = d
+	return c
+}
+
+// https://cloud.google.com/iot/docs/how-tos/exponential-backoff
+func ExponentialBackoff(wait time.Duration) time.Duration {
+	time.Sleep(wait)
+	return wait * 2
+}
+
+func NopBackoff(d time.Duration) time.Duration {
+	return 0
+}
+
+func NewGatewayClient(url string) *GatewayClient {
 	return &GatewayClient{
-		baseUrl: baseUrl,
+		url:        url,
+		backoff:    ExponentialBackoff,
+		maxRetries: 11, // ~34 minutes with ExponentialBackoff (block time on mainnet is 20-30 minutes)
+		minWait:    1 * time.Second,
 	}
 }
 
-// `buildQueryString` builds the query url with encoded parameters
+// buildQueryString builds the query url with encoded parameters
 func (c *GatewayClient) buildQueryString(endpoint string, args map[string]string) string {
-	base, err := url.Parse(c.baseUrl)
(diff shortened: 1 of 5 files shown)


utils/network.go
```diff
@@ -29,13 +29,13 @@ func (n Network) String() string {
...
+		return "https://alpha4.starknet.io/feeder_gateway/"
 	case MAINNET:
-		return "https://alpha-mainnet.starknet.io"
+		return "https://alpha-mainnet.starknet.io/feeder_gateway/"
 	case GOERLI2:
-		return "https://alpha4-2.starknet.io"
+		return "https://alpha4-2.starknet.io/feeder_gateway/"
 	case INTEGRATION:
-		return "https://external.integration.starknet.io"
+		return "https://external.integration.starknet.io/feeder_gateway/"
```
omerfirmak, line 38: I considered doing this as well, but I was not sure if there were other endpoints open to public

joshklop: Hmmm, that's I good point that I did not consider. I am not aware of any. If that is the case, we could make another function `GatewayUrl` that returns `network.Url() + "/feeder_gateway/"`.

joshklop:
Made a few tiny adjustments (three lines). This is ready for review.

clients/gateway.go
```diff
@@ -8,23 +8,64 @@ import (
...
+		c.maxWait = d
+	}
+}
+
+func WithMinWait(d time.Duration) GatewayOption {
+	return func(c *GatewayClient) {
+		c.minWait = d
+	}
+}
+
```
omerfirmak, line 44: how about this approach? https://github.com/dgraph-io/badger/blob/83298b771573e8778eb98131913a55e6f30eb11c/options.go#L372

so, something like

```
func (c *Gatewayclient) WithBackoff(b Backoff) *GatewayClient
func (c *Gatewayclient) WithMaxWait(d time.Duration) *GatewayClient

NewGatewayClient().WithBackoff().WithMaxWait()
```

clients/gateway.go
```diff
@@ -8,23 +8,64 @@ import (
...
 }
 
-func NewGatewayClient(baseUrl string) *GatewayClient {
-	return &GatewayClient{
-		baseUrl: baseUrl,
+func NewGatewayClient(url string, opts ...GatewayOption) *GatewayClient {
+	c := &GatewayClient{
+		url:     url,
+		backoff: ExponentialBackoff,
+		maxWait: 30 * time.Minute,
```
omerfirmak, line 55: 30 mins is too long, 1 min is reasonable IMO

clients/gateway.go
```diff
@@ -42,7 +83,21 @@ func (c *GatewayClient) buildQueryString(endpoint string, args map[string]string
 
 // get performs a "GET" http request with the given URL and returns the response body
 func (c *GatewayClient) get(queryUrl string) ([]byte, error) {
-	res, err := http.Get(queryUrl)
+	var res *http.Response
+	var err error
+	wait := c.minWait
+	for {
```
omerfirmak, line 89: Instead of breaking on `maxWait`, I think we should put a limit on number of retries

joshklop: Do you mean replacing `maxWait` with something like `maxRetries`? Just want to make sure I'm understanding.

clients/gateway.go
joshklop: That's much simpler, thanks!

joshklop:
Rebased on develop. Needed to refactor a few things to work nicely with `testsource`.

clients/gateway.go
joshklop: @IronGauntlets That's what I was doing before, but I think I prefer Omer's solution for simplicity. The only potential drawback I can see is that users can change options after creating the gateway, but I don't think that's an issue here.

omerfirmak (approved):

clients/gateway.go
```diff
@@ -44,7 +76,16 @@ func (c *GatewayClient) buildQueryString(endpoint string, args map[string]string
 
 // get performs a "GET" http request with the given URL and returns the response body
 func (c *GatewayClient) get(queryUrl string) ([]byte, error) {
```
omerfirmak, line 78: We probably need `get` to be cancelable as well, if not we might have to wait 34 mins just to get back the control of execution when we make a request. Can you see if we can use https://pkg.go.dev/net/http#NewRequestWithContext ?

clients/gateway.go
```diff
@@ -8,25 +8,57 @@ import (
...
 }
 
-func NewGatewayClient(baseUrl string) *GatewayClient {
+func (c *GatewayClient) WithBackoff(b Backoff) *GatewayClient {
+	c.backoff = b
+	return c
+}
+
+func (c *GatewayClient) WithMaxRetries(d int) *GatewayClient {
+	c.maxRetries = d
```
jelilat, line 31: Trivial. But can we make this descriptive? Say, `num` as in number of retries?

Also maybe `uint` since we can't have negative retries
```suggestion
func (c *GatewayClient) WithMaxRetries(num uint) *GatewayClient {
	c.maxRetries = num
```

jelilat (changes requested):

jelilat (approved):

clients/gateway.go
joshklop: Good point. Since will require modifying a lot of code, it will be handled in a separate PR.

PR #545, opened 2023-02-06
trie-maxheight-fix

anshulforyou:
https://github.com/NethermindEth/juno/pull/514

Description
Checks that the trie height is not more than the max height supported.
Checks for the height of trie before putting the value in the trie

Changes:
Added condition for trie max height as already done in https://github.com/NethermindEth/juno/pull/514
Added check for trie height in the put function of trie
Types of changes
Bugfix (non-breaking change which fixes an issue)
New feature (non-breaking change which adds functionality)
Testing
Requires testing: Yes

Did you write tests??: Yes

Documentation
If this requires a documentation update, did you add one?

maxHeight = 251
specification stated here:


clients/programhash_test.go
```diff
@@ -20,6 +19,11 @@ var (
 )
 
 func TestProgramHash(t *testing.T) {
+	hexToFelt := func(hex string) *felt.Felt {
+		f, _ := new(felt.Felt).SetString(hex)
+		return f
+	}
```
IronGauntlets, line 25:
```suggestion
	hexToFelt := func(hex string) *felt.Felt {
		f, err := new(felt.Felt).SetString(hex)
		require.NoError(t, err)
		return f
	}
```

IronGauntlets (changes requested):
Please change the target branch to `develop`.

Also, please change split the PR into multiple commits, so that the `hexToFelt` are not mixed with Trie changes.

IronGauntlets:
Closing this pull request for now. Feel free to re-open when the above changes have been made.

PR #547, opened 2023-02-06
Add logger interface

joshklop:
Fixes | Closes | Resolves #

## Description

Add logger interface so we can control logging in tests and so anyone importing juno as a library can provide their own logger.

## Changes:

- Add logger interface
- Use zap logger instead of `log` package 

## Types of changes

## Testing

**Requires testing**: Yes

## Documentation

**If this requires a documentation update, did you add one?** Yes, Yes


joshklop:
Badger logs now only print on ERROR and will use the logger passed from the node (keeps formatting consistent). I'll stop making changes until this is reviewed.

node/node.go
```diff
@@ -61,10 +65,19 @@ func New(cfg *Config) (StarknetNode, error) {
...
-	log.Println("Running Juno with config: ", fmt.Sprintf("%+v", *n.cfg))
-
 	var err error
-	n.db, err = db.NewDb(n.cfg.DatabasePath)
+	n.log, err = utils.NewZapLogger(n.cfg.Verbosity)
+	if err != nil {
+		return err
+	}
+
+	fmt.Println("Running Juno with config: ", fmt.Sprintf("%+v", *n.cfg))
```
omerfirmak, line 74: why are we not using the logger here?

utils/log.go
```diff
@@ -0,0 +1,83 @@
...
+	Infow(msg string, keysAndValues ...any)
+	Warnw(msg string, keysAndValues ...any)
+	Errorw(msg string, keysAndValues ...any)
+}
+
+type ZapLogger struct {
+	*zap.SugaredLogger
+}
+
+var _ Logger = (*ZapLogger)(nil)
```
IronGauntlets, line 57: Can you please explain what is happening here?

utils/log.go
```diff
@@ -0,0 +1,83 @@
...
+type Logger interface {
+	SimpleLogger
+	badger.Logger
+}
+
+type SimpleLogger interface {
+	Debugw(msg string, keysAndValues ...any)
+	Infow(msg string, keysAndValues ...any)
+	Warnw(msg string, keysAndValues ...any)
+	Errorw(msg string, keysAndValues ...any)
```
IronGauntlets, line 50: Are these the functions implemented by `zap.SugaredLogger`? Is that why they all end with `w`?

db/badger.go
```diff
@@ -111,8 +111,8 @@ func (t *badgerTxn) Impl() any {
 }
 
 // NewDb opens a new database at the given path
-func NewDb(path string) (DB, error) {
-	opt := badger.DefaultOptions(path)
+func NewDb(path string, log badger.Logger) (DB, error) {
```
IronGauntlets, line 114: Why are we not passing our `utils.Logger`?

utils/log.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+func (l LogLevel) IsValid() bool {
+	return l.String() != ""
+}
+
+type Logger interface {
+	SimpleLogger
+	badger.Logger
+}
```
IronGauntlets, line 44: Is this only being used to satisfy the badger's logger?

utils/log.go
joshklop: Ensures `ZapLogger` fulfills the `Logger` interface. I've seen it in a lot of go codebases, and the Uber style guide [recommends it](https://github.com/uber-go/guide/blob/master/style.md#verify-interface-compliance). Also makes maintenance a tad easier since gopls will complain until the interface is fulfilled.

utils/log.go
joshklop: Yes, it implements those methods. I think the `w` stands for "with" (the message is printed *with* the keys and values).

db/badger.go
joshklop: Because `utils.Logger` includes methods that badger does not require.

node/node.go
joshklop: Ah, forgot that one. Thanks! Will fix.

IronGauntlets (approved):

omerfirmak:
needs rebase

omerfirmak (approved):

PR #548, opened 2023-02-06
BlockStorage refactor

IronGauntlets:
Refactor of block storage and blockchain test. This was easier to implement than to comment.

## Changes:

See commits.

## Types of changes

- Build related changes

## Testing

**Requires testing**: No


omerfirmak:
I found storage stuff to be in a single file easier to maintain and reason about. We could just make BlockStorage private as well. But this works too. I will just remove the direct reference to `cbor` and merge.

PR #549, opened 2023-02-07
Update Pedersen Hash

jelilat:
## Description

Since Consensys has now added the Pedersen hash implementation, this PR updates our Pedersen hash implementation to use it.

## Types of changes

## Testing

**Requires testing**: Yes

**Did you write tests??**: Used existing tests

### DIFF
--- a/core/crypto/pedersen_hash.go
+++ b/core/crypto/pedersen_hash.go
@@ -1,59 +1,21 @@
 package crypto
 
 import (
-	"math/big"
-
 	"github.com/NethermindEth/juno/core/felt"
-	starkcurve "github.com/consensys/gnark-crypto/ecc/stark-curve"
 	"github.com/consensys/gnark-crypto/ecc/stark-curve/fp"
+	pedersenhash "github.com/consensys/gnark-crypto/ecc/stark-curve/pedersen-hash"
 )
-
-var (
-	shiftPoint *starkcurve.G1Jac
-	p0         *starkcurve.G1Jac
-	p1         *starkcurve.G1Jac
-	p2         *starkcurve.G1Jac
-	p3         *starkcurve.G1Jac
-)
-
-func init() {
-	// The curve points come from the [reference implementation].
-	//
-	// [reference implementation]: https://github.com/starkware-libs/cairo-lang/blob/de741b92657f245a50caab99cfaef093152fd8be/src/starkware/crypto/signature/fast_pedersen_hash.py
-
-	x := new(fp.Element)
-	y := new(fp.Element)
-
-	x.SetString("2089986280348253421170679821480865132823066470938446095505822317253594081284")
-	y.SetString("1713931329540660377023406109199410414810705867260802078187082345529207694986")
-	shiftPoint = new(starkcurve.G1Jac).FromAffine(&starkcurve.G1Affine{X: *x, Y: *y})
-
-	x.SetString("996781205833008774514500082376783249102396023663454813447423147977397232763")
-	y.SetString("1668503676786377725805489344771023921079126552019160156920634619255970485781")
-	p0 = new(starkcurve.G1Jac).FromAffine(&starkcurve.G1Affine{X: *x, Y: *y})
-
-	x.SetString("2251563274489750535117886426533222435294046428347329203627021249169616184184")
-	y.SetString("1798716007562728905295480679789526322175868328062420237419143593021674992973")
-	p1 = new(starkcurve.G1Jac).FromAffine(&starkcurve.G1Affine{X: *x, Y: *y})
-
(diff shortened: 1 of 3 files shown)


core/crypto/pedersen_hash.go
```diff
@@ -61,25 +21,6 @@ func PedersenArray(elems ...*felt.Felt) *felt.Felt {
...
-
-	highPart := bigInt.SetUint64(uint64(aBytes[0])) // The top nibble (bits 249-252)
-	lowPart := aBytes[1:]                           // Zero-out the top nibble (bits 249-252)
-
-	m := new(starkcurve.G1Jac).ScalarMultiplication(p2, highPart)
-	var n starkcurve.G1Jac
-	n.ScalarMultiplication(p1, bigInt.SetBytes(lowPart))
-	return m.AddAssign(&n)
+	hash := pedersenhash.Pedersen(&a.Val, &b.Val)
+	return felt.NewFelt(hash)
```
jelilat, line 25: I'm worried that these extra conversions may slow down our hash calculation a bit. Thoughts?

core/felt/felt.go
```diff
@@ -11,12 +11,12 @@ import (
 )
 
 type Felt struct {
-	val fp.Element
+	Val fp.Element
```
omerfirmak, line 14: Use `Impl()` instead of making this field public

core/crypto/pedersen_hash.go
omerfirmak: I doubt it would. Can you try benchmarking it?

core/felt/felt.go
jelilat: Thanks

omerfirmak:
We should use `PedersenArray` from their package as well.

jelilat:
> We should use `PedersenArray` from their package as well.

I didn't do that because of my reservations about the implication of the conversions. Or do you think it's insignificant?

omerfirmak:
> > We should use `PedersenArray` from their package as well.
> 
> I didn't do that because of my reservations about the implication of the conversions. Or do you think it's insignificant?

We do that conversion already since `gnark-crypto` functions work on `fp.Element` type. We are simply doing it a bit earlier in this case.

jelilat:
> > > We should use `PedersenArray` from their package as well.
> > 
> > 
> > I didn't do that because of my reservations about the implication of the conversions. Or do you think it's insignificant?
> 
> We do that conversion already since `gnark-crypto` functions work on `fp.Element` type. We would simply be doing it a bit earlier in this case.

True that. So, there shouldn't be any difference in the hash calculation time as well.

core/crypto/pedersen_hash.go
```diff
@@ -1,85 +1,28 @@
...
 // PedersenArray implements [Pedersen array hashing].
 //
 // [Pedersen array hashing]: https://docs.starknet.io/documentation/develop/Hashing/hash-functions/#array_hashing
 func PedersenArray(elems ...*felt.Felt) *felt.Felt {
-	d := new(felt.Felt)
-	for _, e := range elems {
-		d = Pedersen(d, e)
+	var fpElements []*fp.Element
+	for _, elem := range elems {
+		fpElements = append(fpElements, elem.Impl())
```
omerfirmak, line 15: Instead of appending and causing multiple allocations, use `make` to allocate all the storage needed at once.

omerfirmak: so

``` golang
fpElements := make([]*fp.Element, len(elems))
```

core/crypto/pedersen_hash.go
```diff
@@ -1,85 +1,28 @@
 package crypto
 
 import (
-	"math/big"
-
 	"github.com/NethermindEth/juno/core/felt"
-	starkcurve "github.com/consensys/gnark-crypto/ecc/stark-curve"
 	"github.com/consensys/gnark-crypto/ecc/stark-curve/fp"
+	pedersenhash "github.com/consensys/gnark-crypto/ecc/stark-curve/pedersen-hash"
```
omerfirmak, line 6: how about `gnark` instead of `pedersenhash`?

jelilat: I think `gnark` would have been appropriate if we were importing `github.com/consensys/gnark-crypto`. Imo, `fp` could as well be `gnark` since they are both subsets of `gnark`. 

Maybe we could do `hash`? Then, say `hash.Pedersen()` and `hash.PedersenArray()`. That feels better to me and it eliminates the  repetition of the word `pedersen` which I sense may be unsettling.

omerfirmak: Lets just keep it as is

omerfirmak (approved):

joshklop (approved):

PR #551, opened 2023-02-08
Block hash verification with no sequencer address & remove commitments from core.Block

omerfirmak:
## Description

Apparently some post 0.7.0 blocks with no sequencer address in the header don't use the fallback address, instead they use zero as the sequencer address.

## Changes:

- Try with zero address, then with configured fallback address
- remove commitments from core.Block, add the block body instead 

## Types of changes

## Testing

**Requires testing**: Yes

## Documentation


jelilat (changes requested):

core/block.go
```diff
@@ -94,53 +84,97 @@ func getBlockHashMetaInfo(network utils.Network) *blockHashMetaInfo {
...
 		// Check if the block number is in the unverifiable range
 		if b.Number >= unverifiableRange[0] && b.Number <= unverifiableRange[1] {
-			// If so, return unverifiable block error
-			return nil, &ErrUnverifiableBlock{blockNumber: b.Number}
+			// If so, return success
+			return nil
+		}
+	}
+
+	if b.Number < metaInfo.First07Block {
```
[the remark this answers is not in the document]
omerfirmak: I initially thought it was not possible, but yeah it was doable. See the last commit.

core/block.go
```diff
@@ -94,53 +84,93 @@ func getBlockHashMetaInfo(network utils.Network) *blockHashMetaInfo {
...
+		} else if hash.Equal(b.Hash) {
+			return nil
+		}
+	}
+	return ErrCantVerifyBlockHash
+}
+
+// BlockHash computes the block hash
+func BlockHash(b *Block, network utils.Network) (*felt.Felt, error) {
+	return blockHash(b, network, nil)
```
jelilat, line 118: By automatically passing `overrideSeqAddr` as `nil`, that means that our `BlockHash` function could generate a wrong hash. It won't be able to calculate the correct hash for blocks with `zero` sequencer address. 

Imo, we should remove it altogether. My thoughts are:

1. I don't think we'll need to generate the Block hash anywhere else besides `VerifyBlockHash`
2. We probably shouldn't keep something that we know could produce a wrong value. Especially because we may not be able to detect if the value is wrong outside `VerifyBlockHash`

jelilat (changes requested):

core/block.go
omerfirmak: We are not generating any hashes at this point, all we are doing is verifying them. `BlockHash` is just there for tests essentially. So you wouldnt call BlockHash on a block you fetched from a peer/gateway. You would only call `VerifyBlockHash` on such a block.

> It won't be able to calculate the correct hash for blocks with zero sequencer address

BlockHash assumes all fields are populated, so no `nil` values. With a `nil` field, `BlockHash` will panic. If sequencer address is `felt.Zero` what `BlockHash` generates is correct for that block. It is the caller's job to guarantee that the block passed to `BlockHash` is valid.

omerfirmak: In fact, BlockHash shouldn't accept a Block object. Because it is reasonable to assume a Block to already have it's Hash field populated. It should probably require you to pass the fields separately

jelilat: Got it. Thanks for the explanation.

I think we should leave it as it is though. It might be a bot messy passing the arguments separately.

IronGauntlets (approved):

PR #553, opened 2023-02-08
Change TansactionCommitment input from []*TransactionReceipt to []Transaction 

IronGauntlets:
Transaction commitment is calculated using transactions instead of transaction receipts.

## Changes:

See commits

## Types of changes

- Documentation Update
- 
## Testing

**Requires testing**: No

### DIFF
--- a/core/transaction.go
+++ b/core/transaction.go
@@ -254,11 +254,13 @@ func (d *DeployAccountTransaction) Hash(network utils.Network) (*felt.Felt, erro
 	panic("not implemented")
 }
 
+const commitmentTrieHeight uint = 64
+
 // TransactionCommitment is the root of a height 64 binary Merkle Patricia tree of the
 // transaction hashes and signatures in a block.
 func TransactionCommitment(receipts []*TransactionReceipt) (*felt.Felt, error) {
 	var transactionCommitment *felt.Felt
-	return transactionCommitment, trie.RunOnTempTrie(64, func(trie *trie.Trie) error {
+	return transactionCommitment, trie.RunOnTempTrie(commitmentTrieHeight, func(trie *trie.Trie) error {
 		for i, receipt := range receipts {
 			signaturesHash := crypto.PedersenArray()
 			if receipt.Type == Invoke {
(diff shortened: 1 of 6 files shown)


core/transaction.go
```diff
@@ -220,36 +279,33 @@ func declareTransactionHash(d *DeclareTransaction, network utils.Network) (*felt
...
+func TransactionCommitment(transactions []Transaction) (*felt.Felt, error) {
 	var transactionCommitment *felt.Felt
 	return transactionCommitment, trie.RunOnTempTrie(commitmentTrieHeight, func(trie *trie.Trie) error {
-		for i, receipt := range receipts {
-			signaturesHash := crypto.PedersenArray()
-			if receipt.Type == Invoke {
-				signaturesHash = crypto.PedersenArray(receipt.Signatures...)
+		for i, transaction := range transactions {
+			signatureHash := crypto.PedersenArray()
+			if _, ok := transaction.(*InvokeTransaction); ok {
```
omerfirmak, line 303: this if is not needed anymore, right? signature() will return an empty array for non-invoke txs.

omerfirmak (approved):

core/transaction.go
jelilat: Some other transaction types actually have a Signature (e.g. Declare)

jelilat (approved):

core/transaction.go
omerfirmak: then if must go

core/transaction.go
```diff
@@ -224,18 +279,69 @@ func (d *DeclareTransaction) Hash(network utils.Network) (*felt.Felt, error) {
...
+const commitmentTrieHeight uint = 64
+
+// TransactionCommitment is the root of a height 64 binary Merkle Patricia tree of the
+// transaction hashes and signatures in a block.
+func TransactionCommitment(transactions []Transaction) (*felt.Felt, error) {
+	var transactionCommitment *felt.Felt
+	return transactionCommitment, trie.RunOnTempTrie(commitmentTrieHeight, func(trie *trie.Trie) error {
+		for i, transaction := range transactions {
+			signatureHash := crypto.PedersenArray()
+			if _, ok := transaction.(*InvokeTransaction); ok {
```
omerfirmak, line 303: if should be removed

omerfirmak (changes requested):

core/transaction.go
jelilat: Not really. Although there are other transactions with Signatures, we only use that of `InvokeTransaction` to calculate commitments

omerfirmak: my bad

omerfirmak (approved):

PR #555, opened 2023-02-09
Make clients.GatewayClient.get cancellable

joshklop:
## Description

See [here](https://github.com/NethermindEth/juno/pull/542#discussion_r1101229443).

## Changes:

See commits.

## Types of changes

- Documentation Update

## Testing

**Requires testing**: Yes

**Did you write tests??**: Existing tests are sufficient

## Documentation

**If this requires a documentation update, did you add one?** No, no

### DIFF
--- a/clients/gateway.go
+++ b/clients/gateway.go
@@ -17,6 +17,7 @@ type Backoff func(wait time.Duration) time.Duration
 
 type GatewayClient struct {
 	url        string
+	client     *http.Client
 	backoff    Backoff
 	maxRetries int
 	maxWait    time.Duration
@@ -55,6 +56,7 @@ func NopBackoff(d time.Duration) time.Duration {
 func NewGatewayClient(url string) *GatewayClient {
 	return &GatewayClient{
 		url:        url,
+		client:     http.DefaultClient,
 		backoff:    ExponentialBackoff,
 		maxRetries: 35, // ~35 minutes with default backoff and maxWait (block time on mainnet is 20-30 minutes)
 		maxWait:    time.Minute,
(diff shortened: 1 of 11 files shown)


clients/gateway.go
```diff
@@ -75,15 +78,24 @@ func (c *GatewayClient) buildQueryString(endpoint string, args map[string]string
...
 // get performs a "GET" http request with the given URL and returns the response body
-func (c *GatewayClient) get(queryUrl string) ([]byte, error) {
+func (c *GatewayClient) get(ctx context.Context, queryUrl string) ([]byte, error) {
 	var res *http.Response
 	var err error
+	var body io.Reader
 	wait := c.minWait
 	for i := 0; i <= c.maxRetries; i++ {
-		res, err = http.Get(queryUrl)
+		req, err := http.NewRequestWithContext(ctx, "GET", queryUrl, body)
```
omerfirmak, line 87: we are passing `nil` as the body here, do we need the variable?

node/node.go
```diff
@@ -14,7 +15,7 @@ import (
 )
 
 type StarknetNode interface {
-	Run() error
+	Run(ctx context.Context) error
```
omerfirmak, line 18: We have an explicit `Shutdown()`, so we don't need a `ctx` here. Alternative is getting rid of `Shutdown()`. Same goes for Synchronizer

omerfirmak: I think getting rid of Shutdown is the better option here btw

joshklop: That's a great point! Didn't think of that.

omerfirmak (approved):

IronGauntlets:
I am not an of using context because of the following reasons:

- In most cases, context is used for its quit and done functions. Which are nothing more than closing and reading from a quit channel.
- Due to the context pattern you cannot store it thus you have to pass it to every function in the call stack which may not even call it.
- Context also has a data field which may be used to pass data however all type of safety is lost.

For our use case, it is best to keep the context isolated to the client's package. And use shutdown/close functions where they are required.

omerfirmak:
I thought using context throughout the entire codebase made sense because it allows us to shutdown the entire node with a single context. It looked easier to get it right in the first try compared to per-struct Start/Shutdown functions and quit channels. I am down to anything that would get PRs merged tho :)

omerfirmak (approved):

IronGauntlets (approved):
Approving with the assumption that `ctx` will be removed in #556

PR #556, opened 2023-02-09
 Parallelize sync process 

omerfirmak:
## Description

speeds up syncing

## Changes:

- see commits

## Types of changes

## Testing

**Requires testing**: No

**Did you write tests??**: No


omerfirmak:
Tests hang for a known reason, I will fix them tomorrow

omerfirmak:
Ready for reviews.

blockchain/blockchain.go
```diff
@@ -109,15 +109,13 @@ func (b *Blockchain) Store(block *core.Block, stateUpdate *core.StateUpdate) err
 	})
 }
 
-func (b *Blockchain) VerifyBlock(block *core.Block, stateUpdate *core.StateUpdate) error {
+func (b *Blockchain) IsNextHead(block *core.Block) error {
```
IronGauntlets, line 112: I think the previous name still makes sense since `verifyBlock` doesn't imply that we are going to store it.

blockchain/blockchain.go
```diff
@@ -109,15 +109,13 @@ func (b *Blockchain) Store(block *core.Block, stateUpdate *core.StateUpdate) err
...
 	txn := b.database.NewTransaction(false)
 	defer txn.Discard()
-	return b.verifyBlock(txn, block, stateUpdate)
+	return b.isNextHead(txn, block)
```
IronGauntlets, line 115: We should use `database.View ()` similar to other functions which were recently updated.

blockchain/blockchain_test.go
```diff
@@ -159,7 +159,7 @@ func TestVerifyBlock(t *testing.T) {
 	t.Run("error when block hash does not match state update's block hash", func(t *testing.T) {
 		stateUpdate := &core.StateUpdate{BlockHash: h1}
 		expectedErr := ErrIncompatibleBlockAndStateUpdate{"block hashes do not match"}
-		assert.Equal(t, chain.VerifyBlock(mainnetBlock1, stateUpdate), expectedErr)
+		assert.Equal(t, chain.SanityCheckNewHeight(mainnetBlock1, stateUpdate), expectedErr)
```
IronGauntlets, line 162: Wherever we are using `SanityCheckNewHeight()` for testing it should be separated into a separate set of tests called: `TestSanityCheckNewHeight()`.

sync/sync.go
```diff
@@ -1,81 +1,103 @@
...
-// Shutdown attempts to stop the Synchronizer, should block until loop acknowledges the request
-func (s *Synchronizer) Shutdown() error {
-	if stopped := atomic.CompareAndSwapUint64(&s.running, 1, 0); !stopped {
-		return errors.New("synchronizer is stopped")
+func (s *Synchronizer) SyncBlocks(ctx context.Context) error {
+	fetchers := stream.New().WithMaxGoroutines(runtime.NumCPU())
+	verifiers := stream.New().WithMaxGoroutines(runtime.NumCPU())
+
+	verifierTask := func(block *core.Block, stateUpdate *core.StateUpdate) stream.Callback {
+		err := s.Blockchain.SanityCheckNewHeight(block, stateUpdate)
```
IronGauntlets, line 37: Any particular reason this line is not been called in the `default` case in the `Callback` func?

sync/sync.go
```diff
@@ -1,81 +1,103 @@
...
+	return s.SyncBlocks(ctx)
 }
 
-// Shutdown attempts to stop the Synchronizer, should block until loop acknowledges the request
-func (s *Synchronizer) Shutdown() error {
-	if stopped := atomic.CompareAndSwapUint64(&s.running, 1, 0); !stopped {
-		return errors.New("synchronizer is stopped")
+func (s *Synchronizer) SyncBlocks(ctx context.Context) error {
+	fetchers := stream.New().WithMaxGoroutines(runtime.NumCPU())
+	verifiers := stream.New().WithMaxGoroutines(runtime.NumCPU())
```
IronGauntlets, line 34: What will happen if we were to submit max goroutines higher than the number of CPUs?

sync/sync.go
omerfirmak: callbacks are called from the same goroutine, so if it was called from the callback we wouldnt get any concurrency in verification.

sync/sync.go
omerfirmak: Not much, we would overwhelm our CPU and some of the goroutines would have to wait to be scheduled on a thread. So sweet spot is having exactly NumCPU number of computation heavy goroutines.

omerfirmak: Not much, we would overwhelm our CPU and some of the goroutines would have to wait to be scheduled on a thread. So sweet spot is having exactly NumCPU number of computation heavy goroutines.

blockchain/blockchain.go
omerfirmak: I wanted emphasize that it assumes the block to be sanity checked beforehand. Any suggestions that would have that meaning?

sync/sync.go
IronGauntlets: Shouldn't be `runtime.NumCPU()/2`? Given we have 2 streams?

sync/sync.go
```diff
@@ -1,81 +1,103 @@
...
-	return s.SyncBlocks()
+func (s *Synchronizer) Run(ctx context.Context) error {
+	return s.SyncBlocks(ctx)
 }
 
-// Shutdown attempts to stop the Synchronizer, should block until loop acknowledges the request
-func (s *Synchronizer) Shutdown() error {
-	if stopped := atomic.CompareAndSwapUint64(&s.running, 1, 0); !stopped {
-		return errors.New("synchronizer is stopped")
+func (s *Synchronizer) SyncBlocks(ctx context.Context) error {
```
IronGauntlets, line 32: `SyncBlocks` will never return a non-nil error now. I am not sure if that is something we want.

sync/sync.go
omerfirmak: fetchers are not computation heavy routines. they are mostly blocked, waiting for network.

sync/sync.go
```diff
@@ -1,81 +1,103 @@
...
+			default:
+				if err != nil {
+					log.Printf("Sanity checks failed on Block #%d, Hash: %s", block.Number, block.Hash.Text(16))
+					return
+				}
+				err = s.Blockchain.Store(block, stateUpdate)
+				if err != nil {
+					log.Printf("Failed storing Block #%d, Hash: %s %s", block.Number, block.Hash.Text(16), err.Error())
+					return
+				}
```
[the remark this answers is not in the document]
omerfirmak: see https://github.com/NethermindEth/juno/pull/556/commits/9b32fc5fb64fa985f106af44be90e3b6506a6ab2

blockchain/blockchain.go
IronGauntlets: I cannot think of a better name than `VerifyBlock`. Usually `Is...()` return bool and we are not doing that here. I think keeping it as verify block with a comment to say what it does should be sufficient.

omerfirmak: got it, reverting

sync/sync.go
```diff
@@ -2,80 +2,125 @@ package sync
...
-			// Clean up and return
-			return nil
+		case err := <-errChan:
+			streamCancel() // cancel all running tasks
+			streamCtx, streamCancel = context.WithCancel(syncCtx)
+			nextHeight = err.Height // keep syncing from failed height
+		case <-syncCtx.Done():
+			fetchers.Wait()
+			verifiers.Wait()
+			return syncCtx.Err()
```
joshklop, line 116: Should we be returning `nil` here? Right now, the node prints an error message when I `<Ctrl-c>`.

omerfirmak: Maybe we should check for “context canceled” error and return nil in that case. But I am not sure if that is a special type.

omerfirmak: done

cmd/juno/juno.go
```diff
@@ -108,23 +109,22 @@ func NewCmd(newNodeFn node.NewStarknetNodeFn, quit <-chan os.Signal) *cobra.Comm
 			return err
 		}
 
+		ctx := context.Background()
+		nodeCtx, cancel := context.WithCancel(ctx)
```
IronGauntlets, line 113:
```suggestion
		nodeCtx, cancel := context.WithCancel(context.Background())
```

blockchain/blockchain.go
```diff
@@ -109,15 +109,15 @@ func (b *Blockchain) Store(block *core.Block, stateUpdate *core.StateUpdate) err
 	})
 }
 
-func (b *Blockchain) VerifyBlock(block *core.Block, stateUpdate *core.StateUpdate) error {
-	txn := b.database.NewTransaction(false)
-	defer txn.Discard()
-	return b.verifyBlock(txn, block, stateUpdate)
+// VerifyBlock assumes block to be sanity checked first
```
IronGauntlets, line 112:
```suggestion
// VerifyBlock assumes the block has already been sanity-checked.
```

blockchain/blockchain.go
```diff
@@ -109,15 +109,15 @@ func (b *Blockchain) Store(block *core.Block, stateUpdate *core.StateUpdate) err
...
+func (b *Blockchain) VerifyBlock(block *core.Block) error {
+	return b.database.View(func(txn db.Transaction) error {
+		return b.verifyBlock(txn, block)
+	})
 }
 
-func (b *Blockchain) verifyBlock(txn db.Transaction, block *core.Block,
-	stateUpdate *core.StateUpdate,
-) error {
+// verifyBlock assumes block to be sanity checked first
```
IronGauntlets, line 119:
```suggestion
```

sync/sync_test.go
```diff
@@ -41,7 +42,12 @@ func TestSyncBlocks(t *testing.T) {
 		testDB := db.NewTestDb()
 		bc := blockchain.NewBlockchain(testDB, utils.MAINNET)
 		synchronizer := NewSynchronizer(bc, gw, log)
-		assert.Error(t, synchronizer.SyncBlocks())
+		ctx, cancel := context.WithCancel(context.Background())
+		go func() {
+			time.Sleep(time.Second)
+			cancel()
+		}()
+		assert.EqualError(t, synchronizer.Run(ctx), "context canceled")
```
IronGauntlets, line 50: We never return context cancelled error, so how does this make sense? Also, we should use `errors.Is()/As()` for error comparisons

IronGauntlets:
The commit history is a bit superfluous and some commits can be fixed up as follow 
```
- Use contexts to manage node lifecycle
- Extract block&state update sanity checks to a function
- Use db.View instead of explicit txns
- Parallelize sync process
  |_ Handle errors in sync loop
  |_ Rename IsNexthead to VerifyBlock
  |_ Don't propagate context.Cancelled errors in sync
  |_ Make FetchetTask and verifierTask methods

"|_" is fixup 
```

sync/sync.go
```diff
@@ -3,79 +3,126 @@ package sync
...
+	if h, err := s.Blockchain.Height(); err == nil {
+		nextHeight = h + 1
+	}
+
+	for {
+		select {
+		case err := <-errChan:
+			streamCancel() // cancel all running tasks
+			streamCtx, streamCancel = context.WithCancel(syncCtx)
+			nextHeight = err.Height // keep syncing from failed height
```
IronGauntlets, line 110: We should log here that we are restarting sync

sync/sync_test.go
omerfirmak: Yeap, it doesnt make sense:) fixed

blockchain/blockchain_test.go
omerfirmak: done

omerfirmak:
All done and rebased

IronGauntlets (approved):
Nice work!

sync/sync.go
```diff
@@ -3,79 +3,127 @@ package sync
...
+			s.log.Warnw("Rolling back sync process to failed height", "height", err.Height)
```
jelilat, line 111: Quick question regarding this. Do we at any point stop restarting due to too many errors? Or we just keep restarting until the user stops the sync?

jelilat (approved):
I don't have any comments. Great work!

Just dropped a question in there.

sync/sync.go
IronGauntlets: We never stop, that is supposed to be the expected behaviour #568. However, if there are any panics in the goroutines they only are propagated up when the user sends the signal to stop. Now, this may or may not be a desired behaviour. Also, in general, panics are not detected by the parent go routine.

omerfirmak: the latter

PR #557, opened 2023-02-10
Switch database backend to pebble

omerfirmak:


db/badger.go
```diff
@@ -68,41 +68,38 @@ func (t *badgerTxn) Delete(key []byte) error {
...
-		return nil, err
+		return err
 	}
 
-	var value []byte
-	return value, item.Value(func(val []byte) error {
-		value = append([]byte{}, val...)
-		return nil
+	return item.Value(func(val []byte) error {
+		return cb(val)
```
IronGauntlets, line 101: Maybe I am missing something but in the docs of badger they say you must copy. From [docs (Using key/value pairs)](https://dgraph.io/docs/badger/get-started/#using-key-value-pairs) 

```go
err := db.View(func(txn *badger.Txn) error {
  item, err := txn.Get([]byte("answer"))
  handle(err)

  var valNot, valCopy []byte
  err := item.Value(func(val []byte) error {
    // This func with val would only be called if item.Value encounters no error.

    // Accessing val here is valid.
    fmt.Printf("The answer is: %s\n", val)

    // Copying or parsing val is valid.
    valCopy = append([]byte{}, val...)

    // Assigning val slice to another variable is NOT OK.
    valNot = val // Do not do this.
    return nil
  })
  handle(err)

  // DO NOT access val here. It is the most common cause of bugs.
  fmt.Printf("NEVER do this. %s\n", valNot)

  // You must copy it to use it outside item.Value(...).
  fmt.Printf("The answer is: %s\n", valCopy)

  // Alternatively, you could also use item.ValueCopy().
  valCopy, err = item.ValueCopy(nil)
  handle(err)
  fmt.Printf("The answer is: %s\n", valCopy)

  return nil
})

```

omerfirmak:
```
  // You must copy it to use it outside item.Value(...).
```

we are not using it outside the scope of item.Value

IronGauntlets: Also, since we are going to use pebble. It might be a good idea to drop this commit?

We should be wary of having callbacks calling callbacks. 

And since pebble has reduced our memory usage so much maybe the low-hanging fruit doesn't need to be addressed right now.

omerfirmak:
```
    // Copying or parsing val is valid.
```

we are simply delegating parsing work to `cb`

omerfirmak: > We should be wary of having callbacks calling callbacks.

there is nothing wrong with that. We should keep it, there is no down side.

> much maybe the low-hanging fruit doesn't need to be addressed right now.

it is already done tho

IronGauntlets: > there is nothing wrong with that. We should keep it, there is no down side.

[callback hell](http://callbackhell.com/) is a thing which is why javascript introduced promises. Anyway, for us it is fine.

> we are simply delegating parsing work to `cb`

In our case, we are using `cb` mostly for unmarshalling. Are we sure that unmarshall functions don't mutate `val`?

omerfirmak: > > there is nothing wrong with that. We should keep it, there is no down side.
> 
> [callback hell](http://callbackhell.com/) is a thing which is why javascript introduced promises. Anyway, for us it is fine.

Yeah, I am not fond of callbacks either, but in this case it has performance implications to not use them. We should avoid GC as much as we can.

 
> > we are simply delegating parsing work to `cb`
> 
> In our case, we are using `cb` mostly for unmarshalling. Are we sure that unmarshall functions don't mutate `val`?

There is no reason they would do that. Even if they do, they are mutating a temporary copy. `val` is still in the memory, it is not directly mapped to the disk. So changing `val` doesnt really do anything.

db/badger.go
```diff
@@ -10,11 +11,18 @@ import (
 // golang
 type badgerDb struct {
 	badger *badger.DB
+	wMutex *sync.Mutex
```
IronGauntlets, line 16: Aren't badger transactions thread safe? So Why do we need to only use one transaction at a time? 

I don't think we need to have a reference to a [mutex](https://github.com/uber-go/guide/blob/master/style.md#zero-value-mutexes-are-valid). Although golang will panic if you were to copy a mutex.

omerfirmak: > Aren't badger transactions thread safe? So Why do we need to only use one transaction at a time?

that is mostly because Pebble does not have transactions but only `Batch`s. Batchs dont check for conflicts on between different write batches, so that would create race conditions if we were to allow multiple write "Transactions".

So I made it a requirement on our `DB` interface that it allows only a single update txn at a time.

omerfirmak: > I don't think we need to have a reference to a [mutex](https://github.com/uber-go/guide/blob/master/style.md#zero-value-mutexes-are-valid). Although golang will panic if you were to copy a mutex.

I will take a look, thanks!

db/db_test.go
```diff
@@ -226,3 +233,58 @@ func TestSeek(t *testing.T) {
...
+	}
+
+	wg.Wait()
+	require.NoError(t, testDb.View(func(txn db.Transaction) error {
+		return txn.Get(key, func(bytes []byte) error {
+			assert.Equal(t, byte(200), bytes[0])
+			return nil
+		})
+	}))
+}
```
IronGauntlets, line 290: Please split this case into 2 separate sub tests. One should be called `t.Run("Update()", ...)` and the other `t.Run("NewTransaction()", ...) `

IronGauntlets (changes requested):
Given that we now have 2 implementations of db interface (pebble is not trivial) we should have the following directory structure:

```
db
|-> db.go
|-> badger
  |-> badgerDB.go
  |-> badgerDB_test.go
  |-> badgerTxn.go
  |-> badgerTxn_test.go
|-> pebble
  |-> pebbleDB.go
  |-> pebbleDB_test.go
  |-> pebbleTxn.go
  |-> pebbleTxn_test.go
```

db/db_test.go
omerfirmak: It is intentional that they are in the same test tho. We dont really mix Update and NewTransaction in our code. But they should be respecting only-one-update-at-a-time rule.

ie NewTransaction(true) should block if a Update() is running and vice versa.

omerfirmak:
> 



> Given that we now have 2 implementations of db interface (pebble is not trivial) we should have the following directory structure:
> 
> ```
> db
> |-> db.go
> |-> badger
>   |-> badgerDB.go
>   |-> badgerDB_test.go
>   |-> badgerTxn.go
>   |-> badgerTxn_test.go
> |-> pebble
>   |-> pebbleDB.go
>   |-> pebbleDB_test.go
>   |-> pebbleTxn.go
>   |-> pebbleTxn_test.go
> ```

to do this we need a test suite like [this](https://github.com/ethereum/go-ethereum/blob/master/ethdb/dbtest/testsuite.go). Do you think it is worth the effort or should we just ditch badger? I don't think we should manage a badger backend as well.

IronGauntlets:
>
>> Given that we now have 2 implementations of db interface (pebble is not trivial) we should have the following directory structure:
>> 
>> ```
>> db
>> |-> db.go
>> |-> badger
>>   |-> badgerDB.go
>>   |-> badgerDB_test.go
>>   |-> badgerTxn.go
>>   |-> badgerTxn_test.go
>> |-> pebble
>>   |-> pebbleDB.go
>>   |-> pebbleDB_test.go
>>   |-> pebbleTxn.go
>>   |-> pebbleTxn_test.go
>> ```
>
>to do this we need a test suite like [this](https://github.com/ethereum/go-ethereum/blob/master/ethdb/dbtest/testsuite.go). Do you think it is worth the effort or should we just ditch badger? I don't think we should maintain a badger backend as well.

Let's delete badger.

omerfirmak:
> > > Given that we now have 2 implementations of db interface (pebble is not trivial) we should have the following directory structure:
> > > ```
> > > db
> > > |-> db.go
> > > |-> badger
> > >   |-> badgerDB.go
> > >   |-> badgerDB_test.go
> > >   |-> badgerTxn.go
> > >   |-> badgerTxn_test.go
> > > |-> pebble
> > >   |-> pebbleDB.go
> > >   |-> pebbleDB_test.go
> > >   |-> pebbleTxn.go
> > >   |-> pebbleTxn_test.go
> > > ```
> > 
> > 
> > to do this we need a test suite like [this](https://github.com/ethereum/go-ethereum/blob/master/ethdb/dbtest/testsuite.go). Do you think it is worth the effort or should we just ditch badger? I don't think we should maintain a badger backend as well.
> 
> Let's delete badger.

already done:)

IronGauntlets:
> already done:)

There are still references to badger in `bucket.go` comment and `log.go`. Also, you should run `make tidy` to remove it from out go modules.

omerfirmak:
Oh sorry, cleaned them up too.

db/db_test.go
```diff
@@ -191,24 +204,87 @@ func TestSeek(t *testing.T) {
 	assert.NoError(t, err)
 
 	testDb.View(func(txn db.Transaction) error {
-		next, err := txn.Seek([]byte{0})
+		err := txn.Seek([]byte{0}, func(next *db.Entry) error {
```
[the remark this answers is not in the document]
jelilat: Since [this PR](https://github.com/NethermindEth/juno/pull/567) will actually address this, we may want to approve though.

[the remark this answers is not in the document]
omerfirmak: I will just remove Seek from this PR, since it should be introduced in `Iterator` PR anyways. Thanks!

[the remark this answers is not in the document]
omerfirmak: done @jelilat

IronGauntlets (approved):

PR #560, opened 2023-02-10
Transaction Verification

IronGauntlets:
## Description

- Add L1Handler and Deploy Account Transactions
- Add L1Handler and Deploy Account Hash calculations
- Verify all transaction hashes based on the following rules:
  - Deploy: no verification since it will be deprecated
  - Declare: verify v1 but not v0
  - Invoke: verify v1 but not v0
  - Deploy Account: verify v1 (there is no v0)
  - L1 Hander: verify v0 (with some caveats see comments)
- Some sanity checks for Transactions and Transaction Receipts

## Types of changes

- 
## Testing

**Requires testing**: Yes


blockchain/blockchain.go
```diff
@@ -164,7 +157,22 @@ func (b *Blockchain) verifyBlock(txn db.Transaction, block *core.Block,
...
+	if len(block.Transactions) != len(block.Receipts) {
+		return ErrIncompatibleBlock{"len of transaction do not match len of receipts"}
+	}
+
+	for i := 0; i < len(block.Transactions); i++ {
+		if err = core.VerifyTransactionAndReceipt(block.Transactions[i],
+			block.Receipts[i], b.network); err != nil {
+			return ErrIncompatibleBlock{err.Error()}
+		}
+	}
```
omerfirmak, line 169: dont these belong in `VerifyBlockHash`?

omerfirmak: What I meant was moving these checks in to `VerifyBlockHash`. similar to https://github.com/NethermindEth/juno/pull/556/commits/30813eda425bfbe6dc7203c27b7e802d75bd0cc9

since verifying transaction hashes are ultimately a step in verifying the block hash

IronGauntlets: Yeah you are right, I am making changes now, just rebased

core/transaction.go
```diff
@@ -229,102 +231,131 @@ func (l *L1HandlerTransaction) signature() []*felt.Felt {
...
-			d.MaxFee,
-			network.ChainId(),
-			d.ClassHash,
-		), nil
+		// Due to inconsistencies in version 0 hash calculation we don't verify the hash
+		return nil, nil
 	} else if d.Version.IsOne() {
 		return crypto.PedersenArray(
-			declareFelt,
+			new(felt.Felt).SetBytes([]byte("declare")),
```
omerfirmak, line 275: can we make these constants global vars for each Tx type? could save use some calls to GC. 

We could hash them beforehand as well, so we avoid repeating the same calculations. But I am not sure if that is possible with the existing `PedersenArray` API we have

core/transaction.go
```diff
@@ -229,102 +231,131 @@ func (l *L1HandlerTransaction) signature() []*felt.Felt {
...
 			d.Nonce,
 		), nil
 	}
 	return nil, ErrInvalidTransactionVersion{d, d.Version.Text(10)}
 }
 
-func l1HandlerTransactionHash(l *L1HandlerTransaction, network utils.Network) (*felt.Felt, error) {
+var unverifiableL1HandlerTxM map[felt.Felt]struct{}
+
+func populateUnverifiableL1HandlerTx() {
```
omerfirmak, line 290: It sucks that we have to do this:( Feels like a disgusting hack :D

core/transaction.go
omerfirmak: We should implement an API that allows building a hash incrementally

core/block.go
```diff
@@ -86,6 +86,23 @@ func getBlockHashMetaInfo(network utils.Network) *blockHashMetaInfo {
...
+	if len(b.Transactions) != len(b.Receipts) {
+		return fmt.Errorf("len of transaction do not match len of receipts")
+	}
+
+	for i := 0; i < len(b.Transactions); i++ {
+		if !b.Transactions[i].hash().Equal(b.Receipts[i].TransactionHash) {
+			// todo: extract this into an error type so that more information about transaction hash
+			//  and index can be included. The error should be wrapped by the caller so that in the
+			//  test error.As() can be used for comparison.
+			return fmt.Errorf("transaction hash does not match receipt's hash")
```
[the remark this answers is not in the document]
omerfirmak: This goes for all the error types we have in our codebase already

core/transaction_test.go
```diff
@@ -44,23 +50,19 @@ func TestTransactionHash(t *testing.T) {
 		},
 	}
 
-	gw, closer := testsource.NewTestGateway(utils.MAINNET)
-	defer closer.Close()
-
 	for name, test := range tests {
 		t.Run(name, func(t *testing.T) {
+			gw, closer := testsource.NewTestGateway(test.network)
```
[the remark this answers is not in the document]
IronGauntlets: Before there was only one network. now there are multiple networks. We can create one gateway per network and then switch. I will do that. This was just for convenience. I actually started running into problems where ports were not available while testing.

blockchain/blockchain.go
```diff
@@ -43,6 +43,19 @@ func NewBlockchain(database db.DB, network utils.Network) *Blockchain {
 	}
 }
 
+func (b *Blockchain) Network() utils.Network {
+	return b.network
+}
+
+// StateCommitment returns the latest block state commitment.
+// If blockchain is empty nil is returned.
```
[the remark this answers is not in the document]
IronGauntlets: Yeah, this was a copy-paste. So didn't really put too much thought into it. I will update the comment.

core/block.go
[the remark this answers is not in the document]
IronGauntlets: So fmt doesn't wrap errors until you use `%w` verb.

We really need to sort out our error handling. It's been bugging me for some time. When we get some time I will prepare some material to discuss our approach.

core/transaction.go
IronGauntlets: I will make the global vars. 

I am not sure how building incrementally will help. As we don't share hash inputs.

core/transaction.go
IronGauntlets: Yeah, i am just dreading populating it. We have to sit there and watch for which hash it fails and then ignore it.

core/block_test.go
```diff
@@ -5,6 +5,8 @@ import (
 	_ "embed"
 	"testing"
 
+	"github.com/NethermindEth/juno/core/felt"
+
```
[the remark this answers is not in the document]
IronGauntlets: Will remove

core/transaction.go
omerfirmak: Different types does not share inputs, but for a given type, first round of Pedersen always gives the same result since the strings are constant.

core/transaction_test.go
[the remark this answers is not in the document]
IronGauntlets: So if we want fast tests we cannot share because it creates a race condition on the close functions (when using `t.Parallel`). If we don't use `t.Parallel` it takes 5s to run on my machine, with `t.Parallel` it takes 1s.

IronGauntlets:
@omerfirmak i have made all the changes. Please have another look.

blockchain/blockchain_test.go
```diff
@@ -135,28 +139,30 @@ func TestVerifyBlock(t *testing.T) {
...
+			expectedErr := blockchain.ErrIncompatibleBlock{
+				fmt.Errorf("block number difference between head and incoming block is not 1"),
 			}
-			assert.Equal(t, chain.VerifyBlock(incomingBlock), expectedErr)
+			assert.EqualError(t, chain.VerifyBlock(incomingBlock), expectedErr.Error())
 		})
 
 	t.Run("error when head hash does not match incoming block's parent hash", func(t *testing.T) {
 		incomingBlock := &core.Block{Header: core.Header{ParentHash: h1, Number: 1}}
+		err = chain.VerifyBlock(incomingBlock)
```
omerfirmak, line 152: same, redundant

core/transaction.go
```diff
@@ -186,64 +228,36 @@ func (l *L1HandlerTransaction) signature() []*felt.Felt {
...
 	if i.Version.IsZero() {
-		return crypto.PedersenArray(
-			invokeFelt,
-			i.ContractAddress,
-			i.EntryPointSelector,
-			crypto.PedersenArray(i.CallData...),
-			network.ChainId(),
-		), nil
+		// Due to inconsistencies in version 0 hash calculation we don't verify the hash
+		return nil, nil
```
omerfirmak, line 260: returning `nil` Felt seems dangerous, can we just return `i.Hash` so that we dont have to check for `nil` in `verifyTransactionHash`?

blockchain/blockchain.go
```diff
@@ -15,19 +15,23 @@ import (
...
 }
 
 func (e ErrIncompatibleBlockAndStateUpdate) Error() string {
-	return fmt.Sprintf("incompatible block and state update: %v", e.reason)
+	return fmt.Sprintf("incompatible block and state update: %v", e.Reason)
 }
 
 type ErrIncompatibleBlock struct {
-	reason string
+	Err error
```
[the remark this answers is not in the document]
IronGauntlets: Sure, it took me quite some time to figure out how to wrap errors without using `fmt.Errorf`. Should have read the documentation, lol.

blockchain/blockchain_test.go
```diff
@@ -135,28 +139,30 @@ func TestVerifyBlock(t *testing.T) {
 	t.Run("error if difference between incoming block number and head is not 1",
 		func(t *testing.T) {
 			incomingBlock := &core.Block{Header: core.Header{Number: 10}}
+			err = chain.VerifyBlock(incomingBlock)
```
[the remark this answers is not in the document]
IronGauntlets: yep I will fix

blockchain/blockchain_test.go
IronGauntlets: yep

core/transaction.go
IronGauntlets: I am going to remove this piece of code since we will log for inconsistencies

core/transaction.go
IronGauntlets: ok sure

core/transaction.go
IronGauntlets: We can have this task as a good first issue?

omerfirmak: Sounds good

IronGauntlets:
@omerfirmak @jelilat @joshklop pr is ready for review. The easiest way to review is by going through each commit individually.

blockchain/blockchain.go
```diff
@@ -209,20 +270,87 @@ func getBlockByHash(txn db.Transaction, hash *felt.Felt) (block *core.Block, err
...
+	} else if err = txn.Set(db.ReceiptsByBlockNumberAndIndex.Key(bnIndexBytes), rBytes); err != nil {
+		return err
+	}
 	return nil
 }
+
+// getTransactionBlockNumberAndIndexByHash gets the block number and index for a given transaction hash
+func getTransactionBlockNumberAndIndexByHash(txn db.Transaction, hash *felt.Felt) (bnIndexBytes []byte, err error) {
+	err = txn.Get(db.TransactionBlockNumbersAndIndicesByHash.Key(hash.Marshal()), func(val []byte) error {
+		bnIndexBytes = val
```
omerfirmak, line 377: we have to copy val in to bnIndexBytes, simply assigning the slice is not valid.

omerfirmak: unmarshaling it in to a `txAndReceiptDBKey` works as well

omerfirmak: done

core/transaction.go
```diff
@@ -93,12 +93,12 @@ type TransactionReceipt struct {
 }
 
 type Transaction interface {
-	hash() *felt.Felt
+	Hash() *felt.Felt
```
omerfirmak, line 96: I would've prefered `GetHash`. changing the field name to `TransactionHash` sounds redundant, because it is already a Transaction object

core/transaction.go
```diff
@@ -229,49 +229,33 @@ func (l *L1HandlerTransaction) signature() []*felt.Felt {
 func transactionHash(transaction Transaction, n utils.Network) (*felt.Felt, error) {
 	switch t := transaction.(type) {
 	case *DeclareTransaction:
-		return declareTransactionHash(t, network)
+		return declareTransactionHash(t, n)
```
omerfirmak, line 232: I think these belong to a previous commit?

https://github.com/NethermindEth/juno/pull/560/commits/2387f7eedaf9c5787c177977c04723f65fa7e651#diff-65ddbfca57544ad3aaedea5dbb6e4911bbe40753fb680c41a527dccff59311ebL229

blockchain/blockchain.go
```diff
@@ -248,22 +248,76 @@ func storeBlock(txn db.Transaction, block *core.Block) error {
...
-	})
-	return
+func getBlockByNumberBytes(txn db.Transaction, numBytes []byte) (*core.Block, error) {
+	var txs []core.Transaction
+	var receipts []*core.TransactionReceipt
+
+	block, header := new(core.Block), new(core.Header)
+
+	if err := txn.Get(db.BlockHeadersByNumber.Key(numBytes), func(val []byte) error {
+		return encoder.Unmarshal(val, header)
```
omerfirmak, line 261: we can directly unmarshal in to `&block.Header`

blockchain/blockchain.go
```diff
@@ -248,22 +248,76 @@ func storeBlock(txn db.Transaction, block *core.Block) error {
...
+		if err != nil {
+			return nil, err
+		}
+
+		var tx core.Transaction
+		if err = encoder.Unmarshal(val, &tx); err != nil {
+			return nil, err
+		}
+
+		txs = append(txs, tx)
```
omerfirmak, line 288: same, we can directly append to `block.Transactions`

blockchain/blockchain.go
```diff
@@ -248,22 +248,76 @@ func storeBlock(txn db.Transaction, block *core.Block) error {
...
+		if err != nil {
+			return nil, err
+		}
+
+		receipt := new(core.TransactionReceipt)
+		if err = encoder.Unmarshal(val, receipt); err != nil {
+			return nil, err
+		}
+
+		receipts = append(receipts, receipt)
```
omerfirmak, line 307: same, `block.Receipts`

blockchain/blockchain.go
omerfirmak: done

blockchain/blockchain.go
omerfirmak: done

blockchain/blockchain.go
omerfirmak: done

core/transaction.go
omerfirmak: done

blockchain/blockchain.go
IronGauntlets: > we have to copy val into bnIndexBytes, simply assigning the slice is not valid.

In a previous pull request, I raised this [point](https://github.com/NethermindEth/juno/pull/557#discussion_r1107078125) and you said because of callbacks val does leave scope of the callback so we didn't have to copy.

Also, do unmarshalling functions always create a copy?

IronGauntlets: This also means that there is a bug in our get block functions because it follows the same previous pattern. I will look into it.

omerfirmak: I must have been unclear then.

If we want to use `val` outside of the callback, we must copy. Assume `val` is an array allocated on the stack. So, outside the callback it does not exist.

>  This also means that there is a bug in our get block functions because it follows the same previous pattern. I will look into it.

I see unmarshals only in our block getters, which one are you referring to?

core/transaction.go
IronGauntlets: Go discourages `Get` Prefixes. That is why I don't like to use it until absolutely necessary. But I am happy to change the field to `Hash` and rename this function to `GetHash`.

@jelilat named it `TransactionHash`, I kept it because the thought the interface function `Hash()` would be used more so people wouldn't notice the stutter (that's what go calls `transaction.TransactionHash`).

blockchain/blockchain.go
omerfirmak: > Also, do unmarshalling functions always create a copy?

they allocate a new object and populate it from the given data, they don't use the underlying memory location that `val` points to

core/transaction.go
omerfirmak: Got it, I wasnt suggesting that we should change it. Merely pointing it out

sync/sync.go
```diff
@@ -73,12 +73,17 @@ func (s *Synchronizer) verifierTask(ctx context.Context, block *core.Block, stat
 			return
 		default:
 			if err != nil {
-				s.log.Warnw("Sanity checks failed", "number", block.Number, "hash", block.Hash.Text(16))
-				select {
-				case <-ctx.Done():
-				case errChan <- ErrSyncFailed{block.Number, err}:
+				if errors.As(err, new(core.ErrCantVerifyTransactionHash)) {
+					s.log.Warnw("Sanity checks failed", "number", block.Number, "hash",
```
IronGauntlets, line 77: use debug

blockchain/blockchain.go
```diff
@@ -248,23 +246,66 @@ func storeBlock(txn db.Transaction, block *core.Block) error {
...
-	})
-	return
+	block := new(core.Block)
+	if err := txn.Get(db.BlockHeadersByNumber.Key(numBytes), func(val []byte) error {
+		return encoder.Unmarshal(val, &block.Header)
+	}); err != nil {
+		return nil, err
+	}
+
+	iterator, err := txn.NewIterator()
```
IronGauntlets, line 259: close

core/transaction.go
IronGauntlets: Ok i am not going to change it for now

blockchain/blockchain.go
IronGauntlets: done

core/block.go
```diff
@@ -121,12 +128,50 @@ func VerifyBlockHash(b *Block, network utils.Network) error {
...
+			return []error{err}
 		} else if hash.Equal(b.Hash) {
 			return nil
 		}
 	}
-	return ErrCantVerifyBlockHash
+	return []error{ErrCantVerifyBlockHash}
+}
+
+func verifyTransactionHashes(txs []Transaction, n utils.Network) []error {
```
omerfirmak, line 139: we should benchmark if this is actually faster, we already have `runtime.NumCPU` level of parallelism at block level. If we add parallelism at txn level as well we potentially going to have `runtime.NumCPU ^ 2` number of CPU heavy goroutines in-flight which will lead to starvation.

IronGauntlets: Having tested the sync times it looks like this is worse than sequential verification so I will revert. On `develop` it was around 1 minute faster to sync 429 blocks vs this branch.

IronGauntlets: I will make it sequential

IronGauntlets: done

core/block.go
```diff
@@ -85,7 +85,34 @@ func getBlockHashMetaInfo(network utils.Network) *blockHashMetaInfo {
...
+		if !tx.Hash().Equal(b.Receipts[i].TransactionHash) {
+			return []error{
+				fmt.Errorf(
+					"transaction hash (0x%v) at index: %v does not match receipt's hash (0x%v)",
+					tx.Hash().Text(16), i, b.Receipts[i].TransactionHash),
+			}
+		}
+	}
+
+	var errs []error
```
omerfirmak, line 106: Would making this a chain of errors by wrapping make sense?

omerfirmak: at sync, we can follow the chain and report the "transaction hash failed" ones.

omerfirmak (approved):

blockchain/blockchain.go
```diff
@@ -37,13 +45,26 @@ type Blockchain struct {
...
 	}
 }
 
+func (b *Blockchain) Network() utils.Network {
+	return b.network
+}
+
+// StateCommitment returns the latest block state commitment.
+// If blockchain is empty zero felt is returned.
+func (b *Blockchain) StateCommitment() (commitment *felt.Felt, err error) {
```
jelilat, line 61: Will this be used for verifying state before storing the state update?

core/block_test.go
```diff
@@ -13,94 +16,207 @@ import (
...
 		block119802, err := goerliGW.BlockByNumber(context.Background(), 119802)
 		require.NoError(t, err)
 
-		err = core.VerifyBlockHash(block119802, utils.GOERLI)
-		assert.NoError(t, err)
+		errs := core.VerifyBlockHash(block119802, utils.GOERLI)
+		assert.Equal(t, 0, len(errs))
+	})
+
+	t.Run("error if len of transaction do not match len of receipts", func(t *testing.T) {
```
jelilat, line 193:
```suggestion
	t.Run("error if len of transactions do not match len of receipts", func(t *testing.T) {
```

jelilat (changes requested):

core/block.go
IronGauntlets: Yeah, that is a good idea. I will think about how I can do that

blockchain/blockchain.go
IronGauntlets: No, this is just to make it available for through the blockchain object. I needed this mainly for testing.

testsource/client.go
```diff
@@ -25,6 +25,9 @@ func NewTestClient(network utils.Network) (*clients.GatewayClient, io.Closer) {
 	return client, &srvCloser{srv}
 }
 
+// NewTestGateway Todo: return a close function instead of an interface
+//
+//	func NewTestGateway(network utils.Network) (*gateway.Gateway, func() error)
```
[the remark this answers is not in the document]
IronGauntlets: I just need to rebase

core/block_test.go
IronGauntlets: done

core/block.go
```diff
@@ -85,7 +85,34 @@ func getBlockHashMetaInfo(network utils.Network) *blockHashMetaInfo {
 
 // VerifyBlockHash verifies the block hash. Due to bugs in Starknet alpha, not all blocks have
 // verifiable hashes.
-func VerifyBlockHash(b *Block, network utils.Network) error {
+func VerifyBlockHash(b *Block, network utils.Network) []error {
+	if len(b.Transactions) != len(b.Receipts) {
+		return []error{
+			fmt.Errorf("len of transaction: %v do not match len of receipts: %v",
```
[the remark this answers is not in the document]
IronGauntlets: done

core/block.go
IronGauntlets: You this make me think we should wrap close errors as well instead of prioritising

blockchain/blockchain.go
jelilat: Alright. I just wanted to understand its use case.

jelilat (approved):

PR #561, opened 2023-02-14
Implement L1 Verifier

jelilat:
## Description

Allows us to check if a block has been verified on L1 or not

## Types of changes

## Testing

**Requires testing**: Yes

## Documentation

### DIFF
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -478,7 +478,7 @@ func getReceiptByBlockNumberAndIndex(txn db.Transaction, bnIndex *txAndReceiptDB
 	})
 }
 
-func (b *Blockchain) StoreL1ChainHeight(height uint64, block *core.Block) error {
+func (b *Blockchain) StoreL1ChainHeight(height uint64) error {
 	return b.database.Update(func(txn db.Transaction) error {
 		heightBin := make([]byte, lenOfByteSlice)
 		binary.BigEndian.PutUint64(heightBin, height)
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -39,6 +39,14 @@ func (e ErrIncompatibleBlock) Unwrap() error {
 	return e.Err
 }
 
+type ErrInvalidL1Verification struct {
+	reason string
+}
+
+func (e ErrInvalidL1Verification) Error() string {
+	return fmt.Sprintf("invalid L1 verification: %v", e.reason)
+}
+
 type Reader interface {
 	Height() (height uint64, err error)
 	Head() (head *core.Block, err error)
@@ -82,7 +90,22 @@ func (b *Blockchain) Height() (height uint64, err error) {
 }
 
 func (b *Blockchain) height(txn db.Transaction) (height uint64, err error) {
-	return height, txn.Get(db.ChainHeight.Key(), func(val []byte) error {
+	return height, txn.Get(db.L2ChainHeight.Key(), func(val []byte) error {
+		height = binary.BigEndian.Uint64(val)
+		return nil
+	})
+}
+
+// L1ChainHeight returns the latest L1 verified block height.
+func (b *Blockchain) L1ChainHeight() (height uint64, err error) {
+	return height, b.database.View(func(txn db.Transaction) error {
+		height, err = b.l1ChainHeight(txn)
+		return err
+	})
+}
+
+func (b *Blockchain) l1ChainHeight(txn db.Transaction) (height uint64, err error) {
+	return height, txn.Get(db.L1ChainHeight.Key(), func(val []byte) error {
 		height = binary.BigEndian.Uint64(val)
 		return nil
 	})
@@ -182,7 +205,7 @@ func (b *Blockchain) Store(block *core.Block, stateUpdate *core.StateUpdate, dec
 		// [db.ChainHeight]() -> (BlockNumber)
 		heightBin := make([]byte, lenOfByteSlice)
 		binary.BigEndian.PutUint64(heightBin, block.Number)
-		return txn.Set(db.ChainHeight.Key(), heightBin)
+		return txn.Set(db.L2ChainHeight.Key(), heightBin)
 	})
 }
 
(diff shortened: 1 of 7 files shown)


core/block.go
```diff
@@ -26,6 +26,8 @@ type Header struct {
 	Timestamp *felt.Felt
 	// The version of the Starknet protocol used when creating this block
 	ProtocolVersion string
+	// The status of the block
+	Status string
```
omerfirmak, line 30: core.Block should not have this.

we can decide on a blocks status just by looking at the `L1ChainHeight` we have in database.
```
if L1ChainHeight > block.Number {
 return "L1_VERIFIED"
} else {
 return "L2_VERIFIED"
}

```

omerfirmak: Yeah, it is definitely OK. You can use a GatewayClient, this is just a placeholder anyways.

blockchain/blockchain.go
```diff
@@ -209,3 +233,47 @@ func (b *Blockchain) SanityCheckNewHeight(block *core.Block, stateUpdate *core.S
 
 	return core.VerifyBlockHash(block, b.network)
 }
+
+func (b *Blockchain) StoreL1ChainHeight(height uint64, block *core.Block) error {
+	if err := b.VerifyL1ChainHeight(height); err != nil {
```
omerfirmak, line 238: this shouldn't be creating a new txn. We should drop the call to `VerifyL1ChainHeight` and call `verifyL1ChainHeight` inside the Update() callback

blockchain/blockchain.go
```diff
@@ -209,3 +233,47 @@ func (b *Blockchain) SanityCheckNewHeight(block *core.Block, stateUpdate *core.S
...
+	return b.verifyL1ChainHeight(txn, height)
+}
+
+func (b *Blockchain) verifyL1ChainHeight(txn db.Transaction, height uint64) error {
+	l2Height, err := b.height(txn)
+	if err != nil && !errors.Is(err, db.ErrKeyNotFound) {
+		return err
+	}
+	if height > l2Height {
+		return ErrInvalidL1Verification{
```
omerfirmak, line 260: this is not necessarily true. Our sync could be slower then our L1 verification.

blockchain/blockchain.go
```diff
@@ -209,3 +233,47 @@ func (b *Blockchain) SanityCheckNewHeight(block *core.Block, stateUpdate *core.S
...
+			"l1 height cannot be greater than l2 height",
+		}
+	}
+
+	l1Height, err := b.l1ChainHeight(txn)
+	if err != nil && !errors.Is(err, db.ErrKeyNotFound) {
+		return err
+	}
+	if l1Height+1 != height && l1Height != height {
+		return ErrInvalidL1Verification{
```
omerfirmak, line 270: we shouldnt need to verify blocks one by one from L1. That would require an ethereum archive node. When we implement the actual L1 verification, we should only be listening to new verification events.

For example; if we detect a block on height `N` is verified, if the `N`th block in our storage matches the verified block's hash, we now know just that all previous blocks were L1 verified as well.

node/node.go
```diff
@@ -75,7 +76,7 @@ func New(cfg *Config) (StarknetNode, error) {
 	}
 
 	blockchain := blockchain.NewBlockchain(stateDb, cfg.Network)
-	synchronizer := sync.NewSynchronizer(blockchain, gateway.NewGateway(cfg.Network), log)
+	synchronizer := sync.NewSynchronizer(blockchain, gateway.NewGateway(cfg.Network), clients.NewGatewayClient(cfg.Network.URL()), log)
```
omerfirmak, line 79: L1 verifier should probably be another service. Not embedded inside the synchronizer

sync/l1verifier.go
```diff
@@ -0,0 +1,117 @@
...
+func NewL1Verifier(bc *blockchain.Blockchain, gateway *clients.GatewayClient, log utils.SimpleLogger) *L1Verifier {
+	return &L1Verifier{
+		Blockchain: bc,
+		Gateway:    gateway,
+		log:        log,
+	}
+}
+
+// Run starts the L1Verifier.
+func (v *L1Verifier) Run(syncCtx context.Context, verifier *stream.Stream, nextVerifiedHeight uint64) stream.Callback {
```
omerfirmak: why do you need a stream here?

sync/l1verifier.go
```diff
@@ -0,0 +1,117 @@
...
+					})
+			}
+		}
+	}
+	// }
+	return func() {}
+}
+
+func (v *L1Verifier) verifier(ctx context.Context, nextVerifiedHeight uint64, errChan chan ErrL1VerificationFailed) stream.Callback {
+	return func() {
```
omerfirmak, line 60: this provides no concurrency, callbacks are executed synchronously

sync/sync.go
```diff
@@ -26,14 +28,16 @@ func (e ErrSyncFailed) Error() string {
 type Synchronizer struct {
 	Blockchain   *blockchain.Blockchain
 	StarknetData starknetdata.StarknetData
+	Gateway      *clients.GatewayClient
```
omerfirmak, line 40: synchronizer shouldnt need a GatewayClient reference.

omerfirmak (changes requested):

sync/l1verifier.go
jelilat: That was intentional because `StoreL1ChainHeight` needs to be synchronous just like we do with block storage.What about just moving `GetBlock` so it will be called before the callback?

blockchain/blockchain.go
omerfirmak: Yeap, that is another edge case we need to handle

sync/l1verifier.go
omerfirmak: You shouldn't need to use any streams here. You only need a single goroutine to make this work, not a pool of workers.

sync/sync.go
jelilat: Had to add this cos l1Verifier is called in sync

omerfirmak: see https://github.com/NethermindEth/juno/pull/561#discussion_r1116848109 then

sync/l1verifier.go
jelilat: Got it. Thanks

sync/l1verifier.go
```diff
@@ -0,0 +1,109 @@
...
+		Blockchain: bc,
+		Gateway:    gateway,
+		log:        log,
+	}
+}
+
+// Run starts the L1Verifier.
+func (v *L1Verifier) Run(ctx context.Context) error {
+	errChan := make(chan ErrL1VerificationFailed)
+	l1Stream := stream.New().WithMaxGoroutines(runtime.NumCPU())
```
omerfirmak, line 46: We are not supposed to L1 verify blocks in parallel like we do in sync.

Assume our L2 height is 100 and L1 height is `nil`. The actual L1 verified height is 75, but we havent figured that out yet.

Try using a binary search like approach to find the latest L1 verified block. For example;
- GetBlock(100), and notice it is not L1 verified
- GetBlock(50), and notice it IS L1 verified, make `blockchain` know that at height 50 a block with hash of `H1` was verified
- GetBlock(75)  and notice it IS L1 verified, make `blockchain` know that at height 75 a block with hash of `H2` was verified
- GetBlock(87), not verified
- GetBlock(82), not verified

At some point this search will yield and L1Verifier will realize that 75 is the latest block that is L1 verified and at that point it can start to L1 verify blocks one by one. You dont need a `stream` here, just make sure `Run()` does not block the execution and can be cancelled with ctx.

omerfirmak (changes requested):

sync/l1verifier.go
jelilat: Done

jelilat:
Closing because we'll be considering a new approach

PR #562, opened 2023-02-14
Cap gateway client backoff to max wait time

joshklop:
## Description

Strictly exponential backoff can cause significant delays (e.g., if a block is created while we are in the middle of a 30 minute backoff). By capping the maximum wait time at one minute, we can download blocks soon after they are created.

## Changes:

- Change default backoff function
- Change default number of maximum retries (now 35).

## Types of changes

## Testing

**Requires testing**: Yes

**Did you write tests??**: No (manual testing preferred for these changes; existing tests are sufficient to ensure backoff works as expected)

## Documentation

**If this requires a documentation update, did you add one?** No, No

### DIFF
--- a/clients/gateway.go
+++ b/clients/gateway.go
@@ -19,6 +19,7 @@ type GatewayClient struct {
 	url        string
 	backoff    Backoff
 	maxRetries int
+	maxWait    time.Duration
 	minWait    time.Duration
 }
 
@@ -32,12 +33,16 @@ func (c *GatewayClient) WithMaxRetries(num int) *GatewayClient {
 	return c
 }
 
+func (c *GatewayClient) WithMaxWait(d time.Duration) *GatewayClient {
+	c.maxWait = d
+	return c
+}
+
 func (c *GatewayClient) WithMinWait(d time.Duration) *GatewayClient {
 	c.minWait = d
 	return c
 }
 
-// https://cloud.google.com/iot/docs/how-tos/exponential-backoff
 func ExponentialBackoff(wait time.Duration) time.Duration {
 	time.Sleep(wait)
 	return wait * 2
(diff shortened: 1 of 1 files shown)


clients/gateway.go
```diff
@@ -37,10 +37,13 @@ func (c *GatewayClient) WithMinWait(d time.Duration) *GatewayClient {
 	return c
 }
 
-// https://cloud.google.com/iot/docs/how-tos/exponential-backoff
-func ExponentialBackoff(wait time.Duration) time.Duration {
+func ExponentialUpToOneMinBackoff(wait time.Duration) time.Duration {
 	time.Sleep(wait)
-	return wait * 2
+	wait *= 2
+	if wait > time.Minute {
```
omerfirmak, line 43: I think it makes sense to make this configurable. Block times are different on different networks, so even a minute could be too much for some network.

joshklop: I agree. However, it will require adding a `maxWait` parameter to the `Backoff` function type that will always be have the same value. Since we already have `WithBackoff`, do you think it is enough for users to implement their own?

Alternatively, we could export default functions for the common networks `MainnetBackoff`, `GoerliBackoff`, etc. that wrap a function like `ExponentialBackoffWithMaximum`. 

Ex:
```go
func MainnetBackoff(wait time.Duration) time.Duration {
        ExponentialBackoffWithMaximum(wait, time.Minute)
}
```

omerfirmak: How about;

``` golang
type Backoff func(numOfTriesSoFar uint)  time.Duration
```

and we can make `ExponentialBackoff` a member function of GatewayClient. So it has access to the configuration

joshklop: The proposed function signature looks good to me, but users wouldn't be able to create their own backoff functions if it's a method.

clients/gateway.go
```diff
@@ -32,13 +33,17 @@ func (c *GatewayClient) WithMaxRetries(num int) *GatewayClient {
...
+}
+
 func (c *GatewayClient) WithMinWait(d time.Duration) *GatewayClient {
 	c.minWait = d
 	return c
 }
 
-// https://cloud.google.com/iot/docs/how-tos/exponential-backoff
-func ExponentialBackoff(wait time.Duration) time.Duration {
+func ExponentialUpToOneMinBackoff(wait time.Duration) time.Duration {
```
omerfirmak, line 46: should be renamed back to `ExponentialBackoff`

omerfirmak (approved):

PR #564, opened 2023-02-15
Refactor GatewayClient retry mechanism

omerfirmak:
Old implementation was not quite doing what it was supposed to


clients/gateway_test.go
```diff
@@ -517,11 +517,15 @@ func TestGetClassDefinition(t *testing.T) {
 }
 
 func TestHttpError(t *testing.T) {
+	maxRetries := 2
+	callCount := make(map[string]int)
 	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		soFar := callCount[r.URL.String()]
+		callCount[r.URL.String()] = soFar + 1
```
IronGauntlets, line 524:
```suggestion
		callCount[r.URL.String()]++
```

Not entirely sure the above will work but worth a try.

IronGauntlets (approved):

clients/gateway.go
```diff
@@ -87,32 +86,35 @@ func (c *GatewayClient) buildQueryString(endpoint string, args map[string]string
...
 		}
 	}
-	if err != nil {
-		return nil, err
-	} else if res.StatusCode != http.StatusOK {
-		return nil, errors.New(res.Status)
-	}
-
-	return io.ReadAll(res.Body)
+	return nil, err
```
joshklop, line 117: Will this return a nil error after `maxRetries`? If so, maybe we can create a new error?

clients/gateway_test.go
omerfirmak: works:) thanks!

joshklop (approved):

clients/gateway.go
omerfirmak: It just returns the last error `Do` returned. am I missing something?

IronGauntlets: I believe it will the error on line 105

omerfirmak: or what L105 sets it to

PR #569, opened 2023-02-16
Move all instantiations to node.New

omerfirmak:


node/node.go
```diff
@@ -60,29 +60,34 @@ func New(cfg *Config) (StarknetNode, error) {
...
+	blockchain := blockchain.NewBlockchain(stateDb, cfg.Network)
+	synchronizer := sync.NewSynchronizer(blockchain, gateway.NewGateway(cfg.Network), log)
+	n := &Node{
+		cfg:          cfg,
+		log:          log,
+		db:           stateDb,
+		blockchain:   blockchain,
+		synchronizer: synchronizer,
+	}
+	return n, nil
```
IronGauntlets, line 85:
```suggestion
	return &Node{
		cfg:          cfg,
		log:          log,
		db:           stateDb,
		blockchain:   blockchain,
		synchronizer: synchronizer,
	}, nil
```

IronGauntlets (changes requested):

node/node.go
omerfirmak: done

joshklop (approved):

PR #570, opened 2023-02-17
Implement Poseidon hash

omerfirmak:


core/crypto/poseidon_hash.go
```diff
@@ -0,0 +1,176 @@
...
+import (
+	"math/big"
+
+	"github.com/NethermindEth/juno/core/felt"
+)
+
+var (
+	fullRounds    = 8
+	partialRounds = 83
+)
```
IronGauntlets, line 12:
```suggestion
const (
	fullRounds    = 8
	partialRounds = 83
)
```

core/crypto/poseidon_hash.go
```diff
@@ -0,0 +1,176 @@
...
+
+func permutate(state []felt.Felt) {
+	totalRounds := fullRounds + partialRounds
+	for i := 0; i < totalRounds; i++ {
+		full := (i < fullRounds/2) || (totalRounds-i <= fullRounds/2)
+		round(state, full, i)
+	}
+}
+
+func Poseidon(elems []*felt.Felt) *felt.Felt {
```
IronGauntlets, line 54: can this be the following:

`func Poseidon(elems ...*felt.Felt)`

core/crypto/poseidon_hash_test.go
```diff
@@ -0,0 +1,21 @@
...
+func TestPoseidon(t *testing.T) {
+	msg := []*felt.Felt{
+		new(felt.Felt).SetUint64(0),
+		new(felt.Felt).SetUint64(1),
+		new(felt.Felt).SetUint64(2),
+		new(felt.Felt).SetUint64(3),
+	}
+	assert.Equal(t, strings.ToLower("30ACF0EF3C4549E3684B19652015FD9EADF7BDBA2A1A46CB29B9E18E6622296"),
+		crypto.Poseidon(msg).Text(16))
+}
```
IronGauntlets, line 21: Please add benchmarks similar to Pedersen hash

IronGauntlets (changes requested):

core/crypto/poseidon_hash_test.go
omerfirmak: all done

core/crypto/poseidon_hash.go
```diff
@@ -0,0 +1,177 @@
...
+
+const (
+	fullRounds    = 8
+	partialRounds = 83
+)
+
+func addRoundKeys(state []felt.Felt, round int) {
+	state[0].Add(&state[0], &roundKeys[round][0])
+	state[1].Add(&state[1], &roundKeys[round][1])
+	state[2].Add(&state[2], &roundKeys[round][2])
```
jelilat, line 17:
```suggestion
	for i := 0; i < 3; i++ {
		state[i].Add(&state[i], &roundKeys[round][i])
	}
```

jelilat (changes requested):

core/crypto/poseidon_hash.go
```diff
@@ -0,0 +1,177 @@
...
+var three = new(big.Int).SetUint64(3)
+
+func subWords(state []felt.Felt, full bool) {
+	squared := felt.Felt{}
+	state[2].Mul(&state[2], squared.Mul(&state[2], &state[2]))
+	if full {
+		state[0].Mul(&state[0], squared.Mul(&state[0], &state[0]))
+		state[1].Mul(&state[1], squared.Mul(&state[1], &state[1]))
+	}
+}
```
jelilat, line 29:
```suggestion
func subWords(state []felt.Felt, full bool) {
	state[2].Exp(&state[2], three)
	if full {
		state[0].Exp(&state[0], three)
		state[1].Exp(&state[1], three)
	}
}
```

core/felt/felt.go
```diff
@@ -174,3 +174,33 @@ func (z *Felt) Bits() [4]uint64 {
...
 	return z.val.BigInt(res)
 }
+
+// Double forwards the call to underlying field element implementation
+func (z *Felt) Double(x *Felt) *Felt {
+	z.val.Double(&x.val)
+	return z
+}
+
+// Sub forwards the call to underlying field element implementation
```
jelilat, line 184: I think we should update all the comments her to depict what the function is doing cos it'll help a lot when reading through codes. I personally had to move through 2 layers of files just to know what a function is doing. It becomes a problem when you have to do this multiple times.

jelilat (changes requested):

core/crypto/poseidon_hash.go
```diff
@@ -0,0 +1,177 @@
...
+	partialRounds = 83
+)
+
+func addRoundKeys(state []felt.Felt, round int) {
+	state[0].Add(&state[0], &roundKeys[round][0])
+	state[1].Add(&state[1], &roundKeys[round][1])
+	state[2].Add(&state[2], &roundKeys[round][2])
+}
+
+var three = new(big.Int).SetUint64(3)
```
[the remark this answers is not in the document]
omerfirmak: yeap, will go

core/crypto/poseidon_hash.go
```diff
@@ -0,0 +1,177 @@
...
+
+func permutate(state []felt.Felt) {
+	totalRounds := fullRounds + partialRounds
+	for i := 0; i < totalRounds; i++ {
+		full := (i < fullRounds/2) || (totalRounds-i <= fullRounds/2)
+		round(state, full, i)
+	}
+}
+
+var two = new(felt.Felt).SetUint64(2)
```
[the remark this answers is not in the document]
omerfirmak: it is a constant value, declared it outside to avoid having to create it in every call.

core/crypto/poseidon_hash.go
omerfirmak: Hardcoded `Mul`s are faster then `Exp`

core/crypto/poseidon_hash.go
omerfirmak: loops are essentially branches an incur a performance penalty. See "loop unrolling" and https://dev.to/jobinrjohnson/branchless-programming-does-it-really-matter-20j4

core/felt/felt.go
omerfirmak: ideally, we should get rid of all those wrappers. So it is not really worth to document them properly.

core/crypto/poseidon_hash.go
jelilat: Thanks for the link

core/crypto/poseidon_hash.go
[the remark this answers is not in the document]
omerfirmak: done

core/crypto/poseidon_hash.go
```diff
@@ -0,0 +1,197 @@
...
+
+	rem := len(elems) % 2
+	if rem == 1 {
+		state[0].Add(&state[0], elems[len(elems)-1])
+	}
+	state[rem].Add(&state[rem], new(felt.Felt).SetUint64(1))
+	permutate(state)
+
+	return new(felt.Felt).Set(&state[0])
+}
```
jelilat, line 82: This does not work correctly if the length of the array is 1 or 2. Just like in our pedersen hash function, I think this should also be able to produce the correct hash whether the length is 0, 1, or 2.

jelilat (changes requested):

core/crypto/poseidon_hash.go
omerfirmak: Poseidon2 and PoseidonArray is 2 different functions. They are not supposed to yield the same results.

see https://docs.starknet.io/documentation/architecture_and_concepts/Hashing/hash-functions/

jelilat: Can we have Poseidon1 implemented as well then?

omerfirmak: Of course, but it is not really used anywhere at this point.

jelilat: Was able to do some comparison test for this and yes, it works perfectly. Great work, boss

jelilat: We should try generating tests by running random values through Cairo Lang's Python implementation

core/crypto/poseidon_hash_test.go
```diff
@@ -0,0 +1,26 @@
...
+	if err != nil {
+		b.Errorf("Error occured %s", err)
+	}
+
+	e1, err := new(felt.Felt).SetString("0x208a0a10250e382e1e4bbe2880906c2791bf6275695e02fbbc6aeff9cd8b31a")
+	if err != nil {
+		b.Errorf("Error occured %s", err)
+	}
+
+	var f *felt.Felt
```
IronGauntlets, line 21: Please refer to the section "A note on compiler optimisation" [here](https://dave.cheney.net/2013/06/30/how-to-write-benchmarks-in-go) for why the variable `f` should be declared as a package-level variable.

core/crypto/poseidon_hash_test.go
```diff
@@ -0,0 +1,26 @@
+package crypto_test
+
+import (
+	"testing"
+
+	"github.com/NethermindEth/juno/core/crypto"
+	"github.com/NethermindEth/juno/core/felt"
+)
+
```
IronGauntlets, line 9: Please add PoseidonArray benchmark as well

core/crypto/poseidon_hash_test.go
```diff
@@ -0,0 +1,26 @@
...
+func BenchmarkPoseidon(b *testing.B) {
```
IronGauntlets, line 10: Please try to use [`testing.TB`](https://pkg.go.dev/testing#TB) interface

core/crypto/poseidon_hash_test.go
IronGauntlets, line 9: Please add some tests for Poseidon and PoseidonArray

core/crypto/poseidon_hash.go
```diff
@@ -0,0 +1,197 @@
...
+	state[2].Sub(stateSum, state[2].Add(&state[2], new(felt.Felt).Double(&state[2]))) // stateSum - state[2] * 3
+}
+
+func round(state []felt.Felt, full bool, index int) {
+	addRoundKeys(state, index)
+	subWords(state, full)
+	mixLayer(state)
+}
+
+func permutate(state []felt.Felt) {
```
IronGauntlets, line 43: should we call this `hadesPermutation` to be more specific to spec?

IronGauntlets (changes requested):

core/crypto/poseidon_hash_test.go
omerfirmak: doesnt work for some reason, requires a wrapper of some sort.

IronGauntlets (approved):

PR #572, opened 2023-02-20
Add Database Iterations

jelilat:
## Description

This PR allows us to be able to query the database using prefixes. It will be useful when querying all the`Transactions` and `Receipts` in a `Block`. 

We will now be able to query all `Transactions` using the block number as a prefix.

## Types of changes

## Testing

**Requires testing**: Yes

### DIFF
--- a/db/db.go
+++ b/db/db.go
@@ -6,7 +6,7 @@ import (
 )
 
 // ErrKeyNotFound is returned when key isn't found on a txn.Get.
-var ErrKeyNotFound = errors.New("Key not found")
+var ErrKeyNotFound = errors.New("key not found")
 
 // DB is a key-value database
 type DB interface {
@@ -30,6 +30,8 @@ type DB interface {
 
 // Iterator is an iterator over a DB's key/value pairs.
 type Iterator interface {
+	io.Closer
+
 	// Valid returns true if the iterator is positioned at a valid key/value pair.
 	Valid() bool
 
@@ -42,10 +44,7 @@ type Iterator interface {
 	Key() []byte
 
 	// Value returns the value at the current position.
-	Value() []byte
-
-	// Close closes the iterator.
-	Close() error
+	Value() ([]byte, error)
 
 	// Seek would seek to the provided key if present. If absent, it would seek to the next
 	// key in lexicographical order
--- a/db/db.go
+++ b/db/db.go
@@ -28,16 +28,36 @@ type DB interface {
 	Impl() any
 }
 
-// Entry is a database entry consisting of a Key and Value
-type Entry struct {
-	Key   []byte
-	Value []byte
+// Iterator is an iterator over a DB's key/value pairs.
+type Iterator interface {
+	// Valid returns true if the iterator is positioned at a valid key/value pair.
+	Valid() bool
+
+	// Next moves the iterator to the next key/value pair. It returns whether the
+	// iterator is valid after the call. Once invalid, the iterator remains
+	// invalid.
+	Next() bool
+
+	// Key returns the key at the current position.
+	Key() []byte
+
+	// Value returns the value at the current position.
+	Value() []byte
+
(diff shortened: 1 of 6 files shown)


omerfirmak (changes requested):

db/db.go
```diff
@@ -34,10 +34,41 @@ type Entry struct {
...
+
+	// Value returns the value at the current position.
+	Value() []byte
+
+	// Close closes the iterator.
+	Close() error
+
+	// Seek would seek to the provided key if present. If absent, it would seek to the next
+	// key in lexicographic order
+	Seek(cb func(*Entry) error) error
```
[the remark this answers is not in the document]
jelilat: In this case, our key is the lowerbound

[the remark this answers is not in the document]
omerfirmak: Lower bound is too pebble specific, we would have to implement that ourselves if we ever move to another db backend that doesnt support it.

db/db.go
```diff
@@ -34,10 +34,41 @@ type Entry struct {
 	Value []byte
 }
 
+type IterOptions struct {
```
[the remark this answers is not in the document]
omerfirmak: You can just call iter.Next right after initializing it in NewIterator and make sure it is Valid before returning to the caller

db/pebble/transaction.go
```diff
@@ -93,3 +97,49 @@ func (t *Transaction) Impl() any {
...
+}
+
+// Next : see db.Transaction.Iterator.Next
+func (it *pebbleIterator) Next() bool {
+	return it.iter.Next()
+}
+
+// Seek : see db.Transaction.Iterator.Seek
+func (it *pebbleIterator) Seek(key []byte) bool {
+	it.iter.SetBounds(key, nil)
```
omerfirmak, line 137: Let's not restrict the view of the Iterator by setting bounds, we could need a reverse Iterator at some point.

https://pkg.go.dev/github.com/cockroachdb/pebble?utm_source=godoc#Iterator.SeekGE should do fine

db/pebble/db_test.go
```diff
@@ -221,3 +222,94 @@ func TestConcurrentUpdate(t *testing.T) {
...
+			defer iter.Close()
+
+			iter.Seek([]byte{4})
+			assert.Nil(t, iter.Key())
+		})
+		return nil
+	})
+}
+
+type Entry struct {
```
omerfirmak, line 259: I dont think this is used anymore. There is also another `Entry` in `db.go`, we can remove both of them

omerfirmak (approved):

db/pebble/db_test.go
jelilat: It is actually being used [here](https://github.com/NethermindEth/juno/pull/572/files#diff-e9f741c3b54d069564c343dbbe5833e0e95207c50232497ce32d9a1decc31a07R296-R302). I've removed the one in `db.go`

db/db.go
```diff
@@ -28,16 +28,36 @@ type DB interface {
...
+	// key in lexicographical order
+	Seek(key []byte) bool
 }
 
 // Transaction provides an interface to access the database's state at the point the transaction was created
 // Updates done to the database with a transaction should be only visible to other newly created transaction after
 // the transaction is committed.
 type Transaction interface {
+	// NewIterator returns an iterator over the database's key/value pairs.
+	NewIterator() (Iterator, error)
```
IronGauntlets, line 60: Shouldn't this be defined on db interface instead of the transaction?

jelilat: Nah. The iterations are made on a transaction

omerfirmak: Database is volatile, can change any time. Transactions represent a fixed point in database's history. So, iterators make more sense on a fixed view of the database rather than a one that is possibly changing continuously .

IronGauntlets (approved):

PR #575, opened 2023-02-21
Fix a deadlock when sync shutdown and restart happens at the same time

omerfirmak:


jelilat (approved):
Should probably consider eliminating the code repetition. Maybe a `sendError` function?

sync/sync.go
```diff
@@ -74,14 +74,20 @@ func (s *Synchronizer) verifierTask(ctx context.Context, block *core.Block, stat
 		default:
 			if err != nil {
 				s.log.Warnw("Sanity checks failed", "number", block.Number, "hash", block.Hash.Text(16))
-				errChan <- ErrSyncFailed{block.Number, err}
+				select {
+				case <-ctx.Done():
+				case errChan <- ErrSyncFailed{block.Number, err}:
+				}
```
IronGauntlets, line 80: instead of this, you should close the error channel before returning on line 73 and read from the error channel on line 119.

IronGauntlets (changes requested):

sync/sync.go
IronGauntlets: So when I wrote the comment I thought everyone was getting their own error channel. But that is not the case. We are going to panic because we can't close a channel more than once and based on what I said that will happen. 

If there were panics, then let's keep the original code

omerfirmak: Let's keep it, it works:)

IronGauntlets (approved):

omerfirmak:
> Should probably consider eliminating the code repetition. Maybe a `sendError` function?

Yeah, most probably. I will tidy it up in the future, I just didnt want to leave a potential deadlock in the code.

PR #576, opened 2023-02-21
Refactor pebble iterator

IronGauntlets:
Rename pebbleIterator to iterator and move all its related code to iterator.go. Also, use pebble.iterator.ValueAndErr() since pebble .iterator.Value() is deprecated.

Use io.Closer interface instead of Close() function in Iterator interface.


omerfirmak (approved):
I will let @jelilat merge it

Issue #528, opened 2023-01-30
Refactor Trie Test


jelilat:
- Test names should not have "_"
- Table test are being used incorrectly: they should be separated into subsets, see node_test.go
- Functions such as `Path` and `FindCommonKey` don't need to be public. Thus, they don't need to be tested explicitly.
- There are warning which ignore returned errors, returned errors should not be ignored.
- Add more test cases with different heights
- Add more complicated Put and Delete scenarios

**Additional context**
https://github.com/NethermindEth/juno/blob/develop/core/trie/trie_test.go#L15-L22

liusanchuan:
@jelilat  Can you assign this to me?

jelilat:
Sure. Done

liusanchuan:
Hi @jelilat, can I split this issue into several PR commits?

liusanchuan:
Hi @IronGauntlets , I'm not very sure about the least todo item "Add more test cases with different heights", I plan to "add a max height to Trie", then test create trie with different height. I wonder, is this way meet your require?

IronGauntlets:
Hi @liusanchuan, 

First of all, thank you very much for your contributions!

What I meant by adding different heights is that we should have test cases that create trie of heights and check that the `Get`, `Put` and `Delete` operations work as expected, along with some edge cases. One of which is if you try to insert a key which is higher than the max value (calculated as `2^height -1`) there would be a panic. 

Adding max height is also required. If you would like you can work on these issues together. I believe #545 contains some work towards adding max height already

PR #577, opened 2023-02-22
change Path and FindCommonKey as private

liusanchuan:
Complete the TODO list for trie's test in #528

## Description

Complete the TODO list for trie's test.

## Changes:

- change Path and FindCommonKey to private,
- add more test to trie.Put and trie.FeltToBitSet,
- handle the ignored return error,
- replace all the t.Error with require.NoError,
- replace some improper assert by require,

## Types of changes

## Testing

**Requires testing**: Yes

## Documentation

### DIFF
--- a/core/trie/trie.go
+++ b/core/trie/trie.go
@@ -54,19 +54,19 @@ func RunOnTempTrie(height uint, do func(*Trie) error) error {
 	return do(NewTrie(newMemStorage(), height, nil))
 }
 
-// FeltToBitSet Converts a key, given in felt, to a bitset which when followed on a [Trie],
+// feltToBitSet Converts a key, given in felt, to a bitset which when followed on a [Trie],
 // leads to the corresponding [Node]
-func (t *Trie) FeltToBitSet(k *felt.Felt) *bitset.BitSet {
-	if k == (*felt.Felt)(nil) {
-		return (*bitset.BitSet)(nil)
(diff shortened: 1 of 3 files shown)


core/trie/trie.go
```diff
@@ -57,16 +57,16 @@ func RunOnTempTrie(height uint, do func(*Trie) error) error {
 // FeltToBitSet Converts a key, given in felt, to a bitset which when followed on a [Trie],
 // leads to the corresponding [Node]
 func (t *Trie) FeltToBitSet(k *felt.Felt) *bitset.BitSet {
```
IronGauntlets, line 59: We should make this private as well

liusanchuan: I think the `FeltToBitSet()` is a very useful tool function, otherwise, it's hard to convert a `Felt` to a `Bitset`, as we can't get the Trie's height from out side of Trie package.
And I also make it private in  [commit](https://github.com/NethermindEth/juno/pull/577/commits/1626544d6169f4677d6733166b16e7d74f025372).

IronGauntlets: Yes, however, it is not being used outside of the package anywhere. With regards to its utility, it would be better to make it a static function instead of a method on a trie struct. For our use case, this is not necessary.

liusanchuan: Ohh, I see

IronGauntlets (approved):

omerfirmak (approved):
Thanks!

PR #578, opened 2023-02-22
Class storage

joshklop:
## Description

Stores classes.

## Changes:

There are many places in the Starknet documentation and our codebase where terms like "Contracts", "Classes", "ContractDefinition", "ClassDefinition", "DeployedContract", "DeclaredContract", etc. are used interchangeably.  In addition to storing classes, this PR implements the following conventions:
 
- **Class**: the object that unambiguously defines a contract's semantics. Right now, this is sometimes called the "ContractDefinition" and "ClassDefinition".
- **Contract**: used to refer exclusively to an instance of a class. I.e., this contract's address maps to some storage in the state.
- Classes are **declared** (added to the state via a declare transaction), contracts are **deployed** (added to the state via a deploy transaction).

## Types of changes

## Testing

**Requires testing**: Yes

**Did you write tests??**: No (modified existing tests)

## Documentation

**If this requires a documentation update, did you add one?** Probably necessary to add comments, but none are present right now. It is low-priority to add them in this PR

### DIFF
--- a/blockchain/blockchain_test.go
+++ b/blockchain/blockchain_test.go
@@ -36,7 +36,7 @@ func TestNew(t *testing.T) {
 
 		testDB := pebble.NewMemTest()
 		chain := blockchain.New(testDB, utils.MAINNET)
-		assert.NoError(t, chain.Store(block0, stateUpdate0))
+		assert.NoError(t, chain.Store(block0, stateUpdate0, nil))
 
 		chain = blockchain.New(testDB, utils.MAINNET)
 		b, err := chain.Head()
@@ -62,7 +62,7 @@ func TestHeight(t *testing.T) {
 
 		testDB := pebble.NewMemTest()
 		chain := blockchain.New(testDB, utils.MAINNET)
-		assert.NoError(t, chain.Store(block0, stateUpdate0))
+		assert.NoError(t, chain.Store(block0, stateUpdate0, nil))
 
 		chain = blockchain.New(testDB, utils.MAINNET)
 		height, err := chain.Height()
@@ -82,7 +82,7 @@ func TestGetBlockByNumberAndHash(t *testing.T) {
 		update, err := gw.StateUpdate(context.Background(), 0)
 		require.NoError(t, err)
 
-		require.NoError(t, chain.Store(block, update))
+		require.NoError(t, chain.Store(block, update, nil))
 
 		storedByNumber, err := chain.GetBlockByNumber(block.Number)
 		require.NoError(t, err)
@@ -135,7 +135,7 @@ func TestVerifyBlock(t *testing.T) {
 	mainnetStateUpdate0, err := gw.StateUpdate(context.Background(), 0)
 	require.NoError(t, err)
 
-	require.NoError(t, chain.Store(mainnetBlock0, mainnetStateUpdate0))
+	require.NoError(t, chain.Store(mainnetBlock0, mainnetStateUpdate0, nil))
 
 	t.Run("error if difference between incoming block number and head is not 1",
 		func(t *testing.T) {
@@ -170,7 +170,7 @@ func TestSanityCheckNewHeight(t *testing.T) {
 	mainnetStateUpdate0, err := gw.StateUpdate(context.Background(), 0)
 	require.NoError(t, err)
 
-	require.NoError(t, chain.Store(mainnetBlock0, mainnetStateUpdate0))
+	require.NoError(t, chain.Store(mainnetBlock0, mainnetStateUpdate0, nil))
 
 	t.Run("error when block hash does not match state update's block hash", func(t *testing.T) {
 		mainnetBlock1, err := gw.BlockByNumber(context.Background(), 1)
@@ -208,7 +208,7 @@ func TestStore(t *testing.T) {
 
 	t.Run("add block to empty blockchain", func(t *testing.T) {
 		chain := blockchain.New(pebble.NewMemTest(), utils.MAINNET)
-		require.NoError(t, chain.Store(block0, stateUpdate0))
+		require.NoError(t, chain.Store(block0, stateUpdate0, nil))
 
 		headBlock, err := chain.Head()
 		assert.NoError(t, err)
@@ -234,8 +234,8 @@ func TestStore(t *testing.T) {
 		require.NoError(t, err)
 
 		chain := blockchain.New(pebble.NewMemTest(), utils.MAINNET)
-		require.NoError(t, chain.Store(block0, stateUpdate0))
-		require.NoError(t, chain.Store(block1, stateUpdate1))
+		require.NoError(t, chain.Store(block0, stateUpdate0, nil))
+		require.NoError(t, chain.Store(block1, stateUpdate1, nil))
 
 		headBlock, err := chain.Head()
 		assert.NoError(t, err)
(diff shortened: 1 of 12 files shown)


sync/sync.go
```diff
@@ -57,15 +58,23 @@ func (s *Synchronizer) fetcherTask(ctx context.Context, height uint64, verifiers
 			if err != nil {
 				continue
 			}
+			declaredClasses := make(map[felt.Felt]*core.Class, len(stateUpdate.StateDiff.DeployedContracts))
+			for _, classHash := range stateUpdate.StateDiff.DeclaredClasses {
+				class, err := s.StarknetData.Class(ctx, classHash)
```
omerfirmak, line 63: Could we make DeclaredClasses in StateUpdate a map and move this to gatewayclient?

joshklop: I would love to do that, but it is a very invasive change. 

Since only the class hashes (not the full classes themselves) are present in the state update response from the gateway, we would need to send separate `get_class` requests for each hash. We could do that in two places:

1. Within the state update adapt function (not preferable: limits the caller's ability to parallelize requests and the behavior goes beyond the what the function name indicates).
2. In the `sync` package and pass the received classes into the `starknetdata.Gateway.StateUpdate` function. This leaks the gateway implementation to the `starknetdata` interface, which is what we're trying to avoid.

omerfirmak: > Within the state update adapt function (not preferable: limits the caller's ability to parallelize requests and the behavior goes beyond the what the function name indicates).

I would actually prefer this. Let's see what others think.

joshklop: That's fair. @IronGauntlets what do you think?

starknetdata/gateway/gateway.go
```diff
@@ -307,7 +307,7 @@ func (g *Gateway) StateUpdate(ctx context.Context, blockNumber uint64) (*core.St
 
 func AdaptStateUpdate(response *clients.StateUpdate) (*core.StateUpdate, error) {
 	stateDiff := new(core.StateDiff)
-	stateDiff.DeclaredContracts = response.StateDiff.DeclaredContracts
+	stateDiff.DeclaredClasses = response.StateDiff.DeclaredContracts
```
jelilat, line 310: I think the lingo here is `DeclaredContracts = Classes`. So, we either go with just `Classes` or `DeclaredContracts`

core/state.go
```diff
@@ -118,7 +119,7 @@ func (s *State) putStateStorage(state *trie.Trie) error {
 // updated if an error is encountered during the operation. If update's
 // old or new root does not match the state's old or new roots,
 // [ErrMismatchedRoot] is returned.
-func (s *State) Update(update *StateUpdate) error {
+func (s *State) Update(update *StateUpdate, declaredClasses map[felt.Felt]*Class) error {
```
jelilat, line 122: We actually need to create a new `DeclaredContract` (we can change the name) type because `Class` won't help us fulfill any RPC endpoint. See [here](https://playground.open-rpc.org/?uiSchema%5BappBar%5D%5Bui:splitView%5D=false&schemaUrl=https://raw.githubusercontent.com/starkware-libs/starknet-specs/master/api/starknet_api_openrpc.json&uiSchema%5BappBar%5D%5Bui:input%5D=false&uiSchema%5BappBar%5D%5Bui:darkMode%5D=true&uiSchema%5BappBar%5D%5Bui:examplesDropdown%5D=false).

We need the Abi (not present here), Entrypoints, and program(not here). See the [previous implementation](https://github.com/NethermindEth/juno/blob/jelilat/implement-class-storage/core/state_update.go#L37-L41). It also contains program compression and so on.

jelilat: The major purpose of `Class` type is to help us calculate the class hash.

sync/sync.go
jelilat: I prefer `1.` as well. 

I don't think the behaviour goes beyond what the function indicates. All we are doing is getting all the declared contracts (classes), which are part of the state update. That was what I did in the previous implementation. [Here](https://github.com/NethermindEth/juno/blob/jelilat/implement-class-storage/starknetdata/gateway/gateway.go#L334)

jelilat (changes requested):

core/state.go
joshklop: I agree that we need to modify the class struct to contain other information that users can optionally include in their database (such as ABIs and programs). However, I think that would be outside the scope of this PR, where we're just trying to store classes in the database.

sync/sync.go
IronGauntlets: I would also prefer `1`. However, we don't need to do it in this PR.

core/state.go
IronGauntlets: I agree with @joshklop. We can worry about how best to serve Classes through the rpc in the next phase.

blockchain/blockchain_test.go
```diff
@@ -35,7 +37,8 @@ func TestNewBlockchain(t *testing.T) {
 
 		testDB := pebble.NewMemTest()
 		chain := NewBlockchain(testDB, utils.MAINNET)
-		assert.NoError(t, chain.Store(block0, stateUpdate0))
+		// state update zero has no declared contracts
+		assert.NoError(t, chain.Store(block0, stateUpdate0, emptyDeclaredContracts))
```
IronGauntlets, line 41: Why not pass nil?

sync/sync.go
```diff
@@ -57,15 +58,23 @@ func (s *Synchronizer) fetcherTask(ctx context.Context, height uint64, verifiers
 			if err != nil {
 				continue
 			}
+			declaredClasses := make(map[felt.Felt]*core.Class, len(stateUpdate.StateDiff.DeployedContracts))
```
IronGauntlets, line 61:
```go
len(stateUpdate.StateDiff.DeployedContracts) != len(stateUpdate.StateDiff.DeclaredClasses)
```

Therefore, just make it the following:
```suggestion
			declaredClasses := make(map[felt.Felt]*core.Class)
```

core/state_test.go
```diff
@@ -169,7 +171,8 @@ func TestUpdate(t *testing.T) {
 	testDb := pebble.NewMemTest()
 	state := core.NewState(testDb.NewTransaction(true))
 
-	assert.Equal(t, nil, state.Update(coreUpdate))
+	// coreUpdate has no declared contracts
+	assert.Equal(t, nil, state.Update(coreUpdate, emptyDeclaredContracts))
```
IronGauntlets, line 175: I think passing nil would work just fine. Please update all the tests accordingly.

IronGauntlets (changes requested):

starknetdata/gateway/gateway.go
joshklop: In this case, I think it is preferable to emphasize that these classes are declared in this state diff. Since classes are declared and contracts are deployed, `DeclaredClasses` seems to be a fitting name.

sync/sync.go
joshklop: Thanks! I think that was a typo on my end. It should be `len(stateUpdate.StateDiff.DeclaredClasses)`. Changed in most recent rebase.

sync/sync.go
joshklop: Thanks everyone for the feedback. I think I'll handle this in another PR.

sync/sync_test.go
```diff
@@ -57,8 +57,8 @@ func TestSyncBlocks(t *testing.T) {
 		b0, err := gw.BlockByNumber(context.Background(), 0)
 		require.NoError(t, err)
 		s0, err := gw.StateUpdate(context.Background(), 0)
-		require.NoError(t, err)
-		require.NoError(t, bc.Store(b0, s0))
+		assert.NoError(t, err)
```
IronGauntlets, line 60: This line should still be `require`. Otherwise, the next line would panic if s0 is nil.

IronGauntlets (approved):
LGTM!

jelilat (approved):

PR #511, opened 2023-01-18
First round of RPC support

omerfirmak:

PR #582, opened 2023-02-24
Implement JSONRPC server

omerfirmak:
Implements the server and a HTTP transport layer

extracted from #511 to make it easier to review

### DIFF
--- a/jsonrpc/server.go
+++ b/jsonrpc/server.go
@@ -1,0 +1,342 @@
+// Package jsonrpc implements a JSONRPC2.0 compliant server as described in https://www.jsonrpc.org/specification
+package jsonrpc
+
+import (
+	"bufio"
+	"bytes"
+	"encoding/json"
+	"errors"
+	"io"
+	"reflect"
+	"strings"
+)
+
+// Todo: add rpcErr type which implements Error() to return short string representation of the error. For example:
+//   - ErrInvalidJson.Error() = "Parse error"
+//   - ErrInvalidRequest.Error() = "Invalid Request"
+//   - ErrMethodNotFound.Error() = "Method Not Found"
+//   - ErrInvalidParams.Error() = "Invalid Params"
+//   - ErrInternal.Error() = "Internal Error"
+//
+// More contextual information can then be passed through Data field of Error as follows:
+//
+//	Error {
+//	  Code: ErrInvalidRequest,
+//	  Message: ErrInvalidRequest.Error(),
+//	  Data: err.Error(), // These are golang errors to provide more information
+//	}
+
+const (
+	InvalidJson    = -32700 // Invalid JSON was received by the server.
+	InvalidRequest = -32600 // The JSON sent is not a valid Request object.
+	MethodNotFound = -32601 // The method does not exist / is not available.
+	InvalidParams  = -32602 // Invalid method parameter(s).
+	InternalError  = -32603 // Internal JSON-RPC error.
+)
+
+type request struct {
+	Version string `json:"jsonrpc"`
+	Method  string `json:"method"`
+	Params  any    `json:"params,omitempty"`
+	Id      any    `json:"id,omitempty"`
+}
+
+type response struct {
+	Version string `json:"jsonrpc"`
+	Result  any    `json:"result,omitempty"`
+	Error   *Error `json:"error,omitempty"`
+	Id      any    `json:"id"`
+}
+
+type Error struct {
+	Code    int    `json:"code"`
+	Message string `json:"message"`
+	Data    any    `json:"data,omitempty"`
+}
+
+func (r *request) isSane() error {
+	if r.Version != "2.0" {
+		return errors.New("unsupported RPC request version")
+	}
+	if len(r.Method) <= 0 {
+		return errors.New("no method specified")
+	}
+
+	if r.Params != nil {
+		paramType := reflect.TypeOf(r.Params)
+		if paramType.Kind() != reflect.Slice && paramType.Kind() != reflect.Map {
+			return errors.New("params should be an array or an object")
+		}
+	}
+
+	if r.Id != nil {
+		idType := reflect.TypeOf(r.Id)
+		floating := idType.Name() == "Number" && strings.Contains(r.Id.(json.Number).String(), ".")
+		if (idType.Kind() != reflect.String && idType.Name() != "Number") || floating {
+			return errors.New("id should be a string or an integer")
+		}
+	}
+
+	return nil
+}
+
+type Parameter struct {
+	Name     string
+	Optional bool
+}
+
+type Method struct {
+	Name    string
+	Params  []Parameter
+	Handler any
+}
+
+type Server struct {
+	methods map[string]Method
+}
+
+// NewServer instantiates a JSONRPC server
+func NewServer() *Server {
+	return &Server{
+		methods: make(map[string]Method),
+	}
+}
+
+// RegisterMethod verifies and creates an endpoint that the server recognizes.
+//
+// - name is the method name
+// - handler is the function to be called when a request is received for the
+// associated method. It should have (any, *jsonrpc.Error) as its return type
+// - paramNames are the names of parameters in the order that they are expected
+// by the handler
+func (s *Server) RegisterMethod(method Method) error {
+	handlerT := reflect.TypeOf(method.Handler)
+	if handlerT.Kind() != reflect.Func {
+		return errors.New("handler must be a function")
+	}
+	if handlerT.NumIn() != len(method.Params) {
+		return errors.New("number of function params and param names must match")
+	}
+	if handlerT.NumOut() != 2 {
+		return errors.New("handler must return 2 values")
+	}
+	if handlerT.Out(1) != reflect.TypeOf(&Error{}) {
+		return errors.New("second return value must be a *jsonrpc.Error")
+	}
+
+	s.methods[method.Name] = method
+
+	return nil
+}
+
+// Handle processes a request to the server
+// It returns the response in a byte array, only returns an
+// error if it can not create the response byte array
+func (s *Server) Handle(data []byte) ([]byte, error) {
+	return s.HandleReader(bytes.NewReader(data))
+}
+
+// HandleReader processes a request to the server
+// It returns the response in a byte array, only returns an
+// error if it can not create the response byte array
+func (s *Server) HandleReader(reader io.Reader) ([]byte, error) {
+	bufferedReader := bufio.NewReader(reader)
+	requestIsBatch := isBatch(bufferedReader)
+	res := &response{
+		Version: "2.0",
+	}
+
+	dec := json.NewDecoder(bufferedReader)
+	dec.UseNumber()
+
+	if !requestIsBatch {
+		req := new(request)
+		if jsonErr := dec.Decode(req); jsonErr != nil {
+			res.Error = &Error{Code: InvalidJson, Message: jsonErr.Error()}
+		} else if resObject, handleErr := s.handleRequest(req); handleErr != nil {
+			res.Error = &Error{Code: InvalidRequest, Message: handleErr.Error()}
+		} else {
+			res = resObject
+		}
+	} else {
+		var batchReq []json.RawMessage
+		var batchRes []json.RawMessage
+
+		if batchJsonErr := dec.Decode(&batchReq); batchJsonErr != nil {
+			res.Error = &Error{Code: InvalidJson, Message: batchJsonErr.Error()}
+		} else if len(batchReq) == 0 {
+			res.Error = &Error{Code: InvalidRequest, Message: "empty batch"}
+		} else {
+			for _, rawReq := range batchReq { // todo: handle async
+				var resObject *response
+
+				reqDec := json.NewDecoder(bytes.NewBuffer(rawReq))
+				reqDec.UseNumber()
+
+				req := new(request)
+				if jsonErr := reqDec.Decode(req); jsonErr != nil {
+					resObject = &response{
+						Version: "2.0",
+						Error:   &Error{Code: InvalidRequest, Message: jsonErr.Error()},
+					}
+				} else {
+					var handleErr error
+					resObject, handleErr = s.handleRequest(req)
+					if handleErr != nil {
+						resObject = &response{
+							Version: "2.0",
+							Error:   &Error{Code: InvalidRequest, Message: handleErr.Error()},
+						}
+					}
+				}
+
+				if resObject != nil {
+					if resArr, jsonErr := json.Marshal(resObject); jsonErr != nil {
+						return nil, jsonErr
+					} else {
+						batchRes = append(batchRes, resArr)
+					}
+				}
+			}
+
+			if len(batchRes) == 0 {
+				return nil, nil
+			}
+			return json.Marshal(batchRes)
+		}
+	}
+
+	if res == nil {
+		return nil, nil
+	}
+	return json.Marshal(res)
+}
+
+func isBatch(reader *bufio.Reader) bool {
+	for {
+		char, err := reader.Peek(1)
+		if err != nil {
+			break
+		} else if char[0] == ' ' || char[0] == '\t' || char[0] == '\r' || char[0] == '\n' {
+			if discarded, err := reader.Discard(1); discarded != 1 || err != nil {
+				break
+			}
+			continue
+		} else {
+			return char[0] == '['
+		}
+	}
+
+	return false
+}
+
+func isNil(i any) bool {
+	return i == nil || reflect.ValueOf(i).IsNil()
+}
+
+func (s *Server) handleRequest(req *request) (*response, error) {
+	if err := req.isSane(); err != nil {
+		return nil, err
+	}
+
+	res := &response{
+		Version: "2.0",
+		Id:      req.Id,
+	}
+
+	calledMethod, found := s.methods[req.Method]
+	if !found {
+		res.Error = &Error{
+			Code:    MethodNotFound,
+			Message: "method not found",
+		}
+		return res, nil
+	}
+
+	args, err := buildArguments(req.Params, calledMethod.Handler, calledMethod.Params)
+	if err != nil {
+		res.Error = &Error{
+			Code:    InvalidParams,
+			Message: err.Error(),
+		}
+		return res, nil
+	}
+
+	tuple := reflect.ValueOf(calledMethod.Handler).Call(args)
+	if res.Id == nil { // notification
+		return nil, nil
+	}
+
+	if errAny := tuple[1].Interface(); !isNil(errAny) {
+		res.Error = errAny.(*Error)
+		return res, nil
+	}
+
+	res.Result = tuple[0].Interface()
+	return res, nil
+}
+
+func buildArguments(params, handler any, configuredParams []Parameter) ([]reflect.Value, error) {
+	var args []reflect.Value
+	if isNil(params) {
+		return args, nil
+	}
+
+	handlerType := reflect.TypeOf(handler)
+
+	handlerParamValue := func(param any, t reflect.Type) (reflect.Value, error) {
+		handlerParam := reflect.New(t)
+		valueMarshaled, err := json.Marshal(param) // we have to marshal the value into JSON again
+		if err != nil {
+			return reflect.ValueOf(nil), err
+		}
+		err = json.Unmarshal(valueMarshaled, handlerParam.Interface())
+		if err != nil {
+			return reflect.ValueOf(nil), err
+		}
+
+		return handlerParam.Elem(), nil
+	}
+
+	switch reflect.TypeOf(params).Kind() {
+	case reflect.Slice:
+		paramsList := params.([]any)
+
+		if len(paramsList) != handlerType.NumIn() {
+			return nil, errors.New("missing param in list")
+		}
+
+		for i, param := range paramsList {
+			v, err := handlerParamValue(param, handlerType.In(i))
+			if err != nil {
+				return nil, err
+			}
+			args = append(args, v)
+		}
+		return args, nil
+	case reflect.Map:
+		paramsMap := params.(map[string]any)
+
+		for i, configuredParam := range configuredParams {
+			var v reflect.Value
+			if param, found := paramsMap[configuredParam.Name]; found {
+				var err error
+				v, err = handlerParamValue(param, handlerType.In(i))
+				if err != nil {
+					return nil, err
+				}
+			} else if configuredParam.Optional {
+				// optional parameter
+				v = reflect.New(handlerType.In(i)).Elem()
+			} else {
+				return nil, errors.New("missing non-optional param")
+			}
+
+			args = append(args, v)
+		}
+		return args, nil
+	default:
+		// Todo: return InternalError
+		return nil, errors.New("impossible param type: check request.isSane")
+	}
+}
--- a/jsonrpc/server_test.go
+++ b/jsonrpc/server_test.go
@@ -1,0 +1,333 @@
+package jsonrpc_test
+
+import (
+	"testing"
+
+	"github.com/NethermindEth/juno/jsonrpc"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestServer_RegisterMethod(t *testing.T) {
+	server := jsonrpc.NewServer()
+	tests := map[string]struct {
+		handler    any
+		paramNames []jsonrpc.Parameter
+		want       string
+	}{
+		"not a func handler": {
+			handler: 44,
+			want:    "handler must be a function",
+		},
+		"excess param names": {
+			handler:    func() {},
+			paramNames: []jsonrpc.Parameter{{Name: "param1"}},
+			want:       "number of function params and param names must match",
+		},
+		"missing param names": {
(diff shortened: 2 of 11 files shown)


jsonrpc/server_test.go
```diff
@@ -0,0 +1,147 @@
...
+		"no error return": {
+			handler: func(param1, param2 int) (any, int) { return 0, 0 },
+			want:    "second return value must be a *jsonrpc.Error",
+		},
+	}
+
+	for desc, test := range tests {
+		err := server.RegisterMethod("method", test.paramNames, test.handler)
+		assert.EqualError(t, err, test.want, desc)
+	}
```
IronGauntlets, line 49: Please create sub-tests with `desc` as their name.

jsonrpc/server_test.go
```diff
@@ -0,0 +1,147 @@
...
+
+	err := server.RegisterMethod("method", nil, func(param1, param2 int) (int, *jsonrpc.Error) { return 0, nil })
+	assert.NoError(t, err)
```
IronGauntlets, line 52: Please name this test case and add it as a sub test.

jsonrpc/server_test.go
```diff
@@ -0,0 +1,147 @@
...
+			}{*num * 2}, nil
+		}))
+
+	tests := []struct {
+		req string
+		res string
+	}{
+		{
+			req: `{]`,
+			res: `{"jsonrpc":"2.0","error":{"code":-32600,"message":"invalid character ']' looking for beginning of object key string"},"id":null}`,
```
IronGauntlets, line 73: This test case should return `-32700` which is for invalid json. `-32600` is for well-formed json but an invalid request.

jsonrpc/server.go
```diff
@@ -0,0 +1,283 @@
...
+}
+
+// RegisterMethod verifies and creates and endpoint that server recognizes.
+//
+// - name is the method name
+// - handler is the function to be called when a request is received for the
+// associated method. It should have (any, error) as it's return type
+// - paramNames are the names of parameters in the order that they are expected
+// by the handler
+func (s *Server) RegisterMethod(name string, params []Parameter, handler any) error {
```
IronGauntlets, line 111: According to the comment you should declare the following type:

```go
type HandlerFn func(args ...any) (any, error)
```

You can unexport it if needed. Then change the signature of the function as follows:
```suggestion
func (s *Server) RegisterMethod(name string, params []Parameter, handler HandlerFn) error {
```

This way you can reduce the use of reflect package here.

jsonrpc/server.go
```diff
@@ -0,0 +1,283 @@
...
+	handlerT := reflect.TypeOf(handler)
+	if handlerT.Kind() != reflect.Func {
+		return errors.New("handler must be a function")
+	}
+	if len(params) > 0 && handlerT.NumIn() != len(params) {
+		return errors.New("number of function params and param names must match")
+	}
```
IronGauntlets, line 118: parameters are optional so why are we ensuring `len(params) > 0`?

jsonrpc/server_test.go
```diff
@@ -0,0 +1,147 @@
...
+			res: `{"jsonrpc":"2.0","error":{"code":44,"message":""},"id":22}`,
+		},
+		{
+			req: `{"jsonrpc" : "2.0", "method" : "method",
+					"params" : { "shouldError" : true } , "id" : 22}`,
+			res: `{"jsonrpc":"2.0","error":{"code":-32602,"message":"missing non-optional param"},"id":22}`,
+		},
+		{
+			req: `[]`,
+			res: `[]`,
```
IronGauntlets, line 117: This is an incorrect test case. The specification says the following in the last paragraph of section 6: 
```
If there are no Response objects contained within the Response array as it is to be sent
to the client, the server MUST NOT return an empty Array and should return nothing at all.
``` 

Nothing should be returned to the client.

jsonrpc/server_test.go
```diff
@@ -0,0 +1,147 @@
...
+	server := jsonrpc.NewServer()
+	require.NoError(t, server.RegisterMethod("method", []jsonrpc.Parameter{{Name: "num"}, {Name: "shouldError", Optional: true}, {Name: "msg", Optional: true}},
+		func(num *int, shouldError bool, msg string) (any, *jsonrpc.Error) {
+			if shouldError {
+				return nil, &jsonrpc.Error{Code: 44, Message: msg}
+			}
+			return struct {
+				Doubled int `json:"doubled"`
+			}{*num * 2}, nil
+		}))
```
IronGauntlets, line 65: Please separate this into a sub test case with a name

jsonrpc/server_test.go
```diff
@@ -0,0 +1,147 @@
...
+					"params" : { "num" : 5 } , "id" : 5},
+					{"jsonrpc" : "2.0", "method" : "fail",
+					"params" : { "num" : 5 } , "id" : 7},
+					{"jsonrpc" : "2.0", "method" : "method",
+					"params" : { "num" : 44 } , "id" : 6}]`,
+			res: `[{"jsonrpc":"2.0","result":{"doubled":10},"id":5},{"jsonrpc":"2.0","error":{"code":-32601,"message":"method not found"},"id":7},{"jsonrpc":"2.0","result":{"doubled":88},"id":6}]`,
+		},
+	}
+
+	for _, test := range tests {
```
IronGauntlets, line 142: Please separate this into sub test with names.

jsonrpc/server_test.go
```diff
@@ -0,0 +1,147 @@
...
+		res, err := server.Handle([]byte(test.req))
+		assert.NoError(t, err)
```
IronGauntlets, line 144:
```suggestion
		require.NoError(t, err)
```

Otherwise, there will be a panic on the next line is `res` is nil.

IronGauntlets (changes requested):

jsonrpc/server.go
omerfirmak: We are not ensuring that. We are simply checking that the handler accepts that many arguments; only if user provides params while registering

jsonrpc/server_test.go
omerfirmak: yeap, fixed

jsonrpc/server.go
omerfirmak: I tried to do something similar but sadly it does not work;

``` golang
type HandlerFn func(args ...any) (any, *Error)


node/node.go:91:30: cannot use rpcHandler.ChainId (value of type func() (*felt.Felt, *jsonrpc.Error)) as jsonrpc.HandlerFn value in struct literal
node/node.go:92:34: cannot use rpcHandler.BlockNumber (value of type func() (uint64, *jsonrpc.Error)) as jsonrpc.HandlerFn value in struct literal
node/node.go:93:41: cannot use rpcHandler.BlockNumberAndHash (value of type func() (*rpc.BlockNumberAndHash, *jsonrpc.Error)) as jsonrpc.HandlerFn value in struct literal
node/node.go:94:79: cannot use rpcHandler.GetBlockWithTxHashes (value of type func(id *rpc.BlockId) (*rpc.BlockWithTxHashes, *jsonrpc.Error)) as jsonrpc.HandlerFn value in struct literal
node/node.go:95:74: cannot use rpcHandler.GetBlockWithTxs (value of type func(id *rpc.BlockId) (*rpc.BlockWithTxs, *jsonrpc.Error)) as jsonrpc.HandlerFn value in struct literal
node/node.go:96:87: cannot use rpcHandler.GetTransactionByHash (value of type func(hash *felt.Felt) (*rpc.Transaction, *jsonrpc.Error)) as jsonrpc.HandlerFn value in struct literal
```

comment is outdated tho, updating

jsonrpc/server_test.go
omerfirmak: that is not a testcase, hence the `require`. Added descriptions to following table tests.

jsonrpc/server_test.go
omerfirmak: Thanks, fixed

omerfirmak:
Changes according to comments https://github.com/NethermindEth/juno/compare/ac2c4b8f7325de759e93f4accdbb8fbc8aeb4ae7..32946af23707bff666eaeaddc480f97a2098e522

jsonrpc/server.go
```diff
@@ -0,0 +1,284 @@
...
+		return nil, &Error{Code: InvalidJson, Message: err.Error()}
+	} else if err = req.isSane(); err != nil {
+		return nil, &Error{Code: InvalidRequest, Message: err.Error()}
+	}
+	return req, nil
+}
+
+func (r *request) isSane() error {
+	if r.Version != "2.0" {
+		return errors.New(fmt.Sprintf("unsupported RPC request version"))
```
jelilat, line 59:
```suggestion
		return errors.New("unsupported RPC request version")
```

jelilat: Or probably this 

```suggestion
		return fmt.Errorf("unsupported RPC request version: %s", r.Version)
```

jsonrpc/server.go
```diff
@@ -0,0 +1,284 @@
...
+	if handlerT.Kind() != reflect.Func {
+		return errors.New("handler must be a function")
+	}
+	if len(params) > 0 && handlerT.NumIn() != len(params) {
+		return errors.New("number of function params and param names must match")
+	}
+	if handlerT.NumOut() != 2 {
+		return errors.New("handler must return 2 values")
+	}
+	if handlerT.Out(1) != reflect.TypeOf((*Error)(nil)) {
```
jelilat, line 123:
```suggestion
	if handlerT.Out(1) != reflect.TypeOf(&Error{}) {
```

jelilat (changes requested):

jsonrpc/server.go
omerfirmak: done

jsonrpc/server.go
omerfirmak: done

rpc/types.go
```diff
@@ -0,0 +1,116 @@
...
+	Status BlockStatus `json:"status"`
+	BlockHeader
+	Transactions []*Transaction `json:"transactions"`
+}
+
+// https://github.com/starkware-libs/starknet-specs/blob/a789ccc3432c57777beceaa53a34a7ae2f25fda0/api/starknet_api_openrpc.json#L1109
+type BlockWithTxHashes struct {
+	Status BlockStatus `json:"status"`
+	BlockHeader
+	TxnHashes []*felt.Felt `json:"transactions"`
```
jelilat, line 115: I think this should be `json:transaction_hashes`

omerfirmak: Nope, that is how spec names that field.

rpc/handlers_test.go
```diff
@@ -0,0 +1,299 @@
...
+		assert.Equal(t, gwBlock.SequencerAddress, latestRpc.SequencerAddress)
+		assert.Equal(t, gwBlock.Timestamp, latestRpc.Timestamp)
+		for i := 0; i < len(gwBlock.Transactions); i++ {
+			assert.Equal(t, gwBlock.Transactions[i].Hash(), latestRpc.TxnHashes[i])
+		}
+
+		byHash, err := handler.GetBlockWithTxHashes(&rpc.BlockId{Hash: latestRpc.Hash})
+		assert.Equal(t, latestRpc, byHash)
+		byNumber, err := handler.GetBlockWithTxHashes(&rpc.BlockId{Number: latestRpc.Number})
+		assert.Equal(t, latestRpc, byNumber)
```
jelilat, line 93: We should probably handle the errors here

rpc/types.go
```diff
@@ -0,0 +1,116 @@
...
+	Hash   *felt.Felt `json:"block_hash"`
+}
+
+// https://github.com/starkware-libs/starknet-specs/blob/a789ccc3432c57777beceaa53a34a7ae2f25fda0/api/starknet_api_openrpc.json#L814
+type BlockId struct {
+	Pending bool
+	Latest  bool
+	Hash    *felt.Felt
+	Number  uint64
+}
```
jelilat, line 22: Was thinking we should add `json` tag here as well, but not sure if it's intentional

```

type BlockId struct {
	Pending bool       `json:"pending,omitempty"`
	Latest  bool       `json:"latest,omitempty"`
	Hash    *felt.Felt `json:"hash,omitempty"`
	Number  uint64     `json:"number,omitempty"`
}
```

But the tests seem to be working fine after adding it

omerfirmak: Yeap, no tags was intentional. BlockId has custom unmarshaler. Check out how blockid is defined in the spec from the link above it.

jelilat: Right. Got it.

jelilat (approved):

jsonrpc/server.go
IronGauntlets: I tried playing around with it as well and apparently, the types need to match exactly.

jsonrpc/server.go
```diff
@@ -0,0 +1,283 @@
...
+}
+
+// RegisterMethod verifies and creates an endpoint that the server recognizes.
+//
+// - name is the method name
+// - handler is the function to be called when a request is received for the
+// associated method. It should have (any, *jsonrpc.Error) as its return type
+// - paramNames are the names of parameters in the order that they are expected
+// by the handler
+func (s *Server) RegisterMethod(name string, params []Parameter, handler any) error {
```
IronGauntlets, line 111: Is there a reason why the declaration is not the following:
```suggestion
func (s *Server) RegisterMethod(m Method) error {
```

jsonrpc/server.go
```diff
@@ -0,0 +1,283 @@
...
+		} else {
+			reader.UnreadByte()
+			return char == '['
+		}
+	}
+
+	return false
+}
+
+func isNil(i interface{}) bool {
```
IronGauntlets, line 198:
```suggestion
func isNil(i any) bool {
```

jsonrpc/server.go
omerfirmak: Probably because `RegisterMethod` predates `Method`, will change

jsonrpc/server_pkg_test.go
```diff
@@ -0,0 +1,312 @@
...
+	"bytes"
+	"io"
+	"reflect"
+	"testing"
+
+	"github.com/NethermindEth/juno/core/felt"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
```
IronGauntlets, line 14: All of the following tests can be moved to `server_test.go`. These private functions can all be tested through the `HandleRequest()`. We should not be testing private functions unless we require a complicated setup. You are already defining handlers and incoming requests so I don't see why they cannot be moved to `server_test.go`.

Also, all these tests require names and should be run as sub-tests.

IronGauntlets (changes requested):

jsonrpc/server_pkg_test.go
omerfirmak: Those were the tests I used while developing that functions, it will take some time to make sure we dont lose coverage on the edge cases while moving them.

IronGauntlets: You can note the coverage percentage before moving and then make sure it is the same after moving them. Goland can easily help with that.

IronGauntlets: I just ran it with jsonrpc package with  coverage and it sits at 94.4%

jsonrpc/server_pkg_test.go
```diff
@@ -0,0 +1,312 @@
...
+	tests := []struct {
+		handler    any
+		req        string
+		shouldFail bool
+		errorMsg   string
+	}{
+		{
+			handler: func(param1 int, param2 *arg) (any, error) {
+				assert.Equal(t, 44, param1)
```
IronGauntlets, line 159: Once you run these tests as subtests please make sure `t` refers to the correct test. Currently, `t` refers to the parent tests.

node/node.go
```diff
@@ -97,6 +110,7 @@ func (n *Node) Run(ctx context.Context) (err error) {
 		<-ctx.Done()
 		n.log.Infow("Shutting down Juno...")
 	}()
+	n.http.Run(ctx)
 	return n.synchronizer.Run(ctx)
```
jelilat, line 114: I think this should help us manage the http errors

```suggestion
	syncErrChan := make(chan error)
	httpErrChan := make(chan error)

	go func() {
		syncErrChan <- n.synchronizer.Run(ctx)
	}()
	go func() {
		httpErrChan <- n.http.Run(ctx)
	}()

	syncErr := <-syncErrChan
	httpErr := <-httpErrChan

	if syncErr != nil {
		return syncErr
	}
	if httpErr != nil {
		return httpErr
	}
	return err
```

omerfirmak: We dont want Run to block tho

omerfirmak: Run should only return an error if the service fails to start, so there is no reason to push them thru a channel because it will return immediately.

Any errors that the service may encounter in runtime should be handled within that service.

omerfirmak: There is not much the Node can do if http or synchronizer fails anyways

rpc/types.go
```diff
@@ -0,0 +1,116 @@
+package rpc
+
+import (
+	"encoding/json"
+	"errors"
+
+	"github.com/NethermindEth/juno/core/felt"
+)
+
```
IronGauntlets, line 9: Please separate these types into their own files so it is easier to quickly identify which file contains what.

rpc/types_test.go
```diff
@@ -0,0 +1,76 @@
...
+			},
+		},
+		"hash": {
+			blockIdJson: `{ "block_hash" : "0x123" }`,
+			expectedBlockId: rpc.BlockId{
+				Hash: new(felt.Felt).SetUint64(0x123),
+			},
+		},
+	}
+	t.Parallel()
```
IronGauntlets, line 42: I believe this usage is incorrect you should move it to line 45. unless this was intentional.

rpc/handlers_test.go
```diff
@@ -0,0 +1,299 @@
...
+	canceler()
+	<-syncNodeChan
+}
+
+// implements only GetTransactionByHash from BlockchainReader interface
+// calling any other function from BlockchainReader will panic
+type txGwReader struct {
+	*blockchain.Blockchain
+	gw starknetdata.StarknetData
+}
```
IronGauntlets, line 121: Please rename this to fake blockchain reader

jsonrpc/server.go
```diff
@@ -0,0 +1,279 @@
...
+			}
+		} else if !configuredParams[idx].Optional {
+			return nil, errors.New("missing non-optional param")
+		}
+
+		args = append(args, valueContainer.Elem())
+	}
+
+	return args, nil
+}
```
IronGauntlets, line 279: Some comments would be helpful. It took me a while to understand what was happening here.

node/node.go
```diff
@@ -76,12 +80,21 @@ func New(cfg *Config) (StarknetNode, error) {
...
 		blockchain:   chain,
 		synchronizer: synchronizer,
+		http: jsonrpc.NewHttp(cfg.RpcPort, []jsonrpc.Method{
+			{"starknet_chainId", nil, rpcHandler.ChainId},
+			{"starknet_blockNumber", nil, rpcHandler.BlockNumber},
+			{"starknet_blockHashAndNumber", nil, rpcHandler.BlockNumberAndHash},
+			{"starknet_getBlockWithTxHashes", []jsonrpc.Parameter{{Name: "block_id"}}, rpcHandler.GetBlockWithTxHashes},
+			{"starknet_getBlockWithTxs", []jsonrpc.Parameter{{Name: "block_id"}}, rpcHandler.GetBlockWithTxs},
+			{"starknet_getTransactionByHash", []jsonrpc.Parameter{{Name: "transaction_hash"}}, rpcHandler.GetTransactionByHash},
+		}),
```
IronGauntlets, line 97: registration of methods should be extracted into a function

IronGauntlets (changes requested):

jsonrpc/server_test.go
```diff
@@ -0,0 +1,153 @@
...
+		t.Run(desc, func(t *testing.T) {
+			err := server.RegisterMethod(jsonrpc.Method{"method", test.paramNames, test.handler})
+			assert.EqualError(t, err, test.want, desc)
+		})
+	}
+
+	t.Run("should not fail", func(t *testing.T) {
+		err := server.RegisterMethod(jsonrpc.Method{"method", nil, func(param1, param2 int) (int, *jsonrpc.Error) { return 0, nil }})
+		assert.NoError(t, err)
+	})
```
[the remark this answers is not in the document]
omerfirmak: Probably not intentional but is valid. If the server only means to use list parameters it can omit parameter names.

jsonrpc/server.go
```diff
@@ -0,0 +1,279 @@
...
+		if paramType.Kind() != reflect.Slice && paramType.Kind() != reflect.Map {
+			return errors.New("params should be an array or an object")
+		}
+	}
+
+	if r.Id != nil {
+		idType := reflect.TypeOf(r.Id)
+		floating := idType.Name() == "Number" && strings.Contains(r.Id.(json.Number).String(), ".")
+		if (idType.Kind() != reflect.String && idType.Name() != "Number") || floating {
+			return errors.New("id should be a string or an integer")
```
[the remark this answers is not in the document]
omerfirmak: we allow `nil` id, see L71

jsonrpc/server.go
```diff
@@ -0,0 +1,279 @@
...
+	Version string `json:"jsonrpc"`
+	Result  any    `json:"result,omitempty"`
+	Error   *Error `json:"error,omitempty"`
+	Id      any    `json:"id"`
+}
+
+type Error struct {
+	Code    int    `json:"code"`
+	Message string `json:"message"`
+	Data    any    `json:"data,omitempty"`
```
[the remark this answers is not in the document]
omerfirmak: This is a generic JSONRPC package, so starknet related comments shouldnt go here.

`Data` is produced by handlers and server just passes them along in the response. There is not much it can do with it anyways.

jsonrpc/server.go
```diff
@@ -0,0 +1,279 @@
...
+			if len(paramsList) != paramCount {
+				return nil, errors.New("missing param in list")
+			}
+
+			requestValue = paramsList[idx]
+		case reflect.Map:
+			paramsMap := params.(map[string]any)
+			if len(configuredParams) != paramCount {
+				return nil, errors.New("missing param name")
+			}
```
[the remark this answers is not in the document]
omerfirmak: > Why would we allow registration of a method whose handler type input length differs from the length of parameters the caller provided?

yeah, I dont think there is a reason we should, removed

jsonrpc/server_test.go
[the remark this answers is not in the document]
omerfirmak: ^ removed this feature

jsonrpc/server.go
```diff
@@ -0,0 +1,279 @@
...
+	} else {
+		batchReq := batch{}
+		batchRes := []json.RawMessage{}
+
+		if err := json.NewDecoder(bufferedReader).Decode(&batchReq); err != nil {
+			return nil, err
+		}
+
+		for _, rawReq := range batchReq {
+			if res, err := s.Handle(rawReq); err == nil { // todo: handle async
```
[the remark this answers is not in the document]
omerfirmak: great catch, thanks

rpc/types.go
omerfirmak: What should the separation be based on? Different endpoints might share same data types.

jsonrpc/server_pkg_test.go
omerfirmak: done

jsonrpc/server.go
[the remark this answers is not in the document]
omerfirmak: ended up being a big refactor

IronGauntlets (approved):
Please merge #594 before closing this pr

omerfirmak:
I cherry-picked and squashed first and last commit of #594
I really dont like using a wait group and explicitly handling each service we start, I propose we leave it to another PR so that we can discuss it further.

IronGauntlets:
>I am reluctant to merge the second one, I propose we leave it to another PR so that we can discuss it further.

There was a possibility of go-routine leaks in jsonrpc.http.Run() functions and the documentation of Serve() mentioned that the application should not exit without Shutdown() returning. We relied on the fact that synchroniser will take longer to exit that jsonrpc.http.Run()

IronGauntlets:
Happy for you to merge btw

omerfirmak:
> > I am reluctant to merge the second one, I propose we leave it to another PR so that we can discuss it further.
> 
> There was a possibility of go-routine leaks in jsonrpc.http.Run() functions and the documentation of Serve() mentioned that the application should not exit without Shutdown() returning. We relied on the fact that synchroniser will take longer to exit than jsonrpc.http.Run().

Yeah, we should add a way for Node to handle lifecycles of the services it uses. But I cant seem to find the doc that mentions that, could you share it with me?

IronGauntlets:
>Yeah, we should add a way for Node to handle lifecycles of the services it uses. But I cant seem to find the doc that mentions that, could you share it with me?

https://cs.opensource.google/go/go/+/refs/tags/go1.20.1:src/net/http/server.go;drc=ea4631cc0cf301c824bd665a7980c13289ab5c9d;l=2760-2762

PR #589, opened 2023-02-27
Wrap transaction hash calculation errors

IronGauntlets:
## Description

Change the return type of `VerifyBlockhash` to `error`.  This was done by wrapping multiple errors into one error.

## Changes:

see commit

## Types of changes

## Testing

**Requires testing**: Yes

### DIFF
--- a/core/transaction.go
+++ b/core/transaction.go
@@ -331,19 +331,46 @@ func deployAccountTransactionHash(d *DeployAccountTransaction, n utils.Network) 
 }
 
 type ErrCantVerifyTransactionHash struct {
-	t Transaction
+	t           Transaction
+	hashFailure error
+	next        *ErrCantVerifyTransactionHash
+}
+
+func (e ErrCantVerifyTransactionHash) Unwrap() error {
+	if e.next != nil {
+		return *e.next
+	}
+	return nil
 }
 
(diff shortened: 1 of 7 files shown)


core/transaction.go
```diff
@@ -331,19 +331,47 @@ func deployAccountTransactionHash(d *DeployAccountTransaction, n utils.Network)
...
+	for _, tx := range txs {
+		if err := verifyTransactionHash(tx, n); err != nil {
+			tail.next = err
+			tail = err
+		}
+	}
+	if head.next != nil {
+		return *head.next
+	}
+	return nil
```
omerfirmak, line 367:
```suggestion
func verifyTransactions(txs []Transaction, n utils.Network) error {
	var head *ErrCantVerifyTransactionHash
	for _, tx := range txs {
		if err := verifyTransactionHash(tx, n); err != nil {
			err.next = head
			head = err
		}
	}
	if head != nil {
		return head
	}
	return nil
}
```

If I am not mistaken, this would work the same but is simpler

omerfirmak (approved):

core/transaction.go
IronGauntlets: Thanks is much better, thanks!

PR #594, opened 2023-03-03
Iron gauntlets/jsonrpc server

IronGauntlets:


jsonrpc/server.go
```diff
@@ -262,52 +277,65 @@ func (s *Server) handleRequest(req *request) (*response, error) {
...
-			return nil, errors.New("missing non-optional param")
+			args = append(args, v)
 		}
+		return args, nil
+	case reflect.Map:
+		paramsMap := params.(map[string]any)
+
+		for i, configuredParam := range configuredParams {
+			// optional parameter
+			v := reflect.New(handlerType.In(i)).Elem()
```
omerfirmak, line 322: when named parameters are used, we are creating 2 instances for each parameter.

jsonrpc/http.go
```diff
@@ -18,7 +16,7 @@ type Http struct {
 	log  utils.Logger
 }
 
-func NewHttp(port uint16, methods []Method, log utils.Logger) *Http {
+func New(port uint16, methods []Method, log utils.Logger) *Http {
```
omerfirmak, line 19: We will probably have RPC over Websocket in some point. So we need the Http postfix there.

rpc/handlers.go
```diff
@@ -21,7 +21,7 @@ type Handler struct {
 	chainId *felt.Felt
 }
 
-func NewHandler(bcReader blockchain.Reader, chainId *felt.Felt) *Handler {
+func New(bcReader blockchain.Reader, chainId *felt.Felt) *Handler {
```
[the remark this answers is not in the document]
IronGauntlets: We will rename it later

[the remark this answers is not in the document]
IronGauntlets: Its is go convention not to repeat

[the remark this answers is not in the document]
IronGauntlets: Sorry these were meant for NewHttp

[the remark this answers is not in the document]
omerfirmak: There is no repetition tho, it is `jsonrpc.NewHttp`

jsonrpc/server.go
IronGauntlets: I don't understand how. This line just creates a reflect value of the handler param type, how is it worse than `var v reflect.Value`?

omerfirmak: `reflect.Value` is a container that can encapsulate a value of any type.

`reflect.New` initializes a variable of the type that you pass to it, with that type's zero value. Then wraps it with a `reflect.Value`.   `var v reflect.Value` initializes the container only, not the value inside it. Container (`reflect.Value`) is a lightweight object, essentially a pointer with some metadata.

reflect.New might allocate on heap (twice, since we call it twice), `var v reflect.Value` allocates strictly on stack and since we eliminate a `reflect.New` call, we eliminate a heap allocation as well.

see `reflect.Value` definition; https://cs.opensource.google/go/go/+/refs/tags/go1.20.1:src/reflect/value.go;l=39

IronGauntlets: Thanks for the explanation. 

>reflect.New allocates on heap (twice, since we call it twice), var v reflect.Value allocates strictly on stack and since we eliminate a reflect.New call, we eliminate a heap allocation as well.

Allocation depends on escape analysis, not on keywords. Since the second reference (in handlerParamValue) does not leave the scope of this function it will not be allocated on the heap.

omerfirmak: D had a great set of compile time features, we actually had a really similar construct that generated REST servers like we do with JSONRPC here. All allocations happened on stack with the help of compile time magic. Since these allocations are directly influenced by what user sends us, we should make sure we are not allocating too much to be able to fit the incoming user data.

PR #596, opened 2023-03-06
Add Block Transaction Count Endpoint

jelilat:
## Description

Adds `starknet_getBlockTransactionCount` rpc endpoint

## Types of changes

## Testing

**Requires testing**: Yes

### DIFF
--- a/rpc/handlers.go
+++ b/rpc/handlers.go
@@ -215,3 +215,12 @@ func (h *Handler) GetTransactionByHash(hash *felt.Felt) (*Transaction, *jsonrpc.
 	}
 	return adaptTransaction(txn), nil
 }
+
+// https://github.com/starkware-libs/starknet-specs/blob/a789ccc3432c57777beceaa53a34a7ae2f25fda0/api/starknet_api_openrpc.json#L373
+func (h *Handler) GetBlockTransactionCount(id *BlockId) (int, *jsonrpc.Error) {
+	block, err := h.getBlockById(id)
+	if err != nil {
+		return 0, ErrBlockNotFound
+	}
+	return len(block.Transactions), nil
+}
--- a/node/node.go
+++ b/node/node.go
@@ -98,6 +98,7 @@ func makeHttp(port uint16, rpcHandler *rpc.Handler, log utils.Logger) *jsonrpc.H
 		{"starknet_getBlockWithTxHashes", []jsonrpc.Parameter{{Name: "block_id"}}, rpcHandler.GetBlockWithTxHashes},
 		{"starknet_getBlockWithTxs", []jsonrpc.Parameter{{Name: "block_id"}}, rpcHandler.GetBlockWithTxs},
 		{"starknet_getTransactionByHash", []jsonrpc.Parameter{{Name: "transaction_hash"}}, rpcHandler.GetTransactionByHash},
+		{"starknet_getBlockTransactionCount", []jsonrpc.Parameter{{Name: "block_id"}}, rpcHandler.GetBlockTransactionCount},
 	}, log)
 }
 
--- a/rpc/handlers_test.go
+++ b/rpc/handlers_test.go
@@ -111,6 +111,16 @@ func TestHandler(t *testing.T) {
 		}
 	})
 
+	t.Run("starknet_getBlockTransactionCount", func(t *testing.T) {
+		blockTxnCount, err := handler.GetBlockTransactionCount(&rpc.BlockId{Number: 2})
+		assert.Nil(t, err)
+
+		gwBlock, gwErr := gw.BlockByNumber(ctx, 2)
+		assert.NoError(t, gwErr)
+
+		assert.Equal(t, len(gwBlock.Transactions), blockTxnCount)
+	})
+
 	canceler()
 	<-syncNodeChan
 }


rpc/handlers.go
```diff
@@ -215,3 +215,13 @@ func (h *Handler) GetTransactionByHash(hash *felt.Felt) (*Transaction, *jsonrpc.
...
 }
+
+// https://github.com/starkware-libs/starknet-specs/blob/a789ccc3432c57777beceaa53a34a7ae2f25fda0/api/starknet_api_openrpc.json#L373
+func (h *Handler) GetBlockTransactionCount(id *BlockId) (uint64, *jsonrpc.Error) {
+	block, err := h.getBlockById(id)
+	if err != nil {
+		return 0, ErrBlockNotFound
+	}
+
+	return uint64(len(block.Transactions)), nil
```
omerfirmak, line 226: We probably dont need the cast to uint64 here

omerfirmak (approved):

rpc/handlers.go
omerfirmak: Yeah, I meant we can just return `int`

jelilat: Alright. Done

jelilat:
@omerfirmak made some changes to this cos I don't think we should fetch all the blocks. We won't have to pointlessly fetch block head and receipts, so it should be faster and consume less memory

blockchain/blockchain.go
```diff
@@ -323,6 +339,43 @@ func getBlockByHash(txn db.Transaction, hash *felt.Felt) (block *core.Block, err
...
+	numBytes := make([]byte, lenOfByteSlice)
+	binary.BigEndian.PutUint64(numBytes, number)
+
+	iterator, err := txn.NewIterator()
+	if err != nil {
+		return 0, err
+	}
+	defer func() {
+		// Prioritise closing error over other errors
+		if closeErr := iterator.Close(); closeErr != nil {
```
omerfirmak, line 353: should use `CloseAndWrapOnError`

rpc/handlers.go
```diff
@@ -215,3 +215,31 @@ func (h *Handler) GetTransactionByHash(hash *felt.Felt) (*Transaction, *jsonrpc.
...
+	} else {
+		count, err = h.bcReader.GetBlockTransactionCountByNumber(id.Number)
+	}
+
+	return count, err
+}
+
+// https://github.com/starkware-libs/starknet-specs/blob/a789ccc3432c57777beceaa53a34a7ae2f25fda0/api/starknet_api_openrpc.json#L373
+func (h *Handler) GetBlockTransactionCount(id *BlockId) (uint64, *jsonrpc.Error) {
+	txnCount, err := h.getBlockTxnCountById(id)
```
omerfirmak, line 239: no need for the `getBlockTxnCountById` helper

rpc/handlers.go
```diff
@@ -215,3 +215,31 @@ func (h *Handler) GetTransactionByHash(hash *felt.Felt) (*Transaction, *jsonrpc.
...
+	var err error
+	if id.Latest {
+		if height, e := h.bcReader.Height(); err != nil {
+			return 0, e
+		} else {
+			count, err = h.bcReader.GetBlockTransactionCountByNumber(height)
+		}
+	} else if id.Hash != nil {
+		count, err = h.bcReader.GetBlockTransactionCountByHash(id.Hash)
+	} else {
```
omerfirmak, line 230: id.Pending will hit this branch

blockchain/blockchain.go
```diff
@@ -323,6 +339,43 @@ func getBlockByHash(txn db.Transaction, hash *felt.Felt) (block *core.Block, err
...
+			err = closeErr
+		}
+	}()
+
+	var noOfTransactions uint64
+	prefix := db.TransactionsByBlockNumberAndIndex.Key(numBytes)
+	for iterator.Seek(prefix); iterator.Valid(); iterator.Next() {
```
omerfirmak, line 360: essentially duplicate of `getBlockByNumber`, we should eliminate the duplication

omerfirmak (changes requested):

omerfirmak:
> @omerfirmak made some changes to this cos I don't think we should fetch all the blocks. We won't have to pointlessly fetch block head and receipts, so it should be faster and consume less memory

You still have to iterate over txns, just not unmarshaling them. Iterating over db dominates the runtime anyways, so the benefit we get from this is marginal.

IronGauntlets:
Any particular reason why we are not doing the following:

```go
func (h *Handler) GetBlockTransactionCount(id *BlockId) (uint64, *jsonrpc.Error) {
	block, err := h.getBlockById(id)
	if err != nil {
		return 0, ErrBlockNotFound
	}

	return len(block.Transactions), nil
}
```

I know the above will be wasteful, however, adding `txCount` to the header is out of scope for this PR. 

If we do decide to add `txCount` to the header we should also add `eventCount` along with testcases to ensure there is no mismatch.

jelilat:
> Any particular reason why we are not doing the following:
> 
> ```go
> func (h *Handler) GetBlockTransactionCount(id *BlockId) (uint64, *jsonrpc.Error) {
> 	block, err := h.getBlockById(id)
> 	if err != nil {
> 		return 0, ErrBlockNotFound
> 	}
> 
> 	return len(block.Transactions), nil
> }
> ```
> 
> I know the above will be wasteful, however, adding `txCount` to the header is out of scope for this PR.
> 
> If we do decide to add `txCount` to the header we should also add `eventCount` along with testcases to ensure there is no mismatch.

This is what I did earlier, but I had reservations and made the new changes. If everyone agrees to do it this way, I'll just revert the changes cc @omerfirmak

IronGauntlets (approved):

omerfirmak (approved):

PR #598, opened 2023-03-06
test different trie height

liusanchuan:
add test #528

## Description

test different trie height

## Changes:

- add unit test for different trie height.

## Types of changes

- Other (please describe):  Add more test

## Testing

**Requires testing**: Yes

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ config/
 
 # pyenv
 .python-version
+.gitpod.yml
--- a/core/trie/trie_test.go
+++ b/core/trie/trie_test.go
@@ -684,3 +684,24 @@ func TestOldData(t *testing.T) {
 		return nil
 	})
 }
+
+func TestTrieHeight(t *testing.T) {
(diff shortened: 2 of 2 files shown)


liusanchuan:
Hi @IronGauntlets, 
When I put a key to a trie, (the key's length is longer than the trie's height), it work normal as the longer key was truncked in `func feltToBitSet()`. 
And while I `get()` this key's value, it still be truncked. So the finial result [works noraml.](https://github.com/NethermindEth/juno/pull/598/files#diff-eb1136f09e3388f45c868c2cb7f330934f685ade824a1cada8944805048720d3), no panic as we expect in [issue](https://github.com/NethermindEth/juno/issues/528#issuecomment-1448453145).
And I have 2 proposals for this problem, first one is treat this as normal, just like assign `uint16` to `uint8`.
Second method is adding a check in `func feltToBitSet()`, panic while key out of length.
Which one is do you agree? or is there any better method?

omerfirmak:
Closing in favor of #668

liusanchuan:
sorry about this

PR #599, opened 2023-03-06
Implement starknet_getTransactionReceipt

omerfirmak:


rpc/transaction.go
```diff
@@ -21,3 +22,27 @@ type Transaction struct {
...
+
+type Event struct {
+	From *felt.Felt   `json:"from_address"`
+	Keys []*felt.Felt `json:"keys"`
+	Data []*felt.Felt `json:"data"`
+}
+
+// https://github.com/starkware-libs/starknet-specs/blob/master/api/starknet_api_openrpc.json#L1871
+type TransactionReceipt struct {
+	Type            string     `json:"type"`
```
jelilat, line 39: We should probably also make this an `enum` https://github.com/starkware-libs/starknet-specs/blob/master/api/starknet_api_openrpc.json#L1979

rpc/block.go
```diff
@@ -8,24 +8,24 @@ import (
...
 
 const (
-	BlockStatusPending BlockStatus = iota
-	BlockStatusAcceptedL2
-	BlockStatusAcceptedL1
-	BlockStatusRejected
+	StatusPending Status = iota
+	StatusAcceptedL2
+	StatusAcceptedL1
+	StatusRejected
```
jelilat, line 17: Maybe this?
```suggestion
	Pending Status = iota
	AcceptedOnL2
	AcceptedOnL1
	Rejected
```

rpc/handlers.go
```diff
@@ -215,3 +216,49 @@ func (h *Handler) GetTransactionByHash(hash *felt.Felt) (*Transaction, *jsonrpc.
...
+// https://github.com/starkware-libs/starknet-specs/blob/master/api/starknet_api_openrpc.json#L222
+func (h *Handler) GetTransactionReceiptByHash(hash *felt.Felt) (*TransactionReceipt, *jsonrpc.Error) {
+	txn, rpcErr := h.GetTransactionByHash(hash)
+	if rpcErr != nil {
+		return nil, rpcErr
+	}
+	receipt, blockHash, blockNumber, err := h.bcReader.GetReceipt(hash)
+	if err != nil {
+		return nil, ErrTxnHashNotFound
+	}
```
jelilat, line 229: We could probably make this faster by running `GetTransactionByHash` and `h.bcReader.GetReceipt(hash)` concurrently

rpc/block.go
omerfirmak: With the prefix I find it easier to enumerate all possible "Status" values while actually coding.
after typing only "Status" goland suggests you all the possible values, which I find to be neat:)

also prefix avoids any conflicts. another enum type could have a Pending value as well.

rpc/handlers.go
omerfirmak: Not really worth to spawn a new goroutine. They are both requests to the local db.

rpc/transaction.go
omerfirmak: Thanks, done!

jelilat (approved):

IronGauntlets (approved):

PR #600, opened 2023-03-06
Limit RPC request sizes in HTTP transport layer

omerfirmak:


jsonrpc/http.go
```diff
@@ -10,6 +10,8 @@ import (
 	"github.com/NethermindEth/juno/utils"
 )
 
+const TEN_MB = 10 * 1024 * 1024
```
IronGauntlets, line 13:
```suggestion
const MaxBytes = 10 * 1024 * 1024
```

IronGauntlets (approved):
What made you realise that we need to do this?

jsonrpc/http.go
jelilat:
```suggestion
const MaxBytes = 10 * 1024 * 1024 // 10MB
```

jelilat (approved):

omerfirmak:
> What made you realise that we need to do this?

I was already aware for some time, also @wojciechos pointed it out.

PR #601, opened 2023-03-06
Add profiling pprof flag

wojciechos:
## Description
Adds a --prof flag, which enables the pprof server for continuous profiling. When the --prof flag is provided, the application starts a pprof server that listens on port 9080.

## Changes:

- Added the --prof flag
- Run profiler server [pprof](https://pkg.go.dev/net/http/pprof)

## Types of changes

## Testing

**Requires testing**: Yes

**Did you write tests??**: No

## Documentation


node/node.go
```diff
@@ -102,6 +106,12 @@ func makeHttp(port uint16, rpcHandler *rpc.Handler, log utils.Logger) *jsonrpc.H
 }
 
 func (n *Node) Run(ctx context.Context) (err error) {
+	if n.cfg.Pprof {
+		go func() {
+			n.log.Infow("Starting pprof server on port " + defaultPprofPort)
+			println(http.ListenAndServe("localhost"+defaultPprofPort, nil))
```
IronGauntlets, line 112: Where is this println coming from? is it from the "net/http/pprof" package?

cmd/juno/juno.go
```diff
@@ -76,6 +79,7 @@ func NewCmd(newNodeFn node.NewStarknetNodeFn) *cobra.Command {
 	junoCmd.Flags().String(dbPathF, defaultDbPath, dbPathUsage)
 	junoCmd.Flags().Uint8(networkF, uint8(defaultNetwork), networkUsage)
 	junoCmd.Flags().String(ethNodeF, defaultEthNode, ethNodeUsage)
+	junoCmd.Flags().Bool(pprofF, defaultPprof, pprofUsage)
```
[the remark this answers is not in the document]
wojciechos: Yeah, that's right - tests adjusted.

IronGauntlets:
@wojciechos do we need both `--pprof` and `--metrics` flags? if not then we should replace `metrics` with `pprof`.

wojciechos:
> @wojciechos do we need both `--pprof` and `--metrics` flags? if not then we should replace `metrics` with `pprof`.

I would like to keep both the `--pprof` and `--metrics` flags, as they serve different purposes. The profiler is mainly intended for our internal/developer usage, so maybe we should even hide the `--pprof` flag from the public-facing cli. What do you think?

IronGauntlets (approved):

node/node.go
```diff
@@ -152,8 +166,13 @@ func (n *Node) Run(ctx context.Context) {
 	client := feeder.NewClient(n.cfg.Network.URL())
 	synchronizer := sync.New(n.blockchain, adaptfeeder.New(client), n.log)
 	http := makeHttp(n.cfg.RpcPort, rpc.New(n.blockchain, n.cfg.Network), n.log)
+
 	n.services = []service.Service{synchronizer, http}
 
+	if n.cfg.Pprof {
+		n.services = append(n.services, pprof.New(defaultPprofPort, n.log))
```
wojciechos, line 173: @omerfirmak @IronGauntlets is it expected way of running services like pprof here or should I move it somewhere else?

pprof/pprof.go
```diff
@@ -0,0 +1,51 @@
...
+		started <- struct{}{}
+		err := server.ListenAndServe()
+		if err != nil && err != http.ErrServerClosed {
+			p.log.Errorw("Pprof server error", "err", err)
+		}
+	}()
+
+	go func() {
+		<-ctx.Done()
+		server.Shutdown(context.Background())
```
omerfirmak, line 45: See how `http` handles this, we need pprof to be similar to that.

wojciechos: Not sure what you exactly meant but I made it a little bit simpler and added error handling for shut downing.

pprof/pprof.go
```diff
@@ -0,0 +1,49 @@
...
+		log:     log,
+	}
+}
+
+func (p *Pprof) Run(ctx context.Context) error {
+	if !p.enabled {
+		return nil
+	}
+
+	server := &http.Server{
```
omerfirmak, line 29: this can be moved to New

pprof/pprof.go
omerfirmak: http propagates the shutdown error to the caller, maybe we should do the same here as well.

wojciechos: changed - now error on shutdown is returned

node/node.go
```diff
@@ -26,6 +27,10 @@ type StarknetNode interface {
 
 type NewStarknetNodeFn func(cfg *Config) (StarknetNode, error)
 
+const (
+	defaultPprofPort = ":9010"
```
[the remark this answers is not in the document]
wojciechos: done

pprof/pprof.go
wojciechos: done

pprof/pprof.go
```diff
@@ -0,0 +1,53 @@
...
+		p.log.Infow("Starting pprof...", p.server.Addr)
+		if err := p.server.ListenAndServe(); err != nil && err != http.ErrServerClosed {
+			p.log.Errorw("Pprof server error", "err", err)
+		}
+	}()
+
+	<-ctx.Done()
+
+	p.log.Infow("Shutting down pprof...")
+	err := p.server.Shutdown(context.Background())
```
omerfirmak, line 46: `return p.server.Shutdown(context.Background())` is enough, node will log the error anyways

PR #603, opened 2023-03-07
Fix sync loop not being able to roll itself back to a known good height

omerfirmak:
Assume all our streams are full, and `SyncBlocks` is blocked on `fetchers.Go`.

If the next `verifierTask` callback encounters an error and tries to write it to the `errChan` it will block indefinitely because our main loop is blocked on `fetchers.Go`. This is a deadlock that mostly happens after we restart sync loop a couple of times.

By making `errChan` a buffered channel with capacity 1, we make sure that we get to write at least 1 error to the channel without blocking and immediately free up a slot in our streams. That makes sure `fetchers.Go` will never block indefinitely.

Other verifiers in the `stream` may end up being blocked trying to write to the `errChan`, but `SyncBlocks` loop will immediately acknowledge the first error in the channel and cancel the stream context. So all verifiers and fetchers will eventually terminate.

### DIFF
--- a/sync/sync_test.go
+++ b/sync/sync_test.go
@@ -22,6 +22,7 @@ func TestSyncBlocks(t *testing.T) {
 			require.NoError(t, err)
 
 			height := int(headBlock.Number)
+			assert.Equal(t, 2, height)
 			for height >= 0 {
 				b, err := gw.BlockByNumber(context.Background(), uint64(height))
 				if err != nil {
--- a/sync/sync.go
+++ b/sync/sync.go
@@ -83,6 +83,14 @@ func (s *Synchronizer) fetcherTask(ctx context.Context, height uint64, verifiers
 	}
 }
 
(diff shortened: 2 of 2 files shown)


sync/sync_test.go
```diff
@@ -22,6 +22,7 @@ func TestSyncBlocks(t *testing.T) {
 			require.NoError(t, err)
 
 			height := int(headBlock.Number)
+			assert.True(t, height > 0)
```
IronGauntlets, line 25: we should assert the height to be 2 and if this fails we should increase sleep time

IronGauntlets (changes requested):

jelilat (approved):

sync/sync_test.go
omerfirmak: done

IronGauntlets (approved):

PR #606, opened 2023-03-07
Move test gateway client to clients package

joshklop:
Allows us to remove the testsource package.

## Types of changes

## Testing

**Requires testing**: Yes

**Did you write tests??**: existing tests are sufficient

## Documentation

**If this requires a documentation update, did you add one?** unnecessary


joshklop:
It might make sense to create a `clients/gateway` package (or `clients/feeder` to prevent clashing with `starknetdata/gateway`) since the `testdata` won't apply to the p2p client. I'll hold off on that for now and if anyone has a strong opinion please let me know in the reviews :)

IronGauntlets:
>It might make sense to create a clients/gateway package (or clients/feeder to prevent clashing with starknetdata/gateway) since the testdata won't apply to the p2p client. I'll hold off on that for now and if anyone has a strong opinion please let me know in the reviews :)

One thing to note is that `feeder gateway` and `gateway` are 2 different endpoints. `Feeder gateway` provide starknet information while `gateway` accepts new transactions. We should resolve these naming issues first.

Regarding packages, I would like to see something similar to the following:

```go
clients
  |-> feeder
  |-> gateway
starknetdata
  |-> feeder
  |-> p2p
  |-> starknetdata.go
```

clients/gateway.go
```diff
@@ -60,6 +64,62 @@ func NopBackoff(d time.Duration) time.Duration {
 	return 0
 }
 
+// NewTestGatewayClient returns a client and a function to close a test server.
+func NewTestGatewayClient(network utils.Network) (*GatewayClient, func()) {
```
IronGauntlets, line 68: Remove stutter from `NewTestGatewayClient` and `NewGatewayClient` by renaming them to `NewTestGateway` and `NewGateway`, respectively.

joshklop:
That makes sense @IronGauntlets. In that case, would the package structure look like the following? 
```
clients
  |-> feeder
  |-> gateway
  |-> p2p
starknetdata
  |-> feeder
  |-> gateway
  |-> p2p
  |-> starknetdata.go
```
Since we will usually be importing `clients/...` and `starknetdata/...` in the same places, it might be helpful to make the package names slightly different, like `adaptfeeder`, `adaptp2p`, etc.. Kind of ugly, but more explicit than doing things like `feeder2`.

IronGauntlets:
> Since we will usually be importing clients/... and starknetdata/... in the same places, it might be helpful to make the package names slightly different, like adaptfeeder, adaptp2p, etc.. Kind of ugly, but more explicit than doing things like feeder2.

I would prefer to keep the above structure and just have an import alias to differentiate them when they are imported together. I really don't want to have the word `adapt` in the package name.

```go
import (
  ".../clients/feeder"
  adaptfeeder ".../starknetdata/feeder"
)
```

joshklop:
Finished and rebased, ready for another round of reviews. Pretty much all of the changes are simple renaming/refactoring, limited as much as possible to what is in the commit message. I changed most references from "gateway" to "feeder" so we can unambiguously refer to the "gateway" API in the future.

starknetdata/feeder/gateway.go
```diff
@@ -11,10 +11,10 @@ import (
 )
 
 type Gateway struct {
-	client *feeder.GatewayClient
+	client *feeder.FeederClient
```
IronGauntlets, line 14: `feeder.Feeder` is a sutter.

IronGauntlets (approved):
Nice work!

starknetdata/feeder/feeder.go
```diff
@@ -1,35 +1,28 @@
...
+	return &Feeder{
 		client: client,
 	}
 }
 
-// BlockByNumber gets the block for a given block number from the feeder gateway,
+// BlockByNumber gets the block for a given block number from the feeder,
 // then adapts it to the core.Block type.
-func (g *Gateway) BlockByNumber(ctx context.Context, blockNumber uint64) (*core.Block, error) {
+func (g *Feeder) BlockByNumber(ctx context.Context, blockNumber uint64) (*core.Block, error) {
```
jelilat, line 25: Since we've changed the name to `Feeder`, we should probably also change this

```suggestion
func (f *Feeder) BlockByNumber(ctx context.Context, blockNumber uint64) (*core.Block, error) {
```

starknetdata/feeder/feeder.go
```diff
@@ -160,9 +153,9 @@ func adaptL2ToL1Message(response *clients.L2ToL1Message) *core.L2ToL1Message {
 	}
 }
 
-// Transaction gets the transaction for a given transaction hash from the feeder gateway,
+// Transaction gets the transaction for a given transaction hash from the feeder,
 // then adapts it to the appropriate core.Transaction types.
-func (g *Gateway) Transaction(ctx context.Context, transactionHash *felt.Felt) (core.Transaction, error) {
+func (g *Feeder) Transaction(ctx context.Context, transactionHash *felt.Felt) (core.Transaction, error) {
```
jelilat, line 158: Same as above
```suggestion
func (f *Feeder) Transaction(ctx context.Context, transactionHash *felt.Felt) (core.Transaction, error) {
```

jelilat (changes requested):

jelilat (approved):

PR #607, opened 2023-03-08
Wait for services to shutdown properly

omerfirmak:


node/node.go
```diff
@@ -17,7 +17,7 @@ import (
 )
 
 type StarknetNode interface {
-	Run(ctx context.Context) error
+	Service
```
IronGauntlets, line 20: StarknetNode should not be considered a service. Since it is the one responsible for managing all of Juno's services.

node/service.go
```diff
@@ -0,0 +1,11 @@
...
+
+import "context"
+
+type Service interface {
+	// Run starts the service and returns an error only if service fails to start
+	// Run does not block and can be canceled with the given context
+	Run(ctx context.Context) error
+	// Wait waits for the service to shut down gracefully
+	Wait()
+}
```
IronGauntlets, line 11: This should be moved to root level so that we signify that anyone can implement this

node/node.go
omerfirmak: done

node/service.go
omerfirmak: done

IronGauntlets (approved):

jelilat (approved):

PR #604, opened 2023-03-07
Introduce blockchain.Reader Mock

IronGauntlets:

PR #608, opened 2023-03-08
Implement starknet_getStateUpdate

omerfirmak:
Not tested, waiting for #604 

follows the old structure at https://github.com/starkware-libs/starknet-specs/blob/ed85de6a701c966ec7395b0df4c8bd62f3cddbcd/api/starknet_api_openrpc.json#L844


IronGauntlets:
We should implement the latest state update as per the latest specs so that we won't have change anything following 0.11.0 changes. If we go ahead with this integration network may not work.

omerfirmak:
> We should implement the latest state update as per the latest specs so that we won't have change anything following 0.11.0 changes. If we go ahead with this integration network may not work.

Integration network wont work without #590 anyways. Shouldnt we implement the old state update endpoint for the v0.10.3 release?

IronGauntlets:
>Integration network wont work without https://github.com/NethermindEth/juno/pull/590 anyways. Shouldnt we implement the old state update endpoint for the v0.10.3 release?

I thought we would be increasing work for ourselves by implementing the old version. If we have fields for the latest state_update which are always empty then we will not have to do any more work on the rpc side of things to support 0.11.0 changes, right?

omerfirmak:
> > Integration network wont work without #590 anyways. Shouldnt we implement the old state update endpoint for the v0.10.3 release?
> 
> I thought we would be increasing work for ourselves by implementing the old version. If we have fields for the latest state_update which are always empty then we will not have to do any more work on the rpc side of things to support 0.11.0 changes, right?

We will still have to populate the new fields and do additional testing for those fields.

If we are going to make a v0.10.3 release we should make a proper one. The additional work is negligible at worst. `core` supporting v0.10.3 only while rpc trying to serve v0.11.0 doesnt really make sense to me.

omerfirmak:
Added tests as well, ready for review.

rpc/handlers.go
```diff
@@ -271,3 +273,59 @@ func (h *Handler) GetTransactionReceiptByHash(hash *felt.Felt) (*TransactionRece
 		ContractAddress: contractAddress,
 	}, nil
 }
+
+// https://github.com/starkware-libs/starknet-specs/blob/master/api/starknet_api_openrpc.json#L77
+func (h *Handler) GetStateUpdate(id *BlockId) (*StateUpdate, *jsonrpc.Error) {
+	var update *core.StateUpdate
+	var err error
+	if id.Latest {
+		if head, headErr := h.bcReader.Height(); headErr != nil {
```
IronGauntlets, line 282: head should be rename to height

rpc/handlers_test.go
```diff
@@ -618,3 +618,97 @@ func TestGetTransactionReceiptByHash(t *testing.T) {
...
+		assert.Nil(t, rpcErr)
+		checkUpdate(t, update21656, update)
+	})
+
+	t.Run("by height", func(t *testing.T) {
+		mockReader.EXPECT().GetStateUpdateByNumber(uint64(21656)).Return(update21656, nil)
+
+		_, rpcErr := handler.GetStateUpdate(&rpc.BlockId{Number: uint64(21656)})
+		assert.Nil(t, rpcErr)
+		//		checkUpdate(t, update21656, update)
```
IronGauntlets, line 704: should be uncommented

rpc/handlers_test.go
```diff
@@ -618,3 +618,97 @@ func TestGetTransactionReceiptByHash(t *testing.T) {
...
+	})
+
+	t.Run("by hash", func(t *testing.T) {
+		mockReader.EXPECT().GetStateUpdateByHash(update21656.BlockHash).Return(update21656, nil)
+
+		_, rpcErr := handler.GetStateUpdate(&rpc.BlockId{Hash: update21656.BlockHash})
+		assert.Nil(t, rpcErr)
+		//		checkUpdate(t, update21656, update)
```
IronGauntlets, line 712: same as above

PR #615, opened 2023-03-10
Fix Assert.Equal() Arguments Order in Tests

jelilat:
When using `assert.Equal()`, the expected order for the arguments is the following:

```
func assert.Equal(t assert.TestingT, expected interface{}, actual interface{}, msgAndArgs ...interface{}) bool
```

But we tend to not follow the order in some of our test cases. So, we sometimes do this:

```
func assert.Equal(t assert.TestingT, actual interface{}, expected interface{}, msgAndArgs ...interface{}) bool
```

This PR refactors our tests to follow the correct order.


omerfirmak (approved):
Makes sense

joshklop (approved):

PR #616, opened 2023-03-10
Add --version flag

omerfirmak:
```
$ ./build/juno --version
juno version v0.1.0-424b20e310380df83c10067fa294ee7ae16d53b3
```

prints `<version>-<commit sha>`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@ export CC = clang
 
 juno: ## compile
 	@mkdir -p build
-	@go build -o build/juno ./cmd/juno/
+	@go build -ldflags="-X main.Version=$(shell git describe --tags)" -o build/juno ./cmd/juno/
 
 all: juno
 
--- a/cmd/juno/juno.go
+++ b/cmd/juno/juno.go
@@ -9,6 +9,8 @@ import (
 	"github.com/spf13/viper"
 )
 
+var Version string
+
 const greeting = `
        _                    
       | |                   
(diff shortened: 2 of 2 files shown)


swapnilraj (approved):

wojciechos (approved):

cmd/juno/juno.go
```diff
@@ -2,13 +2,27 @@ package main
...
 	"fmt"
+	"runtime/debug"
 
 	"github.com/NethermindEth/juno/node"
 	"github.com/NethermindEth/juno/utils"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 )
 
+var version = "v0.1.0"
```
IronGauntlets, line 13: We should not be using a string to maintain our version, it is prone to not being updated

cmd/juno/juno.go
```diff
@@ -2,13 +2,27 @@ package main
...
+	if info, ok := debug.ReadBuildInfo(); ok {
+		for _, setting := range info.Settings {
+			if setting.Key == "vcs.revision" {
+				return setting.Value
+			}
+		}
+	}
+	return ""
+}()
+
```
IronGauntlets, line 25: Is there a way to get a tag instead of a commit?

omerfirmak: we can use `git describe --tags`

omerfirmak: done

omerfirmak:
```
$ ./build/juno --version
juno version legacy-260-ge65d841
```

Makefile
```diff
@@ -3,7 +3,7 @@ export CC = clang
 
 juno: ## compile
 	@mkdir -p build
-	@go build -o build/juno ./cmd/juno/
+	@go build -ldflags="-X main.Version=$(shell git describe --tags)" -o build/juno ./cmd/juno/
```
IronGauntlets, line 6: Would this always return the highest semantic flag?

cmd/juno/juno.go
IronGauntlets: Do you know what the 260 means, is it the commit difference from the main?

omerfirmak: Yeap, number of new commits after `legacy`

IronGauntlets (approved):

PR #618, opened 2023-03-10
Remove Duplicate State Update Data Storage

jelilat:
Some of the data stored in `StateUpdatesByBlockNumber` already exists in the database and shouldn't be stored again. Storing them again unnecessarily bloats our database. 

This PR removes duplicate data storage and introduces a new way to store the nonce.

### Changes

- **Nonce**: Introduces historical nonce storage outside `StateUpdate`. Storing nonces in a separate table allows us to be able to serve `starknet_getNonce` in the future.
- **Remove duplicate data storage**: Stop duplicated storage of `block hashes`, `roots`, `declaredClasses`, and `deployedContracts` since they are already stored as part of the `block`.

### DIFF
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -203,7 +203,7 @@ func (b *Blockchain) Store(block *core.Block, stateUpdate *core.StateUpdate, dec
 		if err := b.verifyBlock(txn, block); err != nil {
 			return err
 		}
-		if err := core.NewState(txn).Update(stateUpdate, declaredClasses); err != nil {
+		if err := core.NewState(txn).Update(block.Number, stateUpdate, declaredClasses); err != nil {
 			return err
 		}
 		if err := storeBlockHeader(txn, block.Header); err != nil {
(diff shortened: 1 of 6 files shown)


omerfirmak:
> declaredClasses, and deployedContracts since they are already stored as part of the block.

That is not particularly correct. Txns in blocks could fail during execution, even if included in a block. So It is not really possible to rebuild declaredClasses, and deployedContracts just by looking at deploy and declare txns. 

> Stop duplicated storage of block hashes, roots

not really worth the effort, they are couple of felts per block anyways

jelilat:
> That is not particularly correct. Txns in blocks could fail during execution, even if included in a block. So It is not really possible to rebuild declaredClasses, and deployedContracts just by looking at deploy and declare txns.
> 
I think we should be able to skip failed transactions as it'll reflect in the status that the transaction has failed.
> 
> not really worth the effort, they are couple of felts per block anyways

Imo, these couple of felts add up at the end of the day. It may seem little at the moment, but when we start synching millions of blocks, it'll add up to something huge. We should look for ways to make our db smaller and future-proof

omerfirmak:
3 extra felts per block makes 96MB extra space used per 1M blocks. It is a drop in the bucket.

jelilat:
Closing this given the comments above

PR #619, opened 2023-03-13
Add workflow to publish binaries

wojciechos:
The workflow will build and publish artifacts with binaries for Linux, Darwin, and Windows for every tag push with a version to the repository.

### DIFF
--- a/.github/workflows/build-binaries.yml
+++ b/.github/workflows/build-binaries.yml
@@ -1,0 +1,45 @@
+name: Build Binaries
+
+on:
+  push:
+    tags:
+      - 'v*'
+  workflow_dispatch:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        arch: ['amd64', 'arm64']
+        os: ['linux', 'darwin', 'windows']
+        compiler: ['gcc']
+        go-version: ['1.19']
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+      
+      - name: Install UPX
+        run: sudo apt-get install -y upx-ucl
+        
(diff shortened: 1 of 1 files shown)


.github/workflows/build-binaries.yml
```diff
@@ -0,0 +1,45 @@
+name: Build Binaries
+
+on:
+  push:
+    tags:
+      - 'v*.*.*'
+  workflow_dispatch:
```
IronGauntlets, line 7: I am assuming this workflow would run every time a new tag is pushed or updated?

IronGauntlets (approved):

.github/workflows/build-binaries.yml
wojciechos: just changed to only '*'  - but yeah, the workflow should be triggered when any new tag is pushed

wojciechos: ...or could be triggered manually

IronGauntlets: I don't think we should build on all tags. Should keep the regex at `v*`?

wojciechos: yeah, we can try with only that for now - changed

omerfirmak:
Does it make sense to run this on releases only ? like; https://github.com/bosagora/agora/blob/v0.x.x/.github/workflows/release.yaml#L14

joshklop (approved):

wojciechos:
> Does it make sense to run this on releases only ? like; https://github.com/bosagora/agora/blob/v0.x.x/.github/workflows/release.yaml#L14

Yeah, that make sense but maybe we can wait with that for now as I have to test each binary manually before releasing it.

PR #624, opened 2023-03-14
Remove named return values

joshklop:
Used this regex to find them: `\) \(\w+ \*?[\w|\.]+.*\).*$`

### DIFF
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -98,30 +98,37 @@ func (b *Blockchain) Network() utils.Network {
 
 // StateCommitment returns the latest block state commitment.
 // If blockchain is empty zero felt is returned.
-func (b *Blockchain) StateCommitment() (commitment *felt.Felt, err error) {
+func (b *Blockchain) StateCommitment() (*felt.Felt, error) {
+	var commitment *felt.Felt
 	return commitment, b.database.View(func(txn db.Transaction) error {
+		var err error
 		commitment, err = core.NewState(txn).Root()
 		return err
 	})
 }
 
 // Height returns the latest block height. If blockchain is empty nil is returned.
-func (b *Blockchain) Height() (height uint64, err error) {
+func (b *Blockchain) Height() (uint64, error) {
+	var height uint64
 	return height, b.database.View(func(txn db.Transaction) error {
+		var err error
 		height, err = b.height(txn)
 		return err
 	})
 }
 
-func (b *Blockchain) height(txn db.Transaction) (height uint64, err error) {
+func (b *Blockchain) height(txn db.Transaction) (uint64, error) {
+	var height uint64
 	return height, txn.Get(db.ChainHeight.Key(), func(val []byte) error {
 		height = binary.BigEndian.Uint64(val)
 		return nil
 	})
 }
 
-func (b *Blockchain) Head() (head *core.Block, err error) {
+func (b *Blockchain) Head() (*core.Block, error) {
+	var head *core.Block
 	return head, b.database.View(func(txn db.Transaction) error {
+		var err error
 		head, err = b.head(txn)
 		return err
 	})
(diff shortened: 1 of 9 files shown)


blockchain/blockchain.go
```diff
@@ -394,18 +425,20 @@ func storeStateUpdate(txn db.Transaction, blockNumber uint64, update *core.State
 	return nil
 }
 
-func getStateUpdateByNumber(txn db.Transaction, blockNumber uint64) (update *core.StateUpdate, err error) {
+func getStateUpdateByNumber(txn db.Transaction, blockNumber uint64) (*core.StateUpdate, error) {
 	numBytes := make([]byte, lenOfByteSlice)
 	binary.BigEndian.PutUint64(numBytes, blockNumber)
 
+	update := new(core.StateUpdate)
 	return update, txn.Get(db.StateUpdatesByBlockNumber.Key(numBytes), func(val []byte) error {
```
omerfirmak, line 433: Let's check error explicitly and `return nil, err`; so that we dont return a non-nil `update` when `err != nil`

blockchain/blockchain.go
```diff
@@ -474,15 +507,16 @@ func storeTransactionAndReceipt(txn db.Transaction, number, i uint64, t core.Tra
 }
 
 // getTransactionBlockNumberAndIndexByHash gets the block number and index for a given transaction hash
-func getTransactionBlockNumberAndIndexByHash(txn db.Transaction, hash *felt.Felt) (bnIndex *txAndReceiptDBKey, err error) {
+func getTransactionBlockNumberAndIndexByHash(txn db.Transaction, hash *felt.Felt) (*txAndReceiptDBKey, error) {
+	bnIndex := new(txAndReceiptDBKey)
```
omerfirmak, line 511: same, explicit `return nil, err`

core/contract.go
```diff
@@ -50,14 +50,13 @@ func (c *Contract) Deploy(classHash *felt.Felt) error {
 
 // Nonce returns the number of transactions sent from this contract.
 // Only account contracts can have a non-zero nonce.
-func (c *Contract) Nonce() (nonce *felt.Felt, err error) {
+func (c *Contract) Nonce() (*felt.Felt, error) {
 	key := db.ContractNonce.Key(c.Address.Marshal())
-	err = c.txn.Get(key, func(val []byte) error {
-		nonce = new(felt.Felt)
+	nonce := new(felt.Felt)
```
omerfirmak, line 55: same, `return nil, err`

core/contract.go
```diff
@@ -67,14 +66,13 @@ func (c *Contract) UpdateNonce(nonce *felt.Felt) error {
 }
 
 // ClassHash returns hash of the class that this contract instantiates.
-func (c *Contract) ClassHash() (classHash *felt.Felt, err error) {
+func (c *Contract) ClassHash() (*felt.Felt, error) {
 	key := db.ContractClassHash.Key(c.Address.Marshal())
-	err = c.txn.Get(key, func(val []byte) error {
-		classHash = new(felt.Felt)
+	classHash := new(felt.Felt)
```
omerfirmak, line 71: same, `return nil, err`

core/state.go
```diff
@@ -91,12 +91,12 @@ func (s *State) getStateStorage() (*trie.Trie, error) {
 }
 
 // rootKey returns key to the root node in the given Txn context.
-func (s *State) rootKey() (key *bitset.BitSet, err error) {
-	err = s.txn.Get(db.State.Key([]byte(stateRootKey)), func(val []byte) error {
-		key = new(bitset.BitSet)
+func (s *State) rootKey() (*bitset.BitSet, error) {
+	key := new(bitset.BitSet)
```
omerfirmak, line 95: same, `return nil, err`

core/transaction_storage.go
```diff
@@ -47,17 +47,17 @@ func (t *TransactionStorage) Put(key *bitset.BitSet, value *trie.Node) error {
...
 		return nil, err
 	}
 
+	node := new(trie.Node)
 	err = t.txn.Get(dbKey, func(val []byte) error {
-		node = new(trie.Node)
 		return encoder.Unmarshal(val, node)
 	})
-	return
+	return node, err
```
omerfirmak, line 60: same, `return nil, err`

db/pebble/transaction.go
```diff
@@ -18,25 +18,30 @@ type Transaction struct {
...
 
 // Discard : see db.Transaction.Discard
-func (t *Transaction) Discard() (err error) {
+func (t *Transaction) Discard() error {
 	if t.batch != nil {
-		err = t.batch.Close()
+		if err := t.batch.Close(); err != nil {
+			return err
+		}
 		t.batch = nil
```
omerfirmak, line 26: This changes discard behavior on failure. batch & snapshot wont be set to nil if close fails. It might be better this way, let's see what others think.

IronGauntlets: The behaviour should remain the same because we are using nil to enforce calling Discard multiple times on the same transaction does nothing. If we don't set batch and snapshot to nil we guarantee second and subsequent calls to discard would error. Also, we risk causing deadlocks as we may not call unlock on line 36 when we need to.

db/pebble/db.go
```diff
@@ -64,16 +64,18 @@ func (db *DB) Close() error {
 }
 
 // View : see db.DB.View
-func (d *DB) View(fn func(txn db.Transaction) error) (err error) {
+func (d *DB) View(fn func(txn db.Transaction) error) error {
 	txn := d.NewTransaction(false)
+	var err error
 	defer db.CloseAndWrapOnError(txn.Discard, &err)
```
IronGauntlets, line 70: The behaviour of `db.CloseAndWrapOnError` has changed because the error returned by `fn(txn)` on line 72 will not be set to the error declared on line 69. Thus the error passed to `db.CloseAndWrapOnError` will always be nil.

This is the case in most places where `db.CloseAndWrapOnError` is being used. Either we change how we wrap errors or we set `err` correctly before returning. `db.CloseAndWrapOnError` relied on the named return parameter to work correctly. Since we are no longer using that `db.CloseAndWrapOnError` no longer works as expected.

db/pebble/transaction.go
omerfirmak: It lets you to retry when Discard returns an error tho., that is why I thought it might be better this way.

IronGauntlets: I think for retying we would require more information and it will have to depend on the error that pebble returns. This would be difficult to determine which errors are recoverable and which errors are not recoverable. Therefore, I would like the behaviour to remain the same.

If the behaviour changes then we need to update the tests accordingly.

omerfirmak: > It would be difficult to determine which errors are recoverable and which errors are not. 

that is not a concern of the pebble package tho. That responsibility lies on the caller.

Question here is do we give user the ability to retry or not. My 2 cents is that I think we should.

blockchain/blockchain.go
```diff
@@ -400,6 +400,10 @@ func blockByNumber(txn db.Transaction, number uint64) (*core.Block, error) {
 		block.Receipts = append(block.Receipts, receipt)
 	}
 
+	if err := iterator.Error(); err != nil {
```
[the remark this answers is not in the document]
joshklop: I agree this doesn't match the previous code. As far as I can tell, achieving equivalence would require 

1. changing `return nil, err` statements to something like `return nil, db.CloseAndWrapOnErr(closeFn, err)`
2. "manually" closing the resource and checking for error before returning from the happy path.

To avoid these code smells, we must always release the resource without error and catch any "internal database errors" before returning from the happy path. `defer resource.Discard()` and `resource.Error()` accomplish these respective goals.

I agree it's not a great solution, but I'm not sure how else to separate releasing resources from error handling. I'm definitely open to alternatives.

omerfirmak (approved):
LGTM

joshklop:
Fixed lint errors.

db/pebble/transaction.go
IronGauntlets: > having a dangling snapshot/batch open on the database which we have no reference to is not a pleasant situation to be in.

This is a good point. 

My concern was how would the caller decide that they need to retry without switching on a specific error.

IronGauntlets:
@joshklop please merge #667 which has been rebase onto main

PR #628, opened 2023-03-15
Pass in string instead of uint8 for log level

joshklop:


joshklop:
It works perfectly when I test it manually, but the tests aren't working. Will try again tomorrow.

joshklop:
Stayed up late and found the issue. Since `t.Run` is executed in a separate go routine, there is a race condition on the global `StarknetNode` variable. Removing `t.Run` makes the tests pass. The better solution is probably to use a mock instead of a spy and remove the exported global variable.

PR #631, opened 2023-03-15
Fix invalid params error

jelilat:
json RPC was returning `missing param in list` even when the call didn't require any parameters.

### DIFF
--- a/jsonrpc/server.go
+++ b/jsonrpc/server.go
@@ -311,7 +311,7 @@ func buildArguments(params, handler any, configuredParams []Parameter) ([]reflec
 		paramsList := params.([]any)
 
 		if len(paramsList) != handlerType.NumIn() {
-			return nil, errors.New("missing param in list")
+			return nil, errors.New("missing/unexpected params in list")
 		}
 
 		for i, param := range paramsList {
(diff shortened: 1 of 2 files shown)


omerfirmak (changes requested):
Please provide a testcase as well.

jelilat:
> Please provide a testcase as well.

Done

omerfirmak:
> > Please provide a testcase as well.
> 
> Done

Which one was failing in the existing version?

jelilat:
> > > Please provide a testcase as well.
> > 
> > 
> > Done
> 
> Which one was failing in the existing version?

`missing param`. I updated the error message for missing param, so I needed to update it in the test as well but didn't.

jelilat:
> What I mean is, can you provide a testcase that fails without the changes you made in the code here?

```
curl -X POST http://localhost:6060 -H 'Content-Type: application/json'  -d '{ "jsonrpc":"2.0", "method":"starknet_blockNumber", "params":["1"], "id":1 }'
```

jsonrpc/server.go
```diff
@@ -310,8 +311,10 @@ func buildArguments(params, handler any, configuredParams []Parameter) ([]reflec
 	case reflect.Slice:
 		paramsList := params.([]any)
 
-		if len(paramsList) != handlerType.NumIn() {
-			return nil, errors.New("missing param in list")
```
omerfirmak, line 314: I dont think we need to handle this in two different conditionals.
Let's just update the error message to be generic like; "missing/unexpected params in list" and we are good

jelilat: Done

IronGauntlets (approved):

PR #632, opened 2023-03-15
Allow instantiation of a Trie using Poseidon hash

omerfirmak:


core/trie/trie.go
```diff
@@ -38,14 +38,26 @@ type Trie struct {
...
 }
 
+type NewTrieFunc func(Storage, uint, *bitset.BitSet) *Trie
+
 func NewTrie(storage Storage, height uint, rootKey *bitset.BitSet) *Trie {
+	return newTrie(storage, height, rootKey, crypto.Pedersen)
+}
+
+func NewTriePoseidon(storage Storage, height uint, rootKey *bitset.BitSet) *Trie {
+	return newTrie(storage, height, rootKey, crypto.Poseidon)
```
omerfirmak, line 51: Uh, I really dont want to write tests for this just to be able to merge.

omerfirmak:
I will include this in to a bigger PR.

core/trie/node_test.go
```diff
@@ -170,5 +171,5 @@ func TestNodeHash(t *testing.T) {
 	}
 	path := bitset.FromWithLength(6, []uint64{42})
 
-	assert.Equal(t, true, expected.Equal(node.Hash(path)), "TestTrieNode_Hash failed")
+	assert.Equal(t, true, expected.Equal(node.Hash(path, crypto.Pedersen)), "TestTrieNode_Hash failed")
```
IronGauntlets, line 174: Please add a case for Poseidon

core/trie/trie.go
```diff
@@ -38,14 +38,26 @@ type Trie struct {
 	height  uint
 	rootKey *bitset.BitSet
 	storage Storage
+	hash    func(*felt.Felt, *felt.Felt) *felt.Felt
 }
 
+type NewTrieFunc func(Storage, uint, *bitset.BitSet) *Trie
+
 func NewTrie(storage Storage, height uint, rootKey *bitset.BitSet) *Trie {
```
IronGauntlets, line 46: Please rename `NewTrie` to `NewTriePedersen`

core/trie/trie.go
```diff
@@ -38,14 +38,26 @@ type Trie struct {
 	height  uint
 	rootKey *bitset.BitSet
 	storage Storage
+	hash    func(*felt.Felt, *felt.Felt) *felt.Felt
 }
 
+type NewTrieFunc func(Storage, uint, *bitset.BitSet) *Trie
```
IronGauntlets, line 44: I don't see any usage of this. If this is not being used please remove

IronGauntlets (changes requested):

core/trie/node_test.go
omerfirmak: That would be testing the Poseidon hash's implementation, not TrieNode.Hash. Testing for each dependency (hash function in this case) defeats the purpose of dependency injection.  

and we don't have a test vector for Poseidon tries.

core/trie/trie.go
omerfirmak: done

core/trie/trie.go
omerfirmak: done

core/trie/trie.go
```diff
@@ -38,14 +38,26 @@ type Trie struct {
 	height  uint
 	rootKey *bitset.BitSet
 	storage Storage
+	hash    func(*felt.Felt, *felt.Felt) *felt.Felt
```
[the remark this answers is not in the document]
omerfirmak: done

core/trie/trie.go
```diff
@@ -17,6 +17,8 @@ type Storage interface {
 	Delete(key *bitset.BitSet) error
 }
 
+type HashFunc func(*felt.Felt, *felt.Felt) *felt.Felt
+
```
IronGauntlets, line 21: Any particular reason this is public? 

We should only make it public if `NewTrie` is also public. Which is currently not the case.

IronGauntlets (changes requested):

core/trie/trie.go
omerfirmak: made private

IronGauntlets (approved):

PR #634, opened 2023-03-15
Node setup and configuration refactors

joshklop:
- Simplifies the cobra-centric flow for setting up the node. The key thing to notice here is how we are passing in a `node.Config` to `NewCmd`, populating it in `junoCmd.PreRunE`, and then running the node with the user-provided `run` function.
- Pass in log level and network as strings instead of uint8s. Not only is this more ergonomic, but it also prevents overflows.
- Remove unused configuration parameters.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,6 @@ docker run -it \
   nethermindeth/juno \
   --rpc-port 6060 \
   --db-path /var/lib/juno \
-  --verbosity 0 \
   --network 0
 ```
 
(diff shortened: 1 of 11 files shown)


wojciechos (approved):

jelilat (changes requested):
Should run `go mod tidy`

omerfirmak (approved):
Should be good to merge after addressing linter warnings.

jelilat (approved):
Nice work!

Should be good to go once the linting errors are fixed

PR #637, opened 2023-03-16
Update Rpc Errors

jelilat:


jsonrpc/server.go
```diff
@@ -254,19 +254,13 @@ func (s *Server) handleRequest(req *request) (*response, error) {
 
 	calledMethod, found := s.methods[req.Method]
 	if !found {
-		res.Error = &Error{
-			Code:    MethodNotFound,
-			Message: "method not found",
-		}
+		res.Error = rpcErr(MethodNotFound, "method not found")
```
omerfirmak, line 257: I think we can just pass `nil` as `data` here, right? It looks redundant.

omerfirmak (approved):

wojciechos (approved):

IronGauntlets (changes requested):

jsonrpc/server.go
omerfirmak: should be `nil` not empty string

jelilat: Done

jsonrpc/server.go
```diff
@@ -56,6 +41,21 @@ type Error struct {
 	Data    any    `json:"data,omitempty"`
 }
 
+func rpcErr(code int, data interface{}) *Error {
```
omerfirmak, line 44:
```suggestion
func rpcErr(code int, data any) *Error {
```

IronGauntlets (approved):

PR #638, opened 2023-03-16
Declare v2 and Sierra support

omerfirmak:


omerfirmak:
Disabled https://go-critic.com/overview.html#sloppyreassign

It was just so annoying and made no sense.

blockchain/encoder_initializer.go
```diff
@@ -0,0 +1,34 @@
...
+
+var once sync.Once
+
+func registerCoreTypesToEncoder() {
+	registerAndPanicOnError := func(t reflect.Type) {
+		err := encoder.RegisterType(t)
+		if err != nil {
+			panic(err)
+		}
+	}
```
joshklop, line 19: I think we can move the function body to line 31 and remove the function call.

Would it make more sense to have the registration happen on a call to `pebble.New`? I think that would allow us to remove `sync.Once` magic and keep database-specific encodings separate from the blockchain. I'm not sure if that's the right way to go and would appreciate your take.

clients/feeder/feeder.go
```diff
@@ -384,6 +400,26 @@ type ClassDefinition struct {
...
+		return err
+	}
+
+	if _, found := jsonMap["sierra_program"]; found {
+		c.V1 = new(SierraDefinition)
+		return json.Unmarshal(data, c.V1)
+	} else {
+		c.V0 = new(Cairo0Definition)
+		return json.Unmarshal(data, c.V0)
+	}
```
joshklop, line 420: Pedantic [styling](https://github.com/uber-go/guide/blob/master/style.md#reduce-nesting) suggestion.
```suggestion
        }
        c.V0 = new(Cairo0Definition)
        return json.Unmarshal(data, c.V0)
```

clients/feeder/feeder_test.go
```diff
@@ -311,7 +311,7 @@ func TestBlockWithSequencerAddressUnmarshal(t *testing.T) {
 	assert.Equal(t, "0x5dcd266a80b8a5f29f04d779c6b166b80150c24f2180a75e82427242dab20a9", block.SequencerAddress.String())
 }
 
-func TestClassUnmarshal(t *testing.T) {
+func TestClassV0Unmarshal(t *testing.T) {
```
joshklop, line 314: Could we use [this](https://github.com/starkware-libs/cairo/blob/main/crates/cairo-lang-starknet/test_data/hello_starknet.json) to test v1 classes?

core/class.go
```diff
@@ -52,3 +64,49 @@ func flatten(entryPoints []EntryPoint) []*felt.Felt {
...
+	Index    uint64
+	Selector *felt.Felt
+}
+
+func (c *Cairo1Class) Version() uint64 {
+	return 1
+}
+
+func (c *Cairo1Class) Hash() *felt.Felt {
+	contractClassVersion := fmt.Sprintf("CONTRACT_CLASS_V%s", c.SemanticVersion)
```
joshklop, line 92: We can save an allocation with:
```suggestion

```

joshklop, line 92: We can save an allocation with:
```suggestion

```

joshklop, line 92: We can save an allocation with:
```suggestion

```

core/class.go
```diff
@@ -52,3 +64,49 @@ func flatten(entryPoints []EntryPoint) []*felt.Felt {
...
+	return crypto.PoseidonArray(
+		new(felt.Felt).SetBytes([]byte(contractClassVersion)),
```
joshklop, line 94: I believe we can [avoid](https://github.com/uber-go/guide/blob/master/style.md#prefer-strconv-over-fmt) an allocation from `fmt.Sprintf`. 
```suggestion
	return crypto.PoseidonArray(
		new(felt.Felt).SetBytes([]byte("CONTRACT_CLASS_V"+c.SemanticVersion)),
```

joshklop (changes requested):
Not sure why the first review was closed before I finished. Must've mis-clicked somewhere.

clients/feeder/feeder_test.go
omerfirmak: Since that is not real data from a network, it might be hard to put it together with our existing test data. I thought https://github.com/NethermindEth/juno/pull/638/files#diff-a6ae96f1fa557bcf68f6d1e27521660124c4c1bbcaa8d46193ec26f7047245adR51 would be enough, since it uses the client already. Let me know what you think

blockchain/encoder_initializer.go
omerfirmak: We need to register some of our types because it is a limitation of our encoder library. DB doesnt really care about the data it is storing. For DB they are all byte arrays. 

It made more sense to me to be in blockchain since it is the one that actually uses the `encoder` package. For example; if someone uses our blockchain struct only as a library in their application and implement our `DB` interface themselves. They would still expect to be able to encode/decode these types fine.

`sync.Once` is only there because I wanted to allow multiple `blockchain` structs to be instantiated, without additional calls to `RegisterType`. `RegisterType` is stateful, so it is important that we call it once per type.

rpc/handlers.go
```diff
@@ -151,22 +151,16 @@ func adaptTransaction(t core.Transaction) *Transaction {
...
-		Type:      TxnInvoke,
-		Hash:      t.Hash(),
-		MaxFee:    t.MaxFee,
-		Version:   t.Version,
-		Signature: &sig,
-		Nonce:     t.Nonce,
-		Calldata:  &t.CallData,
-	}
-
-	if t.Version.IsZero() {
```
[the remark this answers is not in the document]
omerfirmak: This is rpc, so it does not affect the sync process. This if was only there to mitigate a feeder gateway quirk, it used to use ContractAddress field for SenderAddress in Invoke V1 txns

core/class.go
omerfirmak: Will apply since it is one less line of code, but I believe this will still allocate since strings are immutable in go.

starknetdata/feeder/feeder_test.go
```diff
@@ -206,7 +208,7 @@ func TestTransaction(t *testing.T) {
 		require.NoError(t, err)
 
 		assert.Equal(t, responseTx.Hash, invokeTx.Hash())
-		assert.Equal(t, responseTx.ContractAddress, invokeTx.ContractAddress)
+		assert.Equal(t, responseTx.SenderAddress, invokeTx.ContractAddress)
```
jelilat, line 211: Should it be this instead?
```suggestion
		assert.Equal(t, responseTx.ContractAddress, invokeTx.SenderAddress)
```

The 2 values here are nil and I'm not sure if a valid invoke transaction should have a nil sender. We should probably also modify our implementation to reflect same names as it were before. (e.g. `assert.Equal(t, responseTx.ContractAddress, invokeTx.ContractAddress)`).

clients/feeder/feeder_test.go
omerfirmak: done, https://github.com/NethermindEth/juno/pull/638/files#diff-ada8ad9f7998cd8be2d874e9605538aa33f4c2a0cfe7b8d16f921743073c1207R337

blockchain/encoder_initializer.go
joshklop: That makes sense. Can we inline `registerAndPanicOnError`?

joshklop (approved):

PR #639, opened 2023-03-16
replace gofumt with golangci-lint

omerfirmak:


Makefile
```diff
@@ -28,14 +28,20 @@ test-cover: ## tests with coverage
 	go test -coverpkg=./... -coverprofile=coverage/coverage.out -covermode=atomic ./...
 	go tool cover -html=coverage/coverage.out -o coverage/coverage.html
 
-install-deps: | install-gofumpt install-mockgen ## install some project dependencies
+install-deps: | install-gofumpt install-mockgen install-golangci-lint## install some project dependencies
```
IronGauntlets, line 31:
```suggestion
install-deps: | install-mockgen install-golangci-lint## install some project dependencies
```

IronGauntlets: Let's delete `all` target since it is not used any more in the juno-build.yml

jelilat (approved):

PR #645, opened 2023-03-16
Refactor Contract

IronGauntlets:


omerfirmak:
I think we should move towards a `Contract` instance being unambiguously "deployed".

omerfirmak:
I added a new commit, let me know what you think. Feel free to drop if it doesnt make sense, or squash in to the previous commit. I based it on the non-refactored `contract.go` so it might be better to checkout to the commit and take a look locally, instead of trying to understand the diff on Github.

core/contract.go
```diff
@@ -106,73 +91,79 @@ func (c *Contract) Nonce() (nonce *felt.Felt, err error) {
...
+}
 
-	return c.storage.Get(key)
+// Root returns the root of the contract storage.
+func (c *Contract) Root() (*felt.Felt, error) {
+	if cStorage, err := storage(c.Address, c.txn); err != nil {
+		return nil, err
+	} else {
+		return cStorage.Root()
+	}
```
IronGauntlets: @omerfirmak, it is idiomatic to do the following:
```go
func (c *Contract) Root() (*felt.Felt, error) {
	if cStorage, err := storage(c.Address, c.txn); err != nil {
		return nil, err
	} 
	return cStorage.Root()
}
```

Also when all the `if`, `else if` and `else` statements are returning there is no need to have `if else/ else` statements. We should prefer the following:

```go
classHashKey := db.ContractClassHash.Key(addr.Marshal())
err := txn.Set(classHashKey, classHash.Marshal())
if err != nil {
	return nil, err
}

c, err := NewContract(addr, txn)
if err != nil {
	return nil, err
}

err = c.UpdateNonce(&felt.Zero)
if err != nil {
	return nil, err
}

return c, nil
```

Over: 

```go
classHashKey := db.ContractClassHash.Key(addr.Marshal())
if err := txn.Set(classHashKey, classHash.Marshal()); err != nil {
	return nil, err
} else if c, err := NewContract(addr, txn); err != nil {
	return nil, err
} else if err = c.UpdateNonce(&felt.Zero); err != nil {
	return nil, err
} else {
	return c, nil
}
```
Please refer to Effective of [here](https://go.dev/doc/effective_go#if) and this [article](https://www.calhoun.io/one-liner-if-statements-with-errors/).

omerfirmak: Got it, thanks!

core/contract.go
```diff
@@ -67,83 +120,80 @@ func (c *Contract) UpdateNonce(nonce *felt.Felt) error {
...
-		return storage.Root()
+		return cStorage.Root()
 	}
 }
 
-// UpdateStorage applies a change-set to the contract storage.
-func (c *Contract) UpdateStorage(diff []StorageDiff) error {
-	storage, err := c.Storage()
+// Update applies a change-set to the contract storage.
+func (c *Contract) Update(diff []StorageDiff) error {
```
omerfirmak, line 137: I thought UpdateStorage made more sense because we have UpdateNonce as well.

omerfirmak (approved):

core/contract.go
IronGauntlets: I have changed `Update` to `UpdateStorage` and `Get` to `Storage`

PR #647, opened 2023-03-20
Implement History

omerfirmak:
Lets us log the changes in the state to be retrieved later by height

We can extend this by adding more functions like `LogStorageValue` and `StorageValueAt`. For example; nonces, class hashes (they are mutable after v0.11.0 with replace syscall) etc

### DIFF
--- a/core/history.go
+++ b/core/history.go
@@ -1,0 +1,78 @@
+package core
+
+import (
+	"bytes"
+	"encoding/binary"
+	"errors"
+
+	"github.com/NethermindEth/juno/core/felt"
+	"github.com/NethermindEth/juno/db"
+)
+
+type History struct {
+	txn db.Transaction
+}
+
+func NewHistory(txn db.Transaction) *History {
+	return &History{txn: txn}
+}
+
+func (h *History) logOldValue(key, value []byte, height uint64) error {
+	return h.txn.Set(binary.BigEndian.AppendUint64(key, height), value)
+}
+
+func (h *History) valueAt(key []byte, height uint64) ([]byte, error) {
+	it, err := h.txn.NewIterator()
+	if err != nil {
+		return nil, err
+	}
+
+	for it.Seek(binary.BigEndian.AppendUint64(key, height)); it.Valid(); it.Next() {
+		seekedKey := it.Key()
+		// seekedKey size should be `len(key) + sizeof(uint64)` and seekedKey should match key prefix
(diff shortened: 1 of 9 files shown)


omerfirmak:
Towards #670

core/history_test.go
```diff
@@ -0,0 +1,68 @@
...
+	"github.com/NethermindEth/juno/core/felt"
+	"github.com/NethermindEth/juno/db/pebble"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestStorageValueAt(t *testing.T) {
+	testDB := pebble.NewMemTest()
+	txn := testDB.NewTransaction(true)
+	defer func() {
```
IronGauntlets, line 16: Please use `t.Cleanup()`

omerfirmak: done

core/history_test.go
```diff
@@ -0,0 +1,68 @@
...
+	t.Run("get value on height that change happened ", func(t *testing.T) {
+		oldValue, err := history.ContractStorageAt(contractAddres, location, 5)
+		require.NoError(t, err)
+		assert.Equal(t, value, oldValue)
+
+		_, err = history.ContractStorageAt(contractAddres, location, 10)
+		assert.EqualError(t, err, "check head state")
+	})
+
+	t.Run("get value between after height 10 ", func(t *testing.T) {
```
jelilat, line 59:
```suggestion
	t.Run("get value after height 10 ", func(t *testing.T) {
```

core/history.go
```diff
@@ -0,0 +1,78 @@
...
+
+	for it.Seek(binary.BigEndian.AppendUint64(key, height)); it.Valid(); it.Next() {
+		seekedKey := it.Key()
+		// seekedKey size should be `len(key) + sizeof(uint64)` and seekedKey should match key prefix
+		if len(seekedKey) != len(key)+8 || !bytes.HasPrefix(seekedKey, key) {
+			break
+		}
+
+		seekedHeight := binary.BigEndian.Uint64(seekedKey[len(key):])
+		if seekedHeight < height {
```
jelilat, line 38: Can't imagine a case where this could happen since our implementation only seeks forward `SeekGE`

jelilat: On another note, I think it's good to check for it regardless

omerfirmak: Yeah, I think you are right. Let's keep it for the sake of being more expressive, it might help down the line when updating the implementation.

omerfirmak:
Closing in favor of #707

PR #648, opened 2023-03-20
Strictly align our errors with uber-go guidelines

omerfirmak:
https://github.com/uber-go/guide/blob/master/style.md#errors

### DIFF
--- a/core/transaction.go
+++ b/core/transaction.go
@@ -12,8 +12,6 @@ import (
 	"github.com/ethereum/go-ethereum/common"
 )
 
-var ErrUnknownTransaction = errors.New("unknown transaction")
-
 type Event struct {
 	Data []*felt.Felt
 	From *felt.Felt
@@ -220,7 +218,7 @@ func transactionHash(transaction Transaction, n utils.Network) (*felt.Felt, erro
 	case *DeployAccountTransaction:
 		return deployAccountTransactionHash(t, n)
 	default:
-		return nil, ErrUnknownTransaction
+		return nil, errors.New("unknown transaction")
 	}
 }
 
--- a/core/transaction.go
+++ b/core/transaction.go
@@ -14,16 +14,6 @@ import (
 
 var ErrUnknownTransaction = errors.New("unknown transaction")
 
-type ErrInvalidTransactionVersion struct {
-	t Transaction
-	v string
-}
-
-func (e ErrInvalidTransactionVersion) Error() string {
-	return fmt.Sprintf("invalid Transaction verions. Type: %v, Version: %v", reflect.TypeOf(e.t),
-		e.t)
-}
-
 type Event struct {
 	Data []*felt.Felt
 	From *felt.Felt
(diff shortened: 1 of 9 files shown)


omerfirmak:
I am not opinionated at all on this matter. So please make any changes (including dropping commits) you deem necessary before approving, instead of leaving requests.

omerfirmak:
Rebased

core/transaction.go
jelilat, line 1: Random. I think this should be `not matched,  dynamic error message`

core/trie/node_test.go
jelilat, line 1: We probably shouldn't be adding unrelated changes

jelilat (commented):
Random. I think this should be `not matched, dynamic error message`

core/transaction.go
```diff
@@ -233,6 +223,10 @@ var (
 	deployAccountFelt = new(felt.Felt).SetBytes([]byte("deploy_account"))
 )
 
+func errInvalidTransactionVersion(t Transaction, version *felt.Felt) error {
+	return fmt.Errorf("invalid Transaction (type: %v) verion: %v", reflect.TypeOf(t), version.Text(felt.Base10))
+}
+
```
IronGauntlets, line 227: let's create the following error (please see error naming section in uber-go):

```go
type invalidTransactionVersionError struct {	
	t Transaction	
	v string	
}	

func (e ErrInvalidTransactionVersion) Error() string {	
	return fmt.Sprintf(""invalid Transaction (type: %v) verion: %v", reflect.TypeOf(t), version.Text(felt.Base10))
}
```

And add a test for an unsupported version which uses require/assert's `ErrorIs`/`ErrorAs`.

core/transaction.go
```diff
@@ -320,23 +312,23 @@ func deployAccountTransactionHash(d *DeployAccountTransaction, n utils.Network)
 			d.Nonce,
 		), nil
 	}
-	return nil, ErrInvalidTransactionVersion{d, d.Version.Text(felt.Base10)}
+	return nil, errInvalidTransactionVersion(d, d.Version)
 }
 
-type ErrCantVerifyTransactionHash struct {
+type CantVerifyTransactionHashError struct {
```
jelilat, line 318: @IronGauntlets thought we wanted to follow the convention `Err...`? Uber Guidelines also suggested the same https://github.com/uber-go/guide/blob/master/style.md#error-naming

IronGauntlets: yeah you are right I forgot to make it private

IronGauntlets:
Please replace `ErrorEqual` with `ErrorAs`/`ErrorIs` where appropriate, this will fulfill the matching condition allowing us to create private and public package-level errors.

omerfirmak:
> Please replace `ErrorEqual` with `ErrorAs`/`ErrorIs` where appropriate, this will fulfill the matching condition allowing us to create private and public package-level errors.

I dont think matching them in tests fulfills the requirement.

core/transaction.go
omerfirmak: We shouldn't create typed errors unless we are explicitly handling them somewhere. Checking them in tests does not warrant a typed error since they are not really "handling" the error.

IronGauntlets (approved):

PR #650, opened 2023-03-21
Fix Node Equality Check

jelilat:


omerfirmak:
Do you think we can get rid of node.Equal and just use assert.DeepEqual in tests?

omerfirmak (approved):

jelilat:
> Do you think we can get rid of node.Equal and just use assert.Equal in tests?

Yes, we can actually. It's only used in tests and assert works well in those cases

joshklop (approved):

PR #651, opened 2023-03-21
Verify block version on blockchain.Store

omerfirmak:


blockchain/blockchain.go
```diff
@@ -236,6 +239,14 @@ func (b *Blockchain) VerifyBlock(block *core.Block) error {
 }
 
 func (b *Blockchain) verifyBlock(txn db.Transaction, block *core.Block) error {
+	if len(block.ProtocolVersion) > 0 {
```
wojciechos, line 242: what do you think about extracting that logic into a separate function? 
Then it could look like this:

```
	if err := checkBlockVersion(block); err != nil {
		return err
	}
```

wojciechos (approved):

blockchain/blockchain.go
```diff
@@ -237,6 +240,14 @@ func (b *Blockchain) VerifyBlock(block *core.Block) error {
...
+		if blockVer, err := semver.NewVersion(block.ProtocolVersion); err != nil {
+			return err
+		} else if blockVer.GreaterThan(supportedStarknetVersion) {
+			return errors.New("unsupported block version")
+		}
```
IronGauntlets, line 248:
```suggestion
		blockVer, err := semver.NewVersion(block.ProtocolVersion)
		if err != nil {
			return err
		}
		if blockVer.GreaterThan(supportedStarknetVersion) {
			return errors.New("unsupported block version")
		}
```

IronGauntlets (approved):

PR #652, opened 2023-03-21
Refactor state

IronGauntlets:


core/contract.go
```diff
@@ -126,11 +124,40 @@ func (c *Contract) ClassHash() (*felt.Felt, error) {
...
-	if cStorage, err := storage(c.Address, c.txn); err != nil {
+	cStorage, err := storage(c.Address, c.txn)
+	if err != nil {
+		return nil, err
+	}
+	return cStorage.Root()
+}
+
+// Commitment returns contract storage commitment
+func (c *Contract) Commitment() (*felt.Felt, error) {
```
omerfirmak, line 135: I still think commitment is a state thing. Contract has no use for commitment, but state does.

core/contract_test.go
```diff
@@ -241,3 +246,32 @@ func TestUpdateAndGet(t *testing.T) {
 		assert.Equal(t, new(felt.Felt), sRoot)
 	})
 }
+
+func TestCommitment(t *testing.T) {
+	// Copied and pasted from contract.go:161 for testing purposes.
+	calculateContractCommitment := func(storageRoot, classHash, nonce *felt.Felt) *felt.Felt {
```
omerfirmak, line 252: The fact that we had to do this strengthens my thoughts on commitment calculation belonging to state. Commitment calculation was tested thru a Update call already.

core/state.go
```diff
@@ -25,11 +20,10 @@ type ErrMismatchedRoot struct {
 }
 
 func (e *ErrMismatchedRoot) Error() string {
-	newOld := " new "
 	if e.IsOld {
-		newOld = " old "
+		return fmt.Sprintf("state's current root: %s does not match state update's old root: %s", e.Got.String(), e.Want.String())
 	}
-	return fmt.Sprintf("mismatched %s root: want %s, got %s", newOld, e.Want.String(), e.Got.String())
+	return fmt.Sprintf("state's new root: %s does not match state update's new root: %s", e.Got.String(), e.Want.String())
```
omerfirmak, line 26: We should get rid of ErrMismatchedRoot, as per uber-go standards.

core/contract.go
IronGauntlets: reverted

core/contract_test.go
IronGauntlets: reverted

core/state.go
IronGauntlets: done

omerfirmak (approved):
LGTM

joshklop (approved):

PR #653, opened 2023-03-22
Verify Class Hashes

jelilat:


blockchain/blockchain.go
```diff
@@ -437,14 +437,42 @@ func stateUpdateByHash(txn db.Transaction, hash *felt.Felt) (update *core.StateU
...
+		vErrs = append(vErrs, ErrIncompatibleBlockAndStateUpdate{
+			errors.New("block's GlobalStateRoot does not match state update's NewRoot"),
+		})
+		return vErrs
 	}
-	return core.VerifyBlockHash(block, b.network)
+
+	errCh := make(chan error, 2)
+
+	go func() {
```
omerfirmak, line 458: we handle parallelism at sync level, so we don't really need these

blockchain/blockchain.go
```diff
@@ -449,14 +449,27 @@ func stateUpdateByHash(txn db.Transaction, hash *felt.Felt) (*core.StateUpdate,
 }
 
 // SanityCheckNewHeight checks integrity of a block and resulting state update
-func (b *Blockchain) SanityCheckNewHeight(block *core.Block, stateUpdate *core.StateUpdate) error {
+func (b *Blockchain) SanityCheckNewHeight(block *core.Block, stateUpdate *core.StateUpdate, classes map[felt.Felt]core.Class,
+) []error {
+	var vErrs []error
```
IronGauntlets, line 454: It doesn't make sense to wrap errors and then also have a list of errors. We should do one or the other.

sync/sync.go
```diff
@@ -84,22 +84,30 @@ func (s *Synchronizer) fetcherTask(ctx context.Context, height uint64, verifiers
...
+						}
+					} else if errors.As(err, new(core.CantVerifyTransactionHashError)) {
+						for ; err != nil; err = errors.Unwrap(err) {
+							s.log.Debugw("Sanity checks failed", "number", block.Number, "hash",
+								block.Hash.ShortString(), "error", err.Error())
+						}
+					} else {
+						s.log.Warnw("Sanity checks failed", "number", block.Number, "hash", block.Hash.ShortString())
+						resetStreams()
+						return
```
IronGauntlets, line 109: Since we ignore the error which fails class hash and transaction verification. Do you think it would be best to log the errors in `SanityCheckNewHeight` and return nil if class or transaction verification fails? 

This would mean that we would have to write separate tests for transaction verification as you have done for class.

core/class.go
```diff
@@ -112,3 +115,80 @@ func flattenSierraEntryPoints(entryPoints []SierraEntryPoint) []*felt.Felt {
...
+	var head *ErrCantVerifyClassHash
+	for hash, class := range classes {
+		if err := verifyClassHash(class, &hash); err != nil {
+			err.next = head
+			head = err
+		}
+	}
+
+	if head != nil {
+		return head
```
IronGauntlets, line 190:
```suggestion
		return *head
```

`error` accepts both `*ErrCantVerifyClassHash` and `ErrCantVerifyClassHash`, however `Is` and `As` matching for some reason frails if it is not `ErrCantVerifyClassHash`

IronGauntlets (changes requested):

sync/sync.go
jelilat: Sounds good to me.

blockchain/blockchain.go
jelilat: Could we then log errors here like you recommended?

core/class.go
jelilat: Fixed this

sync/sync.go
jelilat: Done

blockchain/blockchain.go
```diff
@@ -454,14 +454,35 @@ func stateUpdateByHash(txn db.Transaction, hash *felt.Felt) (*core.StateUpdate,
 }
 
 // SanityCheckNewHeight checks integrity of a block and resulting state update
-func (b *Blockchain) SanityCheckNewHeight(block *core.Block, stateUpdate *core.StateUpdate) error {
+func (b *Blockchain) SanityCheckNewHeight(block *core.Block, stateUpdate *core.StateUpdate, classes map[felt.Felt]core.Class,
+	log utils.SimpleLogger,
```
IronGauntlets, line 458: We should pass the logger to blockchain struct rather than passing it to a function

blockchain/blockchain.go
```diff
@@ -454,14 +454,35 @@ func stateUpdateByHash(txn db.Transaction, hash *felt.Felt) (*core.StateUpdate,
...
 	if !block.Hash.Equal(stateUpdate.BlockHash) {
 		return errors.New("block hashes do not match")
 	}
 	if !block.GlobalStateRoot.Equal(stateUpdate.NewRoot) {
 		return errors.New("block's GlobalStateRoot does not match state update's NewRoot")
 	}
-	return core.VerifyBlockHash(block, b.network)
+
+	if cErr := core.VerifyClassHashes(classes); cErr != nil {
+		for ; cErr != nil; cErr = errors.Unwrap(cErr) {
```
IronGauntlets, line 468: Is there a reason why you are not checking the error type here?

core/class.go
```diff
@@ -112,3 +115,76 @@ func flattenSierraEntryPoints(entryPoints []SierraEntryPoint) []*felt.Felt {
...
+func (e CantVerifyClassHashError) Unwrap() error {
+	if e.next != nil {
+		return *e.next
+	}
+	return nil
+}
+
+func (e CantVerifyClassHashError) Error() string {
+	return fmt.Sprintf("cannot verify class hash: %s", e.hashFailure)
+}
```
jelilat, line 145: With this change, if we pass in something like this `CantVerifyClassHashError{}`, we'll have `cannot verify class hash: %!s(<nil>)`. Is that okay?

@IronGauntlets

IronGauntlets: Yes because we an in complete control of creating `CantVerifyClassHashError` and no one outside of the package should create this error. Therefore, any and all instances of `CantVerifyClassHashError` will habe non-nill `hashFailure`.

core/class.go
```diff
@@ -112,3 +115,76 @@ func flattenSierraEntryPoints(entryPoints []SierraEntryPoint) []*felt.Felt {
...
+
+func verifyClassHash(c Class, hash *felt.Felt) *CantVerifyClassHashError {
+	if c == nil {
+		return &CantVerifyClassHashError{
+			hashFailure: fmt.Errorf("class is nil"),
+		}
+	}
+
+	cHash, err := cairoClassHash(c)
```
omerfirmak, line 158: `Class` is an interface, we can just call `c.Hash()`

omerfirmak: done

omerfirmak (approved):

PR #657, opened 2023-03-23
Fix linting errors

IronGauntlets:


clients/feeder/feeder.go
```diff
@@ -171,10 +172,21 @@ func (c *Client) get(ctx context.Context, queryURL string) ([]byte, error) {
...
 			res, err = c.client.Do(req)
-			if err == nil && res != nil && res.StatusCode == http.StatusOK {
-				return io.ReadAll(res.Body)
-			} else if res != nil && res.StatusCode != http.StatusOK {
-				err = errors.New(res.Status)
+			if err == nil {
+				if res.StatusCode == http.StatusOK {
+					resBytes, err = io.ReadAll(res.Body)
+					if err != nil {
+						return nil, err
```
omerfirmak, line 179: Close missing here?

I think we can just defer `res.Body.Close()` if `err == nil`

core/crypto/poseidon_hash.go
```diff
@@ -41,6 +43,7 @@ func round(state []felt.Felt, full bool, index int) {
 }
 
 func hadesPermutation(state []felt.Felt) {
+	initialiseRoundKeys.Do(setRoundKeys)
```
omerfirmak, line 46: we should benchmark to see if this affects performance

encoder/encoder.go
```diff
@@ -38,22 +41,20 @@ func RegisterType(rType reflect.Type) error {
...
 	return nil
 }
 
-func init() {
-	initEncModes()
-}
-
 // Marshal returns encoding of param v
 func Marshal(v any) ([]byte, error) {
+	initialiseEncoder.Do(initEncAndDecModes)
```
omerfirmak, line 51: This might not work with https://github.com/NethermindEth/juno/commit/3fc38a92541497f3519ba21c19b9a0d0efbec0cb, we should check

omerfirmak (commented):
🫡

core/crypto/poseidon_hash.go
omerfirmak: fast path of `Do` is just an `atomic.Load`, so we are fine

clients/feeder/feeder.go
omerfirmak: > I think we can just defer res.Body.Close() if err == nil

lol, triggers another lint error.

pprof/pprof.go
```diff
@@ -3,12 +3,12 @@ package pprof
 import (
 	"context"
 	"net/http"
+	// #nosec G108
```
jelilat, line 6: What security issue was it complaining about here?

rpc/handlers.go
```diff
@@ -221,18 +221,20 @@ func (h *Handler) TransactionByHash(hash *felt.Felt) (*Transaction, *jsonrpc.Err
 	return adaptTransaction(txn), nil
 }
 
-// BlockTransactionCount https://github.com/starkware-libs/starknet-specs/blob/a789ccc3432c57777beceaa53a34a7ae2f25fda0/api/starknet_api_openrpc.json#L373
-func (h *Handler) BlockTransactionCount(id *BlockId) (uint64, *jsonrpc.Error) {
-	header, err := h.blockHeaderById(id)
+// BlockTransactionCount
+//
+// https://github.com/starkware-libs/starknet-specs/blob/a789ccc3432c57777beceaa53a34a7ae2f25fda0/api/starknet_api_openrpc.json#L373
```
jelilat, line 226: I understand that the function names are descriptive. But can we follow this convention for the functions here?
```suggestion
// BlockTransactionCount returns the number of transactions in a block
// identified by the given BlockID.
//
// It follows the specification found here:
// https://github.com/starkware-libs/starknet-specs/blob/a789ccc3432c57777beceaa53a34a7ae2f25fda0/api/starknet_api_openrpc.json#L373
```

jelilat: Or maybe in another PR since this is just for fixing linting errors.

pprof/pprof.go
IronGauntlets: This was just a reordering of the imports. I believe @wojciechos can give more context on the security issue

rpc/handlers.go
IronGauntlets: In another pr would be best. I don't want to rebase 17 commits

IronGauntlets:
rebased

jelilat (approved):

PR #660, opened 2023-03-24
Testing related refactors

omerfirmak:
See commit by commit


clients/feeder/feeder_test.go
```diff
@@ -15,95 +13,15 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func testClient(clientURL string) *feeder.Client {
-	return feeder.NewClient(clientURL).WithBackoff(feeder.NopBackoff).WithMaxRetries(0)
-}
+func TestDeclareTransactionUnmarshal(t *testing.T) {
+	client, closeFn := feeder.NewTestClient(utils.MAINNET)
+	defer closeFn()
```
joshklop, line 18: Maybe we can use `t.Cleanup` instead of `defer`? I think we can do this in other feeder tests as well.

omerfirmak: Yeap, done

wojciechos (approved):

IronGauntlets (approved):

PR #668, opened 2023-03-27
Trie height

joshklop:


core/trie/trie.go
```diff
@@ -3,13 +3,35 @@ package trie
...
+	return fmt.Sprintf("key %s exceeds trie height %d", e.key.String(), maxHeight)
+}
+
+type errMaxHeight struct {
+	height uint
+}
+
+func (e *errMaxHeight) Error() string {
+	return fmt.Sprintf("max trie height is %d, got: %d", maxHeight, e.height)
+}
```
jelilat, line 33: We should follow [this](https://github.com/uber-go/guide/blob/master/style.md#error-types) convention for the errors here as well

omerfirmak: Yeah, we don't need these. We should use `fmt.Errorf`

core/trie/trie.go
```diff
@@ -43,27 +65,38 @@ type Trie struct {
...
 
-func NewTriePoseidon(storage Storage, height uint, rootKey *bitset.BitSet) *Trie {
+func NewTriePoseidon(storage Storage, height uint, rootKey *bitset.BitSet) (*Trie, error) {
 	return newTrie(storage, height, rootKey, crypto.Poseidon)
 }
 
-func newTrie(storage Storage, height uint, rootKey *bitset.BitSet, hash hashFunc) *Trie {
-	// Todo: set max height to 251 and set max key value accordingly
+func newTrie(storage Storage, height uint, rootKey *bitset.BitSet, hash hashFunc) (*Trie, error) {
+	if height > maxHeight {
```
omerfirmak, line 77: `height > fp.Bits`

core/trie/trie.go
```diff
@@ -43,27 +65,38 @@ type Trie struct {
...
+		return nil, &errMaxHeight{height}
+	}
+
+	// maxKey is 2^maxHeight - 1
+	maxKey = new(felt.Felt).Exp(new(felt.Felt).SetUint64(2), new(big.Int).SetUint64(uint64(maxHeight)))
```
omerfirmak, line 82: this should be a member of `Trie`, if not this applies for all tries of all height

omerfirmak (changes requested):

core/trie/trie.go
```diff
@@ -3,13 +3,35 @@ package trie
...
+	"math/big"
 	"strings"
 
 	"github.com/NethermindEth/juno/core/crypto"
 	"github.com/NethermindEth/juno/core/felt"
 	"github.com/bits-and-blooms/bitset"
 )
 
+var (
+	maxHeight uint = 251
```
omerfirmak, line 15: `maxHeight = felt.Bits`

core/trie/trie.go
```diff
@@ -157,8 +190,13 @@ func (t *Trie) Get(key *felt.Felt) (*felt.Felt, error) {
 }
 
 // Put updates the corresponding `value` for a `key`
+//
+//nolint:gocyclo
 func (t *Trie) Put(key, value *felt.Felt) (*felt.Felt, error) {
-	// Todo: check key is not bigger than max key value for a trie height.
+	if key.Cmp(maxKey) > 0 {
+		return nil, &errMaxKey{key: key}
+	}
```
IronGauntlets, line 198: you can extract this into a method and it should solve the gocylo issue

joshklop: I think we'll still need to check the output of the method, so the number of branches will remain the same. I'll think about how to cleanly refactor `Put` to avoid the gocyclo issue.

joshklop:
Pushed my changes that should resolve most comments. I'll need some more time to think about cleanly refactoring `trie.Put` to avoid gocyclo lint errors.

IronGauntlets:
> Pushed my changes that should resolve most comments. I'll need some more time to think about cleanly refactoring trie.Put to avoid gocyclo lint errors.

Let's create an issue for it and fix it later. For now, keep the `nolint` comment.

core/trie/trie_test.go
```diff
@@ -280,3 +280,11 @@ func TestOldData(t *testing.T) {
 		return nil
 	}))
 }
+
+func TestMaxTrieHeight(t *testing.T) {
+	t.Run("create trie with invalid height", func(t *testing.T) {
+		require.Error(t, trie.RunOnTempTrie(felt.Bits+1, func(_ *trie.Trie) error {
```
IronGauntlets, line 286: this should be `assert`

core/trie/trie_test.go
```diff
@@ -287,4 +287,14 @@ func TestMaxTrieHeight(t *testing.T) {
 			return nil
 		}))
 	})
+
+	t.Run("insert invalid key", func(t *testing.T) {
+		height := uint64(felt.Bits)
+		require.NoError(t, trie.RunOnTempTrie(uint(height), func(tt *trie.Trie) error {
```
IronGauntlets, line 293: This should be `assert` as well

IronGauntlets (changes requested):
Please make the above changes

core/trie/trie.go
```diff
@@ -164,8 +171,13 @@ func (t *Trie) Get(key *felt.Felt) (*felt.Felt, error) {
 }
 
 // Put updates the corresponding `value` for a `key`
+//
+//nolint:gocyclo
 func (t *Trie) Put(key, value *felt.Felt) (*felt.Felt, error) {
-	// Todo: check key is not bigger than max key value for a trie height.
+	if key.Cmp(t.maxKey) > 0 {
```
omerfirmak, line 177: I expect this to have negative effects on performance, but hoping it to be negligible.
`Cmp` implementation is bunch of `if` statements, that many conditionals are bound to mess up branch prediction.

IronGauntlets: I believe modern compilers try to account for this and performance impact is minimal.

IronGauntlets (approved):

PR #688, opened 2023-03-28
Fix pprof test

wojciechos:


pprof/pprof_test.go
IronGauntlets, line 1: Can we make the following change to the test since there is only one case now:

```go
func TestPprofServer(t *testing.T) {
	port := uint16(9050)
	log := utils.NewNopZapLogger()
	url := fmt.Sprintf("http://localhost:%d/debug/pprof/", port)

	profiler := pprof.New(port, log)
	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
	t.Cleanup(cancel)

	go func() {
		err := profiler.Run(ctx)
		require.NoError(t, err)
	}()

	waitForServerReady(t, url, time.Second)
}
```

IronGauntlets (approved):

pprof/pprof_test.go
wojciechos: done

PR #691, opened 2023-03-28
Refactor tests

IronGauntlets:
Please see the commit history.


rpc/handlers.go
```diff
@@ -182,34 +182,35 @@ func adaptDeclareTransaction(t *core.DeclareTransaction) *Transaction {
 
 	if t.Version.Equal(new(felt.Felt).SetUint64(2)) {
 		txn.CompiledClassHash = nil // todo: add when we have support for Declare V2
-	} else if !t.Version.IsZero() && !t.Version.IsOne() {
```
wojciechos, line 185: Was the removal of this specific 'else if' intentional?

wojciechos (approved):

rpc/handlers.go
```diff
@@ -182,34 +182,35 @@ func adaptDeclareTransaction(t *core.DeclareTransaction) *Transaction {
...
 	}
+
 	return txn
 }
 
 func (h *Handler) blockByID(id *BlockID) (*core.Block, error) {
 	if id.Latest {
 		return h.bcReader.Head()
-	} else if id.Hash != nil {
+	}
```
omerfirmak, line 193: Let's make this a `switch`? The conditions we are testing are mutually exclusive, so I believe we should reflect that it code.

rpc/handlers.go
```diff
@@ -182,34 +182,35 @@ func adaptDeclareTransaction(t *core.DeclareTransaction) *Transaction {
...
-		return h.bcReader.BlockByNumber(id.Number)
 	}
+	return h.bcReader.BlockByNumber(id.Number)
 }
 
 func (h *Handler) blockHeaderByID(id *BlockID) (*core.Header, error) {
 	if id.Latest {
 		return h.bcReader.HeadsHeader()
-	} else if id.Hash != nil {
+	}
```
omerfirmak, line 206: this as well

starknetdata/feeder/feeder.go
```diff
@@ -267,11 +267,13 @@ func (f *Feeder) Class(ctx context.Context, classHash *felt.Felt) (core.Class, e
 
 	if response.V1 != nil {
 		return adaptCairo1Class(response.V1)
-	} else if response.V0 != nil {
+	}
+
+	if response.V0 != nil {
```
omerfirmak, line 272: this as well

omerfirmak (changes requested):

rpc/handlers.go
IronGauntlets: Yes since we trust the source of information.

rpc/handlers.go
IronGauntlets: done

rpc/handlers.go
IronGauntlets: done

starknetdata/feeder/feeder.go
IronGauntlets: done

omerfirmak (approved):

Issue #690, opened 2023-03-28
Sync Tests Data Race


IronGauntlets:
```
==================
WARNING: DATA RACE
Write at 0x00c00037bfd8 by goroutine 47623:
  github.com/NethermindEth/juno/sync.TestSyncBlocks.func4.2()
      /Users/irongauntlets/projects/juno/sync/sync_test.go:118 +0x304
  runtime.call32()
      /usr/local/go/src/runtime/asm_arm64.s:479 +0x74
  reflect.Value.Call()
      /usr/local/go/src/reflect/value.go:370 +0x90
  github.com/golang/mock/gomock.(*Call).DoAndReturn.func1()
      /Users/irongauntlets/go/pkg/mod/github.com/golang/mock@v1.6.0/gomock/call.go:132 +0x61c
  github.com/golang/mock/gomock.(*Controller).Call()
      /Users/irongauntlets/go/pkg/mod/github.com/golang/mock@v1.6.0/gomock/controller.go:251 +0xcc
  github.com/NethermindEth/juno/mocks.(*MockStarknetData).StateUpdate()
      /Users/irongauntlets/projects/juno/mocks/mock_starknetdata.go:72 +0x158
  github.com/NethermindEth/juno/sync.(*Synchronizer).fetcherTask()
      /Users/irongauntlets/projects/juno/sync/sync.go:53 +0xcc
  github.com/NethermindEth/juno/sync.(*Synchronizer).syncBlocks.func1()
      /Users/irongauntlets/projects/juno/sync/sync.go:151 +0x68
  github.com/sourcegraph/conc/stream.(*Stream).Go.func1()
      /Users/irongauntlets/go/pkg/mod/github.com/sourcegraph/conc@v0.2.0/stream/stream.go:84 +0x6c
  github.com/sourcegraph/conc/pool.(*Pool).worker()
      /Users/irongauntlets/go/pkg/mod/github.com/sourcegraph/conc@v0.2.0/pool/pool.go:137 +0xac
  github.com/sourcegraph/conc/pool.(*Pool).worker-fm()
      <autogenerated>:1 +0x34
  github.com/sourcegraph/conc/panics.(*Catcher).Try()
      /Users/irongauntlets/go/pkg/mod/github.com/sourcegraph/conc@v0.2.0/panics/panics.go:23 +0x60
  github.com/sourcegraph/conc.(*WaitGroup).Go.func1()
[...]

PR #695, opened 2023-03-29
Fix race condition

jelilat:
Fixes https://github.com/NethermindEth/juno/issues/690


sync/sync_test.go
```diff
@@ -86,9 +87,13 @@ func TestSyncBlocks(t *testing.T) {
 
 		mockSNData := mocks.NewMockStarknetData(mockCtrl)
 
+		var mutex sync.Mutex
```
IronGauntlets, line 90: Let's use a `RWMutex`

IronGauntlets (approved):

sync/sync_test.go
```diff
@@ -86,9 +87,13 @@ func TestSyncBlocks(t *testing.T) {
 
 		mockSNData := mocks.NewMockStarknetData(mockCtrl)
 
+		var rwMutex sync.RWMutex
 		syncingHeight := uint64(0)
```
omerfirmak, line 91: Cool, using an atomic would be even better https://pkg.go.dev/sync/atomic#Int64

omerfirmak (approved):

sync/sync_test.go
jelilat: Will explore this and see if it can be done in a separate PR.

PR #699, opened 2023-03-30
Add l1 verifier

joshklop:
Key additions and modifications:
- Add the `L1HeadKey` key to the database. It holds the latest `core.L1Head`
- Add `blockchain.SetL1Head`
- The `l1` package contains most of the new logic.
- We can mock the `l1.Subscriber` interface in tests. In practice, the interface is implemented by the `l1.StarknetFilterer`, which was generated using go-ethereum's abigen tool: `abigen --abi starknetAbi.json --pkg starknet`. Since we're only using the filterer, we can remove a lot of the generated code if we want.
  - codecov should ignore the generated `l1/contract/starknet.go` file.  

`node.Run` -> `l1.Client.Run` -> (update is received from subscription) -> `blockchain.SetL1Head`


l1/l1.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+	defer sub.Unsubscribe()
+
+	for {
+		select {
+		case err := <-sub.Err():
+			return err
+		case <-ctx.Done():
+			return nil
+		case update := <-updateChan:
```
omerfirmak, line 59: we should check for update.Raw.Removed

https://github.com/ethereum/go-ethereum/blob/master/core/types/log.go#L53-L55

omerfirmak:
I think we should implement a "confirmation period" like approach and only update the L1 head when we are pretty confident that the log will not be removed due to a reorg.

We will have to subscribe to both logs and new blocks. 

- On new logs; 
 If `Removed` is false, save the log and the height that the log was emitted on, in a FIFO queue.
 If `Removed` is true, find that log in the queue and remove

- On new blocks;
See if the oldest log in FIFO queue is atleast `N` blocks old. 
   If yes, pop the log from queue and update the L1 head.
   If no, do nothing.

l1/l1.go
```diff
@@ -0,0 +1,134 @@
...
+		select {
+		case err := <-subLogStateUpdate.Err():
+			return err
+		case err := <-subHeader.Err():
+			return err
+		case <-ctx.Done():
+			return nil
+		case header := <-headerChan:
+		Outer:
+			for {
```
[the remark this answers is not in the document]
joshklop: I think the loop should terminate (see `break Outer` in the `default` case). I added a test and it passes, but maybe I'm missing something?

I agree the order of messages doesn't matter in practice. However, it does allow us to make the tests deterministic. That's why I ended up nesting the `select`s. 

I think waiting 2 epochs for L1 is maybe a bit much? The overwhelming majority of [past reorgs](https://etherscan.io/blocks_forked?p=1) appear to be single-block reorgs. I couldn't find a non-single block reorg in the last 54 days. What do you think?

node/node.go
```diff
@@ -145,7 +149,24 @@ func (n *Node) Run(ctx context.Context) {
...
+	if err != nil {
+		n.log.Errorw("Error finding core contract address for network", "err", err, "network", n.cfg.Network)
+		return
+	}
+	coreContractFilterer, err := contract.NewStarknetFilterer(*coreContractAddress, ethClient)
+	if err != nil {
+		n.log.Errorw("Failed to set up filterer for L1 contract", "err", err, "network", n.cfg.Network, "ethereumNode", n.cfg.EthNode)
+		return
+	}
+	l1Client := l1.NewClient(l1.NewWatcher(ethClient, coreContractFilterer), n.blockchain, n.log)
```
omerfirmak, line 167: I think we can move all this to l1.go
It might make testing a little harder but we can always introduce a helper to make the dependency injection work.

l1/l1.go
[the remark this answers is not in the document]
omerfirmak: > I agree the order of messages doesn't matter in practice. However, it does allow us to make the tests deterministic. That's why I ended up nesting the selects.

I would avoid doing that just for the sake of easier testing. Could we have updated the test instead?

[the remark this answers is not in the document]
joshklop: > doesnt that just jump to start of the loop again?

It should break out of the `Outer` loop. [This test](https://github.com/NethermindEth/juno/blob/37a5d711173450f0cc3bb1f2e7f348d1217a548a/l1/l1_test.go#L122) wouldn't succeed if it didn't behave as expected.

> Could we have updated the test instead?

I don't think so. I feel the tests should be agnostic to the order the messages are delivered. With the nested selects, we only rely on the assumption: "if we receive logs, then we also receive a header around the same time."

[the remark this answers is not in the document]
omerfirmak: > we only rely on the assumption: "if we receive logs, then we also receive a header around the same time."

Also we rely on events coming before the new header, dont we? Nested select is a bit racey, although since headers are pretty frequent it might not make a difference in the end.

Assume you get a header for height `N`, but events for that height `N` arrive after some time. So nested select will exit thru the default case since the events are not in the channels yet and we will only be processing the events after the next header for `N+1`.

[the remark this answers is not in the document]
omerfirmak: Not really. Other than making the flat-select work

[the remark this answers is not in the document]
joshklop: The flat-select has the same issue, right?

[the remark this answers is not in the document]
omerfirmak: Maybe, I might have not really thought this thru. We dont really need to fix this, it is not an issue at all. It is just an implicit `+1` to the confirmation period.

l1/l1.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+}
+
+func (c *Client) updateBlockchain(l1BlockNumber uint64) error {
+	for blockNumber, update := range c.q {
+		// Wait for at least one block just in case the tip of the L1 chain is reorged.
+		if blockNumber > l1BlockNumber-ConfirmationPeriod {
+			continue
+		}
+		if err := c.l2Chain.SetL1Update(update); err != nil {
```
omerfirmak, line 104: we wont be calling `SetL1Update` with ordered `blockNumber`s now.
`SetL1Update` should make sure the incoming update is more recent.

wojciechos:
In order to make our smoke test work with the added L1 verifier, we need to provide an Ethereum node address. I'm considering whether we should use Infura/Alchemy, run our own node, or ask the core team to provide us with one.

joshklop:
- Fixed some issues raised by @omerfirmak 
- Rebased
- Removed unused code that was generated by `abigen`.
- Moved the generated code to `l1/internal`.

l1/l1.go
omerfirmak: Dont we still have the same issue here?

when we `range c.q`, we will loop over the `blockNumber`s in random order

wojciechos:
I got an error while trying to run Juno with an eth-node via HTTPS(it works for other node implementations). It panicked with this error:

`ERROR node/node.go:205 Service error {"name": "*l1.Client", "err": "subscribe to L1 headers: notifications not supported"}`

When using wss, everything works perfectly. Just wondering, is HTTPS not supported on purpose, or is this something we'll look into?

wojciechos:
I guess we should also update README file:

- Run with Docker part - we should add `--eth-node <YOUR-ETH-NODE>` to the command and maybe some short note like: "You should replace `<YOUR-ETH-NODE>` with your actual Ethereum node address. For example, if you're using Infura, your Ethereum node address might look something like this: wss://mainnet.infura.io/ws/v3/your-infura-project-id."
- Mark L1 checkbox in Phase 2

joshklop:
> I guess we should also update README file

Done, thanks!

joshklop:
A things that can be done in this PR or a follow-up:

- In order to test cases involving multiple block headers, we need to replace the subscriber (at least using the current mocking techniques--we can probably use custom mocks and avoid this issue). The only way to do this without manipulating the public API is by putting the tests in the same package as the code.
- More graceful reorg handling (right now we only update the confirmation queue and assume more logs will come in the future--the local L1 head remains the same).
- Handling failed subscriptions.

.github/workflows/smoke-tests.yml
```diff
@@ -40,4 +40,4 @@ jobs:
         env:
           TARGET_BLOCK: 100
           TIMEOUT: 5m
-          JUNO_PARAMETERS: "--network mainnet"
+          JUNO_PARAMETERS: "--network mainnet --eth-node ${{ secrets.ETH_NODE }}"
```
kirugan, line 43: Do you need setup somewhere `secrets.ETH_NODE` or it's done already?

blockchain/blockchain_test.go
```diff
@@ -398,3 +398,26 @@ func TestState(t *testing.T) {
...
+			StateRoot:   new(felt.Felt).SetUint64(2),
+		},
+		{
+			BlockNumber: 2,
+			StateRoot:   new(felt.Felt).SetUint64(3),
+		},
+	}
+
+	for _, head := range heads {
+		t.Run("update L1 head to block "+string(rune(head.BlockNumber)), func(t *testing.T) {
```
kirugan, line 415: I think you wanted to do smth. like `strconv.Itoa(...)`, currently `string(rune(...))` expression returns A letter and some weird rune LTX for block 1 and 2 respectively

l1/internal/contract/starknet.go
```diff
@@ -0,0 +1,103 @@
+// Code generated - DO NOT EDIT.
```
kirugan, line 1: how was it generated?

utils/network.go
```diff
@@ -97,3 +98,22 @@ func (n Network) ChainID() *felt.Felt {
...
+		address = common.HexToAddress("0xc662c410C0ECf747543f5bA90660f6ABeBD9C8c4")
+	case GOERLI:
+		address = common.HexToAddress("0xde29d060D45901Fb19ED6C6e959EB22d8626708e")
+	case GOERLI2:
+		address = common.HexToAddress("0xa4eD3aD27c294565cB0DCc993BDdCC75432D498c")
+	case INTEGRATION:
+		return common.Address{}, errors.New("l1 contract is not available on the integration network")
+	default:
+		// Should not happen.
+		panic(ErrUnknownNetwork)
```
kirugan, line 116: Since there is an error in return signature wdyt about unification i.e. returning this error rather than panicking. As far as I understand both ways lead to same result (node will stop).

node/node.go
```diff
@@ -177,6 +180,26 @@ func (n *Node) Run(ctx context.Context) {
...
+			n.log.Errorw("Error finding core contract address for network", "err", err, "network", n.cfg.Network)
+			return
+		}
+		var ethSubscriber *l1.EthSubscriber
+		ethSubscriber, err = l1.NewEthSubscriber(n.cfg.EthNode, coreContractAddress)
+		if err != nil {
+			n.log.Errorw("Error creating ethWatcher", "err", err)
+			return
+		}
+		confirmationPeriod := 16
```
kirugan, line 198: `confirmationPeriod` is for number of blocks right? Also since it's hardcoded now maybe make it const (we can either keep it here or move it outside the func

l1/eth_subscriber.go
```diff
@@ -0,0 +1,42 @@
...
+
+type EthSubscriber struct {
+	ethClient *ethclient.Client
+	filterer  *contract.StarknetFilterer
+}
+
+var _ Subscriber = (*EthSubscriber)(nil)
+
+func NewEthSubscriber(ethClientAddress string, coreContractAddress common.Address) (*EthSubscriber, error) {
+	ethClient, err := ethclient.Dial(ethClientAddress)
```
kirugan, line 22: There is a `ethClient.Close()` method which is not used atm. I think we should call it during shutdown, but maybe it's not critical

l1/queue.go
```diff
@@ -0,0 +1,63 @@
...
+func (q *queue) Enqueue(update *contract.StarknetLogStateUpdate) {
+	q.q[update.Raw.BlockNumber] = update
+}
+
+func (q *queue) Remove(l1BlockNumber uint64) {
+	delete(q.q, l1BlockNumber)
+}
+
+func (q *queue) Reorg(forkedBlockNumber uint64) {
+	for l1BlockNumber := range q.q {
```
kirugan, line 34: loop over map in Go is randomized, can it lead to some issues here?

.github/workflows/smoke-tests.yml
joshklop: @wojciechos may know.

wojciechos: It's already done.

l1/internal/contract/starknet.go
joshklop: Geth's [abigen](https://geth.ethereum.org/docs/tools/abigen) utility. 
> Abigen creates easy-to-use, type-safe Go packages from Ethereum smart contract definitions known as ABIs. This abstracts away a lot of the complexity of handling smart contract deployment and interaction in Go native applications such as encoding and decoding smart contracts into EVM bytecode

```
abigen --abi coreContract_0.11.1.json --pkg starknet > starknet.go
```

l1/queue.go
joshklop: I don't think the order matters here since we loop over the entire map.

node/node.go
```diff
@@ -182,6 +186,26 @@ func (n *Node) Run(ctx context.Context) {
...
+			n.log.Errorw("Error finding core contract address for network", "err", err, "network", n.cfg.Network)
+			return
+		}
+		var ethSubscriber *l1.EthSubscriber
+		ethSubscriber, err = l1.NewEthSubscriber(n.cfg.EthNode, coreContractAddress)
+		if err != nil {
+			n.log.Errorw("Error creating ethWatcher", "err", err)
+			return
+		}
+		defer ethSubscriber.Close()
```
omerfirmak, line 204: we are immediately closing the `ethSubscriber`, without l1.Run can be called. How does this work?

node/node.go
```diff
@@ -11,17 +11,20 @@ import (
...
 	"github.com/NethermindEth/juno/sync"
 	"github.com/NethermindEth/juno/utils"
+	"github.com/ethereum/go-ethereum/common"
 	"github.com/sourcegraph/conc"
 )
 
 const (
-	defaultPprofPort = uint16(9080)
+	defaultPprofPort   = uint16(9080)
+	confirmationPeriod = uint64(16)
```
kirugan, line 27: constants are untyped so we don't need to specify type here

kirugan (approved):

l1/queue.go
```diff
@@ -0,0 +1,63 @@
...
+	"github.com/NethermindEth/juno/l1/internal/contract"
+)
+
+// queue is a priority queue that implements a "confirmation window."
+// Given an L1 height, all [contract.LogStateUpdate]s emitted at least
+// confirmationPeriod blocks before the height are "confirmed".
+//
+// Its abnormal API reflects the fact that it is only meant to be used
+// by the L1 verifier.
+type queue struct {
```
IronGauntlets, line 13: This queue doesn't ensure that there are no missing blocks, which can happen if subscriptions fail and new blocks were created during the creation of a new subscription.

l1/l1.go
```diff
@@ -0,0 +1,141 @@
...
+					subUpdates.Unsubscribe()
+
+					subUpdates, logStateUpdateChan, err = c.subscribeToUpdates(subscribeCtx, buffer)
+					if err != nil {
+						return err
+					}
+					defer subUpdates.Unsubscribe() //nolint:gocritic
+				case logStateUpdate := <-logStateUpdateChan:
+					c.log.Debugw("Received L1 LogStateUpdate",
+						"number", logStateUpdate.BlockNumber,
```
IronGauntlets, line 108: logStateUpdate.BlockNumber is this Starknet block number or Ethereum Block number?

l1/l1.go
```diff
@@ -0,0 +1,141 @@
...
+				case err := <-subUpdates.Err():
+					c.log.Warnw("L1 update subscription failed, resubscribing", "error", err)
+					subUpdates.Unsubscribe()
+
+					subUpdates, logStateUpdateChan, err = c.subscribeToUpdates(subscribeCtx, buffer)
+					if err != nil {
+						return err
+					}
+					defer subUpdates.Unsubscribe() //nolint:gocritic
+				case logStateUpdate := <-logStateUpdateChan:
```
IronGauntlets, line 106: How can you be sure that the log emitted here belongs to the `l1height` above? 

If logStateUpdate doesn't contain Ethereum Block Number, how can you be sure that looping here only contains events from a single Ethereum block?

IronGauntlets (changes requested):

node/node.go
```diff
@@ -189,6 +193,26 @@ func (n *Node) Run(ctx context.Context) {
...
+	if n.cfg.EthNode == "" {
+		n.log.Warnw("Ethereum node address not found; will not verify against L1")
+	} else {
+		var coreContractAddress common.Address
+		coreContractAddress, err = n.cfg.Network.CoreContractAddress()
+		if err != nil {
+			n.log.Errorw("Error finding core contract address for network", "err", err, "network", n.cfg.Network)
+			return
+		}
+		var ethSubscriber *l1.EthSubscriber
```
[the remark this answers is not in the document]
IronGauntlets: `Subscription` if rename.

l1/l1.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+	defer subHeaders.Unsubscribe()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
+		case err := <-subHeaders.Err():
+			c.log.Warnw("L1 header subscription failed, resubscribing", "error", err)
+			subHeaders.Unsubscribe()
```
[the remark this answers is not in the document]
joshklop: We unsubscribe to release the subscription resources (most notable is the channel).

node/node.go
```diff
@@ -189,6 +193,26 @@ func (n *Node) Run(ctx context.Context) {
...
+		ethSubscriber, err = l1.NewEthSubscriber(n.cfg.EthNode, coreContractAddress)
+		if err != nil {
+			n.log.Errorw("Error creating ethWatcher", "err", err)
+			return
+		}
+		defer ethSubscriber.Close()
```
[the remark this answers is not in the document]
joshklop: `EthSubscriber` isn't a service, its an `l1.Subscriber`. That's why it doesn't implement the `Service` interface.

l1/l1.go
```diff
@@ -0,0 +1,141 @@
...
+	"github.com/NethermindEth/juno/core/felt"
+	"github.com/NethermindEth/juno/l1/internal/contract"
+	"github.com/NethermindEth/juno/service"
+	"github.com/NethermindEth/juno/utils"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/event"
+)
+
+//go:generate mockgen -destination=./mocks/mock_subscriber.go -package=mocks github.com/NethermindEth/juno/l1 Subscriber
+type Subscriber interface {
```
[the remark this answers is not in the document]
joshklop: A similar interface is called `Subscriber` in go-ethereum, so I think I'll leave it for now.

node/node.go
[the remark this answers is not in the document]
joshklop: We need to call EthSubscriber's `Close()` method.

l1/l1.go
```diff
@@ -0,0 +1,141 @@
...
+			return nil
+		case err := <-subHeaders.Err():
+			c.log.Warnw("L1 header subscription failed, resubscribing", "error", err)
+			subHeaders.Unsubscribe()
+
+			subHeaders, headerChan, err = c.subscribeToHeaders(subscribeCtx, buffer)
+			if err != nil {
+				return err
+			}
+			defer subHeaders.Unsubscribe() //nolint:gocritic
```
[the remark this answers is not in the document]
joshklop: > will refer to the same subHeader

We reassign it a few lines above? If I'm missing something, please let me know.

l1/queue.go
joshklop: Right. You think we should add a comment about that somewhere?

l1/l1.go
joshklop: Starknet

l1/l1.go
```diff
@@ -0,0 +1,141 @@
...
+
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
+		case err := <-subHeaders.Err():
+			c.log.Warnw("L1 header subscription failed, resubscribing", "error", err)
+			subHeaders.Unsubscribe()
+
+			subHeaders, headerChan, err = c.subscribeToHeaders(subscribeCtx, buffer)
```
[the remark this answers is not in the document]
joshklop: I think we can drop lost headers. We should probably pick up lost updates, though. Let me know if you disagree.

l1/l1.go
```diff
@@ -0,0 +1,141 @@
...
+				select {
+				case err := <-subUpdates.Err():
+					c.log.Warnw("L1 update subscription failed, resubscribing", "error", err)
+					subUpdates.Unsubscribe()
+
+					subUpdates, logStateUpdateChan, err = c.subscribeToUpdates(subscribeCtx, buffer)
+					if err != nil {
+						return err
+					}
+					defer subUpdates.Unsubscribe() //nolint:gocritic
```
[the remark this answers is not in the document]
joshklop: Thanks. Conversation continued on other thread.

core/l1_head.go
```diff
@@ -0,0 +1,9 @@
+package core
+
+import "github.com/NethermindEth/juno/core/felt"
+
+type L1Head struct {
+	BlockNumber uint64
+	BlockHash   *felt.Felt
+	StateRoot   *felt.Felt
+}
```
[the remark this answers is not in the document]
IronGauntlets: Yeah, I didn't think about that scenario.

node/node.go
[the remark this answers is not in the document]
IronGauntlets: You are correct, we do the same thing for closing the db. Apologies for the misunderstanding.

l1/l1.go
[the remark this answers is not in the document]
IronGauntlets: Here is an example:

```go
import "fmt"

func main() {
	x := 2
	defer add2(&x)

	x = 0

	// This program will print the following:
	// 2 (not 4, as x was reassigned before it could be evaluated,
        // It will print 4 if add2 accepted a value instead of a reference)
}

func add2(x *int) {
	*x += 2
	fmt.Println(*x)
}
```

Therefore, when the line 68 `Unsubscribe()` method is called it will refer to the subHeader which was most recently assigned, not the initial one. And since you explicitly `Unsubscribe()` if an error occurs. I thought that line 88 was redundant.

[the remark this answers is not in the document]
IronGauntlets: After reading the documentation, `Unsubscribe()` can be called any number of times, so there won't be a panic.

l1/queue.go
IronGauntlets: I realised it doesn't matter if there are missing blocks all we care about is the difference between the block number and the confirmation period.

l1/l1.go
IronGauntlets: The reason I am asking is that I fail to see why we are subscribing to the headers when we can't be sure which ethereum block the log belongs to. 

Using the ethereum block header number to determine whether the log won't be removed requires the assumption that we know which block the log was emitted from. Since we don't have that information making any statement about ethereum block confirmation is superfluous.

I will create separate issue/s for my concerns. For now, we can merge it.

l1/l1.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+	defer subHeaders.Unsubscribe()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
+		case err := <-subHeaders.Err():
+			c.log.Warnw("L1 header subscription failed, resubscribing", "error", err)
+			subHeaders.Unsubscribe()
```
IronGauntlets, line 82: Are you sure that `Unsubscribe()` won't close the `headerChan`?

l1/l1.go
omerfirmak: Each event comes with the block number & hash it is associated with https://github.com/NethermindEth/juno/pull/699/files/9d0eb5e9f39a7ffe787ec242ba588e6b50a2b6d3#diff-7f7fd9a26e7cb24ff29bbaa5130bd597f6ec2c15bec9414dfec4863850ca8570R108-R110

l1/queue.go
```diff
@@ -0,0 +1,63 @@
...
+		if l1BlockNumber >= forkedBlockNumber {
+			q.Remove(l1BlockNumber)
+		}
+	}
+}
+
+// MaxConfirmed returns the most recent confirmed update and deletes
+// all other confirmed updates.
+func (q *queue) MaxConfirmed(l1Height uint64) *contract.StarknetLogStateUpdate {
+	maxConfirmed := new(contract.StarknetLogStateUpdate)
```
omerfirmak, line 44:
```suggestion
	var maxConfirmed *contract.StarknetLogStateUpdate
```

lets you remove the `if maxConfirmed.BlockNumber == nil` check below

omerfirmak (approved):
LGTM

l1/l1.go
joshklop: The documentation says `Unsubscribe()` "cancels sending of events, closing the error channel". It should not close the channels it did not create.

Reading the implementation, I can confirm `Unsubscribe()` doesn't close the `headerChan`.

l1/queue.go
joshklop: Thanks! I'll go with something like this.

This particular change causes a panic in the tests since `maxConfirmed.Raw` is `nil` and we use it a few lines down. Adding a separate `maxConfirmedBlockNumber` var makes it work.

l1/queue.go
```diff
@@ -0,0 +1,59 @@
...
+// MaxConfirmed returns the most recent confirmed update and deletes
+// all other confirmed updates.
+func (q *queue) MaxConfirmed(l1Height uint64) *contract.StarknetLogStateUpdate {
+	var maxConfirmed *contract.StarknetLogStateUpdate
+	var maxConfirmedBlockNumber uint64
+	for l1BlockNumber, possibleHead := range q.q {
+		if l1Height-l1BlockNumber >= q.confirmationPeriod {
+			// We are careful to use >= here. If the max block is zero,
+			// this will still work. Mostly helpful for tests.
+			if l1BlockNumber >= maxConfirmedBlockNumber {
```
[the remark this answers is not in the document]
joshklop: Much better. Thanks again!

IronGauntlets (approved):

PR #703, opened 2023-03-31
Position iterator on Next() if neccessary

### LINKED ISSUE #702: `db.Iterator` is always invalid if `Seek` is never called
`NewIterator` returns an invalid Iterator.

```
// NewIter returns an iterator that is unpositioned (Iterator.Valid() will
// return false). The iterator can be positioned via a call to SeekGE,
// SeekPrefixGE, SeekLT, First or Last. Only indexed batches support iterators.
```

omerfirmak:
Fixes #702

### DIFF
--- a/db/pebble/db_test.go
+++ b/db/pebble/db_test.go
@@ -350,3 +350,44 @@ func TestPrefixSearch(t *testing.T) {
 		return nil
 	}))
 }
+
+func TestNext(t *testing.T) {
+	testDB := pebble.NewMemTest()
+	t.Cleanup(func() {
+		require.NoError(t, testDB.Close())
+	})
+
+	txn := testDB.NewTransaction(true)
+	t.Cleanup(func() {
+		require.NoError(t, txn.Discard())
+	})
+	require.NoError(t, txn.Set([]byte{0}, []byte{0}))
+	require.NoError(t, txn.Set([]byte{1}, []byte{1}))
+	require.NoError(t, txn.Set([]byte{2}, []byte{2}))
+
+	t.Run("Next() on new iterator", func(t *testing.T) {
+		it, err := txn.NewIterator()
+		require.NoError(t, err)
+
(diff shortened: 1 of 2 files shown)


jelilat (approved):

IronGauntlets:
I don't think I understand the problem that is being solved here. Are we saying that we want to be able to `Next()` on a new iterator without calling seek?

omerfirmak:
> I don't think I understand the problem that is being solved here. Are we saying that we want to be able to call `Next()` on a new iterator without calling `Seek()`?

Yes

db/pebble/db_test.go
```diff
@@ -350,3 +350,44 @@ func TestPrefixSearch(t *testing.T) {
...
+		})
+
+		t.Run("Next() should validate iterator", func(t *testing.T) {
+			assert.True(t, it.Next())
+		})
+
+		require.NoError(t, it.Close())
+	})
+
+	t.Run("first Next() should work as expected after a Seek()", func(t *testing.T) {
```
IronGauntlets, line 383:
```suggestion
	t.Run("Next() should work as expected after a Seek()", func(t *testing.T) {
```

db/pebble/db_test.go
```diff
@@ -350,3 +350,44 @@ func TestPrefixSearch(t *testing.T) {
...
+		it, err := txn.NewIterator()
+		require.NoError(t, err)
+
+		require.True(t, it.Seek([]byte{0}))
+		require.True(t, it.Next())
+		require.Equal(t, []byte{1}, it.Key())
```
IronGauntlets, line 389: small nitpick: should we also check the value?

omerfirmak: not really, that would be testing the underlying db implementation. Not the change here.

db/pebble/db_test.go
omerfirmak: done

IronGauntlets (approved):

PR #706, opened 2023-03-31
Implement starknet_syncing

jelilat:
Closes https://github.com/NethermindEth/juno/issues/677


starknetdata/starknetdata.go
```diff
@@ -11,7 +11,7 @@ import (
 //
 //go:generate mockgen -destination=../mocks/mock_starknetdata.go -package=mocks github.com/NethermindEth/juno/starknetdata StarknetData
 type StarknetData interface {
-	BlockByNumber(ctx context.Context, blockNumber uint64) (*core.Block, error)
+	BlockByID(ctx context.Context, blockID string) (*core.Block, error)
```
omerfirmak, line 14: We lose type safety here for no real reason. Let's keep it the old way and introduce a `Head` or a similar endpoint

blockchain/blockchain.go
```diff
@@ -57,15 +59,17 @@ var _ Reader = (*Blockchain)(nil)
 
 // Blockchain is responsible for keeping track of all things related to the Starknet blockchain
 type Blockchain struct {
-	network  utils.Network
-	database db.DB
+	database    db.DB
+	memDatabase db.DB
```
omerfirmak, line 63: why do we need a DB here? If all we are storing is temporary stuff we should use in-memory variables instead of a in-memory DB.

omerfirmak (changes requested):

PR #708, opened 2023-03-31
Implement starknet_syncing

jelilat:
Closes https://github.com/NethermindEth/juno/issues/677


jelilat:
@omerfirmak made the changes to this PR

rpc/handlers_test.go
```diff
@@ -135,23 +136,23 @@ func TestBlockTransactionCount(t *testing.T) {
 		assert.Equal(t, rpc.ErrBlockNotFound, rpcErr)
 	})
 
-	t.Run("blockId - latest", func(t *testing.T) {
+	t.Run("blockID - latest", func(t *testing.T) {
```
IronGauntlets, line 139: These changes should go in the previous commit

jelilat: Done

blockchain/blockchain.go
```diff
@@ -15,6 +15,11 @@ import (
 
 const lenOfByteSlice = 8
 
+var (
+	startingBlockNumber *uint64
```
omerfirmak, line 19: We can put all this in `Synchronizer`. I don't think we need `blockchain` to be aware of all this, in fact, I think it shouldn't be.

rpc/handlers_test.go
```diff
@@ -889,3 +890,69 @@ func TestStateUpdate(t *testing.T) {
...
+	t.Run("syncing", func(t *testing.T) {
+		mockReader.EXPECT().StartingBlockNumber().Return(&startingBlock)
+		mockReader.EXPECT().BlockHeaderByNumber(startingBlock).Return(&core.Header{Hash: &felt.Zero}, nil)
+		mockReader.EXPECT().HeadsHeader().Return(&core.Header{Number: 1, Hash: new(felt.Felt).SetUint64(1)}, nil)
+		mockReader.EXPECT().HighestBlockHeader().Return(&core.Header{Number: 2, Hash: new(felt.Felt).SetUint64(2)})
+
+		expectedSyncing := &rpc.SyncState{
+			Status: &rpc.SyncStatus{
+				StartingBlockHash:   &felt.Zero,
+				StartingBlockNumber: "0x" + strconv.FormatUint(startingBlock, 16),
```
omerfirmak, line 947: I think this is a common enough pattern that we can define a `HexUint64` type and implement a custom MarshalJSON function for it.

blockchain/blockchain.go
jelilat: Done

rpc/handlers_test.go
jelilat: I probably don't fully understand your comment. But not really sure we need a custom `HexUint64` type. I have created a `Uint64ToHexStr` function [here](https://github.com/NethermindEth/juno/pull/708/files#diff-df2395a5aa8fcd186ed6c403ff5b4c573d69350914e6bca767cd9b559cd41afeR5)

sync/sync.go
```diff
@@ -122,7 +131,9 @@ func (s *Synchronizer) syncBlocks(syncCtx context.Context) {
 	verifiers := stream.New().WithMaxGoroutines(runtime.NumCPU())
 
 	streamCtx, streamCancel := context.WithCancel(syncCtx)
+
 	nextHeight := s.nextHeight()
+	StartingBlockNumber = &nextHeight
```
omerfirmak, line 136: `nextHeight` constantly changes. So `StartingBlockNumber` will change as well, since it is just a pointer to `nextHeight` . Is stil intentional?

starknetdata/starknetdata.go
```diff
@@ -11,7 +11,8 @@ import (
 //
 //go:generate mockgen -destination=../mocks/mock_starknetdata.go -package=mocks github.com/NethermindEth/juno/starknetdata StarknetData
 type StarknetData interface {
-	BlockByNumber(ctx context.Context, blockNumber uint64) (*core.Block, error)
+	BlockByNumber(ctx context.Context, blockID uint64) (*core.Block, error)
```
omerfirmak, line 14: should be `blockNumber uint64`

starknetdata/starknetdata.go
```diff
@@ -11,7 +11,8 @@ import (
...
+	BlockByID(ctx context.Context, blockID string) (*core.Block, error)
```
omerfirmak, line 15: Instead of a `string`, we should use a more well defined type here to avoid any confusion on what the valid inputs are. Let's either make this `BlockByTag` and define a `BlockTag` enum with Pending and Latest values

or let's just jave `BlockLatest` that does not need any arguments

sync/sync.go
```diff
@@ -15,6 +15,11 @@ import (
 
 var _ service.Service = (*Synchronizer)(nil)
 
+var (
+	StartingBlockNumber *uint64
+	HighestBlockHeader  *core.Header
+)
+
```
omerfirmak, line 22: Unless there is a real good reason, let's avoid package level variables. We can move these to `Synchronizer` struct.

sync/sync.go
```diff
@@ -53,6 +58,10 @@ func (s *Synchronizer) fetcherTask(ctx context.Context, height uint64, verifiers
 			if err != nil {
 				continue
 			}
+			highestBlock, err := s.StarknetData.BlockByID(ctx, "latest")
```
omerfirmak, line 61: This is a bit concerning. We would be fetching two blocks every for block. This could slow down the sync, and create a lot more network traffic that users would be paying for.

sync/sync.go
jelilat: No, it isn't intentional. I've fixed it now.

rpc/handlers_test.go
omerfirmak: Now that I think about it, printing block numbers in base 10 makes more sense to me, WDYT?

sync/sync.go
jelilat: How do you suggest we handle latest block? Probably only poll it at intervals? Or is there a way to listen to new blocks on Starknet?

omerfirmak: > Or is there a way to listen to new blocks on Starknet?

I dont think there is.

How about this;
Let's query the latest block when we are starting to sync and only query it again after we reach that initial latest block.
So after storing a block; if `storedBlockHeight >= highestBlock` is true, query new highest block.

jelilat: Thanks. Will do this then

rpc/handlers_test.go
omerfirmak: my bad then, thanks!

sync/sync.go
jelilat: Done

starknetdata/starknetdata.go
jelilat: Done

starknetdata/starknetdata.go
jelilat: Done

sync/sync.go
jelilat: Done

rpc/sync.go
```diff
@@ -0,0 +1,18 @@
+package rpc
+
+import "github.com/NethermindEth/juno/core/felt"
+
+type SyncState struct {
+	False  *bool       `json:"false,omitempty"`
```
omerfirmak, line 6: Any particular reason this a pointer?

jelilat: We want the field to be omitted from the JSON representation if its value is nil (i.e. when the value is not false). If False was defined as a regular bool field instead of a pointer to a bool, it would be included in the JSON representation with a value of `false` even if it was not set or explicitly set to `false`. 

That was what I discovered when I tried it with a regular `bool`. Since we don't want that to happen according to the [spec](https://github.com/starkware-libs/starknet-specs/blob/a789ccc3432c57777beceaa53a34a7ae2f25fda0/api/starknet_api_openrpc.json#L878), I had to change it.

rpc/sync.go
```diff
@@ -0,0 +1,18 @@
...
+	Status *SyncStatus `json:"sync_status,omitempty"`
+}
+
+// https://github.com/starkware-libs/starknet-specs/blob/a789ccc3432c57777beceaa53a34a7ae2f25fda0/api/starknet_api_openrpc.json#L852
```
omerfirmak, line 10: what I meant in one of my previous comments was something like this;

```golang
type NumAsHex uint64

func (n NumAsHex) MarshalJSON() ([]byte, error) {
   return []byte(fmt.Sprintf("0x%x", value))
}

type SyncStatus struct {
	StartingBlockHash   *felt.Felt `json:"starting_block_hash"`
	StartingBlockNumber NumAsHex     `json:"starting_block_number"`
	CurrentBlockHash    *felt.Felt `json:"current_block_hash"`
	CurrentBlockNumber  NumAsHex     `json:"current_block_number"`
	HighestBlockHash    *felt.Felt `json:"highest_block_hash"`
	HighestBlockNumber  NumAsHex     `json:"highest_block_number"`
}
```

and you dont need the `utils.Uint64ToHexStr`

NumAsHex is used a lot throughout the RPC spec, so we could use it in other endpoints as well.

sync/sync.go
```diff
@@ -122,7 +133,10 @@ func (s *Synchronizer) syncBlocks(syncCtx context.Context) {
 	verifiers := stream.New().WithMaxGoroutines(runtime.NumCPU())
 
 	streamCtx, streamCancel := context.WithCancel(syncCtx)
+
 	nextHeight := s.nextHeight()
+	startingHeight := nextHeight
+	s.StartingBlockNumber = &startingHeight
```
omerfirmak, line 139: any chance we do this with out the `startingHeight`?
also, dont we need to set `s.StartingBlockNumber` to nil when exiting `syncBlocks`?

sync/sync.go
```diff
@@ -17,17 +17,21 @@ var _ service.Service = (*Synchronizer)(nil)
...
 
 func New(bc *blockchain.Blockchain, starkNetData starknetdata.StarknetData, log utils.SimpleLogger) *Synchronizer {
 	return &Synchronizer{
-		Blockchain:   bc,
-		StarknetData: starkNetData,
-		log:          log,
+		Blockchain:          bc,
+		StarknetData:        starkNetData,
+		StartingBlockNumber: nil,
+		HighestBlockHeader:  nil,
```
omerfirmak, line 33: I dont think you need these, they are nil by default anyways.

sync/sync.go
jelilat: Got it now. Thanks

sync/sync.go
jelilat: Done

starknetdata/feeder/feeder.go
```diff
@@ -27,7 +28,18 @@ func New(client *feeder.Client) *Feeder {
...
+		return nil, err
+	}
+
+	return adaptBlock(response)
+}
+
+// BlockLatest gets the latest block from the feeder,
+// then adapts it to the core.Block type.
+func (f *Feeder) BlockLatest(ctx context.Context) (*core.Block, error) {
+	response, err := f.client.Block(ctx, "latest")
```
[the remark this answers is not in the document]
jelilat: Done

sync/sync.go
jelilat: Done

omerfirmak: `s.HighestBlockHeader` needs to be cleared as well. lets do both of those in a defer statement at the beginning of the `syncBlocks`

jelilat: updated it

rpc/handlers.go
```diff
@@ -409,3 +412,41 @@ func (h *Handler) StateUpdate(id *BlockID) (*StateUpdate, *jsonrpc.Error) {
...
+		return defaultSyncState, nil
+	}
+	startingBlockHeader, err := h.bcReader.BlockHeaderByNumber(*startingBlockNumber)
+	if err != nil {
+		return defaultSyncState, nil
+	}
+	head, err := h.bcReader.HeadsHeader()
+	if err != nil {
+		return defaultSyncState, nil
+	}
```
omerfirmak, line 434: should these be returning an error instead of defaultSyncState?

omerfirmak (approved):
Feel free to merge after updating the README

rpc/handlers.go
jelilat: No. The spec doesn't specify an error. It should return false if it's not synching

Issue #711, opened 2023-04-04
Block hash verification error is not prioritised


jelilat:
The current implementation of VerifyBlock in [`core/block.go`](https://github.com/NethermindEth/juno/blob/main/core/block.go#L104-L106) returns an error whenever transaction hash verification fails, leading to the possibility of ignoring block verification errors. This can cause invalid blocks to be added to the chain despite the possibility of block verification errors. 

We should modify the `VerifyBlockHash` function to prioritise block verification errors over transaction hash errors.

jelilat:
Closing due to [this](https://github.com/NethermindEth/juno/pull/712#issuecomment-1495892103)

PR #712, opened 2023-04-04
Prioritise block hash verification error

jelilat:
Fixes https://github.com/NethermindEth/juno/issues/711

### DIFF
--- a/core/block.go
+++ b/core/block.go
@@ -85,9 +85,8 @@ func networkBlockHashMetaInfo(network utils.Network) *blockHashMetaInfo {
 	}
 }
 
-// VerifyBlockHash verifies the block hash. Due to bugs in Starknet alpha, not all blocks have
-// verifiable hashes.
-func VerifyBlockHash(b *Block, network utils.Network) error {
+// VerifyBlock checks that the given block is valid.
+func VerifyBlock(b *Block, network utils.Network) error {
 	if len(b.Transactions) != len(b.Receipts) {
 		return fmt.Errorf("len of transactions: %v do not match len of receipts: %v",
 			len(b.Transactions), len(b.Receipts))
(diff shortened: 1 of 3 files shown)


core/block.go
```diff
@@ -101,10 +100,20 @@ func VerifyBlockHash(b *Block, network utils.Network) error {
...
 
+	if err := verifyBlockHash(b, network); err != nil {
+		return err
+	}
+
 	if err := verifyTransactions(b.Transactions, network); err != nil {
 		return err
 	}
 
+	return nil
```
omerfirmak, line 111:
```suggestion
	return verifyTransactions(b.Transactions, network)
```

omerfirmak (approved):
Nice catch!

core/block.go
jelilat: Thanks for this. Fixed it

IronGauntlets:
Checking transaction hashes before block hash was intentional because if there is an error in transaction hash calculation then the block's hash is invalid. Block hash calculation requires creating transaction commitment which requires transaction hashes.

By checking the transaction hash first and if there is an error we don't need to create a transaction commitment since we know the block hash must be incorrect.

I think what we had before was correct assuming we don't ignore transaction verification errors. 

Now if the caller of VerifyBlockHash chooses to ignore transaction verification errors there is not much we can do. We should not change this piece of code because of what the caller has decided to do.

jelilat:
> Checking transaction hashes before block hash was intentional because if there is an error in transaction hash calculation then the block's hash is invalid. Block hash calculation requires creating transaction commitment which requires transaction hashes.
> 
> By checking the transaction hash first and if there is an error we don't need to create a transaction commitment since we know the block hash must be incorrect.
> 
> I think what we had before was correct assuming we don't ignore transaction verification errors.
> 
> Now if the caller of VerifyBlockHash chooses to ignore transaction verification errors there is not much we can do. We should not change this piece of code because of what the caller has decided to do.

Got you. But in this case, it wasn't the caller who decided that they didn't want to verify blocks. Rather, we are the ones who isn't verifying the block.

Is it safe to say that we aren't actually verifying blocks then?

jelilat:
> Checking transaction hashes before block hash was intentional because if there is an error in transaction hash calculation then the block's hash is invalid. Block hash calculation requires creating transaction commitment which requires transaction hashes.
> 
> By checking the transaction hash first and if there is an error we don't need to create a transaction commitment since we know the block hash must be incorrect.
> 
> I think what we had before was correct assuming we don't ignore transaction verification errors.
> 
> Now if the caller of VerifyBlockHash chooses to ignore transaction verification errors there is not much we can do. We should not change this piece of code because of what the caller has decided to do.

Got you

IronGauntlets:
> Rather, we are the ones who isn't verifying the block.

Yes, but we need to keep in mind the separation of concerns. We need to write a function in a manner which makes sense as to what it is trying to achieve.

> Is it safe to say that we aren't actually verifying blocks then?

This is not true, we are verifying the blocks but our sync package has chosen to make progress despite some errors. Not verifying them would be to not even acknowledge that there is a failure.

Also, after regenesis when we don't ignore errors we will have to revert this change.

PR #646, opened 2023-03-20
 Reflect v0.11.0 changes in StateUpdate

omerfirmak:

PR #713, opened 2023-04-04
Remove separate buckets for trie roots in state

omerfirmak:
as suggested [here](https://github.com/NethermindEth/juno/pull/646#discussion_r1156675877)

I forgot to force push it to that PR

### DIFF
--- a/core/state.go
+++ b/core/state.go
@@ -94,18 +94,19 @@ func (s *State) Root() (*felt.Felt, error) {
 
 // storage returns a [core.Trie] that represents the Starknet global state in the given Txn context.
 func (s *State) storage() (*trie.Trie, func() error, error) {
-	return s.globalTrie(db.StateTrie, db.StateRootKey, trie.NewTriePedersen)
+	return s.globalTrie(db.StateTrie, trie.NewTriePedersen)
 }
 
 func (s *State) classesTrie() (*trie.Trie, func() error, error) {
-	return s.globalTrie(db.ClassesTrie, db.ClassesRootKey, trie.NewTriePoseidon)
+	return s.globalTrie(db.ClassesTrie, trie.NewTriePoseidon)
 }
 
-func (s *State) globalTrie(bucket, rootKeyBucket db.Bucket, newTrie trie.NewTrieFunc) (*trie.Trie, func() error, error) {
-	tTxn := NewTransactionStorage(s.txn, bucket.Key())
+func (s *State) globalTrie(bucket db.Bucket, newTrie trie.NewTrieFunc) (*trie.Trie, func() error, error) {
+	dbPrefix := bucket.Key()
+	tTxn := NewTransactionStorage(s.txn, dbPrefix)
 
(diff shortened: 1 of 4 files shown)


core/state.go
```diff
@@ -94,18 +94,19 @@ func (s *State) Root() (*felt.Felt, error) {
...
-	tTxn := NewTransactionStorage(s.txn, bucket.Key())
+func (s *State) globalTrie(bucket db.Bucket, newTrie trie.NewTrieFunc) (*trie.Trie, func() error, error) {
+	dbPrefix := bucket.Key()
+	tTxn := NewTransactionStorage(s.txn, dbPrefix)
 
 	// fetch root key
-	rootKeyDBKey := rootKeyBucket.Key()
+	rootKeyDBKey := dbPrefix
 	var rootKey *bitset.BitSet
 	err := s.txn.Get(rootKeyDBKey, func(val []byte) error {
```
jelilat, line 111: We probably don't need to reassign this again
```suggestion
	var rootKey *bitset.BitSet
	err := s.txn.Get(dbPrefix, func(val []byte) error {
```

jelilat (changes requested):

core/state.go
```diff
@@ -94,18 +94,19 @@ func (s *State) Root() (*felt.Felt, error) {
...
 }
 
 func (s *State) classesTrie() (*trie.Trie, func() error, error) {
-	return s.globalTrie(db.ClassesTrie, db.ClassesRootKey, trie.NewTriePoseidon)
+	return s.globalTrie(db.ClassesTrie, trie.NewTriePoseidon)
 }
 
-func (s *State) globalTrie(bucket, rootKeyBucket db.Bucket, newTrie trie.NewTrieFunc) (*trie.Trie, func() error, error) {
-	tTxn := NewTransactionStorage(s.txn, bucket.Key())
+func (s *State) globalTrie(bucket db.Bucket, newTrie trie.NewTrieFunc) (*trie.Trie, func() error, error) {
```
jelilat, line 104: We could as well just pass in the key also instead of the whole bucket
```suggestion
func (s *State) globalTrie(key []byte, newTrie trie.NewTrieFunc) (*trie.Trie, func() error, error) {
```

core/state.go
omerfirmak: I think it is better to explicitly emphasize that `rootKeyDBKey` is `dbPrefix` because it is a special case

core/state.go
omerfirmak: We shouldn't allow any arbitrary key. If you need a new global Trie, you shouldnt be able to do that without creating a new bucket for it. Else you might corrupt the DB by accidentally using an existing bucket, that is why we maintain the bucket list

jelilat: Got it. Thanks

jelilat (commented):
pprof test seem to be failing

jelilat (approved):

IronGauntlets (approved):

Issue #717, opened 2023-04-04
Switch RWMutex to atomic in sync_test


jelilat:
See [here](https://github.com/NethermindEth/juno/pull/695#discussion_r1152164035)

PR #718, opened 2023-04-04
Switch to atomic

jelilat:
Closes https://github.com/NethermindEth/juno/issues/717

### DIFF
--- a/sync/sync_test.go
+++ b/sync/sync_test.go
@@ -3,7 +3,7 @@ package sync
 import (
 	"context"
 	"errors"
-	"sync"
+	"sync/atomic"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


IronGauntlets:
Did you run `make test-race`?

IronGauntlets (approved):
I ran `make test-race` and it didn't report any errors.

jelilat:
> Did you run `make test-race`?

yes

IronGauntlets:
@jelilat Please update the commit message. Give more context `Switch to atomic` is not sufficient

IronGauntlets:
> This commit commit uses `atomic` instead of RWMutex to prevent race conditions in `sync_test` since it's more concise and simpler`

`commit` is mentioned twice. 

Also, the Commit subject line should be limited to 50 characters and the body should wrap at 72 characters as mentioned in the `The seven rules of a great Git commit message` section of [this](https://cbea.ms/git-commit/) article.

PR #721, opened 2023-04-06
Only fetch unknown classes in sync process

### LINKED ISSUE #689: Stop unconditionally fetching all classes referenced in a StateUpdate
https://github.com/NethermindEth/juno/blob/e54fe06ba39d84a1f932a937366346a7b1b7c70e/sync/sync.go#L58-L73

We should somehow only fetch the ones we are missing.

omerfirmak:
Fixes #689

### DIFF
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -228,12 +228,12 @@ func (b *Blockchain) Receipt(hash *felt.Felt) (*core.TransactionReceipt, *felt.F
 }
 
 // Store takes a block and state update and performs sanity checks before putting in the database.
-func (b *Blockchain) Store(block *core.Block, stateUpdate *core.StateUpdate, declaredClasses map[felt.Felt]core.Class) error {
+func (b *Blockchain) Store(block *core.Block, stateUpdate *core.StateUpdate, newClasses map[felt.Felt]core.Class) error {
 	return b.database.Update(func(txn db.Transaction) error {
 		if err := b.verifyBlock(txn, block); err != nil {
 			return err
 		}
-		if err := core.NewState(txn).Update(stateUpdate, declaredClasses); err != nil {
+		if err := core.NewState(txn).Update(stateUpdate, newClasses); err != nil {
 			return err
 		}
 		if err := storeBlockHeader(txn, block.Header); err != nil {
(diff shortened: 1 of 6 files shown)


core/state.go
```diff
@@ -246,6 +246,20 @@ func (s *State) putClass(classHash *felt.Felt, class Class) error {
 	return err
 }
 
+// Class returns the class object corresponding to the given classHash
+func (s *State) Class(classHash *felt.Felt) (Class, error) {
```
jelilat, line 250: We should probably add test for this

jelilat (commented):
Looks good to me

core/state.go
omerfirmak: done, with a todo

jelilat (approved):

blockchain/blockchain.go
IronGauntlets, line 1: I don't think it is a good idea for blockchain to do the fetching.

sync/sync.go
```diff
@@ -54,36 +54,23 @@ func (s *Synchronizer) fetcherTask(ctx context.Context, height uint64, verifiers
...
-				referencedClasses[*classHash] = nil
-			}
-			for classHash := range referencedClasses {
-				class, err := s.StarknetData.Class(ctx, &classHash)
-				if err != nil {
-					continue
-				}
-				referencedClasses[classHash] = class
-			}
-
```
IronGauntlets, line 73: By moving this to the blockchain we are not conflating what `fetcherTask` and `verifierTask` are supposed to do.

blockchain/blockchain.go
```diff
@@ -456,23 +456,32 @@ func stateUpdateByHash(txn db.Transaction, hash *felt.Felt) (*core.StateUpdate,
 	})
 }
 
+type OnUnknownClass func(classHash *felt.Felt) (core.Class, error)
```
IronGauntlets, line 459: Although this function type allows us not to import the feeder package however I don't think it is required. I think we can still do the fetching in the sync package.

sync/sync.go
omerfirmak: I know it is not ideal, feel free to suggest an alternative

omerfirmak: This made me realize that we have cases where the verification process halts without no real reason.

We start fetchers and in their callback functions they start the verifiers. Since fetcher callbacks are executed in order, verification tasks are started in order as well. This means that without block N being downloaded, we cannot start verifying block N+1. Which is no ideal, verification should be able to start right away for each individual block after it is downloaded.

omerfirmak: On the other hand, this behavior might make sense because this avoids using up the verifiers tasks for later blocks while the block `HEAD+1` is waiting for a verifier to be available

blockchain/blockchain.go
omerfirmak: Blockchain is not doing the fetching. It is only signalling the caller that it is missing the class for a given hash,  to continue with the verification.

jelilat: I think it is actually, [here](https://github.com/NethermindEth/juno/pull/721/files#diff-37b6e7ed4f299b87b357b54376123393b34ee0699fc035f25bc0621627dcc6baR512)

omerfirmak: what I meant was, the actual network operation is abstract away from `blockchain`

sync/sync.go
jelilat: A suggestion would be, instead of fetching unknown classes, we can fetch unknown class hashes, then fetch the classes in the sync package. Might just be a tad bit slower.

omerfirmak:
I dont want this to delay the release, closing for now.

sync/sync.go
omerfirmak: Now the question is how does `blockchain` know that `sync` successfully fetched all the missing classes.

PR #723, opened 2023-04-07
 Implement starknet_getClass endpoint

omerfirmak:


omerfirmak:
Rebased, depends on #738

rpc/handlers.go
```diff
@@ -450,3 +452,106 @@ func (h *Handler) Syncing() (*SyncState, *jsonrpc.Error) {
...
+
+// https://github.com/starkware-libs/starknet-specs/blob/master/api/starknet_api_openrpc.json#L248
+func (h *Handler) Class(id *BlockID, classHash *felt.Felt) (*Class, *jsonrpc.Error) {
+	state, closer, err := h.stateByID(id)
+	if err != nil {
+		return nil, ErrInternalError
+	}
+	declared, storageErr := state.Class(classHash)
+	if err = closer(); err != nil {
+		return nil, ErrInternalError
```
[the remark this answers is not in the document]
omerfirmak: fixed

omerfirmak:
rebased

core/state_test.go
```diff
@@ -316,3 +316,39 @@ func TestContractIsDeployedAt(t *testing.T) {
...
+	state := core.NewState(txn)
+	su0, err := gw.StateUpdate(context.Background(), 0)
+	require.NoError(t, err)
+	require.NoError(t, state.Update(0, su0, map[felt.Felt]core.Class{
+		*cairo0Hash: cairo0Class,
+		*cairo1Hash: cairo1Class,
+	}))
+
+	gotCairo1Class, err := state.Class(cairo1Hash)
+	require.NoError(t, err)
```
jelilat, line 347: Getting the following error locally

```
Error:      	Received unexpected error:
        	            	cbor: cannot unmarshal map into Go struct field core.DeclaredClass.Class of type core.Class
        	Test:       	TestClass
```

rpc/handlers.go
```diff
@@ -479,11 +481,101 @@ func (h *Handler) Nonce(id *BlockID, address *felt.Felt) (*felt.Felt, *jsonrpc.E
...
+		for _, entryPoint := range c.EntryPoints.External {
+			index := entryPoint.Index
+			rpcClass.EntryPoints.External = append(rpcClass.EntryPoints.External, EntryPoint{
+				Index:    &index,
+				Selector: entryPoint.Selector,
+			})
+		}
+
+	default:
+		return nil, ErrInternalError
```
jelilat, line 577: Not sure we should have an internalError since the spec doesn't define it. We should probably just return `ErrClassHashNotFound` in this case. Maybe log errors so we can track these cases?
```suggestion
		return nil, ErrClassHashNotFound
```

rpc/handlers.go
```diff
@@ -479,11 +481,101 @@ func (h *Handler) Nonce(id *BlockID, address *felt.Felt) (*felt.Felt, *jsonrpc.E
...
+		}
+
+		rpcClass.EntryPoints.External = make([]EntryPoint, 0, len(c.EntryPoints.External))
+		for _, entryPoint := range c.EntryPoints.External {
+			index := entryPoint.Index
+			rpcClass.EntryPoints.External = append(rpcClass.EntryPoints.External, EntryPoint{
+				Index:    &index,
+				Selector: entryPoint.Selector,
+			})
+		}
```
jelilat, line 574: We'll need this for `getClassAt` as well so we should probably make it a function

jelilat (changes requested):

rpc/handlers.go
```diff
@@ -479,11 +481,101 @@ func (h *Handler) Nonce(id *BlockID, address *felt.Felt) (*felt.Felt, *jsonrpc.E
...
+		h.log.Errorw("Error closing state reader in getNonce", "err", closerErr)
 	}
 	if err != nil {
 		return nil, ErrContractNotFound
 	}
 
 	return nonce, nil
 }
+
+// https://github.com/starkware-libs/starknet-specs/blob/master/api/starknet_api_openrpc.json#L248
```
jelilat, line 493:
```suggestion
// Class gets the contract class definition in the given block associated with the given hash
//
// It follows the specification defined here:
// https://github.com/starkware-libs/starknet-specs/blob/master/api/starknet_api_openrpc.json#L248
```

rpc/handlers.go
omerfirmak: I just implemented it, and it is so trivial that I think it doesnt even need testing:)

core/state_test.go
omerfirmak: should be fixed now

node/node.go
```diff
@@ -130,6 +130,16 @@ func makeHTTP(port uint16, rpcHandler *rpc.Handler, log utils.SimpleLogger) *jso
...
 		},
+		{
+			Name:    "starknet_getClass",
+			Params:  []jsonrpc.Parameter{{Name: "block_id"}, {Name: "class_hash"}},
+			Handler: rpcHandler.Class,
+		},
+		{
+			Name:    "starknet_getClassAt",
+			Params:  []jsonrpc.Parameter{{Name: "block_id"}, {Name: "contract_address"}},
+			Handler: rpcHandler.ClassHashAt,
```
jelilat, line 141:
```suggestion
			Handler: rpcHandler.ClassAt,
```

rpc/handlers.go
```diff
@@ -529,3 +530,108 @@ func (h *Handler) ClassHashAt(id *BlockID, address *felt.Felt) (*felt.Felt, *jso
...
+// ClassAt gets the contract class definition in the given block associated instantiated by the given contract address
+//
+// It follows the specification defined here:
+// https://github.com/starkware-libs/starknet-specs/blob/master/api/starknet_api_openrpc.json#L329
+func (h *Handler) ClassAt(id *BlockID, address *felt.Felt) (*Class, *jsonrpc.Error) {
+	classHash, err := h.ClassHashAt(id, address)
+	if err != nil {
+		return nil, err
+	}
+	return h.ClassAt(id, classHash)
```
jelilat, line 636:
```suggestion
	return h.Class(id, classHash)
```

rpc/handlers.go
jelilat: Seen and kinda caught something. We probably should always write tests no matter how trivial. Could help us catch something important.

jelilat (changes requested):

rpc/handlers.go
```diff
@@ -529,3 +530,108 @@ func (h *Handler) ClassHashAt(id *BlockID, address *felt.Felt) (*felt.Felt, *jso
...
+		}
+
+	default:
+		return nil, ErrClassHashNotFound
+	}
+
+	return rpcClass, nil
+}
+
+// ClassAt gets the contract class definition in the given block associated instantiated by the given contract address
```
jelilat, line 627:
```suggestion
// ClassAt gets the contract class definition in the given block instantiated by the given contract address
```

rpc/handlers.go
omerfirmak: Haha, touche. You win:)

omerfirmak: Added the test

jelilat (approved):

PR #730, opened 2023-04-17
Implement starknet_call

omerfirmak:


omerfirmak:
I decided to move away from `starknet_in_rust` in favor of `blockifier` since it looks to be more mature and reliable as a dependency.

PR #735, opened 2023-04-23
Build state history

omerfirmak:


core/history_test.go
```diff
@@ -0,0 +1,96 @@
...
+	testDB := pebble.NewMemTest()
+	txn := testDB.NewTransaction(true)
+	t.Cleanup(func() {
+		require.NoError(t, txn.Discard())
+		require.NoError(t, testDB.Close())
+	})
+
+	history := core.NewHistory(txn)
+
+	contractAddres := new(felt.Felt).SetUint64(123)
```
jelilat, line 23:
```suggestion
	contractAddress := new(felt.Felt).SetUint64(123)
```

core/history_test.go
```diff
@@ -0,0 +1,96 @@
...
```
jelilat, line 80:
```suggestion
	contractAddress := new(felt.Felt).SetUint64(123)
```

core/history_test.go
```diff
@@ -0,0 +1,96 @@
...
+	t.Run("get value on height that change happened ", func(t *testing.T) {
+		oldValue, err := history.ContractStorageAt(contractAddres, location, 5)
+		require.NoError(t, err)
+		assert.Equal(t, value, oldValue)
+
+		_, err = history.ContractStorageAt(contractAddres, location, 10)
+		assert.EqualError(t, err, "check head state")
+	})
+
+	t.Run("get value between after height 10 ", func(t *testing.T) {
```
jelilat, line 59:
```suggestion
	t.Run("get value after height 10 ", func(t *testing.T) {
```

core/history_test.go
jelilat, line 1: We currently don't have tests for `ContractNonce` and `ContractClassHash`. To avoid repetitiveness, I suggest we create a helper function so we can have something like this

```
func testValueAt(t *testing.T, getValueAt func(*felt.Felt, uint64) (*felt.Felt, error), logValue func(*felt.Felt, *felt.Felt, uint64) error, deleteLog func(*felt.Felt, uint64) error) {
	// Add test cases
}
```

jelilat (changes requested):

core/history_test.go
omerfirmak: done, please see the last commit.

wojciechos (approved):

core/history.go
```diff
@@ -0,0 +1,135 @@
...
+func (h *History) valueAt(key []byte, height uint64) ([]byte, error) {
+	it, err := h.txn.NewIterator()
+	if err != nil {
+		return nil, err
+	}
+
+	for it.Seek(logDBKey(key, height)); it.Valid(); it.Next() {
+		seekedKey := it.Key()
+		// seekedKey size should be `len(key) + sizeof(uint64)` and seekedKey should match key prefix
+		if len(seekedKey) != len(key)+8 || !bytes.HasPrefix(seekedKey, key) {
```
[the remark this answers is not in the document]
omerfirmak: Even tho it is a `for` loop, it is only expected to run for 2 iterations at max and typically it will only run a single iteration.

core/history.go
```diff
@@ -0,0 +1,135 @@
...
+
+		val, itErr := it.Value()
+		if err = db.CloseAndWrapOnError(it.Close, itErr); err != nil {
+			return nil, err
+		}
+		// seekedHeight > height
+		return val, nil
+	}
+
+	return nil, db.CloseAndWrapOnError(it.Close, errors.New("check head state"))
```
jelilat, line 64: What do you think about creating an error type for `errors.New("check head state")`? It seems to be referenced a couple of times even outside `history`

jelilat (approved):

core/history.go
omerfirmak: All references are in tests if I am not mistaken. So until we match the error somewhere in production code, it should stay as a plain string error. 

I am working on a separate PR right now, that matches the "check head state" error in actual code. So we will have to make it a global error var in that PR.

PR #738, opened 2023-04-24
Record when a contract was deployed

omerfirmak:
this is useful for snapshots to tell if a contract was deployed before the snapshot block number

### DIFF
--- a/core/state.go
+++ b/core/state.go
@@ -1,6 +1,7 @@
 package core
 
 import (
+	"encoding/binary"
 	"errors"
 	"fmt"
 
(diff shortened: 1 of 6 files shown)


core/state.go
```diff
@@ -390,3 +399,18 @@ func (s *State) updateDeclaredClasses(declaredClasses []DeclaredV1Class) error {
 
 	return classesCloser()
 }
+
+// ContractIsDeployedAt returns if contract at given addr was deployed at blockNumber
+func (s *State) ContractIsDeployedAt(addr *felt.Felt, blockNumber uint64) (bool, error) {
```
wojciechos, line 404: Shouldn't the function name be more like `ContractIsDeployedAtOrBefore`?

omerfirmak: Probably yes, I just wanted to keep the naming consistent with the other functions in the same interface.

How does `ContractIsAlreadyDeployedAt` sound?

wojciechos: Yeah, sound much better to me

omerfirmak: Done

core/state.go
```diff
@@ -390,3 +399,18 @@ func (s *State) updateDeclaredClasses(declaredClasses []DeclaredV1Class) error {
 
 	return classesCloser()
 }
+
+// ContractIsDeployedAt returns if contract at given addr was deployed at blockNumber
```
jelilat, line 403: The comment should be updated as well

jelilat (approved):
Dropped a comment. I think we can merge once that is fixed

PR #741, opened 2023-04-26
Implement getStorageAt and getClassHashAt

jelilat:
Closes https://github.com/NethermindEth/juno/issues/672 and https://github.com/NethermindEth/juno/issues/674

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,9 +139,9 @@ The focus of Phase 2 will be to Verify the state from layer 1 and implement the 
     * [ ] Implement the remaining endpoints:
         * [X] `starknet_syncing`
         * [X] `starknet_getNonce`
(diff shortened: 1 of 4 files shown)


jelilat:
Imo, there are too many duplicated codes. So, still exploring the possibility of refactoring things a bit

omerfirmak:
> Imo, there are too many duplicated codes. So, still exploring the possibility of refactoring things a bit

> Imo, there are too many duplicated codes. So, still exploring the possibility of refactoring things a bit

Duplication is mostly in tests and I think it is inevitable without some overly-engineered-little-to-no-benefit refactors. We can worry about them later

omerfirmak (approved):

PR #742, opened 2023-04-27
Add workflows for deploy juno to test envs

wojciechos:
This PR introduces two GitHub Actions workflows:

- **upgrade_nodes_on_test_envs:** This workflow is triggered on a push to the main branch or when a new release is created. In the case of a push to the main branch, nodes in the "staging environment" will be upgraded to the latest version from main. When a new release is created, nodes in the "release environment" will be upgraded to the new release version.

- **run_node_on_new_vm**: This workflow can be triggered manually through the GitHub UI, or it will run automatically every Friday night. It allows developers to select a branch and network, and then triggers an action that creates a VM and runs a Juno node on it. This workflow is designed to make it easy for developers to test Juno from their branches with just one click, or to have periodic full sync on a weekly basis.


omerfirmak:
Could/Should we somehow handle breaking changes to DB as well? Like, if there are breaking changes, clear the DB before deploying.

Or maybe we should just bite the bullet and start migrating the DB instead.

wojciechos:
> Could/Should we somehow handle breaking changes to DB as well? Like, if there are breaking changes, clear the DB before deploying.
> 
> Or maybe we should just bite the bullet and start migrating the DB instead.

Maybe we can set up the workflow to clear the database when necessary by using tags such as "required_resync" on commits. Are database changes going to be frequent? If not, I would like to be aware of them and observe how the upgraded node fails before manually clearing the DB.

omerfirmak:
> Maybe we can set up the workflow to clear the database when necessary by using tags such as "required_resync" on commits. Are database changes going to be frequent? If not, I would like to be aware of them and observe how the upgraded node fails before manually clearing the DB.

Let's not bother for now. After the next release, we should start migrating the DB if we ever need to introduce a breaking change.

omerfirmak (approved):

PR #746, opened 2023-05-02
 Implement starknet_getEvents

omerfirmak:


starknetdata/feeder/feeder.go
```diff
@@ -80,6 +80,7 @@ func adaptBlock(response *feeder.Block) (*core.Block, error) {
 			SequencerAddress: response.SequencerAddress,
 			TransactionCount: uint64(len(response.Transactions)),
 			EventCount:       eventCount,
+			EventsBloom:      core.EventsBloom(receipts),
```
joshklop, line 83: Should we modify the tests to account for this?

rpc/handlers.go
```diff
@@ -635,3 +642,85 @@ func (h *Handler) ClassAt(id *BlockID, address *felt.Felt) (*Class, *jsonrpc.Err
 	}
 	return h.Class(id, classHash)
 }
+
+// Events gets the events matching a filter
+//
+// It follows the specification defined here:
+// https://github.com/starkware-libs/starknet-specs/blob/master/api/starknet_api_openrpc.json#L586
```
joshklop, line 649:
```suggestion
// https://github.com/starkware-libs/starknet-specs/blob/94a969751b31f5d3e25a0c6850c723ddadeeb679/api/starknet_api_openrpc.json#LL642C25-L642C25
```

starknetdata/feeder/feeder.go
omerfirmak: I think only thing we can test is that the EventsBloom is not nil. Behaviour of the bloom filter is tested in EventFilter already.

jelilat:
Although I haven't had time to review this PR, I'm currently running a Juno node together with checkpoint.fyi and from my test, the `getEvents` endpoint isn't compatible with it. Works well when I use infura

wojciechos:
When sending a _starknet_getEvents_ request with an empty object in the params array, the server encounters a panic due to an invalid memory address or nil pointer dereference.

**Request**:
```
{
    "jsonrpc":"2.0",
    "method":"starknet_getEvents",
    "params": [
        { }
    ],
	"id":1
}
```
**Response**:
Could not get response
Error: socket hang up

**Logs**:
```
runtime error: invalid memory address or nil pointer dereference
goroutine 754474 [running]:
net/http.(*conn).serve.func1()
        /usr/local/go/src/net/http/server.go:1854 +0xbf
panic({0xd80720, 0x14f5050})
        /usr/local/go/src/runtime/panic.go:890 +0x263
github.com/NethermindEth/juno/rpc.(*Handler).Events(0xc0017d0420?, 0x2?)
        /app/rpc/handlers.go:651 +0x58
reflect.Value.call({0xd5a120?, 0xc002340cb0?, 0x7fd40964fa58?}, {0xe6f845, 0x4}, {0xc0014ea510, 0x1, 0xc0016437c0?})
        /usr/local/go/src/reflect/value.go:586 +0xb0b
reflect.Value.Call({0xd5a120?, 0xc002340cb0?, 0xd5a120?}, {0xc0014ea510?, 0xc00038e0c0?, 0x1?})
        /usr/local/go/src/reflect/value.go:370 +0xbc
github.com/NethermindEth/juno/jsonrpc.(*Server).handleRequest(0xc00237e020, 0xc0014c1ac0)
        /app/jsonrpc/server.go:266 +0x235
github.com/NethermindEth/juno/jsonrpc.(*Server).HandleReader(0x515749?, {0x7fd409121968?, 0xc0014c1a40?})
        /app/jsonrpc/server.go:158 +0x4a5
github.com/NethermindEth/juno/jsonrpc.(*HTTP).ServeHTTP(0xc0003671c0, {0xfcfe10, 0xc001116000}, 0xc000138500)
        /app/jsonrpc/http.go:70 +0x17a
net/http.serverHandler.ServeHTTP({0xfccd40?}, {0xfcfe10, 0xc001116000}, 0xc000138500)
        /usr/local/go/src/net/http/server.go:2936 +0x316
net/http.(*conn).serve(0xc00104d4d0, {0xfd05d8, 0xc00cf022d0})
        /usr/local/go/src/net/http/server.go:1995 +0x612
created by net/http.(*Server).Serve
```

omerfirmak:
> Although I haven't had time to review this PR, I'm currently running a Juno node (on this PR) together with checkpoint.fyi and from my test, the `getEvents` endpoint isn't compatible with it. Works well when I use infura.
> 
> Other endpoints like `getBlock` works well with it though

Please provide more information, maybe someone can pick it up while I am gone.

jelilat:
> > Although I haven't had time to review this PR, I'm currently running a Juno node (on this PR) together with checkpoint.fyi and from my test, the `getEvents` endpoint isn't compatible with it. Works well when I use infura.
> > Other endpoints like `getBlock` works well with it though
> 
> Please provide more information, maybe someone can pick it up while I am gone.

Looked again and it's because of the issue @wojciechos mentioned.

omerfirmak:
Ready to be reviewed and merged

blockchain/event_filter.go
```diff
@@ -0,0 +1,198 @@
...
+	"github.com/NethermindEth/juno/db"
+)
+
+var errChunkSizeReached = errors.New("chunk size reached")
+
+type EventFilter struct {
+	txn       db.Transaction
+	FromBlock uint64
+	ToBlock   uint64
+	From      *felt.Felt
```
joshklop, line 18: Personal preference: could we rename `From` to `ContractAddress`? Reads easier in the implementation imho.

blockchain/event_filter.go
```diff
@@ -0,0 +1,198 @@
...
+			possiblyMatches = header.EventsBloom.Test(keyBytes[:])
+			if possiblyMatches {
+				break
+			}
+		}
+
+		// bloom filter says no events match the filter, skip this block entirely if from is not nil
+		if !possiblyMatches {
+			continue
+		}
```
joshklop, line 132: It took me a second to connect this to `core.EventsBloom`. I feel the two snippets should be close to each other in the codebase, since they change together.

What do you think about an `events` package?

- Exports: `Filterer` interface, `Filter` struct, and `Query` struct
- Create a filter with: `events.NewFilter(r blockchain.Reader) *Filter`
- Execute a query with: `filter.Filter(q *Query) []*core.Event`
- Calculate bloom filter with: `events.Bloom(receipts []*core.TransactionReceipt) *bloom.BloomFilter` 

The `rpc` package can manage the continuation token details, while the blockchain is behind an interface and decoupled from event filtering.

joshklop (changes requested):
Does the spec specify if the `from_block` and `to_block` are inclusive or exclusive? I couldn't tell.

core/transaction.go
```diff
@@ -69,6 +70,13 @@ var (
 	_ Transaction = (*L1HandlerTransaction)(nil)
 )
 
+const (
+	// Calculated at https://hur.st/bloomfilter/?n=1000&p=&m=8192&k=
+	// provides 1 in 51 possibility of false positives for approximately 1000 events
```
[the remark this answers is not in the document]
omerfirmak: Yeah, that should 1000 items, not events.

rpc/handlers.go
```diff
@@ -14,12 +14,19 @@ import (
...
+	ErrNoBlock                  = &jsonrpc.Error{Code: 32, Message: "There are no blocks"}
+	ErrInvalidTxIndex           = &jsonrpc.Error{Code: 27, Message: "Invalid transaction index in a block"}
+	ErrClassHashNotFound        = &jsonrpc.Error{Code: 28, Message: "Class hash not found"}
+	ErrInvalidContinuationToken = &jsonrpc.Error{Code: 33, Message: "Invalid continuation token"}
+	ErrPageSizeTooBig           = &jsonrpc.Error{Code: 31, Message: "Requested page size is too big"}
+	ErrInternal                 = &jsonrpc.Error{Code: jsonrpc.InternalError, Message: "Internal error"}
+)
+
+const (
+	MaxEventChunkSize = 10240
```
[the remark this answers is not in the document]
omerfirmak: No reason, we could make it configurable later.

omerfirmak:
> Does the spec specify if the `from_block` and `to_block` are inclusive or exclusive? I couldn't tell.

I think they should be both inclusive. Or you cant search all the blocks, from=0 to=latest shouldnt exclude any blocks. That is how I implemented it at the least

blockchain/event_filter.go
omerfirmak: Why I had to go with this implementation was consistency. All DB access should be handled by a single db.Transaction or we risk returning inconsistent results. If we rely on blockchain.Reader interface to access the DB, it will create separate Transactions for each request. That is why EventFilter has a Transaction reference and uses private blockchain package functions that operate within a Transaction context.

By letting Blockchain create the EventFilter we are ensuring the entire operation is atomicly executed within a single db.Transaction context and the result is consistent.

Think of the scenario where user defines the range in terms of hashes. Between we verify that block hash exists and actually run the filter on that block, that block could be reorged and leave us in a weird situation.

or think of the scenario where we are mid way filtering the events and the entire range that user defined gets reorged. Half of our results will be pre-reorg, the other half will be post-reorg.

blockchain/event_filter.go
```diff
@@ -0,0 +1,194 @@
...
+		txn:       txn,
+		From:      from,
+		Keys:      keys,
+		FromBlock: fromBlock,
+		ToBlock:   toBlock,
+	}
+}
+
+// SetRangeEndBlockByNumber sets an end of the block range by block number
+func (f *EventFilter) SetRangeEndBlockByNumber(end FilterRangeEnd, blockNumber uint64) error {
```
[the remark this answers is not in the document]
omerfirmak: SetRange functions are there because filter range could be defined in terms of block numbers or block hashes. That makes 4 possible combinations (ie both ends are number, both are hashes, only 1 is nber the other one is a hashes). So we either need to define a new variadic type or have 4 NewEventFilter functions. I didnt want to define a variadic type like rpc.BlockId in blockchain package because I considered it to be a unnecessary leakage of RPC into our business logic. 

this was the simplest API that I could think of, open to any suggestions.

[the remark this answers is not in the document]
joshklop: I think we can let the event filter deal exclusively with numbers. The caller should have access to the methods defined in `blockchain.Reader` and can convert hashes to numbers if necessary.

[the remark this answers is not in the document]
omerfirmak: As I explained [here](https://github.com/NethermindEth/juno/pull/746#discussion_r1192975773), we cant rely on blockchain.Reader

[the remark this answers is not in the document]
joshklop: I mean that if they're calling methods on the `Blockchain` struct (i.e., `blockchain.EventFilter`), they should have access to the methods in `blockchain.Reader`.

blockchain/event_filter.go
omerfirmak: I think completely abstracting DB behind the Blockchain still makes sense. But regardless, that would be a huge refactor, maybe in a separate PR after giving it some consideration.

blockchain/event_filter.go
```diff
@@ -0,0 +1,198 @@
...
+	for _, receipt := range receipts {
+		for _, event := range receipt.Events {
+			// if last request was interrupted mid-block, and we are still processing that block, skip events
+			// that were already processed
+			if cToken != nil && header.Number == cToken.fromBlock && processedEvents < cToken.processedEvents {
+				processedEvents++
+				continue
+			}
+
+			if f.contractAddress != nil && !event.From.Equal(f.contractAddress) {
```
IronGauntlets, line 167: What does it mean to have an `f.contractAddress == nil`?

blockchain/event_filter.go
```diff
@@ -0,0 +1,198 @@
...
+}
+
+type FilterRangeEnd byte
+
+const (
+	From FilterRangeEnd = iota
+	To
+)
+
+func newEventFilter(txn db.Transaction, contractAddress *felt.Felt, keys []*felt.Felt, fromBlock, toBlock uint64) *EventFilter {
```
IronGauntlets, line 29: We should ensure that contract address is always non-nil

blockchain/event_filter.go
```diff
@@ -0,0 +1,198 @@
...
+}
+
+func (t *ContinuationToken) String() string {
+	return fmt.Sprintf("%d-%d", t.fromBlock, t.processedEvents)
+}
+
+func (t *ContinuationToken) FromString(str string) error {
+	_, err := fmt.Sscanf(str, "%d-%d", &t.fromBlock, &t.processedEvents)
+	return err
+}
```
IronGauntlets, line 81: I like this simple approach to managing continuation tokens, I thought we may have to do something more complicated and maintain a record of different tokens similar to cookies.

Do you think parse checking is enough to return `ErrInvalidContinuationToken`?

There are no checks/tests to define what happens if a structurally correct token is passed but it is for a different request. The result could be unpredictable. Is it sufficient to assume that it is the user's responsibility to provide a correct token?

rpc/handlers_test.go
```diff
@@ -1341,3 +1343,132 @@ func TestClassAt(t *testing.T) {
...
+			ChunkSize:         100,
+			ContinuationToken: "",
+		},
+	}
+
+	t.Run("filter non-existent", func(t *testing.T) {
+		t.Run("block number", func(t *testing.T) {
+			args.ToBlock = &rpc.BlockID{Number: 55}
+			_, err := handler.Events(args)
+			require.Equal(t, rpc.ErrBlockNotFound, err)
```
IronGauntlets, line 1385: What is the reason for not checking the result of `handler.Events(args)`? At the very least you should check the len of the filtered events which are returned, right?

rpc/handlers_test.go
```diff
@@ -1341,3 +1343,132 @@ func TestClassAt(t *testing.T) {
...
+		_, err := handler.Events(args)
+		require.Nil(t, err)
+	})
+
+	t.Run("filter with no address", func(t *testing.T) {
+		args.ToBlock = &rpc.BlockID{Latest: true}
+		args.Address = nil
+		_, err := handler.Events(args)
+		require.Nil(t, err)
+	})
```
IronGauntlets, line 1414: I don't see how these tests are meaningful, all they are checking is that there is no error but there are no assert statements on what the actual result should be.

I don't know what the behaviour should be if `fromBlock`/`toBlock`/`contractAddress` is missing from the filter (and the spec isn't any better in specifying what it means).

rpc/events.go
```diff
@@ -0,0 +1,32 @@
+package rpc
+
+import "github.com/NethermindEth/juno/core/felt"
+
+type EventsArg struct {
+	*EventFilter
+	*ResultPageRequest
```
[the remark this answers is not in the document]
omerfirmak: No reason, made them values.

rpc/handlers.go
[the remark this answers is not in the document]
omerfirmak: Correct

blockchain/event_filter.go
```diff
@@ -0,0 +1,198 @@
...
+	contractAddress *felt.Felt
+	keys            []*felt.Felt
+}
+
+type FilterRangeEnd byte
+
+const (
+	From FilterRangeEnd = iota
+	To
+)
```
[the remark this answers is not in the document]
omerfirmak: Explained it [here](https://github.com/NethermindEth/juno/pull/746#discussion_r1192975773) and [here](https://github.com/NethermindEth/juno/pull/746#discussion_r1192975773)

blockchain/event_filter.go
```diff
@@ -0,0 +1,198 @@
...
+		}
+
+		header, err := blockHeaderByNumber(f.txn, curBlock)
+		if err != nil {
+			return nil, nil, err
+		}
+
+		// test `from` only by default if it's not nil because empty filter keys means match all
+		possiblyMatches := true
+		if f.contractAddress != nil {
```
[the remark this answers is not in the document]
omerfirmak: > Shouldn't a contract address always be provided?

no, it is optional

> What does it mean to have keys that are not associated with a contract address?

that you only filter by key

> What does it mean to have keys that are not associated with a contract address?

yes, with nil contractAddress and non-nil keys, you get events with the same keys but from different contracts

blockchain/event_filter.go
```diff
@@ -0,0 +1,198 @@
...
+						Event:           event,
+					})
+				} else {
+					// we are at the capacity, return what we have accumulated so far and a continuation token
+					return matchedEventsSofar, processedEvents, errChunkSizeReached
+				}
+			}
+			// count the events we processed for this block to include in the continuation token
+			processedEvents++
+		}
```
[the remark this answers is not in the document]
omerfirmak: Yes, that is the expected behaviour

blockchain/event_filter.go
omerfirmak: No, it is optional

rpc/handlers_test.go
omerfirmak: It is returning an error, so no events will be returned.

rpc/handlers_test.go
omerfirmak: It checks that the arguments have a default value, (ie it is possible to call getEvents with bunch of arguments missing).

in our implementation;

default fromBlock = 0
default toBlock = latest
default contractAddress = nil (which mean what I explained [here](https://github.com/NethermindEth/juno/pull/746#discussion_r1197358092))

blockchain/event_filter.go
omerfirmak: https://github.com/NethermindEth/juno/pull/746#discussion_r1197358092

blockchain/event_filter.go
omerfirmak: > Do you think parse checking is enough to return ErrInvalidContinuationToken?

Yes I think it is. As long as we don't crash (we dont), it is acceptable. User shouldnt misuse the API and expect to get consistent results.

blockchain/event_filter.go
```diff
@@ -0,0 +1,198 @@
...
+		if cToken != nil && curBlock < cToken.fromBlock {
+			continue
+		}
+
+		header, err := blockHeaderByNumber(f.txn, curBlock)
+		if err != nil {
+			return nil, nil, err
+		}
+
+		// test `from` only by default if it's not nil because empty filter keys means match all
```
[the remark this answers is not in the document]
omerfirmak: old comment, removed

rpc/handlers.go
```diff
@@ -14,12 +14,19 @@ import (
...
-	ErrClassHashNotFound = &jsonrpc.Error{Code: 28, Message: "Class hash not found"}
+	ErrBlockNotFound            = &jsonrpc.Error{Code: 24, Message: "Block not found"}
+	ErrContractNotFound         = &jsonrpc.Error{Code: 20, Message: "Contract not found"}
+	ErrTxnHashNotFound          = &jsonrpc.Error{Code: 25, Message: "Transaction hash not found"}
+	ErrNoBlock                  = &jsonrpc.Error{Code: 32, Message: "There are no blocks"}
+	ErrInvalidTxIndex           = &jsonrpc.Error{Code: 27, Message: "Invalid transaction index in a block"}
+	ErrClassHashNotFound        = &jsonrpc.Error{Code: 28, Message: "Class hash not found"}
+	ErrInvalidContinuationToken = &jsonrpc.Error{Code: 33, Message: "Invalid continuation token"}
+	ErrPageSizeTooBig           = &jsonrpc.Error{Code: 31, Message: "Requested page size is too big"}
+	ErrInternal                 = &jsonrpc.Error{Code: jsonrpc.InternalError, Message: "Internal error"}
```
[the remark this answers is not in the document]
omerfirmak: done

blockchain/event_filter.go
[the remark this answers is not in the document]
IronGauntlets: Both links are the same, which other comment where are you referring to?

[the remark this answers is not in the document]
omerfirmak: Oops, updated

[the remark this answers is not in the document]
IronGauntlets: I found the other related comment. I agree with your decision on the range and blockchain having the DB.

rpc/handlers_test.go
IronGauntlets: > default fromBlock = 0
default toBlock = latest

How is this enforced?

IronGauntlets (approved):
LGTM! 

Please clean the commit history before merging.

Issue #716, opened 2023-04-04
Make colour codes configurable


jelilat:
We should add a flag that lets users enable colour codes.

tinoh9:
Hi, can I try this one? Thanks.

omerfirmak:
Hi @tinoh9, sure. We would be glad:) Assigning it to you.

tinoh9:
Should I create a new .go file which stores all color variables, and import it to all relevant string outputs that need to be colorized in CLI?

omerfirmak:
> Should I create a new .go file which stores all color variables, and import it to all relevant string outputs that need to be colorized in CLI?

That sounds too complicated, see the [PR](https://github.com/NethermindEth/juno/pull/705) that disabled the color output. It should as simple as just passing the old value `EncodeLevel` function there.

PR #752, opened 2023-05-03
Add colorized code output flag in CLI

tinoh9:
Based on #716, the purpose of this PR is to add a flag which let users enable colorized code output. Please kindly let me know if I need to modify something.

### DIFF
--- a/cmd/juno/juno.go
+++ b/cmd/juno/juno.go
@@ -31,13 +31,13 @@ const (
 	dbPathF   = "db-path"
 	networkF  = "network"
 	pprofF    = "pprof"
-	colorF    = "color"
+	colourF   = "colour"
 
 	defaultConfig  = ""
 	defaultRPCPort = uint16(6060)
 	defaultDBPath  = ""
 	defaultPprof   = false
-	defaultColor   = true
+	defaultColour  = true
 
 	configFlagUsage   = "The yaml configuration file."
 	logLevelFlagUsage = "Options: debug, info, warn, error."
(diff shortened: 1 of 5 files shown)


cmd/juno/juno.go
```diff
@@ -31,11 +31,13 @@ const (
 	dbPathF   = "db-path"
 	networkF  = "network"
 	pprofF    = "pprof"
+	colorF    = "color"
 
 	defaultConfig  = ""
 	defaultRPCPort = uint16(6060)
 	defaultDBPath  = ""
 	defaultPprof   = false
+	defaultColor   = false
```
omerfirmak, line 40: lets make it `true` by default

utils/log.go
```diff
@@ -91,10 +91,14 @@ func NewNopZapLogger() *ZapLogger {
...
-func NewZapLogger(logLevel LogLevel) (*ZapLogger, error) {
+func NewZapLogger(logLevel LogLevel, color bool) (*ZapLogger, error) {
 	config := zap.NewProductionConfig()
 	config.Encoding = "console"
-	config.EncoderConfig.EncodeLevel = zapcore.CapitalLevelEncoder
+	if color {
+		config.EncoderConfig.EncodeLevel = zapcore.CapitalColorLevelEncoder
+	} else {
+		config.EncoderConfig.EncodeLevel = zapcore.CapitalLevelEncoder
+	}
```
joshklop, line 101: Thanks for the contribution!
```suggestion
	config.EncoderConfig.EncodeLevel = zapcore.CapitalLevelEncoder
	if color {
		config.EncoderConfig.EncodeLevel = zapcore.CapitalColorLevelEncoder
	}
```

cmd/juno/juno.go
tinoh9: So it means the output will be colorized by default? Unless set it to false to disable colorized output.

omerfirmak: Colored by default and `—color false` to disable. Would that work?

tinoh9: Would `--color=false` be okay? Because I found this https://github.com/spf13/cobra/issues/613 that said cannot use `--bool [true|false]` nor the shorthand `-b[true|false]`.

omerfirmak: LGTM

omerfirmak:
Please fix the CI failures

IronGauntlets (approved):

Issue #747, opened 2023-05-02
Use `json.Decoder` in `feeder` to avoid reading the entire response body in to memory


omerfirmak:
We read the entire feeder gateway response in to memory and it is one of biggest memory allocators during sync

https://github.com/NethermindEth/juno/blob/main/clients/feeder/feeder.go#L176

instead, we can just return the body from `get` as a `io.ReaderCloser` and callee can use a `json.Decoder` to stream the response data in to the expected type.

PR #753, opened 2023-05-03
Update response body reading

rafiramadhana:
Issue: #747 

Update response body reading in feeder to use `json.Decoder`.

### DIFF
--- a/clients/feeder/feeder.go
+++ b/clients/feeder/feeder.go
@@ -154,10 +154,9 @@ func (c *Client) buildQueryString(endpoint string, args map[string]string) strin
 }
 
 // get performs a "GET" http request with the given URL and returns the response body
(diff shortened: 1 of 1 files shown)


clients/feeder/feeder.go
```diff
@@ -211,9 +203,10 @@ func (c *Client) StateUpdate(ctx context.Context, blockNumber uint64) (*StateUpd
 	if err != nil {
 		return nil, err
 	}
+	defer body.Close()
```
omerfirmak, line 206: We dont defer statements that can return an error. Can you please explicitly close the body and handle the error on each branch that `return`?

omerfirmak (commented):
Nice, thanks!

omerfirmak (approved):

Issue #743, opened 2023-05-01
Provide access to the underlying KV storage thru a gRPC service


omerfirmak:
Pebble does not allow simultaneous access to the database. So for the users that need direct access to the DB, we should expose a gRPC service that provides the necessary access to the DB.

There are existing examples of such a service like;

https://github.com/ledgerwatch/interfaces/blob/master/remote/kv.proto

fracek:
To go into more details how I use the remote kv api:

 * in [remote/kv.rs])(https://github.com/apibara/dna/blob/3fd1acd9178d005d71adf4dc27022645efe9bf1a/erigon_dna/src/remote/kv.rs) I created a simple abstraction over it, mostly to automate encoding and decoding of key and values.
 * in [remote/db.rs](https://github.com/apibara/dna/blob/3fd1acd9178d005d71adf4dc27022645efe9bf1a/erigon_dna/src/remote/db.rs) I use the abstraction of kv.rs to read data from Erigon's tables.
 * tables are in [erigon/tables.rs](https://github.com/apibara/dna/blob/3fd1acd9178d005d71adf4dc27022645efe9bf1a/erigon_dna/src/erigon/tables.rs), but it's not super related to the remote kv api.

kirugan:
@fracek I checked rs code that you provided and here is resulting grpc spec. Could you please check is it enough? maybe I missed smth.
```
service DB {
  rpc Tx(stream Cursor) returns (stream Pair);
}

enum Op {
  FIRST = 0;
  SEEK = 1;
  SEEK_EXACT = 2;
  LAST = 3;
  NEXT = 4;
  PREV = 5;
}

message Cursor {
  Op op = 1;
  bytes k = 2;
}

message Pair {
  bytes k = 1;
  bytes v = 2;
}
````

**Comment by fracek** [comment]
It's also missing some code needed to implement [RemoteTx::open_cursor](https://github.com/apibara/dna/blob/3fd1acd9178d005d71adf4dc27022645efe9bf1a/erigon_dna/src/remote/kv.rs#L33)

Changes:

```proto
message Cursor {
  Op op = 1;
  string bucket_name = 2; // used to open a cursor in the given table
  uint32 cursor = 3; // used to specify which cursor to use
  bytes k = 4;
  bytes v = 5; // not used
}
```

```proto
message Pair {
  bytes k = 1;
  bytes v = 2;
  uint32 cursor_id = 3; // needed when using multiple cursors over one tx
[...]

PR #754, opened 2023-05-04
gRPC service for kv storage

kirugan:
Issue: #743


grpc/handlers.go
```diff
@@ -0,0 +1,65 @@
...
+			return err
+		}
+	}
+}
+
+func (h handlers) handleTxCursor(
+	cur *gen.Cursor,
+	tx *tx,
+	server gen.KV_TxServer,
+) error {
```
rianhughes, line 36:
```suggestion
func (h handlers) handleTxCursor(cur *gen.Cursor, tx *tx, server gen.KV_TxServer) error {
```

grpc/handlers.go
```diff
@@ -0,0 +1,65 @@
...
+	if err != nil {
+		return err
+	}
+
+	switch cur.Op {
+	case gen.Op_SEEK:
+		key := utils.Flatten(cur.BucketName, cur.K)
+		it.Seek(key)
+	case gen.Op_NEXT:
+		it.Next()
```
rianhughes, line 47: Just a question: how are `it.Next()` and `it.Seek(key)` being used here? Do you not need to return the results to the caller?

kirugan: For both methods return value is `bool` and I tried to follow existing proto spec (there is no field that corresponds to wrong iterator). Probably in this case it's better to return error, will check

grpc/handlers.go
```diff
@@ -0,0 +1,65 @@
...
+)
+
+type handlers struct {
+	db db.DB
+}
+
+func (h handlers) Tx(server gen.KV_TxServer) error {
+	dbTx := h.db.NewTransaction(false)
+	tx := newTx(dbTx)
+	defer tx.cleanup()
```
omerfirmak, line 17: As a rule, we don't defer functions that can error.

See `db.CloseAndWrapOnError` and please use that to explicitly call cleanup on return statements.

grpc/tx.go
```diff
@@ -0,0 +1,40 @@
...
+
+		t.iterators[id] = it
+	}
+
+	return it, nil
+}
+
+func (t *tx) cleanup() error {
+	for _, it := range t.iterators {
+		it.Close()
```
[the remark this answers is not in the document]
kirugan: I updated go to 1.20 in `go.mod` and `README.md` (we already use it in Docker). There is handy function for that - `errors.Join(...)` ❤️

grpc/handlers.go
kirugan: Rewrote it with `CloseAndWrap...` function.

grpc/grpc_test.go
```diff
@@ -0,0 +1,47 @@
...
+		Op: gen.Op_CURRENT,
+	})
+	require.NoError(t, err)
+
+	pair, err := stream.Recv()
+	if err != nil {
+		if err == io.EOF {
+			fmt.Println("disconnected from server")
+		} else {
+			spew.Dump("error", err)
```
omerfirmak, line 42: I think require.NoError does this already, no?

node/node.go
```diff
@@ -149,6 +151,10 @@ func makeHTTP(port uint16, rpcHandler *rpc.Handler, log utils.SimpleLogger) *jso
 	}, log)
 }
 
+func makeGRPC(port uint16, db db.DB, log utils.Logger) *grpc.Server {
+	return grpc.NewServer(port, db, log)
+}
+
 // Run starts Juno node by opening the DB, initialising services.
```
omerfirmak, line 158: we can inline this?

omerfirmak (approved):
I dont think we need to delay this any longer, let's include it in a release and allow apibara to start testing it.

grpc/grpc_test.go
kirugan: `spew.Dump(...)` is very convinient pretty printer in Go, I was debugging what kind of grpc errors grpc can I get

node/node.go
kirugan: sure, I just wanted to mimic http service with its `makeHTTP(...)`. inlined

grpc/handlers.go
joshklop: Was reading through the code and almost posted the same question :)

Is it better to return the error here @kirugan?

joshklop (approved):

joshklop:
Another thought: unit tests for the grpc package could be a good-first-issue.

grpc/handlers.go
kirugan: Ok, I'll return error here. The problem with provided proto file is that there is no error => cursor correspondence, probably Rust code will add it later.

IronGauntlets:
> Another thought: unit tests for the grpc package could be a good-first-issue.

In general, we don't want this to become a practice. I think it is important for the author of the PR to include tests in the PR. I don't think it is fair on others to write tests for code that was authored by someone else as very few people look forward to writing tests. 

We have `codecov/patch` to keep track of this and ideally, we should aim for 80% (it's not always possible that's why it is not a required condition for merging). 1.08% coverage doesn't seem enough.

For now, we can merge it but @kirugan should write tests for it soon.

grpc/handlers.go
```diff
@@ -0,0 +1,65 @@
...
+		if err != nil {
+			return err
+		}
+
+		err = server.Send(&gen.Pair{
+			K:        it.Key(),
+			V:        v,
+			CursorId: cur.Cursor,
+		})
+	default:
```
fracek, line 60: I believe `Op_FIRST` is not handled.

grpc/handlers.go
fracek: IIRC in rust in case of error we get an empty pair response (with only the cursor id set).

fracek:
First of all, thank you for this! I played around with this and here's my thoughts:

 * It would be nice to have a `Version()` method that returns the current Juno version. Not because it's _that_ useful to know the version, but because it makes it straight forward to test if the connection is going to work (easier than setting up bidi stream).
 * It looks like the `Open` operation is not supported? It's useful to get a `cursor_id` assigned by the server.
 * Since there are no buckets, is it possible to have a `seek_range` operator that positions at the first element greater than or equal the provided key? That way `seek_range(table_prefix || 0)` is equivalent to positioning to the first element of the table.
   - If that's the behavior of `seek`, then it would be useful to have a `seek_exact` that returns an empty kv if the key doesn't match exactly.
 * How would I implement something like `last`? 
 * I also added some smaller comments to the discussion.

kirugan:
@fracek 

> It would be nice to have a Version() method ...

Implemented 

> It looks like the Open operation is not supported? It's useful to get a cursor_id assigned by the server.

Initially I thought client should send `cursorId` and if doesn't exist server creates new iterator for new `cursorId`. I've changed this logic, added `open` operation

> If that's the behavior of seek, then it would be useful to have a seek_exact that returns an empty kv if the key doesn't match exactly.

yep, current `seek` works the way you described `seek_range` operation. I've also added `seek_exact`

> How would I implement something like last?

The only way to do it right now is to iterate over all keys with same prefix until you reach next prefix key. I know it's not super convenient, I'll think how to fix it.

> I also added some smaller comments to the discussion.

Thanks

grpc/handlers.go
kirugan: I thought it's the same as `Op_SEEK` + `Op_Current`

kirugan:
I'll merge this PR. The main functionality is implemented. All subsequent edits and features will be added to other pr

Issue #757, opened 2023-05-08
Make block hash calculation compatible with 0.11.1


jelilat:
Note from Starknet

> 0.11.1 update related to block hash calculation: note that in 0.11.1 we start including the signatures of all transaction types except l1_handler.
> In the past, only invoke transactions had signatures and we included only those in the block hash. This persisted even when we had signatures for declare and other types, and is changed now in 0.11.1

rianhughes:
I'll work on this one.

rianhughes:
Leaving this link here for future reference https://github.com/NethermindEth/juno/pull/760

PR #760, opened 2023-05-09
Transaction signatures are included if specified in blockHashMetaInfo

rianhughes:
The blockhash now depends on the signatures of any transaction specified in the blockHashMetaInfo (except for Invoke transaction types which are always included).


rianhughes:
An additional field ''TxTypesForCommitment map[Transaction]bool" has been included in the blockHashMetaInfo struct. This is where  transaction types whose signatures are to be included in the blockhash are specified. 

Note that for backwards compatibility the Invoke-transaction type does not need to be specified in 'TxTypesForCommitment' as it is implicitly included by default (respecting the pre-post 07 Hash logic). 

The blockHashMetaInfo data is passed to pre07Hash(), post07Hash() and transactionCommitment() so that transactionCommitment() knows which transaction signatures to include in the root Hash.

core/block.go
```diff
@@ -39,9 +39,10 @@ type Block struct {
 }
 
 type blockHashMetaInfo struct {
-	First07Block             uint64     // First block that uses the post-0.7.0 block hash algorithm
-	UnverifiableRange        []uint64   // Range of blocks that are not verifiable
-	FallBackSequencerAddress *felt.Felt // The sequencer address to use for blocks that do not have one
+	First07Block             uint64               // First block that uses the post-0.7.0 block hash algorithm
+	UnverifiableRange        []uint64             // Range of blocks that are not verifiable
+	FallBackSequencerAddress *felt.Felt           // The sequencer address to use for blocks that do not have one
+	TxTypesForCommitment     map[Transaction]bool // (Non-invoke) Transaction types to be included in the transactionCommitment
```
IronGauntlets, line 45: Where is this map populated?

IronGauntlets:
> The blockhash now depends on the signatures of any transaction specified in the blockHashMetaInfo (except for Invoke transaction types which are always included).

I don't know what you are trying to do here. If the block hash calculation were incorrect then our tests would have failed. Please look at this [function](https://github.com/NethermindEth/juno/blob/c3e110e2823b1ee80d1136a53b657b6c273f7068/core/block.go#L90) for block hash verification.

Also, have a look at [this](https://github.com/NethermindEth/juno/blob/main/core/block.go) file for tests.

core/transaction.go
```diff
@@ -385,13 +385,17 @@ func verifyTransactionHash(t Transaction, n utils.Network) *CantVerifyTransactio
...
+func transactionCommitment(transactions []Transaction, metaInfo *blockHashMetaInfo) (*felt.Felt, error) {
 	var commitment *felt.Felt
 	return commitment, trie.RunOnTempTrie(commitmentTrieHeight, func(trie *trie.Trie) error {
 		for i, transaction := range transactions {
 			signatureHash := crypto.PedersenArray()
-			if _, ok := transaction.(*InvokeTransaction); ok {
+
+			if txType, ok := transaction.(*InvokeTransaction); ok {
+				signatureHash = crypto.PedersenArray(transaction.Signature()...)
+			} else if metaInfo.TxTypesForCommitment[txType] {
```
IronGauntlets, line 398: This map is currently empty, thus the `else if` case is never executed.

jelilat:
The PR is for this [issue](https://github.com/NethermindEth/juno/issues/757). @rianhughes please link the issue in the PR so people can have context.

core/block.go
jelilat, line 45: You don't need to add a `TxTypesForCommitment`. Only `transaction.go` needs to be updated.

core/transaction.go
```diff
@@ -385,13 +385,17 @@ func verifyTransactionHash(t Transaction, n utils.Network) *CantVerifyTransactio
...
 				signatureHash = crypto.PedersenArray(transaction.Signature()...)
 			}
```
jelilat, line 400: You should do something like this.

```go
if transaction.Version >= 0.11.1 {
  if _, !ok := transaction.(*L1HandlerTransaction); ok {
	signatureHash = crypto.PedersenArray(transaction.Signature()...)
  } 
} else {
   if _, ok := transaction.(*InvokeTransaction); ok {
	signatureHash = crypto.PedersenArray(transaction.Signature()...)
   }
}
```
We use `semver` for versioning

jelilat (changes requested):

core/transaction.go
rianhughes: This is a much nicer solution, thanks for the help. I'll update it now

core/block.go
```diff
@@ -138,15 +139,39 @@ func VerifyBlockHash(b *Block, network utils.Network) error {
...
+		starknetVersion = "0.0.0"
+	} else {
+		starknetVersion = b.Header.ProtocolVersion
+	}
+	blockVersion, err := semver.NewVersion(starknetVersion)
+	if err != nil {
+		return nil, err
+	}
+	return blockVersion, nil
+
```
omerfirmak, line 169: we already have a function that parses and handles some quirks that starknet block versions have. 
Please extract the functionality you need from that function in to a separate function and use that in both places.

https://github.com/NethermindEth/juno/blob/c3e110e2823b1ee80d1136a53b657b6c273f7068/blockchain/blockchain.go#L44-L65

rianhughes: I've extracted the important parts of checkBlockVersion() into a new function that determines if the blocks protocol version is above 0.11.0 blockVersionPost011()

core/transaction.go
```diff
@@ -409,6 +422,30 @@ func transactionCommitment(transactions []Transaction) (*felt.Felt, error) {
...
+		sep := "."
+		digits := strings.Split(protocolVersion, sep)
+		// pad with 3 zeros in case version has less than 3 digits
+		digits = append(digits, []string{"0", "0", "0"}...)
+
+		// get first 3 digits only
+		blockVersion, err = semver.NewVersion(strings.Join(digits[:3], sep))
+		if err != nil {
+			return false, err
+		}
```
omerfirmak, line 444: instead of duplicating this piece here, you should make it a separate function, something like`parseBlockVersion` that takes the `string` and returns a `semver.Version`

core/transaction.go
```diff
@@ -409,6 +422,38 @@ func transactionCommitment(transactions []Transaction) (*felt.Felt, error) {
...
+// blockVersionPost011 determines if the blocks protocolVersion is >=0.11.1.
+func blockVersionPost011(protocolVersion string) (bool, error) {
+
+	versionConstraint, _ := semver.NewConstraint(">=0.11.1")
+	var blockVersion *semver.Version
+	var err error
+
+	if protocolVersion == "" {
+		blockVersion, _ = semver.NewVersion("0.0.0")
+	} else {
```
omerfirmak, line 434: let's move this bit to `parseBlockVersion`

core/transaction.go
```diff
@@ -409,6 +422,38 @@ func transactionCommitment(transactions []Transaction) (*felt.Felt, error) {
 	})
 }
 
+// blockVersionPost011 determines if the blocks protocolVersion is >=0.11.1.
+func blockVersionPost011(protocolVersion string) (bool, error) {
```
omerfirmak, line 426: we dont need this to be a function, we can inline it to transactionCommitment

core/transaction.go
```diff
@@ -409,6 +422,38 @@ func transactionCommitment(transactions []Transaction) (*felt.Felt, error) {
...
+	} else {
+		blockVersion, err = parseBlockVersion(protocolVersion)
+		if err != nil {
+			return false, err
+		}
+	}
+	return versionConstraint.Check(blockVersion), nil
+}
+
+func parseBlockVersion(version string) (*semver.Version, error) {
```
omerfirmak, line 443: please use this function in https://github.com/NethermindEth/juno/blob/c3e110e2823b1ee80d1136a53b657b6c273f7068/blockchain/blockchain.go#L44-L65 as well

rianhughes: Importing the blockchain package into core creates a cyclic dependency (which is why I effectively copied the checkBlockVersion logic into parseBlockVersion. Do you have any suggestions on how I might resolve this cleanly?

core/transaction.go
```diff
@@ -409,6 +424,28 @@ func transactionCommitment(transactions []Transaction) (*felt.Felt, error) {
 	})
 }
 
+// ParseBlockVersion computes the block version, defaulting to "0.0.0" for empty strings
+func ParseBlockVersion(protocolVersion string) (*semver.Version, error) {
+	var blockVersion *semver.Version
+	var err error
+
+	if protocolVersion == "" {
+		blockVersion, _ = semver.NewVersion("0.0.0")
```
omerfirmak, line 433:
```suggestion
		return semver.NewVersion("0.0.0")
```

core/transaction.go
```diff
@@ -409,6 +424,28 @@ func transactionCommitment(transactions []Transaction) (*felt.Felt, error) {
...
+	} else {
+		sep := "."
+		digits := strings.Split(protocolVersion, sep)
+		// pad with 3 zeros in case version has less than 3 digits
+		digits = append(digits, []string{"0", "0", "0"}...)
+
+		// get first 3 digits only
+		blockVersion, err = semver.NewVersion(strings.Join(digits[:3], sep))
```
omerfirmak, line 441:
```suggestion
		return semver.NewVersion(strings.Join(digits[:3], sep))
```

core/transaction.go
omerfirmak: then we dont need the else branch

rianhughes: Much neater, thanks.

core/transaction.go
```diff
@@ -386,13 +388,25 @@ const commitmentTrieHeight uint = 64
...
 		for i, transaction := range transactions {
 			signatureHash := crypto.PedersenArray()
-			if _, ok := transaction.(*InvokeTransaction); ok {
+
+			if blockVersion.GreaterThan(versionConstraint) {
 				signatureHash = crypto.PedersenArray(transaction.Signature()...)
+			} else {
+				if _, ok := transaction.(*InvokeTransaction); ok {
+					signatureHash = crypto.PedersenArray(transaction.Signature()...)
+				}
```
joshklop, line 409:
```suggestion
			} else if _, ok := transaction.(*InvokeTransaction); ok {
				signatureHash = crypto.PedersenArray(transaction.Signature()...)
```

core/transaction.go
```diff
@@ -409,6 +423,21 @@ func transactionCommitment(transactions []Transaction) (*felt.Felt, error) {
 	})
 }
 
+// ParseBlockVersion computes the block version, defaulting to "0.0.0" for empty strings
+func ParseBlockVersion(protocolVersion string) (*semver.Version, error) {
```
joshklop, line 427: This may be a task for a future PR (or totally unnecessary), but we should consider storing the block version as an enum instead of as a string. That way we're not doing regex/string manipulation whenever we calculate a transaction hash.

joshklop (approved):
Great stuff! :rocket:

PR #761, opened 2023-05-10
Make Event Filters Optional

jelilat:


omerfirmak:
Let’s skip the validation commit for now. It can be done in jsonrpc module, that way we dont need to do it per handler.

blockchain/event_filter.go
```diff
@@ -108,8 +110,12 @@ func (f *EventFilter) Events(cToken *ContinuationToken, chunkSize uint64) ([]*Fi
 			return nil, nil, err
 		}
 
-		// test `from` only by default because empty filter keys means match all
-		possiblyMatches := header.EventsBloom.Test(fromBytes[:])
+		// test `from` only by default if it's not nil because empty filter keys means match all
+		var possiblyMatches bool
```
omerfirmak, line 114: This is not quite the same thing. It breaks the “empty keys mean match all” behavior.

omerfirmak (changes requested):
Making `from` optional requires changes to how we calculated the bloom filter as well.

and to_block and from_block should be optional as well.

blockchain/event_filter.go
jelilat: I accounted for it [here](https://github.com/NethermindEth/juno/pull/761/files#diff-299bd13c9d85e40bab3e1c50f08f697257ed323a0db67685f628d71e3c635518R131)

rpc/handlers.go
```diff
@@ -676,7 +676,17 @@ func (h *Handler) Events(args *EventsArg) (*EventsChunk, *jsonrpc.Error) {
...
-	if err = setEventFilterRange(filter, args.EventFilter.FromBlock, args.EventFilter.ToBlock, height); err != nil {
+	fromBlock := args.EventFilter.FromBlock
+	if fromBlock == nil {
+		fromBlock = &BlockID{Number: 0}
+	}
+
+	toBlock := args.EventFilter.ToBlock
+	if toBlock == nil {
+		toBlock = &BlockID{Latest: true}
+	}
```
omerfirmak, line 687: we can move these to blockchain.EventFilter

jelilat: Or maybe `setEventFilterRange`? `id.Latest` panics if it's `nil`

omerfirmak: The reason I think the default behaviour should be defined in the EventFilter object instead of the RPC is that if the EventFilter is ever used outside of the RPC, it should behave reasonably with just the default values

omerfirmak: we can make setEventFilterRange a noop if block id is nil

omerfirmak:
@jelilat I added a commit here, feel free to take a look and merge if you are happy with it

jelilat:
> @jelilat I added a commit here, feel free to take a look and merge if you are happy with it

That's a much better approach. Thanks

Issue #748, opened 2023-05-03
Sanity check the user input to RPC module


omerfirmak:
`json` package doesnt check that all the fields are present when unmarshaling a RPC argument from JSON.
`jsonrpc` checks that the non-optional arguments are there, but doesnt recursively check that the arguments are sane. (no nil ptrs, etc)

So for example, 

calling an endpoint with the following JSON which expects an `Arg` successfully reaches the handler.

``` json
{
   "val1" : "0xDEADBEEF"
}
```

``` golang
type Arg struct   {
   val1 *felt.Felt `json:"val1"`
   val2 *felt.Felt `json:"val2"`
}
```

and if handler assumes `val2` is non-nil, it can easily cause a panic

joshklop:
Do `UnmarshalJSON`s on the appropriate `rpc` types solve this issue? It seems impossible for the `jsonrpc` module to know what fields are optional in a given situation.

We already do this for `BlockID`. Maybe we should do it for `EventsArg` to fix the [getEvents bug](https://github.com/NethermindEth/juno/pull/746#issuecomment-1534706870)?

omerfirmak:
> Do `UnmarshalJSON`s on the appropriate `rpc` types solve this issue? It seems impossible for the `jsonrpc` module to know what fields are optional in a given situation.
> 
> We already do this for `BlockID`. Maybe we should do it for `EventsArg` to fix the [getEvents bug](https://github.com/NethermindEth/juno/pull/746#issuecomment-1534706870)?

validator package uses tags, so that information is encoded in the type. jsonrpc will only need to pass the value it creates to the validator package and it will extract the all the necessary criteria in runtime.

https://github.com/go-playground/validator/blob/master/_examples/simple/main.go

PR #762, opened 2023-05-10
Add UnmarshalJSON to EventArg

joshklop:
See:
- [General issue](https://github.com/NethermindEth/juno/issues/748)
- [Related bug](https://github.com/NethermindEth/juno/pull/746#issuecomment-1534706870) on getEvents Pr

### DIFF
--- a/rpc/events.go
+++ b/rpc/events.go
@@ -1,10 +1,38 @@
 package rpc
 
-import "github.com/NethermindEth/juno/core/felt"
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+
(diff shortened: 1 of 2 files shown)


rpc/events.go
jelilat, line 1: We want to make all fields optional except the `chunk_size`. Starknet will be making changes to the rpc requirement.

Also, @omerfirmak said we should add the validation in a separate PR so that we can do it for all endpoints and not just `starknet_getEvents` (or individual endpoints). https://github.com/NethermindEth/juno/pull/761#issuecomment-1541438294

joshklop:
Closing in favor of #763

PR #763, opened 2023-05-11
Validate structs in jsonrpc

### LINKED ISSUE #748: Sanity check the user input to RPC module

joshklop:
Establishes some correctness guarantees before passing the struct to the respective handler.

Resolves #748

### DIFF
--- a/jsonrpc/server.go
+++ b/jsonrpc/server.go
@@ -9,6 +9,8 @@ import (
 	"io"
 	"reflect"
 	"strings"
+
+	"github.com/go-playground/validator/v10"
 )
 
 const (
@@ -96,6 +98,8 @@ type Method struct {
 type Server struct {
 	methods map[string]Method
 }
+
+var validate = validator.New()
 
 // NewServer instantiates a JSONRPC server
 func NewServer() *Server {
@@ -298,6 +302,12 @@ func buildArguments(params, handler any, configuredParams []Parameter) ([]reflec
 		err = json.Unmarshal(valueMarshaled, handlerParam.Interface())
 		if err != nil {
 			return reflect.ValueOf(nil), err
+		}
+
+		if err = validate.Struct(handlerParam.Elem().Interface()); err != nil {
+			if _, ok := err.(validator.ValidationErrors); ok {
+				return reflect.ValueOf(nil), err
+			}
 		}
 
 		return handlerParam.Elem(), nil
--- a/jsonrpc/server.go
+++ b/jsonrpc/server.go
@@ -293,26 +293,6 @@ func buildArguments(params, handler any, configuredParams []Parameter) ([]reflec
 
 	handlerType := reflect.TypeOf(handler)
 
-	handlerParamValue := func(param any, t reflect.Type) (reflect.Value, error) {
-		handlerParam := reflect.New(t)
-		valueMarshaled, err := json.Marshal(param) // we have to marshal the value into JSON again
-		if err != nil {
-			return reflect.ValueOf(nil), err
-		}
-		err = json.Unmarshal(valueMarshaled, handlerParam.Interface())
-		if err != nil {
-			return reflect.ValueOf(nil), err
-		}
-
-		if err = validate.Struct(handlerParam.Elem().Interface()); err != nil {
-			if _, ok := err.(validator.ValidationErrors); ok {
-				return reflect.ValueOf(nil), err
-			}
-		}
-
-		return handlerParam.Elem(), nil
-	}
-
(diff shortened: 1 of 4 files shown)


joshklop:
This should be ready for review. It is easiest to review the two commits in sequence.

jsonrpc/server.go
```diff
@@ -300,6 +304,19 @@ func buildArguments(params, handler any, configuredParams []Parameter) ([]reflec
 			return reflect.ValueOf(nil), err
 		}
 
+		// Get the underlying value, dereferencing until we reach the "raw" kind.
+		value := handlerParam.Elem()
+		for value.Kind() == reflect.Pointer {
+			value = value.Elem()
```
omerfirmak: this is not safe, we might get a nil pointer after we deference it couple of times. I dont think we need to deference it multiple times, we can just assume there wont be pointers to pointers in any of the arguments.

Validator deferences the value if it is a pointer anyways.
https://github.com/go-playground/validator/blob/c8b98cbcea08971fbead222ddf5fd406a40d40ed/validator_instance.go#L373-L375

to keep things simple here we can call validate.Struct without doing any checks whether if the argument is a struct or not. If it is not a struct validator will return InvalidValidationError and we can just silently ignore that

https://github.com/go-playground/validator/blob/v10.13.0/validator_instance.go#LL378C11-L378C33

jsonrpc/server.go
```diff
@@ -300,6 +304,19 @@ func buildArguments(params, handler any, configuredParams []Parameter) ([]reflec
...
 		}
 
+		// If the argument is a struct, validate it.
+		value := handlerParam
+		// If the argument to the handler is a pointer, we need to dereference
+		// one time since we initialized handlerParam with reflect.New(t).
+		if t.Kind() == reflect.Pointer {
+			value = value.Elem()
+		}
+		if err = validate.Struct(value.Interface()); err != nil {
```
omerfirmak: We can just pass `handlerParam.Elem().Interface()` here, no? Without any of the checks above.

handlerParam is a pointer since it is created by `reflect.New()`. `Elem()` will dereference it and return the value of type that the handler expects it to be in. If the resulting parameter is a pointer `validate.Struct` will deference it to validate the underlying value.

omerfirmak (approved):
Thanks!

jsonrpc/server.go
```diff
@@ -344,3 +334,24 @@ func buildArguments(params, handler any, configuredParams []Parameter) ([]reflec
 	}
 	return args, nil
 }
+
+func parseParam(param any, t reflect.Type) (reflect.Value, error) {
+	handlerParam := reflect.New(t)
+	valueMarshaled, err := json.Marshal(param) // we have to marshal the value into JSON again
```
kirugan, line 340: why do we need to marshal again?

omerfirmak: JSON RPC allows parameters to be sent as a list or an object.

```
{"jsonrpc": "2.0", "method": "subtract", "params": [42, 23], "id": 1}
{"jsonrpc": "2.0", "method": "subtract", "params": {"subtrahend": 23, "minuend": 42}, "id": 3}
```

to figure out which format the request uses, we first [unmarshal](https://github.com/NethermindEth/juno/pull/763/files#diff-b18e347b821e9f1511d44c3ccb83f52bfe315b2a648678802c06781bd62b8e9fL306-L321) the entire "params" field in to a `any` typed variable. That variable either ends up being a `[]any` or `map[string]any`

then each element of the list or map needs to be converted to the type that the handler expects (`t` in this piece of code). So we need a way to convert an object in type `any` to a some arbitrary concrete type `t`. Only way that I could think of when I first wrote this JSONRPC server was to marshal the `any` value back to JSON and then unmarshal it to type `t`

omerfirmak: This is out of scope for this PR, but feel free to suggest any alternatives. It would be really nice to get rid of this round trip to and from JSON.

wojciechos (approved):

PR #764, opened 2023-05-12
Calculate Trie commitment lazily

### LINKED ISSUE #751: Calculate Trie commitment lazily
Currently, we are calculating commitment of a `Trie` on every `Put` call, ie the `propagateValues` function. https://github.com/NethermindEth/juno/blob/6baadd09c38e08d8a1d6fb4f1af3f7bde26d78b7/core/trie/trie.go#L328-L363 This could result in bunch of unnecessary hashing.

What we can do instead is update only the relation between (ie update the children, add new nodes, delete old ones) the nodes of a Trie in the Put function and but dont update the `Value` fields. By keeping track of which nodes were changed, we can calculate the commitment only when the user signals the Trie that it is done updating or when a `Root()` call is made

omerfirmak:
defer commitment calculation until the user calls Trie.Commit or Trie.Root explicitly. This results in %15 less Pedersen calculations during sync

Fixes #751

### DIFF
--- a/core/state.go
+++ b/core/state.go
@@ -166,6 +166,10 @@ func (s *State) globalTrie(bucket db.Bucket, newTrie trie.NewTrieFunc) (*trie.Tr
 
 	// prep closer
 	closer := func() error {
+		if err = gTrie.Commit(); err != nil {
+			return err
+		}
+
 		resultingRootKey := gTrie.RootKey()
 		// no updates on the trie, short circuit and return
 		if resultingRootKey.Equal(rootKey) {
--- a/core/state.go
+++ b/core/state.go
@@ -53,7 +53,7 @@ func NewState(txn db.Transaction) *State {
 
 // putNewContract creates a contract storage instance in the state and stores the relation between contract address and class hash to be
 // queried later with [GetContractClass].
-func (s *State) putNewContract(addr, classHash *felt.Felt, blockNumber uint64) error {
+func (s *State) putNewContract(stateTrie *trie.Trie, addr, classHash *felt.Felt, blockNumber uint64) error {
 	contract, err := DeployContract(addr, classHash, s.txn)
 	if err != nil {
 		return err
@@ -65,7 +65,7 @@ func (s *State) putNewContract(addr, classHash *felt.Felt, blockNumber uint64) e
 		return err
 	}
 
-	return s.updateContractCommitment(contract)
+	return s.updateContractCommitment(stateTrie, contract)
 }
 
 // ContractClassHash returns class hash of a contract at a given address.
@@ -228,16 +228,21 @@ func (s *State) Update(blockNumber uint64, update *StateUpdate, declaredClasses 
 }
 
 func (s *State) updateContracts(blockNumber uint64, diff *StateDiff) error {
+	stateTrie, storageCloser, err := s.storage()
+	if err != nil {
+		return err
+	}
+
 	// register deployed contracts
 	for _, contract := range diff.DeployedContracts {
-		if err := s.putNewContract(contract.Address, contract.ClassHash, blockNumber); err != nil {
+		if err := s.putNewContract(stateTrie, contract.Address, contract.ClassHash, blockNumber); err != nil {
 			return err
 		}
 	}
 
 	// replace contract instances
 	for _, replace := range diff.ReplacedClasses {
-		oldClassHash, err := s.replaceContract(replace.Address, replace.ClassHash)
+		oldClassHash, err := s.replaceContract(stateTrie, replace.Address, replace.ClassHash)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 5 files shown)


core/trie/trie.go
```diff
@@ -191,172 +193,173 @@ func (t *Trie) Put(key, value *felt.Felt) (*felt.Felt, error) {
...
+	if parent.node.Left.Equal(last.key) {
+		siblingKey = parent.node.Right
+	} else {
+		siblingKey = parent.node.Left
+	}
 
-			cur.node.Value = t.hash(left.Hash(leftPath, t.hash), right.Hash(rightPath, t.hash))
+	if len(nodes) == 2 { // sibling should become root
+		t.rootKey = siblingKey
+	} else { // sibling should link to grandparent (len(affectedNodes) > 2)
```
joshklop, line 350: nit: we can remove the else here.

joshklop (commented):
Thinking more generally: erigon only calculates the commitment every N blocks (I think N=50 in their implementation, but can't remember) while syncing. With this PR, we could do something similar.

omerfirmak:
> Thinking more generally: erigon only calculates the commitment every N blocks (I think N=50 in their implementation, but can't remember) while syncing. With this PR, we could do something similar.

Yeap, that is one of the things this PR could enable. But I am not sure how effective it would be in speeding up the sync. Maybe we can give it a try and see. It shouldnt be that hard.

omerfirmak:
I benchmarked this PR against `main` for mainnet block 500, which a decently sized block.

```
before:
BenchmarkMainnetBlock500-16          100         616745108 ns/op        25081993 B/op     637726 allocs/op

after:
BenchmarkMainnetBlock500-16          100         535941507 ns/op        41434650 B/op    1836362 allocs/op
```
There is a considerable speed up. I will try to add #781 on top of this and see how it helps.

omerfirmak:
With a quick implementation of #781 

```
BenchmarkMainnetBlock500-16          100         485868048 ns/op        39204112 B/op    1724185 allocs/op
```

So, overall %22 speed up.

omerfirmak:
```
BenchmarkMainnetBlock500-16          100         293354476 ns/op         8972344 B/op     218413 allocs/op
```

eliminated most of the allocations as well. In this test runtime reduction looks to be around %53

core/state.go
```diff
@@ -224,16 +228,21 @@ func (s *State) Update(blockNumber uint64, update *StateUpdate, declaredClasses
 }
 
 func (s *State) updateContracts(blockNumber uint64, diff *StateDiff) error {
+	stateTrie, storageCloser, err := s.storage()
```
kirugan, line 231: no need to defer `storageCloser()` ?

core/state.go
```diff
@@ -390,16 +399,11 @@ func (s *State) updateContractCommitment(contract *Contract) error {
...
+	if _, err = stateTrie.Put(contract.Address, commitment); err != nil {
 		return err
 	}
 
-	if _, err = state.Put(contract.Address, commitment); err != nil {
-		return err
-	}
-
-	return storageCloser()
+	return nil
```
kirugan, line 406: lets simplify it `return stateTrie.Put(contract.Address, commitment)`

core/trie/transaction_storage.go
```diff
@@ -75,13 +85,25 @@ func (t *TransactionStorage) Get(key *bitset.BitSet) (*Node, error) {
 		return nil, err
 	}
 
+	if t.cache != nil {
```
kirugan, line 88: I think we don't need `t.cache != nil` checks in all methods if the only way we can initiate storage is by calling constructor (`NewTransactionStorage(...)`)

core/trie/transaction_storage.go
```diff
@@ -64,7 +66,15 @@ func (t *TransactionStorage) Put(key *bitset.BitSet, value *Node) error {
 	}
 
 	encodedBytes := buffer.Bytes()
-	return t.txn.Set(encodedBytes[:keyLen], encodedBytes[keyLen:])
+	if err = t.txn.Set(encodedBytes[:keyLen], encodedBytes[keyLen:]); err != nil {
+		return err
+	}
+
+	if t.cache != nil {
+		keyStr := string(encodedBytes[:keyLen])
```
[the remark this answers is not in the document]
omerfirmak: This actually made me realize something.

If `encodedBytes` are all 0, or have 0 somewhere in the middle. How would string() behave it that scenario?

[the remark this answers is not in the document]
omerfirmak: Discussed this in slack, will keep it as it is.

core/trie/transaction_storage.go
```diff
@@ -75,13 +85,25 @@ func (t *TransactionStorage) Get(key *bitset.BitSet) (*Node, error) {
...
+
 	var node *Node
 	if err = t.txn.Get(buffer.Bytes(), func(val []byte) error {
 		node = new(Node)
 		return encoder.Unmarshal(val, node)
 	}); err != nil {
 		return nil, err
 	}
+
+	if t.cache != nil && err == nil {
```
joshklop, line 103: I think we can remove `err == nil`.

omerfirmak:
Addressed all comments

joshklop (approved):

core/trie/transaction_storage.go
omerfirmak: Correct, removed.

PR #766, opened 2023-05-16
Fix shadowed declaration

omerfirmak:


omerfirmak:
Was causing linter to fail on `main`

wojciechos (approved):

joshklop (approved):

PR #768, opened 2023-05-16
Only fetch unknown classes in sync process

### LINKED ISSUE #689: Stop unconditionally fetching all classes referenced in a StateUpdate

### LINKED ISSUE #714: Fetch and store Sierra classes
https://github.com/NethermindEth/juno/blob/8887d659770866ce25d8f1208c89f6d07a072cd1/sync/sync.go#L58-L73

omerfirmak:
Fixes #689
Fixes #714


core/transaction.go
```diff
@@ -340,53 +340,25 @@ func deployAccountTransactionHash(d *DeployAccountTransaction, n utils.Network)
...
-		if err := verifyTransactionHash(tx, n); err != nil {
-			err.next = head
-			head = err
-		}
-	}
-	if head != nil {
-		return *head
+	// blockVersion < 0.11.0
+	// only start verifying transaction hashes after 0.11.0
+	if blockVersion.Compare(semver.MustParse("0.11.0")) == -1 {
```
IronGauntlets, line 351: Please store the version as a `const`.

IronGauntlets (approved):
LGTM! 

I think 3rd commit should be squashed into 2nd commit.

core/transaction.go
omerfirmak: Sadly, `semver.MustParse` can not be assigned to a const

omerfirmak:
> LGTM!
> 
> I think 3rd commit should be squashed into 2nd commit.

I want to keep it in a separate commit in case we ever need that code again. So that we can just revert the commit that deletes the code and not cause any behavior changes.

wojciechos (approved):

Issue #683, opened 2023-03-28
Implement starknet_addInvokeTransaction


jelilat:
Implement the `starknet_addInvokeTransaction` endpoint to follow the Starknet RPC specification defined [here](https://github.com/starkware-libs/starknet-specs/blob/a789ccc3432c57777beceaa53a34a7ae2f25fda0/api/starknet_write_api.json#L11)

rianhughes:
I would like to work on this.

joshklop:
#791

PR #771, opened 2023-05-18
Add invoke tx

rianhughes:


rianhughes:
Related to this GH issue: https://github.com/NethermindEth/juno/issues/683

utils/network.go
```diff
@@ -82,6 +83,23 @@ func (n Network) URL() string {
...
+	case GOERLI2:
+		return "https://alpha4-2.starknet.io/gateway/"
+	case INTEGRATION:
+		return "https://external.integration.starknet.io/gateway/"
+	default:
+		// Should not happen.
+		panic(ErrUnknownNetwork)
+	}
+}
+
```
omerfirmak, line 102: Let's try to reuse `URL` function.

We can make URL to only return the base url like `https://alpha4.starknet.io` and callee can append `feeder_gateway` or `gateway`.

rpc/handlers.go
```diff
@@ -633,3 +639,17 @@ func (h *Handler) ClassAt(id *BlockID, address *felt.Felt) (*Class, *jsonrpc.Err
 	}
 	return h.Class(id, classHash)
 }
+
+// AddInvokeTransaction relays an invoke transaction to the gateway.
+//
+// It follows the specification defined here:
+// https://github.com/starkware-libs/starknet-specs/blob/a789ccc3432c57777beceaa53a34a7ae2f25fda0/api/starknet_write_api.json#L11
+func (h *Handler) AddInvokeTransaction(broadcastInvokeTxn *gateway.BroadcastInvokeTxn) (*core.AddInvokeTxResult, *jsonrpc.Error) {
+	invokeResult, err := h.gatewayClient.AddInvokeTransaction(context.TODO(), broadcastInvokeTxn)
```
omerfirmak, line 648: Context with a timeout should be enough here IMO

node/node.go
```diff
@@ -168,12 +175,15 @@ func (n *Node) Run(ctx context.Context) {
 		}
 	}()
 
-	n.blockchain = blockchain.New(n.db, n.cfg.Network, n.log)
+	network := n.cfg.Network
```
omerfirmak, line 178: why do we need to alias `n.cfg.Network` ? cant we just use `n.cfg.Network`

core/transaction.go
```diff
@@ -63,6 +63,16 @@ type Transaction interface {
...
+type AddInvokeTxResult struct {
+	Hash string `json:"transaction_hash"`
+	Code string `json:"code"`
+}
+type AddInvokeTxError struct {
+	Data    any    `json:"data"`
+	Code    string `json:"code"`
+	Message string `json:"message"`
+}
+
```
omerfirmak, line 75: These dont belong here, they should either be moved to `gateway` or `rpc`, which ever makes more sense

clients/gateway/errors.go
```diff
@@ -0,0 +1,12 @@
...
+var (
+	ErrTransactionLimitExceeded        = "StarknetErrorCode.TRANSACTION_LIMIT_EXCEEDED"
+	ErrOutOfRangeFee                   = "StarknetErrorCode.OUT_OF_RANGE_FEE"
+	ErrInvalidTransactionVersion       = "StarknetErrorCode.INVALID_TRANSACTION_VERSION"
+	ErrInvalidProgram                  = "StarknetErrorCode.INVALID_PROGRAM"
+	ErrDeprecatedTransaction           = "StarknetErrorCode.DEPRECATED_TRANSACTION"
+	ErrInvalidCompiledClassHash        = "StarknetErrorCode.INVALID_COMPILED_CLASS_HASH"
+	ErrCompilationFailed               = "StarknetErrorCode.COMPILATION_FAILED"
+	ErrUnauthorizedEntryPointForInvoke = "StarknetErrorCode.UNAUTHORIZED_ENTRY_POINT_FOR_INVOKE"
+)
```
IronGauntlets, line 12: This should be `const`. Also, where will these errors be used?

core/transaction.go
rianhughes: Ive moved these to gateway since moving them to rpc causes an import cycle error. I've also replaced `AddInvokeTxError` with a new struct `StarknetError` so that it matches what both the sequencer and pathfinder returns

clients/gateway/errors.go
rianhughes: Currently they are not needed, so I've removed them. We might need them in the future if we implement a gateway test Server

utils/network.go
rianhughes: I'm not sure what you mean by reuse the base URL in this package, so I created a new BaseURL() function and used it in both GatewayURL and FeederURL

clients/gateway/gateway.go
```diff
@@ -0,0 +1,101 @@
...
+
+type BroadcastInvokeTxn struct {
+	Type          string   `json:"type"`
+	Version       string   `json:"version"`
+	MaxFee        string   `json:"max_fee"`
+	Signature     []string `json:"signature"`
+	Nonce         string   `json:"nonce"`
+	SenderAddress string   `json:"sender_address"`
+	CallData      []string `json:"calldata"`
+}
```
omerfirmak, line 39: I think these all should be field elements according to the spec. If we are going to unmarshal the Txn into a type rather than just treating as a JSON blob, let's use the actual types of the fields.

rpc/handlers.go
rianhughes: How long should the timeout be, 2 seconds?

kirugan: But +1 for timeout, just keep `context.TODO()` instead of `...Background()`

utils/network.go
```diff
@@ -66,22 +66,33 @@ func (n *Network) UnmarshalText(text []byte) error {
 	return n.Set(string(text))
 }
 
-func (n Network) URL() string {
+// BaseURL returns the base URL without endpoint
+func (n Network) BaseURL() string {
```
kirugan, line 70: since it's used only internally let's unexport it (i.e. `BaseURL()` -> `baseURL()`)

rpc/handlers_test.go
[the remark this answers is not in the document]
rianhughes: I deleted these in a new commit.

clients/gateway/integration_test.go
```diff
@@ -0,0 +1,92 @@
+//go:build integration
+// +build integration
+
+package gateway
```
omerfirmak, line 4: We put our tests in `*_test` packages
```suggestion
package gateway_test
```

rpc/handlers.go
```diff
@@ -734,3 +738,26 @@ func setEventFilterRange(filter *blockchain.EventFilter, fromID, toID *BlockID,
...
 }
+
+// AddInvokeTransaction relays an invoke transaction to the gateway.
+//
+// It follows the specification defined here:
+// https://github.com/starkware-libs/starknet-specs/blob/a789ccc3432c57777beceaa53a34a7ae2f25fda0/api/starknet_write_api.json#L11
+func (h *Handler) AddInvokeTransaction(invokeTx *gateway.BroadcastedInvokeTxn) (*AddInvokeTransactionResponse, *jsonrpc.Error) {
+	if invokeTx.Type == "INVOKE" {
+		invokeTx.Type = "INVOKE_FUNCTION"
+	}
```
omerfirmak, line 749: This shouldnt be necessary, right? It is typed `BroadcastedInvokeTxn` anyways.

rpc/handlers.go
```diff
@@ -734,3 +738,26 @@ func setEventFilterRange(filter *blockchain.EventFilter, fromID, toID *BlockID,
...
+
+	resp, err := h.gatewayClient.AddInvokeTransaction(context.TODO(), invokeTx)
+	if err != nil {
+		return nil, &jsonrpc.Error{
+			Code:    0,
```
omerfirmak, line 754: Are you sure 0 is the right error code here?

rpc/handlers.go
omerfirmak: BroadcastedInvokeTxn is part of the rpc spec. so, IMO, it should be in `rpc` package.

kirugan: The thing about this hack is that rpc spec says `Type` is `INVOKE` and pathfinder follows the spec.
But gateway expects type to be `INVOKE_FUNCTION`. It would be great if we could just pass recevied `BroadcastedInvokeTxn` to gateway but unfortunately things get complicated here.

Regarding moving it to `rpc` package sounds reasonable.

rpc/handlers.go
kirugan: it should be non zero, fixing it

rpc/handlers.go
kirugan: As discussed on daily last Friday: now it's going to be always INVOKE_FUNCTION

kirugan: Regarding moving to `rpc` package: it causes import cycle, which could be solved by introducing private gateway interface in `rpc` package itself:
```
//go:generate mockgen -destination=../../mocks/mock_gateway.go -package=mocks github.com/NethermindEth/juno/clients/gateway Gateway
type gateway interface {
	AddInvokeTransaction(context.Context, *BroadcastedInvokeTxn) (*InvokeResponse, error)
}
```
but in this case type `gateway.InvokeResponse` introduces another cycle :). `InvokeResponse` is type related to gateway communication and shouldn't be moved to `rpc` package. So I think we should keep it as is, we can solve it by copying same type in rpc and gateway packages but it's overkill IMO

rpc/handlers.go
rianhughes: It's not clear what the error code should be in some cases. For example, if the max fee is too high, the sequencer returns an error code of  `{"code": "StarknetErrorCode.OUT_OF_RANGE_FEE", "message": "Fee 0xaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa is out of range"}`, whereas pathfinder (via infura endpoint) returns `{"code":-32603,"message":"Sending V1 invoke transaction to gateway"}`. Should we just specify our own error code?

rianhughes: Any errors related to validation or incorrect versions should be caught prior to this, so it's mainly errors related to high gas fees, etc.

rpc/handlers_test.go
```diff
@@ -1492,3 +1493,53 @@ func TestEvents(t *testing.T) {
 		require.Nil(t, events)
 	})
 }
+
+func TestMe(t *testing.T) {
```
omerfirmak, line 1497: `TestAddInvokeTransaction` maybe?

rpc/handlers.go
rianhughes: I added a set of error codes, following pathfinders approach.

rpc/handlers_test.go
kirugan: yes, sorry forgot to rename it

rpc/handlers.go
```diff
@@ -24,6 +29,10 @@ var (
 	ErrPageSizeTooBig           = &jsonrpc.Error{Code: 31, Message: "Requested page size is too big"}
 	ErrTooManyKeysInFilter      = &jsonrpc.Error{Code: 34, Message: "Too many keys provided in a filter"}
 	ErrInternal                 = &jsonrpc.Error{Code: jsonrpc.InternalError, Message: "Internal error"}
+	ErrCodeInvalidParams        = -32602
+	ErrCodeOutOfRangeFee        = -32603
+	ErrCodeOutOfRangeContract   = -32604
+	ErrCodeUnknownError         = -32605
```
kirugan, line 35: @rianhughes codes -32602 and -32603 already defined in `jsonrpc` package. Also looks like -32604 and -32605 are not part of [predefined errors from the spec](https://www.jsonrpc.org/specification#:~:text=NOT%20be%20included.-,5.1%20Error%20object,-When%20a%20rpc).

rianhughes: Nice catch, I've updated the code to use predefined jsonrpc codes.

kirugan (approved):

clients/gateway/integration_test.go
IronGauntlets, line 1: It is good that you have written some integration tests here. I have not used this type of convention before.

As far as I understand we should try to test as many cases as possible in unit tests before testing them in integration tests and we should not duplicate the same case in both types of tests. In my understanding integration tests are supposed to test the integration of "external collaborators" which have been passed through an interface. For example, a package which requires a DB, should use a memory/fakedb in unit tests to check that the business logic is correct, while the integration test should replace the fake/memory db with an actual instance and make sure the input and output from the DB are as expected.

Please correct me if I have misunderstood anything and if we can introduce an integration test in the rest of the codebase which would benefit us would be amazing.

In my opinion, the tests in this file should be moved to `rpc/handlers_test.go` because the purpose of the gateway package is to make requests to an external endpoint and handle those responses. While the purpose of rpc package is to listen for incoming requests. Also, the gateway mock should be used instead of sending a request to the actual endpoint. This is to avoid having a flaky test (in the event the API response changes).

This file should try to test the code contained within `gateway.go` with the help of `httptest` package to create fake server and request handlers.

clients/gateway/types.go
IronGauntlets, line 1: Let's move this to `clients/gateway/transaction.go` and rename `InvokeResponse` to `InvokeTxResponse`.

rpc/handlers.go
```diff
@@ -32,18 +37,26 @@ const (
...
 }
 
-func New(bcReader blockchain.Reader, synchronizer *sync.Synchronizer, n utils.Network, logger utils.Logger) *Handler {
+func New(
+	bcReader blockchain.Reader,
+	synchronizer *sync.Synchronizer,
+	n utils.Network,
+	gatewayClient gateway.Gateway,
+	logger utils.Logger,
+) *Handler {
```
IronGauntlets, line 53: Let's keep them on a single line. If the linter complains then do the following:

```go
func New(cReader blockchain.Reader, synchronizer *sync.Synchronizer, n utils.Network, gatewayClient gateway.Gateway,
 logger utils.Logger) *Handler {
```

Only move the minimum number of args required to the next line to satisfy `lll`.

clients/gateway/integration_test.go
IronGauntlets: Also, the code coverage for the patch is very low. Following the above should increase the reported coverage.

IronGauntlets (changes requested):

rpc/handlers_test.go
```diff
@@ -1492,3 +1494,77 @@ func TestEvents(t *testing.T) {
...
+		resp, handlerErr := handler.AddInvokeTransaction(invokeTx)
+
+		require.Nil(t, resp)
+		assert.Equal(t, handlerErr.Code, jsonrpc.InvalidParams)
+		assert.Equal(t, handlerErr.Message, "Transaction version '0x0' not supported. Supported versions: '0x1'")
+		assert.Equal(t, handlerErr.Data, nil)
+	})
+}
+
+func generateAddInvokeTx() *gateway.BroadcastedInvokeTxn {
```
rianhughes, line 1552: Is it okay to create this function that returns an invoke transaction? I find it makes the tests more readable

rpc/handlers.go
omerfirmak: `InvokeResponse` can stay in `gateway`. But structs like `BroadcastedTxn`, `InvokeTxnV1` & `BroadcastedInvokeTxn` should be part of the `rpc` package

for example; 

[getEstimateFee](https://github.com/starkware-libs/starknet-specs/blob/aaea417f193bbec87b59455128d4b09a06876c28/api/starknet_api_openrpc.json#L487-L504) has nothing to do with gateway, yet it still uses broadcasted tx types.

Gateway shouldnt depend on and RPC types. Duplicating the types to achieve this is not that big of an issue, we have duplicated types between `feeder` and `core` anyways.

rpc/handlers_test.go
```diff
@@ -1492,3 +1494,77 @@ func TestEvents(t *testing.T) {
...
+			Signature: []*felt.Felt{},
+			Nonce:     nonce,
+		},
+		Type: "INVOKE",
+		InvokeTxnV1: gateway.InvokeTxnV1{
+			SenderAddress: senderAddress,
+			Calldata:      []*felt.Felt{},
+		},
+	}
+}
```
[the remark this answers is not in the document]
omerfirmak: Of course :+1:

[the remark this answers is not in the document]
omerfirmak:
```suggestion
func generateAddInvokeTx() *gateway.BroadcastedInvokeTxn {
	maxFee := new(felt.Felt).SetUint64(1)
	nonce  := new(felt.Felt).SetUint64(1)
	senderAddress := new(felt.Felt).SetUint64(44)
```

try to use SetUint64 for tests where possible, that way you dont have to deal with a possible error

clients/gateway/integration_test.go
rianhughes: I followed your steps and: 
1. moved these tests to `rpc/handlers_test.go`, 
2. created a new file `clients/gateway_test.go` that tests `gateway.go` by using a fake server to handle requests (akin to feeder_test.go).

clients/gateway/gateway.go
```diff
@@ -0,0 +1,110 @@
...
+	"strings"
+	"time"
+
+	"github.com/NethermindEth/juno/utils"
+)
+
+//go:generate mockgen -destination=../../mocks/mock_gateway.go -package=mocks github.com/NethermindEth/juno/clients/gateway Gateway
+type Gateway interface {
+	AddInvokeTransaction(context.Context, *BroadcastedInvokeTxn) (*InvokeResponse, error)
+}
```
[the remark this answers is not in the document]
rianhughes: This has been moved to the rpc package, although it remains Public so that we can use go:generate. 
If you want to keep it private and generate mocks we can still do so, but the mocks would have to be generated by hand I think.

rpc/handlers.go
rianhughes: After recent discussions, structs related to addInvokeTx (eg `BroadcastedTxn`) will be moved to a new package `sequencer`, which `rpc` and `gateway` will import. This means rpc and gateway don't have to import each other.

clients/sequencer/transaction.go
```diff
@@ -0,0 +1,33 @@
+package sequencer
```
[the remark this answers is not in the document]
IronGauntlets: Where you unable to compile due to circular dependency?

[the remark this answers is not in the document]
rianhughes: I can compile okay.

clients/gateway/gateway.go
```diff
@@ -0,0 +1,187 @@
...
+	endpoint := c.url + "/add_transaction"
+
+	body, err := c.post(ctx, endpoint, txn)
+	if err != nil {
+		return nil, err
+	}
+	defer body.Close()
+
+	var resp struct {
+		TransactionHash string `json:"transaction_hash"`
```
omerfirmak, line 119:
```suggestion
		TransactionHash *felt.Felt `json:"transaction_hash"`
```

IronGauntlets:
closing in favour of #791

PR #776, opened 2023-05-23
Make fully synced nodes return false to `starknet_syncing`

jelilat:


jelilat:
@omerfirmak I sense false negatives here since our `highestBlockHeader` is not always accurate. @wojciechos what it means is that we can't rely on this endpoint to know if a node is fully synced or not.

omerfirmak:
> @omerfirmak I sense false negatives here since our `highestBlockHeader` is not always accurate. @wojciechos what it means is that we can't rely on this endpoint to know if a node is fully synced or not.

Yeah, but there is not much we can do about it. `syncing` is a "best effort" endpoint. As long as node "thinks" it is fully synced, it is free to report that.

PR #777, opened 2023-05-23
Fetch and store compiled classes for each sierra class

omerfirmak:


starknetdata/feeder/feeder.go
```diff
@@ -275,15 +275,20 @@ func (f *Feeder) Class(ctx context.Context, classHash *felt.Felt) (core.Class, e
...
 	case response.V1 != nil:
-		return adaptCairo1Class(response.V1)
+		compiledClass, cErr := f.client.CompiledClassDefinition(ctx, classHash)
+		if cErr != nil {
+			return nil, cErr
+		}
+
+		return adaptCairo1Class(response.V1, compiledClass)
 	case response.V0 != nil:
 		return adaptCairo0Class(response.V0)
```
kirugan, line 285: Just curios how definitions worked in V0 ?

IronGauntlets (approved):

starknetdata/feeder/feeder.go
omerfirmak: I am not sure if this answers the question but the main difference is V0 classes contain Cairo Assembly whereas V1 classes contain Sierra programs. Sierra programs are eventually compiled to cairo assembly and that is what the compiledClass in this PR represents.

core/class.go
rianhughes, line 1: So this 'Compiled' field represents the Sierra program? Why is it stored as a json.RawMessage?

kirugan (approved):

PR #778, opened 2023-05-23
Allow migrating the DB with hardcoded revisions

omerfirmak:


migration/migration.go
```diff
@@ -0,0 +1,68 @@
...
+	"errors"
+
+	"github.com/NethermindEth/juno/db"
+)
+
+type revision func(db.DB) error
+
+var revisions = []revision{
+	revision0000,
+}
```
omerfirmak, line 14: Revisions list contains a set of revisions that can be applied to a database. Currently there is only one revision which makes sure the initial DB the node is using is empty.

migration/migration.go
```diff
@@ -0,0 +1,68 @@
...
+		}
+
+		// revision returned with no errors, bump the version
+		if err = targetDB.Update(func(txn db.Transaction) error {
+			var versionBytes [8]byte
+			binary.BigEndian.PutUint64(versionBytes[:], i+1)
+			return txn.Set(db.SchemaVersion.Key(), versionBytes[:])
+		}); err != nil {
+			return err
+		}
```
omerfirmak, line 34: Schema version of the targetDB determines which set of revisions need to be applied to the database. After revision is successfully executed, which may update the database, schema version is incremented by 1.

For example;

Assume an empty DB, which has a schema version of `0`. So we start applying the revisions from the very first one in the list and increment the schema version as we go. After the loop exists we end up with a database which's schema version is `len(revisions)+1`.

After running that Juno version for a while, if the user updates it's Juno version which adds new revisions to the list, `MigrateIfNeeded` will skip the already applied revisions and only apply the new ones. It will be able to do this since the schema version it reads from the database will be non-zero and that is what we use to initialize the `i` loop variable.

migration/migration.go
omerfirmak: After making braking changes to the DB layout, we will have to add new revisions to this list.

omerfirmak (commented):
Let me briefly explain what is going on here;

migration/migration.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	return nil
+}
+
+func SchemaVersion(targetDB db.DB) (uint64, error) {
```
joshklop, line 40: This doesn't appear to be used outside the `migration` package. Could it be unexported?

migration/migration.go
```diff
@@ -0,0 +1,68 @@
...
+	err := txn.Get(db.SchemaVersion.Key(), func(bytes []byte) error {
+		version = binary.BigEndian.Uint64(bytes)
+		return nil
+	})
+	if err != nil && !errors.Is(err, db.ErrKeyNotFound) {
+		return 0, db.CloseAndWrapOnError(txn.Discard, err)
+	}
+
+	return version, db.CloseAndWrapOnError(txn.Discard, nil)
+}
```
joshklop, line 52:
```suggestion
	return version, txn.Get(db.SchemaVersion.Key(), func(bytes []byte) error {
		version = binary.BigEndian.Uint64(bytes)
		return nil
	})
}
```

migration/migration.go
```diff
@@ -0,0 +1,68 @@
...
+func revision0000(targetDB db.DB) error {
+	return targetDB.View(func(txn db.Transaction) error {
+		it, err := txn.NewIterator()
+		if err != nil {
+			return err
+		}
+
+		// not empty if valid
+		if it.Next() {
+			return db.CloseAndWrapOnError(it.Close, errors.New("Initial DB directory should be empty"))
```
joshklop, line 64:
```suggestion
			return db.CloseAndWrapOnError(it.Close, errors.New("Initial DB should be empty"))
```
I think the db directory can have other things in it, but the db itself must be empty.

migration/migration.go
```diff
@@ -0,0 +1,68 @@
...
+	"github.com/NethermindEth/juno/db"
+)
+
+type revision func(db.DB) error
+
+var revisions = []revision{
+	revision0000,
+}
+
+func MigrateIfNeeded(targetDB db.DB) error {
```
joshklop, line 16: Can we pass in the revisions as a parameter here? The associated test appears to be unnecessarily dependent on the `revisions` global variable.

joshklop (changes requested):

migration/migration.go
omerfirmak: done

migration/migration.go
omerfirmak: >The associated test appears to be unnecessarily dependent on the revisions global variable.

not really, it just makes sure that first call to `MigrateIfNeeded` increments the schema version and the second call doesnt. The actual value of what schema version ends up being is not relevant.

migration/migration.go
joshklop: In that case, could the tests be moved to a separate package?

omerfirmak: Yeap, it should've been in `migration_test` in the first place. My bad.

joshklop (approved):

migration/migration.go
```diff
@@ -0,0 +1,68 @@
...
+		return err
+	}
+
+	for i := version; i < uint64(len(revisions)); i++ {
+		if err = revisions[i](targetDB); err != nil {
+			return err
+		}
+
+		// revision returned with no errors, bump the version
+		if err = targetDB.Update(func(txn db.Transaction) error {
```
kirugan, line 28: What happens if the schema update is successful, but the process dies before updating the version here?

omerfirmak: Yeap, that is an edge case that we dont want to be left in. I will make sure everything happens within a transaction context.

omerfirmak: done

migration/migration.go
IronGauntlets: Please add the above explanation as comments

migration/revisions_test.go
IronGauntlets, line 1: Should we not follow our convention of using `*_pkg_test.go` here?

migration/migration.go
IronGauntlets: Please add this explanation as comments

migration/migration.go
```diff
@@ -0,0 +1,66 @@
...
+
+func MigrateIfNeeded(targetDB db.DB) error {
+	version, err := SchemaVersion(targetDB)
+	if err != nil {
+		return err
+	}
+
+	for i := version; i < uint64(len(revisions)); i++ {
+		if err = targetDB.Update(func(txn db.Transaction) error {
+			// revision returned with no errors, bump the version
```
IronGauntlets, line 24: Should this comment be moved to line 24?

IronGauntlets:
Am I correct in understanding that this will not work with existing DB since they don't have a schema version and for it to take effect everyone will have to sync from scratch, right?

omerfirmak:
> 

correct

omerfirmak:
> Am I correct in understanding that this will not work with existing DB since they don't have a schema version and for it to take effect everyone will have to sync from scratch, right?

correct

migration/revisions_test.go
omerfirmak: Yeap, my bad.

IronGauntlets (approved):
LGTM

PR #780, opened 2023-05-25
Add pending block support

omerfirmak:


joshklop (commented):
Is it possible to break this into two or three PRs? e.g., db+blockchain, sync, and RPC.

I think it will help speed up reviews.

sync/sync.go
```diff
@@ -192,18 +201,70 @@ func (s *Synchronizer) syncBlocks(syncCtx context.Context) {
...
+				pendingPollTicker.Stop()
+				pendingPollSem <- struct{}{}
 				return
 			default:
 				streamCtx, streamCancel = context.WithCancel(syncCtx)
 				nextHeight = s.nextHeight()
 				s.log.Warnw("Rolling back sync process", "height", nextHeight)
 			}
-		default:
+		case <-pendingPollTicker.C:
```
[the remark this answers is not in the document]
omerfirmak: >Could the pending block downloader be split into a separate Service that is run from

I thought of doing that too, but I don't think it warrant a new service. IMO, It can be thought of a part of sync process

omerfirmak:
> Is it possible to break this into two or three PRs? e.g., db+blockchain, sync, and RPC.
> 
> I think it will help speed up reviews.

I believe commits are atomic enough that one should be able to review this commit by commit. We can always drop couple of commits from the end to make the PR smaller, do let me know if you think that would be helpful.

joshklop:
> drop couple of commits from the end to make the PR smaller

I think that would be useful @omerfirmak. The last three commits seem to cover the RPC. If you wouldn't mind moving those to a separate branch, I'll do a more thorough review of the remaining commits.

sync/sync.go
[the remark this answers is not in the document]
joshklop: >  I don't think it warrants a new service

I am not fully convinced here. Can you share some of your reasoning? I'm mainly concerned about the concurrency complexity in the synchronizer, which a separate service should alleviate.

Kind of related: do we need to delay downloading the pending block until after we have reached the head? That seems to be a source of complexity.

blockchain/blockchain_test.go
```diff
@@ -486,8 +494,7 @@ func TestEvents(t *testing.T) {
 
 		require.NoError(t, filter.SetRangeEndBlockByHash(blockchain.EventFilterFrom,
 			utils.HexToFelt(t, "0x3b43b334f46b921938854ba85ffc890c1b1321f8fd69e7b2961b18b4260de14")))
-		require.NoError(t, filter.SetRangeEndBlockByHash(blockchain.EventFilterTo,
-			utils.HexToFelt(t, "0x3b43b334f46b921938854ba85ffc890c1b1321f8fd69e7b2961b18b4260de14")))
+		require.NoError(t, filter.SetRangeEndBlockByNumber(blockchain.EventFilterTo, 8))
```
IronGauntlets, line 497: What is this test saying? Why was the hash removed? 

If the previous test is still valid we should create a new test case

rpc/handlers_test.go
```diff
@@ -1492,7 +1501,7 @@ func TestEvents(t *testing.T) {
 	t.Run("filter with no keys", func(t *testing.T) {
 		var allEvents []*rpc.EmittedEvent
 		t.Run("get all events without pagination", func(t *testing.T) {
-			args.ToBlock = &rpc.BlockID{Latest: true}
+			args.ToBlock = &rpc.BlockID{Pending: true}
```
IronGauntlets, line 1504: We should create a new test case, instead of overwriting a previous one, right?

rpc/events.go
```diff
@@ -26,7 +26,7 @@ type EventsChunk struct {
 
 type EmittedEvent struct {
 	*Event
-	BlockNumber     uint64     `json:"block_number"`
-	BlockHash       *felt.Felt `json:"block_hash"`
+	BlockNumber     *uint64    `json:"block_number,omitempty"`
+	BlockHash       *felt.Felt `json:"block_hash,omitempty"`
```
IronGauntlets, line 30: We should have a test case that checks that the response JSON doesn't have these values set, right?

rpc/handlers.go
```diff
@@ -746,8 +746,13 @@ func (h *Handler) Events(args *EventsArg) (*EventsChunk, *jsonrpc.Error) {
...
 	for _, fEvent := range filteredEvents {
+
+		var blockNumber *uint64
+		if fEvent.BlockHash != nil {
+			blockNumber = &fEvent.BlockNumber
+		}
 		emittedEvents = append(emittedEvents, &EmittedEvent{
-			BlockNumber:     fEvent.BlockNumber,
+			BlockNumber:     blockNumber,
 			BlockHash:       fEvent.BlockHash,
```
IronGauntlets, line 756: Should this also be nil for pending event?

sync/sync.go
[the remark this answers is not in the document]
omerfirmak: > I don't think we need fectcherSem (both max go routine and channel with the same buffer size seem redundant)

the problem is `stream.Go` is blocking, and could mess up the polling interval. Now that we poll pending in a separate go routine, that is no longer an issue.

blockchain/pending.go
```diff
@@ -0,0 +1,12 @@
...
+import (
+	"github.com/NethermindEth/juno/core"
+	"github.com/NethermindEth/juno/core/felt"
+)
+
+type Pending struct {
+	Block       *core.Block
+	StateUpdate *core.StateUpdate
+	NewClasses  map[felt.Felt]core.Class
+}
```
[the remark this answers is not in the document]
omerfirmak: I thought "pending block" as a `blockchain` level concept. It isn't something that is core to the starknet protocol, but a stop-gap measure that sequencer implements to look more responsive.

I dont mind moving it to core package tho, your choice.

blockchain/blockchain.go
```diff
@@ -252,6 +254,10 @@ func (b *Blockchain) Store(block *core.Block, stateUpdate *core.StateUpdate, new
 			return err
 		}
 
+		if err := txn.Delete(db.Pending.Key()); err != nil {
+			return err
+		}
```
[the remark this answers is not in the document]
omerfirmak: > Or did you deliberately mean to delete the pending block as soon as a finalised block is stored regardless of whether they are the same or not?

Exactly.

`StorePending` checks that the parent hash of the new pending block matches the head and `Store` does the same thing for a finalized block. So we know that the existing pending block and the new finalized block that we are storing is for the same height. They dont have to be the same block.

sync/sync.go
```diff
@@ -176,8 +181,12 @@ func (s *Synchronizer) syncBlocks(syncCtx context.Context) {
 	}()
 
 	fetchers := stream.New().WithMaxGoroutines(runtime.NumCPU())
+	fetchersSem := make(chan struct{}, runtime.NumCPU())
```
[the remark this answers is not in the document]
omerfirmak: Yes

clients/feeder/testdata/mainnet/block/latest.json
```diff
@@ -1 +1,217 @@
...6319c5c6d71ae0a8480c", "0x0", "0x3", "0x10884171baf1914edc28d7afb619b40a4051cfae78a094a55d230f19e944a28", "0x15543c3708653cda9d418b4ccd3be11368e40636c10c44b18cfe756b6d88b29", "0x3", "0x6", "0x9", "0x10884171baf1914edc28d7afb619b40a4051cfae78a094a55d230f19e944a28", "0x641900", "0x0", "0x1", "0x53c91253bc9682c04929ca02ed00b3e423f6710d2ee7e0d5ebb06f3ecf368a8", "0x641900", "0x0", "0xcd5045903f7c2", "0x0"], "type": "INVOKE_FUNCTION"}, {"transaction_hash": "0x297a827c72720d106ad970d3878084ad1950f59ef6998e8ef9e5ede355c593c", "version": "0x1", "max_fee": "0x3a45f100ed520", "signature": ["0x641228e368a23d569d4b340b370f0a5549f7ba4f5a4f5d9d411197ef196f038", "0xa8b5ec731faebc7b6aaac782585660197ce0f627be607ff54dda0942127c05"], "nonce": "0x33", "sender_address": "0xb8fea23304cc5aba4871011d2b802f09c94904e7e286f908d6be7c1ee47cec", "calldata": ["0x2", "0x68f5c6a61780768455de69077e07e89787839bf8166decfbf92b645209c0fb8", "0x219209e083275171774dab1df80982e9df2096516f06319c5c6d71ae0a8480c", "0x0", "0x3", "0x41fd22b238fa21cfcf5dd45a8548974d8263b3a531a60388411c5e230f97023", "0x3276861cf5e05d6daf8f352cabb47df623eb10c383ab742fcc7abea94d5c5cc", "0x3", "0x9", "0xc", "0x41fd22b238fa21cfcf5dd45a8548974d8263b3a
```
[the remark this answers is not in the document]
omerfirmak: To make sync tests start polling pending block as well. If `latest` is too far in to the future, pending block is not polled.

rpc/block.go
```diff
@@ -74,10 +74,10 @@ func (b *BlockID) UnmarshalJSON(data []byte) error {
 
 // https://github.com/starkware-libs/starknet-specs/blob/a789ccc3432c57777beceaa53a34a7ae2f25fda0/api/starknet_api_openrpc.json#L1072
 type BlockHeader struct {
-	Hash             *felt.Felt `json:"block_hash"`
+	Hash             *felt.Felt `json:"block_hash,omitempty"`
 	ParentHash       *felt.Felt `json:"parent_hash"`
-	Number           uint64     `json:"block_number"`
-	NewRoot          *felt.Felt `json:"new_root"`
+	Number           *uint64    `json:"block_number,omitempty"`
```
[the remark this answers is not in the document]
omerfirmak: They dont

rpc/handlers.go
```diff
@@ -341,13 +378,17 @@ func (h *Handler) TransactionReceiptByHash(hash *felt.Felt) (*TransactionReceipt
 		contractAddress = nil
 	}
 
+	var receiptBlockNumber *uint64
+	if blockHash != nil {
+		receiptBlockNumber = &blockNumber
+	}
 	return &TransactionReceipt{
 		Status:          StatusAcceptedL2, // todo
```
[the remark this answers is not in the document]
omerfirmak: starknet 0.12 will remove Pending status from Transactions

rpc/events.go
omerfirmak: That would be testing the JSON marshaler, no? not our code

rpc/handlers.go
omerfirmak: the trick we have to do for BlockNumber is only needed with fEvent.BlockNumber is a value not a pointer. fEvent.BlockHash is a `*felt.Felt` already and will be nil for pending events.

sync/sync.go
```diff
@@ -192,18 +201,70 @@ func (s *Synchronizer) syncBlocks(syncCtx context.Context) {
...
+				<-fetchersSem
+				return cb
 			})
 			nextHeight++
 		}
 	}
 }
+
+func (s *Synchronizer) pollPending(ctx context.Context) error {
+	if s.HighestBlockHeader != nil {
```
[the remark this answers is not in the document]
omerfirmak: No reason, I think we can just return early if it is nil

blockchain/blockchain_test.go
```diff
@@ -415,15 +415,23 @@ func TestEvents(t *testing.T) {
...
+		}
 	}
 
 	t.Run("filter non-existent", func(t *testing.T) {
 		filter, err := chain.EventFilter(nil, nil)
 
 		t.Run("block number", func(t *testing.T) {
 			err = filter.SetRangeEndBlockByNumber(blockchain.EventFilterTo, uint64(44))
-			require.Error(t, err)
+			require.NoError(t, err)
```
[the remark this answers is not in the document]
omerfirmak: Yes, if it is greater than the latest height we clamp it to `latest+1`

https://github.com/NethermindEth/juno/pull/780/files#diff-299bd13c9d85e40bab3e1c50f08f697257ed323a0db67685f628d71e3c635518R99-R108

rpc/handlers_test.go
omerfirmak: testcase is "get all events without pagination", "all events used" to mean "0 to Latest", but now "0 to Pending" because we can possibly have a pending block and matching events in the pending block now.

blockchain/blockchain.go
```diff
@@ -665,3 +671,51 @@ func (b *Blockchain) EventFilter(from *felt.Felt, keys []*felt.Felt) (*EventFilt
...
+			expectedOldRoot = head.GlobalStateRoot
+		}
+
+		if !expectedParent.Equal(pending.Block.ParentHash) || !expectedOldRoot.Equal(pending.StateUpdate.OldRoot) {
+			return errors.New("pending block parent is not our local HEAD")
+		}
+
+		existingPending, err := pendingBlock(txn)
+		if err == nil && existingPending.Block.TransactionCount >= pending.Block.TransactionCount {
+			return nil // ignore the incoming pending if it has fewer transactions than the one we already have
```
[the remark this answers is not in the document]
omerfirmak: Because of concurrency and network delay, `StorePending` could be called with an older pending block compared to what we have in the DB.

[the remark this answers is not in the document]
omerfirmak: and once a transaction makes it into a pending block, it is never removed. So transaction count is a good indication of pending block age.

blockchain/blockchain_test.go
omerfirmak: same reason in https://github.com/NethermindEth/juno/pull/780#discussion_r1222866023 

we need to include the pending block in the range, cant do that with a hash

blockchain/pending.go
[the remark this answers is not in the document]
IronGauntlets: You are correct

rpc/handlers_test.go
```diff
@@ -244,6 +255,16 @@ func TestBlockWithTxHashes(t *testing.T) {
...
 	})
+
+	t.Run("blockID - pending", func(t *testing.T) {
+		mockReader.EXPECT().Pending().Return(blockchain.Pending{
+			Block: latestBlock,
+		}, nil)
+
+		block, rpcErr := handler.BlockWithTxHashes(&rpc.BlockID{Pending: true})
+		require.Nil(t, rpcErr)
+		checkLatestBlock(t, block)
```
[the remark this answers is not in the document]
omerfirmak: done

rpc/handlers_test.go
```diff
@@ -351,6 +372,20 @@ func TestBlockWithTxs(t *testing.T) {
 
 		checkLatestBlock(t, blockWithTxHashes, blockWithTxs)
 	})
+
+	t.Run("blockID - latest", func(t *testing.T) {
+		mockReader.EXPECT().Pending().Return(blockchain.Pending{
+			Block: latestBlock,
```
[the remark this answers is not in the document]
omerfirmak: all done

blockchain/blockchain_test.go
```diff
@@ -498,3 +498,67 @@ func TestEvents(t *testing.T) {
...
+		expectedPending := blockchain.Pending{
+			Block:       b,
+			StateUpdate: su,
+		}
+		require.NoError(t, chain.StorePending(&expectedPending))
+
+		gotPending, pErr := chain.Pending()
+		require.NoError(t, pErr)
+		assert.Equal(t, expectedPending, gotPending)
+	})
```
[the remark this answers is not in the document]
IronGauntlets: I was just confused based on the expected pending block that you were using and I didn't have context for what you intended.

The earlier comments have made it clearer.

sync/sync.go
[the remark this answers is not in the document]
IronGauntlets: > the messages in the ticker channel keep building up and we end up bombarding the gateway with request for the pending block.

We should be able to consume extra ticks before calling `fetchAndStorePending`.

sync/sync.go
```diff
@@ -191,6 +197,9 @@ func (s *Synchronizer) syncBlocks(syncCtx context.Context) {
 	startingHeight := nextHeight
 	s.StartingBlockNumber = &startingHeight
 
+	pendingSem := make(chan struct{}, 1)
```
IronGauntlets, line 200: Why not manage this in `pollPendinding`?

sync/sync.go
```diff
@@ -208,7 +218,8 @@ func (s *Synchronizer) syncBlocks(syncCtx context.Context) {
 		default:
 			curHeight, curStreamCtx, curCancel := nextHeight, streamCtx, streamCancel
 			fetchers.Go(func() stream.Callback {
-				return s.fetcherTask(curStreamCtx, curHeight, verifiers, curCancel)
+				cb := s.fetcherTask(curStreamCtx, curHeight, verifiers, curCancel)
+				return cb
```
IronGauntlets, line 222: Seems like this change was not necessary

omerfirmak: Yeap, it must have slipped in somehow

sync/sync.go
```diff
@@ -231,3 +242,69 @@ func (s *Synchronizer) revertHead(forkBlock *core.Block) {
...
+			case sem <- struct{}{}:
+				go func() {
+					err := s.fetchAndStorePending(ctx)
+					if err != nil {
+						s.log.Debugw("Error while trying to poll pending block", "err", err)
+					}
+					<-sem
+				}()
+			default:
+				continue
```
IronGauntlets, line 268: I don't think we need to explicitly have `continue` here

sync/sync.go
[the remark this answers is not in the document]
omerfirmak: Hmm, is reading len(channel) amount of messages is enough? Sounds like it could be racey to me. Semaphore helps with graceful shutdown as well. So I am still voting in favor of it:)

sync/sync.go
```diff
@@ -199,6 +208,7 @@ func (s *Synchronizer) syncBlocks(syncCtx context.Context) {
 				streamCancel()
 				fetchers.Wait()
 				verifiers.Wait()
+				pendingSem <- struct{}{}
```
IronGauntlets, line 211: I don't understand why we are sending pendingSem when we streamCtx is done

omerfirmak: this https://github.com/NethermindEth/juno/pull/780#discussion_r1227946183

to ensure there is no pending poll routine in flight

rpc/handlers_test.go
```diff
@@ -162,6 +163,18 @@ func TestBlockTransactionCount(t *testing.T) {
 		require.Nil(t, rpcErr)
 		assert.Equal(t, expectedCount, count)
 	})
+
+	t.Run("blockID - pending", func(t *testing.T) {
+		latestBlock.Hash = nil
+		latestBlock.GlobalStateRoot = nil
+		mockReader.EXPECT().Pending().Return(blockchain.Pending{
+			Block: latestBlock,
```
IronGauntlets, line 171: I don't see you setting `latestBlock.Number = nil`, am I missing something?

omerfirmak: `latestBlock.Number` is a `uint64`, not a pointer. So any value you can set it to doesnt really signal if it is a pending or canonical block.

rpc/events.go
IronGauntlets: Possibly, I just thought that we should have test case which shows that our pending event doesn't have `BlockNumber` and `BlockHash`

cmd/juno/juno.go
```diff
@@ -25,21 +26,23 @@ const greeting = `
...
+	colourF              = "colour"
+	pendingPollIntervalF = "pending-poll-interval"
+
+	defaultConfig              = ""
+	defaultRPCPort             = uint16(6060)
+	defaultDBPath              = ""
+	defaultEthNode             = ""
+	defaultPprof               = false
+	defaultColour              = true
+	defaultPendingPollInterval = time.Duration(0)
```
IronGauntlets, line 45: I think the default should be more than 0. Maybe 1s? Otherwise, the ticker will go crazy, if the user forgets to assign the poll interval.

IronGauntlets: Nevermind. Just saw that it is disabled by default.

blockchain/blockchain.go
```diff
@@ -227,6 +244,25 @@ func (b *Blockchain) Receipt(hash *felt.Felt) (*core.TransactionReceipt, *felt.F
...
 		var err error
 		receipt, blockHash, blockNumber, err = receiptByHash(txn, hash)
+
+		// not found in the canonical blocks, try pending
+		if errors.Is(err, db.ErrKeyNotFound) {
+			var pending Pending
+			pending, err = pendingBlock(txn)
+			if err != nil {
+				return err
+			}
```
IronGauntlets, line 255: @joshklop's comment applies here too

blockchain/blockchain_test.go
IronGauntlets: Unresolving because we discussed adding another test case

rpc/handlers_test.go
IronGauntlets: No need to add an extra case for this as your explanation makes sense

rpc/events.go
IronGauntlets: I think there should be some checks on the single event returned in and of it self so that we can have something similar to `checkLatestBlock`. We should have at least one case for non-pending event and one for pending event (without block number and block hash).

IronGauntlets (commented):
Please fix @joshklop comments and add some test cases for events. I have unresolved convo for those cases.

blockchain/blockchain.go
```diff
@@ -213,6 +213,23 @@ func (b *Blockchain) TransactionByHash(hash *felt.Felt) (core.Transaction, error
 	return transaction, b.database.View(func(txn db.Transaction) error {
 		var err error
 		transaction, err = transactionByHash(txn, hash)
+
+		// not found in the canonical blocks, try pending
+		if errors.Is(err, db.ErrKeyNotFound) {
+			var pending Pending
+			pending, err = pendingBlock(txn)
```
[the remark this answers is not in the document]
omerfirmak: Done, thanks once again!

omerfirmak:
> Please fix @joshklop comments and add some test cases for events. I have unresolved convo for those cases.

All done, please take another look @joshklop @IronGauntlets

IronGauntlets (approved):

joshklop (approved):

PR #787, opened 2023-06-03
Make `core.State` calculate commitment lazily

joshklop:
Takes advantage of @omerfirmak's [foundational PR](https://github.com/NethermindEth/juno/commit/11e0d64375ce479cc0d3af409c49c44375656656) that made the trie commitment calculation lazy. 

This shouldn't improve or degrade performance, since the `blockchain` is still calculating the commitment after every block. For now, this PR is primarily a refactor that enables us to defer commitment calculation for multiple blocks if we decide to do so.

The first few commits refactor some code to make the last commit (the one that actually makes `state.Root()` lazy) easier to understand. 

The commit messages have more details.

### DIFF
--- a/core/contract.go
+++ b/core/contract.go
@@ -160,7 +160,7 @@ func (c *Contract) UpdateStorage(diff []StorageDiff, cb OnValueChanged) error {
 	}
 
 	// update contract storage root in the database
-	rootKeyDBKey := db.ContractRootKey.Key(c.Address.Marshal())
+	rootKeyDBKey := db.ContractStorage.Key(c.Address.Marshal())
 	if rootKey := cStorage.RootKey(); rootKey != nil {
 		rootKeyBytes, err := rootKey.MarshalBinary()
 		if err != nil {
@@ -215,7 +215,7 @@ func storage(addr *felt.Felt, txn db.Transaction) (*trie.Trie, error) {
 	addrBytes := addr.Marshal()
 	var contractRootKey *bitset.BitSet
 
-	if err := txn.Get(db.ContractRootKey.Key(addrBytes), func(val []byte) error {
+	if err := txn.Get(db.ContractStorage.Key(addrBytes), func(val []byte) error {
 		contractRootKey = new(bitset.BitSet)
 		return contractRootKey.UnmarshalBinary(val)
 	}); err != nil && !errors.Is(err, db.ErrKeyNotFound) {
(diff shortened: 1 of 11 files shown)


core/trie/trie_test.go
```diff
@@ -298,6 +299,41 @@ func TestMaxTrieHeight(t *testing.T) {
...
+			// Recalculating after no changes should give same result.
+			got, err = tempTrie.Root()
+			require.NoError(t, err)
+			assert.Equal(t, want, got)
+		})
+
+		return nil
+	}))
+}
+
```
omerfirmak: can you explain what is the goal of this test?

core/contract.go
```diff
@@ -160,7 +160,7 @@ func (c *Contract) UpdateStorage(diff []StorageDiff, cb OnValueChanged) error {
 	}
 
 	// update contract storage root in the database
-	rootKeyDBKey := db.ContractRootKey.Key(c.Address.Marshal())
+	rootKeyDBKey := db.ContractStorage.Key(c.Address.Marshal())
```
omerfirmak: This is prone to corrupting the DB.

Contract storage tries use `db.ContractStorage.Key(c.Address.Marshal())` as it's prefix

https://github.com/NethermindEth/juno/blob/5670d77fe656667141be980916f908db17c832a8/core/contract.go#L226

and the rest of the key is essentially arbitrary;

https://github.com/NethermindEth/juno/blob/5670d77fe656667141be980916f908db17c832a8/core/trie/transaction_storage.go#L42-L52

if `key.WriteTo` generates the same byte array as `c.Address.Marshal()`, which is not likely but possible, could cause us to misinterpret the root key as a storage value.

omerfirmak (changes requested):

core/contract.go
joshklop: I got a little mixed up while thinking about this, so correct me if I'm wrong:

The result of `key.WriteTo` is appended to `t.prefix`, which includes `c.Address.Marshal()`. Thus, the `db.ContractStorage+c.Address` key is unused (since `key.WriteTo` cannot be empty) and we are free to use it for the root key.

core/trie/trie_test.go
joshklop: That `trie.Root` matches the spec for an empty trie and after one insertion. Maybe this is just re-testing the node hashing code though...

joshklop:
Now that a database migration is needed, the PR will get quite large. I have submitted #816 and will submit separate PRs for the remaining changes.

PR #788, opened 2023-06-04
Fix RPC not serializing CompiledClassHash for DeclareV2 txns

omerfirmak:


rpc/handlers.go
```diff
@@ -206,18 +206,15 @@ func adaptInvokeTransaction(t *core.InvokeTransaction) *Transaction {
...
-		MaxFee:        t.MaxFee,
-		Version:       t.Version,
-		Signature:     &sig,
-		Nonce:         t.Nonce,
-		ClassHash:     t.ClassHash,
-		SenderAddress: t.SenderAddress,
-	}
-
-	if t.Version.Equal(new(felt.Felt).SetUint64(2)) {
-		txn.CompiledClassHash = nil // todo: add when we have support for Declare V2
```
IronGauntlets, line 220: Can you please give some more context as to why removing this if statement solves the issue?

IronGauntlets (approved):

kirugan (approved):

rpc/handlers.go
omerfirmak: If statement wasn't really needed in the first place. If the transaction version is 2, `t.CompiledClassHash` will be non-nil, else it will be nil. So we can just assign it to `txn.CompiledClassHash` regardless

PR #789, opened 2023-06-04
Use less explicit conversions in feeder adaption layer where applicable 

omerfirmak:
Apparently when both types have exact same fields, there is no need to explicitly enumerate each field when doing conversions

### DIFF
--- a/blockchain/event_filter.go
+++ b/blockchain/event_filter.go
@@ -81,7 +81,7 @@ func (c *ContinuationToken) FromString(str string) error {
 }
 
 type FilteredEvent struct {
-	*core.Event
+	core.Event
 	BlockNumber     uint64
 	BlockHash       *felt.Felt
 	TransactionHash *felt.Felt
--- a/clients/feeder/transaction.go
+++ b/clients/feeder/transaction.go
@@ -67,7 +67,7 @@ type BuiltinInstanceCounter struct {
 
 type TransactionReceipt struct {
 	ActualFee          *felt.Felt          `json:"actual_fee"`
-	Events             []*Event            `json:"events"`
+	Events             []Event             `json:"events"`
 	ExecutionResources *ExecutionResources `json:"execution_resources"`
 	L1ToL2Message      *L1ToL2Message      `json:"l1_to_l2_consumed_message"`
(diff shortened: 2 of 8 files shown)


IronGauntlets:
> Apparently when both types have exact same fields, there is no need to explicitly enumerate each field when doing conversions

If the types are identical then I would say remove the feeder types and just use core types everywhere instead of casting them. However, if it turns out to be confusing them having separate types are ok.

blockchain/event_filter.go
```diff
@@ -81,7 +81,7 @@ func (c *ContinuationToken) FromString(str string) error {
 }
 
 type FilteredEvent struct {
-	*core.Event
+	core.Event
```
IronGauntlets, line 84: What is the reason for removing this reference?

clients/feeder/transaction.go
```diff
@@ -67,7 +67,7 @@ type BuiltinInstanceCounter struct {
 
 type TransactionReceipt struct {
 	ActualFee          *felt.Felt          `json:"actual_fee"`
-	Events             []*Event            `json:"events"`
+	Events             []Event             `json:"events"`
```
IronGauntlets, line 70: Same as above

core/transaction.go
```diff
@@ -42,17 +42,17 @@ type ExecutionResources struct {
...
+	Bitwise    uint64
+	Output     uint64
+	Ecsda      uint64
+	EcOp       uint64
 }
 
 type TransactionReceipt struct {
 	Fee                *felt.Felt
-	Events             []*Event
+	Events             []Event
```
IronGauntlets, line 55: Same as above

rpc/events.go
```diff
@@ -20,12 +20,12 @@ type ResultPageRequest struct {
 }
 
 type EventsChunk struct {
-	Events            []*EmittedEvent `json:"events"`
-	ContinuationToken string          `json:"continuation_token,omitempty"`
+	Events            []EmittedEvent `json:"events"`
```
IronGauntlets, line 23: Same as above

omerfirmak:
Closing this in favor of using https://github.com/jinzhu/copier in a separate PR

PR #791, opened 2023-06-06
Add starknet_AddInvokeTransaction

IronGauntlets:
Please don't look at the intermediate commits. Before merging they will be squashed.


rpc/handlers.go
```diff
@@ -731,3 +742,48 @@ func setEventFilterRange(filter *blockchain.EventFilter, fromID, toID *BlockID,
...
+			Message: err.Error(),
+		}
+	}
+
+	return invokeRes, nil
+}
+
+// getAddInvokeTxCode returns the relevant Code for a given addInvokeTx error
+func getAddInvokeTxCode(err error) int {
+	switch {
```
rianhughes, line 775: @IronGauntlets, I thought you decided it would be best to remove this function, and just return jsonrpc.InvalidParms  (in line 756)?

IronGauntlets: I changed my mind because `h.gatewayClient.AddInvokeTransaction()` can return an error which wraps an HTTP error (see gateway.go:110) which would be incorrect to return as invalid params.

rianhughes (approved):

PR #797, opened 2023-06-07
Add core.BlockNumber type

kirugan:


omerfirmak:
let's change the Number field in header to be of type core.BlockNumber as well?
or, we can make ToBytes function just a standalone helper function without creating the BlockNumber type

omerfirmak:
>or, we can make ToBytes function just a standalone helper function without creating the BlockNumber type

sounds easier to me

IronGauntlets:
Any particular reason why we can't use `cbor` as we do for other structs?

IronGauntlets:
> cbor encodes uints to variable length byte streams. So if it doesnt need the full 8 bytes to represent a uint64, the resulting byte array will be smaller. That will mess up our sorted buckets.

I see, thanks!

@kirugan Instead of having `ToBytes()`, do you think it makes sense to have `MarshalBinary()` and `UnmarshalBinary()`

IronGauntlets:
There are places in `blockchain.go` where we read block number from db, I think that should be replaced with UnmarshalBinary() so that we don't have to import `encoding/binary`

kirugan:
> Instead of having ToBytes(), do you think it makes sense to have MarshalBinary() and UnmarshalBinary()

@IronGauntlets as far as I know, we don't use binary encoding interfaces (`encoding.BinaryMarshaler`, `encoding.BinaryUnmarshaler`) much. Compliance with the interface will bring little benefit IMO, but the ability to use a function call in one line affects readability:
`db.Key(blockNumber.ToBytes())`

Perhaps `ToBytes()` is not the best name for this method and it's better to call it `Bytes()` or something similar. What do you think?

blockchain/blockchain.go
```diff
@@ -507,12 +534,16 @@ func (b *Blockchain) SanityCheckNewHeight(block *core.Block, stateUpdate *core.S
...
-	Number uint64
+	Number core.BlockNumber
 	Index  uint64
 }
 
 func (t *txAndReceiptDBKey) MarshalBinary() []byte {
-	return binary.BigEndian.AppendUint64(binary.BigEndian.AppendUint64([]byte{}, t.Number), t.Index)
+	numBytes, err := t.Number.MarshalBinary()
+	if err != nil {
+		panic(err)
```
kirugan, line 544: two comments:
- `BlockNumber.MarshalBinary()` always returns nil for an error (I added it because it's golang interface signature). So what do you guys think if I add `MustMarshalBinary(...) []byte` to `BlockNumber` ?
_ Do you think I should change return signature in `txAndReceiptDBKey.MarshalBinary()` also?

omerfirmak: >Do you think I should change return signature in txAndReceiptDBKey.MarshalBinary() also?

I think this

omerfirmak:
This overall LGTM. I know I asked for core.BlockNumber to be used everywhere but I am now wondering if this is too big of a changeset just to remove 3 lines being duplicated in a handful of places.

We could achieve the same thing by making the MarshalBinary a standalone helper (something like `MarshalBlockNumber(blockNumber uint64) ([]byte, error)` ) and use that wherever needed.

WDYT @kirugan ?

kirugan:
late comment about this PR: @omerfirmak yep, agree that it turned out a lot of changes just to encapsulate mashaling logic into specific type

Issue #685, opened 2023-03-28
Implement starknet_addDeployAccountTransaction


jelilat:
Implement the`starknet_addDeployAccountTransaction` endpoint to follow the Starknet RPC specification defined [here](https://github.com/starkware-libs/starknet-specs/blob/a789ccc3432c57777beceaa53a34a7ae2f25fda0/api/starknet_write_api.json#L74)

rianhughes:
https://github.com/NethermindEth/juno/pull/794

PR #798, opened 2023-06-07
Add deploy account tx

rianhughes:


rianhughes:
https://github.com/NethermindEth/juno/issues/685

clients/gateway/gateway.go
```diff
@@ -85,12 +93,12 @@ func (c *Client) AddInvokeTransaction(txn json.RawMessage) (json.RawMessage, err
 	}
 	defer body.Close()
 
-	invokeRes, readErr := io.ReadAll(body)
+	deployRes, readErr := io.ReadAll(body)
```
joshklop, line 96: Since this is used for invoke and deploy, can we use a generic name like `res`?

clients/gateway/gateway.go
```diff
@@ -77,6 +77,14 @@ func NewClient(gatewayURL string, log utils.SimpleLogger) *Client {
 }
 
 func (c *Client) AddInvokeTransaction(txn json.RawMessage) (json.RawMessage, error) {
+	return c.addTransaction(txn)
+}
+
+func (c *Client) AddDeployAccountTransaction(txn json.RawMessage) (json.RawMessage, error) {
+	return c.addTransaction(txn)
```
joshklop, line 84: Since the gateway only exposes the `add_transaction` endpoint, can we just have one `AddTransaction` function?

joshklop (changes requested):
Nice work

clients/gateway/gateway.go
```diff
@@ -57,8 +57,8 @@ func newTestServer() *httptest.Server {
 			return
 		}
 
-		txHash := new(felt.Felt).SetBytes([]byte("random"))
-		resp := fmt.Sprintf("{\"code\": \"TRANSACTION_RECEIVED\", \"transaction_hash\": %q}", txHash.String())
+		hash := new(felt.Felt).SetBytes([]byte("random"))
+		resp := fmt.Sprintf("{\"code\": \"TRANSACTION_RECEIVED\", \"transaction_hash\": %q, \"address\": %q}", hash.String(), hash.String())
```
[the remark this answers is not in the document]
rianhughes: Sure! A successful `addInvokeTransaction` method only needs to return a `transaction hash`, whereas the `addDeployAccountTransaction` method requires a` transaction hash` and `contract address`. I just return both here (even for the `addInvokeTransaction`) because we're just using this for test purposes, and it keeps the code simple. The gateway also returns a "code" of type string, which isn't apart of the spec, but I included it anyway because we're mocking the gateways response.

clients/gateway/gateway.go
rianhughes: Yes, good point!

clients/gateway/gateway.go
rianhughes: Yes, it would require the methods to have the same input and output parameters, but it would make the code neater. I'll update the code to reflect this.

rpc/handlers.go
```diff
@@ -781,3 +781,28 @@ func getAddInvokeTxCode(err error) int {
...
+		if strings.Contains(err.Error(), "Class hash not found") {
+			ErrClassHashNotFound.Data = err.Error()
+			return nil, ErrClassHashNotFound
+		}
+		return nil, jsonrpc.Err(getAddInvokeTxCode(err), err.Error())
+	}
+
+	deployResp := new(DeployAccountTxResponse)
+	err = json.Unmarshal(resp, deployResp)
+	if err != nil {
```
omerfirmak, line 803:
```suggestion
	if err = json.Unmarshal(resp, deployResp); err != nil {
```

omerfirmak (approved):

rpc/handlers.go
```diff
@@ -781,3 +781,27 @@ func getAddInvokeTxCode(err error) int {
 		return jsonrpc.InternalError
 	}
 }
+
+// AddDeployAccountTransaction relays an deploy account transaction to the gateway.
+//
+// It follows the specification defined here:
+// https://github.com/starkware-libs/starknet-specs/blob/a789ccc3432c57777beceaa53a34a7ae2f25fda0/api/starknet_write_api.json#L74
+// Note: No checks are performed on the incoming request since we rely on the gateway to perform sanity checks.
+// As this handler is just as a proxy. Any error returned by the gateway is returned to the user as a jsonrpc error.
```
joshklop, line 790:
```suggestion
// It follows the [specification].
// Note: This handler is a proxy. No checks are performed on the incoming request since we rely on the gateway to perform sanity checks. Any error returned by the gateway is returned to the user as a jsonrpc error.
//
// [specification]: https://github.com/starkware-libs/starknet-specs/blob/a789ccc3432c57777beceaa53a34a7ae2f25fda0/api/starknet_write_api.json#L74
```
We can use [godoc formatting](https://go.dev/doc/comment) here.

joshklop (approved):

rpc/handlers.go
rianhughes: Unfortunately, this causes linting errors (on the [spec]:.. line). Also the current version is also consistent with the rest of the doc comments, so I'd prefer to keep it as is for now, if that's okay.

PR #799, opened 2023-06-07
Add declare tx

rianhughes:


rianhughes:
https://github.com/NethermindEth/juno/issues/684

rpc/handlers.go
```diff
@@ -775,9 +779,80 @@ func getAddInvokeTxCode(err error) int {
...
 	}
 }
+
+// AddDeclareTransaction relays a declare transaction to the gateway.
+//
+// It follows the specification defined here:
+// https://github.com/starkware-libs/starknet-specs/blob/a789ccc3432c57777beceaa53a34a7ae2f25fda0/api/starknet_write_api.json#L39
+// Note: The gateway expects the sierra_program to be gzip compressed and 64-base encoded. We perform this operation,
+// and then relay the transaction to the gateway.
+func (h *Handler) AddDeclareTransaction(declareTx *json.RawMessage) (*DeclareTxResponse, *jsonrpc.Error) {
```
omerfirmak, line 795:
```suggestion
func (h *Handler) AddDeclareTransaction(declareTx map[string]any) (*DeclareTxResponse, *jsonrpc.Error) {
```

Let the jsonrpc package do the unmarshaling for you

rpc/handlers.go
```diff
@@ -775,9 +779,80 @@ func getAddInvokeTxCode(err error) int {
...
+			return nil, jsonrpc.Err(jsonrpc.InternalError, errIn.Error())
+		}
+		if errIn = gzipWriter.Close(); errIn != nil {
+			return nil, jsonrpc.Err(jsonrpc.InternalError, errIn.Error())
+		}
+
+		gwSierraProg := base64.StdEncoding.EncodeToString(gzipBuffer.Bytes())
+		if errIn != nil {
+			return nil, jsonrpc.Err(jsonrpc.InternalError, errIn.Error())
+		}
```
omerfirmak, line 830: We do this in https://github.com/NethermindEth/juno/blob/main/starknetdata/feeder/feeder.go#L349-L360 as well. Let's extract that into a function and reuse that

rpc/handlers.go
omerfirmak: might make testing harder tho, so up to you

rpc/handlers.go
rianhughes: I've created a Gzip64Encode() function and placed it in utils.

rpc/handlers.go
rianhughes: Thanks for the suggestion. I'm happy to leave it as is because` AddTransaction()` takes` json.RawMessage`, and if I accepted a `map[string]any`, I think I would have to cast it to `json.RawMessage` anyway to use in `AddTransaction()`

rpc/handlers.go
```diff
@@ -775,9 +776,70 @@ func getAddInvokeTxCode(err error) int {
...
 		return jsonrpc.InvalidParams
 	default:
 		return jsonrpc.InternalError
 	}
 }
+
+// AddDeclareTransaction relays a declare transaction to the gateway.
+//
+// It follows the specification defined here:
+// https://github.com/starkware-libs/starknet-specs/blob/a789ccc3432c57777beceaa53a34a7ae2f25fda0/api/starknet_write_api.json#L39
```
joshklop, line 789: We can use [godoc formatting](https://go.dev/doc/comment#links) here.

rpc/handlers.go
```diff
@@ -775,9 +776,70 @@ func getAddInvokeTxCode(err error) int {
...
+	}
+
+	if version, ok := request["version"]; ok && version == "0x2" {
+		contractClass, ok := request["contract_class"].(map[string]interface{})
+		if !ok {
+			return nil, jsonrpc.Err(jsonrpc.InvalidParams, errors.New("{'contract_class': ['Missing data for required field.']}").Error())
+		}
+		sierraProg, ok := contractClass["sierra_program"]
+		if !ok {
+			return nil, jsonrpc.Err(jsonrpc.InvalidParams, errors.New("{'sierra_program': ['Missing data for required field.']}").Error())
```
joshklop, line 806: I think we can pass in the string directly. No need for `errors.New(...).Error()`. Same for the error a few lines up.

joshklop: Also: is there a reason why the strings are formatted like json objects instead of normal English sentence? It may be the right thing to do, I'm just wondering if there's a specific reason :)

rpc/handlers_test.go
```diff
@@ -1543,10 +1543,89 @@ func TestAddInvokeTransaction(t *testing.T) {
...
+		mockGateway.EXPECT().AddTransaction(tx).Return(nil, errors.New("{'type': ['Missing data for required field.']}"))
+
+		resp, handlerErr := handler.AddDeclareTransaction(&tx)
+		require.Nil(t, resp)
+		assert.Equal(t, handlerErr.Code, expectedErr.Code)
+		assert.Equal(t, handlerErr.Message, expectedErr.Message)
+		assert.Equal(t, handlerErr.Data, expectedErr.Data)
+	})
+
+	t.Run("ErrInvalidContractClass", func(t *testing.T) {
```
joshklop, line 1620:
```suggestion
	t.Run("invalid contract class", func(t *testing.T) {
```

joshklop (changes requested):

rpc/handlers.go
rianhughes: The godoc format causes linting issues, and the current version is also consistent with the rest of the documentation comments in this file. So I think it makes sense to leave it as is for now.

rpc/handlers.go
rianhughes: > Also: is there a reason why the strings are formatted like json objects instead of normal English sentence? It may be the right thing to do, I'm just wondering if there's a specific reason :)

Great question, I use this json-format just because it's what the sequencer would return in this scenario.

rpc/handlers_test.go
```diff
@@ -1543,10 +1543,89 @@ func TestAddInvokeTransaction(t *testing.T) {
...
+			Data:    errors.New("{'type': ['Missing data for required field.']}").Error(),
+		}
+
+		mockGateway.EXPECT().AddTransaction(tx).Return(nil, errors.New("{'type': ['Missing data for required field.']}"))
+
+		resp, handlerErr := handler.AddDeclareTransaction(&tx)
+		require.Nil(t, resp)
+		assert.Equal(t, handlerErr.Code, expectedErr.Code)
+		assert.Equal(t, handlerErr.Message, expectedErr.Message)
+		assert.Equal(t, handlerErr.Data, expectedErr.Data)
```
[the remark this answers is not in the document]
rianhughes: Yes we can!

joshklop (approved):

rpc/handlers.go
```diff
@@ -30,6 +30,7 @@ var (
 	ErrInvalidContinuationToken = &jsonrpc.Error{Code: 33, Message: "Invalid continuation token"}
 	ErrPageSizeTooBig           = &jsonrpc.Error{Code: 31, Message: "Requested page size is too big"}
 	ErrTooManyKeysInFilter      = &jsonrpc.Error{Code: 34, Message: "Too many keys provided in a filter"}
+	ErrInvlaidContractClass     = &jsonrpc.Error{Code: 50, Message: "Invalid contract class"}
```
IronGauntlets, line 33: You are missing [`CLASS_ALREADY_DECLARED`](https://github.com/starkware-libs/starknet-specs/blob/aaea417f193bbec87b59455128d4b09a06876c28/api/starknet_write_api.json#L72)

IronGauntlets (approved):

Issue #796, opened 2023-06-07
Export `jsonrpc.rpcErr` and make use of it in `rpc` package


omerfirmak:
https://github.com/NethermindEth/juno/blob/7ca28f11b102744511831bbe6d92d83bf36f5b92/jsonrpc/server.go#L46-L59

in places like;

https://github.com/NethermindEth/juno/blob/7ca28f11b102744511831bbe6d92d83bf36f5b92/rpc/handlers.go#L755-L769

rianhughes:
Currently the error messages in [juno/rpc/handlers.go] provides useful information to the caller that would be lost if we used `rpcErr`. 

For example, if a user submitted a transaction with an incorrect transaction version, rpcErr wouldn't tell them version was incorrect, only that some parameter they provided was incorrect. It would also return the same error if they used the correct version, but forgot some other field, etc.

omerfirmak:
We can use Data field to return an arbitrary error message

rianhughes:
https://github.com/NethermindEth/juno/pull/802

PR #802, opened 2023-06-08
Export RPCErr

rianhughes:


rianhughes:
https://github.com/NethermindEth/juno/issues/796

jsonrpc/server.go
```diff
@@ -43,7 +43,7 @@ type Error struct {
 	Data    any    `json:"data,omitempty"`
 }
 
-func rpcErr(code int, data any) *Error {
+func RPCErr(code int, data any) *Error {
```
omerfirmak, line 46:
```suggestion
func Err(code int, data any) *Error {
```

omerfirmak (approved):

jsonrpc/server.go
kirugan, line 46: maybe rename it just to `Err(...)` ? it's already clear what is doing outside of package json**rpc**.Err(...)

kirugan (approved):

IronGauntlets (approved):

kirugan:
@rianhughes could you please check that in case we pass an error without calling `.Error()` it was actually sent to a client?
Looks like in this case `jsonrpc.Err(jsonrpc.InternalError, err)` data field is still empty (here is a quick test -> https://go.dev/play/p/N4RvNHAOt0e)

rianhughes:
> @rianhughes could you please check that in case we pass an error without calling `.Error()` it was actually sent to a client? Looks like in this case `jsonrpc.Err(jsonrpc.InternalError, err)` data field is still empty (here is a quick test -> https://go.dev/play/p/N4RvNHAOt0e)

The `handler.AddInvokeTransaction` method seems to be returning the error in the data field okay, without calling `.Error()` in `jsonrpc.Err(jsonrpc.InternalError, err)`. I agree that the playground example doesn't seem to (not sure why), but our code seems to be working as intended.

omerfirmak:
> The handler.AddInvokeTransaction method seems to be returning the error in the data field okay

does it marshal to JSON fine?

rianhughes:
> > The handler.AddInvokeTransaction method seems to be returning the error in the data field okay
> 
> does it marshal to JSON fine?
`
jsonrpc.Err(jsonrpc.InternalError, err)` builds the `jsonrpc.Error` structs as you would expect, ie with the data field populated with the err (without needing to call err.Error())

 see the commit I made here https://github.com/rianhughes/juno/tree/ErrMarhsalCheck, where I test the jsonrpcError has the expected data field.

rianhughes:
> > The handler.AddInvokeTransaction method seems to be returning the error in the data field okay
> 
> does it marshal to JSON fine?

@kirugan nicely caught that it doesn't marshal to JSON okay, and we need to pass in err.Error() instead

PR #806, opened 2023-06-09
Update `starknet_getEvents` to conform spec v0.3.0

### LINKED ISSUE #804: Update getEvents to RPC Spec 3.0
https://github.com/starkware-libs/starknet-specs/pull/55

omerfirmak:
Fixes #804

### DIFF
--- a/blockchain/blockchain_test.go
+++ b/blockchain/blockchain_test.go
@@ -448,7 +448,7 @@ func TestEvents(t *testing.T) {
 
 	from := utils.HexToFelt(t, "0x49d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7")
 	t.Run("filter with no keys", func(t *testing.T) {
-		filter, err := chain.EventFilter(from, [][]*felt.Felt{{}, {}, {}})
+		filter, err := chain.EventFilter(from, [][]felt.Felt{{}, {}, {}})
 		require.NoError(t, err)
 
 		require.NoError(t, filter.SetRangeEndBlockByNumber(blockchain.EventFilterFrom, 0))
@@ -489,7 +489,7 @@ func TestEvents(t *testing.T) {
 
 	t.Run("filter with keys", func(t *testing.T) {
 		key := utils.HexToFelt(t, "0x3774b0545aabb37c45c1eddc6a7dae57de498aae6d5e3589e362d4b4323a533")
-		filter, err := chain.EventFilter(from, [][]*felt.Felt{{key}})
+		filter, err := chain.EventFilter(from, [][]felt.Felt{{*key}})
 		require.NoError(t, err)
 
 		require.NoError(t, filter.SetRangeEndBlockByHash(blockchain.EventFilterFrom,
@@ -507,9 +507,9 @@ func TestEvents(t *testing.T) {
 	})
 
 	t.Run("filter with not matching keys", func(t *testing.T) {
-		filter, err := chain.EventFilter(from, [][]*felt.Felt{
-			{utils.HexToFelt(t, "0x3774b0545aabb37c45c1eddc6a7dae57de498aae6d5e3589e362d4b4323a533")},
-			{utils.HexToFelt(t, "0xDEADBEEF")},
+		filter, err := chain.EventFilter(from, [][]felt.Felt{
+			{*utils.HexToFelt(t, "0x3774b0545aabb37c45c1eddc6a7dae57de498aae6d5e3589e362d4b4323a533")},
+			{*utils.HexToFelt(t, "0xDEADBEEF")},
 		})
 		require.NoError(t, err)
 		require.NoError(t, filter.SetRangeEndBlockByNumber(blockchain.EventFilterFrom, 0))
--- a/blockchain/blockchain_test.go
+++ b/blockchain/blockchain_test.go
@@ -448,7 +448,7 @@ func TestEvents(t *testing.T) {
 
 	from := utils.HexToFelt(t, "0x49d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7")
 	t.Run("filter with no keys", func(t *testing.T) {
-		filter, err := chain.EventFilter(from, []*felt.Felt{})
+		filter, err := chain.EventFilter(from, [][]*felt.Felt{{}, {}, {}})
 		require.NoError(t, err)
 
 		require.NoError(t, filter.SetRangeEndBlockByNumber(blockchain.EventFilterFrom, 0))
@@ -489,7 +489,7 @@ func TestEvents(t *testing.T) {
 
 	t.Run("filter with keys", func(t *testing.T) {
 		key := utils.HexToFelt(t, "0x3774b0545aabb37c45c1eddc6a7dae57de498aae6d5e3589e362d4b4323a533")
-		filter, err := chain.EventFilter(from, []*felt.Felt{key})
+		filter, err := chain.EventFilter(from, [][]*felt.Felt{{key}})
 		require.NoError(t, err)
 
 		require.NoError(t, filter.SetRangeEndBlockByHash(blockchain.EventFilterFrom,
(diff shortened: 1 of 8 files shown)


blockchain/event_filter.go
```diff
@@ -196,3 +197,29 @@ func (e *EventFilter) appendBlockEvents(matchedEventsSofar []*FilteredEvent, hea
...
+	/// ((event.Keys[0] == "V1" OR event.Keys[0] == "V2") AND (event.Keys[2] == "V3")).
+	//
+	// Essentially
+	// for each event.Keys[i], (len(e.keys[i]) == 0 OR event.Keys[i] is in e.keys[i]) should hold
+	for index, eventKey := range eventKeys {
+		// empty filter keys means match all
+		if index >= len(keysMap) || len(keysMap[index]) == 0 {
+			break
+		}
+		if found := keysMap[index][*eventKey]; !found {
```
kirugan, line 219: Do we check that request contains valid event keys? i.e. not nil

omerfirmak: `jsonrpc` package does that before our handler is called

kirugan: Hmm strange I got panic with this curl request:
```
curl --location --request POST 'http://localhost:6060' \
--data-raw '{
	"jsonrpc":"2.0",
	"method":"starknet_getEvents",
	"params": [{
        "chunk_size": 2,
        "keys": [
            [null, null]
        ]
    }],
	"id":2
}'
```

joshklop (commented):
I think the tests need to have pointers turned into values as well (per the last commit).

omerfirmak:
> I think the tests need to have pointers turned into values as well (per the last commit).

I believe I did that wherever necessary. Can you point out which ones I missed?

kirugan (approved):

joshklop:
Also, should we update [RegisterMethod](https://github.com/NethermindEth/juno/blob/7e7fe238c3a0936be12225514e0bf2e62e51feb0/jsonrpc/server.go#L118) to check for pointers? Could be done in a separate PR as well.

omerfirmak:
> > > I think the tests need to have pointers turned into values as well (per the last commit).
> > 
> > 
> > I believe I did that wherever necessary. Can you point out which ones I missed?
> 
> I think there are some in [server_test.go](https://github.com/NethermindEth/juno/blob/7e7fe238c3a0936be12225514e0bf2e62e51feb0/jsonrpc/server_test.go#L74).


> Also, should we update [RegisterMethod](https://github.com/NethermindEth/juno/blob/7e7fe238c3a0936be12225514e0bf2e62e51feb0/jsonrpc/server.go#L118) to check for pointers? Could be done in a separate PR as well.


We are not deprecating pointer arguments, they are still useful for optional arguments.

omerfirmak:
Rebased

blockchain/blockchain_test.go
```diff
@@ -448,7 +448,7 @@ func TestEvents(t *testing.T) {
 
 	from := utils.HexToFelt(t, "0x49d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7")
 	t.Run("filter with no keys", func(t *testing.T) {
-		filter, err := chain.EventFilter(from, []*felt.Felt{})
+		filter, err := chain.EventFilter(from, [][]felt.Felt{{}, {}, {}})
```
IronGauntlets, line 451: Why is the size of slice 3? Can you not just have `[][]felt.Felt{{}}` to get all the keys?

blockchain/event_filter.go
```diff
@@ -102,10 +103,7 @@ func (e *EventFilter) Events(cToken *ContinuationToken, chunkSize uint64) ([]*Fi
 		}
 	}
 
-	filterKeysMap := make(map[felt.Felt]bool, len(e.keys))
-	for _, key := range e.keys {
-		filterKeysMap[*key] = true
-	}
+	filterKeysMaps := makeKeysMap(e.keys)
```
IronGauntlets, line 106:
```suggestion
	filterKeysMaps := makeKeysMaps(e.keys)
```

blockchain/event_filter.go
```diff
@@ -222,3 +222,42 @@ func (e *EventFilter) appendBlockEvents(matchedEventsSofar []*FilteredEvent, hea
...
+	for index, keys := range filterKeys {
+		kMap := make(map[felt.Felt]bool, len(keys))
+		for _, key := range keys {
+			kMap[key] = true
+		}
+		filterKeysMaps[index] = kMap
+	}
+
+	return filterKeysMaps
+}
```
IronGauntlets, line 263: If you are just interested in that something exists in a map it is better to use `struct{}` instead of `bool` (AFAIK).

```suggestion
func makeKeysMap(filterKeys [][]felt.Felt) []map[felt.Felt]struct{} {
	filterKeysMaps := make([]map[felt.Felt]struct{}, len(filterKeys))
	for index, keys := range filterKeys {
		kMap := make(map[felt.Felt]struct{}, len(keys))
		for _, key := range keys {
			kMap[key] = struct{}{}
		}
		filterKeysMaps[index] = kMap
	}

	return filterKeysMaps
}
```

blockchain/event_filter.go
```diff
@@ -102,10 +103,7 @@ func (e *EventFilter) Events(cToken *ContinuationToken, chunkSize uint64) ([]*Fi
...
 	}
 
-	filterKeysMap := make(map[felt.Felt]bool, len(e.keys))
-	for _, key := range e.keys {
-		filterKeysMap[*key] = true
-	}
+	filterKeysMaps := makeKeysMap(e.keys)
 
 	curBlock := e.fromBlock
 	// skip the blocks that we previously processed for this request
```
IronGauntlets, line 109: I couldn't select line 116 for some reason.

I am just curious why the `if` statement on line 117 (and 131) is not `curBlock <= latest` instead of `curBlock != latest + 1`? Is it not possible that the user can send `e.toBlock`and `e.FromBlock` which is larger than our chain height?

IronGauntlets:
IIUC, the purpose of the second commit is to ensure the args passed to `param` field in JSON-RPC request are not pointers. So that nobody can pass `null` which we would unmarshal to `nil` that can lead to a panic?

rpc/events.go
```diff
@@ -8,10 +8,10 @@ type EventsArg struct {
 }
 
 type EventFilter struct {
-	FromBlock *BlockID     `json:"from_block"`
-	ToBlock   *BlockID     `json:"to_block"`
-	Address   *felt.Felt   `json:"address"`
-	Keys      []*felt.Felt `json:"keys"`
+	FromBlock *BlockID      `json:"from_block"`
+	ToBlock   *BlockID      `json:"to_block"`
+	Address   *felt.Felt    `json:"address"`
```
IronGauntlets, line 13: Why are they, not value type? Surely someone could pass null for `FromBlock`, `ToBlock` and `Address` and there is no `omitempty` tag so they will always be present in EventFilter struct.

IronGauntlets: Also, `BlockID.Hash` should not be a pointer either

omerfirmak: They are optional, so they should be pointers

omerfirmak:
> IIUC, the purpose of the second commit is to ensure the args passed to `param` field in JSON-RPC request are not pointers. So that nobody can pass `null` which we would unmarshal to `nil` that can lead to a panic?

Yes, anything that is not optional should be a value type.

blockchain/event_filter.go
omerfirmak: Yes but we update them to be `latest+1` at max

https://github.com/NethermindEth/juno/pull/806/files#diff-299bd13c9d85e40bab3e1c50f08f697257ed323a0db67685f628d71e3c635518L94-L103

IronGauntlets: That is only on `e.toBlock`, there are no checks on `e.fromBlock` to be much larger than the latest.

Also, I think a check should ensure that `from !>= to`.

rpc/events.go
IronGauntlets: According to the spec, nothing is a [required](https://github.com/starkware-libs/starknet-specs/blob/aaea417f193bbec87b59455128d4b09a06876c28/api/starknet_api_openrpc.json#L889) for EventFilter. So shouldn't EventFilter look like the following:

```go
type EventFilter struct {
	FromBlock *BlockID      `json:"from_block"`
	ToBlock   *BlockID      `json:"to_block"`
	Address   *felt.Felt    `json:"address"`
    Keys      *[][]felt.Felt `json:"keys"`
}
```

IronGauntlets: I just realised that since slices are reference types there is no need to have `*` in front of `[][]felt.Felt`. I became confused between this and when we want to differentiate between the zero value of the type vs the missing value.

IronGauntlets (approved):

PR #808, opened 2023-06-11
Add websocket support

joshklop:


joshklop:
This PR uses gorilla/websocket, which has been archived. See #848 for revised implementation using nhooyr.io/websocket.

PR #809, opened 2023-06-12
VM integration

omerfirmak:


adapters/feeder2core/feeder2core.go
```diff
@@ -143,7 +144,7 @@ func AdaptTransaction(transaction *feeder.Transaction) (core.Transaction, error)
 		return AdaptDeclareTransaction(transaction), nil
 	case "DEPLOY":
 		return AdaptDeployTransaction(transaction), nil
-	case "INVOKE_FUNCTION":
+	case "INVOKE_FUNCTION", "INVOKE":
```
joshklop, line 147: nit: should this be in the previous commit?

rpc/handlers.go
```diff
@@ -1038,3 +1038,48 @@ func (h *Handler) Call(call FunctionCall, id BlockID) ([]*felt.Felt, *jsonrpc.Er
...
+
+func (h *Handler) EstimateFee(broadcastedTxns []BroadcastedTransaction, id BlockID) ([]*FeeEstimate, *jsonrpc.Error) {
+	var estimates []*FeeEstimate
+	state, closer, err := h.stateByBlockID(&id)
+	if err != nil {
+		return nil, ErrBlockNotFound
+	}
+	defer func() {
+		if closeErr := closer(); closeErr != nil {
+			h.log.Warnw("Failed to close state in starknet_call", "err", closeErr)
```
joshklop, line 1050: nit: should this be an `Errorw`?

rpc/handlers.go
```diff
@@ -1038,3 +1038,48 @@ func (h *Handler) Call(call FunctionCall, id BlockID) ([]*felt.Felt, *jsonrpc.Er
...
+		}
+	}()
+
+	header, err := h.blockHeaderByID(&id)
+	if err != nil {
+		return nil, ErrBlockNotFound
+	}
+
+	if header.GasPrice == nil {
```
joshklop, line 1059: Should we just return `0` in this case?

rpc/handlers.go
```diff
@@ -1038,3 +1038,48 @@ func (h *Handler) Call(call FunctionCall, id BlockID) ([]*felt.Felt, *jsonrpc.Er
...
+		return nil, jsonrpc.Err(jsonrpc.InternalError, "no gas price history for the given block")
+	}
+
+	for idx := range broadcastedTxns {
+		txn, declaredClass, err := adaptBroadcastedTransaction(&broadcastedTxns[idx])
+		if err != nil {
+			return nil, jsonrpc.Err(jsonrpc.InvalidParams, err.Error())
+		}
+
+		gasConsumed, err := vm.Execute(txn, declaredClass, header.Number, header.Timestamp, state, h.network)
```
joshklop, line 1069: From the spec:

> a sequence of transactions to estimate, running each transaction on the state resulting from applying all the previous ones

Do we update the state somewhere? Since `callContext` uses a `state.Reader`, it doesn't seem that we do. Based on the [blockifier implementation](https://github.com/starkware-libs/blockifier/blob/a3a133af7eaa07197caa90766db11edacf8f46f6/crates/blockifier/src/transaction/transactions.rs#L30-L48), it seems that the state is only modified on their side but never passed to us?

joshklop (changes requested):

clients/feeder/transaction.go
```diff
@@ -6,20 +6,20 @@ import (
...
-	CallData            []*felt.Felt `json:"calldata"`
-	EntryPointSelector  *felt.Felt   `json:"entry_point_selector"`
-	Nonce               *felt.Felt   `json:"nonce"`
-	CompiledClassHash   *felt.Felt   `json:"compiled_class_hash"`
+	Hash                *felt.Felt    `json:"transaction_hash,omitempty" copier:"must,nopanic"`
+	Version             *felt.Felt    `json:"version,omitempty"`
+	ContractAddress     *felt.Felt    `json:"contract_address,omitempty"`
+	ContractAddressSalt *felt.Felt    `json:"contract_address_salt,omitempty"`
+	ClassHash           *felt.Felt    `json:"class_hash,omitempty"`
+	ConstructorCallData *[]*felt.Felt `json:"constructor_calldata,omitempty"`
```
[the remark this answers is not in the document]
omerfirmak: Yeap, we have to differentiate between an empty `constructor_calldata` and `constructor_calldata` field not being present at all

rpc/handlers.go
omerfirmak: > Do we update the state somewhere

Nope, nice catch.

vm/rust/src/lib.rs
```diff
@@ -98,6 +109,106 @@ pub extern "C" fn cairoVMCall(
...
+        let txn = Transaction::from_api(sn_api_txn.clone(), contract_class, None);
+        if txn.is_err() {
+            report_error(reader_handle, txn.unwrap_err().to_string().as_str());
+            return;
+        }
+
+        let res = match txn.unwrap() {
+            Transaction::AccountTransaction(t) => t.execute(&mut state, &block_context),
+            Transaction::L1HandlerTransaction(t) => t.execute(&mut state, &block_context),
+        };
```
IronGauntlets, line 187: Why are we matching for `L1HandlerTransaction`?

IronGauntlets (changes requested):
There are no unit tests. Please add at least some in `rpc/handler_test.go`.

As far as I understand the main issue with testing is not being able to create a state with the blocks that we currently have in the test data, right?

vm/rust/src/lib.rs
```diff
@@ -98,6 +111,106 @@ pub extern "C" fn cairoVMCall(
...
+                    report_error(reader_handle, "missing declared class".to_string().as_str());
+                    return;
+                }
+                let class_json_str = classes.remove(0);
+
+                let mut maybe_cc = contract_class_from_json_str(class_json_str.get());
+                if declare.version() == TransactionVersion(2u32.into()) && maybe_cc.is_err() {
+                    // class json could be sierra
+                    maybe_cc = contract_class_from_sierra_json(class_json_str.get())
+                };
```
IronGauntlets, line 169: Can you please explain why both `contract_class_from_json_str` and `contract_class_from_sierra_json` are required?

AFAIU both function above return `ContractClass` while `contract_class_from_json_str` can return either `ContractClassV0/V1` and we set the class json for `v1` to be cairo0 json while `v2` sets seirra json, so is `let v1_class = ContractClassV1::try_from_json_string(raw_json);` (line in `contract_class_from_json_str`) not sufficient to created `ContractClassV1`?

If not then we should not try to create `ContractClassV1` in `contract_class_from_json_str`

vm/transaction.go
```diff
@@ -0,0 +1,54 @@
...
+		txnMap["Declare"] = map[string]any{
+			"V" + clearQueryBit(t.Version).Text(felt.Base10): t,
+		}
+	default:
+		return nil, errors.New("unsupported txn type")
+	}
+	return json.Marshal(txnMap)
+}
+
+func clearQueryBit(v *felt.Felt) *felt.Felt {
```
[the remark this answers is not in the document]
omerfirmak: It's essentially a measure against replay attacks.

User creates a transaction, signs it and sends it to a node in a RPC request. Without the query bit set, that transaction is perfectly valid. Full node, or anyone in between the user and the full node, can submit that transcation to the sequencer to be included in a block. 

But with the query bit set (which changes the transaction hash and therefore the signature), sequencer knows that this transaction was signed for an RPC request, not to be included in a block.

cc @IronGauntlets

rpc/handlers.go
omerfirmak: Nope, `header.GasPrice == nil` is a limitation of Juno. I added the `GasPrice` field to the block header without a migration, so some blocks don't have that field populated.

rpc/handlers.go
```diff
@@ -1062,3 +1062,57 @@ func (h *Handler) TransactionStatus(hash felt.Felt) (Status, *jsonrpc.Error) {
 
 	return status, nil
 }
+
+func (h *Handler) EstimateFee(broadcastedTxns []BroadcastedTransaction, id BlockID) ([]FeeEstimate, *jsonrpc.Error) {
```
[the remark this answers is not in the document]
omerfirmak: As we have talked before, that information is encoded in the calldata. How it encoded in there is not necessarily standardized. AFAICS, pathinder doesnt return CONTRACT_NOT_FOUND either.

rpc/handlers.go
omerfirmak: Done

rpc/handlers.go
```diff
@@ -1062,3 +1062,57 @@ func (h *Handler) TransactionStatus(hash felt.Felt) (Status, *jsonrpc.Error) {
...
+	}()
+
+	header, err := h.blockHeaderByID(&id)
+	if err != nil {
+		return nil, ErrBlockNotFound
+	}
+
+	if header.GasPrice == nil {
+		return nil, jsonrpc.Err(jsonrpc.InternalError, "no gas price history for the given block")
+	}
```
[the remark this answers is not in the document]
omerfirmak: Done

rpc/handlers.go
```diff
@@ -1062,3 +1062,57 @@ func (h *Handler) TransactionStatus(hash felt.Felt) (Status, *jsonrpc.Error) {
...
+			return nil, jsonrpc.Err(jsonrpc.InvalidParams, aErr.Error())
+		}
+
+		txns = append(txns, txn)
+		if declaredClass != nil {
+			classes = append(classes, declaredClass)
+		}
+	}
+
+	gasesConsumed, err := vm.Execute(txns, classes, header.Number, header.Timestamp, header.SequencerAddress, state, h.network)
```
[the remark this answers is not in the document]
omerfirmak: Done

vm/vm.go
```diff
@@ -7,10 +7,14 @@ package vm
 //					uintptr_t readerHandle, unsigned long long block_number, unsigned long long block_timestamp,
 //					char* chain_id);
 //
+// extern void cairoVMExecute(char* txn_json, char* class_json, uintptr_t readerHandle, unsigned long long block_number,
```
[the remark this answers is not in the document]
omerfirmak: Done

vm/transaction.go
```diff
@@ -0,0 +1,56 @@
...
+	versionWithoutQueryBit := new(felt.Felt).Set(v)
+	// if versionWithoutQueryBit >= queryBit
+	if versionWithoutQueryBit.Cmp(queryVersion) != -1 {
+		versionWithoutQueryBit.Sub(versionWithoutQueryBit, queryVersion)
+	}
```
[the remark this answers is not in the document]
omerfirmak: Query bit is the 128th bit of Version field, which may be set by the user

see https://github.com/NethermindEth/juno/pull/809#discussion_r1254045318

vm/rust/Cargo.toml
```diff
@@ -7,12 +7,12 @@ edition = "2021"
 
 [dependencies]
 serde = "1.0.160"
-serde_json = "1.0.96"
-blockifier = {git = "https://github.com/starkware-libs/blockifier", rev = "d571467"}
+serde_json = { version = "1.0.96", features = ["raw_value"] }
```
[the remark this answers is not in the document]
omerfirmak: It allows us to access `serde_json::value::RawValue` type

vm/rust/src/lib.rs
omerfirmak: `cairoVMExecute` is a general purpose txn execution entrypoint. so, even tho we dont need it for estimateFee, it supports L1 handlers as well.

vm/rust/src/lib.rs
omerfirmak: `contract_class_from_json_str` deals with compiled classes only. It is used in the JunoStateReader as well.

a declared class may be a legacy casm class (ie ContractClassV0), or a Sierra class that we first need to compile to a casm class (ie ContractClassV1)

for the first case, a just reused `contract_class_from_json_str` and the second case is implemented in `contract_class_from_sierra_json`. Which parses the sierra class and compiles it to a `ContractClassV1`

rpc/transaction.go
```diff
@@ -66,21 +66,23 @@ func (t *TransactionType) UnmarshalJSON(data []byte) error {
...
+//
+//nolint:lll
 type Transaction struct {
 	Hash                *felt.Felt      `json:"transaction_hash,omitempty"`
-	Type                TransactionType `json:"type"`
-	Version             *felt.Felt      `json:"version,omitempty"`
-	Nonce               *felt.Felt      `json:"nonce,omitempty"`
-	MaxFee              *felt.Felt      `json:"max_fee,omitempty"`
+	Type                TransactionType `json:"type" validate:"required"`
+	Version             *felt.Felt      `json:"version,omitempty" validate:"required"`
```
IronGauntlets, line 74: Does it make sense to have `omitempty` tag for json while we also have `validate: "required"`? Can we not just have the json tag here?

omerfirmak: Validate tags are for incoming Transactions, json tags are used for marshaling only (ie for marshaling responses to the RPC requests we get). So they are not related.

vm/transaction.go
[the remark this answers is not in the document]
joshklop: Nice! Thanks.

rpc/handlers.go
joshklop: Will it be possible to do a migration in the future that removes this limitation? It would be unfortunate to maintain this forever.

vm/rust/Cargo.toml
```diff
@@ -7,12 +7,14 @@ edition = "2021"
...
+blockifier = {git = "https://github.com/starkware-libs/blockifier", rev = "v0.1.0-rc0"}
 starknet_api = { git = "https://github.com/starkware-libs/starknet-api", rev = "a4c78ff" }
 cairo-vm = "0.6"
-cairo-lang-casm = "2.0.0-rc4"
-cairo-lang-starknet = "2.0.0-rc4"
+cairo-lang-casm = "2.0.0"
+cairo-lang-starknet = "2.0.0"
+
+starknet = { rev = "ada5db0", git = "https://github.com/xJonathanLEI/starknet-rs" }
```
IronGauntlets, line 17: Should we use release for this as well instead of a commit?

vm/rust/src/lib.rs
```diff
@@ -80,9 +93,9 @@ pub extern "C" fn cairoVMCall(
     let mut state = CachedState::new(reader);
     let mut resources = ExecutionResources::default();
     let mut context = EntryPointExecutionContext::new(
-        build_block_context(chain_id_str, block_number, block_timestamp),
+        build_block_context(chain_id_str, block_number, block_timestamp, StarkFelt::default()),
         AccountTransactionContext::default(),
-        MAX_STEPS_PER_TX as u32,
+        4_000_000,
```
IronGauntlets, line 98: Is this an arbitrary number?

rpc/handlers.go
```diff
@@ -1062,3 +1062,67 @@ func (h *Handler) TransactionStatus(hash felt.Felt) (Status, *jsonrpc.Error) {
...
+		rpcErr := *ErrContractError
+		rpcErr.Data = err.Error()
+		return nil, &rpcErr
+	}
+
+	for _, gasConsumed := range gasesConsumed {
+		estimates = append(estimates, FeeEstimate{
+			GasConsumed: gasConsumed,
+			GasPrice:    header.GasPrice,
+			OverallFee:  new(felt.Felt).Mul(gasConsumed, header.GasPrice),
```
IronGauntlets, line 1123: Pending blocks don't have `GasPrice`, so we should set an arbitrary number for pending block

IronGauntlets: Please ignore this, feeder gateway returns the price but rpc doesn't so that's why I was confused.

utils/compression.go
```diff
@@ -17,3 +18,23 @@ func Gzip64Encode(data []byte) (string, error) {
 	}
 	return base64.StdEncoding.EncodeToString(compressedBuffer.Bytes()), nil
 }
+
+func Gzip64Decode(data string) ([]byte, error) {
+	decodedProgram, err := base64.StdEncoding.DecodeString(data)
```
joshklop, line 23: nit: remove references to `program` in this function.

vm/rust/src/lib.rs
omerfirmak: got it from the main branch

https://github.com/starkware-libs/blockifier/blob/main/crates/blockifier/src/abi/constants.rs#L29

vm/rust/Cargo.toml
omerfirmak: Yeap, will do

wojciechos (approved):

rpc/handlers.go
IronGauntlets: We will remove it in the next release.

joshklop (approved):

IronGauntlets (approved):

PR #810, opened 2023-06-12
Increase MaxIdleConnsPerHost for feeder TestClient

IronGauntlets:
On macOS tests often fail with the following error:

"Get "http://127.0.0.1:xxxx/get_{feeder gateway method}?{arg}={value}": dial tcp 127.0.0.1:xxxx: connect: can't assign requested address"

This error make running local tests, in quick succession, difficult because we have to wait for the OS to release ports. Sometimes the sync tests will hang because sync process will keep requesting if there was some error. This problem is further exacerbated by having parallel tests.

Increasing test client's idle conns allow for large concurrent requests to be made from a single test client.

### DIFF
--- a/clients/feeder/feeder.go
+++ b/clients/feeder/feeder.go
@@ -68,7 +68,23 @@ type closeTestClient func()
 // NewTestClient returns a client and a function to close a test server.
 func NewTestClient(network utils.Network) (*Client, closeTestClient) {
 	srv := newTestServer(network)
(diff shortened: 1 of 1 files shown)


wojciechos (approved):

joshklop (approved):
Thanks for the description. Can that be added as a comment above line 72 or 74?

IronGauntlets:
Done

Issue #783, opened 2023-06-01
Start collecting metrics during runtime


omerfirmak:
It could be all sorts of metrics

- JSON RPC request rates, response times, failures
- Gateway response times
- Block interval
- Block import times, verification times, failures
- Db read/write counts, delays

see https://prometheus.io/docs/guides/go-application/

omerfirmak:
and eventually display them in grafana

weiihann:
hey @omerfirmak, would like to contribute to this feature. Can you assign it to me?

rianhughes:
https://github.com/NethermindEth/juno/pull/814

joshklop:
Closing this optimistically since we have some metrics now. If we need a tracking issue for specific metrics, we can create one. But I think adding metrics as-needed is probably the best approach.

PR #812, opened 2023-06-13
Monitor

rianhughes:


rianhughes:
https://github.com/NethermindEth/juno/issues/783

PR #814, opened 2023-06-13
Create monitor

rianhughes:


kirugan:
I think in general such things should be moved to middleware without required changes in any handler

rianhughes:
> > I think in general such things should be moved to middleware without required changes in any handler
> 
> Yeap, should be fairly straightforward to collect all metrics related to RPC in `jsonrpc` package



> > I think in general such things should be moved to middleware without required changes in any handler
> 
> Yeap, should be fairly straightforward to collect all metrics related to RPC in `jsonrpc` package

So, something like jsonrpc this in https://github.com/NethermindEth/juno/pull/814/files?

chichi13:
Has it been abandoned? No metrics added?

IronGauntlets:
> Has it been abandoned? No metrics added?
No @chichi13, please have a look at #919

PR #816, opened 2023-06-14
Remove unnecessary db.ContractRootKey bucket

joshklop:
The commit messages describe all of the changes.

### DIFF
--- a/migration/migration.go
+++ b/migration/migration.go
@@ -1,6 +1,7 @@
 package migration
 
 import (
+	"bytes"
 	"encoding/binary"
 	"errors"
 
@@ -13,6 +14,7 @@ type revision func(transaction db.Transaction) error
 // After making breaking changes to the DB layout, add new revisions to this list.
 var revisions = []revision{
 	revision0000,
+	relocateContractStorageRootKeys,
 }
 
 func MigrateIfNeeded(targetDB db.DB) error {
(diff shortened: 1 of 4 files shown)


migration/migration.go
```diff
@@ -82,3 +85,48 @@ func revision0000(txn db.Transaction) error {
...
+	}
+
+	// Iterate over all entries in the old bucket, moving each one to its new location.
+	oldPrefix := db.ContractRootKey.Key()
+	it.SetOptions(&pebble.IterOptions{
+		LowerBound: oldPrefix,
+		UpperBound: (db.ContractRootKey + 1).Key(), // The prefix that succeeds db.ContractRootKey.
+	})
+	for it.First(); it.Valid(); it.Next() {
+		oldKey := it.Key()
```
omerfirmak, line 108: How about something like;

https://github.com/NethermindEth/juno/blob/b969344250d71aebe8b54ce72fd993ab1fb9c694/blockchain/blockchain.go#L402-L406

that way, we dont need to introduce `SetOptions` or `First`

joshklop: Good idea.

migration/migration.go
```diff
@@ -82,3 +84,49 @@ func revision0000(txn db.Transaction) error {
...
+		return err
+	}
+
+	// Iterate over all entries in the old bucket, moving each one to its new location.
+	oldPrefix := db.Unused.Key()
+	for it.Seek(oldPrefix); it.Valid(); it.Next() {
+		// Stop iterating once we're out of the old bucket.
+		if !bytes.Equal(it.Key()[:len(oldPrefix)], oldPrefix) {
+			break
+		}
```
omerfirmak, line 106: I think CutPrefix ensures this already, so maybe not really needed.

migration/migration_pkg_test.go
```diff
@@ -25,3 +26,54 @@ func TestRevision0000(t *testing.T) {
...
+
+	// Each root-key entry should have been moved to its new location
+	// and the old entry should not exist.
+	for i := 0; i < numberOfContracts; i++ {
+		require.True(t, it.Valid(), "iterator is invalid, should be valid")
+
+		exampleBytes := new(felt.Felt).SetUint64(uint64(i)).Bytes()
+
+		// New entry exists.
+		found := it.Seek(db.ContractStorage.Key(exampleBytes[:]))
```
omerfirmak, line 66: I think Seek can return true even if the key that you are seeking to doesnt exist. So we need to verify that it.Key() is what we expect.

or we can straight up use txn.Get instead of an iterator.

migration/migration.go
```diff
@@ -82,3 +84,49 @@ func revision0000(txn db.Transaction) error {
...
+		if err != nil {
+			return db.CloseAndWrapOnError(it.Close, err)
+		}
+
+		if err := txn.Set(db.ContractStorage.Key(contractAddress), value); err != nil {
+			return db.CloseAndWrapOnError(it.Close, err)
+		}
+		if err := txn.Delete(oldKey); err != nil {
+			return db.CloseAndWrapOnError(it.Close, err)
+		}
```
omerfirmak, line 128: does iterator behave nicely when we are also deleting the keys as we go

joshklop: Nice catch. From the pebble docs:

> // It is also safe to use an iterator concurrently with modifying its
// underlying DB, if that DB permits modification. However, the resultant
// key/value pairs are not guaranteed to be a consistent snapshot of that DB
// at a particular point in time.

I'll just use txn.Get instead of an iterator.

migration/migration_pkg_test.go
joshklop: Yes. However, because this is a test and we are only iterating `numberOfContracts` times, we know the key has to exist. That's why we require `found` to be true.

joshklop: Actually, just took another look and `txn.Get` is a lot simpler than the iterator. Will use that. Thanks!

omerfirmak (approved):

PR #823, opened 2023-06-15
Duplicate iterator buffers

omerfirmak:
The slice that pebble.Iterator returns for Key() and Value() requests are owned by the iterator. So it can be reused or garbage collected anytime.

We should make sure that we own those slices by copying the data first

### DIFF
--- a/db/pebble/iterator.go
+++ b/db/pebble/iterator.go
@@ -19,12 +19,24 @@ func (i *iterator) Valid() bool {
 
 // Key : see db.Transaction.Iterator.Key
 func (i *iterator) Key() []byte {
-	return i.iter.Key()
+	key := i.iter.Key()
+	if key == nil {
+		return nil
+	}
+	buf := make([]byte, len(key))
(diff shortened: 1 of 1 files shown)


IronGauntlets (approved):

db/pebble/iterator.go
```diff
@@ -19,12 +19,20 @@ func (i *iterator) Valid() bool {
 
 // Key : see db.Transaction.Iterator.Key
 func (i *iterator) Key() []byte {
-	return i.iter.Key()
+	key := i.iter.Key()
+	if key == nil {
+		return nil
+	}
+	return append([]byte{}, key...)
```
joshklop, line 26: Might be a micro-optimization, but what do you think about using `copy` instead of `append`?

joshklop (approved):

db/pebble/iterator.go
omerfirmak: Done

wojciechos (approved):

PR #824, opened 2023-06-15
Fix timeout in deploy node workflow

wojciechos:


.github/workflows/upgrade-nodes-on-test-envs.yml
```diff
@@ -68,6 +68,6 @@ jobs:
                     --colour=false \
                     --eth-node ${{ secrets.ETH_NODE }} \
                     $(if [ "$network" = "mainnet" ]; then echo "--pprof"; fi)
-                port=$((port+1))
-                setsid sh -c "docker logs -f juno_$network >> /var/log/juno_logs_$network.log 2>&1" &
-            done
\ No newline at end of file
+            port=$((port+1))
+            nohup sh -c 'docker logs -f juno_$network > /var/log/juno_logs_$network.log 2>&1' &>/dev/null &
```
kirugan, line 72: why ignore stderr ?

wojciechos: stderr is not ignored - the `2>&1` means that the stderr is also redirected

wojciechos: I mean `&>/dev/null` discards both stdout and stderr of the nohup command itself (not the command it runs)

kirugan: Ahh ok got it

kirugan (approved):

omerfirmak (approved):

PR #825, opened 2023-06-15
Implement juno_version json-rpc method

kirugan:


wojciechos:
Seems like current version fails while trying to sync due to: `Error: Invalid Semantic Version`.

kirugan:
@wojciechos seems like there is no version passed to linker, looking into it 👀

Dockerfile
```diff
@@ -3,7 +3,7 @@ FROM golang:1.20.3-alpine AS build
 
 # Install Alpine Dependencies
 RUN apk update && \
-    apk add build-base clang upx
+    apk add build-base clang upx git
```
kirugan, line 6: @wojciechos command from Makefile `go build -ldflags="-X main.Version=$(shell git describe --tags)" -o build/juno ./cmd/juno/` was failing because there were no git binary in build image

kirugan: Not failing entirely I meant flag `-ldflags="..."` wasn't doing anything helpful

rpc/handlers.go
```diff
@@ -46,9 +46,15 @@ type Handler struct {
...
+	version       string
 }
 
-func New(bcReader blockchain.Reader, synchronizer *sync.Synchronizer, n utils.Network, gatewayClient Gateway,
+func New(
+	bcReader blockchain.Reader,
+	synchronizer *sync.Synchronizer,
+	n utils.Network,
+	gatewayClient Gateway,
+	version string,
```
IronGauntlets, line 57: Lets keep this on a single line and wrap when required:

```go
// Please prefer the following over the above
func New(bcReader blockchain.Reader, synchronizer *sync.Synchronizer,n utils.Network,
	gatewayClient Gateway,	version string, logger utils.Logger) *Handler {...}
```

IronGauntlets (approved):
Please make the minor style change in `handler.go`. 

Approving since I don't need to review the change again.

PR #828, opened 2023-06-16
Correctly return block and receipt status in rpc calls

IronGauntlets:


rpc/handlers.go
```diff
@@ -381,11 +398,24 @@ func (h *Handler) TransactionReceiptByHash(hash felt.Felt) (*TransactionReceipt,
...
 	if blockHash != nil {
 		receiptBlockNumber = &blockNumber
+		l1Head, err := h.bcReader.L1Head()
+		if err != nil && !errors.Is(err, db.ErrKeyNotFound) {
+			return nil, ErrInternal
+		}
+
+		if l1Head != nil && l1Head.BlockNumber >= *receiptBlockNumber {
+			status = StatusAcceptedL1
+		}
```
omerfirmak, line 412: Let's make this a helper?

`bcReader.BlockIsL1Verified(blockHash)`

IronGauntlets: Sounds good

rpc/handlers_test.go
```diff
@@ -368,6 +396,27 @@ func TestBlockWithTxs(t *testing.T) {
...
+		blockWithTxs, rpcErr := handler.BlockWithTxs(rpc.BlockID{Number: latestBlockNumber})
+		require.Nil(t, rpcErr)
+
+		assert.Equal(t, blockWithTxHashes.BlockHeader, blockWithTxs.BlockHeader)
+		assert.Equal(t, len(blockWithTxHashes.TxnHashes), len(blockWithTxs.Transactions))
+
+		checkLatestBlock(t, blockWithTxHashes, blockWithTxs)
+	})
+
+	t.Run("blockID - number accepted on l1", func(t *testing.T) {
```
joshklop, line 413: Should we `assert.Equal(t, rpc.StatusAcceptedL1, block.Status)`, similar to the test for `BlockWithTxHashes`?

rpc/handlers_test.go
```diff
@@ -761,6 +811,23 @@ func TestTransactionReceiptByHash(t *testing.T) {
 	block0, err := mainnetGw.BlockByNumber(context.Background(), 0)
 	require.NoError(t, err)
 
+	checkTxReceipt := func(h *felt.Felt, expected string) {
+		t.Helper()
```
joshklop, line 815: Not sure if this matters, but I think `t` should be a parameter since we're calling this from subtests?

rpc/handlers.go
IronGauntlets: Done

rpc/handlers_test.go
IronGauntlets: The `checkLatestBlock` makes sure that the headers are equal for `blockWithTxHashes` and `blockWithTx` and since we check the Status in `BlockWithTxHasehes` we don't need to recheck it in `blockWithTx` explicitly. This was the convention we used for test other fields as well.

rpc/handlers_test.go
IronGauntlets: Done

blockchain/blockchain.go
```diff
@@ -270,16 +272,53 @@ func (b *Blockchain) Receipt(hash *felt.Felt) (*core.TransactionReceipt, *felt.F
...
+	var update *core.L1Head
+	if err := txn.Get(db.L1Height.Key(), func(updateBytes []byte) error {
+		return encoder.Unmarshal(updateBytes, &update)
 	}); err != nil {
 		return nil, err
 	}
 	return update, nil
 }
 
+func (b *Blockchain) IsL1Verified(blochHash *felt.Felt) bool {
```
kirugan, line 293: Should it be called bloc**k**Hash ? And also is it ok if caller receives `false` and makes assumption that block is not verified on L1 (but actually there is a storage failure or smth.)

IronGauntlets: > Should it be called blockHash ?

I think so, what else can I call it?

> also is it ok if caller receives false and makes assumption that block is not verified on L1 (but actually there is a storage failure or smth.)

I see your point. I prefer to call `L1Head() (*core.L1Head, error)` and let the caller decide what to do with it (like I had before) then we can return the error for storage failure. I can write a private func in `handlers.go` to remove repetition, WDYT @omerfirmak?

I think any `Is*()` function, ideally, shouldn't return an error only a bool. However, I don't mind modifying function signature to:

```go
IsL1Verified(blochHash *felt.Felt) (bool, error)
```

kirugan: > I think so, what else can I call it?

but it's bloc**h**Hash now or you already renamed it?

> I prefer to call L1Head() (*core.L1Head, error) and let the caller decide what to do with it 

got it, makes sense

omerfirmak: > I can write a private func in handlers.go to remove repetition, WDYT @omerfirmak?

> I don't mind modifying function signature


Both works fine IMO, up to you.

IronGauntlets: > but it's blochHash now or you already renamed it?

It was always blockHash you must be getting confused with something else

> Both works fine IMO, up to you.

Ok I am changing to how it was before

IronGauntlets:
Fixed comments and rebase

rpc/handlers.go
```diff
@@ -110,17 +111,41 @@ func (h *Handler) BlockWithTxHashes(id BlockID) (*BlockWithTxHashes, *jsonrpc.Er
 		txnHashes[index] = txn.Hash()
 	}
 
+	l1H, err := h.l1Head()
+	if err != nil {
+		return nil, ErrInternal
```
omerfirmak, line 116: could be nice to use json.Err and pass err.Error() as data

rpc/handlers.go
```diff
@@ -110,17 +111,41 @@ func (h *Handler) BlockWithTxHashes(id BlockID) (*BlockWithTxHashes, *jsonrpc.Er
...
+		return nil, err
+	}
+	// nil is returned if l1 head doesn't exist
+	return l1Head, nil
+}
+
+func isL1Verified(n uint64, h *felt.Felt, l1 *core.L1Head) bool {
+	if l1 != nil && l1.BlockHash.Equal(h) && l1.BlockNumber >= n {
+		return true
+	}
```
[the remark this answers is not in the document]
IronGauntlets: I am not sure which scenario you are referring to

rpc/handlers.go
IronGauntlets: Ok will update

rpc/handlers.go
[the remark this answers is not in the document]
IronGauntlets: Since it is a private func is it not reasonable to make the assumption the `h` and `n` will be of the same block?

I initially passed the block but Transaction Receipt by Hash doesn't have access to the full block. We would need to make another db call to get it. So that is why i passed the block hash and number separately.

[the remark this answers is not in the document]
omerfirmak: that is not what I am trying to say. If the block you are trying to check its L1 verification status (let's call it `target` block) is not the L1 head itself, this function will always return false. You are checking that the `target` block and the l1 head has the same hash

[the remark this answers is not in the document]
IronGauntlets: I understand, we can only compare block numbers and not hashes. I am not sure why I did that, I initially only checked the block number.

rpc/handlers.go
IronGauntlets: fixed

rpc/handlers.go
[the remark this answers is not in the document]
IronGauntlets: fixed

rpc/handlers.go
```diff
@@ -110,17 +111,45 @@ func (h *Handler) BlockWithTxHashes(id BlockID) (*BlockWithTxHashes, *jsonrpc.Er
...
+	l1H, jsonErr := h.l1Head()
+	if jsonErr != nil {
+		return nil, jsonErr
+	}
+
 	status := StatusAcceptedL2
-	if id.Pending {
+	if block.Hash != nil && isL1Verified(block.Number, l1H) {
+		status = StatusAcceptedL1
+	} else if id.Pending {
```
omerfirmak, line 122: I
```suggestion
	if id.Pending {
		status = StatusPending
	} else if isL1Verified(block.Number, l1H) {
```

we dont need block.Hash nil-check this way

rpc/handlers.go
```diff
@@ -110,17 +111,45 @@ func (h *Handler) BlockWithTxHashes(id BlockID) (*BlockWithTxHashes, *jsonrpc.Er
...
 }
 
+func (h *Handler) l1Head() (*core.L1Head, *jsonrpc.Error) {
+	l1Head, err := h.bcReader.L1Head()
+	if err != nil && !errors.Is(err, db.ErrKeyNotFound) {
+		return nil, &jsonrpc.Error{
+			Code:    ErrInternal.Code,
+			Message: ErrInternal.Message,
+			Data:    err.Error(),
+		}
```
omerfirmak, line 140: We can use `jsonrpc.Err` helper here.

rpc/handlers.go
```diff
@@ -153,10 +182,18 @@ func (h *Handler) BlockWithTxs(id BlockID) (*BlockWithTxs, *jsonrpc.Error) {
...
+	l1H, jsonErr := h.l1Head()
+	if jsonErr != nil {
+		return nil, jsonErr
+	}
+
 	status := StatusAcceptedL2
-	if id.Pending {
+	if block.Hash != nil && isL1Verified(block.Number, l1H) {
+		status = StatusAcceptedL1
+	} else if id.Pending {
```
omerfirmak, line 193: same, we can revert the branch order

rpc/handlers.go
```diff
@@ -383,11 +420,26 @@ func (h *Handler) TransactionReceiptByHash(hash felt.Felt) (*TransactionReceipt,
...
+
 	if blockHash != nil {
 		receiptBlockNumber = &blockNumber
+
+		l1H, jsonErr := h.l1Head()
+		if jsonErr != nil {
+			return nil, jsonErr
+		}
+
+		if isL1Verified(*receiptBlockNumber, l1H) {
```
omerfirmak, line 433: nit; we can avoid dereferencing

```suggestion
		if isL1Verified(blockNumber, l1H) {
```

omerfirmak (changes requested):

omerfirmak (approved):
Pushed a commit fixing small nits. Feel free to squash merge when you are back

PR #832, opened 2023-06-16
Method juno_getTransactionStatus

kirugan:


rpc/handlers.go
```diff
@@ -996,3 +1000,26 @@ func (h *Handler) AddDeclareTransaction(declareTx json.RawMessage) (*DeclareTxRe
...
+	} else if txErr == ErrTxnHashNotFound {
+		txStatus, err := h.feederClient.Transaction(context.Background(), &hash)
+		if err != nil {
+			return 0, jsonrpc.Err(jsonrpc.InternalError, err.Error())
+		}
+
+		err = status.UnmarshalJSON([]byte(txStatus.Status))
+		if err != nil {
+			return 0, jsonrpc.Err(jsonrpc.InternalError, err.Error())
+		}
```
omerfirmak, line 1019: Let's not do this IMO. If we can't find the transaction in our db, we dont need to query the gateway.

rpc/handlers.go
```diff
@@ -1,11 +1,13 @@
 package rpc
 
 import (
+	"context"
 	"encoding/json"
 	"errors"
 	"strings"
 
 	"github.com/NethermindEth/juno/blockchain"
+	"github.com/NethermindEth/juno/clients/feeder"
```
omerfirmak, line 10: To be consistent, let's define a interface like we did with `Gateway` do avoid importing this package.

rpc/handlers.go
```diff
@@ -996,3 +1000,26 @@ func (h *Handler) AddDeclareTransaction(declareTx json.RawMessage) (*DeclareTxRe
...
 	return h.version, nil
 }
+
+func (h *Handler) TransactionStatus(hash felt.Felt) (Status, *jsonrpc.Error) {
+	var status Status
+
+	receipt, txErr := h.TransactionReceiptByHash(hash)
+	if txErr == nil {
+		status = receipt.Status
+	} else if txErr == ErrTxnHashNotFound {
```
omerfirmak, line 1010: let's use `errors.Is`

kirugan: the problem that `ErrTxnHashNotFound` is not an error but `&jsonrpc.Error{Code: 25, Message: "Transaction hash not found"}`

omerfirmak: Ah, my bad

rpc/handlers_test.go
```diff
@@ -1938,8 +1938,45 @@ func TestPendingTransactions(t *testing.T) {
...
+	t.Run("transaction not found in db", func(t *testing.T) {
+		mockReader := mocks.NewMockReader(mockCtrl)
+		mockReader.EXPECT().TransactionByHash(hash).Return(nil, db.ErrKeyNotFound)
+		handler := rpc.New(mockReader, nil, utils.MAINNET, nil, client, "", nil)
+
+		status, err := handler.TransactionStatus(*hash)
+		require.Nil(t, err)
+		require.Equal(t, rpc.StatusAcceptedL1, status)
+	})
+}
```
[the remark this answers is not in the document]
kirugan: Added as subtests of a subtest "found l1"

rpc/handlers.go
kirugan: We would still need this package because there is a reference to type `TransactionStatus`:
```
type Feeder interface {
	Transaction(context.Context, *felt.Felt) (*feeder.TransactionStatus, error)
}
```
Do you think I need to use interface anyway?

omerfirmak: I dont think we do in that case.

But @IronGauntlets might have stronger opinions in this matter. I will leave it to him.

omerfirmak (approved):

rpc/handlers.go
IronGauntlets: I think for now it is fine. If we were to introduce the interface here we should do it for `Starknet` data as well, which is out of scope of the pr.

omerfirmak (approved):

Issue #827, opened 2023-06-16
Check chain ID of L1 node


joshklop:
When using `--eth-node`, one can can specify an Ethereum node running on Goerli while syncing Starknet mainnet (other network mis-configurations are also possible). 

We should verify that the L1 node and Starknet network are compatible.

Credit to @wojciechos for finding the bug :)

PR #833, opened 2023-06-19
Check that the L1 node is on the correct network

joshklop:
Resolves: #827 

How we achieve this:

- Add a `ChainID` method to the `l1.Subscriber` interface.
- Add a `network` field to the `l1.Client` struct.
- Set the `l1.Client.network` field with the value retrieved from `blockchain.Network()` in `l1.NewClient(...)`.
- Compare the result of `l1.Subscriber.ChainID` to `l1.Client.network` in `l1.Client.checkChainID`, which is called in `l1.Client.Run`.

The behavior can be manually tested with a command like:

```
juno --eth-node <mainnet-eth-node> --network goerli
```

which should log an error and exit.

### DIFF
--- a/l1/l1.go
+++ b/l1/l1.go
@@ -3,6 +3,7 @@ package l1
 import (
 	"context"
 	"fmt"
+	"math/big"
 
 	"github.com/NethermindEth/juno/blockchain"
 	"github.com/NethermindEth/juno/core"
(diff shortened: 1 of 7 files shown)


l1/l1.go
```diff
@@ -56,7 +60,30 @@ func (c *Client) subscribeToUpdates(ctx context.Context,
...
 }
 
+func (c *Client) checkChainID(ctx context.Context) error {
+	gotChainID, err := c.l1.ChainID(ctx)
+	if err != nil {
+		return err
+	}
+	if err != nil {
+		return fmt.Errorf("retrieve Ethereum chain ID: %w", err)
+	}
```
omerfirmak, line 70: what is going on here?

joshklop: Forgot to remove the first error. Fixed, thanks.

omerfirmak (approved):

PR #843, opened 2023-06-20
Publish transaction hash’s generation method.

Zhangguiguang:


IronGauntlets:
Hi @Zhangguiguang, what is the reason you want to make `TransactionHash` public?

JorikSchellekens:
Because of this possibly? https://github.com/coming-chat/wallet-SDK. Think it makes sense as a utility function

Zhangguiguang:
> Hi @Zhangguiguang, what is the reason you want to make `TransactionHash` public?

Because I am writing a wallet sdk that needs to support users to sign offline, I need the transaction hash as the source of the signature.

Zhangguiguang:
> Because of this possibly? https://github.com/coming-chat/wallet-SDK. Think it makes sense as a utility function

Yes, I need the transaction hash. 😄

omerfirmak:
I think we can just make it public without the new testcase.

IronGauntlets:
Apologies for the late reply.

The reason `transactionHash` is private is that we make a lot of assumptions about the Transaction that is passed to it. Please the `transactionHash` implementation for details. 

The most important one to note is that each transaction type has a `TransactionHash` field, which we assume to always be populated. If we were to make it public it is possible these assumptions won't hold and you can run into panics due to nil pointer errors.

IronGauntlets:
closing because #809 will make `TransactionHash` public.

PR #847, opened 2023-06-20
Passing context.Context() to a json-rpc handler

kirugan:


joshklop:
I originally thought I needed a feature like this to implement subscriptions.

Now I'm thinking I may implement it differently, but since I finished the implementation I went ahead and opened a PR #950.

PR #848, opened 2023-06-21
Add websocket support to jsonrpc package

joshklop:
This commit only modifies the jsonrpc package. It doesn't provide a way to serve websocket connections when Juno is started, which can be implemented in a future PR.

Two design decisions:

We use the nhooyr.io/websocket library. The widely used gorilla/websocket library has been archived and geth is already running into issues with it [1]. The most notable consumer of nhooyr.io/websocket that I could find is Tailscale [2] [3].

Unlike geth [4] and cometbft [5], we don't send ping messages to the client. Pro: saves bandwidth. Con: if the client app hangs, the Websocket connection will persist. As of 2017, Chromium is doing the same [6]. Note that the TCP keep-alives sent every 15 seconds [7] by the HTTP server will detect when the TCP connection is down, but will not detect when the application using the TCP connection is stuck. We can change this in the future, but it will slightly complicate the logic and a ping loop is difficult to test since the pong replies are handled by the websocket library.

[1]: https://github.com/ethereum/go-ethereum/issues/27261
[2]: https://github.com/tailscale/tailscale/blob/0f5090c526c2019fae94695b2991cb561e131788/cmd/derper/websocket.go#L13
[3]: https://github.com/nhooyr/websocket/pull/350#issuecomment-1279512932
[4]: https://github.com/ethereum/go-ethereum/blob/6d2aeb43d516fbe2cafd9e65df7ccbd885f861d3/rpc/websocket.go#L336-L359
[5]: https://github.com/cometbft/cometbft/blob/c138e785c992b3a76cb25fdf2e0820e29c56c1af/rpc/jsonrpc/server/ws_handler.go#L427-L432
[6]: https://groups.google.com/a/chromium.org/g/net-dev/c/2RAm-ZYAIYY/m/NMFuHBuHAAAJhttps://groups.google.com/a/chromium.org/g/net-dev/c/2RAm-ZYAIYY/m/NMFuHBuHAAAJ
[7]: https://github.com/golang/go/blob/f3bf18117b284b63f4350a5aa61773a30d91a6d5/src/net/dial.go#L15-L17


jsonrpc/websocket.go
```diff
@@ -0,0 +1,168 @@
...
+type Websocket struct {
+	rpc        *Server
+	log        utils.SimpleLogger
+	connParams *WebsocketConnParams
+	port       uint16
+}
+
+func NewWebsocket(port uint16, methods []Method, log utils.SimpleLogger) *Websocket {
+	ws := &Websocket{
+		rpc:        NewServer(),
```
omerfirmak, line 23: Maybe in the next PR we could make HTTP and WS share the same RPC server

omerfirmak (approved):
LGTM, feel free to self merge.

jsonrpc/websocket.go
joshklop: I agree that we can reduce a lot of code duplication, but I think we will want separate servers since subscribe endpoints can only be served with websockets.

Issue #839, opened 2023-06-19
Handle batch requests concurrently


joshklop:
From the JSON RPC 2.0 [spec](https://www.jsonrpc.org/specification#batch):

> The Server MAY process a batch rpc call as a set of concurrent tasks, processing them in any order and with any width of parallelism.

I think this is what the [todo](https://github.com/NethermindEth/juno/blob/69bfaebd17a4b3c9eb13e286714060983a0ea545/jsonrpc/server.go#L179) in the code is referring to.

kirugan:
Can I work on this one? Sounds super cool

joshklop:
Sounds great to me!

IronGauntlets:
Done #851

PR #851, opened 2023-06-21
Handle batch requests concurrently

kirugan:
Issue #839

### DIFF
--- a/jsonrpc/server.go
+++ b/jsonrpc/server.go
@@ -9,7 +9,6 @@ import (
 	"errors"
 	"io"
 	"reflect"
-	"runtime"
 	"strings"
 	"sync"
 	"time"
@@ -125,11 +124,11 @@ type Validator interface {
 }
 
 // NewServer instantiates a JSONRPC server
-func NewServer(log utils.SimpleLogger) *Server {
+func NewServer(poolMaxGoroutines int, log utils.SimpleLogger) *Server {
 	s := &Server{
 		log:     log,
 		methods: make(map[string]Method),
-		pool:    pool.New().WithMaxGoroutines(runtime.GOMAXPROCS(0)),
+		pool:    pool.New().WithMaxGoroutines(poolMaxGoroutines),
 		requests: prometheus.NewCounterVec(prometheus.CounterOpts{
 			Namespace: "rpc",
 			Subsystem: "server",
--- a/jsonrpc/server.go
+++ b/jsonrpc/server.go
@@ -10,11 +10,13 @@ import (
 	"io"
 	"reflect"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/NethermindEth/juno/metrics"
 	"github.com/NethermindEth/juno/utils"
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/sourcegraph/conc/pool"
 )
 
 const (
@@ -211,52 +213,13 @@ func (s *Server) HandleReader(ctx context.Context, reader io.Reader) ([]byte, er
 		}
 	} else {
 		var batchReq []json.RawMessage
-		var batchRes []json.RawMessage
 
 		if batchJSONErr := dec.Decode(&batchReq); batchJSONErr != nil {
 			res.Error = Err(InvalidJSON, batchJSONErr.Error())
 		} else if len(batchReq) == 0 {
 			res.Error = Err(InvalidRequest, "empty batch")
 		} else {
-			for _, rawReq := range batchReq { // todo: handle async
-				var resObject *response
-
-				reqDec := json.NewDecoder(bytes.NewBuffer(rawReq))
-				reqDec.UseNumber()
-
-				req := new(request)
-				if jsonErr := reqDec.Decode(req); jsonErr != nil {
-					resObject = &response{
-						Version: "2.0",
-						Error:   Err(InvalidRequest, jsonErr.Error()),
-					}
-				} else {
-					var handleErr error
-					resObject, handleErr = s.handleRequest(ctx, req)
-					if handleErr != nil {
-						resObject = &response{
-							Version: "2.0",
-							Error:   Err(InvalidRequest, handleErr.Error()),
-						}
-						if !errors.Is(handleErr, ErrInvalidID) {
-							resObject.ID = req.ID
-						}
-					}
-				}
-
-				if resObject != nil {
-					if resArr, jsonErr := json.Marshal(resObject); jsonErr != nil {
-						return nil, jsonErr
-					} else {
-						batchRes = append(batchRes, resArr)
-					}
-				}
-			}
-
-			if len(batchRes) == 0 {
-				return nil, nil
-			}
-			return json.Marshal(batchRes)
+			return s.handleBatchRequest(ctx, batchReq)
 		}
 	}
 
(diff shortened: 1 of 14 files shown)


jsonrpc/server.go
```diff
@@ -4,13 +4,16 @@ package jsonrpc
...
+	"context"
 	"encoding/json"
 	"errors"
 	"io"
 	"reflect"
 	"strings"
+	"sync"
 
 	"github.com/go-playground/validator/v10"
+	"golang.org/x/sync/errgroup"
```
kirugan, line 16: Pros:
- allows to receive errors from goroutines
Cons:
- there is no way to limit number of spawned goroutines out of the box

omerfirmak: how about https://pkg.go.dev/github.com/sourcegraph/conc/pool#ErrorPool, we use this package already

joshklop (commented):
LGTM. Exciting! I can approve once the conversation about `errgroup`/`conc.ErrorPool` is resolved.

jsonrpc/server.go
kirugan: nice, I'll use this one

jsonrpc/server.go
```diff
@@ -139,13 +142,13 @@ func (s *Server) RegisterMethod(method Method) error {
 // It returns the response in a byte array, only returns an
 // error if it can not create the response byte array
 func (s *Server) Handle(data []byte) ([]byte, error) {
-	return s.HandleReader(bytes.NewReader(data))
+	return s.HandleReader(context.Background(), bytes.NewReader(data))
```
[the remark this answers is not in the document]
kirugan: Sure, I'll add context to `Handle(...)` as well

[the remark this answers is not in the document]
kirugan: added, note that in `handleBatchRequest` context is not used for now (but I'll pass it to handlers in upcoming PR)

jsonrpc/server.go
```diff
@@ -188,14 +188,14 @@ func (s *Server) HandleReader(ctx context.Context, reader io.Reader) ([]byte, er
 	return json.Marshal(res)
 }
 
-func (s *Server) handleBatchRequest(ctx context.Context, batchReq []json.RawMessage) ([]byte, error) {
+func (s *Server) handleBatchRequest(_ context.Context, batchReq []json.RawMessage) ([]byte, error) {
 	var batchRes []json.RawMessage
 	var resMutex sync.Mutex
 
-	g, _ := errgroup.WithContext(ctx)
+	p := pool.New().WithErrors()
```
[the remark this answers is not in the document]
omerfirmak: A simpler alternative; We can keep the old behaviour and unmarshal all the requests at the main go routine and only do the actual handling in parallel. Unmarshaling shouldnt be that demanding anyways.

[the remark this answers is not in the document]
joshklop: I like that.

jsonrpc/server.go
```diff
@@ -169,59 +172,81 @@ func (s *Server) HandleReader(reader io.Reader) ([]byte, error) {
...
-				reqDec := json.NewDecoder(bytes.NewBuffer(rawReq))
-				reqDec.UseNumber()
+			reqDec := json.NewDecoder(bytes.NewBuffer(rawReq))
+			reqDec.UseNumber()
 
-				req := new(request)
-				if jsonErr := reqDec.Decode(req); jsonErr != nil {
+			req := new(request)
+			if jsonErr := reqDec.Decode(req); jsonErr != nil {
+				cancel()
```
omerfirmak, line 209: requests should are handled separately. so even if one of them is not a valid request object, the rest should still be processed.

only exception is if ```		if batchJSONErr := dec.Decode(&batchReq); batchJSONErr != nil {
			res.Error = Err(InvalidJSON, batchJSONErr.Error())``` fails. In that case, we fail the entire batch with invalid json error

jsonrpc/server.go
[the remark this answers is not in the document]
joshklop: > In this case if one request contains wrong data the whole batch will be discarded

I think that is the desired behavior, according to the [spec](https://www.jsonrpc.org/specification#batch):

> If the batch rpc call itself fails to be recognized as an valid JSON or as an Array with at least one value, the response from the Server MUST be a single Response object.

omerfirmak:
@kirugan can we address the comments here? It should be a good addition.

omerfirmak:
@kirugan can you rebase?

jsonrpc/server.go
kirugan: Removed `cancel()`

jsonrpc/server.go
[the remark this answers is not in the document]
kirugan: > I think that is the desired behavior, according to the [spec](https://www.jsonrpc.org/specification#batch):
> 
> >

For instance `[1,2,3]` is a valid JSON, but it contains wrong values so response should be an array with errors for each item. Here is an example from the spec:
```
--> [1,2,3]
<-- [
  {"jsonrpc": "2.0", "error": {"code": -32600, "message": "Invalid Request"}, "id": null},
  {"jsonrpc": "2.0", "error": {"code": -32600, "message": "Invalid Request"}, "id": null},
  {"jsonrpc": "2.0", "error": {"code": -32600, "message": "Invalid Request"}, "id": null}
]
```
in the end I decided to use `[ ]json.RawMessage` in outer layer and replace `ErrorPool` with regular `Pool` so requests are completely independent (and error in one of them won't break the whole batch)

jsonrpc/server.go
```diff
@@ -211,59 +213,76 @@ func (s *Server) HandleReader(ctx context.Context, reader io.Reader) ([]byte, er
...
-				reqDec.UseNumber()
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
 
-				req := new(request)
-				if jsonErr := reqDec.Decode(req); jsonErr != nil {
+	p := pool.New()
+	for _, rawReq := range batchReq {
+		rawReq := rawReq
+		p.Go(func() {
```
omerfirmak, line 242: Is it possible to just call `Handle` here?

jsonrpc/server.go
```diff
@@ -211,59 +213,76 @@ func (s *Server) HandleReader(ctx context.Context, reader io.Reader) ([]byte, er
...
+	var resMutex sync.Mutex
 
-				reqDec := json.NewDecoder(bytes.NewBuffer(rawReq))
-				reqDec.UseNumber()
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
 
-				req := new(request)
-				if jsonErr := reqDec.Decode(req); jsonErr != nil {
+	p := pool.New()
```
omerfirmak, line 239: We should limit concurrency here else it could be a DoS vector.

Can we make pool a shared thing in `Server`? So that all requests use the same pool for concurrency and we dont overwhelm the scheduler at the times of heavy RPC usage.

kirugan: Good point, do you think server should accept size of pool in constructor or we can hardcode its value to some reasonable value?

omerfirmak: `runtime.GOMAXPROCS(0)` should be good enough for now.

kirugan: Done. It requires additional `WaitGroup` var  (`pool.Wait()` might wait for parallel batch request)

omerfirmak: > Done. It requires additional WaitGroup var

Yeap, that is fine.

jsonrpc/server.go
```diff
@@ -211,59 +216,78 @@ func (s *Server) HandleReader(ctx context.Context, reader io.Reader) ([]byte, er
...
+			reqDec := json.NewDecoder(bytes.NewBuffer(rawReq))
+			reqDec.UseNumber()
+
+			req := new(request)
+			if jsonErr := reqDec.Decode(req); jsonErr != nil {
+				resObject = &response{
+					Version: "2.0",
+					Error:   Err(InvalidRequest, jsonErr.Error()),
+				}
+			} else {
```
omerfirmak, line 259: Nit; we could do this outside of go routine, to avoid spinning up new go routines for obviously invalid requests.

jsonrpc/server.go
```diff
@@ -211,59 +216,78 @@ func (s *Server) HandleReader(ctx context.Context, reader io.Reader) ([]byte, er
...
+	var resMutex sync.Mutex
+
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	var wg sync.WaitGroup
+	wg.Add(len(batchReq))
+	for _, rawReq := range batchReq {
+		rawReq := rawReq
+		s.pool.Go(func() {
```
omerfirmak, line 246: one thing to keep in mind with the shared pool is a big batch request can hog the entire pool and cause additional latency for requests from other users.

omerfirmak (approved):

jsonrpc/server.go
kirugan: Rewrote it

jsonrpc/server.go
```diff
@@ -211,59 +216,83 @@ func (s *Server) HandleReader(ctx context.Context, reader io.Reader) ([]byte, er
...
+	defer cancel()
+
+	var wg sync.WaitGroup
+	for _, rawReq := range batchReq {
+		reqDec := json.NewDecoder(bytes.NewBuffer(rawReq))
+		reqDec.UseNumber()
+
+		req := new(request)
+		if err := reqDec.Decode(req); err != nil {
+			addResponse(&response{
```
joshklop, line 261: Does the issue previously discussed still exist? That is: if one request has a decoding error, all requests for the batch should be canceled and only a single response sent back.

kirugan: There are two cases:
1. Entire request is an invalid JSON, example:
```
--> [
  {"jsonrpc": "2.0", "method": "sum", "params": [1,2,4], "id": "1"},
  {"jsonrpc": "2.0", "method"
]
<-- {"jsonrpc": "2.0", "error": {"code": -32700, "message": "Parse error"}, "id": null}
```
2. Inside an array (in batch request) one or more items are not valid according to spec (but it's still valid JSON item):
```
--> [
        {"jsonrpc": "2.0", "method": "sum", "params": [1,2,4], "id": "1"},
        {"jsonrpc": "2.0", "method": "notify_hello", "params": [7]},
        {"jsonrpc": "2.0", "method": "subtract", "params": [42,23], "id": "2"},
        {"foo": "boo"},
        {"jsonrpc": "2.0", "method": "foo.get", "params": {"name": "myself"}, "id": "5"},
        {"jsonrpc": "2.0", "method": "get_data", "id": "9"} 
    ]
<-- [
        {"jsonrpc": "2.0", "result": 7, "id": "1"},
        {"jsonrpc": "2.0", "result": 19, "id": "2"},
        {"jsonrpc": "2.0", "error": {"code": -32600, "message": "Invalid Request"}, "id": null},
        {"jsonrpc": "2.0", "error": {"code": -32601, "message": "Method not found"}, "id": "5"},
        {"jsonrpc": "2.0", "result": ["hello", 5], "id": "9"}
    ]
```
and
```
--> [1]
<-- [
  {"jsonrpc": "2.0", "error": {"code": -32600, "message": "Invalid Request"}, "id": null}
]
```

both cases are handled the way spec defines it.

jsonrpc/server.go
kirugan: Added parameter to control max size of the pool

PR #853, opened 2023-06-22
Never revert L1 head

### LINKED ISSUE #750: Reject any new L2 block that conflicts with the known L1 head
If blockchain realizes that the local chain has diverged from the L1 canonical chain, it should stop accepting any new blocks even if they are perfectly fine and would be accepted normally.

In that case, it should return a special error that the sync package can match and take the necessary actions, like reverting some blocks.

Depends on #699 #707

### LINKED ISSUE #818: Make sure L1 head is not reverted
If a block is L1 head, we should not revert it

joshklop:
Fixes #818 
Fixes #750

### DIFF
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -46,7 +46,9 @@ type Reader interface {
 
 var (
 	ErrParentDoesNotMatchHead = errors.New("block's parent hash does not match head block hash")
-	supportedStarknetVersion  = semver.MustParse("0.12.1")
+	ErrBlockConflictsWithL1   = errors.New("block conflicts with L1 head")
+
+	supportedStarknetVersion = semver.MustParse("0.12.1")
 )
 
 func checkBlockVersion(protocolVersion string) error {
@@ -313,7 +315,7 @@ func (b *Blockchain) Store(block *core.Block, blockCommitments *core.BlockCommit
 	stateUpdate *core.StateUpdate, newClasses map[felt.Felt]core.Class,
 ) error {
 	return b.database.Update(func(txn db.Transaction) error {
-		if err := verifyBlock(txn, block); err != nil {
+		if err := b.verifyBlock(txn, block); err != nil {
 			return err
 		}
 		if err := core.NewState(txn).Update(block.Number, stateUpdate, newClasses); err != nil {
@@ -353,11 +355,11 @@ func (b *Blockchain) Store(block *core.Block, blockCommitments *core.BlockCommit
 // VerifyBlock assumes the block has already been sanity-checked.
 func (b *Blockchain) VerifyBlock(block *core.Block) error {
 	return b.database.View(func(txn db.Transaction) error {
-		return verifyBlock(txn, block)
+		return b.verifyBlock(txn, block)
 	})
 }
 
-func verifyBlock(txn db.Transaction, block *core.Block) error {
+func (b *Blockchain) verifyBlock(txn db.Transaction, block *core.Block) error {
 	if err := checkBlockVersion(block.ProtocolVersion); err != nil {
 		return err
 	}
@@ -378,6 +380,26 @@ func verifyBlock(txn db.Transaction, block *core.Block) error {
 	}
 	if !block.ParentHash.Equal(expectedParentHash) {
 		return ErrParentDoesNotMatchHead
+	}
+
+	head, err := l1Head(txn)
+	switch {
+	case err == nil:
+		var header *core.Header
+		if head.BlockNumber == block.Number {
+			header = block.Header
+		} else {
+			header, err = blockHeaderByNumber(txn, head.BlockNumber)
+		}
+
+		if err == nil && !head.BlockHash.Equal(header.Hash) {
+			return ErrBlockConflictsWithL1
+		} else if !errors.Is(err, db.ErrKeyNotFound) {
+			return err
+		}
+	case errors.Is(err, db.ErrKeyNotFound):
+	default:
+		return err
 	}
 
 	return nil
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -787,11 +787,22 @@ func (b *Blockchain) RevertHead() error {
 	return b.database.Update(b.revertHead)
 }
 
-func (b *Blockchain) revertHead(txn db.Transaction) error {
+func (b *Blockchain) revertHead(txn db.Transaction) error { //nolint:gocyclo
 	blockNumber, err := chainHeight(txn)
 	if err != nil {
 		return err
 	}
+
+	// Don't revert part of the L1-verified chain.
+	// If the L1 head isn't in the db yet, revert.
+	// If the L2 height is less than the L1 height, don't revert.
+	var head *core.L1Head
+	if head, err = l1Head(txn); err != nil && !errors.Is(err, db.ErrKeyNotFound) {
+		return err
+	} else if head != nil && blockNumber <= head.BlockNumber {
+		return errors.New("cannot revert L1-verified block")
+	}
+
 	numBytes := core.MarshalBlockNumber(blockNumber)
 
 	stateUpdate, err := stateUpdateByNumber(txn, blockNumber)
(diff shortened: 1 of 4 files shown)


blockchain/blockchain.go
```diff
@@ -734,6 +734,17 @@ func revertHead(txn db.Transaction) error {
...
 		return err
 	}
+
+	// Don't revert part of the L1-verified chain.
+	// If the L1 head isn't in the db yet, revert.
+	// If the L2 height is less than the L1 height, don't revert.
+	var head *core.L1Head
+	if head, err = l1Head(txn); err != nil && !errors.Is(err, db.ErrKeyNotFound) {
+		return err
+	} else if head != nil && blockNumber <= head.BlockNumber {
```
omerfirmak, line 744: This assumes that our L1 head always agrees with our L2 chain. I don't think we ensure that yet, right?

joshklop: Correct. I think when I first wrote the PR I intended to do the changes separately, but can't remember.

In any case, the latest changes should ensure the L1 and L2 chains always agree.

blockchain/blockchain.go
```diff
@@ -570,6 +581,27 @@ func (b *Blockchain) SanityCheckNewHeight(block *core.Block, stateUpdate *core.S
...
+				L2BlockHash: header.Hash,
+				Height:      head.BlockNumber,
+			}
+		} else if !errors.Is(err, db.ErrKeyNotFound) {
+			return err
+		}
+	} else if !errors.Is(err, db.ErrKeyNotFound) {
+		return err
+	}
+
```
omerfirmak, line 604: This should be done at the time of Store operation to be on the safe side. `SanityCheckNewHeight` is called async to the store&L1 verification process, so it might create race conditions.

omerfirmak: How I think it should work is;

- Before storing a new L2 block, make sure it agrees with the L1 head. If not, reject
- After storing a L1 head, make sure the L2 chain agrees with the new L1 head. If not, revert the L2 chain as many blocks as necessary.

joshklop: > After storing a L1 head, make sure the L2 chain agrees with the new L1 head.  If not, revert the L2 chain as many blocks as necessary.

This one is tough with the current design. Rolling back the L2 chain in `SetL1Head` creates a race between the synchronizer and L1 verifier (please correct me if I'm wrong). At least it is tough to reason about.

The current implementation will eventually resolve any discrepancy between L1 and L2 in `SanityCheckNewHeight` (will move that to `Store` per your first comment).

Does that seem sufficient?

omerfirmak: > Rolling back the L2 chain in SetL1Head creates a race between the synchronizer and L1 verifier (please correct me if I'm wrong).

Not really. We only allow a single write transaction, `SetL1Head` blocks all other `Store` calls. So if `SetL1Head` ends up reverting the L2 chain, next call to `Store` that synchronizer makes will fail and synchronizer will have to restart the sync loop and try to sync new blocks on the new L1 verified chain.

joshklop: Sounds good. I'll add a check to `SetL1Head`.

> We only allow a single write transaction

Do we specify this somewhere, or is it a Pebble thing?

omerfirmak: > Do we specify this somewhere, or is it a Pebble thing?

It is something that I thought would help us reason with concurrency  

https://github.com/NethermindEth/juno/blob/8bb4c7e6e6f2dcddcb22b0b1378866fc624d9c71/db/pebble/db.go#L53

blockchain/blockchain.go
```diff
@@ -48,6 +48,17 @@ var (
...
+	L1HeadHash  *felt.Felt
+	L2BlockHash *felt.Felt
+	Height      uint64
+}
+
+func (e ErrBlockConflictsWithL1) Error() string {
+	return fmt.Sprintf("l2 block hash conflicts with l1 head (height %d): got %s, want %s",
+		e.Height, e.L2BlockHash.String(), e.L1HeadHash.String())
+}
+
```
omerfirmak, line 61: The additional information is not really used by the callee. Let's just make this a string err like `ErrParentDoesNotMatchHead` ?

blockchain/blockchain.go
```diff
@@ -369,6 +380,30 @@ func verifyBlock(txn db.Transaction, block *core.Block) error {
...
+	switch {
+	case err == nil:
+		var header *core.Header
+		if head.BlockNumber == block.Number {
+			header = block.Header
+		} else {
+			header, err = b.BlockHeaderByNumber(head.BlockNumber)
+		}
+
+		if err == nil && !head.BlockHash.Equal(header.Hash) {
```
omerfirmak, line 425: I am not sure if I follow this.

We are checking if `head.BlockHash.Equal(header.Hash)` but `head.BlockNumber == header.Number` is not necessarily always true. So how do we expect `head.BlockHash.Equal(header.Hash)` to be always equal?

joshklop: > `head.BlockNumber == header.Number` is not necessarily always true

We get the matching header from the db in the `else` clause a few lines above.

blockchain/blockchain_test.go
```diff
@@ -595,6 +595,36 @@ func TestRevert(t *testing.T) {
...
+	})
+
+	t.Run("with L1 head, error", func(t *testing.T) {
+		require.NoError(t, chain.SetL1Head(&core.L1Head{
+			BlockNumber: 1,
+		}))
+		require.Error(t, chain.RevertHead())
+	})
+}
+
```
omerfirmak: can we add a case here where L1 head is not nil, but we can revert blocks newer than L1 height?

blockchain/blockchain.go
```diff
@@ -301,14 +302,46 @@ func (b *Blockchain) SetL1Head(update *core.L1Head) error {
...
+				return nil
+			}
+		case errors.Is(err, db.ErrKeyNotFound):
+			return nil
+		default:
+			return err
+		}
+
+		// New L1 head conflicts with L2 chain.
+		// Revert the L2 chain to one block before the new L1 head.
```
omerfirmak, line 321: I dont think this is enough.

We need to revert until the common ancestor of our L1 head and L2 chain.
For example;

Assume there are 2 forks; first one is our local L2 chain, which contains some reorged blocks( `B'X` )

`B0 <- B1 <-  B2 <-  B'3 <-  B'4 <-  B'5 <- B'6` and our L1 head is `B2`



From L1, we get verification on `B5` (since skipping `B3` & `B4` is perfectly valid), which is the canonical block for that slot.
We roll back to the chain below;

`B0 <- B1 <-  B2 <-  B'3 <-  B'4`

`B'3` and `B'4` are still in storage, but they are older then the new L1 head, `B6`. So we will never be able to revert them.

What we should do is, somehow, realize that `B2` is the most recent common block in both forks and revert until that.

omerfirmak (changes requested):

blockchain/blockchain.go
omerfirmak: @joshklop I think the best way forward is to implement beacon-sync method. (ie start from L1 head and sync backwards).

That way, we don't have this problem at all.

joshklop: Agree that beacon sync is good in general, but does it help here? AFAIK, new L1 heads are not guaranteed to be in the L2 chain, so we still need to manage the case where they conflict with our current L2 chain.

omerfirmak: Yeap, it just makes it easier.

When the node first boots, it fetches the L1 head and starts syncing from that. So we for sure know that our first L1 head is compatible with the L2 chain that we are syncing to.

We can keep doing this for each incompatible L1 head that we get. We can just revert until the previous L1 head,  and beacon sync from new L1 head to the previous L1 head.

joshklop: Agree. I see your point. 

While making unrelated changes to accept only finalized L1 blocks, I was thinking about some simplifications we could make to the sync <> L1 integration; I think those would help here as well.

Will return to this once 0.4 and 0.12.1 changes are through.

PR #858, opened 2023-06-24
Only accept finalized blocks on L1

### LINKED ISSUE #826: Handle L1 reorgs
Right now we just store L1 updates in the database. Now that the `blockchain.Blockchain` can revert, we should handle L1 reorgs.

joshklop:
Resolves #826 

Depends on #860


l1/l1.go
```diff
@@ -75,101 +62,86 @@ func (c *Client) checkChainID(ctx context.Context) error {
...
+			if err != nil {
+				c.log.Warnw("Failed to fetch finalised and latest L1 block numbers", "err", err)
+				continue
+			}
 
-			subHeaders, err = c.subscribeToHeaders(ctx, headerChan)
+			if c.finalisedHeight > finalisedL1BlockNumber {
+				// TODO how should we handle such catastrophic events?
+				return errReorgFinalisedBlock
+			}
```
omerfirmak, line 90: This could happen when using a RPC provider. Since RPC providers run multiple nodes and put them behind a load balancer, our requests could be handled by different nodes. So getting a new finalized height that is lower than our c.finalisedHeight is not necessarily a catastrophic failure, it might just be that one of the nodes behind the load balancer could be a bit behind.

So I think we should simply ignore it instead of shutting the service down.

l1/l1.go
```diff
@@ -75,101 +62,86 @@ func (c *Client) checkChainID(ctx context.Context) error {
...
-					c.log.Warnw("L1 update subscription failed, resubscribing", "error", err)
-					subUpdates.Unsubscribe()
-
-					subUpdates, err = c.subscribeToUpdates(ctx, logStateUpdateChan)
-					if err != nil {
-						return err
+			c.log.Debugw("Retrieved L1 head", "latestL1", latestL1BlockNumber, "finalisedL1",
+				finalisedL1BlockNumber, "l2BlockNumber", head.BlockNumber)
+
+			c.finalisedQueue[latestL1BlockNumber] = head
```
omerfirmak, line 102: isnt this `non-finalisedQueue` ? Confused me a little.

l1/l1.go
```diff
@@ -75,101 +62,86 @@ func (c *Client) checkChainID(ctx context.Context) error {
...
+			var maxFinalisedNumber uint64
+			for l1BlockNumber := range c.finalisedQueue {
+				if l1BlockNumber > latestL1BlockNumber {
+					// Ethereum block was reorged. Remove l1Head from queue.
+					delete(c.finalisedQueue, l1BlockNumber)
```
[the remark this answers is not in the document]
omerfirmak: What we should do is, store `latestL1BlockHash` together with the head here; 

`c.finalisedQueue[latestL1BlockNumber] = head`

and make sure that `latestL1BlockHash` is still a part of the canonical chain.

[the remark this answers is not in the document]
omerfirmak: Honestly, I believe the event subscription approach handles this better. Instead of `latest - confirmationPeriod` we would just check the events we have in the queue against the most recent `finalisedL1BlockNumber`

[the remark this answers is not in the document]
joshklop: Good point. I'll add tests for this.

To fix, I think I'll use [`eth_newBlockFilter`](https://docs.infura.io/networks/ethereum/json-rpc-methods/filter-methods/eth_newblockfilter) + [`eth_getFilterChanges`](https://docs.infura.io/networks/ethereum/json-rpc-methods/filter-methods/eth_getfilterchanges) to maintain a slice of block hashes from the finalized block to the latest block. i.e., we keep track of the fork the Ethereum node is on.

l1/l1.go
joshklop: Much better name. Fixed.

l1/ethereum.go
```diff
@@ -0,0 +1,210 @@
...
+			Result: &stateRootString,
+		},
+	}
+	if err := e.batchCallContext(ctx, batch); err != nil {
+		return nil, err
+	}
+
+	// We use slice[2:] to remove the `0x` prefix.
+
+	blockNumber, err := strconv.ParseUint(blockNumberString[2:], felt.Base16, 64)
```
kirugan, line 161: let's use explicitly `strings.TrimPrefix(...)` for blockNumber

l1/ethereum.go
```diff
@@ -0,0 +1,210 @@
...
+}
+
+// ChainID implements part of the L1 interface.
+func (e *Ethereum) ChainID(ctx context.Context) (*big.Int, error) {
+	var result string
+	err := e.client.CallContext(ctx, &result, "eth_chainId")
+	if err != nil {
+		return nil, err
+	}
+	resultNoPrefix := result[2:] // Remove `0x` prefix.
```
kirugan, line 204: same for `TrimPrefix(...)` it's more explicit and has same number of allocations/copying

l1/l1.go
```diff
@@ -75,101 +62,86 @@ func (c *Client) checkChainID(ctx context.Context) error {
...
-			subHeaders.Unsubscribe()
+		case <-ticker.C:
+			latestL1BlockNumber, finalisedL1BlockNumber, latestL1BlockHash, err := c.l1.LatestAndFinalisedHeights(ctx)
+			if err != nil {
+				c.log.Warnw("Failed to fetch finalised and latest L1 block numbers", "err", err)
+				continue
+			}
 
-			subHeaders, err = c.subscribeToHeaders(ctx, headerChan)
+			if c.finalisedHeight > finalisedL1BlockNumber {
```
IronGauntlets, line 87: How is this possible?

l1/l1.go
```diff
@@ -75,101 +62,86 @@ func (c *Client) checkChainID(ctx context.Context) error {
...
+
+			c.nonFinalisedQueue[latestL1BlockNumber] = head
+			var maxFinalisedNumber uint64
+			for l1BlockNumber := range c.nonFinalisedQueue {
+				if l1BlockNumber > latestL1BlockNumber {
+					// Ethereum block was reorged. Remove l1Head from queue.
+					delete(c.nonFinalisedQueue, l1BlockNumber)
+				} else if l1BlockNumber <= finalisedL1BlockNumber && l1BlockNumber >= maxFinalisedNumber {
+					if l1BlockNumber >= maxFinalisedNumber {
+						maxFinalisedNumber = l1BlockNumber
```
IronGauntlets, line 110:
```suggestion
				} else if l1BlockNumber <= finalisedL1BlockNumber && l1BlockNumber >= maxFinalisedNumber {
						maxFinalisedNumber = l1BlockNumber
```

Same condidtion was tested twice, once in and and again as if statement

l1/l1.go
```diff
@@ -75,101 +62,86 @@ func (c *Client) checkChainID(ctx context.Context) error {
...
-					} else {
-						c.confirmationQueue.Enqueue(logStateUpdate)
-					}
-				default:
-					break Outer
 				}
 			}
 
-			if err := c.setL1Height(l1Height); err != nil {
+			maxFinalisedHead := c.nonFinalisedQueue[maxFinalisedNumber]
```
IronGauntlets, line 115: I think it is better to use `x, ok := someMap[someKey]`, where ok is either `true` or `false` depending on whether `someKey` exists or not.

l1/l1.go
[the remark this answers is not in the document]
joshklop: You were right, the event subscriptions are the way to go. Updated.

l1/l1.go
joshklop: Thanks, fixed.

l1/l1.go
joshklop: Thanks. Changed code to use event subscription, these lines are no longer present.

l1/l1.go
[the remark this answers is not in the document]
IronGauntlets: > In practice ethereum nodes never rollback. They just switch to a better fork with atleast the same amount of blocks they have locally, so latestL1BlockNumber will always monotonically increase.

As far as I understand the above is not true because reorg logic is not "longest chain" rather it is the "heaviest chain". Therefore, it is entirely possible for an ethereum to switch to a chain with a lower height but it is "heavier" than a "longer" chain. For `ethash`, the reorg logic would take previous blocks + uncle/omner blocks into account to decide the chain to build on. I assume this was kept for the DAG part of Casper consensus mechanism.

l1/mocks/mock_subscriber.go
IronGauntlets, line 1: Shouldn't this be under `juno/mock` dir?

joshklop: I put it under `l1` since it isn't needed outside the `l1` package. I can move it, but it will increase the diff.

l1/eth_subscriber.go
```diff
@@ -46,6 +58,31 @@ func (s *EthSubscriber) ChainID(ctx context.Context) (*big.Int, error) {
 	return s.ethClient.ChainID(ctx)
 }
 
+func (s *EthSubscriber) FinalisedHeight(ctx context.Context) (uint64, error) {
+	finalisedBlock := make(map[string]any, 0)
+	if err := s.client.CallContext(ctx, &finalisedBlock, "eth_getBlockByNumber", "latest", false); err != nil {
```
IronGauntlets, line 63: Shouldn't this be `finalized`?

joshklop: Yes, nice catch, thanks!

l1/mocks/mock_subscriber.go
IronGauntlets: I think a single place for all the mocks is best

l1/l1.go
```diff
@@ -78,98 +77,127 @@ func (c *Client) checkChainID(ctx context.Context) error {
...
 						"blockHash", logStateUpdate.BlockHash.Text(felt.Base16))
+
 					if logStateUpdate.Raw.Removed {
-						// NOTE: we only modify the local confirmationQueue upon receiving reorged logs.
-						// We assume new logs will soon follow, so we don't notify the l2Chain.
-						c.confirmationQueue.Reorg(logStateUpdate.Raw.BlockNumber)
+						for l1BlockNumber := range c.nonFinalisedLogs {
+							if l1BlockNumber <= logStateUpdate.Raw.BlockNumber {
+								delete(c.nonFinalisedLogs, l1BlockNumber)
+							}
```
IronGauntlets, line 139:
```suggestion
							if l1BlockNumber > logStateUpdate.Raw.BlockNumber {
								delete(c.nonFinalisedLogs, l1BlockNumber)
							}
```

Shouldn't it be the above? I don't exactly understand why we would want to delete blocks lower than the current height? We should delete it and anything above, right?

PR #860, opened 2023-06-26
Use listener instead of port in jsonrpc.HTTP

joshklop:
Allows us to set up HTTP JSON RPC servers in tests without having to choose an open port.

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -3,6 +3,7 @@ package node
 import (
 	"context"
 	"fmt"
+	"net"
 	"path/filepath"
 	"reflect"
 	"time"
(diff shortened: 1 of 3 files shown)


kirugan (approved):

node/node.go
```diff
@@ -220,7 +221,13 @@ func (n *Node) Run(ctx context.Context) {
...
 	gatewayClient := gateway.NewClient(n.cfg.Network.GatewayURL(), n.log)
 
-	http := makeHTTP(n.cfg.RPCPort, rpc.New(n.blockchain, synchronizer, n.cfg.Network, gatewayClient, n.version, n.log), n.log)
+	listener, err := net.Listen("tcp", fmt.Sprintf(":%d", n.cfg.RPCPort))
+	if err != nil {
+		n.log.Errorw("Failed to listen for RPC requests", "port", n.cfg.RPCPort)
+		return
+	}
+	defer listener.Close()
+	http := makeHTTP(listener, rpc.New(n.blockchain, synchronizer, n.cfg.Network, gatewayClient, n.version, n.log), n.log)
```
omerfirmak, line 230: I am not sure about this, our `Run` function is already overly crowded as it is.

Instead of changing `rpc.New`, can we introdue something like `rpc.NewWithListener` that we can use in tests only?

joshklop: I learned that don't need to close the listener (`http.Shutdown` does it for us), which decreases complexity and size of the function.

PR #863, opened 2023-06-27
P2P part1

asdacap:
Part1 of probably several for p2p codes.
- Basic peer discovery via kademlia.
- Peer pool implementation, which will be used by sync code later.


omerfirmak (approved):
I am fine with merging this even after the release.

asdacap:
Lets merge this.

omerfirmak:
Fixed the history, please use rebase only to maintain a linear history. Merge commits makes it realy hard to collaborate on a branch

omerfirmak:
I want to merge this as well, @NethermindEth/nubia-juno please take a look.

node/node.go
```diff
@@ -122,6 +129,17 @@ func New(cfg *Config, version string) (*Node, error) {
 		n.services = append(n.services, grpc.NewServer(n.cfg.GRPCPort, n.version, n.db, n.log))
 	}
 
+	if cfg.P2P {
+		privKeyStr, _ := os.LookupEnv("P2P_PRIVATE_KEY")
```
joshklop, line 133: Is it possible to pass this in as a flag? Not sure if that's worse from a security point of view.

joshklop (approved):

node/node.go
omerfirmak: We can add a flag for using keystore files in the future.

IronGauntlets (approved):

PR #866, opened 2023-06-27
Remove Juno-specific code from l1 package

joshklop:


omerfirmak (approved):
Why is MakeClient juno specific?

joshklop:
Someone can implement the `l1.Subscriber` interface themselves use the `l1.Client` without it.

Maybe a more appropriate commit message would have been "Reduce API surface of l1 package". I'll change it.

omerfirmak:
> Someone can implement the `l1.Subscriber`

that is why we still have, `NewClient`, but I dont mind moving `MakeClient` either

PR #872, opened 2023-06-29
Add special contracts support

IronGauntlets:


core/state.go
```diff
@@ -291,6 +304,26 @@ func (s *State) updateContracts(stateTrie *trie.Trie, blockNumber uint64, diff *
...
+				err = s.putNewContract(stateTrie, &addr, specialContractsClassHash, blockNumber)
+				if err != nil {
+					return err
+				}
+			}
+		}
+	}
+	return nil
+}
+
```
omerfirmak, line 326: a alternative is;

```golang
// updateContractStorage applies the diff set to the Trie of the
// contract at the given address in the given Txn context.
func (s *State) updateContractStorage(stateTrie *trie.Trie, addr *felt.Felt, diff []StorageDiff, onChanged OnValueChanged) error {
	contract, err := NewContract(addr, s.txn)
	if err != nil {
                if !errors.Is(err, ErrContractNotDeployed) || (addr is not special) {
		      return err
                }

                err = s.putNewContract(stateTrie, &addr, specialContractsClassHash, blockNumber)
		if err != nil {
		     return err
		}
	}

```

core/state.go
```diff
@@ -274,6 +282,11 @@ func (s *State) updateContracts(stateTrie *trie.Trie, blockNumber uint64, diff *
 		}
 	}
 
+	err := s.checkAndDeploySpecialContracts(stateTrie, blockNumber, diff.StorageDiffs)
+	if err != nil {
+		return err
+	}
+
```
omerfirmak, line 289: We should do this in reverse in `State.Revert`

omerfirmak (changes requested):

core/state.go
IronGauntlets: done

core/state.go
IronGauntlets: Done

core/state.go
```diff
@@ -356,10 +364,26 @@ func (s *State) Class(classHash *felt.Felt) (*DeclaredClass, error) {
...
 // updateContractStorage applies the diff set to the Trie of the
 // contract at the given address in the given Txn context.
-func (s *State) updateContractStorage(stateTrie *trie.Trie, addr *felt.Felt, diff []StorageDiff, onChanged OnValueChanged) error {
+func (s *State) updateContractStorage(stateTrie *trie.Trie, addr *felt.Felt, diff []StorageDiff, onChanged OnValueChanged,
+	blockNumber uint64,
+) error {
 	contract, err := NewContract(addr, s.txn)
 	if err != nil {
-		return err
+		if _, ok := specialContracts[*addr]; !errors.Is(err, ErrContractNotDeployed) || !ok {
```
joshklop, line 372: nit: personally I think the code is easier to follow if the two cases are separated. Then we don't need to do a lookup in the map if `err` is not `ErrContractNotDeployed`.

```go
if !errors.Is(err, ErrContractNotDeployed) { return err }

if _, ok := specialContracts[*addr]; !ok { return err }
```

core/state_test.go
```diff
@@ -526,6 +571,51 @@ func TestRevert(t *testing.T) {
...
+	su1, err := gw.StateUpdate(context.Background(), 1)
+	require.NoError(t, err)
+
+	// These value were taken from part of integration state update number 299762
+	// https://external.integration.starknet.io/feeder_gateway/get_state_update?blockNumber=299762
+	scKey := utils.HexToFelt(t, "0x492e8")
+	scValue := utils.HexToFelt(t, "0x10979c6b0b36b03be36739a21cc43a51076545ce6d3397f1b45c7e286474ad5")
+	scAddr := new(felt.Felt).SetUint64(1)
+
+	t.Run("revert state update with special contract", func(t *testing.T) {
```
joshklop, line 602: Do we need the subtest?

core/state_test.go
```diff
@@ -526,6 +571,51 @@ func TestRevert(t *testing.T) {
...
+		// update state root
+		su1.NewRoot = utils.HexToFelt(t, "0x2829ac1aea81c890339e14422fe757d6831744031479cf33a9260d14282c341")
```
joshklop, line 604: Was this calculated by hashing `su0.NewRoot` with `0x1`?

IronGauntlets: Yes, I calculated it by running the test with `0x1` and when it failed I copied the root it failed with and updated the `NewRoot` (assumption: our trie's root calculation is correct).

core/state.go
IronGauntlets: I did have it like this initially but I thought the current once followed @omerfirmak suggestion more closely. Given we won't do a lookup, I will change ti back

core/state_test.go
IronGauntlets: No, I thought I would need to write another test case but it was not required.

core/state_test.go
IronGauntlets: No, on line 606 I update the storage diff of `su1` with `0x1` and without updating the root `Update` would fail, as expected. Thus I use the root `Update` fails with as `NewRoot`. I make the assumption that `Update` is correct and `Trie` correctly propagates values.

core/state.go
IronGauntlets: done

core/state_test.go
IronGauntlets: Done

core/state.go
```diff
@@ -492,6 +520,24 @@ func (s *State) Revert(blockNumber uint64, update *StateUpdate) error {
...
 	}
 
+	// purge noClassContracts
+	//
+	// As noClassContracts are not in StateDiff.DeployedContracts we can only purge them if their storage no longer exists.
+	// Updating contracts with reverse diff will eventually lead to the deletion of noClassContract's storage key from db. Thus,
+	// we can use the lack of key's existence as reason for purging noClassContracts.
+
+	for addr := range noClassContracts {
+		err := s.txn.Get(db.ContractStorage.Key(addr.Marshal()), func(val []byte) error {
```
omerfirmak, line 530: This leaks `contract` package implementation. Can we use `if contract.Root() == felt.Zero` instead?

omerfirmak (approved):

core/state.go
IronGauntlets: Done.

Although I don't think its `leakage` as they are in the same package.

PR #885, opened 2023-07-05
Calculate class hash for declare transactions

joshklop:


vm/rust/src/lib.rs
```diff
@@ -166,6 +168,36 @@ pub extern "C" fn cairoVMExecute(
...
+                            transaction_hash: declare.transaction_hash(),
+                            signature: declare.signature(),
+                            nonce: declare.nonce(),
+                            sender_address: declare.sender_address(),
+                            class_hash,
+                        }))
+                    },
+                    _ => sn_api_txn,
+                };
+
```
omerfirmak, line 200: We need class hash to be able to calculate the transaction hash. Which we do on Go side. So Go should be able to calculate the class hash before calling cairoVmExecute.

See how I wrapped the rust code to be called from Go 

https://github.com/NethermindEth/juno/pull/809/commits/a5a6edaef57678f83ebc3479ca94f545f6bd7534

and how it is used

https://github.com/NethermindEth/juno/pull/809/commits/a6f60267fb82e0c34fd21968bdb754bd16e41c63#diff-14aba5c63ad044ac1e8a62d96fed2cd132ad5e4935fc555a935a40c14f5cb0f7R156-R169

omerfirmak:
Closing since the issues is fixed on the `estimate-fee` branch

PR #887, opened 2023-07-05
Recursively validate structs in arrays and maps

omerfirmak:


IronGauntlets:
Can you please give some more context for this PR?

omerfirmak:
to be able to validate 

> Can you please give some more context for this PR?

needed to get jsonrpc package to validate arguments that are not plain structs but array of structs for example.

```
func (h *Handler) EstimateFee(broadcastedTxns []BroadcastedTransaction, id BlockID) ([]FeeEstimate, *jsonrpc.Error) {
```

for instance, to be able to validate each BroadcastedTransaction in EstimateFee

jsonrpc/server_test.go
```diff
@@ -105,15 +105,15 @@ func TestHandle(t *testing.T) {
 		{
 			"validation",
 			[]jsonrpc.Parameter{{Name: "param"}},
-			func(v validationStruct) (int, *jsonrpc.Error) {
-				return v.A, nil
+			func(v []validationStruct) (int, *jsonrpc.Error) {
+				return v[0].A, nil
```
IronGauntlets, line 109: Why not add this as a separate test case? Surely, we want the previous case to still work, right?

jsonrpc/server_test.go
```diff
@@ -105,15 +105,15 @@ func TestHandle(t *testing.T) {
...
 			},
 		},
 		{
 			"validationPointer",
 			[]jsonrpc.Parameter{{Name: "param"}},
-			func(v *validationStruct) (int, *jsonrpc.Error) {
-				return v.A, nil
+			func(v map[string]*validationStruct) (int, *jsonrpc.Error) {
+				return v["expectedkey"].A, nil
```
IronGauntlets, line 116: same as above

jsonrpc/server_test.go
omerfirmak: because it covers all the cases, including the original one.

jsonrpc/server.go
```diff
@@ -354,14 +354,36 @@ func (s *Server) parseParam(param any, t reflect.Type) (reflect.Value, error) {
...
+	} else if kind == reflect.Map {
+		for _, key := range param.MapKeys() {
+			if err := s.validateParam(param.MapIndex(key)); err != nil {
+				return err
+			}
+		}
+	}
+
+	return nil
+}
```
IronGauntlets, line 389: `switch` statement?

jsonrpc/server_test.go
IronGauntlets: That is making the assumption about the code under test. 

Please add it as separate cases. I think it is useful to have these cases separately so that in future it is easier to identity which case failed more easily.

IronGauntlets (approved):

PR #889, opened 2023-07-06
Pass in jsonrpc.Server to HTTP and Websocket bridges

joshklop:
Only register the validator and the methods once when possible.

### DIFF
--- a/jsonrpc/http.go
+++ b/jsonrpc/http.go
@@ -21,14 +21,7 @@ type HTTP struct {
 	listener net.Listener
 }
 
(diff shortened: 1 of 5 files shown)


IronGauntlets:
I am not sure we get much benefit from sharing the same rpc server between http and web sockets.

IronGauntlets:
Also, before this change, the same validator was passed to both the HTTP server and web socket.

omerfirmak (approved):

omerfirmak:
@IronGauntlets it's just neater IMO.

IronGauntlets (approved):

PR #895, opened 2023-07-09
Update types for 0.12.1

joshklop:
Needed for #878. See commit message for details.

### DIFF
--- a/clients/feeder/transaction.go
+++ b/clients/feeder/transaction.go
@@ -1,8 +1,48 @@
 package feeder
 
 import (
+	"errors"
+
 	"github.com/NethermindEth/juno/core/felt"
 )
+
+type ExecutionStatus uint8
+
+const (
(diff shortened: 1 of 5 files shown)


clients/feeder/transaction.go
```diff
@@ -1,9 +1,49 @@
...
+		return errors.New("unknown ExecutionStatus")
+	}
+	return nil
+}
+
+type FinalityStatus uint8
+
+const (
+	AcceptedOnL2 FinalityStatus = iota + 1
+	AcceptedOnL1
```
omerfirmak, line 32: Do we need `Received` as well?

joshklop: Based on my understanding, no we do not. It's used internally by the feeder gateway for transactions that haven't been `REJECTED` or `ACCEPTED_*` yet. That's technically already used and we haven't needed it.

PR #896, opened 2023-07-09
New method starknet_traceTransaction

kirugan:


omerfirmak:
Rebase please

rpc/handlers.go
```diff
@@ -1109,12 +1110,13 @@ func (h *Handler) EstimateFee(broadcastedTxns []BroadcastedTransaction, id Block
 		blockNumber = height + 1
 	}
 
-	gasesConsumed, err := vm.Execute(txns, classes, blockNumber, header.Timestamp, header.SequencerAddress, state, h.network)
+	gasesConsumed, executionInfo, err := vm.Execute(txns, classes, blockNumber, header.Timestamp, header.SequencerAddress, state, h.network)
```
kirugan, line 1113: it's temporary to debug execution info

rpc/trace.go
```diff
@@ -0,0 +1,78 @@
+package rpc
+
+import (
+	"github.com/NethermindEth/juno/core/felt"
+	"github.com/NethermindEth/juno/utils"
+	"github.com/NethermindEth/juno/vm"
+)
+
+type TransactionTrace struct {
+	// todo add the rest of types
```
omerfirmak, line 10: Why are we missing the rest?

kirugan: I've been thinking how to properly declare enum of different transaction types (choice between struct with same field + omitempty or private interface). I decided to use the latter.

rpc/handlers.go
```diff
@@ -1120,3 +1123,49 @@ func (h *Handler) EstimateFee(broadcastedTxns []BroadcastedTransaction, id Block
...
+		return nil, jsonrpc.Err(jsonrpc.InternalError, err.Error())
+	}
+	tx, err := h.bcReader.TransactionByHash(&hash)
+	if err != nil {
+		return nil, jsonrpc.Err(jsonrpc.InternalError, err.Error())
+	}
+
+	id := &BlockID{Number: blockNumber - 1}
+
+	state, closer, err := h.stateByBlockID(id)
```
omerfirmak, line 1139:
```suggestion
	state, closer, err := h.bcReader.StateAtBlockNumber(blockNumber -1)
```

rpc/handlers.go
```diff
@@ -1120,3 +1123,49 @@ func (h *Handler) EstimateFee(broadcastedTxns []BroadcastedTransaction, id Block
...
+	}
+
+	var classes []core.Class
+	if tx, ok := tx.(*core.DeclareTransaction); ok {
+		class, stateErr := state.Class(tx.ClassHash)
+		if stateErr != nil {
+			return nil, jsonrpc.Err(jsonrpc.InternalError, stateErr.Error())
+		}
+		classes = append(classes, class.Class)
+	}
```
omerfirmak, line 1161: This wont work. You need to query classes from the head state. Because `tx.ClassHash` wont be in the state for the block with number `blockNumber - 1` since it was declared on `blockNumber`

rpc/handlers.go
```diff
@@ -1120,3 +1123,49 @@ func (h *Handler) EstimateFee(broadcastedTxns []BroadcastedTransaction, id Block
...
+
+	_, info, err := h.vm.Execute([]core.Transaction{tx}, classes, header.Number, header.Timestamp, header.SequencerAddress, state, h.network)
+	if err != nil {
+		err = fmt.Errorf("vm.Execute: %w", err)
+		return nil, jsonrpc.Err(jsonrpc.InternalError, err.Error())
```
omerfirmak, line 1166:
```suggestion
                rpcErr := *ErrContractError
                rpcErr.Data = err.Error()
                return nil, &rpcErr
```

Let's do what EstimateFee does

rpc/handlers.go
```diff
@@ -1120,3 +1123,49 @@ func (h *Handler) EstimateFee(broadcastedTxns []BroadcastedTransaction, id Block
...
+	var classes []core.Class
+	if tx, ok := tx.(*core.DeclareTransaction); ok {
+		class, stateErr := state.Class(tx.ClassHash)
+		if stateErr != nil {
+			return nil, jsonrpc.Err(jsonrpc.InternalError, stateErr.Error())
+		}
+		classes = append(classes, class.Class)
+	}
+
+	_, info, err := h.vm.Execute([]core.Transaction{tx}, classes, header.Number, header.Timestamp, header.SequencerAddress, state, h.network)
```
omerfirmak, line 1163: You have to execute the entire block, not just the `txn`. So you need fetch the entire block with `blockByID`

rpc/trace.go
```diff
@@ -0,0 +1,143 @@
...
+
+	"github.com/NethermindEth/juno/core"
+	"github.com/NethermindEth/juno/core/felt"
+	"github.com/NethermindEth/juno/jsonrpc"
+	"github.com/NethermindEth/juno/utils"
+	"github.com/NethermindEth/juno/vm"
+)
+
+type TransactionTrace interface {
+	trace()
```
omerfirmak, line 14: We dont need separate types for each transaction.

Let's just use something like

```
type TransactionTrace struct {
	ValidateInvocation    *FunctionInvocation `json:"validate_invocation,omitempty"`
	ExecuteInvocation     *ExecuteInvocation  `json:"execute_invocation,omitempty"`
	FeeTransferInvocation *FunctionInvocation `json:"fee_transfer_invocation,omitempty"`
}
```

and some of the fields can be `nil` according to the transaction type

omerfirmak (changes requested):

rpc/handlers.go
kirugan: Agree sounds reasonable, but why does it work with current implementation:
```
curl --location --request POST 'http://localhost:6060' \
--data-raw '{
    "jsonrpc": "2.0",
    "id": 1,
    "method": "starknet_traceTransaction",
    "params": {
        "transaction_hash": "0x55ba350e706f02c5903cc62f2381d6a7cc6f88efeb9c72a230487fec5f60a50"
    }
}'
```
on commit 7de8ce6a9c8bfeafe04fa5152599a04257382eca, I logged locally `append(classes, class.Class)` logic - it seems working, maybe not the way expected 😄

kirugan: Probably it worked because class is declared at 4883 block and tx is from 4968

rpc/handlers.go
kirugan: Fixed, but now it doesn't work :) the problem that if block contains transaction type that unsupported by `vm.marshalTxn(...)`  then entire execution is failed. Could you please explain why do I need entire block to execute? because a transaction that I requested can rely on another in the same block?

rpc/handlers.go
```diff
@@ -1120,3 +1123,48 @@ func (h *Handler) EstimateFee(broadcastedTxns []BroadcastedTransaction, id Block
...
+	_, _, blockNumber, err := h.bcReader.Receipt(&hash)
+	if err != nil {
+		return nil, jsonrpc.Err(jsonrpc.InternalError, err.Error())
+	}
+	tx, err := h.bcReader.TransactionByHash(&hash)
+	if err != nil {
+		return nil, jsonrpc.Err(jsonrpc.InternalError, err.Error())
+	}
+
+	state, closer, err := h.bcReader.StateAtBlockNumber(blockNumber -1)
```
kirugan: @omerfirmak what if there is a request for tx that resides in 0 block? what should we do in this case

rpc/handlers.go
omerfirmak: >  Probably it worked because class is declared at 4883 block and tx is from 4968

Yeap, indeed. It is just redeclaring a class hash. Which is fine.

rpc/handlers.go
omerfirmak: > because a transaction that I requested can rely on another in the same block?

Yes, exactly.

> if block contains transaction type that unsupported by vm.marshalTxn(...)

Let's add support for them then:)

rpc/handlers.go
omerfirmak: Yes, that is an issue. Let's add a special case in `StateAtBlockNumber`. 

```
var EmptyState := core.NewState(db.NewMemTransaction)

if blockNumber == 0 {
   return EmptyState
}
```

omerfirmak: and in trace handler, we can just use `h.bcReader.StateAtBlockHash(block.ParentHash)` instead of `h.bcReader.StateAtBlockNumber(blockNumber -1)`

rpc/handlers.go
```diff
@@ -1120,3 +1103,71 @@ func (h *Handler) EstimateFee(broadcastedTxns []BroadcastedTransaction, id Block
...
+
+		class, stateErr := headState.Class(declareTx.ClassHash)
+		if stateErr != nil {
+			return nil, jsonrpc.Err(jsonrpc.InternalError, stateErr.Error())
+		}
+		classes = append(classes, class.Class)
+	}
+
+	header := block.Header
+	_, info, err := h.vm.Execute(block.Transactions, classes, header.Number, header.Timestamp, header.SequencerAddress, state, h.network)
```
omerfirmak, line 1158: We need to check for pending block explicitly, since header.Number is not valid for pending blocks. See how we do that in `EstimateFee`

kirugan: Added support

rpc/handlers.go
kirugan: Added

rpc/trace.go
kirugan: Changed type from interface to struct

rpc/handlers.go
kirugan: It was easier than I expected 😄

rpc/trace.go
omerfirmak: Can we make them pointers as well? So it is easier to check if they are empty or no

grpc/gen/kv.pb.go
omerfirmak, line 1: can you remove these changes in `grpc` and `l1` since they are not-related

rpc/trace.go
```diff
@@ -0,0 +1,111 @@
...
+	ClassHash      *felt.Felt           `json:"class_hash,omitempty"`
+	EntryPointType EntryPointType       `json:"entry_point_type,omitempty"`
+	CallType       CallType             `json:"call_type,omitempty"`
+	Result         []felt.Felt          `json:"result,omitempty"`
+	Calls          []FunctionInvocation `json:"calls,omitempty"`
+	Events         []EventContent       `json:"events,omitempty"`
+	Messages       []MsgToL1            `json:"msg_to_l1,omitempty"`
+}
+
+// enum: EXTERNAL, L1_HANDLER, CONSTRUCTOR
```
omerfirmak, line 39: is this a directive to compiler or just a regular comment?

kirugan: Just a regular comment

grpc/gen/kv.pb.go
kirugan: Sure, there were some issues with file watcher in GoLand IDE. Fixed

rpc/trace.go
kirugan: Agree, fixed

vm/vm.go
```diff
@@ -127,13 +150,20 @@ func Execute(txns []core.Transaction, declaredClasses []core.Class, blockNumber,
...
 
-	sequencerAddressBytes := sequencerAddress.Bytes()
+	var address *felt.Felt
+	if sequencerAddress != nil {
+		address = sequencerAddress
+	} else {
+		address = core.NetworkBlockHashMetaInfo(network).FallBackSequencerAddress
+	}
+	sequencerAddressBytes := address.Bytes()
+
```
omerfirmak, line 166: can we move this to caller site? it feels like it doesnt belong here

vm/rust/src/lib.rs
```diff
@@ -205,12 +205,56 @@ pub extern "C" fn cairoVMExecute(
...
+            ).map_err(|e| e.to_string())?;
+            Ok(Transaction::AccountTransaction(AccountTransaction::Declare(declare_tx)))
+        }
+        StarknetApiTransaction::DeployAccount(deploy_account) => {
+            Ok(Transaction::AccountTransaction(AccountTransaction::DeployAccount(deploy_account)))
+        }
+        StarknetApiTransaction::Invoke(invoke) => {
+            Ok(Transaction::AccountTransaction(AccountTransaction::Invoke(invoke)))
+        }
+        StarknetApiTransaction::Deploy(deploy) => {
```
omerfirmak, line 244: Let's not copy and paste `Transaction::from_api` here. 

Instead let's check for the cases that `Transaction::from_api` panic and make sure we error out early don't call it in those cases.

rpc/trace.go
```diff
@@ -0,0 +1,111 @@
...
+
+	"github.com/NethermindEth/juno/core"
+	"github.com/NethermindEth/juno/core/felt"
+	"github.com/NethermindEth/juno/jsonrpc"
+	"github.com/NethermindEth/juno/utils"
+	"github.com/NethermindEth/juno/vm"
+)
+
+type TransactionTrace struct {
+	ValidateInvocation    *FunctionInvocation `json:"validate_invocation,omitempty"`
```
omerfirmak, line 14: Do you think we can just use `vm` Types for this?
This looks way too much code to duplicate

kirugan: Unfortunately no, because `CallInfo` has a structure that depends on format of rust-vm (and naming of json fields) and `FunctionInvocation` type is format of json-rpc. For instance:
```
result.Result = info.Execution.Retdata
```
json rpc field `result` is actually `retdata` inside other struct in vm response

omerfirmak: Can we fix stuff like this in rust types?

For example instead of 

``` rust
#[derive(Debug,Serialize)]
pub struct CallExecution {
    pub retdata: Retdata,
    pub events: Vec<OrderedEvent>,
    pub l2_to_l1_messages: Vec<OrderedL2ToL1Message>,
}
```

can we rename `retdata` to `result` here?

We have 3 almost-exactly same type definitions, 1 in Rust, 2 in Go. If we can at least get rid of one of them. It would be already an huge improvement.

vm/vm.go
kirugan: Moved

rpc/trace.go
omerfirmak: I wouldnt mind even if we straight up used the JSON that we return from rust. Without unmarshaling/marshaling on Go side.

vm/rust/src/lib.rs
kirugan: Ok I added call to `Transaction::from_api`, also decided to make things clean to keep validation in existing function `transaction_from_api(...)`

rpc/handlers.go
```diff
@@ -1120,3 +1103,88 @@ func (h *Handler) EstimateFee(broadcastedTxns []BroadcastedTransaction, id Block
...
+	}
+
+	header := block.Header
+
+	var sequencerAddress *felt.Felt
+	if header.SequencerAddress != nil {
+		sequencerAddress = header.SequencerAddress
+	} else {
+		sequencerAddress = core.NetworkBlockHashMetaInfo(h.network).FallBackSequencerAddress
+	}
```
omerfirmak, line 1173: nit

```suggestion
	sequencerAddress := header.SequencerAddress
	if sequencerAddress == nil {
		sequencerAddress = core.NetworkBlockHashMetaInfo(h.network).FallBackSequencerAddress
	}
```

kirugan: Thank you, it's much cleaner now

rpc/handlers.go
```diff
@@ -1150,3 +1133,88 @@ func (h *Handler) EstimateFee(broadcastedTxns []BroadcastedTransaction, id Block
...
```
omerfirmak, line 1203: this got reverted for some reason, can you make sure nothing else is mistakenly reverted as well?

vm/rust/src/lib.rs
```diff
@@ -209,12 +209,45 @@ pub extern "C" fn cairoVMExecute(
...
     }
 }
 
+fn transaction_from_api(
+    tx: StarknetApiTransaction,
+    contract_class: Option<ContractClass>,
+) -> Result<Transaction, String> {
+    match tx {
+        StarknetApiTransaction::Deploy(deploy) => {
+            return Err(format!("Deploy transaction is not supported (transaction_hash={})", deploy.transaction_hash))
```
jelilat, line 229: This line is causing the following error even in cases where the transaction is not a deploy transaction. 
```
"error":{"code":40,"message":"Contract error","data":"Deploy transaction is not supported 
```

Looking at it, I realised that the error shows up in blocks that contain deploy transactions even though the transaction in question is not a Deploy transaction. 

Do we need to go through the whole block to get the transaction trace for a single transaction?

omerfirmak: > Do we need to go through the whole block to get the transaction trace for a single transaction?

Yes, because transactions get effected from state generated by the previous transactions in the same block. For trace to be accurate we need to execute the entire block

jelilat: Okay

rpc/handlers.go
```diff
@@ -1174,3 +1155,83 @@ func (h *Handler) EstimateMessageFee(msg MsgFromL1, id BlockID) (*FeeEstimate, *
...
 }
+
+// TraceTransaction returns the trace for a given executed transaction, including internal calls
+//
+// It follows the specification defined here:
+// https://github.com/starkware-libs/starknet-specs/blob/1ae810e0137cc5d175ace4554892a4f43052be56/api/starknet_trace_api_openrpc.json#L11
+func (h *Handler) TraceTransaction(hash felt.Felt) (json.RawMessage, *jsonrpc.Error) {
+	_, blockHash, blockNumber, err := h.bcReader.Receipt(&hash)
+	if err != nil {
+		return nil, jsonrpc.Err(jsonrpc.InternalError, err.Error())
```
omerfirmak, line 1166: should be a ErrTxnHashNotFound

rpc/handlers.go
```diff
@@ -1174,3 +1155,83 @@ func (h *Handler) EstimateMessageFee(msg MsgFromL1, id BlockID) (*FeeEstimate, *
...
+	state, closer, err := h.bcReader.StateAtBlockHash(block.ParentHash)
+	if err != nil {
+		return nil, ErrBlockNotFound
+	}
+	defer h.callAndLogErr(closer, "Failed to close state in starknet_traceTransaction")
+
+	headState, headStateCloser, err := h.bcReader.HeadState()
+	if err != nil {
+		return nil, jsonrpc.Err(jsonrpc.InternalError, err.Error())
+	}
```
omerfirmak, line 1183: One thing I didnt think of before,

If txn `hash` is part of the pending block, we need to use `PendingState` here, instead of `HeadState`

vm/vm.go
```diff
@@ -50,6 +50,7 @@ type callContext struct {
 	response []*felt.Felt
 	// amount of gas consumed per transaction during VM execution
 	gasConsumed []*felt.Felt
+	traces      []string
```
omerfirmak, line 53:
```suggestion
	traces      []json.RawMessage
```

so we dont need to convert it when returning from `vm.Execute`?


``` golang
	var traces []json.RawMessage
	for _, trace := range context.traces {
		traces = append(traces, json.RawMessage(trace))
	}
```

utils/slices.go
```diff
@@ -1,10 +1,34 @@
...
+
+func Filter[T any](slice []T, f func(T) bool) []T {
+	var result []T
+	for _, e := range slice {
+		if f(e) {
+			result = append(result, e)
+		}
+	}
+
+	return result
```
omerfirmak, line 33: do we still need these?

vm/rust/src/lib.rs
```diff
@@ -242,12 +241,45 @@ pub extern "C" fn cairoVMExecute(
                 JunoAppendGasConsumed(
                     reader_handle,
                     felt_to_byte_array(&t.actual_fee.0.into()).as_ptr(),
-                )
+                );
+
+                // let info = execution_info::TransactionExecutionInfo::from(t);
```
omerfirmak, line 246: leftover?

utils/slices.go
kirugan: I left these helpers for future use. They quite helpful

rpc/handlers.go
```diff
@@ -1174,3 +1155,91 @@ func (h *Handler) EstimateMessageFee(msg MsgFromL1, id BlockID) (*FeeEstimate, *
...
+
+	_, traces, err := h.vm.Execute(block.Transactions, classes, blockNumber, header.Timestamp,
+		sequencerAddress, state, h.network, []*felt.Felt{})
+	if err != nil {
+		rpcErr := *ErrContractError
+		rpcErr.Data = err.Error()
+		return nil, &rpcErr
+	}
+	trace := traces[txIndex]
```
joshklop, line 1236: Do we need to re-execute the whole block, or just `block.Transactions[:txIndex]`?

omerfirmak: > or just block.Transactions[:txIndex]?

Yeap, nice catch.

rpc/handlers.go
kirugan: yep, thanks

rpc/handlers.go
joshklop: That would make sense to me. If class isn't present in the state before `txIndex`, we shouldn't need to include it.

kirugan: > we shouldn't need to include it.

added

omerfirmak (approved):
LGTM, let's wait for @joshklop's approval as well.

blockchain/blockchain.go
```diff
@@ -715,6 +715,12 @@ func (b *Blockchain) StateAtBlockNumber(blockNumber uint64) (core.StateReader, S
 
 // StateAtBlockHash returns a StateReader that provides a stable view to the state at the given block hash
 func (b *Blockchain) StateAtBlockHash(blockHash *felt.Felt) (core.StateReader, StateCloser, error) {
+	if blockHash.IsZero() {
```
joshklop, line 718: Maybe the caller should create the empty state?

joshklop (approved):
The code looks good to me. Have we done manual tests for this yet?

blockchain/blockchain.go
kirugan: Hm.. not sure, we would implement it over and over again on caller side I guess. @omerfirmak wdyt?

omerfirmak: > the caller should create the empty state

I think it fits well in the API of `blockchain.Reader`, we can keep it here.
We can make `emptyState` a singleton tho, to avoid instantiating it every time. But it's not a priority either.

kirugan: > We can make emptyState a singleton tho, to avoid instantiating it every time. But it's not a priority either.

Good point I'll add it to next PR

PR #897, opened 2023-07-10
Basic documentation scaffold

joshklop:
Just getting the basic structure down for now. The styling/branding with images, etc. will take more time. We can work on the branding and content in parallel now that the structure is in place.

I can create a tracking issue for docs improvements once this is merged.


kirugan:
Overall looks good, but how much of it is autogenerated ? I hope we don't need to write React components ourselves

joshklop:
Almost all of it is auto-generated. We should only have to write markdown for docs.

omerfirmak (approved):
I am fine with merging this. Up to you @joshklop

PR #906, opened 2023-07-11
Add Migration interface

### LINKED ISSUE #884: Add `Migration` interface to migration package
Right now we use a function type for all migrations that takes a single `db.Transaction` argument. However, some migrations require additional state across calls. Turning this function type into a single-method interface should solve the problem, since we can preserve state in a struct instead of a package-level variable.

joshklop:
Closes #884. See that issue and the commit messages for details.


migration/migration.go
```diff
@@ -49,23 +62,30 @@ func MigrateIfNeeded(targetDB db.DB) error {
...
-			if err = revisions[i](txn); err != nil {
-				if errors.Is(err, ErrCallWithNewTransaction) {
-					// rerun the same revision
-					i--
-					return nil
+	for i := version; i < uint64(len(migrations)); i++ {
+		migration := migrations[i]
+		migration.Before()
+		var done bool
+		for !done {
```
omerfirmak, line 69: nit: remove `done` and just `break` after `targetDB.Update` is successful

omerfirmak (approved):
LGTM

PR #907, opened 2023-07-12
Debug log the incoming RPC requests

omerfirmak:


jsonrpc/server.go
```diff
@@ -253,7 +257,11 @@ func isNil(i any) bool {
 }
 
 func (s *Server) handleRequest(req *request) (*response, error) {
-	if err := req.isSane(); err != nil {
+	reqJSON, err := json.Marshal(req)
+	if err == nil {
+		s.log.Debugw("Serving RPC request", "request", string(reqJSON))
```
IronGauntlets, line 262: Instead of `Debug` I think we should make it Info. Also, I think we should do 2 logs one where we just received the request and one where we are about to respond to the request.

IronGauntlets (changes requested):

jsonrpc/server.go
omerfirmak: > we just received the request and one where we are about to respond to the request.

do you mean log the response that we generated?

IronGauntlets: No to the response, unless you feel it is important. I mean to say that we should have some like the following:

`INFO received rpc request string(reqJson)`
`INFO responding to rpc request string(reqJson)`

Something like that, that would also allow us to have some measure of how long an rpc took.

omerfirmak: Done

IronGauntlets (approved):

PR #908, opened 2023-07-12
Add missing From field to rpc.MsgToL1

omerfirmak:


IronGauntlets (approved):
Suprised we missed this

PR #915, opened 2023-07-13
Remove nil gas price check from EstimateFee

IronGauntlets:


joshklop (commented):
Love the change, but not quite understanding. Why can we do this? Can users upgrade with an old db and expect `estimateFee` on old txs to work?

IronGauntlets:
We had this if statement because we did not want to write a migration and it was done for testing purposes.

We have a warning in the v0.4.0 release that users need to re-sync we can remove this if statement. So now we can remove it.

joshklop (approved):

PR #916, opened 2023-07-14
Adjust worker number for sync process dynamically

### LINKED ISSUE #859: Scale down our fetcher and verifier workers in sync process after reaching HEAD
We fetch multiple blocks in parallel during sync. After reaching the HEAD, this is only putting extra load on the gateway with no benefit on our side.

We should be able to disable/enable/scale the level of parallelism in block fetch and verification process.

omerfirmak:
scale them up or down according to if we think we are at the HEAD or not

Fixes #859

### DIFF
--- a/sync/sync.go
+++ b/sync/sync.go
@@ -28,6 +28,8 @@ type Synchronizer struct {
 	log utils.SimpleLogger
 
 	pendingPollInterval time.Duration
+
+	catchUpMode bool
 }
 
 func New(bc *blockchain.Blockchain, starkNetData starknetdata.StarknetData,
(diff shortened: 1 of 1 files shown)


sync/sync.go
```diff
@@ -159,12 +161,18 @@ func (s *Synchronizer) verifierTask(ctx context.Context, block *core.Block, stat
...
 
-			if s.HighestBlockHeader == nil || s.HighestBlockHeader.Number < block.Number {
+			if s.HighestBlockHeader == nil || s.HighestBlockHeader.Number <= block.Number {
 				highestBlock, err := s.StarknetData.BlockLatest(ctx)
 				if err != nil {
 					s.log.Warnw("Failed fetching latest block", "err", err)
 				} else {
 					s.HighestBlockHeader = highestBlock.Header
+
+					isBehind := s.HighestBlockHeader.Number > block.Number
```
kirugan, line 171: Can `s.HighestBlockHeader` be nil here?

omerfirmak: Nope

kirugan (approved):

PR #917, opened 2023-07-14
Make parts of SanityCheckNewHeight run in parallel

omerfirmak:
Before

`BenchmarkMainnetBlock           10         754074777 ns/op         3112717 B/op     118761 allocs/op`

After

`BenchmarkMainnetBlock           10         424905598 ns/op         3677447 B/op     140130 allocs/op`

for some blocks sanity checks take even longer(>1sec), so this should help with our latency when tracking the HEAD

### DIFF
--- a/core/transaction.go
+++ b/core/transaction.go
@@ -4,7 +4,9 @@ import (
 	"encoding/binary"
 	"errors"
 	"fmt"
+	"runtime"
 	"strings"
+	"sync"
 
 	"github.com/Masterminds/semver/v3"
 	"github.com/NethermindEth/juno/core/crypto"
@@ -13,6 +15,7 @@ import (
 	"github.com/NethermindEth/juno/utils"
 	"github.com/bits-and-blooms/bloom/v3"
 	"github.com/ethereum/go-ethereum/common"
+	"github.com/sourcegraph/conc/pool"
 )
 
 type Event struct {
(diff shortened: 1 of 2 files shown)


core/transaction.go
```diff
@@ -428,20 +431,45 @@ func ParseBlockVersion(protocolVersion string) (*semver.Version, error) {
...
+					for idx := curReceiptIdx; idx < curReceiptIdx+receiptPerWorker && idx < len(receipts); idx++ {
+						for _, event := range receipts[idx].Events {
+							eventHash := crypto.PedersenArray(
+								event.From,
+								crypto.PedersenArray(event.Keys...),
+								crypto.PedersenArray(event.Data...),
+							)
+
+							trieMutex.Lock()
+							_, err := trie.Put(new(felt.Felt).SetUint64(curEventIdx), eventHash)
```
kirugan, line 458: microoptimization: compute `new(felt.Felt).SetUint64(...)` before acquiring the lock?

core/transaction.go
```diff
@@ -428,20 +431,45 @@ func ParseBlockVersion(protocolVersion string) (*semver.Version, error) {
...
+		var trieMutex sync.Mutex
+
+		for receiptIdx := range receipts {
+			if receiptIdx%receiptPerWorker == 0 {
+				curReceiptIdx := receiptIdx
+				curEventIdx := eventCount
+
+				workerPool.Go(func() error {
+					for idx := curReceiptIdx; idx < curReceiptIdx+receiptPerWorker && idx < len(receipts); idx++ {
+						for _, event := range receipts[idx].Events {
```
kirugan, line 450: Can we please use reslice so it would be easier to read i.e.:
```
maxIndex := curReceiptIdx + receiptPerWorker
if maxIndex > len(receipts) {
   maxIndex = len(receipts)
}
receipts := receipts[curReceiptIdx:maxIndex]

for _, receipt := range receipts {
	for _, event := range receipt.Events {
```

core/transaction.go
omerfirmak: Done

core/transaction.go
omerfirmak: Yeap, thanks. Done

kirugan (approved):

PR #918, opened 2023-07-16
Notify user that upgrade is available

joshklop:


kirugan:
Josh that's a really great idea 👍

joshklop:
I appreciate it, but I just copied the idea from pathfinder :)

joshklop:
Rebased.

node/node.go
```diff
@@ -115,14 +123,16 @@ func New(cfg *Config, version string) (*Node, error) {
 	if n.cfg.EthNode == "" {
 		n.log.Warnw("Ethereum node address not found; will not verify against L1")
 	} else {
-		ethNodeURL, err := url.Parse(n.cfg.EthNode)
+		var ethNodeURL *url.URL
+		ethNodeURL, err = url.Parse(n.cfg.EthNode)
```
omerfirmak, line 127: why the changes here? and similarly on L134 and L144

upgrader/upgrader.go
```diff
@@ -0,0 +1,106 @@
...
+			if err != nil {
+				u.log.Debugw("Failed to fetch latest release", "err", err)
+				continue
+			} else if resp.StatusCode != http.StatusOK {
+				u.log.Debugw("Failed to fetch latest release", "status", resp.Status)
+				continue
+			}
+
+			latest := new(Release)
+			if err := json.NewDecoder(resp.Body).Decode(&latest); err == nil {
```
omerfirmak, line 74: `latest` is a `*Release`, so you dont need to get a pointer to it here

```suggestion
			if err := json.NewDecoder(resp.Body).Decode(latest); err == nil {
```

upgrader/upgrader.go
```diff
@@ -0,0 +1,106 @@
...
+			}
+
+			timer.Reset(u.delay)
+			resp.Body.Close()
+		}
+	}
+}
+
+func needsUpdate(currentVersion, latestVersion semver.Version) bool {
+	if currentVersion.Major() == latestVersion.Major() {
```
omerfirmak, line 94: why can't we just use https://pkg.go.dev/github.com/Masterminds/semver#Version.GreaterThan

node/node.go
joshklop: Linter started complaining for some reason. Probably a bug in it's parsing logic.

upgrader/upgrader.go
joshklop: It doesn't work well when the version isn't exactly semver (as when building from `main`). 

Will add to test cases to make this clear, and add a comment.

omerfirmak (approved):

PR #919, opened 2023-07-16
Implement some metrics collection

omerfirmak:


jsonrpc/http.go
```diff
@@ -62,6 +75,7 @@ func (h *HTTP) ServeHTTP(writer http.ResponseWriter, req *http.Request) {
 	}
 
 	req.Body = http.MaxBytesReader(writer, req.Body, MaxRequestBodySize)
+	h.requests.Inc()
```
joshklop, line 78: Can we expose callbacks for certain events in the "rpc request/response lifecycle"? Otherwise we require dependents to use prometheus.

omerfirmak: I think I can live with that for now. If we expose callbacks, it will complicate the caller site unnecessarily IMO.

IronGauntlets, line 78: Why are we only incrementing and not decrementing when we have responded to the request?

sync/sync.go
```diff
@@ -28,17 +30,32 @@ type Synchronizer struct {
...
 		Blockchain:          bc,
 		StarknetData:        starkNetData,
 		log:                 log,
 		pendingPollInterval: pendingPollInterval,
+
+		opTimers: prometheus.NewHistogramVec(prometheus.HistogramOpts{
+			Namespace: "sync",
+			Name:      "timers",
+		}, []string{"op"}),
+		blockRate: prometheus.NewCounter(prometheus.CounterOpts{
```
IronGauntlets, line 52: It would be more appropriate to call this `TotalBlocks` since `rate` implies something related to time and this is just a counter.

sync/sync.go
```diff
@@ -133,7 +150,9 @@ func (s *Synchronizer) fetchUnknownClasses(ctx context.Context, stateUpdate *cor
 func (s *Synchronizer) verifierTask(ctx context.Context, block *core.Block, stateUpdate *core.StateUpdate,
 	newClasses map[felt.Felt]core.Class, resetStreams context.CancelFunc,
 ) stream.Callback {
+	timer := prometheus.NewTimer(s.opTimers.WithLabelValues("verify"))
```
IronGauntlets, line 153: lets add the labels as `const`, this applies to `store` and `fetch`

jsonrpc/http.go
omerfirmak: We are not trying to keep track of concurrent requests. We are simply counting incoming requests so that we can calculate the rate from that.

sync/sync.go
omerfirmak: Done

sync/sync.go
omerfirmak: I dont tend to do that when the constant is used only once. Since it is just an indirection that serves no purpose IMO. It is up to you, I don't mind changing it.

jsonrpc/http.go
joshklop: Just to make sure we're on the same page: I am proposing something like 

```go
func (s *Server) WithOnRequest(cb func()) *Server {
  s.onRequest = cb
  return s
}
```

This only needs to be set once on the main jsonrpc server, and we don't need to deal with metrics on every protocol bridge (HTTP, websocket, stdio, IPC, etc.). But I'm very open to other ideas.

omerfirmak: > and we don't need to deal with metrics on every protocol bridge (HTTP, websocket, stdio, IPC, etc.).

I think that is a good metrics tho, to get an idea of which transport protocols are more utilized.

> This only needs to be set once on the main jsonrpc server

Yeap, but that way (if we apply this approach to all packages that collect metrics) we need to move all metric collection code away from the packages and in to `node` package. Since the callbacks and the metrics (counters etc) will be defined on `node` package. That is what I am trying to avoid essentially, to not clutter `node` package.

joshklop: > I think that is a good metrics tho, to get an idea of which transport protocols are more utilized.

Good point.

> we need to move all metric collection code away from the packages and in to node package.

>  That is what I am trying to avoid essentially, to not clutter node package.

Ah ok, that is what I was trying to move towards xD. Move metrics code together into the Juno-specific `node` or `node/metrics` package, instead of sprinkled throughout the codebase. It would be disappointing if virtually every package had to depend on prometheus (which itself pulls in a large number of dependencies).

To avoid excessive use of callbacks, maybe we can compromise with the lightweight [expvar](https://pkg.go.dev/expvar) package? We can combine it with Tailscale's [prometheus exporter](github.com/tailscale/tailscale). It requires us to name metrics a [certain way](https://github.com/tailscale/tailscale/blob/388b12451341494b76976c40467b3785767eddf2/tsweb/varz/varz.go#L34-L36), but I don't think that's too big of an issue.

omerfirmak: > maybe we can compromise with the lightweight [expvar](https://pkg.go.dev/expvar) package?

We will have to spin up or own implementation for more complex types like `Histogram` or `CounterVec`. I am not sure if it is worth it.

Honestly, prometheus' dependencies are not that bad imo https://github.com/prometheus/client_golang/blob/main/go.mod

joshklop: Prometheus depends on protobuf, which pulls in quite a bit (e.g., see [here](https://github.com/tailscale/tailscale/pull/125#issuecomment-593903976)).

Tailscale [has histograms](https://github.com/tailscale/tailscale/blob/388b12451341494b76976c40467b3785767eddf2/metrics/metrics.go#L79).

Maybe you're right and I'm just over-complicating things. Feel free to go with what you think is best.

cmd/juno/juno.go
```diff
@@ -53,6 +54,7 @@ const (
 	defaultP2p                 = false
 	defaultP2pAddr             = ""
 	defaultP2pBootPeers        = ""
+	defaultMetricsPort         = 0
```
IronGauntlets, line 57: Since we are using Prometheus to expose metrics we should by default set it to `9090`, right?

cmd/juno/juno.go
```diff
@@ -159,6 +162,7 @@ func NewCmd(config *node.Config, run func(*cobra.Command, []string) error) *cobr
 	junoCmd.Flags().Bool(p2pF, defaultP2p, p2pUsage)
 	junoCmd.Flags().String(p2pAddrF, defaultP2pAddr, p2PAddrUsage)
 	junoCmd.Flags().String(p2pBootPeersF, defaultP2pBootPeers, p2pBootPeersUsage)
+	junoCmd.Flags().Uint16(metricsPortF, defaultMetricsPort, metricsPortUsage)
```
IronGauntlets, line 165: I think it would be best to have a `bool` flag where by default we don't have metrics enabled and if the user wants to enable it we use port `9090` as that is the expected Prometheus metrics port. I highly doubt someone would want to run it on something other than 9090, if someone does request it then we can make it configurable.

metrics/metrics.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+}
+
+func (m *Metrics) Run(ctx context.Context) error {
+	errCh := make(chan error)
+
+	srv := &http.Server{
+		Addr:    m.listener.Addr().String(),
+		Handler: promhttp.HandlerFor(prometheus.DefaultGatherer, promhttp.HandlerOpts{Registry: prometheus.DefaultRegisterer}),
+		// ReadTimeout also sets ReadHeaderTimeout and IdleTimeout.
```
IronGauntlets, line 38: This comment is misleading as setting `ReadTimeout` doesn't set  `ReadHeaderTimeout` and `IdleTimeout`. Instead `ReadTimeout` is used if `ReadHeaderTimeout` and `IdleTimeout` are not set. IMO, it would be best to remove it.

jsonrpc/http.go
```diff
@@ -19,14 +21,25 @@ type HTTP struct {
...
-	return &HTTP{
+	h := &HTTP{
 		rpc:      rpc,
 		log:      log,
 		listener: listener,
+
+		requests: prometheus.NewCounter(prometheus.CounterOpts{
+			Namespace: "rpc",
+			Subsystem: "http",
+			Name:      "requests",
```
IronGauntlets, line 38: let's make it more specific `http request` and the same for WebSockets.

sync/sync.go
IronGauntlets: I think we should extract them out to a `const`.

IronGauntlets (changes requested):
Overall looks good, requesting some small changes.

jsonrpc/http.go
omerfirmak: It has `Subsystem: "http"` already. WS has `Subsystem: "ws",`

metrics/metrics.go
omerfirmak: changed

cmd/juno/juno.go
omerfirmak: done

cmd/juno/juno.go
omerfirmak: done

jsonrpc/http.go
omerfirmak: > I'm just over-complicating things

Nah, you are absolutely right that we should try to minimize our dependencies. But for juno, `prometheus` doesnt really bring many new dependencies, for example we already depend on protobuf.

and at this point, it is so early to try to optimize for re-usability of our packages. We should instead optimize for Juno overall and bring as many features we need as fast as possible.

metrics/metrics.go
```diff
@@ -0,0 +1,53 @@
...
+	"errors"
+	"net"
+	"net/http"
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+)
+
+var Enabled = false
```
IronGauntlets, line 14: I think we make this private by setting it in the `New()`, having both bool flag and this public global variable seems redundant to me.

IronGauntlets: Will fix in another issue.

IronGauntlets (approved):

cmd/juno/juno.go
```diff
@@ -69,6 +73,8 @@ const (
 	p2pUsage                 = "enable p2p server"
 	p2PAddrUsage             = "specify p2p source address as multiaddr"
 	p2pBootPeersUsage        = "specify list of p2p boot peers splitted by a comma"
+	metricsUsage             = "enable prometheus endpoint"
+	metricsPortUsage         = "The port on which the prometheus server will listen for requests (default: 9090)"
```
wojciechos, line 77: not sure if (default: 9090) is needed here as it is displayed twice, see here:
image

PR #924, opened 2023-07-18
Proto/p2p part2

asdacap:
- Protobuff and serving blocks.
- Should be able to serve blocks.


omerfirmak:
Should we adopt the new proposed protobuf spec?

asdacap:
It does not look like its ready yet.

PR #925, opened 2023-07-18
Return 200 on root GET requests

### LINKED ISSUE #923: Add `/health` endpoint and return 200 for `/`
Hello,

We're using several instances of juno / pathfinder binaries for our Starknet RPCs.
The root path (`/`) of pathfinder return a `405` : 

Do you think it's possible to add a `/health` endpoint returning 200 if the binary is running well for our Load Balancer Health Check ?
Furthermore, do you think it's possible to return `200` when there are no parameters given on the endpoint `/` please ?

omerfirmak:
Otherwise 404

Fixes #923

### DIFF
--- a/jsonrpc/http.go
+++ b/jsonrpc/http.go
@@ -55,9 +55,15 @@ func (h *HTTP) Run(ctx context.Context) error {
 
 // ServeHTTP processes an incoming HTTP request
 func (h *HTTP) ServeHTTP(writer http.ResponseWriter, req *http.Request) {
-	if req.Method != "POST" {
+	if req.Method == "GET" {
+		status := http.StatusNotFound
+		if req.URL.Path == "/" {
+			status = http.StatusOK
+		}
+		writer.WriteHeader(status)
+		return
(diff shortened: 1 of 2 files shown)


jsonrpc/http.go
```diff
@@ -55,7 +55,15 @@ func (h *HTTP) Run(ctx context.Context) error {
...
 // ServeHTTP processes an incoming HTTP request
 func (h *HTTP) ServeHTTP(writer http.ResponseWriter, req *http.Request) {
-	if req.Method != "POST" {
+	if req.Method == "GET" {
+		status := http.StatusNotFound
+		if req.URL.Path == "/" {
+			status = http.StatusOK
+		}
+		writer.WriteHeader(status)
+		req.Close = true
```
kirugan, line 64: [Looks like](https://pkg.go.dev/net/http#Request:~:text=//%20For%20server%20requests%2C%20the%20HTTP%20server%20handles%20this%20automatically) in server handlers we don't need to set it explicitly

kirugan (approved):

PR #926, opened 2023-07-18
New method starknet_simulateTransactions

kirugan:


rpc/simulation.go
```diff
@@ -0,0 +1,28 @@
+package rpc
+
+import "fmt"
+
+type SimulationFlag int
+
+const (
+	SkipValidateFlag SimulationFlag = iota + 1
+	SkipFeeChargeFlag
```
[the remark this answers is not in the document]
kirugan: Return an error when simulation flags passed

rpc/handlers.go
```diff
@@ -1242,6 +1242,64 @@ func (h *Handler) TraceTransaction(hash felt.Felt) (json.RawMessage, *jsonrpc.Er
...
+		if err != nil {
+			// todo or InvalidParams ?
+			return nil, jsonrpc.Err(jsonrpc.InternalError, err.Error())
+		}
+
+		txns = append(txns, tx)
+		classes = append(classes, class)
+
+		paidFeesOnL1 := []*felt.Felt{paidFeeOnL1}
+		gasConsumed, traces, err := h.vm.Execute(txns, classes, header.Number, header.Timestamp,
```
omerfirmak, line 1281: Why are calling `vm.Execute()` for each txn?
Shouldnt we just call it once for the entire txn set?

omerfirmak:
If I am not mistaken, EstimateFee is just SimulateTransactions but it doesnt return the traces.
So what I think we should do is; move the entire function body of EstimateFee to SimulateTransactions and just add traces to it.

Once that is done, we can make EstimateFee call SimulateTransactions.

rpc/handlers.go
kirugan: My bad, fixed

kirugan:
> If I am not mistaken, EstimateFee is just SimulateTransactions but it doesnt return the traces. So what I think we should do is; move the entire function body of EstimateFee to SimulateTransactions and just add traces to it.
> 
> Once that is done, we can make EstimateFee just a wrapper around SimulateTransactions.

Done, the minor issue with logging:
```
defer h.callAndLogErr(closer, "Failed to close state in starknet_estimateFee")
```
We need somehow pass information about which method is called either through `context.Context` or just plain string. WDYT?

omerfirmak:
> We need somehow pass information about which method is called either through `context.Context` or just plain string. WDYT?

I dont think it is worth it. It can just print `Failed to close state in starknet_simulateTransactions`

omerfirmak (approved):
LGTM

PR #931, opened 2023-07-19
Store Transaction and Event commitments

omerfirmak:
Comes with a lengthy migration


blockchain/blockchain.go
```diff
@@ -558,16 +597,18 @@ func stateUpdateByHash(txn db.Transaction, hash *felt.Felt) (*core.StateUpdate,
 }
 
 // SanityCheckNewHeight checks integrity of a block and resulting state update
-func (b *Blockchain) SanityCheckNewHeight(block *core.Block, stateUpdate *core.StateUpdate, newClasses map[felt.Felt]core.Class) error {
+func (b *Blockchain) SanityCheckNewHeight(block *core.Block, stateUpdate *core.StateUpdate,
+	newClasses map[felt.Felt]core.Class,
```
joshklop, line 601: Is there a reason why we aren't adding the commitments to the `core.Header` struct? ~We can calculate them in `feeder2core.AdaptBlock`.~

core/block.go
```diff
@@ -92,57 +92,59 @@ func networkBlockHashMetaInfo(network utils.Network) *blockHashMetaInfo {
...
 	if err := VerifyTransactions(b.Transactions, network, b.ProtocolVersion); err != nil {
-		return err
+		return nil, err
 	}
 
 	metaInfo := networkBlockHashMetaInfo(network)
-
 	unverifiableRange := metaInfo.UnverifiableRange
-	if unverifiableRange != nil {
-		// Check if the block number is in the unverifiable range
```
joshklop, line 119: Can we calculate the commitments here so we don't do the check in the loop?

migration/migration.go
```diff
@@ -299,3 +304,33 @@ func (m *changeTrieNodeEncoding) Migrate(txn db.Transaction) error {
...
 	return iterator.Close()
 }
+
+// calculateBlockCommitments calculates the txn and event commitments for each block and stores them separately
+func calculateBlockCommitments(txn db.Transaction, network utils.Network) error {
+	var txnLock sync.RWMutex
+	workerPool := pool.New().WithErrors().WithMaxGoroutines(runtime.GOMAXPROCS(0))
+	blockNumber := 0
+
+	for {
```
joshklop, line 314:
```suggestion
	for blockNumber := 0; ; blockNumber++ {
```
And then remove `blockNumber++` at the end of the loop.

migration/migration_pkg_test.go
```diff
@@ -160,3 +168,34 @@ func TestChangeTrieNodeEncoding(t *testing.T) {
...
+	}
+
+	require.NoError(t, testdb.Update(func(txn db.Transaction) error {
+		return calculateBlockCommitments(txn, utils.MAINNET)
+	}))
+
+	for i := uint64(0); i < 3; i++ {
+		b, err := chain.BlockCommitmentsByNumber(i)
+		require.NoError(t, err)
+		assert.False(t, b.TransactionCommitment == nil && b.EventCommitment == nil)
```
joshklop, line 199: Shouldn't this be `||` instead of `&&`?

nit: I personally prefer individual `assert.Nil`/`assert.NotNil`s here, but feel free to keep it as it is.

migration/migration_pkg_test.go
```diff
@@ -160,3 +168,34 @@ func TestChangeTrieNodeEncoding(t *testing.T) {
...
+	t.Cleanup(closeFn)
+	gw := adaptfeeder.New(client)
+
+	for i := uint64(0); i < 3; i++ {
+		b, err := gw.BlockByNumber(context.Background(), i)
+		require.NoError(t, err)
+		su, err := gw.StateUpdate(context.Background(), i)
+		require.NoError(t, err)
+
+		b.EventsBloom = nil
```
joshklop, line 188: I don't think we need this line.

joshklop (changes requested):

core/block.go
omerfirmak: I considered that too, but didnt want to duplicate the `blockHash(b, network, overrideSeq)` call.

migration/migration_pkg_test.go
omerfirmak: Earlier blocks don't have event commitments. So we need to check only TransactionCommitment to be non-nil. Will update

migration/migration_pkg_test.go
omerfirmak: Oops, copy pasta error

migration/migration.go
omerfirmak: Thanks:)

core/block.go
joshklop: Can we export the `transactionCommitment` and `eventCommitment` functions, call them from `SanityCheckNewHeight` (or wherever), and pass them down to `blockHash` instead of bubbling them up? Injects the dependencies.

omerfirmak: They are not calculated the same for all block versions and some blocks dont even use `eventCommitment`. Additional logic in `blockHash`, `pre07Hash` and `post07Hash` handles these differences. So we would need to move (if not duplicate, which is worse) those checks out of `blockHash` and into the caller site. 

I want to keep the complexity close to wherever the data is actually used, so I think it is best to keep it as is for now.

joshklop: Sounds good.

omerfirmak:
@joshklop can we merge this if you are happy with it?

joshklop (approved):
LGTM. Sorry for the delay, forgot to re-review last week.

migration/migration_pkg_test.go
```diff
@@ -168,3 +168,32 @@ func TestChangeTrieNodeEncoding(t *testing.T) {
...
+	}
+
+	require.NoError(t, testdb.Update(func(txn db.Transaction) error {
+		return calculateBlockCommitments(txn, utils.MAINNET)
+	}))
+
+	for i := uint64(0); i < 3; i++ {
+		b, err := chain.BlockCommitmentsByNumber(i)
+		require.NoError(t, err)
+		assert.NotNil(t, b.TransactionCommitment)
```
IronGauntlets, line 197: Is there a reason why you only check `TransactionCommitment` and not `EventCommitment`?

migration/migration.go
```diff
@@ -300,3 +304,31 @@ func (m *changeTrieNodeEncoding) Migrate(txn db.Transaction, _ utils.Network) er
...
+
+// calculateBlockCommitments calculates the txn and event commitments for each block and stores them separately
+func calculateBlockCommitments(txn db.Transaction, network utils.Network) error {
+	var txnLock sync.RWMutex
+	workerPool := pool.New().WithErrors().WithMaxGoroutines(runtime.GOMAXPROCS(0))
+
+	for blockNumber := 0; ; blockNumber++ {
+		txnLock.RLock()
+		block, err := blockchain.BlockByNumber(txn, uint64(blockNumber))
+		txnLock.RUnlock()
```
IronGauntlets, line 316: Are you using the Mutex to protect `block` or `blockNumber`? 

Also it looks like to me that you are using `BlockByNumber` only to determine when db no longer have blocks, if that is the case then I think the following code would be more simpler (by removing the need for mutex):

```go
// calculateBlockCommitments calculates the txn and event commitments for each block and stores them separately
func calculateBlockCommitments(txn db.Transaction, network utils.Network) error {
	workerPool := pool.New().WithErrors().WithMaxGoroutines(runtime.GOMAXPROCS(0))

	for blockNumber := uint64(0); ; blockNumber++ {
		block, err := blockchain.BlockByNumber(txn, blockNumber)

			break
		}

		workerPool.Go(func() error {
			block := block
			commitments, err := core.VerifyBlockHash(block, network)
			if err != nil {
				return err
			}
			return blockchain.StoreBlockCommitments(txn, block.Number, commitments)
		})
	}

	return workerPool.Wait()
}
```

IronGauntlets (changes requested):

IronGauntlets:
Also, I haven't understood why this PR is required.

migration/migration_pkg_test.go
omerfirmak: See https://github.com/NethermindEth/juno/pull/931#discussion_r1269063435

migration/migration.go
omerfirmak: As the name implies, mutex is used to protect the txn. We are concurrently doing both read and write operations on it. So it needs to be protected.

omerfirmak:
> Also, I haven't understood why this PR is required.

in P2P we will have to serve event and transaction commitments as part of the block header. We need to store them to be able to do this without constantly recalculating them. Which is really costly.

omerfirmak:
@IronGauntlets, I want to merge this since it will help with the P2P work. Can you approve if my answers satisfies your questions?

migration/migration.go
IronGauntlets: I don't understand why we need to protect the `txn`, should pebble handle concurrent updates and read to batches/snapshot also we have a test for ConcurrentUpdates. 

Also, `block := block` is required

migration/migration_pkg_test.go
IronGauntlets: If we test for the content of the TransactionCommitment then we would be able to tell whether we stored the right commitment to the the right block number. Right now, these tests are only testing for the existence of a Commitment and I think `calculateCommitments` is racey even with mutex because `block` can change in the goroutine by the time it is evaluated.

migration/migration.go
omerfirmak: > Also, block := block is required

I dont think so, the variable is in the `for` loop body. So they are separate instances.

> I don't understand why we need to protect the txn, should pebble handle concurrent updates and read to batches/snapshot also we have a test for ConcurrentUpdates.

Transactions are not concurrency-safe. What `TestConcurrentUpdate` is testing is that db can be read and updated concurrently concurrently thru `db.Update`/`db.View`, which each create separate transactions.

IronGauntlets (approved):

PR #941, opened 2023-07-21
Use ubuntu:23.04 as base docker image

omerfirmak:
Because the default libc in alpine (musl) has really small default stack sizes and we were overflowing it in our CGO calls.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,9 @@
 # Stage 1: Build golang dependencies and binaries
-FROM alpine:3.18 AS build
+FROM ubuntu:23.04 AS build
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,9 +1,9 @@
 # Stage 1: Build golang dependencies and binaries
-FROM alpine:3.18 AS build
+FROM ubuntu:23.04 AS build
 
 # Install Alpine Dependencies
```
wojciechos, line 4: it's not Alpine deps anymore

wojciechos (approved):

PR #943, opened 2023-07-21
Expect status strings in quotes for JSON unmarshaling

omerfirmak:
JSON parser does not strip the quotes before passing the string to the UnmarshalJSON functions

### DIFF
--- a/clients/feeder/transaction.go
+++ b/clients/feeder/transaction.go
@@ -15,9 +15,9 @@ const (
 
 func (es *ExecutionStatus) UnmarshalJSON(data []byte) error {
 	switch string(data) {
-	case "SUCCEEDED":
+	case `"SUCCEEDED"`:
 		*es = Succeeded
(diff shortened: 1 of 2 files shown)


clients/feeder/transaction.go
```diff
@@ -15,9 +15,9 @@ const (
 
 func (es *ExecutionStatus) UnmarshalJSON(data []byte) error {
 	switch string(data) {
-	case "SUCCEEDED":
+	case "\"SUCCEEDED\"":
```
kirugan, line 18: maybe use backticks? i.e. 
```
case `"SUCCEEDED"`
```
but it's up to you

kirugan (approved):

clients/feeder/transaction.go
omerfirmak: Yeap, much better. Thanks for pointing that out.

PR #944, opened 2023-07-21
Store RevertReason in storage

omerfirmak:


joshklop (approved):
I assume we don't need a migration here since `RevertReason` will unmarshal to `""`?

omerfirmak:
> I assume we don't need a migration here since `RevertReason` will unmarshal to `""`?

Yeap, you are correct.

PR #947, opened 2023-07-24
Set Keccak builtin weight

omerfirmak:
Starknet v0.12.1 enables Keccak builtin
got the weight from https://github.com/starkware-libs/cairo-lang/blob/c98fc0b50529185b7018208cb3460191eeb53e0d/src/starkware/cairo/lang/instances.py#L288-L292

### DIFF
--- a/vm/rust/src/class.rs
+++ b/vm/rust/src/class.rs
@@ -1,21 +1,21 @@
+use starknet::core::types::contract::legacy::LegacyContractClass;
 use std::{
     ffi::{c_char, c_uchar, CStr},
     slice,
 };
-use starknet::core::types::contract::legacy::LegacyContractClass;
 
 #[no_mangle]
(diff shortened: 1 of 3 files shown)


vm/rust/src/class.rs
```diff
@@ -1,21 +1,21 @@
...
     let class_json = unsafe { CStr::from_ptr(class_json_str) }.to_str().unwrap();
     let class: Result<LegacyContractClass, serde_json::Error> = serde_json::from_str(class_json);
     if class.is_err() {
-        return
+        return;
     }
     let class_hash = class.unwrap().class_hash();
     if class_hash.is_err() {
-        return
+        return;
```
kirugan, line 16: is there a difference with `;` ?

kirugan (approved):

vm/rust/src/class.rs
omerfirmak: Functionally, probably not. It is just a formatting/style thing. 

It was `cargo fmt` who changed it.

PR #949, opened 2023-07-24
Use legacy `status` field while testnet and mainnet are 0.12

joshklop:
We can transition to the new status fields once the network upgrades to 0.12.1.

### DIFF
--- a/clients/feeder/testdata/integration/transaction/0x45d9c2c8e01bacae6dec3438874576a4a1ce65f1d4247f4e9748f0e7216838.json
+++ b/clients/feeder/testdata/integration/transaction/0x45d9c2c8e01bacae6dec3438874576a4a1ce65f1d4247f4e9748f0e7216838.json
@@ -1,0 +1,31 @@
+{
+    "execution_status": "SUCCEEDED",
+    "finality_status": "ACCEPTED_ON_L2",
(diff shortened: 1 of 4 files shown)


omerfirmak:
How about something like;

``` go
		switch txStatus.FinalityStatus {
		case feeder.AcceptedOnL1:
			status.Finality = TxnAcceptedOnL1
		case feeder.AcceptedOnL2:
			status.Finality = TxnAcceptedOnL2
		default:
                        // pre0.12.1
			if txStatus.Status == "ACCEPTED_ON_L1" {
				status.Finality = TxnAcceptedOnL1
			} else {
				status.Finality = TxnAcceptedOnL2
			}
		}

		switch txStatus.ExecutionStatus {
		case feeder.Succeeded:
			status.Execution = TxnSuccess
		case feeder.Reverted:
			status.Execution = TxnFailure
		default:
                        // pre0.12.1
			status.Execution = TxnSuccess
		}
```

this way, it will work post0.12.1 as well. 

We can just bring back the errors in `default` cases once all networks upgrade to 0.12.1

joshklop:
Nice, I like that.

omerfirmak (approved):
Will need tests for CI to pass.

PR #950, opened 2023-07-24
Allow RPC handlers to accept context as first argument

joshklop:


jsonrpc/server.go
```diff
@@ -145,8 +154,16 @@ func (s *Server) RegisterMethod(method Method) error {
...
 		return errors.New("handler must be a function")
 	}
-	if handlerT.NumIn() != len(method.Params) {
-		return errors.New("number of function params and param names must match")
+	var needsContext bool
+	numArgs := handlerT.NumIn()
+	if numArgs > 0 {
+		if handlerT.In(0).Implements(contextInterface) {
+			numArgs--
+			needsContext = true
```
omerfirmak, line 162: I believe you can directly set `method.needsContext` here, there is no need to a separate  `var needsContext bool`. `method` is a copy anyways.

jsonrpc/server.go
```diff
@@ -316,37 +335,52 @@ func (s *Server) handleRequest(req *request) (*response, error) {
...
 
-		if len(paramsList) != handlerType.NumIn() {
+		if len(paramsList) != numNonContextArgs {
 			return nil, errors.New("missing/unexpected params in list")
 		}
 
 		for i, param := range paramsList {
+			if method.needsContext {
+				i++
+			}
```
omerfirmak, line 370: If you keep the previous `append` approach, this becomes unnecessary. You can just append the context to the args first, then continue appending `paramsList-1` number of actual arguments.

omerfirmak: since the entire list is pre-allocated `args := make([]reflect.Value, 0, len(configuredParams))` appending doesnt cause any allocations

omerfirmak (changes requested):

joshklop:
Resolved comments, rebased, added two more tests.

jsonrpc/server.go
```diff
@@ -316,28 +333,36 @@ func (s *Server) handleRequest(req *request) (*response, error) {
...
-	args := make([]reflect.Value, 0, len(configuredParams))
+func (s *Server) buildArguments(ctx context.Context, params any, method Method) ([]reflect.Value, error) {
 	if isNil(params) {
-		if len(configuredParams) > 0 {
+		if len(method.Params) > 0 {
 			return nil, errors.New("missing non-optional param field")
 		}
 
-		return args, nil
+		return make([]reflect.Value, 0), nil
```
omerfirmak, line 342: Hmm, I think we actually had a bug here.

From the spec;

```
params
    A Structured value that holds the parameter values to be used during the invocation of the method. This member MAY be omitted. 
```

This makes the check above a bit too strict. If all of `method.Params` are optional, we should let `params` be nil. But now, we only allow `params` to be nil for methods with no parameters.

`params` is optional, if it is not given we should return the default values for each argument. Not an empty list.

joshklop: Nice catch. Maybe we can resolve this in a separate PR? Seems like a separate issue if I'm understanding correctly.

Issue #937, opened 2023-07-20
Write API: map `StarknetErrorCode` to new 0.4 spec error codes


joshklop:
Feeder returns a `StarknetErrorCode` that maps to the spec error codes.

PR #951, opened 2023-07-24
Add ErrCode constants to feeder

omerfirmak:
Towards #937

### DIFF
--- a/clients/gateway/gateway.go
+++ b/clients/gateway/gateway.go
@@ -16,6 +16,42 @@ import (
 	"github.com/NethermindEth/juno/utils"
 )
 
+type ErrorCode string
+
+var (
+	BlockNotFound                   ErrorCode = "StarknetErrorCode.BLOCK_NOT_FOUND"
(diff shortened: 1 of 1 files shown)


joshklop (approved):
Where did you find all of these?

joshklop:
Actually just found [this](https://github.com/starkware-libs/cairo-lang/blob/54d7e92a703b3b5a1e07e9389608178129946efc/src/starkware/starknet/definitions/error_codes.py). Should we add the others?

omerfirmak:
> Where did you find all of these?

Stole some of them from Pathfinder, the rest are the new errors from Write API

omerfirmak:
> Actually just found [this](https://github.com/starkware-libs/cairo-lang/blob/master/src/starkware/starknet/definitions/error_codes.py). Should we add the others?

Hmm, let's either add them all or delete most of the existing ones as well and only keep the ones that we will check for write API. WDYT?

joshklop:
Sounds good. I'm fine with both options.

omerfirmak:
> delete most of the existing ones as well and only keep the ones that we will check for write API.

Let's go with this. I will close this PR for now and include this in the PR for #937

PR #952, opened 2023-07-24
Add SubscribeNewHeads

joshklop:
Towards websocket subscriptions


blockchain/blockchain.go
```diff
@@ -765,15 +770,47 @@ func revertHead(txn db.Transaction) error {
...
+
+func removeTxsAndReceipts(txn db.Transaction, blockNumber, numTxs uint64) error {
 	blockIDAndIndex := txAndReceiptDBKey{
 		Number: blockNumber,
 	}
 	// remove txs and receipts
-	for i := uint64(0); i < header.TransactionCount; i++ {
+	for i := uint64(0); i < numTxs; i++ {
 		blockIDAndIndex.Index = i
 		var reorgedTxn core.Transaction
```
kirugan, line 812: we don't need this declaration because below is `:=` assignment now

blockchain/blockchain.go
```diff
@@ -765,15 +770,47 @@ func revertHead(txn db.Transaction) error {
...
+}
+
+func removeTxsAndReceipts(txn db.Transaction, blockNumber, numTxs uint64) error {
 	blockIDAndIndex := txAndReceiptDBKey{
 		Number: blockNumber,
 	}
 	// remove txs and receipts
-	for i := uint64(0); i < header.TransactionCount; i++ {
+	for i := uint64(0); i < numTxs; i++ {
 		blockIDAndIndex.Index = i
```
kirugan, line 811: I think we can make `blockIDAndIndex` declaration here entirely i.e.:
```
blockIDAndIndex := txAndReceiptDBKey{
	Number: blockNumber,
	Index:  i,
}
```

blockchain/blockchain.go
```diff
@@ -765,15 +770,47 @@ func revertHead(txn db.Transaction) error {
 	if err = txn.Delete(db.BlockHeaderNumbersByHash.Key(header.Hash.Marshal())); err != nil {
 		return err
 	}
+	if blockNumber != 0 { // Don't send block before genesis
```
kirugan, line 773: do you think introducing new variable `genesisBlock := blockNumber == 0`
and using it later like:
```
if !genesisBlock {
....
}
...
if genesisBlock {
...
}
```
would improve readability or it's unnecessary?

blockchain/blockchain.go
joshklop: Will compiler know that the final binary doesn't need to include write instructions for `blockIDAndIndex.Number`? 

It's totally a micro-optimization, but I'm curious nonetheless. I think we might as well go with the more efficient version rather than relying on the compiler.

blockchain/blockchain.go
joshklop: I like it.

blockchain/blockchain.go
joshklop: Sorry, I don't understand.

kirugan: this line is not used, before your changes it was necessary because `=` is assignment operator, but `:=` assignment + declaration operator (terminology is mine, but I think you get idea).

blockchain/blockchain.go
kirugan: Short answer: you never know :) compilers are complex beasts and sometimes they apply all sorts of crazy unexpected optimizations.

In this particular example you are right current solution is more optimal than the one I proposed. Here is [output from godbolt](https://godbolt.org/z/xzej7aP41):
```
mov     rax, QWORD PTR [rbp-72]
mov     QWORD PTR [rbp-48], rax
mov     rax, QWORD PTR [rbp-8]
mov     QWORD PTR [rbp-40], rax
```
vs [assignment of just one field](https://godbolt.org/z/zP3x7fneY) 
```
mov     rax, QWORD PTR [rbp-8]
mov     QWORD PTR [rbp-40], rax
```

joshklop: Interesting, thanks. I think I'll leave it as it is.

blockchain/blockchain.go
joshklop: I see now. I didn't realize you were referring to the `var reorgedTxn core.Transaction` line. Fixed now, thanks.

kirugan (approved):

blockchain/blockchain.go
```diff
@@ -758,22 +763,55 @@ func revertHead(txn db.Transaction) error {
...
 	if err = txn.Delete(db.BlockHeaderNumbersByHash.Key(header.Hash.Marshal())); err != nil {
 		return err
 	}
+	if !genesisBlock {
+		var newHeader *core.Header
+		newHeader, err = blockHeaderByNumber(txn, blockNumber-1)
+		if err != nil {
+			return err
+		}
+		b.newHeads.Send(newHeader)
```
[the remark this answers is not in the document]
joshklop: The user can save the most recent received header `A`. Upon receiving a new header `B`, one of two things must be true:

1. `B`'s parent is `A`: `B.ParentHash.Equal(A.Hash)`
2. `B` is `A`: `B.Hash.Equal(A.Hash)`

Since it's simple for a user to distinguish reverted/new blocks, I don't think it's necessary to do anything on our end. We should make that distinction for logs, though, since there isn't a good way for a client to decide. Wdyt?

PR #955, opened 2023-07-24
Add subscription support to jsonrpc package

joshklop:


jsonrpc/http.go
```diff
@@ -53,32 +53,54 @@ func (h *HTTP) Run(ctx context.Context) error {
...
+
+func (c *httpConn) Read(p []byte) (int, error) {
+	c.req.Body = http.MaxBytesReader(c.rw, c.req.Body, MaxRequestBodySize)
+	return c.req.Body.Read(p)
+}
+
+// Write returns the number of bytes of p written, not including the header.
+func (c *httpConn) Write(p []byte) (int, error) {
+	c.rw.Header().Set("Content-Type", "application/json")
+	c.rw.WriteHeader(http.StatusOK)
```
omerfirmak, line 87: dont we need a way to write non-200 status codes?

joshklop: Not here. If we've made it this far, we will only send a 200. `rw.Write` will do it automatically even if we don't.

jsonrpc/memconn.go
```diff
@@ -0,0 +1,29 @@
+package jsonrpc
+
+import "io"
+
+// MemConn is a simple in-memory JSON RPC connection.
+type MemConn struct {
```
omerfirmak, line 10: I think we can avoid adding this, we can keep HandleReader public and use that in tests.

omerfirmak: `Handle` is simple enough that we can skip testing it.

jsonrpc/http.go
```diff
@@ -30,33 +30,53 @@ func NewHTTP(rpc *Server, log utils.SimpleLogger) *HTTP {
...
+	}
+
+	return &httpConn{
+		req: req,
+		rw:  rw,
+	}
+}
+
+func (c *httpConn) Read(p []byte) (int, error) {
+	c.req.Body = http.MaxBytesReader(c.rw, c.req.Body, MaxRequestBodySize)
```
omerfirmak, line 58: we need to move `http.MaxBytesReader` to `accept`, otherwise caller can read more than `MaxRequestBodySize` by reading in chunks.

I think httpConn can store an io.ReaderCloser instead of the entire `http.Request`

jsonrpc/memconn.go
joshklop: > I think we can avoid adding this

Brilliant. Didn't think long enough to realize it wasn't necessary.

> we can keep HandleReader public and use that in tests.

Ended up using `net.Pipe()`, which is what the `MemConn` was mimicking anyway.

jsonrpc/websocket.go
```diff
@@ -97,53 +100,36 @@ func DefaultWebsocketConnParams() *WebsocketConnParams {
...
-		}
-
-		// Send the message to the client.
-		if err := wsc.Write(ctx, resp); err != nil {
-			return err
-		}
-	}
+func (wsc *websocketConn) Read(p []byte) (int, error) {
+	_, r, err := wsc.conn.Read(wsc.ctx)
+	return copy(p, r), err
```
omerfirmak, line 119: There is a possibility of losing part of the data read here. Is there a way to overcome this?

joshklop: Nice catch. I gave it a shot, but would appreciate your review. Not sure how to test this one.

omerfirmak:
@joshklop can you rebase this

jsonrpc/server.go
```diff
@@ -192,10 +192,25 @@ func (s *Server) RegisterMethod(method Method) error {
...
+	}
+	return nil
+}
+
+// handle processes a request to the server
 // It returns the response in a byte array, only returns an
 // error if it can not create the response byte array
-func (s *Server) HandleReader(ctx context.Context, reader io.Reader) ([]byte, error) {
+func (s *Server) handle(ctx context.Context, reader io.Reader) ([]byte, error) {
 	bufferedReader := bufio.NewReader(reader)
```
omerfirmak: unrelated, but this allocates 4K per request. We should get rid of it.

omerfirmak: or pool it for reuse

joshklop: Looks like we only use it for determining if the request is a batch. geth does this:

```go
// isBatch returns true when the first non-whitespace characters is '['
func isBatch(raw json.RawMessage) bool {
	for _, c := range raw {
		// skip insignificant whitespace (http://www.ietf.org/rfc/rfc4627.txt)
		if c == 0x20 || c == 0x09 || c == 0x0a || c == 0x0d {
			continue
		}
		return c == '['
	}
	return false
}
```
Could we unmarshal into `json.RawMessage` instead of using the `Decoder`?

jsonrpc/http.go
```diff
@@ -30,31 +31,51 @@ func NewHTTP(rpc *Server, log utils.SimpleLogger) *HTTP {
...
+	conn := accept(writer, req)
+	if conn == nil {
+		return
+	}
 	h.requests.Inc()
-	resp, err := h.rpc.HandleReader(req.Context(), req.Body)
-	writer.Header().Set("Content-Type", "application/json")
-	if err != nil {
+	if err := h.rpc.Handle(req.Context(), conn); err != nil {
 		writer.WriteHeader(http.StatusInternalServerError)
```
omerfirmak, line 79: this potentially brings back the error fixed [here](https://github.com/NethermindEth/juno/pull/1075)
First call to `conn.Write` in `server.Handle` might successfully write a 200 header and fail to write the body. Then here we will try to write a 500 header again.

jsonrpc/server.go
omerfirmak: We can but that will double to allocations, could be worse than what we currently do.

joshklop: Will try to pool it then :+1:

omerfirmak: Let's just replace the call with `bufio.NewReaderSize(reader, 16)` atleast we allocate less this way.

omerfirmak:
I think we need to move subscription related stuff out of `jsonrpc` to either its own package or `rpc`. This will provide better separation of concerns. The current way it works in this PR looks a bit too intrusive to me.

Also, we had consensus on keeping `jsonrpc` as an implementation of its spec. So I believe we shouldnt expend it beyond that.

What I have in mind looks more or less like;

1) `jsonrpc` is only modified to allow bidirectional communication between method handler and the client (we can pass the "connection" (ie WriterReaderCloser) to handler by piggy backing it to the context)
2) each subscription endpoint comes with a handler like all endpoints do, which declines request or registers/unregisters them to a separate entity. (lets call it SubsServer for the time being) It can also send the initial response (for example, subscribe_txn_status) can return the current txn status at the time of subscribe request then SubsServer will handle all the updates)
3) we emit events from `blockchain` like we do for metrics in #1242
4) SubsServer listens to the events emitted from blockchain and if there is a registered subscription for that event sends it over the connection.

Let me know what you think

joshklop:
I like that. I'll work on it.

joshklop:
Created a new PR to avoid rewriting this PR's history https://github.com/NethermindEth/juno/pull/1266

PR #956, opened 2023-07-25
Map gateway errors to write api rpc errors

omerfirmak:


omerfirmak:
Fixes #937

rpc/handlers.go
```diff
@@ -879,20 +888,35 @@ func (h *Handler) AddInvokeTransaction(invokeTx json.RawMessage) (*AddInvokeTxRe
 }
 
 // getAddInvokeTxCode returns the relevant Code for a given addInvokeTx error
```
kirugan, line 890: this comment is obsolete

omerfirmak: Yeap, fixed. Thanks!

clients/gateway/gateway.go
```diff
@@ -132,19 +144,21 @@ func (c *Client) doPost(ctx context.Context, url string, data any) (*http.Respon
...
-	var err error
-	if gatewayError.Message != "" {
-		err = errors.New(gatewayError.Message)
+func (c *Client) tryDecodeErr(resp io.Reader) error {
+	var gatewayError Error
+	if err := json.NewDecoder(resp).Decode(&gatewayError); err != nil {
+		return err
 	}
-	return err
+	return &gatewayError
```
kirugan, line 163: Here is a problem: json unmarshaling fails only if:

- it's malformed json
- types are not the same as in a struct

So if there is empty json or json without same fields as in a struct - unmarshaling will be ok. On caller side: 
```
err = c.tryDecodeErr(resp.Body)
// decoding failed, at least pass info about http code
if err == nil {
    err = fmt.Errorf("received non 200 status code: %d", resp.StatusCode)
}
```
this condition will be always false. I decided to log `...Decode()` failure instead of passing it as an error in API (thought that "received non 200 status code..." is source of problem and meaningful then decoding of failure message itself)

rpc/handlers.go
```diff
@@ -878,21 +887,35 @@ func (h *Handler) AddInvokeTransaction(invokeTx json.RawMessage) (*AddInvokeTxRe
...
+		return ErrContractBytecodeSizeTooLarge
+	case gateway.DuplicatedTransaction:
+		return ErrDuplicateTx
+	case gateway.InvalidTransactionNonce:
+		return ErrInvalidTransactionNonce
+	case gateway.CompilationFailed:
+		return ErrCompilationFailed
 	default:
-		return jsonrpc.InternalError
+		return jsonrpc.Err(jsonrpc.InvalidParams, err.Error())
```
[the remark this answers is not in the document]
omerfirmak: So since `err.(*gateway.Error)` succeeded, gateway returned us a proper error message. Which means that Juno fulfilled it's job a proxy. We can assume it is an error related to the request rather than an internal error that is related to Juno.

rpc/handlers.go
```diff
@@ -23,19 +24,27 @@ type Gateway interface {
...
+	ErrInternal                     = &jsonrpc.Error{Code: jsonrpc.InternalError, Message: "Internal error"}
+	ErrContractError                = &jsonrpc.Error{Code: 40, Message: "Contract error"}
+	ErrInvalidTransactionNonce      = &jsonrpc.Error{Code: 52, Message: "Invalid transaction nonce"}
+	ErrInsufficientMaxFee           = &jsonrpc.Error{Code: 53, Message: "Max fee is smaller than the validation's actual fee"}
+	ErrInsufficientAccountBalance   = &jsonrpc.Error{Code: 54, Message: "Account balance is smaller than the transaction's max_fee"}
+	ErrValidationFailure            = &jsonrpc.Error{Code: 55, Message: "Account validation failed"}
+	ErrCompilationFailed            = &jsonrpc.Error{Code: 56, Message: "Compilation failed"}
+	ErrContractBytecodeSizeTooLarge = &jsonrpc.Error{Code: 57, Message: "Contract bytecode size is too large"}
+	ErrNonAccount                   = &jsonrpc.Error{Code: 58, Message: "Sender address is not an account contract"}
+	ErrDuplicateTx                  = &jsonrpc.Error{Code: 59, Message: "A transaction with the same hash already exists in the mempool"}
```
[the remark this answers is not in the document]
omerfirmak: Nah, it is just the order we added them here.

clients/gateway/gateway.go
omerfirmak: Nice catch, thank you! Will try to fix and ping you for another review.

omerfirmak: How about [this](https://github.com/NethermindEth/juno/compare/2bc4dfc6fd6ce68151725dad80d93a4af2748eaf..e339b7df7cb64df89988cb6ecc15b55082e2d1a0)

omerfirmak: ping @kirugan

clients/gateway/gateway.go
```diff
@@ -47,7 +64,10 @@ func newTestServer() *httptest.Server {
 
 		// empty request: "{}"
 		emptyReqLen := 4
-		if len(b) <= emptyReqLen {
+		if string(b) == "null" {
```
kirugan, line 67: Is there a test that pass null or you added it just in case?

kirugan (approved):

clients/gateway/gateway.go
omerfirmak: This one does https://github.com/NethermindEth/juno/pull/956/files#diff-e6160c31d5a80924a41c26c6844e59e31250f3ddd0257d370cf6fdfb20a5019fR43-R46

PR #960, opened 2023-07-26
Clarify websocket URL requirements

joshklop:


README.md
```diff
@@ -72,6 +72,7 @@ docker run -d \
 
 You should replace `<YOUR-ETH-NODE> `with your actual Ethereum node address.
 If you're using Infura, your Ethereum node address might look something like: `wss://mainnet.infura.io/ws/v3/your-infura-project-id`.
+Make sure you are using the websocket URL `ws`/`wss` and not the http URL `http`/`https`.
```
kirugan, line 75: @joshklop do we support wss? I haven't found parameter to pass ssl/tls certificate to net.Listener for websocket

joshklop: We don't support it for RPC. This is for connecting to the L1 node. We use geth's RPC client for that.

joshklop:
Thanks @IronGauntlets. We'll need to force merge this one.

README.md
joshklop: Yes, we use websockets for L1 verification. I think @kirugan was asking if we support secure websockets `wss` for RPC.

kirugan: >  @kirugan was asking if we support secure websockets wss for RPC.

yes I thought we use it for RPC, thanks

PR #964, opened 2023-07-27
Adjust ErrInsufficientMaxFee response message to latest starknet spec

wojciechos:
[Latest starknet spec](https://github.com/starkware-libs/starknet-specs/blob/29bab650be6b1847c92d4461d4c33008b5e50b1a/api/starknet_write_api.json#L242C29-L242C112)

### DIFF
--- a/rpc/handlers.go
+++ b/rpc/handlers.go
@@ -24,31 +24,28 @@ type Gateway interface {
 }
 
 var (
-	ErrContractNotFound           = &jsonrpc.Error{Code: 20, Message: "Contract not found"}
-	ErrBlockNotFound              = &jsonrpc.Error{Code: 24, Message: "Block not found"}
-	ErrInvalidTxIndex             = &jsonrpc.Error{Code: 27, Message: "Invalid transaction index in a block"}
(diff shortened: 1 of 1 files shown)


rpc/handlers.go
```diff
@@ -38,7 +38,10 @@ var (
 	ErrClassAlreadyDeclared       = &jsonrpc.Error{Code: 51, Message: "Class already declared"}
 	ErrInternal                   = &jsonrpc.Error{Code: jsonrpc.InternalError, Message: "Internal error"}
 	ErrInvalidTransactionNonce    = &jsonrpc.Error{Code: 52, Message: "Invalid transaction nonce"}
-	ErrInsufficientMaxFee         = &jsonrpc.Error{Code: 53, Message: "Max fee is smaller than the validation's actual fee"}
+	ErrInsufficientMaxFee         = &jsonrpc.Error{
```
IronGauntlets, line 41: let's inline it and if linter complains about `lll` then ignore it

IronGauntlets (approved):

rpc/handlers.go
omerfirmak: done

PR #965, opened 2023-07-27
Unify write API handlers

omerfirmak:


clients/feeder/transaction.go
```diff
@@ -44,22 +45,72 @@ func (fs *FinalityStatus) UnmarshalJSON(data []byte) error {
 	return nil
 }
 
+type TransactionType uint8
+
+const (
+	Invalid TransactionType = iota
```
joshklop, line 51: Why not `iota+1` on the next line? I remember discussing this but forget what the trade-off is.

kirugan: I guess because untyped constant will be converted to Invalid automatically. Like this:
```
...
    return 0, err
```
And later on we can safely compare this returned value with `Invalid` const if we need to

clients/feeder/transaction.go
```diff
@@ -44,22 +45,72 @@ func (fs *FinalityStatus) UnmarshalJSON(data []byte) error {
...
+		return "INVOKE_FUNCTION"
+	case TxnL1Handler:
+		return "L1_HANDLER"
+	default:
+		return "<unknown>"
+	}
+}
+
+func (t TransactionType) MarshalJSON() ([]byte, error) {
+	return []byte(fmt.Sprintf("%q", t.String())), nil
```
kirugan, line 77: in case of %s, %v and %q it's safe to pass just t, without calling `.String()` explicitly

clients/feeder/transaction.go
omerfirmak: > we can safely compare this returned value with Invalid const if we need to

Pretty much

clients/feeder/transaction.go
omerfirmak: Done

rpc/handlers_test.go
```diff
@@ -1742,194 +1742,54 @@ func TestEvents(t *testing.T) {
...
-		expectedDeclareRespB, err := json.Marshal(expectedDeclareResp)
-		require.NoError(t, err)
-
-		mockGateway.EXPECT().AddTransaction(json.RawMessage(declareTxV1)).Return(expectedDeclareRespB, nil)
-
-		resp, handlerErr := handler.AddDeclareTransaction(tx)
-
-		require.Nil(t, handlerErr)
-		assert.Equal(t, *expectedDeclareResp, *resp)
-	})
```
IronGauntlets, line 1848: Can you please explain why you are removing these test cases?

omerfirmak: They provide no additional coverage, both in terms of code coverage or edge case coverage.

AddTransaction works on raw JSON. So it only makes sense to test it for the modifications that it does on the json (ie change invoke type and compress sierra classes) and how it responds to invalid json.

Testing it for all the separate transactions that are just relayed without any modifications is just not necessary because the API of AddTransaction makes no distinction on them.

wojciechos (approved):

PR #966, opened 2023-07-27
Make RPC available with versioned path

omerfirmak:


jsonrpc/http.go
```diff
@@ -16,26 +16,31 @@ const MaxRequestBodySize = 10 * 1024 * 1024 // 10MB
...
+		log:       log,
+		listener:  listener,
 	}
 }
 
 // Run starts to listen for HTTP requests
 func (h *HTTP) Run(ctx context.Context) error {
 	errCh := make(chan error)
 
+	mux := http.NewServeMux()
```
joshklop, line 38: Could we pass the `mux` to the `HTTP` and `Websocket` servers? So we can have versioned routes on both.

omerfirmak: I am done for the day, but feel free to give it a try and update this PR

joshklop: I added the `urlPrefix` to `jsonrpc.Websocket` for now. Reducing the code duplication would require a larger refactor I think.

For another PR, what do you think about using the same HTTP server for `HTTP` and `Websocket`? The `HTTP` handler can serve the root route `/` (and `/v0_4`) and the `Websocket` handler can serve `/websocket` (and `/websocket/v0_4`). This also means we only use one port. Downside is that `ws://localhost:6060/websocket` looks redundant.

omerfirmak: > This also means we only use one port

I think this is pretty neat. We should go for it.

omerfirmak: But maybe there should be standard for this as well?

joshklop (approved):

jsonrpc/http.go
joshklop: A standard would be nice.

Also: maybe we could mux requests ourselves in `ServeHTTP`. If the request has `Upgrade: websocket`, we can send it to the websocket handler. Otherwise, continue using http. Then we wouldn't even need the `/websocket` route. I think this might be what Pathfinder's rpc library already does.

joshklop: I can work on that. Resolving this conversation so we can merge the PR.

PR #970, opened 2023-07-28
Fix post requests sometimes getting context cancelled errors

omerfirmak:


clients/gateway/gateway.go
```diff
@@ -96,28 +96,16 @@ func NewClient(gatewayURL string, log utils.SimpleLogger) *Client {
...
 }
 
 // post performs additional utility function over doPost method
-func (c *Client) post(url string, data any) (io.ReadCloser, error) {
+func (c *Client) post(url string, data any) ([]byte, error) {
 	ctx, cancel := context.WithTimeout(context.Background(), c.timeout)
 	defer cancel()
 
 	resp, err := c.doPost(ctx, url, data)
+	defer resp.Body.Close()
```
kirugan, line 108: if `doPost` returns an error will `resp` be nil ?

omerfirmak: :facepalm: thanks

kirugan (approved):

PR #971, opened 2023-07-28
Use a single HTTP server for RPC

joshklop:


jsonrpc/http.go
```diff
@@ -1,61 +1,23 @@
...
-var _ service.Service = (*HTTP)(nil)
-
 type HTTP struct {
-	rpc       *Server
-	log       utils.SimpleLogger
-	listener  net.Listener
-	urlPrefix string
+	rpc *Server
+	log utils.SimpleLogger
 }
```
omerfirmak, line 14: Instead of moving stuff out of `http.go`, how about moving ws handling to here as well? Since WS is a protocol that is build on top of HTTP, I think that makes sense.

We can have WS to be configurably enabled/disabled on `jsonrpc.HTTP` object.

If enabled, in `ServeHTTP` we can check for `req.URL.Schema` to be "ws" or "wss" and promote the incoming connection to a websocket.

omerfirmak: This way we dont need a separate "/websocket" path either.

joshklop: > how about moving ws handling to here as well?

Very good idea. Will do, thanks.

omerfirmak: But in the current state of this PR, we have a type called jsonrpc.HTTP that doesnt do any HTTP:)

joshklop: True. `HTTPHandler` and `WebsocketHandler` would be better names.

We could also implement `ServeHTTP` and `ServeWebsocket` as receivers on `jsonrpc.Server` like geth. I don't like this as a long-term solution, but it should work for the short-term.

omerfirmak: > HTTPHandler and WebsocketHandler would be better names

How are they different? Because `jsonrpc.Websocket` and `jsonrpc.HTTP` are pretty much the same thing in this PR.

More context on how I am thinking about all this;
My aim for `jsonrpc` was to have a `Server` type that does the actual request handling (ie an implementation of the JSONRPC 2.0 spec) and multiple types like `jsonrpc.HTTP`,  `jsonrpc.Websocket` etc that handle the transport layer which actually listens for requests on a medium and forwards them to a `jsonrpc.Server`.

Juno's `rpc` package consists of just the handlers, it is an implemention of the starknet rpc spec. So it has nothing to do with how we are exposing this functionality (HTTP/WS/IPC). 

So anything that is not directly tied to starknet (transport layer stuff for example) should be in `jsonrpc` package.

joshklop: Appreciate the context.

> Juno's rpc package consists of just the handlers, it is an implemention of the starknet rpc spec. So it has nothing to do with how we are exposing this functionality (HTTP/WS/IPC).

Understood. I like this mental model and will adopt it.

> So anything that is not directly tied to starknet (transport layer stuff for example) should be in jsonrpc package.

My mental model has been: "if it's not directly tied to the `jsonrpc` spec (transport layer stuff for example), it should not be in the `jsonrpc` package."

I've been aiming to move transport-specific code into sub-packages of `jsonrpc` (e.g., `jsonrpc/jwebsocket`) so `jsonrpc` doesn't unnecessarily depend on packages like `nhooyr.io/websocket` and `net/http`.

joshklop: In light of the discussion above, I'll get rid of `rpc/server` and move it to `node`.

Do you have any other improvements in mind?

omerfirmak: > I've been aiming to move transport-specific code into sub-packages of jsonrpc (e.g., jsonrpc/jwebsocket

Hmm, that makes sense too.

> , I'll get rid of rpc/server and move it to node.

Shouldnt we move it to a sub-package in `jsonrpc` then?

joshklop: > Shouldnt we move it to a sub-package in jsonrpc then? Which I think makes the most sense moving forward.

I don't think so. Setting up an `http.Server` in `jsonrpc` (or a sub-package) ties the transport bridge to Juno. E.g., how do we cleanly add new routes like `/` and `/v0_4`?

I think the sub-packages should include minimal handlers that can be pulled in by applications that set up their own servers.

joshklop: Went with the `node/rpc` idea; didn't bother with refactoring `jsonrpc` in this PR. I'll create issues for the other things we discussed.

node/rpc/server.go
```diff
@@ -0,0 +1,214 @@
...
+)
+
+type Server struct {
+	listener net.Listener
+	mux      *http.ServeMux
+}
+
+var _ service.Service = (*Server)(nil)
+
+func New(listener net.Listener, rpcHandler *rpc.Handler, log utils.SimpleLogger) (*Server, error) {
```
omerfirmak, line 25: Now that we are moving towards a single HTTP server, we can just pass a port here instead of a listener, no?

omerfirmak (approved):

node/rpc/server.go
joshklop: Explicit ports cause flaky tests.

omerfirmak: Can't we just pass 0 as port in tests?

PR #975, opened 2023-07-31
Add trie.Key type

### LINKED ISSUE #974: Define our own trie.Key type
We are using a third-party bitset library as our key type. But it is  not properly serialized to a big endian byte array. The maintainer is not acknowledging the issue, so there is no way to fix it upstream.

Define a simple key type like;

``` golang
type Key struct {
   Len byte
   Val felt.Felt
}
```

omerfirmak:
Fixes #974 

Before

```BenchmarkTriePut-16         8191            182102 ns/op            3326 B/op        125 allocs/op```

After

```BenchmarkTriePut-16                 7425            167827 ns/op            2114 B/op         53 allocs/op```

Will need to add a migration as well

### DIFF
--- a/core/transaction.go
+++ b/core/transaction.go
@@ -376,7 +376,7 @@ func VerifyTransactions(txs []Transaction, n utils.Network, protocolVersion stri
 	return nil
 }
 
-const commitmentTrieHeight uint = 64
+const commitmentTrieHeight = 64
 
 // transactionCommitment is the root of a height 64 binary Merkle Patricia tree of the
 // transaction hashes and signatures in a block.
--- a/core/trie/key.go
+++ b/core/trie/key.go
@@ -1,0 +1,119 @@
+package trie
+
+import (
+	"bytes"
+	"encoding/hex"
+	"fmt"
+	"math/big"
+
+	"github.com/NethermindEth/juno/core/felt"
+)
+
+type Key struct {
+	len    uint8
+	bitset [32]byte
+}
+
+func NewKey(length uint8, keyBytes []byte) Key {
+	k := Key{len: length}
+	if len(keyBytes) > len(k.bitset) {
+		panic("bytes does not fit in bitset")
+	}
+	copy(k.bitset[len(k.bitset)-len(keyBytes):], keyBytes)
+	return k
+}
+
+func (k *Key) bytesNeeded() uint {
+	const byteBits = 8
+	return (uint(k.len) + (byteBits - 1)) / byteBits
+}
+
+func (k *Key) inUseBytes() []byte {
+	return k.bitset[len(k.bitset)-int(k.bytesNeeded()):]
+}
+
+func (k *Key) unusedBytes() []byte {
+	return k.bitset[:len(k.bitset)-int(k.bytesNeeded())]
+}
+
+func (k *Key) WriteTo(buf *bytes.Buffer) (int64, error) {
+	if err := buf.WriteByte(k.len); err != nil {
+		return 0, err
+	}
+
+	n, err := buf.Write(k.inUseBytes())
+	return int64(1 + n), err
+}
+
+func (k *Key) UnmarshalBinary(data []byte) error {
+	k.len = data[0]
+	k.bitset = [32]byte{}
+	copy(k.inUseBytes(), data[1:1+k.bytesNeeded()])
+	return nil
+}
+
+func (k *Key) EncodedLen() uint {
+	return k.bytesNeeded() + 1
+}
+
+func (k *Key) Len() uint8 {
+	return k.len
+}
+
+func (k *Key) Felt() felt.Felt {
+	var f felt.Felt
+	f.SetBytes(k.bitset[:])
+	return f
+}
+
+func (k *Key) Equal(other *Key) bool {
+	if k == nil && other == nil {
+		return true
+	} else if k == nil || other == nil {
+		return false
+	}
+	return k.len == other.len && bytes.Equal(k.bitset[:], other.bitset[:])
+}
+
+func (k *Key) Test(bit uint8) bool {
+	const LSB = uint8(0x1)
+	byteIdx := bit / 8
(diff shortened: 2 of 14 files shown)

### CODE REVIEW

File: core/trie/node.go [hunk 1]
```diff
@@ -92,17 +77,16 @@ func (n *Node) UnmarshalBinary(data []byte) error {
...
-	if err != nil {
+	if err := n.Left.UnmarshalBinary(data); err != nil {
+		return err
+	}
+	if err := n.Right.UnmarshalBinary(data[n.Left.EncodedLen():]); err != nil {
 		return err
 	}
-	_, err = n.Right.ReadFrom(stream)
-	return err
+	return nil
```
kirugan, line 91: maybe simplify to `return n.Right.UnmarshalBinary(...)` ?

migration/migration.go
```diff
@@ -226,15 +225,46 @@ func (m *changeTrieNodeEncoding) Before() {
 	}
 }
 
+type node struct {
+	Value *felt.Felt
+	Left  *bitset.BitSet
+	Right *bitset.BitSet
+}
+
+func (n *node) _WriteTo(buf *bytes.Buffer) (int64, error) {
```
kirugan, line 234: Will this migration be applied automatically? or it's just refactoring of old one and in runtime everything is going to be ok ?

core/trie/key.go
```diff
@@ -0,0 +1,119 @@
...
+	length uint8
+	bitset [32]byte
+}
+
+func NewKey(length uint8, keyBytes []byte) Key {
+	k := Key{length: length}
+	if len(keyBytes) > len(k.bitset) {
+		panic("bytes does not fit in bitset")
+	}
+	copy(k.bitset[len(k.bitset)-len(keyBytes):], keyBytes)
```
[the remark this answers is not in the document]
omerfirmak: To maintain a proper big-endian byte array, that way we can use it directly in `Key.Felt()` as well.

and to be able to easily extract individual bits. See `Test`.

core/trie/key.go
```diff
@@ -0,0 +1,119 @@
...
+)
+
+type Key struct {
+	length uint8
+	bitset [32]byte
+}
+
+func NewKey(length uint8, keyBytes []byte) Key {
+	k := Key{length: length}
+	if len(keyBytes) > len(k.bitset) {
```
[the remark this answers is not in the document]
omerfirmak: Yes, that is true.

core/trie/key.go
```diff
@@ -0,0 +1,119 @@
...
+		panic("bytes does not fit in bitset")
+	}
+	copy(k.bitset[len(k.bitset)-len(keyBytes):], keyBytes)
+	return k
+}
+
+func (k *Key) bytesNeeded() uint {
+	const byteBits = 8
+	return (uint(k.length) + (byteBits - 1)) / byteBits
```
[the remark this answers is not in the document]
omerfirmak: We are doing integer division here.

So for `k.length` = 0, we do `(0 + 7) / 8`, which results to `0`
So for `k.length` = 1, we do `(1 + 7) / 8`, which results to `1`
.
.
So for `k.length` = 8, we do `(8 + 7) / 8`, which results to `1`
So for `k.length` = 9, we do `(9 + 7) / 8`, which results to `2`

and so on

so essentially, we are calculating the minimum number of bytes needed to store `k.length` amount of bits.

core/trie/key.go
```diff
@@ -0,0 +1,119 @@
...
+	if err := buf.WriteByte(k.length); err != nil {
+		return 0, err
+	}
+
+	n, err := buf.Write(k.inUseBytes())
+	return int64(1 + n), err
+}
+
+func (k *Key) UnmarshalBinary(data []byte) error {
+	k.length = data[0]
```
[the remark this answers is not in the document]
omerfirmak: If no corruptions in the DB, yes. In a corrupted case, a panic isn't our biggest problem anyways :D

core/trie/node.go
omerfirmak: Thanks, will do.

migration/migration.go
omerfirmak: > or it's just refactoring of old one and in runtime everything is going to be ok ?

Yeap, it is just refactoring an old migration. It should still behave exactly the same.

omerfirmak:
@kirugan ready for a second round of reviews

core/trie/key.go
```diff
@@ -0,0 +1,119 @@
...
+	"bytes"
+	"encoding/hex"
+	"fmt"
+	"math/big"
+
+	"github.com/NethermindEth/juno/core/felt"
+)
+
+type Key struct {
+	length uint8
```
IronGauntlets, line 13: nit: lets change it to `len` as it a go idiom

omerfirmak: Done :+1:

core/transaction.go
```diff
@@ -376,7 +376,7 @@ func VerifyTransactions(txs []Transaction, n utils.Network, protocolVersion stri
 	return nil
 }
 
-const commitmentTrieHeight uint = 64
+const commitmentTrieHeight uint8 = 64
```
kirugan, line 379: Do we need to specify type here? In Go constants are untyped by default

omerfirmak: Removed the type

migration/migration.go
```diff
@@ -232,15 +242,77 @@ func (m *changeTrieNodeEncoding) Before() {
...
+	totalBytes := int64(0)
+
+	valueB := n.Value.Bytes()
+	wrote, err := buf.Write(valueB[:])
+	totalBytes += int64(wrote)
+	if err != nil {
+		return totalBytes, err
+	}
+
+	if n.Left != nil {
```
kirugan, line 265: Is it possible that Left is nil and n.Right is not nil for a node?

core/trie/node.go
omerfirmak: Done

kirugan (approved):

PR #977, opened 2023-07-31
Introduce BucketMigrator

omerfirmak:
Will be used in #975


migration/bucket_migrator.go
```diff
@@ -0,0 +1,106 @@
...
+	m.batchSize = batchSize
+	return m
+}
+
+func (m *BucketMigrator) WithBefore(before func()) *BucketMigrator {
+	m.before = before
+	return m
+}
+
+func (m *BucketMigrator) WithKeyFilter(keyFilter func([]byte) (bool, error)) *BucketMigrator {
```
joshklop, line 63:
```suggestion
func (m *BucketMigrator) WithKeyFilter(keyFilter BucketMigratorKeyFilter) *BucketMigrator {
```

migration/bucket_migrator.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	for iterator.Seek(m.startFrom); iterator.Valid(); iterator.Next() {
+		key := iterator.Key()
+		if !bytes.HasPrefix(key, m.target.Key()) {
+			break
+		}
+
+		if pass, err := m.keyFilter(key); err != nil {
+			return err
```
joshklop, line 86: Do we need to close the iterator here?

migration/bucket_migrator.go
```diff
@@ -0,0 +1,106 @@
...
+
+func (m *BucketMigrator) Migrate(txn db.Transaction) error {
+	remainingInBatch := m.batchSize
+	iterator, err := txn.NewIterator()
+	if err != nil {
+		return err
+	}
+
+	for iterator.Seek(m.startFrom); iterator.Valid(); iterator.Next() {
+		key := iterator.Key()
```
joshklop, line 80: We are doing some interesting things with `key` later on. Should we create a defensive copy to be safe? Same with `value`.

I think we should at least document somewhere that `key` and `value` should not be mutated.

migration/bucket_migrator_test.go
```diff
@@ -0,0 +1,84 @@
...
+				if !bytes.Equal(b, []byte{i}) {
+					return errors.New("shouldve moved")
+				}
+				return nil
+			})
+			if err != nil {
+				return err
+			}
+
+			require.ErrorIs(t, db.ErrKeyNotFound, txn.Get(sourceBucket.Key([]byte{i}), func(b []byte) error { return nil }))
```
joshklop, line 79: nit: can we expand this for readability?

migration/bucket_migrator_test.go
```diff
@@ -0,0 +1,84 @@
...
+			}
+		}
+		return txn.Set(sourceBucket.Key(), []byte{44})
+	}))
+
+	mover.Before()
+	require.True(t, beforeCalled)
+
+	txnBudget := 100
+	for {
```
joshklop, line 38: nit: is it possible to remove this loop from the test? Maybe decreasing the number of elements in the db from `10` to `4` would make that reasonable?

joshklop (changes requested):
Neat idea. Is the plan to transition the `changeTrieNodeEncoding` to be a `BucketMigrator` in a future PR?

omerfirmak:
> Neat idea. Is the plan to transition the `changeTrieNodeEncoding` to be a `BucketMigrator` in a future PR?

We could do that, but I am more interested in writing new ones with this:)

migration/bucket_migrator.go
omerfirmak: We do, thanks!

migration/bucket_migrator.go
omerfirmak: `key` and `value` are copies, so the caller owns these slices.

migration/bucket_migrator_test.go
omerfirmak: Done

migration/bucket_migrator_test.go
omerfirmak: Done

migration/migration.go
```diff
@@ -66,25 +66,28 @@ func MigrateIfNeeded(targetDB db.DB) error {
...
+				return txn.Set(db.SchemaVersion.Key(), versionBytes[:])
+			}); dbErr != nil {
+				return dbErr
+			} else if migrationErr != nil {
+				if !errors.Is(migrationErr, ErrCallWithNewTransaction) {
+					return migrationErr
 				}
-				return nil
-			}); err == nil {
+			} else {
```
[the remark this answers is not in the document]
omerfirmak: Thanks, changed it to this. I bit more complicated to read IMO but I believe this works too.

migration/migration.go
```diff
@@ -66,25 +66,28 @@ func MigrateIfNeeded(targetDB db.DB) error {
 		migration := migrations[i]
 		migration.Before()
 		for {
-			if err = targetDB.Update(func(txn db.Transaction) error {
-				if err = migration.Migrate(txn); err != nil {
-					if !errors.Is(err, ErrCallWithNewTransaction) {
```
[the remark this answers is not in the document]
omerfirmak: > Would the problem also be fixed if we always return err in this case

I dont think we can. We need to be able to `return nil` in `Update` but at the same time be able to check outside of `Update` if `migrationErr` was `ErrCallWithNewTransaction`

joshklop (approved):

PR #988, opened 2023-08-03
enhancement: viper.AutomaticEnv

### LINKED ISSUE #905: Allow Juno to be configured by using environment variables
So instead of using flags or config file, we should be able to use environment variables as configuration as well.

[see](https://pkg.go.dev/github.com/spf13/viper#AutomaticEnv) `viper.AutomaticEnv`

lazyfuhrer:
Closes #905

Allowed Juno to be configured by using environment variables using `viper.AutomaticEnv`


joshklop (changes requested):
Maybe a bit overkill, but I think it would be worthwhile to add tests with environment variables (they would catch things like the panic mentioned below).

Also, I think we need to use `viper.SetEnvKeyReplacer` to ensure environment variables have a `JUNO_` prefix (avoids collisions with the user's existing environment).

omerfirmak:
We need to set env prefix as well.

see https://pkg.go.dev/github.com/spf13/viper#SetEnvPrefix

se can simply set in to "JUNO"

lazyfuhrer:
will keep ya posted 👍🏻

omerfirmak:
Closing because of inactivity

PR #990, opened 2023-08-03
Basic stream handling and gossiping

omerfirmak:


p2p/p2p.go
```diff
@@ -188,3 +201,87 @@ func (s *Service) ListenAddrs() ([]multiaddr.Multiaddr, error) {
...
+// NewStream creates a bidirectional connection to a random peer that implements a set of protocol ids
+func (s *Service) NewStream(ctx context.Context, pids ...protocol.ID) (network.Stream, error) {
+	peers := s.host.Peerstore().Peers()
+	peersCount := peers.Len()
+	if peersCount < 0 {
+		return nil, errors.New("no peers")
+	}
+
+	randomPeerIdx := rand.Intn(peersCount) //nolint: gosec
+	for peerIdx := (randomPeerIdx + 1) % peersCount; ; peerIdx = (peerIdx + 1) % peersCount {
```
asdacap, line 214: What is the end condition here? It will loop forever? Also, you might wanna keep a list of peer with known protocol. I'm not sure how libp2p handle list of peer in its peer store, it could store a lot of peer that does not support the protocol. 

See https://github.com/NethermindEth/juno/pull/815/files#diff-cce63aa475cb9a161db5de4cf7297f3ef5fc058000702e26e6dbcd8c195ac9daR77.

asdacap: Might wanna have a wait and cancellation on context cancelled.

omerfirmak: L223-225 is the end condition. It gives up, after looping once thru the entire peerlist

omerfirmak: > I'm not sure how libp2p handle list of peer in its peer store, it could store a lot of peer that does not support the protocol

Yeap, we can deal with advanced peer management later.

omerfirmak: > Might wanna have a wait and cancellation on context cancelled.

Added explicit checks for them, thanks!


https://github.com/NethermindEth/juno/compare/2ca888f3df1a46114e759f9f6274d7cf50550b9e..6c5c1189572e37f385a713e73db46652e7cf61a8

asdacap (approved):

p2p/p2p.go
```diff
@@ -95,7 +102,7 @@ func makeDHT(p2phost host.Host, network utils.Network, cfgBootPeers string) (*dh
 		}
 	}
 
-	protocolPrefix := protocol.ID(fmt.Sprintf("/starknet/%s", network.String()))
+	protocolPrefix := protocol.ID(fmt.Sprintf("/starknet/%s", snNetwork.String()))
```
kirugan, line 105: can we remove `.String()` please? it's called implicitly for %s

p2p/p2p.go
```diff
@@ -188,3 +203,89 @@ func (s *Service) ListenAddrs() ([]multiaddr.Multiaddr, error) {
 
 	return listenAddrs, nil
 }
+
+// NewStream creates a bidirectional connection to a random peer that implements a set of protocol ids
+func (s *Service) NewStream(ctx context.Context, pids ...protocol.ID) (network.Stream, error) {
+	peers := s.host.Peerstore().Peers()
+	peersCount := peers.Len()
+	if peersCount < 0 {
```
kirugan, line 211: `peers.Len()` has this definition: `func (es IDSlice) Len() int           { return len(es) }`

what is the case when `len(...)` might return negative number?

p2p/p2p.go
```diff
@@ -188,3 +203,89 @@ func (s *Service) ListenAddrs() ([]multiaddr.Multiaddr, error) {
...
+		return nil, errors.New("no peers")
+	}
+
+	randomPeerIdx := rand.Intn(peersCount) //nolint: gosec
```
kirugan, line 215: I haven't found `rand.Seek(...)` in our codebase so probably for the same number of peers `rand.Intn(...)` (it's pseudorandom) will return the same index over and over again. Could there be any issues with it?

p2p/p2p.go
omerfirmak: No idea, but builtin `len` returns an `int` as well

p2p/p2p.go
```diff
@@ -30,18 +33,21 @@ type Service struct {
 	host      host.Host
 	bootPeers string
 	network   utils.Network
+	log       utils.SimpleLogger
 
-	dht *dht.IpfsDHT
+	dht    *dht.IpfsDHT
+	pubsub *pubsub.PubSub
+	topics map[string]*pubsub.Topic
```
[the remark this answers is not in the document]
omerfirmak: I assumed that all `SubscribeToTopic` calls will be made in startup and thru out the normal execution only `PublishOnTopic` calls will be made. But to be on the safe side, will add a RW lock.

p2p/p2p.go
kirugan: I guess my question was why it's `peersCount < 0` and not `peersCount == 0`

omerfirmak: It should've been `<=`

omerfirmak: Fixed

PR #996, opened 2023-08-03
Add buffer to sync's catchUpMode

IronGauntlets:


sync/sync.go
```diff
@@ -197,7 +197,7 @@ func (s *Synchronizer) verifierTask(ctx context.Context, block *core.Block, stat
 				} else {
 					s.HighestBlockHeader = highestBlock.Header
 
-					isBehind := s.HighestBlockHeader.Number > block.Number
+					isBehind := s.HighestBlockHeader.Number > block.Number+uint64(runtime.GOMAXPROCS(0))
```
omerfirmak, line 200: For some machines that institutional users use buffer might be as big as 64-128 blocks since CPUs these days have a lot of hw threads. Should we put a cap on this?

something like?

```suggestion
					isBehind := s.HighestBlockHeader.Number > block.Number+min(16, uint64(runtime.GOMAXPROCS(0)))
```

omerfirmak: Same way, we can put a limit on max fetcherTask number as well.

IronGauntlets: I think it makes sense to add a limit for when to start catch-up mode otherwise 64-128 core CPU won't try to catch unless it is 64-128 blocks behind. However, we should not link it to max fetchers and verifiers because when the sync process initially starts we will not be fully utilising the resources, as a machine with 64-128 cores would only use 16 threads (I have put them into separate commits we can drop the second one if we don't want to limit the fetchers and verifiers).

sync/sync.go
```diff
@@ -265,10 +265,18 @@ func (s *Synchronizer) syncBlocks(syncCtx context.Context) {
...
+	return mProcs
+}
+
 func (s *Synchronizer) setupWorkers() (*stream.Stream, *stream.Stream) {
 	numWorkers := 1
 	if s.catchUpMode {
-		numWorkers = runtime.GOMAXPROCS(0)
+		numWorkers = maxWorkers()
 	}
 	return stream.New().WithMaxGoroutines(numWorkers), stream.New().WithMaxGoroutines(numWorkers)
```
omerfirmak, line 281: Hmm, contrary to what I have done before. I don't think we need to limit `verifierTask` limit.
```suggestion
	return stream.New().WithMaxGoroutines(numWorkers), stream.New().WithMaxGoroutines(runtime.GOMAXPROCS(0))
```

If fetcherTask (even with 1 worker) manages to fetch multiple blocks faster than we should be able to verify them in parallel.

IronGauntlets: Done

omerfirmak (approved):

PR #1016, opened 2023-08-04
New method starknet_traceBlockTransactions

kirugan:


rpc/handlers.go
```diff
@@ -1263,6 +1207,90 @@ func (h *Handler) SimulateTransactions(id BlockID, transactions []BroadcastedTra
...
+func (h *Handler) TraceBlockTransactions(blockHash felt.Felt) ([]TracedBlockTransaction, *jsonrpc.Error) {
+	block, err := h.bcReader.BlockByHash(&blockHash)
+	if err != nil {
+		return nil, ErrBlockNotFound
+	}
+
+	return h.traceBlockTransactions(block)
+}
+
+func (h *Handler) traceBlockTransactions(block *core.Block) ([]TracedBlockTransaction, *jsonrpc.Error) {
```
kirugan, line 1219: there is one flaw with this approach: traceTransaction will execute whole block instead of executing all transactions till provided one (as was before these changes). I'll fix it, just need to figure out correct signature (maybe I'll provided txHash as a hint when to stop or filter function ... TBD)

omerfirmak: Yeah, we can just add an index argument to signify how many transactions to run. Can be done in a separate PR

omerfirmak (approved):
Is it possible to get to %60 patch coverage?

kirugan:
> Is it possible to get to %60 patch coverage?

yep, working on it

rpc/handlers.go
kirugan: Added in this PR

utils/slices.go
```diff
@@ -32,3 +32,13 @@ func Filter[T any](slice []T, f func(T) bool) []T {
 
 	return result
 }
+
+func IndexFunc[T comparable](slice []T, f func(T) bool) int {
```
kirugan, line 36: this function is going to be available in stdlib since go1.21, will remove when we upgrade go version

rpc/handlers.go
```diff
@@ -1213,10 +1214,10 @@ func (h *Handler) TraceBlockTransactions(blockHash felt.Felt) ([]TracedBlockTran
 		return nil, ErrBlockNotFound
 	}
 
-	return h.traceBlockTransactions(block)
+	return h.traceBlockTransactions(block, len(block.Transactions))
 }
 
-func (h *Handler) traceBlockTransactions(block *core.Block) ([]TracedBlockTransaction, *jsonrpc.Error) {
+func (h *Handler) traceBlockTransactions(block *core.Block, maxTransactionIdx int) ([]TracedBlockTransaction, *jsonrpc.Error) {
```
omerfirmak, line 1220: something like `numTxns` sounds better to me, instead of `maxTransactionIdx`. But up to you.

omerfirmak (approved):

omerfirmak:
@kirugan please rebase, let's merge this.

rpc/handlers.go
kirugan: Yep, renamed

Issue #1012, opened 2023-08-04
Make `History` private


IronGauntlets:
[History](https://github.com/NethermindEth/juno/blob/main/core/history.go) is used to keep track of the historical state, however, it is not used outside of the `core` package, therefore, it doesn't need to be public.

amirhasanzadehpy:
Hey! I noticed that there is an open issue #1012 regarding making the `History` struct private. Please let me know if I can proceed with working on this issue.

Thank you!

PR #1021, opened 2023-08-05
made history private

Atoo35:
This PR makes History private
#1012

### DIFF
--- a/core/history.go
+++ b/core/history.go
@@ -15,10 +15,6 @@ type history struct {
 	txn db.Transaction
 }
 
-func NewHistory(txn db.Transaction) *history {
-	return &history{txn: txn}
-}
-
 func logDBKey(key []byte, height uint64) []byte {
 	return binary.BigEndian.AppendUint64(key, height)
 }
--- a/core/history.go
+++ b/core/history.go
@@ -11,27 +11,27 @@ import (
 
 var ErrCheckHeadState = errors.New("check head state")
 
-type History struct {
+type history struct {
 	txn db.Transaction
 }
 
-func NewHistory(txn db.Transaction) *History {
-	return &History{txn: txn}
+func NewHistory(txn db.Transaction) *history {
+	return &history{txn: txn}
 }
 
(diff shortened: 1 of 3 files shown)


omerfirmak (approved):

core/history.go
```diff
@@ -11,27 +11,27 @@ import (
...
 var ErrCheckHeadState = errors.New("check head state")
 
-type History struct {
+type history struct {
 	txn db.Transaction
 }
 
-func NewHistory(txn db.Transaction) *History {
-	return &History{txn: txn}
+func NewHistory(txn db.Transaction) *history {
```
omerfirmak, line 18: we should either make this private or remove it all together

omerfirmak (changes requested):

core/history.go
Atoo35: If we make that private, how to handle the `history_test.go` then? @omerfirmak

omerfirmak: cc @IronGauntlets

IronGauntlets: > we should either make this private or remove it all together

 @Atoo35 Please remove it

>If we make that private, how to handle the history_test.go then?

Please rename the test file to `history_pkg_test.go` (this is not a go convention, just over convention) and change the package name in `history_test.go` from `core_test` to `core`.

Atoo35: How is removing helping? the history test package will error out entirely and so will the state.go. Am I missing something here?

IronGauntlets: There is no need to have a `NewHistory()` as it is a trivial func. Anywhere `NewHistory()` is used it can be replaced with `&history{txn: txn}`. You should only create `New` func when you want to assert all the arguments required to create an instance of the struct, or you do some extra work or if the struct is private and the only way to instantiate is through a public `New` func.

Atoo35:
Thanks a lot for the help guys. I hope this is the correct implementation

omerfirmak (approved):
LGTM, thanks for the contribution!

PR #1022, opened 2023-08-06
Add User-Agent header to all requests sent to feeder and gateway

Brivan-26:
The `User-Agent` header contains the recent Juno version. This can be automated by fetching the latest release tag on the repository, so we don't need to update the header value on each new release. However, this adds an external call on each Http request sent. Please let me know if this trade-off is acceptable.


kirugan:
Version string available inside `node.New(...)` you can pass it to feeder/gateway clients

clients/gateway/gateway.go
```diff
@@ -141,6 +144,7 @@ func (c *Client) doPost(ctx context.Context, url string, data any) (*http.Respon
 	}
 
 	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set("User-Agent", c.version)
```
omerfirmak, line 147: see how user agent should be formatted 

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent#syntax

clients/feeder/feeder.go
```diff
@@ -142,7 +145,7 @@ func newTestServer(network utils.Network) *httptest.Server {
 	}))
 }
 
-func NewClient(clientURL string) *Client {
+func NewClient(clientURL, version string) *Client {
```
joshklop, line 148: Can we have a `WithUserAgent(ua string)` function instead of passing a version in here? That way it's a non-breaking change and more generic.

omerfirmak: Agreed

clients/feeder/feeder.go
```diff
@@ -55,6 +57,11 @@ func (c *Client) WithLogger(log utils.SimpleLogger) *Client {
 	return c
 }
 
+func (c *Client) WithVersion(version string) *Client {
```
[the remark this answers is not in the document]
joshklop: For example, I don't think we should hard-code `Juno... Starknet Client` into the User-Agent field since other Go projects can use this package.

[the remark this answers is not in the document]
joshklop: I believe my suggestion satisfies the issue description. We just need to add the `.WithUserAgent("Juno... Starknet Client")` suffix to wherever we create the client in `node.go`.

joshklop (commented):
Thanks! Can we add sub-tests for the user-agent in `feeder_test.go` and `gateway_test.go` to ensure it is set properly?

joshklop (approved):

node/node.go
```diff
@@ -96,10 +96,9 @@ func New(cfg *Config, version string) (*Node, error) { //nolint:gocyclo
 	}
 
 	chain := blockchain.New(database, cfg.Network, log)
-	client := feeder.NewClient(cfg.Network.FeederURL())
-
+	client := feeder.NewClient(cfg.Network.FeederURL()).WithUserAgent(fmt.Sprintf("Juno/%s Starknet Client", version))
```
joshklop, line 99: nit: we could factor out the user agent to a variable here.

PR #1024, opened 2023-08-06
Trie Proof Nodes and Iterate

asdacap:
- Basic iterate leaf, generate and verify proofs. 
- Iterate using trie visit. Dont wanna mess around with multiple implementation of storage right now.
- Needed for snap sync.

### DIFF
--- a/core/trie/trie.go
+++ b/core/trie/trie.go
@@ -359,11 +359,6 @@ func (t *Trie) put(nodeKey Key, value *felt.Felt, isProof bool) (*felt.Felt, err
 		// Replace if key already exist
 		sibling := nodes[len(nodes)-1]
 		if nodeKey.Equal(sibling.key) {
-			if isProof {
-				// Ignore setting proof. It'll fail verification if not match anyway
-				return nil, nil
-			}
-
 			// we have to deference the Value, since the Node can released back
 			// to the NodePool and be reused anytime
 			old = *sibling.node.Value // record old value to return to caller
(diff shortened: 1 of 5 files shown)


core/trie/trie.go
```diff
@@ -192,16 +192,143 @@ func (t *Trie) Get(key *felt.Felt) (*felt.Felt, error) {
...
+		}
+	}
+
+	return t.doIterate(startValue, thenode.Right, consumer)
+}
+
+type ProofNode struct {
+	Key  *Key
+	Hash *felt.Felt
+}
```
omerfirmak, line 242: We need to come up with a standard way of representing a proof node across different implementation.
I am not sure if pathfinder will be able to consume this proof.

We probably need to use the definition from the spec. https://docs.starknet.io/documentation/architecture_and_concepts/State/starknet-state/#specifications

What spec defines as `path` is a bit different. We calculate that in `path(key, parentKey *Key) Key`

omerfirmak: https://github.com/starknet-io/starknet-p2p-specs/pull/8/files#diff-1895f60c4bbe4eec02ba055cd529fd9ad7cb4b76bad983cbc68a2d1dc0ce0190R6-R21

asdacap: The spec specification for path seems to be for calculating the commitment, not the actual tree path. You can assume this as `JunoSnapProofNode` if you want and rename it later. As for the p2p spec, I thought we would rather make it juno specific first. Or would you rather have me dump the whole snap prototype branch and you guys can figureout how to implement it?

asdacap: I'll assume you'll adapt it accordingly.

Issue #1019, opened 2023-08-04
Make HTTP server address configurable


joshklop:
Right now Juno [listens](https://github.com/NethermindEth/juno/blob/ac24a44f14f6c09db836613f87b87984d54f4a6e/node/node.go#L313) on `0.0.0.0`, which makes it publicly available.

This should be configurable and the default should be `localhost`.

To implement this, we can probably have a single `--http <addr>:<port>` option (this will work for metrics, GRPC, JSON-RPC, etc. when #986 happens).

PR #1025, opened 2023-08-06
make http addr configurable

Exca-DK:
Implemets #1019. 
To note: It includes an extra flag for the address instead of collapsing it with the port. I believe that this is a better approach, as seen in most other packages that have opted for this way.

### DIFF
--- a/cmd/juno/juno.go
+++ b/cmd/juno/juno.go
@@ -28,6 +28,7 @@ Juno is a Go implementation of a Starknet full node client created by Nethermind
 const (
 	configF              = "config"
 	logLevelF            = "log-level"
+	httpHostF            = "http-host"
 	httpPortF            = "http-port"
 	wsPortF              = "ws-port"
 	grpcPortF            = "grpc-port"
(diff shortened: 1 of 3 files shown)


joshklop (approved):
LGTM! Thanks! If you could fix the lint errors, that would be awesome.

We should try to get this and #1036 in the same release so we don't have multiple servers listening on the different default addresses.

omerfirmak:
#1036 is merged, please rebase @Exca-DK

omerfirmak:
Needs another rebase :(

joshklop (approved):

Ehsan-saradar:
@Exca-DK I think it needs another rebase!

omerfirmak:
Hey, we refactored our http servers. Do you mind rebasing and adapting your PR?

Exca-DK:
Hey @omerfirmak . I rebased and adopted this pr.  I also noticed that now every host and port flag is almost identical. They could be collapsed into one and just formatted with %s.

joshklop:
This looks good to me. Mind making changes to the example configuration in `docs/docs/example_config.md`?

> now every host and port flag is almost identical

I agree it's not ideal from a configuration perspective, but it allows for fine-grained access control based on the host and port.

joshklop (approved):

PR #1027, opened 2023-08-07
add missing features to  juno_test.go

aminsato:
Refine configuration handling and update test defaults

i added complete cfgFileContents to section with  text "config file with all settings but without any other flags"

i added complete flags to section with text  "all flags without config file"


IronGauntlets:
@aminsato thank you very much for the pull request.

The purpose of the cmd package tests is to make sure we have the correct precedence order and since we are relying on viper to provide that we don't need to enumerate all the different possible flags.

Therefore, I don't this the pr is necessary.

Issue #984, opened 2023-08-02
Log warning for data migration


wojciechos:
During data migrations, the node become unresponsive. This could potentially confuse or frustrate users who are not aware that a migration is in progress. We need to add warning logs at key stages of the data migration process to alert users about potential node disruptions.

PR #1031, opened 2023-08-07
migration: log stages

Exca-DK:
Implements #984

### DIFF
--- a/migration/migration.go
+++ b/migration/migration.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/binary"
 	"errors"
+	"fmt"
 	"runtime"
 	"sync"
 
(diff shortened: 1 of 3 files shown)


migration/migration.go
```diff
@@ -78,6 +79,9 @@ func MigrateIfNeeded(targetDB db.DB, network utils.Network) error {
 	}
 
 	for i := version; i < uint64(len(migrations)); i++ {
+		if log != nil {
+			log.Infow("applying database migration", "stage", fmt.Sprintf("%d/%d", i+1, uint64(len(migrations))))
```
omerfirmak, line 83:
```suggestion
			log.Infow("applying database migration", "stage", fmt.Sprintf("%d/%d", i+1, len(migrations)))
```

why do you need the casting here?

Exca-DK: Leftover from the previous commit, my bad.

joshklop, line 83:
```suggestion
			log.Infow("Applying database migration", "stage", fmt.Sprintf("%d/%d", i+1, uint64(len(migrations))))
```
Logs should be capitalized.

nit: we don't use the `if log != nil { ... }` and always assume the log is initialized. For consistency and to save a few lines here, I think we can safely use `log` directly. What do you think?

Exca-DK: Yeah, it's alright. I wasn't aware of that assumption.

Issue #1028, opened 2023-08-07
Ungraceful Shutdown on Network Disconnection


wojciechos:
When the network connection is lost, the Juno unexpectedly closes itself instead of handling the network loss gracefully.

**Steps to Reproduce**:

1. Start Juno(with --eth-node flag)
2. While the node is running, disconnect the network.
3. Wait a while and observe that the node shuts down unexpectedly.

**Expected Behavior:**

The node should handle network disconnects gracefully by attempting to reconnect or pausing its operations until the network is available again. It should not shut down completely.

**Current Behavior:**
```10:27:37.756 07/08/2023 +00:00  DEBUG   l1/l1.go:130    Received L1 LogStateUpdate      {"number": "144455", "stateRoot": "21837cb2a61b4105b079a731d4298ad3749fd8a02c94cda325e02b1999f30b1", "blockHash": "607cf2e3f7ace3f1b67aff5cdfb5b6c7ee3255f8ec332ca05622a10bd215674"}
10:28:26.968 07/08/2023 +00:00  WARN    l1/l1.go:107    L1 header subscription failed, resubscribing    {"error": "websocket: close 1006 (abnormal closure): unexpected EOF"}
10:28:26.971 07/08/2023 +00:00  ERROR   node/node.go:339        Service error   {"name": "*l1.Client", "err": "subscribe to L1 headers: dial tcp: lookup mainnet.infura.io on 192.168.65.7:53: no such host"}
github.com/NethermindEth/juno/node.(*Node).Run.func2
        /app/node/node.go:339
github.com/sourcegraph/conc/panics.(*Catcher).Try
        /root/go/pkg/mod/github.com/sourcegraph/conc@v0.2.0/panics/panics.go:23
github.com/sourcegraph/conc.(*WaitGroup).Go.func1
        /root/go/pkg/mod/github.com/sourcegraph/conc@v0.2.0/waitgroup.go:32
10:28:26.971 07/08/2023 +00:00  INFO    node/node.go:348        Shutting down Juno...```

omerfirmak:
Actually, this is a graceful shutdown.

The problem is we shouldn't shutdown the node when we fail to connect to the ETH node. We just need to log the error and retry later.

PR #1033, opened 2023-08-07
Re-attempt re-subscriptions

joshklop:
Fixes: #1028

### DIFF
--- a/l1/l1.go
+++ b/l1/l1.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"math/big"
+	"time"
 
 	"github.com/NethermindEth/juno/blockchain"
 	"github.com/NethermindEth/juno/core"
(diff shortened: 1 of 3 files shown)


l1/l1.go
```diff
@@ -41,24 +43,46 @@ func NewClient(l1 Subscriber, chain *blockchain.Blockchain, log utils.SimpleLogg
...
+		case <-ctx.Done():
+			return nil, fmt.Errorf("context canceled before resubscribe was successful: %w", ctx.Err())
+		default:
+			headerSub, err := c.l1.WatchHeader(ctx, headerChan)
+			if err == nil {
+				return headerSub, nil
+			}
+			c.log.Warnw("Failed to subscribe to L1 headers", "tryAgainIn", c.resubscribeDelay, "err", err)
+			time.Sleep(c.resubscribeDelay)
+		}
```
omerfirmak, line 68: Do you think it is worth extracting this loop to a function to be reused in subscribeToUpdates

omerfirmak (approved):

l1/l1.go
joshklop: Probably more effort than it's worth. The real solution is to drop `WatchHeader` altogether, since we just use it as a `time.Ticker`.

PR #1035, opened 2023-08-07
Remove unnecessary header subscription

joshklop:


joshklop:
Related thought: I would love to have the following interface in the `sync` package. 

```go
type L1 interface {
  WatchL1Head(ctx context.context, sink chan<- *core.L1Head) (event.Subscription, error)
}
```

and refactor the `l1` package to implement this interface for Ethereum (probably rename `l1` package to `ethereum` in the process). 

Benefits:
- By implementing this interface, users could use Juno to deploy custom Starknet instances on other chains (including Starknet)
- Beacon sync and reorg handling become a bit easier, since the sync loop will control the implementation of `L1`.

I've tried to do it locally, but the tests are very difficult to get right. Just mentioning this here in case someone has thoughts.

l1/l1.go
```diff
@@ -12,39 +12,39 @@ import (
...
-		log:              log,
-		network:          chain.Network(),
-		resubscribeDelay: 10 * time.Second,
-		nonFinalisedLogs: make(map[uint64]*contract.StarknetLogStateUpdate, 0),
+		l1:                    l1,
+		l2Chain:               chain,
+		log:                   log,
+		network:               chain.Network(),
+		resubscribeDelay:      10 * time.Second,
+		pollFinalisedInterval: 5 * time.Second,
```
omerfirmak, line 46: 5 secs is probably too frequent. Block time is 12 seconds on ethereum anyways. Even something like a minute is reasonable IMO.

omerfirmak: Ah, forgot to approve. It is up to you to tweak the interval or not :+1:

joshklop: Will tweak the interval, thanks.

PR #1036, opened 2023-08-08
Consolidate HTTP servers

joshklop:
So Juno uses two ports instead of five. Easier to configure one server than four.


omerfirmak (approved):
Very nice. LGTM.

Let's not forget to detail these changes in next release changelogs @wojciechos

PR #1037, opened 2023-08-08
Dont traverse the trie if updating an existing leaf

omerfirmak:
Updating an existing leaf is more much common than writing to a non-existing one. For example; think of how often a contract updates its state as opposed to how often a new contract is deployed. So I tried to optimize for the more common case.

Before;
```
BenchmarkMainnetBlock-16              10         351087330 ns/op         9974978 B/op     161989 allocs/op
--- BENCH: BenchmarkMainnetBlock-16
    blockchain_test.go:789: block num 105770
```

After;
```
BenchmarkMainnetBlock-16              10         315338643 ns/op         8077657 B/op     131497 allocs/op
--- BENCH: BenchmarkMainnetBlock-16
    blockchain_test.go:789: block num 105770
```


core/trie/trie.go
```diff
@@ -234,17 +248,9 @@ func (t *Trie) Put(key, value *felt.Felt) (*felt.Felt, error) {
 			// we have to deference the Value, since the Node can released back
 			// to the NodePool and be reused anytime
 			old = *sibling.node.Value // record old value to return to caller
-			if value.IsZero() {
```
joshklop, line 237: nit: reword the now-outdated comment `Replace if key already exist` a few lines up from here. Maybe `Delete` instead of `Replace`

joshklop (approved):

core/trie/trie.go
omerfirmak: Done :+1:

PR #1039, opened 2023-08-08
add warning user if running out of free disk space

Brivan-26:


node/node.go
```diff
@@ -19,7 +19,7 @@ import (
 	"github.com/NethermindEth/juno/l1"
 	"github.com/NethermindEth/juno/metrics"
 	"github.com/NethermindEth/juno/migration"
-	"github.com/NethermindEth/juno/node/http"
+	junohttp "github.com/NethermindEth/juno/node/http"
```
joshklop, line 22: `node/http` should be available as `junohttp` already.

Brivan-26: Can you elaborate more? You mean provide both imports ?
Btw, that import renaming came form Go's tooling.

joshklop: I mean that `github.com/NethermindEth/juno/node/http` should be enough, we don't need the `junohttp` before the import path.

> import renaming came form Go's tooling.

Interesting! Maybe I'm not aware of a best practice. Can you point me to the tool that did that? My LSP/linting setup doesn't complain, and neither does `golangci-lint run`.

Brivan-26: [Go's official Extension](https://marketplace.visualstudio.com/items?itemName=golang.Go)
I guess the auto renaming import is due to the how the node/http package [was referred](https://github.com/NethermindEth/juno/blob/02e7ebfeaddc7ff9209ed269f8dafaa33e193d07/node/node.go#L101) to

utils/network.go
```diff
@@ -10,7 +10,10 @@ import (
 	"github.com/spf13/pflag"
 )
 
-var ErrUnknownNetwork = errors.New("unknown network (known: mainnet, goerli, goerli2, integration)")
+var (
+	ErrUnknownNetwork = errors.New("unknown network (known: mainnet, goerli, goerli2, integration)")
+	requiredSpace     = map[string]float64{"mainnet": 38.0, "goerli": 34.0, "goerli2": 6.5}
```
eliotstock, line 15: Who is going to update this as the chain grows?

Can it not be calculated, even very roughly, from the current chain height and the storage used per block? (We'd need this to execute after we know the current chain height though.)

Brivan-26: I brought that into discussion on how can we know the size of the chain in advance. Tho, I didn't see an approach for that rather than hard-code it and update it by time. 
Your suggestion seems great, we can know the current chain height by fetching the latest block, but what about `storage used per block`, is there a way to get it ? And if there's, I assume we need to take kind of an average, do you suggest an approach to calculate that average ?

IronGauntlets: The chain size is very arbitrary as `mainnet` and `goerli` are roughly the same size but the difference is the number of blocks is around 650,000.

We can calculate the average size of the block and multiply it with the remaining blocks to be synced.

 Given this is a rough calculation, we may stop someone from syncing when there is enough disk space due to an inaccurate estimate.

IronGauntlets:
If someone runs Juno with an existing db, we need to ensure that it is accounted for when `AvailableDiskSpace` and `RequiredDiskSpace` is calculated. Otherwise, a fully synced peer with less than 30 GB of disk space available will not be able to sync.

IMO, handling all of the different edge cases does not give much benefit over stating disk requirements in the `README.md`

Brivan-26:
I do agree. It is much effort with almost no extra benefit, this issue just attempts to save users from reading a table in the README file.

PR #1042, opened 2023-08-08
node/http: pprof endpoint fix

Exca-DK:


node/http/server.go
```diff
@@ -9,7 +9,7 @@ import (
 	"time"
 
 	"github.com/NethermindEth/juno/db"
-	"github.com/NethermindEth/juno/grpc"
+	junogrpc "github.com/NethermindEth/juno/grpc"
```
Exca-DK, line 12: The rename is due to go default formatter behavior.

joshklop: I believe this is due to VSCode defaults. Are you able to change the VSCode formatter to `gofumpt` for this project?

joshklop:
Interesting that the test did not catch this issue...

node/http/server.go
omerfirmak: Can we revert this change? Since it is unrelated.

node/http/server.go
```diff
@@ -62,7 +62,7 @@ func New(listener net.Listener, metrics bool, database db.DB, version string,
 	mux.Handle("/grpc", grpcHandler)
 	if pprofEnable {
 		// Taken from https://artem.krylysov.com/blog/2017/03/13/profiling-and-optimizing-go-web-applications/
```
joshklop, line 64: Let's remove this comment now that we're changing the code. Almost identical lines are in the stdlib anyway.

node/http/server.go
Exca-DK: Sure.

Exca-DK:
> Interesting that the test did not catch this issue...
> 
> EDIT: just realized this is a change to align behavior with stdlib, not a bug fix.

It is in fact a bugfix. Without the slash it is impossible to access /debug/pprof/heap or /debug/pprof/goroutine

joshklop (approved):
LGTM.

PR #1046, opened 2023-08-09
Log version numbers at the startup

aminsato:


cmd/juno/juno.go
```diff
@@ -83,7 +83,9 @@ func main() {
 	config := new(node.Config)
 	cmd := NewCmd(config, func(cmd *cobra.Command, _ []string) error {
 		fmt.Printf("%s\n\n", greeting)
-
+		// Log the version number
+		fmt.Printf("Version: %s\n", Version)
+
```
IronGauntlets, line 88: Let's not log the versions here, instead let's use [spew](https://pkg.go.dev/github.com/davecgh/go-spew/spew) to pretty print the config being printed on line `node/node.go:165`.

cmd/juno/juno.go
IronGauntlets, line 1: I don't think any change needs to be made in this file. Please have a look at `node/node.go:165` which already prints the version but we need to do in multiple lines, you can use spew for that

IronGauntlets (changes requested):

cmd/juno/juno.go
```diff
@@ -82,8 +82,12 @@ func main() {
...
 	config := new(node.Config)
 	cmd := NewCmd(config, func(cmd *cobra.Command, _ []string) error {
-		fmt.Printf("%s\n\n", greeting)
-
+		log, err := utils.NewZapLogger(utils.INFO, true)
+		if err != nil {
+			return err
+		}
+		// Log the greeting and version
+		log.Info(fmt.Sprintf(greeting, Version))
```
IronGauntlets, line 90: I don't want to instantiate a logger here, please change it to the following 
```suggestion
		fmt.Sprintf(greeting, Version))
```

once you have made this change I will approve

cmd/juno/juno.go
```diff
@@ -82,8 +82,7 @@ func main() {
 
 	config := new(node.Config)
 	cmd := NewCmd(config, func(cmd *cobra.Command, _ []string) error {
-		fmt.Printf("%s\n\n", greeting)
-
+		fmt.Sprintf(greeting, Version))
```
IronGauntlets, line 85:
```suggestion
		fmt.Sprintf(greeting, Version)
```

cmd/juno/juno.go
```diff
@@ -21,7 +21,7 @@ const greeting = `
       | |_   _ _ __   ___   
   _   | | | | | '_ \ / _ \  
  | |__| | |_| | | | | (_) |  
-  \____/ \__,_|_| |_|\___/  
+  \____/ \__,_|_| |_|\___/  %s
 
 Juno is a Go implementation of a Starknet full node client created by Nethermind.`
```
IronGauntlets, line 26:
```suggestion
  \____/ \__,_|_| |_|\___/  %s

Juno is a Go implementation of a Starknet full-node client created by Nethermind.

`
```

cmd/juno/juno.go
```diff
@@ -21,9 +21,11 @@ const greeting = `
       | |_   _ _ __   ___   
   _   | | | | | '_ \ / _ \  
  | |__| | |_| | | | | (_) |  
-  \____/ \__,_|_| |_|\___/  
+  \____/ \__,_|_| |_|\___/  %s
```
IronGauntlets, line 24:
```suggestion
  \____/ \__,_|_| |_|\___/ %s
```

cmd/juno/juno.go
```diff
@@ -82,8 +84,7 @@ func main() {
 
 	config := new(node.Config)
 	cmd := NewCmd(config, func(cmd *cobra.Command, _ []string) error {
-		fmt.Printf("%s\n\n", greeting)
-
+		fmt.Sprintf(greeting, Version)
```
IronGauntlets, line 87:
```suggestion
		fmt.Printf(greeting, Version)
```

cmd/juno/juno.go
```diff
@@ -82,8 +84,7 @@ func main() {
 
 	config := new(node.Config)
 	cmd := NewCmd(config, func(cmd *cobra.Command, _ []string) error {
-		fmt.Printf("%s\n\n", greeting)
-
+		fmt.Printf(greeting, Version)
```
IronGauntlets, line 87:
```suggestion
		fmt.Printf(greeting, Version)
		
```

cmd/juno/juno.go
```diff
@@ -82,8 +84,8 @@ func main() {
...
+
```
IronGauntlets, line 88:
```suggestion
```

PR #1047, opened 2023-08-09
Change History to private

### LINKED ISSUE #1012: Make `History` private

aminsato:
closed #1012
Hi
I change history to private

### DIFF
--- a/core/history.go
+++ b/core/history.go
@@ -11,27 +11,27 @@ import (
 
 var ErrCheckHeadState = errors.New("check head state")
 
-type History struct {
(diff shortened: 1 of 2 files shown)


omerfirmak:
Thanks for the PR! But there is an existing PR that is being reviewed right now. https://github.com/NethermindEth/juno/pull/1021

PR #1054, opened 2023-08-09
Pretty print juno config

IronGauntlets:


IronGauntlets:
@joshklop please review again I added `MarshalJSON()` to `LogLevel` and `Network` so that config would print strings instead of ints.

IronGauntlets:
>Right now we only permit yaml config files. This prints the config in json format. I think it would be useful to copy-paste the printed config to a file.
>
>What do you think about accepting json configs (instead of yaml), or pretty-printing yaml?

I am happy with either, viper can easily consume json or yaml (using `mapstructure` tag makes it easy to unmarshal).

Please create a new issue and add `good first issue` tag to it.

IronGauntlets:
image

kirugan (approved):

cmd/juno/juno.go
```diff
@@ -85,6 +86,13 @@ func main() {
 	config := new(node.Config)
 	cmd := NewCmd(config, func(cmd *cobra.Command, _ []string) error {
 		fmt.Printf(greeting, Version)
+
+		configJSON, err := json.MarshalIndent(config, "", "  ")
```
kirugan, line 90: Cool method, I wasn't aware of it

utils/log_test.go
```diff
@@ -72,6 +72,19 @@ func TestLogLevelUnmarshalText(t *testing.T) {
 	})
 }
 
+func TestLogLevelkMarshalJSON(t *testing.T) {
```
joshklop, line 75:
```suggestion
func TestLogLevelMarshalJSON(t *testing.T) {
```

joshklop: We can iterate over `levelStrings` in the test; that way the test will continue to work if we add/remove log levels. Same for the new `utils.Network` test.

IronGauntlets: Done

IronGauntlets: Make sense. Although, I don't think we need to test trivial functions. I will revert the map changes.

IronGauntlets: Done

wojciechos:
@IronGauntlets seems like PendingPollInterval is displayed in different format than used to be, e.g for 5s it displays 5000000000:

``` 
root@localhost:~# docker run nethermindeth/juno:v0.5.0-rc0-62-g88f476a --pending-poll-interval 5s
      _
      | |
      | |_   _ _ __   ___
  _   | | | | | '_ \ / _ \
 | |__| | |_| | | | | (_) |
  \____/ \__,_|_| |_|\___/ v0.5.0-rc0-62-g88f476a

Running Juno with Config:
{
  "LogLevel": "info",
  "HTTPPort": 6060,
  "DatabasePath": "/var/lib/juno",
  "Network": "mainnet",
  "EthNode": "wss://mainnet.infura.io/ws/v3/81fa89d1cba1494ab1b30f5755fb8989",
  "Pprof": true,
  "Colour": false,
  "PendingPollInterval": 5000000000,
  "Metrics": true,
  "P2P": false,
  "P2PAddr": "",
  "P2PBootPeers": ""
}

PR #1057, opened 2023-08-09
Add base context to HTTP server

joshklop:
The default is `context.Background`, so handlers that use `req.Context()` won't actually be cancelled by default.


kirugan:
But I guess that's the point of code below:
```
select {
	case <-ctx.Done():
		return httpSrv.Shutdown(context.Background())
	case err := <-errCh:
		return err
	}
```
process receives signal -> context passed `Run(...)` are cancelled -> we start shutdown procedure (new connects are not accepted) -> wait till requests that were in flight are processed -> return from `Run(...)` method

joshklop:
Would that work for long-running requests like subscriptions?

kirugan:
> Would that work for long-running requests like subscriptions and our `TransactionStatus` handler?

Yes, I made an example [here](https://gist.github.com/kirugan/168c7e1f7d4e046cf3ec53974740ff21). 

My point is if a user of Juno don't want to wait for all requests to complete then they can easily stop it with kill -9 <pid>, otherwise best practice is to wait till existing requests complete then shutdown.

joshklop:
Got it. If that's the desired behavior, then I agree this change isn't necessary.

joshklop:
I was thinking about this again today @kirugan. Are we sure we want a single websocket subscription to give the appearance of hanging the application on `Ctrl-c`? The handlers still have the chance to clean-up, send close messages, etc. before returning, which should resolve the in-flight requests we are concerned about.

What do you think?

Issue #1056, opened 2023-08-09
Pretty-print YAML config instead of JSON


joshklop:
Right now we pretty-print the config when Juno starts, but it's printed in JSON format. Since we only permit YAML config files, it would be nice to pretty-print in YAML to allow the config to be copy-pasted (and just for general consistency).

aminsato:
HI @joshklop
I have done this #1061
Please review my PR

PR #1061, opened 2023-08-10
Print config with YAML format

aminsato:
@joshklop 
HI

I added the code that print config  in YAML FORMAT
#1056


cmd/juno/juno.go
```diff
@@ -91,6 +92,13 @@ func main() {
...
 			return err
 		}
+
+		yamlConfig, err := yaml.Marshal(config)
+		if err != nil {
+			return err
+		}
+		fmt.Printf("Running Juno with Config:\n%s\n\n", string(yamlConfig))
+
 		fmt.Printf("Running Juno with Config:\n%s\n\n", string(configJSON))
```
omerfirmak, line 102: Instead of printing it separately, you should replace the json here.

aminsato: > Instead of printing it separately, you should replace the json here.

I Change it.
Is that ok?

omerfirmak: Yeap, please run `make format` as well.

omerfirmak: I get;

Screenshot from 2023-08-10 11-21-34

@IronGauntlets @joshklop please take a look if that is what you had in mind.

joshklop: Looks very clean. However, before merging we should aim to make the printed config copy-pastable. That means using the mapstructure keys when unmarshaling (e.g., `db-path` instead of `databasepath`). Otherwise it could be misleading and frustrating for users.

aminsato:
I used mapstructure keys when unmarshaling config

This is sample output 

```
       _                    
      | |                   
      | |_   _ _ __   ___   
  _   | | | | | '_ \ / _ \  
 | |__| | |_| | | | | (_) |  
  \____/ \__,_|_| |_|\___/ v0.5.0-rc0-63-gbd1485b9

Running Juno with Config:
colour: true
db-path: ""
eth-node: ""
http-port: 6060
log-level: 1
metrics: false
network: 1
p2p: false
p2p-addr: ""
p2p-boot-peers: ""
pending-poll-interval: 0s
pprof: false


```

joshklop:
Looking good! It appears the network and log-level got changed to numbers. Let's unmarshal them using their `*.String()` receivers.

nit: I think we can make this a `String()` receiver on the config struct. Should de-clutter the code a bit.

aminsato:
@joshklop 
problem fixed


sample output
```

       _                    
      | |                   
      | |_   _ _ __   ___   
  _   | | | | | '_ \ / _ \  
 | |__| | |_| | | | | (_) |  
  \____/ \__,_|_| |_|\___/ v0.5.0-rc0-63-gbd1485b9

Running Juno with Config:
colour: true
db-path: ""
eth-node: ""
http-port: 6060
log-level: info
metrics: false
network: 1
p2p: false
p2p-addr: ""
p2p-boot-peers: ""
pending-poll-interval: 0s
pprof: false

```

aminsato:
I added MarshalYAMLfot the network.

sample output:
Running Juno with Config:
colour: true
db-path: ""
eth-node: ""
http-port: 6060
log-level: info
metrics: false
network: mainnet
p2p: false
p2p-addr: ""
p2p-boot-peers: ""
pending-poll-interval: 0s
pprof: false

joshklop (approved):
Note to self:

- I just realized that by printing the config in `cmd/juno` we always print an empty db path. We moved it here so the config didn't interrupt the log messages that we are printing in `node.New`. I have some ideas for how to refactor this and move the config log message back to `node.Run`.

Since this PR strictly improves on `main`, I think we can merge it once the CI checks pass.

joshklop:
@omerfirmak Do you mind force-merging this? Codecov won't pass.

Issue #1058, opened 2023-08-09
Refactor `rpc.TransactionStatus` to handle the case when the Tx hash doesn't exist


Brivan-26:
Since some feeder's error codes were specified in v0.12.1 (including [TRANSACTION_NOT_FOUND](https://github.com/starkware-libs/cairo-lang/blob/cc86dfc40202fd74370734ef8b9414a2d78dcc06/src/starkware/starknet/definitions/error_codes.py#L67)), we should cover the case of unexisting Tx hash [here](https://github.com/NethermindEth/juno/blob/c65ecfe2cc4b4472cd1f717218a5bc649281774d/rpc/handlers.go#L1034) and send the appropriate error code

aminsato:
Hi
I have done this #1062
please review my PR
Thanks

PR #1062, opened 2023-08-10
added TransactionNotFound

aminsato:
#1058
Hi @Brivan-26
I added TransactionNotFound cost to cover Tx hash error


jsonrpc/server.go
```diff
@@ -17,11 +17,12 @@ import (
...
-	InvalidRequest = -32600 // The JSON sent is not a valid Request object.
-	MethodNotFound = -32601 // The method does not exist / is not available.
-	InvalidParams  = -32602 // Invalid method parameter(s).
-	InternalError  = -32603 // Internal JSON-RPC error.
+	InvalidJSON         = -32700 // Invalid JSON was received by the server.
+	InvalidRequest      = -32600 // The JSON sent is not a valid Request object.
+	MethodNotFound      = -32601 // The method does not exist / is not available.
+	InvalidParams       = -32602 // Invalid method parameter(s).
+	InternalError       = -32603 // Internal JSON-RPC error.
+	TransactionNotFound = -32004 // Transaction JSON-RPC error.
```
joshklop, line 25: The issue is specific to the Starknet spec and not applicable for all jsonrpc use cases. It should be solved in the [rpc package](https://github.com/NethermindEth/juno/blob/8cf2dff7f2fc2561f4fe5b9cbf863da033d00453/rpc/handlers.go#L31C2-L31C2) and not the `jsonrpc` package.

jsonrpc/server.go
```diff
@@ -56,6 +56,7 @@ func Err(code int, data any) *Error {
 		return &Error{Code: MethodNotFound, Message: "Method Not Found", Data: data}
 	case InvalidParams:
 		return &Error{Code: InvalidParams, Message: "Invalid Params", Data: data}
+
```
joshklop, line 59: Spurious diff.

rpc/handlers.go
```diff
@@ -1031,6 +1033,13 @@ func (h *Handler) TransactionStatus(hash felt.Felt) (*TransactionStatus, *jsonrp
 	case ErrTxnHashNotFound:
 		txStatus, err := h.feederClient.Transaction(context.Background(), &hash)
 		if err != nil {
+			// Check if the error is due to a transaction not being found
+			notFoundErr := fmt.Sprintf("%d %s", http.StatusNotFound, http.StatusText(http.StatusNotFound))
```
joshklop, line 1037: I get an `http.StatusOK` when requesting a tx that doesn't exist, so I don't think this will work. 

I get the following response on testnet ([example](https://alpha4.starknet.io/feeder_gateway/get_transaction?transactionHash=0x2926a3b3c8f1cf08b1597e153928fc49bfaa2c521952610c108808fc21fe285)), which is on v0.12.1:

```
{"finality_status": "NOT_RECEIVED", "status": "NOT_RECEIVED"}
```

and this response on [mainnet](https://alpha-mainnet.starknet.io/feeder_gateway/get_transaction?transactionHash=0x2926a3b3c8f1cf08b1597e153928fc49bfaa2c521952610c108808fc21fe285):

```
{"status": "NOT_RECEIVED"}
```
I think we just need to add this `NOT_RECEIVED` enum value to `feeder.TxnExecutionStatus` (not sure if that's the exact name).

joshklop: Let's add tests for these cases as well. See `handlers_test.go`.

joshklop (changes requested):

aminsato:
@joshklop
I made changes please review them

jsonrpc/server.go
joshklop: Can we remove this change since it is unrelated?

rpc/transaction.go
```diff
@@ -88,6 +88,7 @@ type TxnFinalityStatus uint8
 const (
 	TxnAcceptedOnL1 TxnFinalityStatus = iota + 1
 	TxnAcceptedOnL2
+	TxNotRECEIVED
```
joshklop, line 91: We shouldn't add this here. It isn't part of the [RPC spec](https://github.com/starkware-libs/starknet-specs/blob/6fadca1f1abf48523c0aa799a10061f40f973468/api/starknet_api_openrpc.json#L2532-L2540).

rpc/handlers.go
```diff
@@ -1031,9 +1031,15 @@ func (h *Handler) TransactionStatus(ctx context.Context, hash felt.Felt) (*Trans
 	case ErrTxnHashNotFound:
 		txStatus, err := h.feederClient.Transaction(ctx, &hash)
 		if err != nil {
+			// Handle other internal errors
```
joshklop, line 1034: Let's remove this comment. I don't think it's relevant any more.

joshklop (changes requested):
Looks good overall. We'll need tests for the new cases in order to merge this PR. Can you add tests?

aminsato:
Hi
@joshklop
I removed unnecessary changes and added tests

rpc/handlers.go
aminsato: done

rpc/transaction.go
aminsato: removed

jsonrpc/server.go
aminsato: done

rpc/handlers_test.go
```diff
@@ -1843,16 +1843,19 @@ func TestTransactionStatus(t *testing.T) {
 		network           utils.Network
 		verifiedTxHash    *felt.Felt
 		nonVerifiedTxHash *felt.Felt
+		notFoundTxHash    *felt.Felt
```
joshklop, line 1846: nit: can we move the "not found" case to a separate sub-test? That should reduce the complexity of the test.

joshklop (commented):
Looking good. I think we're almost there.

rpc/handlers_test.go
aminsato: @joshklop 
I Have moved the "not found" case to a separate sub-test

rpc/handlers_test.go
```diff
@@ -1940,6 +1940,58 @@ func TestTransactionStatus(t *testing.T) {
 	}
 }
 
+func TestTransactionNotFoundStatus(t *testing.T) {
```
joshklop, line 1943: Let's make this a sub-test on the top-level `TestTransactionStatus` test (using `t.Run`).

rpc/handlers_test.go
```diff
@@ -1940,6 +1940,58 @@ func TestTransactionStatus(t *testing.T) {
...
+			t.Run("transaction not found", func(t *testing.T) {
+				notFoundTests := map[string]struct {
+					finality      rpc.TxnFinalityStatus
+					hash          *felt.Felt
+					expectedError *jsonrpc.Error
+				}{
+					"not found": {
+						hash: test.notFoundTxHash,
+					},
+				}
```
joshklop, line 1975: There appears to be some unnecessary code here and below. I think we can have just one `for-range` through `tests`, no need for nested loops.

joshklop (changes requested):

rpc/handlers_test.go
aminsato: done

rpc/handlers_test.go
aminsato: done

joshklop (approved):

aminsato:
@joshklop
Hi
Please merge my PR 😢 .

aminsato:
@joshklop
Hi
Please merge my PR 😢 .

aminsato:
@joshklop
Hi
Please merge my PR 😢 .

joshklop:
Hi @aminsato, thank you for your patience. Mind rebasing on `main`? Then we can merge.

aminsato:
> Hi @aminsato, thank you for your patience. Mind rebasing on `main`? Then we can merge.

I rebased on 'main'

Please check

Thanks

joshklop:
I get conflicts when I run `git rebase main`. Another option is to create a separate branch and put the changes in a single commit there.

joshklop:
Closing in favor of #1133

Issue #992, opened 2023-08-03
Create `snapshot` cobra subcommand


IronGauntlets:
The `snapshot` subcommand should take the location of the download and network as an argument and ideally show a progress bar. The subcommand should look something like the following (I am not sure whether cobra subcommands are prefixed with `--` or not):

`./build/juno snapshot --location "path/to/dir" --network "mainnet/goerli/goerli2/integration"`

 As an initial step and simplicity, the `snapshot` subcommand should **not** run Juno after the download is complete. The user should run Juno as normal by providing the location of the downloaded snapshot through `--db-path`.

IronGauntlets:
@JorikSchellekens or @eliotstock 

If the user provides a non-empty db path and the `--snapshot` flag is also provided, what should Juno do? Should it return an error that both flags cannot be set together, or should we just ignore the `--snapshot` flag?

eliotstock:
I'd have thought the user would always want to provide `--db-path` (or be happy with the default) because they care about disk space and want to ensure the volume/partition `juno` is using has room. Personally I'd want to provide both `--db-path` and `--snapshot` and would expect that the snapshot is downloaded and extracted into the `--db-path` then sync starts to add blocks after the last block in the snapshot.

Brivan-26:
For the `--snapshot` value, does the user provide the network name and we map each network to its snapshot, or he provides the snapshot download URL?

eliotstock:
Wait, did we intend `--snapshot` to take a value? I was thinking of it as a boolean defaulting to true which meant "go and grab the snapshot for this chain and extract it before you start syncing". The point of this, to me, is to save the user time and effort so they should not have to know where the snapshot is - Juno should know based on the network, which is already provided separately on the command line (or config file).

I know we provide two snapshots for each network at different heights right now. I'm suggesting we just use the latest one with this flag. If you want the earlier one, it's up to you to grab it manually and extract it.

omerfirmak:
Similarly to,  #993 . I believe there was an agreement on not implementing this. Feel free to reopen if needed.

PR #1067, opened 2023-08-10
Added snapshot command

Brivan-26:
Closes [snapshot Cobra sub command issue](https://github.com/NethermindEth/juno/issues/992)


cmd/juno/juno.go
```diff
@@ -106,6 +106,36 @@ func main() {
 		return nil
 	})
 
+	var location, network string
+
+	snapshotCommand := &cobra.Command{
```
joshklop, line 111: We should probably put this in a separate file `cmd/juno/snapshot.go`. @kirugan probably knows more about Cobra idioms, but that's what I've seen in other projects.

utils/downloads.go
```diff
@@ -0,0 +1,76 @@
...
+	"goerli":  "https://juno-snapshot.s3.us-east-2.amazonaws.com/goerli/juno_goerli_v0.5.0_839969.tar",
+	"goerli2": "https://juno-snapshot.s3.us-east-2.amazonaws.com/goerli2/juno_goerli2_v0.5.0_135973.tar",
+}
+
+const (
+	downloadBufferSize = 8192 // 8KB
+	dirPermissionMod   = 0o770
+)
+
+func DownloadFile(ctx context.Context, network, location string) error {
```
joshklop, line 25: Is there a way we can test this? Maybe using `net/http/httptest`? Would be really nice to have tests for such a large function.

Brivan-26: I thought about that, but I'm not sure if we need to unit test a download file ?

joshklop: Better to unit test it imo. Better than coming back and debugging it later. Also ensures we don't break things if we modify the function in the future.

cmd/juno/juno.go
joshklop: Moving it outside the `main` function also allows us to test it.

utils/downloads.go
joshklop: Thinking out loud: I feel like this is a pretty common operation. Maybe there's a widely used and well-tested package that does something similar? I feel like that would be ideal.

Brivan-26: I did some research about that and I found [this](https://github.com/cavaliergopher/grab). Tho, I thought the download process is a straightforward task that can be accomplished using the standard library(`http.Get()` && `io.Copy` && `out.Write`).

Brivan-26: For the unit test, I agree, I will add a unit test for it.

joshklop: That's fair. Let's go with that.

cmd/juno/juno.go
Brivan-26: I don't see what can be tested since the presence of flag values is already handled by Viper and the required flags are already handled by Cobra. Do you suggest something else to test ?

joshklop: Ah yes I think you're right. Ideally we move as much stuff out of Cobra/Viper's control and rigorously test everything we do control. Apologies for the false-positive review comments, trying to review quickly xD

utils/downloads.go
```diff
@@ -0,0 +1,91 @@
...
+	dirPermissionMod   = 0o770
+)
+
+type Downloader interface {
+	DownloadAndWrite(ctx context.Context, reader io.Reader, writer io.Writer, bar *pb.ProgressBar) error
+}
+
+type RealDownloader struct{}
+
+func (d *RealDownloader) DownloadAndWrite(ctx context.Context, reader io.Reader, writer io.Writer, bar *pb.ProgressBar) error {
```
joshklop, line 31: Why do we need `bar`?

utils/downloads.go
```diff
@@ -0,0 +1,91 @@
...
+	"goerli":  "https://juno-snapshot.s3.us-east-2.amazonaws.com/goerli/juno_goerli_v0.5.0_839969.tar",
+	"goerli2": "https://juno-snapshot.s3.us-east-2.amazonaws.com/goerli2/juno_goerli2_v0.5.0_135973.tar",
+}
+
+const (
+	downloadBufferSize = 8192 // 8KB
+	dirPermissionMod   = 0o770
+)
+
+type Downloader interface {
```
joshklop, line 25: I haven't delved in 100% yet so I could be wrong, but I don't think an interface is necessary here.

I think we can have a signature like this
```go
func(context.Context, url, dir string) error
```
To test, we use `httptest`+`testing.T.TempDir`. This gives us an end-to-end black-box test that runs quickly.

joshklop (changes requested):

utils/downloads.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+type RealDownloader struct{}
+
+func (d *RealDownloader) DownloadAndWrite(ctx context.Context, reader io.Reader, writer io.Writer, bar *pb.ProgressBar) error {
+	buffer := make([]byte, downloadBufferSize)
+	for {
+		select {
+		case <-ctx.Done():
+			return fmt.Errorf("canceled download")
```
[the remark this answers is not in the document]
Brivan-26: Isn't anything that prevents from completing the download is considered an error, or if it comes from the caller it is not? Besides that, the process may be ended by someone else, maybe the OS.

utils/downloads.go
```diff
@@ -0,0 +1,91 @@
...
+		}
+	}
+}
+
+func DownloadFile(ctx context.Context, network, location string, downloader Downloader) error {
+	_, ok := snapshots[network]
+	if !ok {
+		return fmt.Errorf("the %s network is not supported", network)
+	}
+	resp, err := http.Get(snapshots[network]) //nolint
```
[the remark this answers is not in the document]
Brivan-26: Good point, didn't pay attention to it. It's `http.NewRequestWithContext`

utils/downloads.go
```diff
@@ -0,0 +1,91 @@
...
+	"fmt"
+	"io"
+	"net/http"
+	"os"
+	"path/filepath"
+
+	"gopkg.in/cheggaaa/pb.v1"
+)
+
+var snapshots = map[string]string{
```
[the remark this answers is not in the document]
Brivan-26: I guess for a better user experience, the end-user should only provide the network name without worrying to get the snapshot download, if he already bothers him self getting the download link, then I see this command as useless.
That's how I interpreted [the issue](https://github.com/NethermindEth/juno/issues/992)

utils/downloads.go
Brivan-26: For printing the download progress.

utils/downloads.go
Brivan-26: I'm not sure how can we mock different download scenarios without the interface.

utils/downloads.go
joshklop: I see `bar.Finish()`, do we use it elsewhere?

utils/downloads.go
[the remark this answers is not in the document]
joshklop: > if it comes from the caller it is not

Exactly. We don't need to tell the caller that they canceled the context--the caller already knows that. It makes the `if err != nil` idiom hard to use, since the caller always has to check if we're just telling it that we "canceled download".

> the process may be ended by someone else, maybe the OS

That wouldn't cancel the context, those would be fatal errors that would manifest some other way.

Issue #933, opened 2023-07-20
Allow for less than two return values in RPC handlers


joshklop:
Right now RPC handlers must have two return values, and the second return value must be of type `*jsonrpc.Error`. It would convenient if the following three cases were supported:

- zero return values
- one return value (response or error)
- two return values, where the second must be an error <- we only support this right now

This is very similar to geth's RPC server [semantics](https://github.com/ethereum/go-ethereum/blob/7f756dc1185d7f1eeeacb1d12341606b7135f9ea/rpc/doc.go#L29-L30).

IronGauntlets:
I am not sure how useful this would be since returning multiple values with the last value being an error is a go idiom.

Can you please explain which case would benefit from the 2 unimplemented options?

joshklop:
Definitely low-priority.

> Can you please explain which case would benefit from the 2 unimplemented options?

One return value: some simple handlers like [juno_version](https://github.com/NethermindEth/juno/blob/8656316426923c131187873a8bf8df048e5864c8/rpc/handlers.go#L980) never return an error.

Zero return values: Probably useless. I forget what I was thinking here.

joshklop:
Not needed. See linked PR above.

PR #922, opened 2023-07-18
Implement `juno_subscribe_newHeads`

joshklop:
Almost there. Remaining tasks:

- add more tests
- manually test (especially helpful to determine if websocket connection closes under weird circumstances)
- document functions and types
- cleanup git history to help with reviews

PR #1068, opened 2023-08-11
jsonrpc: allow up to 2 returns in handlers

Exca-DK:
This pr implements #933


Exca-DK:
Will rebase against #955 and #922

PR #1075, opened 2023-08-14
jsonrpc,clients/gateway,clients/feeder: fix superfluous headers

Exca-DK:
This pr fixes superfluous responses due to doubleheader writes.  
The issue comes from the fact that Write() method already writes the header. 
Write description:
> ... If WriteHeader has not yet been called, Write calls WriteHeader(http.StatusOK) before writing the data. ...


omerfirmak (approved):
Nice catch, thanks!

Issue #1065, opened 2023-08-10
Allow nil `params` when all parameters of a handler is optional


omerfirmak:
more context https://github.com/NethermindEth/juno/pull/950#discussion_r1289777996

wojciechos:
unfortunately couldn't find that scenario to test it from user perspective

PR #1078, opened 2023-08-14
Allow passing empty/nil params in case all are optional

kirugan:
Issue #1065

### DIFF
--- a/jsonrpc/server.go
+++ b/jsonrpc/server.go
@@ -361,6 +361,7 @@ func (s *Server) handleRequest(ctx context.Context, req *request) (*response, er
 	return res, nil
 }
 
+//nolint:gocyclo
 func (s *Server) buildArguments(ctx context.Context, params any, method Method) ([]reflect.Value, error) {
 	allParamsAreOptional := utils.All(method.Params, func(p Parameter) bool {
 		return p.Optional
@@ -435,7 +436,7 @@ func (s *Server) buildArguments(ctx context.Context, params any, method Method) 
 	return args, nil
 }
 
-func (s *Server) buildDefaultArguments(ctx context.Context, method Method) ([]reflect.Value, error) {
+func (s *Server) buildDefaultArguments(_ context.Context, method Method) ([]reflect.Value, error) {
 	handlerType := reflect.TypeOf(method.Handler)
 
 	numArgs := handlerType.NumIn()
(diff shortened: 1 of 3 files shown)


jsonrpc/server.go
```diff
@@ -399,6 +408,10 @@ func (s *Server) buildArguments(ctx context.Context, params any, method Method)
 	case reflect.Map:
 		paramsMap := params.(map[string]any)
 
+		if len(paramsMap) == 0 && allParamsAreOptional {
```
omerfirmak, line 411: I dont think this is needed, the loop below will construct a list with the default argument anyways.

jsonrpc/server.go
```diff
@@ -385,6 +390,10 @@ func (s *Server) buildArguments(ctx context.Context, params any, method Method)
 	case reflect.Slice:
 		paramsList := params.([]any)
 
+		if len(paramsList) == 0 && allParamsAreOptional {
+			return s.buildDefaultArguments(ctx, method)
+		}
+
```
omerfirmak, line 396: same, I dont think this is needed either. When using list of arguments (instead of a map) optional arguments are not supported in the spec I believe.

jsonrpc/server.go
kirugan: Yep, you're right - passing {} works the same with/without this changes. Will remove

jsonrpc/server.go
kirugan: Without this changes I get this error:
```
{
    "jsonrpc": "2.0",
    "error": {
        "code": -32602,
        "message": "Invalid Params",
        "data": "missing/unexpected params in list"
    },
    "id": 1
}
```

omerfirmak: Can you share the request and the handler definition?

kirugan: request is:
```
{
    "jsonrpc": "2.0",
    "id": 1,
    "method": "kirill_testing",
    "params": []
}
```
registration of this handler:
```
{
			Name:    "kirill_testing",
			Params:  []jsonrpc.Parameter{{Name: "v", Optional: true}},
			Handler: h.Testing,
		},
```
definition: 
```
func (h *Handler) Testing(v string) (string, *jsonrpc.Error) {
	return v, nil
}
```
I made a testing branch `kirugan/allow-nil-params-testing-handler` where this handler resides

omerfirmak: When passing the parameters as a list, parsing with optional parameters is not really straight forward. There is too much ambiguity. We can let it fail IMO.

kirugan: Ok, removed this check

omerfirmak (approved):

PR #1081, opened 2023-08-15
Allow streaming of proto messages over a single network.Stream

omerfirmak:


p2p/starknet/stream.go
```diff
@@ -0,0 +1,16 @@
+package starknet
+
+// Stream represents a series of messages that can be accessed by invoking Stream a number of times. With the last element in the stream
```
joshklop, line 3: Mind finishing off this comment? I'm genuinely interested in knowing what happens at the end :)

joshklop, line 3: Mind finishing off this comment? I'm genuinely interested in knowing what happens at the end :)

joshklop, line 3: Mind finishing off this comment? I'm genuinely interested in knowing what happens at the end :)

joshklop, line 3: Mind finishing off this comment? I'm genuinely interested in knowing what happens at the end :)

joshklop: Is this the iterator implementation you were looking for btw? It's cool how this avoids type assertions. All others that I've seen use an interface with methods like `Next() any` and `Valid() bool`.

p2p/starknet/bytereader.go
```diff
@@ -0,0 +1,21 @@
...
+	"google.golang.org/protobuf/encoding/protodelim"
+)
+
+var _ protodelim.Reader = (*byteReader)(nil)
+
+type byteReader struct {
+	io.Reader
+}
+
+func (r *byteReader) ReadByte() (byte, error) {
```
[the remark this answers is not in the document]
omerfirmak: libp2p gives us a `network.Stream`, which only implements `io.Reader`. But `protodelim.Reader` also requires `io.ByteReader`.

So we need to extend `network.Stream` to support `io.ByteReader` as well.

p2p/starknet/bytereader.go
```diff
@@ -0,0 +1,21 @@
...
+	var b [1]byte
+	if _, err := r.Read(b[:]); err != nil {
```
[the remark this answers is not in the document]
omerfirmak: > I assume this will block until a byte is received, rather than waiting for a complete message. Is that correct?

That is my assumption as well, but I dont think it matters. We are actually interested in the message anyways, we dont really need to know what the first byte is before the entire message arrives.

> Why is this "streaming" method better than the previous non-streaming method?

There is a soft limit of 1MB for protobuf messages. 

``` proto
message List {
 repeated Elems
}
```

When we use `repeated` it essentially allows arbitrarily large messages to be created. Instead of `repeated`, we stream individual `Elems` one by one. Assuming that Elems size is constant, we for sure know that each message is below 1MB in size.

> Is there a name for the "non-streaming" method?

I dont think so

p2p/starknet/stream.go
omerfirmak: > Mind finishing off this comment? I'm genuinely interested in knowing what happens at the end :)

Haha, I must have gotten distracted for some reason :smile: It should continue like, "After stream is entirely consumed of elements, it should return false as its second return value"

> Is this the iterator implementation you were looking for btw? It's cool how this avoids type assertions. All others that I've seen use an interface with methods like Next() any and Valid() bool.

Yeah, I was trying to find a more capable implementation actually. But this will do. Stole it from https://github.com/golang/go/discussions/54245#discussion-4277068, I can not take any credit for it :smile:

p2p/starknet/bytereader.go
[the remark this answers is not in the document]
joshklop: Can we use `bufio.NewReader(stream)`? That will give us an `io.ByteReader` without needing to hand-roll it.

p2p/starknet/stream.go
joshklop: Oh that's an awesome resource. Thanks!

p2p/starknet/bytereader.go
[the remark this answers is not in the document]
omerfirmak: `bufio.NewReader` allocates 4K on the heap. If we pool them, that might be an option as well.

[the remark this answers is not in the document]
omerfirmak: Had a look into libp2p, I think `network.Stream` has some internal buffering as well. So let's not introduce another level of buffering right now.

joshklop (approved):

PR #1087, opened 2023-08-15
Break HTTP routes into multiple servers

joshklop:
For access control purposes.


joshklop:
This is probably easiest to review all at once, rather than commit-by-commit. We can squash merge.

omerfirmak:
We can put Http & ws rpc on the same port, WDYT?

joshklop:
Rebased and made the adjustments mentioned in my last message. If you don't mind taking another quick look @omerfirmak, I would really appreciate it. 

Note that all services except synchronization are disabled by default now. Enabling them all with default ports would look something like:

```
juno --http --ws --metrics --pprof --grpc
```

omerfirmak (approved):

node/http.go
```diff
@@ -16,94 +15,115 @@ import (
...
+		srv: &http.Server{
+			Addr:    listener.Addr().String(),
+			Handler: mux,
+			// ReadTimeout also sets ReadHeaderTimeout and IdleTimeout.
+			ReadTimeout: 30 * time.Second,
+		},
+		listener: listener,
+	}
+}
+
```
omerfirmak, line 126: I dont think we will ever use these in tests, right?
So can we create the listener in these functions so `node.New` has less copy-pasta?

omerfirmak: see https://github.com/NethermindEth/juno/pull/1087/commits/7fff1473398783135ccbe821edd6fc9a2b3c5904
Feel free to revert

joshklop: LGTM. Thanks

PR #1102, opened 2023-08-17
Split L1 from Ethereum

joshklop:
See commit message. The interesting files are `ethereum/ethereum.go` and `l1/l1.go`. `ethereum.go` is a refactored version of `l1.go`, basically. All the other changes result from that refactor.

Ideally we remove the `l1` package entirely in the near future.


joshklop:
Low-priority. May come back to something like this in the future.

Issue #1091, opened 2023-08-16
Add new feeder endpoints to feeder client


omerfirmak:
[get_signature](https://external.integration.starknet.io/feeder_gateway/get_signature?blockNumber=11111), 
[get_public_key](https://external.integration.starknet.io/feeder_gateway/get_public_key), 
[get_state_update_with_block](https://external.integration.starknet.io/feeder_gateway/get_state_update?blockNumber=pending&includeBlock=true)

PR #1110, opened 2023-08-17
New feeder endpoints

Brivan-26:
Closes [Add new feeder endpoints to feeder client](https://github.com/NethermindEth/juno/issues/1091)


clients/feeder/feeder.go
```diff
@@ -127,15 +130,28 @@ func newTestServer(network utils.Network) *httptest.Server {
 		case strings.HasSuffix(r.URL.Path, "get_compiled_class_by_class_hash"):
 			dir = "compiled_class"
 			queryArg = "classHash"
+		case strings.HasSuffix(r.URL.Path, "get_public_key"):
+			dir = "public_key"
+			queryArg = ""
```
omerfirmak, line 135:
```suggestion
			dir = "public_key"
			queryArg = "pk"
                        queryMap[queryArg] = queryArg
```
instead of special casing at L144, WDYT?

Brivan-26: I wanted to keep it consistent with the real endpoint, but this is better to save special case handling at L144.

clients/feeder/feeder.go
```diff
@@ -110,14 +109,18 @@ func newTestServer(network utils.Network) *httptest.Server {
...
 		switch {
 		case strings.HasSuffix(r.URL.Path, "get_block"):
 			dir = "block"
-			queryArg = "blockNumber"
+			queryArg = blockNumberArg
 		case strings.HasSuffix(r.URL.Path, "get_state_update"):
-			dir = "state_update"
-			queryArg = "blockNumber"
+			queryArg = blockNumberArg
+			if _, ok := queryMap["includeBlock"]; !ok {
```
omerfirmak, line 119:
```suggestion
			if includeBlock, ok := queryMap["includeBlock"]; !ok || !includeBlock {
```

omerfirmak (approved):

Issue #1074, opened 2023-08-12
Implement HandleGetTransactions


omerfirmak:
https://github.com/NethermindEth/juno/blob/main/p2p/starknet/handlers.go#L135

kirugan:
This issue was already implemented

PR #1113, opened 2023-08-17
P2P GetTransactions

kirugan:
Issue #1074

### DIFF
--- a/adapters/core2p2p/transaction.go
+++ b/adapters/core2p2p/transaction.go
@@ -7,7 +7,7 @@ import (
 	"github.com/NethermindEth/juno/p2p/starknet/spec"
 )
 
-func AdaptTransaction(transaction core.Transaction) *spec.Transaction { //nolint: gocyclo
+func AdaptTransaction(transaction core.Transaction) *spec.Transaction {
 	if transaction == nil {
 		return nil
 	}
@@ -19,8 +19,6 @@ func AdaptTransaction(transaction core.Transaction) *spec.Transaction { //nolint
 		specTx.Txn = adaptDeployTransaction(tx)
 	case *core.DeployAccountTransaction:
 		switch {
-		case tx.Version == nil:
-			panic("DeployAccount transaction has not set version")
 		case tx.Version.Is(1):
 			specTx.Txn = &spec.Transaction_DeployAccountV1_{
 				DeployAccountV1: &spec.Transaction_DeployAccountV1{
@@ -37,8 +35,6 @@ func AdaptTransaction(transaction core.Transaction) *spec.Transaction { //nolint
 		}
 	case *core.DeclareTransaction:
 		switch {
-		case tx.Version == nil:
-			panic("Declare transaction has not set version field")
 		case tx.Version.Is(0):
 			specTx.Txn = &spec.Transaction_DeclareV0_{
 				DeclareV0: &spec.Transaction_DeclareV0{
@@ -74,8 +70,6 @@ func AdaptTransaction(transaction core.Transaction) *spec.Transaction { //nolint
 		}
 	case *core.InvokeTransaction:
 		switch {
-		case tx.Version == nil:
-			panic("Invoke transaction has not set version field")
 		case tx.Version.Is(0):
 			specTx.Txn = &spec.Transaction_InvokeV0_{
 				InvokeV0: &spec.Transaction_InvokeV0{


p2p/starknet/handlers.go
```diff
@@ -101,75 +101,118 @@ func (h *Handler) reqHandler(req *spec.Request) (Stream[proto.Message], error) {
...
+
+		return &spec.TransactionsResponse{
+			BlockNumber: block.Number,
+			BlockHash:   core2p2p.AdaptHash(block.Hash),
+			Responses: &spec.TransactionsResponse_Transactions{
+				Transactions: &spec.Transactions{
+					Items: utils.Map(block.Transactions, core2p2p.AdaptTransaction),
+				},
+			},
+		}, false
```
joshklop, line 202: Iterator implementation looks good to me overall.

Should we return `true` here, and return `false` only when `it.Err() != nil` or `!it.Valid()`?

p2p/starknet/iterator.go
```diff
@@ -0,0 +1,66 @@
...
+	return true
+}
+
+func (it *iterator) Next() bool {
+	if it.forward {
+		it.blockNumber += it.step
+	} else {
+		it.blockNumber -= it.step
+	}
+	it.limit--
```
joshklop, line 56: Can `limit` underflow?

p2p/starknet/iterator.go
```diff
@@ -0,0 +1,66 @@
...
+
+	return it.Valid()
+}
+
+func (it *iterator) Block() *core.Block {
```
joshklop, line 61: Would it be better to return the error directly, rather than requiring the caller to query `it.Err()`?

p2p/starknet/iterator.go
kirugan: nice catch, thanks!

p2p/starknet/iterator.go
kirugan: Yeah, I was thinking about this for a while. For some time it was a common pattern for iterators to have `Next()`, `HasNext()`, and `Err` (see https://pkg.go.dev/bufio#Scanner for example). I'm not sure about it now, will reconsider it

adapters/core2p2p/transaction.go
```diff
@@ -0,0 +1,185 @@
...
+		case tx.Version.Is(1):
+			specTx.Txn = &spec.Transaction_InvokeV1_{
+				InvokeV1: &spec.Transaction_InvokeV1{
+					Sender:    AdaptFeltToAddress(tx.SenderAddress),
+					MaxFee:    AdaptFelt(tx.MaxFee),
+					Signature: AdaptSignature(tx.Signature()),
+					Calldata:  AdaptFeltSlice(tx.CallData),
+				},
+			}
+		case tx.Version.Is(3):
```
omerfirmak, line 132: Let's not do version 3s for now. They will not be added until 0.13.0

adapters/core2p2p/transaction.go
IronGauntlets, line 1: I think it would be useful to have some tests for this. As in the `starknet_test.go` we only check that we have received the correct number of transactions

IronGauntlets (changes requested):

p2p/starknet/starknet_test.go
```diff
@@ -179,15 +180,138 @@ func TestClientHandler(t *testing.T) {
 	})
 
 	t.Run("get txns", func(t *testing.T) {
-		res, cErr := client.RequestTransactions(testCtx, &spec.TransactionsRequest{})
+		blocks := []*core.Block{
```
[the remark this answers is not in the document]
omerfirmak: If `core2p2p.AdaptTransaction` is covered %100, it should be fine

adapters/core2p2p/transaction.go
```diff
@@ -0,0 +1,165 @@
...
+				DeployAccountV1: &spec.Transaction_DeployAccountV1{
+					MaxFee:      AdaptFelt(tx.MaxFee),
+					Signature:   AdaptAccountSignature(tx.Signature()),
+					ClassHash:   AdaptHash(tx.ClassHash),
+					Nonce:       AdaptFelt(tx.Nonce),
+					AddressSalt: AdaptFelt(tx.ContractAddressSalt),
+					Calldata:    AdaptFeltSlice(tx.ConstructorCallData),
+				},
+			}
+		case tx.Version.Is(3):
```
omerfirmak, line 35: We can omit V3 txns for now I believe. But feel free to ignore that suggestion.

adapters/core2p2p/transaction.go
```diff
@@ -0,0 +1,165 @@
...
+		Deploy: &spec.Transaction_Deploy{
+			ClassHash:   AdaptHash(tx.ClassHash),
+			AddressSalt: AdaptFelt(tx.ContractAddressSalt),
+			Calldata:    AdaptFeltSlice(tx.ConstructorCallData),
+		},
+	}
+}
+
+func adaptL1HandlerTransaction(tx *core.L1HandlerTransaction) *spec.Transaction_L1Handler {
+	if tx.Version != nil && !tx.Version.Is(1) {
```
omerfirmak, line 153: shouldnt this be;

```suggestion
	if tx.Version == nil || !tx.Version.Is(1) {
```

omerfirmak: We can assume it is non-nil btw

IronGauntlets: Yeah because all of the transactions from the feeder have versions

p2p/starknet/starknet_test.go
```diff
@@ -270,13 +350,103 @@ func TestClientHandler(t *testing.T) {
...
+	var f felt.Felt
+	f.SetUint64(v)
 
-	return sl
+	txV := core.TransactionVersion(f)
+	return &txV
+}
+
+func feltSlice(n int) []*felt.Felt {
+	return make([]*felt.Felt, n)
```
IronGauntlets, line 449: nit: consider populating the slice with rand felts

p2p/starknet/starknet_test.go
```diff
@@ -287,3 +457,52 @@ func randFelt(t *testing.T) *felt.Felt {
 
 	return f
 }
+
+func fillFelts[T any](t *testing.T, i T) T {
```
IronGauntlets, line 461: I think this should be moved to `utils` package along with `randFelt` and `feltSlice` as they can be useful in other tests. However, we can do it in another pr.

p2p/starknet/starknet_test.go
```diff
@@ -179,15 +174,100 @@ func TestClientHandler(t *testing.T) {
...
+				Step:      1,
+			},
+		})
 		require.NoError(t, cErr)
 
-		count := uint64(0)
+		var count uint64
 		for txn, valid := res(); valid; txn, valid = res() {
+			if count == numOfBlocks {
+				assert.NotNil(t, txn.GetFin())
```
IronGauntlets, line 260: I think we need to add a testcase to make sure when multiple blocks's transactions are requested then the fin has been sent correctly. However, that can be done in your other pr.

IronGauntlets (approved):
Left some suggestions for future PR. Otherwise, everything looks good.

adapters/core2p2p/transaction.go
```diff
@@ -0,0 +1,131 @@
...
+					Calldata:    AdaptFeltSlice(tx.ConstructorCallData),
+				},
+			}
+		default:
+			panic(fmt.Errorf("unsupported InvokeTransaction version %s", tx.Version))
+		}
+	case *core.DeclareTransaction:
+		switch {
+		case tx.Version == nil:
+			panic("Declare transaction has not set version field")
```
IronGauntlets, line 41: These should be removed as well.

PR #1120, opened 2023-08-18
rpc: Always return sequencer address in BlockHeader

### LINKED ISSUE #1117: Make sure we always return a sequencer address in block header
RPC spec makes the sequencer address required.
Let's just return 0 if core.Header.SequencerAddress is `nil`

Exca-DK:
closes #1117


rpc/handlers.go
```diff
@@ -174,13 +174,18 @@ func adaptBlockHeader(header *core.Header) BlockHeader {
 		blockNumber = &header.Number
 	}
 
+	sequencerAddress := header.SequencerAddress
+	if sequencerAddress == nil {
+		sequencerAddress = new(felt.Felt)
```
omerfirmak, line 179:
```suggestion
		sequencerAddress = &felt.Zero
```

omerfirmak (commented):
Thanks! Do you mind adding a test as well?

Exca-DK:
> Thanks! Do you mind adding a test as well?

What kind of test? Checking returned header or comparing JSON dump?

omerfirmak (approved):
Perfect, thank you!

PR #1121, opened 2023-08-18
rpc: Do not return ContractAddress for Deploy and DeployAccount transactions

### LINKED ISSUE #1118: Do not return ContractAddress for Deploy and DeployAccount transactions
Even tho feeder gives that information, contract_address is not in the RPC spec.

Exca-DK:
This pr closes #1118. 
From What I understood the target was only for transactions related stuff, Receipts still should include the address.

### DIFF
--- a/rpc/handlers.go
+++ b/rpc/handlers.go
@@ -408,21 +408,13 @@ func (h *Handler) TransactionByBlockIDAndIndex(id BlockID, txIndex int) (*Transa
 // It follows the specification defined here:
 // https://github.com/starkware-libs/starknet-specs/blob/master/api/starknet_api_openrpc.json#L222
 func (h *Handler) TransactionReceiptByHash(hash felt.Felt) (*TransactionReceipt, *jsonrpc.Error) {
-	coreTxn, err := h.bcReader.TransactionByHash(&hash)
+	txn, err := h.bcReader.TransactionByHash(&hash)
 	if err != nil {
 		return nil, ErrTxnHashNotFound
 	}
-	txn := adaptTransaction(coreTxn)
 	receipt, blockHash, blockNumber, err := h.bcReader.Receipt(&hash)
 	if err != nil {
 		return nil, ErrTxnHashNotFound
-	}
-
-	switch v := coreTxn.(type) {
-	case *core.DeployTransaction:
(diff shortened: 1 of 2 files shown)


omerfirmak:
I think we can do this more cleanly.

One way to do this is to remove `ContractAddress` from `rpc.Transaction` entirely.
and use `h.bcReader.TransactionByHash(&hash)` in `TransactionReceiptByHash` instead of  `h.TransactionByHash`. Since it needs the ContractAddress information from `core.Transaction`

omerfirmak:
> > I think we can do this more cleanly.
> 
> > 
> 
> > One way to do this is to remove `ContractAddress` from `rpc.Transaction` entirely. and use `h.bcReader.TransactionByHash(&hash)` in `TransactionReceiptByHash` instead of `h.TransactionByHash`. Since it needs the ContractAddress information from `core.Transaction`
> 
> 
> 
> Removing the contract address entirely would affect `INVOKE_TXN_V0` and `L1_HANDLER_TXN` . They include `FUNCTION_CALL` which requires `contract_address`

Ah, I see. Then just dont copy the ContractAddress to rpc.Transaction for Deploy and DeployAccount transactions

rpc/handlers.go
```diff
@@ -405,15 +403,23 @@ func (h *Handler) TransactionByBlockIDAndIndex(id BlockID, txIndex int) (*Transa
...
 		return nil, ErrTxnHashNotFound
 	}
 
+	switch v := coreTxn.(type) {
+	case *core.DeployTransaction:
+		txn.ContractAddress = v.ContractAddress
+	case *core.DeployAccountTransaction:
+		txn.ContractAddress = v.ContractAddress
+	}
+
```
omerfirmak, line 422:
```suggestion
	var contractAddress *felt.Felt
	switch v := coreTxn.(type) {
	case *core.DeployTransaction:
		contractAddress = v.ContractAddress
	case *core.DeployAccountTransaction:
		contractAddress = v.ContractAddress
	}

```

omerfirmak: and we can remove

```
	contractAddress := txn.ContractAddress
	if txn.Type != TxnDeploy && txn.Type != TxnDeployAccount {
		contractAddress = nil
	}
```

PR #1125, opened 2023-08-18
Explicitly calculate actual fee when using skip_fee_charge flag

omerfirmak:
Even tho skip fee charge is enabled, blockifier respects the max_fee field of the transaction. By recalculating the actual fee when it is used we can get an accurate fee estimation.

### DIFF
--- a/vm/rust/src/lib.rs
+++ b/vm/rust/src/lib.rs
@@ -174,6 +174,7 @@ pub extern "C" fn cairoVMExecute(
         felt_to_u128(gas_price_felt),
     );
     let mut state = CachedState::new(reader);
+    let charge_fee = skip_charge_fee == 0;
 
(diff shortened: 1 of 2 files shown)


vm/rust/src/lib.rs
```diff
@@ -244,7 +235,12 @@ pub extern "C" fn cairoVMExecute(
                 );
                 return;
             }
-            Ok(t) => unsafe {
+            Ok(mut t) => unsafe {
+                // we are estimating fee, override actual fee calculation
+                if !charge_fee {
```
joshklop, line 240: Can we move this outside the `unsafe` block?

joshklop (approved):

vm/rust/src/lib.rs
omerfirmak: done

PR #1126, opened 2023-08-18
rpc: ignore nonce for declare_v0

### LINKED ISSUE #1123: Do not return nonce for Declare txn v0
For declare txn v0 we return nonce which is not included in starknet spec.

**Steps to reproduce:**
Request on goerli:
```{
    "id": 1,
    "jsonrpc": "2.0",
    "method": "starknet_getTransactionByHash",
    "params": [
        "0x21933cb48e59c74caa4575a78e89e6046d043505e5600fd88af7f051d3610ca"
    ]
}
```

**Results:**
```
{
    "jsonrpc": "2.0",
    "result": {
        "transaction_hash": "0x21933cb48e59c74caa4575a78e89e6046d043505e5600fd88af7f051d3610ca",
        "type": "DECLARE",
        "version": "0x0",
        "nonce": "0x0",
        "max_fee": "0x0",
        "class_hash": "0x6e8ca9da926f68311da0cb08c3e953ef9952af277e1753d279ad1176c2e4811",
        "sender_address": "0x1",
        "signature": []
    },
    "id": 1
}
```
**Expected results:**
Reponse should not contain nonce according to [Starknet spec](https://github.com/starkware-libs/starknet-specs/blob/0e859ff905795f789f1dfd6f7340cdaf5015acc8/api/starknet_api_openrpc.json#L1608).

Exca-DK:
Closes #1123

### DIFF
--- a/rpc/handlers.go
+++ b/rpc/handlers.go
@@ -224,10 +224,11 @@ func (h *Handler) BlockWithTxs(id BlockID) (*BlockWithTxs, *jsonrpc.Error) {
 }
 
 func adaptTransaction(t core.Transaction) *Transaction {
+	var txn *Transaction
 	switch v := t.(type) {
 	case *core.DeployTransaction:
 		// https://github.com/starkware-libs/starknet-specs/blob/a789ccc3432c57777beceaa53a34a7ae2f25fda0/api/starknet_api_openrpc.json#L1521
-		return &Transaction{
+		txn = &Transaction{
 			Type:                TxnDeploy,
 			Hash:                v.Hash(),
 			ClassHash:           v.ClassHash,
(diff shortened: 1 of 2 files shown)


rpc/handlers.go
```diff
@@ -293,13 +293,19 @@ func adaptInvokeTransaction(t *core.InvokeTransaction) *Transaction {
 // https://github.com/starkware-libs/starknet-specs/blob/a789ccc3432c57777beceaa53a34a7ae2f25fda0/api/starknet_api_openrpc.json#L1340
 func adaptDeclareTransaction(t *core.DeclareTransaction) *Transaction {
 	sig := t.Signature()
+
+	nonce := t.Nonce
+	if t.Version.IsZero() {
+		nonce = nil
+	}
```
omerfirmak, line 300: Do you think we can generalize this and move it to `adaptTransaction()`?

``` golang
if t.Version.IsZero() && type != L1Handler {
 t.Nonce = nil
}
```

rpc/handlers.go
```diff
@@ -267,6 +268,11 @@ func adaptTransaction(t core.Transaction) *Transaction {
 	default:
 		panic("not a transaction")
 	}
+
+	if txn.Version.IsZero() && txn.Type != TxnL1Handler {
+		txn.Nonce = nil
+	}
+	return txn
```
omerfirmak, line 275: @wojciechos I did checked from the spec, but can you double check that this is correct.

wojciechos: Yeah, seems to be correct.

omerfirmak (approved):

PR #1127, opened 2023-08-19
vm/rust/src: Abort execution at reverted tx

### LINKED ISSUE #1108: `SimulateTransactions` should fail if one of the txns gets reverted

Exca-DK:
Should close #1108 if I understood correctly.
Not sure how to test it, `vm_test.go` doesn't have much content that I could base on.

### DIFF
--- a/vm/rust/src/lib.rs
+++ b/vm/rust/src/lib.rs
@@ -244,17 +244,31 @@ pub extern "C" fn cairoVMExecute(
                 );
                 return;
             }
-            Ok(t) => unsafe {
-                JunoAppendActualFee(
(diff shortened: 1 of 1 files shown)


omerfirmak:
Thanks for giving this a shot!

The issue is trying to describe the behaviour that is expected from an RPC endpoint. 
`vm.Execute()` is a general purpose execution API (not a RPC handler) so it shouldnt be affected in any shape or form.

What needs to be done is (in RPC handlers), after calling vm.Execute(), check if any of the txns failed and return an error.

Closing this, since I want to clear our PR backlog a bit. Feel free to open a new one.

PR #1132, opened 2023-08-21
Include starknet-rs tests in CI

### LINKED ISSUE #1114: Add starknet-rs tests to CI
https://github.com/xJonathanLEI/starknet-rs
After staging is updated, we can run starknet-rs testsuite against our goerli node.

quoting jonathan from slack 
> replace all occurrences of https://rpc-goerli-1.starknet.rs/rpc/v0.3 to your local node. The test data were meant for goerli-1 so you need a node for that network. Then run cargo test --all

wojciechos:
closes #1114

[Sample results](https://github.com/NethermindEth/juno/actions/runs/5924870496/job/16063222044#step:4:386) - 4 tests are failing but those issues are already reported

### DIFF
--- a/.github/workflows/upgrade-nodes-on-test-envs.yml
+++ b/.github/workflows/upgrade-nodes-on-test-envs.yml
@@ -93,6 +93,14 @@ jobs:
                 nohup sh -c "docker logs -f juno_$network > /var/log/juno_logs_$network.log 2>&1" &>/dev/null &
             done
 
+    - name: Trigger starknet-rs tests
+      uses: benc-uk/workflow-dispatch@v1
+      with:
+        workflow: starknet-rs-tests.yml
+        token: ${{ secrets.REPOSITORY_DISPATCH_TOKEN }}
+        ref: ${{ github.ref }}
+        inputs: '{"NODE_URL": "http://${{ env.HOST }}:6065"}'
+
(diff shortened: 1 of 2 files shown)


.github/workflows/upgrade-nodes-on-test-envs.yml
```diff
@@ -93,6 +93,14 @@ jobs:
...
             done
 
+    - name: Trigger starknet-rs tests
+      uses: benc-uk/workflow-dispatch@v1
+      with:
+        workflow: starknet-rs-tests.yml
+        token: ${{ secrets.REPOSITORY_DISPATCH_TOKEN }}
+        ref: ${{ github.ref }}
+        inputs: '{"NODE_URL": "http://${{ env.HOST }}:6065"}'
+
```
omerfirmak, line 103: I think we should put some delay here, for node to boot up, catch up to the network and download the pending block as well.

wojciechos: Not sure if it's needed. Before the tests run, the workflow already spends time starting a new agent, checking out the `starknet-rs` repo, and setting up Rust which includes downloading and compiling. This should add enough of a delay on its own. I'd rather not add a fixed delay if we don't need to.

Issue #1139, opened 2023-08-22
Trace should include revert_reason for reverted transactions


omerfirmak:
``` json
  "execute_invocation": {
      "description": "the trace of the __execute__ call or constructor call, depending on the transaction type (none for declare transactions)",
      "oneOf": [
          {
              "$ref": "#/components/schemas/FUNCTION_INVOCATION"
          },
          {
              "type": "object",
              "properties": {
                  "revert_reason": {
                      "name": "revert reason",
                      "description": "the revert reason for the failed execution",
                      "type": "string"
                  }
              }
          }
      ]
  },
```

we just omit "execute_invocation" if the transaction is reverted

an example from goerli

``` json
{
	"jsonrpc":"2.0",
	"method":"starknet_traceTransaction",
	"params":["0x69dcb2c4016907c5e3dc10202a46e0813ce376f18d072846d80bb9d5e6d9658"],
	"id":1
}
```

```json
{
    "jsonrpc": "2.0",
    "result": {
        "validate_invocation": {
            "contract_address": "0x3cab593795e32a6cdcb79ad3c67647d80a4d309839bfbb75cdc15e2423ac12b",
            "entry_point_selector": "0x162da33a4585851fe8d3af3c2a9c60b557814e221e0d4f30ff0b2189d9c7775",
            "calldata": [
                "0x1",
                "0x49d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7",
                "0x83afd3f4caedc6eebf44246fe54e38c95e3179a5ec9ea81740eca5b482d12e",
                "0x0",
                "0x3",
                "0x3",
                "0x439bd9d7bf44db6f065b44c00583542e42afbd0e9183f65f1341f2a21c50f4c",
                "0x5af3107a4000",
                "0x0"
            ],
            "caller_address": "0x0",
            "class_hash": "0x3131fa018d520a037686ce3efddeab8f28895662f019ca3ca18a626650f7d1e",
            "entry_point_type": "EXTERNAL",
            "call_type": "CALL",
            "result": [],
            "calls": [
                {
                    "contract_address": "0x3cab593795e32a6cdcb79ad3c67647d80a4d309839bfbb75cdc15e2423ac12b",
                    "entry_point_selector": "0x162da33a4585851fe8d3af3c2a9c60b557814e221e0d4f30ff0b2189d9c7775",
                    "calldata": [
                        "0x1",
[...]

PR #1140, opened 2023-08-22
Add revert_reason to execute_invocation in trace API

kirugan:
Issue #1139

### DIFF
--- a/vm/rust/src/jsonrpc.rs
+++ b/vm/rust/src/jsonrpc.rs
@@ -14,8 +14,7 @@ pub enum TransactionTrace {
     Common {
         #[serde(skip_serializing_if = "Option::is_none")]
         validate_invocation: Option<FunctionInvocation>,
-        #[serde(skip_serializing_if = "Option::is_none")]
-        execute_invocation: Option<FunctionInvocation>,
+        execute_invocation: ExecuteInvocation,
         #[serde(skip_serializing_if = "Option::is_none")]
         fee_transfer_invocation: Option<FunctionInvocation>,
     },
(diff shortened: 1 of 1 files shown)


vm/rust/src/jsonrpc.rs
```diff
@@ -66,9 +74,12 @@ pub fn new_transaction_trace(tx: StarknetApiTransaction, info: BlockifierTxInfo)
...
                     None => None,
                 },
-                execute_invocation: match info.execute_call_info {
-                    Some(v) => Some(v.into()),
-                    None => None,
+                execute_invocation: match info.revert_error {
+                    Some(str) => ExecuteInvocation::RevertReason(str),
+                    None => match info.execute_call_info {
+                        Some(v) => ExecuteInvocation::ExecuteInvocation(Some(v.into())),
+                        None => ExecuteInvocation::ExecuteInvocation(None),
```
omerfirmak, line 81: dont we need still need         `#[serde(skip_serializing_if = "Option::is_none")]`  and the change below?

```suggestion
                        None => None,
```

kirugan: I trued `#[serde(skip_serializing_if = "Option::is_none")]` with current enum and it didn't work. `None => None` just alone won't work either because `execute_invocation` is enum, but maybe I should make it `Option<ExecuteInvocation>` and get rid of Option in ExecuteInvocation::ExecuteInvocation. wdyt ? also will come up with better naming xD

omerfirmak: > Option<ExecuteInvocation> and get rid of Option in ExecuteInvocation::ExecuteInvocation. wdyt ?

Yeah, I think that makes the most sense.

omerfirmak (approved):

PR #1142, opened 2023-08-22
Update readme with new RPC methods and snapshots

wojciechos:


omerfirmak (approved):
I think we can remove the old ones.

wojciechos:
> I think we can remove the old ones.

done, I've also added trace methods under `Supported Features` section

Issue #1146, opened 2023-08-23
Merge feeder and gateway into client


rianhughes:
https://github.com/NethermindEth/juno/issues/1015

Merge the clients/feeder and clients/gateway packages into the clients package:
- A single Client struct (with private post and get) would implement the Feeder (needs to be defined) and Gateway (needs to be moved from rpc package to the clients package) interface

PR #1151, opened 2023-08-23
Rianhughes/merge feeder gateway clients

rianhughes:
closes: https://github.com/NethermindEth/juno/issues/1146


omerfirmak:
Old enough to close for now, we can revisit it later if needed

Issue #940, opened 2023-07-21
Implement CasmClass (ie Cairo1 Compiled Class) hash calculation


omerfirmak:

IronGauntlets:
Can you please provide some more context why this is required? Isn't `Cairo1Class.Hash()` sufficient for where we need to use class hash?

joshklop:
> Can you please provide some more context why this is required? Isn't `Cairo1Class.Hash()` sufficient for where we need to use class hash?

If I'm not mistaken, we will need this for verification once Juno compiles the classes.

PR #1154, opened 2023-08-23
Abdessamed/implement casm class hash calculation

Brivan-26:
Closes [Implement CasmClass (ie Cairo1 Compiled Class) hash calculation](https://github.com/NethermindEth/juno/issues/940)


core/class.go
```diff
@@ -56,7 +56,28 @@ type Cairo1Class struct {
 	Program         []*felt.Felt
 	ProgramHash     *felt.Felt
 	SemanticVersion string
-	Compiled        json.RawMessage
+	Compiled        CompiledClass
+}
+
+type CompiledClass struct {
+	Bytecode        []*felt.Felt        `json:"bytecode"`
```
joshklop, line 63: We don't need json tags in the core package.

core/class.go
```diff
@@ -56,7 +56,28 @@ type Cairo1Class struct {
...
+	PythonicHints   json.RawMessage     `json:"pythonic_hints"`
+	CompilerVersion string              `json:"compiler_version"`
+	Hints           json.RawMessage     `json:"hints"`
+	EntryPoints     CompiledEntryPoints `json:"entry_points_by_type"`
+	Prime           string              `json:"prime"`
```
joshklop, line 68: Do we need to store the `Prime`? It will be the same for all classes.

core/class.go
```diff
@@ -56,7 +56,28 @@ type Cairo1Class struct {
...
+}
+
+type CompiledEntryPoints struct {
+	External    []CompiledEntryPoint `json:"EXTERNAL"`
+	L1Handler   []CompiledEntryPoint `json:"L1_HANDLER"`
+	Constructor []CompiledEntryPoint `json:"CONSTRUCTOR"`
```
joshklop, line 74: Can there be more than one constructor?

core/class.go
```diff
@@ -56,7 +56,28 @@ type Cairo1Class struct {
...
-	Compiled        json.RawMessage
+	Compiled        CompiledClass
+}
+
+type CompiledClass struct {
+	Bytecode        []*felt.Felt        `json:"bytecode"`
+	PythonicHints   json.RawMessage     `json:"pythonic_hints"`
+	CompilerVersion string              `json:"compiler_version"`
+	Hints           json.RawMessage     `json:"hints"`
+	EntryPoints     CompiledEntryPoints `json:"entry_points_by_type"`
```
joshklop, line 67: We can put the three entry point types in the `CompiledClass` struct directly. Wdyt?

vm/class.go
```diff
@@ -21,21 +21,26 @@ import (
...
+		if err != nil {
+			return nil, err
+		}
+		return json.RawMessage(jsonData), nil
 	default:
 		return nil, errors.New("not a valid class")
 	}
 
-	return json.Marshal(compiledClass)
+	return json.Marshal(compiledCairo0Class)
```
joshklop, line 43: Let's move this up to around line 31.

core/class.go
```diff
@@ -90,6 +121,23 @@ func flattenSierraEntryPoints(entryPoints []SierraEntryPoint) []*felt.Felt {
 	return result
 }
 
+func compiledEntryPoints(entryPoints []CompiledEntryPoint) []*felt.Felt {
```
joshklop, line 124: Can we prepend `flatten` to the function name for consistency?

core/class.go
```diff
@@ -79,6 +100,16 @@ func (c *Cairo1Class) Hash() *felt.Felt {
 	)
 }
 
+func (c *CompiledClass) CompiledClassHash() *felt.Felt {
```
joshklop, line 103: We can rename this to `Hash` to remove the `CompiledClass.CompiledClassHash()` stutter I think.

joshklop (changes requested):
Where is the compiled class hash calculation documented?

Brivan-26:
> Where is the compiled class hash calculation documented?

https://github.com/xJonathanLEI/starknet-rs/blob/e9bdd7efeae9c76ef62a130ccef3f4355a864797/starknet-core/src/types/contract/mod.rs#L376

core/class.go
Brivan-26: For now no, but I'm not sure where I read that in the future it may be. However, I followed the [implementation](https://github.com/xJonathanLEI/starknet-rs/blob/e9bdd7efeae9c76ef62a130ccef3f4355a864797/starknet-core/src/types/contract/mod.rs#L83)

core/class.go
Brivan-26: Totally agree.

core/class.go
joshklop: Let's go with that.

core/class.go
Brivan-26: I had to include JSON tags here for the marshaling happening on the [VM class](https://github.com/NethermindEth/juno/blob/d0e07f0f0cac1e0c656a55de87350a375cb1dd33/vm/class.go#L32), as the VM expects `CompiledClass` as json raw.

joshklop: Is it possible to use the feeder type for marshalling, similar to the Cairo0 compiled class in that example?

If not, we should have a special type in `vm/class.go` for serialization/deserialization in that case.

Brivan-26: Yes, I guess we can use the feeder type. I will go for it.

core/class.go
joshklop: All Starknet contracts should use the same prime. I've [asked](https://t.me/sncorestars/30988) in the core stars channel just to be sure.

core/class.go
joshklop: I mean move the three entry point fields into the `CompiledClass` struct and remove the `CompiledEntryPoints` type entirely.

Brivan-26: I got it, I do agree, it is much simpler.

core/class.go
Brivan-26: I don't feel okay with one thing about this: the `Prime` will be used later in `vm.Class` to adapt from `core.CompiledClass` type to `feeder.CompiledClass` type, we use a constant string to set the prime field? And if the value changes in the future we will need to update that constant.

joshklop: I think that is ok. Converting to the feeder type and passing json to rust is a hack anyway.

Brivan-26: Hahaha, I'm convinced now ;v

core/class.go
```diff
@@ -56,7 +56,25 @@ type Cairo1Class struct {
...
-	Compiled        json.RawMessage
+	Compiled        CompiledClass
+}
+
+type CompiledClass struct {
+	Bytecode        []*felt.Felt    `json:"bytecode"`
+	PythonicHints   json.RawMessage `json:"pythonic_hints"`
+	CompilerVersion string          `json:"compiler_version"`
+	Hints           json.RawMessage `json:"hints"`
+	EntryPoints     struct {
```
joshklop, line 67: We can replace the `EntryPoints` field with the `External`, `L1Handler`, and `Constructor` fields.

core/class.go
```diff
@@ -56,7 +56,25 @@ type Cairo1Class struct {
 	Program         []*felt.Felt
 	ProgramHash     *felt.Felt
 	SemanticVersion string
-	Compiled        json.RawMessage
+	Compiled        CompiledClass
+}
+
+type CompiledClass struct {
+	Bytecode        []*felt.Felt    `json:"bytecode"`
```
joshklop, line 63: We can remove json tags in `core`.

vm/class.go
```diff
@@ -20,22 +20,29 @@ import (
 	"github.com/NethermindEth/juno/utils"
 )
 
+const compiledClassPrime = "0x800000000000011000000000000000000000000000000000000000000000001"
+
 func marshalCompiledClass(class core.Class) (json.RawMessage, error) {
-	var compiledClass any
+	var compiledCairo0Class any
```
joshklop, line 26: Can we remove this line?

joshklop (changes requested):
A few more small things. Let's wait to see if there's an answer on the Cairo prime thing before merging.

core/class.go
Brivan-26: Do you mean, the `External`, `L1Handler`, and `Constructor` fields will be top-level fields of the `CompiledClass` struct?

joshklop: Yes.

Brivan-26: We can do so, but it will not match the definition reference and the response format.

joshklop: This is a core type, it does not need to match.

adapters/feeder2core/feeder2core.go
```diff
@@ -218,7 +219,7 @@ func AdaptDeployAccountTransaction(t *feeder.Transaction) *core.DeployAccountTra
 	}
 }
 
-func AdaptCairo1Class(response *feeder.SierraDefinition, compiledClass json.RawMessage) (core.Class, error) {
+func AdaptCairo1Class(response *feeder.SierraDefinition, compiledClass *feeder.CompiledClass) (core.Class, error) {
```
joshklop, line 222: This can return *core.Cairo1Class. Let's make a separate adapter `AdaptCompiledClass`.

joshklop:
```go
func AdaptCairo1Class(*feeder.SierraDefinition) (*core.Cairo1Class, error)

func AdaptCompiledClass(*feeder.CompiledClass) (*core.CompiledClass, error)
```
We can do something like this^

adapters/feeder2core/feeder2core.go
```diff
@@ -245,14 +246,58 @@ func AdaptCairo1Class(response *feeder.SierraDefinition, compiledClass json.RawM
 		class.EntryPoints.Constructor[index] = core.SierraEntryPoint{Index: v.Index, Selector: v.Selector}
 	}
 
-	if compiledClass != nil {
-		if err = json.Unmarshal(compiledClass, &class.Compiled); err != nil {
-			return nil, err
-		}
+	if class.Compiled, err = AdaptCompiledClass(compiledClass); err != nil {
```
joshklop, line 249: Can the `compiledClass` ever be `nil`?

adapters/feeder2core/feeder2core.go
```diff
@@ -245,14 +246,58 @@ func AdaptCairo1Class(response *feeder.SierraDefinition, compiledClass json.RawM
...
+		return nil, err
 	}
+
 	return class, nil
 }
 
+func AdaptCompiledClass(compiledClass *feeder.CompiledClass) (core.CompiledClass, error) {
```
joshklop, line 256: Let's return a pointer `*core.CompiledClass`.

adapters/feeder2core/feeder2core.go
```diff
@@ -245,14 +246,58 @@ func AdaptCairo1Class(response *feeder.SierraDefinition, compiledClass json.RawM
...
+	compiled := core.CompiledClass{}
+	compiled.Bytecode = compiledClass.Bytecode
+	pyHints, err := json.Marshal(compiledClass.PythonicHints)
```
joshklop, line 259: Can we let `compiled.PythonicHints` be `any` so we don't need to marshal here?

adapters/feeder2core/feeder2core.go
```diff
@@ -245,14 +246,58 @@ func AdaptCairo1Class(response *feeder.SierraDefinition, compiledClass json.RawM
...
+	if err != nil {
+		return core.CompiledClass{}, err
+	}
+	compiled.PythonicHints = pyHints
+	compiled.CompilerVersion = compiledClass.CompilerVersion
+	var success bool
+	compiled.Prime, success = new(big.Int).SetString(compiledClass.Prime, 0)
+	if !success {
+		return core.CompiledClass{}, fmt.Errorf("couldn't convert prime value to big.Int")
```
joshklop, line 268:
```suggestion
		return core.CompiledClass{}, fmt.Errorf("couldn't convert prime value to big.Int: %d", compiled.Prime)
```
or something like that

adapters/feeder2core/feeder2core.go
```diff
@@ -245,14 +246,58 @@ func AdaptCairo1Class(response *feeder.SierraDefinition, compiledClass json.RawM
...
+	}
+	hints, err := json.Marshal(compiledClass.Hints)
```
joshklop, line 270: Same idea as `PythonHints`

adapters/feeder2core/feeder2core.go
Brivan-26: As far as I know, `json.Marshal` doesn't accept types `any` or map[interface{}]interface{}. And we need to marshal `Compiled` field in `vm.Class`

joshklop: Marshal accepts `any` ([docs](https://pkg.go.dev/encoding/json#Marshal)). Did you mean something else?

Brivan-26: I had to clarify better: when marshling nested types like structs, json.Marshal can't marshal fields of type any as they are unsupported

joshklop: It [seems to work](https://play.golang.com/p/6aJgvJJkkeJ).

clients/feeder/testdata/integration/compiled_class/0x6d8ede036bb4720e6f348643221d8672bf4f0895622c32c11e57460b3b7dffc.json
joshklop, line 1: nit: can we condense these to a single line to minimize the diff?

vm/class.go
```diff
@@ -135,6 +138,44 @@ func makeSierraClass(class *core.Cairo1Class) *starknet.SierraDefinition {
 	}
 }
 
+func makeCairo1CompiledClass(coreCompiledClass *core.CompiledClass) feeder.CompiledClass {
```
joshklop, line 141: What do you think about putting an adapter in `adapters/core2sn`?

joshklop (commented):
Apologies if I started reviewing before it was ready. Didn't see it was still wip until I was half-way through :)

core/class_test.go
```diff
@@ -40,6 +40,17 @@ func TestClassV1Hash(t *testing.T) {
...
 
+func TestCompiledClassHash(t *testing.T) {
+	client := feeder.NewTestClient(t, utils.Integration)
+	gw := adaptfeeder.New(client)
+
+	hash := utils.HexToFelt(t, "0x6d8ede036bb4720e6f348643221d8672bf4f0895622c32c11e57460b3b7dffc")
+	class, err := gw.Class(context.Background(), hash)
+	require.NoError(t, err)
+	got := class.(*core.Cairo1Class).Compiled.Hash()
+	assert.Equal(t, "0x18f95714044fd5408d3bf812bcd249ddec098ab3cd201b7916170cfbfa59e05", got.String())
```
joshklop, line 51: Where was this calculated?

core/class.go
```diff
@@ -56,7 +57,24 @@ type Cairo1Class struct {
...
+	Hints           json.RawMessage
+	Prime           *big.Int
+	External        []CompiledEntryPoint
+	L1Handler       []CompiledEntryPoint
+	Constructor     []CompiledEntryPoint
+}
+
+type CompiledEntryPoint struct {
+	Offset   uint64
+	Builtins []string
```
joshklop, line 76: Do you think we should have a `Builtin` enum?

adapters/core2sn/core2sn.go
```diff
@@ -0,0 +1,122 @@
...
+	"github.com/NethermindEth/juno/utils"
+)
+
+func AdaptCompiledClass(coreCompiledClass *core.CompiledClass) starknet.CompiledClass {
+	feederCompiledClass := new(starknet.CompiledClass)
+	feederCompiledClass.Bytecode = coreCompiledClass.Bytecode
+	feederCompiledClass.PythonicHints = coreCompiledClass.PythonicHints
+	feederCompiledClass.CompilerVersion = coreCompiledClass.CompilerVersion
+	feederCompiledClass.Hints = coreCompiledClass.Hints
+	feederCompiledClass.Prime = "0x" + coreCompiledClass.Prime.Text(16) //nolint:gomnd
```
joshklop, line 15:
```suggestion
	feederCompiledClass.Prime = "0x" + coreCompiledClass.Prime.Text(felt.Base16)
```

core/class.go
```diff
@@ -79,6 +97,16 @@ func (c *Cairo1Class) Hash() *felt.Felt {
 	)
 }
 
+func (c *CompiledClass) Hash() *felt.Felt {
+	return crypto.PoseidonArray(
+		new(felt.Felt).SetBytes([]byte("COMPILED_CLASS_V1")),
```
joshklop, line 102: Can we make this a top-level `var`?

vm/class.go
```diff
@@ -13,30 +13,29 @@ import "C"
...
-		compiledClass, err = makeDeprecatedVMClass(c)
+		compiledCairo0Class, err := core2sn.AdaptCairo0Class(c)
 		if err != nil {
 			return nil, err
 		}
+		return json.Marshal(compiledCairo0Class)
 	case *core.Cairo1Class:
-		compiledClass = c.Compiled
+		compiledCairo1Class := core2sn.AdaptCompiledClass(&c.Compiled)
+		// we adapt the core type to the feeder type to avoid using JSON tags in core.Class.CompiledClass
```
joshklop, line 34: nit: can we move this comment above line 33?

core/class.go
joshklop: I believe adding enum variants is a breaking change that should only happen across network versions.

But yeah, probably not worth making this an enum right now.

joshklop (approved):

core/class_test.go
joshklop: Hmmm, I'm seeing a different `compiled_class_hash` https://external.integration.starknet.io/feeder_gateway/get_transaction?transactionHash=0x79a7e03b31e0dbc72c3e092cc6459e8ef4e0c26a7221038028736487933c3

`0x57e51d6856a7f80cbe11a9197bf498efa0a8daf4db4d651569c0ba914f7d8ea`

PR #1156, opened 2023-08-23
Remove pre-v0.12.1 code

joshklop:


joshklop:
Non-essential and requires significant refactor of tests to get right. Real solution is to stop using the feeder entirely and maintain a "mempool" #1134

Issue #1131, opened 2023-08-21
Add BlockAndStateUpdate family of functions to StarknetData


omerfirmak:
We can use the new feeder endpoint and fallback to separate GetBlock & GetStateUpdate requests if the new endpoint is not available yet.

Issue #1089, opened 2023-08-16
Make use of the new `get_state_update_with_block` feeder endpoint


omerfirmak:
https://external.integration.starknet.io/feeder_gateway/get_state_update?blockNumber=pending&includeBlock=true
We can use it both during sync and when polling pending block

PR #1167, opened 2023-08-24
Fetch block and state update at once

Brivan-26:
Closes both [Add BlockAndStateUpdate family of functions to StarknetData](https://github.com/NethermindEth/juno/issues/1131) and [Make use of the new get_state_update_with_block feeder endpoint #1089 ](https://github.com/NethermindEth/juno/issues/1089)

Some refactoring will be necessary once the new endpoint is available on v0.12.2


starknetdata/feeder/feeder.go
```diff
@@ -113,3 +113,35 @@ func (f *Feeder) StateUpdate(ctx context.Context, blockNumber uint64) (*core.Sta
 func (f *Feeder) StateUpdatePending(ctx context.Context) (*core.StateUpdate, error) {
 	return f.stateUpdate(ctx, "pending")
 }
+
+func (f *Feeder) stateUpdateWithBlock(ctx context.Context, blockID string) (*core.StateUpdate, *core.Block, error) {
```
omerfirmak, line 117: you can move the fallback logic to here. It doesnt need to be in `sync` package

Brivan-26: I avoided that to not be forced to add the fallbacks to the `StarknetData` interface, and hence inject them in mocks too.

omerfirmak: This fallback mechanism is a feeder specific detail that we shouldnt leak in to `sync` package. 

> hence inject them in mocks too

only `StarknetData` interface affects the mocks. The fallback that you implement in `Feeder` has no affect on mocks.

Brivan-26: True, thanks

clients/feeder/feeder.go
```diff
@@ -222,6 +222,9 @@ func (c *Client) get(ctx context.Context, queryURL string) (io.ReadCloser, error
 					return res.Body, nil
 				} else {
 					err = errors.New(res.Status)
+					if err.Error() == "400 Bad Request" {
```
joshklop, line 225: Can we check the `res.Status` instead of matching strings?

starknetdata/starknetdata.go
```diff
@@ -18,4 +18,6 @@ type StarknetData interface {
 	Class(ctx context.Context, classHash *felt.Felt) (core.Class, error)
 	StateUpdate(ctx context.Context, blockNumber uint64) (*core.StateUpdate, error)
 	StateUpdatePending(ctx context.Context) (*core.StateUpdate, error)
+	StateUpdateWithBlock(ctx context.Context, blockNumber uint64) (*core.StateUpdate, *core.Block, error)
```
joshklop, line 21: I think this will make a lot of tests so much simpler.

starknetdata/feeder/feeder.go
```diff
@@ -113,3 +113,76 @@ func (f *Feeder) StateUpdate(ctx context.Context, blockNumber uint64) (*core.Sta
...
+	response, err := f.client.StateUpdateWithBlock(ctx, blockID)
+	if err != nil {
+		// TODO: remove once the new feeder is available
+		if err.Error() == "400 Bad Request" {
```
joshklop, line 121: I feel like we should be to use `errors.Is` here, but maybe string matching is necessary.

clients/feeder/feeder.go
```diff
@@ -163,7 +163,7 @@ func handleNotFound(dir, queryArg string, w http.ResponseWriter) {
 	if dir == "transaction" && queryArg == "transactionHash" {
 		w.Write([]byte("{\"finality_status\": \"NOT_RECEIVED\", \"status\": \"NOT_RECEIVED\"}")) //nolint:errcheck
 	} else {
-		w.WriteHeader(http.StatusNotFound)
+		w.WriteHeader(http.StatusBadRequest)
```
joshklop, line 166: I don't think this change is relevant to this PR. Please correct me if I'm wrong.

starknetdata/feeder/feeder.go
```diff
@@ -48,7 +48,7 @@ func (f *Feeder) block(ctx context.Context, blockID string) (*core.Block, error)
 		return nil, err
 	}
 
-	if blockID == "pending" && response.Status != "PENDING" {
+	if blockID == "pending" && response.Status != "PENDING" { //nolint:goconst
```
joshklop, line 51: Probably better to fix the lint error than add a `nolint` here.

starknetdata/feeder/feeder_test.go
```diff
@@ -188,3 +188,43 @@ func TestClassV1(t *testing.T) {
...
+
+func TestStateUpdateWithBlock(t *testing.T) {
+	numbers := []uint64{0, 78541}
+
+	client := feeder.NewTestClient(t, utils.INTEGRATION)
+	adapter := adaptfeeder.New(client)
+	ctx := context.Background()
+
+	for _, number := range numbers {
+		t.Run("integration block number"+strconv.FormatUint(number, 10), func(t *testing.T) {
```
joshklop, line 200:
```suggestion
		t.Run("integration block number "+strconv.FormatUint(number, 10), func(t *testing.T) {
```

starknetdata/feeder/feeder_test.go
```diff
@@ -188,3 +188,43 @@ func TestClassV1(t *testing.T) {
...
+			response, err := client.StateUpdateWithBlock(ctx, strconv.FormatUint(number, 10))
+			require.NoError(t, err)
+			stateUpdate, block, err := adapter.StateUpdateWithBlock(ctx, number)
+			require.NoError(t, err)
+			adaptedBlock, err := feeder2core.AdaptBlock(response.Block)
+			require.NoError(t, err)
+			adaptedStateUpdate, err := feeder2core.AdaptStateUpdate(response.StateUpdate)
+			require.NoError(t, err)
+			assert.Equal(t, adaptedBlock, block)
+			assert.Equal(t, adaptedStateUpdate, stateUpdate)
```
joshklop, line 210: Swap the expected and received values. Expected comes first in `testify` helpers IIRC.

joshklop: Same for other test.

starknetdata/feeder/feeder_test.go
```diff
@@ -188,3 +188,43 @@ func TestClassV1(t *testing.T) {
...
+
+func TestStateUpdatePendingWithBlock(t *testing.T) {
+	client := feeder.NewTestClient(t, utils.INTEGRATION)
+	adapter := adaptfeeder.New(client)
+	ctx := context.Background()
+
+	response, err := client.StateUpdateWithBlock(ctx, "pending")
+	require.NoError(t, err)
+	stateUpdate, block, err := adapter.StateUpdatePendingWithBlock(ctx)
+	require.NoError(t, err)
```
joshklop, line 223: Let's move this above or below the logic to get the `adaptedBlock` & `adaptedStateUpdate` data. Right now it's in between, which makes the test harder to read.

sync/sync_test.go
```diff
@@ -131,6 +131,9 @@ func TestSyncBlocks(t *testing.T) {
 
 			return ret, nil
 		}).AnyTimes()
+		mockSNData.EXPECT().StateUpdateWithBlock(gomock.Any(), gomock.Any()).DoAndReturn(func(_ context.Context, height uint64) (*core.StateUpdate, *core.Block, error) {
```
joshklop, line 134: Can we remove the calls to `mockSNData...BlockByNumber`, etc.?

joshklop (changes requested):

clients/feeder/feeder.go
Brivan-26: The added check here will be removed and just set the error to res.Status like we're doing now as the new endpoint is available on mainnet.

starknetdata/feeder/feeder.go
Brivan-26: I guess string matching is necessary here. But like the first review, this check will be removed in the next commit.

clients/feeder/feeder.go
Brivan-26: I found it relevant for test cases (as the feeder endpoint didn't exist on mainnet) to simulate real scenarios, I changed it to `StatusBadRequest` as the real feeder endpoint returns that status in case of an unexisting endpoint

starknetdata/starknetdata.go
Brivan-26: It did!

sync/sync_test.go
```diff
@@ -94,42 +92,8 @@ func TestSyncBlocks(t *testing.T) {
...
-				ret.BlockHash = new(felt.Felt) // fail sanity checks
-			case 3:
-				ret.OldRoot = new(felt.Felt).SetUint64(1) // fail store
-			default:
-				reqCount = 0
-				atomic.AddUint64(&syncingHeight, 1)
-			}
-
-			return ret, nil
+		mockSNData.EXPECT().StateUpdateWithBlock(gomock.Any(), gomock.Any()).DoAndReturn(func(_ context.Context, height uint64) (*core.StateUpdate, *core.Block, error) {
```
joshklop, line 95: I think we lost test coverage here. In the old implementation, we have multiple cases that exercise different code paths: failing sanity checks, failing store, etc. Can we get that coverage back?

joshklop (commented):
Nice! Almost there I think.

sync/sync_test.go
Brivan-26: Ah.. true, missed that.

omerfirmak: @Brivan-26 any updates on this?

Brivan-26: @omerfirmak Working no adding the coverage back (getting annoying issue when using the new endpoint and trying to solve it).

omerfirmak:
Can you rebase as well?

joshklop (approved):

omerfirmak (approved):

Issue #998, opened 2023-08-03
Create an instance of Prometheus instead of `MustRegister` for metric collection


IronGauntlets:
Prometheus package creates a `defaultRegistry` singleton and provides access to it through `DefaultRegisterer` and `DefaultGatherer` and we use the `DefaultRegisterer` to register various collectors from different packages. 

We have a boolean cli flag to enable/disable metrics which is set to the global boolean variable `Enabled` (both the flag and global variable seem redundant) in the metrics package as a consequence of singleton to register various collectors from other packages. The use of the singleton hides the dependency on the metrics package and couples other packages to the `Enabled` global variable.

A better approach would be to pass the Prometheus Registry to the package which wants to register collectors similar to how loggers are passed to various packages. A metrics interface maybe be beneficial in the future.

Related: #980

Exca-DK:
Hey, I could try to implement this if it's still available.

IronGauntlets:
Yes, please go ahead

Issue #980, opened 2023-08-01
Explore lightweight metrics options


joshklop:
See [here](https://github.com/NethermindEth/juno/pull/919#discussion_r1264800601) for reference. We use prometheus directly for convenience right now, but it would be nice to let users of Juno packages to use their own metrics collection methods.

derrix060:
closing as it's being stale for a long time.  If we ever see the need to explore other metrics options we can create a new issue

PR #1168, opened 2023-08-24
Metrics: Implement metrics abstraction

Exca-DK:
This PR attempts to implement #998 and #980. The general idea is to pass a factory interface along the packages for the registration of metrics. By doing so, it enables abstraction over the current Prometheus metrics system, allowing for potential swapping to a different metrics system in the future. 
Please note that this currently is just a concept; If it's something you would want me to pursue, it can be further developed.


joshklop:
Thanks so much for the initiative.

What do you think of using the Prometheus interfaces directly? E.g., [histogram](https://github.com/prometheus/client_golang/blob/d03abf3a31c973a5bc2c2dc698fb41b661a0f0c5/prometheus/histogram.go#L242), [counter](https://github.com/prometheus/client_golang/blob/d03abf3a31c973a5bc2c2dc698fb41b661a0f0c5/prometheus/counter.go#L34)

We can have nop implementations for each interface to use for defaults, and then have `With<Name>Metric` methods on each struct that requires a metric. For example: 
```go
func (s *Synchronizer) WithNumTransactionsSeenMetric(c prometheus.Counter) *Synchronizer
```
That would also enable us to test that the metrics behave properly and only import `prometheus` in the `node` package.

(apologies for the late review--@omerfirmak and I have been out for the last week or so)

Exca-DK:
> What do you think of using the Prometheus interfaces directly? E.g., [histogram] [counter]

Prometheus doesn't have interfaces defined for vec types so that would need be defined somewhere and imported alongside prometheus. It would also be a little weird to depend on prometheus interfaces meanwhile using another system eg. [codvar](https://pkg.go.dev/expvar) 

> We can have nop implementations for each interface to use for defaults, and then have With<Name>Metric methods on each struct that requires a metric. For example:
func (s *Synchronizer) WithNumTransactionsSeenMetric(c prometheus.Counter) *Synchronizer

That could work but prometheus requires all of the ops to be defined at creation. It would result in package creating synchronizer defining namespace, subsystem etc. for NumTransactionsSeenMetric and then passing it to `synchronizer.WithNumTransactionsSeenMetric` which leaks context to caller. 
Maybe generic contract like this could work:
```go
func (s T) WithLightweightMetrics(c metrics.Factory) T
func (s T) WithExpensiveMetrics(c metrics.Factory) T
```
Could give some configurable options and still keep everything encapsulated while enabling testing. 	
@joshklop

joshklop:
> Prometheus doesn't have interfaces defined for vec types so that would need be defined somewhere and imported alongside prometheus. It would also be a little weird to depend on prometheus interfaces meanwhile using another system eg. codvar

Good points. What if we created interfaces for everything we need (counter, histogram, gauge, etc.) in a `metrics` package? This has the added benefit of not importing prometheus outside the `node` package, which is what we're after anyway.

I'm open to using expvar as well, if we can figure out a good way to wire it up to Prometheus, get histogram support, etc.

> leaks context to caller

Can you elaborate on this? In my mind, the caller should create the namespace, subsystem, etc. to avoid naming collisions and encourage consistency. Plus, it may be necessary to take advantage of prometheus's [testutil](https://stackoverflow.com/questions/63911468/unit-test-using-golang-prometheus-testutil), which may be easier than a huge mock that's only used for metrics.

Exca-DK:
I was imagining metrics like a typical blackbox. By leaking context I mean that the caller has to know some of the internal mechanisms, `node` in that case would have to know what exactly is `NumTransactions` in order to provide correct description for that.  It goes deeper with each nested object that requires metrics eg. `A` has its own metrics namespace and it creates `B` which has yet another namespace, The creator of `A` has to be aware of that to provide proper namespace for `B`. Maybe some of the components could be interfaces that would have slightly different descriptions based on their concrete implementation. That is not possible when the caller is responsible for creation.

joshklop:
That makes sense. I'm curious how large Go projects manage metrics. Tailscale has a custom system that seems to work well, go-ethereum vendors an old metrics library IIRC, ... have you seen any good ways to do it? Ideally our metrics setup is:

- optional: no-op defaults, for example
- testable
- open to multiple implementations
- lightweight (the prometheus package imports protobuf and possibly other large libraries)

Your current implementation is the closest I think, except for transitively importing prometheus.

Exca-DK:
Sadly I do not know the best approach. From what I`ve seen repos supporting multiple exporters usually roll their own metrics types and create adapters for exporting. Haven't seen in those cases a single one supporting labels for prometheus tho.

joshklop:
Jotting down here for future reference: Tailscale's lightweight [metrics package](https://github.com/tailscale/tailscale/blob/main/metrics/metrics.go) has an associated [prometheus exporter](https://github.com/tailscale/tailscale/blob/main/tsweb/varz/varz.go). Maybe we can create a simple interface around it? Haven't looked into it much.

PR #1169, opened 2023-08-25
migration: Gracefull cancelation

### LINKED ISSUE #1163: Graceful shutdown during db migration
If there is a long db migration taking place there is no way to gracefully Juno. Any force shutdown corrupts the DB. 

Juno should be able to gracefully shutdown during the migration and pick up where it left off when Juno is restarted.

Exca-DK:
closes #1163

### DIFF
--- a/migration/migration.go
+++ b/migration/migration.go
@@ -2,6 +2,7 @@ package migration
 
 import (
 	"bytes"
+	"context"
 	"encoding/binary"
 	"errors"
 	"fmt"
@@ -16,27 +17,34 @@ import (
 	"github.com/NethermindEth/juno/encoder"
 	"github.com/NethermindEth/juno/utils"
 	"github.com/bits-and-blooms/bitset"
+	"github.com/fxamacker/cbor/v2"
 	"github.com/sourcegraph/conc/pool"
 )
 
+type schemaMetadata struct {
+	Version           uint64
+	IntermediateState []byte
+}
+
 type Migration interface {
-	Before()
-	Migrate(db.Transaction, utils.Network) error
+	Before(intermediateState []byte) error
+	// Migration should return intermediate state whenever it requests new txn or detects cancelled ctx.
+	Migrate(context.Context, db.Transaction, utils.Network) ([]byte, error)
 }
 
 type MigrationFunc func(db.Transaction, utils.Network) error
 
 // Migrate returns f(txn).
-func (f MigrationFunc) Migrate(txn db.Transaction, network utils.Network) error {
-	return f(txn, network)
+func (f MigrationFunc) Migrate(_ context.Context, txn db.Transaction, network utils.Network) ([]byte, error) {
+	return nil, f(txn, network)
 }
 
 // Before is a no-op.
-func (f MigrationFunc) Before() {}
+func (f MigrationFunc) Before(_ []byte) error { return nil }
 
-// migrations contains a set of migrations that can be applied to a database.
+// defaultMigrations contains a set of migrations that can be applied to a database.
 // After making breaking changes to the DB layout, add new migrations to this list.
-var migrations = []Migration{
+var defaultMigrations = []Migration{
 	MigrationFunc(migration0000),
 	MigrationFunc(relocateContractStorageRootKeys),
 	MigrationFunc(recalculateBloomFilters),
@@ -56,9 +64,13 @@ var migrations = []Migration{
 
 var ErrCallWithNewTransaction = errors.New("call with new transaction")
 
-func MigrateIfNeeded(targetDB db.DB, network utils.Network, log utils.SimpleLogger) error {
+func MigrateIfNeeded(ctx context.Context, targetDB db.DB, network utils.Network, log utils.SimpleLogger) error {
+	return migrateIfNeeded(ctx, targetDB, network, log, defaultMigrations)
+}
+
+func migrateIfNeeded(ctx context.Context, targetDB db.DB, network utils.Network, log utils.SimpleLogger, migrations []Migration) error {
 	/*
-		Schema version of the targetDB determines which set of migrations need to be applied to the database.
+		Schema metadata of the targetDB determines which set of migrations need to be applied to the database.
 		After a migration is successfully executed, which may update the database, the schema version is incremented
 		by 1 by this loop.
 
@@ -73,36 +85,39 @@ func MigrateIfNeeded(targetDB db.DB, network utils.Network, log utils.SimpleLogg
 		new ones. It will be able to do this since the schema version it reads from the database will be
 		non-zero and that is what we use to initialise the i loop variable.
 	*/
-	version, err := SchemaVersion(targetDB)
+	metadata, err := SchemaMetadata(targetDB)
 	if err != nil {
 		return err
 	}
 
-	for i := version; i < uint64(len(migrations)); i++ {
+	for i := metadata.Version; i < uint64(len(migrations)); i++ {
+		if err := ctx.Err(); err != nil {
+			return err
+		}
 		log.Infow("Applying database migration", "stage", fmt.Sprintf("%d/%d", i+1, len(migrations)))
 		migration := migrations[i]
-		migration.Before()
+		if err := migration.Before(metadata.IntermediateState); err != nil {
+			return err
+		}
 		for {
 			var migrationErr error
 			if dbErr := targetDB.Update(func(txn db.Transaction) error {
-				migrationErr = migration.Migrate(txn, network)
-				if migrationErr != nil {
-					if errors.Is(migrationErr, ErrCallWithNewTransaction) {
-						return nil // Run the migration again with a new transaction.
+				metadata.IntermediateState, migrationErr = migration.Migrate(ctx, txn, network)
+				switch {
+				case migrationErr == nil || errors.Is(migrationErr, ctx.Err()):
+					if metadata.IntermediateState == nil {
+						metadata.Version++
 					}
+					return updateSchemaMetadata(txn, metadata)
+				case errors.Is(migrationErr, ErrCallWithNewTransaction):
+					return nil // Run migration again with new transaction.
+				default:
 					return migrationErr
 				}
-
-				// Migration successful. Bump the version.
-				var versionBytes [8]byte
-				binary.BigEndian.PutUint64(versionBytes[:], i+1)
-				return txn.Set(db.SchemaVersion.Key(), versionBytes[:])
 			}); dbErr != nil {
 				return dbErr
 			} else if migrationErr == nil {
 				break
-			} else if !errors.Is(migrationErr, ErrCallWithNewTransaction) {
-				return migrationErr
 			}
 		}
 	}
@@ -110,21 +125,46 @@ func MigrateIfNeeded(targetDB db.DB, network utils.Network, log utils.SimpleLogg
 	return nil
 }
 
-func SchemaVersion(targetDB db.DB) (uint64, error) {
-	version := uint64(0)
+// SchemaMetadata retrieves metadata about a database schema from the given database.
+func SchemaMetadata(targetDB db.DB) (schemaMetadata, error) {
+	metadata := schemaMetadata{}
 	txn, err := targetDB.NewTransaction(false)
 	if err != nil {
-		return 0, nil
+		return metadata, err
 	}
-	err = txn.Get(db.SchemaVersion.Key(), func(bytes []byte) error {
-		version = binary.BigEndian.Uint64(bytes)
+	if err := txn.Get(db.SchemaVersion.Key(), func(b []byte) error {
+		metadata.Version = binary.BigEndian.Uint64(b)
 		return nil
-	})
-	if err != nil && !errors.Is(err, db.ErrKeyNotFound) {
-		return 0, utils.RunAndWrapOnError(txn.Discard, err)
+	}); err != nil && !errors.Is(err, db.ErrKeyNotFound) {
+		return metadata, utils.RunAndWrapOnError(txn.Discard, err)
 	}
 
-	return version, txn.Discard()
+	if err := txn.Get(db.SchemaIntermediateState.Key(), func(b []byte) error {
+		return cbor.Unmarshal(b, &metadata.IntermediateState)
+	}); err != nil && !errors.Is(err, db.ErrKeyNotFound) {
+		return metadata, utils.RunAndWrapOnError(txn.Discard, err)
+	}
+
+	return metadata, txn.Discard()
+}
+
+// updateSchemaMetadata updates the schema in given database.
+func updateSchemaMetadata(txn db.Transaction, schema schemaMetadata) error {
+	var (
+		version [8]byte
+		state   []byte
+		err     error
+	)
+	binary.BigEndian.PutUint64(version[:], schema.Version)
+	state, err = cbor.Marshal(schema.IntermediateState)
+	if err != nil {
+		return err
+	}
+
+	if err := txn.Set(db.SchemaVersion.Key(), version[:]); err != nil {
+		return err
+	}
+	return txn.Set(db.SchemaIntermediateState.Key(), state)
 }
 
 // migration0000 makes sure the targetDB is empty
@@ -227,7 +267,7 @@ type changeTrieNodeEncoding struct {
 	}
 }
 
-func (m *changeTrieNodeEncoding) Before() {
+func (m *changeTrieNodeEncoding) Before(_ []byte) error {
 	m.trieNodeBuckets = map[db.Bucket]*struct {
 		seekTo  []byte
 		skipLen int
@@ -245,6 +285,7 @@ func (m *changeTrieNodeEncoding) Before() {
 			skipLen: 1 + felt.Bytes,
 		},
 	}
+	return nil
 }
 
 type node struct {
@@ -314,7 +355,7 @@ func (n *node) _UnmarshalBinary(data []byte) error {
 	return err
 }
 
-func (m *changeTrieNodeEncoding) Migrate(txn db.Transaction, _ utils.Network) error {
+func (m *changeTrieNodeEncoding) Migrate(_ context.Context, txn db.Transaction, _ utils.Network) ([]byte, error) {
 	// If we made n a trie.Node, the encoder would fall back to the custom encoding methods.
 	// We instead define a cutom struct to force the encoder to use the default encoding.
 	var n node
@@ -371,15 +412,15 @@ func (m *changeTrieNodeEncoding) Migrate(txn db.Transaction, _ utils.Network) er
 
 	iterator, err := txn.NewIterator()
 	if err != nil {
-		return err
+		return nil, err
 	}
 
 	for bucket, info := range m.trieNodeBuckets {
 		if err := migrateF(iterator, bucket, info.seekTo, info.skipLen); err != nil {
-			return utils.RunAndWrapOnError(iterator.Close, err)
+			return nil, utils.RunAndWrapOnError(iterator.Close, err)
 		}
 	}
-	return iterator.Close()
+	return nil, iterator.Close()
 }
 
 // calculateBlockCommitments calculates the txn and event commitments for each block and stores them separately
--- a/migration/migration_pkg_test.go
+++ b/migration/migration_pkg_test.go
@@ -3,6 +3,8 @@ package migration
 import (
 	"bytes"
 	"context"
+	"encoding/binary"
+	"errors"
 	"testing"
 
 	"github.com/NethermindEth/juno/blockchain"
@@ -137,9 +139,10 @@ func TestChangeTrieNodeEncoding(t *testing.T) {
 	}))
 
 	m := new(changeTrieNodeEncoding)
-	m.Before()
+	require.NoError(t, m.Before(nil))
 	require.NoError(t, testdb.Update(func(txn db.Transaction) error {
-		return m.Migrate(txn, utils.Mainnet)
+		_, err := m.Migrate(context.Background(), txn, utils.Mainnet)
+		return err
 	}))
 
 	require.NoError(t, testdb.Update(func(txn db.Transaction) error {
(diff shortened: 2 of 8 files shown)


migration/migration_test.go
```diff
@@ -15,16 +16,27 @@ func TestMigrateIfNeeded(t *testing.T) {
...
+	require.Equal(t, uint64(0), version)
+	ctx = context.Background()
+
+	t.Run("Migration should happen on empty DB", func(t *testing.T) {
+		require.NoError(t, migration.MigrateIfNeeded(ctx, testDB, utils.MAINNET, utils.NewNopZapLogger()))
+	})
+
+	version, err = migration.SchemaVersion(testDB)
+	require.NoError(t, err)
+	require.NotEqual(t, uint64(0), version)
```
Exca-DK, line 36: Linter didn't spot this issue

migration/migration.go
```diff
@@ -19,20 +21,38 @@ import (
 	"github.com/sourcegraph/conc/pool"
 )
 
+type schemaMetadata struct {
+	Version   uint64
+	Done      bool
+	LastState []byte
```
joshklop, line 27: Can we remove the `Done` field and rename `LastState` to `IntermediateState`? If `IntermediateState == nil`, then the db is consistent; otherwise, it's in the middle of a migration.

migration/migration.go
```diff
@@ -19,20 +21,38 @@ import (
...
+}
+
+func (schema *schemaMetadata) fromBytes(data []byte) error {
+	return json.Unmarshal(data, schema)
```
joshklop, line 31: Might be better to use `cbor` for consistency. We can discuss the more important parts of this PR before worrying about that, though.

migration/migration.go
```diff
@@ -19,20 +21,38 @@ import (
...
+func (schema schemaMetadata) toBytes() ([]byte, error) {
+	return json.Marshal(schema)
+}
+
 type Migration interface {
-	Before()
-	Migrate(db.Transaction, utils.Network) error
+	Before(key []byte) error
+	Migrate(context.Context, db.Transaction, utils.Network) error
+	State() []byte
```
joshklop, line 41: Instead of adding the `State` method, can we make `Migrate` return `[]byte, error`, where the `[]byte` slice is the `IntermediateState`?

joshklop (commented):
I agree with the general approach, but I don't think I've thought through all the edge cases.

Let me know what you think of my proposed simplifications and we can go from there. Thanks for the thorough work as always.

Exca-DK:
Thanks for the review! I made the logic slightly more consistent and simpler based on your review and added few docs/comments. 
The expected behavior is: 

> Migration should return intermediate state whenever it requests new txn or detects cancelled ctx.

One thing to note is that `MigrationFunc` doesn't support currently intermediate steps. I saw the discord user being stuck at `calculateBlockCommitments` for hours which resulted in abortion and syncing from snapshot instead. Maybe in other pr they could be transformed to support it or even this one?

migration/migration.go
```diff
@@ -56,47 +73,69 @@ var migrations = []Migration{
...
 		log.Infow("Applying database migration", "stage", fmt.Sprintf("%d/%d", i+1, len(migrations)))
 		migration := migrations[i]
-		migration.Before()
-		for {
+		err := migration.Before(metadata.IntermediateState)
+		if err != nil {
+			return err
+		}
+		var intermediateState []byte
+		for x := uint64(0); ; x++ {
```
joshklop, line 117:
```suggestion
		for {
```

migration/migration.go
```diff
@@ -56,47 +73,69 @@ var migrations = []Migration{
 
 var ErrCallWithNewTransaction = errors.New("call with new transaction")
 
-func MigrateIfNeeded(targetDB db.DB, network utils.Network, log utils.SimpleLogger) error {
+func MigrateIfNeededWithCallback(ctx context.Context, targetDB db.DB, network utils.Network, log utils.SimpleLogger, f func()) error {
```
joshklop, line 76: It appears this is used only for testing purposes. We should test `MigrateIfNeeded` directly if possible. 

That may require removing the `migrations` global variable. If so, one of us can open an issue about refactoring it out somehow.

migration/migration.go
```diff
@@ -56,47 +73,69 @@ var migrations = []Migration{
...
 				}
-
-				// Migration successful. Bump the version.
-				var versionBytes [8]byte
-				binary.BigEndian.PutUint64(versionBytes[:], i+1)
-				return txn.Set(db.SchemaVersion.Key(), versionBytes[:])
+				// Migration successful. Update metadata with new version and none intermediate
+				metadata.Version++
+				metadata.IntermediateState = nil
+				return updateSchemaMetadata(txn, metadata)
```
joshklop, line 138: I think we can replace this with:
```go
metadata.IntermediateState, migrationErr = migration.Migrate(ctx, txn, network)
switch migrationErr {
case ctx.Err():
case nil:
case ErrCallWithNewTransaction:
	return nil // Run migration again with new transaction.
default:
	return migrationErr
}
if metadata.IntermediateState == nil {
	metadata.Version++
}
return updateSchemaMetadata(txn, metadata)
```

migration/migration.go
```diff
@@ -105,27 +144,52 @@ func MigrateIfNeeded(targetDB db.DB, network utils.Network, log utils.SimpleLogg
...
-		return 0, db.CloseAndWrapOnError(txn.Discard, err)
+		return metadata, db.CloseAndWrapOnError(txn.Discard, err)
+	}
+
+	return metadata, db.CloseAndWrapOnError(txn.Discard, nil)
+}
+
+// UpdateSchemaMetadata updates the metadata for a database schema
+// in given database using the specified version number and intermediateState.
+func UpdateSchemaMetadata(targetDB db.DB, version uint64, intermediateState []byte) error {
```
joshklop, line 173: I think we can remove this.

migration/migration.go
```diff
@@ -105,27 +144,52 @@ func MigrateIfNeeded(targetDB db.DB, network utils.Network, log utils.SimpleLogg
...
+		Version:           version,
+		IntermediateState: intermediateState,
 	}
+	return targetDB.Update(func(txn db.Transaction) error {
+		return updateSchemaMetadata(txn, schema)
+	})
+}
 
-	return version, db.CloseAndWrapOnError(txn.Discard, nil)
+func updateSchemaMetadata(txn db.Transaction, schema schemaMetadata) error {
```
joshklop, line 183: We can probably remove this as well.

joshklop (commented):
> Migration should return intermediate state whenever it requests new txn or detects cancelled ctx.

Very helpful, thank you! Can we add this as a comment above the `Migrate` method in the `Migration` interface?

> One thing to note is that MigrationFunc doesn't support currently intermediate steps. I saw the discord user being stuck at calculateBlockCommitments for hours which resulted in abortion and syncing from snapshot instead. Maybe in other pr they could be transformed to support it or even this one?

Good point. It might be easier to remove the stateless migrations entirely. What do you think? Regardless, I think the scope of the PR is large enough. Additional changes should go in issues for follow-up.

migration/migration.go
Exca-DK: Does the switch match the same way as `errors.Is` ? There are some migrations that return wrapped errors, eg. `db.CloseAndWrapOnError(iterator.Close, ErrCallWithNewTransaction)`

Exca-DK: @joshklop What do you think? I don't know if the behavior change is intentional :sweat_smile:

omerfirmak:
We will remove all the existing migrations once mainnet goes thru regenesis, I am not sure if it is worth it to try to make existing migrations cancellable

joshklop:
> I am not sure if it is worth it to try to make existing migrations cancellable

Fair. We can focus on making future migrations cancellable (i.e., moving forward with the changes to the `Migration` interface in this PR), while leaving existing migrations untouched. What do you think?

migration/migration.go
joshklop: I haven't looked into this very deeply, but if there is a behavior change, then one of the implementations should cause a test to fail. Right now, both implementations pass all tests, which indicates we may be missing some edge cases in tests. Please correct me if I'm missing something.

omerfirmak:
> > I am not sure if it is worth it to try to make existing migrations cancellable
> 
> Fair. We can focus on making future migrations cancellable (i.e., moving forward with the changes to the `Migration` interface in this PR), while leaving existing migrations untouched. What do you think?

Makes sense to me.

migration/migration.go
```diff
@@ -73,30 +89,41 @@ func MigrateIfNeeded(targetDB db.DB, network utils.Network, log utils.SimpleLogg
...
 		for {
 			var migrationErr error
 			if dbErr := targetDB.Update(func(txn db.Transaction) error {
-				migrationErr = migration.Migrate(txn, network)
-				if migrationErr != nil {
-					if errors.Is(migrationErr, ErrCallWithNewTransaction) {
-						return nil // Run the migration again with a new transaction.
-					}
+				metadata.IntermediateState, migrationErr = migration.Migrate(ctx, txn, network)
+				switch migrationErr {
```
joshklop, line 111: Let's do proper error unwrapping as you suggested (e.g., using `errors.Is`).

migration/migration.go
```diff
@@ -16,23 +17,38 @@ import (
...
+	"github.com/fxamacker/cbor/v2"
 	"github.com/sourcegraph/conc/pool"
 )
 
+type schemaMetadata struct {
+	Version           uint64
+	IntermediateState []byte
+}
+
+func (schema *schemaMetadata) fromBytes(data []byte) error {
```
joshklop, line 29: I think the idiomatic name for these methods would be `MarshalBinary` and `UnmarshalBinary`. Let's use those if you agree.

migration/migration.go
```diff
@@ -73,30 +89,41 @@ func MigrateIfNeeded(targetDB db.DB, network utils.Network, log utils.SimpleLogg
...
-	for i := version; i < uint64(len(migrations)); i++ {
+	for i := metadata.Version; i < uint64(len(migrations)); i++ {
+		if err := ctx.Err(); err != nil {
+			return err
+		}
 		log.Infow("Applying database migration", "stage", fmt.Sprintf("%d/%d", i+1, len(migrations)))
 		migration := migrations[i]
-		migration.Before()
+		err := migration.Before(metadata.IntermediateState)
+		if err != nil {
```
joshklop, line 104:
```suggestion
		if err := migration.Before(metadata.IntermediateState); err != nil {
```

migration/migration.go
```diff
@@ -73,30 +89,41 @@ func MigrateIfNeeded(targetDB db.DB, network utils.Network, log utils.SimpleLogg
...
-					if errors.Is(migrationErr, ErrCallWithNewTransaction) {
-						return nil // Run the migration again with a new transaction.
-					}
+				metadata.IntermediateState, migrationErr = migration.Migrate(ctx, txn, network)
+				switch migrationErr {
+				case ctx.Err():
+				case nil:
+				case ErrCallWithNewTransaction:
+					return nil // Run migration again with new transaction.
+				default:
```
joshklop, line 116: Now that I think about it, we can simplify this even further:
```go
if migrationErr != nil {
        if errors.Is(migrationErr, ErrCallWithNewTransaction) {
                return nil // Run migration again with new transaction.
        }
        return migrationErr
}
```
This is closer to what you had originally (with one less branch I think). Apologies for the back-and-forth.

migration/migration.go
```diff
@@ -110,18 +137,22 @@ func MigrateIfNeeded(targetDB db.DB, network utils.Network, log utils.SimpleLogg
...
+// SchemaMetadata retrieves metadata about a database schema from the given database.
+func SchemaMetadata(targetDB db.DB) (schemaMetadata, error) {
+	metadata := schemaMetadata{}
 	txn := targetDB.NewTransaction(false)
-	err := txn.Get(db.SchemaVersion.Key(), func(bytes []byte) error {
-		version = binary.BigEndian.Uint64(bytes)
+	err := txn.Get(db.SchemaVersion.Key(), func(b []byte) error {
+		err := metadata.fromBytes(b)
+		if err != nil {
+			metadata.Version = binary.BigEndian.Uint64(b)
```
joshklop, line 147: Can we assume that all unmarshal errors will exist because of our new way of representing the schema? I assume cbor can fail for other reasons.

Do we need a migration to update the schema representation? It may be easier to use a separate key in the database for `IntermediateState` than store all metadata in the same struct.

migration/migration.go
Exca-DK: The metadata won't be updated if ctx is canceled

joshklop: Ah you're right, never mind.

migration/migration.go
```diff
@@ -73,30 +86,35 @@ func MigrateIfNeeded(targetDB db.DB, network utils.Network, log utils.SimpleLogg
...
 				}
-
-				// Migration successful. Bump the version.
-				var versionBytes [8]byte
-				binary.BigEndian.PutUint64(versionBytes[:], i+1)
-				return txn.Set(db.SchemaVersion.Key(), versionBytes[:])
+				if metadata.IntermediateState == nil {
+					metadata.Version++
+				}
+				return updateSchemaMetadata(txn, metadata)
```
joshklop, line 117: Let's move this in the `switch` statement.

migration/migration.go
```diff
@@ -16,23 +17,35 @@ import (
 	"github.com/NethermindEth/juno/encoder"
 	"github.com/NethermindEth/juno/utils"
 	"github.com/bits-and-blooms/bitset"
+	"github.com/fxamacker/cbor/v2"
 	"github.com/sourcegraph/conc/pool"
 )
 
+var (
+	schemaVersionKey           = db.SchemaVersion.Key()
+	schemaIntermediateStateKey = db.SchemaVersion.Key([]byte{1})
```
joshklop, line 26: We should add the new key to `db/buckets.go`. Otherwise we may overwrite this on accident.

migration/migration.go
```diff
@@ -110,18 +128,46 @@ func MigrateIfNeeded(targetDB db.DB, network utils.Network, log utils.SimpleLogg
...
 	}
 
-	return version, db.CloseAndWrapOnError(txn.Discard, nil)
+	if err := txn.Get(schemaIntermediateStateKey, func(b []byte) error {
+		return cbor.Unmarshal(b, &metadata.IntermediateState)
+	}); err != nil && !errors.Is(err, db.ErrKeyNotFound) {
+		return metadata, db.CloseAndWrapOnError(txn.Discard, err)
+	}
+
+	return metadata, db.CloseAndWrapOnError(txn.Discard, nil)
```
joshklop, line 151: No need to `CloseAndWrapOnError` here.

joshklop (commented):
Almost there.

migration/migration_pkg_test.go
```diff
@@ -263,3 +264,74 @@ func TestMigrateTrieNodesFromBitsetToTrieKey(t *testing.T) {
...
+			return updateSchemaMetadata(txn, schemaMetadata{
+				Version:           version,
+				IntermediateState: intermediateState,
+			})
+		}))
+		metadata, err := SchemaMetadata(testDB)
+		require.NoError(t, err)
+		require.Equal(t, version, metadata.Version)
+		require.Equal(t, intermediateState, metadata.IntermediateState)
+		require.True(t, metadata.IntermediateState != nil)
```
joshklop, line 312: Do we need this check?

migration/migration_pkg_test.go
```diff
@@ -263,3 +264,74 @@ func TestMigrateTrieNodesFromBitsetToTrieKey(t *testing.T) {
...
+			return updateSchemaMetadata(txn, schemaMetadata{
+				Version:           version,
+				IntermediateState: intermediateState,
+			})
+		}))
+		metadata, err := SchemaMetadata(testDB)
+		require.NoError(t, err)
+		require.Equal(t, version, metadata.Version)
+		require.Equal(t, intermediateState, metadata.IntermediateState)
+		require.True(t, metadata.IntermediateState == nil)
```
joshklop, line 289:
```suggestion
		require.Nil(t, metadata.IntermediateState)
```

migration/migration_pkg_test.go
```diff
@@ -263,3 +264,74 @@ func TestMigrateTrieNodesFromBitsetToTrieKey(t *testing.T) {
...
+
+func TestSchemaMetadataUpdate(t *testing.T) {
+	t.Run("Intermediate nil", func(t *testing.T) {
+		testDB := pebble.NewMemTest()
+		t.Cleanup(func() {
+			require.NoError(t, testDB.Close())
+		})
+
+		var (
+			intermediateState []byte
```
joshklop, line 276: Can we remove this variable?

migration/migration_pkg_test.go
```diff
@@ -263,3 +264,74 @@ func TestMigrateTrieNodesFromBitsetToTrieKey(t *testing.T) {
...
+			return updateSchemaMetadata(txn, schemaMetadata{
+				Version:           version,
+				IntermediateState: intermediateState,
+			})
+		}))
+		metadata, err := SchemaMetadata(testDB)
+		require.NoError(t, err)
+		require.Equal(t, version, metadata.Version)
+		require.Equal(t, intermediateState, metadata.IntermediateState)
+		require.True(t, metadata.IntermediateState != nil)
```
joshklop, line 335: Same here. I think this check is redundant.

migration/migration_test.go
```diff
@@ -16,17 +17,19 @@ func TestMigrateIfNeeded(t *testing.T) {
...
+		require.NoError(t, migration.MigrateIfNeeded(context.Background(), testDB, utils.MAINNET, utils.NewNopZapLogger()))
 	})
 
-	version, err := migration.SchemaVersion(testDB)
+	var nilIntermediateState []byte
+	meta, err := migration.SchemaMetadata(testDB)
 	require.NoError(t, err)
-	require.NotEqual(t, 0, version)
+	require.NotEqual(t, uint64(0), meta.Version)
+	require.Equal(t, nilIntermediateState, meta.IntermediateState)
```
joshklop, line 27:
```suggestion
	require.Nil(t, meta.IntermediateState)
```
And then we can remove the `nilIntermediateState` variable.

joshklop (approved):
Looks good! Just a few minor adjustments in the tests and we'll be good to merge.

joshklop:
Just realized that codecov is a little low. There are a few places where we could pick up some additional coverage:

*  `node.go`: lines 221-225
* `migrations.SchemaMetadata`: lines 137-138, line 143
* `migrations.MigrateIfNeeded`: lines 91-92, (maybe) lines 96-97

Exca-DK:
```go
dbErr != nil {
    return dbErr
} else if migrationErr == nil {
    break
} else if !errors.Is(migrationErr, ErrCallWithNewTransaction) {
    return migrationErr
}
```
Isn't there a bug since the beginning? `errors.Is(migrationErr, ErrCallWithNewTransaction) ` will never be hit since now and before this pr, with the condition `migrationErr != ErrCallWithNewTransaction && migrationErr != nil` is always returned as is.  This goes into 
```go
// Update : see db.DB.Update
func (d *DB) Update(fn func(txn db.Transaction) error) error {
	txn := d.NewTransaction(true)
	if err := fn(txn); err != nil {
		return utils.RunAndWrapOnError(txn.Discard, err)
	}
	return utils.RunAndWrapOnError(txn.Discard, txn.Commit())
}
```
which ultimately falls into `dbErr != nil` case even though it's the migrationErr, not tx/db.

node/node_test.go
```diff
@@ -56,6 +57,23 @@ func TestNewNode(t *testing.T) {
 		P2PBootPeers:        "",
 	}
 
-	_, err := node.New(config, "v0.3")
+	n, err := node.New(config, "v0.3")
 	require.NoError(t, err)
+
+	ctx, cancel := context.WithCancel(context.Background())
+	cancel()
+
```
joshklop, line 65: Are the changes to this file related to the PR?

joshklop:
> Isn't there a bug since the beginning?

I don't think it's a bug, but definitely worth a refactor.

migration/migration_pkg_test.go
```diff
@@ -263,3 +268,208 @@ func TestMigrateTrieNodesFromBitsetToTrieKey(t *testing.T) {
 	require.Equal(t, felt.Zero, trieNode.Left.Felt())
 	require.Equal(t, felt.Zero, trieNode.Right.Felt())
 }
+
+func TestSchemaMetadata(t *testing.T) {
+	t.Run("failure", func(t *testing.T) {
+		mockCtrl := gomock.NewController(t)
+		t.Cleanup(mockCtrl.Finish)
+		t.Run("at version", func(t *testing.T) {
```
joshklop, line 276: We generally don't test for db errors. Are the new db mocks necessary to increase coverage? We can force-merge if high coverage is unattainable.

node/node_test.go
Exca-DK: it improves coverage of `node.go: lines 237-241` (before merge 221-225 as asked)

migration/migration_pkg_test.go
Exca-DK: Without mocks, it's impossible to cover requested cover in `migrations.SchemaMetadata: lines 137-138, line 143`

Exca-DK:
What's the consensus on that? Should the mocks and tests related to that be removed? As I previously commented, I added them only to increase the requested coverage.

migration/migration_pkg_test.go
joshklop: > Without mocks, it's impossible to cover 

Can we call `SchemaMetadata` on an empty db to cover 138-139? A call to `SchemaMetadata` with only the `db.SchemaVersion` key present should cover 145.

Does that sound good? That should allow us to remove the mocks.

node/node_test.go
Exca-DK: Sure

migration/migration_pkg_test.go
```diff
@@ -248,3 +251,144 @@ func TestMigrateTrieNodesFromBitsetToTrieKey(t *testing.T) {
...
+	before func([]byte) error
+}
+
+func (f testMigration) Migrate(ctx context.Context, txn db.Transaction, network utils.Network) ([]byte, error) {
+	return f.exec(ctx, txn, network)
+}
+
+func (f testMigration) Before(state []byte) error { return f.before(state) }
+
+func TestMigrateIfNeededInternall(t *testing.T) {
```
joshklop, line 344: nit: `TestMigrateIfNeededInternal`

joshklop (approved):

joshklop:
I think we can merge this after a rebase. Thanks!

joshklop:
@Exca-DK Mind rebasing or [squashing](https://stackoverflow.com/questions/30136558/how-to-squash-commits-which-have-merge-commit-in-between) these commits?

Squashing seems a lot easier. I am happy to do it and will list you as a co-author on the commit, but want to give you the opportunity to claim sole authorship.

Exca-DK:
Sure, I will

Issue #1147, opened 2023-08-23
Common adapter for feeder, gateway and rpc


rianhughes:
https://github.com/NethermindEth/juno/issues/1015

Crossing layers in the clean architecture requires adapter functions: input from rpc to core types should have its own adapter and input from feeder and gateway also should have their own adapters to core types (and vice versa). However, due to the commonality between all of the adapters and to reduce duplication, we created feeder2core. These functions can be moved to the utils package with a more generic name.

Further, generalise the feeder2core adapter by renaming it and moving it to the utils package.

joshklop:
Agree that `feeder` and `gateway` should be merged, and that the renaming to `sequencer2core` makes sense.

Is the move to `utils` necessary? imho that makes the package seem more general than it actually is.

PR #1171, opened 2023-08-25
Rianhughes/common feeder2core

rianhughes:
closes https://github.com/NethermindEth/juno/issues/1147

This PR does two things:
1. rename feeder2core to sequencer2core, and move it to utils
2. move network.go from utils to core (to prevent import cycles)


joshklop:
Left a comment on the issue.

PR #1179, opened 2023-08-28
Fix grpc server

joshklop:
Use grpc.Serve, not http.ListenAndServe.

### DIFF
--- a/node/http.go
+++ b/node/http.go
@@ -88,7 +88,7 @@ type grpcService struct {
 }
 
 func (g *grpcService) Run(ctx context.Context) error {
-	errCh := make(chan error, 1)
+	errCh := make(chan error)
 	defer close(errCh)
 
 	portStr := strconv.FormatUint(uint64(g.port), 10)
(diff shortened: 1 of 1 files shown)


joshklop:
Hmm, appears the grpc server hangs on shutdown. Will look into it.

kirugan (approved):

kirugan:
> Hmm, appears the grpc server hangs on shutdown. Will look into it.

Josh the reason was: even if you stop grpc server it returns nil from `Serve(...)` and then goroutines block on write to `errCh`. I added bufferization and now it works.

P.S. Still not sure about `ErrServerStopped` though - in grpc source code it's said that `// Serve called after Stop or GracefulStop.` for this error.

joshklop:
Thanks! I replaced the check for `http.ErrServerStopped` with `err != nil` and made the channel unbuffered. Mind taking another look real quick?

node/http.go
```diff
@@ -81,10 +82,46 @@ func makeMetrics(port uint16) *httpService {
...
 }
 
-func makeGRPC(port uint16, database db.DB, version string) *httpService {
-	grpcHandler := grpc.NewServer()
-	gen.RegisterKVServer(grpcHandler, junogrpc.New(database, version))
-	return makeHTTPService(port, grpcHandler)
+type grpcService struct {
+	srv  *grpc.Server
+	port uint16
+}
```
omerfirmak, line 88: this probably belongs to `grpc` package

Issue #1106, opened 2023-08-17
Upgrade to go 1.21


omerfirmak:
comes with nice generic functions in stdlib

joshklop:
Blocked by quic-go compatibility with go 1.21 ([ref](https://github.com/quic-go/quic-go/wiki/quic-go-and-Go-versions)).

I get this output on `go test` in the `p2p` package:

```
# github.com/quic-go/quic-go/internal/qtls
../../../.go/pkg/mod/github.com/quic-go/quic-go@v0.36.4/internal/qtls/go121.go:5:13: cannot use "The version of quic-go you're using can't be built on Go 1.21 yet. For more details, please see https://github.com/quic-go/quic-go/wiki/quic-go-and-Go-versions." (untyped string constant "The version of quic-go you're using can't be built on Go 1.21 yet. F...) as int value in variable declaration
FAIL	github.com/NethermindEth/juno/p2p [build failed]
```

omerfirmak:
They had support at their main branch, maybe if we bump lipp2p we get the new quic version as well.

PR #1188, opened 2023-08-30
Upgrade Go to 1.21

kirugan:
Issue #1106


IronGauntlets (approved):

.github/workflows/build-binaries.yml
```diff
@@ -40,7 +40,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v4
         with:
-          go-version: 1.20.3
+          go-version: 1.21
```
joshklop, line 43: Could give it the path to the `go.mod` so we don't need to change it ([ref](https://github.com/actions/setup-go#getting-go-version-from-the-gomod-file))?

kirugan: Added `go-version-file`, thanks 👍

.github/workflows/juno-lint.yml
```diff
@@ -16,10 +16,10 @@ jobs:
     steps:
       - uses: actions/setup-go@v3
         with:
-          go-version: '1.20.3'
+          go-version: '1.21'
```
joshklop, line 19: `go-version-file` here? Same for other workflow files.

kirugan: Added to all setup-go actions, thanks!

PR #1193, opened 2023-09-03
Avoid double copy for strings in vm

kirugan:
Functions `C.CString(...)` and `C.Bytes(...)` actually copy data where the former adds \0 at the end of string.
Here is small test how it works:
```
// #include <stdio.h>
// static void test(char* str) {
//.      printf("Call from C: '%s'\n", str);
//.      fflush(stdout);
// }
import "C"
func CopyExample() {
	data := []byte{'s', 't', 'r', 'i', 'n', 'g'}
	original := unsafe.String(unsafe.SliceData(data), len(data))

	str := C.CString(original)
	*str = 'c'
	data[2] = '!'
	C.test(str)

	fmt.Println("Original ", original)
}
```

### DIFF
--- a/vm/string.go
+++ b/vm/string.go
@@ -1,0 +1,9 @@
+package vm
+
+import "C"
+import "unsafe"
+
+func cstring(data []byte) *C.char {
+	str := unsafe.String(unsafe.SliceData(data), len(data))
+	return C.CString(str)
+}
--- a/vm/string.go
+++ b/vm/string.go
@@ -3,6 +3,8 @@ package vm
 import "C"
 import "unsafe"
 
+// cstring creates a null-terminated C string from the given byte slice.
+// the caller is responsible for freeing the underlying memory
 func cstring(data []byte) *C.char {
 	str := unsafe.String(unsafe.SliceData(data), len(data))
 	return C.CString(str)
--- a/vm/class.go
+++ b/vm/class.go
@@ -140,7 +140,7 @@ func Cairo0ClassHash(class *core.Cairo0Class) (*felt.Felt, error) {
 	if err != nil {
 		return nil, err
 	}
-	classJSONCStr := C.CString(string(classJSON))
+	classJSONCStr := cstring(classJSON)
 
 	var hash felt.Felt
 	hashBytes := hash.Bytes()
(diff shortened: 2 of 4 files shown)


vm/string.go
```diff
@@ -0,0 +1,11 @@
+package vm
+
+import "C"
+import "unsafe"
+
+// cstring creates a null-terminated C string from the given byte slice.
+// the caller is responsible for freeing the underlying memory
+func cstring(data []byte) *C.char {
```
kirugan, line 8: By the way, are UTF-8 strings important for us here? 🤔
cc @omerfirmak

joshklop: I believe so. See [here](https://eqlabs.github.io/pathfinder/blog/2023-06-17_mainnet_incident.html#root-cause-of-the-issue)

kirugan: I rechecked `C.CString` implementation:
```
const char *_cgoPREFIX_Cfunc_CString(struct __go_string s) {
	char *p = malloc(s.__length+1);
	if(p == NULL)
		runtime_throw("runtime: C malloc failed");
	memmove(p, s.__data, s.__length);
	p[s.__length] = 0;
	return p;
}
```
which seems fine for `unsafe.String(...)` trick.

Regarding UTF-8: in Rust if you create string with invalid UTF-8 symbol it will panic, but since json valid can contain only valid UTF-8 chars we're fine.

vm/string.go
```diff
@@ -0,0 +1,11 @@
...
+	str := unsafe.String(unsafe.SliceData(data), len(data))
```
omerfirmak, line 9: Should we do https://github.com/golang/go/blob/a03bdc3e6bea34abd5077205371e6fb9ef354481/src/runtime/string.go#L136C9-L136C39 instead?

because `len(data)` is not always equal to `len(str)` with multi-byte characters right?

kirugan: Actually `len(str)` equals `len([]byte(str))`, here is spec - https://go.dev/ref/spec#:~:text=a%20particular%20architecture.-,String%20types,-A%20string%20type

I double checked it with this test:
```
func TestString(t *testing.T) {
	str := "hello 👋"
	bytes := []byte(str)

	var runesCount int
	for range str {
		runesCount++
	}

	fmt.Printf("Str len %d, bytes len %d, runes count %d\n", len(str), len(bytes), runesCount)
}
```

I know that is one of confusing parts of string type in Go :)

omerfirmak (approved):

PR #1201, opened 2023-09-05
Add debug build mode for VM

kirugan:
Local timings for `make all` (i.e. release) vs `make debug`:

make all  2228.00s user 115.56s system 749% cpu 5:12.53 total
make debug  1004.71s user 96.80s system 751% cpu 2:26.56 total

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,8 +6,13 @@ juno: vm ## compile
 	@mkdir -p build
 	@go build -a -ldflags="-X main.Version=$(shell git describe --tags)" -o build/juno ./cmd/juno/
 
+ifeq ($(VM_DEBUG),true)
+vm:
+	$(MAKE) -C vm/rust debug
+else
 vm:
 	$(MAKE) -C vm/rust all
+endif
 
 generate: ## generate
 	mkdir -p mocks
--- a/Makefile
+++ b/Makefile
@@ -25,20 +25,20 @@ clean-testcache:
 	go clean -testcache
 
 test: clean-testcache vm ## tests
-	go test ./...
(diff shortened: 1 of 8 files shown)


kirugan:
If everyone agree with additional command then I'll add support in Dockerfile

Makefile
```diff
@@ -6,8 +6,13 @@ juno: vm ## compile
 	@mkdir -p build
 	@go build -a -ldflags="-X main.Version=$(shell git describe --tags)" -o build/juno ./cmd/juno/
 
+ifeq ($(VM_DEBUG),true)
```
omerfirmak, line 9: should we make this a global swtich and build juno in debug mode as well?

vm/rust/Makefile
```diff
@@ -1,6 +1,11 @@
 all:
 	cargo build --release
 
+debug:
+	cargo build && \
+	mkdir -p target/release/ && \
+	mv -n target/debug/libjuno_starknet_rs.* target/release/
```
omerfirmak, line 7: this might be confusing to both user and the CI caches. Do you think we can move the switch to our linker flags instead? ie `vm.go`

Makefile
kirugan: Discussed in Slack: there are no debug/release builds for go

vm/rust/Makefile
kirugan: Thanks, at first I was sceptical that it's possible to control cgo using `#ifdef` C preprocessor directives, but turned out it supports build constrints in form of `#cgo some_build_constraint_condition ...`

omerfirmak (approved):

PR #1203, opened 2023-09-06
Use pebble branch crl-release-23.1

joshklop:
Previously, I did 
```
go get -u github.com/cockroachdb/pebble
```
which uses the main branch. The [right way](https://github.com/ethereum/go-ethereum/pull/27967/commits/2215c48c7b046e1ade4f3d77bcb63b150f3dce3c) to do it is using the `crl-release-<version>` branches like this:

```
go get -u github.com/cockroachdb/pebble@crl-release-23.1
```

This is blocking #1102


db/pebble/transaction.go
```diff
@@ -123,17 +123,13 @@ func (t *Transaction) Impl() any {
 // NewIterator : see db.Transaction.NewIterator
 func (t *Transaction) NewIterator() (db.Iterator, error) {
 	var iter *pebble.Iterator
-	var err error
 	if t.batch != nil {
-		iter, err = t.batch.NewIter(nil)
+		iter = t.batch.NewIter(nil)
 	} else if t.snapshot != nil {
-		iter, err = t.snapshot.NewIter(nil)
+		iter = t.snapshot.NewIter(nil)
```
kirugan, line 129: what happened with prev error ? 🤔 (i.e. do we need to call some new method to check for it)

joshklop: The error was added after we manually upgraded on its `main` branch. Apparently that behavior wasn't backported to the `crl-release-23.1` branch.

IronGauntlets (approved):

Issue #1198, opened 2023-09-05
Split up the `utils` package


joshklop:
For example, we can have small `network` and `log` packages.

Two benefits:

- Clarity. Better to avoid ambiguous package names like `common`, `utils`, `pkg`, etc.
- Dependency management. A user may depend on a package that only uses `utils/log` and `utils/network`, but that user will have the full `utils` package compiled into the binary.

joshklop:
Note: would like to avoid implementing this until @IronGauntlets (and anyone else who has an opinion) has a say.

rianhughes:
I would also be in favour of this, and think it's also relevant for the final point in the [refactoring PR](https://github.com/NethermindEth/juno/issues/1015) "Extract common types from the feeder package and rpc package to the utils package to reduce duplication."

IronGauntlets:
@joshklop I am happy to split the utils package into multiple sub-packages. It might be better to have them as a separate package at the root level. I don't have a strong opinion on either one

thiagodeev:
Done. The team decided to keep having the utils pkg, however, a refactor was made to clean it up.
https://github.com/NethermindEth/juno/pull/3566
https://github.com/NethermindEth/juno/pull/3567
https://github.com/NethermindEth/juno/pull/3569

PR #1209, opened 2023-09-08
Slog based logging

### LINKED ISSUE #1164: Use golang's `slog` package
Go 1.21.0 introduced [`slog`](https://pkg.go.dev/golang.org/x/exp/slog) which provides structured logging through the standard library, therefore, this should be used to remove Uber's logging package dependency.

Exca-DK:
closes #1164, part of #1198 

The target is 1:1 of old formatting. Not possible with standard slog handlers


joshklop:
Thank you for your effort there @Exca-DK. I don't think it's necessary to match the old format.

If we could get away with using `slog` defaults and remove everything but the interface (and possibly no-op implementation) from `utils/log.go`, that would be amazing.

Exca-DK:
If we want to go with defaults then we will also need to remove the colors flag since slog doesn't support it.

joshklop:
> If we want to go with defaults then we will also need to remove the colors flag since slog doesn't support it.

Thanks for pointing that out. It appears zap is [implementing](https://github.com/uber-go/zap/tree/master/exp/zapslog) its own slog handlers. Shall we continue using zap's nice features and let it handle the slog migration? cc @IronGauntlets

Exca-DK:
Closing since the issue is closed too.

PR #1214, opened 2023-09-11
Make RPC concurrency configurable

omerfirmak:


cmd/juno/juno.go
```diff
@@ -68,6 +69,7 @@ const (
 	defaultMetricsPort         = 9090
 	defaultGRPC                = false
 	defaultGRPCPort            = 6064
+	defaultRPCMaxConcurrency   = 1024
```
joshklop, line 72: Is `1024` better than `2 * runtime.GOMAXPROCS(0)`?

joshklop (approved):
Will users see benefits from tuning this new knob? My gut feeling is that we can set a sane default for all cases, but maybe I'm missing something.

cmd/juno/juno.go
omerfirmak: 1024 is practically unlimited concurrency. Users can decide to limit if they want to.

@wojciechos was seeing `broken pipe` errors in load tests. I am assuming those were related to `2 * runtime.GOMAXPROCS(0)` being too low

omerfirmak:
> Will users see benefits from tuning this new knob? My gut feeling is that we can set a sane default for all cases, but maybe I'm missing something.

HW specs will vary quite a lot, I dont think we can find a good default that will be good for the entire range of HW that users will run Juno on.

PR #1215, opened 2023-09-12
Revert concurrency flag change

joshklop:


omerfirmak:
This is not only for batches. Concurrency limit is shared between all requests.

joshklop:
> This is not only for batches. Concurrency limit is shared between all requests.

Do you mean it is shared between all batch requests or all requests (batch or stand-alone)? Based on my understanding, it appears to be the former. I can adjust the docs to clarify that.

PR #1220, opened 2023-09-12
Give more time for flaky test

joshklop:


omerfirmak (approved):
I am not sure if it will help, since there is a retry loop already.

joshklop:
> I am not sure if it will help, since there is a retry loop already.

Good point. The retry loop is a little buggy though: we end up doing the test four times no matter what (which could lead to us failing on the fourth try). I'll remove the loop now that we're waiting longer.

joshklop:
When I remove the timeout altogether, the test sometimes stalls for 10 minutes (before it is timed out by the go tool itself). This PR clearly isn't a solution; I'll need to investigate further.

omerfirmak:
Originally timeout was longer, I lowered it when I added the retry loop. Lets fix the loop and see how it goes?

p2p/p2p_test.go
```diff
@@ -76,15 +76,20 @@ func TestService(t *testing.T) {
 		require.NoError(t, err)
 		t.Cleanup(closer)
 
-		for i := 0; i < 4; i++ {
+		maxRetries := 4
+	RetryLoop:
```
omerfirmak, line 80: do we need this label?

PR #1229, opened 2023-09-13
rpc: Return error on tx revert in SimulateTransactions

### LINKED ISSUE #1108: `SimulateTransactions` should fail if one of the txns gets reverted

Exca-DK:
closes #1108


omerfirmak:
Traces generated can be multiple MB, searching them with regex is not ideal. You can try do achieve the same result by adding a "hasRevertedTxns" bool in vm.Context and check that in rpc.

Closing this PR for now, feel free to comment on the issue for more clarifications and open another PR with the updated approach.

Exca-DK:
Thanks for the tip, I will follow up with another pr!

PR #1234, opened 2023-09-14
 Account for query bit when switching on txn version for hashing

omerfirmak:


core/transaction.go
```diff
@@ -70,20 +72,65 @@ type Transaction interface {
...
 
 const (
 	// Calculated at https://hur.st/bloomfilter/?n=1000&p=&m=8192&k=
 	// provides 1 in 51 possibility of false positives for approximately 1000 elements
 	eventsBloomLength    = 8192
 	eventsBloomHashFuncs = 6
+	queryBit             = 128
 )
 
+type TransactionVersion felt.Felt
```
joshklop, line 91: Should we add a note here that we need a migration if we change the underlying type of `TransactionVersion`?

omerfirmak: Will do but I think one should already be aware that all `core` types are used as storage objects and be careful when making any changes on them.

vm/transaction.go
```diff
@@ -23,7 +18,8 @@ func marshalTxn(txn core.Transaction) (json.RawMessage, error) {
 		return nil, err
 	}
 
-	t.Version = clearQueryBit(t.Version)
+	versionWithoutQueryBit := (*core.TransactionVersion)(t.Version).WithoutQueryBit()
+	t.Version = versionWithoutQueryBit.AsFelt()
```
joshklop, line 22: Can we do `txn.WithoutQueryBit` before we convert to the feeder type?

joshklop (approved):

vm/transaction.go
omerfirmak: Sadly no, we dont have access to `Version` field thru `core.Transaction` interface.
Alternative could be doing it in the switch below, but it is far from DRY.

PR #1236, opened 2023-09-14
Introduce the jsonrpc.Error interface

joshklop:
Allows for custom error types.

### DIFF
--- a/jsonrpc/http_test.go
+++ b/jsonrpc/http_test.go
@@ -17,7 +17,7 @@ import (
 func TestHTTP(t *testing.T) {
 	method := jsonrpc.Method{
 		Name: "echo",
-		Handler: func(msg string) (string, *jsonrpc.Error) {
+		Handler: func(msg string) (string, jsonrpc.Error) {
 			return msg, nil
 		},
 		Params: []jsonrpc.Parameter{{Name: "msg"}},
--- a/jsonrpc/server.go
+++ b/jsonrpc/server.go
@@ -29,7 +29,8 @@ const (
 var (
 	ErrInvalidID = errors.New("id should be a string or an integer")
 
-	contextInterface = reflect.TypeOf((*context.Context)(nil)).Elem()
+	contextInterface      = reflect.TypeOf((*context.Context)(nil)).Elem()
+	jsonrpcErrorInterface = reflect.TypeOf((*Error)(nil)).Elem()
 )
 
 type request struct {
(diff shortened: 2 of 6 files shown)


omerfirmak:
I will read thru the changes tomorrow but;

1) How do we make sure that the error marshals to spec complaint json?
2) What is the benefit, why do we need this?

joshklop:
> How do we make sure that the error marshals to spec complaint json?

We convert everything to a `*jsonrpc.StaticError` (which has the desired struct tags) in `jsonrpc.*Server.handleRequest`.

> What is the benefit, why do we need this?

So we don't need to modify global error variables to add the `Data` field.

omerfirmak:
> and can create custom types where necessary.

When would that be necessary?

joshklop:
> When would that be necessary?

When we need dynamic error messages (see second commit for examples).

omerfirmak:
> > When would that be necessary?
> 
> 
> 
> When we need dynamic error messages (see second commit for examples).

Error messages are not dynamically created by the interface tho, they are set at the time of instantiation. You are essentially using it the interface as simple getters.

Such an interface only makes sense  to me if the caller (ie jsonrpc.Server) benefits from being able to cast the interface to original concrete error type and makes use of its internal fields and the functions of the interface is generating value (error messages or data) from those fields. I am not sure I see a case where this would be required and be immediately useful for us.

I would love to hear what others think as well. cc: @Iron

joshklop:
Happy to hear what others think as well.

> Error messages are not dynamically created by the interface tho, they are set at the time of instantiation.

This is the problem that the PR attempts to solve. Because we instantiate global error variables, we cannot mutate the messages at runtime.

But there are some cases where the spec demands us to return dynamic error messages (e.g., [UnexpectedError](https://github.com/starkware-libs/starknet-specs/blob/cddef47e2f239c032f2b057217bcd2d1f176b4b0/api/starknet_write_api.json#L294)). The only way to do this is (while maintaining information like the error code) is with custom error types.

The `jsonrpc` package cannot assume what custom error types the handlers will return, but it needs to know the error's `message`, `code`, and `data`. It gets this information through the interface methods.

> You are essentially using it the interface as simple getters.

We could use something like:

```go
type ErrorData interface {
    Error() *jsonrpc.Error
}
```
which feels a little less like a getter, but achieves the same thing.

omerfirmak:
> Happy to hear what others think as well.
> 
> > I am not sure I see a case where this would be required and be immediately useful for us.
> 
> Right now, we instantiate global error variables, which precludes us from mutating messages when the spec [requires it](https://github.com/starkware-libs/starknet-specs/blob/cddef47e2f239c032f2b057217bcd2d1f176b4b0/api/starknet_write_api.json#L294).

We create local copies of them to be mutated when needed. Is there anything obviously wrong with this that I am missing?

joshklop:
> We create local copies of them to be mutated when needed.

We copy the pointer and mutate the underlying value. Even if we copied the value, it would be sub-optimal since `errors.Is` would fail.

omerfirmak:
> > We create local copies of them to be mutated when needed.
> 
> We copy the pointer and mutate the underlying value. Even if we copied the value, it would be sub-optimal since `errors.Is` would fail.

We copy the value here, no?

https://github.com/NethermindEth/juno/pull/1236/files#diff-1f724c5e433be46bfd8dc0962d1796f0dce2a5e22253fe82f69ab6a1eb8df278L1029-L1031

> Even if we copied the value, it would be sub-optimal since errors.Is would fail.

They are not generic `error` types tho, so `errors.Is` wont work on them anyways.

joshklop:
> We copy the value here, no?

Yes we do copy it there. [Here](https://github.com/NethermindEth/juno/blob/7a816ed03ca565b2a1b9e11a3d628a1d7a5393fb/rpc/handlers.go#L973) we do not. Even if we decide the interface isn't necessary, I can fix that in this PR :)

> They are not generic error types tho, so errors.Is wont work on them anyways.

Ah yes you're right. Maybe they should be though? For client-facing projects like starknet.go, being able to use `errors.Is` as usual would be nice.

PR #1237, opened 2023-09-14
Remove unused RPC error

joshklop:


omerfirmak:
Do you think it makes sense to return this for blocks that we cannot successfully run with `blockifier`? Old blocks for example

joshklop:
> Do you think it makes sense to return this for blocks that we cannot successfully run with blockifier? Old blocks for example

No. The spec requires us to indicate why the trace is not available in the `Data` field by setting it to `RECEIVED` or `REJECTED`. i.e., the trace is not available because the transaction is invalid or because the sequencer hasn't traced it yet.

The error shouldn't be part of the spec imho--I think it was mistakenly copied from the feeder, where it would make more sense in the context of caching traces.

omerfirmak (approved):

PR #1240, opened 2023-09-15
Make Synchronizer fields private

omerfirmak:


rpc/handlers.go
```diff
@@ -579,19 +579,19 @@ func (h *Handler) StateUpdate(id BlockID) (*StateUpdate, *jsonrpc.Error) {
 func (h *Handler) Syncing() (*Sync, *jsonrpc.Error) {
 	defaultSyncState := &Sync{Syncing: new(bool)}
 
-	startingBlockNumber := h.synchronizer.StartingBlockNumber
-	if startingBlockNumber == nil {
+	startingBlockNumber, err := h.syncReader.StartingBlockNumber()
+	if err != nil {
```
kirugan, line 583: don't we need to log this error?

kirugan (approved):

rpc/handlers.go
omerfirmak: Not really, it is a static error that we expect sometimes.

PR #1241, opened 2023-09-15
Store empty pending block instead of deleting the key

joshklop:
To decrease the number of `block not found` errors when simulating transactions against the pending block.

### DIFF
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -339,7 +339,7 @@ func (b *Blockchain) Store(block *core.Block, blockCommitments *core.BlockCommit
 			return err
 		}
 
-		if err := txn.Delete(db.Pending.Key()); err != nil {
+		if err := storeEmptyPending(txn, block.Header); err != nil {
 			return err
 		}
 
@@ -822,14 +822,6 @@ func (b *Blockchain) revertHead(txn db.Transaction) error {
 			return err
 		}
 	}
-	if !genesisBlock {
-		var newHeader *core.Header
-		newHeader, err = blockHeaderByNumber(txn, blockNumber-1)
-		if err != nil {
-			return err
-		}
-		b.newHeads.Send(newHeader)
-	}
 
 	if err = removeTxsAndReceipts(txn, blockNumber, header.TransactionCount); err != nil {
 		return err
@@ -840,14 +832,24 @@ func (b *Blockchain) revertHead(txn db.Transaction) error {
 		return err
 	}
 
-	// remove pending
-	if err = txn.Delete(db.Pending.Key()); err != nil {
-		return err
+	// Revert chain height and pending.
+
+	if genesisBlock {
+		if err = txn.Delete(db.Pending.Key()); err != nil {
+			return err
+		}
+		return txn.Delete(db.ChainHeight.Key())
 	}
 
-	// update chain height
-	if genesisBlock {
-		return txn.Delete(db.ChainHeight.Key())
+	var newHeader *core.Header
+	newHeader, err = blockHeaderByNumber(txn, blockNumber-1)
+	if err != nil {
+		return err
+	}
+	b.newHeads.Send(newHeader)
+
+	if err := storeEmptyPending(txn, newHeader); err != nil {
+		return err
 	}
 
 	heightBin := core.MarshalBlockNumber(blockNumber - 1)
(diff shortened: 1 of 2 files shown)


blockchain/blockchain.go
```diff
@@ -854,6 +859,55 @@ func (b *Blockchain) revertHead(txn db.Transaction) error {
...
+			ProtocolVersion:  supportedStarknetVersion.String(),
+			ExtraData:        nil,
+			EventsBloom:      nil,
+			GasPrice:         new(felt.Felt),
+		},
+		Transactions: make([]core.Transaction, 0),
+		Receipts:     make([]*core.TransactionReceipt, 0),
+	}
+	var err error
+	pendingBlock.Hash, _, err = core.BlockHash(pendingBlock, b.network, nil)
```
omerfirmak, line 882: pending block does not have hash

blockchain/blockchain.go
```diff
@@ -854,6 +859,55 @@ func (b *Blockchain) revertHead(txn db.Transaction) error {
 	return txn.Set(db.ChainHeight.Key(), heightBin)
 }
 
+func (b *Blockchain) storeEmptyPending(txn db.Transaction, latestHeader *core.Header) error {
+	pendingBlock := &core.Block{
+		Header: &core.Header{
+			Hash:             nil, // Set below.
+			ParentHash:       latestHeader.Hash,
+			Number:           latestHeader.Number + 1,
+			GlobalStateRoot:  latestHeader.GlobalStateRoot,
```
omerfirmak, line 868: pending block does not have these

blockchain/blockchain.go
```diff
@@ -854,6 +859,55 @@ func (b *Blockchain) revertHead(txn db.Transaction) error {
...
+			SequencerAddress: latestHeader.SequencerAddress,
+			TransactionCount: 0,
+			EventCount:       0,
+			Timestamp:        latestHeader.Timestamp + 1,
+			ProtocolVersion:  supportedStarknetVersion.String(),
```
omerfirmak, line 873: can we get this from the latest block?

blockchain/blockchain.go
```diff
@@ -854,6 +859,55 @@ func (b *Blockchain) revertHead(txn db.Transaction) error {
...
+			ExtraData:        nil,
+			EventsBloom:      nil,
+			GasPrice:         new(felt.Felt),
```
omerfirmak, line 876: `latestHeader.GasPrice` ?

blockchain/blockchain.go
```diff
@@ -854,6 +859,55 @@ func (b *Blockchain) revertHead(txn db.Transaction) error {
...
+			ParentHash:       latestHeader.Hash,
+			Number:           latestHeader.Number + 1,
+			GlobalStateRoot:  latestHeader.GlobalStateRoot,
+			SequencerAddress: latestHeader.SequencerAddress,
+			TransactionCount: 0,
+			EventCount:       0,
+			Timestamp:        latestHeader.Timestamp + 1,
+			ProtocolVersion:  supportedStarknetVersion.String(),
+			ExtraData:        nil,
+			EventsBloom:      nil,
```
omerfirmak, line 875: let's create an empty bloom for this?

blockchain/blockchain.go
```diff
@@ -854,6 +859,55 @@ func (b *Blockchain) revertHead(txn db.Transaction) error {
...
+				ReplacedClasses:   make([]core.ReplacedClass, 0),
+			},
+		},
+		NewClasses: make(map[felt.Felt]core.Class, 0),
+	}
+	emptyPendingBytes, err := encoder.Marshal(emptyPending)
+	if err != nil {
+		return err
+	}
+	return txn.Set(db.Pending.Key(), emptyPendingBytes)
```
omerfirmak, line 908: can we use `StorePending` here?

joshklop:
While Juno is syncing:

* A user can simulate a transaction against the pending state and potentially get a confusing result 
* `starknet_getBlockWithTxs` with the `pending` param will return the empty pending block

Are these behavior changes acceptable?

omerfirmak:
> * `starknet_getBlockWithTxs` with the `pending` param will return the empty pending block

I think that is fine

> * A user can simulate a transaction against the pending state and potentially get a confusing result

This I am not sure of

blockchain/blockchain.go
joshklop: We could, but it will perform some useless verification. Is that acceptable?

omerfirmak: > but it will perform some useless verification

Let's think of them as sanity checks :D

joshklop: Actually, we can't use `StorePending` as-is since it creates a separate transaction.

We can create function `storePending` that is called by `storeEmptyPending` and `StorePending`. Is that worth it? I don't mind either way.

omerfirmak: Let's create a `storePending(txn)` if you dont mind

blockchain/blockchain.go
```diff
@@ -881,36 +883,71 @@ func removeTxsAndReceipts(txn db.Transaction, blockNumber, numTxs uint64) error
 	return nil
 }
 
+func storeEmptyPending(txn db.Transaction, latestHeader *core.Header) error {
+	receipts := make([]*core.TransactionReceipt, 0)
+	pendingBlock := &core.Block{
+		Header: &core.Header{
+			ParentHash:       latestHeader.Hash,
+			SequencerAddress: latestHeader.SequencerAddress,
+			Timestamp:        latestHeader.Timestamp + 1,
```
omerfirmak, line 892: I am having second thoughts about this approach.

Normally, pending block timestamp never changes. But people may observe a change in the timestamp when we replace the empty placeholder with the actual state of the pending block from the sequencer.

The value of timestamp might affect execution of a transaction as well.

Same goes for GasPrice as well. If we return a fee estimate for a pending block request by using latest.GasPrice and the GasPrice increases drastically, the user txn might fail.

Do you think these are relevant concerns or am I overthinking it?

omerfirmak (approved):

PR #1242, opened 2023-09-15
Move tightly coupled metrics out of packages

omerfirmak:
If this makes sense, I will create separate metrics collectors in `metrics` package that implement EventListener interfaces of individual packages.


joshklop:
It makes sense to me. Instead of having monolithic `EventListener` interfaces in each package, could we have 

```
WithOn<Event>(func callback())
```

methods on the structs? This way users and tests don't need to pass a no-op, and have the option to configure a subset of the listeners.

omerfirmak:
> It makes sense to me. Instead of having monolithic `EventListener` interfaces in each package, could we have
> 
> ```
> WithOn<Event>(callback func()) // callback will have different signature as necessary
> ```
> 
> methods on the structs? This way users and tests don't need to pass a no-op, and have the option to configure a subset of the listeners.

I considered doing at the beginning, but opted for the interface at the end since per-event approach would make the caller code more verbose. 

I gave it a try to see how it would look like now, but stuff like below is making me doubt if it is worth it for the additional bit of flexibility. See the last commit.

```
server := jsonrpc.NewServer(1, utils.NewNopZapLogger()).WithValidator(validator.New()).WithOnNewRequest(listener.OnNewRequest).WithOnRequestHandled(listener.OnRequestHandled).WithOnRequestFailed(listener.OnRequestFailed)
```

omerfirmak:
Let me know what you think about the latest push @joshklop

joshklop:
> I considered doing at the beginning, but opted for the interface at the end since per-event approach would make the caller code more verbose.

It looks alright to me. We could put the `.With...` calls on separate lines. 

```go
server := jsonrpc.NewServer(1, utils.NewNopZapLogger())
        .WithValidator(validator.New())
        .WithOnNewRequest(listener.OnNewRequest)
        .WithOnRequestHandled(listener.OnRequestHandled)
        .WithOnRequestFailed(listener.OnRequestFailed)
```

It moves callbacks to the call site, rather than making the caller implement an interface away from the call site. Since the callbacks will probably be concise, I think this makes sense.

It feels similar to Go code that uses [functional options](https://abhinavg.net/2022/12/06/designing-go-libraries/#functional-options) for configuration.

> Let me know what you think about the latest push

What is the `SelectiveListener` for?

omerfirmak:
> > Let me know what you think about the latest push
> 
> What is the `SelectiveListener` for?

It allows caller to selectively listen to events by assigning callbacks for each of them separately.

joshklop:
> Something like SelectiveListener + EventListener interface, gives user the flexibility to use standalone callbacks, implement the interface however they see fit and keeps the overall clutter minimum IMO.

Personally, I feel we could achieve the same with the `With...` methods. We can attach the callbacks in `makeRPCMetrics(server *jsonrpc.Server)`, instead of having an intermediary `SelectiveListener` that fulfills the same purpose.

omerfirmak:
> Personally, I feel we could achieve the same with the `With...` methods

That is correct, until we start to add more and more metrics to packages. With each additional metrics, we will need to chain more `With...` calls, and create more clutter.

joshklop:
> until we start to add more and more metrics to packages. With each additional metric, we will need to chain more With... calls, and create more clutter.

Is the situation different for `SelectiveListener`?

omerfirmak:
Yes, there are no additional calls needed to be made. Just an assignment to function pointers in SelectiveListener

jsonrpc/http.go
```diff
@@ -3,30 +3,30 @@ package jsonrpc
...
-		}),
+		rpc:      rpc,
+		log:      log,
+		listener: &SelectiveListener{},
 	}
-	metrics.MustRegister(h.requests)
+	return h
+}
+
+// WithListener registers an NewRequestListener
```
joshklop, line 27: nit: `an` -> `a`

sync/event_listener.go
```diff
@@ -0,0 +1,25 @@
...
+
+import "time"
+
+type EventListener interface {
+	OnSyncStepDone(op string, blockNum uint64, took time.Duration)
+	OnReorg(blockNum uint64)
+}
+
+type SelectiveListener struct {
+	OnSyncStepDoneCb func(op string, blockNum uint64, took time.Duration)
```
joshklop, line 11: Can we have separate callbacks instead of having the `op` enum?

joshklop (approved):
I still have some lingering doubts about the ergonomics of this approach, but let's move forward and refactor later if necessary.

sync/event_listener.go
omerfirmak: We can, but we currently dont really have a use for it. Should be an easy refactor when we do.

PR #1246, opened 2023-09-18
Add signatures to core.Header

### LINKED ISSUE #1225: Add signatures to `core.Header`
and fetch them from the feeder

omerfirmak:
Fixes #1225


core/block.go
```diff
@@ -37,6 +37,8 @@ type Header struct {
 	EventsBloom *bloom.BloomFilter
 	// Amount of ETH charged per Gas spent
 	GasPrice *felt.Felt
+	// Sequencer signatures
+	Signatures [][]*felt.Felt
```
joshklop, line 41: Do we only need the most recent signatures in memory for serving p2p? Why do we need to store them for every block?

omerfirmak: That is true, syncing peer only needs to verify the most recent signature. But storing them only in-memory is a bit too risky and contradicting to what we have done so far, which is keeping everything in database.

> do we need to store them for every block?

We need to store them for all non-L1-verified blocks. Because chain might get reorged at any point and any of those blocks could become the HEAD. 

Later on, if we end up needing to free some storage, we can remove the signatures of L1-verified blocks.

adapters/feeder2core/feeder2core.go
```diff
@@ -31,6 +31,10 @@ func AdaptBlock(response *feeder.Block) (*core.Block, error) {
 		eventCount += uint64(len(response.Receipts[i].Events))
 	}
 
+	sigs := [][]*felt.Felt{}
+	if sig != nil {
```
joshklop, line 35: Do we need this check, or is it only for old tests?

joshklop (approved):

adapters/feeder2core/feeder2core.go
omerfirmak: We need it, pending blocks dont have signatures.

PR #1248, opened 2023-09-18
Immediately acknowledge panicing services

omerfirmak:
```
and gracefully shutdown the node
```

we used to wait until the user shuts down the node with Ctrl+C


joshklop (approved):
Could add a comment above the `defer`?

PR #1249, opened 2023-09-19
Set default data directory correctly

joshklop:
Towards #1198


cmd/juno/juno.go
```diff
@@ -215,3 +238,19 @@ func NewCmd(config *node.Config, run func(*cobra.Command, []string) error) *cobr
...
+		return filepath.Join(userDataDir, junoDirName), nil
+	}
+
+	userHomeDir, err := os.UserHomeDir()
+	if err != nil {
+		return "", err
+	} else if userHomeDir == "" {
+		return "", errors.New("home directory not found")
+	}
+	return filepath.Join(userHomeDir, ".local", "share", junoDirName), nil
```
omerfirmak, line 255: would it make our lives easier if we used `cwd/juno` as the default datadir?

joshklop: I'd love to do that.

Main downside is that restarting Juno in a different working directory with the same params will not use the existing db. We follow the [XDG base directory](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html#Environment_variables) spec right now, so `XDG_DATA_HOME` should exist. Wdyt?

omerfirmak: > Main downside is that restarting Juno in a different working directory with the same params will not use the existing db

I wouldnt mind that. Using cwd sounds reasonable enough to me. cc: @IronGauntlets

IronGauntlets: Using `cwd` sounds good to me. Whenever I run juno I specify the `cwd` as `datadir`.

node/node.go
```diff
@@ -85,13 +84,6 @@ type Node struct {
 // New sets the config and logger to the StarknetNode.
 // Any errors while parsing the config on creating logger will be returned.
 func New(cfg *Config, version string) (*Node, error) { //nolint:gocyclo,funlen
-	if cfg.DatabasePath == "" {
-		dirPrefix, err := utils.DefaultDataDir()
-		if err != nil {
-			return nil, err
-		}
-		cfg.DatabasePath = filepath.Join(dirPrefix, cfg.Network.String())
-	}
```
IronGauntlets, line 94: Why not set the default db path here instead? if `pwd, err := os.Getwd()` fails then we can return the error.

joshklop: Then we print the incorrect default directory on `--help` and in the DEBUG log message. (also, it would be very nice if all config values are set before we enter `node.New`, rather than changing the defaults on the fly)

joshklop: We can call `os.Getwd()` in the `PreExec` function and return the error. That way `--help` succeeds, but running Juno will fail. I'll implement that.

joshklop: Done.

IronGauntlets (approved):

PR #1250, opened 2023-09-19
Discard database txn if user callback panics

omerfirmak:


db/pebble/db.go
```diff
@@ -109,3 +113,13 @@ func (d *DB) Update(fn func(txn db.Transaction) error) error {
 func (d *DB) Impl() any {
 	return d.pebble
 }
+
+func discardTxnOnPanic(txn db.Transaction) {
+	p := recover()
+	if p != nil {
+		if err := txn.Discard(); err != nil {
```
kirugan, line 120: What happens now when function triggers panic inside db layer? I guess it's going to be not commited anyway, right? Or it causes memory leak?

omerfirmak: `txn.Discard()` panicing is undefined behaviour. Node will try to shutdown gracefully but may corrupt the database.

db/pebble/db_test.go
```diff
@@ -391,3 +391,38 @@ func TestNext(t *testing.T) {
...
+			p := recover()
+			require.NotNil(t, p)
+		}()
+
+		require.NoError(t, testDB.View(func(txn db.Transaction) error {
+			panic("view")
+		}))
+	})
+
+	t.Run("update", func(t *testing.T) {
```
kirugan, line 412: I commented locally this line `defer discardTxnOnPanic(txn)` it the test pass successfully.
Is there a way to test added changes? I think we need somehow pass the same txn to testDB.View insdie defer closure 🤔

omerfirmak: `pebble.Close()` returns an error if you have an open read transaction on the database
but if you have a open write transaction, it doesnt for some reason. if you remove the deferred call from `View()` it should fail.

Let me see if I can better test the change on `Update()` as well.

omerfirmak: Updated the test.

kirugan (approved):

PR #1252, opened 2023-09-19
Enabled feeder client logs and timeout

omerfirmak:


clients/feeder/feeder.go
```diff
@@ -61,6 +61,11 @@ func (c *Client) WithUserAgent(ua string) *Client {
 	return c
 }
 
+func (c *Client) WithTimeout(t time.Duration) *Client {
+	c.client.Timeout = t
```
joshklop, line 65: This timeout includes the time spent reading the body. Is that something we want? ([reference](https://blog.cloudflare.com/the-complete-guide-to-golang-net-http-timeouts/)) Maybe `IdleConnTimeout` would be better?

clients/feeder/feeder.go
```diff
@@ -234,7 +239,7 @@ func (c *Client) get(ctx context.Context, queryURL string) (io.ReadCloser, error
 			if wait > c.maxWait {
 				wait = c.maxWait
 			}
-			c.log.Warnw("failed query to feeder, retrying...", "retryAfter", wait.String())
+			c.log.Debugw("failed query to feeder, retrying...", "retryAfter", wait.String(), "err", err)
```
joshklop, line 242: nit: while we're at it, can we capitalize `failed`?

joshklop (approved):

clients/feeder/feeder.go
omerfirmak: Hmm, I dont see any other configurable timeouts in `http.Client`? Do we need to override its `Transport` ?

joshklop: Yes. `net.Dialer.Timeout` (TCP connection setup timeout) and `http.Client.Timeout` are the only client timeouts not set on the `Transport`.

omerfirmak: Let's not rush doing that, 5secs is plenty of time to download a block and state update.

PR #1256, opened 2023-09-20
Only log L1 url at debug level

### LINKED ISSUE #1254: Only print L1 node url in debug and error logs
To avoid logging API keys by default. 

This will require a careful review of all of our INFO and WARN log messages to ensure we never print the L1 node url. It will also require logging the config at the DEBUG level instead of using `fmt.Printf`.

joshklop:
Fixes #1254

### DIFF
--- a/l1/l1.go
+++ b/l1/l1.go
@@ -69,7 +69,7 @@ func (c *Client) subscribeToUpdates(ctx context.Context, updateChan chan *contra
 			if err == nil {
 				return updateSub, nil
 			}
-			c.log.Warnw("Failed to subscribe to L1 state updates", "tryAgainIn", c.resubscribeDelay, "err", err)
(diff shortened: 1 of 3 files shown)


l1/l1.go
```diff
@@ -122,7 +124,9 @@ func (c *Client) Run(ctx context.Context) error {
 				select {
 				case err := <-updateSub.Err():
 					// TODO can we use geth's event.Resubscribe?
-					c.log.Warnw("L1 update subscription failed, resubscribing", "error", err)
+					// We can't use a warn log level here since we guarantee the L1 url will only be printed
+					// in debug logs and panics (to avoid leaking the API key).
+					c.log.Debugw("L1 update subscription failed, resubscribing", "error", err)
```
IronGauntlets, line 129: I think the repeated comment is overkill. Having it once is fine

IronGauntlets (approved):
Lgtm

PR #1258, opened 2023-09-20
Dont request entire block for only hash and number

joshklop:


sync/sync.go
```diff
@@ -33,15 +33,17 @@ const (
...
 }
 
 // Synchronizer manages a list of StarknetData to fetch the latest blockchain updates
 type Synchronizer struct {
 	blockchain          *blockchain.Blockchain
 	starknetData        starknetdata.StarknetData
 	startingBlockNumber *uint64
-	highestBlockHeader  atomic.Pointer[core.Header]
+	highestBlockNumber  atomic.Uint64
+	highestBlockHash    atomic.Pointer[felt.Felt]
```
omerfirmak, line 46: it feels like these should be together to ensure consistency

omerfirmak:
I feel like we should refrain from adding more methods to StarknetData that we cannot implement in p2p.

joshklop:
> I feel like we should refrain from adding more methods to StarknetData that we cannot implement in p2p.

Makes sense. Closing for now.

PR #1259, opened 2023-09-20
db/pebble: Extract Pebble's native metrics

### LINKED ISSUE #1243: Extract Pebble's native metrics
https://github.com/cockroachdb/pebble/blob/master/metrics.go

Exca-DK:
closes #1243


db/pebble/db.go
```diff
@@ -39,8 +40,13 @@ func New(path string, logger pebble.Logger) (db.DB, error) {
...
 		Name:      "write",
 	})
-	metrics.MustRegister(pDB.readCounter, pDB.writeCounter)
 
+	metrics.MustRegister(
+		pDB.readCounter,
+		pDB.writeCounter,
+	)
+
+	go pDB.meter(5 * time.Second)
```
Exca-DK, line 49: linter complains, should this be exposed as an option or nolint is fine here?

Exca-DK:
I see that the overall metrics were refactored. Any changes you would want to see before I start adapting it to the new structure? I could also split this PR into smaller ones, each representing part of the `pebble.metrics` if that's helpful.

omerfirmak:
> PR into smaller ones, each representing part of the `pebble.metrics` if that's helpful.

thanks, that would be great. separate commits are good enough, no need for multiple PRs.

Exca-DK:
@omerfirmak Would mind giving it a try? Just `LevelsListener` for now.

Exca-DK:
Any thoughts? I don't think that going forward with this pr is appropriate until the base approach is approved since each metric is just a slight variation of that.

Few thoughts:

- Pebble returns some of the metrics as total, I transform them into delta to use it in counter instead of gauge since only increasing gauge doesn't seem logical to me, eg.
``` go 
	// These metrics are only ever increasing, return delta instead of total.
	formatted.Level[i].BytesIn = levels[i].BytesIn - cache.Lvls[i].BytesIn
```

- Some interesting metrics that aren't included in `pebble.Metrics` can be collected through `pebble.EventListener`, eg compaction duration. 
``` go 
  // CompactionBegin is invoked after the inputs to a compaction have been
    // determined, but before the compaction has produced any output.
    CompactionBegin func(CompactionInfo)

    // CompactionEnd is invoked after a compaction has completed and the result
    // has been installed.
    CompactionEnd func(CompactionInfo)
```

- Most of the metrics don't seem useful, is it really worth [ingesting them all](https://github.com/NethermindEth/juno/issues/1243#issuecomment-1723284327) ? It could be way slimmer if we just collected the bare minimum, eg. [etherum for reference](https://github.com/ethereum/go-ethereum/blob/bc6d184872889224480cf9df58b0539b210ffa9e/ethdb/pebble/pebble.go#L477C2-L507C52).

Exca-DK:
Some small linting issues, I will resolve them later. Just figured out that I had different linter version installed and that one wasn't complaining about anything :smile:

omerfirmak:
I think most of the metrics pebble exposes are way too specific to make any sense for us. We definitely don't need metrics from all the levels.

Let's do what geth does and only report a subset if you dont mind.
https://github.com/ethereum/go-ethereum/blob/a8617c6d4dbe0df8c67e1f5c2ecd76c3200dc18d/ethdb/pebble/pebble.go#L58-L70C27

Exca-DK:
Should be ready for review @omerfirmak 

preview of metrics in dashboard during initial sync:
image

db/pebble/db_metrics.go
```diff
@@ -0,0 +1,110 @@
...
+
+	// Create storage for delta calculation.
+	var (
+		compTimes        [2]time.Duration
+		writeDelayCounts [2]uint64
+		writeDelayTimes  [2]time.Duration
+		compWrites       [2]uint64
+		compReads        [2]uint64
+		nWrites          [2]uint64
+	)
```
omerfirmak, line 55: I don't think you need to do this here, no? Delta calculation can be done on Grafana.

Exca-DK: I am not sure. The data provided by `Pebble` is the total cumulative so far, and that sounds to me like a case for `Counter` with delta instead of only ever-increasing `gauge`. For example, `compWrites` and `compReads` represent bytes written and read during compactions. I haven't personally seen a bytes-written metric represented as a `gauge` or, even worse, a `counter` that adds the total instead of the change. `go-ethereum` also reports the delta for that, but they use a slightly different collector called meter, which, from what I understand, accumulates data and resets itself after taking a snapshot.

I've been thinking about whether `memComps`, `lvl0Comps`, `nonLvl0Comps`, and `seekComps` shouldn't be represented as counters with a delta, too, for that reason. That would leave us with only three gauges in total: `DiskSize`, `LevelFiles`, and `ManualMemAlloc`. They represent the current value and can decrease or increase between snapshots.

Exca-DK:
I added changes that align with the behavior described in[comment](https://github.com/NethermindEth/juno/pull/1259#discussion_r1373828207).

Exca-DK:
Hey @joshklop, I noticed you added this to the milestone. Do you want me to resolve the conflicts since this PR has been idle for some time already?

joshklop:
> Hey @joshklop, I noticed you added this to the milestone. Do you want me to resolve the conflicts since this PR has been idle for some time already?

Hi @Exca-DK, that would be super helpful. Thanks for offering.

I think @omerfirmak had some opinions about how we should move forward on this PR, so I'll re-request a review from him.

omerfirmak:
> Hey @joshklop, I noticed you added this to the milestone. Do you want me to resolve the conflicts since this PR has been idle for some time already?

Please do, we will try to get this merged soon. Sorry it took us so long:(

db/pebble/db_metrics.go
omerfirmak: > The data provided by Pebble is the total cumulative so far

That is entirely fine, prometheus counters are expected to monotonically increase. They shouldn't be decremented. Let's just export whatever pebble is reporting, I don't think we need to do delta calculation ourselves.

Exca-DK: Sure, I exported them as they are. Would mind sharing some insights as to why the total? I am also not sure if I shouldn't change the counters into gauges with that change.

kirugan:
Closing in favour of https://github.com/NethermindEth/juno/pull/2187/files (similar changes).

PR #1266, opened 2023-09-21
Add pubsub package

joshklop:
This gets us most of the way to implementing websocket subscriptions. Once this is merged, I'll work on implementing the actual RPC handlers.

### DIFF
--- a/jsonrpc/server.go
+++ b/jsonrpc/server.go
@@ -195,6 +195,17 @@ func (s *Server) Handle(ctx context.Context, data []byte) ([]byte, error) {
 	return s.HandleReader(ctx, bytes.NewReader(data))
 }
 
+type ConnKey struct{}
+
+func ConnFromContext(ctx context.Context) (io.Writer, bool) {
+	conn := ctx.Value(ConnKey{})
+	if conn == nil {
+		return nil, false
+	}
+	w, ok := conn.(io.Writer)
+	return w, ok
+}
+
 // HandleReader processes a request to the server
 // It returns the response in a byte array, only returns an
 // error if it can not create the response byte array
--- a/jsonrpc/websocket.go
+++ b/jsonrpc/websocket.go
@@ -54,9 +54,20 @@ func (ws *Websocket) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	// TODO include connection information, such as the remote address, in the logs.
 
-	wsc := newWebsocketConn(conn, ws.rpc, ws.connParams, ws.listener)
+	wsc := newWebsocketConn(r.Context(), conn, ws.connParams, ws.listener)
 
-	err = wsc.ReadWriteLoop(r.Context())
+	ctx := context.WithValue(wsc.ctx, ConnKey{}, wsc)
+	for {
+		var resp []byte
+		resp, err = ws.rpc.HandleReader(ctx, wsc)
+		if err != nil {
+			break
+		}
+		wsc.listener.OnNewRequest("any")
+		if _, err = wsc.Write(resp); err != nil {
+			break
+		}
+	}
 
 	var errClose websocket.CloseError
 	if errors.As(err, &errClose) {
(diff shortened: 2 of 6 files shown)


jsonrpc/server.go
```diff
@@ -187,10 +187,25 @@ func (s *Server) registerMethod(method Method) error {
...
+		return err
+	}
+	return nil
+}
+
+// handle processes a request to the server
 // It returns the response in a byte array, only returns an
 // error if it can not create the response byte array
-func (s *Server) HandleReader(ctx context.Context, reader io.Reader) ([]byte, error) {
+func (s *Server) handle(ctx context.Context, reader io.Reader) ([]byte, error) {
```
omerfirmak, line 215: How about this,

We keep the old API. `Handle(ctx context.Context, reader io.Reader)`
And still pass `reader` as connection to the handler. In the subscription handler, we try to "promote" it to a `io.ReadWriter` by casting it. (Like WS server tries to promote a http connection to a ws connection.) If cast fails,  reject the request it was probably sent with HTTP.

This way, we dont need to update `jsonrpc/http.go` at all

jsonrpc/server.go
```diff
@@ -187,9 +187,16 @@ func (s *Server) registerMethod(method Method) error {
 	return nil
 }
 
+type connKey struct{}
```
omerfirmak, line 190: We get this context from http package, it might me carrying other values already. can we use something with more context, to avoid collusion?

"jsonrpc.ctx.connection" or something?

pubsub/server.go
```diff
@@ -0,0 +1,75 @@
...
+
+func getRandomID() uint64 {
+	var n uint64
+	for err := binary.Read(rand.Reader, binary.LittleEndian, &n); err != nil; {
+	}
+	return n
+}
+
+func (s *Server) Subscribe(ctx context.Context, sub event.Subscription) uint64 {
+	id := getRandomID()
```
omerfirmak, line 40: why random? could we get away with an incrementing counter?

pubsub/server.go
```diff
@@ -0,0 +1,75 @@
...
+
+var ErrNotFound = errors.New("subscription not found")
+
+type Server struct {
+	log utils.SimpleLogger
+	wg  conc.WaitGroup
+
+	mu            sync.Mutex // protects subscriptions
+	subscriptions map[uint64]event.Subscription
+}
```
omerfirmak, line 23: Can you elaborate on how you imagine this would fit with the rest of the design? This looks like a registry instead of a server to me.

pubsub/server.go
joshklop: I'm almost positive I read about some security issue related to sequential ids. I can't find it right now. Will leave this open for now.

pubsub/server.go
joshklop: Registry is a better name.

I imagine the `rpc.Handler` having a registry. It's handlers create subscriptions and call `h.registry.Subscribe` / `h.registry.Unsubscribe`. The handlers return the responses to initial subscription request (containing the id) and unsubscribe request.

jsonrpc/server.go
joshklop: Collision is impossible https://brabalawuka.cc/posts/study/empty_struct/

I can update the name if you think it's unclear, though.

omerfirmak: Wow, nice. TIL.
Thanks!

pubsub/server.go
omerfirmak: Why do we need a Registry?

Do you intend to emit geth events directly from `blockchain`?

joshklop: The registry keeps track of subscriptions so we can release them when we receive an unsubscribe request.

joshklop: > Do you intend to emit geth events directly from blockchain?

We can do that, or have callbacks. The registry should be agnostic to how we manage events.

jsonrpc/websocket.go
```diff
@@ -54,9 +54,19 @@ func (ws *Websocket) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	for {
+		// Handle will read from the connection once.
+		var resp []byte
+		resp, err = ws.rpc.HandleReader(wsc.ctx, wsc)
+		if err != nil {
+			break
+		}
+		if _, err = wsc.Write(resp); err != nil {
+			break
+		}
```
joshklop, line 68: We'll need to make sure subscriptions don't write a notification before we respond to the initial request.

pubsub/server.go
joshklop: I guess the only issue is leaking information about how many subscriptions have been served. Would you like it to be incrementing? I don't mind either way.

jsonrpc/websocket.go
```diff
@@ -54,9 +54,19 @@ func (ws *Websocket) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	}
```
omerfirmak, line 69: this loop will come back to bite us.

`HandleReader` uses a buffered reader, so it might over-read from the connection. Meaning that, while decoding a request we might consume a part of the second request and at the second iteration of the loop fail to decode the second request since the remaining bytes in the reader is now not a valid request.

omerfirmak: But this issue is not introduced in this PR, so lets just create an issue but we should fix it before fully rolling out WS

jsonrpc/websocket.go
```diff
@@ -95,52 +105,53 @@ func DefaultWebsocketConnParams() *WebsocketConnParams {
...
 	}
+	var n int
+	if len(wsc.buf) > len(p) {
+		n = copy(p, wsc.buf[:len(p)])
+		wsc.buf = wsc.buf[len(p):]
+	} else {
+		n = copy(p, wsc.buf)
+		wsc.buf = nil
+	}
+	wsc.listener.OnNewRequest("any")
```
omerfirmak, line 140: I am not sure if this is the right place for this, moving it to after `HandleReader()` in the loop above makes more sense to me.

pubsub/registry.go
```diff
@@ -0,0 +1,75 @@
...
+		case err := <-sub.Err():
+			if err != nil {
+				r.log.Warnw("Subscription failed", "err", err)
+			}
+		}
+		// Unsubscribe releases the subscription resources.
+		if err := r.Delete(id); err != nil {
+			r.log.Warnw("Failed to unsubscribe", "subscriptionID", id, "err", err)
+		}
+	})
```
omerfirmak, line 56: Spinning up a goroutine just for cleanup is a bit overkill imo.

jsonrpc/server.go
```diff
@@ -195,10 +195,21 @@ func (s *Server) Handle(ctx context.Context, data []byte) ([]byte, error) {
...
+	// store the connection in the context on every request.
+	conn, ok := ctx.Value(connKey{}).(io.ReadWriter)
+	return conn, ok
+}
+
 // HandleReader processes a request to the server
 // It returns the response in a byte array, only returns an
 // error if it can not create the response byte array
 func (s *Server) HandleReader(ctx context.Context, reader io.Reader) ([]byte, error) {
+	connCtx := context.WithValue(ctx, connKey{}, reader)
```
omerfirmak, line 211: I think it makes sense to move this to transport side (ie http, ws) That way we can fix #1271 pretty easily.

omerfirmak (approved):
Without the full implementation I am having hard time seeing how things will fit together, which makes optimizing the design harder for me as well. Let's merge this and iron out any details in the upcoming PRs.

PR #1269, opened 2023-09-22
Towards rpc v0.5.0 support

omerfirmak:


node/http.go
```diff
@@ -74,7 +84,7 @@ func makeRPCOverHTTP(host string, port uint16, servers map[string]*jsonrpc.Serve
 		if listener != nil {
 			httpHandler = httpHandler.WithListener(listener)
 		}
-		mux.Handle(path, httpHandler)
+		mux.Handle(path, exactPathServer(path, httpHandler))
```
kirugan, line 87: hm.... I always thought that `Handle(...)` use strict matching at least for `/path?query=` or you added it so it won't be possible to pass get parameters?

omerfirmak: > I always thought that Handle(...) use strict matching

sadly it doesnt, "/" matches every possible path.

omerfirmak: https://pkg.go.dev/net/http#ServeMux

> Note that since a pattern ending in a slash names a rooted subtree, the pattern "/" matches all paths not matched by other registered patterns, not just the URL with Path == "/".

rpc/handlers.go
```diff
@@ -151,6 +151,13 @@ func (h *Handler) BlockWithTxHashes(id BlockID) (*BlockWithTxHashes, *jsonrpc.Er
 	}, nil
 }
 
+func (h *Handler) LegacyBlockWithTxHashes(id BlockID) (*BlockWithTxHashes, *jsonrpc.Error) {
+	block, err := h.BlockWithTxHashes(id)
+	block.BlockHeader.L1GasPrice = nil
+	block.BlockHeader.StarknetVersion = ""
```
kirugan, line 157: Why do we need explicit assignments here? It's going to zero valued by default

kirugan (approved):

rpc/handlers.go
omerfirmak: they are set by the `h.BlockWithTxHashes`, we need to clear them as those fields are not present in RPC v0.4.0.

Issue #1228, opened 2023-09-13
RPC Trace API v0.5.0-rc1


omerfirmak:
- [x] make `starknet_traceBlockTransactions` get a block id [ref](https://github.com/starkware-libs/starknet-specs/compare/v0.4.0...v0.5.0-rc0#diff-3f1851e8de77613ed83333cbd5bcb76ad771821c6580c7668af2b1c4c87ec3eeR120)
   - [x] add txn type to traces [ref](https://github.com/starkware-libs/starknet-specs/compare/v0.4.0...v0.5.0-rc0#diff-3f1851e8de77613ed83333cbd5bcb76ad771821c6580c7668af2b1c4c87ec3eeR184-R189)
   - [x] Add new method: `starknet_specVersion`
   - [ ] Add [`message_hash`](https://github.com/starkware-libs/starknet-specs/blob/cddef47e2f239c032f2b057217bcd2d1f176b4b0/api/starknet_api_openrpc.json#L2782) field for `L1_HANDLER_TXN_RECEIPT`
   - [ ] https://github.com/starkware-libs/starknet-specs/pull/138
   - [ ] https://github.com/starkware-libs/starknet-specs/pull/140
   
this might not be an exhaustive list, please go thru https://github.com/starkware-libs/starknet-specs/compare/v0.4.0...v0.5.0-rc1# and add any items that I might have missed.

omerfirmak:
Closed in #1272

PR #1272, opened 2023-09-25
Add OrderedMessage and OrderedEvent to traces & txn state diffs

omerfirmak:


omerfirmak:
Closes #1228

vm/rust/src/lib.rs
```diff
@@ -223,10 +221,13 @@ pub extern "C" fn cairoVMExecute(
             return;
         }
 
+        let mut txn_state = CachedState::create_transactional(&mut state);
```
kirugan, line 224: Could you explain why do we need this?

vm/rust/src/jsonrpc.rs
```diff
@@ -32,7 +39,35 @@ pub struct TransactionTrace {
...
+                _ => {}
+            }
+        }
+        if self.fee_transfer_invocation.is_some() {
+            self.fee_transfer_invocation.as_mut().unwrap().make_legacy()
+        }
+        if self.constructor_invocation.is_some() {
+            self.constructor_invocation.as_mut().unwrap().make_legacy()
+        }
+        if self.function_invocation.is_some() {
```
kirugan, line 67: common rust idiom is to use if let in such cases i.e. 
```
if let Some(actualValue) = self.function_invocation {
   // do something with
}
```
but it's up to you

omerfirmak: I knew there was a better way of doing this:)

vm/rust/src/lib.rs
omerfirmak: to be able to generate per-txn state diffs. 

txn_state buffers the changes internally, until `commit()` is called. So we gather all the changes for a single txn, save them separately and then commit them to `state`.

kirugan (approved):

PR #1275, opened 2023-09-26
Fix over-reading websocket requests

joshklop:
The JSON decoder can read past the end of the JSON message and into the next request, causing decoding to fail when the next request is processed.

This commit ensures we only give the decoder a single message at a time.

### DIFF
--- a/jsonrpc/websocket.go
+++ b/jsonrpc/websocket.go
@@ -132,9 +132,8 @@ func (wsc *websocketConn) Write(p []byte) (int, error) {
 	writeCtx, writeCancel := context.WithTimeout(wsc.ctx, wsc.params.WriteDuration)
 	defer writeCancel()
 	// Use MessageText since JSON is a text format.
-	err := wsc.conn.Write(writeCtx, websocket.MessageText, p)
-	if err != nil {
+	if err := wsc.conn.Write(writeCtx, websocket.MessageText, p); err != nil {
 		return 0, err
 	}
-	return len(p), err
+	return len(p), nil
 }
--- a/jsonrpc/websocket.go
+++ b/jsonrpc/websocket.go
@@ -54,16 +54,21 @@ func (ws *Websocket) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	// TODO include connection information, such as the remote address, in the logs.
 
-	wsc := newWebsocketConn(r.Context(), conn, ws.connParams, ws.listener)
+	wsc := newWebsocketConn(r.Context(), conn, ws.connParams)
 
 	ctx := context.WithValue(wsc.ctx, ConnKey{}, wsc)
 	for {
-		var resp []byte
-		resp, err = ws.rpc.HandleReader(ctx, wsc)
+		var msg []byte
+		_, msg, err = wsc.conn.Read(ctx)
 		if err != nil {
 			break
 		}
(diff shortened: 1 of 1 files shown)


jsonrpc/websocket.go
```diff
@@ -54,16 +54,21 @@ func (ws *Websocket) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	// TODO include connection information, such as the remote address, in the logs.
 
-	wsc := newWebsocketConn(r.Context(), conn, ws.connParams, ws.listener)
+	wsc := newWebsocketConn(r.Context(), conn, ws.connParams)
 
 	ctx := context.WithValue(wsc.ctx, ConnKey{}, wsc)
 	for {
+		var msg []byte
+		_, msg, err = wsc.conn.Read(ctx)
```
omerfirmak, line 62: We will allocate a buffer here for each request, on top of the `io.Reader` and `bufio.Reader` allocations that server will do.

I am not sure if this is the most efficient approach.

omerfirmak: I dont think so. 

These approaches also break the bidirectional communication that we introduced in the last PR, no? But I think bidirectional communication between handler and client was a mistake.  Handler should only be able to write to the connection, it shouldnt be able to read from it. Reading should be only available to `jsonrpc.Server`.

Something like would work better IMO.

``` golang
bufferedWSConn := bufio.Reader(wsc.conn)
writerCtx := context.WithValue(key, wsc.conn)
for {
   resp, err := ws.rpc.HandleReader(ctx, reader) 
   if err != nil { break }
   .
   .
   .
}
```

joshklop: With the `bufio` approach, `wsc.conn` must be an `io.Reader`. What do you think the `wsc.conn.Read` method should look like?

> But I think bidirectional communication between handler and client was a mistake.

I agree. I changed the return type of `jsonrpc.ConnFromContext` from `io.ReadWriter` to `io.Writer` before merging (just realized I forgot to update the commit message, unfortunately).

omerfirmak: > . I changed the return type of jsonrpc.ConnFromContext from io.ReadWriter to io.Writer before merging
Nice, I didnt realize that.

> With the bufio approach, wsc.conn must be an io.Reader
Looks like we can just use this? https://pkg.go.dev/nhooyr.io/websocket#NetConn

joshklop: > Looks like we can just use this? https://pkg.go.dev/nhooyr.io/websocket#NetConn

Oh cool, I forgot about that. There is a bug in that function, so we can use Tailscale's workaround I think: https://github.com/tailscale/tailscale/pull/5983

omerfirmak:
This will break #1279, am I right?

joshklop:
> This will break #1279, am I right? Lets fold this in to #1279?

I'm more than willing to do so, but how will this break that?

omerfirmak:
> > This will break #1279, am I right? Lets fold this in to #1279?
> 
> I'm more than willing to do so, but how will this break that?

Because now we are using the `Handle` function, which creates its own `io.Reader` (ie a bytes.Buffer).
and `ConnFromContext` wont be able to cast `bytes.Buffer` to a `io.Writer`

omerfirmak (approved):

PR #1276, opened 2023-09-26
jsonrpc: Support up to 2 return values

### LINKED ISSUE #933: Allow for less than two return values in RPC handlers

Exca-DK:
closes #933

### DIFF
--- a/jsonrpc/server_test.go
+++ b/jsonrpc/server_test.go
@@ -503,11 +503,11 @@ func TestHandle(t *testing.T) {
 		},
 		"echo nil error": {
 			req: `{"jsonrpc": "2.0", "method": "echoError", "params": ["nil"], "id": 1}`,
-			res: `{"jsonrpc":"2.0","id":1}`,
+			res: `{"jsonrpc":"2.0","result":null,"id":1}`,
 		},
 		"void": {
 			req: `{"jsonrpc": "2.0", "method": "void", "id": 1}`,
-			res: `{"jsonrpc":"2.0","id":1}`,
+			res: `{"jsonrpc":"2.0","result":null,"id":1}`,
 		},
 	}
 
--- a/jsonrpc/server_test.go
+++ b/jsonrpc/server_test.go
@@ -36,11 +36,6 @@ func TestServer_RegisterMethod(t *testing.T) {
 			paramNames: []jsonrpc.Parameter{{Name: "param1"}},
 			want:       "number of non-context function params and param names must match",
 		},
-		"no return": {
-			handler:    func(param1, param2 int) {},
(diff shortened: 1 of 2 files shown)


jsonrpc/server_test.go
```diff
@@ -453,6 +493,22 @@ func TestHandle(t *testing.T) {
...
+			req: `{"jsonrpc": "2.0", "method": "echoValue", "params": [10], "id": 1}`,
+			res: `{"jsonrpc":"2.0","result":10,"id":1}`,
+		},
+		"echo error": {
+			req: `{"jsonrpc": "2.0", "method": "echoError", "params": ["echo"], "id": 1}`,
+			res: `{"jsonrpc":"2.0","error":{"code":44,"message":"echo"},"id":1}`,
+		},
+		"echo nil error": {
+			req: `{"jsonrpc": "2.0", "method": "echoError", "params": ["nil"], "id": 1}`,
+			res: `{"jsonrpc":"2.0","id":1}`,
```
joshklop, line 506: I believe `result` is a required member of the response. ([spec](https://www.jsonrpc.org/specification#response_object))

Exca-DK: You are right, [result should be null](https://github.com/ethereum/go-ethereum/blob/master/rpc/testdata/reqresp-noargsrets.js)

omerfirmak:
I am not sure I am following, what is the point?

joshklop:
> I am not sure I am following, what is the benefit?

Described here https://github.com/NethermindEth/juno/issues/933

omerfirmak:
> > I am not sure I am following, what is the benefit?
> 
> Described here #933

"if it ain't broke, don't fix it" - someone wise, probably

Unless we have an immediate use for the feature or cleaning up technical dept that is making our lives harder; I don't really see the point in introducing things like this.

omerfirmak:
> > > I am not sure I am following, what is the benefit?
> > 
> > 
> > Described here #933
> 
> "if it ain't broke, don't fix it" - someone wise, probably
> 
> Unless we have an immediate use for the feature or cleaning up technical dept that is making our lives harder; I don't really see the point in introducing things like this.

Because at the end of the day, we will end up maintaining it for little-to-no benefit on our account.

"best code is no code at all" - someone even wiser

Exca-DK:
If this is no longer wanted the issue and pr can be closed :sweat_smile:

PR #1278, opened 2023-09-26
Use t.Cleanup in pebble.NewMemTest

joshklop:
There are cases where we aren't closing the in-memory db. In some cases, this causes fatal bugs in our tests that we aren't catching.

### DIFF
--- a/db/pebble/db.go
+++ b/db/pebble/db.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"sync"
+	"testing"
 
 	"github.com/NethermindEth/juno/db"
 	"github.com/NethermindEth/juno/utils"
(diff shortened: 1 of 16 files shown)


db/pebble/db.go
```diff
@@ -38,11 +39,16 @@ func NewMem() (db.DB, error) {
 }
 
 // NewMemTest opens a new in-memory database, panics on error
-func NewMemTest() db.DB {
+func NewMemTest(t *testing.T) db.DB {
```
joshklop, line 42: This is the key change.

omerfirmak (approved):
cool

db/pebble/db.go
IronGauntlets: I think you should also add `t.Helper()` as well.

PR #1279, opened 2023-09-27
Implement subscribe new heads

joshklop:


rpc/handlers.go
```diff
@@ -1343,6 +1348,56 @@ func (h *Handler) SpecVersion() (string, *jsonrpc.Error) {
...
+		return 0, jsonrpc.Err(jsonrpc.MethodNotFound, nil)
+	}
+	var id uint64
+	id = h.registry.Add(ctx, event.NewSubscription(func(quit <-chan struct{}) error {
+		defer func() {
+			_ = h.registry.Delete(id)
+		}()
+		hexID := "0x" + strconv.FormatUint(id, 16)
+
+		const bufferSize = 2048
```
omerfirmak, line 1365: this is way too big, do we need a buffered channel?

joshklop: I agree that's too large. It's good to have a small buffer so we don't drop responses if one takes a while to send.

rpc/handlers.go
```diff
@@ -1343,6 +1348,56 @@ func (h *Handler) SpecVersion() (string, *jsonrpc.Error) {
...
+		sink := make(chan *core.Header, bufferSize)
+		sub := h.bcReader.SubscribeNewHeads(sink)
```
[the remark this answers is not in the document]
omerfirmak: I think we can simplify the way we maintain our open subscriptions.

Let me briefly, explain and we can discuss it further.

1) We can remove `pubsub.Registry`, registering `event.Subscription` instances to a registry doesnt really benefit us in an important way imo.
2) `SubscribeNewHeads` could be implemented this way.

 ``` golang
 func (h *Handler) SubscribeNewHeads(ctx context.Context) (ID, *jsonrpc.Error) {
 	w, ok := jsonrpc.ConnFromContext(ctx)
	if !ok {
		return 0, jsonrpc.Err(jsonrpc.MethodNotFound, nil)
	}
	
      id := randomID();
      subscriptionCtx, subscriptionCtxCancel := context.WithCancel(ctx)
       h.subscriptionCanceler[id] = subscriptionCtxCancel

        go func() {
              // handle subscription until subscriptionCtx is canceled
        }()
       return id, nil
}
 
  func (h *Handler) UnsubscribeNewHeads(ctx context.Context, id ID) (ID, *jsonrpc.Error) {
 	canceler, ok := h.subscriptionCanceler[id]
        if !ok {
		return 0, IdNotFound
	}
       h.subscriptionCanceler.delete(id)
       canceler()
       return id, nil
 }
 ```
 
 `h.subscriptionCanceler` should be protected with a mutex, but this should work for the most part.

rpc/handlers.go
omerfirmak: > I made it buffered so we don't drop responses if one takes a while to send

Why would we drop responses?

rpc/handlers.go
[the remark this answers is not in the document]
joshklop: I like that a lot.

Do you think we could save the (reference to) the connection in the map along with the canceler? Then we can enforce that connections only close their own subscriptions?

rpc/handlers.go
joshklop: We won't, my bad :man_facepalming:

joshklop: There is this in the `event.*FeedOf.Subscribe` documentation:

> The channel should have ample buffer space to avoid blocking other subscribers.

But we can save memory for now and benchmark to see if blocking is a problem.

joshklop:
Much better, thanks @omerfirmak!

It is possible that we write a block header before (or while) responding to the original request. Our websocket package permits concurrent writes, so I don't think it's an issue.

rpc/handlers.go
```diff
@@ -1379,6 +1395,82 @@ func (h *Handler) SpecVersion() (string, *jsonrpc.Error) {
...
+		}()
+		for {
+			select {
+			case <-subscriptionCtx.Done():
+				return
+			case err := <-sub.Err():
+				h.log.Warnw("Subscription failed", "err", err)
+				return
+			case header := <-sink:
+				resp, err := json.Marshal(jsonrpc.Request{
```
omerfirmak, line 1437: why `Request` and not a `jsonrpc.response` ?

joshklop: Ethereum abuses the jsonrpc spec. I think the reasoning is that standard JSON-RPC 2.0 requests can be sent to the client without expecting a response; they just happen to contain information that is useful to the client.

I'm not sure how useful that is; I feel any reasonable client would need to implement support for subscriptions anyway.

rpc/handlers.go
```diff
@@ -1379,6 +1395,82 @@ func (h *Handler) SpecVersion() (string, *jsonrpc.Error) {
...
+	subscriptionCtx, subscriptionCtxCancel := context.WithCancel(ctx)
+	h.mu.Lock()
+	defer h.mu.Unlock()
+	h.subscriptions[id] = &subscription{
+		cancel: subscriptionCtxCancel,
+		conn:   w,
+	}
+
+	go func() {
+		sink := make(chan *core.Header)
```
omerfirmak, line 1421: Can you take a look at how `event.Subscription.Send()` behaves when one of the sinks are full/non-buffered (ie blocks on write)

rpc/handlers.go
```diff
@@ -1379,6 +1395,82 @@ func (h *Handler) SpecVersion() (string, *jsonrpc.Error) {
...
+		sub := h.bcReader.SubscribeNewHeads(sink)
```
omerfirmak, line 1422: This event will be triggered both on new blocks and reorg events. So the same block might be emitted multiple times.
Let's do what ETH does and not notify users when chain reorgs? https://docs.alchemy.com/reference/newheads

> NOTE: Chain Reorganizations (ReOrgs)

> When a chain reorganization occurs, this subscription will emit an event containing all new headers (blocks) for the new chain. This means that you may see multiple headers emitted with the same height (block number), and when this happens the later (highest) block number should be taken as the correct one after a reorganization.

rpc/handlers.go
joshklop: It sends on each channel that is subscribed, blocking if it tries to send on a channel that is full.

omerfirmak: So, our `Store()` might block for arbitrarily long if some of our WS connections are busy?

joshklop: Yes. A small buffer sounds reasonable. What do you think about `3`?

joshklop:
> It is possible that we write a block header before (or while) responding to the original request.

I now believe this a problem. Geth seems to fix this by buffering the handler's messages until the original response has been sent. This seems reasonable to me.

To do this, we need the `jsonrpc.Server` to:

1. write on the connection
2. set the connection value on the context on a per-message basis.

For (1), we can re-introduce the connection abstraction idea from before, where the transport is an `io.Writer` that is passed to `jsonrpc.Server.Handle`.

For (2), we can wrap the transport-level `io.Writer` in a `Conn` struct that has a `Write` method that looks something like:

```go
type Conn struct {
         w io.Writer
        activated bool
        buffer [][]byte
}

func (c *Conn) Write(data []byte) (int, error) {
        if c.activated {
                return c.Write(data)
        }
        c.buffer = append(c.buffer, data)
        return len(data), nil
}
```

I plan to implement this first thing tomorrow.

omerfirmak:
> > It is possible that we write a block header before (or while) responding to the original request.
> 
> I now believe this a problem. Geth seems to buffer the handler's messages until the original response has been sent. This seems reasonable to me.
> 
> To do this, we need the `jsonrpc.Server` to:
> 
>     1. write on the connection
> 
>     2. set the connection value on the context on a per-message basis.
> 
> 
> For (1), we can re-introduce the connection abstraction idea from before, where the transport is an `io.Writer` that is passed to `jsonrpc.Server.Handle`.
> 
> For (2), we can wrap the transport-level `io.Writer` in a `Conn` struct:
> 
> ```go
> type Conn struct {
>         w io.Writer
>         activated bool
>         buffer [][]byte
> }
> 
> func (c *Conn) Write(data []byte) (int, error) {
>         if c.activated {
>                 return c.Write(data)
>         }
>         c.buffer = append(c.buffer, data)
>         return len(data), nil
> }
> ```
> 
> I plan to implement this first thing tomorrow.

Can we make an exception for subscription handlers and let them write the initial response as well? Should be much easier to implement.

joshklop:
> Can we make an exception for subscription handlers and let them write the initial response as well? Should be much easier to implement.

I thought about this a little... how would the `jsonrpc` package know if a handler sends the initial response? I don't think we can do it without leaking subscription details to `jsonrpc`, which is what this PR tries to avoid.

omerfirmak:
> > Can we make an exception for subscription handlers and let them write the initial response as well? Should be much easier to implement.
> 
> I thought about this a little... how would the `jsonrpc` package know if a handler sends the initial response? I don't think we can do it without leaking subscription details to `jsonrpc`, which is what this PR tries to avoid.

if we return a nil response and nil error from the handler, `jsonrpc` wont write anything to the connection.

joshklop:
> if we return a nil response and nil error from the handler, jsonrpc wont write anything to the connection. 

Oh nice, I didn't realize that. I'll go with your idea then.

jsonrpc/websocket.go
```diff
@@ -56,20 +57,17 @@ func (ws *Websocket) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
-		resp, err = ws.rpc.Handle(ctx, msg)
-		if err != nil {
+		ws.listener.OnNewRequest("any")
+		if err = ws.rpc.HandleReadWriter(wsc.ctx, wsc); err != nil {
 			break
 		}
-		ws.listener.OnNewRequest("any")
-		if _, err = wsc.Write(resp); err != nil {
+		// From websocket docs: "Read to EOF otherwise connection will hang."
+		if _, err = io.ReadAll(wsc.r); err != nil {
```
omerfirmak, line 70: Do you think it is worth it to use `io.Discard` https://pkg.go.dev/io#Writer to avoid the allocation?

joshklop: How would we use `io.Discard` to flush the reader?

omerfirmak: https://pkg.go.dev/io#Copy

joshklop: `io.Copy` will allocate regardless https://github.com/golang/go/blob/2744155d369ca838be57d1eba90c3c6bfc4a3b30/src/io/io.go#L426

joshklop: Oh missed this https://github.com/golang/go/blob/2744155d369ca838be57d1eba90c3c6bfc4a3b30/src/io/io.go#L634-L635

I'll use `io.Discard` :+1:

blockchain/blockchain.go
```diff
@@ -42,6 +47,8 @@ type Reader interface {
 	EventFilter(from *felt.Felt, keys [][]felt.Felt) (*EventFilter, error)
 
 	Pending() (Pending, error)
+
+	SubscribeNewHeads() HeaderSubscription
```
omerfirmak, line 51: I would rather `blockchain` have an `EventListener` interface since subscriptions are probably only useful for RPC. But no need to bother changing it at the moment.

feed/feed.go
```diff
@@ -0,0 +1,69 @@
...
+func (f *Feed[T]) Subscribe() *Subscription[T] {
+	ch := make(chan T, 1)
+	f.mu.Lock()
+	defer f.mu.Unlock()
+	s := &Subscription[T]{
+		c:  ch,
+		f:  f,
+		id: f.nextID,
+	}
+	f.nextID++
```
omerfirmak, line 53: Let's use [atomics](https://pkg.go.dev/sync/atomic#Uint64.Add) to be safe here

feed/feed.go
```diff
@@ -0,0 +1,69 @@
...
+	s.unsubOnce.Do(func() {
+		s.f.mu.Lock()
+		defer s.f.mu.Unlock()
+		for _, sub := range s.f.subs {
+			if sub.id == s.id {
+				close(sub.c)
+				delete(s.f.subs, sub.id)
+				return
+			}
+		}
```
omerfirmak, line 34: why are we iterating?

feed/feed.go
joshklop: Done, although `f.mu` should be sufficient to protect `nextID`.

omerfirmak: Ah, my bad. Didn't realize that.

blockchain/blockchain.go
joshklop: > I would rather blockchain have an EventListener interface since subscriptions are probably only useful for RPC.

Would the `blockchain.Reader` have a `SetListener(EventListener)` method? The `rpc.Handler` only accepts a `blockchain.Reader`, not a `blockchain.Blockchain`.

omerfirmak: If I were to do it, I would implement `blockchain.EventListener` interface in `rpc.Handler` and set it as the event listener.

joshklop: That sounds nice.

> set it as the event listener

If I'm understanding correctly, we would still need to modify the `blockchain.Reader` interface with a `SetListener` method. That would overwrite the existing listener. Is that a fair trade-off?

omerfirmak: > set it as the event listener.

We can do this in `node` package to avoid modifying `blockchain.Reader`

omerfirmak (approved):
Could use a test that serves multiple subscriptions at the same time.

joshklop:
Added the test with multiple subscribers. This is ready to be merged when the CI checks finish.

PR #1288, opened 2023-10-03
Bob/juno cicd

NetherBob:
Here I have added a timeout minute to be 10 minutes. This will mean, the job will timeout after 10 minutes of running which is more than enough to carry out the commands in the job.


gehlotanish (approved):
LGTM

wojciechos:
@NetherBob, do we really need a 10-minute timeout here? The `deploy_to_staging` job mainly dispatches another workflow with a curl command and won't wait for the triggered workflow to finish - it should take just a few seconds.

BTW instead of using a dispatch, have we considered refactoring the target workflow to be reusable and then using workflow_call to call it directly? I mean [this](https://docs.github.com/en/actions/using-workflows/reusing-workflows).

PR #1300, opened 2023-10-05
return default storage and nonce values for contracts deployed on pending block

omerfirmak:


blockchain/pending.go
```diff
@@ -44,6 +44,9 @@ func (p *PendingState) ContractNonce(addr *felt.Felt) (*felt.Felt, error) {
 		return nonce, nil
 	}
 
+	if p.isDeployedInPending(addr) {
+		return &felt.Zero, nil
```
joshklop, line 48: Is it risky that we use `&felt.Zero` in many places? It seems we might mutate it on accident.

joshklop (approved):

blockchain/pending.go
omerfirmak: > mutate it on accident

Yeah, we should move away from using felt pointers as soon as possible.

PR #1310, opened 2023-10-06
Fix jsonrpc notification handling

joshklop:


joshklop:
Just realized this is fixed in #1279, disregard.

PR #1317, opened 2023-10-09
P2P Get events recent changes

kirugan:


p2p/starknet/handlers.go
```diff
@@ -134,16 +136,34 @@ func (h *Handler) onBlockBodiesRequest(req *spec.BlockBodiesRequest) (Stream[pro
...
-	count := uint64(0)
+	it, err := h.newIterator(req.Iteration)
+	if err != nil {
+		return nil, err
+	}
+
 	return func() (proto.Message, bool) {
-		if count > 3 {
+		if it.Valid() {
 			return nil, false
```
omerfirmak, line 146: we need to send a `Fin` message when we are done.

https://github.com/NethermindEth/juno/blob/main/p2p/starknet/p2p/proto/event.proto#L24

omerfirmak:
```
    optional BlockID id = 1; // may not appear if Fin is sent to end the whole response
```

Id is optional. We can leave it `nil`

p2p/starknet/handlers.go
```diff
@@ -134,16 +136,45 @@ func (h *Handler) onBlockBodiesRequest(req *spec.BlockBodiesRequest) (Stream[pro
...
+			events = append(events, item.Events...)
+		}
+
 		return &spec.EventsResponse{
-			Id: &spec.BlockID{
-				Number: count - 1,
+			Id: core2p2p.AdaptBlockID(block.Header),
+			Responses: &spec.EventsResponse_Events{
+				Events: &spec.Events{
+					Items: utils.Map(events, core2p2p.AdaptEvent),
```
omerfirmak, line 176: We need to chunk the events in case the messages exceeds 1MB, but we can leave it for later.

p2p/starknet/starknet_test.go
```diff
@@ -97,14 +98,37 @@ func TestClientHandler(t *testing.T) {
...
 		require.NoError(t, cErr)
 
-		count := uint64(0)
+		var count uint64
 		for evnt, valid := res(); valid; evnt, valid = res() {
 			assert.Equal(t, count, evnt.Id.Number)
 			count++
 		}
-		require.Equal(t, uint64(4), count)
+		require.Equal(t, uint64(limit), count)
```
omerfirmak, line 132: can we explicitly check for fin message here?

adapters/core2p2p/block.go
IronGauntlets, line 1: Are we not doing to add test for adapters?

p2p/starknet/starknet_test.go
```diff
@@ -97,14 +98,42 @@ func TestClientHandler(t *testing.T) {
...
+		const limit uint64 = 3
+		const firstBlock uint64 = 0
+
+		for blockNumber := firstBlock; blockNumber < limit; blockNumber++ {
+			mockReader.EXPECT().BlockByNumber(blockNumber).Return(&core.Block{
+				Header: &core.Header{
+					Number: blockNumber,
+				},
+				Receipts: []*core.TransactionReceipt{
+					{Events: []*core.Event{}},
```
omerfirmak, line 110: Let's add some events here and make sure they are adapted properly? We will cover AdaptEvents as well that way.
Like, start from no events and add increasingly more events each block.

p2p/starknet/starknet_test.go
```diff
@@ -97,14 +101,100 @@ func TestClientHandler(t *testing.T) {
...
 
-		count := uint64(0)
+		var count uint64
 		for evnt, valid := res(); valid; evnt, valid = res() {
+			if count == limit {
+				expectedFin := &spec.EventsResponse{
+					Responses: &spec.EventsResponse_Fin{},
+				}
+				assert.True(t, proto.Equal(expectedFin, evnt))
+				break
```
omerfirmak, line 158: let's increase the count here as well, otherwise we might miss this branch but the check at L180 will still pass.

p2p/starknet/handlers.go
```diff
@@ -134,16 +136,46 @@ func (h *Handler) onBlockBodiesRequest(req *spec.BlockBodiesRequest) (Stream[pro
...
+		block, err := it.Block()
+		if err != nil {
+			h.log.Errorw("Failed to fetch block", "err", err)
+			return fin()
+		}
+		it.Next()
+
+		var events []*core.Event
+		for _, item := range block.Receipts {
+			events = append(events, item.Events...)
```
omerfirmak, line 170: Instead of doing this, can we straight up build the `[]*spec.Event` to avoid allocating the intermediate `events []*core.Event`

p2p/starknet/handlers.go
```diff
@@ -134,16 +135,47 @@ func (h *Handler) onBlockBodiesRequest(req *spec.BlockBodiesRequest) (Stream[pro
...
+		block, err := it.Block()
+		if err != nil {
+			h.log.Errorw("Failed to fetch block", "err", err)
+			return fin()
+		}
+		it.Next()
+
+		items := make([]*spec.Event, 0, len(block.Receipts))
+		for _, item := range block.Receipts {
+			events := utils.Map(item.Events, core2p2p.AdaptEvent)
```
omerfirmak, line 169: this is another allocation we can avoid by simply looping over item.Events and appending each event to `items` one by one after adapting.

p2p/starknet/starknet_test.go
```diff
@@ -97,14 +101,102 @@ func TestClientHandler(t *testing.T) {
...
 		require.NoError(t, cErr)
 
-		count := uint64(0)
+		var count uint64
 		for evnt, valid := res(); valid; evnt, valid = res() {
+			if count == numOfBlocks {
+				expectedFin := &spec.EventsResponse{
+					Responses: &spec.EventsResponse_Fin{},
+				}
+				assert.True(t, proto.Equal(expectedFin, evnt))
```
IronGauntlets, line 157: Can you please elaborate on why are you not using `assert.Eqaul()`?

p2p/starknet/starknet_test.go
```diff
@@ -97,14 +101,102 @@ func TestClientHandler(t *testing.T) {
...
+						return &spec.Event{
+							FromAddress: adaptFelt(e.From),
+							Keys:        utils.Map(e.Keys, adaptFelt),
+							Data:        utils.Map(e.Data, adaptFelt),
+						}
+					}),
+				},
+			}
+
+			assert.True(t, proto.Equal(expectedEventsResponse.Events, evnt.Responses.(*spec.EventsResponse_Events).Events))
```
IronGauntlets, line 178: Same as above. 

Also, consider using: `event.Responses.GetEvents()` so that type assertion is not required.

PR #1321, opened 2023-10-09
P2P GetBlockHeaders handler

kirugan:


adapters/core2p2p/block.go
```diff
@@ -0,0 +1,40 @@
...
+		ParentHeader:     AdaptHash(header.ParentHash),
+		Number:           header.Number,
+		Time:             timestamppb.New(time.Unix(int64(header.Timestamp), 0)),
+		SequencerAddress: AdaptAddress(header.SequencerAddress),
+		StateDiffs:       nil,
+		State:            nil,
+		ProofFact:        nil, // ???
+		Transactions:     nil,
+		Events:           nil,
+		Receipts:         nil,
```
kirugan, line 30: Where should I get those ? 😭

joshklop: Does it work if we get those values from the db and pass them into the adapter?

kirugan: Sure, I just need to know which values to use for field that are nil

p2p/starknet/handlers.go
```diff
@@ -96,19 +99,48 @@ func (h *Handler) TransactionsHandler(stream network.Stream) {
...
-						Header: &spec.BlockHeader{
-							Number: count - 1,
+						Header: core2p2p.AdaptHeader(header, commitments),
+					},
+				},
+				{
+					HeaderMessage: &spec.BlockHeadersResponsePart_Signatures{
+						Signatures: &spec.Signatures{
+							Block:      core2p2p.AdaptBlockID(header),
+							Signatures: []*spec.ConsensusSignature{}, // todo fill signatures
```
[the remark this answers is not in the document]
omerfirmak: You should be able to use `header.Signatures` without waiting the ECDSA PR.

p2p/starknet/starknet_test.go
```diff
@@ -49,15 +50,39 @@ func TestClientHandler(t *testing.T) {
...
 	t.Run("get block headers", func(t *testing.T) {
-		res, cErr := client.RequestBlockHeaders(testCtx, &spec.BlockHeadersRequest{})
+		const limit uint64 = 2
+		const firstBlockNumber uint64 = 0
+
+		for i := firstBlockNumber; i < limit; i++ {
+			mockReader.EXPECT().BlockHeaderByNumber(i).Return(&core.Header{
+				Number: i,
+			}, nil)
+			mockReader.EXPECT().BlockCommitmentsByNumber(i).Return(&core.BlockCommitments{}, nil)
```
omerfirmak, line 60: let's populate the headers with sane values and make sure they adapt properly

p2p/starknet/iterator.go
```diff
@@ -67,10 +92,35 @@
...
+		block, err = it.bcReader.BlockByHash(it.blockHash)
+		if err != nil {
+			// after first successful fetch by hash
+			// we switch to iteration by block number
+			it.blockHash = nil
+			it.blockNumber = block.Number
+		}
+	} else {
+		block, err = it.bcReader.BlockByNumber(it.blockNumber)
+	}
```
IronGauntlets, line 110: I think it would be useful to have this behaviour tested

p2p/starknet/starknet_test.go
```diff
@@ -53,15 +55,104 @@ func TestClientHandler(t *testing.T) {
...
 
 	t.Run("get block headers", func(t *testing.T) {
-		res, cErr := client.RequestBlockHeaders(testCtx, &spec.BlockHeadersRequest{})
+		type pair struct {
+			header      *core.Header
+			commitments *core.BlockCommitments
+		}
+		pairsPerBlock := []pair{}
+		for i := 0; i < 2; i++ {
+			ui := uint64(i)
```
IronGauntlets, line 64: Consider `for i := uint64(0)...`, so there is no need to create another variable.

IronGauntlets (changes requested):
Most things looks good. I think have some test for iterating using block hash would be useful.

adapters/core2p2p/block.go
```diff
@@ -17,6 +21,37 @@ func AdaptBlockID(header *core.Header) *spec.BlockID {
...
+	return &spec.BlockHeader{
+		ParentHeader:     AdaptHash(header.ParentHash),
+		Number:           header.Number,
+		Time:             timestamppb.New(time.Unix(int64(header.Timestamp), 0)),
+		SequencerAddress: AdaptAddress(header.SequencerAddress),
+		ProofFact:        nil, // not defined yet
+		Receipts:         nil, // not defined yet
+		StateDiffs:       nil, // not defined yet
+		State: &spec.Patricia{
+			Height: uint32(header.Number),
```
omerfirmak, line 41: The `Height` here is the height of the patricia trie. So it should be a fixed `251`. I know, It is weird but, yeah.

p2p/starknet/iterator.go
```diff
@@ -62,15 +87,36 @@ func (it *iterator) Next() bool {
...
+		block, err = it.bcReader.BlockByHash(it.blockHash)
+		if err == nil {
+			// after first successful fetch by hash
+			// we switch to iteration by block number
+			it.blockHash = nil
+			it.blockNumber = block.Number
+		}
+	} else {
+		block, err = it.bcReader.BlockByNumber(it.blockNumber)
+	}
```
omerfirmak, line 106: IMO, we can avoid this additional complexity. We can keep `iterator` operate on block numbers only and make `newIteratorByHash` look up the  block number of a given block hash before creating the iterator. Similar to how `ByHash` functions work in `blockchain` for example;

```golang
func blockHeaderByHash(txn db.Transaction, hash *felt.Felt) (*core.Header, error) {
	var header *core.Header
	return header, txn.Get(db.BlockHeaderNumbersByHash.Key(hash.Marshal()), func(val []byte) error {
		var err error
		header, err = blockHeaderByNumber(txn, binary.BigEndian.Uint64(val))
		return err
	})
}
```

IronGauntlets: I agree

p2p/starknet/iterator.go
```diff
@@ -31,13 +33,33 @@ func newIterator(bcReader blockchain.Reader, blockNumber, limit, step uint64, fo
...
 	}, nil
 }
 
+func newIteratorByHash(bcReader blockchain.Reader, blockHash *felt.Felt, limit, step uint64, forward bool) (*iterator, error) {
+	if step == 0 {
+		return nil, fmt.Errorf("step is zero")
+	}
+	if limit == 0 {
+		return nil, fmt.Errorf("limit is zero")
+	}
```
omerfirmak, line 50: redundant, `newIteratorByNumber` will check these anyways

p2p/starknet/iterator.go
```diff
@@ -6,12 +6,14 @@ import (
...
 	"github.com/NethermindEth/juno/blockchain"
 	"github.com/NethermindEth/juno/core"
+	"github.com/NethermindEth/juno/core/felt"
 	"github.com/NethermindEth/juno/db"
 )
 
 type iterator struct {
 	bcReader blockchain.Reader
 
+	blockHash   *felt.Felt
```
omerfirmak, line 16: why do we still need this?

IronGauntlets (approved):

omerfirmak (approved):
LGTM, apart from the `Fin` comment I made on slack.

PR #1323, opened 2023-10-09
Fixing macOS tests

kirugan:


kirugan:
I'll give it another try

PR #1332, opened 2023-10-11
Send newHeads event only if store succeded for sure

omerfirmak:
and the block is available for users to query via RPC


joshklop (approved):

joshklop:
Would it make more sense for the event to be sent by the synchronizer instead of the blockchain?

omerfirmak:
> Would it make more sense for the event to be sent by the synchronizer instead of the blockchain?

Maybe. We could make `blockchain` just a DB and move all the moving parts to sync.

Issue #1226, opened 2023-09-13
Implement ECDSA signature verification


omerfirmak:
we should be able to use `gnark-crypto` for this

omerfirmak:
Done in #1333

PR #1333, opened 2023-10-11
Add ECDSA signature verification

omerfirmak:


omerfirmak:
Closes #1226

core/crypto/ecsda.go
```diff
@@ -0,0 +1,77 @@
...
+	gnarkPk := (*ecdsa.PublicKey)(k)
+
+	var sigBytes [felt.Bytes * 2]byte
+	rBytes := sig.R.Bytes()
+	copy(sigBytes[0:felt.Bytes], rBytes[:])
+	sBytes := sig.S.Bytes()
+	copy(sigBytes[felt.Bytes:], sBytes[:])
+
+	msgBytes := msg.Bytes()
+	return gnarkPk.Verify(sigBytes[:], msgBytes[:], nil)
```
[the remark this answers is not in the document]
omerfirmak: > I guess everything works fine, but just to double check: sigBytes len is 64, msgBytes len is 32 and that's ok, right?
> 

Yeah, signatures have 2 felt.Felt components. R and S

> Can you please also give me a link to read more about all these calculations and what they mean?

https://brilliant.org/wiki/elliptic-curves/

Our elliptic curve is simply a function in the form of `y^2 = f(x)` Given `x` we can calculate `y^2`. Since it is squared, `y^2` gives us 2 possible values of y, one positive one negative. `yCoord` function simply calculates `y` value from `x`.

core/crypto/ecsda.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/consensys/gnark-crypto/ecc/stark-curve/fp"
+)
+
+type Signature struct {
+	R, S felt.Felt
+}
+
+type PublicKey ecdsa.PublicKey
+
+func NewPublicKey(x *felt.Felt) PublicKey {
```
rianhughes, line 18: Should we allow users to specify the Y coordinate as well?

core/crypto/ecsda.go
```diff
@@ -0,0 +1,77 @@
...
+	return PublicKey(ecdsa.PublicKey{
+		A: starkcurve.G1Affine{
+			X: *x.Impl(),
```
rianhughes, line 21: Maybe we could validate x?

rianhughes: I guess this is done in `return ySquared.Sqrt(&ySquared)`

rianhughes (approved):

PR #1335, opened 2023-10-12
Sync from L1

joshklop:
I'll continue polishing this as I finish the remaining to-dos. Wanted to put things in a PR to track progress for now.

- Finish testing the full sync. Probably will want to have some basic parallelism to speed things up. 
- Add a sub-command, probably with only `--network` and `--db-path` options.
- Factor out the backoff logic from `feeder.go` so we can test against Infura.


rodrigo-pino:
This is an interesting PR and would've loved making it part of Juno but it is:
- 2 years old, which will require updating to every Starknet version
- There seems to be no need for it

PR #1337, opened 2023-10-12
Return TxnStatus instead of TxnFinalityStatus from RPC

joshklop:


rpc/transaction.go
```diff
@@ -66,6 +66,58 @@ func (t *TransactionType) UnmarshalJSON(data []byte) error {
...
+		return TxnStatusReceived
+	case `"REJECTED"`:
+		return TxnStatusRejected
+	case `"ACCEPTED_ON_L2"`:
+		return TxnStatusAcceptedOnL2
+	case `"ACCEPTED_ON_L1"`:
+		return TxnStatusAcceptedOnL1
+	default:
+		// Should not happen
+		panic("unknown transaction status")
```
omerfirmak, line 90: Let's not panic on these

rpc/transaction.go
```diff
@@ -66,6 +66,58 @@ func (t *TransactionType) UnmarshalJSON(data []byte) error {
...
+
+func adaptFinalityStatusToStatus(s TxnFinalityStatus) TxnStatus {
+	switch s {
+	case TxnAcceptedOnL1:
+		return TxnStatusAcceptedOnL1
+	case TxnAcceptedOnL2:
+		return TxnStatusAcceptedOnL2
+	default:
+		// Should not happen
+		panic("unknown finality status")
```
omerfirmak, line 117: Let's not panic on these

rpc/transaction.go
```diff
@@ -66,6 +66,58 @@ func (t *TransactionType) UnmarshalJSON(data []byte) error {
...
+		return TxnStatusReceived, nil
+	case "REJECTED":
+		return TxnStatusRejected, nil
+	case "ACCEPTED_ON_L2":
+		return TxnStatusAcceptedOnL2, nil
+	case "ACCEPTED_ON_L1":
+		return TxnStatusAcceptedOnL1, nil
+	default:
+		// Should not happen
+		return 0, errors.New("unknown feeder status")
```
omerfirmak, line 90: Do you think it makes sense to default to NOT_RECEIVED?

rpc/handlers.go
```diff
@@ -1178,8 +1171,7 @@ func (h *Handler) TransactionStatus(ctx context.Context, hash felt.Felt) (*Trans
 		case feeder.Reverted:
 			status.Execution = TxnFailure
 		default:
-			// pre-0.12.1
-			status.Execution = TxnSuccess
+			return nil, jsonrpc.Err(jsonrpc.InternalError, "unknown execution status from the feeder gateway")
```
omerfirmak, line 1174: Let's not error but omit the field? status.Finality may have some useful information

rpc/transaction.go
joshklop: `NOT_RECEIVED` is not part of the `TXN_STATUS` enum in the spec ([ref](https://github.com/starkware-libs/starknet-specs/blob/c0980ed9aa5a8eb2def8d9b145fba27b4e29b7f9/api/starknet_api_openrpc.json#L2771)). For now, I'll add a comment saying `adaptFeederStatus` assumes `s != "NOT_RECEIVED"`.

joshklop:
After thinking about the problem, I'm having second thoughts about this. The feeder uses the `status` field in mysterious ways, [sometimes](https://alpha-mainnet.starknet.io/feeder_gateway/get_transaction?transactionHash=0x6c40890743aa220b10e5ee68cef694c5c23cc2defd0dbdf5546e687f9982ab1) using it to refer to the finality status and [sometimes](https://external.integration.starknet.io/feeder_gateway/get_transaction?transactionHash=0x19abec18bbacec23c2eee160c70190a48e4b41dd5ff98ad8f247f9393559998) for the execution status.

Because the version on `main` fulfills the spec and because the feeder seems to always include the `finality_status` and `execution_status` fields for received transactions, I'll close this. If I'm missing something, feel free to re-open and let me know what needs changed.

PR #1344, opened 2023-10-16
Cache generated Traces

omerfirmak:


omerfirmak:
> LGTM, it would be cool to compare perf improvement over old approach (tracing limited number of transactions)

That depends on the block we are tracing, but when the cache is hit the response time is around 50ms.

PR #1345, opened 2023-10-17
Add support for Rejected Txns

omerfirmak:


joshklop (approved):
Thanks for taking care of this.

PR #1353, opened 2023-10-18
Implement RemoteDB over gRPC

omerfirmak:


omerfirmak:
To give this a try locally;

master node: `juno --db-path ../juno_mainnet/ --http --grpc`
parasite node `juno --remote-db localhost:6064 --http --http-port 6061`

omerfirmak:
Closing since the idea is shelved for now.

db/remote/transaction.go
```diff
@@ -0,0 +1,76 @@
...
+func (t *transaction) Delete(key []byte) error {
+	return errors.New("read only DB")
+}
+
+func (t *transaction) Get(key []byte, cb func([]byte) error) error {
+	iter, err := t.NewIterator() // we shouldnt need to do this, add a new Op_GET
+	if err != nil {
+		return err
+	}
+	defer iter.Close()
```
joshklop: Check for error

db/remote/iterator.go
```diff
@@ -0,0 +1,71 @@
...
+
+	i.currentK = pair.K
+	i.currentV = pair.V
+	return nil
+}
+
+func (i *iterator) Valid() bool {
+	if len(i.currentK) == 0 && len(i.currentV) == 0 {
+		if err := i.doOpAndUpdate(gen.Op_CURRENT, nil); err != nil {
+			i.log.Debugw("error", "op", gen.Op_CURRENT, "err", err)
```
joshklop, line 41: nit: capitalize log messages.

db/remote/transaction.go
```diff
@@ -0,0 +1,76 @@
...
+		cursorID: pair.CursorId,
+		log:      t.log,
+	}, nil
+}
+
+func (t *transaction) Discard() error {
+	return t.client.CloseSend()
+}
+
+func (t *transaction) Commit() error {
```
joshklop, line 43: Can we have a separate `ReadOnlyDB` interface, and then embed it in the `DB` interface? It should remove the need for these superfluous functions and make the db argument to `rpc.New` more precise.

sync/sync.go
```diff
@@ -239,7 +241,9 @@ func (s *Synchronizer) syncBlocks(syncCtx context.Context) {
 	s.startingBlockNumber = &startingHeight
 
 	pendingSem := make(chan struct{}, 1)
-	go s.pollPending(syncCtx, pendingSem)
+	if !s.readOnlyBlockchain {
```
joshklop, line 244: Can we move everything that needs to be initialized/run when the db is read-only to the top of `syncBlocks`, and then insert
```go
if s.readOnlyBlockchain {
        <-streamCtx.Done
        streamCancel()
        return
}
```

grpc/handlers_test.go
```diff
@@ -94,20 +94,22 @@ func TestHandlers_Tx(t *testing.T) {
 	})
 	cur, err := stream.RecvToClient()
 	require.NoError(t, err)
+	cID := cur.CursorId
```
joshklop, line 97: Is this change necessary?

grpc/handlers.go
```diff
@@ -58,30 +58,38 @@ func (h Handler) Tx(server gen.KV_TxServer) error {
...
+		return server.Send(responsePair)
+	} else if cur.Op == gen.Op_GET {
+		if err := tx.dbTx.Get(cur.K, func(b []byte) error {
+			responsePair.V = append(responsePair.V, b...)
+			responsePair.K = cur.K
+			return nil
+		}); err != nil && !errors.Is(err, db.ErrKeyNotFound) {
+			return err
+		}
+		return server.Send(responsePair)
```
joshklop, line 85: ~How will the receiving end determine if the key was not found?~ next commit explains it.

joshklop (approved):
:bow:

db/db.go
```diff
@@ -14,7 +18,7 @@ type DB interface {
 
 	// NewTransaction returns a transaction on this database, it should block if an update transaction is requested
 	// while another one is still in progress
-	NewTransaction(update bool) Transaction
+	NewTransaction(update bool) (Transaction, error)
```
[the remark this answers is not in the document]
joshklop: Read the commits in order, makes sense now.

db/remote/transaction.go
joshklop: Looking at the subsequent commits, it looks like this wouldn't work well since the db is also passed to the synchronizer.

grpc/handlers_test.go
omerfirmak: Yeah, response to Op_GET doesnt have a CursorID set. So it messes up the next Op_CLOSE call.

grpc/tx.go
```diff
@@ -3,20 +3,23 @@ package grpc
...
 )
 
 type tx struct {
 	dbTx db.Transaction
 	// index is cursorId for an iterator
-	iterators []db.Iterator
+	iterators sync.Map
 }
 
+var itCounter atomic.Uint32
```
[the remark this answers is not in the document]
omerfirmak: Hmm, I think we might be able to. gRPC txns don't have ids so I must have thought that iterators needed global unique IDs.

grpc/handlers_test.go
```diff
@@ -116,4 +114,6 @@ func TestHandlers_Tx(t *testing.T) {
 		assert.Empty(t, cur.K)
 		assert.Empty(t, cur.V)
 	}
+	close(stream.recvToServer) // signal server that we are done
+	<-stream.sentFromServer    // wait for server to shutdown
```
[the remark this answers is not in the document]
omerfirmak: Neat, done.

sync/sync.go
omerfirmak: I wanted to keep using the existing "shutdown path" that sync service followed just to be safe. Pushed a new commit, let me know what you think.

grpc/tx.go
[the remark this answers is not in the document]
omerfirmak: Done

grpc/handlers_test.go
```diff
@@ -75,18 +75,15 @@ func createTxStream(t *testing.T, h Handler) *grpcStreamMock {
 	go func() {
 		err := h.Tx(stream)
 		if err != nil {
-			t.Errorf(err.Error())
+			assert.ErrorContains(t, err, "empty")
```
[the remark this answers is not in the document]
omerfirmak: `h.Tx` doesnt always fail tho. The aim here is to, if it fails, make sure the error is an expected one.

sync/sync.go
joshklop: I think we can move the `fetchers, verifiers := s.setupWorkers()` call below the `if` statement. Otherwise LGTM.

grpc/handlers_test.go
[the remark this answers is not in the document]
joshklop: You sure? It works for me when I change it locally.

Intuitively, the `stream` shouldn't end until `Op_CLOSE`, which is sent last in the test.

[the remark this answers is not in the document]
omerfirmak: I might be wrong. But honestly, it doesnt really matter. Leaving it as is doesnt really hurt.

PR #1354, opened 2023-10-18
Small p2p refactor

IronGauntlets:


node/node.go
```diff
@@ -179,12 +178,9 @@ func New(cfg *Config, version string) (*Node, error) { //nolint:gocyclo,funlen
...
 
 	if cfg.P2P {
-		var privKeyStr string
-		privKeyStr, _ = os.LookupEnv("P2P_PRIVATE_KEY")
-		var p2pService *p2p.Service
-		p2pService, err = p2p.New(cfg.P2PAddr, "juno", cfg.P2PBootPeers, privKeyStr, cfg.Network, log)
+		p2pService, err := p2p.New(cfg.P2PAddr, "juno", cfg.P2PBootPeers, "", cfg.Network, log)
 		if err != nil {
-			return nil, fmt.Errorf("set up p2p service: %w", err)
+			return nil, fmt.Errorf("Failed to set up p2p service: %w", err)
```
joshklop, line 183: nit: I think the previous error message is more idiomatic.

p2p/p2p.go
```diff
@@ -45,17 +44,10 @@ type Service struct {
...
-	bootPeers,
-	privKeyStr string,
-	snNetwork utils.Network,
-	log utils.SimpleLogger,
-) (*Service, error) {
+func New(addr, userAgent, bootPeers, privKeyStr string, snNetwork utils.Network, log utils.SimpleLogger) (*Service, error) {
 	if addr == "" {
-		// 0.0.0.0 will listen on any interface device.
-		addr = fmt.Sprintf("/ip4/0.0.0.0/tcp/%d", defaultSourcePort)
+		// 0.0.0.0/tcp/0 will listen on any interface device and assing a free port.
```
joshklop, line 49:
```suggestion
		// 0.0.0.0/tcp/0 will listen on any interface device and assign a free port.
```

Not a problem with this PR, but I am concerned about making the p2p server accessible over the network by default. What do you think about defaulting to `127.0.0.1`? I'm not convinced that's a reasonable default either, though.

joshklop (approved):

node/node.go
IronGauntlets: Reverting. Could you please send a reference right now we have a mixture of these in the codebase.

p2p/p2p.go
IronGauntlets: It is fine for now.

omerfirmak: Why are we listening on a random port?

PR #1356, opened 2023-10-19
Update fin behaviour in getEvents, getBlockHeaders handlers

kirugan:


p2p/starknet/handlers.go
```diff
@@ -144,6 +137,9 @@ func (h *Handler) onBlockHeadersRequest(req *spec.BlockHeadersRequest) (Stream[p
 						},
 					},
 				},
+				{
+					HeaderMessage: &spec.BlockHeadersResponsePart_Fin{},
```
omerfirmak, line 141: Don't wee need to set block id here?
Fin without a block ID is only used for EOF.

omerfirmak: I think that is a mistake. `BlockHeadersResponsePart` should include the BlockID

omerfirmak:
This was probably fixed/will be fixed in p2p branch, closing for now.

PR #1359, opened 2023-10-23
P2P GetBlockBodies

kirugan:


p2p/starknet/handlers.go
```diff
@@ -150,16 +150,63 @@ func (h *Handler) onBlockHeadersRequest(req *spec.BlockHeadersRequest) (Stream[p
...
+	if err != nil {
+		return nil, err
+	}
+
+	fin := h.newFin(&spec.BlockBodiesResponse{
+		BodyMessage: &spec.BlockBodiesResponse_Fin{},
+	})
+
+	// Either BlockBodiesResponse_Diff, *_Classes, *_Proof, *_Fin
+	var pendingMessages []proto.Message
```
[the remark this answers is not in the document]
omerfirmak: > or state machine
IMO, this

p2p/starknet/handlers.go
```diff
@@ -150,16 +150,63 @@ func (h *Handler) onBlockHeadersRequest(req *spec.BlockHeadersRequest) (Stream[p
...
+
+		state, err := h.bcReader.StateUpdateByNumber(block.Number)
+		if err != nil {
+			h.log.Errorw("Failed to get state", "err", err)
+			return fin()
+		}
+		diff := state.StateDiff
+
+		var contractDiffs []*spec.StateDiff_ContractDiff
+		for _, contract := range diff.DeployedContracts {
```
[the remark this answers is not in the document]
omerfirmak: Yes, pretty much the entire `core.StateDiff` should be adapted.

p2p/starknet/block_bodies.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+	for _, class := range stateDiff.DeclaredV1Classes {
+		cls, err := state.Class(class.ClassHash)
+		if err != nil {
+			return b.finAndSetTerminalState()
+		}
+
+		cairo1Cls := cls.Class.(*core.Cairo1Class)
+		classes = append(classes, core2p2p.AdaptClass(cls.Class, cairo1Cls.Hash()))
+	}
```
IronGauntlets, line 101: I think we also need to check the `deployedContracts` because there have been cases in the older blocks where Contracts were deployed without them being declared (I don't know why, its one of those Starknet things). Please have a look at the `fetchUnknownClasses()` in `sync/sync.go`.

p2p/starknet/block_bodies.go
```diff
@@ -0,0 +1,164 @@
...
+	_ streamState = iota
+
+	sendDiff // initial
+	sendClasses
+	sendBlockFin
+	terminal // final state
+)
+
+type blockBodyIterator struct {
+	bcReader blockchain.Reader
```
IronGauntlets, line 26: I don't think we need to pass the blockchain Reader here as we only use it for 2 calls. Usually, it is not a good idea to pass alarge interface when it is not fully utilised (we should break up the Reader into smaller interfaces, this is for another pr, ofcourse).

Passing `core.StateUpdate` and `core.StateReader` should be enough. We only use the blockchain reader for a single block and the iterator is also created per block, so I don't think it is wise to pass a stuct that is intended to provide data for multiple blocks.

p2p/starknet/block_bodies.go
```diff
@@ -0,0 +1,164 @@
...
+	switch b.state {
+	case sendDiff:
+		msg, valid = b.diff()
+		b.state = sendClasses
+	case sendClasses:
+		msg, valid = b.classes()
+		b.state = sendBlockFin
+	case sendBlockFin:
+		msg, valid = b.fin()
+		b.state = terminal
```
IronGauntlets, line 64: I think this can be refactored to use `finAndSetTerminalState()` as it is doing similar things. It just seems strange that we have fin() which doesn't set the state to terminal, but then after its usage it sets the state to terminal immediately afterwards.

IMO we should only have one `fin()` and based on the current usages of the 2 fin functions i think it is possible.

p2p/starknet/block_bodies.go
```diff
@@ -0,0 +1,164 @@
...
+	var err error
+	b.stateUpdate, err = b.bcReader.StateUpdateByNumber(b.header.Number)
+	if err != nil {
+		b.log.Errorw("Failed to get state", "err", err)
+		return b.finAndSetTerminalState()
+	}
+	diff := b.stateUpdate.StateDiff
+
+	var contractDiffs []*spec.StateDiff_ContractDiff
+	contractPairs := utils.Flatten(diff.DeployedContracts, diff.ReplacedClasses)
```
IronGauntlets, line 124: I don't think this is correct. `DeployedContracts` and `ReplacedClasses` state may or may not be modified. Based on the current implementation you may miss some contracts whose state was changed.

What should be done is we need to range through the `StateDiff.StorageDiffs` and `StateDiff.Nonce` and determine all the contract address whose state was modified. It is possible that a contract storage was modified but not its nonce and vice versa.

I will push a fix.

IronGauntlets (changes requested):

adapters/core2p2p/class.go
```diff
@@ -0,0 +1,26 @@
...
+	switch v := class.(type) {
+	case *core.Cairo0Class:
+		return &spec.Class{
+			CompiledHash: AdaptHash(compiledHash),
+			Definition:   []byte(v.Program),
+		}
+	case *core.Cairo1Class:
+		return &spec.Class{
+			CompiledHash: AdaptHash(compiledHash),
+			Definition:   v.Compiled,
```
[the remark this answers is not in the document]
IronGauntlets: Resolving as I commented for future reference.

p2p/starknet/block_bodies.go
IronGauntlets: I have updated the implementation to make sure there are no duplicate classes. Also, added a test case for it.

p2p/starknet/block_bodies.go
IronGauntlets: Done

IronGauntlets (approved):
I pushed some changes. Please have a look at the latest commits. If they look good please squash and merge.

PR #1362, opened 2023-10-24
jsonrpc: implement unix handler

### LINKED ISSUE #1277: Unix socket support
## Issue
Currently, jsonrpc only supports http and ws.

## Rationale

**Why should this feature exist?**
Unix socket support would enhance the project by providing a more secure and efficient communication mechanism. Unix sockets typically have lower latency and higher throughput than TCP. This feature is particularly useful when hosting multiple users on the same machine since access can be easly restricted.

## Implementation
Do you have ideas regarding the implementation of this feature? yes
Are you willing to implement this feature? yes

**Scope:**
- Implementing unix handler.

- Adding cli-flags (`unix-enabled` and `unix-path` ).

## Additional Information:
Windows would require downloading another [package](https://github.com/microsoft/go-winio).

Exca-DK:
closes #1277


omerfirmak:
Hey, thanks for the contribution!

Do you have an immediate use case for this? Because I am not sure if the marginal speed up that IPC brings, on its own, warrants a new feature to be added over TCP.

Exca-DK:
I usually mount the socket instead of using loopback since it saves 2(docker+interface itself) hops for containerized env and is generally easier to manage.

omerfirmak:
> I usually mount the socket instead of using loopback since it saves 2(docker+interface itself) hops for containerized env and is generally easier to manage.

Fair enough

joshklop:
I'll try to review this sometime next week. May take a bit longer depending on prioritization.

jsonrpc/ipc_all.go
```diff
@@ -0,0 +1,62 @@
...
+	l, err = net.Listen("unix", endpoint)
+	if err != nil {
+		return nil, err
+	}
+	// Set permissions for the socket file to read and write for the owner only (0o600)
+	err = os.Chmod(endpoint, 0o600) //nolint:gomnd
+	return l, err
+}
+
+func UnixDial(ctx context.Context, endpoint string) (net.Conn, error) {
```
joshklop, line 44: This is only used in tests. Can we move it to the `jsonrpc_test` package?

jsonrpc/ipc_all.go
```diff
@@ -0,0 +1,62 @@
...
+	return new(net.Dialer).DialContext(ctx, "unix", endpoint)
+}
+
+// preparePath ensures the existence of directories leading to provided path
+// and removes existing file if any.
+func preparePath(path string) error {
+	var err error
+	err = os.MkdirAll(filepath.Dir(path), 0o751) //nolint:gomnd
```
joshklop, line 52: What is the reasoning for `0o751`?

joshklop (commented):
Made a first pass. Should Juno be able to communicate over multiple unix sockets, each representing a different IPC connection?

jsonrpc/ipc_all.go
Exca-DK: `751` is a stricter version of `755` which is a common default for directories, eg. ubuntu default homes are of such mode. I wasn't sure between `751` and `750` but decided to go with the first one since it's also used in `go-etherum`.

jsonrpc/ipc_all.go
Exca-DK: It can be moved but it will also need to be compiled differently for windows, `ipc_test_windows.go` and `ipc_test_all.go` is fine?

jsonrpc/ipc.go
```diff
@@ -0,0 +1,200 @@
...
+	for {
+		conn, err := i.listener.Accept()
+		if netutil.IsTemporaryError(err) {
+			i.log.Errorw("Failed to accept connection", "err", err)
+			continue
+		} else if err != nil {
+			return err
+		}
+		i.connWg.Add(1)
+		go i.serveConn(ctx, newIpcConn(conn, i.connParams))
```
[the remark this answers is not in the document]
Exca-DK: Yes, this should be safe I think.
```go
errCh := make(chan error, 1)
go func() {
	errCh <- i.run(ctx)
}()

select {
case <-ctx.Done():
case err := <-errCh:
	if !isSocketError(err) {
		i.log.Warnw("Listener failure", "err", err)
	}
}
return i.Stop()
```

Exca-DK:
> Made a first pass. Should Juno be able to communicate over multiple unix sockets, each representing a different IPC connection?

Spawning a socket for each connection is overkill I believe, one is enough. Spawning multiple sockets can be considered for keeping multiple `jsonrpc.server` to align with current http and ws rpc (`jsonrpcServerLegacy` and `jsonrpcServer`) behavior. The problem here is that socket creation can fail in either of those and I didn't know what expected behavior would be in that case so I just spawned the latest one.

jsonrpc/ipc_all.go
joshklop: Thanks for the explanation, sounds good to me. Can you please summarize the reasoning in a comment for future reference? Extracting it out to a constant instead of using `nolint` would be nice as well.

joshklop:
> Spawning a socket for each connection is overkill I believe, one is enough.

Understood, thanks.

> The problem here is that socket creation can fail in either of those and I didn't know what expected behavior would be in that case so I just spawned the latest one.

That's ok, creating TCP sockets can also fail. Let's maintain consistency with websocket and http handlers, and spawn a socket for each server.

jsonrpc/ipc.go
```diff
@@ -0,0 +1,200 @@
...
+}
+
+// WithListener registers a NewRequestListener
+func (i *Ipc) WithListener(listener NewRequestListener) *Ipc {
+	i.events = listener
+	return i
+}
+
+// Run starts the IPC handler in a goroutine and returns any errors encountered during
+// handling. It can be stopped by canceling the provided context or by
```
joshklop, line 63: We can un-export `Stop` and require the caller to use the context. Thoughts?

jsonrpc/ipc.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+	return err
+}
+
+// run continuously accepts and serves incoming connections.
+func (i *Ipc) run(ctx context.Context) error {
+	for {
+		conn, err := i.listener.Accept()
+		if netutil.IsTemporaryError(err) {
+			i.log.Errorw("Failed to accept connection", "err", err)
```
joshklop, line 87: nit: we can demote this to a `Warnw`.

jsonrpc/ipc.go
```diff
@@ -0,0 +1,200 @@
...
+		if err != nil {
+			break
+		}
+	}
+
+	if isSocketError(err) || isPipeError(err) {
+		return
+	}
+
+	i.log.Warnw("Closing unix connection due to internal error", "err", err)
```
joshklop, line 142: We should avoid `unix` language in code that compiles on Windows.

jsonrpc/ipc.go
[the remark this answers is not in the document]
joshklop: I think that works.

What do you think about using a [conc](https://github.com/sourcegraph/conc) waitgroup instead of `sync.WaitGroup`? It's a very small dependency and makes these cases a lot easier to reason about.

jsonrpc/ipc_all.go
joshklop: Hmm, I see `IpcDial` in `ipc_windows.go`. Assuming we don't change anything, does that need to be renamed to `UnixDial` for the tests to build on windows?

Exca-DK: yeah, my bad, forgot about that one when doing a naming refactor from unix to ipc .

jsonrpc/ipc.go
Exca-DK: Sure.

jsonrpc/ipc_all.go
joshklop: No worries, just wanted to make sure I was understanding. To be honest, we won't be able to guarantee any level of support for the windows code, so we might as well drop it. That should simplify things. Wdyt?

Exca-DK: Lets drop it then

joshklop:
No worries. It's flaky sometimes. Thanks for fixing.

jsonrpc/ipc.go
joshklop: Looks like it's still an error

jsonrpc/ipc.go
```diff
@@ -0,0 +1,221 @@
...
+
+// run continuously accepts and serves incoming connections.
+func (i *Ipc) run(ctx context.Context) error {
+	for {
+		conn, err := i.listener.Accept()
+		if netutil.IsTemporaryError(err) {
+			i.log.Errorw("Failed to accept connection", "err", err)
+			continue
+		} else if err != nil {
+			return err
```
joshklop, line 111:
```suggestion
			return fmt.Errorf("accept connection: %v", err)
```

jsonrpc/ipc.go
```diff
@@ -0,0 +1,221 @@
...
+		i.connWg.Done()
+	}
+}
+
+// serveConn handles incoming connection.
+func (i *Ipc) serveConn(ctx context.Context, conn net.Conn) {
+	defer i.cleanupConn(conn)
+	if !i.setupConn(conn) {
+		return
+	}
```
joshklop, line 148: When will this execute?

jsonrpc/ipc.go
```diff
@@ -0,0 +1,221 @@
...
+
+	var err error
+	for err == nil {
+		i.events.OnNewRequest("any")
+		err = i.rpc.HandleReadWriter(ctx, conn)
+		if err != nil {
```
joshklop, line 154:
```suggestion
		if err = i.rpc.HandleReadWriter(ctx, conn); err != nil {
```

jsonrpc/ipc.go
```diff
@@ -0,0 +1,221 @@
...
+	delete(i.conns, conn)
+	if ok {
+		conn.Close()
+		i.connWg.Done()
+	}
+}
+
+// serveConn handles incoming connection.
+func (i *Ipc) serveConn(ctx context.Context, conn net.Conn) {
+	defer i.cleanupConn(conn)
```
joshklop, line 145:
```suggestion
	defer func() {
	        i.mu.Lock()
	        defer i.mu.Unlock()
	        i.cleanupConnNoLock(conn)
	}()
```
Let's remove some indirection.

jsonrpc/ipc.go
```diff
@@ -0,0 +1,221 @@
...
+	if !i.setupConn(conn) {
```
joshklop, line 146: can we inline the call to `setupConn` here? Looks like this is the only place it's used.

jsonrpc/ipc.go
```diff
@@ -0,0 +1,221 @@
...
+	if err := i.initialiseSocket(); err != nil {
+		return err
+	}
+
+	var (
+		errCh = make(chan error, 1)
+		wg    conc.WaitGroup
+	)
+	defer wg.Wait()
+	wg.Go(func() { errCh <- i.run(ctx) })
```
joshklop, line 80: nit: can we inline `run` here?

jsonrpc/ipc.go
```diff
@@ -0,0 +1,221 @@
...
+func (i *Ipc) WithListener(listener NewRequestListener) *Ipc {
+	i.events = listener
+	return i
+}
+
+// Run launches the IPC handler and handles any potential errors.
+// It is the caller's responsibility to cancel the provided context when they wish to
+// gracefully shut down the IPC handler.
+func (i *Ipc) Run(ctx context.Context) error {
+	if err := i.initialiseSocket(); err != nil {
```
joshklop, line 71: If this is the only place that `initialiseSocket` is used, let's inline it.

jsonrpc/ipc.go
```diff
@@ -0,0 +1,221 @@
...
+		log:            log,
+		connParams:     DefaultIpcConnParams(),
+		conns:          make(map[net.Conn]struct{}),
+		events:         &SelectiveListener{},
+		socketCallback: func() {},
+	}
+}
+
+// WithSocketCallback sets the callback for socket creation.
+func (i *Ipc) WithSocketCallback(f func()) *Ipc {
```
joshklop, line 50: Is this mainly for avoiding race conditions in tests?

jsonrpc/ipc_test.go
```diff
@@ -0,0 +1,260 @@
...
+
+	"github.com/NethermindEth/juno/jsonrpc"
+	"github.com/NethermindEth/juno/utils"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func unixDial(ctx context.Context, endpoint string) (net.Conn, error) {
+	return new(net.Dialer).DialContext(ctx, "unix", endpoint)
+}
```
joshklop, line 22:
```suggestion
func unixDial(t *testing.T, ctx context.Context, endpoint string) net.Conn {
	c, err := new(net.Dialer).DialContext(ctx, "unix", endpoint)
	require.NoError(t, err)
	return c
}
```
This may simplify some of the tests.

edit: I see there are asserts for specific errors. Maybe those cases could forgo the helper and call `DialContext` directly?

node/http.go
```diff
@@ -7,6 +7,7 @@ import (
 	"net"
 	"net/http"
 	"net/http/pprof"
+	gopath "path"
```
joshklop, line 10: Let's be safe and use `path/filepath`.

node/http.go
```diff
@@ -112,6 +113,75 @@ func makeRPCOverWebsocket(host string, port uint16, servers map[string]*jsonrpc.
...
+		trimmed, ok := strings.CutPrefix(path, "/")
+		if !ok {
+			continue
+		}
+		var endpoint string
+		if trimmed == "" {
+			endpoint = gopath.Join(dir, "juno.ipc")
+		} else {
+			endpoint = gopath.Join(dir, fmt.Sprintf("juno_%s.ipc", trimmed))
+		}
```
joshklop, line 175:
```suggestion
		endpoint := gopath.Join(dir, fmt.Sprintf("juno_%s.ipc", trimmed))
		if trimmed == "" {
			endpoint = gopath.Join(dir, "juno.ipc")
		}
```

node/http.go
```diff
@@ -112,6 +113,75 @@ func makeRPCOverWebsocket(host string, port uint16, servers map[string]*jsonrpc.
 	return makeHTTPService(host, port, cors.Default().Handler(mux))
 }
 
+type ipcService struct {
```
joshklop, line 116: Can we move the ipc things to `node/ipc.go`? They're technically not http-related.

jsonrpc/ipc.go
Exca-DK: When the handler is stopping and there are still pending connections yet to be scheduled.

jsonrpc/ipc.go
Exca-DK: Yes. It used to be at initialization of handler instead of startup but I changed it in order to follow the http and ws services pattern at `node/http`.

node/http.go
Exca-DK: Websocket and grpc is being kept there too. Maybe `node/ipc.go` > `node/rpc.go` is better solution? It would align with the way they are instantiated.  
```go
if cfg.HTTP {
	services = append(services, makeRPCOverHTTP(cfg.HTTPHost, cfg.HTTPPort, rpcServers, log, cfg.Metrics))
}
if cfg.Websocket {
	services = append(services, makeRPCOverWebsocket(cfg.WebsocketHost, cfg.WebsocketPort, rpcServers, log, cfg.Metrics))
}
if cfg.IPC {
	services = append(services, makeRPCOverIpc(cfg.IPCPath, rpcServers, log, cfg.Metrics))
}
```

jsonrpc/ipc.go
Exca-DK: To be more precise, There can be some connections yet to be scheduled from `go i.serveConn(ctx, newIpcConn(conn, i.connParams))` since closing socket doesn't close the connections.
```go
**// Close stops listening on the Unix address. Already accepted
// connections are not closed.
func (l *UnixListener) Close() error {**
```

node/http.go
joshklop: > Websocket and grpc is being kept there too.

Both of those protocols use http, but I see your point. Let's leave the `ipcService` where it is for now.

> Maybe node/ipc.go > node/rpc.go is better solution? It would align with the way they are instantiated.

I like that idea. Let's leave it out of scope for this PR.

jsonrpc/ipc.go
Exca-DK: There is no guarantee that the `HandleReadWriter` will error. Already accepted connection is perfectly fine when `net.Listener` is closed which means that reads/writes are going to work.

jsonrpc/ipc.go
Exca-DK: Seems like we both missed the loop condition :smile:

jsonrpc/ipc.go
Exca-DK: Sorry, should be fine now.

jsonrpc/ipc_all.go
```diff
@@ -0,0 +1,61 @@
...
+// preparePath ensures the existence of directories leading to provided path
+// and removes existing file if any.
+func preparePath(path string) error {
+	var err error
+	err = os.MkdirAll(filepath.Dir(path), fileModePerms)
+	if err != nil {
+		return err
+	}
+	// Remove any existing file at the specified path.
+	err = os.Remove(path)
```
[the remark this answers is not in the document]
joshklop: [Here](https://github.com/tailscale/tailscale/blob/e78cb9aeb3b32ee97991710c72eb979ebad61477/safesocket/unixsocket.go#L27-L45) is an alternative approach that seems to achieve the best of both worlds: if the socket exists and is in use, shut down with an error, otherwise use the existing socket. Thoughts?

jsonrpc/ipc.go
joshklop: I see, thanks for clarifying. 

Even still, `closing` may become `true` right after this condition fails, so we may accept the connection anyways. If that's correct, can we remove the `closing` field entirely?

jsonrpc/ipc.go
joshklop: What do you think about replacing `NewIpc` with `DialIPC`, which does everything `NewIpc` does plus setting up the listener? That way we know the listener is initialized when we call `Run`.

jsonrpc/ipc.go
```diff
@@ -0,0 +1,198 @@
...
+				return
+			}
+			i.connWg.Add(1)
+			go i.serveConn(ctx, newIpcConn(conn, i.connParams))
+		}
+	})
+
+	select {
+	case <-ctx.Done():
+	case <-ch:
```
joshklop, line 100: Rather than managing a separate `ch` channel, can we use a child context of `ctx`?

node/http.go
```diff
@@ -112,6 +113,73 @@ func makeRPCOverWebsocket(host string, port uint16, servers map[string]*jsonrpc.
...
+// It takes context as a parameter to control the execution and returns an error if any handler fails.
+func (ipc *ipcService) Run(ctx context.Context) error {
+	errCh := make(chan error, len(ipc.handlers))
+	defer close(errCh)
+
+	var wg conc.WaitGroup
+	defer wg.Wait()
+
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
```
joshklop, line 132: If we use `context.WithCancelCause(ctx)` here, I think we can get rid of the error channel.

Instead of `if err := handler.Run(ctx); ... { errCh <- err } ` we use `cancel(handler.Run(ctx))`.

Instead of the `select` at the end of the function we can do something like `return context.Cause(ctx)`.

jsonrpc/ipc.go
Exca-DK: If it becomes true afterwards then stop will close that connection. Stop takes over lock and closes every connection, also marks that field as true. After the lock is released every new connection is ensured to see true. That also means that every connection before stop will be closed by that or either when connection errors (the cleanup thats deferred also takes the same lock and closes if its still there)

joshklop: > If it becomes true afterwards then stop will close that connection.

Right, this is why I think we are better off removing the `i.closing` condition anyway. You mention that stopping the connection while we are reading from it causes an error. Will that be caught at the `if isSocketErr(err)...` line after the `for` loop? If so, I don't see any downsides to removing the `i.closing` field.

jsonrpc/ipc.go
Exca-DK: sure

jsonrpc/ipc.go
Exca-DK: I don't mind reverting that change. As I said I changed it just to allign more or less with the `node/http.go` . (They construct listeneres at startup and initialization of struct never fails)
Regarding the DialIPC I think that it doesn't reflect the behaviour and can be too confusing. The DialIPC might even be required to implement the part you said above about checking if socket is alive

jsonrpc/ipc.go
Exca-DK: Yes, that will be caught by that. What I am trying to say is that there can be such a case:
1. It closes the listener which rejects all of the new connections
2. Some goroutines are yet to be scheduled from go runtime (go serveConn can be executed some time in the future)

Exca-DK: The i.stopping is actually stopped signal meaning that the process of closing and removing all connections has already ended. In that case that connection will never be closed untill it errors out but it will only error out when either  juno encounters error internally on some rpc method or when client disconnects.

node/http.go
Exca-DK: Thanks for the tip. Wll check that out, never actually used `WithCancelCause` :laughing:

jsonrpc/ipc.go
joshklop: Ah my bad, I misunderstood what you meant by setting up the listener "at the initialization of the handler." In this case, I think we can disregard the convention and initialize the listener in `NewIpc`. I think it's ok if `NewIpc` can return an error.

(apologies if my earlier statement that we should "maintain consistency with websocket and http handlers, and spawn a socket for each server" was was confusing here)

> Regarding the DialIPC I think that it doesn't reflect the behaviour and can be too confusing. 

Makes sense.

> The DialIPC might even be required to implement the part you said above about checking if socket is alive

Mind expanding on this? Feel free to move this part of the conversation to the other comment as well.

jsonrpc/ipc_all.go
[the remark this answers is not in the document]
Exca-DK: Sounds good to me (wasn't aware that dial is able to detect stale socket, thought it will simply connect) expect using the existing socket. It has to be recreated in order to net.Listen.

jsonrpc/ipc.go
Exca-DK: Nevermind it, I thought that some simple ack might be required to check for listener but the dial is actually able to determine listener presence.

jsonrpc/ipc.go
joshklop: Ahhh I see now. That's a tricky edge case.

I think using a child context as proposed in https://github.com/NethermindEth/juno/pull/1362#discussion_r1425831083 can help us here.
```go
	i.mu.Lock()
	i.conns[conn] = struct{}{}
	i.mu.Unlock()
	if ctx.Err() != nil {
		return
	}
```
Right now, there are two ways that `i.stop` is called: `ctx` is canceled or `ch` is closed. If we have a single child context, we simplify the invariant to: `i.stop` is called when `childCtx` is canceled. Since the child context is passed to `serveConn`, we can check it instead of `i.closing`.

Does this sound alright?

jsonrpc/ipc_all.go
```diff
@@ -0,0 +1,69 @@
...
+)
+
+var errPathTooLong = errors.New("path too long")
+
+// createListener creates a Unix domain socket at the given path and sets proper permissions.
+func createListener(endpoint string) (net.Listener, error) {
+	var (
+		l   net.Listener
+		err error
+	)
```
joshklop, line 30: Can we delete these?

jsonrpc/ipc_all.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+	// Try connecting first; if it works, then the socket is still live,
+	// so let's abort the creation of a new one.
+	c, err := net.Dial("unix", endpoint)
+	if err == nil {
+		c.Close()
+		return nil, fmt.Errorf("%v: address already in use", endpoint)
+	}
+	err = preparePath(endpoint)
+	if err != nil {
```
joshklop, line 43:
```suggestion
	if err = preparePath(endpoint); err != nil {
```
Let's inline `preparePath` if it's only used here.

jsonrpc/ipc_all.go
```diff
@@ -0,0 +1,69 @@
...
+		err error
+	)
+	// path + terminator
+	if len(endpoint)+1 > maxUnixPathBytes {
+		return nil, errPathTooLong
+	}
+	// Try connecting first; if it works, then the socket is still live,
+	// so let's abort the creation of a new one.
+	c, err := net.Dial("unix", endpoint)
+	if err == nil {
```
joshklop, line 38:
```suggestion
	if c, err := net.Dial("unix", endpoint); err == nil {
```

jsonrpc/ipc_all.go
```diff
@@ -0,0 +1,69 @@
...
+	maxUnixPathBytes = 108
+	// fileModePerms represents the permission mode for directories, which is set to 751.
+	// This mode is stricter than the common default of 755 for directories.
+	fileModePerms = 0o751
+	// socketPerms represents the permission mode for socket files, set to 600. It allows
+	// read and write access for the owner only, ensuring that socket is accessible only to the intended user.
+	socketPerms = 0o600
+)
+
+var errPathTooLong = errors.New("path too long")
```
joshklop, line 23: We can just return this from the one place it's used.

jsonrpc/ipc.go
```diff
@@ -0,0 +1,178 @@
...
+	"github.com/ethereum/go-ethereum/p2p/netutil"
+	"github.com/sourcegraph/conc"
+)
+
+type Ipc struct {
+	rpc    *Server
+	events NewRequestListener
+	log    utils.SimpleLogger
+
+	connWg sync.WaitGroup // connWg is a WaitGroup for tracking active connections.
```
joshklop, line 22: nit: let's use a `conc.WaitGroup` here as well.

jsonrpc/ipc_all.go
joshklop, line 1: Remind me: is there an advantage to keeping `ipc_all.go` separate from `ipc.go` if we don't support windows? If not, let's combine them.

node/http.go
```diff
@@ -112,6 +113,70 @@ func makeRPCOverWebsocket(host string, port uint16, servers map[string]*jsonrpc.
...
+	handlers []*jsonrpc.Ipc
+}
+
+var _ service.Service = (*ipcService)(nil)
+
+// Run is a method of the ipcService struct that starts the execution of IPC handlers in parallel.
+// It takes context as a parameter to control the execution and returns an error if any handler fails.
+func (ipc *ipcService) Run(ctx context.Context) error {
+	errCh := make(chan error, len(ipc.handlers))
+	defer close(errCh)
```
joshklop, line 126: unused

node/http.go
```diff
@@ -112,6 +113,70 @@ func makeRPCOverWebsocket(host string, port uint16, servers map[string]*jsonrpc.
...
+	// Iterate over all registered handlers and execute them concurrently.
+	for i := range ipc.handlers {
+		handler := ipc.handlers[i]
+		wg.Go(func() {
+			cancel(handler.Run(ctx))
+		})
+	}
+
+	<-ctx.Done()
+	return context.Cause(ctx)
```
joshklop, line 142: I realized may need to do a tad more:
```go
	if err := context.Cause(ctx); err == ctx.Err() {
		return nil
	} else {
		return err
	}
```
This way we don't return spurious `context.Canceled`s and flood the logs with errors when a user types Ctrl-c.

node/http.go
```diff
@@ -112,6 +113,70 @@ func makeRPCOverWebsocket(host string, port uint16, servers map[string]*jsonrpc.
...
+		if trimmed == "" {
+			endpoint = filepath.Join(dir, "juno.ipc")
+		}
+		ipcHandler, err := jsonrpc.NewIpc(server, endpoint, log)
+		if err != nil {
+			// ugly hack to not leave stale sockets.
+			ctx, cancel := context.WithCancel(context.Background())
+			cancel()
+			ipcHandler.Run(ctx) //nolint:errcheck
+			return nil, err
```
joshklop, line 170: Is this necessary?

jsonrpc/ipc_test.go
```diff
@@ -0,0 +1,249 @@
...
+		}
+
+		for i := 0; i < items; i++ {
+			select {
+			case <-ctx.Done():
+				t.Fatal(ctx.Err())
+			case err := <-errCh:
+				if err != nil {
+					t.Fatal(err)
+				}
```
joshklop, line 146:
```suggestion
        require.NoError(t, err)
```

jsonrpc/ipc_test.go
```diff
@@ -0,0 +1,249 @@
...
+		for i := 0; i < items; i++ {
+			conn, err := new(net.Dialer).DialContext(ctx, "unix", ep)
+			assert.NoError(t, err)
+			conns[i] = conn
+			go func() {
+				errCh <- exchangeMsg(conn, msg, want)
+			}()
+		}
+
+		for i := 0; i < items; i++ {
```
joshklop, line 139: ~Do we need the for loop?~ I think I get it now. `exchangeMsg` will return `nil` on `errCh` for each item. Is that right?

jsonrpc/ipc_test.go
```diff
@@ -0,0 +1,249 @@
...
+			select {
+			case <-ctx.Done():
+				t.Fatal(ctx.Err())
```
joshklop, line 142: `require.NoError`

jsonrpc/ipc.go
```diff
@@ -0,0 +1,178 @@
...
+		defer cancel()
+		for {
+			conn, err := i.listener.Accept()
+			if netutil.IsTemporaryError(err) {
+				i.log.Warnw("Failed to accept connection", "err", err)
+				continue
+			} else if err != nil {
+				i.log.Warnw("Accept connection", "err", err)
+				return
+			}
```
joshklop, line 80: This warning gets logged on a simple ctrl-c. Is that avoidable?
```
21:58:13.023 14/12/2023 -06:00	WARN	jsonrpc/ipc.go:78	Accept connection	{"err": "accept unix /home/user/repos/juno/juno/juno_v0_5.ipc: use of closed network connection"}
21:58:13.023 14/12/2023 -06:00	WARN	jsonrpc/ipc.go:78	Accept connection	{"err": "accept unix /home/user/repos/juno/juno/juno_v0_4.ipc: use of closed network connection"}
21:58:13.023 14/12/2023 -06:00	WARN	jsonrpc/ipc.go:78	Accept connection	{"err": "accept unix /home/user/repos/juno/juno/juno.ipc: use of closed network connection"}
```

joshklop (approved):
Looks great! Works perfectly on my machine. Just a few minor things left.

Mind if a team member rebases and/or makes small adjustments when we're ready to merge? I'm not sure if we'll fit it into a release right away.

jsonrpc/ipc.go
joshklop: Preferably we avoid this somehow. If it's too difficult, we can demote it to a debug log.

Exca-DK: Sure, go ahead for the adjustment/rebase. Regarding the changes I'll prolly take a look at it later today.

node/http.go
Exca-DK: Yes, there could be other sockets spawned before that error and they should be cleaned up.

jsonrpc/ipc_test.go
Exca-DK: Yes. It's ping-pong test for multiple connections.

node/http.go
joshklop: Since we clean up stale sockets on startup, do you think it's ok to delete this code?

Exca-DK: The cleanup at startup is racy racy  since it's the same pattern as used [here](https://github.com/tailscale/tailscale/blob/e78cb9aeb3b32ee97991710c72eb979ebad61477/safesocket/unixsocket.go#L27-L45)

It's always better to ensure that resources are freed by the same process that's taking them.

node/http.go
Exca-DK: Why `err == ctx.Err()`  instead of `errors.Is` ?

joshklop: No particular reason. Both should work since `context.Cause` is documented to return the same as `ctx.Err()` when the cause is `nil` or when a parent was canceled.

node/http.go
joshklop: Does that mean we should delete the endpoints for the other `srv.handlers` as well?

Exca-DK: Yes and that's what's being done by those few lines.

PR #1363, opened 2023-10-24
Parallelize recursive calls in Trie.updateValueIfDirty

### LINKED ISSUE #904: Compare RWMutex vs BufferedTransaction performance
Right now we use a buffered transaction in the contract storage update. We could use an RWMutex instead. We should investigate which is faster.

See [here](https://github.com/NethermindEth/juno/pull/900#discussion_r1258654639) for reference:

> I expected a lock to have a worse performance impact compared to memory allocations that we need for buffering.
But that is just my gut feeling. We can keep experimenting after merging this.
But regardless, we will have to implement a new type that wraps a transaction and protects it with a RWMutex.

cc @omerfirmak

### LINKED ISSUE #1096: Parallelize recursive calls in Trie.updateValueIfDirty
https://github.com/NethermindEth/juno/blob/9fbb30c23e63122c490daf24f22ec3d0177b1442/core/trie/trie.go#L335-L344
These two recursive calls can be executed in parallel

omerfirmak:
Closes #1096 
Closes #904 

Significant speed up in state updates on my local machine.

Before
```BenchmarkMainnetBlock-16    	      10	 647261272 ns/op	16582017 B/op	  205653 allocs/op```

After
```BenchmarkMainnetBlock-16    	      10	 389617332 ns/op	17668045 B/op	  229347 allocs/op```

### DIFF
--- a/core/trie/trie.go
+++ b/core/trie/trie.go
@@ -12,17 +12,6 @@ import (
 	"github.com/NethermindEth/juno/db"
 )
 
-// Storage is the Persistent storage for the [Trie]
-type Storage interface {
-	Put(key *Key, value *Node) error
-	Get(key *Key) (*Node, error)
-	Delete(key *Key) error
-
-	PutRootKey(newRootKey *Key) error
-	RootKey() (*Key, error)
-	DeleteRootKey() error
-}
-
 type hashFunc func(*felt.Felt, *felt.Felt) *felt.Felt
 
 // Trie is a dense Merkle Patricia Trie (i.e., all internal nodes have two children).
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Review by joshklop — APPROVED** [review]
Why is the SyncTransaction better in this case?

**Comment by omerfirmak** [comment]
> Why is the SyncTransaction better than BufferedTransaction in this case?

Because it avoids triggering multiple allocations for updating the same key.
Assume we used `BufferedTransaction` in `Trie` instead of `SyncTransaction`

at some point we will call `BufferedTransaction.Commit()` which will commit the changes to the original txn. That original transaction is another `BufferedTransaction` that we created in `State.updateContractStorages`. So we are commiting a BufferedTransaction to another BufferedTransaction, which creates additional allocations that we can avoid easily.

File: core/trie/trie.go [hunk 1]
```diff
@@ -6,23 +6,13 @@ import (
...
-type Storage interface {
-	Put(key *Key, value *Node) error
-	Get(key *Key) (*Node, error)
-	Delete(key *Key) error
-
-	PutRootKey(newRootKey *Key) error
-	RootKey() (*Key, error)
-	DeleteRootKey() error
-}
-
```
IronGauntlets, line 25: Can you please explain why this is being removed?

joshklop: From the commit message:

> It wasn't utilized anyways, no need to pay the virtualization cost

PR #1373, opened 2023-10-25
Move subscription to synchronizer

joshklop:


sync/sync.go
```diff
@@ -286,12 +296,13 @@ func (s *Synchronizer) revertHead(forkBlock *core.Block) {
...
 	s.log.Infow("Reorg detected", "localHead", localHead, "forkHead", forkBlock.Hash)
 
-	err = s.blockchain.RevertHead()
+	header, err := s.blockchain.RevertHead()
 	if err != nil {
 		s.log.Warnw("Failed reverting HEAD", "reverted", localHead, "err", err)
 	} else {
 		s.log.Infow("Reverted HEAD", "reverted", localHead)
 	}
+	s.newHeads.Send(header)
```
omerfirmak, line 305: I thought we decided not to send reverted blocks?

sync/sync.go
```diff
@@ -292,6 +302,9 @@ func (s *Synchronizer) revertHead(forkBlock *core.Block) {
 	} else {
 		s.log.Infow("Reverted HEAD", "reverted", localHead)
 	}
+	if newHead, err := s.blockchain.HeadsHeader(); err == nil {
+		s.newHeads.Send(newHead)
+	}
```
omerfirmak, line 307: We didnt do this in the original implementation right? We only sent notifications when head changed in the forward direction. I dont think ethereum does this either. https://docs.alchemy.com/reference/newheads

> When a chain reorganization occurs, this subscription will emit an event containing all new headers (blocks) for the new chain. This means that you may see multiple headers emitted with the same height (block number), and when this happens the later (highest) block number should be taken as the correct one after a reorganization.

joshklop: Ohhh you're right. I misinterpreted the `newHeads` docs as meaning "an update is sent whenever the head changes" rather than "an update is sent whenever a new head is added to the chain".

I think it's fixed now.

omerfirmak (approved):

Issue #1349, opened 2023-10-17
RPC should not be able to starve Sync process


omerfirmak:
When under heavy RPC load, sync can fall behind due to not enough CPU resources being left.

Exca-DK:
If it's not taken I could work on it.

omerfirmak:
Somewhat addressed by #1378

PR #1378, opened 2023-10-26
Throttle number of VM instances RPC can utilize concurrently 

omerfirmak:
Towards #1349

We should add metrics as well If this turns out to be useful

### DIFF
--- a/cmd/juno/juno.go
+++ b/cmd/juno/juno.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"os/signal"
 	"path/filepath"
+	"runtime"
 	"syscall"
 	"time"
 
@@ -54,6 +55,7 @@ const (
 	grpcF                = "grpc"
 	grpcHostF            = "grpc-host"
 	grpcPortF            = "grpc-port"
+	maxVMsF              = "max-vms"
 
 	defaultConfig              = ""
 	defaulHost                 = "localhost"
@@ -100,6 +102,7 @@ const (
 	grpcUsage                = "Enable the HTTP GRPC server on the default port."
 	grpcHostUsage            = "The interface on which the GRPC server will listen for requests."
 	grpcPortUsage            = "The port on which the GRPC server will listen for requests."
+	maxVMsUsage              = "Maximum number for VM instances to be used for RPC calls concurrently"
 )
 
 var Version string
@@ -191,6 +194,7 @@ func NewCmd(config *node.Config, run func(*cobra.Command, []string) error) *cobr
 	// may mutate their values.
 	defaultLogLevel := utils.INFO
 	defaultNetwork := utils.MAINNET
+	defaultMaxVMs := runtime.GOMAXPROCS(0)
 
 	junoCmd.Flags().StringVar(&cfgFile, configF, defaultConfig, configFlagUsage)
 	junoCmd.Flags().Var(&defaultLogLevel, logLevelF, logLevelFlagUsage)
@@ -217,6 +221,7 @@ func NewCmd(config *node.Config, run func(*cobra.Command, []string) error) *cobr
 	junoCmd.Flags().Bool(grpcF, defaultGRPC, grpcUsage)
 	junoCmd.Flags().String(grpcHostF, defaulHost, grpcHostUsage)
 	junoCmd.Flags().Uint16(grpcPortF, defaultGRPCPort, grpcPortUsage)
+	junoCmd.Flags().Uint(maxVMsF, uint(defaultMaxVMs), maxVMsUsage)
 
 	return junoCmd
 }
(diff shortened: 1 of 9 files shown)


rpc/handlers.go
```diff
@@ -1149,6 +1152,12 @@ func makeContractError(err error) *jsonrpc.Error {
 	return &contractErr
 }
 
+func makeUnexpectedError(err error) *jsonrpc.Error {
```
joshklop, line 1155: We could do this for any `*jsonrpc.Error` that needs a data field.

IronGauntlets (changes requested):

IronGauntlets:
Apologies, I meant to "comment" instead of "request changes"

cmd/juno/juno.go
```diff
@@ -217,6 +220,7 @@ func NewCmd(config *node.Config, run func(*cobra.Command, []string) error) *cobr
 	junoCmd.Flags().Bool(grpcF, defaultGRPC, grpcUsage)
 	junoCmd.Flags().String(grpcHostF, defaulHost, grpcHostUsage)
 	junoCmd.Flags().Uint16(grpcPortF, defaultGRPCPort, grpcPortUsage)
+	junoCmd.Flags().Uint(rpcVMMaxConcurrencyF, uint(runtime.GOMAXPROCS(0)), rpcVMMaxConcurrencyUsage)
```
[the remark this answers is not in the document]
omerfirmak: I wanted to do that but the problem is the default is not a constant.

utils/throttler.go
```diff
@@ -0,0 +1,51 @@
...
+	queueLen := t.queue.Add(1)
+	if t.maxQueueLen > 0 && queueLen > t.maxQueueLen {
+		t.queue.Add(-1)
+		return ErrResourceBusy
+	}
+	t.sem <- struct{}{}
+	defer func() {
+		<-t.sem
+	}()
+	t.queue.Add(-1)
```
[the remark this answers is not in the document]
omerfirmak: `queue` is the number of callers that are blocked on this resource. So as soon as they acquire the semaphore and start using the resource (calling the actual vm in this case) they are marked as "off the queue" by subtracting 1 from `queue`.

cmd/juno/juno.go
```diff
@@ -54,6 +55,7 @@ const (
 	grpcF                = "grpc"
 	grpcHostF            = "grpc-host"
 	grpcPortF            = "grpc-port"
+	rpcVMMaxConcurrencyF = "rpc-vm-max-concurrency"
```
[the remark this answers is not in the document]
omerfirmak: done

rpc/handlers.go
omerfirmak: See the last commit, let me know what you think.

joshklop: LGTM.

cmd/juno/juno.go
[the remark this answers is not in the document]
IronGauntlets: Hmm, how about doing something similar for `defaultLogLevel` and `defaultnetwork` as on lines 195 and 196.

jsonrpc/server.go
```diff
@@ -53,6 +53,13 @@ type Error struct {
 	Data    any    `json:"data,omitempty"`
 }
 
+// DuplicateWithData copies the error and sets the data field on the copy
+func (e *Error) DuplicateWithData(data any) *Error {
```
IronGauntlets, line 57: nit: Consider renaming it to `AddData()`

utils/throttler.go
```diff
@@ -0,0 +1,51 @@
...
+)
+
+var ErrResourceBusy = errors.New("resource busy, try again")
+
+type Throttler[T any] struct {
+	resource *T
+	sem      chan struct{}
+	queue    atomic.Int32
+
+	maxQueueLen int32
```
IronGauntlets, line 16: Do you think `Unit32` would be more appropriate?

IronGauntlets (commented):
Left some comments for you to decide upon. Otherwise LGTM!

IronGauntlets (approved):

cmd/juno/juno.go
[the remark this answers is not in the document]
omerfirmak: done

jsonrpc/server.go
omerfirmak: I want to emphasize that it copies the error before adding the data. Will rename it to `CloneWithData()`

utils/throttler.go
omerfirmak: Yeah, I don't like that I had to use signed types there either. I wanted to keep `queue` and `maxQueueLen` same types to avoid running in to weird issues when comparing them due to casting one of them to the other type. And `queue` is a signed type to be able to do `t.queue.Add(-1)`.

PR #1380, opened 2023-10-26
rpc: Return error on tx revert in SimulateTransactions

### LINKED ISSUE #1108: `SimulateTransactions` should fail if one of the txns gets reverted

Exca-DK:
The third time is a charm :sweat_smile:

closes #1108


vm/vm.go
```diff
@@ -95,6 +97,12 @@ func JunoAppendActualFee(readerHandle C.uintptr_t, ptr unsafe.Pointer) {
 	context.actualFees = append(context.actualFees, makeFeltFromPtr(ptr))
 }
 
+//export JunoAppendRevert
+func JunoAppendRevert(readerHandle C.uintptr_t, str *C.char) {
+	context := unwrapContext(readerHandle)
+	context.reverts = append(context.reverts, C.GoString(str))
+}
+
```
omerfirmak, line 105: Revert reasons are already available within the trace, so returning them separately is a bit overkill. Instead of returning the reason as a whole, we can get away with returning just the transaction indexes that are failing.

vm/vm.go
```diff
@@ -209,10 +217,10 @@ func (v *vm) Execute(txns []core.Transaction, declaredClasses []core.Class, bloc
 	C.free(unsafe.Pointer(chainID))
 
 	if len(context.err) > 0 {
-		return nil, nil, errors.New(context.err)
+		return nil, nil, nil, errors.New(context.err)
 	}
 
-	return context.actualFees, context.traces, nil
+	return context.actualFees, context.traces, context.reverts, nil
```
omerfirmak, line 223: instead of adding more return values to `Execute()`, let's define a new custom error type and return that instead.
If needed, rpc.Handler or whoever is calling the VM can check for that specific error type instead of treating all errors the same.

omerfirmak:
Sorry to break it to you but this behaviour is no longer needed, it was decided upon to just return the trace for a reverted transaction.

omerfirmak:
Looks like I talked too soon :)

omerfirmak:
Sorry, but things have changes since you implemented this, so I have to ask another rewrite :/
We currently associate errors with transaction indexes, so that caller knows which txn failed in `cairoVmExecute()`. We can use this existing infrastructure to implement the same things for reverted transactions as well.

https://github.com/NethermindEth/juno/blob/3248d599bc714785dd72074434e8a1cfc6190d61/vm/vm.go#L76-L80

Let's add a new flag to `cairoVmExecute()` that makes it treat reverted txns as errors. It could be called `errorOnRevert`. So instead of reporting it separately as a reverted txn like you do [here](https://github.com/NethermindEth/juno/pull/1380/files#diff-8d744605e0447372940512f1f0ce8f36115ad8adf80941e0eb201bb7697d76b8R262-R267), we will report it like a regular error similar to how we do it [here](https://github.com/NethermindEth/juno/blob/3248d599bc714785dd72074434e8a1cfc6190d61/vm/rust/src/lib.rs#L273-L285) and return early from the `cairoVmExecute()`.

`errorOnRevert` should be set to true only if we are running a txn for fee estimation purposes. Otherwise it should be set to false.

omerfirmak:
Had to go ahead and implement it myself due to time constraints. 
Please feel free to take a look and comment https://github.com/NethermindEth/juno/pull/1518

PR #1384, opened 2023-10-26
rpc: configurable maximum number of block scans

### LINKED ISSUE #1355: Make maximum number of blocks scanned in a single call to `getEvents` configurable
Currently, Juno will happily scan thousands of blocks trying to find an event that matches the filter. The request is terminated early only if the number of events accumulated reaches a limit provided by the user.

This is a pretty easy to execute DoS attack. With a single RPC request, with a filter that does not match any events, one can make Juno scan the entire header chain.

We should make Juno only scan maximum of N blocks before responding to the request with a continuation token.

Exca-DK:
closes #1355


blockchain/event_filter.go
```diff
@@ -149,7 +162,13 @@ func (e *EventFilter) Events(cToken *ContinuationToken, chunkSize uint64) ([]*Fi
...
 		}
 	}
-	return matchedEvents, nil, nil
+
+	// Create a continuation token if there are remaining blocks
+	var rToken *ContinuationToken
+	if remainingScannedBlocks == 0 {
+		rToken = &ContinuationToken{fromBlock: curBlock, processedEvents: processedEvents}
+	}
+	return matchedEvents, rToken, nil
```
omerfirmak, line 171: We can consolidate the early return above in to this as well. They are essentially returning the same thing.
Let's keep this one and remove the previous one.

cmd/juno/juno.go
```diff
@@ -100,6 +101,7 @@ const (
 	grpcUsage                = "Enable the HTTP GRPC server on the default port."
 	grpcHostUsage            = "The interface on which the GRPC server will listen for requests."
 	grpcPortUsage            = "The port on which the GRPC server will listen for requests."
+	rpcMaxBlockScanUsage     = "Maximum number of blocks scanned in single call"
```
omerfirmak, line 104:
```suggestion
	rpcMaxBlockScanUsage     = "Maximum number of blocks scanned in single starknet_getEvents call"
```

blockchain/event_filter.go
```diff
@@ -137,19 +152,20 @@ func (e *EventFilter) Events(cToken *ContinuationToken, chunkSize uint64) ([]*Fi
...
-				}, nil
+				rToken = &ContinuationToken{fromBlock: curBlock, processedEvents: processedEvents}
+				break
 			}
 			return nil, nil, err
 		}
 	}
-	return matchedEvents, nil, nil
+
+	if remainingScannedBlocks == 0 {
```
omerfirmak, line 165:
```suggestion
	if remainingScannedBlocks == 0 && curBlock <= e.toBlock {
```

cmd/juno/juno.go
```diff
@@ -73,6 +74,7 @@ const (
 	defaultMetricsPort         = 9090
 	defaultGRPC                = false
 	defaultGRPCPort            = 6064
+	defaultRPCMaxBlockScan     = 64
```
omerfirmak, line 77: 64 blocks seems a bit too few to me, did you run any tests to see how long 64 blocks take to scan?

Exca-DK: It takes around 10ms on my machine when each block is early skipped due to bloomfilter.

joshklop: drive-by comment: if I'm understanding correctly, this is a significant change to the default behavior. Can we set the default to the max uint64 to minimize breaking changes? Users can explicitly set the limit if they want. I believe geth doesn't set a limit by default, but feel free to fact check me.

Exca-DK: geth doesn't have such option at all. There is an open [issue](https://github.com/ethereum/go-ethereum/issues/24508#issuecomment-1063843495) for that

omerfirmak: > Can we set the default to the max uint64 to minimize breaking changes?

This is not really a breaking change IMO. Users should already be able to handle pagination.

omerfirmak: >  It takes around 10ms on my machine when each block is early skipped due to bloomfilter.

Then, we can bump it to 512 IMO.

omerfirmak:
needs a rebase

rpc/handlers.go
```diff
@@ -126,6 +128,12 @@ func New(bcReader blockchain.Reader, syncReader sync.Reader, n utils.Network,
 	}
 }
 
+// WithFilterLimit sets the maximum number of blocks to scan in a single call for event filtering.
+func (h *Handler) WithFilterLimit(limit uint) *Handler {
```
omerfirmak, line 132: I am not sure about this, should we make `limit` an argument to `New`?

Exca-DK: I can't say for sure, but here's the reasoning as to why I used `With...`

- `limit`  sounds to me like an optional value rather than a core component needed for initialization.

- `New` already contains a significant number of values being passed in.

omerfirmak: Ok, let's follow @joshklop 's advice and make this disabled by default by setting the default value to `math.MaxUint`.
and remove `if h.filterLimit != 0` and `if cfg.RPCMaxBlockScan > 0` checks before calling `WithLimit` on `EventFilter` and `Handler`

omerfirmak (approved):

rpc/handlers_test.go
```diff
@@ -1723,6 +1723,23 @@ func TestEvents(t *testing.T) {
...
 
+	t.Run("filter with limit", func(t *testing.T) {
+		handler = handler.WithFilterLimit(1)
+		key := utils.HexToFelt(t, "0x3774b0545aabb37c45c1eddc6a7dae57de498aae6d5e3589e362d4b4323a533")
+		args.ChunkSize = 100
+		args.Keys = make([][]felt.Felt, 0)
+		args.Keys = append(args.Keys, []felt.Felt{*key})
+		events, err := handler.Events(args)
+		require.Nil(t, err)
+		require.NotEmpty(t, events.ContinuationToken)
```
omerfirmak, line 1734: can we assert the continuation token value? I believe it should be `1-0` ?

omerfirmak (changes requested):

blockchain/event_filter.go
```diff
@@ -137,19 +152,20 @@ func (e *EventFilter) Events(cToken *ContinuationToken, chunkSize uint64) ([]*Fi
...
+				break
 			}
 			return nil, nil, err
 		}
 	}
-	return matchedEvents, nil, nil
+
+	if remainingScannedBlocks == 0 && curBlock <= e.toBlock {
+		rToken = &ContinuationToken{fromBlock: curBlock, processedEvents: processedEvents}
+	}
```
omerfirmak, line 167: Taking another look at this, I don't feel comfortable possibly overwriting the `rToken` set at L158 or using `processedEvents` here since it is returned by appendBlockEvents.

Let's make sure to only set `rToken` here only if it is `nil` and set 0 as its `processedEvents`

PR #1386, opened 2023-10-27
Collect DB latency metrics

omerfirmak:
Should help us compare storages of different machines.

### DIFF
--- a/db/event_listener.go
+++ b/db/event_listener.go
@@ -1,15 +1,17 @@
 package db
 
+import "time"
+
 type EventListener interface {
-	OnIO(write bool)
+	OnIO(write bool, duration time.Duration)
 }
 
 type SelectiveListener struct {
-	OnIOCb func(write bool)
+	OnIOCb func(write bool, duration time.Duration)
 }
 
-func (l *SelectiveListener) OnIO(write bool) {
+func (l *SelectiveListener) OnIO(write bool, duration time.Duration) {
 	if l.OnIOCb != nil {
-		l.OnIOCb(write)
+		l.OnIOCb(write, duration)
 	}
 }
--- a/node/metrics.go
+++ b/node/metrics.go
@@ -1,6 +1,7 @@
 package node
 
 import (
+	"math"
 	"time"
 
 	"github.com/NethermindEth/juno/blockchain"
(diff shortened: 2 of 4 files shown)


node/metrics.go
```diff
@@ -140,7 +160,10 @@ func makeSyncMetrics(syncReader sync.Reader, bcReader blockchain.Reader) sync.Ev
 		OnSyncStepDoneCb: func(op string, blockNum uint64, took time.Duration) {
 			opTimerHistogram.WithLabelValues(op).Observe(took.Seconds())
 			if op == sync.OpStore {
+				fmt.Println("store", blockNum, took)
 				blockCount.Inc()
+			} else if op == sync.OpVerify {
+				fmt.Println("verify", blockNum, took)
```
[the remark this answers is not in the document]
omerfirmak: nope, my bad.

db/event_listener.go
```diff
@@ -1,15 +1,17 @@
 package db
 
+import "time"
+
 type EventListener interface {
-	OnIO(write bool)
+	OnIO(write bool, took time.Duration)
```
IronGauntlets, line 6: `duration` would be better name instead of `took`

omerfirmak: done

node/metrics.go
```diff
@@ -11,22 +12,40 @@ import (
...
+		500,
+		1000, // 1ms
+		2000,
+		3000,
+		4000,
+		5000,
+		10000,
+		50000,
+		math.Inf(0),
+	}
```
IronGauntlets, line 30: Shouldn't buckets be either linear or exponential? this is neither, also before going to infinity, a 1-second bucket would be nice, right?

IronGauntlets (approved):

node/metrics.go
omerfirmak: > Shouldn't buckets be either linear or exponential?

I don't think that is a requirement. But I don't really know if there is a correct set of values. I set these with by trial and error on my local SSD mainly. We can always update them if it is not making sense for the storage options cloud providers are offering. 

> a 1-second bucket would be nice, right?

adding that as well.

Issue #1370, opened 2023-10-25
Sync erroneously emits events even when the step fails


omerfirmak:
https://github.com/NethermindEth/juno/blob/6fc7553478c8cb26508cf2f638e5cf11ba1020fd/sync/sync.go#L160-L173

omerfirmak:
Fixed in #1387

PR #1387, opened 2023-10-28
Fix Event Emission Issue on Sync Step Failure

aminsato:
@omerfirmak 

have a look omer

#1370

### DIFF
--- a/db/event_listener.go
+++ b/db/event_listener.go
@@ -1,15 +1,17 @@
 package db
 
+import "time"
+
 type EventListener interface {
-	OnIO(write bool)
+	OnIO(write bool, duration time.Duration)
 }
 
 type SelectiveListener struct {
-	OnIOCb func(write bool)
+	OnIOCb func(write bool, duration time.Duration)
 }
 
-func (l *SelectiveListener) OnIO(write bool) {
+func (l *SelectiveListener) OnIO(write bool, duration time.Duration) {
 	if l.OnIOCb != nil {
-		l.OnIOCb(write)
+		l.OnIOCb(write, duration)
 	}
 }
--- a/db/event_listener.go
+++ b/db/event_listener.go
@@ -1,17 +1,15 @@
 package db
 
-import "time"
-
 type EventListener interface {
-	OnIO(write bool, duration time.Duration)
+	OnIO(write bool)
 }
 
 type SelectiveListener struct {
-	OnIOCb func(write bool, duration time.Duration)
+	OnIOCb func(write bool)
 }
 
-func (l *SelectiveListener) OnIO(write bool, duration time.Duration) {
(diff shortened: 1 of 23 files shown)


joshklop (commented):
I believe Omer's issue also refers to the call to `OnSyncStepDone(OpVerify...` about 10 lines above.

aminsato:
> I believe Omer's issue also refers to the call to `OnSyncStepDone(OpVerify...` about 10 lines above.

I've made some changes.
Is that ok?

omerfirmak:
Yeah, both `OnSyncStepDone(OpVerify` and `OnSyncStepDone(OpStore` should be put behind a `err == nil` check

aminsato:
> Yeah, both `OnSyncStepDone(OpVerify` and `OnSyncStepDone(OpStore` should be put behind a `err == nil` check

Changed

db/event_listener.go
```diff
@@ -1,17 +1,15 @@
 package db
 
-import "time"
-
 type EventListener interface {
-	OnIO(write bool, duration time.Duration)
+	OnIO(write bool)
```
omerfirmak, line 4: why is this commit reverted?

sync/sync.go
```diff
@@ -167,10 +166,11 @@ func (s *Synchronizer) verifierTask(ctx context.Context, block *core.Block, stat
 				s.log.Warnw("Sanity checks failed", "number", block.Number, "hash", block.Hash.ShortString())
 				resetStreams()
 				return
+			} else {
+				s.listener.OnSyncStepDone(OpVerify, block.Number, time.Since(verifyTimer))
```
omerfirmak, line 170: this should be moved back to it's original position, before the `return func()` statement.

sync/sync.go
```diff
@@ -184,7 +184,10 @@ func (s *Synchronizer) verifierTask(ctx context.Context, block *core.Block, stat
 				}
 				resetStreams()
 				return
+			} else {
```
omerfirmak, line 187: you dont need the else statement here. `if err != nil` branch returns early anyways.

sync/sync.go
aminsato: but  @joshklop commented in [this](https://github.com/NethermindEth/juno/pull/1387#pullrequestreview-1702853868
) to change s.listener.OnSyncStepDone(OpVerify

omerfirmak: I think he meant something different. If you move this line to here, `time.Since(verifyTimer)` won't yield the correct value.

omerfirmak: We still need to check `if err == nil` before calling `OnSyncStepDone(OpVerify`

omerfirmak (approved):

Issue #1385, opened 2023-10-27
L1 Verifier shouldn't shutdown the node when it encounters an error


omerfirmak:
```
ERROR	node/node.go:255	Service error	{"name": "*l1.Client", "err": "retrieve Ethereum chain ID: daily request count exceeded, request rate limited"}
```

We probably should demote such errors to warnings and keep trying.

aidenwong812:
I'd like to handle this task.

josexx96:
I’d like to help with this.

wojciechos:
Fixed by https://github.com/NethermindEth/juno/pull/1388

Ehsan-saradar:
Happy to take this if it's still open — opened #3771, which retries transient chain-ID errors (rate limits, timeouts) as warnings instead of shutting the node down, with a regression test. Feedback welcome!

PR #1388, opened 2023-10-28
Enhancement: Improved L1 Verifier Error Handling

aminsato:
#1385


l1/l1.go
```diff
@@ -78,7 +78,9 @@ func (c *Client) subscribeToUpdates(ctx context.Context, updateChan chan *contra
 func (c *Client) checkChainID(ctx context.Context) error {
 	gotChainID, err := c.l1.ChainID(ctx)
 	if err != nil {
-		return fmt.Errorf("retrieve Ethereum chain ID: %w", err)
+		// Instead of returning an error, log a warning and keep trying.
+		c.log.Warnw("Failed to retrieve Ethereum chain ID: ", err)
+		return nil
```
joshklop, line 83: We need to keep trying (i.e., no `return nil`). We'll need to put this in a loop and backoff when we receive an error.

joshklop (changes requested):

aminsato:
It still hasn't been reviewed by someone.
@omerfirmak 
@joshklop

l1/l1.go
```diff
@@ -76,19 +76,28 @@ func (c *Client) subscribeToUpdates(ctx context.Context, updateChan chan *contra
 }
 
 func (c *Client) checkChainID(ctx context.Context) error {
-	gotChainID, err := c.l1.ChainID(ctx)
-	if err != nil {
-		return fmt.Errorf("retrieve Ethereum chain ID: %w", err)
-	}
+	retryAttempts := 3
+
+	for i := 0; i < retryAttempts; i++ {
```
joshklop, line 81: We should retry forever (never fail).

l1/l1.go
```diff
@@ -76,19 +76,28 @@ func (c *Client) subscribeToUpdates(ctx context.Context, updateChan chan *contra
...
+		gotChainID, err := c.l1.ChainID(ctx)
+		if err != nil {
+			c.log.Warnw("Failed to retrieve Ethereum chain ID: ", err)
+			backOffDuration := time.Duration(i+1) * time.Second
+			time.Sleep(backOffDuration)
```
joshklop, line 86: The backoff should be exponential.

joshklop (changes requested):

l1/l1.go
```diff
@@ -76,19 +76,25 @@ func (c *Client) subscribeToUpdates(ctx context.Context, updateChan chan *contra
...
 func (c *Client) checkChainID(ctx context.Context) error {
-	gotChainID, err := c.l1.ChainID(ctx)
-	if err != nil {
-		return fmt.Errorf("retrieve Ethereum chain ID: %w", err)
-	}
+	for {
+		backOffDuration := time.Duration(1) * time.Second
+		gotChainID, err := c.l1.ChainID(ctx)
+		if err != nil {
+			c.log.Warnw("Failed to retrieve Ethereum chain ID: ", err)
```
joshklop, line 83:
```suggestion
			c.log.Warnw("Failed to retrieve Ethereum chain ID, retrying", err)
```
Let's move this below the backoff calculation and add `"waitFor", backoffDuration` to the log so the user knows how long Juno will wait.

l1/l1.go
```diff
@@ -76,19 +76,25 @@ func (c *Client) subscribeToUpdates(ctx context.Context, updateChan chan *contra
...
+			backOffDuration *= 2
+			time.Sleep(backOffDuration)
```
joshklop, line 85: These lines should be swapped.

l1/l1.go
```diff
@@ -76,19 +76,25 @@ func (c *Client) subscribeToUpdates(ctx context.Context, updateChan chan *contra
 }
 
 func (c *Client) checkChainID(ctx context.Context) error {
-	gotChainID, err := c.l1.ChainID(ctx)
-	if err != nil {
-		return fmt.Errorf("retrieve Ethereum chain ID: %w", err)
-	}
+	for {
+		backOffDuration := time.Duration(1) * time.Second
```
joshklop, line 80: This should go outside the `for` loop, otherwise we reset the backoff every iteration.

joshklop (changes requested):

joshklop:
We can wait for https://github.com/NethermindEth/juno/pull/1399 to be merged so we don't have to calculate backoffs and sleep manually.

l1/l1.go
```diff
@@ -76,19 +76,25 @@ func (c *Client) subscribeToUpdates(ctx context.Context, updateChan chan *contra
 }
 
 func (c *Client) checkChainID(ctx context.Context) error {
-	gotChainID, err := c.l1.ChainID(ctx)
-	if err != nil {
-		return fmt.Errorf("retrieve Ethereum chain ID: %w", err)
-	}
+	backOffDuration := time.Duration(1) * time.Second
```
kirugan, line 79: nit: `backOffDuration := time.Second`

PR #1389, opened 2023-10-28
L1 and pending subscriptions

joshklop:
Depends on #1373


omerfirmak:
do you mind rebasing?

sync/sync.go
```diff
@@ -404,11 +407,15 @@ func (s *Synchronizer) fetchAndStorePending(ctx context.Context) error {
...
+	if err := s.blockchain.StorePending(&blockchain.Pending{
 		Block:       pendingBlock,
 		StateUpdate: pendingStateUpdate,
 		NewClasses:  newClasses,
-	})
+	}); err != nil {
+		return err
+	}
+	s.pendingHeads.Send(pendingBlock.Header)
+	return nil
```
omerfirmak: https://github.com/NethermindEth/juno/blob/23bcc52a4d23a0bb96e0cb0dfcb406105dae1414/blockchain/blockchain.go#L974-L977

there is a case where we return `nil` error, even tho don't update the stored pending block. we should handle this case so that we don't send bogus updates the user.

joshklop: Nice catch. Fixed this using a `bool`, but not a huge fan tbh. You think this warrants putting the subscription things back in `blockchain`?

PR #1395, opened 2023-10-31
Collect read metrics on blockchain

omerfirmak:


node/metrics.go
```diff
@@ -177,3 +177,17 @@ func makeJunoMetrics(version string) {
 		ConstLabels: prometheus.Labels{"version": version},
 	}))
 }
+
+func makeBlockchainMetrics() blockchain.EventListener {
+	reads := prometheus.NewCounterVec(prometheus.CounterOpts{
+		Namespace: "bc",
```
joshklop, line 183: maybe `chain` would be more self-describing than `bc`

joshklop (approved):

node/metrics.go
omerfirmak: changed it to `blockchain`

PR #1397, opened 2023-10-31
Properly set GOMAXPROCS in Kubernetes pod

omerfirmak:
more information at https://levelup.gitconnected.com/know-gomaxprocs-before-deploying-your-go-app-to-kubernetes-7a458fb63af1


node/node.go
```diff
@@ -30,6 +30,7 @@ import (
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/mitchellh/mapstructure"
 	"github.com/sourcegraph/conc"
+	_ "go.uber.org/automaxprocs"
```
joshklop, line 33: I think this should be in `cmd/juno/juno.go`

wojciechos (approved):

node/node.go
omerfirmak: I don't think it matters, but will move it there if you think it makes more sense there.

PR #1401, opened 2023-11-02
Update blockifier

joshklop:


vm/vm.go
```diff
@@ -215,33 +216,37 @@ func (v *vm) Execute(txns []core.Transaction, declaredClasses []core.Class, bloc
 	return context.actualFees, context.traces, nil
 }
 
-func marshalTxnsAndDeclaredClasses(txns []core.Transaction, declaredClasses []core.Class) (json.RawMessage, json.RawMessage, error) {
+func marshalTxnsAndDeclaredClasses(txns []core.Transaction, declaredClasses []core.Class) (json.RawMessage, json.RawMessage, int, error) {
 	txnJSONs := []json.RawMessage{}
-	for _, txn := range txns {
-		txnJSON, err := marshalTxn(txn)
+	var queryBitBitmap int
```
omerfirmak, line 221: Neat idea, but it fails when `len(txns) >= (sizeof(int) * 8)`

omerfirmak:
Gave it a try myself as well, since it is already late for you and we need to get this out the door ASAP.

vm/vm.go
joshklop: Oh my goodness. Sleep deprived coding gone wrong :sweat_smile: 
Thanks!

IronGauntlets (approved):

PR #1407, opened 2023-11-05
Use most up-to-date docker image in docs

joshklop:


README.md
```diff
@@ -68,7 +68,7 @@ docker run -d \
   --name juno \
   -p 6060:6060 \
   -v $HOME/juno:/var/lib/juno \
-  nethermind/juno \
+  nethermindeth/juno \
   --http \
```
wojciechos, line 72: We should stay with nethermind/juno which is our official docker hub. 

nethermindeth/juno is rather for our internal usage or tests

joshklop: Ahhh sounds good. I forgot to pull `nethermind/juno` and thought it wasn't up to date. My bad.

PR #1412, opened 2023-11-06
Specify syn version for rust vm to fix build

wojciechos:


joshklop (approved):

omerfirmak:
Let's merge this to unblock the release. But I really don't like the idea of forcing version of a package that we don't depend on directly.

Created an issue upstream, https://github.com/udoprog/genco/issues/50
Let's monitor the issue and revert this hack as soon as we can.

omerfirmak (approved):

Issue #1328, opened 2023-10-10
Enable syncing from custom feeder URL


joshklop:

Exca-DK:
If this issue is not being worked on, I'd be willing to take it on.

joshklop:
@omerfirmak has worked on it and may put it in a PR.

omerfirmak:
Closing in favor of #1415

PR #1416, opened 2023-11-08
Rianhughes/custom network

rianhughes:
closes https://github.com/NethermindEth/juno/issues/1328


.golangci.yaml
```diff
@@ -27,6 +27,9 @@ linters-settings:
       - octalLiteral
       - whyNoLint
       - unnamedResult
+    settings:
+      hugeParam:
+        sizeThreshold: 89
```
joshklop, line 32: I'm not sure where the warning is coming from, but maybe we should heed the warning and consider using pointers.

utils/network.go
```diff
@@ -6,151 +6,229 @@ import (
...
-		address = common.HexToAddress("0xa4eD3aD27c294565cB0DCc993BDdCC75432D498c")
-	case INTEGRATION:
-		return common.Address{}, errors.New("l1 contract is not available on the integration network")
-	default:
-		// Should not happen.
-		return common.Address{}, ErrUnknownNetwork
-	}
-	return address, nil
+	if n.l1ChainID == nil {
+		return common.Address{}, errors.New("l1 contract is not available on this network")
```
joshklop, line 229:
```suggestion
		return common.Address{}, fmt.Errorf("l1 contract is not available on network %s", n.name)
```
nit: make error message slightly more informative

utils/network.go
```diff
@@ -6,151 +6,229 @@ import (
...
+
+	if network, ok := predefinedNetworks[s]; ok {
+		*n = network
+		return nil
+	}
+
+	return n.setCustomNetwork(s)
+}
+
+func (n *Network) setCustomNetwork(s string) error {
```
joshklop, line 138: I'm not sure how well this will scale. In the future, we will likely need a lot more information for a custom network, e.g., the genesis state and hard fork history. Can we make this accept a json string and unmarshal it to the `Network` struct? That should be easier than manually parsing it.

Just a thought, I'm open to other ideas.

utils/network.go
```diff
@@ -6,151 +6,229 @@ import (
...
-	default:
-		// Should not happen.
-		panic(ErrUnknownNetwork)
-	}
-}
-
 // FeederURL returns URL for read commands
 func (n Network) FeederURL() string {
-	return n.baseURL() + "feeder_gateway/"
+	return n.baseURL + "feeder_gateway/"
```
joshklop, line 211: Do you think we should have a `feederURL` field instead of `baseURL`? As it stands, custom networks must include sequencer endpoints on the `feeder_gateway` path. Same for `gatewayURL`.

utils/network_test.go
```diff
@@ -90,8 +92,37 @@ func TestNetworkSet(t *testing.T) {
 		n := new(utils.Network)
 		require.ErrorIs(t, n.Set("blah"), utils.ErrUnknownNetwork)
 	})
+
+	t.Run("custom network - success", func(t *testing.T) {
```
rianhughes, line 96: Note: this is failing to unmarshal the networkJSON string into a network type (even when only contains name field)

utils/network.go
rianhughes: I implemented the json-string approach

utils/network.go
rianhughes: I replaced the baseURL with FeederURL and GatewayURL fields.

.golangci.yaml
rianhughes: It arose because the Network struct became heavy. I updated the code to use pointers which removes the relevant lint error

utils/network.go
```diff
@@ -62,85 +135,111 @@ func (n *Network) MarshalJSON() ([]byte, error) {
...
 	}
+	return nil
 }
 
-// FeederURL returns URL for read commands
-func (n Network) FeederURL() string {
-	return n.baseURL() + "feeder_gateway/"
+// Unmarshals a json string into a Network struct, and requires all fields to be present
+func (n *Network) UnmarshalJSON(data []byte) error {
+	jsonMap := make(map[string]any)
```
omerfirmak, line 175: Let's define a proper type with json tags instead of using maps?

juno/CURRENT
```diff
@@ -0,0 +1 @@
+MANIFEST-000036
```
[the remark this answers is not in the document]
rianhughes: What is this file anyway? :)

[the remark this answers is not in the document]
rianhughes: removed

utils/network_test.go
```diff
@@ -25,70 +28,140 @@ func TestNetwork(t *testing.T) {
...
+				"feeder_url": "baseURL/feeder_gateway/",
+				"gateway_url": "baseURL/gateway",
+				"chain_id": "SN_CUSTOM",
+				"l1_chain_id": 123,
+				"core_contract_address": "0x1",
+				"block_hash_meta_info": {
+					"first_07_block": 1,
+					"unverifiable_range": [2, 3],
+					"fallback_sequencer_address": "0x0"
+				}
```
omerfirmak, line 117: I don't think we should support these for the custom networks, they are only there because SW operated networks have a lot of baggage.

utils/network_test.go
```diff
@@ -25,70 +28,140 @@ func TestNetwork(t *testing.T) {
...
-			assert.Equal(t, network, *n)
+			assert.Equal(t, network, n)
 		})
 	}
 
-	t.Run("unknown network", func(t *testing.T) {
+	t.Run("custom network - success", func(t *testing.T) {
 		n := new(utils.Network)
-		require.ErrorIs(t, n.Set("blah"), utils.ErrUnknownNetwork)
+		networkJSON := `{
```
omerfirmak, line 106: Passing a JSON from command line feels really impractical to me. I think we either need to make these different command line flags that use have to set one by one or make custom networks only configurable by config files.

omerfirmak:
I am not sure if we need to change all usages of utils.Network to pointers, it is creating so much noise in the PR. It is a small enough struct to be passed by value IMO.

If you disagree, let's extract changes related to that to a separate commit.

joshklop:
Can this be closed in favor of #1542?

rianhughes:
> Can this be closed in favor of https://github.com/NethermindEth/juno/pull/1542?
Yes, closing it now.

Issue #1411, opened 2023-11-06
Update README.md


stranger80:
Update README.md:
- Align Roadmap ("Phase 1/Phase 2/Phase 3/ Phase 4") with https://www.nethermind.io/juno and indicate progress
- Add "Quick start" section to provide the shortest possible sequence to setup and run a Juno node.

PR #1424, opened 2023-11-11
Update Roadmap 

ninja342:
#1411  Roadmap aligned with website


joshklop (approved):

README.md
```diff
@@ -262,6 +262,35 @@ The focus of Phase 2 will be to Verify the state from layer 1 and implement the
...
 
+### Phase 3: Starknet decentralization begins 🚧
+
+<details>
+<summary></summary>
+
+Juno can synchronize Starknet state from other full nodes with the aim of decentralizing Starknet by removing the dependency from the centralized sequencer.
+
+
+Snap sync is implemented, significantly reducing sync times and the implementation of JSON RPC’s Trace API will allow users to better understand its execution.
```
joshklop, line 273: Just realized: we already implement the trace API. Can we remove that clause from this sentence?

joshklop (changes requested):

README.md
ninja342: Updated

joshklop: Thanks. I think it would be better if we removed the part about the trace API. I am concerned readers may interpret it as saying that the trace API isn't implemented.

joshklop (approved):

PR #1427, opened 2023-11-14
RPC v0.6-rc1

joshklop:


core/transaction.go
```diff
@@ -171,6 +237,15 @@ type DeployAccountTransaction struct {
...
 	Nonce *felt.Felt
+
+	// Version 3 fields
+	// See InvokeTransaction for descriptions of the fields.
+	ResourceBounds        map[Resource]*ResourceBounds
+	Tip                   uint64
+	PaymasterData         []*felt.Felt
+	AccountDeploymentData []*felt.Felt
+	NonceDAMode           DataAvailabilityMode
+	FeeDAMode             DataAvailabilityMode
```
omerfirmak, line 248: Do you think it makes sense put all these in a separate struct and embed them in each txn type, instead of duplicating them?

omerfirmak: If it is backwards compatible (ie wont need a DB migration), we can put all common fields in that struct as well.

vm/rust/src/jsonrpc.rs
```diff
@@ -190,6 +195,53 @@ impl From<BlockifierOrderedEvent> for OrderedEvent {
...
+        let applications = |builtin_name: &str| -> Option<usize> {
+            match val.builtin_instance_counter.get(builtin_name) {
+                Some(x) => Some(*x),
+                None => None,
+            }
+        };
+        ExecutionResources {
+            steps: val.n_steps,
+            memory_holes: val.n_memory_holes,
+            range_check_builtin_applications: applications(RANGE_CHECK_BUILTIN_NAME),
```
omerfirmak, line 213: you can do `val.builtin_instance_counter.get(RANGE_CHECK_BUILTIN_NAME).cloned()` if you feel like it looks cleaner

rpc/trace.go
```diff
@@ -123,12 +114,12 @@ func adaptFunctionInvocation(snFnInvocation *starknet.FunctionInvocation, legacy
...
 	}
 	for index := range snFnInvocation.InternalCalls {
-		fnInvocation.Calls = append(fnInvocation.Calls, *adaptFunctionInvocation(&snFnInvocation.InternalCalls[index], legacyTrace))
+		fnInvocation.Calls = append(fnInvocation.Calls, *adaptFunctionInvocation(&snFnInvocation.InternalCalls[index]))
 	}
 	for index := range snFnInvocation.Events {
 		snEvent := &snFnInvocation.Events[index]
 		fnInvocation.Events = append(fnInvocation.Events, OrderedEvent{
-			Order: orderPtr(snEvent.Order),
+			Order: &snEvent.Order,
```
omerfirmak, line 122: We can make `Order` a value now, since it is not optional.

core/transaction.go
joshklop: Just realized the `AccountDeploymentData` field is unused in `DeployAccount` transactions. Fixed that.

> Do you think it makes sense put all these in a separate struct and embed them in each txn type, instead of duplicating them?

Most of the fields serve very different purposes. I'm hesitant to divide the structs. Makes the code harder to read imho.

rpc/handlers.go
```diff
@@ -73,7 +73,6 @@ const (
 
 type traceCacheKey struct {
```
omerfirmak, line 74: we can remove this and use felt.Felt as the key

adapters/sn2core/sn2core.go
```diff
@@ -165,24 +169,39 @@ func AdaptDeclareTransaction(t *starknet.Transaction) *core.DeclareTransaction {
...
+}
+
+func adaptResource(resource starknet.Resource) (core.Resource, error) {
+	switch resource {
+	case starknet.ResourceL1Gas:
+		return core.ResourceL1Gas, nil
+	case starknet.ResourceL2Gas:
+		return core.ResourceL2Gas, nil
+	default:
+		return 0, fmt.Errorf("unknown starknet.Resource %s", resource)
```
omerfirmak, line 203: I think it makes sense to keep the UnmarshalJSON function to catch this. So we don't need to make `AdaptInvokeTransaction` return an error and bubble it up to the caller.

starknet/transaction.go
```diff
@@ -104,22 +104,59 @@ func (t *TransactionType) UnmarshalJSON(data []byte) error {
...
+	Type                  TransactionType              `json:"type,omitempty"`
+	SenderAddress         *felt.Felt                   `json:"sender_address,omitempty"`
+	MaxFee                *felt.Felt                   `json:"max_fee,omitempty"`
+	Signature             *[]*felt.Felt                `json:"signature,omitempty"`
+	CallData              *[]*felt.Felt                `json:"calldata,omitempty"`
+	EntryPointSelector    *felt.Felt                   `json:"entry_point_selector,omitempty"`
+	Nonce                 *felt.Felt                   `json:"nonce,omitempty"`
+	CompiledClassHash     *felt.Felt                   `json:"compiled_class_hash,omitempty"`
+	ResourceBounds        map[Resource]*ResourceBounds `json:"resource_bounds,omitempty"`
+	Tip                   uint64                       `json:"tip,omitempty"`
```
omerfirmak: I think 0 tip is still a valid value for v3 txns. So I believe it should be an `*uint64`

vm/rust/src/lib.rs
```diff
@@ -124,6 +132,7 @@ pub extern "C" fn cairoVMCall(
 #[derive(Deserialize)]
 pub struct TxnAndQueryBit {
     pub txn: StarknetApiTransaction,
+    pub txn_hash: TransactionHash,
```
omerfirmak, line 135: we need to add this to the json we pass from go side, right?

joshklop: Nice catch, thanks.

starknet/transaction.go
joshklop: [It is](https://external.integration.starknet.io/feeder_gateway/get_block?blockNumber=319100). But not sure why we need `uint64` to be a pointer in that case. Feeder always includes the tip, even if it's `0x0`.

adapters/sn2core/sn2core.go
joshklop: `AdaptInvokeTransaction` is public, so we could encounter an invalid `starknet.Resource` even if an `UnmarshalJSON` existed.

`AdaptTransaction` already returns an error so it doesn't matter much as far as Juno goes.

vm/rust/src/lib.rs
```diff
@@ -355,14 +368,14 @@ fn build_block_context(
...
-            .unwrap(),
-        )
-        .unwrap(),
-        gas_price, // fixed gas price, so that we can return "consumed gas" to Go side
+        fee_token_addresses: FeeTokenAddresses {
+            eth_fee_token_address: ContractAddress::try_from(
+                                       StarkHash::try_from(
+                                           "0x049d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7",
+                                           ).unwrap()).unwrap(),
+            strk_fee_token_address: ContractAddress::try_from(StarkHash::try_from("0x0").unwrap()).unwrap(), // TODO what is the real STRK address?
```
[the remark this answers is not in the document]
joshklop: Yeah. You think `0x0` makes sense for now?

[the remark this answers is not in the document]
joshklop: Using `ContractAddress::default()` for now.

adapters/sn2core/sn2core.go
joshklop: Never mind, I see what you mean. For some reason unmarshalling works when I include `UnmarshalText`. The `encoding/json` docs are vague/misleading about this:

> The map's key type must either be any string type, an integer, implement json.Unmarshaler, or implement encoding.TextUnmarshaler.

Anyways, should be fixed now.

starknet/transaction.go
omerfirmak: Tip is marked `omitempty` so if we ever marshal this type to JSON, it will get omitted if the value is 0.

core/transaction.go
omerfirmak: > Most of the fields serve very different purposes

Do fields introduced in V3 have different meanings depending on the txn type?

starknet/block.go
```diff
@@ -9,10 +9,23 @@ type Block struct {
...
+	GasPriceSTRK     *felt.Felt            `json:"strk_l1_gas_price"`
+
+	// TODO we can remove the GasPrice method and the GasPriceLegacy field
+	// once v0.13 lands on mainnet. In the meantime, we include both to support
+	// pre-v0.13 jsons, where `eth_l1_gas_price` was called `gas_price`.
+	GasPriceLegacy *felt.Felt `json:"gas_price"`
+	GasPriceWEI    *felt.Felt `json:"eth_l1_gas_price"`
+}
+
+func (b *Block) GasPrice() *felt.Felt {
```
omerfirmak, line 26: We can call this `GasPriceETH` / `GasPriceWEI` or sth, to avoid confusion.

rpc/handlers.go
```diff
@@ -1872,17 +1842,30 @@ func (h *Handler) LegacyMethods() ([]jsonrpc.Method, string) { //nolint: funlen
...
 		{
 			Name:    "starknet_traceBlockTransactions",
-			Params:  []jsonrpc.Parameter{{Name: "block_hash"}},
-			Handler: h.LegacyTraceBlockTransactions,
+			Params:  []jsonrpc.Parameter{{Name: "block_id"}},
+			Handler: h.TraceBlockTransactions,
+		},
+		{
+			Name:    "starknet_specVersion",
+			Handler: h.SpecVersion,
```
omerfirmak, line 1859: We need `LegacySpecVersion`

core/transaction.go
joshklop: No.

starknet/transaction.go
joshklop: Just realized that I inadvertently fixed this when I changed `Tip` to a `*felt.Felt`, since the feeder formats it as a `felt`.

core/transaction.go
```diff
@@ -21,6 +21,71 @@ import (
 	"golang.org/x/crypto/sha3"
 )
 
+const (
+	maxAmountBits            = 64
+	maxPricePerUintBits      = 128
```
omerfirmak, line 26: unused

adapters/sn2core/sn2core.go
```diff
@@ -156,17 +157,34 @@ func AdaptTransaction(transaction *starknet.Transaction) (core.Transaction, erro
...
+		Nonce:                 t.Nonce,
+		Version:               (*core.TransactionVersion)(t.Version),
+		ClassHash:             t.ClassHash,
+		CompiledClassHash:     t.CompiledClassHash,
+		ResourceBounds:        adaptResourceBounds(t.ResourceBounds),
+		Tip:                   t.Tip,
+		PaymasterData:         t.PaymasterData,
+		AccountDeploymentData: t.AccountDeploymentData,
+		NonceDAMode:           core.DataAvailabilityMode(t.NonceDAMode),
+		FeeDAMode:             core.DataAvailabilityMode(t.FeeDAMode),
```
omerfirmak, line 173: These cannot be casted directly between each other.

core:
``` golang

type DataAvailabilityMode uint32

const (
	DAModeL1 DataAvailabilityMode = iota + 1
	DAModeL2
)
```

starknet:
``` golang

type DataAvailabilityMode uint32

const (
	DAModeL1 DataAvailabilityMode = iota
	DAModeL2
)

```

let's both make them start at 0. This way, we can get rid of core.DataAvailabilityMode.Index() as well.

joshklop: Fixed this and got rid of `core.DataAvailabilityMode.index` in a later commit.

I can rebase this PR to make it easier to review.

omerfirmak: Ah, ok. I will just review the entire change set then.

core/transaction.go
```diff
@@ -205,6 +268,18 @@ type InvokeTransaction struct {
 	Nonce *felt.Felt
 	// The address of the sender of this transaction
 	SenderAddress *felt.Felt
+
+	// Version 3 fields (there was no version 2)
+	ResourceBounds map[Resource]*ResourceBounds
```
omerfirmak, line 273: any particular reason these are pointers?

```suggestion
	ResourceBounds map[Resource]ResourceBounds
```

joshklop: No.

core/transaction.go
```diff
@@ -349,11 +433,35 @@ func invokeTransactionHash(i *InvokeTransaction, n utils.Network) (*felt.Felt, e
...
+	l1Bounds := new(felt.Felt).SetBytes(resourceBounds[ResourceL1Gas].Bytes(ResourceL1Gas))
+	l2Bounds := new(felt.Felt).SetBytes(resourceBounds[ResourceL2Gas].Bytes(ResourceL2Gas))
+	return crypto.PoseidonArray(new(felt.Felt).SetUint64(tip), l1Bounds, l2Bounds)
+}
+
+func dataAvailabilityMode(feeDAMode, nonceDAMode DataAvailabilityMode) uint64 {
+	const dataAvailabilityModeBits = 32
+	return uint64(feeDAMode) + uint64(nonceDAMode)<<dataAvailabilityModeBits
+}
+
```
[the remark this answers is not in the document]
joshklop: I think the standalone functions make sense in this case because the packing is only used for transaction hashes.

joshklop:
Currently waiting for confirmation from Starkware on how to fill in the `FeePayment.Unit` field in v3 transaction receipts. Right now we default to `WEI`.

omerfirmak:
Can you rebase this PR? Let's see how much code we have left to merge :smile:

vm/rust/src/jsonrpc.rs
```diff
@@ -157,6 +158,7 @@ pub fn new_transaction_trace(
                     DeclareTransaction::V0(_) => Some(declare_txn.class_hash()),
                     DeclareTransaction::V1(_) => Some(declare_txn.class_hash()),
                     DeclareTransaction::V2(_) => None,
+                    DeclareTransaction::V3(_) => None,
```
omerfirmak, line 144: We can remove the last 2 cases and replace them with default case `_`

vm/rust/src/lib.rs
```diff
@@ -146,7 +155,6 @@ pub extern "C" fn cairoVMExecute(
     skip_charge_fee: c_uchar,
     skip_validate: c_uchar,
     gas_price: *const c_uchar,
-    legacy_json: c_uchar,
```
omerfirmak, line 149: I think this change belongs to the next commit

vm/rust/src/lib.rs
```diff
@@ -189,9 +197,12 @@ pub extern "C" fn cairoVMExecute(
         block_number,
         block_timestamp,
         sequencer_address_felt,
-        felt_to_u128(gas_price_felt),
+        GasPrices {
+            eth_l1_gas_price: felt_to_u128(gas_price_felt),
+            strk_l1_gas_price: 0,
```
omerfirmak, line 202: shouldn't we pass this from Go side?

core/transaction.go
```diff
@@ -127,6 +127,7 @@ type TransactionReceipt struct {
 type Transaction interface {
 	Hash() *felt.Felt
 	Signature() []*felt.Felt
+	TxVersion() *TransactionVersion
```
omerfirmak, line 130:
```suggestion
	Version() *TransactionVersion
```

vm/rust/src/lib.rs
```diff
@@ -366,15 +388,18 @@ fn build_block_context(
...
-        .unwrap(),
-        gas_price, // fixed gas price, so that we can return "consumed gas" to Go side
+        fee_token_addresses: FeeTokenAddresses {
+            // eth_fee_token_address is the same for all networks
+            eth_fee_token_address: ContractAddress::try_from(
+                                       StarkHash::try_from(
+                                           "0x049d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7",
+                                           ).unwrap()).unwrap(),
+            strk_fee_token_address: match chain_id_str {
+                "SN_GOERLI" => ContractAddress::try_from(StarkHash::try_from("0x04718f5a0fc34cc1af16a1cdee98ffb20c31f5cd61d6ab07201858f4287c938d").unwrap()).unwrap(),
```
omerfirmak, line 398: Confirmed that this will be the same for all networks, like ETH.

omerfirmak (approved):
Squash merging, but please take a look at the latest commits that I added @joshklop

PR #1432, opened 2023-11-20
 Make read-only contract functions standalone

omerfirmak:
Towards better `starknet_call` performance

### DIFF
--- a/core/contract_test.go
+++ b/core/contract_test.go
@@ -63,7 +63,7 @@ func TestNewContract(t *testing.T) {
 	classHash := new(felt.Felt).SetBytes([]byte("class hash"))
 
 	t.Run("cannot create Contract instance if un-deployed", func(t *testing.T) {
-		_, err = core.NewContract(addr, txn)
+		_, err = core.NewContractUpdater(addr, txn)
 		require.EqualError(t, err, core.ErrContractNotDeployed.Error())
 	})
 
@@ -192,6 +192,6 @@ func TestPurge(t *testing.T) {
 	require.NoError(t, err)
 
 	require.NoError(t, contract.Purge())
-	_, err = core.NewContract(addr, txn)
+	_, err = core.NewContractUpdater(addr, txn)
 	assert.ErrorIs(t, err, core.ErrContractNotDeployed)
 }
--- a/core/contract_test.go
+++ b/core/contract_test.go
@@ -75,46 +75,22 @@ func TestNewContract(t *testing.T) {
 		require.EqualError(t, err, core.ErrContractAlreadyDeployed.Error())
 	})
 
-	t.Run("instantiate previously deployed contract", func(t *testing.T) {
-		t.Run("contract with empty storage", func(t *testing.T) {
-			newContract, err := core.NewContract(addr, txn)
-			require.NoError(t, err)
-
-			root, err := newContract.Root()
-			require.NoError(t, err)
-			assert.Equal(t, &felt.Zero, root)
-		})
-		t.Run("contract with non-empty storage", func(t *testing.T) {
-			oldRoot, err := contract.Root()
-			require.NoError(t, err)
-
-			require.NoError(t, contract.UpdateStorage([]core.StorageDiff{{Key: addr, Value: classHash}}, NoopOnValueChanged))
-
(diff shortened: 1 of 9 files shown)


core/contract_test.go
```diff
@@ -75,46 +75,22 @@ func TestNewContract(t *testing.T) {
...
-
-			newContract, err := core.NewContract(addr, txn)
-			require.NoError(t, err)
-
-			root, err := newContract.Root()
-			require.NoError(t, err)
-			assert.NotEqual(t, oldRoot, root)
-		})
-	})
-
```
IronGauntlets, line 101: I don't see why these tests need to be removed.

IronGauntlets (commented):
I think I am missing some context here, can you please answer the following questions:
- How does making a read-only contract function have a performance impact? Is it because multiple read-only db transactions can simultaneously access contract storage?
- How does passing class hash to the `vm.Call` help with performance?

core/contract_test.go
omerfirmak: My reasoning was, since `Root` is not a member function of `Contract` anymore it doesnt really make sense to test its relation with a `NewContract` call

omerfirmak:
> How does making a read-only contract function have a performance impact? Is it because multiple read-only db transactions can simultaneously access contract storage?

`NewContract` makes sure that there is a contract deployed to that address first. But we don't need that behaviour when we are simply reading from the state. It only makes sense to verify that a contract exists when we are updating its state/nonce/classhash.

> How does passing class hash to the vm.Call help with performance?

We fetch the contract class in `handler.Call` to return a certain error message if there is no contract deployed at the address that user is trying to call. `blockifier` needs the same information and it makes a second DB read to get that same class hash. We can just pass it the value we have already and save on a DB read.

core/contract_test.go
IronGauntlets: That is a consequence of splitting write and read functions. I think we still need to maintain the same behaviour and the only way to check the updater has made the changes correctly is by calling the static functions as it is done in the tests below.

joshklop:
> NewContract makes sure that there is a contract deployed to that address first. But we don't need that behaviour when we are simply reading from the state. It only makes sense to verify that a contract exists when we are updating its state/nonce/classhash.

Would it be cleaner to move the `deploy` check to the "write" methods and maintain the `Contract` abstraction?

core/contract_test.go
omerfirmak: That is already done at https://github.com/NethermindEth/juno/pull/1432/files#diff-a0246ffea9132d6506bb64b2651d4410c9942017252ee959741aed9268792178L179-L192 , tho?

vm/rust/src/lib.rs
```diff
@@ -86,7 +92,7 @@ pub extern "C" fn cairoVMCall(
         calldata: Calldata(calldata_vec.into()),
         storage_address: contract_addr_felt.try_into().unwrap(),
         call_type: CallType::Call,
-        class_hash: None,
+        class_hash: class_hash,
```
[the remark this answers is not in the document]
omerfirmak: Yes, at least that is what the API of `CallEntryPoint` suggests

omerfirmak:
> > NewContract makes sure that there is a contract deployed to that address first. But we don't need that behaviour when we are simply reading from the state. It only makes sense to verify that a contract exists when we are updating its state/nonce/classhash.
> 
> So it doesn't impact the performance, it is just a refactor?

It does, we skip the "deployed" check.

omerfirmak:
> > NewContract makes sure that there is a contract deployed to that address first. But we don't need that behaviour when we are simply reading from the state. It only makes sense to verify that a contract exists when we are updating its state/nonce/classhash.
> 
> Would it be cleaner to move the `deploy` check to the "write" methods and maintain the `Contract` abstraction?

I considered that too, but since the check was previously at `NewContract` I didnt feel safe going thru all the usages of it make sure they werent affected.

Also, every time we wanted to read from a contract's state, we had to call NewContract and potentially cause a heap allocation superfluously.

IronGauntlets (commented):
LGTM

IronGauntlets (approved):

PR #1439, opened 2023-11-21
Align all implementations of State to return ErrKeyNotFound if contra…

omerfirmak:
…ct is not deployed

and make sure callers are checking for ErrKeyNotFound


IronGauntlets (approved):
lgtm

PR #1447, opened 2023-11-22
 Add v3 Transaction fields 

omerfirmak:


adapters/sn2core/sn2core.go
```diff
@@ -157,15 +157,32 @@ func AdaptTransaction(transaction *starknet.Transaction) (core.Transaction, erro
...
+		NonceDAMode:           core.DataAvailabilityMode(t.NonceDAMode),
+		FeeDAMode:             core.DataAvailabilityMode(t.FeeDAMode),
+	}
+}
+
+func adaptResourceBounds(rb map[starknet.Resource]starknet.ResourceBounds) map[core.Resource]core.ResourceBounds {
+	coreBounds := make(map[core.Resource]core.ResourceBounds, len(rb))
+	for resource, bounds := range rb {
+		coreBounds[core.Resource(resource)] = core.ResourceBounds{
+			MaxAmount:       bounds.MaxAmount.Uint64(),
```
omerfirmak, line 181: is it safe to assume `bounds.MaxAmount` fits in a uint64?

core/transaction.go
```diff
@@ -21,6 +21,47 @@ import (
...
+const (
+	WEI FeeUnit = iota
+	STRK
+)
+
+// From the RPC spec: The max amount and max price per unit of gas used in this transaction.
+type ResourceBounds struct {
+	MaxAmount uint64
+	// MaxPricePerUnit is technically a uint128
+	MaxPricePerUnit *felt.Felt
```
joshklop, line 62: Something I was wondering while implementing: Should we have a `uint128` type? Might save a little disk space.

joshklop (approved):

adapters/sn2core/sn2core.go
joshklop: It's what the spec says https://github.com/starknet-io/SNIPs/blob/main/SNIPS/snip-8.md

core/transaction.go
omerfirmak: Yeah, could be worthwhile to explore later on.

adapters/sn2core/sn2core.go
joshklop: We will have to do db migrations if we wait. Might be better to delay the release by 1-2 days and just implement them the first time imho.

omerfirmak: > We will have to do db migrations if we wait
Not really, felts are stored as big-endian byte arrays. We can keep using the underlying data in the database, it is no big deal.

Issue #1448, opened 2023-11-22
Implement basic `uint128` type


joshklop:
Or use third-party package if there's a good one. I think `net` contains an un-exported `uint128` implementation we can use as inspiration.

Context: https://github.com/NethermindEth/juno/pull/1447#discussion_r1402009510

joshklop:
Hi! Something similar would make sense (struct with hi, lo uint64s). There is no need to implement receivers and methods we wouldn't use in Juno; adapter functions/receivers to convert to/from `felt.Felt`s are probably all we want right now.

anubhavitis:
I am applying to this issue via [OnlyDust platform](https://app.onlydust.com/p/juno).

### My background and how it can be leveraged
I have over 2 years of experience in building Go-based applications, having worked with companies like CoinSwitch and currently Fileverse. I have designed, built, and deployed services from scratch that cater to hundreds of thousands of users daily. My expertise spans supporting both Web2 applications and Web3 dApps, making me well-equipped to contribute effectively to your project

### How I plan on tackling this issue
There are two ways to approach this:
1. Leverage the external package: https://github.com/lukechampine/uint128
2. Develop our own implementation based on our specific use case.

We can start with the first approach to test our use case and assess the impacts. Later, we can fork the repository and create our own implementation, ensuring robustness and tailoring it to our needs.

derrix060:
Changing the

PR #1452, opened 2023-11-23
first iteration of new custom uint128 type

sarabrajsingh:
Beginning to implement a `uint128` type into Juno that takes inspiration from the `uint128` custom type in the `net` package - https://github.com/inetaf/netaddr/blob/main/uint128.go.

upstream issue - #1448


core/uint128.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+func NewUint128(hi, lo uint64) *Uint128 {
+	return &Uint128{
+		hi: hi,
+		lo: lo,
+	}
+}
+
+func (u *Uint128) Bytes() []byte {
```
joshklop, line 23: Is this used somewhere?

core/uint128.go
```diff
@@ -0,0 +1,78 @@
+package core
```
joshklop, line 1: Let's move this to it's own package. We typically reserve the `core` package for Starknet-related types.

Following [`uint256`](https://github.com/holiman/uint256)'s lead, the package name can be `uint128` and the struct name can be `Int` to avoid stutters.

core/uint128.go
```diff
@@ -0,0 +1,78 @@
...
+	loBytes := make([]byte, 8)
+
+	binary.BigEndian.PutUint64(hiBytes, u.hi)
+	binary.BigEndian.PutUint64(loBytes, u.lo)
+
+	return append(hiBytes, loBytes...)
+}
+
+func (u *Uint128) UnmarshalJSON(data []byte) error {
+	var value interface{}
```
joshklop, line 34: Can `value` be a `string`? That way we don't need to switch on the type later.

joshklop (commented):
Nice work! The approach looks good; left a few small comments.

core/uint128.go
joshklop: We can safely assume it will always be a hex-string.

uint128/uint128.go
```diff
@@ -0,0 +1,122 @@
...
+	"encoding/binary"
+	"encoding/hex"
+	"encoding/json"
+	"fmt"
+	"strings"
+)
+
+type Int [2]uint64
+
+func NewInt64(lo uint64) *Int {
```
joshklop, line 13: Let's use the `Set...` idiom established by `math/big` and our `core/felt` package.

joshklop:
In `core.ResourceBounds.Bytes`, replace `maxPriceBytes[16:]` with `maxPriceBytes`. We needed to explicitly grab the last 128 bits when we were using the 252-bit felt, but that doesn't apply anymore.

uint128/uint128.go
```diff
@@ -0,0 +1,122 @@
...
+
+func (i *Int) Bytes() []byte {
+	hiBytes := make([]byte, 8)
+	loBytes := make([]byte, 8)
+
+	binary.BigEndian.PutUint64(hiBytes, i[1])
+	binary.BigEndian.PutUint64(loBytes, i[0])
+
+	return append(hiBytes, loBytes...)
+}
```
joshklop, line 57:
```suggestion
func (i *Int) Bytes() []byte {
	result := make([]byte, 16)
	binary.BigEndian.PutUint64(result[8:16], i[0])
	binary.BigEndian.PutUint64(result[0:8], i[1])
	return result
}
```
We can save some allocations this way.

uint128/uint128.go
```diff
@@ -0,0 +1,121 @@
...
+	"encoding/json"
+	"fmt"
+	"strings"
+)
+
+type Int [2]uint64
+
+func SetUint64(lo uint64) *Int {
+	// i[1] = high bits
+	// i[0] = low bits
```
joshklop, line 15: Can we move this comment to the `Int` type and add a bit more context?

uint128/uint128.go
```diff
@@ -0,0 +1,121 @@
...
+	"encoding/binary"
+	"encoding/hex"
+	"encoding/json"
+	"fmt"
+	"strings"
+)
+
+type Int [2]uint64
+
+func SetUint64(lo uint64) *Int {
```
joshklop, line 13: This should be a receiver, e.g., `func (i *Int) SetUint64(lo uint64) *Int` like in `math/big` and `core/felt`.

uint128/uint128.go
```diff
@@ -0,0 +1,121 @@
...
+	z.setUint128(0, lo)
+	return z
+}
+func SetUint128(hi, lo uint64) *Int {
+	z := &Int{}
+	z.setUint128(hi, lo)
+	return z
+}
+
+func SetInt128(hex_str string) *Int {
```
joshklop, line 26: Let's do what felt does:

1. Implement two functions: `SetBigInt` and `SetString`
2. In `SetString`, convert the string to a big int and then use `SetBigInt` to convert to `uint128.Int`. This way we automatically support all the string formats that `math/big.Int.SetString` supports

joshklop (changes requested):

uint128/uint128.go
```diff
@@ -0,0 +1,173 @@
...
+// for example, if we have a 128-bit hex value 0x6e58133b38301a6cdfa34ca991c4ba39
+// *Int[0] (lower bits) should be 0xdfa34ca991c4ba39 or dfa34ca991c4ba39
+// *Int[i] (upper bits) should be 0x6e58133b38301a6c or 6e58133b38301a6c
+// an example of a call to this type can be : uint128.Int([]uint64{lo, high})
+type Int [2]uint64
+
+const (
+	Base16     = 16
+	ByteLength = 16
+	BitLength  = 128
```
joshklop, line 26: Let's use `Bits` and `Bytes` to stay consistent with `felt`

uint128/uint128.go
```diff
@@ -0,0 +1,157 @@
...
+)
+
+var bigIntPool = sync.Pool{
+	New: func() interface{} {
+		return new(big.Int)
+	},
+}
+
+func NewInt(x []uint64) (*Int, error) {
```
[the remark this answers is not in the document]
joshklop: I think this still applies

uint128/uint128.go
```diff
@@ -0,0 +1,173 @@
...
+	vv.SetBytes(b[:])
+
+	return vv.Text(base)
+}
+
+func (i Int) String() string {
+	return "0x" + i.Text(Base16)
+}
+
+// MarshalJSON forwards the call to underlying field element implementation
```
joshklop, line 161: nit: remove comment

uint128/uint128.go
```diff
@@ -0,0 +1,173 @@
...
+		}
+	}
+
+	bytes, err := parseHexString(s)
+
+	if err != nil || len(bytes) > ByteLength {
+		return nil, fmt.Errorf("can't fit string=%s into 128-bit uint; %s", s, err)
+	}
+
+	vv.FillBytes(bytes[:])
```
joshklop, line 86: I don't think we need this. That also allows us to remove the `parseHexString` function.

uint128/uint128_test.go
```diff
@@ -0,0 +1,396 @@
...
+					// probably an integer overflow that big.Int can interpret but not a felt.Felt
+					t.Errorf("actual Bytes() length=%d differs from expected Bytes() length=%d", len(actual.Bytes()), len(expected.Bytes()))
+				}
+				t.Errorf("multiplication not equal")
+			}
+		}
+	}
+}
+
+func byteSlicesEqual(b1 []byte, b2 []byte) bool {
```
joshklop, line 386: We can replace this with `bytes.Equal`

core/transaction.go
```diff
@@ -71,7 +71,7 @@ func (rb ResourceBounds) Bytes(resource Resource) []byte {
 		[]byte{0},
 		[]byte(resource.String()),
 		maxAmountBytes,
-		maxPriceBytes[16:], // Last 128 bits.
+		maxPriceBytes, // Last 128 bits.
```
joshklop, line 74: nit: let's remove the comment here for clarity :)

uint128/uint128_test.go
```diff
@@ -0,0 +1,396 @@
...
+			arr := make([]uint64, 2)
+			arr[0] = test.loBits
+			arr[1] = test.hiBits
+			actual, err := NewInt(arr)
+			if err != nil {
+				if test.wantedErr {
+					return
+				}
+				t.Errorf("couldn't marshal []uint64={%v, %v} into *Int", test.loBits, test.hiBits)
+			}
```
joshklop, line 90:
```suggestion
			actual := Int(arr)
```
We can remove the `NewInt` function; Go does the work fo us here.

uint128/uint128_test.go
```diff
@@ -0,0 +1,396 @@
...
+)
+
+func TestUint128Bytes(t *testing.T) {
+	tests := []struct {
+		description  string
+		loBits       uint64
+		hiBits       uint64
+		expected_arr []byte
+		expected_str string
+		wantedErr    bool
```
joshklop, line 19: this is always false

uint128/uint128_test.go
```diff
@@ -0,0 +1,396 @@
...
+	"github.com/consensys/gnark-crypto/ecc/stark-curve/fp"
+)
+
+func TestUint128Bytes(t *testing.T) {
+	tests := []struct {
+		description  string
+		loBits       uint64
+		hiBits       uint64
+		expected_arr []byte
+		expected_str string
```
joshklop, line 18:
```suggestion
		expectedBytes []byte
		expectedString string
```

uint128/uint128.go
```diff
@@ -0,0 +1,173 @@
...
+}
+
+func (i *Int) SetBigInt(v *big.Int) *Int {
+	// we're expecting words to have a length of 2, which represents 2x uint64s in a slice
+	words := v.Bits()
+
+	for idx, word := range words {
+		i[idx] = uint64(word)
+	}
+
```
joshklop, line 66:
```suggestion
        i[0] = words[0]
        i[1] = words[1]
```

uint128/uint128_test.go
```diff
@@ -0,0 +1,371 @@
...
+func TestUint128Bytes(t *testing.T) {
+	tests := []struct {
+		description    string
+		loBits         uint64
+		hiBits         uint64
+		expectedBytes  []byte
+		expectedString string
+	}{
+		{
+			description:    "128-bit #1",
```
joshklop, line 22: We can make the descriptions a little more expressive in some cases. For example, this one can be `"zero"`. The third one could be `"max"`, etc.

uint128/uint128.go
```diff
@@ -0,0 +1,110 @@
...
+	if len(words) > 0 {
+		switch {
+		case len(words)%2 == 0:
+			i[0] = uint64(words[0])
+			i[1] = uint64(words[1])
+		case len(words)%2 == 1:
+			i[0] = uint64(words[0])
+		}
+	}
+
```
joshklop, line 43: We should align with `felt`'s behavior, which performs the modular reduction when `v` cannot fit in a felt. We can take a lot of inspiration from `felt.SetBigInt` and `felt.setBigInt`.

kirugan:
In future we will take this implementation (most likely) - https://github.com/lukechampine/uint128/blob/master/uint128.go

Issue #1358, opened 2023-10-23
Improving Juno docs


Tomi-3-0:
Add links (e.g https://youtu.be/kh7f_Mvgpak?feature=shared) to presentations and slides by Juno team members in the docs to help improve the docs.

LordGhostX:
Closing this issue as I could only find 3 videos on YouTube, and they don't offer any additional information beyond what's already available in the documentation.

- https://www.youtube.com/watch?v=kh7f_Mvgpak
- https://www.youtube.com/watch?v=IxEB_2GSpGA
- https://www.youtube.com/watch?v=2ZrrltCIBug

Maybe when there are more videos, I'll open this issue and create a **Presentations** page.

Issue #1393, opened 2023-10-30
Blogpost: "How to setup and run a Juno node?"


stranger80:
Publish a tutorial/walkthrough blogpost explaining the actions required to run a Juno node.
This is to be published alongside the P2P MVP release and should include P2P node operation aspects (eg. network setup/connectivity required?)

stranger80:
Draft: https://www.notion.so/nethermind/Juno-p2p-overview-ef846befb43444e191db30120d18ea66

JorikSchellekens:
I think this is stale, shall we close it?

stranger80:
not really, this is still chased by Marketing team, and feels worth the time, as per https://www.notion.so/nethermind/Blog-post-part-1-How-to-juno-in-p2p-mode-b12caed7645f4bad80e0ffff9d112019 

The current idea is to publish the current state of P2P as "experimental feature" (not for prod use, and with caveats, etc.), and coordinate the publish with this blogpost - how does that sound?

PR #1457, opened 2023-11-23
docs: added a user_guide.md file

Tomi-3-0:
#1358

### DIFF
--- a/docs/docs/intro.md
+++ b/docs/docs/intro.md
@@ -1,6 +1,6 @@
 ---
-slug: /
-sidebar_position: 1
+slug: /intro
+sidebar_position: 2
 title: Quick Start
 ---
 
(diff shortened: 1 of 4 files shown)


joshklop (commented):
Thank you for the effort here. The docs are very well-written. I made some simple comments  just in case we want to move forward.

Having read this through again, I am a bit hesitant to merge it. Our docs are supposed to tell users *how to use* Juno, not about why it's important or what it does under the hood. Maybe we can re-use this work to satisfy the blog post issue, which may be a better place for this kind of content https://github.com/NethermindEth/juno/issues/1393. Wdyt?

Tomi-3-0:
> Thank you for the effort here. The docs are very well-written. I made some simple comments just in case we want to move forward.
> 
> Having read this through again, I am a bit hesitant to merge it. Our docs are supposed to tell users _how to use_ Juno, not about why it's important or what it does under the hood. Maybe we can re-use this work to satisfy the blog post issue, which may be a better place for this kind of content #1393. Wdyt?

That makes sense

PR #1461, opened 2023-11-24
Add ChainID bucket

joshklop:


migration/migration.go
```diff
@@ -542,3 +543,65 @@ func migrateTrieNodesFromBitsetToTrieKey(target db.Bucket) BucketMigratorDoFunc
 		return txn.Set(append(orgKeyPrefix, keyBuffer.Bytes()...), tempBuf.Bytes())
 	}
 }
+
+func addNetworkBucket(txn db.Transaction, network utils.Network) error { //nolint:gocyclo
+	mainnetGenesisBlockHash, err := new(felt.Felt).SetString("47c3637b57c2b079b93c61539950c17e868a28f46cdef28f88521067f21e943")
+	if err != nil {
+		return fmt.Errorf("mainnet block hash: %v", err)
```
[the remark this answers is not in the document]
joshklop: I chose to err on the safe side (pun intended) and go with `%v` so I didn't have to test possible exported errors. From the go [blog post](https://go.dev/blog/go1.13-errors) on errors:

> wrapping an error makes that error part of your API. If you don’t want to commit to supporting that error as part of your API in the future, you shouldn’t wrap the error.

migration/migration.go
```diff
@@ -542,3 +543,65 @@ func migrateTrieNodesFromBitsetToTrieKey(target db.Bucket) BucketMigratorDoFunc
 		return txn.Set(append(orgKeyPrefix, keyBuffer.Bytes()...), tempBuf.Bytes())
 	}
 }
+
+func addNetworkBucket(txn db.Transaction, network utils.Network) error { //nolint:gocyclo
```
omerfirmak, line 547: I don't we need to do this. Especially since it wont work for custom networks. We can always populate the `Network` bucket (if empty of course) when we are booting up.

omerfirmak: > What about for existing db's?

We can assume user will provide the correct network the first time Juno is restarted.

> I think the migration should always work, even after we start supporting custom networks. Am I missing something?

https://github.com/NethermindEth/juno/pull/1461/files#diff-d71e3f116a6ab5e5b759bc2e997dd9c9a77c30eb037e59ffa612a25a18925ad8R548-R571

Genesis hashes are hard coded, how would it work for arbitrary networks?

omerfirmak: > Genesis hashes are hard coded, how would it work for arbitrary networks?

Oh, just realized , since custom network support will be introduced after this feature we will never run this migration

joshklop: We would on startup, but we can still use the network name, chain ID, or something else to insert in the `Network` bucket.

joshklop: > We can assume user will provide the correct network the first time Juno is restarted and write that to the DB.

We do this in the migration.

omerfirmak:
I would have preferred a more straightforward approach. For example; hash any block from storage with the provided chain id and make sure it yields the expected result.

PR #1464, opened 2023-11-24
Update Trace API for 0.6

omerfirmak:


vm/rust/src/jsonrpc.rs
```diff
@@ -203,18 +244,15 @@ pub struct FunctionInvocation {
     pub calls: Vec<FunctionInvocation>,
     pub events: Vec<OrderedEvent>,
     pub messages: Vec<OrderedMessage>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub execution_resources: Option<ExecutionResources>,
```
omerfirmak, line 248: We are missing this when we adapt traces from the feeder. Do you mind adding that on top of this PR @joshklop?

joshklop (approved):

vm/rust/src/jsonrpc.rs
joshklop: Will do.

vm/rust/src/jsonrpc.rs
```diff
@@ -185,12 +185,53 @@ use blockifier::execution::call_info::OrderedEvent as BlockifierOrderedEvent;
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub ecdsa_builtin_applications: Option<usize>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub bitwise_builtin_applications: Option<usize>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub keccak_builtin_applications: Option<usize>,
+    // https://github.com/starkware-libs/starknet-specs/pull/167
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub segment_arena_builtin: Option<usize>,
+}
```
omerfirmak, line 215: Also, these must be marshaled in to a hex string if json is legacy :/

joshklop: That's a nice catch

omerfirmak: I think I got it confused sorry:/ Trace API v0.5 doesnt have execution resources, so there is no need to have the ability to marshal these into hex strings.

But we do need to support that for receipts.

PR #1467, opened 2023-11-26
Database: Configurable cache size

### LINKED ISSUE #1458: Add customizable cache size option for pebble
The fixed 8MB cache size in Juno's Pebble database seems to be a performance bottleneck, particularly for starknet_call which are db-intensive.
```
From pebble.Options
	// Cache is used to cache uncompressed blocks from sstables.
	//
	// The default cache size is 8 MB.
	Cache *cache.Cache
```

**Proposed Solution:**
Implement a configurable cache size for Pebble, similar to [Geth's approach](https://github.com/ethereum/go-ethereum/blob/5ff929c22fb0315ebdb6efc5a871469f70522850/ethdb/pebble/pebble.go#L183-L222). This would allow us to adjust the cache size dynamically, potentially improving performance.

**First Step:**
- Add a command-line option (`--cache-size-mb`) to set the cache size in Pebble.

Exca-DK:
Closes #1458

### DIFF
--- a/cmd/juno/juno.go
+++ b/cmd/juno/juno.go
@@ -61,6 +61,7 @@ const (
 	maxVMQueueF          = "max-vm-queue"
 	remoteDBF            = "remote-db"
 	rpcMaxBlockScanF     = "rpc-max-block-scan"
+	cacheSizeMbF         = "cache-size-mb"
 
 	defaultConfig              = ""
 	defaulHost                 = "localhost"
@@ -82,6 +83,7 @@ const (
 	defaultGRPCPort            = 6064
 	defaultRemoteDB            = ""
 	defaultRPCMaxBlockScan     = math.MaxUint
+	defaultCacheSizeMb         = 8
 
 	configFlagUsage   = "The yaml configuration file."
 	logLevelFlagUsage = "Options: debug, info, warn, error."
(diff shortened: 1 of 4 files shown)


cmd/juno/juno.go
```diff
@@ -61,6 +61,7 @@ const (
 	maxVMQueueF          = "max-vm-queue"
 	remoteDBF            = "remote-db"
 	rpcMaxBlockScanF     = "rpc-max-block-scan"
+	cacheSizeMbF         = "cache-size-mb"
```
omerfirmak, line 64:
```suggestion
	cacheSizeMbF         = "db-cache-size-mb"
```

cmd/juno/juno.go
```diff
@@ -113,6 +115,7 @@ const (
 	maxVMQueueUsage          = "Maximum number for requests to queue after reaching max-vms before starting to reject incoming requets"
 	remoteDBUsage            = "gRPC URL of a remote Juno node"
 	rpcMaxBlockScanUsage     = "Maximum number of blocks scanned in single starknet_getEvents call"
+	cacheSizeUsage           = "Determines the amount of memory allocated for caching data in the database."
```
omerfirmak, line 118:
```suggestion
	cacheSizeUsage           = "Determines the amount of memory allocated for caching data in the database. (in megabytes)"
```

PR #1470, opened 2023-11-27
Recovery of cairo0 hash logic

kirugan:


omerfirmak:
To sanity check the changes here, can you enable Cairo0 class verification in syncing and start a fresh sync? You need to modify `VerifyClassHashes`

``` golang
func VerifyClassHashes(classes map[felt.Felt]Class) error {
	for hash, class := range classes {
		cairo1Class, ok := class.(*Cairo1Class)
		// cairo0 classes are deprecated and hard to verify their hash, just ignore them
		if !ok {
			return nil
		}

		cHash := cairo1Class.Hash()
		if !cHash.Equal(&hash) {
			return fmt.Errorf("cannot verify class hash: calculated hash %v, received hash %v", cHash.String(), hash.String())
		}
	}
	return nil
}
```

omerfirmak:
oh, that `return nil` below was a mistake :/ It should've been a `continue`

``` golang
		cairo1Class, ok := class.(*Cairo1Class)
		// cairo0 classes are deprecated and hard to verify their hash, just ignore them
		if !ok {
			return nil
		}
```

PR #1475, opened 2023-11-28
Introduce rpc.ErrTransactionExecutionError

omerfirmak:


rpc/handlers.go
```diff
@@ -1537,7 +1560,7 @@ func (h *Handler) traceBlockTransactions(ctx context.Context, block *core.Block,
 		if errors.Is(err, utils.ErrResourceBusy) {
 			return nil, ErrInternal.CloneWithData(err.Error())
 		}
-		return nil, makeContractError(err)
+		return nil, ErrUnexpectedError.CloneWithData(err.Error())
```
joshklop, line 1563: Is `ErrUnexpectedError` only for gateway errors?

rpc/handlers.go
```diff
@@ -1404,7 +1423,11 @@ func (h *Handler) simulateTransactions(id BlockID, transactions []BroadcastedTra
 		if errors.Is(err, utils.ErrResourceBusy) {
 			return nil, ErrInternal.CloneWithData(err.Error())
 		}
-		return nil, makeContractError(err)
+		var txnExecutionError vm.TransactionExecutionError
+		if errors.As(err, &txnExecutionError) {
+			return nil, makeTransactionExecutionError(&txnExecutionError)
+		}
+		return nil, ErrUnexpectedError.CloneWithData(err.Error())
```
joshklop, line 1430: Same here: should this be `ErrInternal`?

rpc/handlers.go
```diff
@@ -1402,7 +1402,7 @@ func (h *Handler) simulateTransactions(id BlockID, transactions []BroadcastedTra
 		state, h.network, paidFeesOnL1, skipFeeCharge, skipValidate, header.GasPrice, header.GasPriceSTRK, legacyTraceJSON)
 	if err != nil {
 		if errors.Is(err, utils.ErrResourceBusy) {
-			return nil, ErrUnexpectedError.CloneWithData(err.Error())
+			return nil, ErrInternal.CloneWithData(err.Error())
```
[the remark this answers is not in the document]
omerfirmak: Mainly because `utils.ErrResourceBusy` is an internal error to Juno, it has nothing to do with user request and/or Starknet RPC

rpc/handlers.go
omerfirmak: Trace is not expected to return ContractError anymore.
When tracing an existing block, we expect vm.Execute to never return an error. Since we know sequencer executed those transactions just fine. Hence, the `ErrUnexpectedError` if vm.Execute returns an unexpected error.

vm/errors.go
```diff
@@ -0,0 +1,16 @@
...
+	Index uint64
+	Cause error
+}
+
+func (e TransactionExecutionError) Unwrap() error {
+	return e.Cause
+}
+
+func (e TransactionExecutionError) Error() string {
+	return fmt.Sprintf("Failed executing transaction #%d: %s", e.Index, e.Cause)
```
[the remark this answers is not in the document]
omerfirmak: done

rpc/handlers.go
joshklop: Ah, that's very good reasoning. Mind adding that in a comment? It doesn't immediately occur to me when reading the code.

joshklop (approved):

PR #1476, opened 2023-11-28
Revert back compiled class struct + migration

kirugan:


migration/migration.go
```diff
@@ -60,6 +64,71 @@ var defaultMigrations = []Migration{
...
+		External    []core.SierraEntryPoint
+		L1Handler   []core.SierraEntryPoint
+	}
+	Program         []*felt.Felt
+	ProgramHash     *felt.Felt
+	SemanticVersion string
+	Compiled        json.RawMessage
+}
+
+func migrateCairo1CompiledClass(txn db.Transaction, key, value []byte, _ utils.Network) error {
```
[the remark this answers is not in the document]
kirugan: Added

IronGauntlets (approved):
LGTM!

I left some small comments.

Let's not merge this PR until we release v0.8.0 to avoid migration headaches for our prod environments. When we merge we should add Mohammed as a co-author especially when squash and merging (he may get some tokens).

Also, what was the issue you were facing with migration?

IronGauntlets:
I am going to merge this branch into the p2p branch

adapters/core2sn/class.go
```diff
@@ -2,12 +2,78 @@ package core2sn
...
 	"github.com/NethermindEth/juno/starknet"
+	"github.com/NethermindEth/juno/utils"
 )
 
-func AdaptSierraEntryPoint(ep core.SierraEntryPoint) starknet.SierraEntryPoint {
-	return starknet.SierraEntryPoint{
-		Index:    ep.Index,
-		Selector: ep.Selector,
+func AdaptCompiledClass(coreCompiledClass *core.CompiledClass) starknet.CompiledClass {
+	feederCompiledClass := new(starknet.CompiledClass)
```
omerfirmak, line 11: Let's not heap allocate this if we are going to return a value from this function by de-referencing it.

adapters/sn2core/sn2core.go
```diff
@@ -281,28 +281,52 @@ func AdaptCairo1Class(response *starknet.SierraDefinition, compiledClass json.Ra
...
 			return nil, err
 		}
+		class.Compiled = *adaptedCompiledClass
 	}
+
 	return class, nil
 }
 
+func AdaptCompiledClass(compiledClass *starknet.CompiledClass) (*core.CompiledClass, error) {
+	compiled := new(core.CompiledClass)
```
omerfirmak, line 295: same here, unnecessarily allocated on heap. Caller immediately de-references it.

core/class.go
```diff
@@ -61,7 +62,24 @@ type Cairo1Class struct {
 	Program         []*felt.Felt
 	ProgramHash     *felt.Felt
 	SemanticVersion string
-	Compiled        json.RawMessage
+	Compiled        CompiledClass
+}
+
+type CompiledClass struct {
+	Bytecode        []*felt.Felt
+	PythonicHints   json.RawMessage
```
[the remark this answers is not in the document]
omerfirmak: I think it is important to signify to the reader that the value stored there is not some arbitrary byte array but valid JSON.

adapters/core2sn/class.go
IronGauntlets: I don't this is necessarily true. IIUC The reference doesn't escape the function so it won't be allocated on the heap

adapters/sn2core/sn2core.go
IronGauntlets: same as above

core/class.go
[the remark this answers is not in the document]
kirugan: > that the value stored there is not some arbitrary byte array but valid JSON.

I agree with Omer here, IMO right now we do not rely on a lot of external types in `core` and since `json.RawMessage` is just stdlib type over `[]byte` it has no overhead and shouldn't cause much trouble (like cyclic dependency)

adapters/core2sn/class.go
kirugan: The number of allocations will be the same (I benchmarked it with `b.ReportAllocs()`. However, in C++ there is optimization called RVO (return value optimization) and NRVO (named return value optimization) which essentially constructs an object in place of caller (without copying). I don't know whether the Go compiler uses it, but just in case I rewrote this function.

adapters/sn2core/sn2core.go
kirugan: In this case there is actually -1 allocation if there is no `new`. Rewrote it

vm/class.go
```diff
@@ -2,30 +2,27 @@ package vm
...
+		compiledCairo0Class, err := core2sn.AdaptCairo0Class(c)
 		if err != nil {
 			return nil, err
 		}
+		return json.Marshal(compiledCairo0Class)
 	case *core.Cairo1Class:
-		compiledClass = c.Compiled
+		// we adapt the core type to the feeder type to avoid using JSON tags in core.Class.CompiledClass
+		compiledCairo1Class := core2sn.AdaptCompiledClass(&c.Compiled)
+		return json.Marshal(compiledCairo1Class)
```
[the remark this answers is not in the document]
kirugan: rewrote this part

core/class_test.go
```diff
@@ -58,8 +58,8 @@ func TestClassV1Hash(t *testing.T) {
 		classHash string
 	}{
 		{
-			// https://external.integration.starknet.io/feeder_gateway/get_class_by_hash?classHash=0x4e70b19333ae94bd958625f7b61ce9eec631653597e68645e13780061b2136c
-			classHash: "0x4e70b19333ae94bd958625f7b61ce9eec631653597e68645e13780061b2136c",
+			// https://external.integration.starknet.io/feeder_gateway/get_class_by_hash?classHash=0x1cd2edfb485241c4403254d550de0a097fa76743cd30696f714a491a454bad5
+			classHash: "0x1cd2edfb485241c4403254d550de0a097fa76743cd30696f714a491a454bad5",
```
omerfirmak, line 62: why did we change this?

Issue #1477, opened 2023-11-28
Refactor `rpc.AddTransaction` to not work on raw json.


omerfirmak:
It should accept `BroadcastedTransaction` as an argument

PR #1478, opened 2023-11-28
Fix AddTransaction looking for a class declaration on non-declare txns

omerfirmak:


rpc/handlers.go
```diff
@@ -1067,15 +1067,15 @@ func (h *Handler) AddTransaction(txnJSON json.RawMessage) (*AddTxResponse, *json
 		return nil, jsonrpc.Err(jsonrpc.InvalidJSON, err.Error())
 	}
 
-	if txnType, typeFound := request["type"]; typeFound && txnType == TxnInvoke.String() {
+	if txnType := request["type"]; txnType == TxnInvoke.String() {
```
wojciechos, line 1070: I also noticed that INVOKE_TXN_V0 also requires "contract_class": https://github.com/starkware-libs/starknet-specs/blob/v0.6.0-rc3/api/starknet_api_openrpc.json#L2482

wojciechos: I'm also wondering, why part of validation we have here, and other here? 
https://github.com/NethermindEth/juno/blob/9448c9f56c33196c8aacc1d86c3cd23d3d7b424b/rpc/transaction.go#L215

rpc/handlers.go
```diff
@@ -1067,15 +1067,15 @@ func (h *Handler) AddTransaction(txnJSON json.RawMessage) (*AddTxResponse, *json
...
 		request["type"] = starknet.TxnInvoke.String()
 
 		updatedReq, errIn := json.Marshal(request)
 		if errIn != nil {
 			return nil, jsonrpc.Err(jsonrpc.InternalError, errIn.Error())
 		}
 		txnJSON = updatedReq
-	} else if version, ok := request["version"]; ok && (version == "0x2" || version == "0x3") {
+	} else if version := request["version"]; txnType == TxnDeclare.String() && (version == "0x2" || version == "0x3") {
 		contractClass, ok := request["contract_class"].(map[string]interface{})
```
wojciechos, line 1079: I'm also wondering, why part of validation we have like above, and other like that? 
https://github.com/NethermindEth/juno/blob/9448c9f56c33196c8aacc1d86c3cd23d3d7b424b/rpc/transaction.go#L215

rpc/handlers.go
omerfirmak: That is `contract_address`

rpc/handlers.go
omerfirmak: No particular reason, it is tech debt that we will have to address at some point. https://github.com/NethermindEth/juno/issues/1477

wojciechos (approved):

PR #1481, opened 2023-11-28
FRI related 0.6 changes

omerfirmak:


rpc/block.go
```diff
@@ -74,6 +74,7 @@ func (b *BlockID) UnmarshalJSON(data []byte) error {
 
 type ResourcePrice struct {
 	InStark *felt.Felt `json:"price_in_strk"`
+	InFri   *felt.Felt `json:"price_in_fri"`
```
joshklop, line 77: ~I think we should remove the `InStark` ([ref](https://github.com/starkware-libs/starknet-specs/blob/50f8f97facc36e6096d84c74b8316b153df4e653/api/starknet_api_openrpc.json#L3752-L3761)).~

Ah never mind, I see it's left for compatibility.

joshklop: Should they be `omitempty` though?

joshklop (approved):

rpc/block.go
omerfirmak: Yeap, nice catch. Both should be `omitempty`.

Issue #1428, opened 2023-11-15
Create `rpc2core` / `core2rpc` packages under `adapters` directory


joshklop:
Right now RPC adapters are cluttering the `rpc` package.

PR #1484, opened 2023-11-28
Unmarshal AddTx request into BroadcastTx

joshklop:


omerfirmak (approved):
:100: 

Let's prioritize https://github.com/NethermindEth/juno/issues/1428 after the release

PR #1486, opened 2023-11-28
Validate the --network option

joshklop:


joshklop:
Hit another roadblock with this approach @omerfirmak. `core.VerifyBlockHash` doesn't verify block hashes before `0.11.0` (because it can't in some cases).

I'm leaning toward sticking with #1461. What do you think?

node/node.go
```diff
@@ -118,6 +119,25 @@ func New(cfg *Config, version string) (*Node, error) { //nolint:gocyclo,funlen
...
+	height, err := chain.Height()
+	if err != nil && !errors.Is(err, db.ErrKeyNotFound) {
+		return nil, fmt.Errorf("get chain height from database: %v", err)
+	}
+	if height > 0 {
+		var block *core.Block
+		block, err = chain.BlockByNumber(height)
+		if err != nil {
+			return nil, fmt.Errorf("get latest block from database: %v", err)
+		}
```
omerfirmak, line 133: Any reason we dont use `Head()` here?

omerfirmak:
> Hit another roadblock with this approach @omerfirmak. `core.VerifyBlockHash` doesn't verify block hashes before `0.11.0` (because it can't in some cases).

It is not the block hashes but the txn hashes that we are not validating pre-0.11.0. But that is not really important for SW operated networks, because they all either have pre-0.7 genesis or post-0.11.0 genesis. 

> not validating txn hashes pre-0.11.0

We will have to add support for this for p2p btw.

joshklop:
The safegaurd implemented by this PR won't work for certain networks in their `UnverifiableRange`:

The below will not work:
 
* switching from any network to integration in block range 0-110511
* switching from any network to goerli in block range 119802-148428

node/node.go
```diff
@@ -118,6 +119,25 @@ func New(cfg *Config, version string) (*Node, error) { //nolint:gocyclo,funlen
...
+		// We assume there is at least one transaction in the blockchain, otherwise VerifyBlockHash
+		// will succeed for any network with only post-0.7 blocks using the standard fallback sequencer address.
```
omerfirmak, line 135: > will succeed for any network with only post-0.7 blocks

Is that correct? post-0.7 blocks all have ChainID in their hash.

omerfirmak:
I believe that is fine. IMO, It is not worth to the additional complexity. Let's bring this up in Slack to see what others think?

node/node.go
joshklop: Pre-0.7 blocks [have](https://github.com/NethermindEth/juno/blob/27c1ac20c794d8c7c95c71daa0669415b628669f/core/block.go#L183) the chain id in their hash, post-0.7 blocks [do not](https://github.com/NethermindEth/juno/blob/27c1ac20c794d8c7c95c71daa0669415b628669f/core/block.go#L214-L231).

PR #1488, opened 2023-11-29
Write proper adapters for core2vm

omerfirmak:
instead of using `copier`


vm/transaction.go
```diff
@@ -58,3 +50,171 @@ func marshalTxn(txn core.Transaction) (json.RawMessage, error) {
...
+	Hash                  *felt.Felt                   `json:"transaction_hash,omitempty" copier:"must,nopanic"`
+	Version               *felt.Felt                   `json:"version,omitempty"`
+	ContractAddress       *felt.Felt                   `json:"contract_address,omitempty"`
+	ContractAddressSalt   *felt.Felt                   `json:"contract_address_salt,omitempty"`
+	ClassHash             *felt.Felt                   `json:"class_hash,omitempty"`
+	ConstructorCallData   *[]*felt.Felt                `json:"constructor_calldata,omitempty"`
+	SenderAddress         *felt.Felt                   `json:"sender_address,omitempty"`
+	MaxFee                *felt.Felt                   `json:"max_fee,omitempty"`
+	Signature             *[]*felt.Felt                `json:"signature,omitempty"`
+	CallData              *[]*felt.Felt                `json:"calldata,omitempty"`
```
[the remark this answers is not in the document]
omerfirmak: We have cases where we want to have an empty list in JSON

vm/transaction.go
```diff
@@ -58,3 +50,171 @@ func marshalTxn(txn core.Transaction) (json.RawMessage, error) {
...
+
+		if tx.Version.Uint64() == 3 {
+			tx.ResourceBounds = utils.Ptr(adaptResourceBounds(t.ResourceBounds))
+			tx.Tip = new(felt.Felt).SetUint64(t.Tip)
+			tx.PaymasterData = &t.PaymasterData
+			tx.AccountDeploymentData = &t.AccountDeploymentData
+			tx.NonceDAMode = utils.Ptr(DataAvailabilityMode(t.NonceDAMode))
+			tx.FeeDAMode = utils.Ptr(DataAvailabilityMode(t.FeeDAMode))
+		}
+	case *core.InvokeTransaction:
```
joshklop, line 148: Do we need to special-case invoke v0s?

vm/transaction.go
```diff
@@ -58,3 +50,171 @@ func marshalTxn(txn core.Transaction) (json.RawMessage, error) {
 	}
 	return result, nil
 }
+
+type Transaction struct {
+	Hash                  *felt.Felt                   `json:"transaction_hash,omitempty" copier:"must,nopanic"`
```
joshklop, line 55: Do we still need the copier tag? We can remove it from the transaction struct in `starknet/transaction.go` I think.

vm/transaction.go
```diff
@@ -58,3 +50,171 @@ func marshalTxn(txn core.Transaction) (json.RawMessage, error) {
...
+			tx.FeeDAMode = utils.Ptr(DataAvailabilityMode(t.FeeDAMode))
+		}
+	case *core.DeployTransaction:
+		return &Transaction{
+			Hash:                t.Hash(),
+			ClassHash:           t.ClassHash,
+			Version:             t.Version.AsFelt(),
+			ContractAddressSalt: t.ContractAddressSalt,
+			ConstructorCallData: &t.ConstructorCallData,
+		}
```
joshklop, line 176:
```suggestion
		return &Transaction{
			Hash:                t.Hash(),
			ClassHash:           t.ClassHash,
			Version:             t.Version.AsFelt(),
			ContractAddressSalt: t.ContractAddressSalt,
			ConstructorCallData: &t.ConstructorCallData,
			ContractAddress:       t.ContractAddress,
		}
```

joshklop (approved):

vm/transaction.go
omerfirmak: Removed.

omerfirmak: Actually, we can get rid of hash field all together.

vm/transaction.go
omerfirmak: https://github.com/NethermindEth/juno/pull/1488/files#diff-a060f37d0efb55c39bb433ff6ccaa6116ac7fa1462a49d28c64719a010e16373L32-L36 was a workaround that we don't need any more.

PR #1496, opened 2023-11-30
Fix using ETH price in FRI estimates

omerfirmak:


rpc/handlers.go
```diff
@@ -1463,14 +1463,24 @@ func (h *Handler) simulateTransactions(id BlockID, transactions []BroadcastedTra
...
 	var result []SimulatedTransaction
 	for i, overallFee := range overallFees {
+		feeUnit := feeUnit(txns[i])
+
+		gasPrice := header.GasPrice
+		if feeUnit == FRI {
+			gasPrice = header.GasPriceSTRK
+			if gasPrice == nil {
+				gasPrice = &felt.Zero
+			}
```
joshklop, line 1473: Should this go outside the enclosing `if`?

joshklop (approved):

rpc/handlers.go
omerfirmak: `GasPrice` is known to be non-nil in all blocks AFAIK

PR #1509, opened 2023-12-02
Discard fee unit in legacy estimate message fee

joshklop:


omerfirmak (approved):

rpc/handlers.go
```diff
@@ -1340,6 +1340,15 @@ func (h *Handler) EstimateMessageFee(msg MsgFromL1, id BlockID) (*FeeEstimate, *
 	return &estimates[0], nil
 }
 
+func (h *Handler) LegacyEstimateMessageFee(msg MsgFromL1, id BlockID) (*FeeEstimate, *jsonrpc.Error) { //nolint:gocritic
+	estimate, rpcErr := h.EstimateMessageFee(msg, id)
```
omerfirmak, line 1344: If you call LegacyEstimateMessageFee here, you wont need to explicitly nillify Unit.

PR #1511, opened 2023-12-02
Allow for 10MB messages over grpc

joshklop:
Got an error message that looked something like this:
```
JunoStateGetCompiledClass failed to read class
{
    "err": "rpc error: code = ResourceExhausted desc = grpc: received message larger than max (5877256 vs. 4194304)"
}
github.com/NethermindEth/juno/vm.JunoStateGetCompiledClass/home/user/repos/juno/vm/state_reader.go:80
```

https://stackoverflow.com/questions/55362342/grpc-received-message-larger-than-max-8653851-vs-4194304

### DIFF
--- a/db/remote/db.go
+++ b/db/remote/db.go
@@ -34,7 +34,12 @@ func New(rawURL string, ctx context.Context, log utils.SimpleLogger, opts ...grp
 }
 
 func (d *DB) NewTransaction(write bool) (db.Transaction, error) {
-	txClient, err := d.kvClient.Tx(d.ctx)
+	const (
+		megabyte = 1 << 20
+		// Some classes are larger than the default of 4MB.
+		maxCallMsgSize = 10 * megabyte
+	)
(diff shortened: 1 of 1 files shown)


db/remote/db.go
```diff
@@ -34,7 +34,12 @@ func New(rawURL string, ctx context.Context, log utils.SimpleLogger, opts ...grp
 }
 
 func (d *DB) NewTransaction(write bool) (db.Transaction, error) {
-	txClient, err := d.kvClient.Tx(d.ctx)
+	const (
+		megabyte = 1 << 20
+		// Some classes are larger than the default of 4MB.
+		maxCallMsgSize = 10 * megabyte
+	)
+	txClient, err := d.kvClient.Tx(d.ctx, grpc.MaxCallSendMsgSize(maxCallMsgSize), grpc.MaxCallRecvMsgSize(maxCallMsgSize))
```
omerfirmak, line 42: any configuration needed on the server side?

PR #1518, opened 2023-12-04
Add errOnRevert flag to vm.Execute

omerfirmak:
to treat reverted txns as hard-failures

### DIFF
--- a/vm/rust/src/lib.rs
+++ b/vm/rust/src/lib.rs
@@ -156,6 +156,7 @@ pub extern "C" fn cairoVMExecute(
     paid_fees_on_l1_json: *const c_char,
     skip_charge_fee: c_uchar,
     skip_validate: c_uchar,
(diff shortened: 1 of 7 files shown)


vm/rust/src/lib.rs
```diff
@@ -283,6 +284,16 @@ pub extern "C" fn cairoVMExecute(
                 return;
             }
             Ok(mut t) => {
+                if t.is_reverted() && err_on_revert != 0{
+                    report_error(
+                        reader_handle,
+                        format!("reverted: {}",t.revert_error.unwrap())
```
joshklop, line 290:
```suggestion
                if t.is_reverted() && err_on_revert != 0 {
                    report_error(
                        reader_handle,
                        format!("reverted: {}", t.revert_error.unwrap())
```

joshklop (approved):

Issue #1512, opened 2023-12-02
Active connections metric


Exca-DK:
## Issue
Juno now has persistent connections (eg. ws conns) with not much info about them. It would be nice if It was known to user how many of them there are. 

## Implementation
Do you have ideas regarding the implementation of this feature? yes
Are you willing to implement this feature? yes

**Scope:**

- Using gauge for tracking the amount of connections
- Extending `NewRequestListener` with `OnNewConnection(conn net.Conn)` and `OnDisconnect(conn net.Conn)`.
- Implementing the logic in each rpc handler.
## Additional Information:
- `OnNewConnection(conn net.Conn)` and `OnDisconnect(conn net.Conn)` would be a nice place for uniform logging eg. ("Accepted connection", "conn", conn.RemoteAddr()) 
- `OnNewConnection(conn net.Conn)` could be further extended into `OnNewConnection(conn net.Conn) net.Conn` which would enable tracking how much bytes were sent/recv.

joshklop:
> Implementing the logic in each rpc handler.

We would only need to implement this in the websocket transport in the `jsonrpc` package, correct? No need to modify the higher-level `rpc` package handlers.

Exca-DK:
yes, I meant `jsonrpc`

PR #1520, opened 2023-12-04
feat: Track amount of rpc connections

Exca-DK:
Implements #1512

### DIFF
--- a/jsonrpc/event_listener.go
+++ b/jsonrpc/event_listener.go
@@ -1,9 +1,14 @@
 package jsonrpc
 
-import "time"
+import (
+	"net"
+	"time"
+)
 
 type NewRequestListener interface {
 	OnNewRequest(method string)
+	OnNewConnection(conn net.Conn)
+	OnDisconnect(conn net.Conn)
 }
 
 type EventListener interface {
@@ -14,6 +19,8 @@ type EventListener interface {
 
 type SelectiveListener struct {
 	OnNewRequestCb     func(method string)
+	OnNewConnectionCb  func(conn net.Conn)
+	OnDisconnectCb     func(conn net.Conn)
 	OnRequestHandledCb func(method string, took time.Duration)
 	OnRequestFailedCb  func(method string, data any)
 }
@@ -21,6 +28,18 @@ type SelectiveListener struct {
 func (l *SelectiveListener) OnNewRequest(method string) {
 	if l.OnNewRequestCb != nil {
 		l.OnNewRequestCb(method)
+	}
+}
+
+func (l *SelectiveListener) OnNewConnection(conn net.Conn) {
+	if l.OnNewConnectionCb != nil {
+		l.OnNewConnectionCb(conn)
+	}
+}
+
+func (l *SelectiveListener) OnDisconnect(conn net.Conn) {
+	if l.OnDisconnectCb != nil {
+		l.OnDisconnectCb(conn)
 	}
 }
 
--- a/jsonrpc/event_listener.go
+++ b/jsonrpc/event_listener.go
@@ -1,14 +1,17 @@
 package jsonrpc
 
 import (
-	"net"
 	"time"
 )
 
+type NewConnectionListener interface {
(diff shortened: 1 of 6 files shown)


jsonrpc/event_listener.go
```diff
@@ -1,9 +1,14 @@
...
 
-import "time"
+import (
+	"net"
+	"time"
+)
 
 type NewRequestListener interface {
 	OnNewRequest(method string)
+	OnNewConnection(conn net.Conn)
```
joshklop, line 10: It seems dangerous to allow `NewRequestListener` implementations to arbitrarily read, write, or close the connection. Do you think it is sufficient to pass in `remoteAddr string` and then include that as a label in our metrics?

Exca-DK: I don't think prometheus will be happy about having so many unique labels and it will be super hard for visualization in grafana. I think that `remoteAddr` should fall into logging instead and then it should be implemented for every handler instead.

Exca-DK: If it's too dangerous then I think that it can be further split into:
- OnNewConnection(), OnDisconnect() - metric for amount of connections. could even be done for all handlers (http,ws,...)
- OnRead, OnWrite - once again, for all handlers - super usefull for tracking egress/ingress since it's very expensive of `aws/gc`

joshklop: > OnNewConnection(), OnDisconnect() - metric for amount of connections. could even be done for all handlers (http,ws,...)

I think `OnNewConnection()`/`OnDisconnect` makes sense for websockets and ipc. Are they redundant with `OnNewRequest`,etc. for http?

> OnRead(amount uint), OnWrite(amount uint)

This is a great idea. We need to work some details out. For example, we read from the connection using `dec.Decode`, which doesn't tell us how much it reads; it may even read past the json request object. Do you mind creating a separate issue for this?

joshklop: I take the point about `decoder.Decode` back: it looks like we can use its `InputOffset` function to determine how much it has read. Regardless, let's reserve the `OnRead`/`OnWrite` idea for a separate PR.

Exca-DK: > I think OnNewConnection()/OnDisconnect makes sense for websockets and ipc. Are they redundant with OnNewRequest,etc. for http?

Not quite since http connection can reuse the same socket. Simply counting `StateNew` 
 and `StateClosed` from that returns the actual amount which can be lower than `OnNewRequest` but never higher
 https://github.com/golang/go/blob/d95e25e83c922aa63fcb1f596d6bdf1786789edb/src/net/http/server.go#L2890C2-L2890C37 
If it's worth it? not sure

node/http.go
```diff
@@ -109,12 +120,12 @@ func makeRPCOverWebsocket(host string, port uint16, servers map[string]*jsonrpc.
 		wsPrefixedPath := strings.TrimSuffix("/ws"+path, "/")
 		mux.Handle(wsPrefixedPath, exactPathServer(wsPrefixedPath, wsHandler))
 	}
-	return makeHTTPService(host, port, cors.Default().Handler(mux))
+	return makeHTTPService(host, port, cors.Default().Handler(mux), nil)
```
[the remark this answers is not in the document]
Exca-DK: Hijacked connections do not transition to state closed.

joshklop (approved):

jsonrpc/event_listener_test.go
```diff
@@ -37,3 +54,11 @@ func (l *CountingEventListener) OnRequestFailed(method string, data any) {
 		data:   data,
 	})
 }
+
+func (l *CountingEventListener) OnNewConnection() {
+	l.OnConnectionCalls++
+}
+
+func (l *CountingEventListener) OnDisconnect() {
+	l.OnConnectionCalls--
```
kirugan, line 63: I'm not sure that we can do ++ or -- here because these methods called concurrently. Should it be atomic types instead ?

PR #1521, opened 2023-12-05
Use ctx in BucketMigrator

kirugan:


migration/bucket_migrator.go
```diff
@@ -72,14 +72,21 @@ func (m *BucketMigrator) Before(_ []byte) error {
...
 		return nil, err
 	}
 
 	for iterator.Seek(m.startFrom); iterator.Valid(); iterator.Next() {
+		select {
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		default:
+			// keep going
+		}
```
joshklop, line 88: I think we would end up logging the innocuous `context.Canceled` error this way. What if we did:
```go
if ctx.Err() != nil {
        return nil
}
```
and then checked `ctx.Err()` at the caller level as well?

joshklop: There is an invariant: if the `metadata.IntermediateState` is nil, then the migration is finished.

kirugan: Regarding logging issue: everything should work fine. Here is migration test that I used:
```
NewBucketMigrator(db.Class, func(tx db.Transaction, k, v []byte, n utils.Network) error {
		fmt.Printf("Migrating %s\n", string(k))
		time.Sleep(time.Second * 3)
		return nil
	}),
```
Here is output
```
Migrating <some_key>
Migrating <some_key>
Migrating <some_key>
Migrating <some_key>
^C14:13:02.853 07/12/2023 +03:00        INFO    node/node.go:264        DB Migration cancelled
```

Also please note that I removed `|| errors.Is(migrationErr, ctx.Err())` case in condition above

joshklop (approved):

migration/bucket_migrator.go
```diff
@@ -72,14 +72,21 @@ func (m *BucketMigrator) Before(_ []byte) error {
...
 		return nil, err
 	}
 
 	for iterator.Seek(m.startFrom); iterator.Valid(); iterator.Next() {
+		select {
+		case <-ctx.Done():
+			return nil, utils.RunAndWrapOnError(iterator.Close, ctx.Err())
+		default:
+			// keep going
+		}
```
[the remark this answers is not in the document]
joshklop: Nice :+1:

IronGauntlets (approved):

weiihann (approved):

PR #1466, opened 2023-11-26
Use maps instead of slices in state diff struct

joshklop:
Makes it much easier to build a state diff incrementally during execution.

PR #1526, opened 2023-12-05
Add block builder state implementation

joshklop:


joshklop:
All functions except for `builder.Update`, `builder.Diff`, and the adapters are probably useless right now. I can delete them if you want, or we can leave them for when we integrate with Blockifier.

omerfirmak:
I think we should reuse `PendingState` for this, as it implements exactly the same logic.

If you feel like #1466 should be merged to allow that, let's get that merged first. Even if with a slow migration.

joshklop:
I have a local branch with similar changes, and modifications to the vm package. Will close this.

Issue #1528, opened 2023-12-06
Sequencer - genesis state definition and validation


rianhughes:
The first part of https://github.com/NethermindEth/juno/issues/1522

Namely, 1) give a concrete definition of the Genesis state, and 2) a set of validation checks

Note: can look at Madaras genesis [doc](https://github.com/keep-starknet-strange/madara/blob/main/docs/genesis.md) and [genesis file](https://github.com/keep-starknet-strange/madara/blob/e1493f0b2bbbb08c271b4c83040bd01695bf3d20/configs/genesis-assets/genesis.json) for inspiration

rianhughes:
type GenesisFile struct {
    ChainID felt.Felt
    InitialSequencerSet []felt.Felt
    State GenesisState
}

type GenesisState struct {
   Classes map[ClassHash]core.Class
   Contracts map[ContractAddress] struct { 
      ClassHash felt.Felt 
      ConstructorArgs []felt.Felt
   }

   // FeeTokenBalance. Requires the feeToken(s) to be deployed at Genesis.
   FeeTokenBalancesWEI map[felt.Felt]felt.Felt
   FeeTokenBalancesSTRK map[felt.Felt]felt.Felt

   // If we want to build the initial state using vm.Call()
   // vm.Call() only takes a single FunctionCall at a time, so we would have to maintain a pending state etc
   // It also doesn't have skipFee flags etc
   Invocations []FunctionCall 
   
   // If we want to build initial state with vm.Execute()
   // vm.Execute() takes an array of transactions as input, so we don't need to maintain pending state etc
   // It also allows users to set gasPrices, and skip validation / feeCharge, etc.
   Txns []Transaction 
   GenesisSkipValidate
   GenesisSkipFeeCharge
   GenesisGasPriceWEI
   GenesisGasPriceSTRK
}

rianhughes:
After internal discussions, we decided to leave these fields out of GenesisState

	// Not necessary, but a convenient way to set ERC20token balances
	ERC20TokenBalances map[TokenContractAddress]struct {
		UserAddress felt.Felt
		Balance     felt.Felt
	}

	// It would be nice if we could also allow users to also set the initial state by either setting
	// 1) FunctionCalls (vm.Call()), or 2) Transactions (vm.Execute())
	// Otherwise users will have to pre-calculate the storage keys and values etc.

	// 1. Setting initial state with vm.Call()
[...]

PR #1530, opened 2023-12-07
Define the genesis state

rianhughes:
closes https://github.com/NethermindEth/juno/issues/1528

### DIFF
--- a/blockbuilder/genesisState.go
+++ b/blockbuilder/genesisState.go
@@ -1,0 +1,46 @@
+package blockbuilder
+
+import (
+	"encoding/json"
+	"errors"
+
+	"github.com/NethermindEth/juno/core"
+	"github.com/NethermindEth/juno/core/felt"
+)
+
+var (
+	ErrChainIDRequired = errors.New("ChainID is required")
+)
+
+type GenesisConfig struct {
+	ChainID                 string           `json:"chain_id"`
+	WhitelistedSequencerSet []felt.Felt      `json:"whitelisted_sequencer_set"`
+	State                   genesisStateData `json:"state"`
+}
+
+type genesisStateData struct {
+	Classes   map[string]genesisClassData    `json:"classes"`
+	Contracts map[string]genesisContractData `json:"contracts"`
+	Storage   map[string][]core.StorageDiff  `json:"storage"`
+}
+
+type genesisClassData struct {
+	Path    string `json:"path"`
+	Version int    `json:"version"`
+}
+
+type genesisContractData struct {
+	ClassHash       felt.Felt    `json:"class_hash"`
+	ConstructorArgs *[]felt.Felt `json:"constructor_args"`
+}
+
+func (g *GenesisConfig) UnmarshalJSON(data []byte) error {
+	type genesisConfig GenesisConfig
+	if err := json.Unmarshal(data, (*genesisConfig)(g)); err != nil {
+		return err
+	}
+	if g.ChainID == "" {
+		return ErrChainIDRequired
+	}
+	return nil
+}
--- a/blockbuilder/genesisState_test.go
+++ b/blockbuilder/genesisState_test.go
@@ -1,0 +1,86 @@
+package blockbuilder_test
+
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/NethermindEth/juno/blockbuilder"
+	"github.com/NethermindEth/juno/core/felt"
+	"github.com/stretchr/testify/require"
+)
+
+func TestGenesisUnmarshal(t *testing.T) {
+
+	genesisJSONStr := `{
+		"chain_id": "SN_GOERLI",
+		"whitelisted_sequencer_set": [
+			"0x1",
+			"0x2"
+		],
+		"state": {
+			"classes": {
+				"0x3": {
+					"path": "./config/account.json",
+					"version": 0
+				}
+			},
+			"contracts": {
+				"0x4": {
+					"class_hash": "0x123",
+					"constructor_args": []
+				  }
+			},
+			"storage": {
+				"0x5": [
+					{
+						"key": "0x0341c1bdfd89f69748aa00b5742b03adbffd79b8e80cab5c50d91cd8c2a79be1",
+						"value": "0x4574686572"
+					}
+				]
+			}
+		}
+	}
+	`
+
+	var genesis blockbuilder.GenesisConfig
+	require.NoError(t, json.Unmarshal([]byte(genesisJSONStr), &genesis))
+
+	t.Run("test chain ID", func(t *testing.T) {
+		require.Equal(t, "SN_GOERLI", genesis.ChainID)
+	})
+
+	t.Run("test whitelisted sequencer set", func(t *testing.T) {
+		require.Equal(t, 2, len(genesis.WhitelistedSequencerSet))
+		require.Equal(t, new(felt.Felt).SetUint64(1).String(), genesis.WhitelistedSequencerSet[0].String())
+	})
+
+	t.Run("test classes", func(t *testing.T) {
+		for classHash, class := range genesis.State.Classes {
+			require.Equal(t, classHash, "0x3")
+			require.Equal(t, class.Path, "./config/account.json")
+			require.Equal(t, class.Version, 0)
+		}
+	})
+
+	t.Run("test contracts", func(t *testing.T) {
+		for contractHash, contract := range genesis.State.Contracts {
+			require.Equal(t, contractHash, "0x4")
+			require.Equal(t, contract.ClassHash.String(), "0x123")
+			require.Equal(t, contract.ConstructorArgs, &[]felt.Felt{})
+		}
+	})
+
+	t.Run("test storage", func(t *testing.T) {
+		for contractAddress, storage := range genesis.State.Storage {
+			require.Equal(t, contractAddress, "0x5")
+			require.Equal(t, storage[0].Key.String(), "0x341c1bdfd89f69748aa00b5742b03adbffd79b8e80cab5c50d91cd8c2a79be1")
+			require.Equal(t, storage[0].Value.String(), "0x4574686572")
+		}
+	})
+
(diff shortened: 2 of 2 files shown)


blockbuilder/genesisState.go
```diff
@@ -0,0 +1,46 @@
...
+type GenesisConfig struct {
+	ChainID                 string           `json:"chain_id"`
+	WhitelistedSequencerSet []felt.Felt      `json:"whitelisted_sequencer_set"`
+	State                   genesisStateData `json:"state"`
+}
+
+type genesisStateData struct {
+	Classes   map[string]genesisClassData    `json:"classes"`
+	Contracts map[string]genesisContractData `json:"contracts"`
+	Storage   map[string][]core.StorageDiff  `json:"storage"`
```
omerfirmak, line 22: I thought we reached consensus on having `FunctionCall`s?

rianhughes: Yup, somehow slipped my mind :D

rianhughes: Added them

blockbuilder/genesisState.go
```diff
@@ -0,0 +1,50 @@
...
+}
+
+func (g *GenesisConfig) UnmarshalJSON(data []byte) error {
+	type genesisConfig GenesisConfig
+	if err := json.Unmarshal(data, (*genesisConfig)(g)); err != nil {
+		return err
+	}
+	if g.ChainID == "" {
+		return ErrChainIDRequired
+	}
```
omerfirmak, line 48: Let's not mix unmarshaling and validation. You can use `validator` package for validation. See https://github.com/NethermindEth/juno/blob/acc9ac0f3c3895b5cacf41741df2d8c47e6da294/rpc/transaction.go#L222-L243

blockbuilder/genesisState_test.go
```diff
@@ -0,0 +1,98 @@
...
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/NethermindEth/juno/blockbuilder"
+	"github.com/NethermindEth/juno/core/felt"
+	"github.com/stretchr/testify/require"
+)
+
+func TestGenesisUnmarshal(t *testing.T) {
```
omerfirmak, line 12: We don't need to test this, we can trust that the standard json library works:)

blockbuilder/genesisState.go
```diff
@@ -0,0 +1,50 @@
...
+type GenesisConfig struct {
+	ChainID                 string           `json:"chain_id"`
+	WhitelistedSequencerSet []felt.Felt      `json:"whitelisted_sequencer_set"`
+	State                   genesisStateData `json:"state"`
+}
+
+type genesisStateData struct {
+	Classes       map[string]genesisClassData    `json:"classes"`
+	Contracts     map[string]genesisContractData `json:"contracts"`
+	Storage       map[string][]core.StorageDiff  `json:"storage"`
```
omerfirmak, line 22: I don't think we have an immediate use for this when we have `FunctionCalls`

blockbuilder/genesisState.go
```diff
@@ -0,0 +1,50 @@
...
+
+type GenesisConfig struct {
+	ChainID                 string           `json:"chain_id"`
+	WhitelistedSequencerSet []felt.Felt      `json:"whitelisted_sequencer_set"`
+	State                   genesisStateData `json:"state"`
+}
+
+type genesisStateData struct {
+	Classes       map[string]genesisClassData    `json:"classes"`
+	Contracts     map[string]genesisContractData `json:"contracts"`
```
omerfirmak, line 21:
```suggestion
	Classes       map[felt.Felt]genesisClassData    `json:"classes"`
	Contracts     map[felt.Felt]genesisContractData `json:"contracts"`
```

blockbuilder/genesisState.go
rianhughes: Removed the Storage field

blockbuilder/genesisState.go
```diff
@@ -0,0 +1,33 @@
+package blockbuilder
+
+import (
+	"github.com/NethermindEth/juno/core/felt"
+	"github.com/go-playground/validator/v10"
```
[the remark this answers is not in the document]
rianhughes: Updated

blockbuilder/genesisState.go
rianhughes: Un-marshalling a map[felt.Felt]struct is giving errors, so we're just going to define the type as a []string, and calculate the classHash ourselves.

rianhughes:
Contracts     map[felt.Felt]genesisContractData `json:"contracts"`
May need changed. Unmarshalling into a map[felt.Felt] doesn't seem to work..

blockbuilder/genesisState.go
```diff
@@ -0,0 +1,33 @@
...
+	"github.com/NethermindEth/juno/validator"
+)
+
+type GenesisConfig struct {
+	ChainID                 string           `json:"chain_id" validate:"required"`
+	WhitelistedSequencerSet []felt.Felt      `json:"whitelisted_sequencer_set"`
+	State                   genesisStateData `json:"state"`
+}
+
+type genesisStateData struct {
```
joshklop, line 14: We should probably export all of the structs needed to build a `GenesisConfig`.

blockbuilder/genesisState.go
```diff
@@ -0,0 +1,33 @@
+package blockbuilder
+
+import (
+	"github.com/NethermindEth/juno/core/felt"
+	"github.com/NethermindEth/juno/validator"
+)
+
+type GenesisConfig struct {
+	ChainID                 string           `json:"chain_id" validate:"required"`
```
joshklop, line 9: I see there was some discussion about testing unmarshalling/validation. We test these things in the rpc package. Should we test this validation here? Some basic cases would be enough to catch the problem with unmarshaling `map[felt.Felt]genesisContractData`, for example.

blockbuilder/genesisState.go
```diff
@@ -0,0 +1,33 @@
...
+type genesisStateData struct {
+	Classes       []string                          `json:"classes"` // []path-to-class.json
+	Contracts     map[felt.Felt]genesisContractData `json:"contracts"`
+	FunctionCalls []functionCall                    `json:"function_calls"`
+}
+
+type genesisContractData struct {
+	ClassHash       felt.Felt   `json:"class_hash"`
+	ConstructorArgs []felt.Felt `json:"constructor_args"`
+}
```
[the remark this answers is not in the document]
joshklop: I think one of y'all suggested that, but I'm not sure we ever made a final decision about it.

blockbuilder/genesisState.go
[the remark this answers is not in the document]
omerfirmak: agreed

blockbuilder/genesisState.go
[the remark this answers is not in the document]
omerfirmak: `deployer_address` is tricky to set in the genesis context tho, no?

blockbuilder/genesisState.go
omerfirmak: > We test these things in the rpc package. Should we test this validation here

We do that because we are trying to replicate an existing schema already. So, testing unmarshaling some sample data we got from another source makes sense. But since we are trying to define a new schema here, handwriting some JSON samples and testing unmarshaling against them doesn't really serve any purpose IMO.

blockbuilder/genesisState.go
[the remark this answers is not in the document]
rianhughes: We could just print the contract addresses we calculate at startup? Eg, maybe we can just set `deployer_address` to zero, and salt to zero / some random number?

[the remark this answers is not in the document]
rianhughes: If we allow them to supply the `salt` value, we can compute a unique address for them, and just print it on startup. This means we can redefine/simplify
`Contracts map[felt.Felt]genesisContractData` 
to 
`Contracts {classHash, constructorArgs, Salt}`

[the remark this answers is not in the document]
rianhughes: I guess we can also allow them to specify `deployer_address`, afaik, we don't actually need this address to be initialised. It's just another argument into the address calculation.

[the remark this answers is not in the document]
omerfirmak: > afaik, we don't actually need this address to be initialised.
Yes, but still it is meant to be in the normal execution. So I am afraid it might come back to bite us at some point.

I don't see a downside of having using hardcode the initial contract addresses.

blockbuilder/genesisState.go
omerfirmak: Feel free to write tests when you deem necessary, more tests don't hurt.

blockbuilder/genesisState.go
```diff
@@ -0,0 +1,33 @@
+package blockbuilder
+
+import (
+	"github.com/NethermindEth/juno/core/felt"
+	"github.com/NethermindEth/juno/validator"
+)
+
+type GenesisConfig struct {
+	ChainID                 string           `json:"chain_id" validate:"required"`
+	WhitelistedSequencerSet []felt.Felt      `json:"whitelisted_sequencer_set"`
```
omerfirmak, line 10: I think it is fine to get rid of this, MVP will include a single centralized sequencer.

blockbuilder/genesisState.go
rianhughes: Closing this because I don't think we need marshalling or validation tests just yet

blockbuilder/genesisState.go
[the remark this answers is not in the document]
rianhughes: Okay, so because the user will have to know the contract address anyway (to use `FunctionCalls` - assuming they want to use them) we might as well make them supply it upfront. In that case I'll leave this as 
`Contracts     map[string]GenesisContractData `

```
type GenesisContractData struct {
	ClassHash       felt.Felt   `json:"class_hash"`
	ConstructorArgs []felt.Felt `json:"constructor_args"`
}
```

PR #1534, opened 2023-12-08
Add barebones builder with mempool validation facilities

omerfirmak:


builder/builder.go
```diff
@@ -0,0 +1,68 @@
...
+type Builder struct {
+	ownAddress felt.Felt
+
+	bc *blockchain.Blockchain
+	vm vm.VM
+
+	log utils.Logger
+}
+
+func New(ownAddr *felt.Felt, bc *blockchain.Blockchain, builderVM vm.VM, log utils.Logger) *Builder {
```
joshklop, line 24: I guess we'll probably want a private key as well (maybe in lieu of the address?). But it's too early to worry about that.

PR #1537, opened 2023-12-11
 Add and implement the StateReadWriter interface

joshklop:


core/state_snapshot.go
```diff
@@ -87,3 +87,7 @@ func (s *stateSnapshot) Class(classHash *felt.Felt) (*DeclaredClass, error) {
 	}
 	return declaredClass, nil
 }
+
+func (s *stateSnapshot) CompiledClassHash(classHash *felt.Felt) (*felt.Felt, error) {
+	return s.state.CompiledClassHash(classHash)
```
joshklop, line 92: This will not work, correct? We probably need a

```go
ClassIsDeclaredAt(classHash *felt.Felt, blockNumber uint64) (bool, error)
```
in the `StateHistoryReader` interface. That may lead to a pretty costly migration, if I'm thinking of this correctly.

core/state.go
```diff
@@ -40,6 +41,7 @@ type StateReader interface {
 	ContractNonce(addr *felt.Felt) (*felt.Felt, error)
 	ContractStorage(addr, key *felt.Felt) (*felt.Felt, error)
 	Class(classHash *felt.Felt) (*DeclaredClass, error)
+	CompiledClassHash(classHash *felt.Felt) (*felt.Felt, error)
```
omerfirmak, line 44: I don't think we need this ATM. `blockifier` never makes this call.

https://github.com/NethermindEth/juno/blob/66d5fa523321f4e60a10d4bdbe2668c77cdc6758/vm/rust/src/juno_state_reader.rs#L159-L165

core/state_snapshot.go
omerfirmak: Yeah, on its own, this wont suffice. But I don't think we need a migration to get it working.
Doing something similar to what `(s *stateSnapshot) Class` is doing should be enough

But we [can remove](https://github.com/NethermindEth/juno/pull/1537#discussion_r1422029257) `CompiledClassHash` all together.
 
``` golang
	declaredClass, err := s.state.Class(classHash)
	if err != nil {
		return nil, err
	}

	if s.blockNumber < declaredClass.At {
		return nil, db.ErrKeyNotFound
	}
```

omerfirmak:
https://github.com/NethermindEth/juno/compare/5ea7b8aa525f66b2960857979bca815826e32bdb..dd4c3be727cd1b6d6509cd7b54adcca8fa974c97#diff-c087fa7dffa5d472e6b7b59eb8f8bafd4208832487a1429e0fcfc38a9fe05167L57

We still need this to be able to build the state diff

omerfirmak (approved):

PR #1538, opened 2023-12-11
Revert adding useless felt.Clone receiver

joshklop:


omerfirmak:
Looks like `Clone()` was pretty handy when writing https://github.com/NethermindEth/juno/pull/1537?

joshklop:
> Looks like Clone() was pretty handy when writing #1537?

True, but we can use `new(felt.Felt).Set(x)`, which maintains the convention that the caller should explicitly allocate. I forgot this existed when I implemented the `Clone` approach.

omerfirmak:
> > Looks like Clone() was pretty handy when writing #1537?
> 
> True, but we can use `new(felt.Felt).Set(x)`, which maintains the convention that the caller should explicitly allocate. I forgot this existed when I implemented the `Clone` approach.

> which maintains the convention that the caller should explicitly allocate

I think Clone() still makes sense, because it they want a value copy they can just do `newFelt := *oldFeltPtr`, `newFelt := oldFelt`. Since felt.Felt is a simple type which doesnt store any references in it. 

To me, an explicit call to `Clone()` conveys the intent of actually doing a heap allocation.  

I would prefer `Clone()` over `new(felt.Felt).Set(x)` since it looks nicer and avoids repetition imo. Approving regardless, feel free to merge or close.

omerfirmak (approved):

joshklop:
> I would prefer Clone() over new(felt.Felt).Set(x) since it looks nicer and avoids repetition imo.

I don't have much of a preference on the aesthetics, so I'll defer here.

PR #1541, opened 2023-12-11
Enable the `JunoStateReader` to write

joshklop:
Depends on #1537


joshklop:
The top-level `transaction.execute` functions from blockifier only accepts a `CachedState` struct, not just any implementation of the `State` trait.

We could:

1. work around the issue by copying and patching blockifier code until we can pass our own `State` implementation to Blockifier.
2. ask Starkware to change the function signature.
3. merge state diffs after we receive them from the Rust side (starting implementation was [here](https://github.com/NethermindEth/juno/pull/1526/files#diff-4b18271a113e10a7dec3c372cb301cb4ace886effe3c419e5b695857da9a38c4R116-R148), and an adapter is on the old `joshklop/sequencer` branch).

I'm leaning toward option 3. What are others' thoughts?

rianhughes:
> The top-level `transaction.execute` functions from blockifier only accept a `CachedState` struct, not just any implementation of the `State` trait.
> 
> We could:
> 
>     1. work around the issue by unraveling blockifier code until we can pass our own `State` implementation to Blockifier.
> 
>     2. ask Starkware to change the `execute` function signatures.
> 
>     3. merge state diffs after we receive them from the Rust side (starting implementation was [here](https://github.com/NethermindEth/juno/pull/1526/files#diff-4b18271a113e10a7dec3c372cb301cb4ace886effe3c419e5b695857da9a38c4R116-R148), and an adapter is on the old `joshklop/sequencer` branch).
> 
> 
> I'm leaning toward option 3. What are others' thoughts?

Could we have a CachedState representation in Juno, pass that into blockifier, and adapt when we need to?

Otherwise, option 3 also seems like the best path forward.

vm/rust/src/juno_state_reader.rs
```diff
@@ -164,6 +187,83 @@ impl StateReader for JunoStateReader {
...
+
+    /// Sets the given compiled class hash under the given class hash.
+    fn set_compiled_class_hash(
+        &mut self,
+        class_hash: ClassHash,
+        compiled_class_hash: CompiledClassHash,
+        ) -> StateResult<()> {
+        let class_hash_bytes = felt_to_byte_array(&class_hash.0);
+        let compiled_class_hash_bytes = felt_to_byte_array(&compiled_class_hash.0);
+        state_read_err(unsafe { JunoStateSetClassHash(self.handle, class_hash_bytes.as_ptr(), compiled_class_hash_bytes.as_ptr()) })
```
omerfirmak, line 244: I believe `JunoStateSetClassHash` is not the write call to make here

omerfirmak:
Any chance we can sanity check/unittest this somehow?

vm/rust/src/juno_state.rs
```diff
@@ -164,6 +192,83 @@ impl StateReader for JunoStateReader {
...
+
+    fn to_state_diff(&mut self) -> CommitmentStateDiff {
+        // return an empty state diff for now. Not needed when we are directly storing all writes anyway.
+        CommitmentStateDiff {
+            address_to_class_hash: IndexMap::default(),
+            storage_updates: IndexMap::default(),
+            class_hash_to_compiled_class_hash: IndexMap::default(),
+            address_to_nonce: IndexMap::default(),
+        }
+    }
```
omerfirmak, line 260: Instead of silently returning incomplete data, let's use `unimplemented` here. I don't think it will be called at all.

joshklop: It's called for traces. Do we want to break those handlers in the sequencer?

omerfirmak: We call it on the `txn_state`, not the JunoState. Traces should still work.

omerfirmak:
> Any chance we can sanity check/unittest this somehow?

Ignore me, genesis state tests should be enough to sanity check these changes.

omerfirmak (approved):
Nice :100:

Pushed a small clean up commit. Feel free to squash merge at will.

Issue #1415, opened 2023-11-08
Make chainId configurable


stranger80:
Allow configuration of custom chainId values in Juno
This means ability to configure a set of "custom" starknet endpoints and starknet specific contract addresses.
May be useful to adapt and use code from https://github.com/nethermindEth/juno/tree/omerfirmak/custom-networks

joshklop:
Related to #1328

PR #1542, opened 2023-12-12
Rianhughes/custom network atmpt3

rianhughes:
closes https://github.com/NethermindEth/juno/issues/1415


blockchain/blockchain.go
```diff
@@ -68,7 +68,7 @@ var _ Reader = (*Blockchain)(nil)
 
 // Blockchain is responsible for keeping track of all things related to the Starknet blockchain
 type Blockchain struct {
-	network  utils.Network
+	network  utils.NetworkKnown
```
omerfirmak, line 71: We should keep using `utils.Network` everywhere, `utils.KnownNetwork` and  `utils.CustomNetwork` shouldn't leak beyond `utils` and `node` package context.

node/node.go
```diff
@@ -118,28 +120,39 @@ func New(cfg *Config, version string) (*Node, error) { //nolint:gocyclo,funlen
...
+	var network utils.Network
+	if cfg.Network == utils.Custom {
+		err = cfg.NetworkCustom.Validate()
+		if err != nil {
+			return nil, err
+		}
+		network = cfg.NetworkCustom
+	} else {
+		network = cfg.Network
+	}
```
omerfirmak, line 132: can we move this logic to `config` type

rianhughes: Pushed this to config.GetNetwork()

validator/validator.go
```diff
@@ -30,11 +29,40 @@ func Validator() *validator.Validate {
...
 				return t.String()
 			}
 			panic("not a TransactionType")
-		}, rpc.TransactionType(0))
+		}, TransactionType(0))
 	})
 	return v
 }
+
+// TransactionType is duplicated here to prevent import cycles
```
omerfirmak, line 41: I think we should be able to avoid doing this. Can you share the import cycle that is hapenning?

rianhughes: Sure, it just arises if you use rpc.TransactionType in validator.go. For example, validator imports rpc, which imports utils (eg class.go), which imports validator (network.go) which imports validator

utils/network.go
```diff
@@ -177,10 +203,101 @@ func (n Network) CoreContractAddress() (common.Address, error) {
...
+	return nil
+}
+
+func (cn NetworkCustom) Validate() error {
+	validate := validator.Validator()
+	if cn.CoreContractAddressVal.String() == "0x0000000000000000000000000000000000000000" {
+		return ErrInvalidCoreContractAddress
+	}
+	return validate.Struct(cn)
+}
```
omerfirmak, line 248: Let's remove this and move the logic to callsite (ie config.GetNetwork())

validator/validator.go
rianhughes: Later commit resolves this. Moving (cn NetworkCustom) Validate() to config.GetNetwork() stops the import cycle.

utils/network.go
```diff
@@ -177,10 +200,93 @@ func (n Network) CoreContractAddress() (common.Address, error) {
...
+		return common.Address{}, errors.New("core contract address is nil")
+	}
+	return *cn.CoreContractAddressVal, nil
+}
+
+func (cn NetworkCustom) ChainID() *felt.Felt {
+	return new(felt.Felt).SetBytes([]byte(cn.ChainIDVal))
+}
+
+func (cn NetworkCustom) ProtocolID() protocol.ID {
```
omerfirmak, line 290: arent these supposed to be pointer receivers, or am I missing something?

rianhughes: For some reason (not sure why), these methods need to be value receivers? Otherwise they don't implement the Network interface

rianhughes: We are calling these methods as if they are value receivers (eg as map values, see [line 60 in network_test.go](https://github.com/NethermindEth/juno/blob/main/utils/network_test.go#L60)). 

The codebase could be updated so that we can redefine these methods as pointer receivers, but I expect it will have a non-trivial codebase. I'm happy to leave it as is (a bit biased tho because this is more work for me :) ), or just pushing it to a separate issue.

cmd/juno/juno.go
```diff
@@ -221,6 +255,12 @@ func NewCmd(config *node.Config, run func(*cobra.Command, []string) error) *cobr
 	junoCmd.Flags().Uint16(wsPortF, defaultWSPort, wsPortUsage)
 	junoCmd.Flags().String(dbPathF, defaultDBPath, dbPathUsage)
 	junoCmd.Flags().Var(&defaultNetwork, networkF, networkUsage)
+	junoCmd.Flags().String(feederURLF, defaultFeederURL, networkCustomFeederUsage)
+	junoCmd.Flags().String(gatewayURLF, defaultGatewayURL, networkCustomGatewayUsage)
+	junoCmd.Flags().String(chainIDF, defaultChainID, networkCustomL2ChainIDUsage)
+	junoCmd.Flags().Int(l1ChainIDF, defaultL1ChainIDInt, networkCustomL1ChainIDUsage)
+	junoCmd.Flags().Int(protocolIDF, defaultProtocolID, networkCustomProtocolIDUsage)
+	junoCmd.Flags().String(coreContractAddressF, defaultCoreContractAddressStr, networkCustomCoreContractAddressUsage)
```
omerfirmak, line 263: do you think we can prefix these with some string to indicate that they are related?

something like;

`--cn.feeder-url`
`--cn.gateway-url`

omerfirmak: https://github.com/NethermindEth/juno/pull/1542/files#diff-e8011febe1f0287e6ed8aba108ce6e448c8db2c49f5afef912db1626eafd0ee3R70-R75

I don't see this change

omerfirmak:
@joshklop has more experience with our CLI package, I will let him have the last say over this.

joshklop:
I am concerned that distinguishing between "known" and "custom" networks everywhere creates unnecessary complexity.

We can have a single `Network` struct with fields for each property of a network.

```go
type Network struct {
        GatewayURL string `json:"gateway_url"`
        FeederURL string `json:"feeder_url"`
        // etc.
}
```

We can have pre-defined variables for known networks.

```go
var Mainnet = Network{
      GatewayURL: "https://alpha-mainnet.starknet.io/gateway/",
      // etc.
}
```

For the command-line interface, we provide two ways to specify a network:

1. the `--network <network-name>` flag for known networks. We can define a `Network` enum in the `cmd/juno` package that fulfills the `TextUnmarshaler` interface so users can use, e.g, `--network mainnet` as usual. This will basically entail moving the existing `utils.Network` enum from `utils` to `cmd/juno` while removing almost all of its methods.
2. a `--custom-network` flag that accepts a json string that we unmarshal into the `Network` struct.

If both flags are specified, we log an error and quit.

Thoughts?

rianhughes:
> I am concerned that distinguishing between "known" and "custom" networks everywhere creates unnecessary complexity.
> 
> We can have a single `Network` struct with fields for each property of a network.
> 
> ```go
> type Network struct {
>         GatewayURL string `json:"gateway_url"`
>         FeederURL string `json:"feeder_url"`
>         // etc.
> }
> ```
> 
> We can have pre-defined variables for known networks.
> 
> ```go
> var Mainnet = Network{
>       GatewayURL: "https://alpha-mainnet.starknet.io/gateway/",
>       // etc.
> }
> ```

Originally we went with this approach, but the Network struct became too heavy (above the 88-byte limit set by the linter). This required passing the network struct as a pointer rather than by value throughout the codebase, which made for a really messy PR (see [here](https://github.com/NethermindEth/juno/tree/rianhughes/custom-network) for the original PR that became messy, and [here](https://github.com/NethermindEth/juno/tree/rianhughes/custom-network-atmpt-2) for the cleaner PR where the Network struct etc is updated, but not the rest of the codebase changes where Network is passed as a pointer instead of by value).

After internal discussions with @omerfirmak, we decided to redefine/rename utils.Network as an interface, that CustomNetwork implements. This means that we can leave the rest of the codebase untouched, and just pass CustomNetwork around as a pointer.

rianhughes:
> For the command-line interface, we provide two ways to specify a network:
> 
>     the --network <network-name> flag for known networks. We can define a Network enum in the cmd/juno package that fulfills the TextUnmarshaler interface so users can use, e.g, --network mainnet as usual. This will basically entail moving the existing utils.Network enum from utils to cmd/juno while removing almost all of its methods. If the user provides this option, we switch on the enum and add the corresponding pre-defined network to the node.Confg.
>     a --custom-network <json> flag that accepts a json string that we unmarshal into the Network struct.
> 
> If both flags are specified, we log an error and quit.

Currently, users can specify default networks via "--network mainnet" etc, so I'm not sure what we gain by moving Network for utils to cmd/juno. As for how the data is passed into --custom-network, I personally prefer specifying these parameters separately, but I'm happy to go with the majority opinion on this. 

@joshklop @omerfirmak

cmd/juno/juno.go
rianhughes: Whoops! Forgot to prefix the actual flags. Should be present now

PR #1569, opened 2023-12-15
Allow user of pending state to obtain underlying state diff

joshklop:
so the builder can commit the state diff to the db


joshklop:
Opted not to clone the state diff since we don't write to it after using it in the pending state.

Added a comment to document that writing the state diff after obtaining it through `PendingState.StateDiff()` is not safe.

omerfirmak:
Merged as part of #1593

PR #1570, opened 2023-12-15
Build receipts

joshklop:


builder/builder.go
```diff
@@ -8,6 +8,7 @@ import (
 	"github.com/NethermindEth/juno/core/felt"
 	"github.com/NethermindEth/juno/db"
 	"github.com/NethermindEth/juno/mempool"
+	"github.com/NethermindEth/juno/rpc"
```
omerfirmak, line 11: I am not sure if this is the most clean way of doing this. `builder` depending on `rpc` is code smell IMO.

Initially, to avoid declaring vm.Trace and rpc.Trace, we decided to return raw JSON from `vm.Execute()`. But we ended up having to declare `rpc.Trace` already. So I believe we need to bite the bullet and duplicate trace definition in `vm` package (or move rpc.Trace to vm) and return a properly typed `Trace` from `vm.Execute()`.

joshklop:
Depends on #1578

builder/builder.go
```diff
@@ -24,10 +25,49 @@ type Builder struct {
...
+func (b *Builder) Receipt(fee *felt.Felt, feeUnit core.FeeUnit, txHash *felt.Felt, reverted bool,
+	trace *vm.TransactionTrace,
+) (*core.TransactionReceipt, error) {
+	executionResources := make([]*vm.ExecutionResources, 0)
+	events := make([]*core.Event, 0)
+	messages := make([]*core.L2ToL1Message, 0)
+	if trace.ConstructorInvocation != nil {
+		executionResources = append(executionResources, trace.ConstructorInvocation.ExecutionResources)
+		events = append(events, utils.Map(trace.ConstructorInvocation.Events, vm2core.AdaptOrderedEvent)...)
+		messages = append(messages, utils.Map(trace.ConstructorInvocation.Messages, vm2core.AdaptMsgToL1)...)
```
omerfirmak, line 42: Don't we need to recursively traverse the call stack and accumulate all the events and messages?

builder/builder.go
```diff
@@ -24,10 +25,49 @@ type Builder struct {
...
+		revertReason = trace.ExecuteInvocation.RevertReason
+	}
+	if trace.FeeTransferInvocation != nil {
+		executionResources = append(executionResources, trace.FeeTransferInvocation.ExecutionResources)
+		events = append(events, utils.Map(trace.FeeTransferInvocation.Events, vm2core.AdaptOrderedEvent)...)
+		messages = append(messages, utils.Map(trace.FeeTransferInvocation.Messages, vm2core.AdaptMsgToL1)...)
 	}
+	return &core.TransactionReceipt{
+		Fee:                fee,
+		FeeUnit:            core.STRK,
```
omerfirmak, line 63:
```suggestion
		FeeUnit:            feeUnit,
```

builder/builder.go
omerfirmak: and we should maintain the order of the events and messages

joshklop:
We can drop the last commit and merge directly to `main`. Thoughts?

builder/builder.go
```diff
@@ -24,12 +25,26 @@ type Builder struct {
...
 		ownAddress: *ownAddr,
-
-		bc: bc,
-		vm: builderVM,
+		bc:         bc,
+		vm:         builderVM,
 	}
 }
 
+func (b *Builder) Receipt(fee *felt.Felt, feeUnit core.FeeUnit, txHash *felt.Felt, reverted bool,
```
omerfirmak, line 33: It doesnt look like this needs to be a member function. Let's make it standalone?

omerfirmak (approved):

omerfirmak:
> We can drop the last commit and merge directly to main. Thoughts?

That works too

PR #1572, opened 2023-12-18
Log failed requests only

omerfirmak:


jsonrpc/server.go
```diff
@@ -438,7 +431,10 @@ func (s *Server) handleRequest(ctx context.Context, req *Request) (*response, er
 	if errAny := tuple[1].Interface(); !isNil(errAny) {
 		res.Error = errAny.(*Error)
 		if res.Error.Code == InternalError {
-			s.listener.OnRequestFailed(req.Method, err)
+			s.listener.OnRequestFailed(req.Method, res.Error)
+			reqJSON, _ := json.Marshal(req)
+			errJSON, _ := json.Marshal(res.Error)
+			s.log.Debugw("failed handing RPC request", "req", string(reqJSON), "res", string(errJSON))
```
joshklop, line 437:
```suggestion
			s.log.Debugw("Failed handing RPC request", "req", string(reqJSON), "res", string(errJSON))
```

joshklop, line 437: If `reqJSON` or `errJSON` are `nil`, we can't convert them using `string(nil)`. I think we should set them to `[]byte{}` if `err != nil`.

omerfirmak: Actually, `string([]byte(nil))` just returns an empty string. So we should be good.

jsonrpc/server.go
```diff
@@ -438,7 +431,10 @@ func (s *Server) handleRequest(ctx context.Context, req *Request) (*response, er
 	if errAny := tuple[1].Interface(); !isNil(errAny) {
 		res.Error = errAny.(*Error)
 		if res.Error.Code == InternalError {
-			s.listener.OnRequestFailed(req.Method, err)
+			s.listener.OnRequestFailed(req.Method, res.Error)
+			reqJSON, _ := json.Marshal(req)
+			errJSON, _ := json.Marshal(res.Error)
+			s.log.Debugw("Failed handing RPC request", "req", string(reqJSON), "res", string(errJSON))
```
joshklop, line 437: Might be good to promote this to a warning.

joshklop (approved):

jsonrpc/server.go
wojciechos: I would rather stay with debug. If we label it as a warning, we might start overlooking real warnings, especially if these pop up a lot.

PR #1577, opened 2023-12-18
Fix dependabot config again

joshklop:
Third time's the charm.

I'll try to wait until every CI check is finished on the PR before merging. Maybe the dependabot action only runs at the end, which is why we only see it fail when we merge.


omerfirmak (approved):

joshklop:
Alas, the dependabot check did not run. We'll see if it works upon merging.

joshklop:
It worked :raised_hands:

PR #1578, opened 2023-12-18
Copy rpc.TransactionTrace to vm package

joshklop:
Needed for sequencer, might as well do it now


vm/vm.go
```diff
@@ -165,7 +166,7 @@ func (v *vm) Call(contractAddr, classHash, selector *felt.Felt, calldata []felt.
 func (v *vm) Execute(txns []core.Transaction, declaredClasses []core.Class, blockNumber, blockTimestamp uint64,
 	sequencerAddress *felt.Felt, state core.StateReader, network utils.Network, paidFeesOnL1 []*felt.Felt,
 	skipChargeFee, skipValidate, errOnRevert bool, gasPriceWEI *felt.Felt, gasPriceSTRK *felt.Felt, legacyTraceJSON bool,
-) ([]*felt.Felt, []json.RawMessage, error) {
+) ([]*felt.Felt, []*TransactionTrace, error) {
```
omerfirmak, line 169:
```suggestion
) ([]*felt.Felt, []TransactionTrace, error) {
```

rpc/handlers.go
```diff
@@ -1368,23 +1368,23 @@ func (h *Handler) LegacyEstimateMessageFee(msg MsgFromL1, id BlockID) (*FeeEstim
 //
 // It follows the specification defined here:
 // https://github.com/starkware-libs/starknet-specs/blob/1ae810e0137cc5d175ace4554892a4f43052be56/api/starknet_trace_api_openrpc.json#L11
-func (h *Handler) TraceTransaction(ctx context.Context, hash felt.Felt) (json.RawMessage, *jsonrpc.Error) {
+func (h *Handler) TraceTransaction(ctx context.Context, hash felt.Felt) (*TransactionTrace, *jsonrpc.Error) {
```
omerfirmak, line 1371: How would you feel about returning a `vm.Trace` here? So instead of duplicating types, we would be moving them to vm.

joshklop: I'd be fine with that.

rianhughes:
By moving `FunctionInvocation` and `TransactionTrace` from `rpc` to `vm`, will RPC-spec changes not influence the `vm`? If so this seems a bit messy to me. Having adapters / duplicated types seems a bit cleaner.

omerfirmak:
> will RPC-spec changes not influence the `vm`?

It is how it has already been so far anyways. This PR allows us to access the trace on Go side easier, instead of dealing with raw json. So it is an improvement imo.

vm/vm.go
```diff
@@ -246,7 +247,14 @@ func (v *vm) Execute(txns []core.Transaction, declaredClasses []core.Class, bloc
 		return nil, nil, errors.New(context.err)
 	}
 
-	return context.actualFees, context.traces, nil
+	traces := make([]TransactionTrace, len(context.traces))
+	for index, traceJSON := range context.traces {
+		if err := json.Unmarshal(traceJSON, &traces[index]); err != nil {
+			return nil, nil, fmt.Errorf("unmarshal trace: %v", err)
```
omerfirmak, line 253: I failed to point this out before merging, but this error should be mapped to `ErrInternal` on the RPC side.

PR #1568, opened 2023-12-15
Rianhughes/build genesis state

rianhughes:
closes https://github.com/NethermindEth/juno/issues/1567

PR #1593, opened 2023-12-21
Rianhughes/build genesis state fresh

rianhughes:
closes https://github.com/NethermindEth/juno/issues/1567
taken from https://github.com/NethermindEth/juno/pull/1568


genesis/genesis.go
```diff
@@ -0,0 +1,146 @@
...
+
+type GenesisConfig struct {
+	ChainID string `json:"chain_id" validate:"required"`
+
+	Classes []string `json:"classes"` // []path-to-class.json
+
+	Contracts map[string]GenesisContractData `json:"contracts"` // address -> {classHash, constructorArgs}
+
+	FunctionCalls []FunctionCall `json:"function_calls"` // list of functionCalls to Call()
+}
```
omerfirmak, line 27:
```suggestion
type GenesisConfig struct {
	ChainID string `json:"chain_id" validate:"required"`
	Classes []string `json:"classes"` // []path-to-class.json
	Contracts map[string]GenesisContractData `json:"contracts"` // address -> {classHash, constructorArgs}
	FunctionCalls []FunctionCall `json:"function_calls"` // list of functionCalls to Call()
}
```

genesis/genesis.go
```diff
@@ -0,0 +1,146 @@
...
+	log utils.Logger,
+) (*core.StateDiff, map[felt.Felt]core.Class, error) {
+	blockTimestamp := uint64(time.Now().Unix())
+
+	newClasses, err := loadClasses(genesisConfig.Classes)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	pendingReader, closer, err := chain.PendingState()
```
omerfirmak, line 59: We can get rid of the `chain` dependency here. See how we create an empty state in `blockchain`

```
		txn := db.NewMemTransaction()
		emptyState := core.NewState(txn)
		return emptyState, txn.Discard, nil
```

PR #1596, opened 2023-12-21
L1->L2 messaging

### LINKED ISSUE #1525: Sequencer - Message handling
- [ ] L1 Message handling

- For MVP, doesnt need to be a full-blown implementation which listens to a core contract. We just need a way to push messages to be handled by the sequencer. Could be a RPC method for now.

joshklop:
Closes #1525

### DIFF
--- a/rpc/handlers.go
+++ b/rpc/handlers.go
@@ -1793,6 +1793,29 @@ func (h *Handler) unsubscribe(sub *subscription, id uint64) {
 	h.mu.Unlock()
 }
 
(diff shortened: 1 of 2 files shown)


joshklop:
Just realized there are some ways this can be simplified. Will re-open when it's ready :sweat_smile:

rpc/handlers.go
```diff
@@ -1793,6 +1793,27 @@ func (h *Handler) unsubscribe(sub *subscription, id uint64) {
 	h.mu.Unlock()
 }
 
+func (h *Handler) AddMsgFromL1(msg core.L1ToL2Message) *jsonrpc.Error {
```
omerfirmak, line 1796: Is it possible to use `rpc.MsgFromL1`?

omerfirmak (approved):
Let's register the handler as well

PR #1598, opened 2023-12-21
Treat internal vm errors as internal rpc errors

joshklop:
See https://github.com/NethermindEth/juno/pull/1578#discussion_r1432860435


omerfirmak:
pushed a simple refactor, let me know if it looks good to you.

rpc/handlers.go
```diff
@@ -1633,12 +1626,9 @@ func (h *Handler) traceBlockTransactions(ctx context.Context, block *core.Block,
...
 		sequencerAddress, state, h.network, paidFeesOnL1, false, false, false, block.Header.GasPrice, block.Header.GasPriceSTRK, legacyJSON)
 	if err != nil {
-		if isVMInternalErr(err) {
-			return nil, ErrInternal.CloneWithData(err.Error())
-		}
 		// Since we are tracing an existing block, we know that there should be no errors during execution. If we encounter any,
-		// report them as unexpected errors
-		return nil, ErrUnexpectedError.CloneWithData(err.Error())
+		// report them as internal errors
+		return nil, ErrInternal.CloneWithData(err.Error())
```
joshklop, line 1631: Are we reverting the decision made [before](https://github.com/NethermindEth/juno/pull/1475#discussion_r1407919925)? If so, we can remove `vm.ErrInternal` entirely.

omerfirmak: I think we can, since we only count internal errors as failed requests now.

omerfirmak: > we can remove vm.ErrInternal entirely.

Yeap, let's do that as well

PR #1599, opened 2023-12-22
rust shorthand

Lawliet-Chan:
We can use rust shorthand in struct.
https://rust-lang.github.io/rfcs/1682-field-init-shorthand.html


omerfirmak:
For consistency I believe it makes more sense to keep it the way it currently is. None of the other fields are using the shorthand notation.

PR #1600, opened 2023-12-22
Use placeholder nonce when serving L1 handlers over RPC

joshklop:
Old L1 handlers don't have nonces, but the spec requires the nonce. We use 0x0 as the placeholder.

A regression test is included.


rpc/handlers.go
```diff
@@ -352,12 +352,15 @@ func AdaptTransaction(t core.Transaction) *Transaction {
 	case *core.DeployAccountTransaction:
 		txn = adaptDeployAccountTrandaction(v)
 	case *core.L1HandlerTransaction:
-		// https://github.com/starkware-libs/starknet-specs/blob/a789ccc3432c57777beceaa53a34a7ae2f25fda0/api/starknet_api_openrpc.json#L1669
+		nonce := v.Nonce
+		if nonce == nil {
+			nonce = new(felt.Felt)
```
omerfirmak, line 357:
```suggestion
			nonce = &felt.Zero```

omerfirmak (approved):

Issue #694, opened 2023-03-29
Remove `nolint: gocyclo` on trie.Put


joshklop:
Will require refactoring the `trie.Put` method to fix the `gocyclo` issue.

aminsato:
Still no review 😕

IronGauntlets:
Closed by #1601

PR #1601, opened 2023-12-24
Resolve gocyclo Issue

aminsato:
see #694


aminsato:
Hi @IronGauntlets 

I resolved the related issue. 

It passed the corresponding test, and the result of gocyclo is 10.

please take a look

IronGauntlets (approved):

PR #1609, opened 2023-12-27
 Implement blockchain.Finalize 

omerfirmak:


sync/sync_test.go
```diff
@@ -115,7 +115,7 @@ func TestSyncBlocks(t *testing.T) {
 			case 2:
 				state.BlockHash = new(felt.Felt) // fail sanity checks
 			case 3:
-				state.OldRoot = new(felt.Felt).SetUint64(1) // fail store
+				state.NewRoot = new(felt.Felt).SetUint64(1) // fail store
```
joshklop, line 118: What is the reason for this change?

core/block.go
```diff
@@ -150,6 +150,14 @@ func VerifyBlockHash(b *Block, network utils.Network) (*BlockCommitments, error)
 	return nil, errors.New("can not verify hash in block header")
 }
 
+// BlockHash computes the block hash
+func BlockHash(b *Block, network utils.Network) (*felt.Felt, *BlockCommitments, error) {
+	if b.Number < NetworkBlockHashMetaInfo(network).First07Block {
+		return pre07Hash(b, network.ChainID())
+	}
+	return post07Hash(b, nil)
```
joshklop, line 158: We can export `blockHash` or do:
```go
return blockHash(b, network, nil)
```

blockchain/blockchain.go
```diff
@@ -336,47 +336,50 @@ func (b *Blockchain) Store(block *core.Block, blockCommitments *core.BlockCommit
...
 		}
 
-		state := core.NewState(txn)
-		if err := state.Update(block.Number, stateUpdate.StateDiff, newClasses); err != nil {
+		if err := core.NewState(txn).Update(block.Number, stateUpdate.StateDiff, newClasses); err != nil {
 			return err
 		}
 
-		if err := b.verifyStateUpdateRoot(state, stateUpdate.NewRoot); err != nil {
+		if err := b.verifyStateUpdateRoot(core.NewState(txn), stateUpdate.NewRoot); err != nil {
```
joshklop, line 347: Is there a reason to use two different state instances?

blockchain/blockchain_test.go
```diff
@@ -797,3 +797,47 @@ func TestMakeStateDiffForEmptyBlock(t *testing.T) {
...
+	require.NoError(t, err)
+	require.Equal(t, uint64(0), h)
+
+	pending, err := chain.Pending()
+	require.NoError(t, err)
+	require.NoError(t, chain.Finalize(&pending))
+	h, err = chain.Height()
+	require.NoError(t, err)
+	require.Equal(t, uint64(1), h)
+}
```
joshklop, line 843: Can we test that the header, txs, and receipts have been stored correctly as well?

core/block.go
omerfirmak: Nice, thanks!

blockchain/blockchain.go
omerfirmak: No real reason. Since `core.State` is just a wrapper around a `db.Transaction` it is safe to assume all changes are persisted in the `txn`. But will change it to use a single state instance to make easier to follow.

blockchain/blockchain_test.go
omerfirmak: Done, that actually uncovered a bug when unmarshaling empty blocks from storage:D

sync/sync_test.go
omerfirmak: I took the liberty to remove OldRoot check, since it is kinda redundant. `verifyBlock` already checks the parent block hash against the local `HEAD`, which already proves that our state root is `OldRoot`

joshklop (approved):

PR #1610, opened 2023-12-27
Expose Sierra2CASM compiler to Go side

omerfirmak:


starknet/compiler.go
```diff
@@ -0,0 +1,48 @@
...
+	casmJSONOrErrorCstr := C.compileSierraToCasm(sierraJSONCstr)
+	casmJSONOrError := C.GoString(casmJSONOrErrorCstr)
+	C.freeCstr(casmJSONOrErrorCstr)
+
+	var casmClass CompiledClass
+	if !json.Valid([]byte(casmJSONOrError)) {
+		return nil, errors.New(casmJSONOrError)
+	}
+
+	if err = json.Unmarshal([]byte(casmJSONOrError), &casmClass); err != nil {
```
joshklop, line 43: Rather than parsing the json twice, can we unmarshal once? If the unmarshal error is [`json.SyntaxError`](https://pkg.go.dev/encoding/json#SyntaxError), we return `errors.New(casmJSONOrError)`.

joshklop (approved):

starknet/compiler.go
omerfirmak: Makes sense, done.

PR #1612, opened 2023-12-28
Start finalising empty blocks in sequencing mode

omerfirmak:
Best reviewed commit by commit

### DIFF
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -50,7 +50,7 @@ type Reader interface {
 
 var (
 	ErrParentDoesNotMatchHead = errors.New("block's parent hash does not match head block hash")
-	supportedStarknetVersion  = semver.MustParse("0.13.0")
+	SupportedStarknetVersion  = semver.MustParse("0.13.0")
 )
 
 func checkBlockVersion(protocolVersion string) error {
@@ -59,7 +59,7 @@ func checkBlockVersion(protocolVersion string) error {
 		return err
 	}
 
-	if blockVer.GreaterThan(supportedStarknetVersion) {
+	if blockVer.GreaterThan(SupportedStarknetVersion) {
 		return errors.New("unsupported block version")
 	}
 
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -1,6 +1,7 @@
 package builder
 
 import (
+	"context"
 	"errors"
 
 	"github.com/NethermindEth/juno/blockchain"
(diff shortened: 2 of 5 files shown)


node/node.go
```diff
@@ -240,19 +229,36 @@ func New(cfg *Config, version string) (*Node, error) { //nolint:gocyclo,funlen
...
+
+	l1Client, err := l1.NewClient(ethSubscriber, chain, log), nil
+	if err != nil {
+		return nil, fmt.Errorf("set up l1 client: %w", err)
+	}
+
+	if cfg.Metrics {
+		l1Client.WithEventListener(makeL1Metrics())
+	}
+	return l1Client, err
```
joshklop, line 275: I think we can `return l1Client, nil`

rpc/handlers.go
```diff
@@ -1271,6 +1280,10 @@ func (h *Handler) TransactionStatus(ctx context.Context, hash felt.Felt) (*Trans
 			Execution: receipt.ExecutionStatus,
 		}, nil
 	case ErrTxnHashNotFound:
+		if h.feederClient == nil {
+			return nil, ErrInternal.CloneWithData("no feeder client configured")
```
joshklop, line 1284: Maybe we should just return `ErrTxnHashNotFound`?

builder/builder.go
```diff
@@ -26,24 +27,94 @@ type Builder struct {
...
+		}
+	}
+}
+
+// InitPendingBlock initialises a new pending block
+func (b *Builder) InitPendingBlock() error {
+	nextHeight := uint64(0)
+	parentHash := felt.Zero
+	parentState := felt.Zero
+	if head, err := b.bc.HeadsHeader(); err == nil {
```
joshklop, line 69: Other than the first time `InitPendingBlock` is called, `head == b.PendingBlock`. What do you think about making this function stateless by passing in the previous block as a param (`nil` if it doesn't exist)? That way we don't need to read from the db unnecessarily.

builder/builder.go
```diff
@@ -26,24 +27,94 @@ type Builder struct {
...
+		case <-time.After(blockInterval):
+			if err := b.Finalise(); err != nil {
+				return err
+			}
+		}
+	}
+}
+
+// InitPendingBlock initialises a new pending block
+func (b *Builder) InitPendingBlock() error {
```
joshklop, line 65: I think we can un-export this.

blockchain/blockchain.go
```diff
@@ -50,7 +50,7 @@ type Reader interface {
 
 var (
 	ErrParentDoesNotMatchHead = errors.New("block's parent hash does not match head block hash")
-	supportedStarknetVersion  = semver.MustParse("0.13.0")
+	SupportedStarknetVersion  = semver.MustParse("0.13.0")
```
joshklop, line 53: I feel like we should add a `SupportedVersion() semver.Version` to the `blockchain.Reader` interface, but can't come up with a good reason. I guess we can deal with the exported global variable when it actually starts getting in the way. Was that your thinking as well?

joshklop (commented):
:tada:

builder/builder.go
```diff
@@ -26,24 +27,94 @@ type Builder struct {
...
 	}
 }
 
 func (b *Builder) Run(ctx context.Context) error {
-	<-ctx.Done()
+	if err := b.InitPendingBlock(); err != nil {
+		return err
+	}
+
+	blockInterval := 5 * time.Second
```
joshklop, line 51: This can be a `const`

blockchain/blockchain.go
omerfirmak: I think a package level exported constant is fine. Since different instances won't have support for different versions, I don't think need to add `SupportedVersion() semver.Version`.

rpc/handlers.go
omerfirmak: Makes sense

builder/builder.go
omerfirmak: It kinda helps with testing. Allows us to not deal with services in tests.

omerfirmak:
Will extract first 2 commits in to a separate PR to be merged on `main`

builder/builder.go
joshklop: > But things get trickier when builder is restarted on a non-empty DB

I was thinking that we can call `b.bc.HeadsHeader` only on startup.

joshklop: ^Is it harder than calling `b.bc.HeadsHeader` on startup and then passing in `b.PendingBlock.Header` thereafter? If not, I think we might as well do it in this PR.

joshklop (approved):

builder/builder.go
omerfirmak: I think we can skip the excursively "empty block" version  of the builder and go straight for normal block building. Closing this PR for now.

PR #1614, opened 2023-12-28
Use http.Client.Timeout instead of a context with deadline

omerfirmak:


joshklop (approved):
If I'm understanding correctly, there are technically two changes here:

1. Replacing the `context.WithTimeout` with a proper `http.Client.Timeout` config so the `net/http` package doesn't prematurely cancel a request.
2. Adding a `context.Context` to the `AddTransaction` method.

Are these two changes related somehow?

omerfirmak:
> If I'm understanding correctly, there are technically two changes here:
> 
>     1. Replacing the `context.WithTimeout` with a proper `http.Client.Timeout` config so the `net/http` package doesn't prematurely cancel a request.
> 
>     2. Adding a `context.Context` to the `AddTransaction` method.
> 
> 
> Are these two changes related somehow?

Second one was triggered by the first change. But we can split them up to two commits.

omerfirmak:
ah #1616 already depends on this, I won't bother splitting the changes to avoid the headache.

PR #1615, opened 2023-12-28
Add gateway client metrics

joshklop:


joshklop:
I think #1616 may be a better way of doing this. Feel free to review whichever you think is better.

omerfirmak:
Let's go with this since getting timeouts right will be a bit trickier with #1616

omerfirmak (approved):

PR #1618, opened 2023-12-29
Add Cairo1 support in Genesis state

omerfirmak:


genesis/genesis.go
```diff
@@ -57,14 +56,15 @@ func GenesisStateDiff(
...
-			if err = genesisState.SetContractClass(&classHash, class); err != nil {
-				return nil, nil, fmt.Errorf("declare v0 class: %v", err)
+		// Sets pending.newClasses, DeclaredV0Classes, (not DeclaredV1Classes)
+		if err = genesisState.SetContractClass(&classHash, class); err != nil {
+			return nil, nil, fmt.Errorf("declare v0 class: %v", err)
+		}
+
+		if cairo1Class, isCairo1 := class.(*core.Cairo1Class); isCairo1 {
+			if err = genesisState.SetCompiledClassHash(&classHash, cairo1Class.Compiled.Hash()); err != nil {
+				return nil, nil, err
```
joshklop, line 66:
```suggestion
				return nil, nil, fmt.Errorf("set compiled class hash: %v", err)
```

genesis/genesis.go
```diff
@@ -113,13 +113,19 @@ func loadClasses(classes []string) (map[felt.Felt]core.Class, error) {
...
-		coreClass, err := sn2core.AdaptCairo0Class(response)
-		if err != nil {
+		var coreClass core.Class
+		if response.V0 != nil {
+			if coreClass, err = sn2core.AdaptCairo0Class(response.V0); err != nil {
+				return nil, err
+			}
+		} else if compiledClass, err := starknet.Compile(response.V1); err != nil {
+			return nil, err
+		} else if coreClass, err = sn2core.AdaptCairo1Class(response.V1, compiledClass); err != nil {
```
joshklop, line 128: nit: might be a tad clearer to use an `else` instead of chaining `else if`s.

joshklop (approved):

PR #1619, opened 2023-12-29
Time-out starknet_getTransactionStatus after 5s

joshklop:
to avoid sending 429s after waiting for 30s, due to feeder rate limiting.


rpc/handlers.go
```diff
@@ -1284,7 +1285,10 @@ func (h *Handler) TransactionStatus(ctx context.Context, hash felt.Felt) (*Trans
 			break
 		}
 
-		txStatus, err := h.feederClient.Transaction(ctx, &hash)
+		const timeout = 5 * time.Second
+		ctxChild, cancelChild := context.WithTimeout(ctx, timeout)
+		txStatus, err := h.feederClient.Transaction(ctxChild, &hash)
+		cancelChild()
```
omerfirmak, line 1291: Instead of this, should we make `feeder.Client` respect the timeout value that we pass to it by `WithTimeout` regardless of how many times it retries to complete the request?

joshklop: Maybe, but I think the current method of setting a per-request timeout and then controlling the overall timeout with a context also makes sense (makes me wonder if a per-request timeout is even necessary, but that's out-of-scope for this PR).

omerfirmak: > controlling the overall timeout with a context

That didn't work for Gateway client tho, that is why we ended up having to do https://github.com/NethermindEth/juno/pull/1614. I am worrying we will end up having to do the same here as well.

PR #1621, opened 2024-01-01
Initialize genesis state at node startup

omerfirmak:


blockchain/blockchain.go
```diff
@@ -776,11 +776,6 @@ func (b *Blockchain) HeadState() (core.StateReader, StateCloser, error) {
 		return nil, nil, err
 	}
 
-	_, err = chainHeight(txn)
```
joshklop, line 779: Is this change related to adding `InitGenesisState`?

blockchain/blockchain.go
```diff
@@ -1149,3 +1144,51 @@ func (b *Blockchain) Finalise(pending *Pending, sign BlockSignFunc) error {
...
+
+func (b *Blockchain) InitGenesisState(stateDiff *core.StateDiff, classes map[felt.Felt]core.Class) error {
+	return b.database.Update(func(txn db.Transaction) error {
+		_, err := genesisState(txn)
+		if !errors.Is(err, db.ErrKeyNotFound) {
+			return errors.New("genesis state already initiliazed")
+		}
+
+		state := core.NewState(txn)
+		if err = state.Update(0, stateDiff, classes); err != nil {
```
joshklop, line 1173: This will log changes to history for block zero, correct? Do we want that?

node/genesis.go
```diff
@@ -0,0 +1,26 @@
...
+
+import (
+	"github.com/NethermindEth/juno/blockchain"
+	"github.com/NethermindEth/juno/genesis"
+	"github.com/NethermindEth/juno/vm"
+)
+
+func buildGenesis(genesisPath string, bc *blockchain.Blockchain, v vm.VM) error {
+	_, err := bc.GenesisState()
+	if err == nil {
```
joshklop, line 11: Would this be a bit safer?
```go
if _, err := bc.GenesisState(); !errors.Is(err, db.ErrNotFound) {
       return err
}
```

blockchain/blockchain.go
joshklop: Ah I see. Now we can have a state without a chain.

node/genesis.go
omerfirmak: Nice, done.

blockchain/blockchain.go
omerfirmak: Yeah, it is not ideal, but doesn't really hurt either. Since it requires some changes to `core.State` API, I opted to deal with it later if needed.

omerfirmak:
Merging assuming these were all the comment @joshklop had, we can revisit any part of this PR again as needed.

PR #1625, opened 2024-01-02
Double http ReadTimeout

joshklop:
Large declare transactions can take more than 30s to read on a stable connection.


joshklop:
Decided to double the `ReadTimeout` after pushing the branch with the current (innacurate) name. Oh well.

omerfirmak (approved):

PR #1629, opened 2024-01-03
Add txns to blocks

omerfirmak:


genesis/genesis.go
```diff
@@ -112,7 +112,9 @@ func GenesisStateDiff(
 			return nil, nil, fmt.Errorf("execute function call: %v", err)
 		}
 	}
-	return genesisState.StateDiff(), newClasses, nil
+
+	genesisStateDiff, genesisClasses := genesisState.StateDiff()
```
joshklop, line 116: We use `genesisClasses` instead of `newClasses` so we capture the classes declared in function calls, correct?

builder/builder.go
```diff
@@ -42,14 +49,78 @@ func New(privKey *ecdsa.PrivateKey, ownAddr *felt.Felt, bc *blockchain.Blockchai
...
+	if b.pendingBlock.Block != nil {
+		return nil
+	}
+
+	bcPending, err := b.bc.Pending()
+	if err != nil {
+		return err
+	}
+
+	b.pendingBlock, err = utils.Clone[blockchain.Pending](bcPending)
```
joshklop, line 89: Is a round-trip through cbor necessary?

builder/builder.go
```diff
@@ -203,3 +207,76 @@ func Receipt(fee *felt.Felt, feeUnit core.FeeUnit, txHash *felt.Felt, trace *vm.
...
+				return err
+			}
+		}
+
+		select {
+		case <-ctx.Done():
+			return nil
+		case <-b.pool.Wait():
+			continue
+		}
```
joshklop, line 224: What do you think about having three cases?

1. `ctx.Done()` -> exit
2. `b.pool.Wait()` -> add tx to the pending block. If after adding the transaction, we reach a certain `maxTransactions` threshold, commit the block.
3. `time.After` -> timeout, commit the block.

joshklop:
Skimmed the changes, asked some questions (not meaning to nit-pick at this stage, just curious :)). Exciting!

blockchain/pending_test.go
```diff
@@ -275,6 +275,7 @@ func TestSetCompiledClassHash(t *testing.T) {
 
 func TestStateDiff(t *testing.T) {
 	state := makeState(t)
-	got := state.StateDiff()
-	require.Equal(t, core.EmptyStateDiff(), got)
+	gotSD, gotC := state.StateDiff()
+	require.Equal(t, core.EmptyStateDiff(), gotSD)
+	require.Equal(t, 0, len(gotC))
```
joshklop, line 280: `require.Empty`

builder/builder.go
```diff
@@ -132,3 +207,76 @@ func Receipt(fee *felt.Felt, feeUnit core.FeeUnit, txHash *felt.Felt, trace *vm.
...
+	b.pendingLock.Lock()
+	defer b.pendingLock.Unlock()
+
+	state := blockchain.NewPendingState(b.pendingBlock.StateUpdate.StateDiff, b.pendingBlock.NewClasses, b.headState)
+	var classes []core.Class
+	if txn.DeclaredClass != nil {
+		classes = append(classes, txn.DeclaredClass)
+	}
+
+	fee, trace, err := b.vm.Execute([]core.Transaction{txn.Transaction}, classes, b.pendingBlock.Block.Number,
```
rianhughes, line 263: vm.Execute() updates the state right? Think we should add a comment here?

builder/builder.go
```diff
@@ -29,27 +31,100 @@ type Builder struct {
...
+
+	go b.listenPool(ctx)
+
+	const blockTime = 5 * time.Second
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
+		case <-time.After(blockTime):
+			if err := b.Finalise(); err != nil {
```
rianhughes, line 72: nit: do you think it would be clearer to have Finalise return nil on a success, and then call b.initPendingBlock() after?

builder/builder.go
```diff
@@ -29,27 +31,100 @@ type Builder struct {
...
+	if err := b.bc.Finalise(&b.pendingBlock, b.Sign); err != nil {
+		return err
+	}
+	b.log.Infow("Finalised block", "number", b.pendingBlock.Block.Number, "hash",
+		b.pendingBlock.Block.Hash.ShortString(), "state", b.pendingBlock.Block.GlobalStateRoot.ShortString())
+
+	if err := b.clearPending(); err != nil {
+		return err
+	}
+	return b.initPendingBlock()
```
rianhughes, line 125: nit: do you think we should just return nil, and call clearPending() / initPendingBlock() outside of Finalise()?

builder/builder.go
```diff
@@ -29,27 +31,100 @@ type Builder struct {
...
 
 		bc:       bc,
+		pool:     pool,
 		vm:       builderVM,
 		newHeads: feed.New[*core.Header](),
+		log:      log,
 	}
 }
 
 func (b *Builder) Run(ctx context.Context) error {
```
rianhughes, line 58: Maybe add a comment / documentation for this? Eg "Mempool transactions are continuously being popped, and run against the pending state. Every N seconds the pending state is applied to the blockchain, and then reset."

builder/builder.go
omerfirmak: This is the only generic enough way we have to deep copy a struct.

omerfirmak: It is a easy way of deep copying a struct. We can replace it with a dependency later on.

builder/builder.go
omerfirmak: Sadly, that doesnt really work without nesting two for loops.
Mainly because `b.pool.Wait()` can block even with txns in the pool. Wait() is only supposed to be called when we consume all the txns in the pool.

builder/builder.go
omerfirmak: We could and actually should in the long run, for better and more granular error handling. But we can keep it the way it is for simplicity for now.

genesis/genesis.go
omerfirmak: I believe the code is functionally exactly the same, I must have done this just for consistency.

builder/builder.go
omerfirmak: Yeap, it does. What is it that you want clarified in the comment exactly?

builder/builder.go
omerfirmak: It will most probably evolve a lot really fast. I am leaving documentation for the later stages.

builder/builder.go
rianhughes: Maybe builder isn't the best place to have this comment, but imo, it would be useful to have in the definition of Execute() (vm/vm.go)

builder/builder.go
```diff
@@ -156,3 +222,76 @@ func Receipt(fee *felt.Felt, feeUnit core.FeeUnit, txHash *felt.Felt, trace *vm.
...
+			return nil
+		case <-b.pool.Wait():
+			continue
+		}
+	}
+}
+
+func (b *Builder) depletePool(ctx context.Context) error {
+	for {
+		userTxn, err := b.pool.Pop()
```
rianhughes, line 245: Does this return db.ErrKeyNotFound is the pool is empty? What do you think about opening a separate PR so that `Pop() `returns `txn,ok,err`?

omerfirmak: > What do you think about opening a separate PR so that Pop() returns txn,ok,err?

What do you think the benefit would be? Compared to checking if `err` is  `db.ErrKeyNotFound`

rianhughes (approved):

PR #1631, opened 2024-01-04
Genesis block

joshklop:


blockchain/blockchain.go
```diff
@@ -1145,50 +1145,54 @@ func (b *Blockchain) Finalise(pending *Pending, sign BlockSignFunc) error {
...
-func genesisState(txn db.Transaction) (*core.StateUpdate, error) {
-	var genesisStateUpdate *core.StateUpdate
-	return genesisStateUpdate, txn.Get(db.GenesisState.Key(), func(b []byte) error {
-		genesisStateUpdate = new(core.StateUpdate)
-		return encoder.Unmarshal(b, genesisStateUpdate)
-	})
-}
+func (b *Blockchain) StoreGenesis(diff *core.StateDiff, classes map[felt.Felt]core.Class) error {
+	receipts := make([]*core.TransactionReceipt, 0)
+	block := &core.Block{
```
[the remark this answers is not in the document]
joshklop: The genesis block has pre-defined fields for now, so it only needs to be defined in one place. If we want different genesis blocks for different networks, we can expand the genesis config and let the `genesis` package build the block.

Because the genesis config doesn't have any influence over the genesis block, I think leaving the block in `StoreGenesis` makes sense.

blockchain/blockchain.go
```diff
@@ -1145,50 +1145,54 @@ func (b *Blockchain) Finalise(pending *Pending, sign BlockSignFunc) error {
...
-		return txn.Set(db.GenesisState.Key(), genesisStateUpdateBytes)
+		if err = storeStateUpdate(txn, 0, &core.StateUpdate{
+			BlockHash: block.Hash,
+			NewRoot:   root,
+			OldRoot:   new(felt.Felt),
+			StateDiff: diff,
+		}); err != nil {
+			return err
+		}
+		return b.storeBlock(txn, block, commitments)
```
omerfirmak, line 1196: We should be easily able to use Finalise here, no? It is essentially the same logic.

omerfirmak: see the last commit

joshklop: Thanks! I did something similar in https://github.com/NethermindEth/juno/pull/1642, but this patch makes that diff simpler so it's good to apply anyway.

PR #1639, opened 2024-01-09
Implement custom network

rianhughes:
closes https://github.com/NethermindEth/juno/issues/1415

Approach: Implement a new "custom-network" flag that accepts a json-string with the network parameters (name, chainId, etc). Users can specify either (not both) network or custom-network flag.


rianhughes:
@joshklop & @omerfirmak if you guys are happy with this approach, then I'll create a new commit that passes network as a pointer to resolve the "heavy-parameter" lint errors.

utils/network.go
```diff
@@ -62,123 +140,70 @@ func (n *Network) MarshalJSON() ([]byte, error) {
...
 }
 
-// FeederURL returns URL for read commands
-func (n Network) FeederURL() string {
-	return n.baseURL() + "feeder_gateway/"
-}
+func (n *Network) Validate() error {
+	if !(n.Name == "CUSTOM" || n.Name == "custom") {
+		return ErrUnknownNetwork
+	}
```
joshklop, line 176: Why do we need this?

utils/network.go
```diff
@@ -62,123 +140,70 @@ func (n *Network) MarshalJSON() ([]byte, error) {
...
+		CoreContractAddress string `json:"core_contract_address"`
+		*Alias
+	}{
+		Alias: (*Alias)(n),
 	}
+	if err := json.Unmarshal(data, &aux); err != nil {
+		return err
+	}
+	n.CoreContractAddress = common.HexToAddress(aux.CoreContractAddress)
+	return nil
```
joshklop, line 194: Why do we need to override `UnmarshalJSON` to get a new format? Apologies if this was discussed on one of the prior PRs, but my intuition says the default behavior should be good here.

utils/network.go
```diff
@@ -62,123 +140,70 @@ func (n *Network) MarshalJSON() ([]byte, error) {
...
-		return common.Address{}, ErrUnknownNetwork
-	}
-	return address, nil
+func (n *Network) UnmarshalText(text []byte) error {
+	return n.Set(string(text))
 }
 
-func (n Network) ChainID() *felt.Felt {
-	return new(felt.Felt).SetBytes([]byte(n.ChainIDString()))
+func (n Network) ChainIDFelt() *felt.Felt {
```
joshklop, line 205: nit: can we rename this `L2ChainIDFelt`?

validator/validator.go
```diff
@@ -30,11 +29,40 @@ func Validator() *validator.Validate {
...
 				return t.String()
 			}
 			panic("not a TransactionType")
-		}, rpc.TransactionType(0))
+		}, TransactionType(0))
 	})
 	return v
 }
+
+// TransactionType is duplicated here to prevent import cycles
```
joshklop, line 41: We really need to move the rpc types to a separate package. I'll try to send a PR for that today.

utils/network.go
```diff
@@ -62,123 +140,70 @@ func (n *Network) MarshalJSON() ([]byte, error) {
...
-	case Integration:
-		return "https://external.integration.starknet.io/"
-	case Sepolia:
-		return "https://alpha-sepolia.starknet.io/"
-	case SepoliaIntegration:
-		return "https://integration-sepolia.starknet.io/"
-	default:
-		// Should not happen.
-		panic(ErrUnknownNetwork)
+func (n *Network) SetCustomNetwork(s string) error {
```
joshklop, line 165: I think we can remove this and just use `UnmarshalJSON` directly.

cmd/juno/juno.go
```diff
@@ -202,8 +205,19 @@ func NewCmd(config *node.Config, run func(*cobra.Command, []string) error) *cobr
...
+		if err := v.Unmarshal(config, viper.DecodeHook(mapstructure.ComposeDecodeHookFunc(
+			mapstructure.TextUnmarshallerHookFunc(), mapstructure.StringToTimeDurationHookFunc()))); err != nil {
+			return err
+		}
+
+		if v.GetString(customNetworkF) != "" {
+			err := config.Network.SetCustomNetwork(v.GetString(customNetworkF))
+			if err != nil {
+				return err
+			}
```
joshklop, line 217: I'm starting to doubt my idea about using a json string on the command line. Can we change it to a file name?
Sorry for the back-and-forth on this part.

utils/network.go
rianhughes: Its because strings aren't automatically unmarshalled into common.Address types (at least when I tried it last - but I can look into it again)

utils/network.go
```diff
@@ -8,49 +8,127 @@ import (
...
+	"github.com/NethermindEth/juno/validator"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/libp2p/go-libp2p/core/protocol"
 	"github.com/spf13/pflag"
 )
 
-var ErrUnknownNetwork = errors.New("unknown network (known: mainnet, goerli, goerli2, integration)")
+var (
+	ErrUnknownNetwork              = errors.New("unknown network (known: mainnet, goerli, goerli2, integration)")
+	ErrInvalidCustomNetworkJSONStr = errors.New("invalid custom-network json string")
```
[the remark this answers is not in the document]
rianhughes: These are also used in network_test.go so can't be inlined / unexported, unless we inline errors in the tests. I'm not sure what the "correct" approach is, but the current implementation seems a bit neater to me. Happy either way.

core/block.go
```diff
@@ -95,7 +95,7 @@ func NetworkBlockHashMetaInfo(network utils.Network) *blockHashMetaInfo {
 		}
 	default:
 		// This should never happen
-		panic(fmt.Sprintf("unknown network: %d", network))
+		panic(fmt.Sprintf("unknown network: %s", network.String()))
```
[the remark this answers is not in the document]
rianhughes: Yup, good point. I'll remove this function.

rianhughes:
Note the above commit where network needs to be passed as a pointer, because it's too heavy.

utils/network.go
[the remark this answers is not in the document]
rianhughes: Update: 
ErrUnknownNetwork is unexported
ErrInvalidCustomNetworkJSONStr is no longer needed

utils/network.go
rianhughes: Update: this is no longer needed, so has been removed

utils/network.go
rianhughes: Update: SetCustomNetwork is no longer needed so has been removed

cmd/juno/juno.go
```diff
@@ -202,8 +224,25 @@ func NewCmd(config *node.Config, run func(*cobra.Command, []string) error) *cobr
...
-			mapstructure.TextUnmarshallerHookFunc(), mapstructure.StringToTimeDurationHookFunc())))
+		if err := v.Unmarshal(config, viper.DecodeHook(mapstructure.ComposeDecodeHookFunc(
+			mapstructure.TextUnmarshallerHookFunc(), mapstructure.StringToTimeDurationHookFunc()))); err != nil {
+			return err
+		}
+
+		// Set custom network
+		if config.Network.String() == "custom" {
+			if v.GetString(cnFeederURLF) == "" {
+				return errors.New("custom networks must specify all `cn` flags (--cn-feeder-url etc)")
```
joshklop, line 235: Can we remove these checks and make `--network` mutually exclusive with `--cn-feeder-url` instead? Then we can allow for custom network names, too.

utils/network.go
```diff
@@ -6,54 +6,132 @@ import (
...
-	default:
-		// Should not happen.
-		panic(ErrUnknownNetwork)
-	}
+func (n *Network) String() string {
+	return n.Name
 }
 
-func (n Network) MarshalYAML() (interface{}, error) {
+func (n *Network) MarshalYAML() (interface{}, error) {
```
joshklop, line 134:
```suggestion
func (n *Network) MarshalYAML() (any, error) {
```
nit

utils/network.go
```diff
@@ -62,23 +140,23 @@ func (n *Network) MarshalJSON() ([]byte, error) {
...
+		"mainnet":             Mainnet,
+		"goerli":              Goerli,
+		"goerli2":             Goerli2,
+		"integration":         Integration,
+		"sepolia":             Sepolia,
+		"sepolia-integration": SepoliaIntegration,
+		"custom":              CustomNetwork,
+	}
+
+	if network, ok := predefinedNetworks[sLower]; ok {
```
joshklop, line 155:
```suggestion
	if network, ok := predefinedNetworks[strings.ToLower(s)]; ok {
```
nit: we can remove the `sLower` variable.

validator/validator.go
joshklop: Actually, I reverted this change locally and it appears to work now. Maybe this change was necessary at some point but is now unnecessary?

cmd/juno/juno.go
rianhughes: In the current implementation, we require network to be set to "custom" if users want to run a custom network, so these can't be made mutually exclusive. Wasn't sure how to do this in a neater way beyond creating a new function etc

validator/validator.go
rianhughes: Yup, nice catch! This is no longer needed, and has been reverted.

cmd/juno/juno.go
```diff
@@ -231,6 +270,12 @@ func NewCmd(config *node.Config, run func(*cobra.Command, []string) error) *cobr
 	junoCmd.Flags().Uint16(wsPortF, defaultWSPort, wsPortUsage)
 	junoCmd.Flags().String(dbPathF, defaultDBPath, dbPathUsage)
 	junoCmd.Flags().Var(&defaultNetwork, networkF, networkUsage)
+	junoCmd.Flags().String(cnFeederURLF, defaultCNFeederURL, networkCustomFeederUsage)
+	junoCmd.Flags().String(cnGatewayURLF, defaultCNGatewayURL, networkCustomGatewayUsage)
+	junoCmd.Flags().Int(cnL1ChainIDF, defaultCNL1ChainID, networkCustomL1ChainIDUsage)
+	junoCmd.Flags().String(cnL2ChainIDF, defaultCNL2ChainID, networkCustomL2ChainIDUsage)
+	junoCmd.Flags().Int(cnProtocolIDF, defaultCNProtocolID, networkCustomProtocolIDUsage)
```
omerfirmak, line 277: looks like unused?

cmd/juno/juno.go
rianhughes: Update: I didn't originally understand, but I'm updated the code so that these flags are mutually exclusive.

cmd/juno/juno.go
rianhughes: Yup, cnProtocolIDF isn't needed. I've removed it.

cmd/juno/juno.go
```diff
@@ -207,8 +228,24 @@ func NewCmd(config *node.Config, run func(*cobra.Command, []string) error) *cobr
...
 		// encoding.TextUnmarshaller interface (see the LogLevel type for an example).
-		return v.Unmarshal(config, viper.DecodeHook(mapstructure.ComposeDecodeHookFunc(
-			mapstructure.TextUnmarshallerHookFunc(), mapstructure.StringToTimeDurationHookFunc())))
+		if err := v.Unmarshal(config, viper.DecodeHook(mapstructure.ComposeDecodeHookFunc(
+			mapstructure.TextUnmarshallerHookFunc(), mapstructure.StringToTimeDurationHookFunc()))); err != nil {
+			return err
+		}
+
+		// Set custom network
+		if v.GetString(cnNameF) != "" {
```
joshklop, line 237:
```suggestion
		if v.IsSet(cnNameF) {
```

cmd/juno/juno.go
```diff
@@ -207,8 +228,24 @@ func NewCmd(config *node.Config, run func(*cobra.Command, []string) error) *cobr
...
+			config.Network = utils.Network{}
+			config.Network.Name = v.GetString(cnNameF)
+			config.Network.FeederURL = v.GetString(cnFeederURLF)
+			config.Network.GatewayURL = v.GetString(cnGatewayURLF)
+			config.Network.L1ChainID = new(big.Int).SetInt64(v.GetInt64(cnL1ChainIDF))
+			config.Network.L2ChainID = v.GetString(cnL2ChainIDF)
+			coreContractAddress := common.HexToAddress(v.GetString(cnCoreContractAddressF))
+			config.Network.CoreContractAddress = coreContractAddress
+		}
```
joshklop, line 246: nit: slightly more idiomatic this way I think.
```go
			config.Network = utils.Network{
			        Name: v.GetString(cnNameF),
			        // rest of the fields here.
			}
```

joshklop, line 246: We need to set the metainfo here right? Otherwise block hash will panic. Or we can check if it is `nil` during the block hash calculation.

rianhughes: I think requiring users to set metaInfo for custom networks probably adds unnecessary complexity. I've updated the code so it shouldn't panic if meta info isn't set.

core/block.go
```diff
@@ -124,38 +72,40 @@ func VerifyBlockHash(b *Block, network utils.Network) (*BlockCommitments, error)
...
 
-	metaInfo := NetworkBlockHashMetaInfo(network)
-	unverifiableRange := metaInfo.UnverifiableRange
-	for _, fallbackSeq := range []*felt.Felt{&felt.Zero, metaInfo.FallBackSequencerAddress} {
-		var overrideSeq *felt.Felt
-		if b.SequencerAddress == nil {
-			overrideSeq = fallbackSeq
-		}
+	metaInfo := network.BlockHashMetaInfo
+	if metaInfo != nil {
```
joshklop, line 76: We should still calculate the block hash if meta info isn't set. We can have a default meta info.

cmd/juno/juno.go
joshklop: We could set the default ourselves so the user doesn't have to specify.

```go
network := utils.Network{
    // ... 
    BlockHashMetaInfo: &utils.BlockHashMetaInfo{
        // ...
    },
}
```

rianhughes: Sure, that works too. I've set the First07Block as 0, fallBackSeqAddress the same as Goerli (and all non-mannet networks), and let UnverifRange unset.

core/block.go
rianhughes: Reverted this change.

cmd/juno/juno.go
```diff
@@ -236,6 +279,12 @@ func NewCmd(config *node.Config, run func(*cobra.Command, []string) error) *cobr
 	junoCmd.Flags().Uint16(wsPortF, defaultWSPort, wsPortUsage)
 	junoCmd.Flags().String(dbPathF, defaultDBPath, dbPathUsage)
 	junoCmd.Flags().Var(&defaultNetwork, networkF, networkUsage)
+	junoCmd.Flags().String(cnNameF, defaultCNName, networkCustomName)
+	junoCmd.Flags().String(cnFeederURLF, defaultCNFeederURL, networkCustomFeederUsage)
+	junoCmd.Flags().String(cnGatewayURLF, defaultCNGatewayURL, networkCustomGatewayUsage)
+	junoCmd.Flags().Int(cnL1ChainIDF, defaultCNL1ChainID, networkCustomL1ChainIDUsage)
```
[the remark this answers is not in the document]
rianhughes: Updated so users must provide it as a string.

cmd/juno/juno.go
```diff
@@ -207,8 +229,33 @@ func NewCmd(config *node.Config, run func(*cobra.Command, []string) error) *cobr
...
+			config.Network = utils.Network{
+				Name:                v.GetString(cnNameF),
+				FeederURL:           v.GetString(cnFeederURLF),
+				GatewayURL:          v.GetString(cnGatewayURLF),
+				L1ChainID:           l1ChainID,
+				L2ChainID:           v.GetString(cnL2ChainIDF),
+				CoreContractAddress: common.HexToAddress(v.GetString(cnCoreContractAddressF)),
+				BlockHashMetaInfo: &utils.BlockHashMetaInfo{
+					First07Block:             0,
+					FallBackSequencerAddress: fallBackSequencerAddress,
```
omerfirmak, line 253: Custom networks shouldn't have fallback sequencer address

rianhughes: I've removed it, and updated the VerifyBlockHash func

PR #1640, opened 2024-01-09
Minor changes for sequencer

joshklop:
Factoring out some of the smaller changes from #1629

### DIFF
--- a/blockchain/pending.go
+++ b/blockchain/pending.go
@@ -120,8 +120,8 @@ func (p *PendingState) SetCompiledClassHash(classHash, compiledClassHash *felt.F
 	return nil
 }
 
-// StateDiff returns the pending state's internal state diff. The returned object will continue to be
+// StateDiffAndClasses returns the pending state's internal data. The returned objects will continue to be
 // read and modified by the pending state.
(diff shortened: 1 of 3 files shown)


blockchain/pending.go
```diff
@@ -122,6 +122,6 @@ func (p *PendingState) SetCompiledClassHash(classHash, compiledClassHash *felt.F
 
 // StateDiff returns the pending state's internal state diff. The returned object will continue to be
 // read and modified by the pending state.
-func (p *PendingState) StateDiff() *core.StateDiff {
-	return p.stateDiff
+func (p *PendingState) StateDiff() (*core.StateDiff, map[felt.Felt]core.Class) {
```
rianhughes, line 125: Do you think it would be more clear if we had a separate `func (p *PendingState) NewClasses()` function?

joshklop: We can rename it to `StateDiffAndClasses`. Wdyt?

rianhughes: Yeah that works too

rianhughes (approved):

PR #1642, opened 2024-01-10
Build blocks

joshklop:
Relies on #1631.

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -2,6 +2,8 @@ package builder
 
 import (
 	"context"
+	"fmt"
+	"time"
 
 	"github.com/NethermindEth/juno/adapters/vm2core"
 	"github.com/NethermindEth/juno/blockchain"
@@ -26,16 +28,24 @@ type Builder struct {
 	ownAddress felt.Felt
 	privKey    *ecdsa.PrivateKey
 
-	bc       *blockchain.Blockchain
-	vm       vm.VM
-	newHeads *feed.Feed[*core.Header]
-	log      utils.Logger
+	bc        *blockchain.Blockchain
+	vm        vm.VM
+	newHeads  *feed.Feed[*core.Header]
+	log       utils.Logger
+	blockTime time.Duration
+
+	listener EventListener
 }
 
-func New(privKey *ecdsa.PrivateKey, ownAddr *felt.Felt, bc *blockchain.Blockchain, builderVM vm.VM, log utils.Logger) *Builder {
+func New(privKey *ecdsa.PrivateKey, ownAddr *felt.Felt, bc *blockchain.Blockchain, builderVM vm.VM,
+	blockTime time.Duration, log utils.Logger,
+) *Builder {
 	return &Builder{
 		ownAddress: *ownAddr,
 		privKey:    privKey,
+		blockTime:  blockTime,
+		log:        log,
+		listener:   &SelectiveListener{},
 
 		bc:       bc,
 		vm:       builderVM,
@@ -43,9 +53,32 @@ func New(privKey *ecdsa.PrivateKey, ownAddr *felt.Felt, bc *blockchain.Blockchai
 	}
 }
 
+func (b *Builder) WithEventListener(l EventListener) *Builder {
+	b.listener = l
+	return b
+}
+
+// Run takes ownership of the Builder.
 func (b *Builder) Run(ctx context.Context) error {
-	<-ctx.Done()
-	return nil
+	ticker := time.NewTicker(b.blockTime)
+	defer ticker.Stop()
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
+		case <-ticker.C:
+			pending, err := b.bc.Pending()
+			if err != nil {
+				return fmt.Errorf("get pending: %v", err)
+			}
+			pending.Block.SequencerAddress = &b.ownAddress
+			if err = b.bc.Finalise(&pending, b.Sign); err != nil {
+				return fmt.Errorf("finalise: %v", err)
+			}
+			b.log.Infow("Finalised block", "number", pending.Block.Number, "hash", pending.Block.Hash.ShortString())
+			b.listener.OnBlockFinalised(pending.Block.Header)
+		}
+	}
 }
 
 // ValidateAgainstPendingState validates a user transaction against the pending state
--- a/builder/builder_test.go
+++ b/builder/builder_test.go
@@ -5,6 +5,7 @@ import (
 	"crypto/rand"
 	"errors"
 	"testing"
+	"time"
 
 	"github.com/NethermindEth/juno/blockchain"
 	"github.com/NethermindEth/juno/builder"
@@ -30,7 +31,7 @@ func TestValidateAgainstPendingState(t *testing.T) {
 	mockVM := mocks.NewMockVM(mockCtrl)
 	bc := blockchain.New(testDB, utils.Integration)
 	seqAddr := utils.HexToFelt(t, "0xDEADBEEF")
-	testBuilder := builder.New(nil, seqAddr, bc, mockVM, utils.NewNopZapLogger())
+	testBuilder := builder.New(nil, seqAddr, bc, mockVM, 0, utils.NewNopZapLogger())
 
 	client := feeder.NewTestClient(t, utils.Integration)
 	gw := adaptfeeder.New(client)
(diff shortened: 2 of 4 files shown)


node/node.go
```diff
@@ -151,7 +151,7 @@ func New(cfg *Config, version string) (*Node, error) { //nolint:gocyclo,funlen
 			return nil, kErr
 		}
 
-		sequencer := builder.New(pKey, new(felt.Felt).SetUint64(1337), chain, nodeVM, log) //nolint: gomnd
+		sequencer := builder.New(pKey, new(felt.Felt).SetUint64(1337), chain, nodeVM, time.Minute, log) //nolint: gomnd
```
rianhughes, line 154: I think this should be (eg) `time.Now().Unix()`, since `time.Minute` always outputs "1m0s"

blockchain/blockchain.go
```diff
@@ -1099,96 +1100,97 @@ type BlockSignFunc func(blockHash, stateDiffCommitment *felt.Felt) ([]*felt.Felt
 
 // Finalise will calculate the state commitment and block hash for the given pending block and append it to the
 // blockchain
-func (b *Blockchain) Finalise(pending *Pending, sign BlockSignFunc) error {
+func (b *Blockchain) Finalise(txs []core.Transaction, receipts []*core.TransactionReceipt, diff *core.StateDiff,
+	classes map[felt.Felt]core.Class, seqAddr *felt.Felt, sign BlockSignFunc,
```
rianhughes, line 1104: I might be missing something, but currently `classes` aren't used or stored.

blockchain/blockchain.go
```diff
@@ -1099,96 +1100,97 @@ type BlockSignFunc func(blockHash, stateDiffCommitment *felt.Felt) ([]*felt.Felt
...
+		pending.Block.TransactionCount = uint64(len(txs))
+		pending.Block.Receipts = receipts
+		for _, receipt := range receipts {
+			pending.Block.EventCount += uint64(len(receipt.Events))
 		}
+		pending.Block.EventsBloom = core.EventsBloom(receipts)
+		pending.Block.Timestamp = uint64(time.Now().Unix())
+		pending.Block.ProtocolVersion = supportedStarknetVersion.String()
+		// TODO gas prices
+		pending.StateUpdate.StateDiff = diff
```
rianhughes, line 1131: Do we not also have to set the NewRoot, OldRoot and BlockHash?

rianhughes: Ah, nvm, just seen that it's set in `storeEmptyPending()` whenever we store a pending block.

blockchain/blockchain.go
```diff
@@ -1099,96 +1100,97 @@ type BlockSignFunc func(blockHash, stateDiffCommitment *felt.Felt) ([]*felt.Felt
 
 // Finalise will calculate the state commitment and block hash for the given pending block and append it to the
 // blockchain
-func (b *Blockchain) Finalise(pending *Pending, sign BlockSignFunc) error {
+func (b *Blockchain) Finalise(txs []core.Transaction, receipts []*core.TransactionReceipt, diff *core.StateDiff,
```
rianhughes, line 1103: Wdty about either:
1. renaming `Finalise() `to (eg) `BuildAndStorePending()`?
2. removing `Finalise() , creating a `BuildPending()` funciton, and just calling `b.finalise()` on that?

joshklop: I think a renaming refactor could be done outside this PR. As long as the name reflects that the pending block is appended to the chain, I'm good.

blockchain/blockchain.go
```diff
@@ -1099,58 +1100,44 @@ type BlockSignFunc func(blockHash, stateDiffCommitment *felt.Felt) ([]*felt.Felt
...
```
omerfirmak, line 1103: what's the benefit of this change here?

blockchain/blockchain.go
```diff
@@ -1099,58 +1100,44 @@ type BlockSignFunc func(blockHash, stateDiffCommitment *felt.Felt) ([]*felt.Felt
...
-
-		if err = b.storeBlock(txn, pending.Block, commitments); err != nil {
-			return err
-		}
-
-		return storeStateUpdate(txn, pending.Block.Number, pending.StateUpdate)
+		pending.Block.EventsBloom = core.EventsBloom(receipts)
+		pending.Block.Timestamp = uint64(time.Now().Unix())
+		pending.Block.ProtocolVersion = supportedStarknetVersion.String()
+		// TODO gas prices
```
omerfirmak, line 1130: blockchain has no means of setting gas prices.

blockchain/blockchain.go
```diff
@@ -1099,58 +1100,44 @@ type BlockSignFunc func(blockHash, stateDiffCommitment *felt.Felt) ([]*felt.Felt
...
-			pending.Block.Signatures = append(pending.Block.Signatures, sig)
-		}
-
-		if err = b.storeBlock(txn, pending.Block, commitments); err != nil {
-			return err
-		}
-
-		return storeStateUpdate(txn, pending.Block.Number, pending.StateUpdate)
+		pending.Block.EventsBloom = core.EventsBloom(receipts)
+		pending.Block.Timestamp = uint64(time.Now().Unix())
```
omerfirmak, line 1128: this should be the time of block creation, not finalization.

blockchain/blockchain.go
joshklop: The goal was to make the function signature communicate what the builder needs to construct. Passing in `pending` feels a little more risky, since many of the fields are overwritten or ignored until the block hash calculation.

But I'll drop that commit for now in an effort to simplify the PR and save refactors for later.

blockchain/blockchain.go
joshklop: Ah good point. We should modify `storeEmptyPending` to take care of these details imo.

rianhughes (approved):

builder/builder.go
```diff
@@ -28,26 +30,62 @@ type Builder struct {
...
-	log      utils.Logger
+	bc        *blockchain.Blockchain
+	vm        vm.VM
+	newHeads  *feed.Feed[*core.Header]
+	log       utils.Logger
+	blockTime time.Duration
+
+	// Returns true when the builder should exit. Called after finalising each block.
+	// Useful for tests.
+	finaliseCb func() bool
```
omerfirmak, line 41: Let's add an EventListener instead of callbacks.

builder/builder.go
```diff
@@ -26,26 +28,59 @@ type Builder struct {
...
+				return fmt.Errorf("get pending: %v", err)
+			}
+			pending.Block.SequencerAddress = &b.ownAddress
+			if err = b.bc.Finalise(&pending, b.Sign); err != nil {
+				return fmt.Errorf("finalise: %v", err)
+			}
+			b.log.Infow("Finalised block", "number", pending.Block.Number, "hash", pending.Block.Hash.ShortString())
+			if b.listener.OnBlockFinalised(pending.Block.Header) {
+				return nil
+			}
```
omerfirmak, line 81: These are just callbacks, not hooks. So they shouldn't affect execution. To terminate the builder execution, we should cancel `ctx`.

joshklop: That makes it pretty difficult to test. If we allow them to be hooks, we can execute a specific number of blocks.

joshklop: I tried other packages with a `Clock` interface that we can mock, but they all brought complexity we don't need imo.

builder/builder_test.go
```diff
@@ -202,3 +203,51 @@ func TestReceipt(t *testing.T) {
...
+	}
+}
+
+func (f *finisher) OnBlockFinalised(_ *core.Header) {
+	if f.i < f.min {
+		f.i++
+	} else {
+		f.cancel()
+	}
+}
```
omerfirmak, line 226: Let's generalize this so that we can easily extend and reuse it.

Maybe something like?

``` golang
var _ builder.EventListener = (*testListener)(nil)

type testListener struct {}

func (l *testListener) WaitForBlockAndDo(blockNumber, do func())
```

`WaitForBlockAndDo` is supposed to block until builder finalizes a block for height `blockNumber` and calls `do`. So to terminate a running builder we can easily call `WaitForBlockAndDo(N, contextCancel)`

We can later on extend testListener to wait for all sorts of events, like txn inclusion for example.

omerfirmak: > Can we wait until we need to generalize to make these decisions

It wont be too long before we need to do that. But, fair enough.

omerfirmak (approved):

PR #1654, opened 2024-01-12
Only return ErrBlockNotFound if it is not found

joshklop:
Did a similar refactor a while ago for blockByID. This commit is for blockHeaderByID.

### DIFF
--- a/rpc/handlers.go
+++ b/rpc/handlers.go
@@ -484,22 +484,34 @@ func (h *Handler) blockByID(id *BlockID) (*core.Block, *jsonrpc.Error) {
 	return block, nil
 }
 
-func (h *Handler) blockHeaderByID(id *BlockID) (*core.Header, error) {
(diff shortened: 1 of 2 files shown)


rpc/handlers.go
```diff
@@ -483,22 +483,35 @@ func (h *Handler) blockByID(id *BlockID) (*core.Block, *jsonrpc.Error) {
...
 	case id.Hash != nil:
-		return h.bcReader.BlockHeaderByHash(id.Hash)
+		header, err = h.bcReader.BlockHeaderByHash(id.Hash)
 	case id.Pending:
-		pending, err := h.bcReader.Pending()
+		var pending blockchain.Pending
+		pending, err = h.bcReader.Pending()
 		if err != nil {
-			return nil, err
+			return nil, ErrInternal.CloneWithData(fmt.Errorf("get pending: %v", err))
```
omerfirmak, line 498: To make things simpler, we can assume Pending() can return db.ErrKeyNotFound as well.

PR #1681, opened 2024-01-22
Fix Juno snapshots path

ymittal:
It looks like the main snapshots directory inside the tar file was renamed from `juno_mainnet` to `juno-mainnet`. I had an older snapshot tar file, which when I unzipped created `snapshots/juno_mainnet`. However, upon downloading the latest, it looks like the main directory is now `snapshots/juno-mainner`. Was pulling my hair out on why my sync was starting from the very beginning haha.


wojciechos:
Thanks for reporting that! We will just update folder name in the snapshot to not make any confusion as same path is already used in other docs or tutorials

PR #1695, opened 2024-01-31
Make gateway timeout configurable

omerfirmak:


kirugan (approved):

cmd/juno/juno.go
```diff
@@ -148,6 +150,7 @@ const (
 	dbCacheSizeUsage         = "Determines the amount of memory (in megabytes) allocated for caching data in the database."
 	dbMaxHandlesUsage        = "A soft limit on the number of open files that can be used by the DB"
 	gwAPIKeyUsage            = "API key for gateway endpoints to avoid throttling" //nolint: gosec
+	gwTimeoutUsage           = "Timeout for requests made to the gateway"          //nolint: gosec
```
wojciechos, line 153: Should we call it gateway timeout? Isn't that releated to feeder client?

wojciechos (approved):

cmd/juno/juno.go
wojciechos: I mean there is separate timeout for gateway(adding txn) hardcoded for 1 minute.

omerfirmak: Yeah, I wasn't sure either. But `gwAPIKey` is used for feeder requests but it is still called gateway API key. I just wanted to keep it consistent with the existing plugin.

PR #1702, opened 2024-02-14
Restructuring p2p handlers

kirugan:


p2p/starknet/handlers.go
```diff
@@ -338,6 +380,7 @@ func newFin(finMsg proto.Message) Stream[proto.Message] {
 	}
 }
 
+// todo change this logic later on
```
kirugan, line 383: I'll do that after this PR is merged

rianhughes (approved):

Issue #1705, opened 2024-02-15
Starknet 0.13.1 interim blockheaders


rianhughes:
See points 1,2,3 and 5 in https://docs.google.com/document/d/16CyAZw8FnaxY8djZPF9cclwjARsj45940z787XtLHAM/edit
(point 4 is in a separate PR)

PR #1706, opened 2024-02-15
update blockheader for v0.13.1

rianhughes:
closes https://github.com/NethermindEth/juno/issues/1705


core/block.go
```diff
@@ -39,6 +39,16 @@ type Header struct {
...
+	// The mode of the L1 data availability
+	L1DAMode string
+	// The gas price for L1 data availability
+	L1DataGasPrice GasPrice
+	// The gas price for L1
+	L1GasPrice GasPrice
+}
+type GasPrice struct {
+	PriceInWei *felt.Felt `json:"price_in_wei"`
+	PriceInFri *felt.Felt `json:"price_in_fri"`
```
omerfirmak, line 51:
```suggestion
	PriceInWei *felt.Felt
	PriceInFri *felt.Felt
```

core/block.go
```diff
@@ -39,6 +39,16 @@ type Header struct {
 	Signatures [][]*felt.Felt
 	// Amount of STRK charged per Gas spent
 	GasPriceSTRK *felt.Felt
+	// The mode of the L1 data availability
+	L1DAMode string
```
omerfirmak, line 43: any chance we can make this an enum?

core/block.go
```diff
@@ -39,6 +39,16 @@ type Header struct {
...
+	// The gas price for L1 data availability
+	L1DataGasPrice GasPrice
+	// The gas price for L1
+	L1GasPrice GasPrice
```
omerfirmak, line 47: these won't be available for old block, we can make them pointers

```suggestion
	L1DataGasPrice *GasPrice
	// The gas price for L1
	L1GasPrice *GasPrice
```

core/block.go
```diff
@@ -39,6 +39,24 @@ type Header struct {
...
+	// The gas price for L1
+	L1GasPrice *GasPrice
+}
+
+type L1DAMode string
+
+const (
+	CallData L1DAMode = "CALLDATA"
+	Blob     L1DAMode = "BLOB"
+)
```
omerfirmak, line 55: our `core` enums are all integers if I am not mistaken, can we make this something similar to;

```golang
type Resource uint32

const (
	ResourceL1Gas Resource = iota + 1
	ResourceL2Gas
)

func (r Resource) String() string {
	switch r {
	case ResourceL1Gas:
		return "L1_GAS"
	case ResourceL2Gas:
		return "L2_GAS"
	default:
		return ""
	}
}
```

core/block.go
```diff
@@ -21,6 +21,11 @@ type Header struct {
 	Number uint64
 	// The state commitment after this block
 	GlobalStateRoot *felt.Felt
+
+	TransactionCommitment *felt.Felt
+
+	EventCommitment *felt.Felt
+
```
omerfirmak, line 28: we do have these in a separate DB bucket (`BlockCommitments`). Let's not add them here for now. It might make sense to fold them in to `core.Header` in a separate PR with a migration.

omerfirmak:
Closing this one as well, so I can request a review from @rianhughes

PR #1708, opened 2024-02-16
update blockifier

rianhughes:
closes https://github.com/NethermindEth/juno/issues/1707

Upgrade the compiler and blockifier dependencies (cairo 2.6.0 / blockifier 0.5.0)

Note: starknet_api also needed to be upgraded to work with the updated cairo-lang-X /blockifier deps


rianhughes:
Note: there's a still a few `todos`, but it would be useful to get some input on them. Eg should we add new parameters to the Call/Execute functions?

vm/rust/src/lib.rs
```diff
@@ -193,28 +251,37 @@ pub extern "C" fn cairoVMExecute(
...
-            strk_l1_gas_price: felt_to_u128(gas_price_strk_felt),
-        },
-        None
-    );
-    let mut state = CachedState::new(reader, GlobalContractCache::default());
+
+    let fee_token_addresses: FeeTokenAddresses = FeeTokenAddresses { // both addresses are the same for all networks
+        eth_fee_token_address: ContractAddress::try_from(StarkHash::try_from("0x049d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7").unwrap()).unwrap(),
+        strk_fee_token_address: ContractAddress::try_from(StarkHash::try_from("0x04718f5a0fc34cc1af16a1cdee98ffb20c31f5cd61d6ab07201858f4287c938d").unwrap()).unwrap(),
+    };
```
omerfirmak, line 261: Let's define this as a constant.

vm/rust/src/lib.rs
```diff
@@ -233,6 +300,31 @@ pub extern "C" fn cairoVMExecute(
...
+                );
+                if let Err(e) = maybe_class_info {
+                    report_error(reader_handle, e.to_string().as_str(), txn_index as i64);
+                    return;
+                }
+                Some(maybe_class_info.unwrap())
+            }
+            _ => None,
+        };
+
```
omerfirmak, line 327: we can move this to the `match` above

vm/rust/src/lib.rs
```diff
@@ -55,7 +53,19 @@ extern "C" {
...
+
+use lazy_static::lazy_static;
+extern crate serde_json;
+use std::sync::Mutex;
+const VERSIONED_CONSTANTS_JSON: &'static str = include_str!("../versioned_constants.json");
+lazy_static! {
+    static ref VERSIONED_CONSTANTS: Mutex<Result<VersionedConstants, serde_json::Error>> = Mutex::new(serde_json::from_str(VERSIONED_CONSTANTS_JSON));
+}
+
```
omerfirmak, line 67: version constants are supposed to be read from the respective JSON according to the SN version that we are currently executing on.

for example; [versioned_constants_13_0.json](https://github.com/starkware-libs/blockifier/blob/main/crates/blockifier/resources/versioned_constants_13_0.json) should be used if the SN version of the block that we are executing on is 0.13.0

vm/vm.go
```diff
@@ -188,37 +196,29 @@ func (v *vm) Execute(txns []core.Transaction, declaredClasses []core.Class, bloc
...
+	gasPriceWEIBytes := getBytesOrSetDefault(gasPriceWEI)
+	gasPriceSTRKBytes := getBytesOrSetDefault(gasPriceSTRK)
+	daGasPriceWEIBytes := getBytesOrSetDefault(daGasPriceWEI)
+	daGasPriceFRIBytes := getBytesOrSetDefault(daGasPriceFRI)
 
-	var skipChargeFeeByte byte
-	if skipChargeFee {
-		skipChargeFeeByte = 1
+	if gasPriceWEI.IsZero() || gasPriceSTRK.IsZero() || daGasPriceWEI.IsZero() || daGasPriceFRI.IsZero() {
+		return nil, nil, errors.New("gas prices must be non-zero")
```
omerfirmak, line 209: Since these are not user provided, it is fine to panic and crash in Rust if these are zero.

vm/vm.go
```diff
@@ -188,37 +196,29 @@ func (v *vm) Execute(txns []core.Transaction, declaredClasses []core.Class, bloc
...
-	}
-	var legacyTraceJSONByte byte
-	if legacyTraceJSON {
-		legacyTraceJSONByte = 1
-	}
+	skipChargeFeeByte := boolToByte(skipChargeFee)
+	skipValidateByte := boolToByte(skipValidate)
+	errOnRevertByte := boolToByte(errOnRevert)
+	legacyTraceJSONByte := boolToByte(legacyTraceJSON)
+	useKzgDAByte := boolToByte(useKzgDA)
```
omerfirmak, line 216: we don't need these to be separate variables anymore, we can inline them in the call to `cairoVMExecute` below.

omerfirmak:
https://github.com/NethermindEth/juno/pull/1708/files#diff-8d744605e0447372940512f1f0ce8f36115ad8adf80941e0eb201bb7697d76b8R417

We need to pass "L1 data gas consumed" to go side as well.

vm/rust/src/lib.rs
rianhughes: Note for myself, [versioned_constants.json](https://github.com/starkware-libs/blockifier/blob/main/crates/blockifier/resources/versioned_constants.json#L458) and [versioned_constants_13_0.json](https://github.com/starkware-libs/blockifier/blob/main/crates/blockifier/resources/versioned_constants_13_0.json#L383) have different formats :/

node/throttled_vm.go
```diff
@@ -20,27 +20,30 @@ func NewThrottledVM(res vm.VM, concurrenyBudget uint, maxQueueLen int32) *Thrott
 }
 
 func (tvm *ThrottledVM) Call(contractAddr, classHash, selector *felt.Felt, calldata []felt.Felt, blockNumber,
-	blockTimestamp uint64, state core.StateReader, network *utils.Network, maxSteps uint64,
+	blockTimestamp uint64, blockVersion string, blockHash *felt.Felt, state core.StateReader,
+	network *utils.Network,
```
omerfirmak, line 24: `maxSteps` was unintentionally removed I believe.

vm/vm.go
```diff
@@ -188,49 +208,36 @@ func (v *vm) Execute(txns []core.Transaction, declaredClasses []core.Class, bloc
...
 		chainID,
 		(*C.char)(unsafe.Pointer(&sequencerAddressBytes[0])),
 		paidFeesOnL1CStr,
 		C.uchar(skipChargeFeeByte),
 		C.uchar(skipValidateByte),
 		C.uchar(errOnRevertByte),
-		(*C.char)(unsafe.Pointer(&gasPriceWEIBytes[0])),
-		(*C.char)(unsafe.Pointer(&gasPriceSTRKBytes[0])),
+		feltToCChar(gasPriceWEI),
+		feltToCChar(gasPriceSTRK),
```
omerfirmak, line 236: why this change?

vm/vm.go
```diff
@@ -258,6 +265,21 @@ func (v *vm) Execute(txns []core.Transaction, declaredClasses []core.Class, bloc
 	return context.actualFees, traces, nil
 }
 
+func feltToCChar(value *felt.Felt) *C.char {
+	if value == nil {
+		value = new(felt.Felt).SetUint64(1)
+	}
+	bytes := value.Bytes()
+	return (*C.char)(unsafe.Pointer(&bytes[0]))
```
omerfirmak, line 273: this won't work, `bytes`will go out of scope as this function returns, and the pointer you are returning will point to garbage.

vm/rust/src/juno_state_reader.rs
```diff
@@ -184,6 +184,26 @@ pub fn ptr_to_felt(bytes: *const c_uchar) -> StarkFelt {
...
 
+// Note: we only consider the last 16 bytes because `GasPrices` are `NonZeroU128`
+pub fn felt_ptr_to_u128(bytes: *const c_uchar) -> u128 {
+    assert!(!bytes.is_null(), "Null pointer provided");
+    let slice = unsafe { slice::from_raw_parts(bytes, 32) };
+    let array: [u8; 16] = slice[16..32].try_into().expect("Slice with incorrect length");
+    u128::from_be_bytes(array)
+}
+
+use std::ffi::CString;
```
omerfirmak, line 195: let's keep imports at the beginning of the file

vm/rust/src/lib.rs
```diff
@@ -55,22 +56,64 @@ extern "C" {
...
+            strk_fee_token_address,
+        }
+    };
+}
+
+lazy_static! {
+    static ref VERSIONED_CONSTANTS: HashMap<String, VersionedConstants> = {
+        let mut m = HashMap::new();
+        m.insert("0.13.0".to_string(), serde_json::from_str(include_str!("../versioned_constants_13_0.json")).unwrap());
+        m.insert("".to_string(), serde_json::from_str(include_str!("../versioned_constants_13_0.json")).unwrap()); // use this by default
```
omerfirmak, line 83: why are they the same file?

rpc/handlers.go
```diff
@@ -1503,9 +1503,9 @@ func (h *Handler) simulateTransactions(id BlockID, transactions []BroadcastedTra
 	if sequencerAddress == nil {
 		sequencerAddress = h.bcReader.Network().BlockHashMetaInfo.FallBackSequencerAddress
 	}
-	overallFees, traces, err := h.vm.Execute(txns, classes, header.Number, header.Timestamp, sequencerAddress,
-		state, h.bcReader.Network(), paidFeesOnL1, skipFeeCharge, skipValidate, errOnRevert, header.GasPrice,
-		header.GasPriceSTRK, legacyTraceJSON)
+	overallFees, traces, err := h.vm.Execute(txns, classes, header.Number, header.Timestamp, header.ProtocolVersion,
+		header.Hash, sequencerAddress, state, h.bcReader.Network(), paidFeesOnL1, skipFeeCharge, skipValidate,
```
omerfirmak, line 1507: blockHash that we are supposed to use here is of a previous block(-10th block), not the block we are currently executing on. So if we are executing on block #10, we should pass the block hash of #0

node/throttled_vm.go
rianhughes: Nope, it was intentional. max steps is now in the versioned_constants file (invoke_tx_max_n_steps, validate_max_n_steps)

omerfirmak: We still need to be able to override that configuration tho.

vm/rust/src/lib.rs
rianhughes: https://github.com/eqlabs/pathfinder/blob/8a983683e1b01cc579d5339012ef23c86b49f0fc/crates/executor/src/execution_state.rs#L50C21-L50C39

node/throttled_vm.go
rianhughes: Cool, I'll re-add it then

vm/rust/src/lib.rs
rianhughes: I updated the code to use https://github.com/eqlabs/pathfinder/blob/main/crates/executor/src/execution_state.rs#L24

node/throttled_vm.go
rianhughes: max_steps overrides invoke_tx_max_n_steps

rianhughes:
[0d6ae1f](https://github.com/NethermindEth/juno/pull/1708/commits/0d6ae1f381d26b4f8641d1c6fb36bd81dd301c18) breaks TestV1Call for some reason? 

Failed to read from state: error parsing JSON string for class hash 0x01338d85d3e579f6944ba06c005238d145920afeb32f94e3a1e234d21e1e9292.

omerfirmak:
Closing in favor of #1728

PR #1713, opened 2024-02-19
Optimize reading felts from state

omerfirmak:
by eliminating a C allocation and a callback to Go from Rust


vm/state_reader.go
```diff
@@ -90,3 +90,8 @@ func JunoStateGetCompiledClass(readerHandle C.uintptr_t, classHash unsafe.Pointe
 
 	return unsafe.Pointer(cstring(compiledClass))
 }
+
+func fillBufferWithFelt(val *felt.Felt, buffer unsafe.Pointer) int {
```
kirugan, line 94: nit: replace `int` with `C.int` ?

onuruci: Since we are converting the result to C.int everwhere we call it, I guess it looks better. I updated it

kirugan (approved):

rianhughes (approved):

PR #1720, opened 2024-02-24
Add bytecode segments to compiled classes

omerfirmak:


core/class_test.go
```diff
@@ -84,12 +84,37 @@ func TestClassV1Hash(t *testing.T) {
...
-	got := class.(*core.Cairo1Class).Compiled.Hash()
-	assert.Equal(t, "0x18f95714044fd5408d3bf812bcd249ddec098ab3cd201b7916170cfbfa59e05", got.String())
+	for _, tt := range tests {
+		t.Run("ClassHash", func(t *testing.T) {
+			hash := utils.HexToFelt(t, tt.classHash)
+			class, err := gw.Class(context.Background(), hash)
+			require.NoError(t, err)
+			got := class.(*core.Cairo1Class).Compiled.Hash()
+			require.NoError(t, err)
+			assert.Equal(t, tt.expectedCompiledClassHash, got.String())
```
rianhughes, line 115: It would ne nice if we could add a test to make sure ByteCodeSegments length works well with that the FGW returns, although it might be overkill?

Otherwise PR looks goof to me!

Issue #1414, opened 2023-11-07
Emit events on Remote DB


omerfirmak:
Currently remote db doesnt emit any events that db.EventListener expects.

https://github.com/NethermindEth/juno/blob/82e738efe30b7a748704330a24bd4811d59cb242/db/remote/db.go#L53-L55

danielcdz:
Hello! I would like to help with this one!
I think this issue can take me about 2 to 5 days to complete, exactly the time limit of the OD Hack!
I have over 3 years of experience as a software engineer especially working with Golang, developing microservices and backend applications, I want to apply this knowledge in web3 and blockchain projects, on the other hand, this is my 2nd OD Hack, in the last one I contributed to Go projects for Nethermind `cairo-vm-go`

PriyankMkwna:
I am applying to this issue via [OnlyDust platform](https://app.onlydust.com/p/juno).

### My background and how it can be leveraged
I am a blockchain and DeFi developer . I am having experience of 8 months . I am currently working in a DeFi startup . Also I have developed algorithmic stablecoin that is pegged to INR . My experience in programming language will help me to solve this issue .

PR #1721, opened 2024-02-26
emit events

aminsato:
I implement emitting events on the remote DB that related to the issue #1414


kirugan:
Closing the PR since all neccesery logic was implemented in #2088

PR #1723, opened 2024-02-26
First batch of questions

kirugan:


starknet/rust/src/lib.rs
```diff
@@ -34,6 +35,7 @@ pub extern "C" fn freeCstr(ptr: *mut c_char) {
         if ptr.is_null() {
             return;
         }
+        // is it force destruction to deallocate memory?
         let _ = CString::from_raw(ptr);
```
omerfirmak, line 39: Yeap, exactly. `CString`takes the ownership of the pointer and releases it as it goes out of scope.

vm/rust/src/juno_state_reader.rs
```diff
@@ -41,6 +41,7 @@ struct CachedContractClass {
     pub cached_on_height: u64,
 }
 
+// why Lazy?
```
omerfirmak, line 44: As far as I can remember, `SizedCache::with_size` cannot be evaluated in compile time. For `CLASS_CACHE` to be `static`we need it to be lazily instantiated.

starknet/rust/src/lib.rs
```diff
@@ -1,6 +1,7 @@
 use std::ffi::{c_char, CStr, CString};
 use cairo_lang_starknet::casm_contract_class::CasmContractClass;
 
+// why return type is mutable ?
```
omerfirmak, line 4: No particular reason, probably because `CString::into_raw` returns a mutable pointer. I should've used `CString::as_ptr`

.github/workflows/juno-test.yml
```diff
@@ -29,6 +29,8 @@ jobs:
       - uses: Swatinem/rust-cache@v2
         with:
           workspaces: |
+            # do we need to add starknet/rust here?
+            # and what is there real benefit of using rust-cache?
```
omerfirmak, line 33: Yeah, I think starknet/rust should be added here as well. The benefit is faster CI builds. Juno's rust components rarely get's changed, so caching rust builds made sense back then.

blockchain/encoder_initializer.go
```diff
@@ -10,6 +10,7 @@ import (
 
 var once sync.Once
 
+// seems like we always run this code, any objections to replace it with init() ?
```
omerfirmak, line 13: linter didn't approve using init()

core/state_update.go
```diff
@@ -59,6 +60,7 @@ func (d *StateDiff) Commitment() *felt.Felt {
 		case 1:
 			return false
 		default:
+			// Is it somewhere in the docs or it was in slack conversations?
```
omerfirmak, line 63: a slack conversation, I believe it was @joshklop who posted it.

core/state_update.go
```diff
@@ -51,6 +51,7 @@ func (d *StateDiff) Commitment() *felt.Felt {
 	for addr := range d.ReplacedClasses {
 		deployedReplacedAddresses = append(deployedReplacedAddresses, addr)
 	}
+	// this will insert number_of_deployed_contracts into slice which then will be moved to the beginning by this sorting, right?
```
omerfirmak, line 54: nope `hashOfDeployedContracts.Update` straight up updates the digest that we are calculating here.

db/pebble/db.go
```diff
@@ -31,8 +31,10 @@ func New(path string, cache uint, maxOpenFiles int, logger pebble.Logger) (db.DB
 		cache = minCache
 	}
 	pDB, err := newPebble(path, &pebble.Options{
-		Logger:       logger,
-		Cache:        pebble.NewCache(int64(cache * megabyte)),
+		Logger: logger,
+		// what were the benefits?
+		Cache: pebble.NewCache(int64(cache * megabyte)),
+		// were there any issues with max opened files?
```
omerfirmak, line 37: has performance benefits, @wojciechos was the one who made the benchmarks on how these flags affected the performance.

db/pebble/transaction.go
```diff
@@ -47,6 +47,7 @@ func (t *Transaction) Discard() error {
 // Commit : see db.Transaction.Commit
 func (t *Transaction) Commit() error {
 	start := time.Now()
+	// is it ok that it will count metrics for failed Commit() calls as well?
```
omerfirmak, line 50: That is not ideal, but if we fail to commit a transactio, we have much bigger problems anyways.

encoder/encoder.go
```diff
@@ -62,6 +62,7 @@ func Unmarshal(b []byte, v any) error {
 	return decMode.Unmarshal(b, v)
 }
 
+// why it's in not test package
```
omerfirmak, line 65: I believe test packages cannot be imported and this is a helper function to be used outside of encoder package

grpc/handlers.go
```diff
@@ -15,6 +15,7 @@ import (
 )
 
 type Handler struct {
+	// why do we need this?
```
omerfirmak, line 18: grpc generator requires it.

grpc/handlers_test.go
```diff
@@ -60,7 +60,8 @@ func (m *grpcStreamMock) RecvToClient() (*gen.Pair, error) {
 
 func (m *grpcStreamMock) Close() {
 	close(m.recvToServer) // signal server that we are done
-	<-m.sentFromServer    // wait for server to shutdown
+	// why RecvToClient() is not enough?
```
omerfirmak, line 63: looks like it is

jemalloc/jemalloc.go
```diff
@@ -1,5 +1,6 @@
 package jemalloc
 
+// Does it mean that Jemalloc will be used as default allocator in Rust?
```
omerfirmak, line 3: It is will be used thruout the entire binary. So, yes, both Rust and Go will use jemalloc.

jemalloc/jemalloc.go
```diff
@@ -22,6 +23,8 @@ unsigned long long _get_jemalloc_active() {
 */
 import "C"
 
+// So jemalloc is turned on by default? do we need to specify somewhere that jemalloc library is required
```
omerfirmak, line 26: >So jemalloc is turned on by default?

yes

>do we need to specify somewhere that jemalloc library is required

yeah, it should be added to requirements in README

jemalloc/jemalloc.go
```diff
@@ -22,6 +23,8 @@ unsigned long long _get_jemalloc_active() {
...
+// And could you explain shortly what benefits we found by using jemalloc
```
omerfirmak, line 27: It resulted in much faster CairoVM execution and allow us to monitor Juno's C memory usage.

rpc/sync.go
```diff
@@ -17,6 +17,15 @@ type Sync struct {
 	HighestBlockNumber  *uint64    `json:"highest_block_num,omitempty"`
 }
 
+// Why did we start using uint64 instead of NumHex, I think the latter is closer to the spec (https://github.com/starkware-libs/starknet-specs/blob/a789ccc3432c57777beceaa53a34a7ae2f25fda0/api/starknet_api_openrpc.json#L862C55-L862C65)
```
omerfirmak, line 20: that link seems to be outdated https://github.com/starkware-libs/starknet-specs/blob/master/api/starknet_api_openrpc.json#L1098

starknet/block.go
```diff
@@ -16,6 +16,7 @@ type Block struct {
 	SequencerAddress *felt.Felt            `json:"sequencer_address"`
 	GasPriceSTRK     *felt.Felt            `json:"strk_l1_gas_price"`
 
+	// Can we remove GasPriceLegacy now? and we can keep GasPriceWEI I guess
```
omerfirmak, line 19: We can, after updating the testdata we have under clients/feeder

starknet/compiler.go
```diff
@@ -1,5 +1,7 @@
 package starknet
 
+// why it's in separate package (not in VM), is it because of cyclic dependencies ?
+
```
omerfirmak, line 4: I believe it is mainly because I found it cleaner to have it here instead of VM.

starknet/compiler.go
```diff
@@ -33,6 +35,7 @@ func Compile(sierra *SierraDefinition) (*CompiledClass, error) {
 
 	casmJSONOrErrorCstr := C.compileSierraToCasm(sierraJSONCstr)
 	casmJSONOrError := C.GoString(casmJSONOrErrorCstr)
+	// can we replace it with C.free call?
```
omerfirmak, line 38: I would advise against doing that. That string is allocated on Rust side, so it is safer to have it freed on Rust as well.

utils/throttler.go
```diff
@@ -16,6 +16,14 @@ type Throttler[T any] struct {
...
 }
 
+// Question regarding the whole architecture, why buffered channel with limited number
+// of workers is not used? resource busy can be implemented with
+//
+//	select {
+//	  case <- dataChannel: ...
+//	  default:
+//	     err = ErrResourceBusy
+//	 }
```
omerfirmak, line 26: that might work, but you would need 2 channels if I am not mistaken. One to act as a semaphore on the resource and one to act as a queue if semaphore is not available. Sounds a bit tricky to get it right without any race conditions.

utils/throttler_test.go
```diff
@@ -24,6 +24,7 @@ func TestThrottler(t *testing.T) {
 		return nil
 	}
 	do := func() {
+		// go require.NoError(t, throttledRes.Do(doer)) ?
```
omerfirmak, line 27: that works too

validator/validator.go
```diff
@@ -14,6 +14,7 @@ var (
 	v    *validator.Validate
 )
 
+// do you think it's safe to replace once with func init()?
```
omerfirmak, line 17: same as the other usage of `once` in blockchain package.

vm/rust/Cargo.toml
```diff
@@ -5,6 +5,7 @@ edition = "2021"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
+# Any upcoming updates that you aware of?
```
omerfirmak, line 8: Not really, I don't keep track of dependency updates apart from starknet related crates.

db/sync_transaction.go
```diff
@@ -5,6 +5,7 @@ import (
 	"sync"
 )
 
+// why do we need this abstraction?
```
omerfirmak, line 8: it is fairly complicated, might be easier to explain on a call.

node/node.go
```diff
@@ -115,6 +115,7 @@ func New(cfg *Config, version string) (*Node, error) { //nolint:gocyclo,funlen
 
 	dbIsRemote := cfg.RemoteDB != ""
 	var database db.DB
+	// do we use it in RPC btw?
```
omerfirmak, line 118: yeap, that is the whole point. Being able to serve RPC without a local DB.

rpc/trace.go
```diff
@@ -18,6 +18,7 @@ func adaptBlockTrace(block *BlockWithTxs, blockTrace *starknet.BlockTrace, legac
 	}
 	traces := make([]TracedBlockTransaction, 0, len(blockTrace.Traces))
 	for index := range blockTrace.Traces {
+		// why not index, feederTrace := range above ?
```
omerfirmak, line 21: I belive that would work too, I might have wanted to avoid having traces copied on each iteration.

rpc/trace.go
```diff
@@ -78,6 +79,7 @@ func adaptFunctionInvocation(snFnInvocation *starknet.FunctionInvocation, legacy
 		fnInvocation.Calls = append(fnInvocation.Calls, *adaptFunctionInvocation(&snFnInvocation.InternalCalls[index], legacyJSON))
 	}
 	for index := range snFnInvocation.Events {
+		// same question
```
omerfirmak, line 82: same answer I believe:)

vm/trace.go
```diff
@@ -199,6 +199,8 @@ type ExecuteInvocation struct {
 }
 
 func (e ExecuteInvocation) MarshalJSON() ([]byte, error) {
+	// in this case RevertReason will be ignored, right?
+	// I don't remember exactly but it's either RevertReason or FunctionInvocation data?
```
omerfirmak, line 203: Yeap, it is essentially a union, it is either a revert reason or function invocation.

core/trie/key.go
```diff
@@ -15,10 +15,12 @@ type Key struct {
 }
 
 func NewKey(length uint8, keyBytes []byte) Key {
+	// length should be less than 32 ? otherwise bytesNeeded() return wrong index
```
omerfirmak, line 18: length is in bits, so it can be as high as 251 in our usecase IIRC.

> otherwise bytesNeeded() return wrong index

why is that?

core/trie/key.go
```diff
@@ -15,10 +15,12 @@ type Key struct {
...
 	k := Key{len: length}
 	if len(keyBytes) > len(k.bitset) {
 		panic("bytes does not fit in bitset")
 	}
+	// so bitset will be zero prefixed in case len(keyBytes) < 32 ?
```
omerfirmak, line 23: Yeap, correct.

core/trie/key.go
```diff
@@ -72,9 +74,11 @@ func (k *Key) Equal(other *Key) bool {
 	} else if k == nil || other == nil {
 		return false
 	}
+	// I think we can compare k.bitset directly to other.bitset without creating slice and call to bytes.Equal, wdyt?
```
omerfirmak, line 77: I think you already merged a PR doing that.

core/trie/key.go
```diff
@@ -72,9 +74,11 @@ func (k *Key) Equal(other *Key) bool {
...
 	return k.len == other.len && bytes.Equal(k.bitset[:], other.bitset[:])
 }
 
+// could you please explain what is going on here?
```
omerfirmak, line 81: it tests if a bit at a certain position is set or not.
Assume 0b101 as the key, you Test(0) would return true, Test(1) would return false

migration/bucket_migrator.go
```diff
@@ -15,6 +15,8 @@ type (
 	BucketMigratorKeyFilter func([]byte) (bool, error)
 )
 
+// AFAIK right now bucket migration is not transactional across buckets, any ideas how to improve that?
```
omerfirmak, line 18: BucketMigrator only works on a single bucket, so I am not sure what you mean here exactly.

migration/migration.go
```diff
@@ -30,6 +30,7 @@ type schemaMetadata struct {
 }
 
 type Migration interface {
+	// this parameter is not used, do we still need it?
```
omerfirmak, line 33: I am not sure, might come in handy in the future.

core/trie/key.go
```diff
@@ -87,6 +91,7 @@ func (k *Key) String() string {
 	return fmt.Sprintf("(%d) %s", k.len, hex.EncodeToString(k.bitset[:]))
 }
 
+// also hard to understand
```
omerfirmak, line 94: It deletes the least significant bits of a Key

So if you have 0b1101 as your Key and call DeleteLSB(2) on it, it will become 0b11

blockchain/pending.go
```diff
@@ -38,6 +38,7 @@ func (p *PendingState) ContractNonce(addr *felt.Felt) (*felt.Felt, error) {
 	if nonce, found := p.stateDiff.Nonces[*addr]; found {
 		return nonce, nil
 	} else if _, found = p.stateDiff.DeployedContracts[*addr]; found {
+		// why it's zero value for deployed contract?
```
omerfirmak, line 41: that is the default, when a new contract is deployed it's nonce get's initiliazed to zero.

node/metrics.go
```diff
@@ -61,6 +61,7 @@ func makeDBMetrics() db.EventListener {
 			300000,
 			500000,
 			1000000,
+			// is it analogue of maximum?
```
omerfirmak, line 64: Yeap, I believe so.

core/felt/felt.go
```diff
@@ -56,6 +56,7 @@ func (z *Felt) UnmarshalJSON(data []byte) error {
 	if len(s) > fp.Bits*3 {
 		return errors.New("value too large (max = Element.Bits * 3)")
 	}
+	// suggestion to create string only when we pass it to .SetString(...)
```
omerfirmak, line 59: Makes sense

core/trie/node_test.go
```diff
@@ -22,6 +22,7 @@ func TestNodeHash(t *testing.T) {
 	node := trie.Node{
 		Value: new(felt.Felt).SetBytes(valueBytes),
 	}
+	// it's 6 because 2^5 is 32 which is not sufficient to store 42 ?
```
omerfirmak, line 25: I believe 5 bits are enough to store 42 as well, this was just a testcase I took from pathfinder repository.

rpc/handlers_test.go
```diff
@@ -1770,6 +1770,7 @@ func TestStateUpdate(t *testing.T) {
 	client := feeder.NewTestClient(t, &utils.Mainnet)
 	mainnetGw := adaptfeeder.New(client)
 
+	// were there something special in 21656 block on mainnet or it could be any blockNumber
```
omerfirmak, line 1773: it probably is special because it has all fields populated.

rpc/block.go
```diff
@@ -17,6 +17,7 @@ const (
 	BlockRejected
 )
 
+// any objections to use MarshalText() instead ?
```
omerfirmak, line 20: if it works fine, no objections

vm/transaction.go
```diff
@@ -122,6 +124,7 @@ type ResourceBounds struct {
 }
 
 func adaptTransaction(txn core.Transaction) *Transaction {
+	// can we safely replace .Uint64() == 3 with .Is(3) call ?
```
omerfirmak, line 127: I believe we can

vm/transaction.go
```diff
@@ -37,6 +38,7 @@ func marshalTxn(txn core.Transaction) (json.RawMessage, error) {
 		txnAndQueryBit.Txn["Declare"] = map[string]any{
 			"V" + t.Version.Text(felt.Base10): t,
 		}
+	// why stored value in Txn is different for these two ?
```
omerfirmak, line 41: That is due to how the object on Rust side expects the JSON to be formatted. take a look at Rust enums.

vm/transaction.go
```diff
@@ -17,6 +17,7 @@ func marshalTxn(txn core.Transaction) (json.RawMessage, error) {
 	t := adaptTransaction(txn)
 	version := (*core.TransactionVersion)(t.Version)
 	txnAndQueryBit := struct {
+		// what is query bit?
```
omerfirmak, line 20: it is for txns sent to a full node for fee estimates and stuff
So, users still sign the txn but with a query bit
Which invalidates the txn in sequencing context
So someone can't just use a txn that user signed for a RPC request.

rpc/transaction.go
```diff
@@ -218,6 +218,8 @@ type ResourceBounds struct {
 
 // https://github.com/starkware-libs/starknet-specs/blob/a789ccc3432c57777beceaa53a34a7ae2f25fda0/api/starknet_api_openrpc.json#L1252
 //
+// Sorry, I think I already asked you why there are *[]*felt.Felt instead of []*felf.Felt, but I can't remember
```
omerfirmak, line 221: because with []*felf.Felt, there is no way to completely omit the field from JSON output. It always gets marshalled to atleast an empty list. *[]*felt.Felt can be entirely omitted if it is nil.

starknet/transaction.go
```diff
@@ -160,6 +160,7 @@ type ResourceBounds struct {
 }
 
 // Transaction object returned by the feeder in JSON format for multiple endpoints
+// can we have just map[Resource]ResourceBounds instead of *map
```
omerfirmak, line 163: same as *[]*felt.Felt and []felt.Felt

starknetdata/feeder/feeder.go
```diff
@@ -61,6 +61,7 @@ func (f *Feeder) block(ctx context.Context, blockID string) (*core.Block, error)
 
 	var sig *starknet.Signature
 	if blockID != pendingID {
+		// btw it should work for all networks, right?
```
omerfirmak, line 64: Yeap, I believe it does.

PR #1726, opened 2024-02-26
Fix resource bound nil pointer error during p2p sync

IronGauntlets:


adapters/core2p2p/transaction.go
```diff
@@ -97,10 +96,9 @@ func AdaptTransaction(transaction core.Transaction) *spec.Transaction {
 					L1Gas:             adaptResourceLimits(tx.ResourceBounds[core.ResourceL1Gas]),
 					L2Gas:             adaptResourceLimits(tx.ResourceBounds[core.ResourceL2Gas]),
 					Tip:               AdaptFelt(new(felt.Felt).SetUint64(tx.Tip)),
-					// todo fill rest of V3 fields:
-					Paymaster:   nil,
-					NonceDomain: "",
-					FeeDomain:   "",
+					Paymaster:         nil,
```
kirugan, line 99: Paymaster is some kind of obsolete field?

IronGauntlets: I am not sure, since it was both in core and p2p spec but their types were different I left it as nil.

adapters/p2p2core/transaction.go
```diff
@@ -71,13 +71,34 @@ func AdaptTransaction(t *spec.Transaction, network *utils.Network) core.Transact
...
+
+		nDAMode, err := strconv.ParseUint(tx.GetNonceDomain(), 10, 32)
+		if err != nil {
+			panic(fmt.Sprintf("Failed to convert Nonce DA mode: %v to uint32", tx.GetNonceDomain()))
+		}
+
+		fDAMode, err := strconv.ParseUint(tx.GetFeeDomain(), 10, 32)
+		if err != nil {
+			panic(fmt.Sprintf("Failed to convert Fee DA mode: %v to uint32", tx.GetFeeDomain()))
+		}
```
kirugan, line 83: I think we shouldn't use panic here because a malicious peer can send any information in these fields, moreover since they are of string type it should be pretty easy to trigger error

kirugan (approved):

adapters/p2p2core/transaction.go
IronGauntlets: I decided to add those panic because I didn't want `Adapt` to return an error. Also, since some of the code is deprecated I didn't want to spend more time trying to do something proper. 

For now only Juno is going to be sending those messages.

PR #1730, opened 2024-02-29
Add new starknet_getBlockWithReceipts method for v0.7

kirugan:


rpc/handlers.go
```diff
@@ -627,6 +661,12 @@ func (h *Handler) TransactionReceiptByHash(hash felt.Felt) (*TransactionReceipt,
 		return nil, ErrTxnHashNotFound
 	}
 
+	return h.transactionReceipt(receipt, txn, blockHash, blockNumber)
+}
+
+func (h *Handler) transactionReceipt(receipt *core.TransactionReceipt, txn core.Transaction,
+	blockHash *felt.Felt, blockNumber uint64,
+) (*TransactionReceipt, *jsonrpc.Error) {
```
omerfirmak, line 669: You can pass in the block status here as a parameter as well, so that when this is called in a loop it doesn't end up making multiple `h.l1head` calls unnecessarily to figure out what the txn status is.

rpc/handlers.go
```diff
@@ -308,19 +317,44 @@ func (h *Handler) BlockWithTxs(id BlockID) (*BlockWithTxs, *jsonrpc.Error) {
...
+	if rpcErr != nil {
+		return nil, rpcErr
 	}
 
-	return &BlockWithTxs{
+	txs := make([]TransactionWithReceipt, len(block.Transactions))
+	for index, txn := range block.Transactions {
+		r := block.Receipts[index]
+		receipt, rpcErr := h.transactionReceipt(r, txn, block.Hash, block.Number)
+		if rpcErr != nil {
```
omerfirmak, line 342: you can pass `nil` as block hash here. ``lock_hash` an` `block_number` fields are optional in `BlockWithReceipt`

jsonrpc/server.go
```diff
@@ -71,7 +71,7 @@ func Err(code int, data any) *Error {
 	case InvalidParams:
 		return &Error{Code: InvalidParams, Message: "Invalid Params", Data: data}
 	default:
-		return &Error{Code: InternalError, Message: "Internal Error", Data: data}
+		return &Error{Code: InternalError, Message: "Internal error", Data: data}
```
kirugan, line 74: "error" it the [spec](https://www.jsonrpc.org/specification#:~:text=%2D32603-,Internal%20error,-Internal%20JSON%2DRPC) + `rpc.InternalError` message already lowercased

rpc/handlers.go
```diff
@@ -343,6 +379,70 @@ func (h *Handler) LegacyBlockWithTxs(id BlockID) (*BlockWithTxs, *jsonrpc.Error)
 	return block, nil
 }
 
+func AdaptReceipt(receipt *core.TransactionReceipt, txn core.Transaction,
```
rianhughes, line 382: What do you think about passing in an rpc.Transaction instead of a core.Transaction? It would save us calling AdaptTransaction inside the function again, and we need to call it outside the function anyway

rpc/handlers.go
```diff
@@ -343,6 +379,70 @@ func (h *Handler) LegacyBlockWithTxs(id BlockID) (*BlockWithTxs, *jsonrpc.Error)
...
+	switch v := txn.(type) {
+	case *core.DeployTransaction:
+		contractAddress = v.ContractAddress
+	case *core.DeployAccountTransaction:
+		contractAddress = v.ContractAddress
+	case *core.L1HandlerTransaction:
+		messageHash = "0x" + hex.EncodeToString(v.MessageHash())
+	}
+
+	var receiptBlockNumber *uint64
```
rianhughes, line 414: Q: what do you think about leaving a comment here stating pending blocks don't have block hashes or block numbers?

rpc/handlers.go
```diff
@@ -343,6 +379,70 @@ func (h *Handler) LegacyBlockWithTxs(id BlockID) (*BlockWithTxs, *jsonrpc.Error)
...
+	}
+
+	return &TransactionReceipt{
+		FinalityStatus:  finalityStatus,
+		ExecutionStatus: es,
+		Type:            AdaptTransaction(txn).Type,
+		Hash:            txn.Hash(),
+		ActualFee: &FeePayment{
+			Amount: receipt.Fee,
+			Unit:   feeUnit(txn),
```
rianhughes, line 433: Q: why can't we take `receipt.FeeUnit`?

rianhughes: Ah right, they are different types

rpc/handlers.go
```diff
@@ -1957,22 +2007,22 @@ func (h *Handler) LegacyMethods() ([]jsonrpc.Method, string) { //nolint: funlen
 		{
 			Name:    "starknet_getBlockWithTxHashes",
 			Params:  []jsonrpc.Parameter{{Name: "block_id"}},
-			Handler: h.LegacyBlockWithTxHashes,
```
rianhughes, line 1960: We're not using Legacy methods anymore, right? Should we remove them?

rpc/handlers_test.go
```diff
@@ -1494,6 +1494,123 @@ func TestTransactionReceiptByHash(t *testing.T) {
...
+		header := resp.BlockHeader
+
+		var transactions []rpc.TransactionWithReceipt
+		for i, tx := range block1.Transactions {
+			receipt := block1.Receipts[i]
+			adaptedTx := rpc.AdaptTransaction(tx)
+
+			transactions = append(transactions, rpc.TransactionWithReceipt{
+				Transaction: adaptedTx,
+				Receipt:     rpc.AdaptReceipt(receipt, tx, rpc.TxnAcceptedOnL1, nil, 0),
```
rianhughes, line 1592: Why don't we pass in block hash and block number here?

rpc/handlers.go
kirugan: Yep, I'll remove them in upcoming PR. I'm a little bit worried that there are a bunch of Legacy methods already :D

kirugan: Want to double check that removing them doesn't affect anything

rpc/handlers.go
kirugan: Good point

rpc/handlers.go
```diff
@@ -327,6 +329,40 @@ func (h *Handler) BlockWithTxs(id BlockID) (*BlockWithTxs, *jsonrpc.Error) {
 	}, nil
 }
 
+func (h *Handler) BlockWithReceipts(id BlockID) (*BlockWithReceipts, *jsonrpc.Error) {
```
rianhughes, line 332: I think we need to return l1_gas and l1_data_gas in Execution resources ([spec](https://github.com/starkware-libs/starknet-specs/blob/master/api/starknet_api_openrpc.json#L4019) vs [juno](https://github.com/NethermindEth/juno/blob/main/rpc/transaction.go#L302)).

rpc/handlers.go
kirugan: Added comment

rpc/handlers_test.go
```diff
@@ -1494,6 +1494,123 @@ func TestTransactionReceiptByHash(t *testing.T) {
...
+				Timestamp:        header.Timestamp,
+				SequencerAddress: header.SequencerAddress,
+				L1GasPrice:       header.L1GasPrice,
+				StarknetVersion:  header.StarknetVersion,
+			},
+			Transactions: transactions,
+		}, resp)
+	})
+
+	t.Run("accepted L1 block", func(t *testing.T) {
```
rianhughes, line 1571: I don't think we're returning execution resources correctly. For example block1 contains [builtin_instance_counter](https://github.com/NethermindEth/juno/blob/main/clients/feeder/testdata/mainnet/block/1.json#L128) which we don't seem to be returning

rpc/handlers.go
kirugan: well, there are a few problems that I faced trying to do that now:
* `feeUnit` already used in a place where only core.Transaction is used (there is no need to call AdaptTransaction on caller side)
* `v.MessageHash()` for L1Handler type not implemented on rpc.Transaction type 

I'll add a comment to try restruct passed data

rpc/handlers_test.go
rianhughes: At least when I print our output, I only get

```
"execution_resources": {
                    "steps": 29
                }
```

rpc/handlers_test.go
rianhughes: Actually, looking at the spec again I don't think we need to pass block hash or block number at all.

rianhughes: In the receipts that is

rianhughes: Ah nvm, we do need it for TransactionReceiptByHash. Yup, you're right!

rpc/handlers.go
```diff
@@ -1528,10 +1588,20 @@ func (h *Handler) simulateTransactions(id BlockID, transactions []BroadcastedTra
...
+			dataGasPrice = header.L1DataGasPrice.PriceInWei
+		}
+
 		estimate := FeeEstimate{
-			GasConsumed: new(felt.Felt).Div(overallFee, gasPrice),
-			GasPrice:    gasPrice,
-			OverallFee:  overallFee,
+			GasConsumed:     new(felt.Felt).Div(overallFee, gasPrice),
+			GasPrice:        gasPrice,
+			DataGasConsumed: new(felt.Felt).Div(overallFee, dataGasPrice), // please double check that
```
kirugan, line 1602: 🙏

rianhughes: overall_fee will now be gas_consumed × gas_price + data_gas_consumed × data_gas_price

rpc/handlers.go
kirugan: I think it's done in recent commit

vm/rust/src/lib.rs
```diff
@@ -287,6 +288,8 @@ pub extern "C" fn cairoVMExecute(
                 }
 
                 let actual_fee = t.actual_fee.0.into();
+                let data_gas_consumed = t.da_gas.l1_data_gas.into();
```
rianhughes, line 291: Why not just return the entre gas vector?

rianhughes: I guess we can calculate the gas_consumed from the other fields

rpc/transaction.go
```diff
@@ -337,25 +347,25 @@ func (r *ExecutionResources) MarshalJSON() ([]byte, error) {
...
+	ActualFee          *FeePayment           `json:"actual_fee"`
+	ExecutionStatus    TxnExecutionStatus    `json:"execution_status"`
+	FinalityStatus     TxnFinalityStatus     `json:"finality_status"`
+	BlockHash          *felt.Felt            `json:"block_hash,omitempty"`
+	BlockNumber        *uint64               `json:"block_number,omitempty"`
+	MessagesSent       []*MsgToL1            `json:"messages_sent"`
+	Events             []*Event              `json:"events"`
+	ContractAddress    *felt.Felt            `json:"contract_address,omitempty"`
+	RevertReason       string                `json:"revert_reason,omitempty"`
+	ExecutionResources *ComputationResources `json:"execution_resources,omitempty"`
```
rianhughes, line 367: Is this not [Execution resources](https://github.com/starkware-libs/starknet-specs/blob/v0.7.0-rc1/api/starknet_api_openrpc.json#L2889)?

rianhughes: I think it's[ function invocation](https://github.com/starkware-libs/starknet-specs/blob/v0.7.0-rc1/api/starknet_trace_api_openrpc.json#L374) that uses computation resources for the "execution resources" tag

rpc/handlers.go
```diff
@@ -709,11 +774,11 @@ func (h *Handler) LegacyTransactionReceiptByHash(hash felt.Felt) (*TransactionRe
 	return receipt, nil
 }
 
-func adaptExecutionResources(resources *core.ExecutionResources) *ExecutionResources {
+func adaptExecutionResources(resources *core.ExecutionResources) *ComputationResources {
```
rianhughes, line 777: nit: we return ComputationResources here in an adapter function

rpc/handlers.go
rianhughes: Recent changes look good

rpc/handlers_test.go
rianhughes: The we still aren't returning all the data in Execution_resources for some reason. We are also not returning [data_availability](https://github.com/starkware-libs/starknet-specs/blob/v0.7.0-rc1/api/starknet_api_openrpc.json#L4037) from what I can see

rianhughes: I'm inserting the following at the end of the "accepted L1 block" test, and comparing it against that's in the block by hand

```
qwe, _ := json.MarshalIndent(transactions, "", "")
fmt.Println(string(qwe))
panic(1)
```

rpc/handlers_test.go
```diff
@@ -1494,6 +1494,123 @@ func TestTransactionReceiptByHash(t *testing.T) {
...
+
+		blockID := rpc.BlockID{Pending: true}
+
+		mockReader.EXPECT().Pending().Return(blockchain.Pending{Block: block0}, nil)
+		mockReader.EXPECT().L1Head().Return(&core.L1Head{}, nil)
+
+		resp, rpcErr := handler.BlockWithReceipts(blockID)
+		header := resp.BlockHeader
+
+		var transactions []rpc.TransactionWithReceipt
```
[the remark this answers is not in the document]
kirugan: renamed

rpc/handlers.go
```diff
@@ -327,6 +347,40 @@ func (h *Handler) BlockWithTxs(id BlockID) (*BlockWithTxs, *jsonrpc.Error) {
...
+	if rpcErr != nil {
+		return nil, rpcErr
+	}
+
+	finalityStatus := TxnAcceptedOnL2
+	if blockStatus == BlockAcceptedL1 {
+		finalityStatus = TxnAcceptedOnL1
+	}
+
+	txs := make([]TransactionWithReceipt, len(block.Transactions))
```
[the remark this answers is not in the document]
kirugan: renamed

PR #1735, opened 2024-03-06
Replace MarshalJSON() with MarshalText() in most cases

kirugan:


rpc/transaction.go
```diff
@@ -179,21 +179,17 @@ const (
 	ResourceL2Gas
 )
 
-func (r Resource) MarshalJSON() ([]byte, error) {
```
kirugan, line 182: this is interesting case: `MarshalJSON()` returns incorrect values (they are not strings on byte level), but somehow it works (because of MarshalText() below I presume)

rianhughes (approved):

IronGauntlets (approved):

PR #1737, opened 2024-03-08
Fix skipped tests in handlers

kirugan:


kirugan:
Decided to do it from scratch here - https://github.com/NethermindEth/juno/pull/1829

PR #1740, opened 2024-03-12
Fix estimateFee for v0.6

kirugan:


rpc/handlers.go
```diff
@@ -1348,7 +1348,7 @@ func (h *Handler) Call(call FunctionCall, id BlockID) ([]*felt.Felt, *jsonrpc.Er
 	}, &vm.BlockInfo{
 		Header:                header,
 		BlockHashToBeRevealed: blockHashToBeRevealed,
-	}, state, h.bcReader.Network(), h.callMaxSteps)
+	}, state, h.bcReader.Network(), h.callMaxSteps, true)
```
IronGauntlets, line 1351: I think for v0_6 this should be false, right?

kirugan: Yep, fixed it

IronGauntlets: does is make sense to have a `call` private func to reduce the duplication and pass v0_6Response to it?

rpc/handlers.go
```diff
@@ -1482,11 +1522,35 @@ func (h *Handler) EstimateMessageFee(msg MsgFromL1, id BlockID) (*FeeEstimate, *
...
+		// Needed to marshal to blockifier type.
+		// Must be greater than zero to successfully execute transaction.
+		PaidFeeOnL1: new(felt.Felt).SetUint64(1),
+	}
+	estimates, rpcErr := h.EstimateFeeV0_6([]BroadcastedTransaction{tx}, nil, id)
+	if rpcErr != nil {
+		if rpcErr.Code == ErrTransactionExecutionError.Code {
+			data := rpcErr.Data.(TransactionExecutionErrorData)
+			return nil, makeContractError(errors.New(data.ExecutionError))
+		}
```
IronGauntlets, line 1549: I think this duplication can be removed.

IronGauntlets (approved):
Apart from the duplication of code in `Call` and `EstimateMessageFee`, LGTM

rpc/handlers.go
kirugan: yep, changed it

rpc/handlers.go
kirugan: Refactored

PR #1743, opened 2024-03-13
Fix issues found by golangci-linter v1.56.2

kirugan:
+ small optimization of `Felt.UnmarshalJSON()`

### DIFF
--- a/core/felt/felt.go
+++ b/core/felt/felt.go
@@ -52,20 +52,19 @@ func (z *Felt) Impl() *fp.Element {
 //
 // [gnark-crypto]: https://github.com/ConsenSys/gnark-crypto/blob/master/ecc/stark-curve/fp/element.go
 func (z *Felt) UnmarshalJSON(data []byte) error {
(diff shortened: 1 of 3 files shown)


core/felt/felt.go
```diff
@@ -52,20 +52,19 @@ func (z *Felt) Impl() *fp.Element {
...
 	}
 
 	// we accept numbers and strings, remove leading and trailing quotes if any
-	if len(s) > 0 && s[0] == '"' {
-		s = s[1:]
+	if len(data) > 0 && data[0] == '"' {
+		data = data[1:]
 	}
-	if len(s) > 0 && s[len(s)-1] == '"' {
-		s = s[:len(s)-1]
```
kirugan, line 65: Reslicing a string causes creation of new string (unless compiler is super smart). So the optimisation is really simple: just defer string construction till the very end.

IronGauntlets (approved):
Lgtm

PR #1756, opened 2024-03-20
Pass RPC errorCode to prometheus metrics

kirugan:


IronGauntlets (approved):

kirugan:
Right now we have following RPC errors:
```
InvalidJSON    = -32700 // Invalid JSON was received by the server.
InvalidRequest = -32600 // The JSON sent is not a valid Request object.
MethodNotFound = -32601 // The method does not exist / is not available.
InvalidParams  = -32602 // Invalid method parameter(s).
InternalError  = -32603 // Internal JSON-RPC error.
```
In which only `InternalError` is the most usual one, right now it's not clear how can we distinguish between different error causes for same internal error. So I'm closing the PR for now.

PR #1760, opened 2024-03-21
Pass max_steps to blockifier code

### LINKED ISSUE #1759: rpc-call-max-steps flag isn't limiting RPC call steps
**Steps to Reproduce:**
1. Set `rpc-call-max-steps` to 0.
2. Run calls that should exceed these limits. E.g(mainnet):
```{
    "jsonrpc": "2.0",
    "method": "starknet_call",
    "params": {
        "block_id": {
            "block_hash": "0x9f59dd27f387b869844285307bd1de75d970f0a69903485e164e3d1609dd42"
        },
        "request": {
            "contract_address": "0x0772FA7347f1D6c16D7d03dEba3a0D2aDDde10c5D9259c784d0d47Fc6D231aBb",
            "entry_point_selector": "0x039dd5dee115f62929cb2e75e8e4c5e964121efc81aabbd2eaf2002005ad536e",
            "calldata": []
        }
    },
    "id": 1
}
```

**Actual Behavior:** calls ignore the `rpc-call-max-steps` setting and proceed without errors.

**Expected Behavior:** Setting `rpc-call-max-steps` to low values should block calls exceeding the cairo step count. Worked fine for <4M steps before(as 4m is max limit set in blockifier), now setting it to 0 or 1 doesn’t impact multi calls - they run as if the limit's not there.

**Notes:** Could be related to the blockifier update?

kirugan:
Fixes #1759

### DIFF
--- a/vm/rust/src/lib.rs
+++ b/vm/rust/src/lib.rs
@@ -74,7 +74,7 @@ pub extern "C" fn cairoVMCall(
     block_info_ptr: *const BlockInfo,
     reader_handle: usize,
     chain_id: *const c_char,
-    _max_steps: c_ulonglong, //todo: enforce this
+    max_steps: c_ulonglong,
 ) {
     let block_info = unsafe { *block_info_ptr };
     let call_info = unsafe { *call_info_ptr };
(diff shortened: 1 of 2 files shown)


wojciechos (approved):

vm/rust/src/lib.rs
```diff
@@ -393,6 +394,11 @@ fn build_block_context(
             hash: BlockHash(StarkFelt::new(block_info.block_hash_to_be_revealed).unwrap()),
         })
     }
+    let mut constants = get_versioned_constants(block_info.version);
+    if let Some(max_steps) = max_steps {
+        constants.invoke_tx_max_n_steps = max_steps as u32;
```
rianhughes, line 399: should we update max_steps to be a c_ulong?

rianhughes: Otherwise looks good!

PR #1770, opened 2024-03-28
Support starknet 0.13.1.1

### LINKED ISSUE #1761: Support SN 0.13.1.1 patch
1. Context
  - In v0.13.1, class declaration fees increased dramatically. Following feedback from the community, we want to solve this before the next scheduled version (planned in Summer).

2. Solution: A mini-version which reduces class declaration fees in the [versioned constants](https://github.com/starkware-libs/blockifier/blob/v0.5.0/crates/blockifier/resources/versioned_constants.json). We may also change max_calldata_length and invoke_tx_max_n_steps, but that’s TBD in upcoming days.
  - :warning: These are execution changes.
  - The steps we have in mind:
    - Mini-version released with its own versioned constants.
    - Full node release which includes the new versioned constants and their handling.

3. Questions
  - Numbering: we prefer v0.13.1.1 over v0.13.2. Any objections?
  - Does anyone who isn’t a full node directly depend on the versioned constants? (Good for us to know in the future.)

kirugan:
Closes #1761


vm/rust/Cargo.toml
```diff
@@ -8,8 +8,8 @@ edition = "2021"
 [dependencies]
 serde = "1.0.171"
 serde_json = { version = "1.0.96", features = ["raw_value"] }
-blockifier = "=0.5.0-rc.3"
-starknet_api = "=0.8.0"
+blockifier = "=0.6.0-rc.0"
```
wojciechos, line 11: Should we bump up cairo-lang-starknet-classes to 2.6.0 [here](https://github.com/NethermindEth/juno/blob/kirugan/update-pebble-1.1.0/starknet/rust/Cargo.toml#L11C1-L11C28)? This would align it with the blockifier version see [here](https://github.com/starkware-libs/blockifier/blob/v0.6.0-rc.0/Cargo.toml#L26).

kirugan: Good point, I don't expect major changes but at least everything will be aligned together. Thanks

wojciechos (approved):

vm/rust/src/lib.rs
```diff
@@ -425,6 +425,8 @@ lazy_static! {
     static ref CONSTANTS: HashMap<String, VersionedConstants> = {
         let mut m = HashMap::new();
         m.insert("0.13.0".to_string(), serde_json::from_slice(include_bytes!("../versioned_constants_13_0.json")).unwrap());
+        m.insert("0.13.1".to_string(), serde_json::from_slice(include_bytes!("../versioned_constants_13_1.json")).unwrap());
```
IronGauntlets, line 428: Should this not be `0.13.1.1`?

vm/rust/versioned_constants_13_1.json
IronGauntlets, line 1: If the blockifier has been updated, do we still need to have this file in our repo? Also, shouldn't the name have been `versioned_constants_13_1`? Or is this coming from Starkware.

IronGauntlets:
Supported starknet version should also be increase to `0.13.1.1`

vm/rust/src/lib.rs
kirugan: I was checking this commit - https://github.com/starkware-libs/blockifier/commit/dfea574bb346ec43de1a4a54e9a44366cecd0a6d#diff-79227bb6cc8b8dd82bfcc791f0a911f73dcfec44d04b56e7dfc940c78311e76c this file (0.13.1) is actually missing right now. Here is how we use it:
```
CONSTANTS.get(&version_str.to_string()).unwrap_or(VersionedConstants::latest_constants()).to_owned()
```
in case version_str file is present we use it's values otherwise we use latest VerionedConstants from blockifier (which is 0.13.1.1 in this PR)

vm/rust/versioned_constants_13_1.json
kirugan: It's taken from commit that I mentioned in prev comment. They used this naming I guess it's correct (emphasizes type of content that resides inside file i.e. json)

kirugan:
> The supported starknet version should also be increased to 0.13.1.1
It won't work `semver.MustParse(...)` won't handle this version because it's not compliant with semver. Technically blocks from 0.13.1.1 are the same as 0.13.1 - there is no change in structure

vm/rust/src/lib.rs
kirugan: I removed this file from the PR

vm/rust/src/lib.rs
```diff
@@ -425,6 +425,7 @@ lazy_static! {
     static ref CONSTANTS: HashMap<String, VersionedConstants> = {
         let mut m = HashMap::new();
         m.insert("0.13.0".to_string(), serde_json::from_slice(include_bytes!("../versioned_constants_13_0.json")).unwrap());
+
         m
```
wojciechos, line 429: seems like unwanted char

IronGauntlets (approved):

PR #1772, opened 2024-03-29
Add removed versioned_constants for 0.13.1

kirugan:


vm/rust/src/lib.rs
```diff
@@ -421,16 +422,60 @@ fn build_block_context(
 }
 
 
+
```
pnowosie, line 425: Oops, 😹 cat jumped on <enter> here ;)

pnowosie (approved):
👍

vm/rust/src/lib.rs
kirugan: Will remove it later, don't want to block release

PR #1773, opened 2024-03-29
Calculate db bucket size

IronGauntlets:


cmd/juno/juno.go
```diff
@@ -381,10 +391,99 @@ func GenP2PKeyPair() *cobra.Command {
...
+			_, err = fmt.Fprintln(cmd.OutOrStdout(), "P2P Public Key:", string(pubHex))
+			if err != nil {
+				return err
+			}
+
+			_, err = fmt.Fprintln(cmd.OutOrStdout(), "P2P PeerID:", id)
+			if err != nil {
+				return err
+			}
+			return nil
```
kirugan, line 403: maybe just `return err` here without if stmt. above?

cmd/juno/juno.go
```diff
@@ -381,10 +391,99 @@ func GenP2PKeyPair() *cobra.Command {
...
+			}
+
+			pebbleDB, err := pebble.New(dbPath)
+			if err != nil {
+				return err
+			}
+
+			var totalSize, stateSizeWithoutHistory, stateSizeWithHistory uint
+
+			_, err = fmt.Fprintln(cmd.OutOrStdout(), "Total number of DB buckets:", uint(db.SchemaIntermediateState)+1)
```
kirugan, line 430: there is a new bucket `db.Peer`, maybe just put additional constants (where all buckets defined) that tells us how many buckets are there ?

kirugan, line 430: there is a new bucket `db.Peer` maybe we should just add variable alongside bucket constants with all of them inside?

cmd/juno/juno.go
```diff
@@ -381,10 +391,99 @@ func GenP2PKeyPair() *cobra.Command {
...
+			if err != nil {
+				return err
+			}
+
+			var bucketSize uint
+			for i := db.StateTrie; i <= db.SchemaIntermediateState; i++ {
```
kirugan, line 436: same thing related to `db.Peer`

cmd/juno/juno.go
```diff
@@ -381,10 +391,99 @@ func GenP2PKeyPair() *cobra.Command {
...
+				}
+
+				_, err = fmt.Fprintln(cmd.OutOrStdout(), uint(i)+1, "Size of", i, "=", bucketSize)
+				if err != nil {
+					return err
+				}
+
+				totalSize += bucketSize
+
+				if i == db.StateTrie || i == db.ContractStorage || i == db.Class || i == db.ContractNonce || i == db.ClassesTrie ||
```
rianhughes, line 449: What do you think about defining a map with these variables, and just checking if i exists in the map?

rianhughes: same for below

db/buckets.go
```diff
@@ -35,7 +35,42 @@ const (
 	Peer // maps peer ID to peer multiaddresses
 )
 
+var bucketStrings = []string{
+	"StateTrie",
```
rianhughes, line 39: why is Peer excluded in bucketStrings?

kirugan: this list was created before `db.Peer` was merged into main

kirugan: Actually I think we shouldn't maintain it but rely on existing tool called `stringer` it can generate string names for a type: https://go.dev/blog/generate 

I can add it with installation command in a separate PR

cmd/juno/juno.go
```diff
@@ -350,7 +355,9 @@ func NewCmd(config *node.Config, run func(*cobra.Command, []string) error) *cobr
 	junoCmd.Flags().Bool(corsEnableF, defaultCorsEnable, corsEnableUsage)
 	junoCmd.Flags().String(versionedConstantsFileF, defaultVersionedConstantsFile, versionedConstantsFileUsage)
 	junoCmd.MarkFlagsMutuallyExclusive(p2pFeederNodeF, p2pPeersF)
+
 	junoCmd.AddCommand(GenP2PKeyPair())
+	junoCmd.AddCommand(DBSize())
```
kirugan, line 360: wdyt about moving non essential commands into separate file or file per command?

cmd/juno/juno.go
kirugan: or we can actually use this tool https://github.com/dmarkham/enumer it will generate string name for every enum value and useful methods such as `BucketValues()[]Bucket` which can be used in methods below like `if utils.AnyOf(i, db.BucketValues()...)`

cmd/juno/juno.go
IronGauntlets: Nice I will do that

db/buckets.go
kirugan: Added here - https://github.com/NethermindEth/juno/pull/1958

cmd/juno/juno.go
IronGauntlets: I have done something like in the utils/slice

```go
func Contains[T comparable](slice []T, e T) bool {
	for _, v := range slice {
		if e == v {
			return true
		}
	}
	return false
}
```

cmd/juno/juno.go
IronGauntlets: Done

cmd/juno/juno.go
kirugan: there is a function with same name in stdlib: https://pkg.go.dev/slices#Contains

kirugan: I proposed AnyOf because the stdlib version lacks a variadic parameter

cmd/juno/juno.go
IronGauntlets: Done

cmd/juno/juno.go
IronGauntlets: Done

cmd/juno/juno.go
IronGauntlets: Done

db/buckets.go
IronGauntlets: Done

cmd/juno/juno.go
IronGauntlets: Done

cmd/juno/juno.go
IronGauntlets: Just saw your comment

IronGauntlets: Changed the Contains code to Anyof

cmd/juno/dbsize.go
```diff
@@ -0,0 +1,88 @@
...
+				}
+
+				_, err = fmt.Fprintln(cmd.OutOrStdout(), uint(b)+1, "Size of", b, "=", bucketSize)
+				if err != nil {
+					return err
+				}
+
+				totalSize += bucketSize
+
+				if utils.AnyOf(db.StateTrie, db.ContractStorage, db.Class, db.ContractNonce,
```
kirugan, line 52: looks like `b` should be first argument, below as well

db/pebble/db.go
```diff
@@ -108,3 +117,32 @@ func (d *DB) Update(fn func(txn db.Transaction) error) error {
...
+		err  error
+		size uint
+		v    []byte
+	)
+
+	const upperBoundofPrefix = 0xff
+	pebbleDB := pDB.Impl().(*pebble.DB)
+	it, err := pebbleDB.NewIter(&pebble.IterOptions{LowerBound: prefix, UpperBound: append(prefix, upperBoundofPrefix)})
+	if err != nil {
+		// No need to call utils.RunAndWrapOnError() size iterator couldn't be created
```
kirugan, line 132: Did you mean `since` instead of `size` ?

kirugan (approved):

cmd/juno/dbsize.go
IronGauntlets: Yep fixed

db/pebble/db.go
IronGauntlets: Yes. fixed

weiihann (approved):

PR #1780, opened 2024-04-02
Remove unused `customNetworkF`

### LINKED ISSUE #1691: unused custom-network flag is being printed
I think [this](https://github.com/NethermindEth/juno/blob/main/node/node.go#L59) and [this](https://github.com/NethermindEth/juno/blob/main/cmd/juno/juno.go#L48) should be removed

IronGauntlets:
Fixes #1691


rianhughes:
I think we can also remove config.CustomNetwork

IronGauntlets:
Done

rianhughes (approved):

PR #1799, opened 2024-04-08
Update pipeline by removing Goerli deployment and switching tests to use Sepolia

wojciechos:
1. Dropped Goerli and integration deployments.
2. Switched tests to Sepolia. Excluding `traceTransaction` test because of a Juno bug + run only rpc releated tests
3. New creds for Sepolia account in use.

It appears that starknet.js includes tests not only releated with the rpc provider scope, hence selecting specific test files for running. Excluded`traceTransaction` for now due to a reported Juno bug and skipped `declare Sierra 1.5.0` since it's rather meant for devnet - it fails on live networks as contracts with the same hash can't be declared twice.

### DIFF
--- a/.github/workflows/build-and-deploy.yml
+++ b/.github/workflows/build-and-deploy.yml
@@ -2,7 +2,7 @@ name: Docker Build, Publish and Deploy
 
 on:
   push:
-    branches: [main, wojciechos/e2e-tests-sepolia]
+    branches: [main]
     tags: ['v*']
   workflow_dispatch:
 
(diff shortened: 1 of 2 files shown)


.github/workflows/starknet-js-tests.yml
```diff
@@ -29,7 +29,7 @@ jobs:
         run: npm ci
   
       - name: Run tests
-        run: npm test
+        run: npm test -- rpcProvider.test.ts transactionReceipt.test.ts rpcChannel.test.ts defaultProvider.test.ts contract.test.ts cairo1v2.test.ts cairo1v2_typed.test.ts cairo1.test.ts account.test.ts account.starknetId.test.ts --testNamePattern="^(?!.*traceTransaction|.*declare Sierra 1.5.0).*$"
```
wojciechos, line 32: It appears that starknet.js includes tests beyond the rpc provider scope, hence selecting specific test files for running. Excluded `traceTransaction` for now due to a reported Juno bug and skipped `declare Sierra 1.5.0` since it's meant for devnet - it fails on live networks as contracts with the same hash can't be declared twice.

.github/workflows/build-and-deploy.yml
```diff
@@ -2,7 +2,7 @@ name: Docker Build, Publish and Deploy
 
 on:
   push:
-    branches: [main]
+    branches: [main, wojciechos/e2e-tests-sepolia]
```
kirugan, line 5: do we need to keep additional branch?

wojciechos: Oh sorry, that was just for testing.

kirugan (approved):

PR #1819, opened 2024-04-10
Small code enhancements

kirugan:


adapters/vm2core/vm2core.go
```diff
@@ -43,29 +44,15 @@ func AdaptOrderedMessageToL1(message vm.OrderedL2toL1Message) *core.L2ToL1Messag
...
-	slices.SortFunc(messages, func(a vm.OrderedL2toL1Message, b vm.OrderedL2toL1Message) int {
-		switch {
-		case a.Order > b.Order:
-			return 1
-		case a.Order < b.Order:
-			return -1
-		default:
-			return 0
-		}
+	slices.SortFunc(messages, func(a, b vm.OrderedL2toL1Message) int {
```
pnowosie, line 47: I'm just curious 🤔 Would this not work?
```go
slices.SortFunc(messages, cmp.Compare)
```

utils/const.go
```diff
@@ -0,0 +1,5 @@
+package utils
+
+const (
+	MB = 1024 * 1024 // 1 Megabyte
```
pnowosie, line 4: I'd prefer be explicit
```suggestion
	Megabyte = 1024 * 1024
```

pnowosie (approved):
Nice 👍

utils/const.go
kirugan: Initially I thought `MB` is better, but just realised that your suggestion is better to read. Renamed, thanks

IronGauntlets (approved):

PR #1822, opened 2024-04-12
TraceLog changes

KananGiovanni:
Trace log changes


.golangci.yaml
```diff
@@ -46,15 +46,6 @@ linters-settings:
     ignored-functions:
       - strings.SplitN
 
-  govet:
```
rianhughes, line 49: Why were these removed?

.golangci.yaml
```diff
@@ -146,10 +138,4 @@ issues:
     - path: pkg/golinters/unused.go
       text: "rangeValCopy: each iteration copies 160 bytes \\(consider pointers or indexing\\)"
 
-run:
```
rianhughes, line 149: Why was this removed?

utils/log.go
```diff
@@ -59,6 +62,8 @@ func (l *LogLevel) Set(s string) error {
 		*l = WARN
 	case "ERROR", "error":
 		*l = ERROR
+	case "TRACE", "trace": // Adding the TRACE case
```
rianhughes, line 65: No need to add the comment :)

utils/log_test.go
```diff
@@ -19,7 +19,7 @@ var levelStrings = map[utils.LogLevel]string{
 
 func TestLogLevelString(t *testing.T) {
 	for level, str := range levelStrings {
-		t.Run("level "+str, func(t *testing.T) { //nolint:goconst
+		t.Run("level "+str, func(t *testing.T) {
```
rianhughes, line 22: You need to add the trace level to 
var levelStrings = map[utils.LogLevel]string{
...
}	utils.TRACE: "trace",

jsonrpc/server.go
```diff
@@ -397,7 +397,11 @@ func isNil(i any) bool {
 }
 
 func (s *Server) handleRequest(ctx context.Context, req *Request) (*response, error) {
+	reqJSON, _ := json.Marshal(req) // Serialise the request early for use in logging
```
rianhughes, line 400: If we marshal here, we don't need to again at line 444

jsonrpc/server.go
```diff
@@ -435,10 +444,13 @@ func (s *Server) handleRequest(ctx context.Context, req *Request) (*response, er
 			reqJSON, _ := json.Marshal(req)
 			errJSON, _ := json.Marshal(res.Error)
 			s.log.Debugw("Failed handing RPC request", "req", string(reqJSON), "res", string(errJSON))
+			s.log.Tracew("Internal error during RPC handling", "req", string(reqJSON), "error", string(errJSON))
 		}
 		return res, nil
 	}
 	res.Result = tuple[0].Interface()
+	resJSON, _ := json.Marshal(res) // Serialise response for logging
```
rianhughes, line 452: This will marshal for all log levels, but only log when the level=trace. I wonder if we can modify this so that we only cal json.Marshal when the log level is trace. Or just drop this log?

rianhughes: Wdyt @kirugan @IronGauntlets

utils/log.go
```diff
@@ -114,7 +124,12 @@ func NewZapLogger(logLevel LogLevel, colour bool) (*ZapLogger, error) {
 	if err != nil {
 		return nil, err
 	}
-	config.Level.SetLevel(level)
+	if logLevel == TRACE {
```
rianhughes, line 127: Maybe something like?

	var level zapcore.Level
	var err error

	if logLevel.String() == TRACE.String() {
		level = LevelTrace
	} else {
		level, err = zapcore.ParseLevel(logLevel.String())
		if err != nil {
			return nil, err
		}
	}
	config.Level.SetLevel(level)

rianhughes: Currently some tests will fail because 
 zapcore.ParseLevel(logLevel.String())
 returns an error when loglevel is trace

.golangci.yaml
KananGiovanni: Because the Golang ljnt keep showing "Warning" and it requires them to be replaced with Shadow configuration.

.golangci.yaml
KananGiovanni: The same issue with Golang lint, it keeps giving warning . Bot of those files keeps giving warnings

jsonrpc/server.go
KananGiovanni: But one is for request and the other for response, can't that cause soem errors?

jsonrpc/server.go
kirugan: Yes we need to marshal at the top and here only in case trace level is enabled. @KananGiovanni please note that we need to add request/response information to the log only -
there is no need to add additional `s.log.Tracew(...)` in this method.

kirugan: > I wonder if we can modify this so that we only call json.Marshal when the log level is trace

We can add method to check current log level in Logger (it's already there in ZapLogger, we will need to modify interface and probably wrapper)

.golangci.yaml
rianhughes: I think we should leave the linter config file as is, and just fix the problems the linter is raising

utils/log_test.go
```diff
@@ -10,6 +10,8 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+const levelLabel = "level "
+
 var levelStrings = map[utils.LogLevel]string{
```
rianhughes, line 15: You need to add the trace level here

utils/log_test.go
```diff
@@ -10,6 +10,8 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+const levelLabel = "level "
```
rianhughes, line 13: This can be removed

utils/network_test.go
```diff
@@ -14,6 +14,8 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+const networkLabel = "network "
```
rianhughes, line 17: What are you trying to do here?

KananGiovanni: The lint was giving an error about the "network" being repeated several times, and required to create const for them so i did and it passed, is it  a problem?

rianhughes: Removing networkLabel and adding //nolint:goconst to line 92 should fix this linting error

KananGiovanni: I actually did it  before adding them, but they kept failing. The lint kept giving an error

pnowosie (approved):
Nice 👍

jsonrpc/server.go
```diff
@@ -397,7 +397,9 @@ func isNil(i any) bool {
 }
 
 func (s *Server) handleRequest(ctx context.Context, req *Request) (*response, error) {
+	s.log.Tracew("Received request", "req", req)
 	if err := req.isSane(); err != nil {
+		s.log.Tracew("Request sanity check failed", "error", err.Error())
```
kirugan, line 402: 1. no need to call `.Error()` on error interface - logger will do it automatically
2. let's rename "error" to "err" (in our codebase "err" used ~ 54 times in logs while "error" just twice)

utils/log.go
```diff
@@ -90,12 +95,27 @@ type SimpleLogger interface {
...
 	Errorw(msg string, keysAndValues ...any)
+	Tracew(msg string, keysAndValues ...any)
 }
 
 type ZapLogger struct {
 	*zap.SugaredLogger
 }
 
+const (
+	TraceLevel = zapcore.Level(-2)
```
kirugan, line 106: Do you think we can make it private? I don't think we might pass somewhere this value explicitly outside of this package

utils/log.go
```diff
@@ -113,9 +133,17 @@ func NewZapLogger(logLevel LogLevel, colour bool) (*ZapLogger, error) {
...
 		enc.AppendString(t.Local().Format("15:04:05.000 02/01/2006 -07:00"))
 	}
-	level, err := zapcore.ParseLevel(logLevel.String())
-	if err != nil {
-		return nil, err
+
+	var level zapcore.Level
+	var err error
+	levelStr := logLevel.String()
+	if levelStr == TRACE.String() {
```
kirugan, line 140: can we do smth. like this: `if logLevel == TRACE {`

jsonrpc/server.go
```diff
@@ -417,6 +420,7 @@ func (s *Server) handleRequest(ctx context.Context, req *Request) (*response, er
 	args, err := s.buildArguments(ctx, req.Params, calledMethod)
 	if err != nil {
 		res.Error = Err(InvalidParams, err.Error())
+		s.log.Tracew("Error building arguments for RPC call", "error", err.Error())
```
kirugan, line 423: "err", err) 🙏

jsonrpc/server.go
```diff
@@ -439,6 +444,7 @@ func (s *Server) handleRequest(ctx context.Context, req *Request) (*response, er
 		return res, nil
 	}
 	res.Result = tuple[0].Interface()
+	s.log.Tracew("Successfully handled RPC request", "req", req)
```
kirugan, line 447: do we need to log response btw ?

PR #1830, opened 2024-04-18
Set default pending-poll-interval to 5s

### LINKED ISSUE #1787: Enable pending-poll by default
Right now pending poll is disabled by default. Let's enable it to 5s(? - confirm with pathfinder)
```
# Sets how frequently pending block will be updated (disabled by default)
pending-poll-interval: 0s
```

IronGauntlets:
fixes #1787


pnowosie (approved):
LGTM 👍

Issue #1794, opened 2024-04-04
Pending Transaction Trace Failure


wojciechos:
Encountering an "txn hash not found" error when using `starknet_traceTransaction` on pending transactions. The transactions are initially visible through `starknet_getTransactionReceipt` and `starknet_traceBlockTransactions - pending`. However, an immediate trace attempt fails.

### Steps to Reproduce
1. Add a transaction.
2. Retrieve its receipt via `starknet_getTransactionReceipt`.
3. Locate the transaction in `starknet_traceBlockTransactions - pending`.
4. Immediately trace the transaction with `starknet_traceTransaction` and note the "txn hash not found" error.
5. After txn is not pending anymore, attempt to trace the transaction again, and it should succeed.

(reproduce by running [starknet.js test](https://github.com/starknet-io/starknet.js/blob/develop/__tests__/rpcProvider.test.ts#L382))

### Expected Behavior
The transaction should be traceable immediately after it's visible in pending transactions without resulting in a "txn hash not found" error.

PR #1833, opened 2024-04-22
Handle pending block case in traceTransaction method

kirugan:
Issue #1794


rpc/handlers.go
```diff
@@ -1505,14 +1505,24 @@ func (h *Handler) TraceTransactionV0_6(ctx context.Context, hash felt.Felt) (*vm
...
 
-	block, err := h.bcReader.BlockByNumber(blockNumber)
-	if err != nil {
-		return nil, ErrBlockNotFound
+	var block *core.Block
+	pendingBlock := blockHash == nil
+	if pendingBlock {
+		pending, err := h.bcReader.Pending()
+		if err != nil {
+			return nil, ErrInternal
```
kirugan, line 1518: The reason it's internal and not `BlockNotFound` is because afaik pending block should be always present and if it's not there are some other problems how we deal with it. @IronGauntlets right?

rpc/handlers.go
```diff
@@ -1505,14 +1505,24 @@ func (h *Handler) TraceTransactionV0_6(ctx context.Context, hash felt.Felt) (*vm
...
+	_, blockHash, _, err := h.bcReader.Receipt(hash)
 	if err != nil {
 		return nil, ErrTxnHashNotFound
 	}
 
-	block, err := h.bcReader.BlockByNumber(blockNumber)
-	if err != nil {
-		return nil, ErrBlockNotFound
+	var block *core.Block
+	pendingBlock := blockHash == nil
```
rianhughes, line 1514: nit: isPendingBlock

rianhughes (approved):

rpc/handlers.go
kirugan: Renamed, thanks

rpc/handlers.go
kirugan: We checked with Aneeque: these handlers should return only TxHashNotFound errors

wojciechos (approved):

PR #1834, opened 2024-04-23
rpc specs and some more tests

### LINKED ISSUE #1802: Improve coverage and check RPC tests against spec
- Go through every handler tests and compare are they enough cover the 0.7 spec
- Add more tests
- Run actual node RPC request to check the actual responses

pnowosie:
Fixes #1802

### DIFF
--- a/rpc/block_test.go
+++ b/rpc/block_test.go
@@ -3,7 +3,6 @@ package rpc_test
 import (
 	"context"
 	"errors"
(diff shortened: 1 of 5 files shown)


pnowosie:
# ⚠️  DO NOT MERGE

It will be easier to apply newly added tests on main in a new PR after one big #1880 is merged

PR #1841, opened 2024-04-25
Small code enhancements

kirugan:


core/contract.go
```diff
@@ -9,7 +9,7 @@ import (
 	"github.com/NethermindEth/juno/db"
 )
 
-const contractStorageTrieHeight = 251
+const ContractStorageTrieHeight = 251
```
pnowosie, line 12: Just curious from where this number comes?
Is it set by us, kind-of big enough value, or it has to be exactly `251`.
Consider adding a comment here.

pnowosie (approved):
Nice 👍

core/contract.go
kirugan: I believe it's always fixed (see https://docs.starknet.io/documentation/architecture_and_concepts/Network_Architecture/starknet-state/#:~:text=storage_root%20is%20the%20root%20of%20another%20Merkle%2DPatricia%20trie%20of%20height%20251%20that%20is%20constructed%20from%20the%20contract%E2%80%99s%20storage.)

kirugan: added

Issue #1450, opened 2023-11-22
Missing version validation for pending blocks


wojciechos:
Juno currently accepts and stores **pending** blocks regardless of whether their versions are supported by the current version of Juno. This behavior is inconsistent with how Juno handles normal (latest) blocks, where there is a version compatibility check.

**Current Behavior:** 
When Juno encounters a pending block with an unsupported version, it still stores this block.

**Result:** 
This leads to operational problems, including panics:
```
	/usr/lib/go-1.21/src/net/http/server.go:3086 +0x5cb
created by net/http.(*Server).Serve in goroutine 10

	/usr/lib/go-1.21/src/net/http/server.go:2009 +0x5f4
net/http.(*conn).serve(0xc0009903f0, {0x3235de0, 0xc00039a030})
	/usr/lib/go-1.21/src/net/http/server.go:2938 +0x8e
net/http.serverHandler.ServeHTTP({0x322cf30?}, {0x3231180?, 0xc000b3a1c0?}, 0x6?)
	/usr/lib/go-1.21/src/net/http/server.go:2136 +0x29
net/http.HandlerFunc.ServeHTTP(0x4b9aa5?, {0x3231180?, 0xc000b3a1c0?}, 0xc000b3a101?)
	/root/go/pkg/mod/github.com/rs/cors@v1.10.1/cors.go:281 +0x184
github.com/NethermindEth/juno/node.makeRPCOverHTTP.(*Cors).Handler.func2({0x3231180, 0xc000b3a1c0}, 0xc000ff7d00)
	/usr/lib/go-1.21/src/net/http/server.go:2514 +0x142
net/http.(*ServeMux).ServeHTTP(0xc00049a2a0?, {0x3231180, 0xc000b3a1c0}, 0xc000ff7d00)
	/usr/lib/go-1.21/src/net/http/server.go:2136 +0x29
net/http.HandlerFunc.ServeHTTP(0xc000f40d20?, {0x3231180?, 0xc000b3a1c0?}, 0x2da8b6b?)
	/app/node/http.go:71 +0x98
github.com/NethermindEth/juno/node.makeRPCOverHTTP.exactPathServer.func1({0x3231180, 0xc000b3a1c0}, 0xc000ff7d00)
	/app/jsonrpc/http.go:49 +0x227
[...]

Issue #1789, opened 2024-04-03
Pending Block and State can become out of sync


IronGauntlets:
Pending Block and State are written to the DB this means that any subsequent runs of Juno when `pending-poll` is disabled would have access to the old Pending Block and State which is no longer relevant.

There are 2 solutions:

- Stop persisting the pending block and state across multiple runs of Juno and only manage it in memory. This may require a migration to remove the old pending block.
- Check if the pending block is outdated and return `ErrBlockNotFound`.

IronGauntlets:
This should be done along with #1787

PR #1844, opened 2024-04-27
Manage pending in memory and move it to synchroniser

IronGauntlets:
Fixes ~parts of~ #1789. ~A migration to remove the pending from DB is still outstanding.~

It would be best to review the commits individually and in order.


IronGauntlets:
This also fixes #1450

IronGauntlets:
Since I will be away for 2 weeks. Please **"Rebase and Merge"** once it is approved.

blockchain/blockchain.go
```diff
@@ -277,26 +253,6 @@ func (b *Blockchain) Receipt(hash *felt.Felt) (*core.TransactionReceipt, *felt.F
...
-				if hash.Equal(t.Hash()) {
-					receipt = pending.Block.Receipts[i]
-					blockHash = nil
-					blockNumber = 0
-					return nil
-				}
-			}
-			return db.ErrKeyNotFound
-		}
-
```
kirugan, line 299: why for tx-s you moved logic to `transactionByHash` but here it's not part of receiptByHash anymore ?

db/buckets.go
```diff
@@ -29,7 +29,7 @@ const (
 	ContractDeploymentHeight
 	L1Height
 	SchemaVersion
-	Pending
+	Unused
```
kirugan, line 32: do you think it's worth adding a comment i.e. it was Pending previously

node/node.go
```diff
@@ -123,7 +123,8 @@ func New(cfg *Config, version string) (*Node, error) { //nolint:gocyclo,funlen
 
 	services := make([]service.Service, 0)
 
-	chain := blockchain.New(database, &cfg.Network)
+	var synchronizer *sync.Synchronizer
+	chain := blockchain.New(database, &cfg.Network, synchronizer.PendingBlock)
```
kirugan, line 127: it causes nil pointer dereference because `synchronizer` is not set before func pointer is passed (`synchronizer.PendingBlock`). Tbh I'm quite suprise that go allows to pass func pointer of variable that is nil

blockchain/blockchain.go
IronGauntlets: I don't think I understand your question.

I removed checking of pending transactions and receipt from both `TransactionByHash` and `Receip`

node/node.go
IronGauntlets: fixed

db/buckets.go
IronGauntlets: done

sync/sync.go
```diff
@@ -41,13 +44,17 @@ type Reader interface {
 	StartingBlockNumber() (uint64, error)
 	HighestBlockHeader() *core.Header
 	SubscribeNewHeads() HeaderSubscription
+
+	Pending() (*Pending, error)
+	PendingBlock() *core.Block
+	PendingState() (core.StateReader, func() error, error)
```
rianhughes, line 50: PendingNewClasses() as well?

sync/sync.go
```diff
@@ -443,3 +465,78 @@ func (s *Synchronizer) SubscribeNewHeads() HeaderSubscription {
...
+}
+
+func (s *Synchronizer) Pending() (*Pending, error) {
+	p := s.pending.Load()
+	if p == nil {
+		return nil, ErrPendingBlockNotFound
+	}
+
+	expectedParentHash := &felt.Zero
+	if head, err := s.blockchain.HeadsHeader(); err == nil {
```
rianhughes, line 508: Should we not handle the error here (when it's not nil)?

rianhughes: Ah, nvm. I guess if there's an error we just return ErrPendingBlockNotFound

sync/sync.go
IronGauntlets: I am not sure what you mean. I just moved the function from Blockchain Reader which referred to Pending.

rianhughes (approved):

blockchain/blockchain.go
```diff
@@ -970,57 +923,44 @@ func (b *Blockchain) StorePending(pending *Pending) error {
 			return ErrParentDoesNotMatchHead
 		}
 
-		if existingPending, err := b.pendingBlock(txn); err == nil {
+		if existingPending, err := pendingBlock(txn); err == nil {
 			if existingPending.Block.TransactionCount >= pending.Block.TransactionCount {
-				return nil // ignore the incoming pending if it has fewer transactions than the one we already have
+				// ignore the incoming pending if it has fewer transactions than the one we already have
```
kirugan: can you please elaborate in what case it is possible to have incoming block with smaller number of transactions ?

blockchain/blockchain.go
```diff
@@ -248,7 +248,7 @@ func (b *Blockchain) TransactionByHash(hash *felt.Felt) (core.Transaction, error
 		// not found in the canonical blocks, try pending
 		if errors.Is(err, db.ErrKeyNotFound) {
 			var pending Pending
-			pending, err = b.pendingBlock(txn)
+			pending, err = pendingBlock(txn)
```
kirugan: should we check for `ErrPendingBlockNotFound` here?

blockchain/blockchain.go
kirugan: Sorry, seems like I was referring to some other commit. I don't see much differences now.

blockchain/blockchain.go
```diff
@@ -40,23 +38,21 @@ type Reader interface {
...
 }
 
 var (
 	ErrParentDoesNotMatchHead = errors.New("block's parent hash does not match head block hash")
+	ErrPendingBlockNotFound   = errors.New("pending block not found")
 	SupportedStarknetVersion  = semver.MustParse("0.13.3")
 )
 
-func checkBlockVersion(protocolVersion string) error {
+func CheckBlockVersion(protocolVersion string) error {
```
kirugan, line 55: So you making this function public because previously it was used only in `blockchain`, but now since `blockchain` is not responsible for storing pending block you need to call it in `sync` (because pending block is stored there). Correct ?

rpc/helpers.go
```diff
@@ -38,15 +39,17 @@ func (h *Handler) blockByID(id *BlockID) (*core.Block, *jsonrpc.Error) {
...
 		block, err = h.bcReader.BlockByHash(id.Hash)
 	case id.Pending:
-		var pending blockchain.Pending
-		pending, err = h.bcReader.Pending()
-		block = pending.Block
+		var pending *sync.Pending
+		pending, err = h.syncReader.Pending()
+		if err == nil {
+			block = pending.Block
+		}
```
kirugan, line 46: Since there is new method `PendingBlock` maybe we should call it here:
```
case id.Pending:
    block = h.syncReader.PendingBlock()
```

sync/sync.go
```diff
@@ -42,13 +45,17 @@ type Reader interface {
...
+	PendingBlock() *core.Block
+	PendingState() (core.StateReader, func() error, error)
 }
 
 // This is temporary and will be removed once the p2p synchronizer implements this interface.
 type NoopSynchronizer struct{}
 
 func (n *NoopSynchronizer) StartingBlockNumber() (uint64, error) {
-	return 0, nil
+	return 0, errors.New("StartingBlockNumber not implemented")
```
kirugan, line 58: nit: to make it consistent maybe we should pass this message "StartingBlockNumber() not implemented"

sync/sync.go
```diff
@@ -71,16 +91,18 @@ type Synchronizer struct {
...
+	pending             atomic.Pointer[Pending]
 	pendingPollInterval time.Duration
 	catchUpMode         bool
 	plugin              junoplugin.JunoPlugin
 }
 
-func New(bc *blockchain.Blockchain, starkNetData starknetdata.StarknetData,
-	log utils.SimpleLogger, pendingPollInterval time.Duration, readOnlyBlockchain bool,
+func New(bc *blockchain.Blockchain, starkNetData starknetdata.StarknetData, log utils.SimpleLogger,
+	pendingPollInterval time.Duration, readOnlyBlockchain bool, database db.DB,
```
kirugan, line 101: I think it breaks existing abstractions. But I see that you need to use it in order to generate state for pending block. Maybe it's possible to rely on memory txn ?

node/node.go
```diff
@@ -129,7 +129,8 @@ func New(cfg *Config, version string) (*Node, error) { //nolint:gocyclo,funlen
 
 	services := make([]service.Service, 0)
 
-	chain := blockchain.New(database, &cfg.Network)
+	synchronizer := new(sync.Synchronizer)
```
kirugan, line 132: Looks like we create synchronizer and they depend on each other. What were other alternatives to puting pending block in synchronizer? (not going to ask you to rewrite it, I understand that this written like this for a reason)

blockchain/event_filter.go
```diff
@@ -144,7 +147,7 @@ func (e *EventFilter) Events(cToken *ContinuationToken, chunkSize uint64) ([]*Fi
 				return nil, nil, err
 			}
 		} else {
-			header = pending.Block.Header
+			header = pending.Header
```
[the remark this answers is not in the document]
IronGauntlets: The nil check is done above on line 125. Since we are setting the limit on the block number in the if statement on line 121, it is not possible for this else case to execute if pending is nil.

rpc/helpers.go
IronGauntlets: We could do that but then I will have to add another nil check after the switch statement to check if pending block was found or not. With the current code we get an error, which makes current error handling more consistent.
Please let me know which you would prefer?

sync/sync.go
IronGauntlets: Done

sync/sync.go
IronGauntlets: The DB is not for storing the pending state; it is for accessing the current state and is a read-only operation. 

The pending state is created by fetching the current state and applying the pending StateDiff. To get access to StateReader, we need a DB transaction. So, unfortunately, I cannot replace it with memory txn unless I store the Starknet state in memory (which is not a good idea).

node/node.go
IronGauntlets: Yeah, this was something I thought about too. I don't think the dependency between the synchroniser and the pending can be removed since the synchroniser updates the pending block and state. One potential solution is to have pending in its own package and let synchroniser import it but I think this will be just a cosmetic change (i.e. nothing meaningful).

blockchain/event_filter.go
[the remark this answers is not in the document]
kirugan: I see, now latest+1 makes sense. thanks

sync/sync.go
kirugan: got it, thanks for clarifications

rpc/helpers.go
kirugan: Agree, let's keep it this way then

kirugan (approved):

PR #1852, opened 2024-05-01
implement getProof and verifyProof on trie

rianhughes:
closes: https://github.com/NethermindEth/juno/issues/1851


core/trie/proof.go
```diff
@@ -0,0 +1,80 @@
...
+func VerifyProof(root *felt.Felt, leafPath *Key, leafHash felt.Felt, proofNodes []ProofNode, hashFunc hashFunc) (Membership, error) {
+	expectedHash := root
+
+	for i, pNode := range proofNodes {
+		pNodeHash := hashFunc(pNode.LeftHash, pNode.RightHash)
+		if !expectedHash.Equal(pNodeHash) {
+			return NonMember, errors.New("proof node does not have expected hash")
+		}
+
+		if leafPath.Test(leafPath.Len() - uint8(i) - 1) { // Todo: are we selecting the correct child here??
```
kirugan, line 68: could you explain how does it work ?

core/trie/proof.go
```diff
@@ -0,0 +1,80 @@
...
+			LeftHash:  leftHash,
+			RightHash: rightHash,
+		}
+	}
+	return proofNodes, nil
+}
+
+// verifyProof checks if `leafPath` leads from `root` to `leafHash` along the `proofNodes`
+// https://github.com/eqlabs/pathfinder/blob/main/crates/merkle-tree/src/tree.rs#L2006
+func VerifyProof(root *felt.Felt, leafPath *Key, leafHash felt.Felt, proofNodes []ProofNode, hashFunc hashFunc) (Membership, error) {
```
kirugan, line 59: Since the presence of an error affects Membership (i.e., if err is nil, then Membership is Member, and vice versa), I propose simplifying the signature:

Either leave only the error (a nil error indicates that the proof is correct, i.e., Member)
Or return a boolean as the first return value.
Personally, I'm in favor of the first option.

core/trie/proof.go
```diff
@@ -0,0 +1,80 @@
...
+// https://github.com/eqlabs/pathfinder/blob/main/crates/merkle-tree/src/tree.rs#L514
+func GetProof(leaf *felt.Felt, tri *Trie) ([]ProofNode, error) {
+	leafKey := tri.feltToKey(leaf)
+	nodesToLeaf, err := tri.nodesFromRoot(&leafKey)
+	if err != nil {
+		return nil, err
+	}
+	proofNodes := make([]ProofNode, len(nodesToLeaf)-1)
+
+	getHash := func(tri *Trie, key *Key) (*felt.Felt, error) {
```
kirugan, line 31: `tri` is always the same: wdyt about removing this parameter and rely on closure (i.e. capturing outer scope with `tri` parameter from `GetProof(...)` ?

core/trie/proof.go
```diff
@@ -0,0 +1,80 @@
...
+type Membership int
+
+const (
+	Member Membership = iota
+	NonMember
+)
+
+type ProofNode struct {
+	LeftHash  *felt.Felt
+	RightHash *felt.Felt
```
kirugan, line 19: image

to better understand what's going on I attached image of a binary tree 😄 
so let's say we need to proof the presence of node 9 in the trie, we will return ProofNodes from root to the leaf (9):
[50, 17, 12, 9]. Question is: since we return just one branch of a trie why do we need to calculate rightHash (in this example) at all?

rianhughes: So we actually only return the proofNodes of [50,17,12] which are the hashes of the left and right child.
To verify these root nodes, we need to also supply the root hash, the leaf hash, and the path down to the leaf hash.

We effectively check that:
- rootHash = H(hash_17,hash_72) 
- hash_17 = H(hash_12, hash_23)
- hash_12 = H(hash_9, hash_14)
- hash_9 = leafHash
where the path was used to select the above nodes. If at any point this fails (eg hash9!=leafHash) then we know the path doesn't lead from the root to the leaf along the provided proofNodes.

core/trie/proof.go
rianhughes: The path can be thought of as a 256 long bit string, and here we return the i'th bit from the right. If it's a 1, we go right, if it's a 0 we go left.

rianhughes: We effectively test if the i'th bit is a 1 or 0. This is done by selecting the relevant byte in [32]byte, and then the relevant bit in that byte (by right shifting it, and then performing a mask (?) operation against unit(1)). Which is then compared against 0, and the result is returned.

core/trie/proof.go
asdacap: Note: You need the edge to the branch also. https://github.com/starknet-io/starknet-p2p-specs/blob/main/p2p/proto/snapshot.proto#L6

core/trie/proof.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+// verifyProof checks if `leafPath` leads from `root` to `leafHash` along the `proofNodes`
+// https://github.com/eqlabs/pathfinder/blob/main/crates/merkle-tree/src/tree.rs#L2006
+func VerifyProof(root *felt.Felt, leafPath *Key, leafHash felt.Felt, proofNodes []ProofNode, hashFunc hashFunc) error {
+	expectedHash := root
+
+	for i, pNode := range proofNodes {
+		pNodeHash := hashFunc(pNode.LeftHash, pNode.RightHash)
+		if !expectedHash.Equal(pNodeHash) {
```
asdacap, line 58: In practice, there are no specific ordering of the proof. So it will need to be sorted manually first.

rianhughes: GetProof should return proofs in order though, so what scenario would VerifyProof be called with non-ordered proofs?

asdacap: Whatever scenario that other nodes come up with. It could be that they put the nodes in a hashset to dedup them for example, which causes it to be unordered.

core/trie/proof.go
```diff
@@ -0,0 +1,112 @@
...
+
+	return proofNodes, nil
+}
+
+// verifyProof checks if `leafPath` leads from `root` to `leafHash` along the `proofNodes`
+// https://github.com/eqlabs/pathfinder/blob/main/crates/merkle-tree/src/tree.rs#L2006
+// func VerifyProof(root *felt.Felt, leafPath *Key, leafHash felt.Felt, proofNodes []ProofNode, hashFunc hashFunc) error {
+// 	expectedHash := root
+
+// 	for i, pNode := range proofNodes {
```
asdacap, line 93: Keep in mind, when using a proof for a range of leaf, you'll need to make sure that it fail when a leaf in the middle of the tree is missing. Which is a bit more complicated.

asdacap (approved):
Aside from the proving need to work for range also instead of just one leaf (can be done later), looks good to me.

core/trie/proof.go
```diff
@@ -0,0 +1,180 @@
...
+	if err != nil {
+		return nil, nil, err
+	}
+
+	rightHash := rNode.Value
+	if isEdge(sNode.key, storageNode{node: rNode, key: sNode.node.Right}) {
+		edgePath := path(sNode.node.Right, sNode.key)
+		rEdge := ProofNode{Edge: &Edge{
+			Path:  &edgePath,
+			Child: rNode.Value,
```
asdacap, line 95: To clarify, the rNode.Value is the hash?

core/trie/proof.go
```diff
@@ -0,0 +1,180 @@
...
+func (pn *ProofNode) Hash() *felt.Felt {
+	switch {
+	case pn.Binary != nil:
+		return crypto.Pedersen(pn.Binary.LeftHash, pn.Binary.RightHash)
+	case pn.Edge != nil:
+		length := make([]byte, len(pn.Edge.Path.bitset))
+		length[len(pn.Edge.Path.bitset)-1] = pn.Edge.Path.len
+		pathFelt := pn.Edge.Path.Felt()
+		lengthFelt := new(felt.Felt).SetBytes(length)
+		return new(felt.Felt).Add(crypto.Pedersen(pn.Edge.Child, &pathFelt), lengthFelt)
```
asdacap, line 26: This might be a serving bottleneck. Something to take note on.

core/trie/proof.go
rianhughes: Yup

PR #1853, opened 2024-05-01
feat: GetNodesFromRoot rpc handler added

bilgin-kocak:
Onboarding exercise 5

Implement a new rpc method “juno_getNodesFromRoot(key felt.Felt)” that returns the set of nodes from the root to the key for the classes Trie. See [nodesFromRoot()](https://github.com/NethermindEth/juno/blob/main/core/trie/trie.go#L141). Remember to implement tests for the new logic

### DIFF
--- a/rpc/handlers.go
+++ b/rpc/handlers.go
@@ -7,11 +7,13 @@ import (
 	"encoding/json"
 	"math"
 	stdsync "sync"
+	"fmt"
 
 	"github.com/NethermindEth/juno/blockchain"
 	"github.com/NethermindEth/juno/clients/feeder"
 	"github.com/NethermindEth/juno/core"
 	"github.com/NethermindEth/juno/core/felt"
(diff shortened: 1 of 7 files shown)


rpc/handlers.go
```diff
@@ -175,6 +177,19 @@ func (h *Handler) SpecVersionV0_6() (string, *jsonrpc.Error) {
...
+	// TODO: Implement this method
+	// Implement a new rpc method “juno_getNodesFromRoot(key felt.Felt)” 
+	// that returns the set of nodes from the root to the key for the classes Trie. 
+	// See nodesFromRoot(). Remember to implement tests for the new logic
+
+	// trie :=
+	// key_ := trie.feltToKey(key)
+	// storageNodes, _ := trie.nodesFromRoot(key_)
+
+	return "0x1", nil
```
kirugan, line 190: set of nodes is slice of some type (probably felts), not string

PR #1856, opened 2024-05-02
Refactor p2p handlers: move commong iterator logic to a helper

kirugan:


p2p/starknet/handlers.go
kirugan, line 1: Added todo-s need to be discussed later (after this PR is merged)

p2p/starknet/handlers.go
```diff
@@ -112,75 +120,64 @@ func (h *Handler) CurrentBlockHeaderHandler(stream network.Stream) {
...
-		if finMsg, ok := fin(); ok {
-			yield(finMsg)
-		}
-	}
+			},
+		}, nil
+	})
 }
 
 func (h *Handler) onBlockBodiesRequest(req *spec.BlockBodiesRequest) (iter.Seq[proto.Message], error) {
```
rianhughes, line 183: Can we use processIterationRequest here?

p2p/starknet/handlers.go
```diff
@@ -223,140 +216,134 @@ func (h *Handler) onBlockBodiesRequest(req *spec.BlockBodiesRequest) (iter.Seq[p
...
+}
+
+// iterationProcessor is an alias for a function that will generate corresponding data
+// given block data for current iteration through blockDataAccessor
+type iterationProcessor = func(it blockDataAccessor) (proto.Message, error)
+
+// processIterationRequest is helper function that simplifies data processing for provided spec.Iteration object
+// caller usually passes iteration object from received request, finMsg as final message to a peer
+// and iterationProcessor function that will generate response for each iteration
+func (h *Handler) processIterationRequest(iteration *spec.Iteration, finMsg proto.Message, f iterationProcessor) (iter.Seq[proto.Message], error) {
```
rianhughes, line 311: nit: can we rename f to getMsg or something?

p2p/starknet/handlers.go
```diff
@@ -71,17 +76,20 @@ func streamHandler[ReqT proto.Message](ctx context.Context, stream network.Strea
 
 	response, err := reqHandler(req.(ReqT))
```
rianhughes, line 77: What do you think about renaming "response" to "requestIterator" or something?

rianhughes: This may be non-standard, but renaming yield in processIterationRequest to processRquest might also make it easier to understand at first glance

p2p/starknet/handlers.go
kirugan: unfortunately now, this handler is different because it sends multiple different messages for 1 request. But spoiler: in updated spec this method is going to be removed.

p2p/starknet/handlers.go
kirugan: > What do you think about renaming "response" to "requestIterator" or something?

I renamed it to "responseIterator" since it's iterator over response.

kirugan: > This may be non-standard, but renaming yield in processIterationRequest to processRequest might also make it easier to understand at first glance

afaik `yield` is standard name for generators (python, js, php and few others have this statement keyword). If that's ok I'll keep it for a while if there are more complains about naming we will rename it

rianhughes (approved):

PR #1863, opened 2024-05-14
Rianhughes/sequencer

rianhughes:
Merging https://github.com/NethermindEth/juno/compare/main...feature/sequencer into main, by cherry-picking


rianhughes:
Note: the token contract should be updated to remove the assertions that the permissioned_minter can't be the zero address (when calling the token constructor blockifier, the sender address is the zero address - I currently get around this by passing the sender address into blockifier - which is a bit hacky).

Issue #1767, opened 2024-03-26
Add trace option to --log-level for In-Depth RPC Logging


wojciechos:
Add a `trace` log level option, more detailed than `debug`, to capture all incoming and outgoing RPC requests

PR #1868, opened 2024-05-16
add trace to CLI help output

LordGhostX:
This PR is a follow-up to https://github.com/NethermindEth/juno/pull/1822 and fully resolves https://github.com/NethermindEth/juno/issues/1767

- Added `trace` option to the `log-level` description in the CLI
- Resolve casing issues (websocket -> WebSocket, GRPC -> gRPC, yaml -> YAML, prometheus -> Prometheus)
- Fix typo in CLI description

### DIFF
--- a/cmd/juno/juno.go
+++ b/cmd/juno/juno.go
@@ -119,7 +119,7 @@ const (
 	defaultCorsEnable               = false
 
 	configFlagUsage                       = "The YAML configuration file."
-	logLevelFlagUsage                     = "Options: debug, info, warn, error, trace."
+	logLevelFlagUsage                     = "Options: trace, debug, info, warn, error."
 	httpUsage                             = "Enables the HTTP RPC server on the default port and interface."
 	httpHostUsage                         = "The interface on which the HTTP RPC server will listen for requests."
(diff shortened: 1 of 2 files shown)


cmd/juno/juno.go
```diff
@@ -118,14 +118,14 @@ const (
 	defaultGwTimeout                = 5 * time.Second
 	defaultCorsEnable               = false
 
-	configFlagUsage                       = "The yaml configuration file."
-	logLevelFlagUsage                     = "Options: debug, info, warn, error."
+	configFlagUsage                       = "The YAML configuration file."
+	logLevelFlagUsage                     = "Options: debug, info, warn, error, trace."
```
wojciechos, line 122: Looks like it's ordered from the most verbose to least. I would put trace in the first place.

LordGhostX: Oh, interesting. I actually followed the order here: https://github.com/NethermindEth/juno/blob/3283b79ef8eae76c374eab6c5c3b7e3b76a43726/utils/log.go#L14-L17

wojciechos: yep, let's update that as well

wojciechos (approved):

Issue #1866, opened 2024-05-15
P2P Sepolia Sync Stuck at Block 57927


wojciechos:
Node in P2P mode gets stuck at block 57927 and is unable to sync further. This issue has been reproduced twice by restarting all nodes and forcing a full re-sync from scratch.

#### Observations

1. The normal node finds 2 peers and starts getting data.
2. It consistently fails at block 57927.
3. Logs show the following error: ``` starknet/client.go:79 Error while reading from stream {"err": "stream reset"}```
6. Got in loop to point 1

#### Current Setup
- 6 feeder p2p-nodes
- 1 p2p-node 

#### More logs

```
10:55:29.967 15/05/2024 +00:00	INFO	p2p/sync.go:119	Start Pipeline	{"Random node height": 66503, "Current height": 57927, "Start": 57928, "End": 58028} |  
-- | --
  |   | 10:55:29.966 15/05/2024 +00:00	DEBUG	p2p/sync.go:631	Random chosen peer's info	{"peerInfo": "{12D3KooWPuESJuHpSPEzxNXVhEPKzsAM2XzQkHF5hGiSLfXytDCw: []}"} |  
  |   | 10:55:29.966 15/05/2024 +00:00	DEBUG	p2p/sync.go:630	Number of peers	{"len": 2} |  
  |   | 10:55:29.966 15/05/2024 +00:00	DEBUG	p2p/sync.go:86	Continuous iteration	{"i": 37906} |  
  |   | 10:55:29.966 15/05/2024 +00:00	DEBUG	p2p/sync.go:258	Received Events	{"blockNumber": 58027} |  
  |   | 10:55:29.963 15/05/2024 +00:00	DEBUG	p2p/sync.go:258	Received Events	{"blockNumber": 58026} |  
  |   | 10:55:29.957 15/05/2024 +00:00	DEBUG	p2p/sync.go:258	Received Events	{"blockNumber": 58025} |  
  |   | 10:55:29.952 15/05/2024 +00:00	DEBUG	p2p/sync.go:258	Received Events	{"blockNumber": 58024} |  
  |   | 10:55:29.950 15/05/2024 +00:00	DEBUG	p2p/sync.go:258	Received Events	{"blockNumber": 58023} |  
  |   | 10:55:29.947 15/05/2024 +00:00	DEBUG	p2p/sync.go:258	Received Events	{"blockNumber": 58022} |  
[...]

PR #1869, opened 2024-05-18
Increase size of possible p2p message for unmarshaller

kirugan:
Issue #1866

### DIFF
--- a/p2p/starknet/client.go
+++ b/p2p/starknet/client.go
@@ -14,6 +14,8 @@ import (
 	"google.golang.org/protobuf/proto"
 )
 
+const unmarshalMaxSize = 15 * utils.Megabyte
+
 type NewStreamFunc func(ctx context.Context, pids ...protocol.ID) (network.Stream, error)
 
 type Client struct {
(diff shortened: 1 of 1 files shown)


p2p/starknet/client.go
```diff
@@ -14,6 +14,8 @@ import (
 	"google.golang.org/protobuf/proto"
 )
 
+const unmarshalMaxSize = 15 * utils.Megabyte
```
wojciechos, line 17: Wondering why the limit is set to 15mb? Are we sure there won't be messages bigger than that?

kirugan: Idk, max message that we saw is 12 Mb, I set it +3 Mb just to be aware of next time when it happens

pnowosie (approved):
👍

p2p/starknet/client.go
pnowosie: Are there a downsides to set it higher? 

I feel we will be hit by this again 🤔 
Should [p2p spec](https://github.com/starknet-io/starknet-p2p-specs) define max message size?
Maybe we should check & set it at the similar level as Ethereum client?

kirugan: > Should [p2p spec](https://github.com/starknet-io/starknet-p2p-specs) define max message size?

That's a good point, initially we discussed it as 2Mb (because there were rumors about max possible size that's not controlled by a caller) but then decided to get rid from the spec. 

I think we should do the following:
- implement recent changes from the spec (it will reduce size of messages)
- come up with some max message size based on starknet constraints (max number of tx-s per block etc.) and what we get from real-world tests

PR #1873, opened 2024-05-21
VerifyProofs

rianhughes:
The PR implements VerifyRangeProof. It accounts for the cases where either 0, 1 or both boundary proofs are supplied. It does require the proof keys to be set.

Note: It does not account for the case where the proof keys are not set. I have opened a new PR that accounts for this case https://github.com/NethermindEth/juno/pull/1901

### DIFF
--- a/core/trie/proof.go
+++ b/core/trie/proof.go
@@ -54,7 +54,7 @@ type Edge struct {
 	Value *felt.Felt // this nodes hash
 }
 
-func isEdge(parentKey *Key, sNode storageNode) bool {
+func isEdge(parentKey *Key, sNode StorageNode) bool {
 	sNodeLen := sNode.key.len
 	if parentKey == nil { // Root
 		return sNodeLen != 0
@@ -64,7 +64,7 @@ func isEdge(parentKey *Key, sNode storageNode) bool {
 
 // Note: we need to account for the fact that Junos Trie has nodes that are Binary AND Edge,
 // whereas the protocol requires nodes that are Binary XOR Edge
-func adaptNodeToSnap(tri *Trie, parentKey *Key, sNode storageNode) (*Edge, *Binary, error) {
+func adaptNodeToSnap(tri *Trie, parentKey *Key, sNode StorageNode) (*Edge, *Binary, error) {
 	isEdgeBool := isEdge(parentKey, sNode)
 
 	var edge *Edge
@@ -88,7 +88,7 @@ func adaptNodeToSnap(tri *Trie, parentKey *Key, sNode storageNode) (*Edge, *Bina
 	}
 
 	rightHash := rNode.Value
-	if isEdge(sNode.key, storageNode{node: rNode, key: sNode.node.Right}) {
+	if isEdge(sNode.key, StorageNode{node: rNode, key: sNode.node.Right}) {
 		edgePath := path(sNode.node.Right, sNode.key)
 		rEdge := ProofNode{Edge: &Edge{
 			Path:  &edgePath,
@@ -97,7 +97,7 @@ func adaptNodeToSnap(tri *Trie, parentKey *Key, sNode storageNode) (*Edge, *Bina
 		rightHash = rEdge.Hash(tri.hash)
 	}
 	leftHash := lNode.Value
-	if isEdge(sNode.key, storageNode{node: lNode, key: sNode.node.Left}) {
+	if isEdge(sNode.key, StorageNode{node: lNode, key: sNode.node.Left}) {
 		edgePath := path(sNode.node.Left, sNode.key)
 		lEdge := ProofNode{Edge: &Edge{
 			Path:  &edgePath,
@@ -258,7 +258,7 @@ func VerifyRangeProof(root *felt.Felt, keys []*felt.Felt, values []*felt.Felt, p
 }
 
 // Only the path down to leaf Key will be set correctly. Not neightbouring keys
-func ProofToPath(proofNodes []ProofNode, leaf *felt.Felt, hashF hashFunc) ([]storageNode, error) {
+func ProofToPath(proofNodes []ProofNode, leaf *felt.Felt, hashF hashFunc) ([]StorageNode, error) {
 
 	shouldSquish := func(parent *ProofNode, child *ProofNode) (uint8, uint8) {
 		if parent == nil || child == nil {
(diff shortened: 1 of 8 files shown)


core/trie/proof.go
```diff
@@ -145,6 +146,22 @@ func GetProof(leaf *felt.Felt, tri *Trie) ([]ProofNode, error) {
 	return proofNodes, nil
 }
 
+func GetProofs(startKey, endKey *felt.Felt, tri *Trie) ([][]ProofNode, error) {
+	oneFelt := new(felt.Felt).SetUint64(1)
+	iterKey := startKey
+	leafRange := new(felt.Felt).Sub(endKey, startKey).Uint64()
+	proofs := make([][]ProofNode, leafRange)
+	for i := range leafRange {
```
asdacap, line 154: How does it iterate over uint64?

core/trie/proof.go
```diff
@@ -175,5 +192,129 @@ func VerifyProof(root *felt.Felt, key *Key, value *felt.Felt, proofs []ProofNode
...
+
+	}
+	return nil, errors.New("failed to get proof value")
+}
+
+// VerifyRangeProof verifies the range proof for the given range of keys.
+// ref: https://github.com/ethereum/go-ethereum/blob/v1.14.3/trie/proof.go#L484
+// Note: this currently assumes that the inner keys do not contain the min/max key (ie both proofs exist) // Todo
+// The first/last key and value must correspond to the left/right proofs //Todo we currently assume both proofs are provided, as above
+func VerifyRangeProof(root *felt.Felt, keys []*Key, values []*felt.Felt, proofs [2][]ProofNode, hash hashFunc) (bool, error) {
```
asdacap, line 243: Note, the `proofs` is a single array, not two. And the `keys` are `*felt.Felt`, meaning its always leaf. Well, I guess if you add a note on that, it does not matter if its a `Key` or `Felt`.

core/trie/proof.go
rianhughes:
```
someuint := uint(10)
for i := range someuint {
    fmt.Println(i)
}
```
prints 0 to 9

core/trie/proof.go
rianhughes: We only require two proofs no? Where each proof is a set of nodes from the root to a key.
Eg [geth](https://github.com/ethereum/go-ethereum/blob/master/trie/proof.go#L539) requires two proofs in their implementation. This is also how I understood it from watching Daniels snapsync video

rianhughes:
Note: I still need to account for edge cases (eg either single or no proof), but it would be useful to have a set of eyes on this now

core/trie/proof.go
```diff
@@ -172,8 +194,272 @@ func VerifyProof(root *felt.Felt, key *Key, value *felt.Felt, proofs []ProofNode
...
 	return expectedHash.Equal(value)
 }
+
+// VerifyRangeProof verifies the range proof for the given range of keys.
+// This is achieved by constructing a trie from the boundary proofs, and the supplied key-values.
+// If the root of the reconstructed trie matches the supplied root, then the verification passes.
+// If the trie is constructed incorrectly then the root will have an incorrect key(len,path), and value,
+// and therefore it's hash will be incorrect.
+// ref: https://github.com/ethereum/go-ethereum/blob/v1.14.3/trie/proof.go#L484
+func VerifyRangeProof(root *felt.Felt, keys, values []*felt.Felt, proofKeys, proofValues [2]*felt.Felt,
```
asdacap, line 210: Note, proof is 1 array, not 2. It got deduplicated into one.

rianhughes: A single proof is an array, but we need two proofs, right?

asdacap: Well, the easiest way to generate the proof is with the left (start query actually, probably will come back later) and right key separately. Then it is combined into one set. I'm guessing this is done because a lot of the time some of the top nodes are going to be duplicated.

core/trie/proof.go
```diff
@@ -172,8 +194,272 @@ func VerifyProof(root *felt.Felt, key *Key, value *felt.Felt, proofs []ProofNode
...
+		}
+		offset += squishedParent
+
+		// Set the value of the current node
+		crntNode.Value = proofNodes[i].Hash(hashF)
+
+		// Set the children of the current node
+		// The child may need compressed. If so, point to its compressed form.
+		childIdx := i + squishedParent + 1
+		if i+2+squishedParent < len(proofNodes)-1 {
```
pnowosie, line 347: This if-else here is pure magic - I cannot understand why we're looking at 3 different cases:
- `i+2+squishedParent < len(proofNodes)-1`
- `i+1+offset == len(proofNodes)-1`
- otherwise

also I feel me might want to refactor the `crntNode` construction into a functions which would do the `leafKey.Test(...)` because a lot of code is duplicated or similar and it's hard to follow what happens

pnowosie (changes requested):

core/trie/proof.go
```diff
@@ -87,7 +143,7 @@ func transformNode(tri *Trie, parentKey *Key, sNode storageNode) (*Edge, *Binary
 	}
 
 	rightHash := rNode.Value
-	if isEdge(sNode.key, storageNode{node: rNode, key: sNode.node.Right}) {
+	if isEdge(sNode.key, StorageNode{node: rNode, key: sNode.node.Right}) {
```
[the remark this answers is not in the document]
rianhughes: 2. Yes, the value is irrelevant here since we just want to know if the left/right child is an edge node, which we can tell using the length of the parent and child keys

core/trie/proof.go
rianhughes: I added some comments and refactored a bit. Let me know if it's still unclear!

core/trie/proof.go
rianhughes: Geth seems to store proofs nodes in a key-value store, so I guess starkware are thinking about dong something similar.

rianhughes: Mind if I open a separate issue for this?

rianhughes: https://github.com/NethermindEth/juno/issues/1898

core/trie/proof.go
rianhughes: Just an update: this logic is / will be simplified in [this](https://github.com/NethermindEth/juno/pull/1901/files) PR

rianhughes:
This PR is / can be replaced by this https://github.com/NethermindEth/juno/pull/1901

core/trie/proof.go
pnowosie: No need to push it here, we can review the changes in PR #1901

rianhughes:
closing for https://github.com/NethermindEth/juno/pull/1901

PR #1874, opened 2024-05-21
Fix feeder public_key decoding

kirugan:


clients/feeder/feeder.go
```diff
@@ -383,13 +383,12 @@ func (c *Client) PublicKey(ctx context.Context) (*felt.Felt, error) {
 	}
 	defer body.Close()
 
-	b, err := io.ReadAll(body)
-	if err != nil {
+	var publicKey string // public key hex string
+	if err = json.NewDecoder(body).Decode(&publicKey); err != nil {
```
kirugan, line 387: this is better because of semantic check while reading response body (i.e. if it's not a string it will fail at this step, not during unmarshaling to felt)

kirugan: before we could receive anything inside bytes, but now it should be fine

pnowosie (approved):
👍

PR #1879, opened 2024-05-23
DO_NOT_MERGE Exercise 6

ashu26jha:
New Sync package using feeder gateway

### DIFF
--- a/sync/sync.go
+++ b/sync/sync.go
@@ -3,6 +3,7 @@ package sync
 import (
 	"context"
 	"errors"
+	"net/http"
 	"runtime"
 	"sync/atomic"
 	"time"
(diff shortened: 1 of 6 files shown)


sync/sync.go
```diff
@@ -297,6 +307,138 @@ func (s *Synchronizer) syncBlocks(syncCtx context.Context) {
...
+	streamCtx, streamCancel := context.WithCancel(syncCtx)
+	for {
+		select {
+		case <-streamCtx.Done():
+			streamCancel()
+			return
+		default:
+			s.log.Infow("Fetching from feeder")
+			blocks, stateUpdate, blockCommitments := s.getBlockNumberDetails(uint64(s.latestBlockHeight))
+			newClass, _ := s.fetchUnknownClasses(syncCtx, &stateUpdate)
```
rianhughes, line 320: You should handle the error

rianhughes:
Generally it looks pretty good, although I would recommend fixing the tests.

sync/sync.go
ashu26jha: I will try to fix this through the manually, if it doesn't work I will switch to the starknetdata interface as you mentioned above!

ashu26jha:
Hey @rianhughes, did the changes  I am able to sync but there are couple of flaws I would like to point out:

1. Can't fix all the lints as http is not added to ignore 
2. It seems to run indefinitely
3. It only works with mainnet

Should I try fixing these or move on?

kirugan:
Closing since it was onboarding exercise

Issue #1758, opened 2024-03-21
Remove Goerli from Juno codebase


rianhughes:
Context: the Goerli network is being deprecated, but Juno tests still rely on it. The tests that rely on Goerli should be updated to use the Sepolia network instead.

RPC based tests that require Goerli:
- [ ] TestClassV0
- [ ] TestTransaction
- [ ] TestHeight
- [ ] TestBlockByNumberAndHash
- [ ] TestEvents
- [ ] TestChainId
- [ ] TestBlockTransactionCount
- [ ] TestBlockWithTxHashes
- [ ] TestEvents
- [ ] TestRpcBlockAdaptation

Note: Goerli is used in other non-rpc tests (sn2core, blockchian, cmd, etc). These should also be updated, but the RPC tests might be a good place for newcomers to start

stefanalfbo:
Hi, all the tests concerning the `rpc` package do not include any of the `goerli` networks anymore.

However the following tests are still using the `goerli` network as a reference:

- [ ] TestTransaction
  - [sn2core_test](https://github.com/NethermindEth/juno/blob/576d342d8049032fa09412c81155b41b5506101f/adapters/sn2core/sn2core_test.go#L265)
  - [feeder_test](https://github.com/NethermindEth/juno/blob/576d342d8049032fa09412c81155b41b5506101f/starknetdata/feeder/feeder_test.go#L106)
- [ ] TestEvents
  - [blockchain_test](https://github.com/NethermindEth/juno/blob/576d342d8049032fa09412c81155b41b5506101f/blockchain/blockchain_test.go#L421)
- [ ] TestDeclareV0TransactionHash
  - [transaction_test](https://github.com/NethermindEth/juno/blob/576d342d8049032fa09412c81155b41b5506101f/core/transaction_test.go#L334)
- [ ] TestBlockHash
  - [block_test](https://github.com/NethermindEth/juno/blob/576d342d8049032fa09412c81155b41b5506101f/core/block_test.go#L17)
- [ ] TestNetwork
  - [network_test](https://github.com/NethermindEth/juno/blob/576d342d8049032fa09412c81155b41b5506101f/utils/network_test.go#L23)
- [ ] TestCoreContractAddress
  - [network_test](https://github.com/NethermindEth/juno/blob/576d342d8049032fa09412c81155b41b5506101f/utils/network_test.go#L144)
- [ ] TestExecute
[...]

PR #1880, opened 2024-05-24
Remove decommissioned networks

### LINKED ISSUE #1753: Remove dependency on Goerli testnets from Juno environments

pnowosie:
Remove decommissioned networks: `Goerli`, `Goerli2` and `Integration`.

Fixes #1753 , https://github.com/NethermindEth/juno/issues/1758


.github/workflows/run-node-on-new-vm.yml
```diff
@@ -7,7 +7,7 @@ on:
         description: 'Network'
         required: false
         default: 'mainnet'
-        enum: ['mainnet', 'goerli', 'goerli2']
+        enum: ['mainnet', 'sepolia']
```
wojciechos, line 10: let's add sepolia-integration as an option here as well

core/state_update_test.go
```diff
@@ -12,33 +13,37 @@ import (
 )
 
 func TestStateDiffCommitment(t *testing.T) {
-	client := feeder.NewTestClient(t, &utils.Integration)
+	client := feeder.NewTestClient(t, &utils.Mainnet)
```
rianhughes, line 16: Why do we replace some Goerli/Integration tests with Mainnet tests, I would have expected them to use Sepolia data

rianhughes: Otherwise it looks good

pnowosie: My reasoning for such data based tests was to draw more on the mainnet data which likely not get off soon (so there will be no need to change the test then).

blockchain/blockchain_test.go
```diff
@@ -443,125 +443,6 @@ func TestState(t *testing.T) {
 	})
 }
 
-func TestEvents(t *testing.T) {
```
kirugan, line 446: Is this test obsolete?

core/block_test.go
```diff
@@ -21,61 +21,13 @@ func TestBlockHash(t *testing.T) {
...
-			chain:  utils.Goerli,
-			name:   "goerli network (post 0.7.0 without sequencer address)",
-		},
-		{
-			// block 1: goerli
-			// "https://alpha4.starknet.io/feeder_gateway/get_block?blockNumber=1",
-			number: 1,
-			chain:  utils.Goerli,
-			name:   "goerli network (pre 0.7.0 without sequencer address)",
-		},
```
kirugan, line 44: are you sure that these tests should be removed? because sequencerFallback is part of our logic in old hashes, maybe there is an alternative in mainnet ?

core/block_test.go
```diff
@@ -145,30 +91,6 @@ func TestBlockHash(t *testing.T) {
 			chain:  utils.Mainnet,
 			name:   "Block 192 with Failing l1 handler transaction version 1",
 		},
-		// "https://external.integration.starknet.io/feeder_gateway/get_block?blockNumber=283364"
-		{
-			number: 283364,
-			chain:  utils.Integration,
-			name:   "Block 283364 with Declare v2",
```
kirugan, line 152: same, specific tx type with version

core/block_test.go
```diff
@@ -204,12 +126,12 @@ func TestBlockHash(t *testing.T) {
 	})
 
 	t.Run("no error if block is unverifiable", func(t *testing.T) {
```
kirugan, line 128: new version of this test doesn't test what it was intended to. Here is metadata of goerli:
```
UnverifiableRange:        []uint64{119802, 148428},
```
119801 is beginning of unverifiable range

kirugan: If we don't need this logic anymore it's even better, but I'm not sure. cc @IronGauntlets

core/class_test.go
```diff
@@ -52,19 +52,24 @@ func TestClassV0Hash(t *testing.T) {
...
+	client := feeder.NewTestClient(t, &utils.Mainnet)
 	gw := adaptfeeder.New(client)
 	tests := []struct {
 		classHash       string
 		checkNoCompiled bool
 	}{
 		{
-			// https://external.integration.starknet.io/feeder_gateway/get_class_by_hash?classHash=0x1cd2edfb485241c4403254d550de0a097fa76743cd30696f714a491a454bad5
-			classHash: "0x1cd2edfb485241c4403254d550de0a097fa76743cd30696f714a491a454bad5",
+			// https://alpha-mainnet.starknet.io/feeder_gateway/get_class_by_hash?classHash=<calss_hash>
```
kirugan, line 62: missing class_hash

pnowosie: The url will be the same besides the `class_hash` parameter. I moved the comment above, and fixed the typo.

core/block_test.go
pnowosie: I've found already 2 mainnet examples in this test if I am not mistaken
https://github.com/NethermindEth/juno/blob/738356956709f4764b32ecbded7ce19e74311ac0/core/block_test.go#L31-L44

blockchain/blockchain_test.go
pnowosie: I moved it into a possibly better place
https://github.com/NethermindEth/juno/blob/c391201aba707d33fa7f7ae4315856a0ca3d26a0/rpc/events_test.go#L27

core/block_test.go
pnowosie: These one I looked for but found only version v3  - let me run my scripts over the weekend 🤔

pnowosie: add Sepolia block `#7`

rpc/transaction_test.go
```diff
@@ -649,28 +647,29 @@ func TestTransactionReceiptByHashV0_6(t *testing.T) {
...
+			"revert_reason":"Error in the called contract (0x037a10f2808c05f4a328bdac9a9344358547ae4676ebddc005e24ff887b188fd):\nError at pc=0:4273:\nGot an exception while executing a hint: Custom Hint Error: Execution failed. Failure reason: 0x54696d657374616d7020697320696e2074686520667574757265 ('Timestamp is in the future').\nCairo traceback (most recent call last):\nUnknown location (pc=0:67)\nUnknown location (pc=0:1997)\nUnknown location (pc=0:2727)\nUnknown location (pc=0:3582)\n",
+			"execution_resources":{"steps":16849,"pedersen_builtin_applications":16,"range_check_builtin_applications":174,"ec_op_builtin_applications":3}
 		}`
 
-		integClient := feeder.NewTestClient(t, &utils.Integration)
+		integClient := feeder.NewTestClient(t, &utils.Sepolia)
 		integGw := adaptfeeder.New(integClient)
 
-		blockWithRevertedTxn, err := integGw.BlockByNumber(context.Background(), 304740)
+		blockWithRevertedTxn, err := integGw.BlockByNumber(context.Background(), 19957)
```
pnowosie, line 669: I've found no difference in feeder response either from old & recents blocks.
So for these 2 new tests "reverted" & "v3 tx" I'm using the same block `19957` here 
(v0.6 api) and in the [following test](https://github.com/NethermindEth/juno/pull/1880/files#diff-92b1f47d3a0b6c346cb6e8bf24d9d0ac8b37a1c80adaa5e74922f9ab358b2ef5R715)

PR #1881, opened 2024-05-28
Add starknet.go tests to CI

wojciechos:


rianhughes:
Looks good, but I think we should wait until starknet.go makes a release where all RPC0.7.0 tests pass

derrix060:
I'd say to just start ignoring the tests that are failing, this way it will be avoing people introducing regression

derrix060:
for example adding a "xfail", so it's expected to fail until someone fixes it

Issue #903, opened 2023-07-11
Better sorting of buffered transactions in storage update


joshklop:
See [here](https://github.com/NethermindEth/juno/pull/900#discussion_r1258620454) for reference.

> Actually if we sort the slice of BufferedTransactions we get back from the routines in ascending contract address order, we might get an additional speed up. But it is not really worth it at this point. Will leave it to another PR.

cc @omerfirmak

kirugan:
Done in PR #1882

PR #1882, opened 2024-05-28
Sort buffered transactions

ashu26jha:
Implements #903, sorting the buffered txns in the increasing contract addresses. They do give slightly better benchmarking results

Benchmarks from the main branch:

```
ashutoshjha@Ashutoshs-MacBook-Air juno % go test ./... -bench=MainnetBlock -benchmem
PASS
ok      github.com/NethermindEth/juno/adapters/sn2core  0.920s
```

Benchmarks from current branch

```
ashutoshjha@Ashutoshs-MacBook-Air juno % go test ./... -bench=MainnetBlock -benchmem
PASS
ok      github.com/NethermindEth/juno/adapters/sn2core  0.727s
```


core/state.go
```diff
@@ -400,12 +400,16 @@ func (s *State) updateContractStorages(stateTrie *trie.Trie, diffs map[felt.Felt
...
+	contractUpdaters := pool.NewWithResults[*db.BufferedTransactionWithAddress]().WithErrors().WithMaxGoroutines(runtime.GOMAXPROCS(0))
 	for _, key := range keys {
-		conractAddr := key
-		updateDiff := diffs[conractAddr]
-		contractUpdaters.Go(func() (*db.BufferedTransaction, error) {
-			return s.updateStorageBuffered(&conractAddr, updateDiff, blockNumber, logChanges)
+		contractAddr := key
+		updateDiff := diffs[contractAddr]
+		contractUpdaters.Go(func() (*db.BufferedTransactionWithAddress, error) {
+			result, err := s.updateStorageBuffered(&contractAddr, updateDiff, blockNumber, logChanges)
```
kirugan, line 408: let's make it simple as before `return s.update....(...)`

core/state.go
```diff
@@ -414,11 +418,13 @@ func (s *State) updateContractStorages(stateTrie *trie.Trie, diffs map[felt.Felt
...
+		return bufferedTxns[i].Address.Cmp(bufferedTxns[j].Address) < 0
+	})
+
 	// flush buffered txns
-	for _, bufferedTxn := range bufferedTxns {
-		if err = bufferedTxn.Flush(); err != nil {
-			return err
-		}
+	for _, bufferedTxnsSlice := range bufferedTxns {
+		bufferedTxnsSlice.Txn.Flush()
```
kirugan, line 427: why there is no error check ?

db/buffered_transaction.go
```diff
@@ -10,13 +12,25 @@ type BufferedTransaction struct {
 	txn     Transaction
 }
 
+type BufferedTransactionWithAddress struct {
+	Txn     *BufferedTransaction
+	Address *felt.Felt
+}
+
```
IronGauntlets, line 19: I don't think we should have this struct because it doesn't offer much functionality over BufferedTransaction, there is no new method being added.

core/state.go
IronGauntlets, line 408: You already have access to the address when calling `s.updateStorageBuffered`, there is no need to return the address inside the result again.

core/state.go
```diff
@@ -400,12 +400,16 @@ func (s *State) updateContractStorages(stateTrie *trie.Trie, diffs map[felt.Felt
...
 	// update per-contract storage Tries concurrently
-	contractUpdaters := pool.NewWithResults[*db.BufferedTransaction]().WithErrors().WithMaxGoroutines(runtime.GOMAXPROCS(0))
+	contractUpdaters := pool.NewWithResults[*db.BufferedTransactionWithAddress]().WithErrors().WithMaxGoroutines(runtime.GOMAXPROCS(0))
 	for _, key := range keys {
-		conractAddr := key
-		updateDiff := diffs[conractAddr]
-		contractUpdaters.Go(func() (*db.BufferedTransaction, error) {
-			return s.updateStorageBuffered(&conractAddr, updateDiff, blockNumber, logChanges)
+		contractAddr := key
+		updateDiff := diffs[contractAddr]
```
IronGauntlets, line 406: Since this is only used once in the function there is no need for `updateDiff`, also since you are already making a shallow copy of the key in contractAddr it should be fine to have `diffs[contractAddr]` as the argument to `s.updatedStorageBuffered`.

core/state.go
```diff
@@ -414,11 +418,13 @@ func (s *State) updateContractStorages(stateTrie *trie.Trie, diffs map[felt.Felt
 		return err
 	}
 
+	sort.Slice(bufferedTxns, func(i, j int) bool {
```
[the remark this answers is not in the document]
IronGauntlets: you could create a private struct inside the function to allow you to use the `sort.Slice` function instead of declaring a `BufferedTransactionWithAddress` which is publically accessible.

IronGauntlets (changes requested):

db/buffered_transaction.go
ashu26jha: Will do this just you mentioned in above comments 👍

core/state.go
```diff
@@ -400,12 +405,16 @@ func (s *State) updateContractStorages(stateTrie *trie.Trie, diffs map[felt.Felt
 	})
 
 	// update per-contract storage Tries concurrently
-	contractUpdaters := pool.NewWithResults[*db.BufferedTransaction]().WithErrors().WithMaxGoroutines(runtime.GOMAXPROCS(0))
+	// contractUpdaters := pool.NewWithResults[*db.BufferedTransaction]().WithErrors().WithMaxGoroutines(runtime.GOMAXPROCS(0))
```
kirugan, line 408: should we remove this comment?

core/state.go
```diff
@@ -414,9 +423,14 @@ func (s *State) updateContractStorages(stateTrie *trie.Trie, diffs map[felt.Felt
...
 	}
 
+	sort.Slice(bufferedTxns, func(i, j int) bool {
+		return bufferedTxns[i].Address.Cmp(bufferedTxns[j].Address) < 0
+	})
+
 	// flush buffered txns
-	for _, bufferedTxn := range bufferedTxns {
-		if err = bufferedTxn.Flush(); err != nil {
+	for _, bufferedTxnsSlice := range bufferedTxns {
```
kirugan, line 431: `bufferedTxnsSlice` -> `txnWithAddress` ? (it's not a slice right?)

kirugan (approved):

IronGauntlets (approved):

PR #1883, opened 2024-05-29
Update network.go for broken docs link

tukideng:
The old URL https://docs.starknet.io/documentation/useful_info/ is returning 404 now.

It should be changed to https://docs.starknet.io/documentation/tools/important_addresses/, this PR does this changes.

By the way, should we remove all Goerli related content too, as starknet states:
> Goerli testnet support is now removed.


kirugan:
Closing the PR since corresponding changes already in main

Issue #1891, opened 2024-06-04
Update VerifyProof to use key rather than keyFelt


rianhughes:
Th proof logic currently only supports leaf keys, it should be updated so that we can generate and verify proofs of inner nodes (which requires using keys rather than felts). I will also add tests to generate and verify proofs of non-set leaf keys. This means that the Generate and Verify proof logic should be able to handle all scenarios.

PR #1893, opened 2024-06-04
Update GetProof and VerifyProof to use Keys

rianhughes:
closes https://github.com/NethermindEth/juno/issues/1891

### DIFF
--- a/core/trie/key.go
+++ b/core/trie/key.go
@@ -3,6 +3,7 @@ package trie
 import (
 	"bytes"
 	"encoding/hex"
+	"errors"
 	"fmt"
 	"math/big"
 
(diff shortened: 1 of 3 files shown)


core/trie/key.go
```diff
@@ -23,9 +24,9 @@ func NewKey(length uint8, keyBytes []byte) Key {
 	return k
 }
 
-func (k *Key) SubKey(n uint8) *Key {
+func (k *Key) SubKey(n uint8) (*Key, error) {
 	if n > k.len {
-		panic("n is greater than the length of the key")
+		return nil, errors.New(fmt.Sprint("cannot subtract key by len %i", n))
```
pnowosie, line 29:
```suggestion
		return nil, errors.New(fmt.Sprint("cannot subtract key of %i from key of length=%i", n, k.len))
```

core/trie/proof.go
```diff
@@ -113,18 +113,18 @@ func transformNode(tri *Trie, parentKey *Key, sNode storageNode) (*Edge, *Binary
...
 		return nil, err
 	}
 	proofNodes := []ProofNode{}
 
 	var parentKey *Key
 
-	for i := 0; i < len(nodesToLeaf); i++ {
-		sNode := nodesToLeaf[i]
+	for i := 0; i < len(nodesFromRoot); i++ {
+		sNode := nodesFromRoot[i]
```
pnowosie, line 126:
```suggestion
	for i, sNode := range nodesFromRoot {
```

pnowosie (approved):
PR title says that `VerifyProof` was changed which isn't the case

Issue #1895, opened 2024-06-05
VerifyProofRange non-set and inner nodes


rianhughes:
Once https://github.com/NethermindEth/juno/pull/1893 and https://github.com/NethermindEth/juno/pull/1873 are merged, then implement a set of tests / any logic to handle the cases where the boundary keys are not set

Issue #1898, opened 2024-06-07
Update proof logic to use merged proofs


rianhughes:
Currently we handle left and right proofs separately but it [seems](https://github.com/starknet-io/starknet-p2p-specs/blob/main/p2p/proto/snapshot.proto#L24) that we need to merge the proofs. Implement the logic to merge and de-merge the proofs

You should implement two functions. 

`func mergeProofPaths(leftPath,rightPath []ProofNode) (mergedPath []ProofNode,error)`
leftPath and rightPath are lists of ProofNodes, which define paths from the root node to some leaf
mergedPath represents the merged list of leftPath and rightPath, with no duplicates.

`func splitProofPath(mergedPath []ProofNode) (leftPath,rightPath []ProofNode,error)`
This function should do the reverse of mergeProofPaths, ie take a mergedPath, and split it into leftPath and rightPath.

It might be useful to look at [this](https://docs.starknet.io/architecture-and-concepts/network-architecture/starknet-state/#example_trie), and an example trie [here](https://github.com/NethermindEth/juno/blob/main/core/trie/proof_test.go#L211-L220). Note that ProofNodes are either Binary or Edge.

PR #1901, opened 2024-06-07
updating verify proof range to handle empty proof keys

rianhughes:
This PR implements VerifyRangeProof.

It builds on an old PR (1873) which didn't handle empty proof keys, hence the name of this PR.

closes https://github.com/NethermindEth/juno/issues/1895
replaces / closes: https://github.com/NethermindEth/juno/pull/1873


core/trie/proof.go
```diff
@@ -165,12 +189,328 @@ func VerifyProof(root *felt.Felt, key *Key, value *felt.Felt, proofs []ProofNode
...
+		parentNode.Left = childKey
+	}
+	return childKey, nil
+}
+
+// ProofToPath returns a set of storage nodes from the root to the end of the proof path.
+// It will contain the hashes of the children along the path, but only the key of the children
+// along the path. The final node must contain the hash of the leaf if the leaf is set.
+// It will not contain the leaf node even if it is set.
+func ProofToPath(proofNodes []ProofNode, leafKey *Key, hashF hashFunc) ([]StorageNode, error) {
```
pnowosie, line 344: The code is much more readable 🎉  Great job!

Now, I cannot understand what is written in the fn's comment.
So we are given `proofNodes` and constructing `storageNodes` following along the path  to the single leaf node specified by given `leafKey`.
These `storageNodes` will have hashes of the children and keys? - not get this  `"but only the key of the children along the path"`.

Also: `"The final node must contain the hash of the leaf ..."` - which node is the final node, root or the nodes at the bottom of the trie 🤔 

And: `"It will not contain the leaf node even if it is set."`
Hmm, the leaf is set because we have the `leafKey` right? Also the leaf node refers to the `storageNodes` we are constructing here.

Can you please try to improve this comment?

core/trie/proof.go
```diff
@@ -165,12 +189,328 @@ func VerifyProof(root *felt.Felt, key *Key, value *felt.Felt, proofs []ProofNode
...
+	return false
+}
+
+func getLeftRightHash(parentInd int, proofNodes []ProofNode) (*felt.Felt, *felt.Felt) {
+	parent := &proofNodes[parentInd]
+	if parent.Binary != nil {
+		parentBinary := parent.Binary
+		return parentBinary.LeftHash, parentBinary.RightHash
+	} else {
+		parentBinary := proofNodes[parentInd+1].Binary
```
pnowosie, line 420: Don't we need ro check that `proofNodes[parentInd+1].Binary != nil` as we did with the previous node?
also `parentInd+1 < len(proofNodes)` 🤔 

Maybe we should assign the `parent` and have a single return:

```
parent := &proofNodes[parentInd]
if parent.Binary == nil {
    parent := &proofNodes[parentInd+1]
}
return parent.Binary.LeftHash, parent.Binary.RightHash
```

rianhughes: > Don't we need ro check that proofNodes[parentInd+1].Binary != nil as we did with the previous node?
Nope, if proofNode[i] is an edge then we know proofNode[i+1] is Binary (it's not possible for an edge node to follow an edge node).

> Maybe we should assign the parent and have a single return
Yup, this is much nicer!

> parentInd+1 < len(proofNodes)
Good point. Let me think about that

core/trie/proof.go
rianhughes: > These storageNodes will have hashes of the children and keys? - not get this "but only the key of the children along the path".

The storageNodes will have the hashes of the children, but only the key of the child along the path outlined by the proof.

> Also: "The final node must contain the hash of the leaf ..." - which node is the final node, root or the nodes at the bottom of the trie 🤔

Ah sorry, this is a confusing sentence. I was just trying to comment that we don't include the leaf node in the generated storage nodes. But if the path outlined by the proof leads the whole way to the leaf, and the leaf is set, then we should include it's hash.

> Hmm, the leaf is set because we have the leafKey right?

Nope. The Key that the proof node points to may not be set, but we can still have the key.

>  Also the leaf node refers to the storageNodes we are constructing here.

Yup

rianhughes: I'll update the comment

core/trie/proof.go
rianhughes: >    parentInd+1 < len(proofNodes)

I added a check for this

core/trie/proof.go
```diff
@@ -165,12 +189,329 @@ func VerifyProof(root *felt.Felt, key *Key, value *felt.Felt, proofs []ProofNode
...
 }
+
+// VerifyRangeProof verifies the range proof for the given range of keys.
+// This is achieved by constructing a trie from the boundary proofs, and the supplied key-values.
+// If the root of the reconstructed trie matches the supplied root, then the verification passes.
+// If the trie is constructed incorrectly then the root will have an incorrect key(len,path), and value,
+// and therefore it's hash won't match the expected root.
+// ref: https://github.com/ethereum/go-ethereum/blob/v1.14.3/trie/proof.go#L484
+func VerifyRangeProof(root *felt.Felt, keys, values []*felt.Felt, proofKeys [2]*Key, proofValues [2]*felt.Felt,
+	proofs [2][]ProofNode, hash hashFunc,
```
asdacap, line 218: What is `proofKeys` and `proofValues`? What is the difference from `proofs`?

rianhughes: Proofs are the set of nodes from the root to the boundary / proof keys and values

rianhughes: I explicitly labelled the proof keys and values for clarity (I find this is a bit easier to understand than placing the proof key as the first element in keys iff the proof key has actually been set - like is done in [geth](https://github.com/ethereum/go-ethereum/blob/master/trie/proof.go#L450))

core/trie/node.go
```diff
@@ -45,18 +53,38 @@ func (n *Node) WriteTo(buf *bytes.Buffer) (int64, error) {
...
 	}
 
+	if n.LeftHash == nil && n.RightHash == nil {
+		return totalBytes, nil
+	} else if (n.LeftHash != nil && n.RightHash == nil) || (n.LeftHash == nil && n.RightHash != nil) {
+		return totalBytes, errors.New("cannot store only one lefthash or righthash")
+	}
+
+	leftHashB := n.LeftHash.Bytes()
+	wrote, err = buf.Write(leftHashB[:])
```
asdacap, line 75: This seems like an encoding change. Is this intended?

core/trie/proof.go
rianhughes: > the left proof should be generated based on the start key query (not actual first leaf that exist)
Currently the left proof key would just correspond to the start key query. And the last proofKey would correspond to the end key query. Similarly with the proof Values (which can be nil if they don't exist)

> Remember, only one proof array, ..
Yup, but that should be a simple matter of merging the proof nodes into a single array, and then unsplitting them again, right? I've opened an [issue](https://github.com/NethermindEth/juno/issues/1898) for this

rianhughes: Also, regarding there only being a single proof, geth does store the proof nodes in a key-value database, but I believe there are two boundary proofs, which you can see [here](https://github.com/ethereum/go-ethereum/blob/v1.14.3/trie/proof.go#L548-L569). 

Afaiu, when there are key-value pairs to the left and right of the given range we need two boundary proofs. This is because we need to compute the correct shape of the sub-trie, and store the relevant hashes so that we can recompute the root hash for that sub trie. Having only a single boundary proof would no allow us to do this.

rianhughes:
@asdacap, @pnowosie  in geth, the VerifyRangeProof executes the verification logic, but also logic to determine if there are more leaves to the right (see [here](https://github.com/ethereum/go-ethereum/blob/v1.14.3/trie/proof.go#L478-L479)). Imo it would be neater to separate these out into two different functions, hence why I didn't insert it in this implementation of VerifyRangeProof. What are your thoughts?

asdacap:
You can separate them internally if you want. But at high level, I think its easier to combine them, otherwise, need to pass the keys of proof explicitly, which is implementation details IMO.

asdacap (approved):

pnowosie (approved):
✅ Great

PR #1902, opened 2024-06-10
Add database CLI tools

weiihann:
### Description
This PR adds a new CLI tool for database-related operations (i.e. `juno db [commands]`). Under the `db` category, there are 2 subcommand:
1. `juno db info` - display the blockchain information stored in the database
```
{
  "network": "sepolia",
  "chain_height": 71958,
  "latest_block_hash": "0x69fa4f792a98cab20d5fe5040401a8377ae75234ee2446590e1b68d3abfa9e2",
  "latest_state_root": "0x7c5dd3a676f5b6f603227b1f1b77f8d95a96b05abc93c09f0715d27c7169f1c",
  "l1_height": 66367,
  "l1_block_hash": "0x335b4136b1da8317a1a72450ddb2df813ae828cb2ee4347d4c21d2d8adee0dc",
  "l1_state_root": "0x4ff66961f76f33cedc65124cbb3021e370cd966434c13245f1d75015e136ba8"
}
```
3. `juno db size` - display storage information for each bucket in the db
```
Calculating size of StateTrie, remaining buckets: 23
Calculating size of Peer, remaining buckets: 22
Calculating size of ContractClassHash, remaining buckets: 21
Calculating size of ContractStorage, remaining buckets: 20
Calculating size of Class, remaining buckets: 19
Calculating size of ContractNonce, remaining buckets: 18
Calculating size of ChainHeight, remaining buckets: 17
Calculating size of BlockHeaderNumbersByHash, remaining buckets: 16
Calculating size of BlockHeadersByNumber, remaining buckets: 15
Calculating size of TransactionBlockNumbersAndIndicesByHash, remaining buckets: 14
Calculating size of TransactionsByBlockNumberAndIndex, remaining buckets: 13
Calculating size of ReceiptsByBlockNumberAndIndex, remaining buckets: 12
Calculating size of StateUpdatesByBlockNumber, remaining buckets: 11
Calculating size of ClassesTrie, remaining buckets: 10
Calculating size of ContractStorageHistory, remaining buckets: 9
Calculating size of ContractNonceHistory, remaining buckets: 8
Calculating size of ContractClassHashHistory, remaining buckets: 7
Calculating size of ContractDeploymentHeight, remaining buckets: 6
Calculating size of L1Height, remaining buckets: 5
Calculating size of SchemaVersion, remaining buckets: 4
Calculating size of Pending, remaining buckets: 3
Calculating size of BlockCommitments, remaining buckets: 2
Calculating size of Temporary, remaining buckets: 1
Calculating size of SchemaIntermediateState, remaining buckets: 0
+-----------------------------------------+------------+----------+
|                 BUCKET                  |    SIZE    |  COUNT   |
+-----------------------------------------+------------+----------+
| StateTrie                               | 9.44 MiB   |   143032 |
| Peer                                    | 0.00 B     |        0 |
| ContractClassHash                       | 4.43 MiB   |    71516 |
| ContractStorage                         | 662.14 MiB |  5970298 |
| Class                                   | 5.80 GiB   |    11139 |
| ContractNonce                           | 4.43 MiB   |    71516 |
| ChainHeight                             | 9.00 B     |        1 |
| BlockHeaderNumbersByHash                | 2.88 MiB   |    73536 |
| BlockHeadersByNumber                    | 115.86 MiB |    73536 |
| TransactionBlockNumbersAndIndicesByHash | 91.42 MiB  |  1956435 |
| TransactionsByBlockNumberAndIndex       | 4.03 GiB   |  1956435 |
| ReceiptsByBlockNumberAndIndex           | 4.82 GiB   |  1956435 |
| StateUpdatesByBlockNumber               | 688.26 MiB |    73536 |
| ClassesTrie                             | 1.14 MiB   |    17370 |
| ContractStorageHistory                  | 848.86 MiB |  8477125 |
| ContractNonceHistory                    | 43.72 MiB  |   628061 |
| ContractClassHashHistory                | 57.53 KiB  |      807 |
| ContractDeploymentHeight                | 2.80 MiB   |    71516 |
| L1Height                                | 0.00 B     |        0 |
| SchemaVersion                           | 9.00 B     |        1 |
| Pending                                 | 1.56 MiB   |        1 |
| BlockCommitments                        | 8.56 MiB   |    73536 |
| Temporary                               | 0.00 B     |        0 |
| SchemaIntermediateState                 | 2.00 B     |        1 |
+-----------------------------------------+------------+----------+
|                  TOTAL                  | 17.08 GIB  | 21625833 |
+-----------------------------------------+------------+----------+
+-----------------+----------+----------+
|      STATE      |   SIZE   |  COUNT   |
+-----------------+----------+----------+
| Without history | 6.46 GiB |  6267501 |
| With history    | 7.33 GiB | 15373494 |
+-----------------+----------+----------+
```


kirugan:
Can we also show to which network db corresponds to? it would be great if there is some snapshot stored without identification

weiihann:
> Can we also show to which network db corresponds to? it would be great if there is some snapshot stored without identification

I think that's a good addition. From my understanding, the network ID isn't stored directly in the db. Let me figure out how to pull that information.

kirugan:
> > Can we also show to which network db corresponds to? it would be great if there is some snapshot stored without identification
> 
> I think that's a good addition. From my understanding, the network ID isn't stored directly in the db. Let me figure out how to pull that information.

That's right you can modify existing approach for your case:
```
// Verify that cfg.Network is compatible with the database.
	head, err := chain.Head()
	if err != nil && !errors.Is(err, db.ErrKeyNotFound) {
		return nil, fmt.Errorf("get head block from database: %v", err)
	}
	if head != nil {
		// We assume that there is at least one transaction in the block or that it is a pre-0.7 block.
		if _, err = core.VerifyBlockHash(head, &cfg.Network); err != nil {
			return nil, errors.New("unable to verify latest block hash; are the database and --network option compatible?")
		}
	}
```
You should iterate over networks to figure out (`core.VerifyBlockHash(...) == nil`) to which one it corresponds

weiihann:
> You should iterate over networks to figure out (`_, err := core.VerifyBlockHash(...); err == nil`) to which one it corresponds
Added the network field.

clients/feeder/feeder.go
```diff
@@ -128,10 +128,8 @@ func newTestServer(t *testing.T, network *utils.Network) *httptest.Server {
 		assert.Equal(t, []string{"API_KEY"}, r.Header["X-Throttling-Bypass"])
 		assert.Equal(t, []string{"Juno/v0.0.1-test Starknet Implementation"}, r.Header["User-Agent"])
 
-		wd, err := os.Getwd()
```
weiihann, line 131: a

clients/feeder/feeder.go
```diff
@@ -128,10 +128,8 @@ func newTestServer(t *testing.T, network *utils.Network) *httptest.Server {
...
 		require.NoError(t, err)
 
-		base := wd[:strings.LastIndex(wd, "juno")+4]
```
weiihann, line 134: Also I did a small refactor on the `newTestServer` function. Essentially we shouldn't retrieve the directory path with the hardcoded "juno" string. There are 2 issues with this:

1. The directory name might not even be "juno". Users may choose to set some arbitrary directory name.
2. We have a directory of `juno/cmd/juno`, since `strings.LastIndex` returns the last index at which a substring occurs, it will return `juno/cmd/juno` instead of just `juno`. The simple solution to this is just use `strings.Index`

Both of the above scenarios will cause the tests to fail. The refactor solution is to traverse up the directory tree from the current working directory until the target directory is found, solving the issues above.

weiihann:
TODO for myself: do not find the target directory using working directory, binary will not be able to run anywhere. Perhaps use embed directive.

weiihann:
Calculating DB size has some differences between this implementation and https://github.com/NethermindEth/juno/pull/1773:
1. Formats the data size into metric prefixes (i.e. MiB, GiB, etc)
2. Adds count of each KV pair
3. Outputs result in table format
4. Put this subcommand under the subcommand of `db` `./juno db size` instead of `./juno db-size`

weiihann:
Tests fail because https://github.com/NethermindEth/juno/pull/1978 has to be merged.

cmd/juno/dbcmd.go
```diff
@@ -0,0 +1,202 @@
...
+
+func getNetwork(head *core.Block, stateDiff *core.StateDiff) string {
+	networks := map[string]*utils.Network{
+		"mainnet":             &utils.Mainnet,
+		"sepolia":             &utils.Sepolia,
+		"goerli":              &utils.Goerli,
+		"goerli2":             &utils.Goerli2,
+		"integration":         &utils.Integration,
+		"sepolia-integration": &utils.SepoliaIntegration,
+	}
```
kirugan, line 193: you don't need map here, just `[]utils.Network{ ...}` name of a network can be retrieved through `Name` field

utils/size.go
```diff
@@ -0,0 +1,22 @@
...
+type DataSize float64
+
+//nolint:mnd
+func (d DataSize) String() string {
+	switch {
+	case d >= 1099511627776:
+		return fmt.Sprintf("%.2f TiB", d/1099511627776)
+	case d >= 1073741824:
+		return fmt.Sprintf("%.2f GiB", d/1073741824)
+	case d >= 1048576:
```
kirugan, line 16: let's replace it with constants from const.go (there is only 1 for Megabyte, but you can add the rest)

utils/const.go
```diff
@@ -1,5 +1,8 @@
 package utils
 
 const (
+	KiloByte = 1024
```
kirugan, line 4: Kilo**b**yte ?

utils/const.go
```diff
@@ -1,5 +1,8 @@
...
 	Megabyte = 1024 * 1024
```
kirugan, line 5: you can refer to other constants:
```
Megabyte = 1024 * Kilybyte
Gigabyte = 1024 * Megabyte
....
```

db/pebble/db_test.go
```diff
@@ -425,7 +426,8 @@ func TestCalculatePrefixSize(t *testing.T) {
 
 		s, err := pebble.CalculatePrefixSize(context.Background(), testDB, []byte("0"))
 		require.NoError(t, err)
-		assert.Equal(t, uint(0), s)
+		assert.Equal(t, uint(0), s.Count)
+		assert.Equal(t, utils.DataSize(0), s.Size)
```
kirugan, line 430: do you think `assert.Zero(...)` is more clear way to check that?

kirugan: same for rest of cases in this file below

cmd/juno/dbcmd_test.go
```diff
@@ -0,0 +1,52 @@
...
+	client := feeder.NewTestClient(t, &utils.Mainnet)
+	gw := adaptfeeder.New(client)
+	block0, err := gw.BlockByNumber(context.Background(), 0)
+	require.NoError(t, err)
+
+	stateUpdate0, err := gw.StateUpdate(context.Background(), 0)
+	require.NoError(t, err)
+
+	dbPath := t.TempDir()
+	testDB, _ := pebble.New(dbPath)
```
kirugan, line 45: do we need to check for error here?

cmd/juno/dbcmd_test.go
```diff
@@ -0,0 +1,52 @@
...
+	chain := blockchain.New(testDB, &utils.Mainnet)
+	assert.NoError(t, chain.Store(block0, &emptyCommitments, stateUpdate0, nil))
```
kirugan, line 47: `require.NoError(...)` ?

cmd/juno/dbcmd_test.go
kirugan: btw why can't we use `pebble.NewMemTest(...)` here ?

cmd/juno/dbcmd.go
```diff
@@ -0,0 +1,202 @@
...
+}
+
+func dbInfo(cmd *cobra.Command, args []string) error {
+	dbPath, err := cmd.Flags().GetString(dbPathF)
+	if err != nil {
+		return err
+	}
+
+	if _, err = os.Stat(dbPath); os.IsNotExist(err) {
+		fmt.Println("Database path does not exist")
```
kirugan, line 64: Maybe replace it `fmt.Fprintln(cmd.OutOrStdout(), <string>)`

cmd/juno/dbcmd.go
```diff
@@ -0,0 +1,202 @@
...
+	} else {
+		fmt.Printf("Failed to get the latest L1 block information: %v\n", err)
+	}
+
+	jsonData, err := json.Marshal(info)
+	if err != nil {
+		return fmt.Errorf("marshal JSON: %w", err)
+	}
+
+	fmt.Println(string(jsonData))
```
kirugan, line 107: `fmt.Fprintln(cmd.OutOrStdout(), ...)` ?

cmd/juno/dbcmd.go
```diff
@@ -0,0 +1,202 @@
...
+	if err != nil {
+		return err
+	}
+
+	if dbPath == "" {
+		return fmt.Errorf("--%v cannot be empty", dbPathF)
+	}
+
+	if _, err = os.Stat(dbPath); os.IsNotExist(err) {
+		fmt.Println("Database path does not exist")
```
kirugan, line 123: same as above

cmd/juno/dbcmd.go
```diff
@@ -0,0 +1,202 @@
...
+		items [][]string
+	)
+
+	for _, b := range db.BucketValues() {
+		fmt.Fprintf(cmd.OutOrStdout(), "Calculating size of %s, remaining buckets: %d\n", b, len(db.BucketValues())-int(b)-1)
+		bucketItem, err := pebble.CalculatePrefixSize(cmd.Context(), pebbleDB.(*pebble.DB), []byte{byte(b)})
+		if err != nil {
+			return err
+		}
+		items = append(items, []string{b.String(), bucketItem.Size.String(), fmt.Sprintf("%d", bucketItem.Count)})
```
kirugan, line 151: you don't need to create temp. slice i.e.:
```
items = append(items, b.String(), bucketItem.Size.String(), fmt.Sprintf("%d", bucketItem.Count))
```

weiihann: `items` is `[][]string`, so I had to put the list of strings in `[]string{}` before appending it.

It's used in `table.Appendbulk`.

cmd/juno/dbcmd_test.go
weiihann: since we are testing the CLI, we need an actual db path that can be resolved and opened.

cmd/juno/dbcmd.go
kirugan: I see, makes sense

kirugan (approved):

cmd/juno/dbcmd.go
```diff
@@ -0,0 +1,202 @@
...
+	l1Head, err := chain.L1Head()
+	if err == nil {
+		info.L1Height = l1Head.BlockNumber
+		info.L1BlockHash = l1Head.BlockHash
+		info.L1StateRoot = l1Head.StateRoot
+	} else {
+		fmt.Printf("Failed to get the latest L1 block information: %v\n", err)
+	}
+
+	jsonData, err := json.Marshal(info)
```
rianhughes, line 102: nit: jsonData, err := json.MarshalIndent(info, "  ", "  ")

cmd/juno/dbcmd.go
```diff
@@ -0,0 +1,202 @@
...
+	return &cobra.Command{
+		Use:   "size",
+		Short: "Calculate database size information for each data type",
+		Long:  `This subcommand retrieves and displays the storage of each data type stored in the database.`,
+		RunE:  dbSize,
+	}
+}
+
+func dbInfo(cmd *cobra.Command, args []string) error {
+	dbPath, err := cmd.Flags().GetString(dbPathF)
```
rianhughes, line 58: What do you think about checking the default path $HOME/juno if no flag is set?

kirugan: do we store it at this location if user doesn't specify `--db-path` ?

weiihann: The default db path (for db-related subcommands) is `""`.

I added a change to pass the default db path from `junoCmd` to `dbCmd`.

rianhughes (approved):

PR #1903, opened 2024-06-12
Get `AddressRange` & `ContractRange`

ashu26jha:
Currently this PR only covers `AddressRange`, `ContractRange` is mostly based upon the `AddressRange`. Will be adding `ContractRange` after the tests and correcting the implementation if needed.

This is mostly based this [branch](https://github.com/NethermindEth/juno/tree/proto/p2p-snap-sync)


IronGauntlets:
why is the child hash logic change?

IronGauntlets:
Also, how much of this code is Ashraf's?

ashu26jha:
> Also, how much of this code is Ashraf's?

Most of code is based on Ashraf's work (the branch I added in description)
I have fixed it work of new ProofNode definitions, fixed few Todo's. I have to write tests also

PR #1904, opened 2024-06-12
File path support for p2p private key

weiihann:
### Description
This PR adds filepath support to two existing P2P-related tools and flags, namely `genp2pkeypair` and `--p2p-private-key`.

### Example
Command: `juno genp2pkeypair --file nodekey`
- This would generate a file containing the key pairs information.
- Note that users can still do `juno genp2pkeypair` just like before.

Command: `juno --p2p-private-key nodekey`
- This would read the private key given in the `nodekey` file.
- Note that users can also just parse the private key (i.e. `juno --p2p-private-key <private-key>`) just like before.
- The content of the file should either be just the private key or contains the "P2P Private Key:" prefix. Examples are shown below:
```
# Example 1: Pure private key
<private-key>

# Example 2: With prefix
P2P Private Key: <private-key>
```

### Rationale
It's more of a UX nicety to allow users to generate key pairs in a file and parse the file to use as a private key, rather than having to copy pasting manually.

### DIFF
--- a/cmd/juno/juno.go
+++ b/cmd/juno/juno.go
@@ -253,7 +253,7 @@ func NewCmd(config *node.Config, run func(*cobra.Command, []string) error) *cobr
 		v.SetEnvPrefix("JUNO")
 		v.SetEnvKeyReplacer(strings.NewReplacer(".", "_", "-", "_"))
 		if err := v.BindPFlags(cmd.Flags()); err != nil {
-			return nil //nolint:nilerr
+			return nil
 		}
 
 		// TextUnmarshallerHookFunc allows us to unmarshal values that satisfy the
--- a/cmd/juno/juno.go
+++ b/cmd/juno/juno.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"bufio"
 	"context"
 	"fmt"
 	"math"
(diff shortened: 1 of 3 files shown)


cmd/juno/juno.go
```diff
@@ -148,7 +149,7 @@ const (
 		"These peers can be either Feeder or regular nodes."
 	p2pFeederNodeUsage = "EXPERIMENTAL: Run juno as a feeder node which will only sync from feeder gateway and gossip the new" +
 		" blocks to the network."
-	p2pPrivateKeyUsage   = "EXPERIMENTAL: Hexadecimal representation of a private key on the Ed25519 elliptic curve."
+	p2pPrivateKeyUsage   = "EXPERIMENTAL: Hexadecimal representation of a private key on the Ed25519 elliptic curve. Also accepts a file path."
```
IronGauntlets, line 152: I am unsure whether it is a good idea to have a flag accept 2 different input types.

weiihann: I see. My thought process is having one flag can abstract out the input process.

That could be confusing. Perhaps it's better to have a separate flag to accept file path for private key. Let me work on that.

derrix060 (approved):

derrix060 (changes requested):
There are some conflicts... @weiihann could you rebase and fix those ? :)

weiihann:
On a second thought, I find this feature irrelevant as users can just use environment variables.

PR #1905, opened 2024-06-13
cmd/juno: add --instance flag

weiihann:
### Description
This PR adds a new flag called `--instance` to initialize port numbers automatically based on the instance count. This flag affects the following ports:
- `http-port`
- `grpc-port`
- `metrics-port`
- `pprof-port`
- `ws-port` 

### Example
Each instance count will increment the default value of the respective ports by 10. For example:

The default value of each port is as follows:
- `grpc-port` = 6064
- `http-port` = 6060
- `metrics-port` = 9090
- `pprof-port` = 6062
- `ws-port` = 6061

Command: `./juno --instance 2`
Then the ports would have the following numbers:
- `grpc-port` = 6074
- `http-port` = 6070
- `metrics-port` = 9100
- `pprof-port` = 6072
- `ws-port` = 6071

### Rationale
This feature is useful for initializing multiple instances of Juno nodes, without having to care for the port conflict issues.

### DIFF
--- a/cmd/juno/juno.go
+++ b/cmd/juno/juno.go
@@ -118,7 +118,7 @@ const (
 	defaultCallMaxSteps             = 4_000_000
 	defaultGwTimeout                = 5 * time.Second
 	defaultCorsEnable               = false
-	defaultInstance                 = 1
+	defaultInstance                 = 0
 	defaultInstanceInc              = 10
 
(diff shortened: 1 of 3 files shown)


weiihann:
Closing this, additional flags are harder to maintain.

PR #1906, opened 2024-06-14
node: include semantic version in agent version

weiihann:
This PR adds the semantic version of the node to the user agent version such that it follows the appropriate format, as shown in the [Starknet P2P specifications](https://github.com/starknet-io/starknet-p2p-specs/blob/main/p2p/starknet-p2p.md#identifying-nodes).

**Before**: `juno`
**After**: `juno/<semantic-version>`

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -7,6 +7,7 @@ import (
 	"net/url"
 	"reflect"
 	"runtime"
+	"strings"
 	"time"
 
 	"github.com/Masterminds/semver/v3"
@@ -40,6 +41,7 @@ const (
 	upgraderDelay    = 5 * time.Minute
 	githubAPIUrl     = "https://api.github.com/repos/NethermindEth/juno/releases/latest"
 	latestReleaseURL = "https://github.com/NethermindEth/juno/releases/latest"
+	clientName       = "juno"
 )
 
 // Config is the top-level juno configuration.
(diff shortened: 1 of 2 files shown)


node/node.go
```diff
@@ -359,3 +362,12 @@ func (n *Node) Run(ctx context.Context) {
 func (n *Node) Config() Config {
 	return *n.cfg
 }
+
+func parseVersion(version string, log *utils.ZapLogger) string {
```
kirugan, line 366: why do we need this function?

weiihann: The function mainly targets the following scenarios:

1. Versions unobtainable due to missing tags.
2. Version string by default comes with the number of commits and the commit ID (e.g. `v0.11.9-36-g11edfa86`). The function should sanitize it so that it becomes a proper semantic version format.

node/node.go
```diff
@@ -369,3 +372,12 @@ func (n *Node) Run(ctx context.Context) {
...
+	semVer, err := semver.NewVersion(version)
+	if err != nil {
+		log.Warnw("Failed to parse Juno version", "version", version, "err", err)
+		return "0.0.0"
+	}
+	return strings.Split(semVer.String(), "-")[0]
```
kirugan, line 382: IMO it would be more clear if it's:
```
return fmt.Sprintf("%d.%d.%d", semVer.Major(), semver.Minor(), semver.Patch())
```

node/node.go
kirugan: since this logic is mostly p2p related, wdyt about moving this function there and passing `version` string in `p2p.New(...)` ?

node/node.go
```diff
@@ -369,3 +372,12 @@ func (n *Node) Run(ctx context.Context) {
 func (n *Node) Config() Config {
 	return *n.cfg
 }
+
+func parseVersion(version string, log *utils.ZapLogger) string {
+	semVer, err := semver.NewVersion(version)
+	if err != nil {
+		log.Warnw("Failed to parse Juno version", "version", version, "err", err)
```
kirugan, line 379: I was hesitating, but let's not log this error, my arguments:
* we already ignore version string validity in `juno_version` method
* it's mostly static error i.e. it's passed during build time, user has no control over it and we will probably find it before release

kirugan (approved):

IronGauntlets (approved):
lgtm, 

Please make sure to modify the commit message when squashing to follow [git commit message guidelines ](https://cbea.ms/git-commit/).

PR #1909, opened 2024-06-17
Onboarding exercises

artem-streltsov:
Do not merge

### DIFF
--- a/core/trie/trie.go
+++ b/core/trie/trie.go
@@ -101,15 +101,15 @@ func (t *Trie) GetNodesFromRoot(key *Key) ([]storageNode, error) {
 }
 
 func (t *Trie) ParseNodes(nodes []storageNode) ([]map[string]string, error) {
-    result := make([]map[string]string, len(nodes))
-    for i, node := range nodes {
-        result[i] = map[string]string{
-            "key":   fmt.Sprintf("%v", node.key),
-            "value": fmt.Sprintf("%v", node.node.Value),
-        }
-    }
+	result := make([]map[string]string, len(nodes))
+	for i, node := range nodes {
+		result[i] = map[string]string{
+			"key":   node.key.String(),
+			"value": node.node.Value.String(),
+		}
+	}
 
-    return result, nil
+	return result, nil
 }
 
 // findCommonKey finds the set of common MSB bits in two key bitsets.
--- a/core/trie/trie.go
+++ b/core/trie/trie.go
@@ -97,7 +97,7 @@ func (t *Trie) ConvertFeltToKey(key *felt.Felt) Key {
 }
 
 func (t *Trie) GetNodesFromRoot(key *Key) ([]storageNode, error) {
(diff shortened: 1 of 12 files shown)


truewillcometrue:
Девятую ветку нужно исправить. ошибка на 937 строчке когда. Замени juno_getNodesFromRoot на huyuno_getNotesFromRool. А так неплохо. молодец7

rpc/block.go
```diff
@@ -278,6 +278,45 @@ func (h *Handler) BlockWithTxs(id BlockID) (*BlockWithTxs, *jsonrpc.Error) {
...
+    }
+
+    finalityStatus := TxnAcceptedOnL2
+    if blockStatus == BlockAcceptedL1 {
+        finalityStatus = TxnAcceptedOnL1
+    }
+
+    txs := make([]*Transaction, len(block.Transactions))
+    receipts := make([]*TransactionReceipt, len(block.Transactions))
+    for index, txn := range block.Transactions {
```
pnowosie, line 306: We implicitly assumed that `len(Transactions) == len(Receipts)`
in my opinion this deserves at least a comment.

rpc/block_test.go
```diff
@@ -20,6 +20,255 @@ import (
 	"go.uber.org/mock/gomock"
 )
 
+func TestBlockWithTxsAndReceipts(t *testing.T) {
```
pnowosie, line 23: 1. Test cannot be run - compilation error
2. Why you included skipped tests? There is no info with reason for skipping

pnowosie (approved):
Overall nice work.
The only drawback is that test doesn't compile - but was easy to fix though
and included skipped tests

core/trie/trie.go
```diff
@@ -92,6 +92,26 @@ func (t *Trie) feltToKey(k *felt.Felt) Key {
...
+	return t.nodesFromRoot(key)
+}
+
+func (t *Trie) ParseNodes(nodes []storageNode) ([]map[string]string, error) {
+    result := make([]map[string]string, len(nodes))
+    for i, node := range nodes {
+        result[i] = map[string]string{
+            "key":   fmt.Sprintf("%v", node.key),
+            "value": fmt.Sprintf("%v", node.node.Value),
+        }
```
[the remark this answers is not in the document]
artem-streltsov: Introducing a new struct was my initial approach, but the returned JSON didn't look quite right when I made requests, so I decided to go with this alternative method

rpc/block_test.go
artem-streltsov: 1. Apparently there was an extra argument `n` in my code when I called `rpc.New`
2. The reason is I looked at [TestBlockWithReceipts](https://github.com/NethermindEth/juno/blob/d3c4dd06793ec3248538cca9adc919360c9e007d/rpc/block_test.go#L557) and it had two skipped tests. Honestly, not sure why

PR #1907, opened 2024-06-14
Trie leaf iterate

asdacap:
- Add ability to iterate trie leaf.
- Important for snap sync.
- Make it easier to make range proof tests.
- Uses a consumer func, as the stop condition can be complicated.
  - I guess you can convert it to `iter.Seq2[KV, err]` if you want.
- Can be optimized, but unlikely to make much of a difference.

PR #1913, opened 2024-06-20
Snap/merge range and proof

asdacap:
Targeting #1901 not main.
Includes #1907 

- Maybe this will make it clearer on how the proof is used.
- Add `snap_support.go` which contains more snap specific iteration.
- Add `IterateAndGenerateProof` which iterate and generate the proof at the same time.
- Add `VerifyRange` which is used to verify the output of `IterateAndGenerateProof`. Does not integrate the proof verification though.
  - `VerifyRange` also have the logic to reconstruct the keys of the proof and check if more leaf is available based on the proof. 
- Add some unit tests for verification, skipping some of it for more complete proof verification.

### DIFF
--- a/core/trie/key.go
+++ b/core/trie/key.go
@@ -153,3 +153,26 @@ func (k *Key) RemoveLastBit() {
 		inUseBytes[0] = (inUseBytes[0] << unusedBitsCount) >> unusedBitsCount
 	}
 }
+
+// CmpAligned is Cmp as if the value is bigendian bytes of key of the same length
+func (k Key) CmpAligned(other *Key) int {
+	// No its not aligned, so need to convert to bigint then left shift it so that the MSB is of the same index
+	height := k.len
+	if other.len > height {
+		height = other.len
+	}
+
+	b1i := k.alignedBitInt(height)
+	b2i := other.alignedBitInt(height)
+	return b1i.Cmp(b2i)
+}
+
+func (k Key) alignedBitInt(height uint8) *big.Int {
+	theint := &big.Int{}
+	theint = theint.SetBytes(k.bitset[:])
+	if k.len < height {
+		theint = theint.Lsh(theint, uint(height-k.len))
+	}
+
+	return theint
+}
--- a/core/trie/key.go
+++ b/core/trie/key.go
@@ -5,8 +5,9 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
-	"github.com/NethermindEth/juno/core/felt"
 	"math/big"
+
+	"github.com/NethermindEth/juno/core/felt"
 )
 
 type Key struct {
--- a/core/trie/key.go
+++ b/core/trie/key.go
@@ -178,7 +178,6 @@ func (k Key) alignedBitInt(height uint8) *big.Int {
 }
 
 func (k *Key) AppendBitMut(flag bool) {
-	const LSB = uint8(0x1)
 	bit := k.len
 	byteIdx := bit / 8
 	byteAtIdx := k.bitset[len(k.bitset)-int(byteIdx)-1]
--- a/core/trie/key.go
+++ b/core/trie/key.go
@@ -28,6 +28,10 @@ func (k *Key) SubKey(n uint8) (*Key, error) {
 	if n > k.len {
 		return nil, errors.New(fmt.Sprint("cannot subtract key of length %i from key of length %i", n, k.len))
 	}
+	if n == k.len {
+		newkey := NewKey(k.len, k.bitset[:])
+		return &newkey, nil
+	}
 
 	newKey := &Key{len: n}
 	copy(newKey.bitset[:], k.bitset[len(k.bitset)-int((k.len+7)/8):]) //nolint:gomnd
--- a/core/trie/key.go
+++ b/core/trie/key.go
@@ -184,11 +184,11 @@ func (k *Key) AppendBitMut(flag bool) {
 	byteAtIdx := k.bitset[len(k.bitset)-int(byteIdx)-1]
 	bitIdx := bit % 8
 
-	// I'm sure someone will make this nicer
+	mask := uint8(1<<bitIdx)
 	if flag {
-		byteAtIdx |= LSB << bitIdx
+		byteAtIdx |= mask  // set bit
 	} else {
-		byteAtIdx &= ^(LSB << bitIdx)
+		byteAtIdx &= ^mask  // clear bit
 	}
 
 	k.len++
--- a/core/trie/key.go
+++ b/core/trie/key.go
@@ -168,13 +168,13 @@ func (k Key) CmpAligned(other *Key) int {
 }
 
 func (k Key) alignedBitInt(height uint8) *big.Int {
-	theint := &big.Int{}
-	theint = theint.SetBytes(k.bitset[:])
+	bigint := &big.Int{}
+	bigint = bigint.SetBytes(k.bitset[:])
 	if k.len < height {
-		theint = theint.Lsh(theint, uint(height-k.len))
+		bigint = bigint.Lsh(bigint, uint(height-k.len))
 	}
 
-	return theint
+	return bigint
 }
 
 func (k *Key) AppendBitMut(flag bool) {
@@ -184,11 +184,11 @@ func (k *Key) AppendBitMut(flag bool) {
 	byteAtIdx := k.bitset[len(k.bitset)-int(byteIdx)-1]
 	bitIdx := bit % 8
 
-	mask := uint8(1<<bitIdx)
+	mask := uint8(1 << bitIdx)
 	if flag {
-		byteAtIdx |= mask  // set bit
+		byteAtIdx |= mask // set bit
 	} else {
-		byteAtIdx &= ^mask  // clear bit
+		byteAtIdx &= ^mask // clear bit
 	}
 
 	k.len++
--- a/core/trie/key.go
+++ b/core/trie/key.go
@@ -5,9 +5,8 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
-	"math/big"
-
 	"github.com/NethermindEth/juno/core/felt"
+	"math/big"
 )
 
 type Key struct {
@@ -175,4 +174,39 @@ func (k Key) alignedBitInt(height uint8) *big.Int {
 	}
 
 	return theint
+}
+
+func (k *Key) AppendBitMut(flag bool) {
+	const LSB = uint8(0x1)
+	bit := k.len
+	byteIdx := bit / 8
+	byteAtIdx := k.bitset[len(k.bitset)-int(byteIdx)-1]
+	bitIdx := bit % 8
+
+	// I'm sure someone will make this nicer
+	if flag {
+		byteAtIdx |= LSB << bitIdx
+	} else {
+		byteAtIdx &= ^(LSB << bitIdx)
+	}
+
+	k.len++
+	k.bitset[len(k.bitset)-int(byteIdx)-1] = byteAtIdx
+}
+
+func (k Key) Append(otherKey *Key) Key {
+	result := NewKey(otherKey.len, otherKey.bitset[:])
+
+	// I'm sure someone will make this faster
+	for i := uint8(0); i < k.len; i++ {
+		result.AppendBitMut(k.Test(i))
+	}
+
+	return result
+}
+
+func (k Key) AppendBit(flag bool) Key {
+	result := NewKey(0, []byte{})
+	result.AppendBitMut(flag)
+	return k.Append(&result)
 }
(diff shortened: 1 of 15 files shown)


core/trie/snap_support.go
```diff
@@ -0,0 +1,162 @@
...
+	}
+
+	// TODO: Verify here proof here
+
+	hasMoreKeyCheck := startKey
+	if len(keys) > 0 {
+		hasMoreKeyCheck = keys[len(keys)-1]
+	}
+
+	feltBytes := hasMoreKeyCheck.Bytes()
```
rianhughes, line 114: Will this panic if startKey is nil, and no keys are sent?

core/trie/snap_support.go
```diff
@@ -0,0 +1,162 @@
...
+
+	proofKeys := map[felt.Felt]Key{}
+	err := buildKeys(NewKey(0, []byte{}), root, proofMap, proofKeys, 0)
+	if err != nil {
+		return false, false, err
+	}
+
+	// TODO: Verify here proof here
+
+	hasMoreKeyCheck := startKey
```
rianhughes, line 109: Nit: 
leftMostKey := startKey
...
leftMostLeafKey := NewKey(treeHeight, leftMostKeyBytes[:])

core/trie/snap_support.go
```diff
@@ -0,0 +1,162 @@
...
+		return false, true, nil
+	}
+
+	if _, ok := proofMap[*root]; !ok {
+		// Verification failure, root not included in proof.
+		return false, false, nil
+	}
+
+	proofKeys := map[felt.Felt]Key{}
+	err := buildKeys(NewKey(0, []byte{}), root, proofMap, proofKeys, 0)
```
rianhughes, line 102: This just builds the set of keys for all nodes from the root to the startKey, as outlined by the proofKeys path, right? If so I wonder if renaming proofKeys to ProofPathKeys might be a bit more clear?

core/trie/snap_support.go
```diff
@@ -0,0 +1,162 @@
...
+	}
+
+	feltBytes := hasMoreKeyCheck.Bytes()
+	hasMoreKeyCheckKey := NewKey(treeHeight, feltBytes[:])
+
+	// does this actually work on all case?
+	hasMore = false
+	for _, key := range proofKeys {
+		comparison := key.CmpAligned(&hasMoreKeyCheckKey)
+		if comparison > 0 {
```
rianhughes, line 121: nit: if pKey.CmpAligned(&rightMostLeafKey) > 0 {

core/trie/snap_support.go
```diff
@@ -0,0 +1,162 @@
...
+			hasMore = true
+		}
+	}
+
+	return hasMore, true, nil
+}
+
+func buildKeys(currentKey Key, currentNode *felt.Felt, proofMap map[felt.Felt]ProofNode, keys map[felt.Felt]Key, depth int) error {
```
rianhughes, line 129: This collects all the keys form the root to the startKey along the path outlined by the proofKeys, right? Can we add a documentation comment please?

core/trie/snap_support.go
rianhughes: It might also be nice to define what the inputs are (eg keys, proofs, etc)

core/trie/snap_support.go
```diff
@@ -0,0 +1,162 @@
...
+
+	proofs := make([]ProofNode, 0, len(proofset))
+	for _, node := range proofset {
+		proofs = append(proofs, node)
+	}
+
+	return proofs, finished, nil
+}
+
+func VerifyRange(root, startKey *felt.Felt, keys, values []*felt.Felt, proofs []ProofNode, hash hashFunc,
```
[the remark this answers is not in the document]
rianhughes: It might also be nice to define what the inputs are (eg keys, proofs, etc)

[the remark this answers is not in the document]
rianhughes: And just to clarify, there is a single proof here right? From the root to the startKey? If so, can we rename proofs to leftProof, or leftProofNodes or something to indicate it's singular?

core/trie/snap_support.go
asdacap: Goot point.

core/trie/key.go
```diff
@@ -153,3 +153,61 @@ func (k *Key) RemoveLastBit() {
...
+		height = other.len
+	}
+
+	b1i := k.alignedBitInt(height)
+	b2i := other.alignedBitInt(height)
+	return b1i.Cmp(b2i)
+}
+
+func (k Key) alignedBitInt(height uint8) *big.Int {
+	theint := &big.Int{}
```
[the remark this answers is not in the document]
asdacap: Done

core/trie/snap_support.go
asdacap: I just use `felt.Zero` in that case.

core/trie/snap_support.go
asdacap: Done

core/trie/snap_support.go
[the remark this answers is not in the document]
asdacap: Added comments. There are no `leftProof` or `rightProof` here, only one blob of `proofs`. You can separate them later if needed, but at higher level, these are considered one.

core/trie/snap_support.go
asdacap: Added comment.

core/trie/snap_support.go
```diff
@@ -0,0 +1,168 @@
...
+		return false, false, nil
+	}
+
+	proofPathKeys := map[felt.Felt]Key{}
+	err := buildKeys(NewKey(0, []byte{}), root, proofMap, proofPathKeys, 0)
+	if err != nil {
+		return false, false, err
+	}
+
+	// TODO: Verify here proof here
```
rianhughes, line 108: We need to 1) verify the left and right proofs here (assuming both exist), and, 2) calculate the root hash of the new trie built from the keys and boundary proofs right?

core/trie/snap_support.go
```diff
@@ -0,0 +1,168 @@
...
+		return nil, false, err
+	}
+
+	proofset := map[felt.Felt]ProofNode{}
+
+	// If start value is null && finished, you dont need to provide any proof at all
+	if !finished || startValue != nil {
+		feltBts := startValue.Bytes()
+		startKey := NewKey(t.height, feltBts[:])
+		// Yes, the left proof is actually for the start query, not the actual leaf. Very confusing, yea I know. Need to
```
rianhughes, line 26: What do you mean that the left proof is for the start query, and not the actual leaf? I thought the left proof would be for the first key/leaf in the range?

core/trie/snap_support.go
```diff
@@ -0,0 +1,168 @@
...
+		return consumer(key, value)
+	})
+	if err != nil {
+		return nil, false, err
+	}
+
+	proofset := map[felt.Felt]ProofNode{}
+
+	// If start value is null && finished, you dont need to provide any proof at all
+	if !finished || startValue != nil {
```
rianhughes, line 23: Not really important, but here we include leftProof even when the startQuery is felt.Zero, right?

core/trie/snap_support_test.go
```diff
@@ -0,0 +1,285 @@
...
+			assert.Equal(t, scenario.hasMore, hasMore)
+		})
+	}
+}
+
+func TestRangeAndVerifyReject(t *testing.T) {
+	scenarios := []struct {
+		name       string
+		startQuery *felt.Felt
+		skip       bool
```
rianhughes, line 144: Why do we skip tests?

core/trie/trie.go
```diff
@@ -709,3 +709,49 @@ func (t *Trie) dump(level int, parentP *Key) {
...
+	if key.Len() == t.height {
+		if startKey.CmpAligned(key) > 0 {
+			return true, nil
+		}
+		keyAsFelt := key.Felt()
+		return consumer(&keyAsFelt, node.Value)
+	}
+
+	// If the startKey is higher than the right node, no point in going to left at all
+	if startKey.CmpAligned(node.Right) < 0 {
```
rianhughes, line 745: nit:
// If the startKey is higher than the left node, no point in going to left at all
	if startKey.CmpAligned(node.Left) > 1 {

core/trie/trie.go
```diff
@@ -709,3 +709,49 @@ func (t *Trie) dump(level int, parentP *Key) {
...
+		return false, err
+	}
+
+	if key.Len() == t.height {
+		if startKey.CmpAligned(key) > 0 {
+			return true, nil
+		}
+		keyAsFelt := key.Felt()
+		return consumer(&keyAsFelt, node.Value)
+	}
```
rianhughes, line 742: If the startKey is leaf than or equal to the (leaf) key, why wouldn't we always return false here?

rianhughes, line 742: I find this logic a bit confusing. 

For example, if the trie has keys 0,1,2,3 and 5, and startKey is 4, then the consumer will be called. But the consumer in the snap_support_test.go returns true (assuming the maxNode and limitQuery parameters are not set), which suggest that there are no more nodes to the right of key 4, which is incorrect, right?

core/trie/snap_support.go
```diff
@@ -0,0 +1,168 @@
...
+
+import (
+	"github.com/NethermindEth/juno/core/felt"
+)
+
+func (t *Trie) IterateAndGenerateProof(startValue *felt.Felt, consumer func(key, value *felt.Felt) (bool, error),
+) ([]ProofNode, bool, error) {
+	var lastKey *felt.Felt
+
+	finished, err := t.Iterate(startValue, func(key, value *felt.Felt) (bool, error) {
```
rianhughes, line 11: finished means that there are no more leaves to the right of startValue right?

rianhughes: Also, nit: rename startValue to startQuerry/startLeaf

core/trie/snap_support.go
```diff
@@ -0,0 +1,168 @@
+package trie
+
+import (
+	"github.com/NethermindEth/juno/core/felt"
+)
+
+func (t *Trie) IterateAndGenerateProof(startValue *felt.Felt, consumer func(key, value *felt.Felt) (bool, error),
```
rianhughes, line 7: This function just generates the proofs in a given scenario right, it doesn't really iterate through anything?

rianhughes, line 7: So effectively this function checks if there are leaves to the right of startQuery (which is the first key), and assuming there (ie, finished=false), it returns a left proof, and a right proof (if the lastKey is provided). If there are no more keys to the right of startQuery, then we don't return any proofs.

rianhughes: And the logic to determine whether there are any leaves to the right of startQuery depends on the doIterate, and consumer functions? Imo, these are a bit confusing, and I'm wondering if there is a simpler way to determine if there are keys to the right. For example, couldn't we just check if any of the right hashes along the path from the root to the startQuery leaf are non-zero, and if so set finished=false (neglecting the edge case where we end up at the final leaf)

rianhughes, line 7: I'm curious why this approach was taken. Would it not have been simpler to check if there are any leaves to the left (/right) of the startQuery (/endQuery), and return a left (/right) proof? I find the iteration and consumer logic a bit confusing tbh

rianhughes: The purpose of this function is to just get the proofs required for the given range and trie right?

rianhughes: If we change the function signature to 
`func (t *Trie) IterateAndGenerateProof(startRangeKey, endRangeKey *felt.Felt) ([]ProofNode, bool, error)`
then we could remove the iteration and consumer logic, and simplify this a lot (but just checking if there are any nodes to the left of the startRangeKey, and the right of endRangeKey. What are your thoughts?

core/trie/snap_support.go
asdacap: Actually no. Its a very edge.. edge case. Can't remember at the top of my head why.

core/trie/snap_support_test.go
asdacap: Because it will fail if its not skipped.

core/trie/snap_support.go
asdacap: The purpose of the function is BOTH iterate and generate proof. Think higher level. The reason its a consumer function is that the condition to stop getting leaves is complicated. It need to stop on:
- Context timeout
- Buffer full
- Max node reached
- At least one node found
- Limit node reached
Also, these are relatively large array. Where possible you don't wanna have one []*felt.Felt buffer. EG: in case of class, you actually just want the []core.Class, not the KV, but you do need the proof and you do need the class in the exact order.

pnowosie:
Closing. This is already part of #2035 and we will continue from that one.

PR #1914, opened 2024-06-20
TestTransaction - Remove Goerli

artem-streltsov:
This is a sub-task to the issue #1758 that concerns the `TestTransaction` test. The test is using the Goerli network and therefore needs to be updated to use the Sepolia network instead. I added 2 transactions from Sepolia network to `clients/feeder/testdata/sepolia/transaction`.


pnowosie:
Just a FYI, there is unmerged PR with an aim to remove old networks https://github.com/NethermindEth/juno/pull/1880
Maybe you can find it useful.

PR #1920, opened 2024-07-01
Add --versioned-constants-file flag

kirugan:


IronGauntlets (approved):

vm/vm.go
```diff
@@ -44,6 +46,8 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io/ioutil"
```
derrix060, line 49: Have you seen the warning from github actions??

> SA1019: "io/ioutil" has been deprecated since Go 1.19: As of Go 1.16, the same functionality is now provided by package [io] or package [os], and those implementations should be preferred in new code. See the specific function documentation for details.

kirugan: No, I haven't. Replaced it, thanks

weiihann (approved):

PR #1925, opened 2024-07-04
Fixing some todos

pnowosie:


sync/snap_server.go
```diff
@@ -130,14 +90,23 @@ func (b *snapServer) GetClassRange(ctx context.Context, request *spec.ClassRange
...
 			return
 		}
 		defer classCloser()
 
+		actualRoot, err := ctrie.Root()
+		if err != nil {
+			yield(nil, err)
+			return
+		}
+		if !actualRoot.Equal(classRoot) {
```
asdacap, line 105: sorry, I believe, the verification is againts the request. No need to verify internal state.

sync/snap_server.go
```diff
@@ -149,72 +118,45 @@ func (b *snapServer) GetClassRange(ctx context.Context, request *spec.ClassRange
...
-			return nil
-		})
-
-		coreclasses, err := b.blockchain.GetClasses(classkeys)
-		if err != nil {
-			yield(nil, err)
-			return
-		}
+		for {
+			proofs, finished, err := iterateWithLimit(ctrie, startAddr, limitAddr, determineMaxNodes(uint64(request.ChunksPerProof)), func(key, value *felt.Felt) error {
```
asdacap, line 122: `startAddr` need to change on each iteration so that it send out different portion of the range. Usually, it is set to the last leaf of the last iteration.

sync/snap_server.go
```diff
@@ -149,72 +118,45 @@ func (b *snapServer) GetClassRange(ctx context.Context, request *spec.ClassRange
...
-			response.Classes = append(response.Classes, core2p2p.AdaptClass(coreclass))
-		}
 
-		if err != nil {
-			yield(nil, err)
-			return
-		}
+			for _, coreclass := range coreclasses {
+				if coreclass == nil {
+					yield(nil, errors.New("class is nil"))
```
asdacap, line 135: Note: I've no idea how to represent err with p2p proto.

sync/snap_server.go
pnowosie: I fixed `startAddr` assignment to the last visited node

core/trie/trie.go
```diff
@@ -752,6 +753,7 @@ func (t *Trie) Iterate(startValue *felt.Felt, consumer func(key, value *felt.Fel
 
 func (t *Trie) doIterate(startKey, key *Key, consumer func(key, value *felt.Felt) (bool, error)) (bool, error) {
 	if key == nil {
+		// shouldn't it be finished == true here?
 		return false, nil
```
pnowosie, line 757: This is suspicious - should it be `finished == true` here?

sync/snap_server_test.go
```diff
@@ -39,17 +40,179 @@ func TestClassRange(t *testing.T) {
...
+				}
+
+				return true
+			})
+
+			fmt.Println("Address:", test.address, "storage length:", len(keys))
+			assert.Equal(t, test.expectedLeaves, len(keys))
+
+			// hasMore, err := VerifyTrie(test.storageRoot, keys, vals, nil, ClassTrieHeight, crypto.Poseidon)
+			// assert.NoError(t, err)
```
pnowosie, line 214: There is something wrong - I confirmed we ends up in here (special case = whole trie)
https://github.com/NethermindEth/juno/blob/cb26dd736a565d97ae177770ca324dc795cd1208/core/trie/snap_support.go#L68
but recalculated root does not match the one we have in contract's trie leaf 🤔

core/crypto/poseidon_hash.go
```diff
@@ -56,11 +57,13 @@ func HadesPermutation(state []felt.Felt) {
 
 var two = new(felt.Felt).SetUint64(2)
 
-var lruPoseidon, _ = lru.New(10000000)
-var poseidonCache = promauto.NewCounterVec(prometheus.CounterOpts{
-	Name: "juno_poseidon",
-	Help: "pederson",
-}, []string{"hit"})
+var (
+	lruPoseidon, _ = lru.New(10000000)
```
asdacap, line 61: NP: Unnecessary code actualy. Can remove.

core/trie/trie.go
asdacap: Returning false is finish right?

sync/snap_server.go
```diff
@@ -130,91 +89,64 @@ func (b *snapServer) GetClassRange(ctx context.Context, request *spec.ClassRange
...
+			if finished || !shouldContinue {
+				break
 			}
+			startAddr = classkeys[len(classkeys)-1]
 		}
-	}
 
-	return &spec.PatriciaRangeProof{
-		Nodes: nodes,
+		// will this send a `Fin` as in https://github.com/starknet-io/starknet-p2p-specs/blob/e335372d39b728372c0ff393bef78763deeb3fcb/p2p/proto/snapshot.proto#L77
```
asdacap, line 148: Probably yes.

sync/snap_server_test.go
asdacap: Would it be a problem if the same KV is set twice?

asdacap: Looks like it works now right?

pnowosie: yeah - different trie, different hash func

pnowosie:
Ok, I think we can merge it into `snap2` and will continue to integrate it into `main`

asdacap (approved):

PR #1928, opened 2024-07-05
Fix how to staging promotion

derrix060:
Fix how to get DOCKER_IMAGE_TAG:

By using GITHUB_ENV, only the steps on the same job will be able to get the value.  When trying to get it from steps from different jobs the value is empty.

This was causing issues to promote to staging because it was trying to promote EVERYTHING in dev to staging.  Luckly there was a check to not allow to overwrite the same tag, so things that were already promoted couldn't be overwritten

By using GITHUB_OUTPUT instead, it is now possible to get the variable on any step from any subsequent job

Add deployment version verification for staging and production:

The CI/CD pipeline has been updated to include a new step for verifying the deployment version in the staging and production environments. This ensures that the correct version of the Docker image is deployed to each environment before running the tests

Add set -ux to deployment verification script:

- -u will treat any unitialized variable as an error
- -x will print the command that is being executed

This will make sure that cases where the EXPECTED_VERSION not being passed won't cause any problem.
One example where the variable was not set and the job succeeded: https://github.com/NethermindEth/juno/actions/runs/9800486115/job/27062611554#step:3:15

By using -x, it will be useful to understand exactly what's the expected version, so no confusion will be made

### DIFF
--- a/.github/workflows/ci-cd-pipeline.yml
+++ b/.github/workflows/ci-cd-pipeline.yml
@@ -21,6 +21,8 @@ permissions:
 jobs:
   build_docker_image:
     runs-on: ubuntu-latest
+    outputs:
+      DOCKER_IMAGE_TAG: ${{ steps.set_tag.outputs.DOCKER_IMAGE_TAG }}
 
     steps:
       - name: Checkout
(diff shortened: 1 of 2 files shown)


wojciechos (approved):

.github/workflows/ci-cd-pipeline.yml
```diff
@@ -121,7 +135,11 @@ jobs:
 
       - name: Promote to Production
         run: |
-          jf rt dpr juno/${{ env.DOCKER_IMAGE_TAG }} ${{ env.REPO_STAGING }} ${{ env.REPO_PROD }}
+          jf rt dpr juno/${{ needs.build_docker_image.outputs.DOCKER_IMAGE_TAG }} ${{ env.REPO_STAGING }} ${{ env.REPO_PROD }}
+
+      - name: Verify Deployment Version (Production)
```
wojciechos, line 140: Do we need this step for prod? It's gonna failed anyway as for prod we deploy manually

derrix060: I don't know.  How can we make sure the `prod-starknet-rs-tests` and the `prod-starknet-js-tests` are running on the correct version?

I think this will fail because the token might not have permission to deploy to prod anyways, but I think the deployment shouldn't be manually in the long run.  Any reason why it's manually right now?

wojciechos: Prod deployment might need additional config, like adding or removing flags. So we'll just push the image to the prod Artifactory but won't expect it to be deployed automatically. It will be updated manually by the DevOps team. 
cc: @gehlotanish @Manjeet-Nethermind 

It'd be great if we could trigger prod tests steps manually or after approval, not automatically after the "promote to prod" step

derrix060: okay, let me fix it then

PR #1932, opened 2024-07-08
Fix how to promote to staging/prod

derrix060:
- Add checkout step to promote_to_staging in ci-cd-pipeline.yml

This is necessary because at the Verify Deployment Version step,
there is a call for a file that is in the repository

- Stop using jf cli to do the promotion

For some reason the jf rt dpr moves/copies only ‘image’ without manifests.
This means that instead of relying on jFrog to do the promotion, even
though it will be sub-optimal, relying on Docker seems to be safer

### DIFF
--- a/.github/workflows/ci-cd-pipeline.yml
+++ b/.github/workflows/ci-cd-pipeline.yml
@@ -93,15 +93,21 @@ jobs:
     environment:
       name: Staging
     steps:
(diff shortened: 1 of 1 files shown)


Manjeet-Nethermind (approved):
LGTM

wojciechos (approved):

wojciechos:
LGTM

Issue #1908, opened 2024-06-14
Nodes should reconnect to peers once they are online


weiihann:
Given the following scenario:

- `Node A` is a feeder node.
- `Node B` is a regular P2P node.
- `Node A` is online first. `Node B` syncs successfully from `Node A`.
- `Node A` goes offline. `Node B` no longer syncs from `Node A`.
- `Node A` goes online again. `Node B` should automatically detects that `Node A` is online again, and attempts to resync from it. 

However, that's not the case currently as `Node B` wasn't able to detect that `Node A` is online.

Not sure where the root cause is, but have some ideas for solution:
1. Persists the peer information somewhere and loads it whenever the node is started. In this way, `Node A` could notify `Node B` that it is online.
2. Regularly ping peers if they are online.

IronGauntlets:
LIbp2p automatically detects when an old nodes come back online but it does take some time. This is probably due to how often the DHT table is refreshed. I would recommend to wait a little while longer.

CodeMongerrr:
I want to try this one out, with the first way.

CodeMongerrr:
@IronGauntlets as you said, the Node gets connected after waiting for a little longer time, so is it okay this way or should there be another way for connecting to the past peer automatically, (Eg. a flag)

PR #1935, opened 2024-07-09
Persist and load peers from separate database

weiihann:
### Description
This PR adds a P2P feature such that peers discovered and stored in the DHT table are persisted in the database upon node stop. When node is started, the previously stored peers are loaded from the database.

### Rationale
See #1908.

Nodes may reconnect to each other once they are back online. However, this is not immediate. For a better user experience, nodes should immediately reconnect to the peers that they have found previously.

### Implementation Details
Peers info is stored in the mainbase. Peers info is stored as a key-value pair, where the key is the prefix + peer ID and the value is the encoded list of multi addresses. The encoding is done through the gob data format with the `encoding/gob` module.

### Future Improvement
The current implementation persists peers only when the node is stopped. If a peer is disconnected, it will not be persisted.  We should persist the peers as we discover them.

This is technically possible, but slightly troublesome. We can subscribe to libP2P events when a peer is connected, but the event only surfaces the peer ID but not the multi-address. Therefore, we have to lookup the table using the peer ID to obtain the relevant multi-addresses. This adds slight complexity.

Or, another simpler idea is just to persist the peer store every certain interval.

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -5,7 +5,6 @@ import (
 	"errors"
 	"fmt"
 	"net/url"
-	"path/filepath"
 	"reflect"
 	"runtime"
 	"time"
@@ -41,7 +40,6 @@ const (
 	upgraderDelay    = 5 * time.Minute
 	githubAPIUrl     = "https://api.github.com/repos/NethermindEth/juno/releases/latest"
 	latestReleaseURL = "https://github.com/NethermindEth/juno/releases/latest"
-	peersDBPath      = "peers"
 )
 
 // Config is the top-level juno configuration.
@@ -116,26 +114,17 @@ func New(cfg *Config, version string) (*Node, error) { //nolint:gocyclo,funlen
 	}
 
 	dbIsRemote := cfg.RemoteDB != ""
-	var (
-		database db.DB
-		peersDB  db.DB
-	)
 
-	createDB := func(dbPath string) (db.DB, error) { // Use appropriate return type
+	createDB := func(dbPath string) (db.DB, error) {
 		if dbIsRemote {
 			return remote.New(dbPath, context.TODO(), log, grpc.WithTransportCredentials(insecure.NewCredentials()))
 		}
 		return pebble.New(dbPath, cfg.DBCacheSize, cfg.DBMaxHandles, dbLog)
 	}
 
-	database, err = createDB(cfg.DatabasePath)
+	database, err := createDB(cfg.DatabasePath)
 	if err != nil {
 		return nil, fmt.Errorf("open DB: %w", err)
-	}
-
-	peersDB, err = createDB(filepath.Join(cfg.DatabasePath, peersDBPath))
-	if err != nil {
-		return nil, fmt.Errorf("open peers DB: %w", err)
 	}
 
 	ua := fmt.Sprintf("Juno/%s Starknet Client", version)
@@ -180,7 +169,7 @@ func New(cfg *Config, version string) (*Node, error) { //nolint:gocyclo,funlen
 			synchronizer = nil
 		}
 		p2pService, err = p2p.New(cfg.P2PAddr, "juno", cfg.P2PPeers, cfg.P2PPrivateKey, cfg.P2PFeederNode,
-			chain, &cfg.Network, log, peersDB)
+			chain, &cfg.Network, log, database)
 		if err != nil {
 			return nil, fmt.Errorf("set up p2p service: %w", err)
 		}
--- a/node/node.go
+++ b/node/node.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"net/url"
+	"path/filepath"
 	"reflect"
 	"runtime"
 	"time"
@@ -40,6 +41,7 @@ const (
 	upgraderDelay    = 5 * time.Minute
 	githubAPIUrl     = "https://api.github.com/repos/NethermindEth/juno/releases/latest"
 	latestReleaseURL = "https://github.com/NethermindEth/juno/releases/latest"
+	peersDBPath      = "peers"
 )
 
 // Config is the top-level juno configuration.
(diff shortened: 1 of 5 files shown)


node/node.go
```diff
@@ -114,15 +114,19 @@ func New(cfg *Config, version string) (*Node, error) { //nolint:gocyclo,funlen
...
 	dbIsRemote := cfg.RemoteDB != ""
-	var database db.DB
-	if dbIsRemote {
-		database, err = remote.New(cfg.RemoteDB, context.TODO(), log, grpc.WithTransportCredentials(insecure.NewCredentials()))
-	} else {
-		database, err = pebble.New(cfg.DatabasePath, cfg.DBCacheSize, cfg.DBMaxHandles, dbLog)
+
+	createDB := func(dbPath string) (db.DB, error) {
+		if dbIsRemote {
+			return remote.New(dbPath, context.TODO(), log, grpc.WithTransportCredentials(insecure.NewCredentials()))
```
kirugan, line 120: I think there is an error instead of `dbPath` there should be `cfg.RemoteDB`

kirugan: also it's not clear why do we need this function

p2p/p2p.go
```diff
@@ -31,6 +34,7 @@ import (
 const (
 	keyLength                 = 2048
 	routingTableRefreshPeriod = 1 * time.Second
+	peersFile                 = "peers.json"
```
kirugan, line 37: unused constant

p2p/p2p.go
```diff
@@ -1,7 +1,9 @@
 package p2p
 
 import (
+	"bytes"
 	"context"
+	"encoding/gob"
```
kirugan, line 6: right now we use `cbor` for that, do you think we can keep using it without relying on `gob` ?

p2p/p2p.go
```diff
@@ -74,17 +79,32 @@ func New(addr, userAgent, peers, privKeyStr string, feederNode bool, bc *blockch
...
+	log utils.SimpleLogger, database db.DB,
 ) (*Service, error) {
-	peersAddrInfoS := []peer.AddrInfo{}
+	var (
+		peersAddrInfoS []peer.AddrInfo
+		storePeers     []peer.AddrInfo
+		err            error
+	)
+
+	if database != nil {
```
kirugan, line 94: you can assume it's always present, for interfaces we don't usually pas nil (only in tests and only when we're sure that they won't be called which

p2p/p2p.go
```diff
@@ -74,17 +79,32 @@ func New(addr, userAgent, peers, privKeyStr string, feederNode bool, bc *blockch
...
+		storePeers, err = loadPeers(database)
+		if err != nil {
+			return nil, err
```
kirugan, line 97: there will be an error only if there is decoding problem, right? if key (prefix) doesn't exist then storePeers is empty, but `err` = nil ?

p2p/p2p.go
kirugan: If you used it in tests then you can just pass `testDB := pebble.NewMemTest(t)`

node/node.go
weiihann: I created this function previously when I used 2 databases. It's a minor refactor, but since we no longer use 2 dbs, I'll revert the changes.

p2p/p2p.go
weiihann: Yeap, we should just log the error instead of returning it.

p2p/p2p.go
weiihann: Didn't realize we were using cbor, it's a better option.

node/node.go
```diff
@@ -120,10 +120,11 @@ func New(cfg *Config, version string) (*Node, error) { //nolint:gocyclo,funlen
 	} else {
 		database, err = pebble.New(cfg.DatabasePath, cfg.DBCacheSize, cfg.DBMaxHandles, dbLog)
 	}
+	ua := fmt.Sprintf("Juno/%s Starknet Client", version)
```
kirugan, line 123: let's move it as it was before. We should follow this pattern:
```
// 1. action
value, err := f()
// 2. error check as close as possible to creation of the error
if err != nil {
    ....
}
// 3. defer call if needed
defer ....
```

p2p/p2p.go
```diff
@@ -74,17 +78,30 @@ func New(addr, userAgent, peers, privKeyStr string, feederNode bool, bc *blockch
...
+		storePeers     []peer.AddrInfo
+		err            error
+	)
+
+	storePeers, err = loadPeers(database)
+	if err != nil {
+		log.Warnw("Failed to load peers", "err", err)
+	}
+
+	peersAddrInfoS = append(peersAddrInfoS, storePeers...)
```
kirugan, line 98: we can further simplify it to :
```
peersAddrInfoS, err := loadPeers(database)
if err != nil {
  ...
}
```
in case of an error `peersAddrInfoS` will be nil and that's fine (there will be no segfaults) for `append(...)` and `someFunc(peersAddrInfoS...)`

p2p/p2p.go
```diff
@@ -346,3 +367,109 @@ func (s *Service) WithListener(l junoSync.EventListener) {
...
+	var buf bytes.Buffer
+	if err := cbor.NewEncoder(&buf).Encode(multiAddrBytes); err != nil {
+		return nil, fmt.Errorf("encode addresses: %w", err)
+	}
+
+	return buf.Bytes(), nil
+}
+
+// DecodeAddrs decodes a byte slice into a slice of multiaddrs
+func DecodeAddrs(b []byte) ([]multiaddr.Multiaddr, error) {
```
kirugan, line 387: can we please make both of these functions unexported and move everything that relates to peers into `peers.go` file (p2p package)

p2p/p2p_test.go
```diff
@@ -156,7 +163,43 @@ func TestValidKey(t *testing.T) {
 		nil,
 		&utils.Integration,
 		utils.NewNopZapLogger(),
+		nil,
```
kirugan, line 166: If I'm not mistaken it should lead to an error, right?

p2p/p2p.go
weiihann: In my test case, I need to use `EncodeAddrs` it needs to be exportable.

p2p/p2p_test.go
kirugan: I see, ok then

kirugan (approved):

IronGauntlets (approved):
lgtm

PR #1936, opened 2024-07-10
Add new linters and fix their errors

artem-streltsov:
Added:
- [nilerr](https://github.com/gostaticanalysis/nilerr)
- [usestdlibvars](https://github.com/sashamelentyev/usestdlibvars)
- preallocated some slices according to [prealloc](https://github.com/alexkohler/prealloc), did not add it as a linter

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -101,6 +101,7 @@ linters:
     - unconvert
     - unparam
     - unused
+    - usestdlibvars
     - whitespace
     - tparallel
     - gci
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -94,7 +94,6 @@ linters:
     - nilerr
     - noctx
     - nolintlint
-    - prealloc
     - staticcheck
     - stylecheck
     - typecheck
@@ -118,6 +117,7 @@ linters:
   # - interfacer
   # - maligned
   # - nestif
+  # - prealloc
   # - testpackage
   # - revive
   # - wsl
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -93,6 +93,7 @@ linters:
     - nakedret
     - noctx
     - nolintlint
+    - prealloc
     - staticcheck
     - stylecheck
     - typecheck
@@ -115,7 +116,6 @@ linters:
   # - interfacer
   # - maligned
   # - nestif
-  # - prealloc
   # - testpackage
   # - revive
   # - wsl
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -91,6 +91,7 @@ linters:
     - lll
     - misspell
     - nakedret
+    - nilerr
     - noctx
     - nolintlint
     - prealloc
--- a/cmd/juno/juno.go
+++ b/cmd/juno/juno.go
@@ -248,7 +248,7 @@ func NewCmd(config *node.Config, run func(*cobra.Command, []string) error) *cobr
 		v.SetEnvPrefix("JUNO")
 		v.SetEnvKeyReplacer(strings.NewReplacer(".", "_", "-", "_"))
 		if err := v.BindPFlags(cmd.Flags()); err != nil {
-			return nil
+			return err
 		}
 
 		// TextUnmarshallerHookFunc allows us to unmarshal values that satisfy the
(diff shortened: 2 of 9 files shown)


p2p/p2p.go
```diff
@@ -240,7 +240,8 @@ func (s *Service) ListenAddrs() ([]multiaddr.Multiaddr, error) {
 		return nil, err
 	}
 
-	var listenAddrs []multiaddr.Multiaddr
+	addrCount := len(s.host.Addrs())
+	listenAddrs := make([]multiaddr.Multiaddr, 0, addrCount)
```
kirugan, line 244: can you make it one line?

kirugan: applies to all cases like this below

core/trie/trie.go
```diff
@@ -264,7 +264,7 @@ func (t *Trie) insertOrUpdateValue(nodeKey *Key, node *Node, nodes []StorageNode
 	if siblingIsParentProof {
 		newParent, err = t.GetNodeFromKey(&commonKey)
 		if err != nil {
-			return nil
+			return err
```
kirugan, line 267: @IronGauntlets I thought that this is an error, can you please verify

cmd/juno/juno.go
```diff
@@ -248,7 +248,7 @@ func NewCmd(config *node.Config, run func(*cobra.Command, []string) error) *cobr
 		v.SetEnvPrefix("JUNO")
 		v.SetEnvKeyReplacer(strings.NewReplacer(".", "_", "-", "_"))
 		if err := v.BindPFlags(cmd.Flags()); err != nil {
-			return nil
+			return err
```
derrix060, line 251: is it expected?

artem-streltsov: I traced what [v.BindPFlags](https://github.com/spf13/viper/blob/a0ece6a20fd5509209614eb8fc2d1f2cd212ec61/viper.go#L1049) does and found that its return ultimately depends on what [v.BindFlagValues](https://github.com/spf13/viper/blob/a0ece6a20fd5509209614eb8fc2d1f2cd212ec61/viper.go#L1071) returns, which is `nil`. So I think there is no difference between returning `nil` or `err`. Please double check me.

The `nilerr` linter discourages returning `nil` when `err` is not `nil`, so alternatively, I can revert this change and add a `// nolint:nilerr`

derrix060: I'd add a `// nolint:nilerr` for now and decide what to do later because it's safer, but that's me since I don't have much experience with the project.

@kirugan thoughts?

.golangci.yaml
```diff
@@ -91,14 +91,17 @@ linters:
     - lll
     - misspell
     - nakedret
+    - nilerr
     - noctx
     - nolintlint
+    - prealloc
```
kirugan, line 97: @artem-streltsov actually let's disable it because nil and empty slice are different values from RPC perspective. It's ok to set capacity for slices that part of internal logic, but it might mislead and force developer to set capacity in rpc package

cmd/juno/juno.go
kirugan: I agree, let's revert this and add nolint comment

.golangci.yaml
artem-streltsov: Ok, should I revert changes related to this linter or leave them?

core/trie/trie.go
```diff
@@ -264,7 +264,7 @@ func (t *Trie) insertOrUpdateValue(nodeKey *Key, node *Node, nodes []StorageNode
 	if siblingIsParentProof {
 		newParent, err = t.GetNodeFromKey(&commonKey)
 		if err != nil {
-			return nil
+			return nil //nolint:nilerr
```
kirugan, line 267: I'm not sure that this is not an error actually

kirugan (approved):

rianhughes (approved):

cmd/juno/juno.go
kirugan: Btw I just checked impl in Viper and seems like they have a but in there. TL;DR: they always return error=nil for this function

PR #1788, opened 2024-04-03
Rename trie/TransactionStorage to Storage (#1783)

nsiregar:
Rename `trie/TransactionStorage` to `Storage` to resolve issue #1783

PR #1937, opened 2024-07-10
Bump version of github actions - Testing from a FORK

derrix060:
- actions/setup-go: Bumping it to v5.
With the v5, cache is done by default, so there is no need to
specify cache: true

- actions/checkout: bumping it to v4
A lot of different improvements, but the one that made me decide
to upgrade was to use node20 instead of node16.
Node16 has en end of life on 11 Sep 2023

- codecov/codecov-action: bumping it to v4 (v4.5.0)
It was hard-coded v4.2.0, meaning that the fixes around detecting
if running from a fork (v4.3, v4.4 and v.4.5) were not backported,
therefore making the job not being able to detect when it was
running from a fork


.github/workflows/upload-codecov.yml
derrix060, line 1: This file won't be necessary anymore since uploading to codecov will happen where it's supposed to

kirugan (approved):

derrix060:
For those who have no idea what's going on, Creating a PR from a forked repo (like this PR) used to fail to upload files to codecov.  For example https://github.com/NethermindEth/juno/pull/1788

This PR passing the all the checks (including `codecov/patch` and `codecov/project`) means that it's working as desired :)

derrix060:
If you are waiting for codecov to write a new message, it doesn't.  They only update the message that they sent before:

image

wojciechos (approved):

PR #1939, opened 2024-07-10
Safely enable uploading report to Codecov

derrix060:
This is done by saving the coverage output as an workflow artifact then loading it on a separate job and uploading it to codecov.

The reason why this is necessary is because when running the juno-test workflow from a fork / untrusted dev, the secrets is not available. This is done in order to secure the secrets to being exposed from an atacker that might want to create a PR and get them. More info: https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

The second workflow will be trigger once the juno-test is completed so it only needs to download the artifact and upload to codecov. Since this workflow source code is only from main, in order for it to contain malicious code, it would've been required to go through a PR.

I've also added a small script to comment to the PR, so it's clear that the codecov will be uploaded shortly.  This can be reverted later on if it starts to become too spammy.

### DIFF
--- a/.github/workflows/juno-test.yml
+++ b/.github/workflows/juno-test.yml
@@ -50,11 +50,10 @@ jobs:
       - name: Benchmark
         run: make benchmarks
 
-      - name: Upload coverage to Codecov
-        uses: codecov/codecov-action@v4.2.0
+      - uses: actions/upload-artifact@v4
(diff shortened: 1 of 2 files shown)


kirugan (approved):

kirugan:
Testing new workflow - merging without waiting for 1 more approval

PR #1940, opened 2024-07-11
onboarding task

onuruci:
Simple sequential sync package.

Onboarding task. Understood the concurrent sync algorithm with workers.


rianhughes:
I'm happy with this solution

PR #1943, opened 2024-07-11
Fix how to upload coverage to codecov

derrix060:
When I copied the Upload coverage to Codecov step from the juno-test I forgot to remove the condition that it would only do it when running on ubuntu-latest machine.

Also, the Comment on PR didn't work, so I'm just removing it since it doesn't provide much benefit


kirugan (approved):

kirugan:
Continue testing pipelines in github - merging without second approval & not finished pipelines for this PR

PR #1944, opened 2024-07-12
Remove duplicate libraries linked on MacOS

kirugan:
These changes fix this warning on MacOS:
```
ld: warning: ignoring duplicate libraries: '-ldl', '-lm'
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,8 @@ endif
 
 ifeq ($(shell uname -s),Darwin)
 	export CGO_LDFLAGS=-framework Foundation -framework SystemConfiguration
+else
+	export CGO_LDFLAGS=-ldl -lm
 endif
 
 rustdeps: vm core-rust compiler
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -12,6 +12,8 @@ endif
 
 ifeq ($(shell uname -s),Darwin)
 	export CGO_LDFLAGS=-framework Foundation -framework SystemConfiguration
```
kirugan, line 14: So `-ldl` is dynamic linking library and `-lm` is math library. Both are not needed for Mac OS because the OS use different linking mechanism and math is part of C stdlib anyway. But we should also check requirment for `-lm` on other platforms though

derrix060 (approved):

Makefile
kirugan: I'll add it as full git message

rianhughes (approved):

PR #1945, opened 2024-07-12
Add telemetry API

kirugan:


cmd/juno/juno.go
```diff
@@ -202,6 +203,7 @@ func main() {
 	})
 
 	if err := cmd.ExecuteContext(ctx); err != nil {
+		fmt.Fprintf(os.Stderr, "error: %v\n", err)
```
IronGauntlets, line 206: I think this is redundant because cobra prints the error

telemetry/service.go
```diff
@@ -0,0 +1,84 @@
...
+
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
+		case <-ticker.C:
+			err := s.sendTelemetry(ctx)
+			if err != nil {
+				s.log.Errorw("Failed to send telemetry", "err", err)
+			}
```
IronGauntlets, line 44: Do you think we should sleep for a bit before the next iteration of the loop?

kirugan: I thought that ticker should do that, though if sending takes 5 sec next request will start immediately. We can update ticker once request is complete, wdyt?

cmd/juno/juno.go
kirugan: Thanks, forgot about it.

Issue #1004, opened 2023-08-04
Websocket for Transaction Receipt 


IronGauntlets:
1. connect with a specific tx_hash, and immediately receive the current known receipt (if any)
2. then new messages for every status change and/or data finality availability (i.e. included block id/hash)

PR #1947, opened 2024-07-12
Attempt at transaction receipts websocket

artem-streltsov:
Attempt at #1004. 

I have followed the implementation of `SubscribeNewHeads` from #1279.

Currently, for demonstration, I send some transaction receipt (not relevant to the given txHash) to subscribers whenever a new block is added. To test it, run:
1. `./build/juno --ws --ws-port 6061 --ws-host 0.0.0.0 --db-path $HOME/snapshots/juno_mainnet/`
2. In a different terminal tab, run `websocat -v ws://localhost:6061`
3. Send a request like `{"jsonrpc": "2.0","method": "juno_subscribeTransactionReceipts","params": {"transaction_hash": "0x18e25397b012015930bd446cedb8e79b89dd848d402104657383d1df786db1b"},"id": 1}`

Now, I think I need to add logic for fetching the actual transaction receipt / transaction status. Should I just poll for it every few seconds when `juno_subscribeTransactionReceipts` is called?

Am I heading in the right direction?

### DIFF
--- a/sync/sync.go
+++ b/sync/sync.go
@@ -34,6 +34,10 @@ type HeaderSubscription struct {
 	*feed.Subscription[*core.Header]
 }
 
+type ReceiptSubscription struct {
+	*feed.Subscription[*core.TransactionReceipt]
+}
+
 // Todo: Since this is also going to be implemented by p2p package we should move this interface to node package
 //
 //go:generate mockgen -destination=../mocks/mock_synchronizer.go -package=mocks -mock_names Reader=MockSyncReader github.com/NethermindEth/juno/sync Reader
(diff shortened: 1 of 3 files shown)


sync/sync.go
```diff
@@ -229,6 +240,7 @@ func (s *Synchronizer) verifierTask(ctx context.Context, block *core.Block, stat
 			}
 
 			s.newHeads.Send(block.Header)
+			s.newReceipts.Send(block.Receipts[0])
```
artem-streltsov, line 243: Need to move this somewhere where it will make more sense and send the actual transaction receipt for the given transaction hash. Any ideas?

PR #1953, opened 2024-07-16
Separate tests with coverage and without coverage

derrix060:
Running code coverage slows down the run by ~20%.
The commit separates the tests into two different jobs based on the operating system. The "Tests (Coverage)" job runs the tests with coverage and is only triggered on the "ubuntu-latest" OS, while the "Tests (No Coverage)" job runs the tests without coverage on other OSes. This change improves the test workflow and ensures that coverage only runs on the os that does uploads the result to Codecov.


.github/workflows/juno-test.yml
```diff
@@ -42,8 +42,13 @@ jobs:
...
-        
-      - name: Tests
+
+      - name: Tests (Coverage)
+        if: matrix.os == 'ubuntu-latest'
+        run: make test-cover
+
+      - name: Tests (No Coverage)
+        if: matrix.os != 'ubuntu-latest'
         run: make test-cover
```
wojciechos, line 52: Shouldn't be `make test` here?

.github/workflows/juno-test.yml
```diff
@@ -42,8 +42,13 @@ jobs:
...
         run: brew install jemalloc
-        
-      - name: Tests
+
+      - name: Tests (Coverage)
+        if: matrix.os == 'ubuntu-latest'
+        run: make test-cover
+
+      - name: Tests (No Coverage)
+        if: matrix.os != 'ubuntu-latest'
```
kirugan, line 51: also question regarding target definition:
```
test: clean-testcache rustdeps ## tests
	go test $(GO_TAGS) ./...

test-cover: rustdeps ## tests with coverage
        mkdir -p coverage
        go test $(GO_TAGS) -coverpkg=./... -coverprofile=coverage/coverage.out -covermode=atomic ./...
        go tool cover -html=coverage/coverage.out -o coverage/coverage.html
```

kirugan:
Quick comment so we won't forget: we need to change settings of repo to check for new pipeline names

.github/workflows/juno-test.yml
derrix060: 🤦

.github/workflows/juno-test.yml
derrix060: seems like it make sense, let me add it

kirugan (approved):

wojciechos (approved):

PR #1954, opened 2024-07-16
Daniil/rewrite-functional-iterators-into-for-loop-statement

### LINKED ISSUE #1952: Rewrite functional iterators into for loop statement
In go 1.23 [they accepted](https://tip.golang.org/doc/go1.23) functional iterators. We haven't used for loop statement because it wasn't clear what final interface would look like. 
One can find all functional iterators by searching usages of `iter.Seq` or `iter.Seq2` types.
Example:
```
eventsIt(func(res *spec.EventsResponse) bool {
	switch v := res.EventMessage.(type) {
	case *spec.EventsResponse_Event:
		events = append(events, v.Event)
		return true
	case *spec.EventsResponse_Fin:
		return false
	default:
		s.log.Warnw("Unexpected EventMessage from getEvents", "v", v)
		return false
	}
})
```
to be:
```
loop:
for res := range eventsIt {
	switch v := res.EventMessage.(type) {
	case *spec.EventsResponse_Event:
		events = append(events, v.Event)
	case *spec.EventsResponse_Fin:
                break loop // mind the loop label, because just break key word doesn't affect iteration
	default:
		s.log.Warnw("Unexpected EventMessage from getEvents", "v", v)
		break loop
	}
})
```
or
```
for res := range eventsIt {
	switch v := res.EventMessage.(type) {
	case *spec.EventsResponse_Event:
		events = append(events, v.Event)
                continue // as far as I remember continue works fine inside select
	case *spec.EventsResponse_Fin:
                // will break in the end
	default:
		s.log.Warnw("Unexpected EventMessage from getEvents", "v", v)
	}
        // comment why
	break
})
```

P.S. First you need to add `GOEXPERIMENT=rangefunc` env var. I suggest to put it on top of Makefile

AnkushinDaniil:
fixes #1952


Makefile
```diff
@@ -1,5 +1,7 @@
 .DEFAULT_GOAL := help
 
+export GOEXPERIMENT=rangefunc
```
kirugan, line 3: let's add comment on top `todo remove it after update to go 1.23`

p2p/sync.go
```diff
@@ -409,26 +409,25 @@ func (s *syncService) genHeadersAndSigs(ctx context.Context, blockNumber uint64)
...
 			default:
 				s.log.Warnw("Unexpected HeaderMessage from getBlockHeaders", "v", v)
-				return false
+				break loop
 			}
 
 			select {
 			case <-ctx.Done():
-				return false
+				break loop
```
kirugan, line 427: label was not necessary, but let's keep it as is

AnkushinDaniil: [done](https://github.com/NethermindEth/juno/pull/1954/commits/3e10bbbb4d92e126b5da30f74ab415f41527032e)

Makefile
AnkushinDaniil: [done](https://github.com/NethermindEth/juno/pull/1954/commits/3e10bbbb4d92e126b5da30f74ab415f41527032e)

kirugan: comments in Makefile starts with `#`

kirugan (approved):

derrix060 (changes requested):
Failing on lint

AnkushinDaniil:
Currently [golangci-lint does not support GOEXPERIMENT=rangefunc](https://github.com/golangci/golangci-lint/issues/4627), we have to wait for an update.

IronGauntlets (approved):

PR #1955, opened 2024-07-17
Implement Juno sequencer

rianhughes:
This PR is to merge the sequencer logic into main. 

Currently it supports two modes, `make sequencer` which will start Juno with an empty state (which can't really be used for anything given we need accounts and fee tokens to submit transactions), and `make sequencer-with-accounts` which builds a genesis state with some pre-funded accounts.


rianhughes:
Actually, I don't think main need this right now. It's more relevant for the sequencer, so closing for now

PR #1956, opened 2024-07-17
Add test race detection to pipeline

wojciechos:


derrix060 (approved):

derrix060:
I think that currently the tests are failing, but if they are not I'm happy with the change :)

utils/throttler_test.go
```diff
@@ -13,18 +15,22 @@ import (
 func TestThrottler(t *testing.T) {
```
kirugan, line 15: In practice this test was working fine, but (and I agree with -race) it could lead to race conditions

sync/sync_test.go
```diff
@@ -142,7 +137,6 @@ func TestSyncBlocks(t *testing.T) {
 }
 
 func TestReorg(t *testing.T) {
-	t.Parallel()
```
kirugan, line 145: @IronGauntlets if I leave `t.Parallel()` here and in `TestSyncBlocks` (top level) it leads from time to time to bugs (idk why :( )

IronGauntlets: I'm happy for you to remove it. I am not sure why this happens.

IronGauntlets (approved):

PR #1965, opened 2024-07-23
P2P sync update

IronGauntlets:


core/state_update.go
```diff
@@ -35,6 +35,21 @@ func EmptyStateDiff() *StateDiff {
...
+	var length uint64
+
+	for _, storageDiff := range d.StorageDiffs {
+		length += uint64(len(storageDiff))
+	}
+	length += uint64(len(d.Nonces))
+	length += uint64(len(d.DeployedContracts))
+	length += uint64(len(d.DeclaredV0Classes))
+	length += uint64(len(d.DeclaredV1Classes))
+	length += uint64(len(d.ReplacedClasses))
```
kirugan, line 48: maybe we should make `length` var as int and cast it into `uint64` at the bottom?
I don't think that in practice sum of all these fields will reach 2^32 (max value of int)

adapters/core2p2p/block.go
```diff
@@ -44,12 +46,15 @@ func AdaptHeader(header *core.Header, commitments *core.BlockCommitments) *spec.
 			Root:    AdaptHash(commitments.EventCommitment),
 		},
 		// todo fill receipts
```
kirugan, line 48: do we need to replace this comment?

adapters/core2p2p/block.go
```diff
@@ -44,12 +46,15 @@ func AdaptHeader(header *core.Header, commitments *core.BlockCommitments) *spec.
...
-		Receipts:        nil,
+		Receipts:        AdaptHash(&felt.Zero),
 		ProtocolVersion: header.ProtocolVersion,
 		GasPriceFri:     AdaptUint128(header.GasPrice),
 		Signatures:      utils.Map(header.Signatures, AdaptSignature),
 		// todo(kirill) set these fields
```
kirugan, line 53: and remove this one ?

adapters/core2p2p/block.go
IronGauntlets: Replaced with a more specific comment

adapters/core2p2p/block.go
IronGauntlets: Done

core/state_update.go
IronGauntlets: Done

kirugan (approved):

rianhughes (approved):

PR #1978, opened 2024-07-26
Use embed directive to find testdata directory

weiihann:
### Description
This PR modifies the implementation of finding the target test data directory using the embed directive.

### Rationale
The existing implementation relies on the current working directory, which has a dependency on where the binary is run. Errors can occur if the binary is not run where it's supposed to be, causing confusion. Using an embed directive ensures consistent results regardless of where the binary file is run.

Another issue is that if a user changes the name of the repository from `juno` to whatever, the tests will fail. Also, any tests that rely on `NewTestClient` in `cmd/juno` will fail because the base directory extracted will be `../juno/cmd/juno` instead of `../juno`.

Using an embed directive increases the size of the binary file. I did a comparison of the binary size before and after:
Before: 101133458 byte
After: 101133522 bytes
Difference: 64 bytes

### DIFF
--- a/clients/feeder/feeder.go
+++ b/clients/feeder/feeder.go
@@ -2,13 +2,13 @@ package feeder
 
 import (
 	"context"
+	"embed"
 	"encoding/json"
 	"errors"
 	"io"
 	"net/http"
 	"net/http/httptest"
 	"net/url"
-	"os"
 	"path/filepath"
 	"strings"
 	"testing"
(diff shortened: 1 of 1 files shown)


clients/feeder/feeder.go
```diff
@@ -18,10 +18,14 @@ import (
 	"github.com/NethermindEth/juno/starknet"
 	"github.com/NethermindEth/juno/utils"
 	"github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/require"
 )
 
-var ErrDeprecatedCompiledClass = errors.New("deprecated compiled class")
+var (
+	//go:embed testdata
+	testData embed.FS
```
kirugan, line 25: I don't think we need it because this data will be part of produced binary (not test one, but the one we use in production and releases)

weiihann:
Closing this, using embed sounds a little too scary even though the compiler did optimize and exclude the test files.

PR #1982, opened 2024-07-29
Add fuzz tests for utils

AnkushinDaniil:
No bugs found

### DIFF
--- a/utils/compression_test.go
+++ b/utils/compression_test.go
@@ -19,3 +19,13 @@ func TestGzip64(t *testing.T) {
 	require.NoError(t, err)
 	assert.Equal(t, bytes, decompBytes)
 }
+
+func FuzzGzip64(f *testing.F) {
+	f.Fuzz(func(t *testing.T, data []byte) {
+		compressed, err := utils.Gzip64Encode(data)
+		require.NoError(t, err)
+		decompressed, err := utils.Gzip64Decode(compressed)
+		require.NoError(t, err)
+		assert.Equal(t, data, decompressed)
+	})
+}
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -160,3 +160,6 @@ pathfinder: juno-cached
     	--p2p-peers=/ip4/127.0.0.1/tcp/8888/p2p/12D3KooWF1JrZWQoBiBSjsFSuLbDiDvqcmJQRLaFQLmpVkHA9duk \
     	--p2p-private-key="54a695e2a5d5717d5ba8730efcafe6f17251a1955733cffc55a4085fbf7f5d2c1b4009314092069ef7ca9b364ce3eb3072531c64dfb2799c6bad76720a5bdff0" \
     	--metrics-port=9094
+
+test-fuzz: ## run fuzzing script
```
derrix060, line 164: should we start running those in CI as well?

AnkushinDaniil: We can, but there are some caveats:
- It runs one test at a time, which means the duration is $$n\times k$$ where $$n$$ - seconds for one test, $$k$$ - number of tests.
- If it fails, it creates a file with fail input and this test will be executed in case of `go test`.

kirugan (approved):

Makefile
kirugan: @derrix060 there is something called "OSS-Fuzz". As far as I understand it's 24/7 fuzzing infra - you can use something like this I guess. See https://github.com/golang/go/issues/50192 for more info

IronGauntlets (approved):

PR #1984, opened 2024-07-30
merge and split proofs

onuruci:
Implemented merge and split proof logic. 

Takes left and right path and returns a mergedPath without any duplicates and all elements included.

Split logic does the reverse and returns exact left and right paths before the merge.

#1898


onuruci:
Split logic assumes that paths were merged in a specified format.

[...commonNodes, leftNode, rightNode, ...]

onuruci:
As for tests,

MergePaths tests checks that merged path does not contain any duplicates and contains all values in left and right paths.

SplitPaths tests checks that splitted left path and right path are the same with the ones merged together.

rianhughes:
> Split logic assumes that paths were merged in a specified format.
I believe the merged proof nodes may be unordered unfortunately

onuruci:
Then how are we going to split them to left and right paths when it is unordered. To explain it better, only order the merged proof should keep is that first it should have all common nodes, than first split left node and first split right node then ordering does not matter of course preserving the order of the paths root to node.

[...commonNodes, firstLeftNode, firstRightNode, ...restOfNodes ( keeping the order of root to node )]

This way we can capture where the split happens and build right and left paths accurately.

pnowosie:
## Conclusion regarding the `ProofNodes` ordering:

- `MergePaths ` - we can return the suggested array of: `[...commonNodes, leftNode, rightNode, ...]` or anything which seems to be more convenient, because we do not impose specific ordering (more 👇 )|
- `MergePaths` cannot assume any ordering but assumes we can detect the root node and then move along.

## Root node detection algorithm: 
**Given:** 
- `mergedPath`the merged proof paths,
- `rootHash` - the node hash at the root node

1. Populate the map `nodeHashes map[Felt]ProofNode` by calculating the hash of each node in `mergedPath`
2. Get the root node: `rootNode := nodeHashes[rootHash]`
3. Following the root node we can discover the very next node / nodes
  - if `rootNode is Edge` the next child is ``

core/trie/proof.go
```diff
@@ -133,6 +133,163 @@ func transformNode(tri *Trie, parentKey *Key, sNode StorageNode) (*Edge, *Binary
...
+
+func SplitProofPath(mergedPath []ProofNode, rootHash *felt.Felt, hash hashFunc) ([]ProofNode, []ProofNode, error) {
+	leftPath := []ProofNode{}
+	rightPath := []ProofNode{}
+
+	if len(mergedPath) == 0 {
+		return leftPath, rightPath, nil
+	}
+
+	nodeHashes := make(map[string]ProofNode)
```
pnowosie, line 223: I think felt (result of hash) should work in map as well?
```suggestion
	nodeHashes := make(map[felt.Felt]ProofNode)
```

core/trie/proof.go
```diff
@@ -133,6 +133,163 @@ func transformNode(tri *Trie, parentKey *Key, sNode StorageNode) (*Edge, *Binary
 	return edge, binary, nil
 }
 
+func traverseNodesAndAppendToPath(path *[]ProofNode, currNode *ProofNode, nodeHashes map[string]ProofNode) error {
```
pnowosie, line 136: this function & SplitPath have a lot code in common which makes me think something is way overcomplicated 🤔

core/trie/proof_test.go
```diff
@@ -252,6 +252,45 @@ func build4KeyTrie(t *testing.T) *trie.Trie {
...
+		seen[pNode.Hash(crypto.Pedersen).String()] = true
+	}
+	return true
+}
+
+func containsAll(proofNodes, expectedProofNodes []trie.ProofNode) bool {
+	for _, pNode := range expectedProofNodes {
+		found := false
+		for _, p := range proofNodes {
+			if p.Hash(crypto.Pedersen).String() == pNode.Hash(crypto.Pedersen).String() {
```
pnowosie, line 270: `Felt.Equal` method should be used here

core/trie/proof.go
```diff
@@ -133,6 +133,163 @@ func transformNode(tri *Trie, parentKey *Key, sNode StorageNode) (*Edge, *Binary
...
+	whileLoop := true
+	for whileLoop {
+		switch {
+		case currNode.Binary != nil:
+			expectedLeftHash := currNode.Binary.LeftHash
+			expectedRightHash := currNode.Binary.RightHash
+
+			nodeLeft, leftExist := nodeHashes[expectedLeftHash.String()]
+			nodeRight, rightExist := nodeHashes[expectedRightHash.String()]
+			if leftExist && rightExist {
```
pnowosie, line 247: Both left & right exist so this is the last "common" node.
What we do next is we're giving **this node** to `traverseNodesAndAppendToPath` (line 281 & 285 👇 )
and then in traverse it should blow up on this:

https://github.com/NethermindEth/juno/blob/e4c49a46cea749649969fb9999cb8688c4272fef/core/trie/proof.go#L146-L147

So this tells me we have no test with common path 🤔

pnowosie (changes requested):
See comment in Split... function - something tells me that we haven't tested it properly.
...and the code duplication might suggest the whole thing is overcomplicated.

I will try to think about better approach

core/trie/proof.go
onuruci: Yes we don't have a test with common path I will add it but also In common paths I guess both left and right paths will not exist in the same time since there will be no split

onuruci: This case actually shouldn't happen in traverse.

pnowosie: Yes - my bad, split appends not the binary node that has both of left&right but the left & right themselves - so traverse works on them

pnowosie:
@onuruci I've added additional logging to the test `TestSplitProofPaths/4Key_Trie_two_common_ancestors...` and I think I now have a clarity how to implement split in more readable fashion.

```
* Proofs *
**********
[Left]
0: Edge (0x1f4e0)
1: Binary 0x768de - 0x4fe3b
2: Binary 0x57166 - 0x60fbd
3: Binary 0x4 - 0x5 (Hash: 0x57166 )
[Right]
0: Edge (0x1f4e0)
1: Binary 0x768de - 0x4fe3b
2: Binary 0x57166 - 0x60fbd
3: Edge (0x6) (Hash: 0x60fbd )

* Merged *
**********
0: Edge (0x1f4e0)
1: Binary 0x768de - 0x4fe3b
2: Binary 0x57166 - 0x60fbd
3: Binary 0x4 - 0x5 (Hash: 0x57166 )
4: Edge (0x6) (Hash: 0x60fbd )

* Splits *
**********
[Left]
0: Edge (0x1f4e0)
1: Binary 0x768de - 0x4fe3b
2: Binary 0x57166 - 0x60fbd
3: Binary 0x4 - 0x5 (Hash: 0x57166 )
[Right]
0: Edge (0x1f4e0)
1: Binary 0x768de - 0x4fe3b
2: Binary 0x57166 - 0x60fbd
3: Edge (0x6) (Hash: 0x60fbd )
```

What we can observe is (not only about this case, but in general):
- there will always be at least one common node (root),
- then along the path split happens dividing nodes into left and right paths
- no more common nodes are possible after the split - if we split higher the trie the paths never meet later together
- ...so, there will be only one "split" node, the binary node that both left & right children are in hash map.

onuruci:
Current implementation is very similiar with the implementation what you last explained. Though we have code repetition, we can handle all traverse operations in traverse function. I think what you explained is the correct way since we need to traverse the nodes three times one for common nodes until we find a split and two other for left and right paths.

In tests we generate two paths, merge and split them we check if the splitted ones are same with the merged ones. I will add the test case you mentioned also will add more test cases with shuffled mergePaths. I also will add more comments and enhance the readability.

pnowosie:
> Current implementation is very similiar with the implementation what you last explained. Though we have code repetition, we can handle all traverse operations in traverse function. I think what you explained is the correct way since we need to traverse the nodes three times one for common nodes until we find a split and two other for left and right paths.

Yes, but recursive function would greatly simplify the "while loop", and also we should assume data correctness and do some hash nodes map validation before we call traverse.

So the traverse would be trivial: 
- find a node that corresponds given hash, 
- add it to the given slice, 
- recursively call itself on the existing child. 

If node is binary and both children exists in map, it adds the node  and stop, returning the node hash - so we can continue calling for left and right paths.

core/trie/proof.go
onuruci: We are using felt.Felt with the new commit

core/trie/proof.go
onuruci: Changed this with recursive approach which traverses all paths

core/trie/proof.go
```diff
@@ -133,39 +133,34 @@ func transformNode(tri *Trie, parentKey *Key, sNode StorageNode) (*Edge, *Binary
...
-				currNode = &righNode
-				*path = append(*path, *currNode)
-			} else {
-				whileLoop = false
-			}
+// Begins from the root node and traverses the merged proof path
+// Until it finds a split node, adds nodes to commonPath
+// Then continues with traversing left and right paths separetaly
+// Assumes there is no circular paths and the split happens at most once
+func traverseNodes(currNode *ProofNode, path, leftPath, rightPath *[]ProofNode, nodeHashes map[felt.Felt]ProofNode) {
```
pnowosie, line 140: I still think you try to do too much here :)
How about:
```go
func traverseNodes(currHash *felt.Felt, path *[]ProofNode, nodeHashes map[felt.Felt]ProofNode) {
... just populate one slice here
    if leftExist && rightExist { return }
...
```

so at the bottom of `Split...` func where you copy common to left and rigth you could:
```go
traverseNodes(rootHash, &commonPath, nodeHashes)
// this stops on split node which is the last in commonPath

splitNode := commonPath[-1] // forgive me pythonic code 😛
traverseNodes(&splitNode.Binary.LeftHash, &leftPath, nodeHashes)
traverseNodes(&splitNode.Binary.RightHash, &rightPath, nodeHashes)

// copy commons to both left & right
```

Does it make sense or I'm missing something? 🤔

pnowosie (changes requested):
It's close to complete ;) Best luck!

core/trie/proof.go
```diff
@@ -212,82 +207,71 @@ func MergeProofPaths(leftPath, rightPath []ProofNode, hash hashFunc) ([]ProofNod
...
+		}
+
+		nodeHashes[*nodeHash] = node
 	}
 
-	leftPath = append(leftPath, currNode)
-	rightPath = append(rightPath, currNode)
+	// validates that the split happens at most once
+	splitHappened := false
+	for _, node := range mergedPath {
```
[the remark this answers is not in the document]
onuruci: I separated checks to different functions and built a check for cycles. It takes the root hash and traverses through existent nodes in the merged path. If it visits a node more than once which will show us that there is a cycle it returns with an error.

core/trie/proof.go
```diff
@@ -133,6 +133,206 @@ func transformNode(tri *Trie, parentKey *Key, sNode StorageNode) (*Edge, *Binary
 	return edge, binary, nil
 }
 
+// Checks if there happens at most one split in the merged path
+func splitCheck(mergedPath []ProofNode, nodeHashes map[felt.Felt]ProofNode) error {
```
pnowosie, line 137:
```suggestion
func pathSplitOccurredCheck(mergedPath []ProofNode, nodeHashes map[felt.Felt]ProofNode) error {
```

core/trie/proof.go
```diff
@@ -133,6 +133,206 @@ func transformNode(tri *Trie, parentKey *Key, sNode StorageNode) (*Edge, *Binary
...
+			if err := noCycleCheck(edgeNode, nodeHashes, visitedHashes, hash); err != nil {
+				return err
+			}
+		}
+	}
+
+	return nil
+}
+
+func rootExists(rootHash *felt.Felt, nodeHashes map[felt.Felt]ProofNode) (ProofNode, error) {
```
pnowosie, line 201:
```suggestion
func rootNodeExistsCheck(rootHash *felt.Felt, nodeHashes map[felt.Felt]ProofNode) (ProofNode, error) {
```

pnowosie (commented):
I will redo CR - just posted some renames

core/trie/proof.go
```diff
@@ -133,6 +133,206 @@ func transformNode(tri *Trie, parentKey *Key, sNode StorageNode) (*Edge, *Binary
...
+		return currNode, errors.New("root hash not found in the merged path")
+	}
+
+	return currNode, nil
+}
+
+// Begins from the root node and traverses the merged proof path
+// Until it finds a split node, adds nodes to commonPath
+// Then continues with traversing left and right paths separetaly
+// Assumes there is no circular paths and the split happens at most once
```
pnowosie, line 213: This gets outdated.
```suggestion
// traverseNodes traverses the merged proof path starting at `currNode` 
// and adds nodes to `path` slice. It stops when the split node is added 
// or the path is exhausted, and `currNode` children are not included
//  in the path (nodeHashes)
```

core/trie/proof.go
```diff
@@ -133,6 +133,206 @@ func transformNode(tri *Trie, parentKey *Key, sNode StorageNode) (*Edge, *Binary
...
+		if leftExist && rightExist {
+			return
+		} else if leftExist {
+			traverseNodes(nodeLeft, path, nodeHashes)
+		} else if rightExist {
+			traverseNodes(nodeRight, path, nodeHashes)
+		}
+	} else if currNode.Edge != nil {
+		edgeNode, exist := nodeHashes[*currNode.Edge.Child]
+		if !exist {
```
pnowosie, line 232: reverse the condition (so it much the prev 👆 )
```suggestion
		if exist {
		    traverseNodes(edgeNode, path, nodeHashes)
		}
```

core/trie/proof.go
```diff
@@ -133,6 +133,206 @@ func transformNode(tri *Trie, parentKey *Key, sNode StorageNode) (*Edge, *Binary
...
+			return
+		}
+		traverseNodes(edgeNode, path, nodeHashes)
+	}
+}
+
+// Remove duplicates and merges proof paths into a single path
```
pnowosie, line 239:
```suggestion
// MergeProofPaths removes duplicates and merges proof paths into a single path
```
also add "this produces the nodes in specific order - [ common, leftPath, rightPath ] "...
but order is not important hence the `Split...` can discover ... yeah something like this - as we both forget about this detail :P sooner than later

pnowosie (approved):
Ok, this is very close - so I can ✅ now
Please make sure that function's comments follows the go convention:
```go
// funcName does something ...
func funcName() {
```

as a low hanging fruit is to implement simple tests that returns errors in the places marked by "codecov"

onuruci:
I adjusted function comments and covered remaining test cases but the one with the circular merge path because I couldn't produce a circular path. The issue is that we don't use pointers for reaching child nodes instead we keep hash of it in the ProodNode. As we point nodes to each other hash of the node changes thus circle breaks. Here is an example of it.

Let's say we have three nodes ```p1, p2, p3``` and we want to produce a circle from it.
We simply point them to each other first ```p2->p3``` then ```p1->p3``` and this gives us ```p1->p2->p3```
In order to have a cycle we need ```p3->p1``` but when we make ```p3```to point ```p1``` the hash of the ```p3```changes and ```p2->p3```will no longer be valid.

In order to apply this test I can build a dummy hash function where we can easily predict values and use that in this test case to build a circular path.

pnowosie:
> Let's say we have three nodes `p1, p2, p3` and we want to produce a circle from it.

Oh - I can see a point: we calculate hash of the node which points to previous node `A`, but then we need to adjust the reference in it's parent & parent parent... - popping up to the `A` itself. This is clever - I did not think about this 👍

I think we can remove the `noCycleCheck(...)` and add a comment in `Split...()` that cycles are impossible.

core/trie/proof.go
```diff
@@ -133,6 +133,213 @@ func transformNode(tri *Trie, parentKey *Key, sNode StorageNode) (*Edge, *Binary
...
+	if len(mergedPath) == 0 {
+		return leftPath, rightPath, nil
+	}
+
+	currNode, err := rootNodeExistsCheck(rootHash, nodeHashes)
+	if err != nil {
+		return leftPath, rightPath, err
+	}
+
+	visitedHashes := make(map[felt.Felt]bool)
```
rianhughes, line 318: can this be pushed into noCycleCheck?

core/trie/proof.go
```diff
@@ -133,6 +133,213 @@ func transformNode(tri *Trie, parentKey *Key, sNode StorageNode) (*Edge, *Binary
...
+	}
+
+	visitedHashes[*nodeHash] = true
+
+	if node.Binary != nil {
+		leftHash := node.Binary.LeftHash
+		rightHash := node.Binary.RightHash
+
+		leftNode, leftExist := nodeHashes[*leftHash]
+		rightNode, rightExist := nodeHashes[*rightHash]
```
rianhughes, line 181: nit:
remove 
leftHash := node.Binary.LeftHash
rightHash := node.Binary.RightHash
just set directly
leftNode, leftExist := nodeHashes[*node.Binary.LeftHash]
rightNode, rightExist := nodeHashes[*node.Binary.RightHash]

core/trie/proof.go
```diff
@@ -133,6 +133,213 @@ func transformNode(tri *Trie, parentKey *Key, sNode StorageNode) (*Edge, *Binary
...
+		if node.Edge != nil {
+			continue
+		}
+
+		leftHash := node.Binary.LeftHash
+		rightHash := node.Binary.RightHash
+
+		_, leftExists := nodeHashes[*leftHash]
+		_, rightExists := nodeHashes[*rightHash]
+
```
rianhughes, line 151: nit: same as previous nit

core/trie/proof.go
```diff
@@ -133,6 +133,213 @@ func transformNode(tri *Trie, parentKey *Key, sNode StorageNode) (*Edge, *Binary
...
+// or the path is exhausted, and `currNode` children are not included
+// in the path (nodeHashes)
+func traverseNodes(currNode ProofNode, path *[]ProofNode, nodeHashes map[felt.Felt]ProofNode) {
+	*path = append(*path, currNode)
+
+	if currNode.Binary != nil {
+		expectedLeftHash := currNode.Binary.LeftHash
+		expectedRightHash := currNode.Binary.RightHash
+
+		nodeLeft, leftExist := nodeHashes[*expectedLeftHash]
```
rianhughes, line 226: nit: and here :)

core/trie/proof.go
```diff
@@ -133,6 +133,213 @@ func transformNode(tri *Trie, parentKey *Key, sNode StorageNode) (*Edge, *Binary
...
+	for ; i < minLen; i++ {
+		merged = append(merged, leftPath[i], rightPath[i])
+	}
+
+	// Add the rest of the nodes from the longest path
+	if len(leftPath) > minLen {
+		merged = append(merged, leftPath[i:]...)
+	} else if len(rightPath) > minLen {
+		merged = append(merged, rightPath[i:]...)
+	}
```
rianhughes, line 285: can we simplify to this?
merged = append(merged, leftPath[i:]...)
merged = append(merged, rightPath[i:]...)

core/trie/proof_test.go
```diff
@@ -869,3 +918,244 @@ func TestVerifyRangeProof(t *testing.T) {
...
+
+		mergedProofs, _, err := trie.MergeProofPaths(proofs[0], proofs[1], crypto.Pedersen)
+		require.NoError(t, err)
+
+		containsLeft := containsAll(mergedProofs, proofs[0])
+		require.True(t, containsLeft)
+		containsRight := containsAll(mergedProofs, proofs[1])
+		require.True(t, containsRight)
+		checkDuplicates := noDuplicates(mergedProofs)
+		require.True(t, checkDuplicates)
```
rianhughes, line 945: nit:
require.True(t, containsAll(mergedProofs, proofs[0]))
		require.True(t, containsAll(mergedProofs, proofs[1]))
		require.True(t, noDuplicates(mergedProofs))

core/trie/proof.go
onuruci: Yes it was trying to follow a format since we are not using this format anymore and ordering is not important in splitProof we can simplify this

core/trie/proof_test.go
```diff
@@ -252,6 +252,55 @@ func build4KeyTrie(t *testing.T) *trie.Trie {
 	return tempTrie
 }
 
+func noDuplicates(proofNodes []trie.ProofNode) bool {
+	seen := make(map[string]bool)
+	for _, pNode := range proofNodes {
```
rianhughes, line 257: I think we can set the len here 	seen := make(map[string]bool, len(proofNodes))
Also, just curious, why map[string]bool instead of map[*felt.Felt]bool? Saves you calling the String() method and is a tiny bit more readable.

onuruci: We changed it to felt.Felt in code but I forgot to change this in tests

core/trie/proof_test.go
rianhughes: same for all tests below

rianhughes:
Nice work!

rianhughes (approved):

PR #1989, opened 2024-07-31
Add execution stepcount in header - X-Cairo-Steps

obasekiosa:
Todo:

- [ ] Find create rpc header code

### DIFF
--- a/rpc/simulation.go
+++ b/rpc/simulation.go
@@ -106,7 +106,7 @@ func (h *Handler) simulateTransactions(id BlockID, transactions []BroadcastedTra
 		BlockHashToBeRevealed: blockHashToBeRevealed,
 	}
 	useBlobData := !v0_6Response
-	overallFees, dataGasConsumed, traces, err := h.vm.Execute(txns, classes, paidFeesOnL1, &blockInfo,
+	overallFees, dataGasConsumed, traces, numSteps, err := h.vm.Execute(txns, classes, paidFeesOnL1, &blockInfo,
 		state, h.bcReader.Network(), skipFeeCharge, skipValidate, errOnRevert, useBlobData)
 	if err != nil {
 		if errors.Is(err, utils.ErrResourceBusy) {
--- a/rpc/trace.go
+++ b/rpc/trace.go
@@ -262,7 +262,7 @@ func (h *Handler) traceBlockTransactions(ctx context.Context, block *core.Block,
 	}
 
 	useBlobData := !v0_6Response
-	overallFees, dataGasConsumed, traces, err := h.vm.Execute(block.Transactions, classes, paidFeesOnL1, &blockInfo, state, network, false,
+	overallFees, dataGasConsumed, traces, numSteps, err := h.vm.Execute(block.Transactions, classes, paidFeesOnL1, &blockInfo, state, network, false,
 		false, false, useBlobData)
 	if err != nil {
 		if errors.Is(err, utils.ErrResourceBusy) {
--- a/vm/rust/src/lib.rs
+++ b/vm/rust/src/lib.rs
@@ -61,6 +61,7 @@ extern "C" {
     fn JunoAppendResponse(reader_handle: usize, ptr: *const c_uchar);
     fn JunoAppendActualFee(reader_handle: usize, ptr: *const c_uchar);
     fn JunoAppendDataGasConsumed(reader_handle: usize, ptr: *const c_uchar);
+    fn JunoAddExecutionSteps(reader_handle: usize, execSteps: c_longlong);
 }
 
 #[repr(C)]
@@ -348,6 +349,7 @@ pub extern "C" fn cairoVMExecute(
 
                 let actual_fee = t.transaction_receipt.fee.0.into();
                 let data_gas_consumed = t.transaction_receipt.da_gas.l1_data_gas.into();
+                let execution_steps = t.n_steps; // todo: n_steps may not be in this object check later
 
                 let trace =
                     jsonrpc::new_transaction_trace(&txn_and_query_bit.txn, t, &mut txn_state);
@@ -370,6 +372,7 @@ pub extern "C" fn cairoVMExecute(
                         reader_handle,
                         felt_to_byte_array(&data_gas_consumed).as_ptr(),
                     );
+                    JunoAddExecutionSteps(reader_handle, execution_steps)
                 }
                 append_trace(reader_handle, trace.as_ref().unwrap(), &mut trace_buffer);
             }
--- a/vm/vm.go
+++ b/vm/vm.go
@@ -95,6 +95,7 @@ type callContext struct {
 	actualFees      []*felt.Felt
 	traces          []json.RawMessage
 	dataGasConsumed []*felt.Felt
+	executionSteps  uint64
 }
 
 func unwrapContext(readerHandle C.uintptr_t) *callContext {
@@ -136,6 +137,12 @@ func JunoAppendActualFee(readerHandle C.uintptr_t, ptr unsafe.Pointer) {
 func JunoAppendDataGasConsumed(readerHandle C.uintptr_t, ptr unsafe.Pointer) {
 	context := unwrapContext(readerHandle)
 	context.dataGasConsumed = append(context.dataGasConsumed, makeFeltFromPtr(ptr))
+}
+
+//export JunoAddExecutionSteps
+func JunoAddExecutionSteps(readerHandle C.uintptr_t, execSteps C.long) {
+	context := unwrapContext(readerHandle)
+	context.executionSteps += uint64(execSteps) // todo: assumes initial default value is zero even in ffi
 }
 
 func makeFeltFromPtr(ptr unsafe.Pointer) *felt.Felt {
--- a/vm/vm.go
+++ b/vm/vm.go
@@ -65,7 +65,7 @@ type VM interface {
 		maxSteps uint64, useBlobData bool) ([]*felt.Felt, error)
 	Execute(txns []core.Transaction, declaredClasses []core.Class, paidFeesOnL1 []*felt.Felt, blockInfo *BlockInfo,
 		state core.StateReader, network *utils.Network, skipChargeFee, skipValidate, errOnRevert, useBlobData bool,
-	) ([]*felt.Felt, []*felt.Felt, []TransactionTrace, error)
+	) ([]*felt.Felt, []*felt.Felt, []TransactionTrace, uint64, error)
 }
 
 type vm struct {
@@ -267,7 +267,7 @@ func (v *vm) Call(callInfo *CallInfo, blockInfo *BlockInfo, state core.StateRead
 func (v *vm) Execute(txns []core.Transaction, declaredClasses []core.Class, paidFeesOnL1 []*felt.Felt,
 	blockInfo *BlockInfo, state core.StateReader, network *utils.Network,
 	skipChargeFee, skipValidate, errOnRevert, useBlobData bool,
-) ([]*felt.Felt, []*felt.Felt, []TransactionTrace, error) {
+) ([]*felt.Felt, []*felt.Felt, []TransactionTrace, uint64, error) {
 	context := &callContext{
 		state: state,
 		log:   v.log,
@@ -277,12 +277,12 @@ func (v *vm) Execute(txns []core.Transaction, declaredClasses []core.Class, paid
 
 	txnsJSON, classesJSON, err := marshalTxnsAndDeclaredClasses(txns, declaredClasses)
 	if err != nil {
-		return nil, nil, nil, err
+		return nil, nil, nil, 0, err
 	}
 
 	paidFeesOnL1Bytes, err := json.Marshal(paidFeesOnL1)
 	if err != nil {
-		return nil, nil, nil, err
+		return nil, nil, nil, 0, err
 	}
 
 	paidFeesOnL1CStr := cstring(paidFeesOnL1Bytes)
@@ -331,23 +331,23 @@ func (v *vm) Execute(txns []core.Transaction, declaredClasses []core.Class, paid
 
 	if context.err != "" {
 		if context.errTxnIndex >= 0 {
-			return nil, nil, nil, TransactionExecutionError{
+			return nil, nil, nil, context.executionSteps, TransactionExecutionError{
 				Index: uint64(context.errTxnIndex),
 				Cause: errors.New(context.err),
 			}
 		}
-		return nil, nil, nil, errors.New(context.err)
+		return nil, nil, nil, context.executionSteps, errors.New(context.err)
 	}
 
 	traces := make([]TransactionTrace, len(context.traces))
 	for index, traceJSON := range context.traces {
 		if err := json.Unmarshal(traceJSON, &traces[index]); err != nil {
-			return nil, nil, nil, fmt.Errorf("unmarshal trace: %v", err)
+			return nil, nil, nil, context.executionSteps, fmt.Errorf("unmarshal trace: %v", err)
 		}
 		//
 	}
 
-	return context.actualFees, context.dataGasConsumed, traces, nil
+	return context.actualFees, context.dataGasConsumed, traces, context.executionSteps, nil
 }
 
 func marshalTxnsAndDeclaredClasses(txns []core.Transaction, declaredClasses []core.Class) (json.RawMessage, json.RawMessage, error) { //nolint:lll
--- a/node/throttled_vm.go
+++ b/node/throttled_vm.go
@@ -33,13 +33,14 @@ func (tvm *ThrottledVM) Call(callInfo *vm.CallInfo, blockInfo *vm.BlockInfo, sta
 func (tvm *ThrottledVM) Execute(txns []core.Transaction, declaredClasses []core.Class, paidFeesOnL1 []*felt.Felt,
 	blockInfo *vm.BlockInfo, state core.StateReader, network *utils.Network, skipChargeFee, skipValidate, errOnRevert,
 	useBlobData bool,
-) ([]*felt.Felt, []*felt.Felt, []vm.TransactionTrace, error) {
+) ([]*felt.Felt, []*felt.Felt, []vm.TransactionTrace, uint64, error) {
 	var ret []*felt.Felt
 	var traces []vm.TransactionTrace
 	var dataGasConsumed []*felt.Felt
-	return ret, dataGasConsumed, traces, tvm.Do(func(vm *vm.VM) error {
+	var numSteps uint64
+	return ret, dataGasConsumed, traces, numSteps, tvm.Do(func(vm *vm.VM) error {
 		var err error
-		ret, dataGasConsumed, traces, err = (*vm).Execute(txns, declaredClasses, paidFeesOnL1, blockInfo, state, network,
+		ret, dataGasConsumed, traces, numSteps, err = (*vm).Execute(txns, declaredClasses, paidFeesOnL1, blockInfo, state, network,
 			skipChargeFee, skipValidate, errOnRevert, useBlobData)
 		return err
 	})
--- a/vm/vm_test.go
+++ b/vm/vm_test.go
@@ -203,7 +203,7 @@ func TestExecute(t *testing.T) {
 	state := core.NewState(txn)
 
 	t.Run("empty transaction list", func(t *testing.T) {
-		_, _, _, err := New(false, nil).Execute([]core.Transaction{}, []core.Class{}, []*felt.Felt{}, &BlockInfo{
+		_, _, _, _, err := New(false, nil).Execute([]core.Transaction{}, []core.Class{}, []*felt.Felt{}, &BlockInfo{
 			Header: &core.Header{
 				Timestamp:        1666877926,
 				SequencerAddress: utils.HexToFelt(t, "0x46a89ae102987331d369645031b49c27738ed096f2789c24449966da4c6de6b"),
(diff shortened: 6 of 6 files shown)


vm/vm.go
```diff
@@ -324,23 +331,23 @@ func (v *vm) Execute(txns []core.Transaction, declaredClasses []core.Class, paid
...
 		}
-		return nil, nil, nil, errors.New(context.err)
+		return nil, nil, nil, context.executionSteps, errors.New(context.err)
 	}
 
 	traces := make([]TransactionTrace, len(context.traces))
 	for index, traceJSON := range context.traces {
 		if err := json.Unmarshal(traceJSON, &traces[index]); err != nil {
-			return nil, nil, nil, fmt.Errorf("unmarshal trace: %v", err)
+			return nil, nil, nil, context.executionSteps, fmt.Errorf("unmarshal trace: %v", err)
```
kirugan, line 345: in case of errors we return zero values (i.e. `0`)

obasekiosa: even if the vm ran for some time?

obasekiosa: even in cases where the errors came as a result of the transactions
not the system?

kirugan: > even in cases where the errors came as a result of the transactions
not the system?

yes, in general we should use this approach

rpc/simulation.go
```diff
@@ -106,7 +106,7 @@ func (h *Handler) simulateTransactions(id BlockID, transactions []BroadcastedTra
 		BlockHashToBeRevealed: blockHashToBeRevealed,
 	}
 	useBlobData := !v0_6Response
-	overallFees, dataGasConsumed, traces, err := h.vm.Execute(txns, classes, paidFeesOnL1, &blockInfo,
+	overallFees, dataGasConsumed, traces, _, err := h.vm.Execute(txns, classes, paidFeesOnL1, &blockInfo,
```
kirugan, line 109: do you think we don't need numSteps here?

vm/vm.go
```diff
@@ -324,23 +331,23 @@ func (v *vm) Execute(txns []core.Transaction, declaredClasses []core.Class, paid
 
 	if context.err != "" {
 		if context.errTxnIndex >= 0 {
-			return nil, nil, nil, TransactionExecutionError{
+			return nil, nil, nil, context.executionSteps, TransactionExecutionError{
```
kirugan, line 334: that's also an error, I think we should get rid of `context.executionSteps` here

vm/vm.go
```diff
@@ -324,23 +331,23 @@ func (v *vm) Execute(txns []core.Transaction, declaredClasses []core.Class, paid
...
+		return nil, nil, nil, 0, errors.New(context.err)
 	}
 
 	traces := make([]TransactionTrace, len(context.traces))
 	for index, traceJSON := range context.traces {
 		if err := json.Unmarshal(traceJSON, &traces[index]); err != nil {
-			return nil, nil, nil, fmt.Errorf("unmarshal trace: %v", err)
+			return nil, nil, nil, 0, fmt.Errorf("unmarshal trace: %v", err)
 		}
 		//
```
kirugan, line 347: can you please remove this useless comment ?

rpc/simulation.go
obasekiosa: this was to ensure it compiles in CI

added it there without using the variable would cause linter to scream

vm/vm.go
```diff
@@ -138,6 +139,12 @@ func JunoAppendDataGasConsumed(readerHandle C.uintptr_t, ptr unsafe.Pointer) {
 	context.dataGasConsumed = append(context.dataGasConsumed, makeFeltFromPtr(ptr))
 }
 
+//export JunoAddExecutionSteps
+func JunoAddExecutionSteps(readerHandle C.uintptr_t, execSteps C.ulonglong) {
+	context := unwrapContext(readerHandle)
+	context.executionSteps += uint64(execSteps) // assumes initial default value is zero even in ffi
```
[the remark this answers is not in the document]
obasekiosa: so i added this earlier to make sure (I checked later) that the initial values on object creation is zero(0) and not some random data in memory, since it was being called over ffi.
I didn't want to assume the variable would be initialized to zero just because go does initialize its variables on creation.

Its irrelevant now

vm/vm.go
obasekiosa: I'm surprised this is here it's already removed in previous pushes,
might be a github thing.

vm/rust/src/lib.rs
```diff
@@ -348,6 +349,7 @@ pub extern "C" fn cairoVMExecute(
 
                 let actual_fee = t.transaction_receipt.fee.0.into();
                 let data_gas_consumed = t.transaction_receipt.da_gas.l1_data_gas.into();
+                let execution_steps = t.transaction_receipt.resources.vm_resources.n_steps.try_into().unwrap();
```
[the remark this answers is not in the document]
obasekiosa: The problem here is that resources.n_steps is of type usize

while the `C.ulonglong `is of type `u64`.

`u64` does not implement `From<usize>`

rpc/trace.go
```diff
@@ -150,72 +155,81 @@ func (h *Handler) traceTransaction(ctx context.Context, hash *felt.Felt, v0_6Res
...
-	return traceResults[txIndex].TraceRoot, nil
+	return traceResults[txIndex].TraceRoot, header, nil
 }
 
-func (h *Handler) TraceBlockTransactions(ctx context.Context, id BlockID) ([]TracedBlockTransaction, *jsonrpc.Error) {
+func (h *Handler) TraceBlockTransactions(ctx context.Context, id BlockID) ([]TracedBlockTransaction, http.Header, *jsonrpc.Error) {
 	block, rpcErr := h.blockByID(&id)
 	if rpcErr != nil {
-		return nil, rpcErr
+		execStepsHeader := http.Header{}
```
obasekiosa, line 187: all `execStepHeader` will be renamed  to `httpHeader`.

jsonrpc/http.go
```diff
@@ -46,8 +47,11 @@ func (h *HTTP) ServeHTTP(writer http.ResponseWriter, req *http.Request) {
 
 	req.Body = http.MaxBytesReader(writer, req.Body, MaxRequestBodySize)
 	h.listener.OnNewRequest("any")
-	resp, err := h.rpc.HandleReader(req.Context(), req.Body)
+	resp, header, err := h.rpc.HandleReader(req.Context(), req.Body)
+
 	writer.Header().Set("Content-Type", "application/json")
+	maps.Copy(writer.Header(), header) // overwrites duplicate headers
```
rianhughes, line 53: Do you think we could just do something like this instead?

writer.Header().Set(rpc.ExecutionStepsHeader, header.Get(rpc.ExecutionStepsHeader))

obasekiosa: we could but then we'd be heavily breaking boundaries, and not general enough.

now the http service knows about specific information concerning just a single handler.

the idea is these meta-data (headers and info returned and not part of json response but related to it - though for now just headers) can be returned by any handler.

jsonrpc/server.go
```diff
@@ -359,20 +377,33 @@ func (s *Server) handleBatchRequest(ctx context.Context, batchReq []json.RawMess
...
+
+	// merge headers
+	finalHeaders := http.Header{}
+	for _, header := range headers {
+		for k, v := range header {
+			for _, e := range v {
+				finalHeaders.Add(k, e)
+			}
+		}
+	}
```
rianhughes, line 397: We are only looking to add a single header, so I don't think we need to account for multiple headers, imo

jsonrpc/server.go
```diff
@@ -255,7 +264,8 @@ func (s *Server) HandleReadWriter(ctx context.Context, rw io.ReadWriter) error {
 		activated: activated,
 	}
 	msgCtx := context.WithValue(ctx, ConnKey{}, conn)
-	resp, err := s.HandleReader(msgCtx, rw)
+	// header is unnecessary for read-writer(websocket)
+	resp, _, err := s.HandleReader(msgCtx, rw)
```
rianhughes, line 268: The returned header from s.HandleReader is never used, so I don't think it should be returned

rianhughes, line 268: Does HandleReader need to return the header if it's never used?

rianhughes: Ah I see, h.rpc.HandleReader(req.Context(), req.Body) uses it. Nvm

jsonrpc/server.go
```diff
@@ -359,20 +377,33 @@ func (s *Server) handleBatchRequest(ctx context.Context, batchReq []json.RawMess
...
 	// according to the spec if there are no response objects server must not return empty array
 	if len(responses) == 0 {
-		return nil, nil
+		return nil, finalHeaders, nil
 	}
 
-	return json.Marshal(responses)
+	result, err := json.Marshal(responses)
+
+	return result, finalHeaders, err // todo: fix batch request aggregate header
```
rianhughes, line 406: Does this ToDo need fixed before merging into main?

jsonrpc/server.go
```diff
@@ -430,22 +463,28 @@ func (s *Server) handleRequest(ctx context.Context, req *Request) (*response, er
...
 			errJSON, _ := json.Marshal(res.Error)
 			s.log.Debugw("Failed handing RPC request", "req", string(reqJSON), "res", string(errJSON))
 		}
-		return res, nil
+		return res, header, nil
 	}
 	res.Result = tuple[0].Interface()
 
-	return res, nil
+	return res, header, nil
```
rianhughes, line 487: Do we not have to set header = tuple[1].Interface() if the len(tuple)==3 here?

jsonrpc/server.go
rianhughes, line 397: Just making a note for myself to come back to this, I don't fully understand what's happening here

obasekiosa: so whats happening here is 
the headers returned from all the calls to the different handlers (batch calls) are being merged into one.

result would look like this 
```go

// header returned by first call
{
"Header 1":  "value from call 1",
"Header 3":  "value from call 1"
}

// header returned by second call 
{
"Header 2":  "value from call 2",
"Header 3":  "value from call 2"
}


// final header on combined result
{
 "Header 1": "value from call 1",
 "Header 2": "value from call 1, value from call 2",
 "Header 3": "value from call 2"
}
```


its a work around seeing that to support single return headers for batch requests as it is currently implemented **may** require a complete refactor to remove its dependency on single request calls.

rpc/simulation_test.go
```diff
@@ -39,8 +39,10 @@ func TestSimulateTransactions(t *testing.T) {
 		}, mockState, n, true, false, false, false).
 			Return([]*felt.Felt{}, []vm.TransactionTrace{}, nil)
 
-		_, err := handler.SimulateTransactions(rpc.BlockID{Latest: true}, []rpc.BroadcastedTransaction{}, []rpc.SimulationFlag{rpc.SkipFeeChargeFlag})
+		_, httpHeader, err := handler.SimulateTransactions(rpc.BlockID{Latest: true}, []rpc.BroadcastedTransaction{}, []rpc.SimulationFlag{rpc.SkipFeeChargeFlag})
 		require.Nil(t, err)
+		require.NotNil(t, httpHeader)
+		require.NotEmpty(t, httpHeader.Get(rpc.ExecutionStepsHeader))
```
rianhughes, line 45: Do you think we could remove t.Skip() above and re-introduce these tests? Eg:

t.Run("ok with zero values, skip fee", func(t *testing.T) {
		stepsUsed := uint64(123)
		mockVM.EXPECT().Execute([]core.Transaction{}, nil, []*felt.Felt{}, &vm.BlockInfo{
			Header: headsHeader,
		}, mockState, n, true, false, false, true).
			Return([]*felt.Felt{}, []*felt.Felt{}, []vm.TransactionTrace{}, stepsUsed, nil)

		_, httpHeader, err := handler.SimulateTransactions(rpc.BlockID{Latest: true}, []rpc.BroadcastedTransaction{}, []rpc.SimulationFlag{rpc.SkipFeeChargeFlag})
		require.Nil(t, err)
		require.NotNil(t, httpHeader)
		require.Equal(t, fmt.Sprint(stepsUsed), httpHeader.Get(rpc.ExecutionStepsHeader))
	})

rpc/trace_test.go
```diff
@@ -157,11 +163,15 @@ func TestTraceTransaction(t *testing.T) {
...
 		mockVM.EXPECT().Execute([]core.Transaction{tx}, []core.Class{declaredClass.Class}, []*felt.Felt{},
-			&vm.BlockInfo{Header: header}, gomock.Any(), &utils.Mainnet, false, false, false, true).Return(overallFee, consumedGas, []vm.TransactionTrace{*vmTrace}, nil)
+			&vm.BlockInfo{Header: header}, gomock.Any(), &utils.Mainnet, false, false,
+			false, true).Return(overallFee, consumedGas, []vm.TransactionTrace{*vmTrace}, numSteps, nil)
 
-		trace, err := handler.TraceTransaction(context.Background(), *hash)
+		trace, httpHeader, err := handler.TraceTransaction(context.Background(), *hash)
 		require.Nil(t, err)
+		require.NotNil(t, httpHeader)
+		require.NotEmpty(t, httpHeader.Get(rpc.ExecutionStepsHeader))
```
rianhughes, line 174: nit: we can check it's 0 right?

obasekiosa: yes, though the idea was, it wasn't being tested for or expected.
So as log as it returned a value that was enough for a couple of the tests.

updated now though.

Also most of the tests with t.Skip where apparently previously done purposefully.

jsonrpc/server.go
obasekiosa: no it doesn't its there to say the way handle batch request is now isn't optimal, better to refactor later.

obasekiosa:
@rianhughes 

what do we do about this? Refactor?

```sh
Error: rpc/estimate_fee.go:113:38: hugeParam: msg is heavy (112 bytes); consider passing it by pointer (gocritic)
func (h *Handler) estimateMessageFee(msg MsgFromL1, id BlockID, f estimateFeeHandler) (*FeeEstimate,
                                     ^
jsonrpc/server.go:216: File is not `gofumpt`-ed (gofumpt)
	//todo: guard for this in the code! don't depend on devs finding out about this!
rpc/estimate_fee.go:114: File is not `gofumpt`-ed (gofumpt)
	http.Header, *jsonrpc.Error) { //nolint:gocritic
rpc/trace.go:143: File is not `gofumpt`-ed (gofumpt)
	http.Header, *jsonrpc.Error) {
make: *** [Makefile:89: lint] Error 1
Error: Process completed with exit code 2.
```

rpc/trace_test.go
obasekiosa: this has been fixed in main and merged into this branch

obasekiosa:
> @rianhughes
> 
> what do we do about this? Refactor?
> 
> ```shell
> Error: rpc/estimate_fee.go:113:38: hugeParam: msg is heavy (112 bytes); consider passing it by pointer (gocritic)
> func (h *Handler) estimateMessageFee(msg MsgFromL1, id BlockID, f estimateFeeHandler) (*FeeEstimate,
>                                      ^
> jsonrpc/server.go:216: File is not `gofumpt`-ed (gofumpt)
> 	//todo: guard for this in the code! don't depend on devs finding out about this!
> rpc/estimate_fee.go:114: File is not `gofumpt`-ed (gofumpt)
> 	http.Header, *jsonrpc.Error) { //nolint:gocritic
> rpc/trace.go:143: File is not `gofumpt`-ed (gofumpt)
> 	http.Header, *jsonrpc.Error) {
> make: *** [Makefile:89: lint] Error 1
> Error: Process completed with exit code 2.
> ```

fixed!

rpc/simulation_test.go
```diff
@@ -34,55 +34,65 @@ func TestSimulateTransactionsV0_6(t *testing.T) {
...
 			Header: headsHeader,
 		}, mockState, n, true, false, false, false).
-			Return([]*felt.Felt{}, []*felt.Felt{}, []vm.TransactionTrace{}, nil)
+			Return([]*felt.Felt{}, []*felt.Felt{}, []vm.TransactionTrace{}, stepsUsed, nil)
 
-		_, err := handler.SimulateTransactionsV0_6(rpc.BlockID{Latest: true}, []rpc.BroadcastedTransaction{}, []rpc.SimulationFlag{rpc.SkipFeeChargeFlag})
+		_, httpHeader, err := handler.SimulateTransactionsV0_6(rpc.BlockID{Latest: true}, []rpc.BroadcastedTransaction{}, []rpc.SimulationFlag{rpc.SkipFeeChargeFlag})
 		require.Nil(t, err)
+		require.NotNil(t, httpHeader)
+		require.Equal(t, httpHeader.Get(rpc.ExecutionStepsHeader), "123")
```
rianhughes, line 46: nit: we should use stepsUsed here (similar for other tests below)

rpc/trace_test.go
```diff
@@ -246,10 +256,12 @@ func TestTraceTransaction(t *testing.T) {
...
 		mockVM.EXPECT().Execute([]core.Transaction{tx}, []core.Class{declaredClass.Class}, []*felt.Felt{},
 			&vm.BlockInfo{Header: header}, gomock.Any(), &utils.Mainnet, false, false, false, true).
-			Return(overallFee, consumedGas, []vm.TransactionTrace{*vmTrace}, nil)
+			Return(overallFee, consumedGas, []vm.TransactionTrace{*vmTrace}, uint64(123), nil)
 
-		trace, err := handler.TraceTransaction(context.Background(), *hash)
+		trace, httpHeader, err := handler.TraceTransaction(context.Background(), *hash)
 		require.Nil(t, err)
+		require.NotNil(t, httpHeader)
+		require.Equal(t, httpHeader.Get(rpc.ExecutionStepsHeader), "123")
```
rianhughes, line 264: nit: I think it's better to use a variable to hold 123 etc

rianhughes (approved):

obasekiosa:
anything more needed here @kirugan

vm/rust/src/lib.rs
```diff
@@ -350,6 +351,13 @@ pub extern "C" fn cairoVMExecute(
 
                 let actual_fee = t.transaction_receipt.fee.0.into();
                 let data_gas_consumed = t.transaction_receipt.da_gas.l1_data_gas.into();
+                let execution_steps = t
+                    .transaction_receipt
+                    .resources
+                    .vm_resources
+                    .n_steps
+                    .try_into()
+                    .unwrap_or(u64::MAX);
```
kirugan, line 360: why it's u64::MAX if conversion is failed?

rpc/trace_test.go
```diff
@@ -382,15 +398,19 @@ func TestTraceTransactionV0_6(t *testing.T) {
...
 		assert.Equal(t, vmTrace, trace)
 	})
 }
 
-func TestTraceBlockTransactionsV0_6(t *testing.T) {
+func TestTraceBlockTransactions(t *testing.T) {
+	t.Skip() // todo: fix calls later
```
kirugan, line 412: unskip?

rpc/trace_test.go
```diff
@@ -382,15 +398,19 @@ func TestTraceTransactionV0_6(t *testing.T) {
...
+		trace, httpHeader, err := handler.TraceTransactionV0_6(context.Background(), *hash)
 		require.Nil(t, err)
+		require.NotNil(t, httpHeader)
+		require.Equal(t, httpHeader.Get(rpc.ExecutionStepsHeader), "123")
 		assert.Equal(t, vmTrace, trace)
 	})
 }
 
-func TestTraceBlockTransactionsV0_6(t *testing.T) {
+func TestTraceBlockTransactions(t *testing.T) {
```
kirugan, line 411: why changing name of this test? and calling function below?

jsonrpc/server.go
obasekiosa: yea!

not so related Side note: an additional question the question is when does it make sense to aggregate the  different headers values in to one value (like total execution steps)

vm/rust/src/lib.rs
obasekiosa: the only time a conversion actually fails is when coming from a higher range to a lower range.
it fails in the case of overflow.

rpc/trace_test.go
obasekiosa: came as a result of a merge from main.

vm/rust/src/lib.rs
kirugan: makes sense, thanks!

jsonrpc/server.go
```diff
@@ -205,6 +213,7 @@ type connection struct {
 	w         io.Writer
 	activated <-chan struct{}
 
+	// todo: guard for this in the code! don't depend on devs finding out about this!
```
kirugan, line 216: I don't think we should guard this because `connection` created upon websocket connection to one client, also `initialErr` is set during initial setup

obasekiosa: all update this now

obasekiosa: before i make the push.
I the idea was to make the error thread safe by wrapping in a mutex guarded error object.
that gets set after the inner lock has been acquired

```go
type ThreadSafeError struct {
 error
 lock
}

func (t *ThreadSafeError) SetError() {
 acquire lock
 set error
 release lock
}
```

kirugan (approved):

obasekiosa:
lgtm 👍🏾 

PS: oh switching to asserts so the other tests still run?

jsonrpc/server.go
kirugan: I'm not sure that if we needed we would need to do this anyway. Because error is an interface which in turn is just a pointer to a struct, every update of a pointer is atomic even if you don't guard it with mutexes

obasekiosa: oh cool!, I didn't know this!
do you have any go doc that describes this more.

PR #2005, opened 2024-07-31
Search test data directory path dynamically

weiihann:
### Description
This PR modifies the implementation of finding the target test data directory. The issue with the existing implementation is that if a user changes the repository's name from juno to whatever, the tests will fail. Also, any tests that rely on NewTestClient in `cmd/juno` will fail because the base directory extracted will be `../juno/cmd/juno` instead of `../juno`.

### DIFF
--- a/clients/feeder/feeder.go
+++ b/clients/feeder/feeder.go
@@ -128,10 +128,8 @@ func newTestServer(t *testing.T, network *utils.Network) *httptest.Server {
 		assert.Equal(t, []string{"API_KEY"}, r.Header["X-Throttling-Bypass"])
 		assert.Equal(t, []string{"Juno/v0.0.1-test Starknet Implementation"}, r.Header["User-Agent"])
 
-		wd, err := os.Getwd()
 		require.NoError(t, err)
 
-		base := wd[:strings.LastIndex(wd, "juno")+4]
 		queryArg := ""
 		dir := ""
 		const blockNumberArg = "blockNumber"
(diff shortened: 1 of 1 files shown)


kirugan (approved):

clients/feeder/feeder.go
```diff
@@ -447,3 +449,20 @@ func (c *Client) BlockTrace(ctx context.Context, blockHash string) (*starknet.Bl
...
+	for root := wd; ; root = filepath.Dir(root) {
+		if _, err := os.Stat(filepath.Join(root, targetRelPath)); err == nil {
+			return filepath.Join(root, targetRelPath), nil
+		} else if !os.IsNotExist(err) {
+			return "", err
+		} else if newRoot := filepath.Dir(root); newRoot == root {
+			return "", os.ErrNotExist
+		}
+	}
+}
```
IronGauntlets, line 468: I think the function can be simplified as follows:

```suggestion
func findTargetDirectory(targetRelPath string) (string, error) {
	root, err := os.Getwd()
	if err != nil {
		return "", err
	}

	for {
		targetPath := filepath.Join(root, targetRelPath)
		if _, err := os.Stat(targetPath); err == nil {
			return targetPath, nil
		} else if !os.IsNotExist(err) {
			return "", err
		}

		newRoot := filepath.Dir(root)
		if newRoot == root {
			return "", os.ErrNotExist
		}
		root = newRoot
	}
}
```

PR #2011, opened 2024-08-01
Implement p2p over public ip

derrix060:


wojciechos (approved):

.gitignore
```diff
@@ -8,4 +8,4 @@ config/
 .envrc
 # Default path for Juno DB files.  It will get created if you follow the
 # README and/or run `./build/juno` command.
-juno/
+.juno/
```
kirugan, line 11: why do we need this change?

p2p/p2p.go
```diff
@@ -66,12 +66,46 @@ func New(addr, version, peers, privKeyStr string, feederNode bool, bc *blockchai
 		return nil, err
 	}
 
+	var publicMultiAddr multiaddr.Multiaddr = nil
```
kirugan, line 69: you don't need to assign nil explicitly because `multiaddr.Multiaddr` is interface and zero value is nil

.gitignore
derrix060: https://github.com/NethermindEth/juno/pull/2011/commits/4aad791c01e513476c80363567054f06f7c3e3b7

You can see the commit message, basically without this the files in `cmd/juno` would be ignored

p2p/p2p.go
derrix060: oh, so just become something like this?

```suggestion
	var publicMultiAddr multiaddr.Multiaddr
```

kirugan: yep, exactly

.gitignore
kirugan: it shouldn't be the case. For instance here is a PR that change files in the directory - https://github.com/NethermindEth/juno/pull/1902/files

Btw If I specify `.dir` in .gitignore does it mean that it will ignore hidden directory or `./dir/` basically?

derrix060: ha, maybe I missed something here, let me revert this

kirugan (approved):

PR #2012, opened 2024-08-01
Start using gitlab hosted runners for ARM

derrix060:
Those runners have proper auto-scaling and are able to out-perform the self-hosted one that we had.

### DIFF
--- a/.github/workflows/build-binaries.yml
+++ b/.github/workflows/build-binaries.yml
@@ -16,7 +16,7 @@ jobs:
             architecture: amd64
           - os: macOS-latest
             architecture: amd64
-          - os: self-hosted
+          - os: ubuntu-arm64-4-core
             architecture: arm64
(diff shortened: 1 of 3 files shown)


kirugan (approved):

kirugan (approved):

wojciechos (approved):

wojciechos (changes requested):
We should also update the build [docker workflow](https://github.com/NethermindEth/juno/blob/main/.github/workflows/docker-image-build-push.yml#L47). I'm also wondering how long it would take to build the image with this machine. We used an ARM machine to build the ARM image because building it on an AMD machine takes ages

derrix060:
well spoted, let me update

derrix060:
I've changed to use the 4core machine for the build, it should improve the compilation time

wojciechos (approved):

PR #2030, opened 2024-08-06
Daniil/rust-linter-issues-fix

AnkushinDaniil:


starknet/rust/src/lib.rs
```diff
@@ -1,34 +1,36 @@
...
+    let sierra_class = serde_json::from_str(sierra_json_str).map_err(|err| err.to_string());
     if let Err(e) = sierra_class {
-        return raw_cstr(e)
+        return raw_cstr(e);
     }
 
-    let casm_class = CasmContractClass::from_contract_class(sierra_class.unwrap(), true, usize::MAX)
-        .map_err(|err| err.to_string());
+    let casm_class =
+        CasmContractClass::from_contract_class(sierra_class.unwrap(), true, usize::MAX)
```
kirugan, line 15: we should rewrite `if let` above to match stmt. so we won't call `sierra_class.unwrap()` here
hint: you can shadow existing variable with the same name in Rust:
```
let value = 111; // int
let value = value > 10; // bool
```

starknet/rust/src/lib.rs
```diff
@@ -1,34 +1,36 @@
...
+            .map_err(|err| err.to_string());
     if let Err(e) = casm_class {
-        return raw_cstr(e)
+        return raw_cstr(e);
     }
 
     let casm_json = serde_json::to_string(&casm_class.unwrap());
```
kirugan, line 21: same thing with `casm_class` and `casm_json`

kirugan:
Please run `make format` after you will make changes

vm/rust/src/juno_state_reader.rs
```diff
@@ -189,8 +189,6 @@ impl UpdatableState for JunoStateReader {
 
 pub fn felt_to_byte_array(felt: &StarkFelt) -> [u8; 32] {
     felt.to_bytes_be()
-        .try_into()
-        .expect("StarkFelt not [u8; 32]")
```
kirugan, line 193: probably it was a mistake in the first place

kirugan (approved):

starknet/rust/src/lib.rs
```diff
@@ -1,34 +1,35 @@
-use std::ffi::{c_char, CStr, CString};
 use cairo_lang_starknet_classes::casm_contract_class::CasmContractClass;
+use std::ffi::{c_char, CStr, CString};
 
 #[no_mangle]
+#[allow(clippy::not_unsafe_ptr_arg_deref)]
 pub extern "C" fn compileSierraToCasm(sierra_json: *const c_char) -> *mut c_char {
     let sierra_json_str = unsafe { CStr::from_ptr(sierra_json) }.to_str().unwrap();
```
obasekiosa, line 7: might need an unwrap_or here.
Possible error handling.

starknet/rust/src/lib.rs
```diff
@@ -1,34 +1,35 @@
...
+        Ok(value) => value,
+        Err(e) => return raw_cstr(e.to_string()),
+    };
 
-    let casm_json = serde_json::to_string(&casm_class.unwrap());
-    if let Err(e) = casm_json {
-        return raw_cstr(e.to_string())
-    }
-    return raw_cstr(casm_json.unwrap())
+    raw_cstr(casm_json)
```
obasekiosa, line 24: Do we really want to return the error msg and the result as indistiguishable return values?

how does the calling function tell the difference. It seems a bit unusal. We could pass in a value to receive the result and return an error if any. 

What do you think @kirugan @AnkushinDaniil 

heres the calling code.
```go
func Compile(sierra *SierraDefinition) (*CompiledClass, error) {
	sierraJSON, err := json.Marshal(SierraDefinition{
		EntryPoints: sierra.EntryPoints,
		Program:     sierra.Program,
		Version:     sierra.Version,
	})
	if err != nil {
		return nil, err
	}

	sierraJSONCstr := C.CString(string(sierraJSON))
	defer func() {
		C.free(unsafe.Pointer(sierraJSONCstr))
	}()

	casmJSONOrErrorCstr := C.compileSierraToCasm(sierraJSONCstr)
	casmJSONOrError := C.GoString(casmJSONOrErrorCstr)
	C.freeCstr(casmJSONOrErrorCstr)

	var casmClass CompiledClass
	if err = json.Unmarshal([]byte(casmJSONOrError), &casmClass); err != nil {
		var syntaxErr *json.SyntaxError
		if errors.As(err, &syntaxErr) {
			return nil, errors.New(casmJSONOrError)
		}
		return nil, err
	}

	return &casmClass, nil
}
```

vm/rust/src/lib.rs
```diff
@@ -528,16 +530,16 @@ lazy_static! {
 #[allow(static_mut_refs)]
 fn get_versioned_constants(version: *const c_char) -> VersionedConstants {
     let version_str = unsafe { CStr::from_ptr(version) }.to_str().unwrap();
-    let version = StarknetVersion::from_str(&version_str)
-        .unwrap_or(StarknetVersion::from_str(&"0.0.0").unwrap());
+    let version = StarknetVersion::from_str(version_str)
+        .unwrap_or(StarknetVersion::from_str("0.0.0").unwrap());
```
obasekiosa, line 534: why do we have unwraps inside an unwrap_or

Can we have matches across the version strings that go straight to the StarknetVersion with a default branch for that returns the value for starknetversion value for "0.0.0"?

_currently a single missed typo can cause runtime errors_

vm/rust/src/lib.rs
```diff
@@ -528,16 +530,16 @@ lazy_static! {
...
 
     if let Some(constants) = unsafe { &CUSTOM_VERSIONED_CONSTANTS } {
         constants.clone()
-    } else if version < StarknetVersion::from_str(&"0.13.1").unwrap() {
+    } else if version < StarknetVersion::from_str("0.13.1").unwrap() {
```
obasekiosa, line 538: same here i think a couple of constant fixed values could do. 

or having an UNKNOWN version type 


here is the implementation code block. 
```rust
pub struct StarknetVersion(u8, u8, u8, u8);

...

impl FromStr for StarknetVersion {
    type Err = anyhow::Error;

    fn from_str(s: &str) -> Result<Self, Self::Err> {
        if s.is_empty() {
            return Ok(StarknetVersion::new(0, 0, 0, 0));
        }

        let parts: Vec<_> = s.split('.').collect();
        anyhow::ensure!(
            parts.len() == 3 || parts.len() == 4,
            "Invalid version string, expected 3 or 4 parts but got {}",
            parts.len()
        );

        let a = parts[0].parse()?;
        let b = parts[1].parse()?;
        let c = parts[2].parse()?;
        let d = parts.get(3).map(|x| x.parse()).transpose()?.unwrap_or(0);

        Ok(StarknetVersion(a, b, c, d))
    }
}
```

obasekiosa (commented):
It looks good, just a few questions and suggestions to sort out, then we can merge.

vm/rust/src/lib.rs
AnkushinDaniil: I don't think any changes are needed here.
@kirugan What do you think?

core/rust/src/lib.rs
```diff
@@ -5,6 +5,7 @@ use std::{
 };
 
 #[no_mangle]
+#[allow(clippy::not_unsafe_ptr_arg_deref)]
 pub extern "C" fn Cairo0ClassHash(class_json_str: *const c_char, hash: *mut c_uchar) {
     let class_json = unsafe { CStr::from_ptr(class_json_str) }.to_str().unwrap();
```
obasekiosa, line 10: same thing here... unwrap without handling the possibility of error.

obasekiosa (approved):
outside some personal opinionated requests this lgtm

vm/rust/src/lib.rs
kirugan: Agree, let's keep it as is

starknet/rust/src/lib.rs
```diff
@@ -1,34 +1,61 @@
-use std::ffi::{c_char, CStr, CString};
 use cairo_lang_starknet_classes::casm_contract_class::CasmContractClass;
+use std::ffi::{c_char, CStr, CString};
 
 #[no_mangle]
-pub extern "C" fn compileSierraToCasm(sierra_json: *const c_char) -> *mut c_char {
-    let sierra_json_str = unsafe { CStr::from_ptr(sierra_json) }.to_str().unwrap();
+#[allow(clippy::not_unsafe_ptr_arg_deref)]
+pub extern "C" fn compileSierraToCasm(sierra_json: *const c_char, result: *mut *mut c_char) -> u8 {
```
kirugan, line 6: why `result` type is *mut *mut ?

derrix060 (dismissed):
Please rebase your branch and only pick the commits you actually authored :)

There is also conflicts.

BTW it's easier to rebase your branch then merging main into yours when trying to keep up-to-date

starknet/rust/src/lib.rs
AnkushinDaniil: Rust doesn't allow you to directly return a dynamically allocated string because Rust's memory management system would deallocate the string when the function returns.

PR #2034, opened 2024-08-08
feat: persist native compiled contracts

xrvdg:
Adds persistence for native compiled contracts. The directory where these are stored can be set via the environment variable `JUNO_NATIVE_CACHE_DIR`, otherwise the current active directory is used. 

The cache is invalidated by changing the revision of cairo native. The logic for getting the cairo native revision is in `build.rs`. 

Merging into the native2.6.4-blockifier branch as the 2.7.x branch is not ready yet.

### DIFF
--- a/vm/rust/src/juno_state_reader.rs
+++ b/vm/rust/src/juno_state_reader.rs
@@ -1,6 +1,6 @@
 use std::{
     ffi::{c_char, c_uchar, c_void, CStr},
-    slice,
+    fs, slice,
     sync::Mutex,
 };
 use blockifier::execution::contract_class::{ContractClass, NativeContractClassV1};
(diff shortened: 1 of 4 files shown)


vm/rust/src/juno_state_reader.rs
```diff
@@ -251,7 +272,106 @@ fn native_try_from_json_string(
...
+
+/// Load a contract that is already compiled.
+///
+/// Returns None if the contract does not exist at the output_path.
+///
+/// To compile and load a contract use [persist_from_native_module] instead.
+fn load_compiled_contract(
+    sierra_program: &Program,
+    library_output_path: &PathBuf,
+) -> Option<Result<AotNativeExecutor, Box<dyn std::error::Error>>> {
```
rodrigo-pino, line 295: Just general feedback in case you didn't know, there is a library for returning dynamic errors: `anyhow` not important for this PR just sharing in case you didn't know

rodrigo-pino (approved):

Issue #1828, opened 2024-04-16
Implement SnapSync for Juno


stranger80:
Tasks:

- [x] Leaf proof and verification
  - ...including the iterator (Rian) 
- [x] Leaf range (Pawel & Rian) https://github.com/NethermindEth/juno/issues/1871
- [x] Refactor the original PoC code to align with the new range query interface (Ashraf)
- [x] Offline copy state via range (Ashraf)
  - Build the trie based on range query responses
- [x] (somewhere at this point) ability to prove any state (or `n`-most recent states)

- [x] 1. **[4 days + 2 days?](Actual: 12 days) ** Integrate with p2p (Ashraf & Pawel) 
  - handle p2p messages, refactor snap_server - **2 days** 
  - add startup flags & start snap syncher at a node start  - **1 day**
  - prepare synced "sepolia" node and make fresh snap sync - to observe p2p is working - **1 day**
  - fix storage ranges issues - **2 days?**
  
- [x] 2 **[2 days + 2 days?]/Actual: 8 days)** Perform static sync on cloud VM that contains
  - feeder node with sepolia snapshot and syncing disabled 
  - fresh node syncing from above

- [x] 3. **[1 day]/(Actual: 2 days)** Implement a **hidden** CLI command flag in Juno to explicitly launch it in SnapSync mode
(in review https://github.com/NethermindEth/juno/pull/2186)
  - The flag has no effect on the p2p handlers launched - Juno will launch all p2p handlers (including SnapSync specific ones) in any case
  - The flag will determine behaviour of the node when launching (the scenario of node launching in p2p mode is explained here: (link required) )
  - Merge SnapSync codebase into ``main``

Status on 7th Oct:
- All above have been completed and yielded additional improvement work (as mentioned in point 4. below)
- Final PR for task 2. requires merging (#2120)
- Elapsed actual time spent was 22 days, so the accuracy of original estimations was 11 days est  vs 22 days act. 
- We extrapolate this to all remaining estimations.

- [ ] 4. **[2 days]** Implement robustness improvements in P2P handlers
  - Blanket retry mechanism on P2P layer
  - ...anything else?

- [ ] 5. **[]**  Snapshot Layer
  - Design doc presented by @weiihann (https://www.notion.so/nethermind/Snapshot-Layer-Design-Doc-112360fc38d08045b412c2fd8bdedec3?pvs=4)
  - Need estimation - but seems like a big thing.
  
[...]

PR #2035, opened 2024-08-08
Basic snapsync algorithm

pnowosie:
This is [p2p/snap2](https://github.com/NethermindEth/juno/pull/new/pnowosie/basic-snapsync-algorithm) branch rebased onto current main.

During a work on p2p in [snap-sync-p2p branch](https://github.com/NethermindEth/juno/tree/pnowosie/snap-sync-p2p) I've refactored this code in a better way - I will rebase it here and then we can review.

🚧  Needs to be cleaned & aligned to code standards
🚧  Need more tests to comply with codecov


pnowosie:
As agreed in #1828 - snapsync PR will include p2p

PR #2042, opened 2024-08-09
Set macOS deployment target in Makefile

derrix060:
The idea is to avoid warnings like:
```
ld: warning: object file XX was built for newer macOS
version (14.4) than being linked (14.0)
```


kirugan:
@derrix060 good job!

kirugan (approved):

AnkushinDaniil (approved):

PR #2043, opened 2024-08-11
Add docs for subcommands

### LINKED ISSUE #2041: Docs: add information about db command
Recently we merged #1902 and it's worth mentioning `db` command in our [docs](https://juno.nethermind.io/) and how users can use it.

In order to update the website you first need to make changes in `docs/` directory

weiihann:
Closes #2041


derrix060 (approved):
LGTM.  I'd add a reference to this documentation in Readme.md though

LordGhostX (approved):
LGTM!

- Restructured the subcommands section a bit
- Released a new version (0.12.0), so the changes are live once this is merged

I also agree with the suggestion of adding the subcommands to the README

kirugan:
@LordGhostX but in 0.12.0 juno version we didn't add this command, it's only in main. Do you think it might lead to confussion?

LordGhostX:
@kirugan that's a good point. I thought the feature was live and was released in the last version (my bad)

I'll make a follow-up PR to remove the 0.12.0 docs release, then we do it another time

PR #2048, opened 2024-08-12
Some CI improvements

derrix060:


wojciechos (approved):

rianhughes (approved):

.github/workflows/juno-lint.yml
```diff
@@ -22,7 +22,5 @@ jobs:
       - name: Install Jemalloc (Ubuntu or self-hosted)
         run: sudo apt-get update -qq && sudo apt-get install -y libjemalloc-dev libjemalloc2 -y
       - name: golangci-lint
-        uses: golangci/golangci-lint-action@v6
-        with:
-          # Optional: version of golangci-lint to use in form of v1.2 or v1.2.3 or `latest` to use the latest version
-          version: v1.59.1
+        run: |
+          make lint
```
kirugan, line 26: is `golangci-lint-action` is the same as `make lint` ?

derrix060: For now yes, but if we ever change what make lint does, we don't have to worry about it.  An example of change is to bump the version of golangci-lint

PR #2051, opened 2024-08-14
Rianhughes/plugin2

rianhughes:
This branch should be merged into main, so that Juno can support plugins that follow the defined interface


sync/sync.go
```diff
@@ -72,6 +73,7 @@ type Synchronizer struct {
 
 	pendingPollInterval time.Duration
 	catchUpMode         bool
+	plugin              *junoplugin.JunoPlugin
```
kirugan, line 76: interface is pointer itself, we don't need `*` here

blockchain/blockchain.go
```diff
@@ -828,6 +828,26 @@ func (b *Blockchain) RevertHead() error {
...
+		if err != nil {
+			return err
+		}
+		stateUpdate, err := stateUpdateByNumber(txn, blockNumber)
+		if err != nil {
+			return err
+		}
+		state := core.NewState(txn)
+		reverseStateDiff, err = state.BuildReverseDiff(blockNumber, stateUpdate.StateDiff)
+		if err != nil {
```
derrix060, line 844: nit:

if err is not nil it will return error, otherwise it will return nil, then you could change to

```suggestion
		return err
```

weiihann:
I'd suggest to add some documentations so that users are aware of this feature

weiihann (approved):

cmd/juno/juno.go
```diff
@@ -170,6 +172,7 @@ const (
 		"The upper limit is 4 million steps, and any higher value will still be capped at 4 million."
 	corsEnableUsage             = "Enable CORS on RPC endpoints"
 	versionedConstantsFileUsage = "Use custom versioned constants from provided file"
+	pluginPathUsage             = "Path to the plugins .so file"
```
kirugan, line 175: do you think `Path to the plugin .so file` would be more correct? I would thought that it's possible to put several plugins in the same .so file

plugin/plugin.go
```diff
@@ -0,0 +1,43 @@
...
+	"plugin"
+
+	"github.com/NethermindEth/juno/core"
+	"github.com/NethermindEth/juno/core/felt"
+)
+
+//go:generate mockgen -destination=../mocks/mock_plugin.go -package=mocks github.com/NethermindEth/juno/plugin JunoPlugin
+type JunoPlugin interface {
+	Init() error
+	Shutdown() error // Todo: Currently this function will never be called.
```
kirugan, line 14: why we don't call it ?

plugin/plugin_test.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/NethermindEth/juno/mocks"
+	junoplugin "github.com/NethermindEth/juno/plugin"
+	adaptfeeder "github.com/NethermindEth/juno/starknetdata/feeder"
+	"github.com/NethermindEth/juno/sync"
+	"github.com/NethermindEth/juno/utils"
+	"github.com/stretchr/testify/require"
+	"go.uber.org/mock/gomock"
+)
+
+const timeout = time.Second
```
kirugan, line 20: seems like it's local constant can we put it inside `TestPlugin(...)` ?

plugin/plugin_test.go
```diff
@@ -0,0 +1,80 @@
...
+		plugin.EXPECT().NewBlock(block, su, gomock.Any())
+	}
+	bc := blockchain.New(testDB, &utils.Integration)
+	synchronizer := sync.New(bc, integGw, utils.NewNopZapLogger(), 0, false).WithPlugin(plugin)
+
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	require.NoError(t, synchronizer.Run(ctx))
+	cancel()
+
+	t.Run("resync to mainnet with the same db", func(t *testing.T) {
```
kirugan, line 49: Can you please elaborate on this? I'm don't understand why we need it

sync/sync.go
```diff
@@ -180,6 +188,47 @@ func (s *Synchronizer) fetchUnknownClasses(ctx context.Context, stateUpdate *cor
...
+		return
+	}
+	fromSU, err := s.blockchain.StateUpdateByNumber(fromBlock.Number)
+	if err != nil {
+		s.log.Warnw("Failed to retrieve the reverted blockchain head state-update for the plugin", "err", err)
+		return
+	}
+	reverseStateDiff, err := s.blockchain.GetReverseStateDiff()
+	if err != nil {
+		s.log.Warnw("Failed to retrieve reverse state diff", "head", fromBlock.Number, "hash", fromBlock.Hash.ShortString(), "err", err)
```
kirugan, line 207: why there is no return stmt here?

sync/sync.go
```diff
@@ -205,6 +254,9 @@ func (s *Synchronizer) verifierTask(ctx context.Context, block *core.Block, stat
 					// revert the head and restart the sync process, hoping that the reorg is not deep
 					// if the reorg is deeper, we will end up here again and again until we fully revert reorged
 					// blocks
+					if s.plugin != nil {
```
kirugan, line 257: you don't need this check because it's already inside `s.handlePluginRevertBlock()`

vm/rust/src/juno_state_reader.rs
```diff
@@ -111,7 +117,8 @@ impl StateReader for JunoStateReader {
     fn get_class_hash_at(&self, contract_address: ContractAddress) -> StateResult<ClassHash> {
         let addr = felt_to_byte_array(contract_address.0.key());
         let mut buffer: [u8; 32] = [0; 32];
-        let wrote = unsafe { JunoStateGetClassHashAt(self.handle, addr.as_ptr(), buffer.as_mut_ptr()) };
+        let wrote =
+            unsafe { JunoStateGetClassHashAt(self.handle, addr.as_ptr(), buffer.as_mut_ptr()) };
```
kirugan, line 121: is this result of `make format` ?

cmd/juno/juno.go
rianhughes: I guess you can have multiple plugins (/go types that satisfy the JunoPlugin interface), but Juno is only going to load one of them. Specifically the one whose variable name is "JunoPluginInstance"

rianhughes: IIRC, this approach simplified loading and using the plugin, but if you think it's necessary to load multiple plugins, then I should update the load function

plugin/plugin.go
rianhughes: I've updated the PR to handle plugin.Shutdown()

plugin/plugin_test.go
rianhughes: Sure, I just wanted to add a test where we sync a couple of blocks, then revert some, then resume syncing again. I seen that this was implemented in one of the reorg tests so just re-used some test code from there. The specific network we are using doesn't matter, just that the pluging functions are called as expected.

sync/sync.go
rianhughes: Yup, thanks

sync/sync.go
rianhughes: It was a bug, nice catch

vm/rust/src/juno_state_reader.rs
rianhughes: Yup

cmd/juno/juno.go
kirugan: >but if you think it's necessary to load multiple plugins

I meant maybe we should change comment to "Path to the plugin~s~ .so file" ?

sync/sync.go
```diff
@@ -317,7 +374,9 @@ func (s *Synchronizer) revertHead(forkBlock *core.Block) {
 	}
 
 	s.log.Infow("Reorg detected", "localHead", localHead, "forkHead", forkBlock.Hash)
-
+	if err != nil {
+		s.log.Warnw("Failed getting reverse state-diff, err: ", err)
+	}
```
kirugan, line 379: is this error check relevant? if so can we put it to if branch above i.e.:
```
if err == nil {
   ...
} else {
    s.log.Warnw(...)
}
```

sync/sync.go
```diff
@@ -231,6 +282,12 @@ func (s *Synchronizer) verifierTask(ctx context.Context, block *core.Block, stat
 			s.newHeads.Send(block.Header)
 			s.log.Infow("Stored Block", "number", block.Number, "hash",
 				block.Hash.ShortString(), "root", block.GlobalStateRoot.ShortString())
+			if s.plugin != nil {
+				err := (s.plugin).NewBlock(block, stateUpdate, newClasses)
```
kirugan, line 286: parenthesis are not required, can we remove them?

sync/sync.go
rianhughes: It's no longer relevant, sorry about that

sync/sync.go
```diff
@@ -231,6 +282,12 @@ func (s *Synchronizer) verifierTask(ctx context.Context, block *core.Block, stat
 			s.newHeads.Send(block.Header)
 			s.log.Infow("Stored Block", "number", block.Number, "hash",
 				block.Hash.ShortString(), "root", block.GlobalStateRoot.ShortString())
+			if s.plugin != nil {
```
kirugan, line 285: flow is a little bit different compared if block reverted: check `s.plugin != nil` happens inside `verifierTask`,  but for revert it's inside `handlePluginRevertBlock`. I don't have preferences on the way it should be handled, either in separate functions or in `verifierTask`

kirugan: Also for new block: we first store it and then notify plugin.
For revertion: we first notify plugin and then revert it in db

sync/sync.go
```diff
@@ -180,6 +188,48 @@ func (s *Synchronizer) fetchUnknownClasses(ctx context.Context, stateUpdate *cor
...
+		}
+		toSU, err = s.blockchain.StateUpdateByNumber(toBlock.Number)
+		if err != nil {
+			s.log.Warnw("Failed to retrieve the parents state-update for the plugin", "err", err)
+			return
+		}
+	}
+	err = (s.plugin).RevertBlock(
+		&junoplugin.BlockAndStateUpdate{Block: fromBlock, StateUpdate: fromSU},
+		&junoplugin.BlockAndStateUpdate{Block: toBlock, StateUpdate: toSU},
```
kirugan, line 226: maybe in case of `blockNumber == 0` it's better to pass nil explicitly here? because structure in this case won't have a lot of value (both values will be nil)

node/node.go
```diff
@@ -351,6 +366,16 @@ func (n *Node) Run(ctx context.Context) {
 		return
 	}
 
+	if n.plugin != nil {
+		go func() {
+			<-ctx.Done()
+			err := n.plugin.Shutdown()
```
kirugan, line 372: I think we should wait while `Shutdown()` function completes, right now there is no waitgroup. Maybe it's better to create some wrapper-service that calls plugin `Shutdown` in `Run(ctx)` wdyt?

docs/docs/plugins.md
```diff
@@ -0,0 +1,82 @@
...
+	NewBlock(block *core.Block, stateUpdate *core.StateUpdate, newClasses map[felt.Felt]core.Class) error
+	RevertBlock(from, to *BlockAndStateUpdate, reverseStateDiff *core.StateDiff) error
+}
+```
+
+**Init**: Called when the plugin is initialized. This can be used to set up database connections or any other necessary resources.
+
+**Shutdown**: Called when the Juno node is shut down. This can be used to clean up resources like database connections.
+
+**NewBlock**: Triggered when a new block is synced by the Juno client. Juno will send the block, the corresponding state update, and any new classes. Importantly, Juno waits for the plugin to finish processing this function call before continuing. This ensures that the plugin completes its task before Juno proceeds with the blockchain sync.
```
kirugan, line 26: > Importantly, Juno waits for the plugin to finish processing this function call before continuing.

Thanks 👍 this is major thing that we should point out to plugin developers

docs/docs/plugins.md
```diff
@@ -0,0 +1,82 @@
+---
+title: Juno Plugins
```
kirugan, line 2: do you think it's worth to mention separate repo with plugin example ?

core/state.go
```diff
@@ -559,19 +563,26 @@ func (s *State) Revert(blockNumber uint64, update *StateUpdate) error {
 		return err
 	}
 
+	if err = s.purgeNoClassContracts(); err != nil {
```
kirugan, line 566: was it after `purge deployed contracts` and does it matter btw ?

core/state.go
```diff
@@ -559,19 +563,26 @@ func (s *State) Revert(blockNumber uint64, update *StateUpdate) error {
...
 			return fmt.Errorf("purge contract: %v", err)
 		}
 	}
 
+	return s.verifyStateUpdateRoot(update.OldRoot)
+}
+
+func (s *State) purgeNoClassContracts() error {
 	// purge noClassContracts
 	//
```
kirugan, line 582: I think we can remove these 2 lines of comment - function name self explanatory

core/state.go
```diff
@@ -657,7 +667,7 @@ func (s *State) purgeContract(addr *felt.Felt) error {
 	return storageCloser()
 }
 
-func (s *State) buildReverseDiff(blockNumber uint64, diff *StateDiff) (*StateDiff, error) {
```
kirugan, line 660: are you sure that logic left the same?

sync/sync.go
rianhughes: > For revertion: we first notify plugin and then revert it in db

Yup, I wanted to ping the plugin after juno reverts, but this turned out to be a bit messy. The reason is that after Juno reverts, we no longer have the block and state-update that just got reverted. I could collect these before juno reverts, and then pass them into the logic that pings the plugin, but this felt like it would be messy. So I just decided to ping the plugin first.

rianhughes: >flow is a little bit different compared if block reverted

Yup, I'll update the code to be more consistent (handle the nil inside the plugin handler)

docs/docs/plugins.md
rianhughes: Currently it's private, so I just opted for placing the plugin example in the readme. Happy either way though

sync/sync.go
rianhughes: >flow is a little bit different compared if block reverted: check s.plugin != nil happens inside verifierTask, but for revert it's inside handlePluginRevertBlock

Updated to be more consistent (move nil handle outside of revert handler, since SendBlock() doesn't have a handler to put it in)

node/node.go
rianhughes: Yup, sounds good to me. Just upgraded the plugin to a service.

core/state.go
rianhughes: Yup, pretty sure the logic is left the same. I just extracted the deletion logic to a separate function, and they range over the same `blockNumber` and `stateDiff`, so everything should be the exact same (it's a lot clearer when comparing the changes in VSCode).

sync/sync.go
```diff
@@ -180,6 +188,49 @@ func (s *Synchronizer) fetchUnknownClasses(ctx context.Context, stateUpdate *cor
...
+		if err != nil {
+			s.log.Warnw("Failed to retrieve the parents state-update for the plugin", "err", err)
+			return
+		}
+		toBlockAndStateUpdate = &junoplugin.BlockAndStateUpdate{
+			Block:       toBlock,
+			StateUpdate: toSU,
+		}
+	}
+	err = (s.plugin).RevertBlock(
```
kirugan, line 225: can we remove paranthesis?

PR #2053, opened 2024-08-14
Remove listener check

kirugan:


node/http.go
```diff
@@ -82,9 +82,7 @@ func makeRPCOverHTTP(host string, port uint16, servers map[string]*jsonrpc.Serve
 	mux := http.NewServeMux()
 	for path, server := range servers {
 		httpHandler := jsonrpc.NewHTTP(server, log)
-		if listener != nil {
-			httpHandler = httpHandler.WithListener(listener)
-		}
+		httpHandler.WithListener(listener)
```
kirugan, line 85: zero value of listener is `nil` in this case inner logic treat listener as non initialized

obasekiosa: seems that on creation websocket object requires a valid listener.
and does no checks on usage within serveHTTP

```go
// node/websocket.go
func NewWebsocket(rpc *Server, log utils.SimpleLogger) *Websocket {
  ws := &Websocket{
    rpc:        rpc,
    log:        log,
    connParams: DefaultWebsocketConnParams(),
    listener:   &SelectiveListener{},
  }
  
  return ws
}

...
.
.
.
// node/websocket.go - serveHTTP
for {
  _, wsc.r, err = wsc.conn.Reader(wsc.ctx)
  if err != nil {
	  break
  }
  ws.listener.OnNewRequest("any")
  if err = ws.rpc.HandleReadWriter(wsc.ctx, wsc); err != nil {
	  break
  }
  // From websocket docs: "Read to EOF otherwise connection will hang."
  if _, err = io.Copy(io.Discard, wsc.r); err != nil {
	  break
  }
 }
```

obasekiosa: should we turn this todo into an issue?

```go
if err != nil {
		ws.log.Errorw("Failed to upgrade connection", "err", err)
		return
	}

	// TODO include connection information, such as the remote address, in the logs.

	wsc := newWebsocketConn(r.Context(), conn, ws.connParams)

```

kirugan:
```
listener: &SelectiveListener{},
```
🤦  yes, you're right. Ok let's keep it as is now, I'll close this PR

PR #2059, opened 2024-08-16
Add VM concurrency support

### LINKED ISSUE #2016: Add vm concurrency support
We should allow users to use the latest Blockifier feature, which is concurrent execution. We should add a Go flag (boolean) vm-concurrency-mode that will trigger parallel execution in Blockifier. This will require restructuring the Rust code that handles transactions.

Keep in mind that concurrent transactions may modify shared data structures (maps, slices), and we should protect them with mutexes on the Go side (CGo).

AnkushinDaniil:
Fixes #2016

This PR adds a new flag, `VMConcurrencyMode` and changes `cairoVMExecute`.

### DIFF
--- a/cmd/juno/juno.go
+++ b/cmd/juno/juno.go
@@ -82,6 +82,7 @@ const (
 	callMaxStepsF           = "rpc-call-max-steps"
 	corsEnableF             = "rpc-cors-enable"
 	versionedConstantsFileF = "versioned-constants-file"
+	vmConcurrencyModeF      = "vm-concurrency-mode"
 
 	defaultConfig                   = ""
 	defaulHost                      = "localhost"
@@ -119,6 +120,7 @@ const (
 	defaultGwTimeout                = 5 * time.Second
 	defaultCorsEnable               = false
 	defaultVersionedConstantsFile   = ""
+	defaultVMConcurrencyMode        = false
 
 	configFlagUsage                       = "The YAML configuration file."
 	logLevelFlagUsage                     = "Options: trace, debug, info, warn, error."
@@ -170,6 +172,7 @@ const (
 		"The upper limit is 4 million steps, and any higher value will still be capped at 4 million."
 	corsEnableUsage             = "Enable CORS on RPC endpoints"
 	versionedConstantsFileUsage = "Use custom versioned constants from provided file"
+	vmConcurrencyModeUsage      = "Enable VM concurrency mode"
 )
 
 var Version string
@@ -355,6 +358,7 @@ func NewCmd(config *node.Config, run func(*cobra.Command, []string) error) *cobr
 	junoCmd.Flags().Bool(corsEnableF, defaultCorsEnable, corsEnableUsage)
 	junoCmd.Flags().String(versionedConstantsFileF, defaultVersionedConstantsFile, versionedConstantsFileUsage)
 	junoCmd.MarkFlagsMutuallyExclusive(p2pFeederNodeF, p2pPeersF)
+	junoCmd.Flags().Bool(vmConcurrencyModeF, defaultVMConcurrencyMode, vmConcurrencyModeUsage)
 
 	junoCmd.AddCommand(GenP2PKeyPair(), DBCmd(defaultDBPath))
 
--- a/node/node.go
+++ b/node/node.go
@@ -66,6 +66,7 @@ type Config struct {
 	PendingPollInterval    time.Duration  `mapstructure:"pending-poll-interval"`
 	RemoteDB               string         `mapstructure:"remote-db"`
 	VersionedConstantsFile string         `mapstructure:"versioned-constants-file"`
+	VMConcurrencyMode      bool           `mapstructure:"vm-concurrency-mode"`
 
 	Metrics     bool   `mapstructure:"metrics"`
 	MetricsHost string `mapstructure:"metrics-host"`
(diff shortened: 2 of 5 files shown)


AnkushinDaniil:
Tested by running `./build/juno --http --vm-concurrency-mode` and requesting
```json
{
	"jsonrpc":"2.0",
	"method":"starknet_traceTransaction",
	"params":["0x5634f2847140263ba59480ad4781dacc9991d0365145489b27a198ebed2f969"],
	"id":1
}
```
Response body
```json
{
  "jsonrpc": "2.0",
  "result": {
    "type": "INVOKE",
    "execute_invocation": {
      "contract_address": "0x31c9cdb9b00cb35cf31c05855c0ec3ecf6f7952a1ce6e3c53c3455fcd75a280",
      "calldata": [
        "0x20cfa74ee3564b4cd5435cdace0f9c4d43b939620e4a0bb5076105df0a626c6",
        "0x5aee31408163292105d875070f98cb48275b8c87e80380b78d30647e05854d5"
      ],
      "caller_address": "0x0",
      "result": [],
      "calls": [
        {
          "contract_address": "0x20cfa74ee3564b4cd5435cdace0f9c4d43b939620e4a0bb5076105df0a626c6",
          "calldata": [
            "0x5aee31408163292105d875070f98cb48275b8c87e80380b78d30647e05854d5",
            "0x7e5"
          ],
          "caller_address": "0x31c9cdb9b00cb35cf31c05855c0ec3ecf6f7952a1ce6e3c53c3455fcd75a280",
          "result": [],
          "calls": [],
          "events": [],
          "messages": [],
          "execution_resources": { "steps": 25 }
        },
        {
          "contract_address": "0x20cfa74ee3564b4cd5435cdace0f9c4d43b939620e4a0bb5076105df0a626c6",
          "calldata": [
            "0x5aee31408163292105d875070f98cb48275b8c87e80380b78d30647e05854d5"
          ],
          "caller_address": "0x31c9cdb9b00cb35cf31c05855c0ec3ecf6f7952a1ce6e3c53c3455fcd75a280",
          "result": ["0x7e5"],
          "calls": [],
          "events": [],
          "messages": [],
          "execution_resources": { "steps": 31 }
        }
      ],
      "events": [],
      "messages": [],
      "execution_resources": { "steps": 178 }
    }
  },
  "id": 1
}
```

AnkushinDaniil:
The current problem is [this](https://github.com/NethermindEth/juno/actions/runs/10450945256/job/28936774568) bug. It happened in sequential mode.
My hypotheses were:
1. Transaction preparation loop:
In the updated code, transactions are first prepared and stored in a `Vec<Transaction>` before being executed. This is different from the old version, where each transaction was processed immediately after preparation. This change in flow can cause the error.
1. State management:
In the old code, the `CachedState` is modified transactionally and committed after each transaction. In the new code, transactions are prepared, executed in a batch, and then the state is committed. Delaying the commit of state changes until all transactions are executed might lead to issues.
1. The `charge_fee` and `validate` are hardcoded as `true` in the transaction executor, this might cause the error. I changed them both to `false`, it doesn't help.

AnkushinDaniil:
To make it work in my branch:
1. in `vm/rust` run `make clean`
1. [here](https://github.com/starkware-libs/sequencer/blob/cee1f5b985302cf3e5f0474491e6d53c76934840/crates/blockifier/src/blockifier/transaction_executor.rs#L95) and [here](https://github.com/starkware-libs/sequencer/blob/cee1f5b985302cf3e5f0474491e6d53c76934840/crates/blockifier/src/concurrency/worker_logic.rs#L133) `validate: false`
1. in `vm/rust` run `make all`
1. make juno
1. `./build/juno --http --network sepolia`  and ran a test query
1. `./build/juno --http --network sepolia --vm-concurrency-mode `  and ran a test query
test query: 
```json
{
    "id": 1,
    "jsonrpc": "2.0",
    "method": "starknet_estimateFee",
    "params": {
        "request": [
            {
            "type": "INVOKE",
            "sender_address": "0xf9e998b2853e6d01f3ae3c598c754c1b9a7bd398fec7657de022f3b778679",
            "calldata": [
                "0x1",
                "0x41a78e741e5af2fec34b695679bc6891742439f7afb8484ecd7766661ad02bf",
                "0x1987cbd17808b9a23693d4de7e246a443cfe37e6e7fbaeabd7d7e6532b07c3d",
                "0x4",
                "0x16342ade8a7cc8296920731bc34b5a6530f5ee1dc1bfd3cc83cb3f519d6530a",
                "0x77295b07f4b1d4e4fdb18b26faf95d5bdb0087da065b7b0b49d0470d9cbf852",
                "0x1",
                "0x0"
            ],
            "version": "0x100000000000000000000000000000001",
            "signature": [],
            "nonce": "0x2a91",
            "max_fee": "0x0"
            }
        ],
        "block_id": "pending",
        "simulation_flags": [
            "SKIP_VALIDATE"
        ]
    }
}
```

AnkushinDaniil:
[Pathfinder](https://github.com/eqlabs/pathfinder) team encountered [the same problem](https://github.com/eqlabs/pathfinder/issues/2125#issuecomment-2273434728)

rianhughes:
It would be good to add a test for the concurrency mode if possible

AnkushinDaniil:
I tried to measure performance by running multiple transactions and found a new bug.
The error is `Transaction with no fee transfer info must have zero fee.`
This happens [here](https://github.com/AnkushinDaniil/sequencer/blob/c529904b1c57963e1183aa66f2bca5c2bfc67cd9/crates/blockifier/src/concurrency/fee_utils.rs#L62-L63) only in concurrency mode. According to blockifier, in both modes `L1HandlerTransaction` must have `fee_transfer_call_info: None` , i.e. problem in `tx_execution_info.receipt.fee`, it's not equal to 0.

AnkushinDaniil:
> Also curious what the performance of simulating a block in concurrency mode is as well.

I roughly measured the performance difference. I used a slice of different txs from sepolia block №84662 with length 89.
I measured the time of `cairoVMExecute` execution.
Results:
1. Original version: 100%
1. Concurrency support: sequential mode: ~99,59%
1. Concurrency support: concurrent mode: ~115,76%

I'm trying to get a profile of each version, but I get a weird result
I get `ExternalCode` only in concurrent mode

image

Issue #2052, opened 2024-08-14
Add readiness endpoint for synced node


kirugan:
This feature was requested from [tg channel](https://t.me/StarknetJuno/1195). 
We need to add `/ready/synced` (we can discuss actual endpoint path later) endpoint that will return 200 http code if node is fully synced (withing N blocks maybe?) otherwise it should return 503 http code. Similar to what [pathfinder does](https://github.com/eqlabs/pathfinder?tab=readme-ov-file#synced)

obasekiosa:
Why have a dedicated enpoint when 
we can use the HEAD/OPTIONS method for all rpc endpoints to check if they are ready (like a pre-flight check).
Use status code or header value that says "how synced the node is"?

kirugan:
>  Use status code or header value that says "how synced the node is"?

How is this different from `/ready/synced` endpoint? instead of sending the header from all endpoints we can have dedicated endpoint for that.

obasekiosa:
> > Use status code or header value that says "how synced the node is"?
> 
> How is this different from `/ready/synced` endpoint? instead of sending the header from all endpoints we can have dedicated endpoint for that.

it would reduce the number of calls to just one and eliminate an extra endpoint.

kirugan:
We decided to keep `/ready/sync` only (see https://t.me/StarknetJuno/1224)

PR #2060, opened 2024-08-19
added /ready and /ready/sync endpoints

onuruci:
Implements #2052

Added /ready and /ready/sync endpoints to jsonrpcServer.

/ready endpoint return 200 http code if jsonrpcServer is ready to be used and returns 503 code else.

/ready/sync first does the same check with /ready endpoint and return 503 if not ready. If node is synced within last 6 blocks returns 200 http code and if not returns 503.


rpc/monitor.go
```diff
@@ -0,0 +1,36 @@
...
+		w.WriteHeader(http.StatusServiceUnavailable)
+		return
+	}
+	highestBlockHeader := h.syncReader.HighestBlockHeader()
+	if highestBlockHeader == nil {
+		w.WriteHeader(http.StatusServiceUnavailable)
+		return
+	}
+
+	if head.Number+6 >= highestBlockHeader.Number {
```
kirugan, line 31: let's move magic number (`6`) to a package level constant in this file

onuruci: moved in the new commit

rpc/monitor.go
```diff
@@ -0,0 +1,46 @@
...
+)
+
+func (h *Handler) HandleReadyRequest(w http.ResponseWriter, r *http.Request) {
+	if h.ready {
+		w.WriteHeader(http.StatusOK)
+	} else {
+		w.WriteHeader(http.StatusServiceUnavailable)
+	}
+}
+
```
rianhughes, line 18: nit:
func (h *Handler) HandleReadyRequest(w http.ResponseWriter, r *http.Request) {
	if h.ready {
		w.WriteHeader(http.StatusOK)
	}
	w.WriteHeader(http.StatusServiceUnavailable)
}

node/node.go
```diff
@@ -271,6 +272,8 @@ func New(cfg *Config, version string) (*Node, error) { //nolint:gocyclo,funlen
 		log.Warnw("Failed to parse Juno version, will not warn about new releases", "version", version)
 	}
 
+	rpcHandler.SetReady()
```
rianhughes, line 275: I wonder if we should set this in rpcHandler.Run()??

onuruci: Yes I think that will be better

onuruci: Updated this to setting rpcHandler ready in rpcHandler.run and updated tests so that they check readiness after rpcHandler.run

rianhughes (approved):

node/http.go
```diff
@@ -96,7 +102,7 @@ func makeRPCOverHTTP(host string, port uint16, servers map[string]*jsonrpc.Serve
 }
 
 func makeRPCOverWebsocket(host string, port uint16, servers map[string]*jsonrpc.Server,
-	log utils.SimpleLogger, metricsEnabled bool, corsEnabled bool,
+	log utils.SimpleLogger, metricsEnabled bool, corsEnabled bool, rpchandler *rpc.Handler,
```
kirugan, line 105: the reason why we need `/ready` and `/ready/sync` paths are readiness checks in kubernetes kind of environment. There is no need for now to have it on websocket. So I propose to revert this changes for websockets

node/http.go
```diff
@@ -86,6 +87,11 @@ func makeRPCOverHTTP(host string, port uint16, servers map[string]*jsonrpc.Serve
 			httpHandler = httpHandler.WithListener(listener)
 		}
 		mux.Handle(path, exactPathServer(path, httpHandler))
+
+		prefixedPath := strings.TrimSuffix(path, "/")
+
+		mux.HandleFunc(prefixedPath+"/ready", rpchandler.HandleReadyRequest)
+		mux.HandleFunc(prefixedPath+"/ready/sync", rpchandler.HandleReadySyncRequest)
```
kirugan, line 94: I think the best way to structure it would be: register `/ready` path in special http readiness handler (see `servers` variable). It would give us these benefits:

1. reduce number of readiness paths
2. reuse same logic for path handling assignments
3. separation of conerns: jsonrpc server struct should handle only corresponding requests and separate structure for readiness checks should handle its own kind of requests

wdyt?

node/http.go
onuruci: Removed it in 294c5eefcfc3ecd6465dbde1be62b0d5f8497314

Makefile
```diff
@@ -15,7 +15,6 @@ ifeq ($(shell uname -s),Darwin)
 
 	# Set macOS deployment target in order to avoid linker warnings linke
 	# "ld: warning: object file XX was built for newer macOS version (14.4) than being linked (14.0)"
-	export MACOSX_DEPLOYMENT_TARGET=$(shell sw_vers --productVersion)
```
kirugan, line 18: removed by mistake?

onuruci: Yes

node/http.go
onuruci: This sounds like the correct way. I separated the logic from jsonrpc server and moved it into a special http handler. Reduced the number of paths serving for readiness.

jsonrpc/readiness.go
```diff
@@ -0,0 +1,56 @@
...
+func NewReadinessHandlers(bcReader blockchain.Reader, syncReader sync.Reader) *ReadinessHandlers {
+	return &ReadinessHandlers{
+		bcReader:   bcReader,
+		syncReader: syncReader,
+	}
+}
+
+func (h *ReadinessHandlers) HandleReady(w http.ResponseWriter, _ *http.Request) {
+	w.WriteHeader(http.StatusOK)
+}
```
IronGauntlets, line 26: Since this function always returns http.StatusOK, do you think it is a good idea to inline this function on line 215 in node/node.go?

jsonrpc/readiness.go
```diff
@@ -0,0 +1,56 @@
...
+
+	if head.Number > highestBlockHeader.Number {
+		return false
+	}
+
+	if head.Number+SyncBlockRange >= highestBlockHeader.Number {
+		return true
+	} else {
+		return false
+	}
```
IronGauntlets, line 55:
```suggestion
	if head.Number+SyncBlockRange >= highestBlockHeader.Number {
		return true
	}
	return false
```

kirugan: good point, do you think it will be more readable as this:
```
return head.Number+SyncBlockRange >= highestBlockHeader.Number
```

jsonrpc/readiness.go
IronGauntlets, line 1: I think these files should not be under the node package, I don't think it is a good idea for jsonrpc package to import blockchain and sync packages.

jsonrpc/readiness.go
IronGauntlets: Yes, that is even better.

jsonrpc/readiness.go
IronGauntlets: Potentially under `node/http.go`file.

jsonrpc/readiness.go
kirugan: I asked about `/ready` endpoint in [tg](https://t.me/StarknetJuno/1224). Seems like we can safely remove it and just rely on / root path.

jsonrpc/readiness.go
kirugan: >  I don't think it is a good idea for jsonrpc package to import blockchain and sync packages.

good point, I'll move them to `http.go` are you ok with `ReadinessHandlers` structure though? + most likely I'll make it unexported

IronGauntlets (approved):
LGTM

PR #2070, opened 2024-08-19
Revert cors library back to v1.10.1

kirugan:


kirugan:
Cors issues wasn't due to the update

PR #2073, opened 2024-08-19
Replace custom code with maps package functions

kirugan:
Because of recent Go 1.23 update now we can use new functions from `maps` and remove our own (`utils.MapKeys` and `utils.MapValues`)

### DIFF
--- a/core/state.go
+++ b/core/state.go
@@ -5,6 +5,7 @@ import (
 	"encoding/binary"
 	"errors"
 	"fmt"
+	"maps"
 	"runtime"
 	"slices"
 	"sort"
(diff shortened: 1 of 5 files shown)


core/state.go
```diff
@@ -397,11 +398,8 @@ func (s *State) updateContractStorages(stateTrie *trie.Trie, diffs map[felt.Felt
 
 	// sort the contracts in decending diff size order
 	// so we start with the heaviest update first
-	keys := make([]felt.Felt, 0, len(diffs))
-	for key := range diffs {
-		keys = append(keys, key)
-	}
-	slices.SortStableFunc(keys, func(a, b felt.Felt) int { return len(diffs[a]) - len(diffs[b]) })
+	mapKeys := maps.Keys(diffs)
+	keys := slices.SortedStableFunc(mapKeys, func(a, b felt.Felt) int { return len(diffs[a]) - len(diffs[b]) })
```
weiihann, line 402: Minor nit pick:

`mapKeys` are not used anywhere, might be better to just do

```
keys := slices.SortedStableFunc(maps.Keys(diffs), func(a, b felt.Felt) int { return len(diffs[a]) - len(diffs[b]) })
```

core/state_update.go
```diff
@@ -143,12 +143,8 @@ func (d *StateDiff) Commitment() *felt.Felt {
...
 
 func sortedFeltKeys[V any](m map[felt.Felt]V) []felt.Felt {
-	keys := make([]felt.Felt, 0, len(m))
-	for addr := range m {
-		keys = append(keys, addr)
-	}
-	slices.SortFunc(keys, func(a, b felt.Felt) int { return a.Cmp(&b) })
-	return keys
+	keys := maps.Keys(m)
+	return slices.SortedFunc(keys, func(a, b felt.Felt) int { return a.Cmp(&b) })
```
weiihann, line 147: Same as previous

IronGauntlets (approved):

rianhughes (approved):

core/state.go
kirugan: done

core/state_update.go
kirugan: done

PR #2078, opened 2024-08-20
Make starknet_getTransactionReceipt and other function calls response conform to the 0.7.1 spec

### LINKED ISSUE #1805: `starknet_getTransactionReceipt` response doesn't conform to the 0.7.1 spec
I noticed that in some cases the response to `starknet_getTransactionReceipt` doesn't follow the spec for RPC 0.7.1.

For example the following request on mainnet

```json
{
	"id": 0,
	"jsonrpc": "2.0",
	"method": "starknet_getTransactionReceipt",
	"params": ["0x6b1e0c9f74d338fe8d992a6a67307275e9be9e311bfe5297c43bfd4ef16069e"]
}
```

Returns the following response

```json
{
	"jsonrpc": "2.0",
	"result": {
		"type": "INVOKE",
		"transaction_hash": "0x6b1e0c9f74d338fe8d992a6a67307275e9be9e311bfe5297c43bfd4ef16069e",
		"actual_fee": {
			"amount": "0x1106f22bd5d36",
			"unit": "WEI"
		},
		"execution_status": "SUCCEEDED",
		"finality_status": "ACCEPTED_ON_L1",
		"block_hash": "0x2bf7e9abd8b35622f7bfdb8c7a8379766f80bd4e7009b146acb2ced0004d44c",
		"block_number": 500000,
		"messages_sent": [],
		"events": [/* skip events */],
		"execution_resources": {
			"steps": 36134,
			"memory_holes": 9031,
			"pedersen_builtin_applications": 67,
			"range_check_builtin_applications": 2241,
			"bitwise_builtin_applications": 29,
			"segment_arena_builtin": 6
		}
	},
	"id": 0
}
```

Notice how the `execution_resources.data_availability` field is missing, [but it's marked as required in the spec](https://github.com/starkware-libs/starknet-specs/blob/v0.7.1/api/starknet_api_openrpc.json#L3901).

obasekiosa:
Fixes #1805 

## Changes

- Ensures `execution_resources.data_availability` is not omited for non `v0_6` api
- Adds default values of 0 for both `execution_resource.data_availability.l1_gas` and  `execution_resource.data_availability.l1_data_gas`
- Fixes mismatch api version calls in transaction tests..

#### What types of changes does your code introduce?

- [x] Bugfix (a non-breaking change that fixes an issue)
- [ ] New feature (a non-breaking change that adds functionality)
- [ ] Breaking change (a change that causes existing functionality not to work as expected)
- [ ] Optimization
- [ ] Refactoring
- [ ] Documentation update
- [ ] Build-related changes
- [ ] Other: Adds templates to improve contribution flow.

#### Requires testing

- [x] Yes
- [ ] No

#### If yes, did you write tests?

- [x] Yes
- [ ] No

#### Notes on testing
Some tests where updated, to match the required calls.
This is because some test calls previously made use of both v6 and v7 api within the same setup for the same test case.

#### Requires documentation update

- [ ] Yes
- [x] No

#### Requires explanation in Release Notes

- [ ] Yes
- [x] No

## Remarks
Updates the api response for `starknet_getTransactionReceipt` to match the spec for v0.7.
Since the field is required for every call in which `execution_resources` is returned it affects a couple of other remote procedures also.

### DIFF
--- a/rpc/helpers.go
+++ b/rpc/helpers.go
@@ -91,10 +91,7 @@ func adaptExecutionResources(resources *core.ExecutionResources, v0_6Response bo
 	if resources == nil {
 		if !v0_6Response {
 			return &ExecutionResources{
-				DataAvailability: &DataAvailability{
-					L1Gas:     uint64(0),
-					L1DataGas: uint64(0),
-				},
+				DataAvailability: &DataAvailability{},
 			}
 		} else {
 			return &ExecutionResources{}
@@ -117,10 +114,7 @@ func adaptExecutionResources(resources *core.ExecutionResources, v0_6Response bo
 	}
 	if !v0_6Response {
 		if resources.DataAvailability == nil {
-			res.DataAvailability = &DataAvailability{
-				L1Gas:     uint64(0),
-				L1DataGas: uint64(0),
-			}
+			res.DataAvailability = &DataAvailability{}
 		} else {
 			res.DataAvailability = &DataAvailability{
 				L1Gas:     resources.DataAvailability.L1Gas,
--- a/rpc/helpers.go
+++ b/rpc/helpers.go
@@ -89,12 +89,13 @@ func (h *Handler) blockHeaderByID(id *BlockID) (*core.Header, *jsonrpc.Error) {
 
 func adaptExecutionResources(resources *core.ExecutionResources, v0_6Response bool) *ExecutionResources {
 	if resources == nil {
+		var dataAvailability *DataAvailability
 		if !v0_6Response {
-			return &ExecutionResources{
-				DataAvailability: &DataAvailability{},
-			}
-		} else {
-			return &ExecutionResources{}
+			dataAvailability = &DataAvailability{}
+		}
+
+		return &ExecutionResources{
+			DataAvailability: dataAvailability,
 		}
 	}
 
--- a/rpc/helpers.go
+++ b/rpc/helpers.go
@@ -89,7 +89,16 @@ func (h *Handler) blockHeaderByID(id *BlockID) (*core.Header, *jsonrpc.Error) {
 
 func adaptExecutionResources(resources *core.ExecutionResources, v0_6Response bool) *ExecutionResources {
 	if resources == nil {
-		return &ExecutionResources{}
+		if !v0_6Response {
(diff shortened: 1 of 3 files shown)


rianhughes:
Looks good, but it would be nice if we had a test or two for cases where the data_availability fields are non-zero (ie RPCv07 responses for post 0.13.2 blocks)

rianhughes (commented):
Looks good, but it would be nice if we had a test or two for cases where the data_availability fields are non-zero (ie RPCv07 responses for post 0.13.2 blocks)

rpc/helpers.go
```diff
@@ -89,7 +89,16 @@ func (h *Handler) blockHeaderByID(id *BlockID) (*core.Header, *jsonrpc.Error) {
 
 func adaptExecutionResources(resources *core.ExecutionResources, v0_6Response bool) *ExecutionResources {
 	if resources == nil {
-		return &ExecutionResources{}
+		if !v0_6Response {
+			return &ExecutionResources{
+				DataAvailability: &DataAvailability{
+					L1Gas:     uint64(0),
+					L1DataGas: uint64(0),
```
kirugan, line 96: Let's use just `&DataAvailability{}` you don't need explicitly set these values to zero.
Here is good talk about this - https://www.youtube.com/watch?v=PAAkCSZUG1c&t=385s&ab_channel=TheGoProgrammingLanguage

rpc/helpers.go
```diff
@@ -106,10 +115,17 @@ func adaptExecutionResources(resources *core.ExecutionResources, v0_6Response bo
...
 	}
-	if !v0_6Response && resources.DataAvailability != nil {
-		res.DataAvailability = &DataAvailability{
-			L1Gas:     resources.DataAvailability.L1Gas,
-			L1DataGas: resources.DataAvailability.L1DataGas,
+	if !v0_6Response {
+		if resources.DataAvailability == nil {
+			res.DataAvailability = &DataAvailability{
+				L1Gas:     uint64(0),
+				L1DataGas: uint64(0),
```
kirugan, line 122: same

kirugan (approved):

rianhughes (approved):

PR #2088, opened 2024-08-25
Emit events on remote db

### LINKED ISSUE #1414: Emit events on Remote DB

RuneRogue:
So first I have added an listener field in DB struct then added code to emit event when a new transaction is created , view operation , update operation and when db is closed . I have also updated the WithListener function .

Fixed #1414

### DIFF
--- a/db/remote/db.go
+++ b/db/remote/db.go
@@ -2,6 +2,8 @@ package remote
 
 import (
 	"context"
+	"math"
+	"time"
 
 	"github.com/NethermindEth/juno/db"
 	"github.com/NethermindEth/juno/grpc/gen"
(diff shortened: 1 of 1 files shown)


PriyankMkwna:
I have made some changes that you have suggested.

db/remote/db.go
```diff
@@ -13,11 +14,11 @@ import (
 var _ db.DB = (*DB)(nil)
 
 type DB struct {
-	ctx context.Context
-
+	ctx        context.Context
 	grpcClient *grpc.ClientConn
 	kvClient   gen.KVClient
 	log        utils.SimpleLogger
+	listener   db.EventListener
```
kirugan, line 21: there are some improvements that we can do:
1. set `listener` to `SelectiveListener` in `New` function by default
2. remove nil checks for `d.listener`

db/remote/db.go
```diff
@@ -48,10 +55,19 @@ func (d *DB) View(fn func(txn db.Transaction) error) error {
 }
 
 func (d *DB) Update(fn func(txn db.Transaction) error) error {
-	return db.Update(d, fn)
+	start := time.Now()
+
+	err := db.Update(d, fn)
+
+	if d.listener != nil {
+		d.listener.OnCommit(time.Since(start))
```
kirugan, line 63: after removing `d.listener` check consider using `defer func() { d.listener.OnCommit(...)} ` as in the rest of codebase (but only here)

kirugan (changes requested):
get rid of nil checks + add defer in 1 place

PriyankMkwna:
Updated dg.go as you have suggested.

kirugan (approved):

AnkushinDaniil (approved):

rianhughes (approved):

kirugan:
@RuneRogue it's almost done: there is one minor change - see linter issues:
```
db/remote/db.go:31: File is not `gci`-ed with --skip-generated -s standard -s default (gci)
		OnIOCb:     func(write bool, duration time.Duration) {}, 
db/remote/db.go:51: File is not `gci`-ed with --skip-generated -s standard -s default (gci)
	
make: *** [Makefile:89: lint] Error 1
```
you can verify it locally with `make lint`

kirugan:
@RuneRogue thanks for contribution. Please check OnlyDust

Issue #2092, opened 2024-08-27
Return Total Gas Consumed from blockifier 


rianhughes:
TotalGasConsumed: nil, // todo: fill after 0.13.2

kirugan:
good catch! thanks

PR #2093, opened 2024-08-27
return receipts from blockifier (total gas)

rianhughes:
closes https://github.com/NethermindEth/juno/issues/2092

We need to return the total gas consumed (`TransactionReceipt.Gas`) from blockifier to correctly compute the receipts hash. This is particularly relevant for the sequencer, when it wants to shadow sepolia (the sequencer needs to compute the receipt hashes, but can't because it doesn't have access to the total gas consumed). Also note, I decided to return

```
pub struct TransactionReceipt {
    pub fee: Fee,
    pub gas: GasVector,
    pub da_gas: GasVector,
    // pub resources: TransactionResources, // Todo: not needed currently, can't serialize.
}
```

rather than just the `gas` field itself.


rianhughes:
Actually, I don't think main need this right now. It's more relevant for the sequencer, so closing for now

PR #2097, opened 2024-08-28
Hash-check

AnkushinDaniil:


kirugan:
Found what lead to hashes mismatch. TL;DR: 
There is a mismatches of `state diff commitment`,`transaction commitment` and `event commitment` (last two because of different hashing function pedersen instead of poseidon)

PR #2111, opened 2024-09-02
Implement Cairo0 Class Hash

weiihann:
### Description
This PR replaces the implementation of the Cairo0 class hash function from FFI to non-FFI (i.e. implement it directly in Golang).

### Rationale
Why this function was implemented using FFI:
- Implementing it in Golang is a pain. There's a specific formatting that we need to adhere to, and there are certain edge cases that we need to handle.

Using FFI removes the inconveniences. However, this means that other applications using Juno as a library must also do Rust compilation. It is not library-friendly at all means. While we also have `starknet/rust` and `vm/rust`, removing `core/rust` would at least reduce the surface area of using FFI, making it more library-friendly.

Now that new Cairo0 classes are no longer available, we can rewrite it in a more brute-force manner. That is, do a bunch of string formatting. The performance comparison shows that it is about 27% slower than using FFI (due to a bunch of string formatting), I think it's still worth switching over to the non-FFI implementation.

### Test suites
Tested this non-FFI implementation against all cairo0 class hashes on Starknet mainnet and sepolia, all passes.

### FFI vs NoFFI
```
                                                                                  │   ffi.txt    │              noffi.txt               │
                                                                                  │    sec/op    │    sec/op      vs base               │
ClassV0Hash/0x54328a1075b8820eb43caf0caa233923148c983742402dcfc38541dd843d01a-11    30.51m ± ∞ ¹    38.95m ± ∞ ¹  +27.65% (p=0.008 n=5)
ClassV0Hash/0x36c7e49a16f8fc760a6fbdf71dde543d98be1fee2eda5daff59a0eeae066ed9-11    18.85m ± ∞ ¹    22.64m ± ∞ ¹  +20.05% (p=0.008 n=5)
ClassV0Hash/0x07db5c2c2676c2a5bfc892ee4f596b49514e3056a0eee8ad125870b4fb1dd909-11   3.426m ± ∞ ¹    4.404m ± ∞ ¹  +28.54% (p=0.008 n=5)
ClassV0Hash/0x0772164c9d6179a89e7f1167f099219f47d752304b16ed01f081b6e0b45c93c3-11   37.59m ± ∞ ¹    47.24m ± ∞ ¹  +25.68% (p=0.008 n=5)
ClassV0Hash/0x028d1671fb74ecb54d848d463cefccffaef6df3ae40db52130e19fe8299a7b43-11   5.424m ± ∞ ¹    6.992m ± ∞ ¹  +28.91% (p=0.008 n=5)
ClassV0Hash/0x4367b26fbb92235e8d1137d19c080e6e650a6889ded726d00658411cc1046f5-11    7.116m ± ∞ ¹    8.472m ± ∞ ¹  +19.07% (p=0.008 n=5)
ClassV0Hash/0x35c753cbb572d7d0cbd6f725e4bc5f631d82003f3ac743160705b9ecd1befa7-11    829.8µ ± ∞ ¹   1157.0µ ± ∞ ¹  +39.44% (p=0.008 n=5)
ClassV0Hash/0x4fda56652f3dfe9c59904dd0fdc0fa5965b1e560e1641a06e132fdf90a96424-11    17.15m ± ∞ ¹    20.92m ± ∞ ¹  +21.96% (p=0.008 n=5)
ClassV0Hash/0x56fcc016a8ef4bd6310fa95de87c479a086fa1bf1934148c2bdf3b4cdbe4ac5-11    1.513m ± ∞ ¹    2.078m ± ∞ ¹  +37.36% (p=0.008 n=5)
geomean                                                                             7.336m          9.351m        +27.46%
```


core/class.go
```diff
@@ -46,8 +57,270 @@ func (c *Cairo0Class) Version() uint64 {
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Combine both ABI and Program JSON strings
+	var hintedClassHashJSON strings.Builder
+	hintedClassHashJSON.Grow(len(formattedABI) + len(formattedSpacesProgramStr))
+	hintedClassHashJSON.WriteString("{\"abi\": ")
+	hintedClassHashJSON.WriteString(formattedABI)
+	hintedClassHashJSON.WriteString(", \"program\": ")
```
kirugan, line 198: use backtick strings: `` `, "program": ` ``

kirugan: so there is no need to escape double quote

core/class.go
```diff
@@ -46,8 +57,270 @@ func (c *Cairo0Class) Version() uint64 {
...
+
+	var program Program
+	err = json.Unmarshal(definition.Program, &program)
+	if err != nil {
+		return nil, err
+	}
+
+	externalEntryPointElements := make([]*felt.Felt, 0, len(definition.EntryPoints.External)*2)
+	l1HandlerEntryPointElements := make([]*felt.Felt, 0, len(definition.EntryPoints.L1Handler)*2)
+	constructorEntryPointElements := make([]*felt.Felt, 0, len(definition.EntryPoints.Constructor)*2)
```
kirugan, line 75: why capacity is x2 of original data ?

core/class.go
```diff
@@ -46,8 +57,270 @@ func (c *Cairo0Class) Version() uint64 {
...
+	var hintedClassHashErr error
+
+	wg.Add(6) //nolint:mnd
+
+	go func() {
+		defer wg.Done()
+		for _, ep := range definition.EntryPoints.External {
+			externalEntryPointElements = append(externalEntryPointElements, ep.Selector, ep.Offset)
+		}
+		externalEntryPointHash = crypto.PedersenArray(externalEntryPointElements...)
```
kirugan, line 92: can we rewrite these kind of functions into something like:
```
for _, ep := range definition.EntryPoints.External {
   externalEntryPointsDigest.Update(ep.Selector, ep.Offset)
}
```
and later on call `externalEntryPointsDigest.Finish()` at the very bottom of this method?

Pros: we don't allocate additional memory
Cons: if there is a mistake potentially we waste more resources calculating digest

utils/strings_test.go
```diff
@@ -0,0 +1,87 @@
...
+
+func TestToPythonicJSON(t *testing.T) {
+	tests := []struct {
+		name     string
+		input    string
+		expected string
+	}{
+		{
+			name:     "Poop emoji as UTF-16 surrogate pair",
+			input:    `{"message":"Hello 💩!"}`,
```
kirugan, line 19: 😄

core/class.go
```diff
@@ -46,8 +57,270 @@ func (c *Cairo0Class) Version() uint64 {
...
+	err = json.Unmarshal(definition.Program, &program)
+	if err != nil {
+		return nil, err
+	}
+
+	// Use goroutines to parallelize hash computations
+	var wg sync.WaitGroup
+	var externalEntryPointHash, l1HandlerEntryPointHash, constructorEntryPointHash, builtInsHash, dataHash *felt.Felt
+	var hintedClassHash *felt.Felt
+	var hintedClassHashErr error
```
kirugan, line 77: I think we can improve this logic by:
1. replace `sync.WaitGroup` with `conc.NewWaitGroup()`
2. move common logic for EntryPoints in lambda and resuse for `External`, `L1Handler` and `Constructor`
3. as a consequence remove all digest variables below

i.e.:
```
	var externalEntryPointHash, l1HandlerEntryPointHash, constructorEntryPointHash, builtInsHash, dataHash *felt.Felt
	var hintedClassHash *felt.Felt
	var hintedClassHashErr error

	entryPointsDigest := func(entryPoints []starknet.EntryPoint) *felt.Felt {
		var epDigest crypto.PedersenDigest
		for _, ep := range entryPoints {
			epDigest.Update(ep.Selector, ep.Offset)
		}
		return epDigest.Finish()
	}
	wg := conc.NewWaitGroup()
	wg.Go(func() {
		externalEntryPointHash = entryPointsDigest(definition.EntryPoints.External)
	})
	wg.Go(func() {
		l1HandlerEntryPointHash = entryPointsDigest(definition.EntryPoints.L1Handler)
	})
	wg.Go(func() {
		constructorEntryPointHash = entryPointsDigest(definition.EntryPoints.Constructor)
	})
	wg.Go(func() {
		var builtInsDigest crypto.PedersenDigest
		for _, builtIn := range program.Builtins {
			builtInHex := hex.EncodeToString([]byte(builtIn))
			builtInFelt, err := new(felt.Felt).SetString("0x" + builtInHex)
			if err != nil {
				return
			}
			builtInsDigest.Update(builtInFelt)
		}
		builtInsHash = builtInsDigest.Finish()
	})
	wg.Go(func() {
		hintedClassHash, hintedClassHashErr = computeHintedClassHash(definition.Abi, definition.Program)
	})
	wg.Go(func() {
		var dataDigest crypto.PedersenDigest
		for _, data := range program.Data {
			dataFelt, err := new(felt.Felt).SetString(data)
			if err != nil {
				return
			}
			dataDigest.Update(dataFelt)
		}
		dataHash = dataDigest.Finish()
	})
	wg.Wait()
```

kirugan: and btw. are you sure that `.SetString("0x" + builtInHex)` is necessery?

core/class.go
```diff
@@ -46,8 +57,270 @@ func (c *Cairo0Class) Version() uint64 {
...
+		dataHash,
+	)
+
+	return clashHash, nil
+}
+
+// computeHintedClassHash calculates the hinted class hash by hashing the JSON combination
+// of ABI and Program.
+func computeHintedClassHash(abi, program json.RawMessage) (*felt.Felt, error) {
+	var mProgram Program
```
kirugan, line 163: what does "m" mean in `mProgram` ?

core/class.go
```diff
@@ -46,8 +57,270 @@ func (c *Cairo0Class) Version() uint64 {
...
+	if err != nil {
+		return nil, err
+	}
+	formattedABI, err := utils.ToPythonicJSON(stringifyABI)
+	if err != nil {
+		return nil, err
+	}
+
+	// Combine both ABI and Program JSON strings
+	var hintedClassHashJSON strings.Builder
```
kirugan, line 194: I propose to replace it with `bytes.Buffer` it has almost the same API, but there is no need to cast to []byte from string at the bottom because we can just call `buffer.Bytes()`

core/class.go
```diff
@@ -46,8 +57,270 @@ func (c *Cairo0Class) Version() uint64 {
...
+	if err != nil {
+		return "", err
+	}
+
+	// Determine the type of the JSON data
+	var jsonData any
+	if err := json.Unmarshal(jsonBytes, &jsonData); err != nil {
+		return "", err
+	}
+
```
kirugan, line 312: isn't `jsonData` is the same as `value` at this point ?

core/class.go
weiihann: Yeah, `hex.EncodeToString` outputs the hexadecimal representation without `0x` prefix. `felt.SetString` requires the parsed string to have `0x` prefix. You can check `felt_test.go` and verify the usage.

Here's a quick code snippet that you can verify hex encoding output in [go playground](https://go.dev/play/):
```
package main

import (
	"encoding/hex"
	"fmt"
)

func main() {
	// Sample data to encode
	data := []byte("Hello, Hex Encoding!")

	// Encode the data to hexadecimal string
	hexString := hex.EncodeToString(data)

	// Print the result
	fmt.Printf("Original: %s\n", data)
	fmt.Printf("Hex encoded: %s\n", hexString)
}
```

core/class.go
weiihann: modified program :D

core/class.go
weiihann: I changed `value` from `any` to `json.RawMessage`

Issue #2114, opened 2024-09-02
Juno traces overestimate data_availability


rianhughes:
```
{
    "jsonrpc": "2.0",
    "method": "starknet_traceTransaction",
    "params": {
        "transaction_hash": "0x7da7639e19a311e93cf83d87e2f92ea1d0e805b6b7a39f7e2975367a0362039"
    },
    "id": 1
}
```

Returns
```
"data_availability": {
                "l1_gas": 3750551,
                "l1_data_gas": 0
            }
```
but both the receipt, and pathfinder return

```
  "data_availability": {
                "l1_gas": 2754,
                "l1_data_gas": 0
            }
```

PR #2118, opened 2024-09-04
Fix traces (fees, state diff, events)

rianhughes:
closes https://github.com/NethermindEth/juno/issues/2114

Solution: 
1. pass data gas consumed directly from blockifier to users
2. set abi-len to zero for v0 classes


wojciechos (approved):

rpc/trace.go
```diff
@@ -292,32 +292,11 @@ func (h *Handler) traceBlockTransactions(ctx context.Context, block *core.Block,
...
-					dataGasPrice = header.L1DataGasPrice.PriceInFri
-				case WEI:
-					dataGasPrice = header.L1DataGasPrice.PriceInWei
-				}
-			}
-
-			dataGasFee := new(felt.Felt).Mul(dataGasConsumed[index], dataGasPrice)
-			gasConsumed := new(felt.Felt).Sub(overallFees[index], dataGasFee)
-			gasConsumed = gasConsumed.Div(gasConsumed, gasPrice) // division by zero felt is zero felt
-
```
kirugan, line 317: nice, finally we got rid of it

vm/rust/src/lib.rs
```diff
@@ -60,7 +60,7 @@ extern "C" {
     fn JunoAppendTrace(reader_handle: usize, json_trace: *const c_void, len: usize);
     fn JunoAppendResponse(reader_handle: usize, ptr: *const c_uchar);
     fn JunoAppendActualFee(reader_handle: usize, ptr: *const c_uchar);
-    fn JunoAppendDataGasConsumed(reader_handle: usize, ptr: *const c_uchar);
+    fn JunoAppendDataGasConsumed(reader_handle: usize, ptr: *const c_uchar, ptr: *const c_uchar);
```
kirugan, line 63: I think we should rename the function because it's no longer about data gas, but rather about gas consumed (L1 and L1 data gas). So, JunoAppendGasConsumed seems more appropriate, in my opinion.

kirugan (approved):

PR #2121, opened 2024-09-05
Improve build binaries pipeline

derrix060:
Changes made:
- build macOS arm and x86
- rename architecture from amd64 to x86_64, which is much easier to distinct from arm
- remove unnecessary step for self-hosted which is not being used anymore
- create ARTIFACT_NAME env var, so it's much easier to read the code
- remove unnecessary id
- add trigger on PR that change the build binaries files

### DIFF
--- a/.github/workflows/build-binaries.yml
+++ b/.github/workflows/build-binaries.yml
@@ -5,6 +5,11 @@ on:
     tags:
       - 'v*'
   workflow_dispatch:
+  pull_request:
+    branches:
+      - main
+    paths:
+      - .github/workflows/build-binaries.yml
 
 jobs:
   build:
@@ -13,11 +18,9 @@ jobs:
       matrix:
         include:
           - os: ubuntu-latest
-            architecture: amd64
-          - os: macOS-latest
-            architecture: amd64
+          - os: macos-13
           - os: ubuntu-arm64-4-core
-            architecture: arm64
+          - os: macos-latest
     runs-on: ${{ matrix.os }}
 
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/build-binaries.yml
```diff
@@ -13,11 +18,13 @@ jobs:
       matrix:
         include:
           - os: ubuntu-latest
-            architecture: amd64
-          - os: macOS-latest
-            architecture: amd64
+            architecture: x86_64
+          - os: macos-13
```
kirugan, line 22: why it's not latest ? 🤔

.github/workflows/build-binaries.yml
```diff
@@ -5,6 +5,11 @@ on:
     tags:
       - 'v*'
   workflow_dispatch:
+  pull_request:
+    branches:
+      - main
+    paths:
+      - .github/workflows/build-binaries.yml
```
kirugan, line 12: does this mean that the workflow run after PR is made on main branch ?

.github/workflows/build-binaries.yml
derrix060: so we can create a x86 for mac

.github/workflows/build-binaries.yml
derrix060: it means that this workflow will run on PRs that change this file (for example this one)

.github/workflows/build-binaries.yml
derrix060: Basically now we will create mac x86, mac arm, linux x86 and linux arm

wojciechos (approved):

derrix060:
Results:  https://github.com/NethermindEth/juno/actions/runs/10718771544?pr=2121

image

.github/workflows/build-binaries.yml
kirugan: makes sense, thanks

kirugan (approved):

PR #2127, opened 2024-09-09
Fix p2p header adapters

kirugan:


adapters/core2p2p/block.go
```diff
@@ -45,16 +45,15 @@ func AdaptHeader(header *core.Header, commitments *core.BlockCommitments,
...
-		Receipts:        AdaptHash(&felt.Zero),
+		Receipts:        AdaptHash(commitments.ReceiptCommitment),
 		ProtocolVersion: header.ProtocolVersion,
-		GasPriceFri:     AdaptUint128(header.GasPrice),
+		GasPriceFri:     AdaptUint128(header.GasPriceSTRK),
 		Signatures:      utils.Map(header.Signatures, AdaptSignature),
 		StateDiffCommitment: &spec.StateDiffCommitment{
 			StateDiffLength: stateDiffLength,
-			Root:            AdaptHash(stateDiffCommitment),
+			Root:            AdaptHash(stateDiffCommitment), // todo commitments
```
kirugan, line 54: I want to debug why it works for pathfinder if we pass `commitments.StateDiffCommitment` here

kirugan: I passed incorrect value initially it should be not `stateDiff.Commitment()`, but `stateDiff.Hash()`

rianhughes:
`func (d *StateDiff) Commitment() *felt.Felt {...}`

doesn't seem to be used anywhere, can we remove it?

rianhughes (approved):

kirugan:
> doesn't seem to be used anywhere, can we remove it?

I think yes, but I'm checking with Aneeque. Will remove in next PR if we don't need it anymore

AnkushinDaniil (approved):

Issue #2131, opened 2024-09-11
RPC fix - txn status


rianhughes:
Junos  [adaptTransactionStatus()](https://github.com/NethermindEth/juno/blob/main/rpc/transaction.go#L778-L787) incorrectly handles the ["NOT_RECEIVED"](https://github.com/NethermindEth/juno/blob/main/starknet/transaction.go#L38) finality status that the FGW can return. Currently we return an error "unknown finality status 3" when we receive this response from the FGW.

However, the rpc spec doesn't require it (see [here](https://github.com/starkware-libs/starknet-specs/blob/v0.7.1/api/starknet_api_openrpc.json#L247) and [here](https://github.com/starkware-libs/starknet-specs/blob/v0.7.1/api/starknet_api_openrpc.json#L3200)), so it should just be ignored / we should not return an error.

PR #2132, opened 2024-09-11
correct rpc transaction status

rianhughes:
close https://github.com/NethermindEth/juno/issues/2131


rpc/transaction.go
```diff
@@ -791,8 +793,6 @@ func adaptTransactionStatus(txStatus *starknet.TransactionStatus) (*TransactionS
 		status.Execution = TxnSuccess
 	case starknet.Reverted:
 		status.Execution = TxnFailure
-	case starknet.Rejected:
-		status.Finality = TxnStatusRejected
```
kirugan, line 795: are you sure that `Rejected` is part of finality status? according to this image it should be in `ExecutionStatus`:
image

rianhughes: https://docs.starknet.io/architecture-and-concepts/network-architecture/transaction-life-cycle/

rianhughes: This doc suggests it should be apart of Finality Status

rianhughes: Same with the spec https://github.com/starkware-libs/starknet-specs/blob/v0.7.1/api/starknet_api_openrpc.json#L3200-L3210

rianhughes:
It's not clear that any FullNode can actually serve a finality status that is Received or Rejected, since:

1. The FGW doesn't include the finality status in the receipts in a block, and 

2. the FGW "get_transaction_receipt" endpoint is deprecated 

https://alpha-mainnet.starknet.io/feeder_gateway/get_transaction_receipt?transactionHash=0x1b59029be1c10c12298f1bb272b1c3c622e1e89a379e18f77ae33bcb3886379
returns
{"code": "StarknetErrorCode.DEPRECATED_ENDPOINT", "message": "Endpoint is deprecated from version 0.12.3."}.

By default we assume that the finality status is Accepted_on_l2, and check if the transaction has been included in a block behond the L1 head to set if to Accepted_on_l1.

rianhughes:
This endpoint is a bit strange, because the FGW can return "NOT_RECEIVED" as the finality status, but the RPC spec does not allow returning this value. This means users who query for the transaction status of a transaction that is unknown to the FGW will receive an unofficial error (eg "unknown FinalityStatus.."), rather than the "NOT_RECEIVED" finality status (which the docs suggest).

eg https://alpha-mainnet.starknet.io/feeder_gateway/get_transaction?transactionHash=0x1b59029be1c10c12298f1bb272b1c3c622e1e89a379e18f77ae33bcb3886379

rianhughes:
Closing since this PR is incorrect

PR #2134, opened 2024-09-11
Bump Go version to 1.23.1

AnkushinDaniil:
https://groups.google.com/g/golang-announce/c/K-cEzDeCtpc

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,7 +1,7 @@
 module github.com/NethermindEth/juno
 
 // if version specified as "1.22" (without bugfix) it breaks CodeQL github build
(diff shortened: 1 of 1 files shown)


derrix060 (approved):

go.mod
```diff
@@ -1,7 +1,7 @@
 module github.com/NethermindEth/juno
 
 // if version specified as "1.22" (without bugfix) it breaks CodeQL github build
-go 1.23.0
+go 1.23.1
```
kirugan, line 4: do you think we should update docs as well?

PR #2144, opened 2024-09-13
Add Starknet Go tests to CI/CD pipeline

wojciechos:
- Integrate starknet-go-tests workflow into development, staging, and production stages
- Ensure Go tests run after validation/promotion in each stage

tested [here](https://github.com/NethermindEth/juno/actions/runs/10853254695/job/30121571258)

### DIFF
--- a/.github/workflows/ci-cd-pipeline.yml
+++ b/.github/workflows/ci-cd-pipeline.yml
@@ -98,7 +98,6 @@ jobs:
     uses: ./.github/workflows/starknet-go-tests.yml
     secrets:
       TEST_RPC_URL: ${{ secrets.DEV_SEPOLIA_URL }}/v0_7
-      TEST_ACCOUNT_PRIVATE_KEY: ${{ secrets.TEST_ACCOUNT_PRIVATE_KEY }}
 
   promote_to_staging:
     needs: [build_docker_image, validate_dev]
@@ -149,7 +148,6 @@ jobs:
     uses: ./.github/workflows/starknet-go-tests.yml
     secrets:
       TEST_RPC_URL: ${{ secrets.STAGING_SEPOLIA_URL }}/v0_7
-      TEST_ACCOUNT_PRIVATE_KEY: ${{ secrets.TEST_ACCOUNT_PRIVATE_KEY }}
 
   promote_to_production:
     needs: [build_docker_image, promote_to_staging]
@@ -199,4 +197,3 @@ jobs:
     uses: ./.github/workflows/starknet-go-tests.yml
     secrets:
       TEST_RPC_URL: ${{ secrets.PROD_SEPOLIA_URL }}/v0_7
-      TEST_ACCOUNT_PRIVATE_KEY: ${{ secrets.TEST_ACCOUNT_PRIVATE_KEY }}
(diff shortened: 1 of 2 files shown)


.github/workflows/starknet-go-tests.yml
```diff
@@ -0,0 +1,33 @@
+name: starknet-go tests
+
+on:
+  workflow_call:
+    secrets:
+      TEST_RPC_URL:
+        required: false
```
derrix060, line 7: what happens when this is not provided?  I'm leaning toward thinking that these two are required...

.github/workflows/starknet-go-tests.yml
```diff
@@ -0,0 +1,33 @@
...
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout the repo
+        uses: actions/checkout@v4
+        with:
+          repository: NethermindEth/starknet.go
+          ref: main
```
derrix060, line 19: nit: this could be an argument to this workflow, defaulting to main, so you can test releases there without changing much stuff around :)

.github/workflows/starknet-go-tests.yml
```diff
@@ -0,0 +1,33 @@
...
+
+      - name: Setup Go
+        uses: actions/setup-go@v4
+        with:
+          go-version: '1.23'
```
derrix060, line 24: Does it need to be the same as juno?  if so, It needs to be 1.23.1: https://github.com/NethermindEth/juno/blob/main/go.mod#L4

.github/workflows/ci-cd-pipeline.yml
```diff
@@ -137,6 +144,13 @@ jobs:
       TEST_ACCOUNT_ADDRESS: ${{ secrets.TEST_ACCOUNT_ADDRESS }}
       TEST_ACCOUNT_PRIVATE_KEY: ${{ secrets.TEST_ACCOUNT_PRIVATE_KEY }}
 
+  staging-starknet-go-tests:
+    needs: [promote_to_staging]
+    uses: ./.github/workflows/starknet-go-tests.yml
+    secrets:
+      TEST_RPC_URL: ${{ secrets.STAGING_SEPOLIA_URL }}/v0_7
+      TEST_ACCOUNT_PRIVATE_KEY: ${{ secrets.TEST_ACCOUNT_PRIVATE_KEY }}
```
derrix060, line 152: Are we planning to have different test accounts depending on the environment?  If no, I'd say to remove this as argument and just use in the starknet-go-tests the secret directly :)

.github/workflows/starknet-go-tests.yml
wojciechos: right, it should be rather required - updated

.github/workflows/starknet-go-tests.yml
wojciechos: thanks, argument added

.github/workflows/starknet-go-tests.yml
wojciechos: this go version is for building starknet.go proj, not juno, so don't have to be same

.github/workflows/ci-cd-pipeline.yml
wojciechos: Yep, would be great to have different accounts per env in the future, as right now we can't run them parallel at all, because some of tests might fail(e.g due to invalid nonce while adding the txn)

derrix060 (approved):

PR #2145, opened 2024-09-13
feat: Add GitHub Actions workflow to find the smallest supported Rust version

derrix060:
The idea of this workflow is to ensure we have up-to-date documentation on
the minimum required version of rust.  It will only run on PRs that touch either
this workflow or rust files

Since there is no easy way to have a workflow to be required and only run on
some cases, it will send a message to the PR every time the minimum version
changes

### DIFF
--- a/.github/workflows/find-smallest-rust.yml
+++ b/.github/workflows/find-smallest-rust.yml
@@ -1,0 +1,42 @@
+name: Minimum Supported Rust
+
+on:
+  pull_request:
+    branches:
+      - main
+    paths:
+      - .github/workflows/find-smallest-rust.yml
+      - core/rust/*
+      - vm/rust/*
+      - starknet/rust/*
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  find-smallest-rust:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Find smallest supported Rust version
+        id: rust-version
+        uses: derrix060/detect-rust-minimum-version@v1
+        with:
+          paths: core/rust/,vm/rust/,starknet/rust/
+      - name: Send notification on PR
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const msrv = '${{ steps.rust-version.outputs.highest-msrv }}'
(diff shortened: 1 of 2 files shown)


.github/workflows/find-smallest-rust.yml
```diff
@@ -0,0 +1,42 @@
...
+  cancel-in-progress: true
+
+jobs:
+  find-smallest-rust:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Find smallest supported Rust version
+        id: rust-version
+        uses: derrix060/detect-rust-minimum-version@v1
```
wojciechos, line 25: should we consider moving the action from your personal account to the NethermindEth org? It might help with better control, maintenance, and security in the long run.

derrix060: well, this is actually not nethermind specific, it is just a simple wrapper around cargo-msrv which can be used by any organisation

.github/workflows/find-smallest-rust.yml
```diff
@@ -0,0 +1,42 @@
...
+          paths: core/rust/,vm/rust/,starknet/rust/
+      - name: Send notification on PR
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const msrv = '${{ steps.rust-version.outputs.highest-msrv }}'
+            const previous_msrv = '1.80.1'
+
+            if (msrv < previous_msrv) {
+              github.rest.issues.createComment({
```
kirugan, line 36: so it will post comment only if new msrv is < 1.80.1 (for now) right? if msrv changes to 1.90 it won't be noticed ?

.github/workflows/find-smallest-rust.yml
wojciechos: Yep, it’s generic, but still, moving your action to the NethermindEth org could help with more control and maintenance. I remember the core team had a similar case and preferred hosting things internally rather than relying on non-official actions hosted on individual accounts.

wojciechos (approved):

.github/workflows/find-smallest-rust.yml
derrix060: woow, thanks for spotting, it should actually by if `msrv != previous_msrv`, I wanted to comment in case there is any change

PR #2146, opened 2024-09-15
Fix p2p sync for 0.13.2 blocks

kirugan:


core/block.go
```diff
@@ -131,7 +131,7 @@ func VerifyBlockHash(b *Block, network *utils.Network, stateDiff *StateDiff) (*B
 // and by then issues with unverifiable block hash were resolved.
 // In future, this may no longer be required.
 // Todo: Pass stateDiff so that p2p layer can calculate post 0.13.2 Block Hash
-func BlockHash(b *Block) (*felt.Felt, error) {
+func BlockHash(b *Block) (*felt.Felt, error) { // todo remove function?
```
kirugan, line 134: @IronGauntlets wdyt ? looks like we used it for spec without block hashes

IronGauntlets: I think this should be used to check the block hash given by the block in the p2p spec. I think it is good to keep it still.

kirugan: As far as I remember we already check it:
1. in p2p we call `SanityCheckNewHeight` :
```
commitments, err := s.blockchain.SanityCheckNewHeight(coreBlock, stateUpdate, newClasses)
```
2. inside of it we call `core.VerifyBlockHash` 
3. which in turn verifies provided hash with `blockHash` function

IronGauntlets: yeah, you are right, after adding the block hash to the block in p2p spec this function is not required.

IronGauntlets (approved):

AnkushinDaniil (approved):

PR #2147, opened 2024-09-16
Remove repeated close host call

weiihann:
The previous line already does `defer s.host.Close()`, so there's no need to do `return s.host.Close()` again.


rianhughes (approved):

p2p/p2p.go
```diff
@@ -268,7 +268,7 @@ func (s *Service) Run(ctx context.Context) error {
 	if err := s.dht.Close(); err != nil {
 		s.log.Warnw("Failed stopping DHT", "err", err.Error())
 	}
-	return s.host.Close()
+	return nil
```
kirugan, line 271: can you please log error in `s.host.Close()` call in defer ?

kirugan (approved):

PR #2149, opened 2024-09-16
Add a filter for nodes to exclude a host node

AnkushinDaniil:


p2p/p2p.go
```diff
@@ -408,6 +408,11 @@ func (s *Service) persistPeers() error {
 
 	store := s.host.Peerstore()
 	peers := store.Peers()
+
+	peers = utils.Filter(peers, func(peerID peer.ID) bool {
+		return peerID != s.host.ID()
+	})
+
```
weiihann, line 415: Nice catch!

Perhaps it's better to do just this
```suggestion
	peers := utils.Filter(store.Peers(), func(peerID peer.ID) bool {
		return peerID != s.host.ID()
	})
```

weiihann (approved):

derrix060 (approved):
nice catch!

a small nit, there is already a foor loop on the peers, we could just do a check there if is not host and move on to the next, instead of iterating over the peers list twice, but that's not a big deal since the list will be small :)

PR #2154, opened 2024-09-16
chore: Update golangci-lint to version 1.61.0

AnkushinDaniil:


kirugan (approved):

weiihann (approved):
Just for context, I believe why loop variables are recreated inside the loop scope (`s := s`) is because of how closures work in Golang.

But in Go1.22 onwards, they changed so that each loop variable is unique to each iteration, so something like `s := s` is no longer needed.

PR #2155, opened 2024-09-16
Refactor persistPeers function

AnkushinDaniil:
As @derrix060 [mentioned](https://github.com/NethermindEth/juno/pull/2149#pullrequestreview-2306463964), we need to avoid iterating peers twice.

### DIFF
--- a/p2p/p2p.go
+++ b/p2p/p2p.go
@@ -407,10 +407,11 @@ func (s *Service) persistPeers() error {
 	}
 
 	store := s.host.Peerstore()
-	peers := utils.Filter(store.Peers(), func(peerID peer.ID) bool {
-		return peerID != s.host.ID()
-	})
+	peers := store.Peers()
 	for _, peerID := range peers {
+		if peerID == s.host.ID() {
+			continue
+		}
 		peerInfo := store.PeerInfo(peerID)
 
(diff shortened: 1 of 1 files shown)


derrix060 (approved):

p2p/p2p.go
```diff
@@ -407,10 +407,11 @@ func (s *Service) persistPeers() error {
...
 
 	store := s.host.Peerstore()
-	peers := utils.Filter(store.Peers(), func(peerID peer.ID) bool {
-		return peerID != s.host.ID()
-	})
+	peers := store.Peers()
 	for _, peerID := range peers {
+		if peerID == s.host.ID() {
+			continue
+		}
```
kirugan, line 414: There is `s.log.Infow("Stored peers", "num", len(peers))` below in this case it won't log correct value

weiihann:
While you're at it, there's a similar code pattern in the `p2p/sync.go`. That can be refactored too.
https://github.com/NethermindEth/juno/blob/main/p2p/sync.go#L632-L649

p2p/p2p.go
AnkushinDaniil: Yes, you're right. I think it's ok to have 2 iterations in this case.

PR #289, opened 2022-07-16
Upgrade GH Actions

D-DePablos:
Upgraded relevant GH Actions in respective files.

Set node version properly.

PR #2156, opened 2024-09-16
Native contract executor

PearsonWhite:
Corresponding sequencer PR: NethermindEth/sequencer#19

PR: #2156

### DIFF
--- a/vm/rust/src/juno_state_reader.rs
+++ b/vm/rust/src/juno_state_reader.rs
@@ -7,18 +7,15 @@ use blockifier::{
     state::state_api::{StateReader, StateResult},
 };
 use cached::{Cached, SizedCache};
-use cairo_lang_sierra::{program::Program, program_registry::ProgramRegistry};
-use cairo_native::{
-    context::NativeContext, error::Error as NativeError, executor::AotNativeExecutor,
-    metadata::gas::GasMetadata, module::NativeModule,
-};
-use libloading::Library;
+use cairo_native::executor::contract::ContractExecutor;
+use cairo_native::OptLevel;
 use once_cell::sync::Lazy;
 use serde::Deserialize;
 use starknet_api::core::{ClassHash, CompiledClassHash, ContractAddress, Nonce};
 use starknet_api::state::StorageKey;
 use starknet_types_core::felt::Felt;
 use std::cell::RefCell;
+use std::sync::Arc;
 use std::{
     ffi::{c_char, c_uchar, c_void, CStr},
     fs,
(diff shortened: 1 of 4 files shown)


xrvdg (changes requested):
`ContractExecutor::new` does not do any caching to disk. The `compile_and_load`, `load_compiled_contract` and others needs to be reworked with `ContractExecutor::save` and `ContractExecutor::load`.

PearsonWhite:
> `ContractExecutor::new` does not do any caching to disk. The `compile_and_load`, `load_compiled_contract` and others needs to be reworked with `ContractExecutor::save` and `ContractExecutor::load`.

ContractExecutor's entrypoints are not public, so we must either:
1. Make them public. Then we can remove entry_points and Hash from NativeContractClassV1. (It looks like NativeContractClassV1 doesn't use PartialEq for anything). This would make NativeContractClassV1 a nothing more than a wrapper around ContractExecutor.
2. Create save/load functions for NativeContractClassV1 to save/load class hash. The de/serialization would de/serialize hash, entrypoints, and then use (executor.save/load).

xrvdg:
> ContractExecutor's entrypoints are not public, so we must either:

Those don't have to be public, it's handled by the load/save of `ContractExecutor`. The `NativeContractEntryPoints` are there to have a translation for starknet selector to function id. We are in talks with LambdaClass to have the interface different, but for now it is like this. 

> It looks like NativeContractClassV1 doesn't use PartialEq for anything

The `PartialEq` follows from the `PartialEq` on `ContractClass` which is due to a test.

vm/rust/src/juno_state_reader.rs
```diff
@@ -311,68 +288,16 @@ fn native_try_from_json_string(
...
-    }
-
-    library_output_path
-        .is_file()
-        .then_some(load(sierra_program, library_output_path))
+    let executor = ContractExecutor::load(&library_output_path).unwrap_or({
+        let executor = ContractExecutor::new(&sierra_program, OptLevel::Default)?;
+        executor.save(&library_output_path)?;
+        executor
+    });
```
xrvdg, line 298: This will always compile and save the contract as the argument to `unwrap_or` is eagerly evaluated. Use `unwrap_or_else` instead.

xrvdg: Furthermore, `load` and `save` [needlessly borrow](https://rust-lang.github.io/rust-clippy/master/index.html#/needless_borrow) the `library_output_path`.

xrvdg (changes requested):
Three points to address:
- caching has no effect if it's always invalidated
- format the code with `cargo fmt`. 
- edit the Makefile so that Juno can be built.

vm/rust/src/juno_state_reader.rs
PearsonWhite: Changed to `or_else` so I could still use "?" operator. Had to specify the type though since it couldn't figure it out.

xrvdg (approved):
The change itself looks good and I've approved it. But before merging please look into the following regression first. 

Block 633333 fails with:
> execute transaction #289: failed txn 1132083247661400189175640804679018936217786098286704013683829309015708481258 reason: L1 gas bounds (max amount: 1038, max price: 67469438047820) exceed balance (44664767982987702).

PR #2163, opened 2024-09-19
Add metrics for gossipsub

weiihann:
Adds additional Prometheus metrics for P2P related operations


derrix060 (approved):
That's really cool!  Nice work

Makefile
```diff
@@ -110,6 +110,7 @@ feedernode: juno-cached
 	./build/juno \
 	--network=sepolia \
 	--log-level=debug \
+	--metrics \
```
kirugan, line 113: this command is only for devs: we don't need metrics flag here

p2p/p2p.go
```diff
@@ -244,7 +245,7 @@ func (s *Service) Run(ctx context.Context) error {
 		return err
 	}
 
-	s.pubsub, err = pubsub.NewGossipSub(ctx, s.host)
+	s.pubsub, err = pubsub.NewGossipSub(ctx, s.host, pubsub.WithRawTracer(s.gossipTracer))
```
kirugan, line 248: if caller doesn't call `WithGossipTracer()` property `gossipTracer` is nil which will lead to:
```
return func(p *PubSub) error {
		if p.tracer != nil {
			p.tracer.raw = append(p.tracer.raw, tracer)
		} else {
                        // this branch
			p.tracer = &pubsubTracer{raw: []RawTracer{tracer}, pid: p.host.ID(), idGen: p.idGen}
		}
		return nil
	}
```
so I think we should pass this option only if `s.gossipTracer` is not nil, something like this:
```
var options []GossipOption
if s.gossipTracer != nil {
    options = append(options, pubsub.WithRawTracer(s.gossipTracer))
}
s.pubsub, err = pubsub.NewGossipSub(ctx, s.host, options...)
```

p2p/metrics.go
```diff
@@ -1,18 +1,111 @@
...
 
-func runMetrics(store peerstore.Peerstore) {
-	numberOfPeersGauge := prometheus.NewGaugeFunc(prometheus.GaugeOpts{
-		Namespace: "p2p",
-		Name:      "number_of_peers",
-	}, func() float64 {
-		peersLen := store.Peers().Len()
-		return float64(peersLen)
-	})
+var knownAgentVersions = []string{"juno", "papyrus", "pathfinder", "madara"}
```
kirugan, line 8: let's move it to `p2p/gossip_tracer.go` because it's not used in this file

p2p/gossip_tracer.go
```diff
@@ -0,0 +1,115 @@
...
+}
+
+func agentNameFromPeerID(pid peer.ID, store peerstore.Peerstore) string {
+	rawAgent, err := store.Get(pid, "AgentVersion")
+	agent, ok := rawAgent.(string)
+	if err != nil || !ok {
+		return unknownAgent
+	}
+	for _, knownAgent := range knownAgentVersions {
+		if strings.Contains(strings.ToLower(agent), knownAgent) {
```
kirugan, line 110: nit: let's move strings.ToLower out of the loop:
```
agent = strings.ToLower(agent)
for ...
```

p2p/gossip_tracer.go
```diff
@@ -0,0 +1,115 @@
...
+type gossipTracer struct {
+	host    host.Host
+	metrics *P2PMetrics
+}
+
+func NewGossipTracer(h host.Host) *gossipTracer {
+	return &gossipTracer{host: h, metrics: NewP2PMetrics()}
+}
+
+func (g gossipTracer) AddPeer(p peer.ID, proto protocol.ID) {
```
kirugan, line 30: is there any reason receiver (i.e. `g`) is not a pointer?

p2p/gossip_tracer.go
```diff
@@ -0,0 +1,115 @@
...
+	for _, msg := range rpc.Publish {
+		pubCtr.WithLabelValues(*msg.Topic).Inc()
+	}
+}
+
+func agentNameFromPeerID(pid peer.ID, store peerstore.Peerstore) string {
+	rawAgent, err := store.Get(pid, "AgentVersion")
+	agent, ok := rawAgent.(string)
+	if err != nil || !ok {
+		return unknownAgent
```
kirugan, line 107: since this constant don't need to be exposed elsewhere, maybe put it inside this function?

p2p/gossip_tracer.go
```diff
@@ -0,0 +1,115 @@
...
+	g.setRPCMetrics(g.metrics.PSRPCSubDrop, g.metrics.PSRPCPubDrop, g.metrics.PSRPCDrop, rpc)
+}
+
+func (g gossipTracer) setRPCMetrics(subCtr prometheus.Counter, pubCtr, ctrlCtr *prometheus.CounterVec, rpc *pubsub.RPC) {
+	subCtr.Add(float64(len(rpc.Subscriptions)))
+	if rpc.Control != nil {
+		ctrlCtr.WithLabelValues("ihave").Add(float64(len(rpc.Control.Ihave)))
+		ctrlCtr.WithLabelValues("iwant").Add(float64(len(rpc.Control.Iwant)))
+		ctrlCtr.WithLabelValues("graft").Add(float64(len(rpc.Control.Graft)))
+		ctrlCtr.WithLabelValues("prune").Add(float64(len(rpc.Control.Prune)))
```
kirugan, line 96: there is also `idontwant` or we ommitted it on purpose?

kirugan (changes requested):
Nothing critical, left few comments

p2p/metrics.go
```diff
@@ -1,18 +1,111 @@
...
+		}, []string{"topic"}),
+		PSThrottlePeer: promauto.NewCounterVec(prometheus.CounterOpts{
+			Name: "p2p_pubsub_throttle_total",
+			Help: "The number of times the peer has been throttled",
+		}, []string{"topic"}),
+		PSRPCRecv: promauto.NewCounterVec(prometheus.CounterOpts{
+			Name: "p2p_pubsub_rpc_recv_total",
+			Help: "The number of messages received via rpc for a particular control message",
+		}, []string{"control_message"}),
+		PSRPCSubRecv: promauto.NewCounter(prometheus.CounterOpts{
```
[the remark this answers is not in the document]
weiihann: For this metric as well as `PSRPCSubDrop` and `PSRPCSubSent`, we are only interested in knowing the total number of subscription messages sent, drop and received.

p2p/gossip_tracer.go
weiihann: If we use pointer receiver, then it would not implement the `RawTracer` interface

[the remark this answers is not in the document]
weiihann: We are just implementing the `RawTracer` interface here

p2p/gossip_tracer.go
weiihann: Oh I missed it, let me add it

p2p/metrics.go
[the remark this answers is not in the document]
kirugan: I see, ok

p2p/gossip_tracer.go
```diff
@@ -0,0 +1,218 @@
...
+		ctrlCtr.WithLabelValues("idontwant").Add(float64(len(rpc.Control.Idontwant)))
+	}
+
+	for _, msg := range rpc.Publish {
+		pubCtr.WithLabelValues(*msg.Topic).Inc()
+	}
+}
+
+func agentNameFromPeerID(pid peer.ID, store peerstore.Peerstore) string {
```
kirugan, line 202: since we use only `.Get(...)` can we change type from `peerstore.Peerstore` to `PeerMetadata`

p2p/gossip_tracer.go
```diff
@@ -0,0 +1,218 @@
...
+	"github.com/libp2p/go-libp2p/core/host"
+	"github.com/libp2p/go-libp2p/core/peer"
+	"github.com/libp2p/go-libp2p/core/peerstore"
+	"github.com/libp2p/go-libp2p/core/protocol"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
+var (
+	_                  = pubsub.RawTracer(&gossipTracer{})
```
kirugan, line 16: nit: change it to`_ pubsub.RawTracer = (*gossipTracer)(nil)` ?

p2p/gossip_tracer.go
```diff
@@ -0,0 +1,218 @@
...
+func (g *gossipTracer) Prune(p peer.ID, topic string) {
+	g.psTopicsPrune.WithLabelValues(topic).Inc()
+}
+
+func (g *gossipTracer) DeliverMessage(msg *pubsub.Message) {
+	g.psMsgDeliver.WithLabelValues(*msg.Topic).Inc()
+}
+
+func (g *gossipTracer) ValidateMessage(msg *pubsub.Message) {
+	g.psMsgValidate.WithLabelValues(*msg.Topic).Inc()
```
kirugan, line 156: just double checking: `msg.Topic` is always non nil, right ?

kirugan (approved):

PR #2165, opened 2024-09-19
Optimize receiptCommitment function for parallel processing

### LINKED ISSUE #2164: Improve performance of receiptCommitment
Right now `receiptCommitment()` function doesn't exploit parallel execution as other commitment functions (like `eventCommitmentPoseidon`). Because of PR #2133 users may observe slower times of sync non 0.13.2 blocks than before if we keep it as is.

AnkushinDaniil:
This pull request optimizes the `receiptCommitment` function by introducing parallel processing. Instead of processing each receipt sequentially, the receipts are divided among multiple workers, improving performance.

Fixes #2164

### DIFF
--- a/core/receipt.go
+++ b/core/receipt.go
@@ -1,9 +1,13 @@
 package core
 
 import (
+	"runtime"
+	"sync"
+
 	"github.com/NethermindEth/juno/core/crypto"
 	"github.com/NethermindEth/juno/core/felt"
 	"github.com/NethermindEth/juno/core/trie"
+	"github.com/sourcegraph/conc/pool"
 )
 
 type GasConsumed struct {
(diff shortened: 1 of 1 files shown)


core/receipt.go
```diff
@@ -62,22 +66,40 @@ func messagesSentHash(messages []*L2ToL1Message) *felt.Felt {
...
+			startIdx := receiptIdx
+			endIdx := min(startIdx+receiptsPerWorker, len(receipts))
+			workerPool.Go(func() error {
+				for i, receipt := range receipts[startIdx:endIdx] {
+					receiptTrieKey := new(felt.Felt).SetUint64(uint64(receiptIdx + i))
+					receiptHash := receipt.hash()
+
+					trieMutex.Lock()
+					_, err := trie.Put(receiptTrieKey, receiptHash)
+					trieMutex.Unlock()
```
kirugan, line 85: I think current impl is not the most efficient one, we need to write benchmarks for that. You can add them in `receipt_pkg_test.go`. First for old impl, then for new one. After that I'll tell what else we can improve

core/receipt.go
```diff
@@ -62,12 +65,33 @@ func messagesSentHash(messages []*L2ToL1Message) *felt.Felt {
...
 	var commitment *felt.Felt
-
 	return commitment, trie.RunOnTempTriePoseidon(commitmentTrieHeight, func(trie *trie.Trie) error {
-		for i, receipt := range receipts {
-			receiptTrieKey := new(felt.Felt).SetUint64(uint64(i))
-			_, err := trie.Put(receiptTrieKey, receipt.hash())
-			if err != nil {
+		numWorkers := min(runtime.GOMAXPROCS(0), len(receipts))
+		results := make([]*felt.Felt, len(receipts))
+		wg := sync.WaitGroup{}
```
kirugan, line 71: let's rely on zero value: `var wg sync.WaitGroup`

core/receipt.go
```diff
@@ -62,12 +65,33 @@ func messagesSentHash(messages []*L2ToL1Message) *felt.Felt {
...
+				defer wg.Done()
+				for i := range jobs {
+					results[i] = receipts[i].hash()
+				}
+			}()
+		}
+
+		wg.Wait()
+
+		key := new(felt.Felt)
```
kirugan, line 91: just double checking: are you sure that `trie.Put(...)` don't mess with key pointer ?

AnkushinDaniil: I am sure, but I think it would be safer to change.

pnowosie (approved):
👍

kirugan (approved):

core/receipt_pkg_test.go
```diff
@@ -0,0 +1,98 @@
...
+	"slices"
+	"testing"
+
+	"github.com/NethermindEth/juno/core/felt"
+	"github.com/NethermindEth/juno/utils"
+	"github.com/stretchr/testify/require"
+)
+
+func BenchmarkReceiptCommitment(b *testing.B) {
+	fromHex := func(hex string) *felt.Felt {
```
rianhughes, line 13: nit: func BenchmarkReceiptCommitment(t*testing.B) {

core/receipt_pkg_test.go
```diff
@@ -0,0 +1,98 @@
...
+
+import (
+	"slices"
+	"testing"
+
+	"github.com/NethermindEth/juno/core/felt"
+	"github.com/NethermindEth/juno/utils"
+	"github.com/stretchr/testify/require"
+)
+
```
rianhughes, line 11: Can we add a small test where we compute and check the receipt commiement?

rianhughes (approved):

PR #2166, opened 2024-09-20
Add log encoding option for JSON logging.

mcdee:
This option allows selection of either CONSOLE (the default, and the current behaviour) or JSON for log output, allowing for easier parsing of output for structued log backends.


dan-catalano-vc:
this is a great addition. hope this is merged in soon.

curious why you chose to call it encoding, instead of format

dan-catalano-vc:
https://github.com/NethermindEth/juno/issues/2256

mcdee:
> curious why you chose to call it encoding, instead of format

Encoding is the term used internally in the logging.  Also, format can be the format for individual fields (e.g. timestamp format) rather than the general encoding of the log entry.

pnowosie (commented):
LGTM - however linter fixes required

mcdee:
@pnowosie should be fixed, but workflows need running to confirm.

pnowosie:
> @pnowosie should be fixed, but workflows need running to confirm.

Oops - right ;)

db/pebble/db.go
```diff
@@ -36,11 +36,11 @@ func New(path string) (db.DB, error) {
 	return newPebble(path, nil)
 }
 
-func NewWithOptions(path string, cacheSizeMB uint, maxOpenFiles int, colouredLogger bool) (db.DB, error) {
+func NewWithOptions(path string, cacheSizeMB uint, maxOpenFiles int, colouredLogger bool, logEncoding utils.LogEncoding) (db.DB, error) {
```
kirugan, line 39: I think it would be better to pass logger instead of passing two settings for it's initialization. So proposed signature is:
```
func NewWithOptions(path string, cacheSizeMB uint, maxOpenFiles int, log utils.Logger) (db.DB, error) {
```
from the caller side it would be:
```
} else {
                dbLogger, err := utils.NewZapLogger(...)
                // handling of err above
		database, err = pebble.NewWithOptions(cfg.DatabasePath, cfg.DBCacheSize, cfg.DBMaxHandles, cfg.Colour)
	}
```

utils/log.go
```diff
@@ -86,6 +86,65 @@ func (l *LogLevel) UnmarshalText(text []byte) error {
...
+
+// The following are necessary for Cobra and Viper, respectively, to unmarshal log level
+// CLI/config parameters properly.
+var (
+	_ pflag.Value              = (*LogEncoding)(nil)
+	_ encoding.TextUnmarshaler = (*LogEncoding)(nil)
+)
+
+const (
+	CONSOLE LogEncoding = iota
```
kirugan, line 104: Can we rename it to `Console` please? according to standard go naming convention for const

db/pebble/db.go
mcdee: Addressed.

utils/log.go
mcdee: That isn't part of this PR, `CONSOLE` and some other constants in `utils/log.go` weren't created with this commit.  I suggest a separate issue/PR if you want to address that.

mcdee:
Is there any interest in merging this?

PR #2167, opened 2024-09-21
Update native juno for sw/main rebase

PearsonWhite:
Corresponding Juno updates for NethermindEth/sequencer#23 which rebases on top of some gas calculation refactoring from sw/main.

These changes also rely on native contract executor changes from NethermindEth/juno#2156


rodrigo-pino:
Closing, it was meant as part of the development of Cairo Native integration

PR #2176, opened 2024-09-24
Refactor ProofNode to use interface

weiihann:
### Description
This PR refactors the `ProofNode` struct into `ProofNode` interface where `Binary` and `Edge` implements it.

### Rationale
Throughout the codebase, I noticed a similar pattern with `ProofNode`, that is:

```
if proofNode.Binary != nil {
  // ... do something
} else if proofNode.Edge != nil {
  // ... do something
}
```

This suggests that a `ProofNode` can be either a `Binary` node or an `Edge` node, but not both simultaneously. Therefore, `ProofNode` should be represented as an interface. This approach makes the code more intuitive and easier to read.

### DIFF
--- a/core/trie/proof.go
+++ b/core/trie/proof.go
@@ -12,7 +12,6 @@ var (
 	ErrChildHashNotFound = errors.New("can't determine the child hash from the parent and child")
 )
 
-// https://github.com/starknet-io/starknet-p2p-specs/blob/main/p2p/proto/snapshot.proto#L6
 type ProofNode interface {
 	Hash(hash hashFunc) *felt.Felt
 	Len() uint8
--- a/core/trie/proof.go
+++ b/core/trie/proof.go
@@ -153,7 +153,7 @@ func pathSplitOccurredCheck(mergedPath []ProofNode, nodeHashes map[felt.Felt]Pro
 				splitHappened = true
 			}
 		default:
-			return ErrUnknownProofNode
+			return fmt.Errorf("%w: %T", ErrUnknownProofNode, node)
 		}
 	}
 	return nil
@@ -451,22 +451,23 @@ func compressNode(idx int, proofNodes []ProofNode, hashF hashFunc) (int, uint8, 
 	}
 
 	child := proofNodes[idx+1]
-
+	_, isChildBinary := child.(*Binary)
+	isChildEdge := !isChildBinary
 	switch parent := parent.(type) {
 	case *Edge:
-		if _, ok := child.(*Binary); !ok {
+		if isChildEdge {
 			break
 		}
 		return 1, parent.Len(), nil
 	case *Binary:
-		if _, ok := child.(*Edge); !ok {
+		if isChildBinary {
 			break
 		}
 		childHash := child.Hash(hashF)
 		if parent.LeftHash.Equal(childHash) || parent.RightHash.Equal(childHash) {
 			return 1, child.Len(), nil
 		}
-		return 0, 0, errors.New("can't determine the child hash from the parent and child")
+		return 0, 0, ErrChildHashNotFound
 	}
 
 	return 0, 1, nil
@@ -513,7 +514,6 @@ func ProofToPath(proofNodes []ProofNode, leafKey *Key, hashF hashFunc) ([]Storag
 
 		// Set the key of the current node
 		compressParent, compressParentOffset, err := compressNode(i, proofNodes, hashF)
-		fmt.Printf("compressParent: %d, compressParentOffset: %d\n", compressParent, compressParentOffset)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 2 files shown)


core/trie/proof.go
```diff
@@ -139,19 +141,20 @@ func transformNode(tri *Trie, parentKey *Key, sNode StorageNode) (*Edge, *Binary
...
+			if leftExists && rightExists {
+				if splitHappened {
+					return errors.New("split happened more than once")
+				}
+				splitHappened = true
 			}
-
-			splitHappened = true
+		default:
+			return ErrUnknownProofNode
```
kirugan, line 156: can we pass what type it was? i.e. something like that `fmt.Errorf("%w: %T", ErrUnknownProofNode, node)`

core/trie/proof.go
```diff
@@ -438,27 +441,32 @@ func ensureMonotonicIncreasing(proofKeys [2]*Key, keys []*felt.Felt) error {
...
+		}
 		childHash := child.Hash(hashF)
-		if parent.Binary.LeftHash.Equal(childHash) || parent.Binary.RightHash.Equal(childHash) {
-			return 1, child.Edge.Path.len, nil
-		} else {
-			return 0, 0, errors.New("can't determine the child hash from the parent and child")
+		if parent.LeftHash.Equal(childHash) || parent.RightHash.Equal(childHash) {
+			return 1, child.Len(), nil
 		}
+		return 0, 0, errors.New("can't determine the child hash from the parent and child")
```
kirugan, line 469: I think you meant smth like this `return 0, 0, ErrChildHashNotFound` ?

core/trie/proof.go
```diff
@@ -505,6 +513,7 @@ func ProofToPath(proofNodes []ProofNode, leafKey *Key, hashF hashFunc) ([]Storag
 
 		// Set the key of the current node
 		compressParent, compressParentOffset, err := compressNode(i, proofNodes, hashF)
+		fmt.Printf("compressParent: %d, compressParentOffset: %d\n", compressParent, compressParentOffset)
```
kirugan, line 516: debug print ?

core/trie/proof.go
```diff
@@ -438,27 +441,32 @@ func ensureMonotonicIncreasing(proofKeys [2]*Key, keys []*felt.Felt) error {
...
-		return 1, parent.Edge.Path.len, nil
-	case parent.Binary != nil && child.Edge != nil:
+	switch parent := parent.(type) {
+	case *Edge:
+		if _, ok := child.(*Binary); !ok {
+			break
+		}
+		return 1, parent.Len(), nil
+	case *Binary:
+		if _, ok := child.(*Edge); !ok {
```
kirugan, line 462: I think you can move checking is it binary or edge before switch like that:
```
_, isChildBinary := child.(*Binary)
isChildEdge = !isChildBinary
```
and use them of just one variable later in switch cases

core/trie/proof.go
```diff
@@ -558,14 +568,19 @@ func skipNode(pNode ProofNode, pathNodes []StorageNode, hashF hashFunc) bool {
...
+
+	switch parent := parent.(type) {
+	case *Binary:
+		return parent.LeftHash, parent.RightHash, nil
+	case *Edge:
 		if parentInd+1 > len(proofNodes)-1 {
 			return nil, nil, errors.New("cant get hash of children from proof node, out of range")
 		}
-		parent = &proofNodes[parentInd+1]
+		return proofNodes[parentInd+1].(*Binary).LeftHash, proofNodes[parentInd+1].(*Binary).RightHash, nil
```
kirugan, line 580: maybe
```
parentBinary := proofNodes[parentInd+1].(*Binary)
return parentBinary.LeftHash, parentBinary.RightHash, nil
```
?

rianhughes (approved):

AnkushinDaniil (approved):

core/trie/proof.go
```diff
@@ -342,11 +345,11 @@ func VerifyProof(root *felt.Felt, key *Key, value *felt.Felt, proofs []ProofNode
...
 			}
 
-			if !proofNode.Edge.Path.Equal(subKey) {
+			if !proofNode.Path.Equal(subKey) {
 				return false
 			}
-			expectedHash = proofNode.Edge.Child
-			remainingPath.Truncate(251 - proofNode.Edge.Path.Len()) //nolint:mnd
+			expectedHash = proofNode.Child
+			remainingPath.Truncate(251 - proofNode.Path.Len()) //nolint:mnd
```
kirugan, line 352: should it be `globalTrieHeight` or `ContractSotrageTrieHeight` ?

kirugan: actually let's do it in another PR

kirugan (approved):

PR #2182, opened 2024-09-26
Refactor transaction into batch and snapshot

weiihann:
### Description
This PR removes the Transaction struct and added `batch` and `snapshot` which implements the `Transaction` interface. `batch` is used for read-write operations while `snapshot` is used for read-only operations.

### Rationale
A transaction can only be a `batch` or a `snapshot`, not both at the same time. The following pattern may be unnecessary:
```
if t.batch != nil {
  ...do something with batch
} else if t.snapshot != nil {
  ...do something with snapshot 
}
```

It's clearer to just separate `batch` and `snapshot` into different structs and use them accordingly.


db/db.go
```diff
@@ -77,9 +77,6 @@ type Transaction interface {
 	// Get fetches the value for the given key, should return ErrKeyNotFound if key is not present
 	// Caller should not assume that the slice would stay valid after the call to cb
 	Get(key []byte, cb func([]byte) error) error
-
-	// Impl returns the underlying transaction object
-	Impl() any
```
kirugan, line 82: @IronGauntlets do you remember why this method was introduced in the first place? how is this different from `db.(customType)` ?

db/pebble/batch.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+// Commit : see db.Transaction.Commit
+func (b *batch) Commit() error {
+	start := time.Now()
+	defer func() { b.listener.OnCommit(time.Since(start)) }()
+
+	if b.batch == nil {
+		return ErrDiscardedTransaction
+	}
```
kirugan, line 50: let's move this check up to beginning of this method (same as in `Discard`)

kirugan: time difference that is passed to `b.listener.OnCommit` is negligible

kirugan (approved):

db/db.go
IronGauntlets: I wasn't too sure about this when Omer implemented it. If it is not used anywhere we can delete it.

db/pebble/db.go
```diff
@@ -18,6 +19,10 @@ const (
 	minCacheSizeMB = 8
 )
 
+var (
+	ErrDiscardedTransaction = errors.New("discarded txn")
+	ErrReadOnlyTransaction  = errors.New("read-only transaction")
```
IronGauntlets, line 24: Why the disparity betweeen `txn` vs `transaction`?

db/pebble/common.go
```diff
@@ -0,0 +1,34 @@
...
+	defer listener.OnIO(false, time.Since(start)) //nolint:govet
+	if err != nil {
+		if errors.Is(err, pebble.ErrNotFound) {
+			return db.ErrKeyNotFound
+		}
+		return err
+	}
+
+	return utils.RunAndWrapOnError(closer.Close, cb(val))
+}
```
IronGauntlets, line 34: Would a generic be a better idea?

IronGauntlets (approved):
In general, it looks good to me. I have let some comments

db/pebble/common.go
kirugan: Tbh I don't think generic will bring a lot of value. The code will become a little bit harder to read

Issue #2178, opened 2024-09-25
RPC08 - starknet_getMessagesStatus


rianhughes:
[starknet_getMessagesStatus](https://github.com/starkware-libs/starknet-specs/blob/v0.7.1/api/starknet_api_openrpc.json)

https://docs.starknet.io/architecture-and-concepts/network-architecture/messaging-mechanism/#l1-l2-messages

see: https://github.com/eqlabs/pathfinder/issues/2182

Solution:
1. Listen to LogMessageToL2 logs from the core contract on L1. Compute the associated message hash, and store the eth transaction hash. Note that a single L1 transaction can induce multiple messages.
Populate db with:    `map[eth transaction hash][]message hash`
2. When syncing L2. Compute the message hash for l1 handler transactions. 
Populate db with:  `map[message hash]l1_handler_transaction_hash`

Overall this allows us to map eth_txn_hash to l1_handler_txn_hash, which we can then serve over RPC.

rianhughes:
L1 transaction hash is computed as 
```
keccak256(
    abi.encodePacked(
        uint256(FromAddress),
        ToAddress,
        Nonce,
        Selector,
        Payload.length,
        Payload
    )
);
```

PR #2184, opened 2024-09-27
Implement starknet_getMessageStatus

rianhughes:
closes https://github.com/NethermindEth/juno/issues/2178


blockchain/blockchain.go
```diff
@@ -622,6 +651,21 @@ func stateUpdateByHash(txn db.Transaction, hash *felt.Felt) (*core.StateUpdate,
...
+		if len(val) == 0 {
+			return db.ErrKeyNotFound
+		}
+		raw = val
+		return nil
+	})
+	if err != nil {
+		return nil, err
+	}
+	return new(felt.Felt).SetBytes(raw), nil
```
pnowosie, line 666: Are we sure it'd fit 251-bytes of Felt?

pnowosie (approved):
Just one question

blockchain/blockchain.go
rianhughes: Yup, the value is a felt (l1 handler transaction hash).

db/buckets.go
```diff
@@ -33,6 +33,7 @@ const (
 	BlockCommitments
 	Temporary // used temporarily for migrations
 	SchemaIntermediateState
+	L1HandlerTxnHash // maps l1 handler msg hash to l1 handler txn hash
```
kirugan, line 36: please run `make generate` it should update `buckets_enumer.go` file automatically + seems like we need migration

rpc/transaction.go
```diff
@@ -224,8 +224,9 @@ type Transaction struct {
 }
 
 type TransactionStatus struct {
-	Finality  TxnStatus          `json:"finality_status"`
-	Execution TxnExecutionStatus `json:"execution_status,omitempty"`
+	Finality      TxnStatus          `json:"finality_status"`
+	Execution     TxnExecutionStatus `json:"execution_status,omitempty"`
+	FailureReason string             `json:"failure_reason,omitempty"`
```
kirugan, line 229: is it true that we get this value by unmarshaling of some response or we should set it explicitly somewhere? I found `adaptTransactionStatus` function that set 2 out of 3 fields

rianhughes: Good point, the spec requires that [getTransactionStatus](https://github.com/starkware-libs/starknet-specs/blob/v0.8.0-rc0/api/starknet_api_openrpc.json#L225) should return the failure reason. Should I update this PR to reflect those changes, or would it be cleaner to do in a separate PR?

rianhughes: The `revert_error` is used for both rejected and reverted transactions

db/buckets.go
rianhughes: I've ran `make generate` to update the `buckets_enumer.go`. Just need to think about and handle the migration.

rianhughes: added a migration function to populate the new bucket for historical blocks that exist in the db

rpc/transaction.go
kirugan: Thanks, it's strange that we hasn't used `txStatus.RevertError` before

blockchain/blockchain.go
```diff
@@ -622,6 +651,21 @@ func stateUpdateByHash(txn db.Transaction, hash *felt.Felt) (*core.StateUpdate,
 	})
 }
 
+func l1MsgHashes(txn db.Transaction, l1HandlerMsgHash *common.Hash) (*felt.Felt, error) {
```
kirugan, line 654: By `Hashes` in name did you mean `common.Hash` and `l1HandlerTxnHash` ?

blockchain/blockchain_test.go
```diff
@@ -238,7 +239,13 @@ func TestStore(t *testing.T) {
 		got0Update, err := chain.StateUpdateByHash(block0.Hash)
 		require.NoError(t, err)
 		assert.Equal(t, stateUpdate0, got0Update)
+
+		nonExistentMsgHash := common.HexToHash("0xcoffeebabe")
```
kirugan, line 243: > 0xcoffeebabe

👍 😄

migration/migration.go
```diff
@@ -464,6 +465,29 @@ func calculateBlockCommitments(txn db.Transaction, network *utils.Network) error
 	return workerPool.Wait()
 }
 
+func calculateL1MsgHashes(txn db.Transaction, n *utils.Network) error {
```
kirugan, line 468: few ideas how to improve performance:
1. remove mutex
2. use fixed pool of goroutines with `workerPool.Go()` (if it's used in a loop then it constantly using some resources for creating new goroutine and shutting it down)

something like this:
```
workerPool := ...

head, err := blockchain.Head()
if err != nil {
    ... // in case ErrKeyNotFound - return nil
}
blocks := make(chan uint64, 1024) // actual capacity of buffered channel is actually not super important
go func() {
    for block := range head.Number {
         blocks <- block
    }
    close(blocks)
}()
for range numOfWorkers {
    workerPool.Go(func() error {
        for blockNumber := range blocks {
            ... := blockchain.BlockByNumber(txn, blockNumber)
            ... := blockchain.StoreL1HandlerMsgHashes(txn, block)
        }
    })
}
```

kirugan: Also we can significantly improve it if we fetch only necessary data (i.e. transactions). For that we need:
1. receive list of transactions in `StoreL1HandlerMsgHashes` instead of whole block
2. make `transactionsByBlockNumber` public function

blockchain/blockchain.go
```diff
@@ -622,6 +651,21 @@ func stateUpdateByHash(txn db.Transaction, hash *felt.Felt) (*core.StateUpdate,
...
+	var raw []byte
+	err := txn.Get(db.L1HandlerTxnHashByMsgHash.Key(l1HandlerMsgHash.Bytes()), func(val []byte) error {
+		if len(val) == 0 {
```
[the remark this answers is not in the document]
rianhughes: In principle it shouldn't (I don't think we should ever store a key-value pair where the value is empty, just wanted to be cautious).

blockchain/blockchain.go
rianhughes: I'll update the name to `l1HandlerTxnHashByMsgHash` so that it's hopefully a bit more clear. It takes in an l1 handler msg hash (common.Hash) and returns the corresponding l1 handler transaction hash (felt)

blockchain/blockchain_test.go
rianhughes: 🫡

migration/migration.go
rianhughes: I wonder if we could also apply these changes to `calculateBlockCommitments`?

rianhughes: I updated both `calculateL1MsgHashes` and `calculateBlockCommitments`

rianhughes:
note: `processBlocks()` isn't updating the db..

blockchain/blockchain.go
```diff
@@ -34,6 +35,7 @@ type Reader interface {
 	Receipt(hash *felt.Felt) (receipt *core.TransactionReceipt, blockHash *felt.Felt, blockNumber uint64, err error)
 	StateUpdateByNumber(number uint64) (update *core.StateUpdate, err error)
 	StateUpdateByHash(hash *felt.Felt) (update *core.StateUpdate, err error)
+	L1HandlerTxnHash(msgHash *common.Hash) (l1HandlerTxnHashes *felt.Felt, err error)
```
kirugan, line 38: `l1HandlerTxnHashes` - there is only one hash right? should it be `l1HandlerTxnHash` ?

grpc/gen/kv.pb.go
kirugan, line 1: can you please remove grpc changes from this PR, I suppose they are modified because of `make generate`

rpc/l1.go
```diff
@@ -0,0 +1,112 @@
...
+
+var ErrL1ClientNotFound = jsonrpc.Err(jsonrpc.InternalError, fmt.Errorf("L1 client not found, cannot serve starknet_getMessage"))
+
+type LogMessageToL2 struct {
+	FromAddress *common.Address
+	ToAddress   *common.Address
+	Nonce       *big.Int
+	Selector    *big.Int
+	Payload     []*big.Int
+	Fee         *big.Int
```
kirugan, line 22: why they are not felt-s ?

rpc/l1.go
```diff
@@ -0,0 +1,112 @@
...
+func (h *Handler) messageToL2Logs(ctx context.Context, txHash *common.Hash) ([]*common.Hash, *jsonrpc.Error) {
+	if h.ethClient == nil {
+		return nil, ErrL1ClientNotFound
+	}
+
+	receipt, err := h.ethClient.TransactionReceipt(ctx, *txHash)
+	if err != nil {
+		return nil, ErrTxnHashNotFound
+	}
+	logMsgToL2SigHash := common.HexToHash("0xdb80dd488acf86d17c747445b0eabb5d57c541d3bd7b6b87af987858e5066b2b")
```
kirugan, line 90: if it's static maybe move it outside of this method?

rpc/l1.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+	tmp := common.BytesToHash(hash.Sum(nil))
+	return &tmp
+}
+
+type MsgStatus struct {
+	L1HandlerHash  *felt.Felt
+	FinalityStatus TxnStatus
+	FailureReason  string
+}
```
IronGauntlets, line 53: Do you think we need to have `json` tags here? The field names do not match the names required in the spec. Also the failure reason should be omitted if it is empty.

```suggestion
type MsgStatus struct {
	L1HandlerHash  *felt.Felt `json:"transaction_hash"`
	FinalityStatus TxnStatus  `json:"finality_status"`
	FailureReason  string     `json:"failure_reason,omitempty"`
}
```

node/node.go
```diff
@@ -199,6 +199,16 @@ func New(cfg *Config, version string) (*Node, error) { //nolint:gocyclo,funlen
 	}
 
 	rpcHandler := rpc.New(chain, syncReader, throttledVM, version, log).WithGateway(gatewayClient).WithFeeder(client)
+	if cfg.EthNode == "" {
+		log.Warnw(rpc.ErrL1ClientNotFound.Message)
```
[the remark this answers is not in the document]
rianhughes: Sure, I can just pass the message directly in that case. Don't think there's a need to create a dedicated variable for it

rpc/l1.go
rianhughes: The starknet contract ABI returns uint256s (big.Ints) and addresses (common.Address)
```
{
  "anonymous": false,
  "inputs": [
    {
      "indexed": true,
      "internalType": "address",
      "name": "fromAddress",
      "type": "address"
    },
    {
      "indexed": true,
      "internalType": "uint256",
      "name": "toAddress",
      "type": "uint256"
    },
    {
      "indexed": true,
      "internalType": "uint256",
      "name": "selector",
      "type": "uint256"
    },
    {
      "indexed": false,
      "internalType": "uint256[]",
      "name": "payload",
      "type": "uint256[]"
    },
    {
      "indexed": false,
      "internalType": "uint256",
      "name": "nonce",
      "type": "uint256"
    },
    {
      "indexed": false,
      "internalType": "uint256",
      "name": "fee",
      "type": "uint256"
    }
  ],
  "name": "LogMessageToL2",
  "type": "event"
}
```

rpc/l1.go
rianhughes: Yes, thank you!

blockchain/blockchain.go
rianhughes: Yup!

grpc/gen/kv.pb.go
rianhughes: Note to self: I still need to do this

blockchain/blockchain.go
```diff
@@ -115,12 +117,12 @@ func (b *Blockchain) Height() (uint64, error) {
...
 	return height, b.database.View(func(txn db.Transaction) error {
 		var err error
-		height, err = chainHeight(txn)
+		height, err = ChainHeight(txn)
 		return err
 	})
 }
 
-func chainHeight(txn db.Transaction) (uint64, error) {
+func ChainHeight(txn db.Transaction) (uint64, error) {
```
IronGauntlets, line 125: I don't think it is a good idea to make this function public because now the caller is expected to provide a transaction which can have unintended consequences on the state of the DB if we don't carefully think about who has access to the DB. 

When the blockchain struct was designed a conscious decision was made that the public API of the Blockchain would not require a transaction, which essentially meant that only the blockchain (and migration) was responsible for writing to the DB.

As far as I have seen this change is due to migration, I don't think migration necessitates making this function public, as migration code is temporary. Instead, we can copy-paste the code or look at other migrations to take inspiration. In general, migration requirements should have no impact on the design of the rest of the code base.

IronGauntlets: After further investigation, it looks like this is the case with other functions which were made public for migration reasons (which I had missed earlier) too. I think those functions should also be made private (in a different PR though)

blockchain/blockchain.go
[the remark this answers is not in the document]
IronGauntlets: Since you are in control over storing the hash isn't it safe you assume that you will not store an empty hash?

Such a check should exit in the store function, which then leads you to assume that there will be no key which has an empty value.

Also, `db.ErrKeyNotFound` is an incorrect error to return since the DB found the key but there was no value. Outside of the db package `ErrKeyNotFound` should only be used to check an error not return an error.

Also, it would be worth checking if the DB creates an entry if the value is nil. I believe some tests are dealing with this in the DB package.

IMO this `if` statement should be removed as it is superfluous code.

blockchain/blockchain.go
```diff
@@ -622,6 +651,21 @@ func stateUpdateByHash(txn db.Transaction, hash *felt.Felt) (*core.StateUpdate,
 	})
 }
 
+func l1HandlerTxnHashByMsgHash(txn db.Transaction, l1HandlerMsgHash *common.Hash) (*felt.Felt, error) {
+	var raw []byte
+	err := txn.Get(db.L1HandlerTxnHashByMsgHash.Key(l1HandlerMsgHash.Bytes()), func(val []byte) error {
+		if len(val) == 0 {
+			return db.ErrKeyNotFound
+		}
+		raw = val
```
IronGauntlets, line 660: This is just a shallow copy which can have unintended consequences, you should do a deep copy. If you notice whenever we fetch data from DB almost always there is an unmarshal function that serves the same purpose as a deep copy.

blockchain/blockchain.go
```diff
@@ -589,6 +605,19 @@ func blockByHash(txn db.Transaction, hash *felt.Felt) (*core.Block, error) {
 	})
 }
 
+func StoreL1HandlerMsgHashes(dbTxn db.Transaction, blockTxns []core.Transaction) error {
+	for _, txn := range blockTxns {
+		if l1Handler, ok := (txn).(*core.L1HandlerTransaction); ok {
+			l1HandlerTxnHashBytes := txn.Hash().Bytes()
+			err := dbTxn.Set(db.L1HandlerTxnHashByMsgHash.Key(l1Handler.MessageHash()), l1HandlerTxnHashBytes[:])
```
IronGauntlets, line 612: Do you know why the `MessageHash()` doesn't return a `*felt.Felt`?

Also, instead of the following:

```go
l1HandlerTxnHashBytes := txn.Hash().Bytes()
err := dbTxn.Set(db.L1HandlerTxnHashByMsgHash.Key(l1Handler.MessageHash()), l1HandlerTxnHashBytes[:])
```

I think it should be 

```go
err := dbTxn.Set(db.L1HandlerTxnHashByMsgHash.Key(l1Handler.MessageHash().Marshal()), tx.Hash().Marshal())
```

I am checking why MessageHash() is not a felt? I think we can make it a felt.

blockchain/blockchain.go
IronGauntlets: You can use the unmarshal function on felt added in PR: #2246

blockchain/blockchain.go
rianhughes: Are you okay to leave this as public for now then?

blockchain/blockchain.go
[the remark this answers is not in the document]
rianhughes: Actually, if the key-value doesn't exist `val` will have length 0, in which case we should return `ErrKeyNotFound` imo.

blockchain/blockchain.go
rianhughes: Thanks for catching this and the suggestion, I've used the above PR to make a hard-copy.

blockchain/blockchain.go
rianhughes: I'm not sure why MessageHash() doesn't return a felt, although it looks like there was never a need for it to. I think changing it is outside the scope of this PR though.

I implemented your suggestion though

node/node.go
```diff
@@ -199,6 +199,16 @@ func New(cfg *Config, version string) (*Node, error) { //nolint:gocyclo,funlen
...
+	if cfg.EthNode == "" {
+		log.Warnw("L1 client not found, cannot serve starknet_getMessage")
+	} else {
+		ethClient, err := l1.NewETHClient(cfg.EthNode)
+		if err != nil {
+			return nil, err
+		}
+		rpcHandler.WithETHClient(ethClient)
+	}
+
```
IronGauntlets, line 211: I think this code is somewhat being repeated on line 285, I think there is a better way to reduce the repetition.

rpc/handlers.go
```diff
@@ -78,6 +83,7 @@ type Handler struct {
 	feederClient  *feeder.Client
 	vm            vm.VM
 	log           utils.Logger
+	ethClient     l1.EthClient
```
IronGauntlets, line 86: I think this should be `l1.EthSubscriber`

rpc/l1.go
```diff
@@ -0,0 +1,114 @@
...
+	"math/big"
+
+	"github.com/NethermindEth/juno/core/felt"
+	"github.com/NethermindEth/juno/jsonrpc"
+	"github.com/ethereum/go-ethereum/common"
+	"golang.org/x/crypto/sha3"
+)
+
+var (
+	ErrL1ClientNotFound = jsonrpc.Err(jsonrpc.InternalError, fmt.Errorf("L1 client not found, cannot serve starknet_getMessage"))
```
IronGauntlets, line 15:
```suggestion
	ErrL1ClientNotFound = jsonrpc.Err(jsonrpc.InternalError, fmt.Errorf("11 client not found, cannot serve starknet_getMessage"))
```

IronGauntlets (changes requested):
I think there is some repetition around the ethereum client which can be removed.

With regards to migration, it is going to be quite large (@kirugan) and since this has to do with Hashing, it can potentially take quite some time (I would expect it to take several hours) but it will be faster than resyncing the whole chain.

I would prefer to sync from scratch and create a new snapshot so we can avoid adding this migration (if we have time). The rpc may be down if this is not planned well, as an upgrade with block juno until migration is complete which cannot stopped.

rpc/l1.go
```diff
@@ -0,0 +1,114 @@
...
+type LogMessageToL2 struct {
+	FromAddress *common.Address
+	ToAddress   *big.Int
+	Nonce       *big.Int
+	Selector    *big.Int
+	Payload     []*big.Int
+	Fee         *big.Int
+}
+
+func (l *LogMessageToL2) HashMessage() *common.Hash {
```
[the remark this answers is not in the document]
rianhughes: Nope, I've made it private

node/node.go
rianhughes: Yup, good point, I'll combine them

l1/eth_subscriber.go
```diff
@@ -41,12 +41,30 @@ func NewEthSubscriber(ethClientAddress string, coreContractAddress common.Addres
...
+	// TODO replace with our own client once we have one.
+	// Geth pulls in a lot of dependencies that we don't use.
+	client, err := rpc.DialContext(ctx, ethClientAddress)
+	if err != nil {
+		return nil, err
+	}
+	ethClient := ethclient.NewClient(client)
+	return ethClient, nil
+}
+
```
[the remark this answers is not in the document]
rianhughes: The RPC handler only needs access to the `TransactionReceipt` method and I thought it might have been neater to constrain `ethClient` using a new interface. But I updated the code to remove the EthClient interface, and grant the RPC handler access to `Subscriber` instead (no need to give it access to everything in `EthSubscriber` imo)

rpc/handlers.go
rianhughes: See the above comment (I updated it to use `l1.Subscriber` - no need to give the RPC handler access to everything in `l1.EthSubscriber` imo)

blockchain/blockchain.go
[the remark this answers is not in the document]
rianhughes: Yup, you're right! I've committed your changes

l1/eth_subscriber.go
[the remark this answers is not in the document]
IronGauntlets: I think you are right we should not give access to more than necessary but it is not the correct place to do this. 

You can achieve the same behaviour by moving the `EthClient` to the rpc's package and in the node package pass the subscriber since the subscriber will implement the `EthClient` interface.

IronGauntlets (approved):

rianhughes:
The above changes LGTM

kirugan (approved):

PR #2185, opened 2024-09-29
Initial implementation of Tendermint

IronGauntlets:
All of the Tendermint algorithm has been implemented.

Outstanding tasks are described in various "todos". 

Once sufficient tests have been added I will split the PR into multiple commits and it will be ready to merge.


consensus/tendermint/messages.go
```diff
@@ -0,0 +1,126 @@
...
+
+type Message[V Hashable[H], H Hash, A Addr] interface {
+	Proposal[V, H, A] | Prevote[H, A] | Precommit[H, A]
+}
+
+type Proposal[V Hashable[H], H Hash, A Addr] struct {
+	Height     uint
+	Round      uint
+	ValidRound *uint
+	Value      *V
```
rianhughes, line 20: V would be the new block?

consensus/tendermint/messages.go
```diff
@@ -0,0 +1,126 @@
...
+
+func newMessages[V Hashable[H], H Hash, A Addr]() messages[V, H, A] {
+	return messages[V, H, A]{
+		proposals:  make(map[uint]map[uint]map[A][]Proposal[V, H, A]),
+		prevotes:   make(map[uint]map[uint]map[A][]Prevote[H, A]),
+		precommits: make(map[uint]map[uint]map[A][]Precommit[H, A]),
+	}
+}
+
+// Todo: ensure duplicated messages are ignored.
```
rianhughes, line 63: What’s the difference between a duplicate message and a malicious equivocating message?

consensus/tendermint/messages.go
```diff
@@ -0,0 +1,126 @@
...
+	if _, ok := m.proposals[p.Height][p.Round]; !ok {
+		m.proposals[p.Height][p.Round] = make(map[A][]Proposal[V, H, A])
+	}
+
+	sendersProposals, ok := m.proposals[p.Height][p.Round][p.Sender]
+	if !ok {
+		sendersProposals = []Proposal[V, H, A]{}
+	}
+
+	m.proposals[p.Height][p.Round][p.Sender] = append(sendersProposals, p)
```
rianhughes, line 78: I guess we don't need to worry about this array being spammed by a malicious node? Does slashing depend on the number of equivocating messages / cap out at a certain point?

consensus/tendermint/tendermint.go
```diff
@@ -0,0 +1,637 @@
...
+	prevote
+	precommit
+)
+
+type timeoutFn func(round uint) time.Duration
+
+type Addr interface {
+	// Ethereum Addresses are 20 bytes
+	~[20]byte | felt.Felt
+}
```
rianhughes, line 26: Do you think it would be worth distinguishing between Ethereum and Starknet addresses / felts?

consensus/tendermint/tendermint.go
```diff
@@ -0,0 +1,637 @@
...
+type Hash interface {
+	~[32]byte | felt.Felt
+}
+
+// Hashable's Hash() is used as ID()
+type Hashable[H Hash] interface {
+	Hash() H
+}
+
+type Application[V Hashable[H], H Hash] interface {
```
rianhughes, line 37: The notes mentioned an additional ID() method. Is it no longer needed?

// Id() returns the id of the value which is a unique identifier of  // the value being consider for the current height. The votes include // the id of the value not the value itself
Id(value any) any

consensus/tendermint/tendermint.go
```diff
@@ -0,0 +1,637 @@
...
+	// Valid returns true if the provided value is valid according to the application context.
+	Valid(V) bool
+}
+
+type Blockchain[V Hashable[H], H Hash, A Addr] interface {
+	// Height return the current blockchain height
+	Height() uint
+
+	// Commit is called by Tendermint when a block has been decided on and can be committed to the DB.
+	Commit(uint, V, []Precommit[H, A])
```
rianhughes, line 50: What happens if Commit() fails / generates an error? Do we just log it / panic?

consensus/tendermint/tendermint.go
```diff
@@ -0,0 +1,637 @@
...
+
+	// Valid returns true if the provided value is valid according to the application context.
+	Valid(V) bool
+}
+
+type Blockchain[V Hashable[H], H Hash, A Addr] interface {
+	// Height return the current blockchain height
+	Height() uint
+
+	// Commit is called by Tendermint when a block has been decided on and can be committed to the DB.
```
rianhughes, line 49: Given Commit constrains Value to be a block, do you think it's worth updating the V / Value notation to be B / Block / *core.Block? Just for readability

consensus/tendermint/tendermint.go
```diff
@@ -0,0 +1,637 @@
...
+	Commit(uint, V, []Precommit[H, A])
+}
+
+type Validators[A Addr] interface {
```
rianhughes, line 53: Do you think we'll need a function to return all the validators for a given height? Or can we retrieve this information from somewhere else?

consensus/tendermint/tendermint.go
```diff
@@ -0,0 +1,637 @@
...
+
+	// ValidatorVotingPower returns the voting power of the a single validator. This is also required to implement
+	// various thresholds. The assumption is that a single validator cannot have voting power more than f.
+	ValidatorVotingPower(validatorAddr A) uint
+
+	// Proposer returns the proposer of the current round and height.
+	Proposer(height, round uint) A
+}
+
+type Slasher[M Message[V, H, A], V Hashable[H], H Hash, A Addr] interface {
```
rianhughes, line 65: Any validator (or entity that has the relevant messages etc) can call the L2 contract?

consensus/tendermint/tendermint.go
```diff
@@ -0,0 +1,637 @@
...
+	// Equivocation informs the slasher that a validator has sent conflicting messages. Thus it can decide whether to
+	// slash the validator and by how much.
+	Equivocation(msgs ...M)
+}
+
+type Listener[M Message[V, H, A], V Hashable[H], H Hash, A Addr] interface {
+	// Listen would return consensus messages to Tendermint which are set // by the validator set.
```
rianhughes, line 72: typo

consensus/tendermint/tendermint.go
```diff
@@ -0,0 +1,637 @@
...
+	Proposer(height, round uint) A
+}
+
+type Slasher[M Message[V, H, A], V Hashable[H], H Hash, A Addr] interface {
+	// Equivocation informs the slasher that a validator has sent conflicting messages. Thus it can decide whether to
+	// slash the validator and by how much.
+	Equivocation(msgs ...M)
+}
+
+type Listener[M Message[V, H, A], V Hashable[H], H Hash, A Addr] interface {
```
rianhughes, line 71: So consensus messages come in over P2P, undergo some checks etc, and then are passed into the Tendermint pkg via Listener?

consensus/tendermint/tendermint.go
```diff
@@ -0,0 +1,637 @@
...
+			case p := <-t.listeners.ProposalListener.Listen():
+				if p.Height < t.state.height {
+					continue
+				}
+
+				if p.Height > t.state.height { //nolint:staticcheck
+					continue
+				}
+
+				if p.Round > t.state.round {
```
rianhughes, line 190: Can the proposers round also be behind ours? Like with the height above?

consensus/tendermint/tendermint.go
```diff
@@ -0,0 +1,637 @@
...
+				proposalFromProposer := p.Sender == t.validators.Proposer(p.Height, p.Round)
+				vr := p.ValidRound
+
+				if validProposal {
+					// Add the proposal to the message set even if the sender is not the proposer,
+					// this is because of slahsing purposes
+					t.messages.addProposal(p)
+				}
+
+				_, prevotesForHR, precommitsForHR := t.messages.allMessages(p.Height, p.Round)
```
rianhughes, line 211: How can there be prevotes and precommits at the proposal stage?? I thought the flow is prosoal -> prevote -> precommit..

rianhughes: Ah, I didn't realise that proposers could use the prevotes (and precommits?) of previous rounds to help speed up the flow in the current round.

consensus/tendermint/messages.go
```diff
@@ -0,0 +1,126 @@
...
+//  of the validator set and would need access to the blockchain which may not be a good idea.
+
+type Message[V Hashable[H], H Hash, A Addr] interface {
+	Proposal[V, H, A] | Prevote[H, A] | Precommit[H, A]
+}
+
+type Proposal[V Hashable[H], H Hash, A Addr] struct {
+	Height     uint
+	Round      uint
+	ValidRound *uint
```
rianhughes, line 19: Nit: mybe this is too early for these types of comments, but it might be good to make a comment that nil means -1 or the round if not valid. Same with Value.

consensus/tendermint/tendermint.go
```diff
@@ -0,0 +1,637 @@
...
+					t.line55(p.Round)
+				}
+
+				// The below shouldn't panic because it is expected Proposal is well-formed. However,
+				// there need to be a way to distinguish between nil and zero value.
+				// This is expected to be handled by the p2p layer.
+				vID := (*p.Value).Hash()
+				fmt.Println("got proposal with valueID and validRound", vID, p.ValidRound)
+				validProposal := t.application.Valid(*p.Value)
+				proposalFromProposer := p.Sender == t.validators.Proposer(p.Height, p.Round)
```
rianhughes, line 202: nit: we could continue if proposalFromProposer==false right? Might just help simplify the if conditions later on

consensus/tendermint/tendermint.go
```diff
@@ -0,0 +1,637 @@
...
+
+					for addr, valPrevotes := range prevotesForHR {
+						for _, p := range valPrevotes {
+							if *p.ID == vID {
+								vals = append(vals, addr)
+							}
+						}
+					}
+
+				}
```
rianhughes, line 361: What are these lines doing??

consensus/tendermint/tendermint.go
```diff
@@ -0,0 +1,637 @@
...
+	for addr := range precommits {
+		vals[addr] = struct{}{}
+	}
+
+	if t.validatorSetVotingPower(slices.Collect(maps.Keys(vals))) > f(t.validators.TotalVotingPower(t.state.height)) {
+		t.startRound(futureR)
+	}
+}
+
+func (t *Tendermint[V, H, A]) validatorSetVotingPower(vals []A) uint {
```
rianhughes, line 614: nit: rename something like `validatorsTotalVotingPower`, I keep reading this as a Set function. Maybe it's just me, so feel free to ignore :)

consensus/tendermint/tendermint.go
```diff
@@ -0,0 +1,637 @@
...
+						proposal   Proposal[V, H, A]
+						precommits []Precommit[H, A]
+						vals       []A
+					)
+
+					for _, prop := range proposalsForHR[t.validators.Proposer(p.Height, p.Round)] {
+						if (*prop.Value).Hash() == *p.ID {
+							proposal = prop
+						}
+					}
```
rianhughes, line 427: Maybe I'm missing something here, but can we break early?

consensus/tendermint/tendermint.go
```diff
@@ -0,0 +1,637 @@
...
+	t.state.round = r
+	t.state.step = propose
+
+	t.state.line34Executed = false
+	t.state.line36Executed = false
+	t.state.line47Executed = false
+
+	if p := t.validators.Proposer(t.state.height, r); p == t.nodeAddr {
+		var proposalValue *V
+		if t.state.validValue != nil {
```
rianhughes, line 505: Is state.validValue not always nil? I don't see where it's set

consensus/tendermint/messages.go
IronGauntlets: Equivocated messages are when a validator sends conflicting messages for example a validator can not send both prevote(v) and prevote(nil) for the same height and round. Duplicated messages are when a message is received multiple times this can happen due to how gossiping. Therefore, a prevote(v) for a height and round can be received multiple times.

consensus/tendermint/messages.go
IronGauntlets: We need to limit how many equivocated messages are stored as this is an attack vector. However, since the slashing conditions are unclear, I have not yet put a limit on them.

I think it is a design decision whether to slash more if a higher number of equivocated messages have been received. Sending even one should have a large consequence as slashing is the only incentive present to try to limit the number of faulty nodes to less than F.

consensus/tendermint/tendermint.go
IronGauntlets: If it is just through a wrapper than no because we should aim to represent type as close to the underlying type.

consensus/tendermint/tendermint.go
IronGauntlets: In the blockchain context ID is the Hash which is why `V` is Hashable, therefore, there is no need to have ID as a function on the interface, rather it is on the Value itself which makes more sense.

When I originally wrote the interfaces some things regarding go generics weren't clear to me which is why they were changed later on.

consensus/tendermint/tendermint.go
IronGauntlets: This is something I was hoping to decide later on. However, the idea is that we tell the caller that a consensus has been reached for the current height. After that, it is up to the caller what they want to do with that information.

Then the consensus algorithm optimistically moves to the next height but if it turns out for some reason `Height()` has not been incremented then the caller should decide what to do. The caller will be the synchroniser.

consensus/tendermint/tendermint.go
IronGauntlets: So far I didn't need the full set of validators that why I didn't add it. If for some reason we need it we can extend the interface. We should keep our interfaces as small as possible.

consensus/tendermint/tendermint.go
IronGauntlets: Yes, note you don't need a quorum of validators because it is assumed that signatures cannot be forged. Therefore, even one validator sending 2 equivocated messages is sufficient for slashing.

consensus/tendermint/tendermint.go
IronGauntlets: yeah, some basic checks of being well-formed converted from protocol buffers to go types.

consensus/tendermint/tendermint.go
```diff
@@ -0,0 +1,637 @@
...
+	// slash the validator and by how much.
+	Equivocation(msgs ...M)
+}
+
+type Listener[M Message[V, H, A], V Hashable[H], H Hash, A Addr] interface {
+	// Listen would return consensus messages to Tendermint which are set // by the validator set.
+	Listen() <-chan M
+}
+
+type Broadcaster[M Message[V, H, A], V Hashable[H], H Hash, A Addr] interface {
```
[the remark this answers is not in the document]
IronGauntlets: In libp2p there is publish/subscribe mechanism which allows people to publish messages on a particular topic and nodes which subscribe to that topic will receive messages which are published on that topic.

Therefore yes, a non-validator can receive consensus messages if they want.

consensus/tendermint/tendermint.go
IronGauntlets: Technically, yes but it is an extremely unlikely scenario.

 At any moment in time, you need 2/3 of the network to make progress. If a node falls behind it means that node is not required to form a quorum. If this does happen, there are mechanisms in place to help the proposer catch up if that takes too long then timeout will eventually change the round which changes the proposer thus allowing the network to make progress.

consensus/tendermint/messages.go
IronGauntlets: I will remove them when the PR is ready, it was just present so that I don't forget why I made a certain decision and what the considerations were for that decision.

Not sure what you mean by nil means that no round reached a quorum. A nil validRound doesn't mean that.

consensus/tendermint/tendermint.go
IronGauntlets: These lines are fetching all the validators which prevote for `vID`, so that I check if there is a quorum.

consensus/tendermint/tendermint.go
IronGauntlets: Its personal preference, I will change the name.

consensus/tendermint/tendermint.go
IronGauntlets: The code has been moved/changed so I am not following, could you please make the comment again?

consensus/tendermint/tendermint.go
IronGauntlets: Please see the latest changes, it is set when upon condition on line 36 becomes true.

consensus/tendermint/propose_test.go
```diff
@@ -0,0 +1,536 @@
...
+			Vote: Vote[felt.Felt, felt.Felt]{
+				Height: expectedHeight,
+				Round:  rPrime,
+				ID:     utils.Ptr(round4Value.Hash()),
+				Sender: *val3,
+			},
+		}
+
+		algo.messages.addPrevote(val3Prevote)
+		prevoteListener := listeners.PrecommitListener.(*senderAndReceiver[Precommit[felt.Felt, felt.Felt], value,
```
rianhughes, line 149: typo. Also, do you think we could collapse the above three tests into a single test? It seems most of the code is repeated

IronGauntlets: I was trying to make the tests independent. If we start sharing variables, it becomes difficult to debug because of side effects. I have variables that don't change at the top of the test.

consensus/tendermint/tendermint.go
```diff
@@ -0,0 +1,431 @@
...
+	tm.Timer = time.AfterFunc(duration, func() {
+		select {
+		case <-t.quit:
+		case t.timeoutsCh <- tm:
+		}
+	})
+	t.scheduledTms = append(t.scheduledTms, tm)
+}
+
+func (t *Tendermint[_, H, A]) OnTimeoutPropose(h, r uint) {
```
rianhughes, line 343: So imagine that we start up a node, and it never recieves messages from other nodes. We start round 0, schedule the propose timeout, which eventually gets hit, and we end up here. Then we cast a `nil` prevote, set `step` to `prevote` and continue. We will end up back at `startRound(0)` again, reset everything, and the process will repeat indefinitely, right? We would effectively be casting `nil` prevotes with `round=0` forever? (I would have thought the round would increase..)

consensus/tendermint/tendermint_test.go
```diff
@@ -0,0 +1,282 @@
...
+				Sender: *nodeAddr,
+			},
+		}
+
+		prevoteBroadcaster := broadcasters.PrevoteBroadcaster.(*senderAndReceiver[Prevote[felt.Felt, felt.Felt], value,
+			felt.Felt, felt.Felt])
+
+		algo.Start()
+		time.Sleep(time.Millisecond)
+		algo.Stop()
```
rianhughes, line 230: Removing these two lines seems to have no effect. Which is weird, I would have thought there were needed, and don't understand how algo.Stop() is triggered..

consensus/tendermint/messages.go
rianhughes: Makes sense, thanks!

consensus/tendermint/tendermint.go
IronGauntlets: The scenario which you have describe would not happen because to start prevote timeout and precommit timeout the nodes need to receives at least 2/3 prevote/precommit (respectively) which could be for a particular value or nil, therefore, in the above scenario the node will remain in round 0 at prevote step until it receives more messages from other nodes.

consensus/tendermint/propose_test.go
```diff
@@ -0,0 +1,536 @@
...
+
+		assert.Equal(t, 0, len(algo.messages.proposals))
+		assert.Equal(t, 0, len(algo.messages.prevotes))
+		assert.Equal(t, 0, len(algo.messages.precommits))
+	})
+
+	t.Run("Line 22: receive a new proposal before timeout expiry", func(t *testing.T) {
+	})
+	//
+	//t.Run("Line 28: receive an old proposal before timeout expiry", func(t *testing.T) {
```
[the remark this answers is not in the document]
IronGauntlets: yes

consensus/tendermint/tendermint.go
```diff
@@ -0,0 +1,436 @@
...
+
+	lockedValue *V
+	lockedRound int
+	validValue  *V
+	validRound  int
+
+	// The following are round level variable therefore when a round changes they must be reset.
+	line34Executed bool
+	line36Executed bool
+	line47Executed bool
```
infrmtcs, line 160: I understand that we're strictly following the pseudocode, but I feel that this is not easy to read and memorize. Is it possible to name these differently?

consensus/tendermint/tendermint.go
```diff
@@ -0,0 +1,436 @@
...
+	return totalVotingPower
+}
+
+// Todo: add separate unit tests to check f and q thresholds.
+func f(totalVotingPower uint) uint {
+	// note: integer division automatically floors the result as it return the quotient.
+	return (totalVotingPower - 1) / 3
+}
+
+func q(totalVotingPower uint) uint {
```
infrmtcs, line 427: Shouldn't it be `2f + 1`?

consensus/tendermint/tendermint.go
```diff
@@ -0,0 +1,436 @@
...
+		timeoutsCh:       make(chan timeout),
+		proposalsCh:      make(chan Proposal[V, H, A]),
+		prevotesCh:       make(chan Prevote[H, A]),
+		precommitsCh:     make(chan Precommit[H, A]),
+		quit:             make(chan struct{}),
+	}
+}
+
+func (t *Tendermint[V, H, A]) Start() {
+	t.wg.Add(1)
```
infrmtcs, line 194: I'm wondering if we can remove wait group here. By doing so, we can let the caller decides how to run the algorithm.

consensus/tendermint/tendermint.go
```diff
@@ -0,0 +1,436 @@
...
+	case precommit:
+		return "precommit"
+	default:
+		return "unknown"
+	}
+}
+
+const (
+	maxFutureHeight = uint(5)
+	maxFutureRound  = uint(5)
```
infrmtcs, line 35: I think these should be configurable, probably inside the `Tendermint` struct

consensus/tendermint/messages.go
```diff
@@ -0,0 +1,110 @@
...
+	Sender A
+}
+
+type Prevote[H Hash, A Addr] struct {
+	Vote[H, A]
+}
+
+type Precommit[H Hash, A Addr] struct {
+	Vote[H, A]
+}
```
infrmtcs, line 31: Unless we're adding more fields to this struct, I think this is cheaper:
```suggestion
type Prevote[H Hash, A Addr] Vote[H, A]

type Precommit[H Hash, A Addr] Vote[H, A]
```

rodrigo-pino: I agree, I think it is better in sense of readability. I am not sure there would be performance improvements (@infrmtcs do you mean by cheaper that it was going to be more performant?)

consensus/tendermint/tendermint.go
IronGauntlets: That is not exactly the case. Please read the Threshold section of Juno's Tendermint implementation.

consensus/tendermint/tendermint.go
IronGauntlets: I don't think I follow.

The `Start()` is not blocking, I add to the wait group so that when the user calls `Stop()` then the go routine that is started in the `Start()` returns before `Stop()` returns.

consensus/tendermint/tendermint.go
IronGauntlets: I don't think we should as this can lead to an attack vector, so I didn't want to make it configurable for the time being.

consensus/tendermint/messages.go
```diff
@@ -0,0 +1,110 @@
...
+// messages keep tracks of all the proposals, prevotes, precommits by creating a map structure as follows:
+// height->round->address->[]Message
+
+// Todo: would the following representation of message be better:
+//
+//	  height -> round -> address -> ID -> Message
+//	How would we keep track of nil votes? In golan map key cannot be nil.
+//	It is not easy to calculate a zero value when dealing with generics.
+type messages[V Hashable[H], H Hash, A Addr] struct {
+	proposals  map[uint]map[uint]map[A][]Proposal[V, H, A]
```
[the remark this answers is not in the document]
IronGauntlets: Decided to go with your solution

consensus/tendermint/messages.go
IronGauntlets: done

consensus/tendermint/tendermint.go
infrmtcs: By configurable, I didn't mean the node user can configure it via command flag. What I'm understanding here is that the limit of future height and future round is not hardcoded in the tendermint algorithm, but by our application level. Then even if we have this and disallow the node to customize this, it should be on other level.

consensus/tendermint/tendermint_test.go
```diff
@@ -0,0 +1,277 @@
...
+		{15982, 10655, 5327},
+		{301234, 200823, 100411},
+		{301235, 200824, 100411},
+		{301236, 200824, 100411},
+	}
+
+	for _, test := range tests {
+		assert.Equal(t, test.q, q(test.n))
+		assert.Equal(t, test.f, f(test.n))
+	}
```
infrmtcs, line 273: I think what you'd want to assert here for q function is
- For correctness: 2 * q(test.n) > test.n + f(test.n)
- For efficiency: 2 * (q(test.n) - 1) <= test.n + f(test.n)

consensus/tendermint/propose.go
```diff
@@ -0,0 +1,246 @@
...
+
+	if p.R < t.state.r {
+		// Except line 49 all other upon condition which refer to the proposals expect to be acted upon
+		// when the current round is equal to the proposal's round.
+		return
+	}
+
+	t.line22(vr, proposalFromProposer, validProposal, vID)
+	t.line28WhenProposalIsReceived(p, vr, proposalFromProposer, vID, validProposal)
+	t.line36WhenProposalIsReceived(p, validProposal, proposalFromProposer, prevotesForHR, vID)
```
infrmtcs, line 64: I think we can also have better naming for these lines

consensus/tendermint/tendermint.go
infrmtcs: I'd proposed using a simplified version of this:
```
	return (totalVotingPower + f(totalVotingPower)) / 2 + 1
```

infrmtcs: This is based on the original assumption 2Q > N +f

consensus/tendermint/tendermint.go
```diff
@@ -0,0 +1,434 @@
...
+			Sender:     t.nodeAddr,
+		}
+
+		t.messages.addProposal(proposalMessage)
+		t.broadcasters.ProposalBroadcaster.Broadcast(proposalMessage)
+	} else {
+		t.scheduleTimeout(t.timeoutPropose(r), propose, t.state.h, t.state.r)
+	}
+
+	go t.processFutureMessages(t.state.h, t.state.r)
```
infrmtcs, line 283: Is there any reason why we concurrency here? The algorithm is mostly sequential, so I think we can avoid this mutex and 3 channels for future by moving the events out of `futureMessages`, maybe to a field instead of pushing them to 3 channels. That would avoid a mutex, a concurrent goroutine and context switching over 3 more channels.

consensus/tendermint/tendermint.go
```diff
@@ -0,0 +1,434 @@
...
+	PrevoteBroadcaster   Broadcaster[Prevote[H, A], V, H, A]
+	PrecommitBroadcaster Broadcaster[Precommit[H, A], V, H, A]
+}
+
+type Tendermint[V Hashable[H], H Hash, A Addr] struct {
+	nodeAddr A
+
+	state state[V, H] // Todo: Does state need to be protected?
+
+	messages       messages[V, H, A]
```
infrmtcs, line 123: I'm understanding that currently we're iterating over all messages to calculate aggregates. I think this can be more performant by storing only the aggregates.

consensus/tendermint/messages.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+// messages keep tracks of all the proposals, prevotes, precommits by creating a map structure as follows:
+// height->round->address->[]Message
+
+// Todo: would the following representation of message be better:
+//
+//	  height -> round -> address -> ID -> Message
+//	How would we keep track of nil votes? In golan map key cannot be nil.
+//	It is not easy to calculate a zero value when dealing with generics.
```
infrmtcs, line 45: I think it's possible with
```
func ZeroValue[T any]() T {
	var zero T
	return zero
}
```
Although it's not necessary a good way, depending whether zero value is not a valid value.

consensus/tendermint/tendermint.go
infrmtcs: As far as I understand, we can store:
- Proposed value by (height, round)
- Prevote counter and Precommit counter by (height, round) and by (height, round, ID)
Each counter consists of a total count and set of nodes sending it. We can even improve this part by replace Address as key by node index in current instance and use bit array instead of set.

consensus/tendermint/tendermint.go
IronGauntlets: I don't think it is worth doing it right now, since I am not sure how it will be used.

consensus/tendermint/tendermint_test.go
IronGauntlets: If i have already manually calculated the q_min and max_f, I don't see the reason to add the equations.

consensus/tendermint/propose.go
IronGauntlets: I disagree, we want to have as close mapping to the lines in the paper as possible. Coming up with another name will only convolude the meaning and make it difficult to debug when the name is not clear enough to indicate which lines from the algorithm are being implemented.

consensus/tendermint/tendermint.go
IronGauntlets: Maybe it is possible, i didn't try to think of a way to avoid it but couldn't come up with one. The main reason for it was that we need to process the future messages as soon as possible and the easiest way to do that way through channels also while processing the future messages we cannot hold the processing of the new messages which arrive.

consensus/tendermint/messages.go
IronGauntlets: That is a possible solution but I am not sure whether it would work

consensus/tendermint/tendermint.go
IronGauntlets: Yes, these are certainly some optimisations which can be made, for the time being, I was aiming for simplicity. So having q or f is not the only thing which is important, the sender of the message is important too. So depending on how one designs the map there are different trades off. We probably need to think about our access patterns.

consensus/tendermint/propose.go
IronGauntlets: Ok I will update the names with part of the condition which ends with rule.

consensus/tendermint/tendermint.go
```diff
@@ -0,0 +1,434 @@
...
+	go func() {
+		defer t.wg.Done()
+
+		t.startRound(0)
+
+		// Todo: check message signature everytime a message is received.
+		// For the time being it can be assumed the signature is correct.
+
+		for {
+			select {
```
infrmtcs, line 208:
```
			if (there are future messages) {
						select {
						case <- t.timeoutsCh: ...
						default:
									handle N messages
						}
			}
```

consensus/tendermint/tendermint.go
```diff
@@ -0,0 +1,434 @@
...
+	validValue  *V
+	validRound  int
+
+	// The following are round level variable therefore when a round changes they must be reset.
+	timeoutPrevoteScheduled       bool // line34 for the first time condition
+	timeoutPrecommitScheduled     bool // line47 for the first time condition
+	lockedValueAndOrValidValueSet bool // line36 for the first time condition
+}
+
+func New[V Hashable[H], H Hash, A Addr](addr A, app Application[V, H], chain Blockchain[V, H, A], vals Validators[A],
```
rianhughes, line 167: Just thinking aloud here, but given this implementation is quite generic, would it require much additional work to run it against ethereums consensus network?

consensus/tendermint/tendermint.go
```diff
@@ -0,0 +1,434 @@
...
+}
+
+type Slasher[M Message[V, H, A], V Hashable[H], H Hash, A Addr] interface {
+	// Equivocation informs the slasher that a validator has sent conflicting messages. Thus it can decide whether to
+	// slash the validator and by how much.
+	Equivocation(msgs ...M)
+}
+
+type Listener[M Message[V, H, A], V Hashable[H], H Hash, A Addr] interface {
+	// Listen would return consensus messages to Tendermint which are set // by the validator set.
```
rianhughes, line 93:
```suggestion
	// Listen would return consensus messages to Tendermint which are set by the validator set.
```

consensus/tendermint/messages.go
```diff
@@ -0,0 +1,107 @@
...
+
+type Message[V Hashable[H], H Hash, A Addr] interface {
+	Proposal[V, H, A] | Prevote[H, A] | Precommit[H, A]
+}
+
+type Proposal[V Hashable[H], H Hash, A Addr] struct {
+	H          height
+	R          round
+	ValidRound int
+	Value      *V
```
rianhughes, line 20: Can we update Proposal so that Value isn't assigned to the heap? Eg something like this

```
// OptionalValue represents a value that may or may not be present
// This is stack-allocated and avoids heap allocations that pointers cause
type OptionalValue[V any] struct {
    Valid bool
    Value V
}


type Proposal[V Hashable[H], H Hash, A Addr] struct {
    H          height
    R          round
    ValidRound int
    Value      OptionalValue[V]  // Now stack-allocated instead of *V

    Sender A
}

```

consensus/tendermint/messages.go
```diff
@@ -0,0 +1,107 @@
...
+	Precommit[H Hash, A Addr] Vote[H, A]
+)
+
+type Vote[H Hash, A Addr] struct {
+	H  height
+	R  round
+	ID *H
+
+	Sender A
+}
```
rianhughes, line 36: Similar to stack assignement comment above

consensus/tendermint/messages.go
```diff
@@ -0,0 +1,107 @@
...
+
+func newMessages[V Hashable[H], H Hash, A Addr]() messages[V, H, A] {
+	return messages[V, H, A]{
+		proposals:  make(map[height]map[round]map[A][]Proposal[V, H, A]),
+		prevotes:   make(map[height]map[round]map[A][]Prevote[H, A]),
+		precommits: make(map[height]map[round]map[A][]Precommit[H, A]),
+	}
+}
+
+func addMessages[T any, A Addr](storage map[height]map[round]map[A][]T, msg T, s A, h height, r round) {
```
rianhughes, line 60: nit

```suggestion
func addMessages[T any, A Addr](storage map[height]map[round]map[A][]T, msg T, h height, r round, a A) {
```

consensus/tendermint/tendermint.go
rianhughes, line 167: nit

```suggestion
func New[V Hashable[H], H Hash, A Addr](nodeAddr A, app Application[V, H], chain Blockchain[V, H, A], vals Validators[A],
```

consensus/tendermint/messages.go
IronGauntlets: Although heap allocation should be avoided we must also remember that if the value is a very big struct we don't want to copy it every time it is passed. Since we are using generics, it is impossible to tell how big the value will be.

consensus/tendermint/messages.go
IronGauntlets: Similar to above, it is most likely the case that the hash is bigger than its reference 4 words vs 1 word

consensus/tendermint/messages.go
rianhughes: But in practice we will know what the value is right? It's just going to be a block (header & txns) right?

rianhughes (approved):

consensus/tendermint/propose.go
```diff
@@ -0,0 +1,246 @@
...
+		_, prevotesForHVr, _ := t.messages.allMessages(p.H, round(vr))
+
+		var vals []A
+		for addr, valPrevotes := range prevotesForHVr {
+			for _, p := range valPrevotes {
+				if *p.ID == vID {
+					vals = append(vals, addr)
+				}
+			}
+		}
```
infrmtcs, line 171: nit: I think we can have a function for these lines, so each line will have single level of abstraction.

consensus/tendermint/propose.go
```diff
@@ -0,0 +1,246 @@
...
+	var vals []A
+
+	for addr, valPrecommits := range precommitsForHR {
+		for _, p := range valPrecommits {
+			if *p.ID == vID {
+				precommits = append(precommits, p)
+				vals = append(vals, addr)
+			}
+		}
+	}
```
infrmtcs, line 94: nit: same as below, a function for these lines

consensus/tendermint/prevote.go
```diff
@@ -0,0 +1,231 @@
...
+			}
+		}
+
+		for addr, valPrevotes := range prevotesForHR {
+			for _, v := range valPrevotes {
+				if *v.ID == *p.ID {
+					vals = append(vals, addr)
+				}
+			}
+		}
```
infrmtcs, line 205: nit: same as below, probably 2 common functions for these logic

IronGauntlets:
Due to generics and different inputs, reducing duplication is difficult. I have reduced it a little bit, please suggest if there is a better way.

infrmtcs (approved):
Let's merge and then I'll make other refactor PRs to address the remaining issues

rodrigo-pino (approved):
Quick review to make it non blocking and continue our Consensus work

PR #2194, opened 2024-10-03
Starknet's `getStorageProof` rpc method

### LINKED ISSUE #2180: RPC08 - starknet_getStorageProof
Specification - https://github.com/starkware-libs/starknet-specs/blob/v0.8.0-rc0/api/starknet_api_openrpc.json#L910

Changes in PR https://github.com/starkware-libs/starknet-specs/pull/232

pnowosie:
Fixes #2180
Tests are based on [proof-refactor PR](https://github.com/NethermindEth/juno/pull/2252)


rpc/contract.go
```diff
@@ -27,22 +27,3 @@ func (h *Handler) Nonce(id BlockID, address felt.Felt) (*felt.Felt, *jsonrpc.Err
 
 	return nonce, nil
 }
-
```
pnowosie, line 30: Refactor: Moved to a better file: `rpc/storage,go`

blockchain/pending.go
```diff
@@ -65,3 +67,22 @@ func (p *PendingState) Class(classHash *felt.Felt) (*core.DeclaredClass, error)
...
+}
+func (s *PendingState) StorageTrie() (*trie.Trie, func() error, error) {
+	return nil, nopCloser, featureNotImplemented
+}
+func (s *PendingState) StorageTrieForAddr(*felt.Felt) (*trie.Trie, error) {
+	return nil, featureNotImplemented
+}
+func (s *PendingState) StateAndClassRoot() (*felt.Felt, *felt.Felt, error) {
+	return nil, nil, featureNotImplemented
+}
```
pnowosie, line 83: It looks like a bad idea to extend a StateReader which is implemented by other state variants like StateHistoryReader.
Otherwise - for tests we mock StateHistoryReader so in case of separate interface testing might be a bit harsh - ideas?

core/trie/trie.go
```diff
@@ -13,7 +13,7 @@ import (
 	"github.com/NethermindEth/juno/db"
 )
 
-type hashFunc func(*felt.Felt, *felt.Felt) *felt.Felt
+type HashFunc func(*felt.Felt, *felt.Felt) *felt.Felt
```
rianhughes, line 16: Why is thus public? Can we not keep it private?

core/trie/proof.go
```diff
@@ -54,6 +54,11 @@ func (e *Edge) Len() uint8 {
 	return e.Path.Len()
 }
 
+func (e *Edge) PathInt() uint64 {
```
rianhughes, line 57: Is this function still needed?

pnowosie: It depends how this is decided https://github.com/starkware-libs/starknet-specs/pull/232#discussion_r1791659441

rpc/storage.go
```diff
@@ -0,0 +1,278 @@
...
+		return nil, ErrInternal.CloneWithData(err)
+	}
+	defer h.callAndLogErr(stateCloser, "Error closing state reader in getStorageProof")
+
+	head, err := h.bcReader.Head()
+	if err != nil {
+		return nil, ErrInternal.CloneWithData(err)
+	}
+
+	if !(id.Latest || head.Number == id.Number) {
```
rianhughes, line 53: Should we also allow the users to specify the blockhash in the BlockID?

rianhughes: Alos, from the users perspective, this method will only seem "stable" if they set blockID="latest", as they are unlikely to know the block number or block hash that the juno node is on, right?

rpc/handlers.go
```diff
@@ -58,6 +58,9 @@ var (
 
 	// These errors can be only be returned by Juno-specific methods.
 	ErrSubscriptionNotFound = &jsonrpc.Error{Code: 100, Message: "Subscription not found"}
+
+	// TODO[pnowosie]: Update the error while specification describe it
+	ErrBlockNotRecentForProof = &jsonrpc.Error{Code: 1001, Message: "Block is not sufficiently recent for storage proofs"}
```
rianhughes, line 63: What do you think about responding that the user should try set Block ID to "latest"?

blockchain/pending.go
```diff
@@ -65,3 +67,22 @@ func (p *PendingState) Class(classHash *felt.Felt) (*core.DeclaredClass, error)
...
+
+var (
+	featureNotImplemented = errors.New("feature not implemented for a historical state")
```
rianhughes, line 86: nit: errFeatureNotImplemented

rpc/storage.go
pnowosie: For now - yes. 
Maybe I'll stick to "latest" as an only available option - otherwise it will work very strange from the users perspective 🤔

rpc/handlers.go
```diff
@@ -392,6 +395,11 @@ func (h *Handler) MethodsV0_6() ([]jsonrpc.Method, string) { //nolint: funlen
 			Params:  []jsonrpc.Parameter{{Name: "contract_address"}, {Name: "key"}, {Name: "block_id"}},
 			Handler: h.StorageAt,
 		},
+		{
+			Name:    "starknet_getStorageProof",
```
rianhughes, line 399: Should we also support this in `Methods()` (for rpcv7)? I guess it should techincally only be for rpcv8, but we don't have a function for that atm..

rpc/storage.go
```diff
@@ -0,0 +1,273 @@
...
+// TODO[pnowosie]: link to specs
+type StorageProofResult struct {
+	ClassesProof           []*HashToNode   `json:"classes_proof"`
+	ContractsProof         *ContractProof  `json:"contracts_proof"`
+	ContractsStorageProofs [][]*HashToNode `json:"contracts_storage_proofs"`
+	GlobalRoots            *GlobalRoots    `json:"global_roots"`
+}
+
+func getClassesProof(reader core.StateReader, classes []felt.Felt) ([]*HashToNode, error) {
+	ctrie, _, err := reader.ClassTrie()
```
rianhughes, line 149: Super nit: cTrie, feel free to ignore

rianhughes: same for sTrie below etc

rpc/storage.go
```diff
@@ -0,0 +1,273 @@
...
+		if err != nil {
+			return nil, err
+		}
+		result = append(result, nodes...)
+	}
+	return result, nil
+}
+
+func getContractsProof(reader core.StateReader, contracts []felt.Felt) (*ContractProof, error) {
+	strie, _, err := reader.StorageTrie()
```
rianhughes, line 165: why don't we handle the closer functions?

rpc/storage.go
```diff
@@ -0,0 +1,273 @@
...
+// StorageKeys represents an item in `contracts_storage_keys. parameter
+// https://github.com/starkware-libs/starknet-specs/blob/647caa00c0223e1daab1b2f3acc4e613ba2138aa/api/starknet_api_openrpc.json#L938
+type StorageKeys struct {
+	Contract felt.Felt   `json:"contract_address"`
+	Keys     []felt.Felt `json:"storage_keys"`
+}
+
+// MerkleNode represents a proof node in a trie
+// https://github.com/starkware-libs/starknet-specs/blob/647caa00c0223e1daab1b2f3acc4e613ba2138aa/api/starknet_api_openrpc.json#L3632
+type MerkleNode interface{}
```
rianhughes, line 94: Maybe we can constrain this somehow, eg something like GetNodeType() NodeType??

rianhughes: Or merge into a single type??

rpc/storage.go
```diff
@@ -0,0 +1,273 @@
...
+
+// TODO[pnowosie]: link to specs
+type MerkleEdgeNode struct {
+	Path   *felt.Felt `json:"path"`
+	Length int        `json:"length"`
+	Child  *felt.Felt `json:"child"`
+}
+
+// TODO[pnowosie]: link to specs
+type MerkleLeafNode struct {
```
rianhughes, line 110: Is this being used?

rpc/storage.go
```diff
@@ -0,0 +1,273 @@
...
+		return nil, ErrInternal.CloneWithData(err)
+	}
+
+	result := &StorageProofResult{
+		GlobalRoots: &GlobalRoots{
+			ContractsTreeRoot: storageRoot,
+			ClassesTreeRoot:   classRoot,
+			BlockHash:         head.Hash,
+		},
+	}
```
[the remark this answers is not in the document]
rianhughes: Ahm nvv, didn't see there were in this PR https://github.com/starkware-libs/starknet-specs/pull/232

rpc/storage.go
```diff
@@ -0,0 +1,273 @@
...
+		if err != nil {
+			return nil, err
+		}
+		result.Nodes = append(result.Nodes, nodes...)
+	}
+
+	return result, nil
+}
+
+func addLeafDataIfExists(reader core.StateReader, contract *felt.Felt) (*LeafData, error) {
```
rianhughes, line 192: another super nit: getLeafData() /getLeafDataIfExists()

rpc/storage.go
pnowosie: This is read-only and closer tries to update the root if it had changed.
BTW for the StateReader (r-o) I should drop the closer fn from the response but later for the snap-sync we will use it to put ranges into tries...

rpc/storage.go
pnowosie: Ah yes - this is [another objection](https://github.com/starkware-libs/starknet-specs/pull/232#discussion_r1791657811) to the specs. Thanks for the reminder 👍

rpc/storage.go
pnowosie: Yes, I was thinking about some artificial function - but now I have and idea: conversion to our ProofNode

rpc/handlers.go
```diff
@@ -58,6 +58,12 @@ var (
 
 	// These errors can be only be returned by Juno-specific methods.
 	ErrSubscriptionNotFound = &jsonrpc.Error{Code: 100, Message: "Subscription not found"}
+
+	// TODO[pnowosie]: Update the error while specification describe it
+	ErrStorageProofNotSupported = &jsonrpc.Error{
+		Code:    42,
+		Message: "the node doesn't support storage proofs for blocks that are too far in the past. Use 'latest' as block id",
```
rianhughes, line 65: nit: Juno only supports `starknet_getStorageProof` when block id is set to `latest`

starknet/.gitignore
```diff
@@ -0,0 +1 @@
+**/rust/
```
rianhughes, line 1: Why the changes in the .gitignore files?

core/trie/proof.go
rianhughes: This is still unused / unneeded?

core/trie/trie.go
rianhughes: I think you can keep this function private, but expose it via t.HashFunc()?

starknet/.gitignore
pnowosie: Good question, these are rust untracked files - I belive it's starknet/compiler/* that appears out of nowhere after I compiled the all thing

```
Untracked files:
  (use "git add <file>..." to include in what will be committed)
	starknet/_gitignore
	starknet/compiler/rust/Cargo.lock
	starknet/compiler/rust/target/
	starknet/rust/
```

blockchain/blockchain.go
```diff
@@ -769,6 +770,17 @@ func (b *Blockchain) HeadState() (core.StateReader, StateCloser, error) {
 	return core.NewState(txn), txn.Discard, nil
 }
 
+func (b *Blockchain) HeadTrie() (core.TrieReader, StateCloser, error) {
```
rianhughes, line 773: Should we add a listener (b.listener.OnRead(..)) and doc comment?

rpc/storage_test.go
```diff
@@ -0,0 +1,319 @@
...
+		assert.Equal(t, expectedStorage, storage)
+	})
+}
+
+func TestStorageProof(t *testing.T) {
+	// dummy values
+	var (
+		blkHash     = utils.HexToFelt(t, "0x11ead")
+		clsRoot     = utils.HexToFelt(t, "0xc1a55")
+		stgRoor     = utils.HexToFelt(t, "0xc0ffee")
```
rianhughes, line 107: typo

rpc/storage_test.go
```diff
@@ -0,0 +1,319 @@
...
+		mockReader.EXPECT().StateAtBlockNumber(uint64(0)).Return(mockState, nopCloser, nil)
+		mockState.EXPECT().ContractStorage(gomock.Any(), gomock.Any()).Return(expectedStorage, nil)
+
+		storage, rpcErr := handler.StorageAt(felt.Zero, felt.Zero, rpc.BlockID{Number: 0})
+		require.Nil(t, rpcErr)
+		assert.Equal(t, expectedStorage, storage)
+	})
+}
+
+func TestStorageProof(t *testing.T) {
```
rianhughes, line 102: I wonder if we should add a test that matches some data/response that was generated outside of Juno. Currently we are testing that the method is consistent with everything else inside Juno, which is great, but imo we should also test that the RPC method generates the expected output.

rpc/storage.go
```diff
@@ -0,0 +1,297 @@
...
+		}
+		result = append(result, nodes)
+	}
+
+	return result, nil
+}
+
+func getProof(t *trie.Trie, elt *felt.Felt) ([]*HashToNode, error) {
+	feltBytes := elt.Bytes()
+	key := trie.NewKey(core.ContractStorageTrieHeight, feltBytes[:])
```
rianhughes, line 263: maybe we can export the tries height, and use t.Height() here?

rpc/storage.go
```diff
@@ -0,0 +1,297 @@
...
+	nodes, err := trie.GetProof(&key, t)
+	if err != nil {
+		return nil, err
+	}
+
+	// adapt proofs to the expected format
+	hashnodes := make([]*HashToNode, len(nodes))
```
rianhughes, line 270: nit: hashNodes (sorry for all the nit comments!)

starknet/.gitignore
kirugan: these files shouldn't be tracked because we have .gitignore in /starknet/compiler/rust dir

kirugan: you removed it in this PR

rpc/handlers.go
kirugan, line 1: `starknet_getStorageProof` should be only in 0.8 endpoint. I'll create PR so we all in sync how we should enable new handlers

core/trie/trie.go
pnowosie: Yeah, when I added:
`func (t *Trie) HashFunc() HashFunc {`
compiler wanted me to expose it's return type public

rpc/storage_test.go
pnowosie: I've added `TestStorageRoots` - a bit less of what you're asking but some starting point

rpc/storage.go
rianhughes: Good point, maybe we can remove it since it's not being used, and to prevent any potential problems in the future?

core/trie/trie.go
rianhughes: It's not giving me any compilation problems (see [here](https://github.com/NethermindEth/juno/commit/d51bc26601c0c388b6d1343658a0da09ddca5460)), but feel free to ignore me if I'm doing something wrong :)

pnowosie: Maybe it wasn't compiler but linter. Anyway I'm leaving as is - publicly exposed function uses public type.
Thanks for the suggestions 👍

rpc/storage.go
pnowosie: I will leave for now and try to refactor once me merge the snap-sync pr (the same methods with same return types)

rpc/handlers.go
```diff
@@ -233,6 +238,13 @@ func (h *Handler) Methods() ([]jsonrpc.Method, string) { //nolint: funlen, dupl
 			Params:  []jsonrpc.Parameter{{Name: "contract_address"}, {Name: "key"}, {Name: "block_id"}},
 			Handler: h.StorageAt,
 		},
+		{
+			Name: "starknet_getStorageProof",
+			Params: []jsonrpc.Parameter{
+				{Name: "block_id"}, {Name: "classes", Optional: true}, {Name: "contracts", Optional: true}, {Name: "storage_keys", Optional: true},
```
kirugan, line 244: according to https://github.com/starkware-libs/starknet-specs/blob/a789ccc3432c57777beceaa53a34a7ae2f25fda0/api/starknet_api_openrpc.json#L110

all parameters are not optional and there is no `storage_keys`

rianhughes (approved):

weiihann:
I don't think it's a good idea to add `HeadTrie` method to the blockchain's Reader interface for several reasons:

There's a clear hierarchical relationship where:
- `Blockchain` contains `State`
- `State` internally uses `Trie` for implementation
- `Blockchain` shouldn't need to know about these implementation details of `State`

The current `HeadTrie()` implementation is just a wrapper around `HeadState()`, which suggests this abstraction isn't providing additional value.

Instead of having a separate TrieReader interface, I'd suggest:
   - Remove the `TrieReader` interface entirely
   - Move the trie-related methods directly into `StateReader`
   - This better reflects that trie operations are a capability of the state, not the blockchain

The problem with this approach is that we can only support trie operations on the latest state. Historical trie access (via `PendingState` and `stateSnapshot`) are not allowed. I think that's fine since the RPC method will return an error if the block ID is not `latest`

pnowosie:
> I don't think it's a good idea to add `HeadTrie` method to the blockchain's Reader interface for several reasons:
> ... Move the trie-related methods directly into StateReader

@weiihann Thanks for this comment.
This was my initial implementation - trie related methods was added to the `StateReader`. Then I spotted a few problems which were the need to implement these methods for all StateReaders (where it would be possible only for the (current) `State` not historical ones.
The another thing was that we only mock the `StateHistoryReader` - and there was something related to it in tests I cannot recall right now.

I discussed this issue with @kirugan and we agreed that we address this once we allow juno to prove historical states

pnowosie:
## Differences / Issues spotted comparing with PF

### 1. Params are not optional (why?)
Request
```
{
  "jsonrpc": "2.0",
  "method": "starknet_getStorageProof",
  "params": [
    "latest",
    ["0x1adadae7ce7ed3f5c39f275ef4758562e540f27ab0184b24e2b13861988751e"]
  ],
  "id": 1
}
```
Response
"message": "Invalid Params",

### 2. Response does not filter out duplicate nodes in hash HashToNode mapping

Request
```
{
  "jsonrpc": "2.0",
  "method": "starknet_getStorageProof",
  "params": [
    "latest",
    ["0x1adadae7ce7ed3f5c39f275ef4758562e540f27ab0184b24e2b13861988751e", 
     "0x1adadae7ce7ed3f5c39f275ef4758562e540f27ab0184b24e2b13861988751e"],
    [],
    []
  ],
  "id": 1
}
```

Response
```
{
  "jsonrpc": "2.0",
  "result": {
    "classes_proof": [
      {
        "node_hash": "0x217d3fbc3bccaca79f86feb5e9dbbfa531ab481503594df9ca426245a65a12f",
        "node": { ... }
      },
// ...
      {
        "node_hash": "0x1c2825874f9b479403ea627428397852d86c67d955be072d68de143a98bb923",
        "node": { ...
          "child": "0x35f5f9b5523cdbd190f2d7a7310aadcfcf0231a9a580786b9cd5db8b512de2"
        }
      },
      {
        "node_hash": "0x217d3fbc3bccaca79f86feb5e9dbbfa531ab481503594df9ca426245a65a12f",
         "node": { ... }
      },
// ...
      {
        "node_hash": "0x1c2825874f9b479403ea627428397852d86c67d955be072d68de143a98bb923",
        "node": { ...
          "child": "0x35f5f9b5523cdbd190f2d7a7310aadcfcf0231a9a580786b9cd5db8b512de2"
        }
      }
    ],
  // ...
  "id": 1
}
```

core/trie/proof.go
```diff
@@ -316,11 +316,14 @@ func GetProof(key *Key, tri *Trie) ([]ProofNode, error) {
...
 // https://github.com/eqlabs/pathfinder/blob/main/crates/merkle-tree/src/tree.rs#L2006
-func VerifyProof(root *felt.Felt, key *Key, value *felt.Felt, proofs []ProofNode, hash hashFunc) bool {
+func VerifyProof(root *felt.Felt, key *Key, value *felt.Felt, proofs []ProofNode, hash HashFunc) bool {
 	expectedHash := root
 	remainingPath := NewKey(key.len, key.bitset[:])
 	for i, proofNode := range proofs {
 		if !proofNode.Hash(hash).Equal(expectedHash) {
+			proofHash := proofNode.Hash(hash)
+			fmt.Println("Proof verification failure", "node", proofHash, "expected", expectedHash)
+			fmt.Printf("%v\n", proofNode)
```
weiihann, line 326: please remove these debugging lines

core/trie/proof.go
```diff
@@ -345,7 +348,8 @@ func VerifyProof(root *felt.Felt, key *Key, value *felt.Felt, proofs []ProofNode
 				return true
 			}
 
-			if !proofNode.Path.Equal(subKey) {
+			if !proofNode.Path.Equal(subKey) && !subKey.Equal(&Key{}) {
+				fmt.Println("Proof verification failure", "node", proofNode.Path, "expected", subKey)
```
weiihann, line 352: please remove these debugging lines

weiihann:
> I discussed this issue with @kirugan and we agreed that we address this once we allow juno to prove historical states

So there are 2 approaches we can take:

1. Implements `TrieReader` interface and return `HeadTrie` in Blockchain `Reader` interface
2. Add Trie access methods in `StateReader` interface

For the 1st approach, we can agree on the architecture such that a Blockchain `Reader` shouldn't know the underlying `State` implementation (i.e. `Trie). Adding `HeadTrie` method to `Reader` indicates we are coupling `Trie` and `Blockchain` together, which shouldn't happen. So essentially this is a technical debt.

For the 2nd approach, the architecture makes sense, the issue is that we have an unsupported feature for historical trie access. It's unsure when we will support historical trie access as it's a **huge refactor** on the state, trie and db modules.

I'd rather go with an unsupported feature and detect it early rather than incur technical debts and make the modules even more coupled and harder to refactor.

core/trie/key.go
```diff
@@ -28,6 +28,9 @@ func (k *Key) SubKey(n uint8) (*Key, error) {
 	if n > k.len {
 		return nil, errors.New(fmt.Sprint("cannot subtract key of length %i from key of length %i", n, k.len))
 	}
+	if n == k.len {
+		return &Key{}, nil
+	}
```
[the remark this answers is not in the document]
pnowosie: I found a bug in proof-verification here in the test:
https://github.com/NethermindEth/juno/blob/7252e10a551f35b03a04c047e6f9943e46b4106b/rpc/storage_test.go#L166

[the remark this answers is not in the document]
weiihann: I see, that actually doesn't fix the other test cases. It'll be clearer in my `VerifyProof` implementation.

core/trie/proof.go
```diff
@@ -134,153 +165,6 @@ func transformNode(tri *Trie, parentKey *Key, sNode StorageNode) (*Edge, *Binary
 	return edge, binary, nil
 }
 
-// pathSplitOccurredCheck checks if there happens at most one split in the merged path
```
pnowosie, line 137: Note: This functions are reported unused by linter, hence I deleted them.
Maybe further PR #2252 will bring them back.

core/trie/proof.go
```diff
@@ -314,133 +198,84 @@ func GetProof(key *Key, tri *Trie) ([]ProofNode, error) {
...
-	return expectedHash.Equal(value)
-}
-
-// VerifyRangeProof verifies the range proof for the given range of keys.
-// This is achieved by constructing a trie from the boundary proofs, and the supplied key-values.
-// If the root of the reconstructed trie matches the supplied root, then the verification passes.
-// If the trie is constructed incorrectly then the root will have an incorrect key(len,path), and value,
-// and therefore it's hash won't match the expected root.
-// ref: https://github.com/ethereum/go-ethereum/blob/v1.14.3/trie/proof.go#L484
-func VerifyRangeProof(root *felt.Felt, keys, values []*felt.Felt, proofKeys [2]*Key, proofValues [2]*felt.Felt,
```
pnowosie, line 365: Will be added back as of #2252 with proper test coverage. It isn't used anywhere and hence ongoing proof refactor it is safe to drop for now.

core/trie/proof_test.go
```diff
@@ -183,113 +184,270 @@ func build3KeyTrie(t *testing.T) *trie.Trie {
 	return tempTrie
 }
 
-func build4KeyTrie(t *testing.T) *trie.Trie {
```
pnowosie, line 186: Some of these deleted functions will be bring back with #2252.
For this PR they are unused.

PR #2197, opened 2024-10-04
Refactor retrieve state history API

weiihann:
Let's start straight away with an example:

Block 0: Alice initially has 10 STRK
Block 10: Alice now has 20 STRK
Block 25: Alice now has 30 STRK

We store state history (i.e. buckets which are `ContractNonceHistory`, `ContractStorageHistory`, `ContractClassHashHistory`) in the following key-value format:
```
(prefix, contract address, block height) = old value
```

In our given example, let's pretend Alice is the contract address, then we will have the following key-value pairs in the database:
```
(prefix, Alice, 10) = 10
(prefix, Alice, 25) = 20
```

Ok, now we have the database in mind, let's dive into the code. Look at the `func (h *history) valueAt(key []byte, height uint64)` method. The first intuition of this method is that it returns the history at a given key with the given height. This intuition is further solidified by the methods that uses it, which are `history.ContractStorageAt`, `history.ContractNonceAt` and `history.ContractClassHashAt` (check the comments). So I would expect something like:

```
valueAt((prefix+Alice), 10) = 10
valueAt((prefix+Alice), 25) = 20
``` 

Now, if you look at the implementation. That's actually not the case. What the method actually does is that it returns the value at given key **AFTER** the given height. Let me repeat it, it is **AFTER** a given block height.

The original intention of doing this is that, we can represent the history of a given state for a range of block heights. So something like:
- From block 0 to 9 = Alice has 10 STRK
- From block 10 to 24 = Alice has 20 STRK

Finally, let's check if the original intention was actually used. Look at the usage of `ContractClassHashAt`, `ContractNonceAt` and `ContractStorageAt`. They have this pattern:
```
oldValue, err := s.ContractStorageAt(&addr, &key, blockNumber-1)
oldNonce, err = s.ContractNonceAt(&addr, blockNumber-1)
classHash, err = s.ContractClassHashAt(&addr, blockNumber-1) 
```

So actually, the usage of these methods implies that all we really want to get is just the value of a state at a given key with the given height! So why do we bother with representing the history of a given state for a range of block heights in the first place?

Hence, this PR reimplements the methods to suit their intent.

### DIFF
--- a/core/history.go
+++ b/core/history.go
@@ -26,7 +26,7 @@ func (h *history) deleteLog(key []byte, height uint64) error {
 	return h.txn.Delete(logDBKey(key, height))
 }
 
-// valueAt returns the value at the given height for the given key if it exists
+// valueAt returns the old value at the given height for the given key if it exists
 func (h *history) valueAt(key []byte, height uint64) ([]byte, error) {
 	var value []byte
 	err := h.txn.Get(logDBKey(key, height), func(val []byte) error {
--- a/core/history.go
+++ b/core/history.go
@@ -1,13 +1,11 @@
 package core
 
 import (
-	"bytes"
 	"encoding/binary"
 	"errors"
 
 	"github.com/NethermindEth/juno/core/felt"
 	"github.com/NethermindEth/juno/db"
-	"github.com/NethermindEth/juno/utils"
 )
 
 var ErrCheckHeadState = errors.New("check head state")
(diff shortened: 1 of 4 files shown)


weiihann:
On a second thought, maybe it just depends on you look at it:

My perspective:
When I read through the code, I expect `history.valueAt`, `history.ContractStorageAt` etc would give me the old value where changes occur at a given block height, just like how it's portrayed in the database.

The previous implementation:
`history.valueAt` implies the actual value at the given block height. So that's why we do `s.ContractStorageAt(&addr, &key, blockNumber-1)` because it actually does give me the value at that given block height.

Personally, I prefer the current changes that I have for this PR because the methods speak for themselves. But I'm not sure if my perspective is the same as others :)

weiihann:
If we choose to go with the current set of changes, perhaps it's even clearer to change the `StateHistoryReader` interface methods to:
```
type StateHistoryReader interface {
	StateReader

	OldContractStorageAt(addr, key *felt.Felt, blockNumber uint64) (*felt.Felt, error)
	OldContractNonceAt(addr *felt.Felt, blockNumber uint64) (*felt.Felt, error)
	OldContractClassHashAt(addr *felt.Felt, blockNumber uint64) (*felt.Felt, error)
	OldContractIsAlreadyDeployedAt(addr *felt.Felt, blockNumber uint64) (bool, error)
}
```

weiihann:
...`stateSnapshot` also implements the `StateHistoryReader` interface..............

PR #2198, opened 2024-10-07
Fix core/state_test tests failure

weiihann:
If we run the test in the following command:
```
go test ./core/state_test.go -v
```

We would get test errors as such:
```
...
cbor: cannot unmarshal map into Go struct field core.DeclaredClass.Class of type core.Class
```

But if we run the test with the following command, no errors are found:
```
go test ./core -v
```

The `cbor` encoder and decoder that we are using are initialized globally. In the case when are running tests, running all the tests in the `core` package would initialize the encoder/decoder correctly, but some struct types were registered properly if we only run the `state_test.go`.

This PR attempts to fix the proper registering and deregistering types in the `state_test.go` file.

### DIFF
--- a/core/state_test.go
+++ b/core/state_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"os"
 	"reflect"
 	"testing"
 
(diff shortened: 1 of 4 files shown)


core/state_test.go
```diff
@@ -380,15 +405,6 @@ func TestClass(t *testing.T) {
 	cairo1Class, err := gw.Class(context.Background(), cairo0Hash)
 	require.NoError(t, err)
 
-	err = encoder.RegisterType(reflect.TypeOf(cairo0Class))
-	if err != nil {
-		require.Contains(t, err.Error(), "already exists in TagSet")
-	}
-	err = encoder.RegisterType(reflect.TypeOf(cairo1Hash))
```
rianhughes, line 387: This was registering a felt with the encoder right? Should we not include it in the set of types we register in TestMain?

encoder/encoder.go
```diff
@@ -50,6 +50,11 @@ func RegisterType(rType reflect.Type) error {
 	return nil
 }
 
+func DeregisterType(rType reflect.Type) {
+	ts.Remove(rType)
+	initEncAndDecModes()
+}
+
```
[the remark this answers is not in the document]
kirugan: also I'm not sure that we need `initEncAndDecModes()` in tear down step

core/state_test.go
```diff
@@ -25,6 +26,30 @@ var (
...
 
+func TestMain(m *testing.M) {
+	types := []reflect.Type{
+		reflect.TypeOf(core.DeclareTransaction{}),
+		reflect.TypeOf(core.DeployTransaction{}),
+		reflect.TypeOf(core.InvokeTransaction{}),
+		reflect.TypeOf(core.L1HandlerTransaction{}),
+		reflect.TypeOf(core.DeployAccountTransaction{}),
+		reflect.TypeOf(core.Cairo0Class{}),
+		reflect.TypeOf(core.Cairo1Class{}),
```
kirugan, line 37: nit: move reflect.TypeOf into `range types` loop, change `types` var to `txTypes = []core.Transaction{...}`

encoder/encoder.go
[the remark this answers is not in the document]
weiihann: Actually we don't even need to deregister

core/state_test.go
weiihann: I'm not sure why it was registered in the first place, because I was trying to mimic the registration behavior in `blockchain/encoder_initializer.go`, which doesn't register the `felt` type.

kirugan (approved):

rianhughes (approved):

kirugan (approved):

PR #2204, opened 2024-10-08
Add starknet_getCompiledCasm

kirugan:
Here are several class hashes that I used to test this PR:
1. cairo0 (0x5f18f9cdc05da87f04e8e7685bd346fc029f977167d5b1b2b59f69a7dacbfc8)
2. cairo1 (0x0631509cbd3a29d2ab5fe39e26b9fdbb7eac2db19d01ffb9582a444ddc072ba8)

Test script to check cairo0 classes:
```
func TestMe(t *testing.T) {
	fd, err := os.Open("") // path to list of cairo0 class hashes (each line - class hash)
	require.NoError(t, err)

	sc := bufio.NewScanner(fd)

	var classHashes []*felt.Felt
	for sc.Scan() {
		classHash, err := new(felt.Felt).SetString(sc.Text())
		require.NoError(t, err)

		classHashes = append(classHashes, classHash)
	}
	require.NoError(t, sc.Err())

	db, err := pebble.New("") // path to your db
	require.NoError(t, err)

	chain := blockchain.New(db, &utils.Sepolia)
	st, closer, err := chain.HeadState()
	require.NoError(t, err)
	defer closer()

	for _, classHash := range classHashes {
		declaredClass, err := st.Class(classHash)
		require.NoError(t, err)

		cairo0 := declaredClass.Class.(*core.Cairo0Class)
		_, err = adaptCairo0Class(cairo0)
		if err == nil {
			fmt.Println("Seems good", classHash)
		} else {
			fmt.Println("Seems bad", classHash, err)
		}
	}
}
```


rpc/executables.go
```diff
@@ -0,0 +1,132 @@
...
+		EntryPointsByType: EntryPointsByType{
+			Constructor: utils.Map(class.Constructors, adaptEntryPoint),
+			External:    utils.Map(class.Externals, adaptEntryPoint),
+			L1Handler:   utils.Map(class.L1Handlers, adaptEntryPoint),
+		},
+		Prime:                  cairo0.Prime,
+		Bytecode:               cairo0.Data,
+		CompilerVersion:        "",
+		Hints:                  nil,
+		BytecodeSegmentLengths: nil,
```
kirugan, line 92: tbh I have no idea how to get these for cairo0

rpc/executables.go
```diff
@@ -0,0 +1,155 @@
...
+
+	hintRunnerZero "github.com/NethermindEth/cairo-vm-go/pkg/hintrunner/zero"
+	"github.com/NethermindEth/cairo-vm-go/pkg/parsers/zero"
+	"github.com/NethermindEth/juno/core"
+	"github.com/NethermindEth/juno/core/felt"
+	"github.com/NethermindEth/juno/jsonrpc"
+	"github.com/NethermindEth/juno/utils"
+)
+
+type CasmEntryPoint struct {
```
pnowosie, line 14: Maybe it's good idea to add a comment with a link to spec - similar to other rpc methods?

rpc/executables.go
```diff
@@ -0,0 +1,155 @@
...
+}
+
+func (h *Handler) CompiledCasm(classHash *felt.Felt) (*CasmCompiledContractClass, *jsonrpc.Error) {
+	state, stateCloser, err := h.bcReader.HeadState()
+	if err != nil {
+		return nil, jsonrpc.Err(jsonrpc.InternalError, err.Error())
+	}
+	defer h.callAndLogErr(stateCloser, "failed to close state reader")
+
+	declaredClass, err := state.Class(classHash)
```
pnowosie, line 43: Shouldn't we return [`CLASS_HASH_NOT_FOUND`](https://github.com/starkware-libs/starknet-specs/blob/19ab9d4df4ae3acc1a52cde5a43a7cace08bcc4b/api/starknet_executables.json#L37C50-L37C70) in case class not found?

rpc/executables.go
```diff
@@ -0,0 +1,155 @@
...
+		return CasmEntryPoint{
+			Offset:   ep.Offset,
+			Selector: ep.Selector,
+			Builtins: nil,
+		}
+	}
+
+	result := &CasmCompiledContractClass{
+		EntryPointsByType: EntryPointsByType{
+			Constructor: utils.Map(class.Constructors, adaptEntryPoint),
```
pnowosie, line 107: `utils.Map()` - wow, nice. I haven't noticed we have this

rpc/executables.go
```diff
@@ -0,0 +1,155 @@
...
+		Prime:                  cairo0.Prime,
+		Bytecode:               bytecode,
+		CompilerVersion:        cairo0.CompilerVersion,
+		Hints:                  json.RawMessage(rawHints),
+		BytecodeSegmentLengths: nil, // Cairo 0 classes don't have this field (it was introduced since Sierra 1.5.0)
+	}
+	return result, nil
+}
+
+func adaptCompiledClass(class *core.CompiledClass) *CasmCompiledContractClass {
```
pnowosie, line 120: Maybe it's obvious to everybody but it takes me a few seconds to recall - so please add a comment it's for cairo1 classes (feel free to ignore)

utils/ordered_map.go
```diff
@@ -0,0 +1,26 @@
...
+
+import (
+	"cmp"
+	"iter"
+	"slices"
+)
+
+func OrderMap[K cmp.Ordered, T any](m map[K]T) iter.Seq2[K, T] {
+	// 1. collect keys
+	keys := make([]K, 0, len(m))
```
pnowosie, line 11: I've seen we use `maps.Keys` e.g. in state_updates, wonder if we could use it here as well?

pnowosie (approved):
Great work 👍

rpc/executables.go
kirugan: yep, it's used across all of our codebase in adapters

utils/ordered_map.go
kirugan: Rewrote it with `maps.Keys(...)` and `slices.Sorted(...)`. Former action actually return an iterator, but that's ok

rpc/executables.go
kirugan: good point, thanks

rpc/executables.go
kirugan: I think it's ok without comment. The only way to get a compiled class is to pass corresponding field from cairo1 class

weiihann (approved):

PR #2206, opened 2024-10-09
feat: Check for rust version in Makefile

derrix060:
- Enhanced MSRV check in GitHub workflow to include Makefile updates.
- Added a new target in Makefile to ensure Rust version compliance.
- Updated dependencies and formatting targets to include Rust version check.
- Fixed minor formatting issues in Makefile comments.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,8 +13,8 @@ endif
 ifeq ($(shell uname -s),Darwin)
 	export CGO_LDFLAGS=-framework Foundation -framework SystemConfiguration
 
-	# Set macOS deployment target in order to avoid linker warnings linke
-	# "ld: warning: object file XX was built for newer macOS version (14.4) than being linked (14.0)"
+	# Set macOS deployment target in order to avoid linker warnings linke
+	# "ld: warning: object file XX was built for newer macOS version (14.4) than being linked (14.0)"
 	export MACOSX_DEPLOYMENT_TARGET=$(shell sw_vers --productVersion)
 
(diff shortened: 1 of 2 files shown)


wojciechos:
Looks like pipeline [started to fail](https://github.com/NethermindEth/juno/actions/runs/11287790309/job/31394327428?pr=2206#step:3:907) because of `error: 'cargo msrv' requires a subcommand but one was not provided`. Shouldn't it be `cargo msrv` with find subcommand [here](https://github.com/derrix060/detect-rust-minimum-version/blob/main/action.yml#L62)?

wojciechos (approved):

Makefile
```diff
@@ -91,7 +98,7 @@ lint: install-golangci-lint
 tidy: ## add missing and remove unused modules
 	 go mod tidy
 
-format: ## run go & rust formatters
+format: check-rust ## run go & rust formatters
```
kirugan, line 101: it's not critical for format target, maybe remove `check-rust` from here?

derrix060: doesn't it need to have rust to format?  From what I saw it formats both go and rust

kirugan: it does (but in theory having `rustfmt` doesn't mean that you need `rustc`). But my point is that we don't need specific rust version in order to format rust codebase.

Rust from version to version doesn't change language significantly, they use so-called editions to introduce breaking changes into the language. And editions are controlled by cargo.toml

kirugan (approved):

PR #2210, opened 2024-10-11
starknet_subscribeTransactionStatus websocket method

### LINKED ISSUE #2209: Implement starknet_subscribeTransactionStatus websocket method
See spec https://github.com/starkware-libs/starknet-specs/blob/v0.8.0-rc0/api/starknet_ws_api.json

pnowosie:
Fixes #2209 

**Comments:**
- In the meantime this PR was in review @IronGauntlets added `utils/check.go` (#2340)
I renamed  `utils/check.go` to `utils/nil.go` because of [this suggestion](https://github.com/NethermindEth/juno/pull/2210#discussion_r1886588597)


rpc/handlers.go
```diff
@@ -312,9 +328,20 @@ func (h *Handler) Methods() ([]jsonrpc.Method, string) { //nolint: funlen, dupl
...
 			Handler: h.SubscribeNewHeads,
 		},
+		{
+			Name:    "starknet_subscribePendingTransactions",
+			Params:  []jsonrpc.Parameter{{Name: "transaction_details", Optional: true}, {Name: "sender_address", Optional: true}},
+			Handler: h.SubscribePendingTxs,
+		},
+		{
+			Name:    "starknet_subscribeTransactionStatus",
+			Params:  []jsonrpc.Parameter{{Name: "transaction_hash"}, {Name: "block"}},
```
weiihann, line 342: `block` should be optional as per the specs

rpc/events.go
```diff
@@ -63,37 +81,373 @@ func (h *Handler) SubscribeNewHeads(ctx context.Context) (uint64, *jsonrpc.Error
...
+		},
+	})
+	if err != nil {
+		return err
+	}
+	_, err = w.Write(resp)
+	return err
+}
+
+func (h *Handler) SubscribeTxnStatus(ctx context.Context, txHash felt.Felt, _ *BlockID) (*SubscriptionID, *jsonrpc.Error) {
```
weiihann, line 375: As per the specs, we also need to support historical blocks. Or is it not possible?

jsonrpc/server.go
```diff
@@ -498,7 +508,12 @@ func (s *Server) buildArguments(ctx context.Context, params any, method Method)
 		addContext = 1
 	}
 
-	if isNil(params) {
+	isNilOrEmpty, err := isNilOrEmpty(params)
+	if err != nil {
+		return nil, err
+	}
+
+	if isNilOrEmpty {
```
kirugan, line 516: why do we need this ?

rpc/events.go
```diff
@@ -9,11 +9,20 @@ import (
 	"github.com/NethermindEth/juno/jsonrpc"
 )
 
+const (
+	MaxBlocksBack        = 1024
```
kirugan, line 13: There is new constant introduced in Aneeque's PR `maxBlocksBack` in `rpc/handlers.go`. If both of them related consider to reuse existing one

rpc/events.go
kirugan: maybe we should return error in this case? otherwise users will be confused why it's behaving differently from other  possible impl.

rpc/events.go
```diff
@@ -100,6 +105,87 @@ func (h *Handler) SubscribeNewHeads(ctx context.Context) (uint64, *jsonrpc.Error
...
+		}
+	)
+
+	w, ok := jsonrpc.ConnFromContext(ctx)
+	if !ok {
+		return nil, jsonrpc.Err(jsonrpc.MethodNotFound, nil)
+	}
+
+	id := h.idgen()
+	subscriptionCtx, subscriptionCtxCancel := context.WithCancel(ctx)
```
kirugan, line 128: do you need to call `defer subscriptionCtxCancel()` ?

rpc/events.go
```diff
@@ -100,6 +105,87 @@ func (h *Handler) SubscribeNewHeads(ctx context.Context) (uint64, *jsonrpc.Error
...
+	headerSub := h.newHeads.Subscribe()
+	sub.wg.Go(func() {
+		defer func() {
+			h.unsubscribe(sub, id)
+			statusSub.Unsubscribe()
+			headerSub.Unsubscribe()
+		}()
+
+		if err := h.sendTxnStatus(sub.conn, wrapResult(lastKnownStatus), id); err != nil {
+			h.log.Warnw("Error while sending Txn status", "txHash", txHash)
```
kirugan, line 154: maybe correct severity would be `Error` instead of `Warnw` ? Also please add `"err", err` as last arguments

rpc/events.go
```diff
@@ -100,6 +105,87 @@ func (h *Handler) SubscribeNewHeads(ctx context.Context) (uint64, *jsonrpc.Error
...
+		lastSendStatus = lastKnownStatus
+
+		for {
+			select {
+			case <-subscriptionCtx.Done():
+				return
+			case <-headerSub.Recv():
+				lastKnownStatus, rpcErr = h.TransactionStatus(subscriptionCtx, txHash)
+				if rpcErr != nil {
+					h.log.Warnw("Failed to get Tx status", "txHash", txHash, "rpcErr", rpcErr)
```
kirugan, line 166: same question: maybe it should be an error ?

rpc/events.go
```diff
@@ -100,6 +105,87 @@ func (h *Handler) SubscribeNewHeads(ctx context.Context) (uint64, *jsonrpc.Error
...
+				if *lastKnownStatus != *lastSendStatus {
+					if err := h.sendTxnStatus(sub.conn, wrapResult(lastKnownStatus), id); err != nil {
+						h.log.Warnw("Error while sending Txn status", "txHash", txHash)
+						return
+					}
+					lastSendStatus = lastKnownStatus
+				}
+
+				// Stop when final status reached and notified
+				if isFinal(lastSendStatus) {
```
kirugan, line 179: is it according to the spec ?

utils/check.go
kirugan, line 1: do you think `utils/nil.go` would be better naming?

rpc/handlers.go
```diff
@@ -512,7 +523,7 @@ func (h *Handler) MethodsV0_7() ([]jsonrpc.Method, string) { //nolint: funlen
 			Handler: h.SpecVersionV0_7,
 		},
 		{
-			Name:    "juno_subscribeNewHeads",
+			Name:    "starknet_subscribeNewHeads",
```
kirugan, line 526: why do you need to update `starknet_subscribeNewHeads` in this PR ?

pnowosie: I'll revert 👍 - I based on Han's PR - and then reverted when Anique's PR get merged

jsonrpc/server.go
pnowosie: Not necessary for mine - will come in Han's.
I will revert.

rpc/events.go
pnowosie: It's not in written the spec.
I remember we are talking about the final statuses (=no next status in diagram)
I cannot recall why - if it was me who deducted that the connection should be dropped in this case or it was in the group conversation 🤔

rpc/events.go
pnowosie: It's called by `sub.unsubscribe()`

rpc/events.go
IronGauntlets, line 1: Can you please move SubscribeTxnStatus to subscription.go?

rpc/events_test.go
IronGauntlets, line 1: Please move the following test to subscription_test.go

rpc/events.go
```diff
@@ -217,3 +298,30 @@ func setEventFilterRange(filter blockchain.EventFilterer, fromID, toID *BlockID,
...
 }
+
+type NewTransactionStatus struct {
+	TransactionHash *felt.Felt         `json:"transaction_hash"`
+	Status          *TransactionStatus `json:"status"`
+}
+
+// sendHeader creates a request and sends it to the client
+func (h *Handler) sendTxnStatus(w jsonrpc.Conn, status *NewTransactionStatus, id uint64) error {
+	resp, err := json.Marshal(jsonrpc.Request{
```
IronGauntlets, line 309: Please. use SubscriptionResponse struct instead of jsonrpc.Request

rpc/events.go
```diff
@@ -100,6 +100,87 @@ func (h *Handler) SubscribeNewHeads(ctx context.Context) (uint64, *jsonrpc.Error
...
+	if rpcErr != nil {
+		h.log.Errorw("Failed to get Tx status", "txHash", &txHash, "rpcErr", rpcErr)
+		return nil, rpcErr
+	}
+
+	h.mu.Lock()
+	h.subscriptions[id] = sub
+	h.mu.Unlock()
+
+	statusSub := h.txnStatus.Subscribe()
```
IronGauntlets, line 139: No one is sending on this subscription nor is anyone reading from it so why is this required?

rpc/events.go
```diff
@@ -100,6 +100,87 @@ func (h *Handler) SubscribeNewHeads(ctx context.Context) (uint64, *jsonrpc.Error
...
+			h.log.Errorw("Error while sending Txn status", "txHash", txHash, "err", err)
+			return
+		}
+		lastSendStatus = lastKnownStatus
+
+		for {
+			select {
+			case <-subscriptionCtx.Done():
+				return
+			case <-headerSub.Recv():
```
IronGauntlets, line 158: HeaderSub is not enough to get all of the transaction status in particular AcceptedOnL1 since that is updated in a separate go routine.

pnowosie: So header Recv here is more than a timer tick than taking the Tx status.
In other words:
- I'm getting the current status of Tx (the way rpc method does)
- I should check it again in a while to send an eventual update
  - I can wait some arbitrary time period
  - I can wait until next block which more likely brings some changes (but not in all cases as you mentioned)
- I'm getting the current status of Tx (the way rpc method does)

rpc/events.go
pnowosie: Good catch 👍

IronGauntlets (approved):

IronGauntlets:
I have approved the pr because 2 people worked on it, therefore we only need one more approval.

weiihann (approved):

PR #2211, opened 2024-10-11
Implement `subscribeNewHeads`,`subscriptionReorg`,`subscribePendingTransactions`

### LINKED ISSUE #2205: Implement starknet_subscribeNewHeads websocket method
See https://github.com/starkware-libs/starknet-specs/blob/v0.8.0-rc0/api/starknet_ws_api.json

### LINKED ISSUE #2207: Implement starknet_subscriptionNewHeads websocket method
See https://github.com/starkware-libs/starknet-specs/blob/v0.8.0-rc0/api/starknet_ws_api.json

### LINKED ISSUE #2232: Implement starknet_subscribePendingTransactions websocket method
See the spec - https://github.com/starkware-libs/starknet-specs/blob/v0.8.0-rc0/api/starknet_ws_api.json#L176

weiihann:
### Description
Fixes #2205 
Fixes #2207
Fixes #2232

The difference between the previous implementation of `juno_subscribeNewHeads` is that it supports streaming historical block headers up to 1024 blocks ago, as per the specification.


rpc/events.go
```diff
@@ -44,14 +50,27 @@ type EventsChunk struct {
 	ContinuationToken string          `json:"continuation_token,omitempty"`
 }
 
+type SubscriptionID struct {
+	ID uint64 `json:"subscription_id"`
+}
+
+func (s *SubscriptionID) UnmarshalJSON(data []byte) error {
```
kirugan, line 57: why do we need custom `UnmarshalJSON` ?

weiihann: We need subscription id as a param to `SubscriptionNewHeads`

weiihann: deleted

jsonrpc/server.go
```diff
@@ -426,6 +426,19 @@ func isNil(i any) bool {
 	return i == nil || reflect.ValueOf(i).IsNil()
 }
 
+func isNilOrEmpty(i any) (bool, error) {
+	if isNil(i) {
+		return true, nil
+	}
+
+	switch reflect.TypeOf(i).Kind() {
+	case reflect.Slice, reflect.Array, reflect.Map:
```
pnowosie, line 435: Should we support string type as well 🤔 

Also - wdyt to expose it & `isNil()` publicly from `utils`?

sync/sync.go
```diff
@@ -381,6 +422,19 @@ func (s *Synchronizer) revertHead(forkBlock *core.Block) {
...
+
+	if s.currReorg == nil { // first block of the reorg
+		s.currReorg = &ReorgData{
+			StartBlockHash: localHead,
+			StartBlockNum:  head.Number,
+			EndBlockHash:   localHead,
+			EndBlockNum:    head.Number,
+		}
+	} else { // not the first block of the reorg, adjust the starting block
+		s.currReorg.StartBlockHash = localHead
```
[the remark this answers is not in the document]
weiihann: Could you explain further? I didn't get what you meant.

jsonrpc/server.go
weiihann: Yeah we could, since `IsNil()` is used on other places

weiihann: > Should we support string type as well 🤔

No, A direct string value is not valid in `params`.

jsonrpc/server.go
```diff
@@ -498,7 +508,12 @@ func (s *Server) buildArguments(ctx context.Context, params any, method Method)
 		addContext = 1
 	}
 
-	if isNil(params) {
+	isNilOrEmpty, err := isNilOrEmpty(params)
```
kirugan, line 511: Why we started to check for emptiness ?

sync/sync.go
```diff
@@ -66,6 +76,26 @@ func (n *NoopSynchronizer) SubscribeNewHeads() HeaderSubscription {
...
+// ReorgData represents data about reorganised blocks, starting and ending block number and hash
+type ReorgData struct {
+	// StartBlockHash is the hash of the first known block of the orphaned chain
+	StartBlockHash *felt.Felt `json:"starting_block_hash"`
+	// StartBlockNum is the number of the first known block of the orphaned chain
+	StartBlockNum uint64 `json:"starting_block_number"`
+	// The last known block of the orphaned chain
+	EndBlockHash *felt.Felt `json:"ending_block_hash"`
+	// Number of the last known block of the orphaned chain
+	EndBlockNum uint64 `json:"ending_block_number"`
```
kirugan, line 96: I don't think it should be part of `sync/sync.go` package. Presence of json tags means it should be part of `rpc` package, but I don't think that it should be just referenced  here

rpc/subscriptions.go
```diff
@@ -182,3 +168,328 @@ func sendEvents(ctx context.Context, w jsonrpc.Conn, events []*blockchain.Filter
...
+				return
+			}
+		}
+	}
+}
+
+// filterTxs filters the transactions based on the getDetails flag.
+// If getDetails is true, response will contain the transaction details.
+// If getDetails is false, response will only contain the transaction hashes.
+func (h *Handler) filterTxs(pendingTxs []core.Transaction, getDetails bool, senderAddr []felt.Felt) interface{} {
```
kirugan, line 287: please change `interface{}` to `any`

rpc/subscriptions.go
```diff
@@ -182,3 +168,328 @@ func sendEvents(ctx context.Context, w jsonrpc.Conn, events []*blockchain.Filter
...
+		return nil, nil, ErrTooManyBlocksBack
+	}
+
+	return startHeader, latestHeader, nil
+}
+
+// sendHistoricalHeaders sends a range of headers from the start header until the latest header
+func (h *Handler) sendHistoricalHeaders(
+	ctx context.Context,
+	startHeader *core.Header,
```
kirugan, line 389: nit: `startHeader, latestHeader *core.Header`? up to you

jsonrpc/server.go
weiihann: My understanding of the RPC behavior is that if `params` do not exist (i.e. they are empty), then we use the default provided that the method params are all optional. 

For example:
`{"jsonrpc": "2.0", "method": "starknet_subscribeNewHeads", "id": 1}` <- this should work

But our current implementation doesn't check for empty so it will fail.

rpc/handlers.go
```diff
@@ -66,6 +66,7 @@ var (
 	ErrUnsupportedTxVersion            = &jsonrpc.Error{Code: 61, Message: "the transaction version is not supported"}
 	ErrUnsupportedContractClassVersion = &jsonrpc.Error{Code: 62, Message: "the contract class version is not supported"}
 	ErrUnexpectedError                 = &jsonrpc.Error{Code: 63, Message: "An unexpected error occurred"}
+	ErrTooManyAddressesInFilter        = &jsonrpc.Error{Code: 67, Message: "Too many addresses in filter sender_address filter"}
```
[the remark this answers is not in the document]
weiihann: Yeah I followed the specs

sync/sync.go
weiihann: I created a new `ReorgEvent` struct in the `rpc` package with the JSON tags, do check again

jsonrpc/server.go
kirugan: makes sense, thanks

sync/sync.go
kirugan: so we don't need these json tags here anymore, right ?

weiihann: yup, just removed

kirugan (approved):

jsonrpc/server.go
```diff
@@ -422,8 +422,17 @@ func isBatch(reader *bufio.Reader) bool {
...
+	if utils.IsNil(i) {
+		return true, nil
+	}
+
+	switch reflect.TypeOf(i).Kind() {
+	case reflect.Slice, reflect.Array, reflect.Map:
+		return reflect.ValueOf(i).Len() == 0, nil
+	default:
+		return false, fmt.Errorf("impossible param type: check request.isSane")
+	}
```
IronGauntlets, line 435: It would be great if you could add a test for this case. I think you already describe an example for it below.

rpc/subscriptions.go
```diff
@@ -182,3 +168,339 @@ func sendEvents(ctx context.Context, w jsonrpc.Conn, events []*blockchain.Filter
...
+			pendingTxsSub.Unsubscribe()
+		}()
+
+		h.processPendingTxs(subscriptionCtx, getDetails != nil && *getDetails, senderAddr, pendingTxsSub, w, id)
+	})
+
+	return &SubscriptionID{ID: id}, nil
+}
+
+func (h *Handler) processPendingTxs(
```
IronGauntlets, line 262: nit: fit as many on single line as possible before wrapping at 120 characters?

rpc/subscriptions.go
```diff
@@ -182,3 +168,339 @@ func sendEvents(ctx context.Context, w jsonrpc.Conn, events []*blockchain.Filter
...
+			if t.SenderAddress.Equal(&addr) {
+				return true
+			}
+		}
+	}
+
+	return false
+}
+
+func (h *Handler) sendPendingTxs(w jsonrpc.Conn, result interface{}, id uint64) error {
```
IronGauntlets, line 341:
```suggestion
func (h *Handler) sendPendingTxs(w jsonrpc.Conn, result any, id uint64) error {
```

sync/sync.go
```diff
@@ -66,6 +76,26 @@ func (n *NoopSynchronizer) SubscribeNewHeads() HeaderSubscription {
...
+type ReorgBlockRange struct {
+	// StartBlockHash is the hash of the first known block of the orphaned chain
+	StartBlockHash *felt.Felt
+	// StartBlockNum is the number of the first known block of the orphaned chain
+	StartBlockNum uint64
+	// The last known block of the orphaned chain
+	EndBlockHash *felt.Felt
+	// Number of the last known block of the orphaned chain
+	EndBlockNum uint64
+}
```
IronGauntlets, line 97: nit: I think due to rebase this code ended up between NoopSynchronizer's methods please move it above or below.

sync/sync.go
```diff
@@ -494,6 +548,9 @@ func (s *Synchronizer) fetchAndStorePending(ctx context.Context) error {
 		return err
 	}
 
+	// send the pending transactions to the feed
+	s.pendingTxsFeed.Send(pendingBlock.Transactions)
```
IronGauntlets, line 552: I don't think this is the correct place to send the pending transactions, please move it below line 618. This is because polling pending block can return the same block multiple times and we don't want to send the same set of transactions to the user multiple times. Only in the `StorePeniding()` do we determine to update the pending block hence base on that decision we should send the pending transactions.

rpc/subscriptions_test.go
```diff
@@ -326,7 +330,533 @@ func TestSubscribeEvents(t *testing.T) {
 	})
 }
 
-func marshalSubscriptionResponse(e *EmittedEvent, id uint64) ([]byte, error) {
+type fakeSyncer struct {
+	newHeads   *feed.Feed[*core.Header]
+	reorgs     *feed.Feed[*sync.ReorgBlockRange]
+	pendingTxs *feed.Feed[[]core.Transaction]
+}
```
IronGauntlets, line 337: I am a little confused why we have this? Is mockSyncReader not sufficient?

IronGauntlets: Please ignore this, I realised the MockSyncReader isn't enough to send values through the channel.

rpc/subscriptions_test.go
```diff
@@ -326,7 +330,533 @@ func TestSubscribeEvents(t *testing.T) {
...
+	t.Cleanup(mockCtrl.Finish)
+
+	mockChain := mocks.NewMockReader(mockCtrl)
+	syncer := newFakeSyncer()
+	handler, server := setupRPC(t, ctx, mockChain, syncer)
+
+	testCases := []struct {
+		name            string
+		subscribeMethod string
+		ignoreFirst     bool
```
IronGauntlets, line 593: I am a little confused about this. Could you explain when you would want to `ignoreFirst` and when you wouldn't?

rpc/subscriptions_test.go
```diff
@@ -326,7 +330,533 @@ func TestSubscribeEvents(t *testing.T) {
...
+
+			id := uint64(1)
+			handler.WithIDGen(func() uint64 { return id })
+
+			got := sendWsMessage(t, ctx, conn, subMsg(tc.subscribeMethod))
+			require.Equal(t, subResp(id), got)
+
+			if tc.ignoreFirst {
+				_, _, err := conn.Read(ctx)
+				require.NoError(t, err)
```
IronGauntlets, line 622: Do you think just checking for no error is meaningful enough?

sync/sync_test.go
```diff
@@ -197,102 +231,27 @@ func TestSubscribeNewHeads(t *testing.T) {
...
 	cancel()
 
-	head, err := bc.HeadsHeader()
-	require.NoError(t, err)
 	pending, err := synchronizer.Pending()
 	require.NoError(t, err)
-	assert.Equal(t, head.Hash, pending.Block.ParentHash)
-}
-
-func TestPending(t *testing.T) {
```
IronGauntlets, line 223: Why remove the subtest?

rpc/subscriptions_test.go
weiihann: In `subscribeNewHeads`, the server will also send the latest header stored in the db (done via `sendHistoricalHeaders()`), while processing the new headers sent through the feed. Because this test case wants to test if the headers sent through the feed is working or not, that's why I had an `ignoreFirst` flag.

In `subscribeEvents`, this doesn't happen if I understand correctly.

rpc/subscriptions_test.go
weiihann: Yes because I just want to test the header sent through the feed, so the first response received can be ignored.

sync/sync_test.go
weiihann: Do you mean `TestPending`? It wasn't removed, it was moved to a [separate location](https://github.com/NethermindEth/juno/blob/9104a8cc3d06d6a3196796905143763bd1a6d31a/sync/sync_test.go#L188)

sync/sync_test.go
```diff
@@ -197,102 +231,27 @@ func TestSubscribeNewHeads(t *testing.T) {
...
-		assert.Equal(t, expectedPending, gotPending)
-	})
-
-	t.Run("get pending state", func(t *testing.T) {
-		_, pendingStateCloser, pErr := synchronizer.PendingState()
-		t.Cleanup(func() {
-			require.NoError(t, pendingStateCloser())
-		})
-		require.NoError(t, pErr)
-	})
```
IronGauntlets, line 297: Sorry I meant why these cases were removed?

jsonrpc/server_test.go
```diff
@@ -475,6 +482,10 @@ func TestHandle(t *testing.T) {
 			res:              `{"jsonrpc":"2.0","error":{"code":-32603,"message":"Internal error"},"id":1}`,
 			checkFailedEvent: true,
 		},
+		"empty optional param": {
+			req: `{"jsonrpc": "2.0", "method": "emptyOptionalParam", "params": {}, "id": 1}`,
+			res: `{"jsonrpc":"2.0","result":0,"id":1}`,
+		},
```
IronGauntlets, line 488: Sorry to be pedantic. I thought we wanted to check for the following case:

`req: {"jsonrpc": "2.0", "method": "emptyOptionalParam", "id": 1},`, right?

It would be good to add this too.

sync/sync_test.go
weiihann: My bad... That was accidental, added them back

rpc/subscriptions_test.go
IronGauntlets: > In subscribeEvents, this doesn't happen if I understand correctly.

Well, this depends on the filter. If the filter is empty, then all the historical and/or current block/s events are streamed to the client; if it is not nil, only the events that match the filter are. In this case, since the header is empty, which makes it EventBloom nil indicating that the block has no events, no such first message is sent.

I think it is fine as is, no change is required.

IronGauntlets (approved):

kirugan (approved):

PR #2217, opened 2024-10-15
Update starknet.go, starknet-js and starknet-rs tests

wojciechos:


.github/workflows/starknet-go-tests.yml
```diff
@@ -6,7 +6,7 @@ on:
       ref:
         description: 'The branch, tag or SHA to checkout'
         required: false
-        default: 'v0.7.2'
+        default: '593313d1eda42a8e4b428652bddb5d77fc62faf3'
```
kirugan, line 9: is it because commit is more reliable?

kirugan (approved):

.github/workflows/starknet-go-tests.yml
wojciechos: We need Starknet.go with this fix: https://github.com/NethermindEth/starknet.go/pull/639. However, it won’t be merged until the RPC SaaS endpoint is updated - so for now I just hardcoded the commit

derrix060 (approved):

PR #2219, opened 2024-10-16
fix txn status bug - fgw can return not_recieved

rianhughes:
closes https://github.com/NethermindEth/juno/issues/2131


rianhughes:
Nvm, we actually handle this correctly.

Issue #2218, opened 2024-10-16
Missing Root-level execution_resources in Trace Response


wojciechos:
The RPC method for traces does not return the root-level execution_resources in the response, [see spec](https://github.com/starkware-libs/starknet-specs/blob/v0.7.1/api/starknet_trace_api_openrpc.json#L239).

**Steps to Reproduce**:
1. Make a request to Sepolia using the following:
   ```json
   {
     "jsonrpc": "2.0",
     "method": "starknet_traceTransaction",
     "params": ["01936a09e5aaee208fc0f7cc826e126d421c3ac9aca2c789605e1e919e399185"],
     "id": 1
   }

2. In the response, the **execution_resources** field is missing at the root level.

**Expected Behavior:**
	•	The root-level execution_resources should be included in the trace response.
	•	Checked against Pathfinder, which returns the field correctly.

PR #2222, opened 2024-10-16
Add missing execution resources to fetched traces

kirugan:


rpc/trace.go
```diff
@@ -215,6 +215,20 @@ func (h *Handler) traceBlockTransactions(ctx context.Context, block *core.Block,
...
+			if err != nil {
+				return nil, httpHeader, err
+			}
+
+			if !v0_6Response {
+				// add execution resources on root level
+				for index, trace := range result {
+					executionResources := trace.TraceRoot.TotalExecutionResources()
+					// fgw doesn't provide this data in traces endpoint
+					executionResources.DataAvailability = &vm.DataAvailability{}
```
rianhughes, line 227: I guess DataAvailability shouldn't be zero. I think we can fetch it from the Receipts

rianhughes:
closes https://github.com/NethermindEth/juno/issues/2218

rpc/trace.go
kirugan: As you suggested I've added data availability from receipts and it works for some blocks (for very old ones like sepolia #7 it doesn't work)

rianhughes:
LGTM, I'll approve when the tests are fixed

IronGauntlets (approved):

rianhughes (approved):

PR #2223, opened 2024-10-17
Fix upper bound issue in `CalculatePrefixSize`

weiihann:
### Description
This PR modifies the `db size` command to also check for leftover key-value pairs in the database. It also fixes an upper bound bug as the `UpperBound` iterator option is exclusive, not inclusive.

### DIFF
--- a/db/pebble/db.go
+++ b/db/pebble/db.go
@@ -17,6 +17,7 @@ const (
 	// minCache is the minimum amount of memory in megabytes to allocate to pebble read and write caching.
 	// This is also pebble's default value.
 	minCacheSizeMB = 8
+	byteLimit      = 0xff
 )
 
 var (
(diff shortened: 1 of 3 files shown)


wojciechos (approved):

db/pebble/db.go
```diff
@@ -17,6 +17,7 @@ const (
 	// minCache is the minimum amount of memory in megabytes to allocate to pebble read and write caching.
 	// This is also pebble's default value.
 	minCacheSizeMB = 8
+	byteLimit      = 0xff
```
pnowosie, line 20: IMO better describes / more  idiomatic - but subjective. Feel free to ignore
```suggestion
	maxByte = ^byte(0)
```

pnowosie (approved):
Small hint

PR #2224, opened 2024-10-17
Refactor state and contract modules

weiihann:
### Description
This PR has 2 major changes:

1. Combine `ContractNonce`, `ContractClassHash`, and `ContractDeploymentHeight` into a single bucket called `Contract`.
Separation of the contract fields in different buckets creates data duplication. Instead, we can encode them into a single object and store them in a single bucket. The slight performance degradation of encoding/decoding is insignificant as the fields are small. 
Migration codes are added along with the unit tests.

2. Refactor `State.Update`, `State.Revert` and the removal of `history`.
In the previous implementation, after applying each state diff (i.e. nonce, class hash, storage), the contract commitment is recalculated. This indicates that the higher the number of individual state diffs, the more times we need to recalculate the contract commitment. 
In this new implementation, a contract's commitment is calculated only once after all state diffs are applied to a contract object. Now, the number of contract commitment calculations depends on the number of contracts updated.
 

### Benchmark Results
```
goos: darwin
goarch: arm64
pkg: github.com/NethermindEth/juno/core
cpu: Apple M3 Pro
               │   new.txt   │               old.txt               │
               │   sec/op    │   sec/op     vs base                │
StateUpdate-11   3.316m ± 2%   4.146m ± 2%  +25.03% (p=0.000 n=10)

               │   new.txt    │               old.txt                │
               │     B/op     │     B/op      vs base                │
StateUpdate-11   144.1Ki ± 0%   182.4Ki ± 0%  +26.53% (p=0.000 n=10)

               │   new.txt   │               old.txt               │
               │  allocs/op  │  allocs/op   vs base                │
StateUpdate-11   3.037k ± 0%   4.056k ± 0%  +33.54% (p=0.000 n=10)
```


weiihann:
Latest changes [here](https://github.com/NethermindEth/juno/pull/2535)

PR #2226, opened 2024-10-18
Create trivy.yml

yevh:
## Changes

Trivy is added. Trivy is a lightweight image scanning tool that may discover vulnerabilities and misconfigurations. Actions pinned to specific commit hashes:

- trivy-action@5681af892cd0f4997658e2bacc62bd0a894cf564 #v0.27.0
- checkout@d632683dd7b4114ad314bca15554477dd762a938 #v4.2.0
- codeql-action/init@5618c9fc1e675841ca52c1c6b1304f5255a905a0 #v2.19.0
- setup-dotnet@6bd8b7f7774af54e05809fcc5431931b3eb1ddee #v4.0.1

### DIFF
--- a/.github/workflows/trivy.yml
+++ b/.github/workflows/trivy.yml
@@ -18,7 +18,7 @@ jobs:
       security-events: write 
       actions: read
     name: Build
-    runs-on: "ubuntu-20.04"
+    runs-on: "ubuntu-latest"
     steps:
       - name: Checkout code
         uses: actions/checkout@eef61447b9ff4aafe5dcd4e0bbf5d482be7e7871 #v4.2.1
--- a/.github/workflows/trivy.yml
+++ b/.github/workflows/trivy.yml
@@ -30,7 +30,7 @@ jobs:
         uses: aquasecurity/trivy-action@5681af892cd0f4997658e2bacc62bd0a894cf564 #v0.27.0
         with:
           image-ref: 'docker.io/my-organization/my-app:${{ github.sha }}'
-          format: 'template'
+          format: 'sarif'
           template: '@/contrib/sarif.tpl'
           output: 'trivy-results.sarif'
           severity: 'CRITICAL,HIGH'
--- a/.github/workflows/trivy.yml
+++ b/.github/workflows/trivy.yml
@@ -1,0 +1,43 @@
+name: Container Security
+
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
+    branches: [ "main" ]
+  schedule:
+    - cron: '29 19 * * 4'
+
+permissions:
+  contents: read
+
+jobs:
+  build:
+    permissions:
+      contents: read 
+      security-events: write 
+      actions: read
+    name: Build
+    runs-on: "ubuntu-20.04"
+    steps:
+      - name: Checkout code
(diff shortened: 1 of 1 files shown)


.github/workflows/trivy.yml
```diff
@@ -0,0 +1,43 @@
...
+  contents: read
+
+jobs:
+  build:
+    permissions:
+      contents: read 
+      security-events: write 
+      actions: read
+    name: Build
+    runs-on: "ubuntu-20.04"
```
derrix060, line 21: that's quite old version, why not `ubuntu-latest`?

.github/workflows/trivy.yml
```diff
@@ -0,0 +1,43 @@
...
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@eef61447b9ff4aafe5dcd4e0bbf5d482be7e7871 #v4.2.1
+
+      - name: Build an image from Dockerfile
+        run: |
+          docker build -t docker.io/my-organization/my-app:${{ github.sha }} .
```
derrix060, line 28:
```suggestion
          docker build -t docker.io/nethermind/juno:${{ github.sha }} .
```

derrix060, line 28: why not using a image that get's properly built instead of building it again?

.github/workflows/trivy.yml
```diff
@@ -0,0 +1,43 @@
...
+      - name: Run Trivy vulnerability scanner
+        uses: aquasecurity/trivy-action@5681af892cd0f4997658e2bacc62bd0a894cf564 #v0.27.0
```
derrix060, line 30: 0.28.0 is the latest one, we should probably be using that instead

.github/workflows/trivy.yml
```diff
@@ -0,0 +1,43 @@
...
+        with:
+          image-ref: 'docker.io/my-organization/my-app:${{ github.sha }}'
+          format: 'template'
```
derrix060, line 33: format template is deprecated, I'd change it to be:
```suggestion
          format: 'sarif'
```

.github/workflows/trivy.yml
yevh: @derrix060 The main reason for building is to ensure transparency and to make sure the image is always built from the latest code. But this is also an option.

.github/workflows/trivy.yml
yevh: @derrix060 is indeed quite old, but this version is mentioned in all their action examples and is commonly used for this action, at least as far as I know. However, we can certainly try `ubuntu-latest`

.github/workflows/trivy.yml
derrix060: I think we could hook this step in the pipeline where we build the images, so before publishing the image we check that it's okay, what do you think?

.github/workflows/trivy.yml
yevh: @derrix060 for any security-related actions, in case we have an internal adaptation, please refer to the [trusted commit SHA](https://github.com/NethermindEth/trivy)

.github/workflows/trivy.yml
yevh: @derrix060 nice catch! Thank you a lot!

derrix060 (approved):

PR #2227, opened 2024-10-21
Override network config with the one in database

weiihann:
For an existing chain database, if a user provides a different network configuration via `--network`, automatically switches the network config that matches the one in the database.


weiihann:
it's better to explicitly tell users that they are wrong

PR #2228, opened 2024-10-21
Pass required libbz2 to linker to fix compilation error on macOS

kirugan:
For some reason macOS build started to failing. Seem like they update default linked libraries and now it's required to pass `libbz2`

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,3 +1,3 @@
 # do not copy local database files
 p2p-dbs/
-juon/
+juno/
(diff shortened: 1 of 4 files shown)


derrix060 (approved):

wojciechos (approved):

derrix060 (approved):
readme update can come later

kirugan:
> We need to add this as a dependency on Readme

It's not dependency per se because it's only required on macOS due to recent updates for linker. Initially I added corresponding flag in `Makefile` but that led to some warnings like `duplicate library provided` (compilation was successful though). So that's why I putted it in CGo declaration.

Maybe it was bad idea. wdyt?

Issue #2237, opened 2024-10-24
Update ResourceBounds to use spec.Uint128


derrix060:
Superseeding https://github.com/NethermindEth/juno/issues/1448, there is now a function that will create a uint128: https://github.com/NethermindEth/juno/blob/65c0d6c693d99b880a635a6ea7e6bbba09e27570/adapters/core2p2p/felt.go#L49

The goal for this ticket is to update the ResourceBounds in transaction.go to use that

RuneRogue:
I am a blockchain and DeFi developer . I am having experience of 1 year . I am currently working as a Blockchain Researcher and Developer. My Experience in programming language like Go , Rust , Java will help me to solve this issue.

PR #2242, opened 2024-10-27
Update ResourceBounds to use spec.Uint128

RuneRogue:
Fixing #2237

### DIFF
--- a/core/transaction.go
+++ b/core/transaction.go
@@ -64,12 +64,14 @@ type ResourceBounds struct {
 func (rb ResourceBounds) Bytes(resource Resource) []byte {
 	maxAmountBytes := make([]byte, 8) //nolint:mnd
 	binary.BigEndian.PutUint64(maxAmountBytes, rb.MaxAmount)
-	maxPriceBytes := rb.MaxPricePerUnit.Bytes()
+	maxPriceBytes := make([]byte, 16)
+	binary.BigEndian.PutUint64(maxPriceBytes[0:8], rb.MaxPricePerUnit.High)
+	binary.BigEndian.PutUint64(maxPriceBytes[8:16], rb.MaxPricePerUnit.Low)
 	return slices.Concat(
 		[]byte{0},
 		[]byte(resource.String()),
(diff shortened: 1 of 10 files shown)


RuneRogue:
I need your help . I was trying hard to solve some error but still I was not able to solve this error .
`Received unexpected error:
            	            	json: cannot unmarshal string into Go struct field ResourceBounds.transaction.resource_bounds.max_price_per_unit of type spec.Uint128`
            	            	
 I am getting this error during the test cover .

core/transaction.go
```diff
@@ -57,18 +58,20 @@ const (
...
 	// MaxPricePerUnit is technically a uint128
-	MaxPricePerUnit *felt.Felt
+	MaxPricePerUnit *spec.Uint128
 }
 
 func (rb ResourceBounds) Bytes(resource Resource) []byte {
 	maxAmountBytes := make([]byte, 8) //nolint:mnd
 	binary.BigEndian.PutUint64(maxAmountBytes, rb.MaxAmount)
-	maxPriceBytes := rb.MaxPricePerUnit.Bytes()
+	maxPriceBytes := make([]byte, 16) //nolint:mnd
```
derrix060, line 67: I've seen this logic before (https://github.com/NethermindEth/juno/pull/2242/files#diff-8fcd72f4c5e1f056a6875680f66dfbfc8578908ff4fe889c8f1fc21e94e9c8bbR157) please DRY and create a function to do these steps, properly unittested :)

RuneRogue: Made changes

RuneRogue:
> Screenshot 2024-10-27 203256 I need your help . I was trying hard to solve some error but still I was not able to solve this error . I am getting this error during the test .

want your help me to solve this issue .

pnowosie:
What is a reason to use p2p/spec/Uint128 in rpc and even worse in the core modules?

I'm against this change, because it puts additional hurdle to client libraries that now would need to handle non-standard integer type. Also p2p/spec/Uimt128 is protobuf-specific which should not be used in rpc imo.

PR #2245, opened 2024-10-31
Create a dynamic logger (Fixes #1181)

### LINKED ISSUE #1181: Dynamic logging lvl
## Issue
Currently, logging is only set at the bootstrap with the defined log level. It would be beneficial to have the capability to dynamically change that through RPC. Furthermore, for greater extendability, it would be valuable to make it configurable per module.

## Rationale

**Why should this feature exist?**
Enhanced debugging and observability. Often, managing nodes requires specific logging for a limited duration. Maintaining a low level for the entire node lifetime is unnecessary. 
Additionally, some loggers in the codebase have hardcoded lvl such as the one in DB.

**What are the use cases?**
The feature would be particularly useful for debugging modules eg. enabling more verbose logs for the p2p module while keeping the rest of the loggers unchanged or temporarily lowering the global logging level.

## Implementation
Do you have ideas regarding the implementation of this feature? yes
Are you willing to implement this feature? yes

derrix060:
See commit messages for more information

### DIFF
--- a/cmd/juno/juno.go
+++ b/cmd/juno/juno.go
@@ -84,6 +84,8 @@ const (
 	corsEnableF             = "rpc-cors-enable"
 	versionedConstantsFileF = "versioned-constants-file"
 	pluginPathF             = "plugin-path"
+	logHostF                = "log-host"
+	logPortF                = "log-port"
 
 	defaultConfig                   = ""
 	defaulHost                      = "localhost"
@@ -123,6 +125,7 @@ const (
 	defaultCorsEnable               = false
 	defaultVersionedConstantsFile   = ""
 	defaultPluginPath               = ""
+	defaultLogPort                  = 0
 
 	configFlagUsage                       = "The YAML configuration file."
 	logLevelFlagUsage                     = "Options: trace, debug, info, warn, error."
@@ -176,6 +179,8 @@ const (
 	corsEnableUsage             = "Enable CORS on RPC endpoints"
 	versionedConstantsFileUsage = "Use custom versioned constants from provided file"
 	pluginPathUsage             = "Path to the plugin .so file"
+	logHostUsage                = "The interface on which the log level HTTP server will listen for requests."
+	logPortUsage                = "The port on which the log level HTTP server will listen for requests."
 )
 
 var Version string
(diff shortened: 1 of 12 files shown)


cmd/juno/juno.go
```diff
@@ -123,6 +125,7 @@ const (
 	defaultCorsEnable               = false
 	defaultVersionedConstantsFile   = ""
 	defaultPluginPath               = ""
+	defaultLogPort                  = 0
```
wojciechos, line 128: What do you think about setting it e.g. to 9091 by default instead of random? I guess it would keep things more predictable.

cmd/juno/juno.go
```diff
@@ -176,6 +179,8 @@ const (
 	corsEnableUsage             = "Enable CORS on RPC endpoints"
 	versionedConstantsFileUsage = "Use custom versioned constants from provided file"
 	pluginPathUsage             = "Path to the plugin .so file"
+	logHostUsage                = "The interface on which the log level HTTP server will listen for requests."
```
wojciechos, line 182: Should we add the logUsage flag to keep it consistent with the rest of the configuration? Like metricsUsage, grpcUsage, or httpUsage?

cmd/juno/juno.go
derrix060: 0 doesn't mean random here, it acutally means that it's disabled :)

cmd/juno/juno.go
```diff
@@ -84,6 +84,8 @@ const (
 	corsEnableF             = "rpc-cors-enable"
 	versionedConstantsFileF = "versioned-constants-file"
 	pluginPathF             = "plugin-path"
```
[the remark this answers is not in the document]
derrix060: wdym?  I've added this `logHostF` and `logPortF`

Do you mean a boolean flag to enable/disable the log host?

cmd/juno/juno.go
derrix060: See https://github.com/NethermindEth/juno/pull/2245/files/55eb191d431b1e1e536555cae68c7b49557545cf#diff-375d57e386f20eaa5f09f02bb9d28bfc48ac3dca18d0325f59492208219e5618R240

pnowosie (approved):
Nice one 👍

cmd/juno/juno.go
wojciechos: ohh, gotcha, thanks

cmd/juno/juno.go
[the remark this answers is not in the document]
wojciechos: Yeep, but indeed it might be a little bit redundant, but the rest of the configuration includes it. Perhaps it’s better to simplify the others instead

PR #2249, opened 2024-11-04
Make eth-node flag as required

kirugan:


node/node.go
```diff
@@ -293,26 +298,26 @@ func New(cfg *Config, version string) (*Node, error) { //nolint:gocyclo,funlen
 	return n, nil
 }
 
-func newL1Client(cfg *Config, chain *blockchain.Blockchain, log utils.SimpleLogger) (*l1.Client, error) {
```
derrix060, line 296: +1 to stop passing the whole config :)

derrix060 (approved):

cmd/juno/juno.go
```diff
@@ -327,6 +330,7 @@ func NewCmd(config *node.Config, run func(*cobra.Command, []string) error) *cobr
 	junoCmd.Flags().String(cnCoreContractAddressF, defaultCNCoreContractAddressStr, networkCustomCoreContractAddressUsage)
 	junoCmd.Flags().IntSlice(cnUnverifiableRangeF, defaultCNUnverifiableRange, networkCustomUnverifiableRange)
 	junoCmd.Flags().String(ethNodeF, defaultEthNode, ethNodeUsage)
+	junoCmd.Flags().Bool(noEthNodeF, defaultNoEthNode, noEthNodeUsage)
```
IronGauntlets, line 333: You should make `no-eth-node` and `eth-node` mutually exclusive so that both cannot be passed at the same time. Otherwise, some can pass both these flags which may lead to unexpected behaviour in node.go.

wojciechos (approved):

node/node.go
```diff
@@ -273,10 +274,14 @@ func New(cfg *Config, version string) (*Node, error) { //nolint:gocyclo,funlen
 	}
 
 	if n.cfg.EthNode == "" {
-		n.log.Warnw("Ethereum node address not found; will not verify against L1")
+		if n.cfg.NoEthNode {
+			n.log.Warnw("Ethereum node address not found; will not verify against L1")
+		} else {
+			return nil, fmt.Errorf("--eth-node flag is required, you can use --no-eth-node to disable L1 verification")
+		}
```
IronGauntlets, line 281: I think we can simplify this if condition as follows:
```go

if !n.cfg.NoEthNode{
  if n.cfg.EthNode == "" {
    return nil, fmt.Errorf("ethereum node address not found; will not verify against L1. Use --no-eth-node flag if L1 verification is not required")
  }
  // Create L1Client
}

```

cmd/juno/juno.go
```diff
@@ -47,6 +47,7 @@ const (
 	dbPathF                 = "db-path"
 	networkF                = "network"
 	ethNodeF                = "eth-node"
+	noEthNodeF              = "no-eth-node"
```
IronGauntlets, line 50: Let rename it `disable-l1-verfication`.

cmd/juno/juno.go
```diff
@@ -145,6 +147,7 @@ const (
 	colourUsage                           = "Use `--colour=false` command to disable colourized outputs (ANSI Escape Codes)."
 	ethNodeUsage                          = "WebSocket endpoint of the Ethereum node. To verify the correctness of the L2 chain, " +
 		"Juno must connect to an Ethereum node and parse events in the Starknet contract."
+	noEthNodeUsage           = "Disables the connection to the Ethereum node."
```
IronGauntlets, line 150: The description should be: "L1 verification is disabled since an Ethereum node is not provided."

IronGauntlets (approved):
I am approving since the changes requested are simple.

cmd/juno/juno.go
kirugan: Agree, added

cmd/juno/juno.go
kirugan: renamed

rianhughes (approved):

PR #2250, opened 2024-11-04
Fixes getStorageAt undeployed contract response

weiihann:
`Trie.Get` by default returns no error and zero value for non-existent key. To check if a contract is deployed, we have no choice by to check if the nonce also exists (which is a redundant DB IO).

### DIFF
--- a/rpc/contract.go
+++ b/rpc/contract.go
@@ -1,7 +1,10 @@
 package rpc
 
 import (
+	"errors"
+
 	"github.com/NethermindEth/juno/core/felt"
+	"github.com/NethermindEth/juno/db"
 	"github.com/NethermindEth/juno/jsonrpc"
 )
 
@@ -43,7 +46,11 @@ func (h *Handler) StorageAt(address, key felt.Felt, id BlockID) (*felt.Felt, *js
 	// the returned value is always zero and error is nil.
 	_, err := stateReader.ContractNonce(&address)
 	if err != nil {
-		return nil, ErrContractNotFound
+		if errors.Is(err, db.ErrKeyNotFound) {
+			return nil, ErrContractNotFound
+		}
+		h.log.Errorw("Failed to get contract nonce", "err", err)
+		return nil, ErrInternal
 	}
 
 	value, err := stateReader.ContractStorage(&address, &key)
--- a/rpc/contract.go
+++ b/rpc/contract.go
@@ -39,6 +39,13 @@ func (h *Handler) StorageAt(address, key felt.Felt, id BlockID) (*felt.Felt, *js
 	}
 	defer h.callAndLogErr(stateCloser, "Error closing state reader in getStorageAt")
 
+	// This checks if the contract exists because if a key doesn't exist in contract storage,
+	// the returned value is always zero and error is nil.
+	_, err := stateReader.ContractNonce(&address)
+	if err != nil {
+		return nil, ErrContractNotFound
+	}
+
 	value, err := stateReader.ContractStorage(&address, &key)
 	if err != nil {
 		return nil, ErrContractNotFound
--- a/rpc/contract.go
+++ b/rpc/contract.go
@@ -44,7 +44,7 @@ func (h *Handler) StorageAt(address, key felt.Felt, id BlockID) (*felt.Felt, *js
 
 	// This checks if the contract exists because if a key doesn't exist in contract storage,
 	// the returned value is always zero and error is nil.
-	_, err := stateReader.ContractNonce(&address)
+	_, err := stateReader.ContractClassHash(&address)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


core/trie/trie.go
```diff
@@ -195,8 +195,8 @@ func (t *Trie) Get(key *felt.Felt) (*felt.Felt, error) {
 	storageKey := t.feltToKey(key)
 	value, err := t.storage.Get(&storageKey)
 	if err != nil {
-		if errors.Is(err, db.ErrKeyNotFound) {
-			return &felt.Zero, nil
+		if errors.Is(err, db.ErrKeyNotFound) { // a non-existent key returns a zero value
+			return &felt.Zero, db.ErrKeyNotFound
```
kirugan, line 199: why do we need to pass `felt.Zero` in this case?

weiihann: As mentioned in the comment, a key that does not exists in the trie by default has a zero value

kirugan: yes, but I'm wondering how useful it would be if usual pattern is:
```
v, err := func()
if err != nil 
     return err
}
```
since we started to return `db.ErrKeyNotFound`  error

rpc/contract.go
```diff
@@ -39,6 +39,13 @@ func (h *Handler) StorageAt(address, key felt.Felt, id BlockID) (*felt.Felt, *js
 	}
 	defer h.callAndLogErr(stateCloser, "Error closing state reader in getStorageAt")
 
+	// This checks if the contract exists because if a key doesn't exist in contract storage,
+	// the returned value is always zero and error is nil.
+	_, err := stateReader.ContractNonce(&address)
+	if err != nil {
```
kirugan, line 45: We already discussed it in slack, but I think it's worth to discuss it here again:
do you think that we should handle non ErrKeyNotFound errors differently? i.e.:
```
_, err := stateReader.ContractNonce(&address)
if err != nil {
       return nil, ErrContractNotFound
   } else {
       s.log.Errorw("Failed to get contract nonce", "err", err)
       return nil, ErrInternal
   }
}
```
?

kirugan (approved):

rpc/contract.go
weiihann: Yeah this is better, makes it more robust.

rpc/contract.go
```diff
@@ -39,6 +42,17 @@ func (h *Handler) StorageAt(address, key felt.Felt, id BlockID) (*felt.Felt, *js
...
+		if errors.Is(err, db.ErrKeyNotFound) {
+			return nil, ErrContractNotFound
+		}
+		h.log.Errorw("Failed to get contract nonce", "err", err)
```
IronGauntlets, line 52: As far as i remember logging an Error would cause Juno to shut down, is that the behaviour you want?

IronGauntlets (commented):
It seems to me that the problem you are describing would be better solved by returning an error from `stateReader.ContractStorage` when the key doesn't exist, right?

weiihann:
> It seems to me that the problem you are describing would be better solved by returning an error from `stateReader.ContractStorage` when the key doesn't exist, right?

If the key doesn't exist, it will return a zero value for `ContractStorage` (see `Trie.Get`). This behavior of `Trie` is correct because a non-existent key by default has a zero value, so we shouldn't modify it to return the key not found error.

The problem is that we need a way to check if a contract exists. Since we put contract fields (i.e. nonce, balance, storage) in different buckets, we have no choice but to add an additional DB lookup (in this case using `ContractNonce`) to check if a contract is deployed or not.

If we condensed the contract fields into a single bucket (see #2224), we only need to do a single DB lookup to check if a contract exists or not.

IronGauntlets (approved):

rpc/contract.go
```diff
@@ -39,6 +42,17 @@ func (h *Handler) StorageAt(address, key felt.Felt, id BlockID) (*felt.Felt, *js
 	}
 	defer h.callAndLogErr(stateCloser, "Error closing state reader in getStorageAt")
 
+	// This checks if the contract exists because if a key doesn't exist in contract storage,
+	// the returned value is always zero and error is nil.
+	_, err := stateReader.ContractNonce(&address)
```
IronGauntlets, line 47: I think it would be better to use `ContractClassHash` to be more explicit as this is what we use in the [`deployed()`](https://github.com/NethermindEth/juno/blob/e4240acb490f6f213193459de66182ca24ff2810/core/contract.go#L82)

IronGauntlets (changes requested):

rpc/contract.go
weiihann: changed

IronGauntlets (approved):

PR #2252, opened 2024-11-05
Fix and refactor trie proof logics

weiihann:
This PR primarily does the following:
- Refactor `VerifyProof` as it fails on some edge cases
- Refactor `VerifyRangeProof` to cover most edge cases

A new data structure `OrderedSet` is used for proof generation and verification. The previous implementation used slices to contain proof nodes, which introduced complexity, unnecessary duplication, and inefficiency. The `OrderedSet` structure acts as a key-value database, simplifying the proof logic and improving readability.

`VerifyRangeProof` is not fully completed. For some specific missing key in the range (i.e. gapped range), the function would deem the edge case as valid but it's not supposed to be. The root cause still needs further investigation, but it's most likely due to the `Trie` structure and how we perform the root hash calculation. A separate PR will be raised to tackle it.

_Credits: some implementation details are based on go-ethereum_


core/trie/proof.go
```diff
@@ -54,623 +55,586 @@ func (e *Edge) Len() uint8 {
...
+// Both of them can be existent or non-existent.
+// This function also handles the following special cases:
+//
+//   - All elements proof: In this case the proof can be nil, but the range should be all the leaves in the trie.
+//   - One element proof: In this case, both left and right edge proofs are the same and the range should contain only one element.
+//   - Zero element proof: In this case a single edge proof is enough to verify. If there are still more elements, the proof is invalid.
+//
+// If the range proof is invalid, this function returns false and an error.
+// Otherwise, it will return a flag to indicate if there are more elements.
+//
```
pnowosie, line 197:
```suggestion
// VerifyRangeProof checks the validity of given key-value pairs and range proof against a provided root hash.
// The key-value pairs should be consecutive (no gaps) and monotonically increasing.
// The range proof contains two edge proofs: one for the first key and another for the last key.
// Both edge proofs can be for existent or non-existent keys.
// This function handles the following special cases:
//
//   - All elements proof: The proof can be nil if the range includes all leaves in the trie.
//   - Single element proof: Both left and right edge proofs are identical, and the range contains only one element.
//   - Zero element proof: A single edge proof suffices for verification. The proof is invalid if there are additional elements.
//
// The function returns a boolean indicating if there are more elements and an error if the range proof is invalid.
//
```
Please double-check whether I am not going too far here.

core/trie/proof.go
```diff
@@ -54,623 +55,586 @@ func (e *Edge) Len() uint8 {
...
+// TODO(weiihann): Given a binary leaf and a left-sibling first key, if the right sibling is removed, the proof would still be valid.
+// Conversely, given a binary leaf and a right-sibling last key, if the left sibling is removed, the proof would still be valid.
+// Range proof should not be valid for both of these cases, but currently is, which is an attack vector.
+// The problem probably lies in how we do root hash calculation.
+func VerifyRangeProof(root, first *felt.Felt, keys, values []*felt.Felt, proof *ProofNodeSet) (bool, error) { //nolint:funlen,gocyclo
+	// Ensure the number of keys and values are the same
+	if len(keys) != len(values) {
+		return false, fmt.Errorf("inconsistent proof data, keys: %d, values: %d", len(keys), len(values))
```
pnowosie, line 205:
```suggestion
		return false, fmt.Errorf("inconsistent length of proof data, keys: %d, values: %d", len(keys), len(values))
```

core/trie/proof.go
```diff
@@ -54,623 +55,586 @@ func (e *Edge) Len() uint8 {
...
+		if i < len(keys)-1 && keys[i].Cmp(keys[i+1]) > 0 {
+			return false, errors.New("keys are not monotonic increasing")
+		}
 
-		if leftNode.Hash(hash).Equal(rightNode.Hash(hash)) {
-			merged = append(merged, leftNode)
-		} else {
-			break
+		if values[i] == nil || values[i].Equal(&felt.Zero) {
+			return false, errors.New("range contains deletion")
```
pnowosie, line 215:
```suggestion
			return false, errors.New("range contains empty leaf")
```
Yes, might be a result of the deletion, but let's be more straight

core/trie/proof.go
```diff
@@ -54,623 +55,586 @@ func (e *Edge) Len() uint8 {
...
-	rightPath := []ProofNode{}
-	nodeHashes := make(map[felt.Felt]ProofNode)
-
-	for _, node := range mergedPath {
-		nodeHash := node.Hash(hash)
-		_, nodeExists := nodeHashes[*nodeHash]
-
-		if nodeExists {
-			return leftPath, rightPath, errors.New("duplicate node in the merged path")
+	// Special case: no edge proof at all, given range is the whole leaf set in the trie
```
pnowosie, line 219:
```suggestion
	// Special case: no edge proof provided; the given range contains all leaves in the trie
```

The `set` suggests no-ordering

core/trie/proof.go
```diff
@@ -54,623 +55,586 @@ func (e *Edge) Len() uint8 {
...
+//   - Zero element proof: In this case a single edge proof is enough to verify. If there are still more elements, the proof is invalid.
+//
+// If the range proof is invalid, this function returns false and an error.
+// Otherwise, it will return a flag to indicate if there are more elements.
+//
+// TODO(weiihann): Given a binary leaf and a left-sibling first key, if the right sibling is removed, the proof would still be valid.
+// Conversely, given a binary leaf and a right-sibling last key, if the left sibling is removed, the proof would still be valid.
+// Range proof should not be valid for both of these cases, but currently is, which is an attack vector.
+// The problem probably lies in how we do root hash calculation.
+func VerifyRangeProof(root, first *felt.Felt, keys, values []*felt.Felt, proof *ProofNodeSet) (bool, error) { //nolint:funlen,gocyclo
```
pnowosie, line 202: Why we need `first` here - which is if I'm correct `keys[0]`?
I noticed it is used in a special empty keys case - which is very weird imo - what's the purpose to proving empty range?

core/trie/proof.go
```diff
@@ -54,623 +55,586 @@ func (e *Edge) Len() uint8 {
...
-	rootHash := leftPath[0].Hash(hash)
-
-	// Get duplicates and insert by one
-	i := 0
-	for i = 0; i < minLen; i++ {
-		leftNode := leftPath[i]
-		rightNode := rightPath[i]
+	// Ensure all keys are monotonic increasing and values contain no deletions
+	for i := 0; i < len(keys); i++ {
+		if i < len(keys)-1 && keys[i].Cmp(keys[i+1]) > 0 {
```
pnowosie, line 210:
```suggestion
	// Ensure all keys are monotonically increasing and values contain no deletions
	for i := 0; i < len(keys)-1; i++ {
		if keys[i].Cmp(keys[i+1]) > 0 {
```

core/trie/proof.go
```diff
@@ -54,623 +55,586 @@ func (e *Edge) Len() uint8 {
...
-// ref: https://github.com/ethereum/go-ethereum/blob/v1.14.3/trie/proof.go#L484
-func VerifyRangeProof(root *felt.Felt, keys, values []*felt.Felt, proofKeys [2]*Key, proofValues [2]*felt.Felt,
-	proofs [2][]ProofNode, hash hashFunc,
-) (bool, error) {
-	// Step 0: checks
-	if len(keys) != len(values) {
-		return false, fmt.Errorf("inconsistent proof data, number of keys: %d, number of values: %d", len(keys), len(values))
+	// In all other cases, we require two edge paths available.
+	// First, ensure that the last key is greater than the first key
+	if last.Cmp(first) <= 0 {
```
pnowosie, line 279: Redundant, impossible case. Please notice we have `monotonically increasing` check.
...or I totally missed what is `first`

core/trie/trie.go
```diff
@@ -11,8 +11,11 @@ import (
 	"github.com/NethermindEth/juno/core/crypto"
 	"github.com/NethermindEth/juno/core/felt"
 	"github.com/NethermindEth/juno/db"
+	"github.com/NethermindEth/juno/utils"
 )
 
+const globalTrieHeight = 251 // TODO(weiihann): this is declared in core also, should be moved to a common place
```
pnowosie, line 17: Yeah, I noticed this also when working on snap-sync.
IMO should be single place which a public constant

pnowosie (commented):
1st part of review

core/trie/key.go
```diff
@@ -136,20 +134,74 @@ func (k *Key) Truncate(length uint8) {
...
+}
+
+func (k *Key) Bytes() [32]byte {
+	var result [32]byte
+	copy(result[:], k.bitset[:])
+	return result
+}
+
+// WithLeastSignificantBitsFrom returns a copy of the key with the least significant bits replaced by the other key.
+func (k *Key) WithLeastSignificantBitsFrom(other *Key) (*Key, error) {
```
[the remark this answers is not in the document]
weiihann: Ahh yes I was using them then decide to not use it anymore. Will remove them.

core/trie/key.go
```diff
@@ -96,31 +77,48 @@ func (k *Key) Equal(other *Key) bool {
...
-}
-
-// DeleteLSB right shifts and shortens the key
-func (k *Key) DeleteLSB(n uint8) {
+// ShiftRight removes n least significant bits from the key by performing a right shift
+// operation and reducing the key length. For example, if the key contains bits
+// "1111 0000" (length=8) and n=4, the result will be "1111" (length=4).
+//
+// The operation is destructive - it modifies the key in place.
+func (k *Key) ShiftRight(n uint8) {
```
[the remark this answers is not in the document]
weiihann: Yup it should be internal. I think the panic check can be removed, so the caller must ensure that the `k.len >= n`.

core/trie/proof.go
weiihann: In snap sync, a client provide the start key (here I use `first`) and the number of leaves for the range. In the returned keys, `first` may or may not equal to `key[0]`. If it's equal, then it's an existent edge proof, otherwise non-existent.

core/trie/proof.go
weiihann: As mentioned in the previous comment, `first` may or may not equal to `key[0]`, so the monotonically increasing check doesn't apply here.

utils/orderedset.go
```diff
@@ -0,0 +1,69 @@
...
+	"sync"
+)
+
+// OrderedSet is a thread-safe data structure that maintains both uniqueness and insertion order of elements.
+// It combines the benefits of both maps and slices:
+// - Uses a map for O(1) lookups and to ensure element uniqueness
+// - Uses a slice to maintain insertion order and enable ordered iteration
+// The data structure is safe for concurrent access through the use of a read-write mutex.
+type OrderedSet[K comparable, V any] struct {
+	nodeSet  map[K]V
```
pnowosie, line 13: It looks like we can go with 
```go
nodeSet map[K]struct{}
```
to save memory and we don't keep value twice. WDYT?

pnowosie (approved):
Nice work! 👍

core/trie/proof.go
```diff
@@ -54,623 +55,585 @@ func (e *Edge) Len() uint8 {
...
-
-	// Get duplicates and insert by one
-	i := 0
-	for i = 0; i < minLen; i++ {
-		leftNode := leftPath[i]
-		rightNode := rightPath[i]
+	// Ensure all keys are monotonically increasing and values contain no deletions
+	for i := 0; i < len(keys); i++ {
+		if i < len(keys)-1 && keys[i].Cmp(keys[i+1]) > 0 {
+			return false, errors.New("keys are not monotonic increasing")
```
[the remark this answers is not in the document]
weiihann: This is invalid because we will skip the last value check.

utils/orderedset.go
weiihann: Ah actually we can remove `nodeSet` completely, because we already have `nodePos` to track the position of the item in `nodeList`.

core/trie/key.go
```diff
@@ -136,20 +134,46 @@ func (k *Key) Truncate(length uint8) {
...
+}
+
+func findDivergentBit(longerKey, shorterKey *Key) uint8 {
+	divergentBit := uint8(0)
+	for divergentBit <= shorterKey.Len() &&
+		longerKey.IsBitSet(longerKey.Len()-divergentBit) == shorterKey.IsBitSet(shorterKey.Len()-divergentBit) {
+		divergentBit++
 	}
+	return divergentBit
+}
```
rianhughes, line 169: Do you think we should add a check that the length of shorterKey is less than longerKey? Currently I think this might panic if they are inserted in the incorrect order

core/trie/key.go
```diff
@@ -136,20 +134,46 @@ func (k *Key) Truncate(length uint8) {
...
-	inUseBytes := k.inUseBytes()
-	unusedBitsCount := 8 - (k.len % 8)
-	if unusedBitsCount != 8 && len(inUseBytes) > 0 {
-		inUseBytes[0] = (inUseBytes[0] << unusedBitsCount) >> unusedBitsCount
+// findCommonKey finds the set of common MSB bits in two key bitsets.
+func findCommonKey(longerKey, shorterKey *Key) (Key, bool) {
+	divergentBit := findDivergentBit(longerKey, shorterKey)
+	commonKey := *shorterKey
+	commonKey.shiftRight(shorterKey.Len() - divergentBit + 1)
+	return commonKey, divergentBit == shorterKey.Len()+1
```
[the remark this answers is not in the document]
rianhughes: I wonder if we can return early if there is no commonKey (and maybe save performing a deep copy)?

core/trie/node.go
```diff
@@ -138,3 +139,62 @@ func (n *Node) UnmarshalBinary(data []byte) error {
...
+	return fmt.Sprintf("Node{Value: %s, Left: %s, Right: %s, LeftHash: %s, RightHash: %s}", n.Value, n.Left, n.Right, n.LeftHash, n.RightHash)
+}
+
+// Update the receiver with non-nil fields from the `other` Node.
+// If a field is non-nil in both Nodes, they must be equal, or an error is returned.
+//
+// This method modifies the receiver in-place and returns an error if any field conflicts are detected.
+//
+//nolint:gocyclo
+func (n *Node) Update(other *Node) error {
```
rianhughes, line 153: If `other` contains a non-nil `value`, but conflicting `LeftHashes`, wouldn't the `receiver`s `value` be incorrectly updated?`

core/trie/trie.go
```diff
@@ -145,14 +125,85 @@ func (sn *StorageNode) Key() *Key {
...
-	return sn.node
+func (sn *StorageNode) Value() *felt.Felt {
+	return sn.node.Value
+}
+
+func (sn *StorageNode) String() string {
+	return fmt.Sprintf("StorageNode{key: %s, node: %s}", sn.key, sn.node)
+}
+
+func (sn *StorageNode) Update(other *StorageNode) error {
```
rianhughes, line 136: Similar to another comment, can key be (incorrectly) updated but the node not?

core/trie/trie.go
```diff
@@ -406,7 +458,8 @@ func (t *Trie) PutWithProof(key, value *felt.Felt, lProofPath, rProofPath []Stor
 	} else {
 		// Since we short-circuit in leaf updates, we will only end up here for deletions
 		// Delete if key already exist
-		sibling := nodes[len(nodes)-1]
+		last := nodes[len(nodes)-1]
+		sibling := last
```
rianhughes, line 462: Why `last`?

core/trie/trie.go
```diff
@@ -373,13 +425,13 @@ func (t *Trie) Put(key, value *felt.Felt) (*felt.Felt, error) {
 }
 
 // Put updates the corresponding `value` for a `key`
-func (t *Trie) PutWithProof(key, value *felt.Felt, lProofPath, rProofPath []StorageNode) (*felt.Felt, error) {
+func (t *Trie) PutWithProof(key, value *felt.Felt, proof []*StorageNode) (*felt.Felt, error) {
```
rianhughes, line 428: `proof` contains both the left and right proof, right?

core/trie/node.go
weiihann: Ahh yes thanks for spotting. Fixed this part.

core/trie/trie.go
weiihann: Fixed.

core/trie/trie.go
weiihann: Fixed.

core/trie/trie.go
weiihann: Yup

core/trie/key.go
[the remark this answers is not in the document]
weiihann: If shorterKey is not a subset of longerKey, that means they diverge at the first MSB of both keys. With the current logic, it is a bug because `shiftRight` will panic. So we should just return a nil key. But honestly this shouldn't happen at all.

core/trie/key.go
weiihann: If we look at the current usage of `findDivergentBit`, longerKey will always be of length `251`, and the max shorterKey can go is also `251`. So I don't think we need to add this check.

core/trie/proof_test.go
```diff
@@ -1,1139 +1,827 @@
...
-			},
-		}
+	t.Run("both non-existent edge proofs", func(t *testing.T) {
+		t.Parallel()
 
-		zeroFeltBytes := new(felt.Felt).SetUint64(0).Bytes()
-		leafkey := trie.NewKey(251, zeroFeltBytes[:])
-		sns, err := trie.ProofToPath(proofNodes, &leafkey, crypto.Pedersen)
+		start := 100
+		first, last := decrementFelt(records[start].key), incrementFelt(records[start].key)
```
rianhughes, line 338: Would this break if by chance first and last correspond to records[start-1].key, records[start+1].key? I guess it's unlikely, but possible? (same for other tests that use decrementFelt etc)

core/trie/proof_test.go
```diff
@@ -1,1139 +1,827 @@
...
-			&trie.Binary{
-				LeftHash:  utils.HexToFelt(t, "0x06E08BF82793229338CE60B65D1845F836C8E2FBFE2BC59FF24AEDBD8BA219C4"),
-				RightHash: utils.HexToFelt(t, "0x04F9B8E66212FB528C0C1BD02F43309C53B895AA7D9DC91180001BDD28A588FA"),
-			},
-			&trie.Edge{
-				Path:  &path1,
-				Child: utils.HexToFelt(t, "0xcc"),
-			},
+		first := decrementFelt(records[start].key)
+		if start != 0 && first.Equal(records[start-1].key) {
```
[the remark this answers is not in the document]
weiihann: This condition basically ensures that `first` is a non-existent key. `first == records[start].key - 1` and if first == records[start-1].key`, it's an existent key so we skip this key.

core/trie/proof_test.go
weiihann: Very unlikely. The possibility of that happening is about `(1/2^32)^3`. I used a constant seed for the random test cases so this should be fine.

core/trie/proof_test.go
```diff
@@ -1,1139 +1,827 @@
...
+	require.NoError(t, err)
 
-		proofs, err := trie.GetBoundaryProofs(proofKeys[0], proofKeys[1], tri)
-		require.NoError(t, err)
+	cases := []struct {
+		pos int
+		err bool
+	}{
+		{len(records) - 1, false},
+		{500, true},
```
rianhughes, line 456: I guess this is testing the special case:
`// Special case: there is a provided proof but no key-value pairs, make sure regenerated trie has no more values`
Why is the proof invalid if there are additional elements? Would this not be equivalent to proving non-inclusion?

weiihann: You're right, I'm not sure why geth restricts that there should be no more values given a zero element proof. I'm keen on removing the `hasRightElement` condition for this case, but before that let me try to ask them on this.

core/trie/proof_test.go
rianhughes: Yup, good point :D

rianhughes (approved):

rianhughes:
Very nice work :)

core/trie/proof_test.go
weiihann: I got the answer. It's a snap sync specification detail.

Based on Ethereum's snap sync [specs](https://github.com/ethereum/devp2p/blob/master/caps/snap.md#getaccountrange-0x00), it says:

> If no accounts exist between startingHash and limitHash, then the first (if any) account after limitHash must be provided.

So it's guaranteed that empty-range proof can only exist if no more elements are on the right side. Else it's a invalid response.

I would just keep the implementation as it is for now with the added comments. Once we formalize the specification we can decide if we want to keep it or change it.

utils/orderedset.go
```diff
@@ -0,0 +1,67 @@
...
+	// Update existing entry
+	if pos, exists := ps.itemPos[key]; exists {
+		ps.items[pos] = value
+		return
+	}
+
+	// Insert new entry
+	ps.itemPos[key] = len(ps.items)
+	ps.items = append(ps.items, value)
+	ps.size++
```
kirugan, line 38: why do we need size? can we use `len(ps.items)` instead ?

PR #2254, opened 2024-11-06
Add subscribeEvents and subcriptionEvent

IronGauntlets:


rpc/subscriptions.go
```diff
@@ -0,0 +1,186 @@
...
+	defer func() {
+		h.callAndLogErr(filter.Close, "Error closing event filter in events subscription")
+	}()
+
+	if err = setEventFilterRange(filter, &BlockID{Number: from}, &BlockID{Number: to}, to); err != nil {
+		h.log.Warnw("Error setting event filter range", "err", err)
+		return
+	}
+
+	filteredEvents, cToken, err := filter.Events(nil, subscribeEventsChunkSize)
```
pnowosie, line 123: How about `nextToken` istead of `cToken`? - I'd like to make the name more descriptive.

rpc/subscriptions_test.go
```diff
@@ -0,0 +1,338 @@
...
+		serverConn, clientConn := net.Pipe()
+		t.Cleanup(func() {
+			require.NoError(t, serverConn.Close())
+			require.NoError(t, clientConn.Close())
+		})
+
+		subCtx := context.WithValue(context.Background(), jsonrpc.ConnKey{}, &fakeConn{w: serverConn})
+
+		// Note the end of the window doesn't need to be tested because if requested block number is more than the
+		// head, a block not found error will be returned. This behaviour has been tested in various other test, and we
```
pnowosie, line 119:
```suggestion
		// head, a block not found error will be returned. This behaviour has been tested in various other tests, and we
```

rpc/subscriptions_test.go
```diff
@@ -0,0 +1,338 @@
+package rpc
```
pnowosie, line 1: Usualy tests go to `rpc_test` pkg. Can you add a comment with a reason here?

pnowosie (commented):
Some minors comments & ask for details

rpc/handlers.go
```diff
@@ -65,12 +65,15 @@ var (
 	ErrUnsupportedTxVersion            = &jsonrpc.Error{Code: 61, Message: "the transaction version is not supported"}
 	ErrUnsupportedContractClassVersion = &jsonrpc.Error{Code: 62, Message: "the contract class version is not supported"}
 	ErrUnexpectedError                 = &jsonrpc.Error{Code: 63, Message: "An unexpected error occurred"}
+	ErrTooManyBlocksBack               = &jsonrpc.Error{Code: 68, Message: "Cannot go back more than 1024 blocks"}
```
kirugan, line 68: do you think it's worth to pass actual constant here i.e.:
```
Message: fmt.Sprintf("Cannot go back more than %d blocks", maxBlocksBack)
```

rpc/subscriptions.go
```diff
@@ -0,0 +1,186 @@
...
+				case <-subscriptionCtx.Done():
+					return
+				case header := <-headerSub.Recv():
+
+					h.processEvents(subscriptionCtx, w, id, header.Number, header.Number, fromAddr, keys)
+				}
+			}
+		}()
+
+		h.processEvents(subscriptionCtx, w, id, requestedHeader.Number, headHeader.Number, fromAddr, keys)
```
kirugan, line 99: Could you elaborate what is difference between this call here and another one in a goroutine above

rpc/subscriptions.go
```diff
@@ -0,0 +1,186 @@
...
+func (h *Handler) processEvents(ctx context.Context, w jsonrpc.Conn, id, from, to uint64, fromAddr *felt.Felt, keys [][]felt.Felt) {
+	filter, err := h.bcReader.EventFilter(fromAddr, keys)
+	if err != nil {
+		h.log.Warnw("Error creating event filter", "err", err)
+		return
+	}
+
+	defer func() {
+		h.callAndLogErr(filter.Close, "Error closing event filter in events subscription")
+	}()
```
kirugan, line 116: seems like lambda is redundant here:
```
defer h.callAndLogErr(filter.Close, "...")
```

IronGauntlets: Done

rpc/subscriptions.go
IronGauntlets: The one in the goroutine checks for events from new blocks as it listens for the new heads subscription.

This one fetches the events from current and/or old blocks especially if the requested blockID requests events from block `curHeight-1024`

rpc/handlers.go
IronGauntlets: Done

rpc/subscriptions_test.go
IronGauntlets: Yes but we don't have to follow it if its is difficult to use components which make network/db operations. Therefore, I use `package rpc` for mocking which allows me to easily make particular scenarios on events possible.

rpc/subscriptions_test.go
IronGauntlets: Done

rpc/subscriptions.go
IronGauntlets: `c` is short for continuation, same convention is used in other event related functions.

rpc/subscriptions.go
```diff
@@ -0,0 +1,186 @@
+package rpc
```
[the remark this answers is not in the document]
IronGauntlets: I think we should move all the subscription related code to `subscriptions.go` and other prs should follow the same convention.

kirugan (approved):

pnowosie (approved):

PR #2275, opened 2024-11-19
Remove supported block version check

wojciechos:
Allows processing of all versions without explicit checks

### DIFF
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -8,7 +8,6 @@ import (
 	"sync/atomic"
 	"time"
 
-	"github.com/Masterminds/semver/v3"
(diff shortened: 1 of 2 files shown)


blockchain/blockchain.go
```diff
@@ -382,7 +365,7 @@ func (b *Blockchain) VerifyBlock(block *core.Block) error {
 }
 
 func verifyBlock(txn db.Transaction, block *core.Block) error {
-	if err := checkBlockVersion(block.ProtocolVersion); err != nil {
+	if _, err := core.ParseBlockVersion(block.ProtocolVersion); err != nil {
```
wojciechos, line 368: Wondering if we still need this or should also remove the block parsing check?

PR #2277, opened 2024-11-20
Websocket shutdown logic

kirugan:


jsonrpc/websocket.go
```diff
@@ -54,7 +57,19 @@ func (ws *Websocket) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	ctx, cancel := context.WithCancel(r.Context())
+	defer cancel()
+	go func() {
+		select {
+		case <-ws.shutdown:
+			cancel()
+		case <-ctx.Done():
+			// in case websocket connection is closed and server is not in shutdown mode
+			// we need to release this goroutine from waiting
+		}
```
pnowosie, line 69: World you mind to add `return` here to make it more clear we intend to exit the goroutine?
For me it'd more readable 🤔

jsonrpc/websocket_test.go
```diff
@@ -20,7 +20,7 @@ func testConnection(t *testing.T, ctx context.Context, method jsonrpc.Method, li
 	require.NoError(t, rpc.RegisterMethods(method))
 
 	// Server
-	srv := httptest.NewServer(jsonrpc.NewWebsocket(rpc, utils.NewNopZapLogger()))
+	srv := httptest.NewServer(jsonrpc.NewWebsocket(rpc, nil, utils.NewNopZapLogger()))
```
pnowosie, line 23: According to [this article](https://www.codingexplorations.com/blog/mastering-nil-channels-in-go-enhance-your-concurrency-control) `nil` channel will block indefinitely, which is I think counter-intuitive.
Maybe it's worth to add a comment that it will block?

rpc/events_test.go
```diff
@@ -347,7 +347,7 @@ func TestMultipleSubscribeNewHeadsAndUnsubscribe(t *testing.T) {
 		Params:  []jsonrpc.Parameter{{Name: "id"}},
 		Handler: handler.Unsubscribe,
 	}))
-	ws := jsonrpc.NewWebsocket(server, log)
+	ws := jsonrpc.NewWebsocket(server, nil, log)
```
pnowosie, line 350: Also comment about the `nil`-channel behaviour? (see prev comment)

pnowosie (approved):
👍 Just small concerns

IronGauntlets (approved):
LGTM. Please point the branch to main.

jsonrpc/websocket_test.go
kirugan: That's true. The main point here is that in tests we pass nil for arguments in constructor that for anything that is not relevant to a test. 

Nil shutdown channel won't cause any trouble here either because goroutine that is listening for it will be canceled once call `ServeHTTP` is done. Here is more detailed walkthrough:
```
ctx, cancel := context.WithCancel(r.Context())
defer cancel() // 2. context is cancelled when function finished its execution (`ServeHTTP`)
go func() {
		select {
		case <-ws.shutdown: // 1. blocked indefinetely
			cancel()
		case <-ctx.Done(): // 3. since context is canceled on step 2 we exit here, releasing gorutine
		}
}()
```

pnowosie (approved):

PR #2285, opened 2024-11-25
Block's version verifies only the major version

pnowosie:
Block's protocol version verification checks only the major part of the version

### DIFF
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -53,7 +53,7 @@ type Reader interface {
 
 var (
 	ErrParentDoesNotMatchHead = errors.New("block's parent hash does not match head block hash")
-	supportedStarknetVersion  = semver.MustParse("0.13.3")
+	SupportedStarknetVersion  = semver.MustParse("0.13.3")
 )
 
 func checkBlockVersion(protocolVersion string) error {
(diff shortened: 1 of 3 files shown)


blockchain/blockchain.go
```diff
@@ -62,7 +62,8 @@ func checkBlockVersion(protocolVersion string) error {
 		return err
 	}
 
-	if blockVer.GreaterThan(supportedStarknetVersion) {
+	// We only care about major version changes ignoring minor and patch
+	if blockVer.Major() > SupportedStarknetVersion.Major() {
```
kirugan, line 66: ah, I meant we should ignoring patch version, sorry for confusion

blockchain/blockchain.go
```diff
@@ -62,7 +62,9 @@ func checkBlockVersion(protocolVersion string) error {
 		return err
 	}
 
-	if blockVer.GreaterThan(supportedStarknetVersion) {
+	// We ignore changes in patch part of the version
+	blockVerMM, suportedVerMM := blockVer.IncMinor(), SupportedStarknetVersion.IncMinor()
+	if blockVerMM.GreaterThan(&suportedVerMM) {
```
kirugan, line 67: wdyt about adding this function:
```
func CopyWithoutPatch(v *semver.Version) *semver.Version {
	if v == nil {
		return nil
	}

	return semver.New(v.Major(), v.Minor(), 0, v.Prerelease(), v.Metadata())
}
```
to `utils` ?

kirugan: we can update function name later, but I think it would be clear how we treat block version and supported versions

IronGauntlets (approved):

kirugan (approved):

PR #2289, opened 2024-11-28
Fix spamming errors when requesting data from Juno that is not (yet) …

AnkushinDaniil:
This pull request includes changes to the `p2p/starknet/handlers.go` file to improve error handling and add a new import. The most important changes include adding a new import for the `db` package and modifying error handling to specifically check for `db.ErrKeyNotFound`.

### Import Addition:
* Added the `db` package to the import list in `p2p/starknet/handlers.go`.

### Error Handling Improvements:
* Updated the `processIterationRequest` function to log errors only if they are not `db.ErrKeyNotFound`.
* Updated the `processIterationRequestMulti` function to log errors only if they are not `db.ErrKeyNotFound`.

https://demerzelsolutions.slack.com/archives/C03HL8DH52N/p1732706848085409?thread_ts=1732286243.285989&cid=C03HL8DH52N

### DIFF
--- a/p2p/starknet/handlers.go
+++ b/p2p/starknet/handlers.go
@@ -4,6 +4,7 @@ package starknet
 import (
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"iter"
 	"sync"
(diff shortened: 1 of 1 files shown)


p2p/starknet/handlers.go
kirugan, line 1: it's better to use `errors.Is(...)` function

pnowosie (approved):
Nice catch 👍

wojciechos (approved):

PR #2290, opened 2024-11-28
Fix panic when tx hash is nil

### LINKED ISSUE #2239: Juno panics when receiving Transaction message with nil transaction_hash in P2P sync
Juno panics and exits when it receives a Transaction message with a `nil` `transaction_hash` in P2P sync. According to the Starnet P2P spec, `transaction_hash` is required, but this could happen due to bugs or even malicious nodes. Juno should handle such cases gracefully without panicking.

### **Steps to Reproduce:**  
1. Set up Juno to sync via P2P with another client (e.g Pathfinder<=v0.14.4).
2. If a Transaction message with a `nil` `transaction_hash` is received, Juno panics.

### **Expected Behavior:**  
Juno should not panic when receiving a Transaction message with a missing or `nil` `transaction_hash`. It should log an error or reject the message, allowing the node to continue running.

### **Logs:**
```
(*core.DeclareTransaction)(0xc00081cb40)({
 TransactionHash: (*felt.Felt)(<nil>),
 ClassHash: (*felt.Felt)(0xc0005012c0)(0x5c478ee27f2112411f86f207605b2e2c58cdb647bac0df27f660ef2252359c6),
 SenderAddress: (*felt.Felt)(0xc0005012e0)(0x1),
 MaxFee: (*felt.Felt)(0xc000501300)(0x0),
 TransactionSignature: ([]*felt.Felt) <nil>,
 Nonce: (*felt.Felt)(<nil>),
 Version: (*core.TransactionVersion)(0xc000501320)(0x0),
 CompiledClassHash: (*felt.Felt)(<nil>),
 ResourceBounds: (map[core.Resource]core.ResourceBounds) <nil>,
 Tip: (uint64) 0,
 PaymasterData: ([]*felt.Felt) <nil>,
 AccountDeploymentData: ([]*felt.Felt) <nil>,
 NonceDAMode: (core.DataAvailabilityMode) 0,
 FeeDAMode: (core.DataAvailabilityMode) 0
})
panic: TX hash 0 is nil

goroutine 242 [running]:
github.com/NethermindEth/juno/p2p.(*syncService).adaptAndSanityCheckBlock.func1()
	/app/p2p/sync.go:297 +0xee5
created by github.com/NethermindEth/juno/p2p.(*syncService).adaptAndSanityCheckBlock in goroutine 229
	/app/p2p/sync.go:271 +0x1eb
```

AnkushinDaniil:
Fix #2239 
Replaced the `panic` call with an error log

### DIFF
--- a/p2p/sync.go
+++ b/p2p/sync.go
@@ -18,7 +18,6 @@ import (
 	junoSync "github.com/NethermindEth/juno/sync"
 	"github.com/NethermindEth/juno/utils"
 	"github.com/NethermindEth/juno/utils/pipeline"
-	"github.com/davecgh/go-spew/spew"
 	"github.com/libp2p/go-libp2p/core/host"
 	"github.com/libp2p/go-libp2p/core/network"
 	"github.com/libp2p/go-libp2p/core/peer"
@@ -293,8 +292,8 @@ func (s *syncService) adaptAndSanityCheckBlock(ctx context.Context, header *spec
 			for i, r := range receipts {
 				txHash := coreTxs[i].Hash()
 				if txHash == nil {
-					spew.Dump(coreTxs[i])
-					panic(fmt.Errorf("TX hash %d is nil", i))
+					s.log.Errorw("Transaction hash is nil", "index", i)
+					return
 				}
 				coreReceipts = append(coreReceipts, p2p2core.AdaptReceipt(r, txHash))
 			}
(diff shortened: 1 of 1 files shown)


p2p/sync.go
```diff
@@ -293,8 +292,8 @@ func (s *syncService) adaptAndSanityCheckBlock(ctx context.Context, header *spec
 			for i, r := range receipts {
 				txHash := coreTxs[i].Hash()
 				if txHash == nil {
-					spew.Dump(coreTxs[i])
-					panic(fmt.Errorf("TX hash %d is nil", i))
+					s.log.Errorw("Transaction hash is nil", "index", i)
+					return
```
kirugan, line 296: Panic acted like an assertion here. The reason it was introduced in the first place that there were from time to time different problems related to final hash, but since then it became more stable so TL;DR: let's remove this check completely

kirugan (approved):

pnowosie (approved):
So it looks like nil check was unnecessary 🤔

PR #2294, opened 2024-11-28
Remove transaction_hash from getBlockWithReceipts

IronGauntlets:
`starknet_getBlockWithReceipts` was added in 0.7.x rpc spec and its array of `transaction` doesn't contain `transaction_hash` instead it contains the `transaction_hash`.


wojciechos (approved):

pnowosie (approved):
> `starknet_getBlockWithReceipts` was added in 0.7.x rpc spec and its array of `transaction` doesn't contain `transaction_hash` instead it contains the `transaction_hash`.

I understand the PR's title but not this comment 🤔 
Otherwise 👍

PR #2295, opened 2024-11-28
Update benchmarks

IronGauntlets:
Please view each commit separately.

### DIFF
--- a/core/crypto/ecdsa_test.go
+++ b/core/crypto/ecdsa_test.go
@@ -57,6 +57,8 @@ func TestVerify(t *testing.T) {
 	}
 }
 
+var benchVerifyR bool
+
 func BenchmarkVerify(b *testing.B) {
 	signature := crypto.Signature{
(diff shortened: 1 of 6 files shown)


core/crypto/ecdsa_test.go
```diff
@@ -57,6 +57,8 @@ func TestVerify(t *testing.T) {
 	}
 }
 
+var benchVerifyR bool
```
pnowosie, line 60: Why R? I'm curious about its origin - perhaps it's time for a creative renaming! Any suggestions for what R could stand for? (Wrong answers only, of course!)

IronGauntlets: R is for Result.

pnowosie, line 60:
```suggestion
var benchVerifyResult bool
```

pnowosie (approved):
My only consern is rename `*R -> *Result` if possible.
Thanks

core/crypto/ecdsa_test.go
IronGauntlets: I think it is fine as is.

AnkushinDaniil (approved):

PR #2303, opened 2024-12-04
Skip error logs for FGW responses with NOT_RECEIVED status

wojciechos:
Do not log an error when FGW returns NOT_RECEIVED as this is a common scenario and should not generate misleading or confusing log entries for users.

### DIFF
--- a/rpc/transaction.go
+++ b/rpc/transaction.go
@@ -585,13 +585,11 @@ func (h *Handler) TransactionStatus(ctx context.Context, hash felt.Felt) (*Trans
 
 		status, err := adaptTransactionStatus(txStatus)
 		if err != nil {
-			h.log.Errorw("Failed to adapt transaction status", "err", err)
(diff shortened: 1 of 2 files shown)


rianhughes (approved):

rpc/transaction.go
```diff
@@ -751,6 +752,8 @@ func adaptTransactionStatus(txStatus *starknet.TransactionStatus) (*TransactionS
 		status.Finality = TxnStatusAcceptedOnL2
 	case starknet.Received:
 		status.Finality = TxnStatusReceived
+	case starknet.NotReceived:
+		return nil, fmt.Errorf("%s", ErrTxnHashNotFound.Message)
```
kirugan, line 756: maybe it would better to create custom error and reuse it here and above? using rpc error (specifically error message) in my opinion a little bit tricky

wojciechos: yep, sounds cleaner - updated

PR #2306, opened 2024-12-05
Remove obsolete migrations

kirugan:


core/felt/felt.go
```diff
@@ -26,7 +26,6 @@ const (
 )
 
 const (
-	Limbs = fp.Limbs // number of 64 bits words needed to represent a Element
```
IronGauntlets, line 29: I am not sure why this is being removed.

migration/migration.go
```diff
@@ -46,26 +36,12 @@ func (f MigrationFunc) Migrate(_ context.Context, txn db.Transaction, network *u
 // Before is a no-op.
 func (f MigrationFunc) Before(_ []byte) error { return nil }
 
+// todo comment
+const startMigrationVersion = 17
```
IronGauntlets, line 40: Would it be a good idea to start from 0 again? There are a few other migrations that need to be merged soon. Maybe it is a good idea to merge it after that?

core/felt/felt.go
kirugan: it was used only in migration I thought that it was introduced only for migrations

migration/migration.go
kirugan: > There are a few other migrations that need to be merged soon. Maybe it is a good idea to merge it after that?

You mean merge this PR after all migrations in `main` ? Sure.

> Would it be a good idea to start from 0 again?
In this case we need to do something like this:
```
if migration.Version > 0 {
    if migration.Version < 17 {
         return err
    } else migration.Version == 17 { 
        migration.Version = 0
    } else {
        // above 17 can't be ?
    }
}
```

Also when we apply migrations there is "progress bar" for migrations:
```
Applying migration 16/17
```
if you think it doesn't worth to maintain sequential order even if we remove obsolete migrations (I'm ok with that) - let's do that

pnowosie (approved):
👍

IronGauntlets (approved):
I have rebased the branch on top of `main` and made some functions in blockchain private.

Also, I don't think `startSchemaVersion` needs to be updated since it should have been 16 before (please confirm I am correct).

Also, please don't squash and merge, instead rebase and merge.

PR #2309, opened 2024-12-07
Add schema_version to output of db info command

kirugan:


cmd/juno/dbcmd.go
```diff
@@ -97,6 +99,12 @@ func dbInfo(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("failed to get the state update: %v", err)
 	}
 
+	schemaMeta, err := migration.SchemaMetadata(database)
```
kirugan, line 102: maybe we should move it to separate package idk, what is your take on this ?

weiihann: which package are you looking at? This seems good to me

kirugan: Something like `db.SchemaMetadata` instead of `migration`

weiihann: I think migration should have its own package. Perhaps move it to `db` and treat it as a subpackage.

kirugan: Actually let's keep it as is. I think schema makes sense only for migrations and in this place which is not significant.

IronGauntlets (approved):

cmd/juno/dbcmd.go
```diff
@@ -97,6 +99,12 @@ func dbInfo(cmd *cobra.Command, args []string) error {
...
+	if err != nil {
+		return fmt.Errorf("failed to get schema metadata: %v", err)
+	}
```
rodrigo-pino, line 105: How much attention should we pay to this warning?

cmd/juno/dbcmd.go
```diff
@@ -97,6 +99,12 @@ func dbInfo(cmd *cobra.Command, args []string) error {
...
+	info.SchemaVersion = schemaMeta.Version
+
```
rodrigo-pino, line 107: Nitpick: is new line (107) intentional?

rodrigo-pino (approved):

cmd/juno/dbcmd.go
kirugan: not really, let me remove it

cmd/juno/dbcmd.go
kirugan: it's just an error that most likely indicate that there are some

PR #2315, opened 2024-12-10
Improve binary build workflow for cross-platform releases

wojciechos:
Tested here: https://github.com/NethermindEth/juno/actions/runs/12258083548

- Fix mac-os build fail
- Add proper architecture handling in matrix configuration
- Implement caching for Go modules and Rust dependencies
- Streamline dependency installation for both Linux and macOS
- Improve binary artifact handling and checksums
- Add retention policy for build artifacts
- Split build steps for better clarity and maintainability

### DIFF
--- a/.github/workflows/build-binaries.yml
+++ b/.github/workflows/build-binaries.yml
@@ -18,9 +18,13 @@ jobs:
       matrix:
         include:
           - os: ubuntu-latest
+            arch: amd64
           - os: macos-13
+            arch: amd64
           - os: ubuntu-arm64-4-core
+            arch: arm64
           - os: macos-latest
+            arch: arm64
     runs-on: ${{ matrix.os }}
 
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/build-binaries.yml
```diff
@@ -29,46 +33,67 @@ jobs:
...
 
-      - name: Upload Artifact
+      - name: Upload artifact
         uses: actions/upload-artifact@v4
         with:
           name: ${{ env.ARTIFACT_NAME }}
           path: |
             ${{ env.ARTIFACT_NAME }}
             ${{ env.ARTIFACT_NAME }}.sha256
+          retention-days: 30
```
kirugan, line 99: does it mean that binaries will be stored for 30 days only?

wojciechos: yep, do we need them longer? It won't remove those which are in our releases

kirugan: > yep, do we need them longer? It won't remove those which are in our releases

got it, I was worrying about these ones

.github/workflows/build-binaries.yml
```diff
@@ -29,46 +33,67 @@ jobs:
...
 
       - name: Install dependencies (macOS)
         if: runner.os == 'macOS'
-        run: brew install cargo-c jemalloc
-
-      - name: Set up Go
-        uses: actions/setup-go@v5
-        with:
-          go-version-file: go.mod
+        run: brew install jemalloc
```
kirugan, line 69: we don't need cargo-c because it was installed above on "Set up Rust" step ?

.github/workflows/build-binaries.yml
```diff
@@ -29,46 +33,67 @@ jobs:
...
+        run: make juno
+
+      - name: Compress binary (Linux)
+        if: runner.os == 'Linux'
         run: |
-          make juno
-          if [[ "${{ runner.os }}" != "macOS" ]]; then
-            upx build/juno
-          fi
+          upx build/juno
```
kirugan, line 77: can you remind me why we don't use `upx` for macOS ?

kirugan (approved):

.github/workflows/build-binaries.yml
wojciechos: Right, it's redundant - removed.

.github/workflows/build-binaries.yml
wojciechos: UPX and macOS don’t always work nice. Apple’s Gatekeeper can flag UPX-packed binaries as suspicious, and macOS already compresses binaries, so using UPX doesn’t add much. Instead, it might just cause more headaches than it’s worth.

kirugan: thanks 🙇

rianhughes (approved):

.github/workflows/build-binaries.yml
```diff
@@ -29,46 +33,67 @@ jobs:
...
+            starknet/compiler/rust
+
       - name: Get latest tag
         run: echo "TAG=$(git describe --tags)" >> $GITHUB_ENV
 
       - name: Get artifact name
-        run: echo "ARTIFACT_NAME=juno-${{ env.TAG }}-${{ runner.os }}-$(uname -m)" >> $GITHUB_ENV
+        run: |
+          OS_NAME=$([ "${{ runner.os }}" == "macOS" ] && echo "darwin" || echo "linux")
+          echo "ARTIFACT_NAME=juno-${{ env.TAG }}-${OS_NAME}-${{ matrix.arch }}" >> $GITHUB_ENV
```
derrix060, line 59: why relying on `matrix.arch` which is hard-coded instead of `uname -m` which is dynamic and will always reflect the correct arch?

.github/workflows/build-binaries.yml
```diff
@@ -29,46 +33,67 @@ jobs:
...
+            core/rust
+            starknet/compiler/rust
+
       - name: Get latest tag
         run: echo "TAG=$(git describe --tags)" >> $GITHUB_ENV
 
       - name: Get artifact name
-        run: echo "ARTIFACT_NAME=juno-${{ env.TAG }}-${{ runner.os }}-$(uname -m)" >> $GITHUB_ENV
+        run: |
+          OS_NAME=$([ "${{ runner.os }}" == "macOS" ] && echo "darwin" || echo "linux")
```
derrix060, line 58: could probably also change the `runner.os` to be `uname -o`

PR #2319, opened 2024-12-12
Remove size in OrderedSet

weiihann:


kirugan (approved):

AnkushinDaniil (approved):

utils/orderedset.go
```diff
@@ -22,46 +21,45 @@ func NewOrderedSet[K comparable, V any]() *OrderedSet[K, V] {
 	}
 }
 
-func (ps *OrderedSet[K, V]) Put(key K, value V) {
-	ps.lock.Lock()
-	defer ps.lock.Unlock()
+func (o *OrderedSet[K, V]) Put(key K, value V) {
```
pnowosie, line 24: One tiny extension, let it return `bool` value indicating that value was new. 
Returns `True` if the element was added to the set; `False` if the element was already present.

```suggestion
func (o *OrderedSet[K, V]) Put(key K, value V) bool {
```

pnowosie (approved):
👍 Nice

utils/orderedset.go
pnowosie: Or on the 2nd thought we might return the previous value on update (?) 
WDYT? 🤔

weiihann: This is a valid suggestion, but we have no use case of returning `bool`/`V` for now. So I'll leave it to future update if we ever need it.

PR #2320, opened 2024-12-12
refactor: use errors.New to replace fmt.Errorf with no parameters

ChengenH:


weiihann (approved):
LGTM, thanks for the contribution!

ChengenH:
> LGTM, thanks for the contribution!

Thank you for the review!

derrix060 (changes requested):
Can you please explain why the changes (in your commit message)?  Why is it better to use errors.New instead of fmt.Errorf?  Specially because I can see that on the files you added it, the `fmt` package is still being imported, so not much gain there.

ChengenH:
> Can you please explain why the changes (in your commit message)? Why is it better to use errors.New instead of fmt.Errorf? Specially because I can see that on the files you added it, the `fmt` package is still being imported, so not much gain there.
> 
> If it's just a linting thing, then please create a linting rule to avoid people using fmt.Errorf in the future when errors.New would be prefered, otherwise one would need to create this PR again to fix the few new instances

errors.New has better performance, although slightly, since it does not involve the overhead of formatting the string. It is also more concise when there are no formatting arguments.
this will not affect performance and readability too much.

ChengenH:
> Can you please explain why the changes (in your commit message)? Why is it better to use errors.New instead of fmt.Errorf? Specially because I can see that on the files you added it, the `fmt` package is still being imported, so not much gain there.
> 
> If it's just a linting thing, then please create a linting rule to avoid people using fmt.Errorf in the future when errors.New would be prefered, otherwise one would need to create this PR again to fix the few new instances

Additionally, I noticed that in most projects I read, errors without formatting parameters are typically handled using errors.New.

IronGauntlets (approved):

PR #2322, opened 2024-12-15
Implement `BitArray` and replace `trie.Key`

weiihann:
This PR implements a new data structure `BitArray` to replace `trie.Key`. The reasons being:

1. Misleading representation
`trie.Key` doesn't actually just represent a Trie key in our use case. But rather, it’s a special data type that we use to just represent a fixed size of bit array with the number of bits used specified.

2. Inefficient operations
The current implementation of the bitwise operations is scattered all around the place. We implement ourselves for some, and for others, we rely on conversion. `BitArray` implements everything from scratch without unnecessary conversion.

Regarding `BitArray`:
- It uses a little-endian format to perform bitwise operations more efficiently. All available operations are multiple times faster than `trie.Key`
- It is alloc-free
- It supports all use cases for trie and proofs (e.g. derive path from two bit arrays, check if path matches, etc)
- Adds built-in support for conversion (e.g. felt to bit array, bytes to bit array, etc)

_Credits: some implementation details are inspired by `https://github.com/holiman/uint256`_


core/trie/bitarray.go
```diff
@@ -0,0 +1,397 @@
...
+	maxBitArray   = [4]uint64{math.MaxUint64, math.MaxUint64, math.MaxUint64, math.MaxUint64}
+	emptyBitArray = new(BitArray)
+)
+
+// BitArray is a structure that represents a bit array with length representing the number of used bits.
+// It uses a little endian representation to do bitwise operations of the words efficiently.
+// For example, if len is 10, it means that the 2^9, 2^8, ..., 2^0 bits are used.
+// The max length is 255 bits (uint8), because our use case only need up to 251 bits for a given trie key.
+// Although words can be used to represent 256 bits, we don't want to add an additional byte for the length.
+type BitArray struct {
```
rodrigo-pino, line 27:
```suggestion
// Represents a bit array with length representing the number of used bits.
// It uses a little endian representation to do bitwise operations of the words efficiently.
// For example, if len is 10, it means that the 2^9, 2^8, ..., 2^0 bits are used.
// The max length is 255 bits (uint8), because our use case only need up to 251 bits for a given trie key.
// Although words can be used to represent 256 bits, we don't want to add an additional byte for the length.
type BitArray struct {
```

Just a small suggestion, I think there is no need to preface with `BitArray is a structure` since you can infer that from the definition

core/trie/bitarray.go
```diff
@@ -0,0 +1,397 @@
...
+	len   uint8     // number of used bits
+	words [4]uint64 // little endian (i.e. words[0] is the least significant)
+}
+
+func (b *BitArray) Felt() *felt.Felt {
+	bs := b.Bytes()
+	return new(felt.Felt).SetBytes(bs[:])
```
rodrigo-pino, line 34: Why return a pointer to a Felt instead of the Felt itself?

core/trie/bitarray.go
```diff
@@ -0,0 +1,397 @@
...
+	bits8     = 8
+)
+
+var (
+	maxBitArray   = [4]uint64{math.MaxUint64, math.MaxUint64, math.MaxUint64, math.MaxUint64}
+	emptyBitArray = new(BitArray)
+)
+
+// BitArray is a structure that represents a bit array with length representing the number of used bits.
+// It uses a little endian representation to do bitwise operations of the words efficiently.
```
rodrigo-pino, line 23: The same couldn't be achieved with big endian?

core/trie/bitarray.go
```diff
@@ -0,0 +1,397 @@
...
+
+	long, short = b, x
+	if b.len < x.len {
+		long, short = x, b
+	}
+
+	return long.Rsh(long, long.len-short.len).Equal(short)
+}
+
+// Truncate sets b to the first 'length' bits of x (starting from the least significant bit).
```
rodrigo-pino, line 115:
```
// Truncate sets the bit array ...
```
Suggestion to be inline with the previous functions which never mention `b`.

I would also omit the name of the function from the description since it can be infered from the function name. Smth like:

```
// Sets the bit array ...
```

core/trie/bitarray.go
```diff
@@ -0,0 +1,397 @@
...
+		mask := maxUint64 >> (64 - b.len)
+		binary.BigEndian.PutUint64(res[24:32], b.words[0]&mask)
+	}
+
+	return res
+}
+
+// EqualMSBs checks if two bit arrays share the same most significant bits, where the length of
+// the check is determined by the shorter array. Returns true if either array has
+// length 0, or if the first min(b.len, x.len) MSBs are identical.
```
rodrigo-pino, line 81: Since it is written from the point of view of a method I would change the description to represent it. Something along the lines of:

```
// Checks if the current bit array shares the same most significant bits with
// another bit array ...
```

Just suggestions though, feel free dismiss any part of it

rodrigo-pino (commented):
Just some comments. Let. me know what you think. Thanks for detailing the implementation!

core/trie/bitarray.go
rodrigo-pino: I've seen that the code style starts always with the function name, so I think is better to maintain it

core/trie/bitarray.go
```diff
@@ -0,0 +1,397 @@
...
+	emptyBitArray = new(BitArray)
+)
+
+// BitArray is a structure that represents a bit array with length representing the number of used bits.
+// It uses a little endian representation to do bitwise operations of the words efficiently.
+// For example, if len is 10, it means that the 2^9, 2^8, ..., 2^0 bits are used.
+// The max length is 255 bits (uint8), because our use case only need up to 251 bits for a given trie key.
+// Although words can be used to represent 256 bits, we don't want to add an additional byte for the length.
+type BitArray struct {
+	len   uint8     // number of used bits
```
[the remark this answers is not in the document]
weiihann: Both statements are valid, it depends on how you look at it. I didn't want to put "position of the most significant bit" as I thought readers may be confused with the default most significant bit of `255`.

core/trie/bitarray.go
weiihann: Yeah, better for maintainability as well, will change it for the methods too

core/trie/bitarray.go
weiihann: Big endian works too. However, usually when dealing with low level things like bitwise operations, we use little endian format. If you check most libraries like `https://github.com/holiman/uint256` or `math/big`, they use little endian format. The reason being that most CPUs are little endian so it aligns naturally with how they handle the data. It also simplifies bitwise operations.

However, the tradeoff is that it's slightly harder to debug because we are used to big endian format.

core/trie/bitarray.go
```diff
@@ -0,0 +1,458 @@
...
+	words [4]uint64 // little endian (i.e. words[0] is the least significant)
+}
+
+func NewBitArray(length uint8, val uint64) *BitArray {
+	return new(BitArray).SetUint64(length, val)
+}
```
rodrigo-pino, line 33: `BitArray` escapes to heap in this case. What do you think of returning it as a value type instead?

core/trie/bitarray.go
```diff
@@ -0,0 +1,458 @@
...
+	// Compare only the first min(b.len, x.len) bits
+	minLen := b.len
+	if x.len < minLen {
+		minLen = x.len
+	}
+
+	return new(BitArray).MSBs(b, minLen).Equal(new(BitArray).MSBs(x, minLen))
+}
+
+// Sets b to the most significant 'n' bits of x.
```
rodrigo-pino, line 164:
```suggestion
// Sets the bit array to the most significant 'n' bits of x.
```
Suggesting this approach because someone reading the description might be unaware that `b` refers to the bit array itself

core/trie/bitarray.go
```diff
@@ -0,0 +1,458 @@
...
+//	MSBs(x, 0) = 0 (len=0)
+func (b *BitArray) MSBs(x *BitArray, n uint8) *BitArray {
+	if n >= x.len {
+		return b.Set(x)
+	}
+
+	return b.Rsh(x, x.len-n)
+}
+
+// Sets b to the longest sequence of matching most significant bits between two bit arrays.
```
rodrigo-pino, line 180:
```suggestion
// Sets the bit array to the longest sequence of matching most significant bits between two bit arrays.
```

core/trie/bitarray.go
```diff
@@ -0,0 +1,458 @@
...
+	//   0001 (difference at last position)
+	// We can then use the position of the first set bit and right-shift to get the common MSBs
+	diff := long.len - short.len
+	b.Rsh(long, diff).Xor(b, short)
+	divergentBit := findFirstSetBit(b)
+
+	return b.Rsh(short, divergentBit)
+}
+
+// Sets b = x >> n and returns b.
```
rodrigo-pino, line 218: Marking it here as well, inline with the previous suggestions

core/trie/bitarray.go
```diff
@@ -0,0 +1,458 @@
...
+func (b *BitArray) SetBytes(length uint8, data []byte) *BitArray {
+	b.setBytes32(data)
+	b.len = length
+	return b
+}
+
+// Sets b to the uint64 representation of a bit array.
+func (b *BitArray) SetUint64(length uint8, data uint64) *BitArray {
+	b.words[0] = data
+	b.len = length
```
rodrigo-pino, line 362: Can the length be inferred from data? What happens if length doesn't represent the data?

core/trie/proof.go
```diff
@@ -40,14 +40,14 @@ func (b *Binary) String() string {
...
 
 func (e *Edge) Hash(hash hashFunc) *felt.Felt {
-	length := make([]byte, len(e.Path.bitset))
-	length[len(e.Path.bitset)-1] = e.Path.len
+	var length [32]byte
+	length[31] = e.Path.len
 	pathFelt := e.Path.Felt()
-	lengthFelt := new(felt.Felt).SetBytes(length)
+	lengthFelt := new(felt.Felt).SetBytes(length[:])
 	return new(felt.Felt).Add(hash(e.Child, &pathFelt), lengthFelt)
```
rodrigo-pino, line 51: This felt escapes to heap as well. I know this is not part of the PR so need to change it. I am making emphasis on this because some month back, the Cairo VM got twice as fast by removing this type of code. Just a note

core/trie/bitarray.go
```diff
@@ -0,0 +1,458 @@
...
+		b.words[0] = (b.words[0] >> n) | (b.words[1] << (64 - n))
+		b.words[1] = (b.words[1] >> n) | (b.words[2] << (64 - n))
+		b.words[2] = (b.words[2] >> n) | (b.words[3] << (64 - n))
+		b.words[3] >>= n
+	}
+
+	return b
+}
+
+// Sets b = x ^ y and returns b.
```
rodrigo-pino, line 263: It happens all along the comments. Just minor nitpicks with the comment, if you agree with the suggestion, could you update the remaining ones?

core/trie/proof.go
weiihann: Why does `lengthFelt` escape to the heap? It's not returned anywhere

core/trie/bitarray.go
```diff
@@ -0,0 +1,458 @@
...
+
+	"github.com/NethermindEth/juno/core/felt"
+)
+
+const (
+	maxUint64 = uint64(math.MaxUint64) // 0xFFFFFFFFFFFFFFFF
+	bits8     = 8
+)
+
+var emptyBitArray = new(BitArray)
```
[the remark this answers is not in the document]
weiihann: I think the heap allocation is fine for this since it's intended to be shared, immutable and zero instance used in different places. I don't think we need to go for duplications.

core/trie/bitarray.go
weiihann: `length` cannot be inferred from the `data`. For example:

Given uint64 of `0b11111111` (8 bits) and length of `10`, this means that the actual bit array is illustrated as `0b0011111111`.

core/trie/proof.go
weiihann: Oh I see it now, you mean the returned `new(felt.Felt)`

weiihann: Yeah I'd avoid changing it in this PR, will add a TODO

core/state.go
```diff
@@ -139,10 +139,11 @@ func (s *State) globalTrie(bucket db.Bucket, newTrie trie.NewTrieFunc) (*trie.Tr
...
 	// fetch root key
 	rootKeyDBKey := dbPrefix
-	var rootKey *trie.Key
+	var rootKey *trie.BitArray
 	err := s.txn.Get(rootKeyDBKey, func(val []byte) error {
-		rootKey = new(trie.Key)
-		return rootKey.UnmarshalBinary(val)
+		rootKey = new(trie.BitArray)
+		rootKey.UnmarshalBinary(val)
+		return nil
```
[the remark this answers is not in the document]
weiihann: The approach that you suggested wouldn't work because `rootKey` still need to be used in the caller function (i.e. `globalTrie`). I tried something like:

```
	var rootKey trie.BitArray
	err := s.txn.Get(rootKeyDBKey, func(val []byte) error {
		rootKey.UnmarshalBinary(val)
		return nil
	})
``` 

Tests failed though. I'm just gonna leave a TODO there and avoid breaking things for now.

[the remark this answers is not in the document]
weiihann: That said, the entire fetch rootKey thing shouldn't even be there in the first place.

core/trie/bitarray.go
rodrigo-pino: Should there be a check when length is < 8 bits? Since it is clearly non intended by the programmer we can panic at than instance. What do you think?

weiihann: `length` < 8 bits, this is a valid input.

rodrigo-pino: Given a uint64 `0b11111111` would length < 8 be still valid?

weiihann: Hmm in that case it may cause some unintended behaviours (for example, `rsh`). My latest changes now clear the unused bits and ensure that they are zeros.

AnkushinDaniil (approved):

core/trie/bitarray.go
```diff
@@ -0,0 +1,655 @@
...
+	b.words[0] = x.words[0] ^ y.words[0]
+	b.words[1] = x.words[1] ^ y.words[1]
+	b.words[2] = x.words[2] ^ y.words[2]
+	b.words[3] = x.words[3] ^ y.words[3]
+	return b
+}
+
+// Checks if two bit arrays are equal
+func (b *BitArray) Equal(x *BitArray) bool {
+	// TODO(weiihann): this is really not a good thing to do...
```
rianhughes, line 350: Does this need addressed before merging?

core/trie/bitarray.go
```diff
@@ -0,0 +1,655 @@
...
+func (b *BitArray) clear() *BitArray {
+	b.len = 0
+	b.words[0], b.words[1], b.words[2], b.words[3] = 0, 0, 0, 0
+	return b
+}
+
+// Truncates the bit array to the specified length, ensuring that any unused bits are all zeros.
+//
+//nolint:mnd
+func (b *BitArray) truncateToLength() {
```
rianhughes, line 584: Maybe add an example?

// Example:
//   b := &BitArray{
//       len: 5,
//       words: [4]uint64{
//           0xFFFFFFFFFFFFFFFF, // full of 1s
//           0xFFFFFFFFFFFFFFFF,// full of 1s
//           0x0,
//           0x0,
//       },
//   }
//   b.truncateToLength()
//   // Result:
//   // b.words[0] = 0x000000000000001F
//   // b.words[1] = 0x0
//   // b.words[2] = 0x0
//   // b.words[3] = 0x0

core/trie/bitarray.go
```diff
@@ -0,0 +1,655 @@
...
+
+func (b *BitArray) Len() uint8 {
+	return b.len
+}
+
+// Returns the bytes representation of the bit array in big endian format
+func (b *BitArray) Bytes() []byte {
+	var res [32]byte
+
+	b.truncateToLength()
```
rianhughes, line 52: why do we need to call truncateToLength? When would the word-bits and length disagree?

core/trie/bitarray.go
```diff
@@ -0,0 +1,655 @@
...
+// It uses a little endian representation to do bitwise operations of the words efficiently.
+// For example, if len is 10, it means that the 2^9, 2^8, ..., 2^0 bits are used.
+// The max length is 255 bits (uint8), because our use case only need up to 251 bits for a given trie key.
+// Although words can be used to represent 256 bits, we don't want to add an additional byte for the length.
+type BitArray struct {
+	len   uint8     // number of used bits
+	words [4]uint64 // little endian (i.e. words[0] is the least significant)
+}
+
+func NewBitArray(length uint8, val uint64) BitArray {
```
rianhughes, line 31: Why do we only set the first word?

core/trie/bitarray.go
```diff
@@ -0,0 +1,655 @@
...
+	b.words[0] = binary.BigEndian.Uint64(res[24:32])
+}
+
+func (b *BitArray) setBytes32(data []byte) {
+	_ = data[31]
+	b.words[3] = binary.BigEndian.Uint64(data[0:8])
+	b.words[2] = binary.BigEndian.Uint64(data[8:16])
+	b.words[1] = binary.BigEndian.Uint64(data[16:24])
+	b.words[0] = binary.BigEndian.Uint64(data[24:32])
+}
```
rianhughes, line 530: This can panic right? Also what is the purpose of _ = data[31]?

core/trie/bitarray.go
```diff
@@ -0,0 +1,655 @@
...
+		mask := maxUint64 >> (192 - n)
+		b.words[2] &= mask
+		b.words[3] = 0
+	default:
+		mask := maxUint64 >> (256 - uint16(n))
+		b.words[3] &= mask
+	}
+
+	return b
+}
```
rianhughes, line 105: I wonder if it would be more efficient to assign the relevant words from x in each case, rather than update the words in b.Set(x) and the switch statement? Eg something like
```
switch {
	case n == 0:
		b.words = [4]uint64{0, 0, 0, 0}
	case n <= 64:
		b.words[0] = x.words[0] & (maxUint64 >> (64 - n))
		b.words[1], b.words[2], b.words[3] = 0, 0, 0
	case n <= 128:
		b.words[0] = x.words[0]
		b.words[1] = x.words[1] & (maxUint64 >> (128 - n))
		b.words[2], b.words[3] = 0, 0
	case n <= 192:
		b.words[0] = x.words[0]
		b.words[1] = x.words[1]
		b.words[2] = x.words[2] & (maxUint64 >> (192 - n))
		b.words[3] = 0
	default:
		b.words[0] = x.words[0]
		b.words[1] = x.words[1]
		b.words[2] = x.words[2]
		b.words[3] = x.words[3] & (maxUint64 >> (256 - uint16(n)))
	}
```

core/trie/bitarray.go
weiihann: Unfortunately changing this would mean I have to change the other parts that are using this method. This might break stuff, so I'd rather use the same behaviour as`trie.Key`

core/trie/bitarray.go
weiihann: Hmm yeah this should be unnecessary now, will remove it.

core/trie/bitarray.go
weiihann: It's just a convenient function, used in test only (for now).

core/trie/bitarray.go
weiihann: It's a compiler optimization thing, for details see https://golang.org/issue/14808

core/trie/bitarray.go
weiihann: Yes! This is a better approach :)

core/trie/bitarray.go
```diff
@@ -0,0 +1,655 @@
...
+	return n + 1, err
+}
+
+// Deserialises the BitArray from a bytes buffer in the following format:
+// - First byte: length of the bit array (0-255)
+// - Remaining bytes: the necessary bytes included in big endian order
+// Example:
+//
+//	[0x0A, 0x03, 0xFF] -> BitArray{len: 10, words: [4]uint64{0x03FF}}
+func (b *BitArray) UnmarshalBinary(data []byte) {
```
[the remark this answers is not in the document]
weiihann: Yup, I think we should

core/trie/bitarray.go
```diff
@@ -0,0 +1,681 @@
...
+	if carry == 1 {
+		return -1
+	}
+
+	if d0|d1|d2|d3 == 0 {
+		return 0
+	}
+
+	return 1
+}
```
rianhughes, line 681: Two points related to this:
1. why did you go for this approach over something where we effectievly compare each word directly? eg
```
// Lengths are equal, compare words from the most significant to the least significant
    for i := len(b.words) - 1; i >= 0; i-- {
        if b.words[i] < x.words[i] {
            return -1
        }
        if b.words[i] > x.words[i] {
            return 1
        }
    }

    // All words are equal
    return 0
```

2. The b.truncateLength() in SetUint64() causes this to break. At least the following test fails
```
func TestCmp2(t *testing.T) {
	b := NewBitArray(1, 1000)
	x := NewBitArray(1, 100)
	require.Equal(t, b.Cmp(&x), int(1))

	b = NewBitArray(1, 100)
	require.Equal(t, b.Cmp(&x), int(0))

	b = NewBitArray(1, 10)
	require.Equal(t, b.Cmp(&x), int(-1))

}
```

core/trie/bitarray.go
```diff
@@ -0,0 +1,681 @@
...
+var emptyBitArray = new(BitArray)
+
+// Represents a bit array with length representing the number of used bits.
+// It uses a little endian representation to do bitwise operations of the words efficiently.
+// For example, if len is 10, it means that the 2^9, 2^8, ..., 2^0 bits are used.
+// The max length is 255 bits (uint8), because our use case only need up to 251 bits for a given trie key.
+// Although words can be used to represent 256 bits, we don't want to add an additional byte for the length.
+type BitArray struct {
+	len   uint8     // number of used bits
+	words [4]uint64 // little endian (i.e. words[0] is the least significant)
```
rianhughes, line 29: Think it's worth adding a note that the uint64 is big endian for all us dummys out there? :) Seems like it can be relevant when understanding the low level operations of BitArray

core/trie/bitarray.go
rianhughes: Ah, nvm about the test. I think the len should have been higher

rianhughes: Although I'm wondering if the len should be determined by the uint64 argument in `NewBitArray(len uint8, data uint64)`?

core/trie/bitarray.go
weiihann: You're not a dummy :)

Anyways, the endian-ness of uint64 depends on the actual CPU architecture used. But golang abstracts this out for us. So when we perform bitwise operations, we should think of it in terms of big endian, because it's more intuitive for humans, but the underlying implementation could be the opposite. This applies to the usage of BitArray too.

core/trie/bitarray.go
```diff
@@ -0,0 +1,681 @@
...
+}
+
+func NewBitArray(length uint8, val uint64) BitArray {
+	var b BitArray
+	b.SetUint64(length, val)
+	return b
+}
```
rianhughes, line 36: suggestion
```
func NewBitArray(length uint8, word0 uint64) BitArray {
	var b BitArray
	b.SetUint64(length, word0)
	return b
}
```

core/trie/bitarray.go
weiihann: Not really, in our use case we may want leading 0 bits too.

rianhughes: Yup, after thinking about it a bit more I can see why this is the case. Thanks

core/trie/bitarray.go
rianhughes: Hm, interesting

core/trie/bitarray.go
rianhughes: >The current implementation is faster because it leverages CPU's native subtraction-with-carry operations in a single pass with fewer branches, rather than performing multiple sequential comparisons that could cause branch prediction misses and pipeline stalls.

Cool, I wasn't aware of this. Nice one

core/trie/bitarray.go
```diff
@@ -0,0 +1,681 @@
...
+	// but we still shift `n` bits
+	if n > maxUint8-x.len {
+		b.len = maxUint8
+	} else {
+		b.len = x.len + n
+	}
+
+	switch {
+	case n == 0:
+		return b
```
rianhughes, line 279: not that important, but I think we can remove this case

core/trie/bitarray.go
```diff
@@ -0,0 +1,679 @@
...
+
+// Lsh sets the bit array to x << n and returns the bit array.
+//
+//nolint:mnd
+func (b *BitArray) Lsh(x *BitArray, n uint8) *BitArray {
+	b.Set(x)
+
+	if x.len == 0 || n == 0 {
+		return b
+	}
```
rianhughes, line 267: What do you think about moving b.Set(x) into the if statement? The words get overwritten below if this if-statmenet doesn't get triggered

core/trie/bitarray.go
```diff
@@ -0,0 +1,679 @@
...
+	if y.len == 0 {
+		return b.Set(x)
+	}
+
+	// First copy x
+	b.Set(x)
+
+	// Then shift left by y's length and OR with y
+	return b.Lsh(b, y.len).Or(b, y)
+}
```
rianhughes, line 323: If y has full len, then we can also exit early by just returning b.Set(y)?

rianhughes: Also, Lsh calls b.Set(x) (and even if this is moved into the if statement as commented above - it still effectievly sets X), soo I think we can do something like this
```
// First copy x
	// b.Set(x)

	// Then shift left by y's length and OR with y
	return b.Lsh(x, y.len).Or(b, y)
```

core/trie/bitarray.go
weiihann: No, it's intended. We have to first set `b` to `x`, so that the switch statement gets executed. If `x` is 0 bits, we can't do shifting, so we return it, which means at this point `b` is also 0 bits.

core/trie/bitarray.go
weiihann: Yup, both are valid, will make the change

core/trie/bitarray.go
rianhughes: Sorry, not sure I understand (or was clear enough). I mean something like this
```
func (b *BitArray) Lsh(x *BitArray, n uint8) *BitArray {
	if x.len == 0 || n == 0 {
	        b.Set(x)
		return b
	}

	// If the result will overflow, we set the length to the max length
	// but we still shift `n` bits
	if n > maxUint8-x.len {
		b.len = maxUint8
	} else {
		b.len = x.len + n
	}

	switch {
	case n >= 192:
		b.lsh192(x)
		n -= 192
		b.words[3] <<= n
	case n >= 128:
	...
	}
	```

Comment (rianhughes) [reply core/trie/bitarray.go:267]: Also, if `x.len==0`, is there a difference between calling `b.clear()` and `b.Set(x)`? Basically, is there ever a case where a `BitArray` has zero length, but non-zero words?

File: core/trie/bitarray.go [hunk 29]
```diff
@@ -0,0 +1,676 @@
...
+
+// Interprets the data as the big-endian bytes, sets the bit array to that value and returns it.
+// If the data is larger than 32 bytes, only the first 32 bytes are used.
+func (b *BitArray) SetBytes(length uint8, data []byte) *BitArray {
+	b.setBytes32(data)
+	b.len = length
+	b.truncateToLength()
+	return b
+}
```
rianhughes, line 478: This panics if data isn't at least 32 bytes

weiihann: Yup, sounds right, added the new changes

rianhughes: That new `SetBytes` function is wild lol :) Guess it's more performant than handling in a loop or something?

rianhughes (approved):
Very nice work!!

core/trie/bitarray.go
weiihann: yeah hahahah

PR #2337, opened 2024-12-18
Bootnode Registry

AnkushinDaniil:
This pull request introduces significant changes to the `l1` package, primarily focusing on integrating the IP Address Registry functionality. The most important changes include adding the IP Address Registry ABI, updating the `EthSubscriber` to handle IP address events, and modifying the `Client` to subscribe to these events and forward them to the P2P layer.

### IP Address Registry Integration:

* [`l1/abi/ip_address_registry.json`](diffhunk://#diff-9a0f51c7ab77a11006ee230d16a75ae09a3217db364385f1f5c379ea02c073a8R1-R217): Added the ABI for the IP Address Registry contract, defining functions, errors, and events related to IP address management.
* [`l1/eth_subscriber.go`](diffhunk://#diff-33d45c810a23e4517685e6a1c1080a1bc281ce34e77dee7b4c1357eed882d1edR11): Updated the `EthSubscriber` struct to include the IP Address Registry and its filterer. Added methods to watch for IP address additions and removals, and to get the list of IP addresses. [[1]](diffhunk://#diff-33d45c810a23e4517685e6a1c1080a1bc281ce34e77dee7b4c1357eed882d1edR11) [[2]](diffhunk://#diff-33d45c810a23e4517685e6a1c1080a1bc281ce34e77dee7b4c1357eed882d1edR25-R31) [[3]](diffhunk://#diff-33d45c810a23e4517685e6a1c1080a1bc281ce34e77dee7b4c1357eed882d1edL37-R85)
* [`l1/l1.go`](diffhunk://#diff-7f7fd9a26e7cb24ff29bbaa5130bd597f6ec2c15bec9414dfec4863850ca8570R1): Added methods to the `Client` to subscribe to IP address events and forward them to the P2P layer. Modified the `Client` constructor to accept a channel for forwarding these events. [[1]](diffhunk://#diff-7f7fd9a26e7cb24ff29bbaa5130bd597f6ec2c15bec9414dfec4863850ca8570R1) [[2]](diffhunk://#diff-7f7fd9a26e7cb24ff29bbaa5130bd597f6ec2c15bec9414dfec4863850ca8570R14-R29) [[3]](diffhunk://#diff-7f7fd9a26e7cb24ff29bbaa5130bd597f6ec2c15bec9414dfec4863850ca8570R45-R50) [[4]](diffhunk://#diff-7f7fd9a26e7cb24ff29bbaa5130bd597f6ec2c15bec9414dfec4863850ca8570R60) [[5]](diffhunk://#diff-7f7fd9a26e7cb24ff29bbaa5130bd597f6ec2c15bec9414dfec4863850ca8570R96-R131) [[6]](diffhunk://#diff-7f7fd9a26e7cb24ff29bbaa5130bd597f6ec2c15bec9414dfec4863850ca8570L112-R169) [[7]](diffhunk://#diff-7f7fd9a26e7cb24ff29bbaa5130bd597f6ec2c15bec9414dfec4863850ca8570R181) [[8]](diffhunk://#diff-7f7fd9a26e7cb24ff29bbaa5130bd597f6ec2c15bec9414dfec4863850ca8570R230-R295)

### DIFF
--- a/l1/eth_subscriber.go
+++ b/l1/eth_subscriber.go
@@ -50,6 +50,18 @@ func (s *EthSubscriber) WatchLogStateUpdate(ctx context.Context, sink chan<- *co
 	return s.filterer.WatchLogStateUpdate(&bind.WatchOpts{Context: ctx}, sink)
 }
 
+func (s *EthSubscriber) WatchIPAdded(ctx context.Context, sink chan<- *contract.BootnodeRegistryIPAdded) (event.Subscription, error) {
+	return s.bootnodeRegistryFilterer.WatchIPAdded(&bind.WatchOpts{Context: ctx}, sink)
+}
+
+func (s *EthSubscriber) WatchIPRemoved(ctx context.Context, sink chan<- *contract.BootnodeRegistryIPRemoved) (event.Subscription, error) {
+	return s.bootnodeRegistryFilterer.WatchIPRemoved(&bind.WatchOpts{Context: ctx}, sink)
+}
+
+func (s *EthSubscriber) GetIPAddresses(ctx context.Context, ip common.Address) ([]string, error) {
+	return s.bootnodeRegistry.GetIPAddresses(&bind.CallOpts{Context: ctx})
+}
+
 func (s *EthSubscriber) ChainID(ctx context.Context) (*big.Int, error) {
 	reqTimer := time.Now()
 	chainID, err := s.ethClient.ChainID(ctx)
--- a/l1/eth_subscriber.go
+++ b/l1/eth_subscriber.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/NethermindEth/juno/l1/contract"
+	"github.com/NethermindEth/juno/utils"
 	"github.com/ethereum/go-ethereum/accounts/abi/bind"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
@@ -17,15 +18,17 @@ import (
 )
 
 type EthSubscriber struct {
-	ethClient *ethclient.Client
-	client    *rpc.Client
-	filterer  *contract.StarknetFilterer
-	listener  EventListener
+	ethClient                *ethclient.Client
+	client                   *rpc.Client
+	filterer                 *contract.StarknetFilterer
+	listener                 EventListener
+	bootnodeRegistry         *contract.BootnodeRegistry
+	bootnodeRegistryFilterer *contract.BootnodeRegistryFilterer
 }
 
 var _ Subscriber = (*EthSubscriber)(nil)
 
-func NewEthSubscriber(ethClientAddress string, coreContractAddress common.Address) (*EthSubscriber, error) {
+func NewEthSubscriber(ethClientAddress string, network *utils.Network) (*EthSubscriber, error) {
 	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
 	defer cancel()
 
@@ -34,15 +37,34 @@ func NewEthSubscriber(ethClientAddress string, coreContractAddress common.Addres
 		return nil, err
 	}
 	ethClient := ethclient.NewClient(client)
-	filterer, err := contract.NewStarknetFilterer(coreContractAddress, ethClient)
+	filterer, err := contract.NewStarknetFilterer(network.CoreContractAddress, ethClient)
 	if err != nil {
 		return nil, err
 	}
+
+	var (
+		bootnodeRegistry         *contract.BootnodeRegistry
+		bootnodeRegistryFilterer *contract.BootnodeRegistryFilterer
+	)
+	if network.BootnodeRegistry != emptyBootnodeRegistry {
+		fmt.Println("Bootnode registry is not empty")
+		bootnodeRegistry, err = contract.NewBootnodeRegistry(network.BootnodeRegistry, ethClient)
+		if err != nil {
+			return nil, err
+		}
+		bootnodeRegistryFilterer, err = contract.NewBootnodeRegistryFilterer(network.BootnodeRegistry, ethClient)
+		if err != nil {
+			return nil, err
+		}
+	}
+
 	return &EthSubscriber{
-		ethClient: ethClient,
-		client:    client,
-		filterer:  filterer,
-		listener:  SelectiveListener{},
+		ethClient:                ethClient,
+		client:                   client,
+		filterer:                 filterer,
+		listener:                 SelectiveListener{},
+		bootnodeRegistry:         bootnodeRegistry,
+		bootnodeRegistryFilterer: bootnodeRegistryFilterer,
 	}, nil
 }
 
--- a/node/node.go
+++ b/node/node.go
@@ -174,6 +174,33 @@ func New(cfg *Config, version string, logLevel *utils.LogLevel) (*Node, error) {
 		services = append(services, plugin.NewService(p))
 	}
 
+	throttledVM := NewThrottledVM(vm.New(false, log), cfg.MaxVMs, int32(cfg.MaxVMQueue))
+
+	var syncReader sync.Reader = &sync.NoopSynchronizer{}
+	if synchronizer != nil {
+		syncReader = synchronizer
+	}
+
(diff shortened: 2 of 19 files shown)


utils/network.go
```diff
@@ -22,6 +22,7 @@ type Network struct {
 	L1ChainID           *big.Int           `json:"l1_chain_id" validate:"required"`
 	L2ChainID           string             `json:"l2_chain_id" validate:"required"`
 	CoreContractAddress common.Address     `json:"core_contract_address" validate:"required"`
+	IPAddressRegistry   *common.Address    `json:"ip_address_registry"`
```
weiihann, line 25: I don't think a pointer is needed here. A value type will do. For comparison, we can do `!= common.Address{}` instead of `!= nil`. So we don't need any unnecessary dereferencing when parsing this parameter in other methods.

For cleaner semantics, we can also consider declaring `const emptyIPAddressRegistry = common.Address{}` so the reader immediately knows what the check is for.

node/node.go
```diff
@@ -170,6 +170,35 @@ func New(cfg *Config, version string) (*Node, error) { //nolint:gocyclo,funlen
...
+		var l1Client *l1.Client
+		l1Client, err = newL1Client(cfg.EthNode, cfg.Metrics, chain, log, l1ToP2P)
+		if err != nil {
+			return nil, fmt.Errorf("create L1 client: %w", err)
+		}
+		services = append(services, l1Client)
+		rpcHandler.WithL1Client(l1Client.L1())
+	} else {
+		close(l1ToP2P)
+	}
```
weiihann, line 200: We don't need to close the channel, just declare it as `var`. For example:

```
var l1ToP2P chan p2p.IPAddressRegistryEvent
if !cfg.DisableL1Verification {
if cfg.EthNode == "" {
	return nil, fmt.Errorf("ethereum node address not found; Use --disable-l1-verification flag if L1 verification is not required")
}

l1ToP2P = make(chan p2p.IPAddressRegistryEvent)
var l1Client *l1.Client
l1Client, err = newL1Client(cfg.EthNode, cfg.Metrics, chain, log, l1ToP2P)
if err != nil {
	return nil, fmt.Errorf("create L1 client: %w", err)
}
services = append(services, l1Client)
rpcHandler.WithL1Client(l1Client.L1())
}
```

utils/network.go
weiihann: Also, add a check for `listenForL1Events` in `Service`. For example:

```
if s.l1events != nil {
   go s.listenForL1Events()
}
```

p2p/p2p.go
```diff
@@ -248,6 +266,43 @@ func (s *Service) Run(ctx context.Context) error {
...
+			case Remove:
+				if err := s.host.Network().ClosePeer(peerInfo.ID); err != nil {
+					s.log.Warnw("Failed to disconnect from peer", "peer", registryEvent.Address, "err", err)
+				} else {
+					s.log.Debugw("Disconnected from peer", "peer", registryEvent.Address)
+				}
+				s.host.Peerstore().RemovePeer(peerInfo.ID)
+				s.log.Debugw("Removed peer from peerstore", "peer", peerInfo.ID)
+				s.dht.RoutingTable().RemovePeer(peerInfo.ID)
+				s.log.Debugw("Removed peer from routing table", "peer", peerInfo.ID)
```
weiihann, line 300: For the debugging logs, please help to specify that the peer removal is specifically from L1 events.

That said, maybe we should create wrapper structs so we can customize logs directly.

p2p/p2p.go
```diff
@@ -360,9 +415,13 @@ func loadPeers(database db.DB) ([]peer.AddrInfo, error) {
...
 		prefix := db.Peer.Key()
+		var peerIDBytes []byte
 		for it.Seek(prefix); it.Valid(); it.Next() {
-			peerIDBytes := it.Key()[len(prefix):]
-			peerID, err := peer.IDFromBytes(peerIDBytes)
+			peerIDBytes = it.Key()
+			if !bytes.HasPrefix(peerIDBytes, prefix) {
+				break
+			}
+			peerID, err := peer.IDFromBytes(peerIDBytes[len(prefix):])
```
weiihann, line 424: Why is this change needed? The db iterator should give keys which match the prefix, so we don't need to do the prefix check again.

weiihann:
I think the name `IpAddress` may be too vague on what this component actually does. In fact, the address from L1 is not an IP address but rather a `multiaddr`. 

If I understand correctly, there are some peers registered on the L1 smart contract. So we listen to those and adjust the peerstore accordingly. If I'm not wrong, these peers should be bootnodes, so `BootnodeRegistry` would be more suitable.

Also, not sure if this has been discussed before, who is allowed to add/remove the peers on L1?

p2p/p2p.go
AnkushinDaniil: We need to do this, without this prefix check the peer storage won't work. db iterator gives keys that don't match the prefix

utils/network.go
AnkushinDaniil: [Change pointer to value](https://github.com/NethermindEth/juno/pull/2337/commits/4f66c6d4e993a4c6d0e51230b47214cd86e4431f)
[Change channel initialization](https://github.com/NethermindEth/juno/pull/2337/commits/26957c09bf77551f5e98564abb294a87c37c84f7)

node/node.go
AnkushinDaniil: [Change channel initialization](https://github.com/NethermindEth/juno/pull/2337/commits/26957c09bf77551f5e98564abb294a87c37c84f7)

p2p/p2p.go
AnkushinDaniil: [Specify that the peer removal is specifically from L1 events](https://github.com/NethermindEth/juno/pull/2337/commits/1f1be6286764cba71e5ffd4bd9d7d0d994c9f5d2)

p2p/p2p.go
weiihann: I see the bug now. The `NewIterator` should create an iterator with lower bound specified. Let me create a PR real quick.

weiihann: Please check this:
https://github.com/NethermindEth/juno/pull/2357

AnkushinDaniil: [Remove prefix check](https://github.com/NethermindEth/juno/pull/2337/commits/478e8cd923adeb81ac2ee7ed701a953a5e7cd2ed)

l1/eth_subscriber.go
```diff
@@ -17,15 +18,17 @@ import (
...
-	ethClient *ethclient.Client
-	client    *rpc.Client
-	filterer  *contract.StarknetFilterer
-	listener  EventListener
+	ethClient                 *ethclient.Client
+	client                    *rpc.Client
+	filterer                  *contract.StarknetFilterer
+	listener                  EventListener
+	ipAddressRegistry         *contract.IPAddressRegistry
+	ipAddressRegistryFilterer *contract.IPAddressRegistryFilterer
```
weiihann, line 26: I believe this should also be renamed to `BootnodeRegistry`? Alongside the other places.

Also, the data parsed is not IP address, but rather libp2p's multiaddr. Please rename the IP addresses to multiaddrs whichever is appropriate.

weiihann:
This is good work, but I have second thoughts on whether we should incorporate this feature in Juno. I see a few problems with this feature:

1. Responsibility of the L1 peer store
By default, our internal team should be responsible for adding/removing/maintaining the peer store on L1. That means we have to maintain the L1 account that does all these things. This adds extra work because we have to ensure that this piece of information is passed on to a different person. If we lose access to this account, we have to redeploy a new one and update the address in Juno.

2. Security
If the L1 account's private key is leaked, a malicious 3rd party may add their own node or remove the good nodes. This feature then becomes a potential attack vector, which is way too unnecessary. It's also not easy to detect if the private key is leaked or not. Perhaps we can do multisig, but is it worth the extra effort?

Bootnodes are meant to help new users to find new peers. They are centralized and trusted. I don't see the benefits of putting these info on-chain, except that it is cool. The simplest and most straightforward way is to directly put this information into our codebase and call it a day. No need to overengineer and put extra effort into a feature that doesn't bring many benefits.

PR #2340, opened 2024-12-18
Add IsNil()

IronGauntlets:
IsNil() checks if the underlying value of the interface is nil.

### DIFF
--- a/utils/check.go
+++ b/utils/check.go
@@ -1,0 +1,34 @@
+package utils
+
+import "reflect"
+
+// IsNil checks if the underlying value of the interface is nil.
+//
+// In Golang, an interface is boxed by an underlying type and value; both of them need to be nil for the interface to be
+// nil. See the following examples:
+//
+//	var i interface{}
+//	fmt.Println(i == nil) // true
+//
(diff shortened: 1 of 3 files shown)


utils/check.go
```diff
@@ -0,0 +1,34 @@
...
+//	fmt.Println(i == nil) // true
+//
+//	var p *int
+//	var i interface{} = p
+//	fmt.Println(i == nil) // false!
+//
+// A solution for this is to use i == nil || reflect.ValueOf(i).IsNil()), however, this can cause a panic as not all
+// reflect.Value has IsNil() defined. Therefore, default is to return false. For example, reflect.Array cannot be nil,
+// hence calling IsNil() will cause a panic.
+func IsNil(i interface{}) bool {
```
kirugan, line 20: nit: `interface{}` -> `any`

IronGauntlets: I just wanted to make it explicit that we are talking about `interface{}` as that what the comments mention. If you think `any` is better I will change.

rianhughes (approved):

PR #2341, opened 2024-12-19
implement a mempool for the sequencer

rianhughes:
This PR implements a mempool required for the sequencer.


mempool/mempool.go
```diff
@@ -0,0 +1,231 @@
...
+	DeclaredClass core.Class
+}
+
+const (
+	poolLengthKey = "poolLength"
+	headKey       = "headKey"
+	tailKey       = "tailKey"
+)
+
+// Pool stores the transactions in a linked list for its inherent FCFS behaviour
```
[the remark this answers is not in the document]
rianhughes: Yup, first come first served!

mempool/mempool.go
```diff
@@ -0,0 +1,231 @@
...
+	txPushed  chan struct{}
+}
+
+func New(poolDB db.DB) *Pool {
+	return &Pool{
+		db:        poolDB,
+		validator: func(_ *BroadcastedTransaction) error { return nil },
+		txPushed:  make(chan struct{}, 1),
+	}
+}
```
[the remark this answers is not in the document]
rianhughes: So running `go build -gcflags="-m" mempool.go` suggests that returning `*Pool` will cause an escape to the heap. Changing the return type to `Pool` doesn't seem to prevent an escape to the heap since `validator` is still allocated there.

Returning the pointer makes it cheaper to pass around Pool. But leaving it allocated to the heap means the garbage collector will have to check it every time, throughout the entire life time of the Juno program. Honestly, I'm not sure how big of a deal this is. If the main goal of this suggestion is to prevent `Pool` from being allocated to the heap, then we would have to decouple `validator` from `Pool`.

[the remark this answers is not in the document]
jeffrydegrande: In this case, where your struct has multiple, potentially complex members Id go with a pointer purely to avoid creating e .g undesired copies of things like mutexes now or in the future.

mempool/mempool.go
```diff
@@ -0,0 +1,253 @@
...
+	return txn.Set([]byte(headKey), head.Marshal())
+}
+
+func (p *Pool) tailHash(txn db.Transaction) (*felt.Felt, error) {
+	var tail *felt.Felt
+	return tail, txn.Get([]byte(tailKey), func(b []byte) error {
+		tail = new(felt.Felt).SetBytes(b)
+		return nil
+	})
+}
```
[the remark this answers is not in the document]
rianhughes: Yup, implemented this change (and a similar one for headHash)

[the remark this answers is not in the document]
rianhughes: Also, just fyi, `new(felt.Felt)` seems to cause a heap allocation

mempool/mempool.go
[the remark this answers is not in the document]
rodrigo-pino: I ran this locally on my PC and it says the function escapes to the heap, but I don't think the entire object does. Usually is good enough to trust the go compiler escape analysis, for example if you return by value an object big enough it will move it to the heap without any consideration, even if you meant to return it as a value type.

Regarding the object on the heap, if it is going to be there forever, then the runtime has optimization for those, where each object has kind of a lifetime. After it is checked multiple time, it starts being checked less recently until there is a point where it doesn't get checked at all until a few cycles have gone by.

@jeffrydegrande point is a valid one and I hadn't considered it. I've little experience with concurrent Go so I cannot provide much argument either in favor or against.

As a final point, an object which holds pointer to others is not uncommon. Think of go slices (`[]T`) where they are basically a triplet with (`len`, `cap`, `ptr`) where `len` is the len property, the `cap` how much memory they have reserved and `ptr` which point to the actual place in memory. The triplet (usually) lives on the stack while the elements of the array live on the heap.

At the end, I believe this specific case won't matter in the long term. I would suggest to return it as a value type (I believe this route would be novel to you) and see what happens. Worst case scenario, we switch it back. Nonetheless, @rianhughes choose what you think is best!

mempool/mempool.go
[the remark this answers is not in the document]
rodrigo-pino: The same could be applied to `headHash`. Think that this optimization is most important with transactions that are getting executed a lot. In the previous example, with `New` for `Pool` then if it only executed once it will make no real difference; on the other hand I believe this will be executed on a per transaction basis.

mempool/mempool.go
```diff
@@ -0,0 +1,253 @@
...
+	DeclaredClass core.Class
+}
+
+const (
+	poolLengthKey = "poolLength"
+	headKey       = "headKey"
+	tailKey       = "tailKey"
+)
+
+// Pool stores the transactions in a linked list for its inherent FCFS behaviour
```
rodrigo-pino, line 27: Should this description be on top of the Pool type defintiion (?)

weiihann:
From my understanding, the current implementation seems to work as such:
- Receives new tx
- Validates the tx
- Puts new tx into DB directly
- The order is FCFS

This architecture is simple and straightforward, but I'm not sure how effective it would work it would be in real life. Here are a few things we need to consider:

1. In-memory mempool
It's way too expensive to store and retrieve txs only through the DB. There should be an in-memory implementation of it for fast access.

2. Order of sequencing
The current method is FCFS, but how about other factors like fees?

3. Transaction validation
I'm not fully aware what the `ValidatorFunc` actually contains, but few things we need to check for:
- Account balance checks
- Nonce validation
- Gas limit validation

There should also be a separation between pending txs and ready txs. Ready tx means tx that has higher nonce than the next immediate nonce of an account. Ready tx means the next immediate one and it's ready to be executed.

Mempool design and architecture could get complex, as it can be exploited for MEV. I'm not sure how far we want to take it at the current stage, but I'll suggest to look into existing architecture like [Erigon's mempool design](https://erigon.gitbook.io/erigon/advanced-usage/txpool) and go-ethereum one and see how they do certain things.

mempool/mempool.go
```diff
@@ -0,0 +1,253 @@
...
+			return err
+		}
+		nextTxn = headElem.Txn
+		return nil
+	})
+}
+
+// Remove removes a set of transactions from the pool
+func (p *Pool) Remove(hash ...*felt.Felt) error {
+	return errors.New("not implemented")
```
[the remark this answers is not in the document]
rianhughes: Yup, every time a block is added to the chain, we'll have to remove any txns in the mempool that have already been included in the new block

mempool/mempool.go
```diff
@@ -0,0 +1,253 @@
...
+}
+
+// WithValidator adds a validation step to be triggered before adding a
+// BroadcastedTransaction to the pool
+func (p *Pool) WithValidator(validator ValidatorFunc) *Pool {
+	p.validator = validator
+	return p
+}
+
+func (p *Pool) rejectDuplicateTxn(userTxn *BroadcastedTransaction) error {
```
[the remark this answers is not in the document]
rianhughes: Yup, good point with the private-public ordering. I'll move it to the bottom. 

As for the second comment, it's okay that the transaction already exists in the mempool, we just don't want to have duplicates. Plus it can be useful for users to know the txn already exists in the mempool.

mempool/mempool.go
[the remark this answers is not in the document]
rianhughes: Cool, I'll leave this as it is then. Thanks for the interesting viewpoint and comments though :)

mempool/mempool.go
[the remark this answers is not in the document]
rianhughes: I've updated the code such that tail is moved to the stack

mempool/mempool.go
```diff
@@ -0,0 +1,254 @@
...
+type BroadcastedTransaction struct {
+	Transaction   core.Transaction
+	DeclaredClass core.Class
+}
+
+const (
+	poolLengthKey = "poolLength"
+	headKey       = "headKey"
+	tailKey       = "tailKey"
+)
```
weiihann, line 25: What's the reason of using this prefix approach instead of declaring new buckets? Doing this may accidentally corrupt the database when we add more buckets (i.e. byte(h) == 104). I know it's unlikely that we will add that many buckets, but better to stick to the current implementation. Also, other database features such as the `db size` command will also need to know all prefixes in advance.

mempool/mempool.go
```diff
@@ -0,0 +1,254 @@
...
+type storageElem struct {
+	Txn      BroadcastedTransaction
+	NextHash *felt.Felt
+}
+
+// Pool stores the transactions in a linked list for its inherent FCFS behaviour
+type Pool struct {
+	db        db.DB
+	validator ValidatorFunc
+	txPushed  chan struct{}
```
weiihann, line 36: What is this field used for? It seems to me that it's only used to enforce sequential processing. Wouldn't a `mutex` be more suitable?

mempool/mempool.go
```diff
@@ -0,0 +1,254 @@
...
+	return item, err
+}
+
+func (p *Pool) putElem(txn db.Transaction, itemKey *felt.Felt, item *storageElem) error {
+	itemBytes, err := encoder.Marshal(item)
+	if err != nil {
+		return err
+	}
+	return txn.Set(itemKey.Marshal(), itemBytes)
+}
```
weiihann, line 238: I don't think we can store transaction hash as the key directly. It must be prefixed first.

mempool/mempool_test.go
```diff
@@ -0,0 +1,131 @@
...
+	pool := mempool.New(testDB)
+	blockchain.RegisterCoreTypesToEncoder()
+
+	t.Run("empty pool", func(t *testing.T) {
+		l, err := pool.Len()
+		require.NoError(t, err)
+		assert.Equal(t, uint64(0), l)
+
+		_, err = pool.Pop()
+		require.ErrorIs(t, err, db.ErrKeyNotFound)
```
rodrigo-pino, line 29: What do you think of returning a custom error when Pop is called with an empty list. This comes from the idea that as a user of the mempool getting a `db.ErrKeyNotFound` might not be super intuitive at a first glance after a Pop.

I am not sure which would be the best way to implement it, assuming we would have an in-memory Mempool then it would be a check for the length at the beginning and returning a custom error message. 

Final note, I am still not very familiar with the codebase, so what I am proposing might have some cons I don't see. At the very least, what do you think of adding to the function description that we will return a `db.ErrKeyNotFound` when Pop is empty.

mempool/mempool_test.go
```diff
@@ -0,0 +1,131 @@
...
+
+		l, err := pool.Len()
+		require.NoError(t, err)
+		assert.Equal(t, i, l)
+	}
+
+	// consume some (remove 1,2, keep 3)
+	for i := uint64(1); i < 3; i++ {
+		txn, err := pool.Pop()
+		fmt.Println("txn", txn.Transaction.Hash().String())
```
rodrigo-pino, line 48: Rogue print 👀

mempool/mempool.go
rianhughes: Yup good point, I've switched to the bucket implementation.

mempool/mempool.go
rianhughes: Yup good point, it's now prefixed

mempool/mempool.go
rianhughes: The `Builder` will use this field (specifically the `Wait()` function) to wait for new transactions to be added to the mempool, before trying to pop any more functions

```
func (b *Builder) listenPool(ctx context.Context) error {
	for {
		if err := b.depletePool(ctx); err != nil { **// This will pop transaction from the pool**
			if !errors.Is(err, db.ErrKeyNotFound) {
				return err
			}
		}
		select {
		case <-ctx.Done():
			return nil
		case <-b.pool.Wait():
			continue
		}
	}
}
```

mempool/mempool_test.go
rianhughes: Yup, I'm happy to return a custom error. I've updated the code so that it returns a "transaction pool is empty error"

mempool/bucket.go
```diff
@@ -0,0 +1,21 @@
...
+	Head   Bucket = iota // key of the head node
+	Tail                 // key of the tail node
+	Length               // number of transactions
+	Node
+)
+
+// Key flattens a prefix and series of byte arrays into a single []byte.
+func (b Bucket) Key(key ...[]byte) []byte {
+	return append([]byte{byte(b)}, slices.Concat(key...)...)
+}
```
weiihann, line 21: Can't do this, you have to put it in `db/buckets.go`

mempool/mempool.go
```diff
@@ -0,0 +1,372 @@
...
+	"github.com/NethermindEth/juno/encoder"
+	"github.com/NethermindEth/juno/utils"
+)
+
+var ErrTxnPoolFull = errors.New("transaction pool is full")
+
+type storageElem struct {
+	Txn      BroadcastedTransaction
+	NextHash *felt.Felt   // persistent db
+	Next     *storageElem // in-memory
```
weiihann, line 21: I don't think this is required? `Next` is a `storageElem`, which contains `BroadcastedTransaction`, which contains an interface `Transaction`. Simply calling `Next.Txn.Hash()` will give you the next hash. I'd recommend creating a `Hash()` method for `storageElem`.

mempool/mempool.go
```diff
@@ -0,0 +1,372 @@
...
+
+type BroadcastedTransaction struct {
+	Transaction   core.Transaction
+	DeclaredClass core.Class
+}
+
+type txnList struct {
+	head *storageElem
+	tail *storageElem
+	len  uint16
```
weiihann, line 32: I'm slightly concerned with overflowing and underflowing `txnList.len`. I would prefer `txnList.len` to be `int` instead. It doesn't affect the memory, because `len uint16` has to be padded anyways. What I'm referring to is:

type txnList struct {
    head *storageElem  // 8 bytes (64-bit pointer)
    tail *storageElem  // 8 bytes (64-bit pointer)
    len  uint16        // 2 bytes (+ 6 bytes padded)
    mu   sync.Mutex    // 8 bytes (on 64-bit systems)
}

mempool/mempool.go
```diff
@@ -0,0 +1,372 @@
...
+}
+
+// Pool stores the transactions in a linked list for its inherent FCFS behaviour
+type Pool struct {
+	log         utils.SimpleLogger
+	state       core.StateReader
+	db          db.DB // persistent mempool
+	txPushed    chan struct{}
+	txnList     *txnList // in-memory
+	maxNumTxns  uint16
```
weiihann, line 43: This can be `int` too

rianhughes: Why change from uint16?

mempool/bucket.go
weiihann: Yeah, I would say single db is fine. Splitting different dbs is usually when we want to tune the performance of individual type, but it's much more complex to handle. I'd say we are not at that stage yet.

mempool/mempool.go
weiihann: It's more of a Go-consistency kind of thing. Unless we have a very specific reason to limit the number to `2^16 - 1`, int is fine to be more consistent with Go patterns. We also avoid potential type conversions. No memory savings as well due to memory alignment of `uint16` in this struct pattern.

mempool/mempool.go
rianhughes: It's needed to connect the nodes / values (`{Txn, NextHash}`) in the db. Since we can't store the `Next *storageElem` in the db directly, we store `NextHash` instead

rianhughes: I agree that `NextHash` could be removed if we only had an in-memory mempool though

mempool/mempool.go
rianhughes: Okay, I've updated len and maxNumTxns to be `int`s

mempool/bucket.go
rianhughes: I've updated the code so that the txns are stored in the main db

rodrigo-pino (commented):
Haven't checked functionality. Will do that on a second round. I am sharing now a bit of stylistic changes which I think will improve the code. Let me know what you think

mempool/mempool.go
```diff
@@ -0,0 +1,365 @@
...
+	"github.com/NethermindEth/juno/encoder"
+	"github.com/NethermindEth/juno/utils"
+)
+
+var ErrTxnPoolFull = errors.New("transaction pool is full")
+
+type storageElem struct {
+	Txn      BroadcastedTransaction
+	NextHash *felt.Felt   // persistent db
+	Next     *storageElem // in-memory
```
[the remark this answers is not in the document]
rianhughes: Yup, exactly. I'll add a comment

mempool/mempool.go
```diff
@@ -0,0 +1,365 @@
...
+		}
+		pLen, err := p.lenDB(dbTxn)
+		if err != nil {
+			return err
+		}
+		return p.updateLen(dbTxn, pLen+1)
+	})
+}
+
+// Push queues a transaction to the pool and adds it to both the in-memory list and DB
```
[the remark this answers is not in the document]
rianhughes: Updated

mempool/mempool.go
```diff
@@ -0,0 +1,365 @@
...
+
+var ErrTxnPoolFull = errors.New("transaction pool is full")
+
+type storageElem struct {
+	Txn      BroadcastedTransaction
+	NextHash *felt.Felt   // persistent db
+	Next     *storageElem // in-memory
+}
+
+type BroadcastedTransaction struct {
```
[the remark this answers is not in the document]
rodrigo-pino: I think any ordering is good. In this case I am not sure there is one. If we were to use the top down approach then we would have:
```
struct Pool { ... }

struct txnList { ... }

struct storageElem { ... }

struct BroadcastedTransactions { ... }
```

I am ok with whatever you choose. My personal preference is Bottom Up in this case where the most simple definitions stay on top and the most complex one (being Pool) stays at the Bottom. Also, when reading Pool you don't have to scroll down since you should have already read anything you needed to understand.

mempool/mempool.go
```diff
@@ -0,0 +1,365 @@
...
+	txPushed    chan struct{}
+	txnList     *txnList // in-memory
+	maxNumTxns  int
+	dbWriteChan chan *BroadcastedTransaction
+	wg          sync.WaitGroup
+}
+
+// New initialises the Pool and starts the database writer goroutine.
+// It is the responsibility of the user to call the cancel function if the context is cancelled
+func New(mainDB db.DB, state core.StateReader, maxNumTxns int, log utils.SimpleLogger) (*Pool, func() error, error) {
```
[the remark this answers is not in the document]
rianhughes: 1. I think renaming the function `NewPool` is a bit unnecessary, since it will be imported as `pool.New()`, and it's the only New function in the pkg.

2. I agree the `New` function can be simplified. I've push the `loadFromDB` function to a new method, which also lets the user decide if they want to restore the pool from the db. However, I've kept the closer function, and dbwriter go-routine. This probably can be made optional (and therefore simplify the `New` function further), but it adds complexity, and imo it is unnecessary atm. We can always implement this change later if we think its necessary.

mempool/mempool.go
```diff
@@ -0,0 +1,365 @@
...
+			p.txnList.mu.Unlock()
+
+			currentHash = curElem.NextHash
+		}
+
+		return nil
+	})
+}
+
+func (p *Pool) handleTransaction(userTxn *BroadcastedTransaction) error {
```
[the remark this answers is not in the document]
rianhughes: Yup, `writeTransactionToDB` is better

mempool/mempool.go
```diff
@@ -0,0 +1,365 @@
...
+				p.txnList.head = newNode
+				p.txnList.tail = newNode
+			}
+			p.txnList.len++
+			p.txnList.mu.Unlock()
+
+			currentHash = curElem.NextHash
+		}
+
+		return nil
```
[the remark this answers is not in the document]
rianhughes: Yup good point, I've created a push function for txnList

mempool/mempool.go
```diff
@@ -0,0 +1,365 @@
...
+	mu   sync.Mutex
+}
+
+// Pool stores the transactions in a linked list for its inherent FCFS behaviour
+type Pool struct {
+	log         utils.SimpleLogger
+	state       core.StateReader
+	db          db.DB // to store the persistent mempool
+	txPushed    chan struct{}
+	txnList     *txnList // in-memory
```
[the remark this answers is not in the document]
rianhughes: It helps ensure different parts of the program are working on the same instance, and also makes passing pool around much lighter

mempool/mempool.go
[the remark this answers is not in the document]
rianhughes: Yeah, true, I'll go with bottom up then, because it allows me to keep the methods just below the types

mempool/mempool.go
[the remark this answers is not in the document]
rodrigo-pino: I understand, missed point 1 and agree with point 2 👍

mempool/mempool.go
```diff
@@ -0,0 +1,357 @@
...
+			tailValue = nil
+		}
+		if err := p.putdbElem(dbTxn, userTxn.Transaction.Hash(), &storageElem{
+			Txn: *userTxn,
+		}); err != nil {
+			return err
+		}
+		if tailValue != nil {
+			// Update old tail to point to the new item
+			var oldTailElem storageElem
```
rodrigo-pino, line 156: I believe this single line here is redundant?

mempool/mempool.go
```diff
@@ -0,0 +1,357 @@
...
+		}
+	case *core.L1HandlerTransaction:
+		// todo: verification of the L1 handler nonce requires checking the
+		// message nonce on the L1 Core Contract.
+	}
+	return nil
+}
+
+// Pop returns the transaction with the highest priority from the in-memory pool
+func (p *Pool) Pop() (BroadcastedTransaction, error) {
```
rodrigo-pino, line 253: This function should call to `txnList` `pop` method

mempool/mempool.go
```diff
@@ -0,0 +1,357 @@
...
+		tail.SetBytes(b)
+		return nil
+	})
+}
+
+func (p *Pool) updateTail(txn db.Transaction, tail *felt.Felt) error {
+	return txn.Set(db.MempoolTail.Key(), tail.Marshal())
+}
+
+func (p *Pool) dbElem(txn db.Transaction, itemKey *felt.Felt) (storageElem, error) {
```
rodrigo-pino, line 341: I would suggest the name `readdbPoolTxn(...)`

mempool/mempool.go
```diff
@@ -0,0 +1,357 @@
...
+	var item storageElem
+	keyBytes := itemKey.Bytes()
+	err := txn.Get(db.MempoolNode.Key(keyBytes[:]), func(b []byte) error {
+		return encoder.Unmarshal(b, &item)
+	})
+	return item, err
+}
+
+func (p *Pool) putdbElem(txn db.Transaction, itemKey *felt.Felt, item *storageElem) error {
```
rodrigo-pino, line 350: And here `writedbPoolTxn(...)`. Also, I believe the `itemKey` field here is redundant since it can obtained from `item`.

mempool/mempool.go
rianhughes: I think readDBTxn might be a bit misleading since we store more than the user Txn, but I'll add the read / write prefix for clarity

mempool/mempool.go
rianhughes: I feel like these two lines just help anyone who is unfamiliar with the code understand what it's trying to do
// Update old tail to point to the new item
// Empty list, make new item both the head and the tail

mempool/mempool.go
```diff
@@ -0,0 +1,357 @@
...
+var ErrTxnPoolFull = errors.New("transaction pool is full")
+
+type BroadcastedTransaction struct {
+	Transaction   core.Transaction
+	DeclaredClass core.Class
+}
+
+// storageElem defines a node for both the
+// in-memory and persistent linked-list
+type storageElem struct {
```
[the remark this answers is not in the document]
rianhughes: Yup, good point

mempool/mempool.go
rodrigo-pino: I'm resolving this conversation since it is outdated given the latest changes

mempool/mempool.go
```diff
@@ -0,0 +1,362 @@
...
+
+	if err != nil && errors.Is(err, db.ErrKeyNotFound) {
+		return 0, nil
+	}
+	return l, err
+}
+
+func (p *Pool) updateLen(txn db.Transaction, l int) error {
+	return txn.Set(db.MempoolLength.Key(), new(big.Int).SetInt64(int64(l)).Bytes())
+}
```
rodrigo-pino, line 318: This method is used only once, I think we can safely inline it inside `writeToDB` since it is not used anywhere else.

mempool/mempool.go
```diff
@@ -0,0 +1,362 @@
...
+
+func (p *Pool) Wait() <-chan struct{} {
+	return p.txPushed
+}
```
rodrigo-pino, line 322: This method is never used so far. Will it be useful later?

mempool/mempool.go
```diff
@@ -0,0 +1,362 @@
...
+
+func (p *Pool) headHash(txn db.Transaction, head *felt.Felt) error {
+	return txn.Get(db.MempoolHead.Key(), func(b []byte) error {
+		head.SetBytes(b)
+		return nil
+	})
+}
```
rodrigo-pino, line 329: `headHash` shouldn't be a method since it doesn't interact with Pool. I think you should change it to a function

mempool/mempool.go
```diff
@@ -0,0 +1,362 @@
...
+
+func (p *Pool) updateHead(txn db.Transaction, head *felt.Felt) error {
```
rodrigo-pino, line 331: This applies to the every other method below this line. They are define as method of `p *Pool` but they don't interact with it in any way. 

Suggestion: let's move them to a new script inside the same module "mempool/db_utils.go" or with a similar name and keep everything else the same. What do you think?

mempool/mempool_test.go
```diff
@@ -0,0 +1,226 @@
...
+		txn, err := pool.Pop()
+		require.NoError(t, err)
+		require.Equal(t, i, txn.Transaction.Hash().Uint64())
+		require.Equal(t, int(3-i), pool.Len())
+	}
+
+	// push multiple to non empty (push 4,5. now have 3,4,5)
+	for i := uint64(4); i < 6; i++ {
+		senderAddress := new(felt.Felt).SetUint64(i)
+		state.EXPECT().ContractNonce(senderAddress).Return(new(felt.Felt).SetUint64(0), nil)
```
rodrigo-pino, line 82: I believe `SetUint64(0)` is not necessary. If you want to be explicit, I think returning `&felt.Zero` is a better option

rodrigo-pino (commented):
LGTM! Adding some final comments

mempool/mempool.go
rianhughes: Yup, it will be used by the Builder service so it knows when to stop popping txns from the pool

mempool/mempool.go
rianhughes: Yup, good point

mempool/mempool_test.go
rianhughes: I've updated it to &felt.Zero, but just out of curiosity, why do you think `new(felt).SetUint64(0)` is suboptimal? Just because it might get allocated on the heap?

rodrigo-pino: Oh, in this instance it was from a legibility point of view. It is easier to read that your returning zero than creating a new felt and setting it's value to zero

rodrigo-pino (approved):
LGTM!

mempool/db_utils.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+func setDBElem(txn db.Transaction, item *dbPoolTxn) error {
+	itemBytes, err := encoder.Marshal(item)
+	if err != nil {
+		return err
+	}
+	keyBytes := item.Txn.Transaction.Hash().Bytes()
+	return txn.Set(db.MempoolNode.Key(keyBytes[:]), itemBytes)
+}
```
weiihann, line 50: I would prefer the notation of `DBElem` to be more explicit. Since we are reading and writing transactions, simply something like `readTxn` and `setTxn` would be clearer.

mempool/mempool_test.go
```diff
@@ -0,0 +1,226 @@
...
+	}
+	persistentPool, err := pebble.New(dbPath)
+	if err != nil {
+		return nil, nil, err
+	}
+	closer := func() {
+		// The db should be closed by the mempool closer function
+		os.RemoveAll(dbPath)
+	}
+	return persistentPool, closer, nil
```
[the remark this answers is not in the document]
rianhughes: In `TestRestoreMempool` the mempool is opened,  shutdown, and then re-opened. When shutting down, it will close the persistent db, and using an in-memory db will cause the "persistent" txns to get lost. Using a persistent db solves this problem.

mempool/mempool.go
```diff
@@ -0,0 +1,302 @@
...
+		if err != nil {
+			if errors.Is(err, db.ErrKeyNotFound) {
+				return nil
+			}
+			return err
+		}
+		// loop through the persistent pool and push nodes to the in-memory pool
+		currentHash := headVal
+		for currentHash != nil {
+			curDBElem, err := readTxn(txn, currentHash)
```
weiihann, line 136: please rename `curDBElem` to `curTxn` as well

weiihann (approved):

PR #2347, opened 2024-12-24
fix: errors.New use error

RiceChuan:
fix: It is recommended to use the fmt.Errorf method with parameters instead of the errors.New method without parameters.

### DIFF
--- a/core/felt/felt.go
+++ b/core/felt/felt.go
@@ -88,12 +88,12 @@ func (z *Felt) SetString(number string) (*Felt, error) {
 
 	if _, ok := vv.SetString(number, 0); !ok {
 		if _, ok := vv.SetString(number, Base16); !ok {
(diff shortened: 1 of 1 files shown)


weiihann:
Thanks for the PR. The edits merely change from `errors.New` to `fmt.Errorf` without removing the string concatenation. Please use the proper conventional formatting, for example `fmt.Errorf(...%v...)`.

Also, other places have the same issue as well:
- [here](https://github.com/NethermindEth/juno/blob/main/jsonrpc/server.go#L563C1-L564C1)

[This error](https://github.com/NethermindEth/juno/blob/main/db/remote/transaction.go#L44) is reused multiple times, please create a constant and replace them.

It'll be great if you can help to make the changes :)

rodrigo-pino:
@RiceChuan could you please update the PR. We might close it if it stalls for too long.

PR #2349, opened 2024-12-24
fix: errors.New use error 

RiceChuan:
fix: It is recommended to use the fmt.Errorf method with parameters instead of the errors.New method without parameters.


weiihann:
duplicates

PR #2357, opened 2025-01-02
Add bound options to `txn.NewIterator`

weiihann:
The old pattern is as follows:

```
for iterator.Seek(prefix); iterator.Valid(); iterator.Next() {
  if !bytes.HasPrefix(iterator.Key(), prefix) {
	break
  }
}
```

The prefix check is redundant because PebbleDB offers a built-in iterator with prefix match. Hence we simply add the bound options in the `NewIterator` method and create PebbleDB's iterator options accordingly.

### DIFF
--- a/db/pebble/batch.go
+++ b/db/pebble/batch.go
@@ -88,7 +88,7 @@ func (b *batch) Get(key []byte, cb func([]byte) error) error {
 }
 
 // NewIterator : see db.Transaction.NewIterator
-func (b *batch) NewIterator() (db.Iterator, error) {
+func (b *batch) NewIterator(lowerBound []byte, withUpperBound bool) (db.Iterator, error) {
 	var iter *pebble.Iterator
 	var err error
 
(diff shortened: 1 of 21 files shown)


AnkushinDaniil (approved):

db/pebble/batch.go
```diff
@@ -88,15 +88,20 @@ func (b *batch) Get(key []byte, cb func([]byte) error) error {
 }
 
 // NewIterator : see db.Transaction.NewIterator
-func (b *batch) NewIterator() (db.Iterator, error) {
+func (b *batch) NewIterator(lowerBound []byte, withUpperBound bool) (db.Iterator, error) {
```
IronGauntlets, line 91: I am confused under which conditions you wouldn't want an upperbound?

IronGauntlets (approved):
lgtm

IronGauntlets (changes requested):
Actually i forgot to mention, since the behaviour of the iterator has changed (as it now accepts an upperbound) it should be accompanied with relevant tests.

As far as I can see there are no tests which check that the interator return false for `iter.Valid()` if the key goes past the upper bound.

Also, a test to check that the `First()` is correctly impelemented would also be required.

db/pebble/batch.go
IronGauntlets: Please ignore

IronGauntlets:
Also, another test to check would be that if a prefix is not passed than the first value returned by the iterator is at the first value in the DB (please add this if it is not already implemented)

weiihann:
> Actually i forgot to mention, since the behaviour of the iterator has changed (as it now accepts an upperbound) it should be accompanied with relevant tests.
> 
> As far as I can see there are no tests which check that the interator return false for `iter.Valid()` if the key goes past the upper bound.
> 
> Also, a test to check that the `First()` is correctly impelemented would also be required.

Added, please check the latest commit

IronGauntlets (approved):

PR #2361, opened 2025-01-07
Add block.hash 0.13.4 calculation

kirugan:


rpc/transaction.go
```diff
@@ -184,7 +184,7 @@ func (r *Resource) UnmarshalJSON(data []byte) error {
 	case `"l2_gas"`:
 		*r = ResourceL2Gas
 	default:
-		return fmt.Errorf("unknown Resource: %q", string(data))
+		return fmt.Errorf("unknown Resource3: %q", string(data))
```
rodrigo-pino, line 187: Rogue 3?

starknet/transaction.go
```diff
@@ -120,10 +121,12 @@ func (r *Resource) UnmarshalText(text []byte) error {
 	switch string(text) {
 	case "L1_GAS":
 		*r = ResourceL1Gas
+	case "L1_DATA_GAS":
+		*r = ResourceL1DataGas
 	case "L2_GAS":
 		*r = ResourceL2Gas
 	default:
-		return fmt.Errorf("unknown resource: %q", string(text))
+		return fmt.Errorf("unknown resource3: %q", string(text))
```
rodrigo-pino, line 129: And here too

core/transaction.go
```diff
@@ -653,6 +668,21 @@ func transactionCommitmentPedersen(transactions []Transaction, protocolVersion s
 	return calculateCommitment(transactions, trie.RunOnTempTriePedersen, hashFunc)
 }
 
+// transactionCommitmentPoseidon0134 handles empty signatures compared to transactionCommitmentPoseidon:
+// empty signatures interpreted as [] instead of [0]
```
rodrigo-pino, line 672:
```suggestion
// empty signatures are interpreted as [] instead of [0]
```

core/transaction.go
```diff
@@ -653,6 +668,21 @@ func transactionCommitmentPedersen(transactions []Transaction, protocolVersion s
...
+
+		if txSignature := transaction.Signature(); len(txSignature) > 0 {
+			digest.Update(txSignature...)
+		}
+
+		return digest.Finish()
+	})
+}
+
 func transactionCommitmentPoseidon(transactions []Transaction) (*felt.Felt, error) {
```
rodrigo-pino, line 686: I would suggest to update the function name to include 0132 at the end for clarity

rodrigo-pino (commented):
It looks good to me. Cross checked with the pre-release notes and it matches. I believe the only thing missing is a couple of tests

starknet/transaction.go
kirugan: sorry, it was for debugging purposes 😄

kirugan: fixed in both places

core/transaction.go
kirugan: I can add comment saying that Poseidon was used in < 0.13.2 version, wdyt?

kirugan: actually I thought about it one more time and decide to rename it :)

AnkushinDaniil (approved):

rodrigo-pino (approved):

IronGauntlets (approved):

PR #2371, opened 2025-01-13
implement logic to build genesis state for the sequencer

rianhughes:
This PR implements the `genesis` pkg that builds a `state diff` from declaring classes, deploying contracts and invoking functions. The sequencer can then commit this `state diff` to state, so that upon startup, a set of account contracts have already been deployed and funded (it is not possible to achieve this from an empty state by submitting transactions, since Starknet requires accounts to be funded to perform any actions on the state - originally this wasn't a requirement in Starknet). This is particularly useful as it allows users to start the sequencer with any number of accounts pre-deployed, and pre-funded, etc, and opens up the possibility of using Juno as a devnet

In essence:
1. we load classes, then declare them, 
2. execute the constructor functions using the vm, which writes the changes to the pending state
3. execute any functions of interest (eg token transfers) using the vm (where similarly the vm writes the changes to the pending state)
4. extract the state-diff and new classes from the pending state.


genesis/genesis.go
```diff
@@ -0,0 +1,250 @@
...
+	g.FunctionCalls = aux.FunctionCalls
+	g.BootstrapAccounts = aux.BootstrapAccounts
+	g.Contracts = make(map[felt.Felt]GenesisContractData)
+	for k, v := range aux.Contracts {
+		key, err := new(felt.Felt).SetString(k)
+		if err != nil {
+			return err
+		}
+		g.Contracts[*key] = v
+	}
```
weiihann, line 71: I don't think we need to do dereferencing? Just do:

```
	for k, v := range aux.Contracts {
                var key felt.Felt
		key, err = key.SetString(k)
		if err != nil {
			return err
		}
		g.Contracts[key] = v
	}
```

genesis/genesis.go
```diff
@@ -0,0 +1,251 @@
...
+		}
+
+		if cairo1Class, isCairo1 := class.(*core.Cairo1Class); isCairo1 {
+			if err = genesisState.SetCompiledClassHash(&classHash, cairo1Class.Compiled.Hash()); err != nil {
+				return nil, nil, fmt.Errorf("set compiled class hash: %v", err)
+			}
+		}
+	}
+
+	constructorSelector, err := new(felt.Felt).SetString("0x28ffe4ff0f226a9107253e17a904099aa4f63a02a5621de0576e5aa71bc5194")
```
weiihann, line 125: Please make `0x28ffe4ff0f226a9107253e17a904099aa4f63a02a5621de0576e5aa71bc5194` a constant with a comment describing what this value is

core/state_update.go
```diff
@@ -199,3 +199,41 @@ func noncesDigest(nonces map[felt.Felt]*felt.Felt, digest *crypto.PoseidonDigest
...
+
+	mergeStorageDiffs(oldStateDiff.StorageDiffs, newStateDiff.StorageDiffs)
+	mergeMaps(oldStateDiff.Nonces, newStateDiff.Nonces)
+	mergeMaps(oldStateDiff.DeployedContracts, newStateDiff.DeployedContracts)
+	mergeMaps(oldStateDiff.DeclaredV1Classes, newStateDiff.DeclaredV1Classes)
+	mergeMaps(oldStateDiff.ReplacedClasses, newStateDiff.ReplacedClasses)
+	oldStateDiff.DeclaredV0Classes = append(oldStateDiff.DeclaredV0Classes, newStateDiff.DeclaredV0Classes...)
+
+	return oldStateDiff
+}
```
weiihann, line 239: I'd suggest creating a `Merge` method for `StateDiff` struct. Also, I believe `mergeMaps` can be replaced with `maps.Copy()`

genesis/genesis.go
rianhughes: Ah right, in your approach key is allocated to the stack and not the heap, whereas in the original case key is allocated to the heap?

rianhughes: Is the motivation for this change so that key is allocated to the stack rather than the heap?

rianhughes: The suggested change doesn't work, since `SetString` only operates on a pointer 
`func (z *Felt) SetString(number string) (*Felt, error) {`
and initialising a nil pointer will result in a panic when calling `SetString` (ie using `var key *felt.Felt`)

sync/pending.go
```diff
@@ -65,3 +69,77 @@ func (p *PendingState) Class(classHash *felt.Felt) (*core.DeclaredClass, error)
...
+
+	p.newClasses[*classHash] = class
+	if class.Version() == 0 {
+		p.stateDiff.DeclaredV0Classes = append(p.stateDiff.DeclaredV0Classes, classHash.Clone())
+	} // assumption: SetCompiledClassHash will be called for Cairo1 contracts
+	return nil
+}
+
+func (p *PendingStateWriter) SetCompiledClassHash(classHash, compiledClassHash *felt.Felt) error {
+	// assumption: SetContractClass was called for classHash and succeeded
```
AnkushinDaniil, line 132: Can we make these assumptions a part of the function’s doc comment?

genesis/genesis.go
```diff
@@ -0,0 +1,257 @@
...
+					ConstructorCallData: *txn.ConstructorCallData,
+					Version:             (*core.TransactionVersion)(txn.Version),
+				},
+				MaxFee:               txn.MaxFee,
+				TransactionSignature: *txn.Signature,
+				Nonce:                txn.Nonce,
+			}
+
+		default:
+			panic("transaction type not supported")
```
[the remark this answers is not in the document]
rianhughes: Nice

sync/pending.go
```diff
@@ -65,3 +69,77 @@ func (p *PendingState) Class(classHash *felt.Felt) (*core.DeclaredClass, error)
...
+	p.stateDiff.DeclaredV1Classes[*classHash] = compiledClassHash.Clone()
+	return nil
+}
+
+// StateDiffAndClasses returns the pending state's internal data. The returned objects will continue to be
+// read and modified by the pending state.
```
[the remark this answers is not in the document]
rianhughes: I think users should only be able to modify the state and classes via the interface. For example, to set a CairoV0 contract the user should call SetContractClass, and not just add it to newClasses etc

genesis/genesis.go
```diff
@@ -0,0 +1,257 @@
...
+
+type FunctionCall struct {
+	ContractAddress    felt.Felt   `json:"contract_address"`
+	EntryPointSelector felt.Felt   `json:"entry_point_selector"`
+	Calldata           []felt.Felt `json:"calldata"`
+}
+
+func (g *GenesisConfig) Validate() error {
+	validate := validator.Validator()
+	return validate.Struct(g)
```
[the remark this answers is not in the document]
rianhughes: Yup!

genesis/genesis.go
```diff
@@ -0,0 +1,257 @@
...
+
+func (g *GenesisConfig) UnmarshalJSON(data []byte) error {
+	type auxConfig struct {
+		Classes           []string                       `json:"classes"`            // []path-to-class.json
+		Contracts         map[string]GenesisContractData `json:"contracts"`          // address -> {classHash, constructorArgs}
+		FunctionCalls     []FunctionCall                 `json:"function_calls"`     // list of functionCalls to Call()
+		BootstrapAccounts []Account                      `json:"bootstrap_accounts"` // accounts to prefund with strk token
+		Txns              []*rpc.Transaction             `json:"transactions"`       // declare NOT supported
+	}
+	var aux auxConfig
```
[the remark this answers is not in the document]
rianhughes: Nice

core/state_update.go
```diff
@@ -39,6 +39,24 @@ func (d *StateDiff) Length() uint64 {
...
+func (d *StateDiff) Merge(incoming *StateDiff) {
+	mergeStorageDiffs := func(oldMap, newMap map[felt.Felt]map[felt.Felt]*felt.Felt) {
+		for addr, newAddrStorage := range newMap {
+			if oldAddrStorage, exists := oldMap[addr]; exists {
+				maps.Copy(oldAddrStorage, newAddrStorage)
+			} else {
+				oldMap[addr] = newAddrStorage
+			}
+		}
+	}
```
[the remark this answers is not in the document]
rianhughes: We can, but imo the current form is more readable

genesis/genesis.go
```diff
@@ -0,0 +1,257 @@
...
+		if err = genesisState.SetContractClass(&classHash, class); err != nil {
+			return nil, nil, fmt.Errorf("declare v0 class: %v", err)
+		}
+
+		if cairo1Class, isCairo1 := class.(*core.Cairo1Class); isCairo1 {
+			if err = genesisState.SetCompiledClassHash(&classHash, cairo1Class.Compiled.Hash()); err != nil {
+				return nil, nil, fmt.Errorf("set compiled class hash: %v", err)
+			}
+		}
+	}
```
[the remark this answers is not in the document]
rianhughes: >Could we use descriptive variable names instead of this comment?
Not sure I understand sorry. Could you give me an example?

> I don't quite understand why only DeclaredV0Classes in newClasses.
Good question! Basically we inherit this structure from blockifier / starknet. To allow blockifier to make changes to the pending state we need to implement a certain interface (see `impl State for JunoStateReader`). We use that interface here to declare classes. Note this is how it seems to be done in blockifier / starknet, it first calls `set_contract_class` then if its a CairoV1 class, `set_compiled_class_hash`.

core/state_update.go
```diff
@@ -199,3 +217,14 @@ func noncesDigest(nonces map[felt.Felt]*felt.Felt, digest *crypto.PoseidonDigest
...
+
+func EmptyStateDiff() *StateDiff {
+	return &StateDiff{
+		StorageDiffs:      make(map[felt.Felt]map[felt.Felt]*felt.Felt),
+		Nonces:            make(map[felt.Felt]*felt.Felt),
+		DeployedContracts: make(map[felt.Felt]*felt.Felt),
+		DeclaredV0Classes: make([]*felt.Felt, 0),
+		DeclaredV1Classes: make(map[felt.Felt]*felt.Felt),
+		ReplacedClasses:   make(map[felt.Felt]*felt.Felt),
+	}
```
cicr99, line 229: This is what I was mentioning in the comment above about StorageDiff. The same way we could consider using type alias like Address or ClassHash for example, and so DeployedContracts map would be: `map[Address]ClassHash`.
 This allows the code to be more readable, you could also make it safer as you restrict the specific type that must be passed to a function (not just any felt), and you could even create methods on top of these structures. Wdyt?

genesis/genesis.go
```diff
@@ -0,0 +1,256 @@
...
+	Txns              []*rpc.Transaction                `json:"transactions"`       // Declare NOT supported
+}
+
+type Account struct {
+	Address felt.Felt `json:"address"`
+	PubKey  felt.Felt `json:"publicKey"`
+	PrivKey felt.Felt `json:"privateKey"`
+}
+
+func Read(path string) (*GenesisConfig, error) {
```
cicr99, line 39: Is this function only used for testing purposes?

genesis/genesis.go
```diff
@@ -0,0 +1,256 @@
...
+					ContractAddressSalt: txn.ContractAddressSalt,
+					ContractAddress:     txn.SenderAddress,
+					ClassHash:           txn.ClassHash,
+					ConstructorCallData: *txn.ConstructorCallData,
+					Version:             (*core.TransactionVersion)(txn.Version),
+				},
+				MaxFee:               txn.MaxFee,
+				TransactionSignature: *txn.Signature,
+				Nonce:                txn.Nonce,
+			}
```
cicr99, line 200: Aren't there adapters for rpc to core?

genesis/genesis.go
```diff
@@ -0,0 +1,256 @@
...
+
+		default:
+			return nil, nil, fmt.Errorf("unsupported transaction type: %v", txn.Type)
+		}
+
+		_, _, trace, _, _, err := v.Execute([]core.Transaction{coreTxn}, nil, []*felt.Felt{new(felt.Felt).SetUint64(1)},
+			&blockInfo, genesisState, network, true, false, true)
```
cicr99, line 207: Why are you calling `Execute` here with only one transaction? Could it be done at the end and pass all the transactions at once? I believe it should return the trace for each transaction, right?
Also now I'm not so sure, why are we using `Call` above and `Execute` here?

sync/pending.go
```diff
@@ -65,3 +71,79 @@ func (p *PendingState) Class(classHash *felt.Felt) (*core.DeclaredClass, error)
...
+}
+
+// SetContractClass writes a new CairoV0 class to the PendingState
+// Assumption: SetCompiledClassHash should be called for CairoV1 contracts
+func (p *PendingStateWriter) SetContractClass(classHash *felt.Felt, class core.Class) error {
+	if _, err := p.Class(classHash); err == nil {
+		return errors.New("class already declared")
+	} else if !errors.Is(err, db.ErrKeyNotFound) {
+		return fmt.Errorf("get class: %v", err)
+	}
```
cicr99, line 125: Perhaps consider leaving a comment here as it's a bit confusing when you read it for the first time 😅

genesis/genesis.go
```diff
@@ -0,0 +1,256 @@
...
+	"github.com/NethermindEth/juno/starknet"
+	"github.com/NethermindEth/juno/starknet/compiler"
+	"github.com/NethermindEth/juno/sync"
+	"github.com/NethermindEth/juno/utils"
+	"github.com/NethermindEth/juno/validator"
+	"github.com/NethermindEth/juno/vm"
+)
+
+// The constructor entrypoint for cairo contracts
+const constructorSelector = "0x28ffe4ff0f226a9107253e17a904099aa4f63a02a5621de0576e5aa71bc5194"
```
cicr99, line 23: Is it a good idea to save this value in a constant somewhere in the project? I see it's being used some other places, so perhaps is good to have these sort of values saved so they are easier to use

vm/rust/src/lib.rs
```diff
@@ -325,7 +350,7 @@ pub extern "C" fn cairoVMExecute(
                 }
 
                 // we are estimating fee, override actual fee calculation
-                if t.transaction_receipt.fee.0 == 0 {
+                if t.transaction_receipt.fee.0 == 0 && !is_l1_handler_txn {
```
cicr99, line 353: Is there an specific behavior when `t.transaction_receipt.fee.0 == 0` but the other condition is not met? Just making sure we didn't miss any case here

genesis/genesis.go
rianhughes: It only seems to be used in two tests (`core/crypto/pedersen_hash_test.go`, and `clients/feeder/feeder_test.go`), so I don't think making this a global variable across packages is necessary at this point

genesis/genesis.go
rianhughes: We can execute all the transactions together, I'll update the code

genesis/genesis.go
rianhughes: Nope, there are no rpc to core adapters

sync/pending.go
rianhughes: Updated

genesis/genesis.go
rianhughes: Nope, it'll also be used by the `node` and `builder` pkgs later on

core/state_update.go
```diff
@@ -39,6 +39,24 @@ func (d *StateDiff) Length() uint64 {
 	return uint64(length)
 }
 
+func (d *StateDiff) Merge(incoming *StateDiff) {
+	mergeStorageDiffs := func(oldMap, newMap map[felt.Felt]map[felt.Felt]*felt.Felt) {
```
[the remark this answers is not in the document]
rianhughes: I agree that having a custom type for this (and similar types) is more readable. However, given it's used throughout the code, and not just in genesis, I think it would be best to update this in a separate PR. It would also make a good first issue

vm/rust/src/lib.rs
rianhughes: Nope, this just ensures that we don't override transaction fees for `l1_handler_txns` (this has more to do with the "shadow" feature of the sequencer, where we re-execute Sepolia blocks and ensure we get the correct new block and new state update etc)

genesis/genesis.go
cicr99: Thanks for the explanation 🙌

core/state_update.go
rodrigo-pino: This makes sense to me too. Can this text be part of the issue description 🙏 with extra details if possible

core/state_update.go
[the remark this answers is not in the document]
cicr99: Issue #2396 created for this

core/state_update.go
cicr99: This is also added in issue #2396

cicr99:
Even though the PR is approved, I'm adding the `blocked` label up until all the changes regarding the new blockifier release are finished. This takes priority and will probably cause conflicts with the changes done to the vm pkg in this PR.

rodrigo-pino (changes requested):
Just putting a blocking comment here to make sure it doesn't get merged by accident during this release process. 

I would also like to give a final review before moving ahead with the PR.

adapters/vm2core/vm2core.go
```diff
@@ -39,3 +40,51 @@ func AdaptOrderedEvents(events []vm.OrderedEvent) []*core.Event {
...
+		newStorageDiffs[sd.Address] = entries
+	}
+
+	newNonces := make(map[felt.Felt]*felt.Felt)
+	for _, nonce := range stateDiff.Nonces {
+		newNonce := nonce.Nonce
+		newNonces[nonce.ContractAddress] = &newNonce
+	}
+
+	newDeployedContracts := make(map[felt.Felt]*felt.Felt)
```
weiihann, line 64:
```suggestion
	newDeployedContracts := make(map[felt.Felt]*felt.Felt, len(stateDiff.DeployedContracts))
```

adapters/vm2core/vm2core.go
```diff
@@ -39,3 +40,51 @@ func AdaptOrderedEvents(events []vm.OrderedEvent) []*core.Event {
...
+	for _, dc := range stateDiff.DeployedContracts {
+		ch := dc.ClassHash
+		newDeployedContracts[dc.Address] = &ch
+	}
+
+	newDeclaredV1Classes := make(map[felt.Felt]*felt.Felt)
```
weiihann, line 70:
```suggestion
	newDeclaredV1Classes := make(map[felt.Felt]*felt.Felt, len(stateDiff.DeclaredClasses))
```

adapters/vm2core/vm2core.go
```diff
@@ -39,3 +40,51 @@ func AdaptOrderedEvents(events []vm.OrderedEvent) []*core.Event {
...
+	for _, dc := range stateDiff.DeclaredClasses {
+		cch := dc.CompiledClassHash
+		newDeclaredV1Classes[dc.ClassHash] = &cch
+	}
+
+	newReplacedClasses := make(map[felt.Felt]*felt.Felt)
```
weiihann, line 76:
```suggestion
	newReplacedClasses := make(map[felt.Felt]*felt.Felt, len(stateDiff.ReplacedClasses))
```

genesis/genesis.go
```diff
@@ -0,0 +1,258 @@
...
+			ContractAddress: &contractAddress,
+			ClassHash:       classHash,
+			Selector:        &entryPointSelector,
+			Calldata:        fnCall.Calldata,
+		}
+		blockInfo := vm.BlockInfo{
+			Header: &core.Header{
+				Number:    0,
+				Timestamp: 0,
+			},
```
weiihann, line 166: I'd suggest defining a global variable called `genesisHeader` or `emptyHeader = core.Header{Number: 0, Timestamp: 0}` and replace it to improve code clarity. `core.Header{Number: 0, Timestamp: 0}` is used in other places as well so please replace it accordingly too.

genesis/genesis.go
```diff
@@ -0,0 +1,258 @@
...
+	"encoding/json"
+	"fmt"
+	"os"
+
+	"github.com/NethermindEth/juno/adapters/sn2core"
+	"github.com/NethermindEth/juno/adapters/vm2core"
+	"github.com/NethermindEth/juno/core"
+	"github.com/NethermindEth/juno/core/felt"
+	"github.com/NethermindEth/juno/db"
+	rpc "github.com/NethermindEth/juno/rpc/v6"
```
weiihann, line 13: Is it intentional to use `rpc/v6` in this PR and then move the common parts to a common package later on? Or can we just do it here?

vm/state.go
```diff
@@ -91,6 +92,69 @@ func JunoStateGetCompiledClass(readerHandle C.uintptr_t, classHash unsafe.Pointe
 	return unsafe.Pointer(cstring(compiledClass))
 }
 
+//export JunoStateSetStorage
+func JunoStateSetStorage(readerHandle C.uintptr_t, addr, key, value unsafe.Pointer) unsafe.Pointer {
+	context := unwrapContext(readerHandle)
+	addrFelt := makeFeltFromPtr(addr)
+	keyFelt := makeFeltFromPtr(key)
+	valueFelt := makeFeltFromPtr(value)
+	state := context.state.(StateReadWriter)
```
weiihann, line 101: I don't think this is a good design. The issues are:
- Runtime Type Checking: The code relies on runtime type assertions (context.state.(StateReadWriter)) which can panic if the assertion fails.
- Implicit Behavior: The code's behavior changes based on whether the state implements write methods, but this isn't immediately obvious from function signatures.

I'd suggest changing `state core.StateReader` to an interface defined directly in the `vm` package. It's usually a good practice to define interfaces in the consumer package (in this case, `vm`) rather than the implementor package (in this case, `core`).

so something like:

```
package vm

type State interface {
	StateReader
	StateWriter
}

type StateReader interface {
	ContractClassHash(addr *felt.Felt) (*felt.Felt, error)
	ContractNonce(addr *felt.Felt) (*felt.Felt, error)
	ContractStorage(addr, key *felt.Felt) (*felt.Felt, error)
	Class(classHash *felt.Felt) (*DeclaredClass, error)
}

type StateWriter interface {
	SetStorage(contractAddress, storageKey, value *felt.Felt) error
	IncrementNonce(contractAddress *felt.Felt) error
	SetClassHash(contractAddress, classHash *felt.Felt) error
	SetContractClass(classHash *felt.Felt, contractClass core.Class) error
	SetCompiledClassHash(classHash *felt.Felt, compiledClassHash *felt.Felt) error
}

type callContext struct {
	// state that the call is running on
	state State
        ...other fields
}
```

genesis/genesis.go
rianhughes: Yes it was intentional to use rpcv6. I think it makes sense to refactor the rpc pkg in a separate PR (which would only involve moving rpc.TransactionType to the rpccore pkg, since rpc.Transaction is different across the rpc versions).

vm/rust/src/jsonrpc.rs
```diff
@@ -58,6 +65,79 @@ struct StateDiff {
     replaced_classes: Vec<ReplacedClass>,
 }
 
+impl StateDiff {
+    pub fn from_state_maps(state_maps: &StateMaps) -> Self {
+        Self {
+            storage_diffs: state_maps
+                .storage
+                .iter()
```
rodrigo-pino, line 73: This can be `into_iter`, meaning that you are "consuming" the `storage` field in the function. This will also be possible after you take ownership of `state_map` as well, otherwise I think it is not possible

vm/rust/src/jsonrpc.rs
```diff
@@ -58,6 +65,79 @@ struct StateDiff {
...
+                .fold(
+                    IndexMap::<StarkFelt, Vec<Entry>>::new(),
+                    |mut acc, ((address, key), value)| {
+                        let starkfelt_address = Felt::from(address.clone()).into(); // Convert ContractAddress -> StarkFelt
+                        let entry = Entry {
+                            key: Felt::from(key.clone()).into(), // Convert StorageKey -> StarkFelt
+                            value: Felt::from(value.clone()).into(), // Convert Felt -> StarkFelt
```
rodrigo-pino, line 80: You shouldn't need all this `clone()` calls, especially if you use `into_iter` like suggested above. If you need `clone()` calls too much, as a general rule in Rust, you are doing something wrong. I am aware the team forte is Go and not Rust that's why I am adding a bit more context.

vm/rust/src/jsonrpc.rs
```diff
@@ -58,6 +65,79 @@ struct StateDiff {
...
+                    },
+                )
+                .into_iter()
+                .map(|(address, storage_entries)| StorageDiff {
+                    address,
+                    storage_entries,
+                })
+                .collect(),
+
+            nonces: state_maps
```
rodrigo-pino, line 97: The same comments as before (in the context of this function) apply to all the following fields declaration

vm/rust/src/jsonrpc.rs
```diff
@@ -58,6 +65,79 @@ struct StateDiff {
     replaced_classes: Vec<ReplacedClass>,
 }
 
+impl StateDiff {
+    pub fn from_state_maps(state_maps: &StateMaps) -> Self {
+        Self {
+            storage_diffs: state_maps
```
rodrigo-pino, line 71: nitpick: that you can initialize all the types in "one line" of code doesn't mean it is always a good idea. It's more readable if we initialize each field independently (using the same variable name as the field) and then initialize the struct. Just a suggestion

vm/rust/src/lib.rs
```diff
@@ -169,9 +172,27 @@ pub extern "C" fn cairoVMCall(
         ..Default::default()
     };
 
+    // Allow users to call CONSTRUCTOR entry point type
+    // which has fixed entry_point_felt "0x28ffe4ff0f226a9107253e17a904099aa4f63a02a5621de0576e5aa71bc5194"
+    let constuctor_entry_point_felt =
+        StarkFelt::from_hex("0x28ffe4ff0f226a9107253e17a904099aa4f63a02a5621de0576e5aa71bc5194");
```
rodrigo-pino, line 178: Can a contract have several constructors? If that's the case then I think it cannot be fixed.

Secondly, initialize the value type directly, don't set it during execution. Another improvement can be to set it as `static` in a global function

vm/rust/src/lib.rs
```diff
@@ -169,9 +172,27 @@ pub extern "C" fn cairoVMCall(
...
+        Ok(felt) => {
+            if felt.eq(&entry_point_selector_felt) {
+                entry_point.entry_point_type = EntryPointType::Constructor
+            }
+        }
+        Err(err) => {
+            // This should not happen
+            report_error(reader_handle, err.to_string().as_str(), -1, 1);
+            return;
+        }
```
rodrigo-pino, line 189: And remove the need for this unnecesary branch completly

vm/rust/src/lib.rs
```diff
@@ -220,8 +241,22 @@ pub extern "C" fn cairoVMCall(
...
             }
+            // We only need to return the state_diff when creating the genesis state_diff.
+            // Calling this for all other RPC requests is a waste of resources.
+            if return_state_diff == 1 {
+                match state.to_state_diff() {
+                    Ok(state_diff) => {
+                        let json_state_diff =
+                            jsonrpc::StateDiff::from_state_maps(&state_diff.state_maps);
+                        append_state_diff(reader_handle, &json_state_diff, &mut writer_buffer);
+                    }
```
rodrigo-pino, line 252: I don't like the idea of having flags to signal stuff that should happen or not. I believe this functions are due for refactoring but we will address that later on. Just signaling here

rodrigo-pino (commented):
It's a very big PR so I left a lot of comments. Nonetheless, my general thoughts are that it is very good quality and the quantity of the comments is just due it's size.

genesis/genesis.go
```diff
@@ -0,0 +1,265 @@
...
+		Txns              []*rpc.Transaction             `json:"transactions"`       // declare NOT supported
+	}
+	if err := json.Unmarshal(data, &aux); err != nil {
+		return err
+	}
+
+	g.Classes = aux.Classes
+	g.FunctionCalls = aux.FunctionCalls
+	g.BootstrapAccounts = aux.BootstrapAccounts
+	g.Contracts = make(map[felt.Felt]GenesisContractData)
```
weiihann, line 71: Do add the length when declaring the map since we already know beforehand

genesis/genesis.go
```diff
@@ -0,0 +1,265 @@
...
+
+	classhashToSierraVersion := map[felt.Felt]string{}
+	contractAddressToSierraVersion := map[felt.Felt]string{}
+	for classHash, class := range newClasses {
+		// Sets pending.newClasses, DeclaredV0Classes, (not DeclaredV1Classes)
+		if err = genesisState.SetContractClass(&classHash, class); err != nil {
+			return nil, nil, fmt.Errorf("declare v0 class: %v", err)
+		}
+
+		if cairo1Class, isCairo1 := class.(*core.Cairo1Class); isCairo1 {
```
weiihann, line 123: `Class` interface exposes the `Version()` method, I think it's better to use to check if it's version 1

genesis/genesis.go
```diff
@@ -0,0 +1,265 @@
...
+	return validator.Validator().Struct(g)
+}
+
+// GenesisState builds the genesis state given the genesis-config data.
+func GenesisStateDiff( //nolint:funlen,gocyclo
+	config *GenesisConfig,
+	v vm.VM,
+	network *utils.Network,
+	maxSteps uint64,
+) (*core.StateDiff, map[felt.Felt]core.Class, error) {
```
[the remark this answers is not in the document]
weiihann: I don't think so, all the fields in `StateDiff` are already reference types, it's meant to be mutable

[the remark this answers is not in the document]
rodrigo-pino: I don't understand how the fields of `StateDiff` being either references or value types affect anything. 

`StateDiff` current definition is a set of references, I don't understand why this set of references should be located in the heap. Maybe it should, but then perform the allocation later when it is really needed.

genesis/genesis.go
rodrigo-pino: I am planning to delete that method in the future, let's leave the type assertion

genesis/genesis.go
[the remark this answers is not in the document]
rianhughes: What is the motivation for returning by value? 

As I understand it, returning by value involves making a hard copy of the value, and (sometimes) allocating it to the stack, which has certain performance benefits (no GC overhead, faster allocation, etc). However, if the type is large (like core.StateDiff), or has many reference types (like maps), then 1. copying will be expensive, 2. the type is potentially too large to go on the stack, 3. the maps will be allocated to the heap anyway. So returning by value here makes more sense to me (although I could be wrong)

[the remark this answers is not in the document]
rianhughes: (Also, this function will only be called once, for the sequencer, so I'm not sure if we would see any noticeable difference. But it's an interesting problem to think about anyway)

adapters/vm2core/vm2core.go
```diff
@@ -39,3 +40,52 @@ func AdaptOrderedEvents(events []vm.OrderedEvent) []*core.Event {
 	})
 	return utils.Map(events, AdaptOrderedEvent)
 }
+
+func AdaptStateDiff(stateDiff *vm.StateDiff) *core.StateDiff {
```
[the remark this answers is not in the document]
rianhughes: I updated the function so that the caller is expected to pass in the core.StateDiff. If the caller assigns the core.StateDiff via "var toStateDiff core.StateDiff", then the current implementation should be quite efficient, since we only allcoate the maps once, to the pre-determined size.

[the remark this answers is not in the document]
rianhughes: Also, I was using *felt.Felt just out of (a bad) habit.

genesis/genesis.go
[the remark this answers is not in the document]
rodrigo-pino: I think in this specific case by returning by value you are actually saving a heap allocation. Nonetheless, every constructor should return a value. Even if it is not like that, the question is why would you preemptively allocate in the heap? If it is meant to be moved to the heap then let it be done by the user of this function, let it be their choice.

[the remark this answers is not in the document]
rodrigo-pino: you have to think of the heap as a very precious resource. Think that the GC consumes 30% typical of a Go program, the idea is to keep that percent as low as possible. Once you allocate something from the heap, it is not the only the cost of allocating there (very expensive), there is also the cost of maintaining it there.

Stack copies, are very very cheap. If you have to chose between heap allocation and two stack copies, the latter is always better.

genesis/genesis.go
```diff
@@ -0,0 +1,265 @@
...
+	Number:    0,
+	Timestamp: 0,
+}
+
+type GenesisConfig struct {
+	Classes           []string                          `json:"classes"`            // []path-to-class.json
+	Contracts         map[felt.Felt]GenesisContractData `json:"contracts"`          // address -> {classHash, constructorArgs}
+	FunctionCalls     []FunctionCall                    `json:"function_calls"`     // list of functionCalls to Call()
+	BootstrapAccounts []Account                         `json:"bootstrap_accounts"` // accounts to prefund with strk token
+	Txns              []*rpc.Transaction                `json:"transactions"`       // Declare NOT supported
```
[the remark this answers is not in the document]
rianhughes: I'll update this

genesis/genesis.go
```diff
@@ -0,0 +1,265 @@
...
+	if err = config.UnmarshalJSON(file); err != nil {
+		return nil, err
+	}
+	return &config, err
+}
+
+func (g *GenesisConfig) UnmarshalJSON(data []byte) error {
+	var aux struct {
+		Classes           []string                       `json:"classes"`            // []path-to-class.json
+		Contracts         map[string]GenesisContractData `json:"contracts"`          // address -> {classHash, constructorArgs}
```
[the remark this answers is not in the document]
rianhughes: This has been updated to use the alias (so no longer use a string)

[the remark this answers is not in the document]
rianhughes: Golang doesn't seem to allow unmarshalling into a map with a custom type as a key

genesis/genesis.go
```diff
@@ -0,0 +1,265 @@
...
+		FunctionCalls     []FunctionCall                 `json:"function_calls"`     // list of functionCalls to Call()
+		BootstrapAccounts []Account                      `json:"bootstrap_accounts"` // accounts to prefund with strk token
+		Txns              []*rpc.Transaction             `json:"transactions"`       // declare NOT supported
+	}
```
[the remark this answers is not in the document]
rianhughes: Golang doesn't seem to allow unmarshalling into a map with a custom type as a key

genesis/genesis.go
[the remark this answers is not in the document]
rianhughes: > Even if it is not like that, the question is why would you preemptively allocate in the heap? 

The main motivation for returning a pointer explicitly here is because it's cheaper to copy the pointer, than the struct itself (namely all the internal map headers, etc). And this type will be explicitly allcoated to the heap anyway, so I'm not sure how returning by value is better in this case?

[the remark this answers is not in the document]
rianhughes: Actually, maybe one solution would be to force the caller to create and pass in the core.StateDiff (similar to what we now do for AdaptStateDiff)? This should help prevent heap allocation

genesis/genesis.go
```diff
@@ -0,0 +1,265 @@
...
+	EntryPointSelector felt.Felt   `json:"entry_point_selector"`
+	Calldata           []felt.Felt `json:"calldata"`
+}
+
+func (g *GenesisConfig) Validate() error {
+	return validator.Validator().Struct(g)
+}
+
+// GenesisState builds the genesis state given the genesis-config data.
+func GenesisStateDiff( //nolint:funlen,gocyclo
```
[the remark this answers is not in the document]
rianhughes: Done

genesis/genesis.go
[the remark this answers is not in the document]
rianhughes: After internal discussions, it became clear that returning a pointer in this case is the wrong thing to do. Since core.StateDiff is quite light, it makes sense to return the value, which will points to the internal fields. Returning by reference adds an additional layer of indirection which just isn't needed.

adapters/vm2core/vm2core.go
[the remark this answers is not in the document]
rianhughes: I actually reverted this change. I just return the result by value, seems simpler and cleaner.

vm/rust/src/lib.rs
rianhughes: A contract can only have a single constructor, which is also standardised to have the same entrypoint selector.
Eg see here https://starknet-by-example.voyager.online/getting-started/basics/constructor/

Yup, I'll declare this const at the top and remove the dead code

vm/rust/src/lib.rs
rianhughes: Yup, I'll declare this const at the top and remove the dead code

vm/rust/src/jsonrpc.rs
```diff
@@ -58,6 +65,79 @@ struct StateDiff {
     replaced_classes: Vec<ReplacedClass>,
 }
 
+impl StateDiff {
+    pub fn from_state_maps(state_maps: &StateMaps) -> Self {
```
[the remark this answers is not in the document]
rianhughes: I've updated the code to implement the `From` trait

vm/rust/src/jsonrpc.rs
rianhughes: Thanks for pointing this out, I've updated the code so that we no longer need to clone, and can consume the stateMap

vm/rust/src/jsonrpc.rs
rianhughes: Updated

adapters/vm2core/vm2core.go
```diff
@@ -39,3 +40,50 @@ func AdaptOrderedEvents(events []vm.OrderedEvent) []*core.Event {
 	})
 	return utils.Map(events, AdaptOrderedEvent)
 }
+
+func AdaptStateDiff(fromStateDiff *vm.StateDiff) core.StateDiff {
+	var result core.StateDiff
```
rodrigo-pino, line 45: rename `result` to `stateDiff`. If you follow the upcoming suggestions we don't need this Var

adapters/vm2core/vm2core.go
```diff
@@ -39,3 +40,50 @@ func AdaptOrderedEvents(events []vm.OrderedEvent) []*core.Event {
...
+	if fromStateDiff == nil {
+		return result
+	}
```
rodrigo-pino, line 48: I know this pattern is what it's generally used but I prefer not to handle nil cases implicitly in the code. I think they should be handled explicitly. I believe they sum up for better readability for the long term (projects such as this) even if you have to write a bit more code.

  For example, the caller of this function should check that it's input parameter is not nil and the function. The function instead will assume if it gets called, there will be always something to adapt.

Of course, I think this are among the most subjective things so feel free to ignore. The idea is that `nil` is an edge case that you should handle explicitly: "this function adapts from one to the other, it is not the role of this function to adapt nothing" (again, very subjective)

Finally, if you are staying with this approach, I think is better to just return `core.StateDiff{}`

rodrigo-pino (approved):

adapters/vm2core/vm2core.go
rianhughes: renamed to `toStateDiff`

adapters/vm2core/vm2core.go
rianhughes: Yeah I see what you mean, and agree that explicitly handling the nil case is better for readability etc. However, I think checks like this help catch panics that arise in unexpected ways, so would personally prefer to keep it.

For example, we call this function after getting the CallResult from Vm.
```
type CallResult struct {
	StateDiff       StateDiff
}
```
Currently it's not possible to call `AdaptStateDiff` with a nil `StateDiff`. However if `CallResult` gets updated to this for some reason,
```
type CallResult struct {
	StateDiff       *StateDiff
}
```
then it might be possible to call the adapter function with `nil`, in some edge cases that we might not currently cover. And I don't think that whoever is responsible for updating CallResult will be thinking about effects like this across the entire codebase. 

Ofc this is a made-up example, but generally I'm for safety checks like this that can catch panics from unknown changes / edge cases. It's also a pattern we are currently using throughout the codebase, and I think there's value in keeping to a single approach.

rianhughes:
Failing test fixed by https://github.com/NethermindEth/juno/pull/2645

genesis/genesis.go
```diff
@@ -0,0 +1,367 @@
...
+}
+
+func deployContract(
+	v vm.VM,
+	network *utils.Network,
+	maxSteps uint64,
+	genesisState *sync.PendingStateWriter,
+	address felt.Felt,
+	contractData GenesisContractData,
+	constructorSelector *felt.Felt,
```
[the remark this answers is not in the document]
rianhughes: Yup, that's a good point. However, it would mean instantiating it every time we call `deployContract` in `deployContracts`. I'll leave it as it is for now if that's okay?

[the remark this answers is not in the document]
hudem1: Ok yeah sure!

core/state_update.go
```diff
@@ -39,6 +64,24 @@ func (d *StateDiff) Length() uint64 {
 	return uint64(length)
 }
 
+func (d *StateDiff) Merge(incoming *StateDiff) {
```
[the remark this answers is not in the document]
rianhughes: Ah nice catch!! MergeStateDiffs Can be removed completely. 

>  maps.Copy which overwrites only the oldMap's entries whose key is present in the newMap

it should also add key-value pairs from src to dst, even when dst doesn't have those key-value pairs.
```
func Copy(dst map[k]v, src map[k]v) { // simplified this for readability 
	for k, v := range src {
		dst[k] = v
	}
}
```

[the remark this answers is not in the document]
hudem1: Oh okay i see you just removed `MergeStateDiffs`, only `Merge` is useful then!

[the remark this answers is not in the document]
hudem1: And yeah, I read the doc for `maps.Copy` and indeed i missed the part where it was saying it writes also the src's entries whose key doesn't exist in dst. So `Merge` and `MergeStateDiffs` were doing the same thing then, thats why you removed your custom one, ie without `maps.Copy`, then

PR #2374, opened 2025-01-14
Skip tests BlockWithReceipts due to invalid test asserts

wojciechos:
The tests are incorrect because they expect transaction_hash in the transaction object, while the spec only requires it in the receipt. Skipping them for now in our pipeline, and will unskip them once the framework fixes the tests.


AnkushinDaniil (approved):

rodrigo-pino (approved):
How do we make sure that we don't forget to unskip them in the future?

PR #2378, opened 2025-01-15
Fix trace pending txns

wojciechos:
Fix support for tracing transactions in the pending block.

### DIFF
--- a/rpc/trace.go
+++ b/rpc/trace.go
@@ -11,6 +11,7 @@ import (
 	"github.com/NethermindEth/juno/blockchain"
 	"github.com/NethermindEth/juno/core"
 	"github.com/NethermindEth/juno/core/felt"
+	"github.com/NethermindEth/juno/db"
 	"github.com/NethermindEth/juno/jsonrpc"
 	"github.com/NethermindEth/juno/starknet"
 	"github.com/NethermindEth/juno/sync"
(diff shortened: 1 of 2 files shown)


rpc/trace.go
```diff
@@ -141,7 +142,7 @@ func (h *Handler) traceTransaction(ctx context.Context, hash *felt.Felt) (*vm.Tr
 	httpHeader := http.Header{}
 	httpHeader.Set(ExecutionStepsHeader, "0")
 
-	if err != nil {
+	if err != nil && !errors.Is(err, db.ErrKeyNotFound) {
```
derrix060, line 145: super nit

can't it be written as:

```suggestion
	if !errors.Is(err, db.ErrKeyNotFound) {
```

wojciechos: I guess it's common go pattern we use it everywhere in Juno (see [node.go](https://github.com/NethermindEth/juno/blob/d3ff578e4bbe21c85c7718986490107ab40a2a1d/node/node.go#L137), [trie.go](https://github.com/NethermindEth/juno/blob/d3ff578e4bbe21c85c7718986490107ab40a2a1d/core/trie/trie.go#L67), [sync.go](https://github.com/NethermindEth/juno/blob/d3ff578e4bbe21c85c7718986490107ab40a2a1d/p2p/sync/sync.go#L339) etc.)

rodrigo-pino: I think if you remove the first check then a `nil` err will turn true the second condition, which might not be the intended behaviour

rodrigo-pino (approved):
Looks good to me!

rianhughes (approved):

PR #2380, opened 2025-01-15
Set correct DHT protocol to match Starknet P2P v0.1

wojciechos:
Updated DHT protocol naming to align with Starknet P2P v0.1 spec and added a test.

Changes:
	•	Fixed incorrect DHT protocol naming:
	•	Before: /starknet/kad/1.0.0 (not matching any Starknet P2P spec)
	•	After: /starknet/kad/SN_MAIN/1.0.0 (matches [v0.1)](https://github.com/starknet-io/starknet-p2p-specs/blob/v0.1/p2p/proto/protocols.md#protocols-briefing)
	•	Added a test to validate the correct protocol format.


AnkushinDaniil (commented):
Have you tested sync with Pathfinder?

wojciechos:
> Have you tested sync with Pathfinder?

@AnkushinDaniil Yep, I tested it, but Pathfinder hasn’t updated to the latest P2P spec, which is why they can’t sync with each other. See [here](https://github.com/eqlabs/pathfinder/blob/main/crates/p2p/src/sync.rs#L21)

AnkushinDaniil (approved):

AnkushinDaniil (approved):

derrix060 (approved):

PR #2386, opened 2025-01-18
Make internal blockchain methods private

### LINKED ISSUE #2257: Blockchain's public functions with db's transaction should be private
For context please see discussion [here](https://github.com/NethermindEth/juno/pull/2184/files#r1824697619).

Exca-DK:
Closes #2257 

This PR unexports private blockchain functions that rely on transactions by copying the required functions to the migration package.

### DIFF
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -128,12 +128,12 @@ func (b *Blockchain) Height() (uint64, error) {
 	var height uint64
 	return height, b.database.View(func(txn db.Transaction) error {
 		var err error
-		height, err = ChainHeight(txn)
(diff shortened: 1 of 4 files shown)


rodrigo-pino:
There is a pro which is Juno's API becomes simpler, but there is a con that there is repetition, updating a method somewhere will require re-updating it in the migration module, if forgotten, it can cause bugs 🤔

Exca-DK:
Yes, I copied it over as suggested in [discussion](https://github.com/NethermindEth/juno/pull/2184/files#r1824697619)

rodrigo-pino:
Internally we decided not to have duplicated methods at the cost of public methods that should be private

PR #2387, opened 2025-01-19
Peer CIDR based filtering

Exca-DK:
This PR introduces a connection gater for P2P that enables CIDR-based filtering of peers. With this feature, peers outside the specified allowed ranges can be blocked, reducing the risk of netscan-related bans in datacenter environments.


rodrigo-pino:
Setting as no stale. It is not entirely necessary now be we can come back to it in the future

PR #2388, opened 2025-01-20
Reduce number of allocations in p2p

kirugan:
TL;DR: instead of allocating response message on every iteration we allocate it once before iterations and reset it once we done with it after each iteration

Yesterday, I was watching a video about generics and thought that this function (particularly the reflection part) could be rewritten. Unfortunately, after 10 minutes of trial and error, I didn't succeed, but at least I found a way to reduce allocations and reflection calls.


AnkushinDaniil (approved):
p2p sync works

rodrigo-pino (approved):
LGTM! Let's run the P2P tests before merging

PR #2402, opened 2025-01-26
Added custom type for Deployed contracts throughout the repo wherever necessary

### LINKED ISSUE #2396: Custom types for commonly used structures
There are some places in the code where structures such as `map[felt.Felt]map[felt.Felt]*felt.Felt` to refer to `StorageDiff` or `map[felt.Felt]*felt.Felt` to refer to others like `DeployedContracts`, are used.
We could define instead a custom type like:
```
type StorageDiff map[felt.Felt]map[felt.Felt]*felt.Felt
```
Or consider creating types such as `Address` or `ClassHash` for felts, so `DeployedContracts` map would look like `map[Address]ClassHash`.

This will allow the code to be more readable and easier to work with (when creating new structures or passing them as arguments in functions). It would also make it safer as you restrict the specific type that must be used (not just any felt for example), and it even allows to create methods on top of these structures.

Check out these two comments for reference:
- https://github.com/NethermindEth/juno/pull/2371#discussion_r1920152402
- https://github.com/NethermindEth/juno/pull/2371#discussion_r1925141010

Deeptanshu-sankhwar:
Fixes #2396 

This PR introduces the structure `type StorageDiff map[felt.Felt]map[felt.Felt]*felt.Felt` for refactoring `StorageDiff` initially as `map[felt.Felt]map[felt.Felt]*felt.Felt`.

It integrates the new struct throughout the repo wherever needed.

### DIFF
--- a/adapters/p2p2core/state.go
+++ b/adapters/p2p2core/state.go
@@ -33,7 +33,7 @@ func AdaptStateDiff(reader core.StateReader, contractDiffs []*gen.ContractDiff, 
 
 	var deployedContracts, replacedClasses []addrToClassHash
 	// addr -> {key -> value, ...}
-	storageDiffs := make(map[felt.Felt]map[felt.Felt]*felt.Felt)
+	storageDiffs := make(core.StorageDiff)
 	nonces := make(map[felt.Felt]*felt.Felt)
 	for _, diff := range contractDiffs {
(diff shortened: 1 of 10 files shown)


rodrigo-pino:
We were looking for something with a bit more depth. I am going to close this PR for the time being. We can re-open it later or create a new one.

PR #2408, opened 2025-01-29
chore: changed the link to the shield test

Olexandr88:
in my opinion, this link is more suitable here


rodrigo-pino:
> in my opinion, this link is more suitable here

Could explain why?

Olexandr88:
> Could explain why?

since the shield is called test, if you think it's inappropriate, rename the shield to build

rodrigo-pino (approved):
Makes sense!

wojciechos (approved):

PR #2409, opened 2025-01-30
Web socket updates

IronGauntlets:
This PR introduces the most recent changes to websocket specs, please see each commit.

Todo:
- Fix sending of pending events. Currently, only when a new block is added the latest events are sent, however, there cannot be any duplicates and we need to make sure that the events are sent once the pending block is updated.
- Add subscription block id, so users cannot call ws subscription with `blockID = pending`.
- Add max websocket connections

### DIFF
--- a/jsonrpc/websocket.go
+++ b/jsonrpc/websocket.go
@@ -78,9 +78,7 @@ func (ws *Websocket) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	}
 
 	// Ensure we decrease the connection count when the connection closes
-	defer func() {
-		ws.connCount.Add(-1)
-	}()
+	defer ws.connCount.Add(-1)
 
 	// TODO include connection information, such as the remote address, in the logs.
 
--- a/jsonrpc/websocket.go
+++ b/jsonrpc/websocket.go
@@ -5,21 +5,29 @@ import (
 	"io"
 	"net/http"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/NethermindEth/juno/utils"
 	"github.com/coder/websocket"
 )
 
-const closeReasonMaxBytes = 125
+const (
+	closeReasonMaxBytes = 125
+	maxConns            = 2048 // TODO: an arbitrary default number, should be revisited after monitoring
+)
 
 type Websocket struct {
 	rpc        *Server
 	log        utils.SimpleLogger
 	connParams *WebsocketConnParams
 	listener   NewRequestListener
+	shutdown   <-chan struct{}
 
-	shutdown <-chan struct{}
+	// Add connection tracking
+	connCount int
+	maxConns  int
+	connMu    sync.RWMutex
 }
 
 func NewWebsocket(rpc *Server, shutdown <-chan struct{}, log utils.SimpleLogger) *Websocket {
@@ -29,8 +37,15 @@ func NewWebsocket(rpc *Server, shutdown <-chan struct{}, log utils.SimpleLogger)
 		connParams: DefaultWebsocketConnParams(),
 		listener:   &SelectiveListener{},
 		shutdown:   shutdown,
+		maxConns:   maxConns,
 	}
 
+	return ws
+}
+
+// WithMaxConnections sets the maximum number of concurrent websocket connections
+func (ws *Websocket) WithMaxConnections(max int) *Websocket {
+	ws.maxConns = max
 	return ws
 }
 
@@ -49,11 +64,29 @@ func (ws *Websocket) WithListener(listener NewRequestListener) *Websocket {
 // ServeHTTP processes an HTTP request and upgrades it to a websocket connection.
 // The connection's entire "lifetime" is spent in this function.
 func (ws *Websocket) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	// Check connection limit
+	ws.connMu.Lock()
+	if ws.connCount >= ws.maxConns {
+		ws.connMu.Unlock()
+		ws.log.Warnw("Max websocket connections reached", "maxConns", ws.maxConns)
+		http.Error(w, "Too many connections", http.StatusServiceUnavailable)
+		return
+	}
+	ws.connCount++
+	ws.connMu.Unlock()
+
 	conn, err := websocket.Accept(w, r, nil /* TODO: options */)
 	if err != nil {
 		ws.log.Errorw("Failed to upgrade connection", "err", err)
 		return
 	}
+
+	// Ensure we decrease the connection count when the connection closes
+	defer func() {
+		ws.connMu.Lock()
+		ws.connCount--
+		ws.connMu.Unlock()
+	}()
 
 	// TODO include connection information, such as the remote address, in the logs.
 
--- a/jsonrpc/websocket.go
+++ b/jsonrpc/websocket.go
@@ -5,7 +5,7 @@ import (
 	"io"
 	"net/http"
 	"strings"
-	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/NethermindEth/juno/utils"
@@ -25,9 +25,8 @@ type Websocket struct {
 	shutdown   <-chan struct{}
 
 	// Add connection tracking
-	connCount int
-	maxConns  int
-	connMu    sync.RWMutex
+	connCount atomic.Int32
+	maxConns  int32
 }
 
 func NewWebsocket(rpc *Server, shutdown <-chan struct{}, log utils.SimpleLogger) *Websocket {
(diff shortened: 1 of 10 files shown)


rpc/helpers.go
```diff
@@ -60,22 +60,34 @@ func (h *Handler) blockByID(id *BlockID) (*core.Block, *jsonrpc.Error) {
...
-		header, err = h.bcReader.HeadsHeader()
-	case id.Hash != nil:
-		header, err = h.bcReader.BlockHeaderByHash(id.Hash)
-	case id.Pending:
-		var pending *sync.Pending
-		pending, err = h.syncReader.Pending()
-		if err == nil {
-			header = pending.Block.Header
+	switch id := id.(type) {
+	case *BlockID:
```
rodrigo-pino, line 67: It is confusing having a struct named `BlockID` and an interface named `blockID`. 
Also, this function has some repeated code.

Proposal: let's change the function to be generic on the interface `blockID` (using the same name, but ideally, it will have another less confusing, perhaps `subscriptionID`)
```go
interface blockID {
  // methods to get `Latest`, `Hash` and `Pending`
}
```
Have `BlockID` and `SubscriptionBlockID` implement the interface.

rpc/subscriptions.go
```diff
@@ -518,17 +589,28 @@ func (h *Handler) sendPendingTxs(w jsonrpc.Conn, result any, id uint64) error {
 
 // resolveBlockRange returns the start and latest headers based on the blockID.
 // It will also do some sanity checks and return errors if the blockID is invalid.
-func (h *Handler) resolveBlockRange(blockID *BlockID) (*core.Header, *core.Header, *jsonrpc.Error) {
+func (h *Handler) resolveBlockRange(id blockID) (*core.Header, *core.Header, *jsonrpc.Error) {
```
rodrigo-pino, line 592: The same principle can be applied here

rodrigo-pino (commented):
LGTM overall, just some nitpicks

rpc/helpers.go
```diff
@@ -60,22 +60,23 @@ func (h *Handler) blockByID(id *BlockID) (*core.Block, *jsonrpc.Error) {
 	return block, nil
 }
 
-func (h *Handler) blockHeaderByID(id *BlockID) (*core.Header, *jsonrpc.Error) {
+func (h *Handler) blockHeaderByID(id BlockIdentifier) (*core.Header, *jsonrpc.Error) {
```
rodrigo-pino, line 63: I failed to realize that `blockHeaderByID` was a method and not a function in my previous suggestions (so applying generics makes it very hard). With that in context, the previous approach seemed fine to me as well. I think neither is objectively better than the other. Feel free to leave the one you prefer.

rpc/subscriptions.go
```diff
@@ -24,15 +26,75 @@ var (
 	subscribeTxStatusTickerDuration = 5 * time.Second
 )
 
+var (
+	_ BlockIdentifier = (*SubscriptionBlockID)(nil)
+	_ BlockIdentifier = (*BlockID)(nil)
+)
```
rodrigo-pino, line 32: Are these vars left overs?

rpc/subscriptions.go
```diff
@@ -68,23 +126,54 @@ func (h *Handler) SubscribeEvents(ctx context.Context, fromAddr *felt.Felt, keys
...
+					}
+
+					for i, r := range b.Receipts {
+						for _, e := range r.Events {
+							fe := blockchain.FilteredEvent{
+								Event:           e,
+								BlockNumber:     header.Number,
+								BlockHash:       header.Hash,
+								TransactionHash: b.Transactions[i].Hash(),
+							}
```
rodrigo-pino, line 170: Since this is a public function, is there any check that verifies that length of  `b.Receipts` and `b.Transactions` matches?

If this safety is implicitly guaranteed because it is based on the correctness of the network then I think we don't need to add a check.

rpc/subscriptions.go
```diff
@@ -68,23 +126,54 @@ func (h *Handler) SubscribeEvents(ctx context.Context, fromAddr *felt.Felt, keys
...
 				case header := <-headerSub.Recv():
-					h.processEvents(subscriptionCtx, w, id, header.Number, header.Number, fromAddr, keys)
+					// During syncing the events from the new head still need to be sent as there is no pending block.
+					// However, it is not easy to tell when the node is syncing.
+					// To solve this issue, we can send the events regardless, and if the node is done syncing, then the
+					// latest header events would have been sent when the pending block was updated. Hence,
+					// trying to resend the event should be of no consequences and the map can be safely emptied.
+					h.processEvents(subscriptionCtx, w, id, header.Number, header.Number, fromAddr, keys, eventsPreviouslySent)
+
+					b, err := h.bcReader.BlockByNumber(header.Number)
```
rodrigo-pino, line 157: Could `b` be renamed to `block`

rpc/subscriptions.go
```diff
@@ -68,23 +126,54 @@ func (h *Handler) SubscribeEvents(ctx context.Context, fromAddr *felt.Felt, keys
...
 		wg.Go(func() {
+			// Stores the transaction hash -> number of events
+			eventsPreviouslySent := make(map[blockchain.FilteredEvent]struct{})
+
 			for {
 				select {
 				case <-subscriptionCtx.Done():
 					return
 				case header := <-headerSub.Recv():
-					h.processEvents(subscriptionCtx, w, id, header.Number, header.Number, fromAddr, keys)
```
rodrigo-pino, line 87: It seems that the entire `case header` code is not being tested. Could you please add a test that follows this execution path.

rpc/helpers.go
weiihann: The current one seems cleaner, will leave it as it is

rpc/subscriptions.go
weiihann: No, it's just compile-time interface checks.

rpc/subscriptions.go
weiihann: Yes it's guaranteed to be equal

rpc/subscriptions_test.go
```diff
@@ -187,7 +207,7 @@ func TestSubscribeEvents(t *testing.T) {
 
 		mockChain.EXPECT().HeadsHeader().Return(&core.Header{Number: b1.Number}, nil)
 		mockChain.EXPECT().BlockHeaderByNumber(b1.Number).Return(b1.Header, nil)
-		mockChain.EXPECT().EventFilter(fromAddr, keys).Return(mockEventFilterer, nil)
+		mockChain.EXPECT().EventFilter(gomock.Any(), gomock.Any()).Return(mockEventFilterer, nil)
```
AnkushinDaniil, line 210: Why do we need this change? The old version seems more accurate.
Same question for analogous changes.

AnkushinDaniil (commented):
First round of review, I'll continue later

rpc/subscriptions_test.go
weiihann: Reverted this part since it doesn't fix the intermittent test error

rpc/block.go
```diff
@@ -67,6 +67,22 @@ type BlockID struct {
...
+	return b.Pending
+}
+
+func (b *BlockID) GetHash() *felt.Felt {
+	return b.Hash
+}
+
+func (b *BlockID) GetNumber() uint64 {
+	return b.Number
+}
```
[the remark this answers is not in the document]
weiihann: We can't make the fields private because external packages (such as `rpc_test`) still need to set it directly. Creating setter methods for them seem overkill and overengineering. For now I think this is good enough, changing public to private and using interface methods seems more like a nice-to-have semantics kind of thing but we would have to make a lot of unnecessary changes.

jsonrpc/websocket.go
```diff
@@ -49,12 +64,30 @@ func (ws *Websocket) WithListener(listener NewRequestListener) *Websocket {
...
+	// Check connection limit
+	ws.connMu.Lock()
+	if ws.connCount >= ws.maxConns {
+		ws.connMu.Unlock()
+		ws.log.Warnw("Max websocket connections reached", "maxConns", ws.maxConns)
+		http.Error(w, "Too many connections", http.StatusServiceUnavailable)
+		return
+	}
+	ws.connCount++
+	ws.connMu.Unlock()
```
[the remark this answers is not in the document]
weiihann: ahh yes atomic is a better alternative

jsonrpc/websocket.go
```diff
@@ -49,12 +63,25 @@ func (ws *Websocket) WithListener(listener NewRequestListener) *Websocket {
...
 	conn, err := websocket.Accept(w, r, nil /* TODO: options */)
 	if err != nil {
 		ws.log.Errorw("Failed to upgrade connection", "err", err)
 		return
 	}
 
+	// Ensure we decrease the connection count when the connection closes
+	defer func() {
+		ws.connCount.Add(-1)
+	}()
```
AnkushinDaniil, line 83:
```suggestion
	defer ws.connCount.Add(-1)
```

rodrigo-pino (approved):
LGTM

rpc/block.go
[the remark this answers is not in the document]
AnkushinDaniil: We have a lot of cases in the code where we use these public fields to get data, shouldn't we update the code and use these methods?

AnkushinDaniil (approved):

PR #2413, opened 2025-02-03
split rpc into v7 and v8 subpkgs

rianhughes:
This PR splits the RPC pkg into sub packages for each RPC version. It also adds support for RPCv6.

Todo in a separate PR (this one already contains significant changes): delete the repeated code 

Note for self: re-add juno subscriptions to top handler


rpc/rpccore/rpccore.go
rodrigo-pino, line 1: I think it could be called `core` only. I think it shouldn't be a module but a single file, given it's not too big.

Nit: I liked more the name `common` because it made more sense that it was the things in common each rpc version had. For me, core represents something that it is integral of the RPC package, in this case, the rpc code that sends and receive requests, not the utilities we used for that. Nonetheless, this is all very subjective feel free to stick with this.

rianhughes:
created a new PR with a single commit (rebasing involved conflicts)

PR #2417, opened 2025-02-04
Move starknet_getStorageProof method from RPC v0.7 to v0.8

wojciechos:
According to the Starknet spec, starknet_getStorageProof should be part of RPC v0.8 and does not exist in v0.7. This PR moves the method to the correct version for spec compliance.


derrix060 (approved):

rodrigo-pino (approved):

rodrigo-pino:
@rianhughes please take this into account for #2413. Hope that merging this won't cause to much issue on that branch.

Issue #2418, opened 2025-02-04
Optional Parameters Not Working in Positional Calls


wojciechos:
When calling certain methods with positional parameters and omitting optional ones, Juno returns an error indicating invalid parameters.

**Sample request:**
```
{
    "jsonrpc": "2.0",
    "method": "starknet_getStorageProof",
    "params": [
        "latest"
    ],
    "id": 0
}
```

**Current Response:**
```
{
    "jsonrpc": "2.0",
    "error": {
        "code": -32602,
        "message": "Invalid Params",
        "data": "missing/unexpected params in list"
    },
    "id": 0
}
```

**Expected Behavior:**
The request should be processed successfully since the [omitted parameters are optional](https://github.com/starkware-libs/starknet-specs/blob/28ad3f56c147967b63ecfc78db5f4ba1df993e05/api/starknet_api_openrpc.json#L898).

PR #2421, opened 2025-02-05
Fix positional optional params

weiihann:
This PR fixes optional parameters in positional format.

How to reproduce:
```
{
    "jsonrpc": "2.0",
    "method": "starknet_getStorageProof",
    "params": [
        "latest"
    ],
    "id": 0
}
```

Expected behaviour:
- Return a valid output

Actual behaviour:
```
"error": {
    "code": -32602,
    "message": "Invalid Params",
    "data": "missing/unexpected params in list"
}
```

### DIFF
--- a/jsonrpc/server.go
+++ b/jsonrpc/server.go
@@ -115,6 +115,10 @@ type Method struct {
 	// The method takes a context as its first parameter.
 	// Set upon successful registration.
 	needsContext bool
+
+	// The number of optional parameters in the method.
+	// Set upon successful registration.
+	optionalCount int
 }
 
 type Server struct {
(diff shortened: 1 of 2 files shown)


wojciechos (approved):

jsonrpc/server.go
```diff
@@ -115,6 +115,10 @@ type Method struct {
 	// The method takes a context as its first parameter.
 	// Set upon successful registration.
 	needsContext bool
+
+	// The number of optional parameters in the method.
+	// Set upon successful registration.
+	optionalCount int
```
rodrigo-pino, line 121: This is just a nit, for me it makes more sense to stored the `requiredParametersCount` directly rather than the `optionalCount`. The goal of this last field is to know the required parameters, so why not stored that directly?

In the end, it is all the same, but I think our brains (or at least mine) finds this approach more straight forward. What do you think?

rodrigo-pino (commented):
LGTM! Just one tiny nit.

jsonrpc/server.go
weiihann: `requiredParametersCount` is slightly cleaner

wojciechos:
Closes [2418](https://github.com/NethermindEth/juno/issues/2418)

PR #2422, opened 2025-02-05
Use `sync.Map` for subscriptions map

weiihann:


AnkushinDaniil (approved):
LGTM
I think we need to rewrite tests to isolate test cases and not use the same mocks for different test cases.
`make test-race` still fails.
Should we create an issue for that?

weiihann:
> LGTM I think we need to rewrite tests to isolate test cases and not use the same mocks for different test cases. `make test-race` still fails. Should we create an issue for that?

Yeah, the `make test-race` result seems somewhat scary. We should create an issue to fix all the data race issues.

rodrigo-pino (approved):

PR #2427, opened 2025-02-05
Weiihann/fix test race

weiihann:


jsonrpc/websocket.go
```diff
@@ -49,12 +63,23 @@ func (ws *Websocket) WithListener(listener NewRequestListener) *Websocket {
 // ServeHTTP processes an HTTP request and upgrades it to a websocket connection.
 // The connection's entire "lifetime" is spent in this function.
 func (ws *Websocket) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	// Check connection limit
+	if ws.connCount.Load() >= ws.maxConns {
```
infrmtcs, line 67: I think race condition can cause the connection count going over the limit. For example, 2 requests reach this at the same time when the `connCount` is 2047, both of them will pass this check and acquire a new connection.

I think a simpler option to consider is to use https://pkg.go.dev/golang.org/x/sync@v0.11.0/semaphore.

PR #2429, opened 2025-02-06
feat(onboarding): Add rpc endpoint juno_getBlockWithTxnHashesAndReceipts

hudem1:
Onboarding implementation task 1: endpoint `juno_getBlockWithTxnHashesAndReceipts`

Note: I saw other helper function whose visibility was public, so, I changed the visibility of `adaptBlockHeader` to be able to use it in the tests. Let me know if it's something we don't want to do, I'll revert it and change my test.

I set it as draft just to make sure it doesn't get merged.

### DIFF
--- a/rpc/block.go
+++ b/rpc/block.go
@@ -154,6 +154,13 @@ type BlockWithReceipts struct {
 	Transactions []TransactionWithReceipt `json:"transactions"`
 }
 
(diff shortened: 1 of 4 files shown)


rodrigo-pino:
LGTM! Closing

PR #2433, opened 2025-02-07
Fix inconsistent connection count

Exca-DK:
This pr moves deferred function up so that active connection counting stays consistent, even when Accept fails e.g. due to `too many open files`


AnkushinDaniil (approved):

rodrigo-pino (approved):
Good catch, thank you!

PR #2436, opened 2025-02-10
move rpc handler blockHashAndNumber from v7 and v8 to v6

rianhughes:


rodrigo-pino:
This looks good. There are some dependency updates that are unrelated with the PR. I wonder if we could merge a separate PR with those first.

rodrigo-pino (approved):

Issue #2437, opened 2025-02-10
Cleanup rpc pkg


rianhughes:
Currently there is a lot of duplicated code in the rpc `v6`, `v7` and `v8` sub packages. We should remove duplicated code (types, functions, tests) where possible. For example, the `blockHashAndNumber` function (and associated type `BlockHashAndNumber`) has not changed between `v6`, `v7` and `v8`. This means that we can remove this function (and type) from `v7` and `v8`, and have the top level handler use the code in `v6` instead (specifically update the `MethodsV0_8` and `MethodsV0_7` functions). See this [PR](https://github.com/NethermindEth/juno/pull/2436/files) as an example. We should do this for all RPC methods. I would suggest doing the following:

1. Create a PR for each RPC method (this will make it easier to track changes, work on, and review)
2. If the RPC method has not changed between (eg) `v6`, `v7` and `v8`, then we can just re-use the `v6` code in `v7` and `v8`. This means the corresponding code (function, types, tests) should be deleted in `v7` and `v8`.
3. If the RPC function has changed, then we should not remove the function. However, if the associated type has not changed, then we should probably delete it.
4. To keep track of all the changes being made, and still to be made, to the RPC pkg, please add a new bullet point here and link the associated PR.

- [x] [starknet_chainId](https://github.com/NethermindEth/juno/pull/2446)
- [x] [starknet_blockNumber](https://github.com/NethermindEth/juno/pull/2445)
- [x] [starknet_blockHashAndNumber](https://github.com/NethermindEth/juno/pull/2436/files)
- [x] [BlockStatus enum](https://github.com/NethermindEth/juno/pull/2447) (Not really an RPC endpoint but useful for other endpoints, such as starknet_getBlockWithTxHashes)
- [x] [Header struct](https://github.com/NethermindEth/juno/pull/2448) (Not really an RPC endpoint but useful for other endpoints, such as starknet_getBlockWithTxHashes)
- [x] [starknet_getBlockWithTxHashes](https://github.com/NethermindEth/juno/pull/2449)
- [x] [Refactor type Transaction struct](https://github.com/NethermindEth/juno/issues/2458) (Not really an RPC endpoint but useful for other endpoints)
- [x] [starknet_syncing](https://github.com/NethermindEth/juno/pull/2459)
- [x] [starknet_getNonce](https://github.com/NethermindEth/juno/pull/2463)
- [x] [starknet_getClassAt](https://github.com/NethermindEth/juno/pull/2465)
[...]

PR #2448, opened 2025-02-10
Reuse rpcv6 BlockHeader for v7 and v8

hudem1:
clean rpc pkgs according to issue #2437 

- rpcv7's `Header` is the same as v6, so, i just removed it and reused v6's entirely.
- rpcv8's `Header` just has one more field `L2GasPrice`. So, I embedded rpcv6.Header into rpcv8.Header. Let me know whether it is ok to do that. I thought it could be nice as it reuses v6's Header and reduces code dup and makes it clear what new fields have been added, but when instantiating a v8 Header, we have to define an rpcv6 Header field inside, so it might be a bit strange maybe? ie:
```go
rpcv8.BlockHeader {
   BlockHeader: rpcv6.BlockHeader { ... },
   L2GasPrice: ...,
}
```
And also, as you can see in the tests (in file `subscriptions_test.go`) that i had to fix, it changes the order of the fields in the stringified json. Could that be an issue ? I don't think so but still asking to make sure it's fine!


rpc/v8/block.go
```diff
@@ -278,55 +245,57 @@ func adaptBlockHeader(header *core.Header) BlockHeader {
...
+			Timestamp:        header.Timestamp,
+			SequencerAddress: sequencerAddress,
+			L1GasPrice: &rpcv6.ResourcePrice{
+				InWei: header.L1GasPriceETH,
+				InFri: nilToZero(header.L1GasPriceSTRK),
+			},
+			L1DataGasPrice:  &l1DataGasPrice,
+			L1DAMode:        &l1DAMode,
+			StarknetVersion: header.ProtocolVersion,
 		},
```
rodrigo-pino, line 297: Although I agree with this idea of composing structs, they are also part of our public API. 

The objects being composed  mean that eventually, when rpc v6 gets deprecated and subsequently rpcv7, this struct has to change. This means that anyone importing this code will have to update their existing code to interact with rpcv8. 

I don't know how big a deal this is, maybe it is irrelevant if nobody uses Juno like that. @rianhughes what are your thoughts on this?

hudem1: I am not sure I understand what you mean. When rpcv6 and v7 get deprecated, users running juno nodes will have to fetch the latest juno code version for all rpcv8 endpoints using rpcv6 code anyway, right ? So, I am not sure I understand why this example in particular is different ?
When v6 and v7 get deprecated, we will create the struct in rpcv8 pkg. So, if they fetch the latest juno code, they won't have to update anything, right ?

rodrigo-pino: It's a very minor worry, and I don't think it is a real one either, let's bypass it.

rodrigo-pino (approved):

rianhughes (approved):

rpc/v8/block.go
rianhughes: After this update, any users who want to build an rpcv8 BlockHeader will have to do so like
```

BlockHeader{
		BlockHeader: rpcv6.BlockHeader{
			Hash:             header.Hash,
			ParentHash:       header.ParentHash,
			...
			StarknetVersion: header.ProtocolVersion,
		},
		L2GasPrice: &l2GasPrice,
	}
}
```

and if rpcv6 gets deprecated and the fields get inlined, then they will create BlockHeaders like this
```

BlockHeader{
			Hash:             header.Hash,
			ParentHash:       header.ParentHash,
			...
			StarknetVersion: header.ProtocolVersion,
		L2GasPrice: &l2GasPrice,
	}
}
```

Imo while we should try and minimise the changes users need to make with each Juno release, I don't think there are many users directly importing our RPC pkg (or at least I imagine they should use something like starknet.go), and it's a minor change on their part so I'm okay to compose the structs.

PR #2453, opened 2025-02-11
Fix concurrent RPC handlers

### LINKED ISSUE #2450: Subscribtions stop Pushing Updates
Subscriptions only push data to till latest block and then stop. When a new block appears, no further events are received.  

#### Steps to Reproduce  

Initial conditions: Synced node

1. Connect via ws and subscribe to new block headers:  
```
   {
       "jsonrpc": "2.0",
       "method": "starknet_subscribeNewHeads",
       "params": { "block_id": "latest" },
       "id": 1
   }
```

2.	Receive expected response:
```
{ "jsonrpc": "2.0", "result": 12153704810663065403, "id": 1 }
```

3. First event (works):
```
{"jsonrpc": "2.0", "method": "starknet_subscriptionNewHeads", "params": { "result": { "block_hash": "0x67d613..." } } }
```
4. **No further events when new blocks sync.**

weiihann:
`rpcv6Handler.Run()` blocks the rest of the RPC handlers from running, they should be running concurrently in goroutines.

Fixes #2450

### DIFF
--- a/rpc/handlers.go
+++ b/rpc/handlers.go
@@ -13,6 +13,7 @@ import (
 	"github.com/NethermindEth/juno/sync"
 	"github.com/NethermindEth/juno/utils"
 	"github.com/NethermindEth/juno/vm"
+	"golang.org/x/sync/errgroup"
 )
 
 type Handler struct {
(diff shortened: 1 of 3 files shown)


rpc/handlers.go
```diff
@@ -77,15 +80,50 @@ func (h *Handler) WithGateway(gatewayClient rpccore.Gateway) *Handler {
 }
 
 func (h *Handler) Run(ctx context.Context) error {
```
AnkushinDaniil, line 82: Can we use errgroup? 
```go
func (h *Handler) Run(ctx context.Context) error {
	eg, egCtx := errgroup.WithContext(ctx)

	eg.Go(func() error { return h.rpcv6Handler.Run(egCtx) })
	eg.Go(func() error { return h.rpcv7Handler.Run(egCtx) })
	eg.Go(func() error { return h.rpcv8Handler.Run(egCtx) })

	return eg.Wait()
}
```

AnkushinDaniil (approved):

rpc/handlers.go
weiihann: That's cleaner

rodrigo-pino (approved):

Issue #2392, opened 2025-01-21
RPC08 - add Support for ENTRYPOINT_NOT_FOUND Error in starknet_call


wojciechos:
In RPC spec v0.8, a new error ENTRYPOINT_NOT_FOUND was introduced. It should return error code 21 when the specified entry_point_selector in starknet_call does not exist in the contract:

https://github.com/starkware-libs/starknet-specs/commit/85811f92c22c37dcbe8582238cb4c715c8832cc6

**Sample request on sepolia:**
```
{
    "jsonrpc": "2.0",
    "method": "starknet_call",
    "params": {
        "request": {
            "contract_address": "0x78d8ada537ab1132e600b7553c3e574336c4e0f24fb584b8b2d5faff8a37590",
            "entry_point_selector": "0x11c957afc7e84fbf8aac7ff9c79910a53a64edf9a779b18b411fbb1982fb84f", //does not exist
            "calldata": []
        },
        "block_id": "pending"
    },
    "id": 1
}
```

**Expected Response:**
```{
    "jsonrpc": "2.0",
    "error": {
        "code": 21,
        "message": "Requested entrypoint does not exist in the contract"
    },
    "id": 1
}

Exca-DK:
Is this PR still up for grabs? And should it be tackled after https://github.com/NethermindEth/juno/pull/2419 ?

rodrigo-pino:
@Exca-DK yes, up for grabs if interested!

PR #2455, opened 2025-02-11
rpc: handle the entrypoint not found error

rianhughes:
closes: https://github.com/NethermindEth/juno/issues/2392

The `ENTRYPOINT_NOT_FOUND` error is handled at the RPC level because it's less invasive relative to explicitly propogating it up from `vm/rust/lib.rs`, to ` vm/vm.go` and finally `rpc`. 

Also note that I reverted two changes made to the `vm`, specifically in `vm/rust/lib.rs`, which were introduced when trying to solve failing CI/CD tests related to this error. 

1. I removed the early return in `cairoVMExecute` because this change was introduced when trying to solve failing tests that were caused by `ENTRYPOINT_NOT_FOUND`, which is related to `starknet_call`, and has nothing to do with `cairoVMExecute`. This early return may cause unintended problems/errors.
2. I removed the early return in `cairoVMCall` because we can just handle this case in `rpc`.

CI/CD: https://github.com/NethermindEth/juno/actions/runs/13310265576

Note: because blockifier now requires us to handle errors differently, the format of the `Data` field in `Contract_Errors`, may be different (specifically in [this](https://github.com/NethermindEth/juno/pull/2455/files#diff-457536e4359e7f7299d13c55e148e57625d4230cdfe8c04c1cf0626ff82378cdR436) case). This may / may not be a problem for v6 and v7 responses. Ideally, blockifier would return the formatted error in the `call_info` (when `call_info.execution_failed==true`) instead of relying on each full_node client to implement it separately.


rpc/v7/trace.go
```diff
@@ -432,5 +432,8 @@ func (h *Handler) Call(funcCall FunctionCall, id BlockID) ([]*felt.Felt, *jsonrp
 		}
 		return nil, makeContractError(err)
 	}
+	if len(res) != 0 && res[0].String() == rpccore.EntrypointNotFoundFelt {
+		return nil, rpccore.ErrContractError
+	}
```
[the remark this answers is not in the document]
rodrigo-pino: Also, I am in favor on making this condition a small function and a part of `rpccore.go`. The function name describing what is checking for, it seems so hard to understand right now with no context.

rpc/v6/trace.go
rodrigo-pino, line 1: Are only the `call` method affected? I know we discussed this before but can you remind me why for example `simulateTx` shouldn't have this check?

rodrigo-pino: Also, why `Call` only calls `call`? Why don't we move the definition of `call` inside `Call`?

rpc/v7/trace.go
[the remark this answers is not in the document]
rianhughes: >I am probably missing something, but can we claim with 100% certainty that EntryPointNotFoundFelt will be always one the first posiition?

I would need to double check if this is guaranteed, but it seems to be how blockifier 0.13.4 behaves, and also how pathfinder handle it

>Also, I am in favor on making this condition a small function and a part of rpccore.go. The function name describing what is checking for, it seems so hard to understand right now with no context.

After some updates this code is only in v8, but I added a comment for context

rpc/v6/trace.go
rianhughes: > Are only the call method affected? I know we discussed this before but can you remind me why for example simulateTx shouldn't have this check?

Sure, the only way to call `vm.Call()` is via `starknet_call`. Other rpc methods like `starknet_estimateFee`, `starknet_simulateTransaction` go through `vm.Execute()`. These two call different execution functions (CallEntryPoint.execute() vs Transaction.execute()), and only the former calls blockifiers `execute_entry_point_call_wrapper` function, which is the origin of the change in behaviour.

rianhughes: > Also, why Call only calls call? Why don't we move the definition of call inside Call?

It was a side effect of restructuring the rpc pkg

rpc/v7/trace.go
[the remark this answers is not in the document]
rianhughes: It looks like we are guaranteed that res[0] is the entry point felt.

https://github.com/starkware-libs/sequencer/blob/main/crates/blockifier/src/execution/execution_utils.rs#L98

AnkushinDaniil (approved):

vm/rust/src/error_stack.rs
AnkushinDaniil, line 1: What is this file for?

AnkushinDaniil, line 1: I removed it and local tests passed

AnkushinDaniil (changes requested):

vm/vm.go
```diff
@@ -31,10 +31,15 @@ type ExecutionResults struct {
 	NumSteps         uint64
 }
 
+type CallResult struct {
+	Result          []*felt.Felt
+	ExecutionFailed bool
+}
```
rodrigo-pino, line 37: 👌

rodrigo-pino (commented):
If we are going to report an error every time execution failed, why would we need to add an extra flag to execution failed instead of returning an error directly?

vm/rust/src/error_stack.rs
rianhughes: Yup, it should be removed, I was testing something out

rianhughes:
> If we are going to report an error every time execution failed, why would we need to add an extra flag to execution failed instead of returning an error directly?

It's because the error handling it being done in both rust and go. If we did all the error handling in rust, then we could remove `ExecutionFailed`, and just return an error directly (without any more error handling on the go side). I briefly tried this in Rust, but found it was messy, so I just opted to return a flag and handle it on the Golang side. Doing all of this in Rust could be a nice follow-up PR, but I'm happy with the current approach.

AnkushinDaniil (approved):

rodrigo-pino:
From the description of your PR:

> and has nothing to do with cairoVMExecute. This early return may cause unintended problems/errors.

What sort of problems could we have?

> I removed the early return in cairoVMCall because we can just handle this case in rpc.

Why is it better handling it in the RPC package?

rodrigo-pino:
> Doing all of this in Rust could be a nice follow-up PR,

Is it not how it currently is right now. Am I missing something?

Issue #2393, opened 2025-01-21
RPC08 - Change offset to be integer


wojciechos:
In the RPC v0.8 spec, the offset field was updated to be an integer instead of a num as hex - https://github.com/starkware-libs/starknet-specs/commit/78ba4b27aea03b7a45325a7e0fddbd427b5537ab

**Sample request - sepolia:**
```
{
    "jsonrpc": "2.0",
    "method": "starknet_getCompiledCasm",
    "params": {
        "class_hash": "0x00eb46cb744cc47f6e6241e50724d595874187bfd7b63ef3998ee75b24c47152"
    },
    "id": 1
}
```

Expected response:
```
...
"CONSTRUCTOR": [
    {
        "offset": 4236, // should be an integer
        "selector": "0x28ffe4ff0f226a9107253e17a904099aa4f63a02a5621de0576e5aa71bc5194",
        "builtins": [
            "range_check",
            "poseidon"
        ]
    }
],
...
```

wojciechos:
Oh, sorry. I forgot to close the issue. This change was actually reverted in the latest spec: https://github.com/starkware-libs/starknet-specs/pull/275

wojciechos:
Close due to reverted spec: https://github.com/starkware-libs/starknet-specs/pull/275

PR #2456, opened 2025-02-11
fix: Display offset field as int for starknet_getCompiledCasm endpoint

hudem1:
Solves issue #2393


wojciechos:
Oh, sorry. I forgot to close the issue. This change was actually reverted in the latest rc-3 spec: https://github.com/starkware-libs/starknet-specs/pull/275Oh.

Issue #2391, opened 2025-01-20
RPC08 - INSUFFICIENT_RESOURCES_FOR_VALIDATE error


wojciechos:
In RPC 0.8, the [INSUFFICIENT_MAX_FEE ](https://github.com/NethermindEth/juno/blob/main/rpc/handlers.go#L58C2-L58C23) error is renamed to INSUFFICIENT_RESOURCES_FOR_VALIDATE with an updated message:
	•	Old: "Max fee is smaller than the minimal transaction cost (validation plus fee transfer)"
	•	New: "The transaction’s resources don’t cover validation or the minimal transaction fee."
See spec [PR](https://github.com/starkware-libs/starknet-specs/pull/217/files)

For RPC 0.7, keep the error as INSUFFICIENT_MAX_FEE.

Exca-DK:
Is this PR still up for grabs? And should it be tackled after https://github.com/NethermindEth/juno/pull/2419 ?

rodrigo-pino:
Hi @Exca-DK, yes it is!

PR #2457, opened 2025-02-11
fix: Use INSUFFICIENT_RESOURCES_FOR_VALIDATE instead of INSUFFICIENT_MAX_FEE in v8

hudem1:
Solves issue #2391

Note: In the rpccore.go file, now the 2 errors INSUFFICIENT_MAX_FEE and INSUFFICIENT_RESOURCES_FOR_VALIDATE have the same code (53), but they are never both used in the same pkg


clients/gateway/gateway.go
rodrigo-pino, line 1: Could we revert the indentation changes in this file

rpc/rpccore/rpccore.go
rodrigo-pino, line 1: Also in this file as well

hudem1: Done

clients/gateway/gateway.go
hudem1: ah yeah i had on-save formatting. Done!

rodrigo-pino, line 1: Uff now they aren't formatted, can you do `make format` and push

rpc/v8/transaction.go
```diff
@@ -679,8 +679,8 @@ func makeJSONErrorFromGatewayError(err error) *jsonrpc.Error {
 		return rpccore.ErrClassHashNotFound
 	case gateway.ClassAlreadyDeclared:
 		return rpccore.ErrClassAlreadyDeclared
-	case gateway.InsufficientMaxFee:
-		return rpccore.ErrInsufficientMaxFee
+	case gateway.InsufficientResourcesForValidate:
+		return rpccore.ErrInsufficientResourcesForValidate
```
rodrigo-pino, line 683: Is it possible to add a unit test for this?

clients/gateway/gateway.go
hudem1: the `make format` actually gives the same result as my format-on-save. So, `make format` creates again all the indentation you wanted to remove

rpc/v8/transaction.go
hudem1: Done

AnkushinDaniil (approved):

clients/gateway/gateway.go
rodrigo-pino: It was my fault, I didn't realize you had modified the file. It was correct the first.

rodrigo-pino (approved):

PR #2460, opened 2025-02-12
Fix storage proof RPC method to support head block

weiihann:
Accepts the request as long as the block ID matches the head block

### DIFF
--- a/rpc/v8/storage.go
+++ b/rpc/v8/storage.go
@@ -80,14 +80,15 @@ type StorageProofResult struct {
 func (h *Handler) StorageProof(id BlockID,
 	classes, contracts []felt.Felt, storageKeys []StorageKeys,
 ) (*StorageProofResult, *jsonrpc.Error) {
-	// We do not support historical storage proofs for now
-	if !id.Latest {
-		return nil, rpccore.ErrStorageProofNotSupported
-	}
-
 	head, err := h.bcReader.Head()
 	if err != nil {
 		return nil, rpccore.ErrInternal.CloneWithData(err)
+	}
+
+	// We do not support historical storage proofs for now
+	// Ensure that the block requested is the head block
+	if !headOnly(id, head) {
+		return nil, rpccore.ErrStorageProofNotSupported
 	}
 
 	state, closer, err := h.bcReader.HeadState()
@@ -181,6 +182,20 @@ func processStorageKeys(storageKeys []StorageKeys) ([]StorageKeys, *jsonrpc.Erro
 	}
 
 	return uniqueStorageKeys, nil
+}
+
+// Returns true only if the block ID requested matches the head block
+func headOnly(id BlockID, head *core.Block) bool {
+	switch {
+	case id.IsLatest():
+		return true
+	case id.IsPending():
+		return false
+	case id.GetHash() != nil:
+		return id.GetHash().Equal(head.Hash)
+	default:
+		return id.GetNumber() == head.Number
+	}
 }
 
 func getClassProof(tr *trie.Trie, classes []felt.Felt) ([]*HashToNode, error) {
--- a/rpc/v8/storage_test.go
+++ b/rpc/v8/storage_test.go
@@ -227,15 +227,30 @@ func TestStorageProof(t *testing.T) {
 		require.Equal(t, root, proof.GlobalRoots.ClassesTreeRoot)
 		require.Equal(t, root, proof.GlobalRoots.ContractsTreeRoot)
 	})
-	t.Run("error whenever not latest block is requested", func(t *testing.T) {
+	t.Run("error whenever block number is not the latest", func(t *testing.T) {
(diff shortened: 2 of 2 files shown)


rpc/v8/storage.go
```diff
@@ -80,16 +80,17 @@ type StorageProofResult struct {
...
-	}
-
 	head, err := h.bcReader.Head()
 	if err != nil {
 		return nil, rpccore.ErrInternal.CloneWithData(err)
 	}
 
+	// We do not support historical storage proofs for now
+	// Ensure that the block requested is the head block
+	if !id.IsLatest() && (id.GetNumber() != head.Number || (id.GetHash() != nil && !id.GetHash().Equal(head.Hash))) {
```
thiagodeev, line 90: In this condition, the block hash is only verified if the block number is equal to `head.Number`. If the 'blockNumber' is not equal to 'headNumber', it will satisfy the condition immediately with `id.GetNumber() != head.Number` thanks to the OR operator, and block hash will not be evaluated.
We need to verify that all of them are not true.

Sugestion: 
``` golang
if !id.IsLatest() && 
   id.GetNumber() != head.Number && 
   (id.GetHash() != nil && !id.GetHash().Equal(head.Hash)) {
    return nil, rpccore.ErrStorageProofNotSupported
}
```

rpc/v8/storage_test.go
thiagodeev, line 1: The previous bug (the original and the one from the "if" condition) wouldn't have happened if there was a test calling the `StorageProof` method with multiple inputs and asserting the correctness of the outputs for each input. Do you think this is viable?

thiagodeev (changes requested):
Thanks @weiihann!
There's a bug in the "if" condition

rpc/v8/storage_test.go
weiihann: I think it was initially intended that we reject all requests if `!id.Latest`, as we had tests on rejecting such request even if `req.blockNum == head.blockNum`. But as you mentioned we should accept `hash` and `number` if they match the head block, so the logic is modified here.

Regarding multiple inputs of the block id, only one field is unmarshalled, see the following:
```
func (b *BlockID) UnmarshalJSON(data []byte) error {
	if string(data) == `"latest"` {
		b.Latest = true
	} else if string(data) == `"pending"` {
		b.Pending = true
	} else {
		jsonObject := make(map[string]json.RawMessage)
		if err := json.Unmarshal(data, &jsonObject); err != nil {
			return err
		}
		hash, ok := jsonObject["block_hash"]
		if ok {
			b.Hash = new(felt.Felt)
			return json.Unmarshal(hash, b.Hash)
		}

		number, ok := jsonObject["block_number"]
		if ok {
			return json.Unmarshal(number, &b.Number)
		}

		return errors.New("cannot unmarshal block id")
	}
	return nil
}
```

rpc/v8/storage.go
weiihann: I don't think that's right either, please see the latest changes

rpc/v8/storage_test.go
thiagodeev: Sorry, I mean having test cases with multiple calls for `StorageProof` to test the method's behavior in each scenario.
Here's an example:

``` golang
testcases := []testcase{
  { // block id = latest
   ID: BlockID{Latest: true}
   ExpectedResp: resp1,
   ExpectedErr: nil,
  },
  { // block id = old number
   ID: BlockID{Number: 566}
   ExpectedResp: resp2,
   ExpectedErr: err1,
  },
  { // block id = old hash
   ID: BlockID{Hash: "0xab67878fc878d87e876f8a876b7887c6876d876a86b687"}
   ExpectedResp: resp3,
   ExpectedErr: err2,
  },
}

for _, test := range testSet {
  resp, err := h.StorageProof(test.ID, nil, nil, nil)
  if err != nil {
    require.EqualErrors(t, test.ExpectedErr, err)
  }
  assert.NoErr(t, err)
  assert.Equal(t, test.ExpectedResp, resp)
}
```

thiagodeev (approved):
Dumb approve just to no-blocking it. I'll need to be OOO for a time.

Still need review

rpc/v8/storage_test.go
weiihann: Ah, I've added them already

rodrigo-pino (approved):
LGTM!

rianhughes:
Should we add a test for the "latest" case,

```
t.Run("no error when block hash matches head", func(t *testing.T) {
		proof, rpcErr := handler.StorageProof(rpc.BlockID{Latest: true}, nil, nil, nil)
		require.Nil(t, rpcErr)
		require.NotNil(t, proof)
	})
```

weiihann:
> Should we add a test for the "latest" case?,

I think it's not needed since the other test cases already cover that scenario

thiagodeev:
LGTM

PR #2462, opened 2025-02-12
fix: Remove pending blocks from event subscription

infrmtcs:
`starknet_subscribeEvents` shouldn't return events from pending blocks.


rodrigo-pino (approved):
Looks good. The test name `""Events from pending block without duplicates"`, should be modified as well to describe the new thing the test is doing.

infrmtcs:
Superseded by #2510

PR #2464, opened 2025-02-12
fix: Fix starknet_getStorageAt handler in edge cases

hudem1:
The specs for get_StorageAt is the same for v6, v7 and v8: https://github.com/starkware-libs/starknet-specs/blob/v0.6.0/api/starknet_api_openrpc.json#L136

The possible errors are:
- ErrBlockNotFound
- ErrContractNotFound

## Modifications

1. In our v6, when the contract did not exist, it returned a zero-valued field (as if the contract existed but not the key). I added a check for contract existence.

2. In our v6, v7, v8, when an error is returned when trying to retrieve the value of a key (call to `ContractStorage`), we used to return `ErrContractNotFound`. But this check is now done before (thanks to point 1), and this call can actually never return `ErrContractNotFound` error, so, I changed to return an `ErrInternal` instead.

## Question

This bug was not seen before because in the tests, we can "cheat" by mocking calls to return whatever value we want. Shouldn't we use something like `NewMemTest` ? I've seen this used elsewhere, and I believe using this avoids mocking calls and subsequently allows to get the actual returned value that will be returned in practice too. But this would require setting up tests a bit more I guess (for example, adding a contract with a storage key to test key existence than just mocking a call, etc.).  What do you think ?

### DIFF
--- a/rpc/v6/contract.go
+++ b/rpc/v6/contract.go
@@ -1,7 +1,10 @@
 package rpcv6
 
 import (
+	"errors"
+
 	"github.com/NethermindEth/juno/core/felt"
+	"github.com/NethermindEth/juno/db"
 	"github.com/NethermindEth/juno/jsonrpc"
 	rpccore "github.com/NethermindEth/juno/rpc/rpccore"
 )
@@ -29,7 +32,7 @@ func (h *Handler) Nonce(id BlockID, address felt.Felt) (*felt.Felt, *jsonrpc.Err
 	return nonce, nil
 }
 
-// StorageAt gets the value of the storage at the given address and key.
+// StorageAt gets the value of the storage at the given address and key for a given block.
 //
 // It follows the specification defined here:
 // https://github.com/starkware-libs/starknet-specs/blob/a789ccc3432c57777beceaa53a34a7ae2f25fda0/api/starknet_api_openrpc.json#L110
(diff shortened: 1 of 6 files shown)


rpc/v6/contract.go
```diff
@@ -40,9 +43,22 @@ func (h *Handler) StorageAt(address, key felt.Felt, id BlockID) (*felt.Felt, *js
 	}
 	defer h.callAndLogErr(stateCloser, "Error closing state reader in getStorageAt")
 
+	// Check if the contract exists because if it doesn't, the call to `ContractStorage` below
+	// will just return a zero-valued felt with a nil error (not an `ErrContractNotFound` error)
+	_, err := stateReader.ContractClassHash(&address)
```
rodrigo-pino, line 48: Here you are checking for the existence of a contract by retrieving the whole class hash. Think that class hash can be several mbs of Sierra Code. I believe this might hurt performance a bit (and memory). Is no cheaper way of doing this first check?

hudem1: The classhash is a felt, so, doesnt it retrieve only about the size of [4]uint64 ? Maybe there is something I am not understanding/seeing ? If so, I can definitely change to get nonce if it's better instead (it returns a ErrKeyNotFound error if the contract is not found)

rodrigo-pino: Ah, I see you are correct. I misunderstood when I first read

rpc/v7/storage_test.go
```diff
@@ -134,11 +134,21 @@ func TestStorageAt(t *testing.T) {
 	t.Run("non-existent key", func(t *testing.T) {
 		mockReader.EXPECT().HeadState().Return(mockState, nopCloser, nil)
 		mockState.EXPECT().ContractClassHash(&felt.Zero).Return(nil, nil)
-		mockState.EXPECT().ContractStorage(gomock.Any(), gomock.Any()).Return(nil, db.ErrKeyNotFound)
+		mockState.EXPECT().ContractStorage(gomock.Any(), gomock.Any()).Return(&felt.Zero, nil)
+
+		storage, rpcErr := handler.StorageAt(felt.Zero, felt.Zero, rpcv7.BlockID{Latest: true})
```
rianhughes, line 139: nit: rename as storageValue, or sVal or something

rianhughes (approved):

rodrigo-pino (approved):

PR #2469, opened 2025-02-13
Refactor: Reuse v6 starknet_getEvents handler for v7 and v8

hudem1:
RPC pkg cleanup according to issue #2437

Reuse v6 handler for both v7 and v8


rodrigo-pino (approved):
LGTM

AnkushinDaniil (approved):

PR #2470, opened 2025-02-13
Binary search fee estimation

AnkushinDaniil:
It may be easier to review separate commits


Cargo.toml
rodrigo-pino, line 1: This doesn't seem like a good place to have this file. Why create a rust's Cargo here 👀 ?

rodrigo-pino (commented):
Perhaps I am missing something but I am not sure why there are these changes in the rust code. 

For me the only change required with the new estimate fee was to call `vm call` several times with different amount of gas until you find the closest approximation.

Cargo.toml
AnkushinDaniil: I decided to have the same version of the dependencies in all the rust code we have. Earlier update of some library required to update it everywhere

AnkushinDaniil:
> Perhaps I am missing something but I am not sure why there are these changes in the rust code.
> 
> For me the only change required with the new estimate fee was to call vm call several times with different amount of gas until you find the closest approximation.

I made these changes in rust because there are a lot of structs and functions that are convenient to use, also we would lose performance if doing marshalling and unmarshalling through cgo on every iteration

Cargo.toml
rodrigo-pino: Ok, I understand your reason and I think it is a valid one. Let's not do the changes for now and discuss how to do it. I don't think Juno should have Cargo files in its roots. At least for now.

rodrigo-pino: Beyond that I agree with all your refactoring made inside the rust code

rodrigo-pino:
The refactoring is making it hard to review. Let's leave it as is, but for next PRs please let's just have them do one thing at a time 🙏. Either introduce a feature xor refactor xor bug fix xor etc...

AnkushinDaniil:
> Also, not big deal, and it is fine as it is right now, but let's try to avoid it for the future.

Got it! I really thought that reviewing each commit and the refactored code would help to understand the changes. I'll definitely make separate PRs next time.

AnkushinDaniil:
PR without refactoring: #2473

rodrigo-pino:
Closing in favor of #2473

PR #2471, opened 2025-02-13
Refactor: Reuse rpcv6 juno_version handler for v7 and v8

hudem1:

All 3 rpc versions' `juno_version` endpoint reference the same handler method in the "master" handler


rianhughes (approved):
Importing the "errors" pkg should fix the tests and lint errors. Approving now to unblock this PR

rpc/handlers.go
```diff
@@ -197,7 +197,7 @@ func (h *Handler) MethodsV0_8() ([]jsonrpc.Method, string) { //nolint: funlen
 		},
 		{
 			Name:    "juno_version",
-			Handler: h.rpcv8Handler.Version,
+			Handler: h.rpcv6Handler.Version,
```
rodrigo-pino, line 200: The `Version` method shouldn't be part of any rpc. It should be a property of the "Master Handler" defined in `rpc/handlers.go` and it should have that method. No need to have the info replicated in three different structs.

Note that this is an exception since "juno_version" is not part of any specs. Also, every rpc should still expose the "juno_version" rpc, but they should all point to the one defined in the "Master Handler".

hudem1: Done!

rodrigo-pino:
Closing this because it became stale and further changes will affect it

PR #2483, opened 2025-02-14
add validation for rpcv8.Transaction Resource bounds

rianhughes:
resources bounds must have all fields present for rpc v8 calls ([spec](https://github.com/starkware-libs/starknet-specs/blob/436e6307bab3938339e4bf469f14dee79b50802d/api/starknet_api_openrpc.json#L3503))


AnkushinDaniil (approved):

validator/validator.go
```diff
@@ -16,10 +17,28 @@ var (
...
+	return true
+}
+
 // Validator returns a singleton that can be used to validate various objects
 func Validator() *validator.Validate {
 	once.Do(func() {
 		v = validator.New()
+
+		if err := v.RegisterValidation("resource_bounds_required", validateResourceBounds); err != nil {
+			log.Fatalf("failed to register validation: %v", err)
```
AnkushinDaniil, line 39: I don't think we need to use 'log' here
```suggestion
			panic("failed to register validation: " + err.Error())
```

validator/validator.go
```diff
@@ -16,10 +17,28 @@ var (
 	v    *validator.Validate
 )
 
+func validateResourceBounds(fl validator.FieldLevel) bool {
+	req, ok := fl.Parent().Interface().(rpcv8.Transaction)
+	if !ok {
+		return false
+	}
+
+	if (req.Version.String() == "0x3" || req.Version.String() == "0x100000000000000000000000000000003") &&
```
AnkushinDaniil, line 26: I think we can avoid calling the method multiple times
```suggestion
	version := req.Version.String()
	if (version == "0x3" || version == "0x100000000000000000000000000000003") &&
```

AnkushinDaniil (approved):

rodrigo-pino (approved):

validator/validator.go
rodrigo-pino: Why do you think `panic` is better than `log.Fatalf` in this case? Just curious. cc @AnkushinDaniil

AnkushinDaniil: I don't see a big difference between `panic` and `log.Fatalf` here, but we use `panic` in this package for all other cases that we don't expect, so it seems that we can make the code simpler by using `panic`

rpc/v8/transaction_test.go
```diff
@@ -1403,3 +1404,107 @@ func TestResourceUnmarshalJSON(t *testing.T) {
...
+		"resource_bounds": {
+			"l2_gas": {
+				"max_amount": "0x0",
+				"max_price_per_unit": "0x0"
+			},
+			"l1_gas": {
+				"max_amount": "0x0",
+				"max_price_per_unit": "0x0"
+			},
+			"l1_data_gas": {
```
ninja342, line 1466: l1_data_gas is not in the spec.

PR #2486, opened 2025-02-14
fix: Distinguish between ErrTxnHashNotFound and ErrInternal

hudem1:
rpc pkg cleanup according to issue #2437

- **for v6:** I think it could be nice to add this condition to distinguish between ErrTxnHashNotFound and ErrInternal.

- **for v7:** We cannot refactor as the logic is different from v6

- **for v8:** The logic + data structures are the same with v7. So theoretically, we could refactor by reusing v7's handler. But some v8 tests use this v8 handler to compare and test other handlers. For example, it is used to test `TestBlockWithTxs` by comparing the returned txs. So, we cannot really remove the handler.

### DIFF
--- a/rpc/v6/transaction.go
+++ b/rpc/v6/transaction.go
@@ -11,6 +11,7 @@ import (
 	"github.com/NethermindEth/juno/clients/gateway"
 	"github.com/NethermindEth/juno/core"
 	"github.com/NethermindEth/juno/core/felt"
+	"github.com/NethermindEth/juno/db"
 	"github.com/NethermindEth/juno/jsonrpc"
 	rpccore "github.com/NethermindEth/juno/rpc/rpccore"
 	"github.com/NethermindEth/juno/starknet"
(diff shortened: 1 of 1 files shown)


AnkushinDaniil (commented):
We should also be handling the 'pending' block in 'v6' in the same way. As far as I know, this difference happened because the old code for 'v6' was copied in #2419 . You can use the same code from 'v7' and continue refactoring.

hudem1:
> We should also be handling the `pending` block in `v6` in the same way. As far as I know, this difference happened because the old code for `v6` was copied in #2419 . You can use the same code from `v7` and continue refactoring.

Ah yes indeed, good catch! I updated v6 code, and I could then reuse v6 handler for v7.

AnkushinDaniil (approved):

rpc/v7/transaction_test.go
```diff
@@ -476,21 +169,76 @@ func TestTransactionByBlockIdAndIndex(t *testing.T) {
...
+		txn2, rpcErr := rpcv6Handler.TransactionByHash(*latestBlock.Transactions[index].Hash())
 		require.Nil(t, rpcErr)
 
-		assert.Equal(t, txn1, txn2)
+		assert.Equal(t, txn1, AdaptV6TxToV7(t, txn2))
 	})
 }
 
+// Convert a v6 transaction object to a v7 transaction object
+func AdaptV6TxToV7(t *testing.T, tx *rpcv6.Transaction) *rpc.Transaction {
```
rianhughes, line 193: I think we can un-export this

rianhughes (changes requested):
Just need to un-export `AdaptV6TxToV7` and then it looks good to me

rpc/v7/transaction_test.go
hudem1: Done!

rianhughes (approved):

rpc/v7/transaction_test.go
Omor12-2: Transfer coins Ethereum

PR #2491, opened 2025-02-15
Refactor: Reuse starknet_call v7 handler for v8

hudem1:

- v7: cannot refactor
- v8: reuse v7 handler


rpc/v7/trace.go
```diff
@@ -404,7 +407,7 @@ func (h *Handler) Call(funcCall FunctionCall, id BlockID) ([]*felt.Felt, *jsonrp
 
 	declaredClass, err := state.Class(classHash)
 	if err != nil {
-		return nil, rpccore.ErrClassHashNotFound
+		return nil, rpccore.ErrInternal.CloneWithData(err)
```
rodrigo-pino, line 410: The reason for this internal error is that if the class hash is not found, it means that state is not working correctly?

rodrigo-pino (approved):

rpc/v7/trace.go
hudem1: I changed to internal error because it is not one of the 3 possible errors [the specs](https://github.com/starkware-libs/starknet-specs/blob/v0.7.1/api/starknet_api_openrpc.json#L599-L609) state, which are:
 - CONTRACT_NOT_FOUND
- CONTRACT_ERROR
- BLOCK_NOT_FOUND

AnkushinDaniil (approved):

rpc/v7/trace.go
rodrigo-pino: Could you check if it is possible for a declared class not exist after after using the `classHash` specified here

hudem1: My message was not very clear; i changed to an internal error because the CONTRACT_NOT_FOUND error was checked before when retrieving the classhash for a contract address. So, I figured if the call to `state.Class(classHash)` failed, it would mean an internal error because logically, the class should exist in the DB if we an address is associated to its classhash.

Now from what I saw in the code, a pair (address, classhash) is stored in DB when:
- a user-contract is deployed --> in that case, there is indeed an additional chunk of code registering classes that are present in stateDiff.deployedContracts (as well as stateDiff.declaredClasses)
- a system-contract is deployed --> but this seems to be a bit special as the only possible system-contracts are the following:
```var (
	systemContractsClassHash = new(felt.Felt).SetUint64(0)

	systemContracts = map[felt.Felt]struct{}{
		*new(felt.Felt).SetUint64(1): {},
		*new(felt.Felt).SetUint64(2): {},
	}
)
```
By the way, what exactly are system contracts ? From what i've read, they are deployed at the protocol level and provide low-level functionalities essential for the blockchain’s operation.
- a contract replaces the class it instantiates (stateDiff.ReplacedClasses) --> in that case, the pair (address, classHash) is updated but there doesn't seem to be a check for the associated DeclaredClass existence. We receive that info from the feeder, and update the pair without checking that the associated DeclaredClass does exist. We also replace the classhash for an address when there is a block revert. I don't see a check for DeclaredClass existence there either.

Maybe we could add this check here to make sure the DeclaredClass exists when updating the pair (address, clashash) ?

hudem1: @rodrigo-pino, you left a comment after approving, so, I just wanna make sure it's okay to merge this PR :)

Also, what exactly are system contracts ? And is it normal there doesn't seem to be a check for class existence when replacing the classhash a contract refers to ?

PR #2496, opened 2025-02-16
Refactor: starknet_estimateMessageFee

hudem1:

Reuse v6 `SimulationFlag` and `MsgFromL1` types for v7 and v8

As i was looking into refactoring `starknet_estimateMessageFee` endpoint, i just found those 2 types to refactor. I don't know if we really want to refactor those, but i opened a PR in case. Let me know if we don't necessarily want to refactor those


AnkushinDaniil (approved):

rpc/v7/estimate_fee.go
```diff
@@ -65,9 +66,9 @@ func (f FeeEstimate) MarshalJSON() ([]byte, error) {
 *****************************************************/
 
 func (h *Handler) EstimateFee(broadcastedTxns []BroadcastedTransaction,
-	simulationFlags []SimulationFlag, id BlockID,
+	simulationFlags []rpcv6.SimulationFlag, id BlockID,
```
rianhughes, line 69: Hmm, it looks like this is a bug actually. The spec distinguishes between SIMULATION_FLAG_FOR_ESTIMATE_FEE (starknet_estFee) and SIMULATION_FLAG (starknet_sumTxns). This seems to be the case for all rpc versions. Not a critical bug, but should probably be fixed

rpc/v8/estimate_fee.go
```diff
@@ -45,9 +46,9 @@ type FeeEstimate struct {
 *****************************************************/
 
 func (h *Handler) EstimateFee(broadcastedTxns []BroadcastedTransaction,
-	simulationFlags []SimulationFlag, id BlockID,
+	simulationFlags []rpcv6.SimulationFlag, id BlockID,
```
rianhughes, line 49: Similar bug here (we accept SKIP_FEE_CHARGE, which isn't allowed by the spec)

rianhughes:
LGTM, just need to address the starknet_estFee flags bug

rpc/v7/estimate_fee.go
hudem1: Yes true!

As i'm making the changes, i'm seeing that we are explicitly adding the `SkipFreeChargeFlag` to the simulationFlags before calling simulateTransactions in EstimateFee handler, [here](https://github.com/NethermindEth/juno/blob/main/rpc/v6/estimate_fee.go#L45).

It makes me think that we knew the flag wasn't gonna be added by the user, so we added it explicitly. I agree the flag is not part of the specs for the estimateFee, but I'm wondering why then we explicitly added it. 

Also, I see the specs mention this (for simulating tx, not estimate fee): `By default, the sequencer behavior is replicated locally (enough funds are expected to be in the account, and fee will be deducted from the balance before the simulation of the next transaction)`  What do they mean by "before the simulation of the next tx" ? why isn't the fee deducted from the user's balance right away?

rianhughes: Ah right, sorry I was being stupid. Passing in `SkipFreeChargeFlag` by default makes sense, the whole purpose of this endpoint is to determine the fee, so it would be bad ux for the request to fail because the tx fee was set too low.

> What do they mean by before the simulation of the next tx ? why isn't the fee deducted from the user's balance right away?

I think it is deducted from the users account right away. If you look into the blockifier you will effectively see something like:```

for txn in txns_to_execute{
   execute(txn)
   take_fee_tokens_from_user(txn)
}
```

rianhughes (approved):

PR #2498, opened 2025-02-17
bump veresions

rianhughes:
Workflow builds are failing in

### DIFF
--- a/vm/rust/Cargo.toml
+++ b/vm/rust/Cargo.toml
@@ -9,8 +9,8 @@ edition = "2021"
 serde = "1.0.208"
 serde_json = { version = "1.0.125", features = ["raw_value"] }
 blockifier = "=0.14.0-rc.0"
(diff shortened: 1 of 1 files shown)


AnkushinDaniil (commented):
I think we need to update all the rust code we have. We have `cairo-lang-starknet-classes` in `juno/starknet...`. 
I also think we need to consider creating a common `Cargo.toml` in the root as I did in #2470. @rodrigo-pino

rodrigo-pino:
> I also think we need to consider creating a common Cargo.toml

I find it creates more clutter. If we had several rust crates it would make sense to me, but they are only two as far as I know and they get updated once every few months.

If it is truly a problem then we can create a rust folder with all the rust implementation there. I think is better how it is right now though.

rianhughes:
Sorry, this was a test PR, wasn't meant to be reviewed. This fix I was after is here https://github.com/NethermindEth/juno/pull/2500

PR #2502, opened 2025-02-17
fix: Cyclic dependencies when setting up pending subscription

infrmtcs:
Synchronizer and blockchain reader are having cyclic dependencies, and not initialized correctly, specifically `pendingBlockFn` are not configured.


rodrigo-pino (approved):
Created issue #2511 to make as a follow up to this PR

PR #2503, opened 2025-02-17
refactor: Use pointer for optional block number in FilteredEvent

infrmtcs:
Pending transaction events are expected not to have block number, hence `EmittedEvent` should have nil block number in that case, which requires `FilteredEvent` to accept nil block number too.


rodrigo-pino (approved):

rianhughes (approved):
Looks good, although might be worth adding a small test

PR #2509, opened 2025-02-17
Refactor: Reuse v6 juno_unsubscribe handler for v7 and v8 (starknet_unsubscribe)

hudem1:

As side notes:
- reuse v6 (custom) juno_unsubscribe for v7 as well as for v8 (starknet_unsubscribe)
- i had to rename v6 `events_test.go`'s package name to `rpcv6` instead of `rpcv6_test` in order to access `handler.subscriptions` in tests (as it was done in v8 `subscriptions_test.go`).


rpc/v8/subscriptions_test.go
```diff
@@ -882,107 +882,6 @@ func TestSubscribePendingTxs(t *testing.T) {
 	})
 }
 
-func TestUnsubscribe(t *testing.T) {
```
infrmtcs, line 885: These tests actually server the `starknet_unsubscribe`  instead of `juno_unsubscribe`. I think we'll eventually deprecate `juno_unsubscribe` because it's not currently in v8, so we shouldn't move this test.

PR #2510, opened 2025-02-17
fix: Emit pending transaction events

infrmtcs:
Depends on #2502 and #2503


rpc/v8/subscriptions.go
```diff
@@ -159,20 +166,24 @@ func (h *Handler) SubscribeEvents(ctx context.Context, fromAddr *felt.Felt, keys
...
 							}
-
-							delete(eventsPreviouslySent, fe)
+							delete(eventsPreviouslySent, sentEvent)
 						}
 					}
-				case pending := <-pendingSub.Recv():
-					h.processEvents(subscriptionCtx, w, id, pending.Number, pending.Number, fromAddr, keys, eventsPreviouslySent)
+				case <-pendingSub.Recv():
+					height, err := h.bcReader.Height()
```
infrmtcs, line 179: I think this can cause race condition, when the height increases after this call

weiihann (approved):
LGTM

rpc/v8/subscriptions.go
rodrigo-pino: It can, it looks like that for me. One idea that comes to mind is that we can send the block directly to process event, instead of doing to separate requests 🤔.

rodrigo-pino (approved):

rpc/v8/subscriptions.go
rodrigo-pino: Resolving and creating #2516

PR #2514, opened 2025-02-18
Fix simulate transaction

AnkushinDaniil:
This PR fixes an issue that arose after #2473, transaction simulation fails due to `charge_fee` and `validate` flags


rodrigo-pino (commented):
This bug fix doesn't make too much sense to me. What's the reason for this changes?

AnkushinDaniil:
> This bug fix doesn't make too much sense to me. What's the reason for this changes?

This fix should fix the `simulate_transaction` endpoint and e2e tests.

rodrigo-pino:
Solved in #2517

PR #2517, opened 2025-02-19
Rianhughes/fix sim txn

rianhughes:
This PR introduces some fixes for failing simulateTransaction and traceTransaction calls.

1. We derive the sierra Version directly from the sierra code.
2. When running the Binary search, we skip some validation checks. Otherwise blockifier requires the user to have enough l2 gas to pay the `MAX fee` amount. Which shouldn't be required when performing Binary search. We then execute on the original transaction after the Binary Search is complete.


rianhughes:
CICD https://github.com/NethermindEth/juno/actions/runs/13415134619

rianhughes:
CICD https://github.com/NethermindEth/juno/actions/runs/13415893437

vm/class.go
```diff
@@ -37,9 +40,47 @@ func marshalClassInfo(class core.Class) (json.RawMessage, error) {
...
 }
+
+// Parse Sierra version from the JSON representation of the program.
+//
+// Sierra programs contain the version number in two possible formats.
+// For pre-1.0-rc0 Cairo versions the program contains the Sierra version
+// "0.1.0" as a shortstring in its first Felt (0x302e312e30 = "0.1.0").
+// For all subsequent versions the version number is the first three felts
+// representing the three parts of a semantic version number.
+func parseSierraVersion(prog []*felt.Felt) (string, error) {
```
AnkushinDaniil, line 61:
```suggestion
// TODO: There should be an implementation in the blockifier. If there is, move it to the rust part.
func parseSierraVersion(prog []*felt.Felt) (string, error) {
```

vm/rust/src/execution.rs
```diff
@@ -106,14 +107,23 @@ fn get_gas_vector_computation_mode<S>(
 where
     S: UpdatableState,
 {
-    let original_tx = transaction.clone();
     let initial_gas_limit = extract_l2_gas_limit(transaction);
+    let mut original_transaction = transaction.clone();
 
     // Simulate transaction execution with maximum possible gas to get actual gas usage.
     set_l2_gas_limit(transaction, GasAmount::MAX);
+    match transaction {
```
AnkushinDaniil, line 115:
```suggestion
// TODO: Consider getting the upper bound from the balance and not changing the execution flags
    match transaction {
```

vm/rust/src/execution.rs
```diff
@@ -179,17 +189,25 @@ where
...
             Err(SimulationError::ExecutionError(error)) => return Err(error),
         };
+    tx_state.abort();
 
     // If the computed gas limit exceeds the initial limit, revert the transaction.
+    // The L2 gas limit is set to zero to prevent the transaction execution from succeeding
+    // in the case where the user defined gas limit is less than the required gas limit
     if l2_gas_limit > initial_gas_limit {
-        return original_tx.execute(state, block_context);
+        set_l2_gas_limit(&mut original_transaction, GasAmount(0));
```
AnkushinDaniil, line 198: Why can't we just run the original tx?
```suggestion

```

rianhughes: I don't think so. I think the original txn can still succeed, even though the user set `l2_gas` is below what we find in BD (`l2_gas_limit`). I think this is why PF explicitly set it to zero.

vm/class.go
rianhughes: +1.
My implementation was taken from PF's `parse_sierra_version()`

vm/rust/src/execution.rs
```diff
@@ -179,17 +189,25 @@ where
...
+        return original_transaction.execute(state, block_context);
     }
 
+    set_l2_gas_limit(&mut original_transaction, initial_gas_limit);
+    let mut simulated_state = CachedState::<_>::create_transactional(state);
```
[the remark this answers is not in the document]
rianhughes: I did try running this directly on the `state`, but `state` doesn't have a `commit` fn, so I just reverted to using the approach in `simulate_execution`. Tbh, maybe we don't need to call commit on state, and can use it directly. But I felt short on time so went with this approach which seems to work fine.

vm/class.go
```diff
@@ -37,9 +40,48 @@ func marshalClassInfo(class core.Class) (json.RawMessage, error) {
 		classInfo.Class = core2sn.AdaptCompiledClass(c.Compiled)
 		classInfo.AbiLength = uint32(len(c.Abi))
 		classInfo.SierraLength = uint32(len(c.Program))
-		classInfo.SierraVersion = c.Compiled.CompilerVersion
+		sierraVersion, err := parseSierraVersion(c.Program)
+		if err != nil {
+			return nil, err
+		}
+		classInfo.SierraVersion = sierraVersion
```
rodrigo-pino, line 47: 👍

vm/rust/src/execution.rs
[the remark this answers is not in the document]
AnkushinDaniil: We call commit in `cairoVMExecute`

vm/class.go
```diff
@@ -37,9 +40,48 @@ func marshalClassInfo(class core.Class) (json.RawMessage, error) {
...
+// "0.1.0" as a shortstring in its first Felt (0x302e312e30 = "0.1.0").
+// For all subsequent versions the version number is the first three felts
+// representing the three parts of a semantic version number.
+// TODO: There should be an implementation in the blockifier. If there is, move it to the rust part.
+func parseSierraVersion(prog []*felt.Felt) (string, error) {
+	if len(prog) == 0 {
+		return "", errors.New("failed to parse sierra version in classInfo")
+	}
+
+	pre01, err := new(felt.Felt).SetString("0x302e312e30")
```
rodrigo-pino, line 67: One of the small optimizations we did in the Cairo VM was to initialize the felts with values directly, for example, if we knew that the number was always going to be 32 we would initialize the felt like that
```go
const f = Felt { <values> }; // the values here are not straightforward since they use another math
```

I believe this is something we can do here. Parsing from string is usually very slow and also takes runtime memory.

Doesn't need to be solved as part of this PR if we can create an issue with this description

vm/rust/src/execution.rs
```diff
@@ -106,14 +107,24 @@ fn get_gas_vector_computation_mode<S>(
 where
     S: UpdatableState,
 {
-    let original_tx = transaction.clone();
     let initial_gas_limit = extract_l2_gas_limit(transaction);
+    let mut original_transaction = transaction.clone();
 
     // Simulate transaction execution with maximum possible gas to get actual gas usage.
     set_l2_gas_limit(transaction, GasAmount::MAX);
+// TODO: Consider getting the upper bound from the balance and not changing the execution flags
```
rodrigo-pino, line 115: An issue for this one please 🙏

vm/rust/src/execution.rs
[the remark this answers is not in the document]
rodrigo-pino: Issue for this as well 🙏

rodrigo-pino (approved):

rianhughes:
Follow up issues https://github.com/NethermindEth/juno/issues/2518

vm/class.go
rianhughes: https://github.com/NethermindEth/juno/issues/2520

AnkushinDaniil (approved):

PR #2522, opened 2025-02-19
Remove cashed state

### LINKED ISSUE #2518: Can we get rid of cached state in final binary search executiuon?
Follow up issue for https://github.com/NethermindEth/juno/pull/2517

Can we get rid of the cached state and execute direction on the state??

See here https://github.com/NethermindEth/juno/pull/2517#discussion_r1961924721

AnkushinDaniil:
Fixes #2518 

Waiting for testing on ci/cd...


hudem1 (approved):

rodrigo-pino (commented):
I am unsure, there is not test code that proves anything broke before or after. What is this fixing?

AnkushinDaniil:
> I am unsure, there is not test code that proves anything broke before or after. What is this fixing?

It fixes #2518. It may be cleaner to review by commits. The first commit is just formatting, we don't have a format checker for rust in ci/cd. The second commit just removes the extra layer of state.

PR #2525, opened 2025-02-20
Optimize sierra version

### LINKED ISSUE #2520: optimise felt call when getting sierra version
https://github.com/NethermindEth/juno/pull/2517#discussion_r1961917407

Exca-DK:
Closes #2520 

```      
$ benchstat MASTER PR
                             │    MASTER     │                 PR                 │
                             │    sec/op     │   sec/op     vs base               │
SierraVersion/pre_0.1.0-32     138.200n ± 3%   2.083n ± 1%  -98.49% (p=0.002 n=6)
SierraVersion/after_0.1.0-32    224.55n ± 1%   53.22n ± 1%  -76.30% (p=0.002 n=6)
geomean                          176.2n        10.53n       -94.02%
```


AnkushinDaniil (approved):

vm/class.go
```diff
@@ -64,11 +65,6 @@ func parseSierraVersion(prog []*felt.Felt) (string, error) {
 		return "", errors.New("failed to parse sierra version in classInfo")
 	}
 
-	pre01, err := new(felt.Felt).SetString("0x302e312e30")
-	if err != nil {
-		return "", err
-	}
-
```
rodrigo-pino, line 71: I would initialize the value here. There is no need to go through the SetString method.

vm/class.go
```diff
@@ -51,6 +50,8 @@ func marshalClassInfo(class core.Class) (json.RawMessage, error) {
 	return json.Marshal(classInfo)
 }
 
+var pre01, _ = new(felt.Felt).SetString("0x302e312e30")
```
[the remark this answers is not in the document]
Exca-DK: It is possible, updated the code as requested. Code is slightly slower:
```
Before:
BenchmarkSierraVersion/pre_0.1.0-32             536441556                2.278 ns/op           0 B/op          0 allocs/op
After:
BenchmarkSierraVersion/pre_0.1.0-32             100000000               10.58 ns/op            0 B/op          0 allocs/op
```

vm/class.go
AnkushinDaniil: Wouldn't it be better to initialise the value as a global?

rodrigo-pino: The other benchmarks are faster because the initialization are done before the `timer.Reset` in the benchmark test and the initialization is only done once.

The cost now comes in the form of memory which should be negligible as well. This would be a question of should we start hurting readability to get those extra 8 nano seconds. I think in this case we can live without them and save developer time in the future. They understand that variable is only needed where it is.

rodrigo-pino (approved):
Looks good, thank you @Exca-DK

AnkushinDaniil (approved):

PR #2530, opened 2025-02-20
Refactor flags, defaults and usage

AnkushinDaniil:
This pull request includes several changes to the `cmd/juno` package, primarily focused on refactoring the code to use a new `setting` struct for flag and default value management.

### DIFF
--- a/cmd/juno/dbcmd.go
+++ b/cmd/juno/dbcmd.go
@@ -39,7 +39,7 @@ func DBCmd(defaultDBPath string) *cobra.Command {
 		Long:  `This command allows you to perform database operations.`,
 	}
 
-	dbCmd.PersistentFlags().String(dbPathF, defaultDBPath, dbPathUsage)
+	dbCmd.PersistentFlags().String(dbPath.flag, defaultDBPath, dbPath.usage)
 	dbCmd.AddCommand(DBInfoCmd(), DBSizeCmd(), DBRevertCmd())
 	return dbCmd
 }
@@ -75,7 +75,7 @@ func DBRevertCmd() *cobra.Command {
 }
 
 func dbInfo(cmd *cobra.Command, args []string) error {
-	dbPath, err := cmd.Flags().GetString(dbPathF)
+	dbPath, err := cmd.Flags().GetString(dbPath.flag)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


weiihann:
From my understanding, this PR seems to be "cleaner" in the sense of eliminating repetitive flag name/usage string declarations. However, the generic usage seems to complicate things further.

I would suggest the following registry and group pattern instead:
```
type Flag struct {
    Name         string
    DefaultValue any
    Usage        string
    Required     bool
}

type FlagGroup struct {
    Name  string
    Flags []Flag
}

type Registry struct {
    groups     map[string]FlagGroup
    defaults   map[string]any
    validators map[string]func(any) error
}

type ConfigHelper struct {
    registry *Registry
    cmd      *cobra.Command
}

func (c *ConfigHelper) ConfigureCommand(cmd *cobra.Command) error {
    c.cmd = cmd
    
    for _, group := range c.registry.groups {
        for _, flag := range group.Flags {
            switch v := flag.DefaultValue.(type) {
            case string:
                cmd.Flags().String(flag.Name, v, flag.Usage)
            // ... handle other types
            }
            
            if flag.Required {
                cmd.MarkFlagRequired(flag.Name)
            }
        }
    }
    
    return nil
}

// Define flag groups for better organization
var (
    NetworkFlags = FlagGroup{
        Name: "Network Configuration",
        Flags: []Flag{
            {
                Name:         "network",
                DefaultValue: utils.Mainnet,
                Usage:        "Options: mainnet, sepolia, sepolia-integration",
            },
            {
                Name:         "cn-name",
                DefaultValue: "",
                Usage:        "Custom network name",
            },
            // ... other network related flags
        },
    }
    ... // more groups
)

func NewCmd(config *node.Config, run func(*cobra.Command, []string) error) *cobra.Command {
    registry := flags.NewRegistry()
    
    // Register all flag groups
    registry.RegisterGroup(flags.NetworkFlags)
    registry.RegisterGroup(flags.HTTPFlags)
    // ... register other groups
    
    // Add flag validators
    registry.AddValidator("network", validateNetwork)
    // ... add other validators
    
    ...add other things 
}
```

To complete this is much more work, but grouping flags keep it organized and we can expand our CLI tools easily.

rodrigo-pino:
Closing because using global vars doesn't seem the right approach to me just for the CLI. They take runtime memory

PR #2532, opened 2025-02-20
refactor: Reuse v7 handlers for v8

infrmtcs:


infrmtcs:
This was only meant to be example of how to refactor, and we decided to separate the code to avoid dealing with this kind of issue, so close the PR.

PR #2533, opened 2025-02-20
fix: Subscription reorg test is missing mock expect

infrmtcs:


weiihann (approved):

rodrigo-pino:
Why was it working before?

infrmtcs:
This is a fix for the flaky test behaviours before the recent changes in subscription. The reasons were:
- The test is missing mock calls for some objects.
- It's supposed to fail and report to test runner, however gomock try to convert a nil pointer to string and panic instead.
- Because of race condition, most of the time the panic was not caught and ignored. I reproduced the behaviours by adding some random sleep.

rodrigo-pino (approved):

PR #2534, opened 2025-02-21
Refactor: Reuse v6 starknet_add-Invoke/DeployAccount/Declare-Transaction handler for v7

hudem1:
rpc pkg cleanup to #2437 

- Fix a little missing validation field for the `SenderAddress` field inside v6 `Transaction` type.

- Reuse v6 handler for v7 for `starknet_addInvokeTransaction`, `starknet_addDeployAccountTransaction` and `starknet_addDeclareTransaction`


AnkushinDaniil (approved):

rianhughes:
Why can't we delete the v8 `AddTransaction` method? Is it because we will need to restrict the method to only allow v3 transactions?

hudem1:
> Why can't we delete the v8 `AddTransaction` method? Is it because we will need to restrict the method to only allow v3 transactions?

Yes this, and there are many other different validations b/w rpc v7 and v8 txs.
And also, v8 tx has the additional variant `ResourceL1DataGas` for `Resource` enum.

PR #2546, opened 2025-02-25
refactor: Separate matcher logic in EventFilter

infrmtcs:


weiihann:
Just double checking, there are no other changes to the matching logic besides them to `EventMatcher` right?

blockchain/event_matcher.go
```diff
@@ -0,0 +1,141 @@
...
+	contractAddress *felt.Felt
+	keysMap         []map[felt.Felt]struct{}
+}
+
+func NewEventMatcher(contractAddress *felt.Felt, keys [][]felt.Felt) *EventMatcher {
+	return &EventMatcher{
+		contractAddress: contractAddress,
+		keysMap:         makeKeysMaps(keys),
+	}
+}
```
rodrigo-pino, line 21: I think it is safe to return the type by value here.

infrmtcs:
@weiihann Yeah, I only split it in order to be able to call the matcher functions outside of a stateful `EventFilter`. My approach is to combine with #2543 and rewrite the event processing to process the block directly received from subscription and avoid querying DB if possible.

blockchain/event_filter.go
```diff
@@ -26,13 +24,12 @@ type EventFilterer interface {
...
-	fromBlock       uint64
-	toBlock         uint64
-	contractAddress *felt.Felt
-	keys            [][]felt.Felt
-	maxScanned      uint // maximum number of scanned blocks in single call.
-	pendingBlockFn  func() *core.Block
+	txn            db.Transaction
+	fromBlock      uint64
+	toBlock        uint64
+	matcher        *EventMatcher
```
rodrigo-pino, line 30: I would also have `EventMatcher` here instead of `*EventMatcher` since it is not meant to be shared or be optional. Wdyt?

rodrigo-pino (approved):

blockchain/event_filter.go
infrmtcs: Agree, the fields are just a pointer and a slice anyway. Let me make a quick change.

weiihann (approved):

PR #2547, opened 2025-02-25
Add drop policy option to keep last event for Feed

infrmtcs:
This PR adds an option for feed subscription to drop earliest item instead of last item. Rationale behind this is:
- The feed subscription is required to be non-blocking, so that any subscription failure cannot affect the other ones.
- When there are multiple events sent before the subscription receives, the last events will be dropped. This cause 2 consequences:
  - The subscription is more like a signal that data has changed, because the subcriptions cannot rely on the event data itself for latest changes.
  - This is why subscriptions like new heads and pending, while receiving the block, still needs to query database for chain height in order to know the blocks it may missed.
- Although another solution is to use a bigger buffer instead of 1, but this may change the assumption about the cost of each subscription.


weiihann (approved):
LGTM, I'll appreciate it if you add comments explaining the difference between `Subscribe` and `SubscribeKeepLast`

rodrigo-pino (approved):

PR #2548, opened 2025-02-25
Ensure order of delivery for event subscription

infrmtcs:
Depends on #2547

### DIFF
--- a/rpc/v8/subscriptions.go
+++ b/rpc/v8/subscriptions.go
@@ -130,9 +130,9 @@ func (h *Handler) SubscribeEvents(ctx context.Context, fromAddr *felt.Felt, keys
 	}
 	h.subscriptions.Store(id, sub)
 
-	newHeadsSub := h.newHeads.Subscribe()
-	reorgSub := h.reorgs.Subscribe() // as per the spec, reorgs are also sent in the events subscription
-	pendingSub := h.pendingBlock.Subscribe()
+	newHeadsSub := h.newHeads.SubscribeKeepLast()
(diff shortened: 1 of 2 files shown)


rpc/v8/subscriptions.go
```diff
@@ -130,9 +130,9 @@ func (h *Handler) SubscribeEvents(ctx context.Context, fromAddr *felt.Felt, keys
 	}
 	h.subscriptions.Store(id, sub)
 
-	headerSub := h.newHeads.Subscribe()
-	reorgSub := h.reorgs.Subscribe() // as per the spec, reorgs are also sent in the events subscription
-	pendingSub := h.pendingBlock.Subscribe()
+	headerSub := h.newHeads.SubscribeKeepLast()
+	reorgSub := h.reorgs.SubscribeKeepLast() // as per the spec, reorgs are also sent in the events subscription
+	pendingSub := h.pendingBlock.SubscribeKeepLast()
```
weiihann, line 135: Some other subscription methods (i.e. `subscribeNewHeads`, `subscribeTransactionStatus`, `subscribePendingTxs`) also rely on these feeds. Do they need to use `SubscribeKeepLast` as well?

infrmtcs: I think it's likely yes. I plan to get it done for an endpoint (`starknet_subscribeEvents`) first, then convert it to a reusable form, then use it with other endpoints.

rodrigo-pino (approved):

weiihann (approved):

PR #2550, opened 2025-02-25
Refactor endpoints in juno namespace to a dedicated package

infrmtcs:


infrmtcs:
`juno_subscribeNewHeads` and `juno_unsubscribe` are superseded by `starknet_subscribeNewHeads` and `starknet_unsubscribe`, so they're going to be deprecated when v6 and v7 are deprecated. We decided to leave them as they're right now. The only remaining endpoint in `juno` namespace is `juno_version`, which is quite trivial, so it's not worth refactoring for now.

PR #2563, opened 2025-02-25
fix: ExecutionResources match specs

hudem1:
This PR is another (improved, cleaner) approach compared to [PR](https://github.com/NethermindEth/juno/pull/2497)

I created a `TransactionTrace` type specific to each pkg so that we can easily:
- match the specs (for `ExecutionResources` in this case)
- deprecate pkgs when needed

I created a file `adapters` within each rpc pkg when we need to adapt from/to a type that exists in the rpc pkg.
I thought it could nice to centralize the specific (v6/v7/v8) rpc adapters there, because i saw several adapters a bit spread all over the pkg (and this for each rpc pkg).

Also, there is some cleaning that can be done, such as in `vm.trace.go` and `vm.trace_test.go` where a lot of code can be removed now, or also moving existing rpc specific adapters to the new `adapters` file, but I thought it might be better to do that in another PR afterwards as this one is already pretty big.


rpc/v6/adapters.go
```diff
@@ -0,0 +1,299 @@
...
+	"github.com/NethermindEth/juno/starknet"
+	"github.com/NethermindEth/juno/utils"
+	"github.com/NethermindEth/juno/vm"
+)
+
+/****************************************************
+		VM Adapters
+*****************************************************/
+
+func AdaptVMTransactionTrace(trace *vm.TransactionTrace) *TransactionTrace {
```
rianhughes, line 16: Can we move this file to the adapters pkg? Similar for v7 and v8. Or is this what you meant by "moving existing rpc specific adapters to the new adapters file"?

rpc/v7/trace.go
```diff
@@ -24,115 +24,43 @@ var traceFallbackVersion = semver.MustParse("0.13.1")
...
-			trace.FunctionInvocation = fnInvocation
-		}
-
-		traces = append(traces, TracedBlockTransaction{
-			TransactionHash: &feederTrace.TransactionHash,
-			TraceRoot:       &trace,
-		})
-	}
-	return traces, nil
+type TransactionTrace struct {
```
rianhughes, line 27: This type seems largely similar to that in rpcv6, maybe we can re-use it?

rpc/v7/trace.go
```diff
@@ -24,115 +24,43 @@ var traceFallbackVersion = semver.MustParse("0.13.1")
...
-		ClassHash:          snFnInvocation.ClassHash,
-		EntryPointType:     snFnInvocation.EntryPointType,
-		CallType:           snFnInvocation.CallType,
-		Result:             snFnInvocation.Result,
-		Calls:              make([]vm.FunctionInvocation, 0, len(snFnInvocation.InternalCalls)),
-		Events:             make([]vm.OrderedEvent, 0, len(snFnInvocation.Events)),
-		Messages:           make([]vm.OrderedL2toL1Message, 0, len(snFnInvocation.Messages)),
-		ExecutionResources: adaptFeederExecutionResources(&snFnInvocation.ExecutionResources),
-	}
+type ExecuteInvocation struct {
```
rianhughes, line 38: This type seems largely similar to that in rpcv6, maybe we can re-use it?

rpc/v6/adapters.go
hudem1: Yeah definitely i can move it there!
I thought moving it there would mean having some adapter versioning in the adapter pkg, I wasnt sure we wanted this, but if thats fine, i will move those adapters file there!

(No, i meant that there are some other adapter functions (that were there before this PR) in each rpc pkg, and I was thinking of move those fcts into this new file as a little refactor/cleanup PR after this one)

rpc/v7/trace.go
hudem1: I think that's the whole goal of this PR :) We wanted each rpc pkg to have their own `TransactionTrace` types so that its much easier to personalize it for each rpc version (In this case, the `ExecutionResources` field), and that there as few dependencies inter-pkg as possible to make it easier when deprecating pkgs

rpc/v7/trace.go
hudem1: Ah yes for this one we could reuse v6 one indeed because the underlying `FunctionInvocation` field is the same as v6 !

rpc/v8/transaction.go
```diff
@@ -268,9 +268,13 @@ type ComputationResources struct {
...
-	L1Gas     uint64 `json:"l1_gas"`
+	InnerExecutionResources
 	L1DataGas uint64 `json:"l1_data_gas"`
-	L2Gas     uint64 `json:"l2_gas"`
+}
+
+type InnerExecutionResources struct {
+	L1Gas uint64 `json:"l1_gas"`
+	L2Gas uint64 `json:"l2_gas"`
 }
```
rianhughes, line 278: Why split this into distinct types?

rpc/v8/simulation.go
```diff
@@ -174,20 +175,21 @@ func createSimulatedTransactions(
 		l1DataGasPriceStrk = gasPrice.PriceInFri
 	}
 
-	simulatedTransactions := make([]SimulatedTransaction, len(overallFees))
+	simulatedTransactions := make([]SimulatedTransaction, 0, len(overallFees))
```
rianhughes, line 178: Why change this from
	simulatedTransactions := make([]SimulatedTransaction, len(overallFees))
    simulatedTransactions[i] = ...
to
	simulatedTransactions := make([]SimulatedTransaction, 0, len(overallFees))
    simulatedTransactions = append(simulatedTransactions, ..)

rpc/v7/helpers.go
```diff
@@ -179,3 +180,39 @@ func (h *Handler) stateByBlockID(id *BlockID) (core.StateReader, blockchain.Stat
...
+		t.ConstructorInvocation,
+		t.ValidateInvocation,
+		t.FeeTransferInvocation,
+		executeInvocation,
+		t.FunctionInvocation,
+	}, func(i *FunctionInvocation) bool { return i == nil })
+}
+
+func (t *TransactionTrace) TotalComputationResources() ComputationResources {
+	total := ComputationResources{}
```
rianhughes, line 200: Can we move these to where we define the TransactionTrace  please?

rpc/v6/adapters.go
rianhughes: Imo it's cleaner to have the adapters in the adapters pkg. I guess we'll need something like `vm2rpcv6`, `vm2rpcv7` and `vm2rpcv8`. And similarly for the starknet to rpc conversions.. which will mean we'll have like 6 new pkgs. What do you think @rodrigo-pino ?

rpc/v6/simulation.go
```diff
@@ -137,20 +136,10 @@ func (h *Handler) simulateTransactions(id BlockID, transactions []BroadcastedTra
...
-				new(felt.Felt).SetUint64(dataGasConsumed[i].L1DataGas), header.L1DAMode)
-			traces[i].ExecutionResources = &vm.ExecutionResources{
-				ComputationResources: trace.TotalComputationResources(),
-				DataAvailability:     &da,
-			}
-		}
-
-		result[i] = SimulatedTransaction{
-			TransactionTrace: &traces[i],
+		result = append(result, SimulatedTransaction{
```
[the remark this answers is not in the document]
rianhughes: Running this suggests that the "Preallocated" approach is much faster than the "Append" approach. Locally I get 3.65ns/operation vs 22.17ns/operation. The additional overhead seems to come from the append function 
```
func BenchmarkAppend(b *testing.B) {
	for n := 0; n < b.N; n++ {
		result := make([]rpcv7.SimulatedTransaction, 0, 10)
		for i := 0; i < 10; i++ {
			result = append(result, rpcv7.SimulatedTransaction{})
		}
	}
}

func BenchmarkPreallocated(b *testing.B) {
	for n := 0; n < b.N; n++ {
		result := make([]rpcv7.SimulatedTransaction, 10)
		for i := 0; i < 10; i++ {
			result[i] = rpcv7.SimulatedTransaction{}
		}
	}
}
```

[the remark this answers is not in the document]
hudem1: Oh very cool, didn't expect that !
Good idea, I will make some benchmarks next time instead of relying on my theoretical reasoning only!

I'll update the code then to use this approach! :)

rpc/v6/adapters.go
hudem1: (there would also be core2rpc x3)

rpc/v8/transaction.go
hudem1: Because in v8 the `ExecutionResources` in `FunctionInvocation` is different from the `ExecutionResources` in `TransactionTrace`. And the latter includes the 2 fields from the former + an additional (`L1DataGas`)

rpc/v8/simulation.go
hudem1: I thought it'd be more efficient as we discussed above, but your benchmarking proved me wrong, so, i'll put it back :)

rpc/v6/simulation.go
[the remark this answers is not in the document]
rodrigo-pino: `append` is assigning result[i] with extra steps so it is always going to be a bit more expensive. What can happen if the thing being added is too big that the operation cost of both increase such that the 18 ns are neglibile

rpc/v6/adapters.go
rodrigo-pino: After seeing the whole code I agree with @rianhughes. I think we should tackle it is a separate PR since this one is already high priority. 

I would organize it as `vm2rpc`, inside I would then have three pkgs `v6`, `v7` and `v8`.

rodrigo-pino (commented):
I signaled a couple of things for code in `v6`. I believe this thing should be re-applied to `v7` and `v8`.

Overall, the code looks good. Only important thing actually missing are tests

rpc/v6/simulation.go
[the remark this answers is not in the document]
infrmtcs: I think go slice behaves similar to C++ vector, dynamic resize when appending to a full capacity will allocate new memory with twice the size and copy the existing elements to the new chunk of memory. That's why append will be more expensive if we know in advance the size of the slice.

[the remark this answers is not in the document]
hudem1: Yes but i thought the extra cost of `append` came only from dynamic reallocation. So I thought if we knew the exact final size of the slice, we wouldn't exceed it and get this extra cost (because no reallocation would be needed).

And as the other method creates some (useless) initial zero-valued elements, I thought the no-realloc append would end up being more efficient.

But actually, it seems like there `append` still causes some other cost for some reason (unrelated to realloc).

rpc/v7/simulation.go
```diff
@@ -163,18 +165,19 @@ func (h *Handler) simulateTransactions(id BlockID, transactions []BroadcastedTra
...
 
-		trace := traces[i]
+		trace := AdaptVMTransactionTrace(&vmTraces[i])
 
-		traces[i].ExecutionResources = &vm.ExecutionResources{
+		// Add execution resources on the trace root level (from all the fct invocations)
+		trace.ExecutionResources = &ExecutionResources{
 			ComputationResources: trace.TotalComputationResources(),
-			DataAvailability: &vm.DataAvailability{
+			DataAvailability: &DataAvailability{
```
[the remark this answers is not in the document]
hudem1: Actually, we shouldn't have `DataAvailability` in v6. Right now, the code in v6 using `DataAvailibilty` is dead code.

I planned to include that removal in a cleaning PR after merging this one, because it requires removing quite a bit of code. And as this PR is already a bit big and we want to merge it quickly and there is some other cleaning i wanna do, i thought it would be better to do that in another PR.

rpc/v8/trace_test.go
```diff
@@ -46,12 +46,12 @@ func TestTraceFallback(t *testing.T) {
...
...a5ec9ea81740eca5b482d12e","calldata":["0x1176a1bd84444c89232ec27754698e5d2e7e1a7f1539f12027f28b23ec9f3d8","0x3b2d25cd7bccc","0x0"],"caller_address":"0x58b7ee817bd2978c7657d05d3131e83e301ed1aa79d5ad16f01925fd52d1da7","class_hash":"0xd0e183745e9dae3e4e78a8ffedcce0903fc4900beace4e0abf192d4c202da3","entry_point_type":"EXTERNAL","call_type":"CALL","result":["0x1"],"calls":[{"contract_address":"0x49d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7","entry_point_selector":"0x83afd3f4caedc6eebf44246fe54e38c95e3179a5ec9ea81740eca5b482d12e","calldata":["0x1176a1bd84444c89232ec27754698e5d2e7e1a7f1539f12027f28b23ec9f3d8","0x3b2d25cd7bccc","0x0"],"caller_address":"0x58b7ee817bd2978c7657d05d3131e83e301ed1aa79d5ad16f01925fd52d1da7","class_hash":"0x28d7d394810ad8c52741ad8f7564717fd02c10ced68657a81d0b6710ce22079","entry_point_type":"EXTERNAL","call_type":"DELEGATE","result":["0x1"],"calls":[],"events":[],"messages":[],"execution_resources":{"l1_gas":0,"l2_gas":0}}],"events":[],"messages":[],"execution_resources":{"l1_gas":0,"l2_gas":0}},"execution_resources":{"l1_gas":0,"l2_gas":0,"l1_data_gas":0}},"transaction_hash":"0xbc984e8e1fe594dd518a3a51db4f338437a5d2fbdda772d4426b532a67ffff"}]`,
```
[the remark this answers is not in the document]
hudem1: Because now the function tested does not return the same result anymore as execution_resources is not the same anymore

rpc/v6/adapters.go
```diff
@@ -0,0 +1,299 @@
...
+	return &TransactionTrace{
```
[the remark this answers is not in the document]
hudem1: Okay sure!
But just for my understanding, isn't returning a pointer more efficient? because if we return the value, it will create a copy, right ? (and the 2 places where i call this function want a pointer to the type)

hudem1:
> I signaled a couple of things for code in `v6`. I believe this thing should be re-applied to `v7` and `v8`.
> 
> Overall, the code looks good. Only important thing actually missing are tests

I think the tests already exist, i adapted the existing ones to match new ExecutionResources type. Should i add some more ?

rpc/v6/simulation.go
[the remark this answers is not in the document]
rodrigo-pino: `append` each time you add a new element will also has an addition (increasing the length from `i` to `i+1`), while assigning is just that.

Because you are already creating the slice with the right capacity I don't think a slice regrow isn't happening here.

rpc/v6/adapters.go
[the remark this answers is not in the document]
rodrigo-pino: If you return a pointer, you are copying the value to the heap (which is `X` times more expensive) than copying it to the stack, and then returning a pointer to it. You don't want to touch the or store anything of the `heap` if you really don't need to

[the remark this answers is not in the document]
hudem1: ohhh right! I get it, makes sense!

rpc/v6/simulation.go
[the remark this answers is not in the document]
hudem1: True there is this additional operation. But i feel like there must be more to it to justify this big perfomance difference

rpc/v7/trace.go
```diff
@@ -376,14 +264,50 @@ func (h *Handler) fetchTraces(ctx context.Context, blockHash *felt.Felt) ([]Trac
...
-	traces, aErr := adaptBlockTrace(rpcBlock, blockTrace)
+	traces, aErr := AdaptFeederBlockTrace(rpcBlock, blockTrace)
 	if aErr != nil {
 		return nil, rpccore.ErrUnexpectedError.CloneWithData(aErr.Error())
 	}
 
 	return traces, nil
 }
 
+func (t *TransactionTrace) TotalComputationResources() ComputationResources {
```
rianhughes, line 275: pls move this to the struct definition

rpc/v7/trace.go
```diff
@@ -376,14 +264,50 @@ func (h *Handler) fetchTraces(ctx context.Context, blockHash *felt.Felt) ([]Trac
...
+		total.Poseidon += r.Poseidon
+		total.SegmentArena += r.SegmentArena
+		total.MemoryHoles += r.MemoryHoles
+		total.Steps += r.Steps
+	}
+
+	return total
+}
+
+func (t *TransactionTrace) allInvocations() []*rpcv6.FunctionInvocation {
```
rianhughes, line 296: this too

rpc/v7/trace.go
hudem1: oh okay i thought you initially meant the file where trace is defined. I'll move it closer to the struct definition then!

rpc/v8/trace.go
```diff
@@ -24,106 +24,43 @@ var traceFallbackVersion = semver.MustParse("0.13.1")
...
+	CallerAddress      felt.Felt                 `json:"caller_address"`
+	ClassHash          *felt.Felt                `json:"class_hash,omitempty"`
+	EntryPointType     string                    `json:"entry_point_type,omitempty"`
+	CallType           string                    `json:"call_type,omitempty"`
+	Result             []felt.Felt               `json:"result"`
+	Calls              []FunctionInvocation      `json:"calls"`
+	Events             []vm.OrderedEvent         `json:"events"`
+	Messages           []vm.OrderedL2toL1Message `json:"messages"`
+	ExecutionResources *InnerExecutionResources  `json:"execution_resources,omitempty"`
 }
```
rianhughes, line 64: You're missing a field
```
"is_reverted": {
                "title": "Is Reverted",
                "description": "true if this inner call panicked",
                "type": "boolean"
              }
```

rodrigo-pino:
Closing, is being replaced by

#2581
#2591 
#2600

PR #2564, opened 2025-02-25
test(trie): fix data race in range proof

Exca-DK:
The issue was caused by a shared error variable being accessed concurrently across subtests running in parallel.

Validate:
```
$ go test -run=^TestOneElementRangeProof$ github.com/NethermindEth/juno/core/trie -v -race
```


weiihann (approved):
Good catch, thanks!

infrmtcs (approved):

PR #2565, opened 2025-02-25
test(cmd): safer and simpler cmd test env

Exca-DK:
This change replaces manual environment variable handling with `t.Setenv`, which simplifies the logic and ensures proper cleanup.

`t.Setenv` automatically restores the previous value (or removes it if it didn’t exist) using `t.Cleanup`, making the tests more robust, even in case of a crash.

Also makes my linter happy since its been screaming for some time for usage of `os.setenv` in tests.

Ref: [t.Setenv](https://github.com/golang/go/blob/b38b0c0088039b03117b87eee61583ac4153f2b7/src/testing/testing.go#L1327)

### DIFF
--- a/cmd/juno/juno_test.go
+++ b/cmd/juno/juno_test.go
@@ -7,7 +7,6 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
-	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


weiihann (approved):
LGTM, thanks for the PR!

PR #2566, opened 2025-02-26
fix(rpc): Functions in the rpc package were sending the wrong version to vm.Call

rodrigo-pino:
Made `SierraVersion` a method of the `Class` interface where `Cairo0` and `Cairo1` classes implement it. The implementation is the previous function `parseSierra` divided.

Added a restriction to `AdaptCairoClass` to check for the Sierra program size to be at least 3. What other checks could we do?


core/class.go
```diff
@@ -50,10 +52,16 @@ func (c *Cairo0Class) Hash() (*felt.Felt, error) {
...
+}
+
 // Cairo1Class unambiguously defines a [Contract]'s semantics.
 type Cairo1Class struct {
-	Abi         string
-	AbiHash     *felt.Felt
+	Abi     string
+	AbiHash *felt.Felt
+	// TODO: will implement this on a follow up PR commit to avoid the migration
+	// EntryPoints     SierraEntryPointsByType
```
rodrigo-pino, line 64: I don't think we should use anonymous struct. Plan to do it in a follow up, but it is a nit

core/class.go
```diff
@@ -50,10 +52,16 @@ func (c *Cairo0Class) Hash() (*felt.Felt, error) {
 	return cairo0ClassHash(c)
 }
 
+func (c *Cairo0Class) SierraVersion() string {
+	return "0.1.0"
```
rodrigo-pino, line 56: Which source we have on this value (cc: @rianhughes). Looking at the Blockifier code, deprecated Sierra is 0.0.0 👀. It actually doesn't affect the blockifier execution, because literally everything below 2.9.0 is executed with the VM.

core/class_test.go
```diff
@@ -263,3 +263,24 @@ func TestSegmentedBytecodeHash(t *testing.T) {
...
+		require.Equal(t, "0.1.0", sierraVersion)
+	})
+
+	t.Run("cairo one should return based on the program data", func(t *testing.T) {
+		class := core.Cairo1Class{
+			Program: []*felt.Felt{
+				new(felt.Felt).SetUint64(7),
+				new(felt.Felt).SetUint64(3),
+				new(felt.Felt).SetUint64(11),
+			},
```
rodrigo-pino, line 280: I don't like to create a custom program every time the version is required. I've also noticed that we have the patter of creating felts and creating array of felts all around the code. What do you think of having testing utilities for this kind of things

vm/rust/src/execution.rs
rodrigo-pino, line 1: This file only changed because I applied formatting.

vm/vm_test.go
```diff
@@ -50,7 +50,7 @@ func TestV0Call(t *testing.T) {
 		ContractAddress: contractAddr,
 		ClassHash:       classHash,
 		Selector:        entryPoint,
-	}, &BlockInfo{Header: &core.Header{}}, testState, &utils.Mainnet, 1_000_000, "")
+	}, &BlockInfo{Header: &core.Header{}}, testState, &utils.Mainnet, 1_000_000, simpleClass.SierraVersion())
```
rodrigo-pino, line 53: I don't know why we should pass the Sierra version explicitly, don't the rust code or the go code have access to the Sierra version too?

core/class.go
rianhughes: What do you think about removing the Todos and handling them in a seperate PR?

core/class.go
```diff
@@ -108,6 +123,26 @@ func (c *Cairo1Class) Hash() (*felt.Felt, error) {
...
+	const base = 10
+	var buf [32]byte
+	b := buf[:0]
+	b = strconv.AppendUint(b, c.Program[0].Uint64(), base)
+	b = append(b, '.')
+	b = strconv.AppendUint(b, c.Program[1].Uint64(), base)
+	b = append(b, '.')
+	b = strconv.AppendUint(b, c.Program[2].Uint64(), base)
+	return string(b)
+}
```
rianhughes, line 144: This will panic if len(c.Program)<3, right? I imagine such contracts shouldn't exist, but wondering if we should have a safety check anyway (eg for special contracts that might be forced onto the network)

adapters/sn2core/sn2core.go
```diff
@@ -264,27 +264,40 @@ func AdaptDeployAccountTransaction(t *starknet.Transaction) *core.DeployAccountT
...
-	class.ProgramHash = crypto.PoseidonArray(class.Program...)
-
-	class.Abi = response.Abi
-	class.AbiHash = crypto.StarknetKeccak([]byte(class.Abi))
-
-	adapt := func(ep starknet.SierraEntryPoint) core.SierraEntryPoint {
-		return core.SierraEntryPoint{Index: ep.Index, Selector: ep.Selector}
+	// A Sierra program minimum size should be three to contain the version
+	if len(response.Program) < 3 {
+		return nil, errors.New("sierra program size is too small")
```
rianhughes, line 269: nit: "Sierra program length must be at least 3 Felts"

vm/vm_test.go
rianhughes: It seems that the sierra version is passed to set the gas cost for cairoVMCall, but PF don't seem to do this. @AnkushinDaniil, why does our implementation differ? Maybe we can drop passing the sierra version into the vm completely?

core/class_test.go
rianhughes: Like `Utils.CreateArrFelts(numFelts)`? Sure, that would help make the code neater

core/class.go
rodrigo-pino: I see what you mean. Fixing it now

core/class.go
rodrigo-pino: This will never panic. The check is made when we adapt the type, which I thought was the sole constructor, but we can construct types as well from P2P. I am going to see how to fix that.

Even an empty compiled contract includes the Sierra Version.

rpc/v7/trace_test.go
```diff
@@ -560,17 +560,26 @@ func TestCall(t *testing.T) {
...
-		mockState.EXPECT().Class(classHash).Return(&core.DeclaredClass{Class: &core.Cairo1Class{}}, nil)
+		mockState.EXPECT().Class(classHash).Return(&core.DeclaredClass{Class: &cairoClass}, nil)
 		mockReader.EXPECT().Network().Return(n)
 		mockVM.EXPECT().Call(&vm.CallInfo{
 			ContractAddress: contractAddr,
 			ClassHash:       classHash,
 			Selector:        selector,
 			Calldata:        calldata,
-		}, &vm.BlockInfo{Header: headsHeader}, gomock.Any(), &utils.Mainnet, uint64(1337), "").Return(expectedRes, nil)
+		}, &vm.BlockInfo{Header: headsHeader}, gomock.Any(), &utils.Mainnet, uint64(1337), cairoClass.SierraVersion()).Return(expectedRes, nil)
```
hudem1, line 582: shouldn't we here use the hardcoded version string like "3.0.0" because you are actually testing the call to `SierraVersion`, right ?

rpc/v7/trace_test.go
```diff
@@ -598,17 +607,25 @@ func TestCall(t *testing.T) {
...
```
hudem1, line 628: Same comment as above with "3.0.0"

adapters/sn2core/sn2core.go
```diff
@@ -264,27 +264,40 @@ func AdaptDeployAccountTransaction(t *starknet.Transaction) *core.DeployAccountT
...
+
+		Abi:     response.Abi,
+		AbiHash: crypto.StarknetKeccak([]byte(response.Abi)),
+
+		Compiled: coreCompiledClass,
+
+		EntryPoints: struct {
+			Constructor []core.SierraEntryPoint
+			External    []core.SierraEntryPoint
+			L1Handler   []core.SierraEntryPoint
```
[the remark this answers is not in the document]
rodrigo-pino: yes, I think we should, but not on this PR. I want for it to be done on a separate commit. This way, the next release won't need an unnecesary migration again

utils/slices.go
```diff
@@ -21,6 +21,20 @@ func Map[T1, T2 any](slice []T1, f func(T1) T2) []T2 {
 	return result
 }
 
+// The same as Map but the function receives a reference type
+func MapByRef[T1, T2 any](slice []T1, f func(*T1) T2) []T2 {
+	if slice == nil {
+		return nil
+	}
```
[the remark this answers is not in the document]
rodrigo-pino: I think the intent is to return nil. In my opinion we should panic in this kind of instance since it is never the developer intention to map over a nil slice. 

I would like to that refactor but I think this PR is not the place to do it right now. See that `MapByRef` is just a copy of `Map`. If we were to change one, we should change the other as well for the sake of consistency

rpc/v7/trace_test.go
rodrigo-pino: The test for Sierra Version are in other place. I am just passing the sierra version the contract says it has

hudem1: Oh okay, it thought it was some additional way of testing `SierraVersion` call (in addition to testing the vm.Call here). But yeah makes sense that the test explicitly for `SierraVersion` suffices and then we can safely assume it works and can be used in other tests.

core/class.go
```diff
@@ -123,15 +127,18 @@ func (c *Cairo1Class) Hash() (*felt.Felt, error) {
...
 // For all subsequent versions the version number is the first three felts
 // representing the three parts of a semantic version number.
 func (c *Cairo1Class) SierraVersion() string {
-	// It is defined in this weird way because it is fast :)
+	if c.Program[0].Equal(&SierraVersion010) {
+		return "0.1.0"
+	}
+
 	const base = 10
 	var buf [32]byte
```
hudem1, line 150: just out of curiosity, here, isnt 32 bytes a bit big for the returned string? won't it be just about 5 letters, like "1.6.0" ? is there a reason why we use that many bytes ?

adapters/sn2core/sn2core_test.go
```diff
@@ -564,11 +564,24 @@ func TestClassV1(t *testing.T) {
 		}
 	})
 
-	t.Run("sierra class too small", func(t *testing.T) {
+	// TODO: A test is missing where we adapt a class with Sierra Version 0.1.0
```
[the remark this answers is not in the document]
rodrigo-pino: I've already added this. Maybe you reviewed old changes 🤔

core/class_test.go
rodrigo-pino: Yes, something like that, that takes variable  number of args

```go
Program: utils.FeltArray(7, 3, 11) // returns []felt.Felt 
```

core/class.go
rodrigo-pino: Today they are 5 bytes. Tomorrow it could be a bit more, "123.23.12" which in that case it would also be enough. The idea is for there to be enough space for any type of version. The cost of this is negligible.

Also, I think we should represent SierraVersion as it's own type and not a string but that's also for a future PR

hudem1:
How about the call to `vm.Call` from v6 pkg https://github.com/NethermindEth/juno/blob/main/rpc/v6/trace.go#L382 ?
Where can i verify whether we need to pass the sierra version for that rpc version ?

vm/vm_test.go
AnkushinDaniil: I think it's possible. Initially I implemented the same approach as Pathfinder, but later we decided to add a Sierra version to get initial gas from versioned constants.

rodrigo-pino:
> Where can i verify whether we need to pass the sierra version for that rpc version ?

We need to pass it always! You just catched a bug. Let me see how to solve it in this P.R

vm/vm_test.go
rodrigo-pino: We can address it up in a follow up PR

core/class.go
rianhughes: >Even an empty compiled contract includes the Sierra Version.

Okay, if that's guaranteed to be the case then I think this is good

rianhughes:
LGTM up to the rpcv6 bug

PR #2572, opened 2025-02-26
feat(ci): add pre-merge e2e checks workflow with concurrency controls

wojciechos:
This PR improves our CI/CD pipeline by adding pre-merge e2e checks and implementing better concurrency controls for shared dev environment.

## Changes:

- **Added pre-merge e2e checks workflow** that runs on PRs and merge groups targeting the main branch
- **Extracted development environment pipeline** into a reusable workflow to reduce duplication
- **Implemented concurrency controls** to prevent race conditions in shared environments
- **Set appropriate workflow permissions** following security best practices
- **Added test runs to concurrency groups** for better resource management

## Benefits:

- **Catch issues earlier** by running e2e tests before merging to main
- **Prevent race conditions** when multiple workflows try to use the same shared environment
- **Reduce duplication** in workflow definitions for better maintainability
- **Improve security** with proper permission settings
- **Speed up builds** with optimized Docker image building process


.github/workflows/pre-merge-e2e-checks.yml
```diff
@@ -0,0 +1,88 @@
...
+    secrets:
+      STARKNET_RPC: ${{ secrets.DEV_SEPOLIA_URL }}/v0_7
+
+  dev-starknet-js-tests:
+    needs: [validate_dev]
+    uses: ./.github/workflows/starknet-js-tests.yml
+    secrets:
+      TEST_RPC_URL: ${{ secrets.DEV_SEPOLIA_URL }}/v0_7
+      TEST_ACCOUNT_ADDRESS: ${{ secrets.TEST_ACCOUNT_ADDRESS }}
+      TEST_ACCOUNT_PRIVATE_KEY: ${{ secrets.TEST_ACCOUNT_PRIVATE_KEY }}
```
infrmtcs, line 82: Does it cause conflict if we run multiple instances of this test? I think we may consider adding job-level concurrency group for this.

.github/workflows/pre-merge-e2e-checks.yml
```diff
@@ -0,0 +1,88 @@
...
+          context: .
+          platforms: "linux/amd64"
+          push: true
+          tags: |
+            ${{ env.DOCKER_REGISTRY }}/${{ env.REPO_DEV }}/juno:${{ env.DOCKER_IMAGE_TAG }}
+
+  validate_dev:
+    permissions:
+      id-token: write
+      contents: write
```
infrmtcs, line 57: I think the steps don't modify anything, so probably we don't need these lines.

infrmtcs (approved):

.github/workflows/pre-merge-e2e-checks.yml
```diff
@@ -0,0 +1,88 @@
+name: Pre-merge e2e checks
+
+permissions: {}
```
derrix060, line 3: NIT: I don't think this is required :)

```suggestion
```

.github/workflows/pre-merge-e2e-checks.yml
```diff
@@ -0,0 +1,88 @@
...
+env:
+  DOCKER_REGISTRY: nethermind.jfrog.io
+  REPO_DEV: nubia-oci-local-dev
+
+on:
+  merge_group:
+    branches: [main]
+
+concurrency:
+  group: shared_dev_environment
```
derrix060, line 14: this won't run at the same time as the ci-cd-pipeline, is this expected?

.github/workflows/pre-merge-e2e-checks.yml
```diff
@@ -0,0 +1,88 @@
...
+  cancel-in-progress: false
+
+jobs:
+  build_docker_image:
```
derrix060, line 18: this job is VERY similar to the ci-cd-pipeline, I think it's a good idea to keep them both the same so there is no need to support both and if something changes in one, remember to change in the other one as well...

This can be done with either creating a composite, or just a "shared" workflow

.github/workflows/pre-merge-e2e-checks.yml
derrix060: there is already a concurrency group for this workflow, so it won't run multiple instances of this test :)

derrix060 (approved):

.github/workflows/pre-merge-e2e-checks.yml
wojciechos: Yep, totally agree - done.

.github/workflows/pre-merge-e2e-checks.yml
wojciechos: yep, it's expected

.github/workflows/dev-environment-pipeline.yml
```diff
@@ -0,0 +1,119 @@
...
+
+      - name: Setup Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Build Docker Image
+        uses: docker/build-push-action@v6
+        with:
+          context: .
+          platforms: "linux/amd64"
+          push: false
```
derrix060, line 51: why not pushing it here?  You can set it to push to multiple tags :)

.github/workflows/dev-environment-pipeline.yml
```diff
@@ -0,0 +1,119 @@
...
+          outputs: type=docker,dest=/tmp/juno-image.tar
```
derrix060, line 52: maybe the problem could be that the type is set to docker instead of OCI, let me branch off this and play around

.github/workflows/dev-environment-pipeline.yml
```diff
@@ -0,0 +1,119 @@
...
+          path: /tmp
+
+      - name: Setup Docker Buildx
+        uses: docker/setup-buildx-action@v3
+        id: buildx
+
+      - name: Login to registry
+        run: |
+          docker login ${{ env.DOCKER_REGISTRY }} -u ${{ secrets.ARTIFACTORY_NUBIA_USERNAME }} -p ${{ secrets.ARTIFACTORY_NUBIA_TOKEN_DEVELOPER }}
+          echo "Testing login:" && docker push ${{ env.DOCKER_REGISTRY }}/${{ env.REPO_DEV }}/juno:latest --dry-run || echo "Login failed"
```
derrix060, line 86: if the login failed, it should fail the job, not silently print a failure here...

Also, `--dry-run` was not recognized as a command

.github/workflows/dev-environment-pipeline.yml
```diff
@@ -0,0 +1,119 @@
...
+            
+      - name: Verify Deployment Version (Dev)
+        run: |
+          bash .github/workflow-scripts/verify_deployment.sh ${{ secrets.DEV_SEPOLIA_URL }} ${{ needs.build.outputs.DOCKER_IMAGE_TAG }}
+
+  starknet-rs:
+    needs: [deploy]
+    uses: ./.github/workflows/starknet-rs-tests.yml
+    secrets:
+      STARKNET_RPC: ${{ secrets.DEV_SEPOLIA_URL }}/v0_7
```
derrix060, line 105: nit: maybe the version could be an input, in case one want to change it, defaulting it to 0_7 of course

derrix060 (changes requested):

.github/workflows/dev-environment-pipeline.yml
wojciechos: Right, saw you already updated it, thanks!

.github/workflows/dev-environment-pipeline.yml
wojciechos: yeah, this was just some mess for debugging purposes

.github/workflows/dev-environment-pipeline.yml
wojciechos: yep, great idea - also saw you already added it, thanks!

.github/workflows/dev-environment-pipeline.yml
```diff
@@ -0,0 +1,128 @@
...
+          NEW_TAG=${{ env.DOCKER_REGISTRY }}/${{ env.REPO_DEV }}/juno:latest
+          docker buildx imagetools create -t $NEW_TAG $OLD_TAG
+            
+      - name: Verify Deployment Version (Dev)
+        run: |
+          bash .github/workflow-scripts/verify_deployment.sh ${{ secrets.DEV_SEPOLIA_URL }} ${{ needs.build.outputs.DOCKER_IMAGE_TAG }}
+
+  starknet-rs:
+    concurrency:
+      group: shared_dev_environment
```
derrix060, line 103: I just thought about a case where there might be two different jobs waiting to be executed, then one of them push to development cluster, than the other one push, then the tests for the first try to run the tests, but it will be on the second job.


One way to fix that is to split this pipeline into

1: build a container
2: push the image and run the tests (the whole workflow using the concurrency, not per job

.github/workflows/dev-environment-pipeline.yml
```diff
@@ -0,0 +1,128 @@
...
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    outputs:
+      DOCKER_IMAGE_TAG: ${{ steps.set_tag.outputs.DOCKER_IMAGE_TAG }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
```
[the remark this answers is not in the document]
wojciechos: Yep, it's needed to get proper tag

.github/workflows/dev-environment-pipeline.yml
wojciechos: Yeah, I was thinking about this and assumed that if the tests share the same concurrency setting and depend on it, they would be enforced to run before other workflows. I even tested this case and it worked as expected, but tbh still not sure if that gonna always work this way. 

The issue with "push the image and run the tests (the whole workflow using concurrency, not per job)” is that workflow_call cannot be run within a step. Only as separate job.

.github/workflows/dev-environment-pipeline.yml
```diff
@@ -0,0 +1,128 @@
...
+        required: true
+      TEST_ACCOUNT_PRIVATE_KEY:
+        required: true
+
+permissions:
+  contents: read  
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
```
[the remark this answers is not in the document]
wojciechos: but it only cancels runs in the same group `${{ github.workflow }}-${{ github.ref }}`. So, for a PR, it'd cancel and rerun pre-merge tests if someone adds a commit (same refs/pull/123), but it won’t touch CI/CD on main or other PRs since their refs are different.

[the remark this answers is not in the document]
derrix060: ah true, I thought the github.ref would be the git commit, but it's actually the branch

image

.github/workflows/dev-environment-pipeline.yml
derrix060: I think it's still okay:

workflow CI:
  - job: build image
  - job: run dev test
  - ...
  
merge queue:
  - job: build image
  - job: run dev test

derrix060 (approved):

derrix060 (changes requested):
image

derrix060 (approved):

PR #2573, opened 2025-02-26
fix(rpc/v7): add support for `l1_data_gas` resource type

AnkushinDaniil:


weiihann (approved):

wojciechos:
Shouldn’t we add this to rpc/v6 as well?

hudem1:
It seems like I'm missing something but i thought this field was not supported in [v7](https://github.com/starkware-libs/starknet-specs/blob/master/api/starknet_api_openrpc.json#L3406-L3421) ?

rpc/v7/transaction.go
```diff
@@ -173,19 +175,24 @@ func (r Resource) MarshalText() ([]byte, error) {
 		return []byte("l1_gas"), nil
 	case ResourceL2Gas:
 		return []byte("l2_gas"), nil
+	case ResourceL1DataGas:
```
rianhughes, line 178: Give rpcv7 and rpcv6 don't have any concept of l1_data_gas, do you not think it would be cleaner to update
`func adaptResourceBounds()` so that we only loop over L1_GAS and L2_GAS?

I feel like the problem arises because we loop over all three `core.Resource` enums in `adaptResourceBounds()`, but `rpcv7.Resource` only has two enums (similar for rpcv6)

hudem1: that's what i did in my [PR](https://github.com/NethermindEth/juno/pull/2575)

rianhughes: This solution will return an `l1_data_gas` field which shouldn't be present imo

AnkushinDaniil: Yes, you're right

AnkushinDaniil:
#2575

rpc/v7/transaction.go
rianhughes: Something like this https://github.com/NethermindEth/juno/pull/2576

Issue #2574, opened 2025-02-26
API error with starknet_getBlockWithReceipts


mcdee:
Juno version 0.13.0 on sepolia testnet.

Attempting the following request to the API:

```
curl -H 'Content-type: application/json' --data '{"method":"starknet_getBlockWithReceipts","params":{"block_id":{"block_number":551637}},"id":0,"jsonrpc":"2.0"}' http://localhost:6060/v0_7
```

Returns a 500 error.  Logs for juno below:

```
11:02:44.801 26/02/2025 +00:00  ERROR   jsonrpc/http.go:56      Handler failure {"err": "json: encoding error for type \"map[rpcv7.Resource]rpcv7.ResourceBounds\": \"unknow
n Resource 3\""}
github.com/NethermindEth/juno/jsonrpc.(*HTTP).ServeHTTP
        /home/runner/work/juno/juno/jsonrpc/http.go:56
github.com/NethermindEth/juno/node.makeRPCOverHTTP.exactPathServer.func1
        /home/runner/work/juno/juno/node/http.go:76
net/http.HandlerFunc.ServeHTTP
        /opt/hostedtoolcache/go/1.23.4/x64/src/net/http/server.go:2220
net/http.(*ServeMux).ServeHTTP
        /opt/hostedtoolcache/go/1.23.4/x64/src/net/http/server.go:2747
net/http.serverHandler.ServeHTTP
        /opt/hostedtoolcache/go/1.23.4/x64/src/net/http/server.go:3210
net/http.(*conn).serve
        /opt/hostedtoolcache/go/1.23.4/x64/src/net/http/server.go:2092
```

rodrigo-pino:
Looking into it now. Thank you

hudem1:
Hi @mcdee,

The issue should be fixed now.
Don't hesitate to let us know if you encounter any other issue.

Thank you!

PR #2575, opened 2025-02-26
fix(rpc v6 and v7): conversion from core.ResourceL1DataGas unsupported

hudem1:
Fixes bug in issue: https://github.com/NethermindEth/juno/issues/2574


rodrigo-pino (commented):
Hugo, please add a new test that checks the error

weiihann:
Does v6 need to add this check as well?

hudem1:
> Hugo, please add a new test that checks the error

Oh yes of course!

wojciechos (approved):

infrmtcs (approved):

PR #2581, opened 2025-02-28
fix(rpc): TransactionTrace matches rpc v6 specs

hudem1:
As agreed, i divide #2563 into 3, one PR for each of the 3 rpc pkgs

Here is the PR 1/3 for rpcv6 pkg: create rpcv6.TransactionTrace instead of using vm.TransactionTrace

### Main modifications:

- `TransactionTrace.ExecutionResources`
--> _no root-level `ExecutionResources`, and inner ones only have `ComputationResources` (no `DataAvailability`)_

- `TransactionTrace.FunctionInvocation.Events`
--> _no `From` field_

- `TransactionTrace.FunctionInvocation.Messages`
--> _all fields are required, `To` is a felt, and set `From` when adapting feeder trace (with `ContractAddress`)_

- `TransactionTrace.FunctionInvocation`
--> _all fields are required_

- Update a test using `Goerli Integration` to use `Sepolia` instead, to provide more accurate tests


rpc/v6/adapters.go
```diff
@@ -0,0 +1,301 @@
...
+)
+
+/****************************************************
+		VM Adapters
+*****************************************************/
+
+func AdaptVMTransactionTrace(trace *vm.TransactionTrace) TransactionTrace {
+	var validateInvocation *FunctionInvocation
+	if trace.ValidateInvocation != nil {
+		validateInvocation = utils.Ptr(AdaptVMFunctionInvocation(trace.ValidateInvocation))
```
rodrigo-pino, line 19: Don't using utils.Ptr. Same problem we discussed before. Pre-empetively allocating in the heap is not a good idea

hudem1: Oh okay, there too. So, i have to change the whole `TransactionTrace` type then to not contain pointers ?

rpc/v6/trace.go
```diff
@@ -24,106 +24,42 @@ const excludedVersion = "0.13.1.1"
...
-	return traces, nil
+type TransactionTrace struct {
+	Type                  TransactionType     `json:"type"`
+	ValidateInvocation    *FunctionInvocation `json:"validate_invocation,omitempty"`
+	ExecuteInvocation     *ExecuteInvocation  `json:"execute_invocation,omitempty"`
+	FeeTransferInvocation *FunctionInvocation `json:"fee_transfer_invocation,omitempty"`
+	ConstructorInvocation *FunctionInvocation `json:"constructor_invocation,omitempty"`
+	FunctionInvocation    *FunctionInvocation `json:"function_invocation,omitempty"`
+	StateDiff             *StateDiff          `json:"state_diff,omitempty"`
 }
```
rodrigo-pino, line 35: Type defintion should go on top. Usually type definitions are first, followed up by their methods or functions related with them.

rpc/v6/adapters.go
```diff
@@ -0,0 +1,314 @@
...
+	stateDiff := AdaptVMStateDiff(trace.StateDiff)
+
+	return TransactionTrace{
+		Type:                  TransactionType(trace.Type),
+		ValidateInvocation:    validateInvocation,
+		ExecuteInvocation:     executeInvocation,
+		FeeTransferInvocation: feeTransferInvocation,
+		ConstructorInvocation: constructorInvocation,
+		FunctionInvocation:    functionInvocation,
+		StateDiff:             &stateDiff,
```
rodrigo-pino, line 56: Why don't use the `&` directly here
```suggestion

	return TransactionTrace{
		Type:                  TransactionType(trace.Type),
		ValidateInvocation:    &validateInvocation,
		ExecuteInvocation:     &executeInvocation,
		FeeTransferInvocation: &feeTransferInvocation,
		ConstructorInvocation: &constructorInvocation,
		FunctionInvocation:    &functionInvocation,
		StateDiff:             &stateDiff,
```

hudem1: Because it'd mean we have that for example:

```go
var validateInvocation FunctionInvocation
if trace.ValidateInvocation != nil {
	validateInvocation := AdaptVMFunctionInvocation(trace.ValidateInvocation)
}
```

And if `trace.ValidateInvocation` is nil, we'd still have created and assigned a zero-valued `FunctionInvocation` to `TransactionTrace.ValidateInvocation`

rpc/v6/trace.go
rodrigo-pino: I think this type should be define in adapters too. Also, from what I've read in the code `StateDiff` field is not meant to be empty (seeing the adapter functions). Should it be optional here?

hudem1: Oh you wanna move the definition to the adapters file ? when i check existing adapters file, they dont hold the types definition

rodrigo-pino: I see. Should then they stay here or move to core? @rianhughes 

My opinion is that they should stay here.

rpc/v6/adapters.go
rodrigo-pino: Talked a bit more we Hugo and we come to the realization that `utils.Ptr` makes sense here

rpc/v6/trace.go
hudem1: Also, given that this type is specific to rpcv6, i think it should be in v6 pkg

rpc/v6/transaction.go
```diff
@@ -263,6 +263,8 @@ type DataAvailability struct {
 	L1DataGas uint64 `json:"l1_data_gas"`
 }
 
+// TODO: specs say no DataAvailability
+// I will remove this in another PR because it causes a bit of code deletion
```
rodrigo-pino, line 267: Please create an issue and assign it to yourself, since it is a critical bug

rpc/v6/trace.go
hudem1: Actually, `StateDiff` is not mandatory. I got confused because it's a field common to all types of TransactionTrace, but it is not required.

On the other hand, some fields are required for the different trace types, so, i just added a commit with tests for those

rpc/v6/transaction.go
rodrigo-pino: If by the addition of this PR means some dead code is "generated", then let's delete it in this PR. If it is independent of this PR then let's make it in that other PR

rodrigo-pino (commented):
The code looks real good. I would like to add a real world test. Is this possible? Something that without this commit would cause the old code to fail

rpc/v6/transaction.go
hudem1: It's independent of this PR, this dead code was there before

rodrigo-pino: Great, then on another PR 👌

hudem1:
> The code looks real good. I would like to add a real world test. Is this possible? Something that without this commit would cause the old code to fail

I'm not sure what you mean by real world test. Do you mean an rpc call like `starknet_traceTransaction` for example ? If so, we do have tests for the different endpoints involving `TransactionTrace` type. Those tests wouldn't pass anymore without this commit as I've had to update them. That being said, I can definitely add some other tests if necessary !

hudem1:
In this PR, I also fixed `FunctionInvocation.Messages` by creating a struct specific to v6 instead of reusing vm's one :
- I made the `from_address` field mandatory, and fixed tests accordingly.
- The `to_address` field should be a felt. From the VM, we get a string, so, I had to convert it to a felt in the rpc v6 pkg. I think we could even potentially change this `string` type for a `felt` directly in the vm pkg instead. But that's something we could do after the 2 other PRs for v7 and v8 `TransactionTrace`.

rpc/v6/trace_test.go
```diff
@@ -59,20 +64,120 @@ func TestTraceFallback(t *testing.T) {
...
 			require.Nil(t, jErr)
-			jsonStr, err := json.Marshal(trace)
+			jsonStr, err := json.Marshal(traces)
 			require.NoError(t, err)
 			assert.JSONEq(t, test.want, string(jsonStr))
 		})
 	}
 }
 
+func TestTransactionTraceValidation(t *testing.T) {
```
rianhughes, line 89: Why no Declare case?

rianhughes (approved):

rpc/v6/trace_test.go
hudem1: [Declare class](https://github.com/starkware-libs/starknet-specs/blob/v0.6.0/api/starknet_trace_api_openrpc.json#L200) only requires "type" field, so, i cant really test it like the others.

That being said, instead of just testing missing fields (which Declare can't fail because it only requires `Type`), i could test that Declare does not have the fields it is not supposed to have like "execute_invocation", or "constructor_invocation" for example.

I will update the code actually to make sure `execute_invocation` is not set for any other trace type it's supposed to be! Same for the other fields!

rodrigo-pino: > i could test that Declare does not have the fields it is not supposed to have like "execute_invocation", or "constructor_invocation" for example.

These are the kind of checks we could have at compile time using generics (I think). Let's leave it for the future.

> It makes me realize the json validation only checks for missing fields, but not for excess fields. I need to update the code actually to make sure execute_invocation (for example) is not set for any other trace type it's supposed to be! Same for the other fields!

This is a good point, but I am not sure how important it is. A good and a bad thing is that we are deprecating rpcv06 in the future. This kind of testing gives us soundness but it is also an error that doesn't affect anyone. Given the high priority of this PR I think they shouldn't be included.

> Because here (and in the other cases, Transaction, etc.), the returned answer will get marshalled ignoring any type validation. Do you agree ? If I'm not mistaken, then I'll bring up this topic in the chat with everybody

Very good point, let's discuss it more on the chat tomorrow

rodrigo-pino (approved):

Issue #2579, opened 2025-02-28
`starknet_traceBlockTransaction` is still broken


rodrigo-pino:
Request
```
{
  "jsonrpc": "2.0",
  "method": "starknet_traceBlockTransactions",
  "params": [{"block_number": 549692}],
  "id": 1
}
```

Response
```
[
    {
        "jsonrpc": "2.0",
        "error": {
            "code": 63,
            "message": "An unexpected error occurred",
            "data": "execute transaction #10: failed txn 1276060426555445730023792470690681141059779074492672238940284418143503600119 reason: Resource L1Gas bounds (max amount: 1758, max price: 722702311890453) exceed balance (1256650064627212219)."
        },
        "id": 1
    }
]
```

infrmtcs:
Fixed by #2583

PR #2583, opened 2025-02-28
fix: VM fee issue due to selection of wrong versioned constants

infrmtcs:


vm/rust/src/lib.rs
```diff
@@ -711,11 +711,11 @@ fn get_versioned_constants(version: *const c_char) -> VersionedConstants {
...
     } else if version < StarknetVersion::from_str("0.13.2").unwrap() {
         CONSTANTS.get(&"0.13.1.1".to_string()).unwrap().to_owned()
-    } else if version < StarknetVersion::from_str("0.13.2.1").unwrap() {
-        CONSTANTS.get(&"0.13.2".to_string()).unwrap().to_owned()
     } else if version < StarknetVersion::from_str("0.13.3").unwrap() {
-        CONSTANTS.get(&"0.13.3".to_string()).unwrap().to_owned()
+        CONSTANTS.get(&"0.13.2".to_string()).unwrap().to_owned()
     } else if version < StarknetVersion::from_str("0.13.4").unwrap() {
+        CONSTANTS.get(&"0.13.3".to_string()).unwrap().to_owned()
+    } else if version < StarknetVersion::from_str("0.13.5").unwrap() {
```
rodrigo-pino, line 718: What do you think of having `<=` comparison so our brains can pick this kind of errors better.

infrmtcs: I think the intended behaviour is: Use 1.3.4 until next version appears. Let me try with >= to see if it works.

infrmtcs: I'm wondering if we can use `VersionedConstants` directly from blockifier instead of putting them in our repo and load them ourselves in Rust code. I pushed another commit overriding the original one to use it and it works.

rodrigo-pino (approved):

wojciechos:
As we migrated from using our own JSON constants files to versioned constants directly from the blockifier, can we clean up and remove the [versioned_constants_*](https://github.com/NethermindEth/juno/blob/main/vm/rust/resources/) files?

infrmtcs:
I initially thought about that, but currently they're being used as test fixtures for `SetVersionedConstants` (a function to load a custom versioned constants file) too. I make this PR minimal to get the bug fix first, then I want to check how the versions differ. If they all have the same structure, we can probably remove all but one and keep it as the test.

wojciechos (approved):

rodrigo-pino:
Fixes #2579

rodrigo-pino:
Fix #2579

PR #2589, opened 2025-03-01
feat(makefile): Suggest `test-cover` rule addition

hudem1:
Just a small PR to suggest an addition to the Makefile `test-cover` rule to be able to target a specific pkg when running the coverage instead of having to run all tests pkgs by default

Usage:
```bash
make test-cover PKG=path/to/pkg
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,6 +29,8 @@ else
 	NPROCS = $(shell grep -c 'processor' /proc/cpuinfo)
 endif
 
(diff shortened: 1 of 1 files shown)


rodrigo-pino (approved):

infrmtcs (approved):
LGTM, but I think we can apply this to other commands like test or benchmark

PR #2591, opened 2025-03-02
Hudem1/fix/v7/execution resources

hudem1:
As agreed, i divide #2563 into 3, one PR for each of the 3 rpc pkgs.

Please first review #2581

Here is the PR 2/3 for rpcv7 pkg: create rpcv7.TransactionTrace instead of using vm.TransactionTrace


rpc/v7/trace_test.go
```diff
@@ -15,63 +15,206 @@ import (
...
+			},
+		}
+
+		AssertTracedBlockTransactions(t, &utils.Integration, tests)
+	})
+
+	t.Run("Sepolia", func(t *testing.T) {
```
rianhughes, line 55: Nice

rpc/v7/trace_test.go
```diff
@@ -15,63 +15,206 @@ import (
...
+	}
+
+	invalidInvokeTransactionTrace := rpcv7.TransactionTrace{
+		Type: rpcv7.TxnInvoke,
+	}
+
+	validDeployAccountTransactionTrace := rpcv7.TransactionTrace{
+		Type:                  rpcv7.TxnDeployAccount,
+		ConstructorInvocation: &rpcv6.FunctionInvocation{},
+	}
```
rianhughes, line 144: Do we not also need execution_resources?

```
{
                        "name": "DEPLOY_ACCOUNT_TXN_TRACE",
                        "type": "object",
                        "description": "the execution trace of a deploy account transaction",
                        "properties": {
                           ...
                        },
                        "required": [
                            "type",
                            "execution_resources",
                            "constructor_invocation"
                        ]
                    },

```

rpc/v7/trace_test.go
```diff
@@ -15,63 +15,206 @@ import (
...
+
+	invalidDeployAccountTransactionTrace := rpcv7.TransactionTrace{
+		Type: rpcv7.TxnDeployAccount,
+	}
+
+	validL1HandlerTransactionTrace := rpcv7.TransactionTrace{
+		Type:               rpcv7.TxnL1Handler,
+		FunctionInvocation: &rpcv6.FunctionInvocation{},
+	}
```
rianhughes, line 153: Also need execution_resources here
```
                        "required": [
                            "type",
                            "function_invocation",
                            "execution_resources"
                        ]
   ```

File: rpc/v7/trace_test.go [hunk 4]
```diff
@@ -15,63 +15,206 @@ import (
...
+		name     string
+		trace    rpcv7.TransactionTrace
+		wantErr  bool
+		expected string
+	}{
+		{
+			name:     "valid INVOKE tx",
+			trace:    validInvokeTransactionTrace,
+			wantErr:  false,
+			expected: `{"type":"INVOKE","execute_invocation":{"revert_reason":""},"execution_resources":{"steps":0,"data_availability":null}}`,
```
rianhughes, line 169: does data_availability not have to be populated?

rpc/v7/trace_test.go
rianhughes: Looks like the test data has execution_resources

rpc/v7/trace_test.go
```diff
@@ -15,63 +15,206 @@ import (
...
+		t.Run(test.name, func(t *testing.T) {
+			err := validate.Struct(test.trace)
+
+			if test.wantErr {
+				assert.Error(t, err, "Expected validation to fail, but it passed")
+			} else {
+				assert.NoError(t, err, "Expected validation to pass, but it failed")
+
+				// Check marshalling (check required fields)
+				j, _ := json.Marshal(test.trace)
```
rianhughes, line 215: It's good practice to check the error

rpc/v7/trace_test.go
```diff
@@ -289,13 +435,16 @@ func TestTraceTransaction(t *testing.T) {
...
 
+		// Root level execution resources should be the sum of inner execution resources
 		vmTrace.ExecutionResources = &vm.ExecutionResources{
-			// other of fields are zero
+			ComputationResources: vm.ComputationResources{
+				Steps: 6,
+			},
 			DataAvailability: &vm.DataAvailability{
 				L1Gas: 1,
 			},
```
rianhughes, line 445: Both fields should be present, right?

rpc/v7/trace_test.go
```diff
@@ -250,14 +395,11 @@ func TestTraceTransaction(t *testing.T) {
 			"poseidon": 0,
 			"segmentarena": 0,
 			"memoryholes": 0,
-			"steps": 0,
-			"data_availability": {
-				"l1_gas": 1,
-				"l1_data_gas": 1
-			}
+			"steps": 2
```
rianhughes, line 398: why remove data_availability?

rianhughes: Right now it looks like we don't have data_availability in the vmJsonTrace, and set it "by hand" later. But I think we should be able to set it in the json data. What do you think? (Not sure if there was a good reason this was originally done - but it seems a bit hacky)

rpc/v7/trace_test.go
rianhughes: It looks like it's set to zero by default. However, lets set it to 1 to be more explicit (you'll also need to set it in consumedGas above)

rpc/v7/trace_test.go
```diff
@@ -481,19 +652,492 @@ func TestTraceBlockTransactions(t *testing.T) {
 				NumSteps:         stepsUsed,
 			}, nil)
 
+		result, httpHeader, err := handler.TraceBlockTransactions(context.Background(), rpcv7.BlockID{Hash: blockHash})
```
rianhughes, line 655: Can we update the "execution_resources" field in the json data to include both the data_availability field and the steps field?

eg "execution_resources": {"data_availability": {}}, -> "execution_resources": {"data_availability": {}, "steps":0},

rianhughes: And maybe add a failure test if either of these fields are not present?

rpc/v7/trace_test.go
```diff
@@ -15,63 +15,206 @@ import (
...
 
+			// Test it successfully returns the traces of a block (when a feeder client is set)
 			handler = handler.WithFeeder(client)
-			trace, httpHeader, jErr := handler.TraceBlockTransactions(context.Background(), rpcv7.BlockID{Number: test.blockNumber})
+			traces, httpHeader, jErr := handler.TraceBlockTransactions(context.Background(), rpcv7.BlockID{Number: test.blockNumber})
+			if n == &utils.Sepolia && description == "newer block" {
+				// For the newer block test, we test 3 of the block traces (INVOKE, DEPLOY_ACCOUNT, DECLARE)
+				traces = []rpcv7.TracedBlockTransaction{traces[0], traces[7], traces[11]}
+			}
+
```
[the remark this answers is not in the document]
hudem1: Because there are 12 traces in the block, so, it'd mean creating 12 expected traces. I think it'd take a lot of time to create those 12 expected traces for not much more benefit as I've chosen 3 traces of different types, so it covers the different cases. And also, for maintenance, it'd mean more work too when you make a small change to the code, you'd have to update all 12 expected traces.

rpc/v7/trace_test.go
```diff
@@ -15,63 +15,206 @@ import (
...
+			jsonStr, err := json.Marshal(traces)
 			require.NoError(t, err)
-			assert.JSONEq(t, test.want, string(jsonStr))
+			assert.JSONEq(t, test.trace, string(jsonStr))
+		})
+	}
+}
+
+func TestTransactionTraceValidation(t *testing.T) {
```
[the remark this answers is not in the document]
hudem1: No worries, here is the link to the [comment](https://github.com/NethermindEth/juno/pull/2581#discussion_r1977882394). TLDR; I cannot really create an invalid `Declare` trace as its only requirement is the `type` field and the validation only checks for missing fields and not excess ones.

That said, I still test the adaptation of a `Declare` tx in `TestAdaptVMTransactionTrace` and `TestAdaptFeederTransactionTrace` test. It does not exactly perform the same test as here it purely deals with validation & marshalling, but those 2 functions are the only 2 places v7.TransactionTrace instances are created

hudem1:
I'm replying to this [msg](https://github.com/NethermindEth/juno/pull/2591#discussion_r1979717922).

(I don't know why I dont have the option to reply directly on the thread or to "quote reply", hence providing the link to your msg)

Yes it's hand crafted. In this test (`TestTransactionTraceValidation`), I test the validation of TransactionTrace instances and that their marshalled representation indeed contains all required fields

rpc/v7/trace_test.go
hudem1: Yep true its more explicit!

rpc/v7/trace_test.go
hudem1: I removed the `data_availability` here because this corresponds to the inner call's ExecutionResources, which shouldn't have `data_availability` (only the trace root-level ExecutionResources struct should have). Actually, the VM could still theoretically return some data_availability (even for inner call's ExecutionResources) because the vm.ExecutionResources type has data_availability field (even though it practice it seems to always return nil data_availability), but anyway it will just get ignored in my fct when adapting the trace.FunctionInvocation(s) where i adapt only ComputationResources. 

I see what you mean but, in practice, the trace's (root-level) ExecutionResources coming from the VM is nil, and the data_availability part is returned in another distinct field here:
```go
type ExecutionResults struct {
	OverallFees      []*felt.Felt
	DataAvailability []core.DataAvailability <-- here
	GasConsumed      []core.GasConsumed
	Traces           []TransactionTrace
	NumSteps         uint64
}
```
So setting the ExecutionResources in the json will not be useful as it will get overriden by the data in this field (in the above ExecutionResults struct). So that's why we are kinda forced to "override" (in the test, just as in the code actually) the trace's (root-level) ExecutionResources.

I hope this makes sense! :)

rpc/v7/trace_test.go
hudem1: If I understand correctly, you're mentioning root-level `ExecutionResources`. As explained above (in a reply to some other comments), the VM trace's root-leve `ExecutionResources` is set to nil in practice (from what i've seen), and gets overriden anyway by the sum of inner `ExecutionResources` and DataAvailability coming from `ExecutionResults` struct. So I think it doesnt really matter what it's set to.

That being said, I still updated the test:
- giving the root-level `ExecutionResources` very large values to show they are not to be found in the expected trace --> they get overriden
- giving inner calls a non-zero `steps` field in their `ExecutionResources` to clearly show that the root-level `ExecutionResources` is the sum of all inner `ExecutionResources`
- giving non-zero values to `ExecutionResults.DataAvailability` to show those values are used in the returned trace (root-level `ExecutionResources`)


Now concerning your 2nd point:

The root-level `ExecutionResources` can be nil, i made a nil check, and anyway it gets overriden as explained.

But are we guaranteed VM trace inner calls' `ExecutionResources`  cannot be nil ? I think it has to be set by the VM (ie cannot be nil) and must contain at least the `steps` field (used in v6, v7 and v8). But I think thats maybe smth to test in the VM pkg: making sure that `vm.Execute()` returns a trace whose inner calls' `ExecutionResources` always contain at least the `steps` field.

The RPC pkg should assume the correctness of the VM's result, right ? I quickly tested here to set `execution_resources` to `null` in the json trace (simulated returned trace from the VM), and when i run the test (with nil inner ExecutionResources) i get a nil pointer dereference when `TotalComputationResources` gets called, which assumes the inner calls' `ExecutionResources` is not nil. Should I make a nil check there ? But if so, then we will probably not catch that the VM returns a nil inner `ExecutionResources`, which is not supposed to happen, right ?

rpc/v7/trace_test.go
rianhughes: Makes sense, thanks!

So basically:
1. blockifier no longer places  `data_availability` in the `execution_resources`. We extract it form the `receipt`, and pass it from blockifier to the vm via the `callContext`
2. the vm trace still has a `data_availability` field,  but we explicitly set the traces `data_availability` using the value we get from the `callContext`

rianhughes: Created an issue to keep trace of this https://github.com/NethermindEth/juno/issues/2615

rpc/v7/trace_test.go
rianhughes: Yup, the above comment explained why my original question / comment was incorrect (basically `data_availability` is no longer handled by the trace, but returned separately). Thanks!

rianhughes (approved):

rpc/v7/trace_test.go
hudem1: hahah true, an issue to modify this is a good idea!

PR #2600, opened 2025-03-03
Hudem1/fix/v8/execution resources

hudem1:
As agreed, i divide https://github.com/NethermindEth/juno/pull/2563 into 3, one PR for each of the 3 rpc pkgs.

Please first review https://github.com/NethermindEth/juno/pull/2591 as this one is rebased on it.

Here is the PR 3/3 for rpc v8 pkg: create rpcv8.TransactionTrace instead of using vm.TransactionTrace


hudem1:
The PR is ready for review, but don't merge yet. I wanna verify smth with reverted transactions

hudem1:
> The PR is ready for review, but don't merge it yet. I wanna verify smth with reverted transactions

I guess all transactions types can fail, but I could only find examples of reverted INVOKE txs on Voyager.

In that case, in the `execute_invocation` trace field:
```go
type ExecuteInvocation struct {
	RevertReason        string `json:"revert_reason"`
	*FunctionInvocation `json:",omitempty"`
}
```
- the `revert_reason` field is not empty and contains the reason
- the `function_invocation` field is empty (and therefore does not show `"is_reverted": true`)

I wanted to find examples of traces where their inner calls reverted and showed `"is_reverted": true`, but i cannot find any, it's always set to false. So, I couldn't fully test that aspect in practice

rodrigo-pino (approved):

rpc/v8/trace.go
```diff
@@ -135,11 +73,11 @@ func adaptFeederExecutionResources(resources *starknet.ExecutionResources) *vm.E
 //
 // It follows the specification defined here:
 // https://github.com/starkware-libs/starknet-specs/blob/1ae810e0137cc5d175ace4554892a4f43052be56/api/starknet_trace_api_openrpc.json#L11
-func (h *Handler) TraceTransaction(ctx context.Context, hash felt.Felt) (*vm.TransactionTrace, http.Header, *jsonrpc.Error) {
+func (h *Handler) TraceTransaction(ctx context.Context, hash felt.Felt) (*TransactionTrace, http.Header, *jsonrpc.Error) {
 	return h.traceTransaction(ctx, &hash)
 }
 
-func (h *Handler) traceTransaction(ctx context.Context, hash *felt.Felt) (*vm.TransactionTrace, http.Header, *jsonrpc.Error) {
+func (h *Handler) traceTransaction(ctx context.Context, hash *felt.Felt) (*TransactionTrace, http.Header, *jsonrpc.Error) {
```
rianhughes, line 80: We can inline this, right?

rpc/v8/trace_test.go
```diff
@@ -15,63 +15,207 @@ import (
...
 	"github.com/NethermindEth/juno/vm"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"go.uber.org/mock/gomock"
 )
 
+type expectedBlockTrace struct {
+	blockHash   string
+	blockNumber uint64
+	trace       string
```
rianhughes, line 34: nit: could we update this to something like "wantTrace" or "want"?

rpc/v8/trace_test.go
```diff
@@ -15,63 +15,207 @@ import (
...
-			mockReader.EXPECT().BlockByHash(utils.HexToFelt(t, test.hash)).DoAndReturn(func(_ *felt.Felt) (block *core.Block, err error) {
+			mockReader.EXPECT().BlockByHash(utils.HexToFelt(t, test.blockHash)).DoAndReturn(func(_ *felt.Felt) (block *core.Block, err error) {
 				return mockReader.BlockByNumber(test.blockNumber)
 			}).Times(2)
+
+			// Test it returns an error if no feeder client is set
 			handler := rpc.New(mockReader, nil, nil, "", nil)
 			_, httpHeader, jErr := handler.TraceBlockTransactions(context.Background(), rpc.BlockID{Number: test.blockNumber})
 			require.Equal(t, rpccore.ErrInternal.Code, jErr.Code)
 			assert.Equal(t, httpHeader.Get(rpc.ExecutionStepsHeader), "0")
```
rianhughes, line 112: I don't think we need to run this multiple times. Maybe we can move it to a seperate test and run it once?

rianhughes (approved):
Approving since my comments are small nits, and the CI/CD e2e tests failed from the known issue where blockifier doesn't return non-zero steps etc.

PR #2523, opened 2025-02-19
Add `ExecutionError`

AnkushinDaniil:
> [!IMPORTANT]
> Review after #2522  

#### Summary
This PR introduces a new error handling mechanism and refactors the existing code.

#### Changes
1. **New Error Type**:
   - Added `ExecutionError` enum to encapsulate different error types, including `TransactionExecutionError` and `CustomError`

2. **Error Handling Refactoring**:
   - Implemented `fmt::Display` and `From` traits for `ExecutionError` and `SimulationError` to provide meaningful error messages and conversions.
   - Simplified the error propagation using the `?` operator and the `From` trait implementations.

3. **Function Updates**:
   - Updated `execute_transaction` to use the new `ExecutionError` type.
   - Refactored `is_l2_gas_accounting_enabled`, `get_gas_vector_computation_mode`, `simulate_execution`, `set_l2_gas_limit`, and `extract_l2_gas_limit` functions to handle errors using the new error types.

PR #2601, opened 2025-03-03
fix(rpc): failed json marshalling in starknet_call rpc v6-7

### LINKED ISSUE #2599: JSON Marshaling Error in starknet_call
When making a `starknet_call` request (v0.6 and v0.7) and expecting error message, the server responds with a 500 Internal Server Error and Juno logs show the following error:

```
13:54:56.288 03/03/2025 +00:00 ERROR jsonrpc/http.go:56 Handler failure {"err": "json: error calling MarshalJSON for type json.RawMessage: invalid character 'x' after top-level value"}
```

**Sample Request on sepolia:**
```
{
    "jsonrpc": "2.0",
    "method": "starknet_call",
    "params": {
        "request": {
            "contract_address": "0x78d8ada537ab1132e600b7553c3e574336c4e0f24fb584b8b2d5faff8a37590",
            "entry_point_selector": "0x11c957afc7e84fbf8aac7ff9c79910a53a64edf9a779b18b411fbb1982fb84f",
            "calldata": []
        },
        "block_id": "pending"
    },
    "id": 1
}
```

Expected:
```
{
    "jsonrpc": "2.0",
    "error": {
        "code": 40,
        "message": "Contract error",
        "data": {
            "revert_error": "execution failed"
        }
    },
    "id": 1
}
```

**Root Cause:**

I guess it occurs in the error handling of the [Call ](https://github.com/NethermindEth/juno/blob/f81a84f6496262b09b787d951c8edeea63307606/rpc/v7/trace.go#L394) method. When a contract call fails, the code attempts to create a JSON error response by directly converting a string to a [json.RawMessage](https://github.com/NethermindEth/juno/blob/f81a84f6496262b09b787d951c8edeea63307606/rpc/v7/trace.go#L434)

This issue was likely introduced in PR #2523

AnkushinDaniil:
Fixes #2599
[CI/CD](https://github.com/NethermindEth/juno/actions/runs/13634961985)

Since we started using `json.RawMessage` in [this](https://github.com/NethermindEth/juno/pull/2523) PR, and `blockifier` returns `0x454e545259504f494e545f4e4f545f464f554e44` in the case of `ENTRYPOINT_NOT_FOUND`, we should handle this case explicitly.

### DIFF
--- a/rpc/v6/trace.go
+++ b/rpc/v6/trace.go
@@ -304,7 +304,7 @@ func (h *Handler) Call(funcCall *FunctionCall, id *BlockID) ([]*felt.Felt, *json
 	if res.ExecutionFailed {
 		var strErr string
 		if len(res.Result) != 0 && res.Result[0].String() == rpccore.EntrypointNotFoundFelt {
-			strErr = `"execution failed"`
+			strErr = rpccore.ExecutionFailed
 		} else {
 			strErr = `"` + utils.FeltArrToString(res.Result) + `"`
 		}
--- a/rpc/v6/trace.go
+++ b/rpc/v6/trace.go
@@ -303,10 +303,12 @@ func (h *Handler) Call(funcCall *FunctionCall, id *BlockID) ([]*felt.Felt, *json
 	}
 	if res.ExecutionFailed {
 		var strErr string
-		if len(res.Result) != 0 && res.Result[0].String() == rpccore.EntrypointNotFoundFelt {
-			strErr = rpccore.ExecutionFailed
-		} else {
-			strErr = `"` + utils.FeltArrToString(res.Result) + `"`
+		if len(res.Result) != 0 {
+			if res.Result[0].String() == rpccore.EntrypointNotFoundFelt {
+				strErr = rpccore.ExecutionFailed
+			} else {
+				strErr = `"` + utils.FeltArrToString(res.Result) + `"`
+			}
 		}
 		return nil, MakeContractError(json.RawMessage(strErr))
 	}
--- a/rpc/v6/trace.go
+++ b/rpc/v6/trace.go
@@ -302,14 +302,17 @@ func (h *Handler) Call(funcCall *FunctionCall, id *BlockID) ([]*felt.Felt, *json
 		return nil, MakeContractError(json.RawMessage(err.Error()))
 	}
 	if res.ExecutionFailed {
+		// the blockifier 0.13.4 update requires us to check if the execution failed,
+		// and if so, return ErrEntrypointNotFound if res.Result[0]==EntrypointNotFoundFelt,
+		// otherwise we should wrap the result in ErrContractError
 		var strErr string
 		if len(res.Result) != 0 {
 			if res.Result[0].String() == rpccore.EntrypointNotFoundFelt {
-				strErr = rpccore.ExecutionFailed
-			} else {
-				strErr = `"` + utils.FeltArrToString(res.Result) + `"`
+				return nil, rpccore.ErrEntrypointNotFound
 			}
+			strErr = `"` + utils.FeltArrToString(res.Result) + `"`
 		}
+		// Todo: There is currently no standardised way to format these error messages
 		return nil, MakeContractError(json.RawMessage(strErr))
 	}
 	return res.Result, nil
--- a/rpc/v6/trace.go
+++ b/rpc/v6/trace.go
@@ -302,7 +302,13 @@ func (h *Handler) Call(funcCall *FunctionCall, id *BlockID) ([]*felt.Felt, *json
 		return nil, MakeContractError(json.RawMessage(err.Error()))
 	}
 	if res.ExecutionFailed {
-		return nil, MakeContractError(json.RawMessage(utils.FeltArrToString(res.Result)))
+		var strErr string
+		if len(res.Result) != 0 && res.Result[0].String() == rpccore.EntrypointNotFoundFelt {
+			strErr = `"execution failed"`
+		} else {
+			strErr = `"` + utils.FeltArrToString(res.Result) + `"`
+		}
+		return nil, MakeContractError(json.RawMessage(strErr))
 	}
 	return res.Result, nil
 }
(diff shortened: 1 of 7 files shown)


rodrigo-pino (commented):
Tests that checks this for both v6 and v7 version please. Preferably they check all the execution paths of the new code

rpc/v6/trace_test.go
```diff
@@ -490,7 +490,7 @@ func TestCall(t *testing.T) {
 			Result:          []*felt.Felt{utils.HexToFelt(t, rpccore.EntrypointNotFoundFelt)},
 			ExecutionFailed: true,
 		}
-		expectedErr := rpc.MakeContractError(json.RawMessage(rpccore.EntrypointNotFoundFelt))
+		expectedErr := rpc.MakeContractError(json.RawMessage(rpccore.ExecutionFailed))
```
rodrigo-pino, line 493: I am a bit confused by this. Shouldn't there errors which failed due `EntryPointNotFound` return this type of error as well? @rianhughes

rodrigo-pino:
We really need this testing infra, every time I see we change the expected value of a unit test I am very cautious. It means we are testing what we think it should be the result and not what the result is.

rpc/v7/trace.go
```diff
@@ -431,7 +431,13 @@ func (h *Handler) Call(funcCall FunctionCall, id BlockID) ([]*felt.Felt, *jsonrp
 		return nil, MakeContractError(json.RawMessage(err.Error()))
 	}
 	if res.ExecutionFailed {
-		return nil, MakeContractError(json.RawMessage(utils.FeltArrToString(res.Result)))
+		var strErr string
+		if len(res.Result) != 0 && res.Result[0].String() == rpccore.EntrypointNotFoundFelt {
+			strErr = rpccore.ExecutionFailed
+		} else {
+			strErr = `"` + utils.FeltArrToString(res.Result) + `"`
```
rianhughes, line 438: Technically this will override empty errors with non-zero data. It shouldn't really cause a problem, but might be worth fixing now just in case

rianhughes: Similar with the v6 case

rpc/v6/trace_test.go
rianhughes: I'm not entirely sure what you mean. When the user calls an entry point that doesn't exist blockifier returns 0x454e545259504f494e545f4e4f545f464f554e44 in the result (along with an indicator that the execution failed). To be consistent with the old rpc specs, we need to override this error

rpc/v6/trace.go
```diff
@@ -391,7 +391,13 @@ func (h *Handler) Call(funcCall *FunctionCall, id *BlockID) ([]*felt.Felt, *json
...
 	}
 	if res.ExecutionFailed {
-		return nil, MakeContractError(json.RawMessage(utils.FeltArrToString(res.Result)))
+		var strErr string
+		if len(res.Result) != 0 && res.Result[0].String() == rpccore.EntrypointNotFoundFelt {
+			strErr = rpccore.ExecutionFailed
+		} else {
+			strErr = `"` + utils.FeltArrToString(res.Result) + `"`
+		}
+		return nil, MakeContractError(json.RawMessage(strErr))
```
rianhughes, line 400: @AnkushinDaniil do you know why the old version was returning the json marshalling error? I don't understand why it just doesn't return "contract error : entrypoint not found"

rianhughes: Imo, we shouldn't "entrypoint not found" this with "execution failed", its bad UX. Do we need to do this to not break the old SDK tests??

rpc/v6/trace_test.go
rianhughes: I'm not sure what blockifier used to return in this case tbh

rpc/v7/trace.go
AnkushinDaniil: [Handle empty data case](https://github.com/NethermindEth/juno/pull/2601/commits/bce16018ec65746b11bfbe568a700ff1d0306756)

rpc/v6/trace.go
AnkushinDaniil: [Use ErrEntrypointNotFound in v6 and v7](https://github.com/NethermindEth/juno/pull/2601/commits/1b4e3d749f8899d33c71660f1b9fa415bd723282)

wojciechos: > CI/CD results for [this branch](https://github.com/NethermindEth/juno/actions/runs/13653179577) and [main](https://github.com/NethermindEth/juno/actions/runs/13654408748/job/38170825978) are the same

Looks like the tests fail on both not due to invalid JSON respone but because of an internal error - likely still causing a 500 server error because of some ERRs on juno side(?)


> Imo, we shouldn't override "entrypoint not found" with "execution failed", its bad UX. Do we need to do this to not break the old SDK tests??

We’ve been returning "execution failed" before, so there’s a tiny chance that someone relies on it. However, it's still within the spec, we should be fine moving with it. So to be sure, for v6-7 do you mean something like this, right?
```
{
    "jsonrpc": "2.0",
    "error": {
        "code": 40,
        "message": "Contract error",
        "data": {
            "revert_error": "entrypoint not found"
        }
    },
    "id": 1
}
```

wojciechos: I deployed the version from this branch to dev, and there is no ERR and 500 anymore but the response for v0_7 and v0_8 is:
```
{
    "jsonrpc": "2.0",
    "error": {
        "code": 21,
        "message": "Requested entrypoint does not exist in the contract"
    },
    "id": 1
}
```
This does not match the v0_6-v0_7 spec making starknet.go tests fail, so we can't break compatibility changing that behavior. 

Maybe response like below should work for v0.6 - v.0.7? 
```
{
    "jsonrpc": "2.0",
    "error": {
        "code": 40,
        "message": "Contract error",
        "data": {
            "revert_error": "Requested entrypoint does not exist in the contract"
        }
    },
    "id": 1
}
```

Reference: [Starknet RPC Spec](https://github.com/starkware-libs/starknet-specs/blob/v0.7.1/api/starknet_api_openrpc.json#L599).

AnkushinDaniil: [Update error](https://github.com/NethermindEth/juno/pull/2601/commits/29dc1e8f7f345f3313397eeff32593e97095734a)

rianhughes (approved):

rodrigo-pino (approved):

PR #2605, opened 2025-03-04
chore(rpc/v8): use parsed logger

weiihann:


weiihann:
.......can I please bypass this code coverage..............

Issue #2610, opened 2025-03-04
Error sending v3 invoke transaction: "unknown resource"


thiagodeev:
I tried to send this invoke v3 txn to Sepolia with Juno, but it's returning an error. I was able to simulate this same transaction by passing the `SKIP_VALIDATE` flag without any problem, but I'm not being able to send it.

Juno's version: compiled from main branch
URL: localhost:6060/rpc/v0_8

Request:
``` json
{
	"jsonrpc": "2.0",
	"method": "starknet_addInvokeTransaction",
	"params": [
		{
			"account_deployment_data": [],
			"calldata": [
				"0x1",
				"0x669e24364ce0ae7ec2864fb03eedbe60cfbc9d1c74438d10fa4b86552907d54",
				"0x2f0b3c5710379609eb5495f1ecd348cb28167711b73609fe565a72734550354",
				"0x2",
				"0xffffffff",
				"0x0"
			],
			"fee_data_availability_mode": "L1",
			"nonce": "0x2a",
			"nonce_data_availability_mode": "L1",
			"paymaster_data": [],
			"resource_bounds": {
				"l1_data_gas": {
					"max_amount": "0x1e0",
					"max_price_per_unit": "0x2fd4c7f850021"
				},
				"l1_gas": {
					"max_amount": "0x0",
					"max_price_per_unit": "0x1f9f53d0b24e7"
				},
				"l2_gas": {
					"max_amount": "0x1519e0",
					"max_price_per_unit": "0x18c6dbde3"
				}
			},
			"sender_address": "0x36d67ab362562a97f9fba8a1051cf8e37ff1a1449530fb9f1f0e32ac2da7d06",
			"signature": [
				"0x201d6db3eeb8f59986668bf9dc61d59c8302d4f2ae9a4ae30de63c2b959368f",
				"0x2f202229c7a41e71d62f547fe3f4359139cfbe095eeba29ecaa34d0e64ffc04"
			],
			"tip": "0x0",
			"type": "INVOKE",
			"version": "0x3"
		}
	],
	"id": 4
}
```

Response:
``` json
{
	"jsonrpc": "2.0",
	"error": {
		"code": -32603,
		"message": "Internal error",
		"data": "marshal transaction: json: encoding error for type \"map[starknet.Resource]starknet.ResourceBounds\": \"unknown resource\""
	},
	"id": 4
}
```

PR #2611, opened 2025-03-04
Add 'L1_DATA_GAS' missing field from marshal func of the 'starknet.Resource' type

thiagodeev:
Solves https://github.com/NethermindEth/juno/issues/2610

It was a missing field from the `starknet.Resource` type. The `rpc.Resource` was correct, but the `starknet.Resource` was missing a field.


starknet/transaction_test.go
```diff
@@ -43,3 +43,88 @@ func TestUnmarshalFinalityStatus(t *testing.T) {
...
+			name:     "error",
+			resource: starknet.Resource(0),
+			err:      true,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := tt.resource.MarshalText()
+			if tt.err {
+				require.Error(t, err)
```
hudem1, line 79: nit: Really a small detail but we could almost verify the error message is the one expected as well (same for the fct `TestResourceUnmarshalJSON` below)

hudem1 (approved):

starknet/transaction_test.go
thiagodeev: Sure, I'll do it.
But I just copied these tests from the ones in the `rpc` pkg, so they are in the same way haha.

Also, I'm currently change the `TestAddTransaction` from the `rpc` pkg. It's out of date, which is why this bug went unnoticed.

rianhughes (approved):

starknet/transaction_test.go
hudem1: Oh okay we should update the test there too then!

thiagodeev: Done

AnkushinDaniil (approved):

core/transaction_test.go
```diff
@@ -219,17 +219,16 @@ func TestVerifyTransactionHash(t *testing.T) {
...
 	ctx := context.Background()
 
-	// Test data was obtained by playing with Papyrus's implementation before v0.13 was released on integration.
 	tests := map[string]struct {
 		tx   func(hash *felt.Felt) core.Transaction
 		want *felt.Felt
 	}{
 		"invoke": {
-			// https://external.integration.starknet.io/feeder_gateway/get_transaction?transactionHash=0x49728601e0bb2f48ce506b0cbd9c0e2a9e50d95858aa41463f46386dca489fd
+			// https://alpha-sepolia.starknet.io/feeder_gateway/get_transaction?transactionHash=0x76b52e17bc09064bd986ead34263e6305ef3cecfb3ae9e19b86bf4f1a1a20ea
```
rodrigo-pino, line 231: Why did we changed the test data? Could we delete the old test data?

thiagodeev: This test and the others were using Integration data, which is inaccessible and deprecated, so we can't update them with new Integration txns.
Some tests include transactions of all versions, and it would be a bit difficult and slow to find an example of each txn type and version in Sepolia. For these tests, I just updated the data with the new `L1_DATA_GAS` field, but for the tests like this one above, that required only v3 txns, it was easier and faster to update the entire test data (which is the best approach IMO).

>Could we delete the old test data?

No, it's being used in other tests. That's why I updated them to have the new `L1_DATA_GAS` field, so that we could at least check if the txns are being Marshaled/Unmarshaled correctly using the same outdated data.

PR #2612, opened 2025-03-05
refactor: Use a common function for send subscription response

infrmtcs:


rpc/v8/subscriptions.go
```diff
@@ -757,13 +707,17 @@ type SubscriptionTransactionStatus struct {
 }
 
 // sendTxnStatus creates a response and sends it to the client
-func (h *Handler) sendTxnStatus(w jsonrpc.Conn, status SubscriptionTransactionStatus, id uint64) error {
+func sendTxnStatus(w jsonrpc.Conn, status SubscriptionTransactionStatus, id uint64) error {
+	return sendResponse("starknet_subscriptionTransactionsStatus", w, id, status)
+}
+
+func sendResponse(method string, w jsonrpc.Conn, id uint64, result any) error {
```
rodrigo-pino, line 714: Could we use generics here 👀 . Is it worth it?

rodrigo-pino (approved):
Amazing!

rpc/v8/subscriptions.go
```diff
@@ -579,21 +566,8 @@ func (h *Handler) filterTxBySender(txn core.Transaction, senderAddr []felt.Felt)
...
-			"result":          result,
-		},
-	})
-	if err != nil {
-		return err
-	}
-
-	_, err = w.Write(resp)
-	return err
+func sendPendingTxs(w jsonrpc.Conn, result any, id uint64) error {
```
hudem1, line 569: Just out of curiosity, why changing all those send functions as standalone fcts and not methods on Handler ?

hudem1 (approved):

rpc/v8/subscriptions.go
infrmtcs: I think currently not worth it, because there are only 1 argument with type T and it's not constrained by any interface.

rpc/v8/subscriptions.go
infrmtcs: In general, a method is just a function with a receiver argument. When we write a function, if it doesn't need any argument, we should remove the argument. In this case the unnecessary argument is the receiver itself, so I removed it.

hudem1: Okok i see, makes sense!

PR #2617, opened 2025-03-05
fix: Reverted INVOKE tx returns a non-nil ExecuteInvocation in trace

### LINKED ISSUE #2616: Missing execute_invocation in reverted starknet_traceTransaction response
After [PR #2591](https://github.com/NethermindEth/juno/pull/2591), the [starknet.rs test ](https://github.com/NethermindEth/juno/actions/runs/13674190195/job/38231620962#step:4:511)**`jsonrpc_trace_invoke_reverted`** started failing:

- The response for `starknet_traceTransaction` is missing the **`execute_invocation`** 
- According to the Starknet specification, this field **should be present** in the response - [spec for reference](https://github.com/starkware-libs/starknet-specs/blob/v0.7.1/api/starknet_trace_api_openrpc.json#L204)
- On the released version before this PR, `execute_invocation` was included correctly.

## Example Request on Sepolia:
```json
[
    {
        "jsonrpc": "2.0",
        "method": "starknet_traceTransaction",
        "params": {
            "transaction_hash": "02f00c7f28df2197196440747f97baa63d0851e3b0cfc2efedb6a88a7ef78cb1"
        },
        "id": 1
    }
]
```

## Current Behavior (After PR #2591):
- The response **does not** include `execute_invocation`.

hudem1:
Fixes #2616

_Bug explanation_
The `FunctionInvocation` field in the returned trace from the feeder is set to nil when the tx gets reverted

(I added some feeder testdata with a block containing a reverted tx, hence the new files!)

### DIFF
--- a/clients/feeder/testdata/sepolia/block/18.json
+++ b/clients/feeder/testdata/sepolia/block/18.json
@@ -1,0 +1,737 @@
+{
+    "block_hash": "0x5beb56c7d9a9fc066e695c3fc467f45532cace83d9979db4ccfd6b77ca476af",
+    "parent_block_hash": "0x641077df21f3d3167e62ddf3feea236e2644e427940673114a30f2eb5c445c6",
+    "block_number": 18,
+    "state_root": "0x4c7f8d53ff361055ca6b224bff96bae036e1694c7d0efd0bef5fc633eef720d",
+    "transaction_commitment": "0x5eba4f898340f70f68337edb05f13b2c613e4523d61de7baec8e3d273d84afb",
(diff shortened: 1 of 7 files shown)


hudem1:
Question also!

The `httpHeader` is correctly set when the trace is returned by the VM, see:
```go
httpHeader.Set(ExecutionStepsHeader, strconv.FormatUint(executionResult.NumSteps, 10))
```

But it seems set to 0 when returned by the feeder. Shouldn't we set it there too ? What is it used for actually ?

rodrigo-pino (approved):

rodrigo-pino:
Bypassing rules and going ahead with merge since is quite small and fix e2e tests

PR #2613, opened 2025-03-05
fix: Avoid race condition between reorg and head in subscription

infrmtcs:

PR #2619, opened 2025-03-06
fix(rpc/v8): wrong tx status subscription hang for a long time

infrmtcs:
Depends on #2613


clients/feeder/testdata/sepolia-integration/transaction/0x1111.json
```diff
@@ -1,7 +1,7 @@
 {
     "revert_error": "some error",
-    "execution_status": "REJECTED",
-    "finality_status": "RECEIVED",
+    "finality_status": "ACCEPTED_ON_L2",
+    "execution_status": "REVERTED",
```
infrmtcs, line 4: Mempool is not available yet, so technically Rejected is not possible. We should instead test the Reverted case for now.

rpc/v8/subscriptions.go
```diff
@@ -216,35 +216,41 @@ func (h *Handler) SubscribeEvents(ctx context.Context, fromAddr *felt.Felt, keys
...
+		onPending: func(ctx context.Context, id uint64, sub *subscription, pending *core.Block) error {
 			return h.processEvents(ctx, w, id, nextBlock, nextBlock, fromAddr, keys, eventsPreviouslySent)
 		},
 	}
 	return h.subscribe(ctx, w, subscriber)
 }
 
+type errorTxnHashNotFound struct {
+	txHash felt.Felt
+}
```
rodrigo-pino, line 245: one suggestion
```suggestion
type errorTxnHashNotFound felt.Felt
```

One nit: move it to the top of the file where we usually have all the definitions

rpc/v8/subscriptions.go
```diff
@@ -253,138 +259,79 @@ func (h *Handler) SubscribeTransactionStatus(ctx context.Context, txHash felt.Fe
...
-				curStatus, rpcErr = h.TransactionStatus(ctx, txHash)
-				if rpcErr != nil {
-					if rpcErr != rpccore.ErrTxnHashNotFound {
-						return 0, rpcErr
-					}
-					continue
-				}
-				timeout.Stop()
-				break txNotFoundLoop
+	return h.subscribe(ctx, w, subscriber{
```
rodrigo-pino, line 264: nit: subscriber defined on top

rpc/v8/subscriptions.go
```diff
@@ -253,138 +259,79 @@ func (h *Handler) SubscribeTransactionStatus(ctx context.Context, txHash felt.Fe
...
-	id := h.idgen()
-	subscriptionCtx, subscriptionCtxCancel := context.WithCancel(ctx)
-	sub := &subscription{
-		cancel: subscriptionCtxCancel,
-		conn:   w,
+// If the error is transaction not found that means the transaction has not been submitted to the feeder gateway,
+// therefore, we need to wait for a specified time and at regular interval check if the transaction has been found.
+// If the transaction is found during the timout expiry, then we continue to keep track of its status otherwise the
+// websocket connection is closed after the expiry.
+func (h *Handler) getInitialTxStatus(ctx context.Context, sub *subscription, id uint64, txHash felt.Felt, lastStatus *TxnStatus) error {
```
rodrigo-pino, line 293: `txHash` should be passed as a reference, unless you intentionally didn't.

`*txStatus` should be passed by value

rpc/v8/subscriptions_test.go
```diff
@@ -1087,11 +1000,24 @@ func assertNextMessage(t *testing.T, conn net.Conn, id SubscriptionID, e *Emitte
 	assert.Equal(t, string(resp), string(got))
 }
 
-func assertNextMessages(t *testing.T, conn net.Conn, id SubscriptionID, emittedEvents []*EmittedEvent) {
+func assertNextTxnStatus(t *testing.T, conn net.Conn, id SubscriptionID, txHash felt.Felt, finality TxnStatus, execution TxnExecutionStatus, failureReason string) {
```
rodrigo-pino, line 1003: Because this is a test it is not immportant, but txHash should be a `*felt.Felt`

clients/feeder/testdata/sepolia-integration/transaction/0x1111.json
rodrigo-pino: I am a bit out of context here. Could you provide more information so I know what you mean?

rodrigo-pino (commented):
All looks good

clients/feeder/testdata/sepolia-integration/transaction/0x1111.json
infrmtcs: After I double checked again, Rejected is still available, so I revert this change and add a new test for reverted case.

rpc/v8/subscriptions.go
rodrigo-pino: Yes, I agree it is usually worth it to be "pure". Makes code easier to read in the long term. Thank you!

rpc/v8/subscriptions.go
```diff
@@ -216,24 +224,24 @@ func (h *Handler) SubscribeEvents(ctx context.Context, fromAddr *felt.Felt, keys
...
 			if err := h.processEvents(ctx, w, id, nextBlock, head.Number, fromAddr, keys, eventsPreviouslySent); err != nil {
 				return err
 			}
 			nextBlock = head.Number + 1
 			return nil
 		},
-		onPending: func(ctx context.Context, id uint64, pending *core.Block) error {
+		onPending: func(ctx context.Context, id uint64, sub *subscription, pending *core.Block) error {
 			return h.processEvents(ctx, w, id, nextBlock, nextBlock, fromAddr, keys, eventsPreviouslySent)
 		},
```
rodrigo-pino, line 246: In this case can we set `sub` as an `_` since it is not meant to be used

rodrigo-pino (approved):

rpc/v8/subscriptions_test.go
```diff
@@ -1087,11 +1017,24 @@ func assertNextMessage(t *testing.T, conn net.Conn, id SubscriptionID, e *Emitte
 	assert.Equal(t, string(resp), string(got))
 }
 
-func assertNextMessages(t *testing.T, conn net.Conn, id SubscriptionID, emittedEvents []*EmittedEvent) {
+func assertNextTxnStatus(t *testing.T, conn net.Conn, id SubscriptionID, txHash *felt.Felt, finality TxnStatus, execution TxnExecutionStatus, failureReason string) {
+	t.Helper()
+
+	assertNextMessage(t, conn, id, "starknet_subscriptionTransactionsStatus", SubscriptionTransactionStatus{
```
weiihann, line 1023: As per the spec, the subscription response name should be `starknet_subscriptionTransactionStatus`. Please change it in `sendTxStatus` as well.

weiihann (changes requested):

weiihann (approved):

PR #2620, opened 2025-03-07
feat(rpc): add better infrastructure for testing

AnkushinDaniil:
This PR introduces infrastructure for testing RPC endpoints using the real VM instead of mocks.

- Add `Cairo` contracts for testing:
  - Universal deployer
  - Account contract
  - ERC20 token contract
  - Contract for binary search testing
- Implement `test_blockchain` for creating a blockchain with predeployed contracts and preset balances
- Add tests for `starknet_estimateFee` using real VM execution


cairo/Makefile
rodrigo-pino, line 1: I understand that idea of this file is to make life easier, but not everyone uses the same install methods. Let's remove it and let everyone else to the Cairo Compiler/Scarb their own way

Makefile
```diff
@@ -179,3 +179,9 @@ pathfinder: juno-cached ## Run a node to sync from pathfinder feedernode. P2P us
 
 test-fuzz: ## Run fuzzing script
 	./scripts/fuzz_all.sh
+
+cairo: ## Compile Cairo to Sierra
+	$(MAKE) -C cairo build
+
+cairo-clean: ## Clean Sierra files
+	$(MAKE) -C cairo clean
```
rodrigo-pino, line 187: Let's not expose this command

rodrigo-pino (commented):
I think this PR represents a good approximation of what we want to do. There are only two things which I think shouldn't go in:

1. All contract defintion shouldn't depend on Scarb
2. All contracts shouldn't be under the same Scarb project umbrella

Now, other minor thing, every compilation artifact which is not a .json or a .sierra should be deleted, unless they hinder the program.

cairo/scarb/target/CACHEDIR.TAG
```diff
@@ -0,0 +1,3 @@
+Signature: 8a477f597d28d172789f06886806bc55
```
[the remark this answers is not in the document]
AnkushinDaniil: [Remove CACHEDIR.TAG](https://github.com/NethermindEth/juno/pull/2620/commits/0d6a9a093f3a7e02acacfc4f4dc0841b60ac3dc0)

blockchain/test_blockchain.go
```diff
@@ -0,0 +1,269 @@
...
+	"github.com/NethermindEth/juno/db"
+	"github.com/NethermindEth/juno/db/pebble"
+	"github.com/NethermindEth/juno/starknet"
+	"github.com/NethermindEth/juno/starknet/compiler"
+	"github.com/NethermindEth/juno/utils"
+	"github.com/stretchr/testify/require"
+)
+
+type testBlockchain struct {
+	*Blockchain
```
rodrigo-pino, line 22: Why `*Blockchain`, is that field optional?

blockchain/test_blockchain.go
```diff
@@ -0,0 +1,269 @@
...
+	t                        *testing.T
+	account, deployer, erc20 *TestClass
+	addr2classHash           map[felt.Felt]*felt.Felt
```
rodrigo-pino, line 25: Why the value type is a pointer of `*felt.Felt`

blockchain/test_blockchain.go
```diff
@@ -0,0 +1,269 @@
...
+	}
+	genesisStateUpdate := &core.StateUpdate{
+		BlockHash: genesisBlock.Hash,
+		NewRoot:   &felt.Zero,
+		OldRoot:   &felt.Zero,
+		StateDiff: &core.StateDiff{},
+	}
+	require.NoError(t, chain.Store(genesisBlock, &core.BlockCommitments{}, genesisStateUpdate, nil))
+
+	prefix := "../../cairo/scarb/target/dev/"
```
rodrigo-pino, line 56: constant

blockchain/test_blockchain.go
```diff
@@ -0,0 +1,269 @@
...
+)
+
+type testBlockchain struct {
+	*Blockchain
+	t                        *testing.T
+	account, deployer, erc20 *TestClass
+	addr2classHash           map[felt.Felt]*felt.Felt
+}
+
+func NewTestBlockchain(t *testing.T, protocolVersion string) *testBlockchain {
```
rodrigo-pino, line 28: Why are you returning a `*testBlockchain` instead of `testBlockchain`?

blockchain/test_blockchain.go
```diff
@@ -0,0 +1,269 @@
...
+	"github.com/NethermindEth/juno/starknet"
+	"github.com/NethermindEth/juno/starknet/compiler"
+	"github.com/NethermindEth/juno/utils"
+	"github.com/stretchr/testify/require"
+)
+
+type testBlockchain struct {
+	*Blockchain
+	t                        *testing.T
+	account, deployer, erc20 *TestClass
```
rodrigo-pino, line 24: Are they optional fields, shared fields? Why are they pointers and not the values itself?

blockchain/test_blockchain.go
```diff
@@ -0,0 +1,269 @@
...
+		root, err = core.NewTestState(txn).RootWithStateUpdate(blockNumber, stateUpdate, newClasses)
+		return err
+	}))
+	return root
+}
+
+type testAccount struct {
+	t       *testing.T
+	address *felt.Felt
+	balance *felt.Felt
```
rodrigo-pino, line 116: Are this field optional or shared between other variables? If not, please make them value types.

blockchain/test_blockchain.go
```diff
@@ -0,0 +1,269 @@
...
+type TestClass struct {
+	t        *testing.T
+	hash     *felt.Felt
+	class    *core.Cairo1Class
+	accounts []*testAccount
+	snClass  *starknet.SierraDefinition
+	complied *starknet.CompiledClass
+}
+
+func NewClass(t *testing.T, path string) *TestClass {
```
rodrigo-pino, line 132: Why this `TestClass` constructor is returning a pointer?

blockchain/test_blockchain.go
```diff
@@ -0,0 +1,269 @@
...
+		balance: balance,
+	})
+}
+
+func (b *testBlockchain) Prepare(t *testing.T, classes []*TestClass) {
+	t.Helper()
+
+	newClasses := make(map[felt.Felt]core.Class, len(classes))
+	deployedContracts := make(map[felt.Felt]*felt.Felt)
+	balances := make(map[felt.Felt]*felt.Felt)
```
rodrigo-pino, line 173: This types are ok being `*felt.Felt` for example.

blockchain/test_blockchain.go
```diff
@@ -0,0 +1,269 @@
...
+
+func (b *TestClass) AddAccount(address, balance *felt.Felt) {
+	b.accounts = append(b.accounts, &testAccount{
+		t:       b.t,
+		address: address,
+		balance: balance,
+	})
+}
+
+func (b *testBlockchain) Prepare(t *testing.T, classes []*TestClass) {
```
rodrigo-pino, line 168: This `classes` is not clear that it should be of type `[]*TestClass` but instead of `[]TestClass`

blockchain/test_blockchain.go
```diff
@@ -0,0 +1,269 @@
...
+	require.NoError(t, json.NewDecoder(file).Decode(intermediate))
+
+	snClass := &starknet.SierraDefinition{
+		Abi:         string(intermediate.Abi),
+		EntryPoints: intermediate.EntryPoints,
+		Program:     intermediate.Program,
+		Version:     intermediate.Version,
+	}
+
+	compliedClass, err := compiler.Compile(snClass)
```
rodrigo-pino, line 246: complied -> compiled

blockchain/test_blockchain.go
```diff
@@ -0,0 +1,269 @@
...
+	require.NoError(t, err)
+
+	class, err := sn2core.AdaptCairo1Class(snClass, compliedClass)
+	require.NoError(t, err)
+
+	return snClass, compliedClass, class
+}
+
+// https://github.com/eqlabs/pathfinder/blob/7664cba5145d8100ba1b6b2e2980432bc08d72a2/crates/common/src/lib.rs#L124
+func fromNameAndKey(t *testing.T, name string, key *felt.Felt) *felt.Felt {
```
rodrigo-pino, line 256: Why are you returning a reference type?

blockchain/test_blockchain.go
```diff
@@ -0,0 +1,269 @@
...
+	t.Helper()
+
+	intermediate := crypto.StarknetKeccak([]byte(name))
+	byteArr := crypto.Pedersen(intermediate, key).Bytes()
+	value := new(big.Int).SetBytes(byteArr[:])
+
+	maxAddr, ok := new(big.Int).SetString("0x7ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff00", 0)
+	require.True(t, ok)
```
rodrigo-pino, line 264: Why are you using `BigInt` here, can the maximum address be bigger than a felt?

cairo/compiler/src/binary_search.cairo
```diff
@@ -0,0 +1,137 @@
+/// Simple contract for managing balance.
```
rodrigo-pino, line 1: Add a comment explaining the repo where it was originally taken this code from

blockchain/test_blockchain.go
AnkushinDaniil: [Remove pointer](https://github.com/NethermindEth/juno/pull/2620/commits/bfcda25fa4d21208506c86b33b5df82d7675cc15)

blockchain/test_blockchain.go
AnkushinDaniil: [Remove pointers](https://github.com/NethermindEth/juno/pull/2620/commits/1a7742f3fabc80215b7d176e46580039eb8b04f4)

blockchain/test_blockchain.go
AnkushinDaniil: [Remove pointers](https://github.com/NethermindEth/juno/pull/2620/commits/1a7742f3fabc80215b7d176e46580039eb8b04f4)

blockchain/test_blockchain.go
AnkushinDaniil: [Remove pointers](https://github.com/NethermindEth/juno/pull/2620/commits/1a7742f3fabc80215b7d176e46580039eb8b04f4)

blockchain/test_blockchain.go
AnkushinDaniil: What do you mean?

blockchain/test_blockchain.go
AnkushinDaniil: [Remove pointers](https://github.com/NethermindEth/juno/pull/2620/commits/1a7742f3fabc80215b7d176e46580039eb8b04f4)

blockchain/test_blockchain.go
AnkushinDaniil: [Remove pointers](https://github.com/NethermindEth/juno/pull/2620/commits/1a7742f3fabc80215b7d176e46580039eb8b04f4)

blockchain/test_blockchain.go
AnkushinDaniil: [Add const](https://github.com/NethermindEth/juno/pull/2620/commits/9b15fcdea0697e9dbcacaefccc54c7d461dfc5ae)

blockchain/test_blockchain.go
AnkushinDaniil: [Remove testBlockchain pointer](https://github.com/NethermindEth/juno/pull/2620/commits/38e6d42c8faea7a0370dfd5d6e04709a8587fd4f)

blockchain/test_blockchain.go
AnkushinDaniil: [Remove pointers](https://github.com/NethermindEth/juno/pull/2620/commits/1a7742f3fabc80215b7d176e46580039eb8b04f4)

blockchain/test_blockchain.go
AnkushinDaniil: [Fix typo](https://github.com/NethermindEth/juno/pull/2620/commits/aa856c9f720df1af0f2c989d44f0eaeea617789e)

blockchain/test_blockchain.go
AnkushinDaniil: I use it for the `Rem` method

.gitignore
```diff
@@ -18,3 +18,4 @@ MANIFEST-*
 OPTIONS-*
 CURRENT
 LOCK
+*/CACHEDIR.TAG
```
rodrigo-pino, line 21: I think we should keep this tag it might help us later with github actions to check if they were any changes in the repo.

blockchain/test_blockchain.go
```diff
@@ -0,0 +1,273 @@
...
+	)
+	chain.erc20 = NewClass(t, prefix+"juno_ERC20Upgradeable.contract_class.json")
+	chain.erc20.AddAccount(
+		*utils.HexToFelt(t, "0x049d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7"),
+		felt.Felt{},
+	)
+	chain.erc20.AddAccount(
+		*utils.HexToFelt(t, "0x04718f5a0fc34cc1af16a1cdee98ffb20c31f5cd61d6ab07201858f4287c938d"),
+		felt.Felt{},
+	)
```
rodrigo-pino, line 77: Why not use felt.Zero here instead of felt.Felt. No real difference but a bit a more legible

blockchain/test_blockchain.go
```diff
@@ -0,0 +1,273 @@
...
+
+	chain.Prepare(t, []TestClass{chain.account, chain.deployer, chain.erc20})
+
+	return chain
+}
+
+// Use pointers because of the "hugeParam" linter
```
rodrigo-pino, line 84: Using pointers is the right way to do it, no need to specify that it is because of the linter

blockchain/test_blockchain.go
```diff
@@ -0,0 +1,273 @@
...
+
+func (b *testBlockchain) DeployerAddress() felt.Felt {
+	return b.deployer.accounts[0].address
+}
+
+func (b *testBlockchain) AccountClassHash() felt.Felt {
+	return b.account.hash
+}
+
+func (b *testBlockchain) ClassHashByAddress(address felt.Felt) felt.Felt {
```
rodrigo-pino, line 97: `address` should be a `*felt.Felt`

blockchain/test_blockchain.go
```diff
@@ -0,0 +1,273 @@
...
+			L1DAMode:         core.Blob,
+			L1DataGasPrice:   &core.GasPrice{PriceInWei: utils.HexToFelt(t, "0x2"), PriceInFri: utils.HexToFelt(t, "0x2")},
+			L2GasPrice:       &core.GasPrice{PriceInWei: utils.HexToFelt(t, "0x1"), PriceInFri: utils.HexToFelt(t, "0x1")},
+		},
+	}
+
+	require.NoError(t, b.Store(block, &core.BlockCommitments{}, stateUpdate, newClasses))
+}
+
+func classFromFile(t *testing.T, path string) (*starknet.SierraDefinition, *starknet.CompiledClass, *core.Cairo1Class) {
```
rodrigo-pino, line 225: change the order of results to `Cairo1Class`, `SierraDefinition` and `CompiledClass`, that's the compilation order: source file -> sierra file -> casm file.

rodrigo-pino (approved):

cairo/compiler/src/binary_search.cairo
```diff
@@ -0,0 +1,137 @@
...
+#[starknet::contract]
+mod HelloStarknet {
```
infrmtcs, line 3: nit: The module name doesn't align with the file name

utils/test.go
```diff
@@ -32,3 +36,14 @@ func HexToUint64(t testing.TB, hexStr string) uint64 {
...
+func NumToFelt[N constraints.Integer](t testing.TB, n N) *felt.Felt {
+	t.Helper()
+	if n < 0 {
+		t.Fatalf("NumToFelt received a negative number: %v", n)
+	}
+
+	v := uint64(n)
+	f := felt.FromUint64(v)
+	return &f
+}
```
rianhughes, line 49: Why not just use `new(felt.Felt).SetUint64(n)`?

core/state.go
```diff
@@ -223,14 +223,22 @@ func (s *State) Update(blockNumber uint64, update *StateUpdate, declaredClasses
...
 	}
 
+	if err := s.processStateDiff(blockNumber, update, declaredClasses); err != nil {
+		return err
+	}
+
+	return s.verifyStateUpdateRoot(update.NewRoot)
+}
+
+func (s *State) processStateDiff(blockNumber uint64, update *StateUpdate, declaredClasses map[felt.Felt]Class) error {
```
rianhughes, line 233: nit: imo the name should be updated given we also process the new classes. Maybe something like `processStateDiffAndClasses`??

Also, why was this pushed to a seperate function?

blockchain/testchain/utils.go
```diff
@@ -0,0 +1,43 @@
...
+	root := path.Join(path.Dir(b), "..", "..")
+
+	return path.Join(root, "starknet_contracts")
+}
+
+func coreContractGeneralPath() string {
+	return path.Join(contractsFolder(), "core")
+}
+
+func GetCoreSierraContractPath(contractName string) string {
```
rianhughes, line 24: Could we add a small comment stating what these files are for? Eg `sierra contracts generated by scarb from the contracts in XYZ` or something? Similar for GetSierraContractPath

blockchain/testchain/utils.go
```diff
@@ -0,0 +1,43 @@
...
+	return path.Join(
+		coreContractGeneralPath(),
+		"target",
+		"release",
+		fmt.Sprintf("starknet_core_%s.contract_class.json", contractName),
+	)
+}
+
+func contractGeneralPath() string {
```
rianhughes, line 33: nit: can we inline this / put it in GetSierraContractPath? It seems neater than making it an external function, which isn't needed atm

blockchain/testchain/class.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/NethermindEth/juno/core/hash"
+	"github.com/NethermindEth/juno/starknet"
+	"github.com/NethermindEth/juno/starknet/compiler"
+	"github.com/stretchr/testify/require"
+)
+
+type classDefinition struct {
+	t         *testing.T
+	Hash      hash.ClassHash
+	Instances []deployedContract
```
rianhughes, line 21: nit: rename Instances to deployedContracts

rianhughes: similar for methods like AddInstance

blockchain/testchain/testchain.go
```diff
@@ -0,0 +1,173 @@
...
+	var root *felt.Felt
+	require.NoError(t, b.database.Update(func(txn db.Transaction) (err error) {
+		root, err = core.NewTestState(txn).RootWithStateUpdate(blockNumber, stateUpdate, newClasses)
+		return err
+	}))
+	return root
+}
+
+// Adds each class definition to the testchain, it then make sure they are deployed and with the right balance
+func (b *Testchain) Declare(classes ...*classDefinition) {
```
rianhughes, line 112: I don't think we should deviate from the starknet terminology. Declare is specifically for setting classes in the class trie. Deploying Contracts is for setting contracts in the contracts trie. Given we are doing both here I think we should either 1. split these into distinct functions, or 2. update the name to DeclareAndDeploy

blockchain/testchain/testchain.go
```diff
@@ -0,0 +1,173 @@
...
+	"github.com/NethermindEth/juno/utils"
+	"github.com/stretchr/testify/require"
+)
+
+// Test blockchain used
+type Testchain struct {
+	t *testing.T
+	blockchain.Blockchain
+	database    db.DB
+	Account     classDefinition
```
rianhughes, line 22: nit: rename newClassesAndContracts

blockchain/testchain/testchain.go
rianhughes: The current naming related to classDefintion / testnchain.Accounts should be updatd imo, right now it's not as clear as it could be, which could add confusing for people who are unfamiliar with how Starknet works.

blockchain/testchain/testchain.go
```diff
@@ -0,0 +1,173 @@
...
+	Deployer    classDefinition
```
rianhughes, line 23: nit: ContractDeployer / UDC

blockchain/testchain/testchain.go
```diff
@@ -0,0 +1,173 @@
...
+	erc20       classDefinition
```
rianhughes, line 24: Is this the feeToken? Maybe we should have one for ETH and another for STRK? Blockifier typically expects both (eg iirc it first checks the ETH contract before STRK, and if the ETH contract isn't there, it won't check the STRK contract)

rianhughes: Ah, you just used the erc20 contract for both. Cool

blockchain/testchain/testchain.go
```diff
@@ -0,0 +1,173 @@
...
+	chain.Deployer = NewClass(t, GetCoreSierraContractPath("UniversalDeployer"))
+	chain.Deployer.AddInstance(
+		utils.HexTo[address.ContractAddress](t, "0xc02"),
+		&felt.Zero,
+	)
+	chain.erc20 = NewClass(t, GetCoreSierraContractPath("ERC20Upgradeable"))
+	chain.erc20.AddInstance(
+		utils.HexTo[address.ContractAddress](
+			t,
+			"0x049d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7",
```
rianhughes, line 72: Can we push these to consts? Eh ETHContractAddress / STRKAddress etc. Similar with the deployer adress and account address

blockchain/testchain/testchain.go
```diff
@@ -0,0 +1,173 @@
...
+	require.NoError(t, b.database.Update(func(txn db.Transaction) (err error) {
+		root, err = core.NewTestState(txn).RootWithStateUpdate(blockNumber, stateUpdate, newClasses)
+		return err
+	}))
+	return root
+}
+
+// Adds each class definition to the testchain, it then make sure they are deployed and with the right balance
+func (b *Testchain) Declare(classes ...*classDefinition) {
+	// This have to remain the same due to compatibility with old code
```
rianhughes, line 113: typo: This has to

blockchain/testchain/testchain.go
```diff
@@ -0,0 +1,173 @@
...
+	// This should be class hash to class
+	classDefinitions := make(map[felt.Felt]core.Class, len(classes))
+	// This should be: contract address to class hash
+	deployedContracts := make(map[felt.Felt]*felt.Felt) // use pointer because of DeployedContracts field
+	// This should be storage key to balance
+	balances := make(map[felt.Felt]*felt.Felt) // use pointer because of StorageDiff field
+
+	for _, classDef := range classes {
```
rianhughes, line 121: Can we add a comment for this block?

blockchain/testchain/testchain.go
```diff
@@ -0,0 +1,173 @@
...
+	return root
+}
+
+// Adds each class definition to the testchain, it then make sure they are deployed and with the right balance
+func (b *Testchain) Declare(classes ...*classDefinition) {
+	// This have to remain the same due to compatibility with old code
+	// This should be class hash to class
+	classDefinitions := make(map[felt.Felt]core.Class, len(classes))
+	// This should be: contract address to class hash
+	deployedContracts := make(map[felt.Felt]*felt.Felt) // use pointer because of DeployedContracts field
```
rianhughes, line 117: sorry for all the nits!! newContracts / newDeployedContracts

blockchain/testchain/contract.go
```diff
@@ -0,0 +1,50 @@
...
+	return &c.address
+}
+
+func (c *deployedContract) Balance() *felt.Felt {
+	return &c.balance
+}
+
+func (c *deployedContract) BalanceKey() felt.Felt {
+	return feltFromNameAndKey(c.t, "ERC20_balances", c.address.AsFelt())
+}
```
rianhughes, line 29: Nice

core/state.go
rianhughes: Ah, because we use it to calculate the new root in testBlockchain

blockchain/testchain/testchain.go
```diff
@@ -0,0 +1,173 @@
...
+	blockchain.Blockchain
+	database    db.DB
+	Account     classDefinition
+	Deployer    classDefinition
+	erc20       classDefinition
+	ClassHashes map[address.ContractAddress]hash.ClassHash
+}
+
+// Creates a new test chain with a default account, erc20 contracts and universal deployer
+func New(t *testing.T) Testchain {
```
rianhughes, line 29: I like this!

blockchain/testchain/contract.go
```diff
@@ -0,0 +1,50 @@
...
+
+// https://github.com/eqlabs/pathfinder/blob/7664cba5145d8100ba1b6b2e2980432bc08d72a2/crates/common/src/lib.rs#L124
+func feltFromNameAndKey(t *testing.T, name string, key *felt.Felt) felt.Felt {
```
rianhughes, line 32: Could we add a small func doc indicating that this is to get the kehy of a contracts storage, or something?

blockchain/testchain/class.go
```diff
@@ -0,0 +1,83 @@
...
+
+	return classDefinition{
+		t:      t,
+		Hash:   hash.ClassHash(*classHash),
+		Class:  classDef,
+		Sierra: sierra,
+	}
+}
+
+func (t *classDefinition) AddInstance(address *address.ContractAddress, balance *felt.Felt) {
```
rianhughes, line 41: nit: deployFundedContract / deployFundedClassInstance

blockchain/testchain/class.go
```diff
@@ -0,0 +1,83 @@
...
+	file, err := os.Open(path)
+	require.NoError(t, err)
+	defer file.Close()
+
+	intermediate := new(struct {
+		Abi         json.RawMessage            `json:"abi"`
+		EntryPoints starknet.SierraEntryPoints `json:"entry_points_by_type"`
+		Program     []*felt.Felt               `json:"sierra_program"`
+		Version     string                     `json:"contract_class_version"`
+	})
```
rianhughes, line 63: Do we only support cairoV1 classes??

blockchain/testchain/class.go
```diff
@@ -0,0 +1,83 @@
...
+
+	snCasm, err := compiler.Compile(&snSierra)
+	require.NoError(t, err)
+
+	junoClass, err := sn2core.AdaptCairo1Class(&snSierra, snCasm)
+	require.NoError(t, err)
+
+	// TODO: I don't like this too much (the deference using *), but every method that's currently
+	//       returning a reference type will eventually return a value type. Since this is for
+	//       testing is not critical.
```
rianhughes, line 81: Maybe we could pass in a Class instance and populate it in the function?

rpc/v8/estimate_fee_test.go
```diff
@@ -106,3 +113,455 @@ func TestEstimateFee(t *testing.T) {
...
+			broadcastedTransactions: []rpc.BroadcastedTransaction{
+				createDeclareTransaction(
+					t,
+					account,
+					&class.Hash,
+					class.Class.Compiled.Hash(),
+					&class.Sierra,
+				),
+			},
+			expected: []rpc.FeeEstimate{
```
rianhughes, line 151: Where did this test data come from?

rpc/v8/estimate_fee_test.go
```diff
@@ -106,3 +113,455 @@ func TestEstimateFee(t *testing.T) {
...
+type executionError struct {
+	ClassHash       string `json:"class_hash"`
+	ContractAddress string `json:"contract_address"`
+	Error           any    `json:"error"`
+	Selector        string `json:"selector"`
+}
+
+const executeEntryPointSelector = "0x15d40a3d6ca2ac30f4031e42be28da9b056fef9bb7357ac5e85627ee876e5ad"
+
+func TestEstimateFeeWithVMDeclare(t *testing.T) {
```
rianhughes, line 133: Nice work with the tests

utils/test.go
rodrigo-pino: For me it's better this way since I am using a designated constructor.  When creating a type I personally don't like the approach of create a new blank type and set a value, I think that should be the job of a type constructor

blockchain/testchain/utils.go
```diff
@@ -0,0 +1,43 @@
+package testchain
+
+import (
+	"fmt"
+	"path"
+	"runtime"
+)
+
+func contractsFolder() string {
+	// Only dumb way (that I know of) for reliably getting the root of a go project :(
```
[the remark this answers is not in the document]
rodrigo-pino: findTargetDirectory will always be dependant on how it is called. If it is called during normal execution the `os.Getwd()` func call returns the root, if it calls during testing it returns a different root (the directory where the test is being executed).

This way the root is always the same regardless or where the test executing it is located.

blockchain/testchain/utils.go
rodrigo-pino: I don't understand what you mean by external function. 

I think inlining it hurts readability.

blockchain/testchain/testchain.go
rodrigo-pino: I agree that the naming can be improved, if you have any suggestion please share.

Regarding with people who are unfamiliar with how Starknet works, I am not sure I follow, I don't think this the place to learn. This Juno testing infrastructure. It should make sense based on that you already know how Starknet works.

blockchain/testchain/testchain.go
rodrigo-pino: I am not sure I follow what I should rename

blockchain/testchain/testchain.go
rodrigo-pino: The Deployer class definition, I don't think `ContractDeployer` makes sense because it is a class definition and there are clear difference in Starknet between classes and contracts. Also, UDC makes less sense for me since it is an acronym for one type of deployer contract which if you don't know it makes less sense.

blockchain/testchain/testchain.go
rodrigo-pino: Daniil's work

blockchain/testchain/testchain.go
rodrigo-pino: Don't worry, very appreciated!

blockchain/testchain/contract.go
rodrigo-pino: I myself don't fully understand yet the full usability of this function. I will have to investigate more

blockchain/testchain/class.go
rodrigo-pino: I think `deploy` makes more sense. I think specifying the address shouldn't be required either. Let me see if I can get around that

blockchain/testchain/class.go
rodrigo-pino: I don't think we should specify it, in the Testchain anything different than Cairo1 doesn't exist. If we specify then it means there is another type of class (which exists, but not in this context). That's how I interprete it at least.

blockchain/testchain/class.go
rodrigo-pino: It would definitely helps performance, I think in this case performance is not so important. I think after we change to value types the only required change to this function would be removing the `*` deference operator.

rpc/v8/estimate_fee_test.go
rodrigo-pino: I believe it was populated with the VM results, very hard to predict otherwise. We are assuming that the Blockifier is correct.

rpc/v8/estimate_fee_test.go
rodrigo-pino: Daniil's work

PR #2628, opened 2025-03-11
feat(trie): implement hash based scheme for triedb

MaksymMalicki:


rodrigo-pino:
For PR names please use [conventional commit style](https://www.conventionalcommits.org/en/v1.0.0/)

PR #2631, opened 2025-03-11
fix: restore exact version pinning for Starknet dependencies

wojciechos:
## Changes
- Restored exact version pinning (using `=` prefix) for `blockifier` and `starknet_api` dependencies
- Changed from `blockifier = "0.14.0-rc.1"` to `blockifier = "=0.14.0-rc.1"`
- Changed from `starknet_api = "0.14.0-rc.1"` to `starknet_api = "=0.14.0-rc.1"`

## Motivation
This PR fixes an accidental removal of the exact version pinning (`=` prefix) that occurred previously. Exact version pinning is critical for these Starknet dependencies to ensure build reproducibility and prevent unexpected behavior from dependency version drift.

## Impact
- Restores consistent versioning strategy across all pinned dependencies
- Ensures build determinism and reproducibility
- Prevents potential issues that could arise from unintended version updates

### DIFF
--- a/vm/rust/Cargo.toml
+++ b/vm/rust/Cargo.toml
@@ -8,11 +8,11 @@ edition = "2021"
 [dependencies]
 serde = "1.0.208"
 serde_json = { version = "1.0.125", features = ["raw_value"] }
-blockifier = "0.14.0-rc.1"
+blockifier = "=0.14.0-rc.1"
 cairo-lang-casm = "=2.10.1"
 cairo-lang-starknet-classes = "=2.10.0"
 cairo-lang-runner = { version = "2.10.0" }
-starknet_api = "0.14.0-rc.1"
+starknet_api = "=0.14.0-rc.1"
 cairo-vm = "=1.0.2"
 starknet-types-core = { version = "0.1.6", features = ["hash", "prime-bigint"] }
 indexmap = "2.1.0"


rodrigo-pino:
Did not having this cause any type of error? I think in theory this shouldn't happen. Also, if we want people to have the exact same rust dependencies should we commit the `Cargo.lock` instead?

hudem1 (approved):

wojciechos:
> Did not having this cause any type of error? I think in theory this shouldn't happen.

Without the = prefix, Cargo follows SemVer compatibility rules, meaning:
For blockifier = "0.14.0-rc.1", Cargo might accept 0.14.0-rc.3 or other compatible pre-releases which actually happened see [those logs ](https://github.com/NethermindEth/juno/actions/runs/13790586109/job/38569117101#step:9:192)from main branch

> Also, if we want people to have the exact same rust dependencies should we commit the Cargo.lock instead?

I'm not much familiar with how exactly it works, but yep, sounds like better approach

wojciechos:
> Did not having this cause any type of error? I think in theory this shouldn't happen. Also, if we want people to have the exact same rust dependencies should we commit the `Cargo.lock` instead?

Without the = prefix, Cargo follows SemVer compatibility rules, meaning:
For blockifier = "0.14.0-rc.1", Cargo might accept 0.14.0-rc.3 or other compatible pre-releases which actually happened see [those logs ](https://github.com/NethermindEth/juno/actions/runs/13790586109/job/38569117101#step:9:192)from main branch

I'm not much familiar with how exactly Cargo.lock works, but yep, sounds like better approach

rodrigo-pino:
Maybe we can directly in favor of https://github.com/NethermindEth/juno/pull/2632

wojciechos:
in favor of https://github.com/NethermindEth/juno/pull/2632

PR #2639, opened 2025-03-12
refactor(rpc): remove unused code

hudem1:
a PR to clean / reorganize a bit

- Removing deadcode concerning Data Availability in rpcv6 & vm/trace.go
- Moving rpc adapters to dedicated file
- Returning value type instead of (heap) pointers by default for adapters


rpc/v6/adapters.go
```diff
@@ -339,3 +345,392 @@ func adaptFeederExecutionResources(resources *starknet.ExecutionResources) Compu
 		SegmentArena: builtins.SegmentArena,
 	}
 }
+
+func adaptToFeederResourceBounds(rb *map[Resource]ResourceBounds) *map[starknet.Resource]starknet.ResourceBounds {
```
rodrigo-pino, line 349: No need for pointers, either as input or output parameters. `map[K]V` is already a pointer by default

rpc/v6/adapters.go
```diff
@@ -339,3 +345,392 @@ func adaptFeederExecutionResources(resources *starknet.ExecutionResources) Compu
...
+	for resource, bounds := range *rb {
+		feederResourceBounds[starknet.Resource(resource)] = starknet.ResourceBounds{
+			MaxAmount:       bounds.MaxAmount,
+			MaxPricePerUnit: bounds.MaxPricePerUnit,
+		}
+	}
+	return &feederResourceBounds
+}
+
+func adaptToFeederDAMode(mode *DataAvailabilityMode) *starknet.DataAvailabilityMode {
```
rodrigo-pino, line 363: Return value type here.

rodrigo-pino (commented):
I see is still a draft and has the small issues we have discussed recently of the use of `T` vs `*T`. I'll let you fix them. I now this is a refactor but I am ok with extend this function to remove these when straightforward

rpc/v6/adapters.go
hudem1: Ah yeah, I just moved those functions from the other file to this file.
Yeah for sure, I'll make the changes to all functions needing it!

rianhughes:
This approach looks good to me

rodrigo-pino:
Closing because further changes will cancel this out

PR #2643, opened 2025-03-14
feat(onboarding): add juno_getBlockWithTxnHashesAndReceipts

varex83:


rpc/v8/block.go
```diff
@@ -158,6 +165,47 @@ func (h *Handler) BlockWithReceipts(id BlockID) (*BlockWithReceipts, *jsonrpc.Er
 	}, nil
 }
 
+func (h *Handler) BlockWithTxHashesAndReceipts(id BlockID) (*BlockWithTxHashesAndReceipts, *jsonrpc.Error) {
```
rodrigo-pino, line 168: Here you should return `BlockWithTxHashesAndReceipts` instead of `*Block...AndReceipts`. The reason is that the latter always forces a heap allocation which is sub-optimal. Read the recent group interactions in slack to understand a bit more

rpc/v8/block.go
```diff
@@ -91,6 +91,22 @@ type BlockWithReceipts struct {
...
+	BlockHeader
+	TxnHashes    []*felt.Felt             `json:"transactionsHashes"`
+	Transactions []TransactionWithReceipt `json:"transactions"`
+}
+
+func EmptyBlockWithTxHashesAndReceipts() BlockWithTxHashesAndReceipts {
+	return BlockWithTxHashesAndReceipts{
+		Status:       rpcv6.BlockPending,
+		BlockHeader:  BlockHeader{},
+		TxnHashes:    []*felt.Felt{},
```
rodrigo-pino, line 105: It looks good, why is this an array of `*felt.Felt` instead of `felt.Felt`. Which would be the best use of it

varex83: Just was looking at implementation of `BlockWithTxHashes` :D , will fix

PR #2644, opened 2025-03-14
fix: traceBlockTransactions shouldn't fail because of wrong class cache

infrmtcs:


vm/vm.go
```diff
@@ -223,16 +224,6 @@ func (v *vm) Call(callInfo *CallInfo, blockInfo *BlockInfo, state core.StateRead
...
 
-	var concurrencyModeByte byte
-	if v.concurrencyMode {
-		concurrencyModeByte = 1
-	}
-	var structuredErrStackByte byte
-	if structuredErrStack {
-		structuredErrStackByte = 1
-	}
-	C.setVersionedConstants(C.CString("my_json"))
```
infrmtcs, line 234: This is not related, but I'm feeling that it's a debug line that we forgot to remove.

rodrigo-pino: @AnkushinDaniil

rodrigo-pino (approved):

AnkushinDaniil (approved):

PR #2650, opened 2025-03-17
fix(rpc): handle updated entry point not found error format

AnkushinDaniil:
After the blockifier update, the format of the `entry_point_not_found` error has changed to return only the error's code. This change affected the behavior of RPC v6-7. To address this, explicit error handling has been implemented to properly manage the new error format.

### DIFF
--- a/rpc/v6/trace.go
+++ b/rpc/v6/trace.go
@@ -306,7 +306,7 @@ func (h *Handler) Call(funcCall *FunctionCall, id *BlockID) ([]*felt.Felt, *json
 		if len(res.Result) != 0 {
 			if res.Result[0].String() == rpccore.EntrypointNotFoundFelt {
 				strErr = fmt.Sprintf(
-					"Entry point EntryPointSelector(%s) not found in contract.",
+					rpccore.ErrEPSNotFound,
 					funcCall.EntryPointSelector.String(),
 				)
 			} else {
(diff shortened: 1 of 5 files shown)


hudem1 (approved):

wojciechos (approved):

rpc/v6/trace.go
```diff
@@ -304,7 +305,10 @@ func (h *Handler) Call(funcCall *FunctionCall, id *BlockID) ([]*felt.Felt, *json
 		var strErr string
 		if len(res.Result) != 0 {
 			if res.Result[0].String() == rpccore.EntrypointNotFoundFelt {
-				strErr = rpccore.ErrEntrypointNotFound.Message
+				strErr = fmt.Sprintf(
+					"Entry point EntryPointSelector(%s) not found in contract.",
```
rianhughes, line 309: Maybe we can push this string to rpccore?

const ErrEPSNotFound = "Entry point EntryPointSelector(%s) not found in contract."

strErr = fmt.Sprintf(
					rpccore.ErrEPSNotFound,
					funcCall.EntryPointSelector.String(),
				)

rianhughes (approved):

rodrigo-pino (commented):
Tests please that assert the error message for both v6, v7 and v8

rodrigo-pino (changes requested):
Since there are many approvals writing a blocking comment for now

rodrigo-pino (approved):
Oh, I missed the files, they were tests already. All good 👍

PR #2660, opened 2025-03-19
feat(rpc/v0.8.0): change Offset type from *felt.Felt to uint64 in CasmEntryPoint

wojciechos:
Changed Offset field type to match Starknet v0.8.0 specification, where the offset for `starknet_getCompiledCasm` response is now represented as an integer.

Reference: https://github.com/starkware-libs/starknet-specs/blob/v0.8.0/api/starknet_executables.json#L136


rodrigo-pino (approved):
LGTM

PR #2661, opened 2025-03-19
refactor(db): modularize db interfaces to make it cleaner and more performant

weiihann:
### Summary
This PR modularizes the interfaces related to database, making it easier to extend functionalities and use the proper tools to ensure better performance.

### Problem Statement
The old pattern of performing CRUD operation is as follows:
```
DB -> NewTransaction -> Perform the operation
```
**The good** - Operations are guaranteed to be atomic. If a transaction fails, the data will not be flushed. If a transaction succeeds, it's guaranteed to be flushed to the database.
**The bad:**
- For updates: Internally, a `Transaction` is an `IndexedBatch`. `IndexedBatch` is only useful when you need to read from both the in-memory batch and also the disk, but the writes will be slower. However, our operations only need to read from the database, but not the in-memory batch. We should just use the regular `Batch` for all our updates for performance.
- For reads: Internally, a `Transaction` creates a `Snapshot`. `Snapshot` is only useful if we want to read at a specific point in time. For atomic operations, it does guarantee that whatever you read in the `Snapshot` is what you get at a specific block height. However, creating too many long-live snapshots jeopardizes the node performance because it will delay database compactions.
- Concurrent writes to the db are not possible. Or at least, to do concurrent writes, we use `BufferedTransaction` to store the changes then flush to the underlying batch. Essentially we want to do the same thing as a `Batch` but because the `Transaction` interface isn't sufficient so `BufferedTransaction` is a hack. This hack works but it generates heap allocations. So why not just use the regular PebbleDB `Batch`?

### Changes
- Introduce a set of interfaces: `KeyValueStore`, `KeyValueReader`, `KeyValueWriter`, `Batch`, `Batcher`, `Snapshot`, `Snapshotter`, `Iterable`, `KeyValueRangeDeleter`. The design helps us to extend functionalities easily.
- Introduce a custom memory database. Tests should use this database. We can also use PebbleDB's memory database, but it's harder to debug.
- Avoid using `Snapshot`. As far as I can tell, there are no operations that require us to read at a specific point in time.
- A bunch of other code clean-ups and structuring.

### Best Practices
- If you need to read from the database, parse `db.KeyValueStore` into your component and read directly from it.
- If you need to write, create a `Batch`, apply the writes, then flush it at the end.
- If you need to write concurrently in a single process, create a `Batch` for each goroutine, apply the writes accordingly, merge them, then flush in one go. Note that multiple batches can be committed concurrently, so different parts of Juno can commit batch at the same time. Internally, PebbleDB handles flushing the data to the disk in a serialized manner.

### Impacts
- We should see a slight improvement in node performance and RPC as we no longer use `Snapshot`.
- We will not see an improvement in sync speed because we are still using `IndexedBatch`. The new state refactor will omit this and use `Batch` only, which further improves the performance.
- No migrations are needed.

### How to review this PR
- Please do not review state and trie related files "too seriously". In this PR, I try to replicate the old behaviour, which may include bad practices. Those files will be nuked in the state refactor PRs.
- Please pay extra attention to the following:
  - CRUD patterns (i.e. `core/accessors.go`, `blockchain/accessors.go`, `db/schema.go`)
  - The DB interfaces


utils/slices.go
```diff
@@ -85,3 +85,12 @@ func FeltArrToString(arr []*felt.Felt) string {
 	}
 	return strings.Join(res, ", ")
 }
+
+func CopySlice[T any](slice []T) []T {
```
infrmtcs, line 89: Should we use this https://pkg.go.dev/slices#Clone instead?

infrmtcs:
> Avoid using Snapshot. As far as I can tell, there are no operations that require us to read at a specific point in time.

I think we should carefully reconsider this for these reasons:
- If network has reorg, a block number can be mapped to different block/block hash.
- If new head is ingested during the execution, the heads loaded by 2 queries are actually 2 different blocks.

db/database.go
```diff
@@ -0,0 +1,58 @@
+package db
+
+import "io"
+
+// Exposes a read-only interface to the database
+type KeyValueReader interface {
```
infrmtcs, line 6: I'm wondering if we can have generic types instead, something like `KeyValueReader[K dbkey, V dbvalue]` where the constraints require encoder for key and encoder + decoder for value.
Benefits: We don't have to write all operations manually for all buckets. Instead we only need to provide the encoders and decoders.

weiihann: That sounds interesting, can you give a more detailed example?

weiihann:
> > Avoid using Snapshot. As far as I can tell, there are no operations that require us to read at a specific point in time.
> 
> I think we should carefully reconsider this for these reasons:
> 
> * If network has reorg, a block number can be mapped to different block/block hash.
> * If new head is ingested during the execution, the heads loaded by 2 queries are actually 2 different blocks.

You're right, we can still use a snapshot for most operations. But this solution isn't going to be good for the node performance. Ideally, we should handle those scenarios on the app layer.

core/accessors.go
```diff
@@ -0,0 +1,116 @@
...
+		return nil
+	})
+	return number, err
+}
+
+func DeleteBlockHeaderNumberByHash(w db.KeyValueWriter, hash *felt.Felt) error {
+	return w.Delete(db.BlockHeaderNumbersByHashKey(hash))
+}
+
+func GetBlockHeaderByNumber(r db.KeyValueReader, number uint64) (*Header, error) {
```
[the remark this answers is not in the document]
weiihann: Yeah... that's an issue. I think it's better to put these accessors files in a single package and group them accordingly. If not, we're gonna lose track and have duplicates.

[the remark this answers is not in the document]
weiihann: Scrap that, there will be circular dependency

core/contract.go
```diff
@@ -115,27 +115,23 @@ func (c *ContractUpdater) Purge() error {
...
 // ContractNonce returns the amount transactions sent from this contract.
 // Only account contracts can have a non-zero nonce.
-func ContractNonce(addr *felt.Felt, txn db.Transaction) (*felt.Felt, error) {
-	key := db.ContractNonce.Key(addr.Marshal())
+func ContractNonce(addr *felt.Felt, txn db.IndexedBatch) (*felt.Felt, error) {
 	var nonce *felt.Felt
-	if err := txn.Get(key, func(val []byte) error {
-		nonce = new(felt.Felt)
-		nonce.SetBytes(val)
+	err := txn.Get(db.ContractNonceKey(addr), func(val []byte) error {
```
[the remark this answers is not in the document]
weiihann: Missed this, added

core/contract.go
```diff
@@ -177,22 +173,18 @@ func ContractStorage(addr, key *felt.Felt, txn db.Transaction) (*felt.Felt, erro
...
 
 // ContractClassHash returns hash of the class that the contract at the given address instantiates.
-func ContractClassHash(addr *felt.Felt, txn db.Transaction) (*felt.Felt, error) {
-	key := db.ContractClassHash.Key(addr.Marshal())
+func ContractClassHash(addr *felt.Felt, txn db.IndexedBatch) (*felt.Felt, error) {
 	var classHash *felt.Felt
-	if err := txn.Get(key, func(val []byte) error {
-		classHash = new(felt.Felt)
-		classHash.SetBytes(val)
+	err := txn.Get(db.ContractClassHashKey(addr), func(val []byte) error {
```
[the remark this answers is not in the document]
weiihann: Missed this, added

db/database.go
```diff
@@ -0,0 +1,58 @@
...
+	Has(key []byte) (bool, error)
+	// If a given key exists, the callback will be called with the value
+	// Example:
+	//
+	//	var value []byte
+	//	db.Get([]byte("key"), func(v []byte) error {
+	//		value = v
+	//		return nil
+	//	})
+	Get(key []byte, cb func(value []byte) error) error
```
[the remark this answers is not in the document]
weiihann: Primarily to reduce heap allocations. If we do `Get(key []byte) ([]byte, error)` then we might get allocations overhead. The code is slightly less readable but I think it's worth it.

core/trie/storage.go
```diff
@@ -124,11 +122,12 @@ func (t *Storage) DeleteRootKey() error {
 
 func (t *Storage) SyncedStorage() *Storage {
 	return &Storage{
-		txn:    db.NewSyncTransaction(t.txn),
+		txn:    db.NewSyncBatch(t.txn),
 		prefix: t.prefix,
 	}
 }
 
 func newMemStorage() *Storage {
```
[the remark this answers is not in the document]
weiihann: You mean the `SyncBatch` or `SyncedStorage`? Both are required in the main execution path.

db/memory/db_test.go
```diff
@@ -0,0 +1,15 @@
+package memory
```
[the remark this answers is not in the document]
weiihann: `TestKeyValueStoreSuite` in `db/testutil.go` will run the test suites to ensure that both `pebbleDB` and `memoryDB` have the same behavior.

core/accessors.go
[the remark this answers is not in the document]
weiihann: I think it's better to put all the accessors in core. Ideally they should be in a separate subpackage but sadly we can't do that

db/database.go
infrmtcs: I have a raw example here, it still needs to be refined, but it illustrate my rough idea: https://github.com/NethermindEth/juno/commit/d0eb8115e429c8bd7c8767fb90ad0ca5eb5b9809

core/trie/storage.go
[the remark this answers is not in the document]
MaksymMalicki: Actualy, I meant the memory storage 😅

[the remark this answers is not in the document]
weiihann: Ah, it's still used by some other non-tests files

db/database.go
weiihann: I do like that the generics approach offer type safety, reduction in boilerplate code (i.e. accessors) and more intuitive API in some cases.

However, when I attempt to plug in this idea, it's not as simple as it seems. Primarily, there are too many modular interfaces in which `KeyValueStore` encapsulates. Creating typed versions of all these interfaces, for each bucket, is way too complex and it may be unnecessary. Besides that, there are certain cases where we need to access only parts of the keys. For example, perform DB iteration based on a prefix of a prefix.

I'm not strongly against this idea. In fact, I do think that there should be better ways of handling all the accessor functions. If we can come up with an architecture such that it is modular, clean, ease of extension and type safety, that would be awesome.

db/memory/db.go
```diff
@@ -0,0 +1,201 @@
...
+	}, nil
+}
+
+func (d *Database) NewSnapshot() db.Snapshot {
+	if d.db == nil {
+		panic(errDBClosed)
+	}
+	return d.copy()
+}
+
```
rianhughes, line 156: This snapshot implementation seems very heavy given we copy the entire DB. What do you think about keeping all writes in the db (so don't overwrite like we currently do), and stamping them with a `snapshotID`, which just a write counter (so the 5th write would have `snapshotID=5`). When we create a snapshot which just refer to the underlying db, but with (eg) `snapshotID=8`. Reads will only be able to see the k-v pairs if they were written with `snapshotID`s (eg) less than `8`. This way we get stable view of the DB, even as we continue to write to it. It also saves us making a heavy read. Baiscally something like this:

```
type VersionedValue struct {
	snapshotID       uint64 // the curSnapshotID the k-v was inserted into the db 
	val       []byte
	tombstone bool // tracks if the value was deleted
}

type Database struct {
	data       map[string][]VersionedValue // key -> sorted versions (desc seq)
	curSnapshotID uint64
	lock       sync.RWMutex
}


func (db *Database) NewSnapshot() *Snapshot {
	db.lock.Lock()
	defer db.lock.Unlock()

	seq := db.currentSeq
	db.snapshots[seq] = struct{}{}
	return &Snapshot{seq: seq, db: db}
}

func (db *Database) Set(key, val []byte) {
	db.lock.Lock()
	defer db.lock.Unlock()

	db.currentSeq++
	v := VersionedValue{
		seq: db.currentSeq,
		val: append([]byte{}, val...), 
	}
	db.data[string(key)] = append([]VersionedValue{v}, db.data[string(key)]...)
}
func (s *Snapshot) Get(key []byte) ([]byte, bool) {
	s.db.lock.RLock()
	defer s.db.lock.RUnlock()

	versions := s.db.data[string(key)]
	for _, v := range versions {
		if v.seq <= s.seq {
			if v.tombstone {
				return nil, false
			}
			return v.val, true
		}
	}
	return nil, false
}

```

rianhughes: We can also implement a basic compaction function to drop old data, given we only ever need a stable view of the DB at the point the snapshot was created

weiihann: PebbleDB's snapshot doesn't actually replicate the DB. How it works internally is that:
- each DB commit has a unique and incremental sequence number
- a snapshot creates a read-only view at the latest sequence number
- When reading from the DB, it will go through each key-value and skip the key-value with higher sequence number

The mechanism that you've described is somewhat similar. The problem with this approach is that we have to maintain this read-only view in memory, which could be quite expensive, especially when the state diff is larger. This read-only view will only get larger and larger as the chain progresses due to trie growth.

MaksymMalicki (approved):

rodrigo-pino (commented):
First round of reviews

db/memory/db.go
rianhughes: Yup, that was the mechanism I was trying to describe, you did a much nicer job!

The db shouldn't grow indefinitely in size if we implement compaction. Basically remove the old data once it is no longer referenced. For example, if the snapshot is tied to a sequence number of 10, we can delete all data prior to 10. If there are no snapshots then we delete all the old data, and the db should be the same as the overwrite implementation (until a new snapshot is created anyway). If this is done periodically then it should prevent state growth (depending on how long lived the snapshots are). Also, in the current approach, as the state diff gets larger, creating snapshots gets more and more expensive, and scales with the number of read txns.

db/pebble/snapshot.go
```diff
@@ -1,80 +1,70 @@
...
-
-	return nil
+	return true, closer.Close()
 }
 
-// Commit : see db.Transaction.Commit
-func (s *snapshot) Commit() error {
-	return ErrReadOnlyTransaction
-}
+//nolint:dupl
```
rodrigo-pino, line 32: 👀

db/testutil.go
rodrigo-pino, line 1: Doesn't look like an appropriate name for testing files

migration/migration.go
rodrigo-pino, line 1: Do we have to do migrations after this update? What will be the next time we have to do a migration?

rodrigo-pino (commented):
Second round. Looks very good and it is easy to read, thank you!

builder/builder.go
```diff
@@ -106,17 +106,12 @@ func (b *Builder) PendingBlock() *core.Block {
...
-	}
-
 	pending, err := b.Pending()
 	if err != nil {
-		return nil, nil, utils.RunAndWrapOnError(txn.Discard, err)
+		return nil, nil, err
 	}
 
-	return sync.NewPendingState(pending.StateUpdate.StateDiff, pending.NewClasses, b.headState), txn.Discard, nil
+	return sync.NewPendingState(pending.StateUpdate.StateDiff, pending.NewClasses, b.headState), func() error { return nil }, nil
```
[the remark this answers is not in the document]
weiihann: This state closer thing should actually be removed across all the state-related methods. But removing it will modify the interfaces and need a bunch of other changes which are unrelated to this PR. In the upcoming state refactor task, this should be removed. So I think we don't need to explicitly declare a global var.

I'll add a TODO to remove this in the future.

blockchain/event_filter.go
```diff
@@ -13,7 +13,7 @@ import (
 
 var errChunkSizeReached = errors.New("chunk size reached")
 
-//go:generate mockgen -destination=../mocks/mock_event_filterer.go -package=mocks github.com/NethermindEth/juno/blockchain EventFilterer
+// // go:generate mockgen -destination=../mocks/mock_event_filterer.go -package=mocks github.com/NethermindEth/juno/blockchain EventFilterer
```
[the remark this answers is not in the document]
weiihann: oops 😝

blockchain/blockchain.go
```diff
@@ -647,187 +350,64 @@ func (b *Blockchain) SanityCheckNewHeight(block *core.Block, stateUpdate *core.S
...
-	var r *core.TransactionReceipt
-	err := txn.Get(db.ReceiptsByBlockNumberAndIndex.Key(bnIndex.MarshalBinary()), func(val []byte) error {
-		return encoder.Unmarshal(val, &r)
-	})
-	return r, err
-}
-
 type StateCloser = func() error
 
+var noopStateCloser = func() error { return nil }
```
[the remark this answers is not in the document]
weiihann: It'll be removed in the future so I think it doesn't matter :)

blockchain/blockchain.go
```diff
@@ -290,88 +240,68 @@ func (b *Blockchain) SubscribeL1Head() L1HeadSubscription {
...
-
-	return update, b.database.View(func(txn db.Transaction) error {
-		var err error
-		update, err = l1Head(txn)
-		return err
+	var l1Head *core.L1Head
+	err := b.database.Get(db.L1Height.Key(), func(data []byte) error {
+		return encoder.Unmarshal(data, &l1Head)
 	})
-}
```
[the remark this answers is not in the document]
weiihann: Ah good catch. I'll add an accessor function for it.

db/memory/db.go
weiihann: Hmm I don't think I quite get what you're trying to do. 

A PebbleDB's snapshot is a read-only view of the DB at a specific point in time. It doesn't perform any sorts of copy. The problem of having long-lived snapshot is that, it prevents PebbleDB from performing compactions. Compactions occur to remove the obsolete data by merging the the underlying SSTables. If the compaction is delayed, the compaction process takes much longer to execute, which could halt other processes because it is a CPU-heavy process. Imagine if a user submits a request for `starknet_getClass` and the response time is >10s because a compaction is taking place.

What you're describing seems to be some sort of garbage collector and buffering mechanism. Basically, buffer the changes in-memory and persist them when the buffer size hits a certain threshold. This is exactly what we are doing with the in-memory cache for trie (and in the future, in-memory cache for state). I think it's better to have this mechanism at the components level (i.e. state, trie) because we don't need to do decoding.

core/state.go
```diff
@@ -163,17 +163,24 @@ func (s *State) globalTrie(bucket db.Bucket, newTrie trie.NewTrieFunc) (*trie.Tr
...
 	}
 
+	rootKey := new(trie.BitArray)
+	if len(val) > 0 {
+		err = rootKey.UnmarshalBinary(val)
+		if err != nil {
+			return nil, nil, err
+		}
+	}
+
```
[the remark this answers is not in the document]
weiihann: yeah good catch I have no idea why I did it that way. Reverted it to the original implementation.

db/pebble/batch.go
```diff
@@ -2,103 +2,85 @@ package pebble
...
-	start := time.Now()
-	defer func() { b.listener.OnCommit(time.Since(start)) }()
-	return utils.RunAndWrapOnError(b.Discard, b.batch.Commit(pebble.Sync))
+func (b *batch) DeleteRange(start, end []byte) error {
+	return b.batch.DeleteRange(start, end, pebble.Sync)
 }
 
-// Set : see db.Transaction.Set
-func (b *batch) Set(key, val []byte) error {
+//nolint:dupl
```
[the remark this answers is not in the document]
weiihann: The linter determines that `batch.Get` has duplicated codes in `snapshot.Get`. We can move the duplicated parts to a common function, but we have to define a `getter` interface. IMO the code redirection of using an interface to reduce the duplicates doesn't make the code easier to read.

db/pebble/snapshot.go
weiihann: Same reason as above

migration/migration.go
weiihann: No migrations are needed. The behaviour of how most components interact with the database is pretty much the same.

db/testutil.go
weiihann: It just provides a tester function that the implementation of `KeyValueStore` can use to test if it matches the behaviour.

core/accessors.go
```diff
@@ -0,0 +1,495 @@
...
+		if vErr != nil {
+			return nil, utils.RunAndWrapOnError(it.Close, vErr)
+		}
+
+		var tx Transaction
+		if err = encoder.Unmarshal(val, &tx); err != nil {
+			return nil, utils.RunAndWrapOnError(it.Close, err)
+		}
+
+		txs = append(txs, tx)
```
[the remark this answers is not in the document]
weiihann: We can store additional metadata (e.g. length of the transactions of a given block) in the DB. This is hyper-optimization already, and I don't think we can justify it right now unless we do profiling.

Other than storing the metadata in the DB, there's no way of knowing the size beforehand.

db/pebble/batch.go
```diff
@@ -2,103 +2,85 @@ package pebble
...
 )
 
-var _ db.Transaction = (*batch)(nil)
+var _ db.Batch = (*batch)(nil)
 
 type batch struct {
 	batch    *pebble.Batch
-	lock     *sync.Mutex
+	db       *DB
+	size     int // size of the batch in bytes
```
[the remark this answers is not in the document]
weiihann: It works nicely with `len()` so that's why `int` is used.

db/remote/iterator.go
```diff
@@ -59,6 +59,10 @@ func (i *iterator) First() bool {
 	return len(i.currentK) > 0 || len(i.currentV) > 0
 }
 
+func (i *iterator) Prev() bool {
+	panic("not implemented")
```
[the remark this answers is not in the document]
weiihann: Adding a support for it will take more time but I'm not sure if we are even using the `remoteDB` feature now.

mempool/mempool.go
```diff
@@ -120,8 +119,8 @@ type Pool struct {
 
 // New initialises the Pool and starts the database writer goroutine.
 // It is the responsibility of the caller to execute the closer function.
-func New(mainDB db.DB, bc blockchain.Reader, maxNumTxns int, log utils.SimpleLogger) *Pool {
-	pool := Pool{
+func New(mainDB db.KeyValueStore, bc blockchain.Reader, maxNumTxns int, log utils.SimpleLogger) *Pool {
+	pool := &Pool{
```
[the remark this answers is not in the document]
weiihann: which generated code are you referring to?

db/batch.go
```diff
@@ -0,0 +1,38 @@
...
+	// Creates a write-only batch with a pre-allocated size
+	NewBatchWithSize(size int) Batch
+}
+
+// Same as Batch, but allows for reads from the batch and the disk.
+// Use this only if you need to read from both the in-memory and on-disk data.
+// Write operations will be slower.
+// Ideally, IndexedBatch should not be used at all, because it's better to write using regular batch
+// and read directly from the database.
+type IndexedBatch interface {
```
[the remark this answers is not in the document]
weiihann: I'll add more comments on that

db/memory/iterator.go
```diff
@@ -0,0 +1,77 @@
...
+import (
+	"bytes"
+	"errors"
+
+	"github.com/NethermindEth/juno/db"
+)
+
+var _ db.Iterator = (*iterator)(nil)
+
+type iterator struct {
```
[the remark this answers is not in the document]
weiihann: The description is provided in the `Iterator` interface in `db/iterator.go`

db/bufferbatch.go
```diff
@@ -0,0 +1,83 @@
+package db
+
+import (
+	"slices"
+)
+
+// TODO: DO NOT USE THIS! This is meant to be a temporary replacement for buffered transaction.
+// After state refactor, we can remove this.
+type BufferBatch struct {
```
rianhughes, line 9: Should we be using this in `updateContractStorages` and `updateStorageBuffered`?

blockchain/blockchain.go
```diff
@@ -647,187 +341,64 @@ func (b *Blockchain) SanityCheckNewHeight(block *core.Block, stateUpdate *core.S
...
-	_, err = core.ChainHeight(txn)
+	_, err := core.GetChainHeight(txn)
 	if err != nil {
-		return nil, nil, utils.RunAndWrapOnError(txn.Discard, err)
+		return nil, nil, err
 	}
 
-	return core.NewState(txn), txn.Discard, nil
+	return core.NewState(txn), noopStateCloser, nil
 }
```
rianhughes, line 359: Previously we would return a Snapshot, which provided a stable view of the DB. The IndexedBatch type breaks this right? 

I imagine this isn't a problem for reading historical data since it's fixed. This change should only affect the Sync and Builder services, although I can't think of a reason they would need a snapshot of the pending state. So it should be fine. Just wanted to confirm this.

core/state.go
```diff
@@ -51,18 +51,18 @@ type StateReader interface {
...
+	txn db.IndexedBatch
 }
 
-func NewState(txn db.Transaction) *State {
+func NewState(txn db.IndexedBatch) *State {
 	return &State{
 		history: &history{txn: txn},
 		txn:     txn,
 	}
 }
```
rianhughes, line 62: Do you know why we use transactions to access historical data that is fixed, instead of keying the DB directly with db.Get() etc? Afaik, the transactions shouldn't be needed when the db is immutable, and keying directly should save a lot of overhead (txns place a lot more pressure on the HEAP, GC, and can affect the database performance in subtle ways - eg preventing compaction etc)

rianhughes: Afaik, we should only need batched transactions for things like Sync, Builder and when simulating transactions. I don't think we ever care about snapshots for these very short lived states.

db/memory/db.go
rianhughes: Sorry, I'll try to be more clear. Currently if we want to get a snapshot of the in-memory DB, we need to copy the entire db, which seems expensive, and we do this for every snapshot we create. I'm just wondering if we can replicate how pebble implements snapshots, here in our in-memory DB.

db/snapshot.go
```diff
@@ -0,0 +1,13 @@
...
+// If you don't need to read at a specific time, use the database directly.
+type Snapshot interface {
+	KeyValueReader
+	Iterable
+}
+
+// Produces a read-only snapshot of the database
+type Snapshotter interface {
+	NewSnapshot() Snapshot
+}
```
rianhughes, line 13: Do we need Snapshot? We don't seem to use it anywhere outside of tests. Maybe we can just delete it?

core/state.go
rianhughes: Oh we actually do key directly for most data.

rianhughes: Actually, the `rpc` layer gathers `state` data via a transaction which isn't necessary since it 1) never modifies state and 2) never accesses mutable state (most of the time, the pending block might be an exception here). The `rpc` layer should be able to access state directly by keying, although this is an optimisation for a different PR.

rianhughes: It would also be cleaner to update the core.State type to expose either a ReadState or MutableState. Currently the `rpc` layer can access the states Update function which isn't great.

rianhughes:
Up to the comments above, this PR LGTM

core/accessors.go
```diff
@@ -0,0 +1,495 @@
...
+}
+
+func DeleteTxBlockNumIndexByHash(w db.KeyValueWriter, hash *felt.Felt) error {
+	return w.Delete(db.TxBlockNumIndexByHashKey(hash))
+}
+
+func GetTxByBlockNumIndex(r db.KeyValueReader, blockNum, index uint64) (Transaction, error) {
+	var tx Transaction
+	err := r.Get(db.TxByBlockNumIndexKey(blockNum, index), func(data []byte) error {
+		return encoder.Unmarshal(data, &tx)
```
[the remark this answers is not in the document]
weiihann: So I've tried `tx` and it broke most of the things, so we still need to pass `&tx`. Not sure how the unmarshaller works under the hood.

db/bufferbatch.go
weiihann: Yeah it was meant as a replacement for `BufferedTransaction` but after the state refactor this should be removed.

blockchain/blockchain.go
weiihann: Yeah this should be fine.

db/snapshot.go
weiihann: Hmm i think we can still expose this functionality, if in the future we want to use it.

mempool/mempool.go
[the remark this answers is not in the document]
rodrigo-pino: Ah more clearly, in the compiled code. Creating it like `Pool` and returning `&p` is the same in the compilation output as creating `&Pool` and returning `p`. The way the variable is created doesn't affect where it gets stored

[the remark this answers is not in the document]
rodrigo-pino: Just a nit, it is not importnat

db/memory/db.go
weiihann: Ah I got you. In that case I think the versioned strategy you mentioned works quite nicely. But the memoryDB is meant to be a in-memory data store for temporary usage such as for tests. It should not be long-lived. If you need long-lived in-memory key-value store, it's essentially a cache, and there are better cache implementations than this. So I don't think we need to complicate things further.

rianhughes (approved):

PR #2663, opened 2025-03-19
feat(core): add new address and hash types

rodrigo-pino:


rodrigo-pino:
Bypassing rules, only adding new types that will be used later by the code

PR #2664, opened 2025-03-19
refactor(felt): organize felt and add new ctors

rodrigo-pino:


rodrigo-pino:
Bypassing rules, it just re-structures the Felt script and add new constructors

PR #2667, opened 2025-03-20
refactor(vm): refactor loading of the versioned constants

MaksymMalicki:
From what I’ve seen in Madara and Pathfinder, they implemented a map of starknet versions to default versioned constants (saved locally). Then, they override the defaults with files provided by user, with given versions, specified in .yaml or a json file (CLI flag).

This PR is a simple modification, so that our CUSTOM_VERSIONED_CONSTANTS static var hold a map of custom files, and in get_versioned_constants, fetches from the map of customs first, or directly from blockifier, if given version doesn’t appear in CUSTOM_VERSIONED_CONSTANTS

### DIFF
--- a/vm/testdata/versioned_constants/custom_versioned_constants_multiple.json
+++ b/vm/testdata/versioned_constants/custom_versioned_constants_multiple.json
@@ -1,0 +1,4 @@
+{
+    "0.13.2": "./testdata/versioned_constants/0_13_2.json",
+    "0.13.2.1": "./testdata/versioned_constants/0_13_2.json"
+}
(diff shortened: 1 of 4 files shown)


vm/testdata/versioned_constants/custom_versioned_constants_multiple.json
```diff
@@ -0,0 +1,4 @@
+{
+    "0.13.2": "./testdata/versioned_constants/0_13_2.json",
+    "0.13.2.1": "./testdata/versioned_constants/0_13_2.json"
```
infrmtcs, line 3: I think you can recover a version from this PR: https://github.com/NethermindEth/juno/pull/2597/files

MaksymMalicki: I just wanted to test, if a JSON file with multiple keys is loading properly. Doesn't really matter what is the value here. Should I still change the duplicate for some other version, as you pointed out

rodrigo-pino (approved):

PR #2669, opened 2025-03-21
fix(rpc): get_storage_proof should return BLOCK_NOT_FOUND in some cases

infrmtcs:


rpc/v8/storage.go
```diff
@@ -163,24 +163,34 @@ func processStorageKeys(storageKeys []StorageKeys) ([]StorageKeys, *jsonrpc.Erro
...
-			return nil
+		header, err := h.bcReader.BlockHeaderByHash(id.GetHash())
+		if err != nil {
+			if errors.Is(err, db.ErrKeyNotFound) {
+				return rpccore.ErrBlockNotFound
+			}
+			return rpccore.ErrInternal.CloneWithData(err)
 		}
-		return rpccore.ErrStorageProofNotSupported
+		blockNumber = header.Number
```
MaksymMalicki, line 182: a small question out of curiosity, is the header.Number any different than BlockID.Number? I'm trying to understand, why do we have to get the information from the DB

rodrigo-pino (approved):

MaksymMalicki (approved):

weiihann (approved):

PR #2673, opened 2025-03-21
fix(rpc): fix processedEvents not handled if starting from pending

### LINKED ISSUE #2666: Unexpected Continuation Token Behavior with pending Blocks in starknet_getEvents
# Description

When querying events with starknet_getEvents using from_block: "pending" and to_block: "pending", the returned continuation_token remains the same across subsequent requests, preventing proper pagination. This issue does not occur when using "latest" or specific block numbers.

# How to Reproduce

- Juno version: v0.14.2
- Request 1:
```
{
  "id": 1,
  "jsonrpc": "2.0",
  "method": "starknet_getEvents",
  "params": [
    {
      "from_block": "pending",
      "to_block": "pending",
      "chunk_size": 1
    }
  ]
}
```
response might include
```
{
  "result": {
    "continuation_token": "1245472-1",
    "events": [...]
  }
}
```
- Request 2 (using the received continuation_token)
```
{
  "id": 1,
  "jsonrpc": "2.0",
  "method": "starknet_getEvents",
  "params": [
    {
      "from_block": "pending",
      "to_block": "pending",
      "chunk_size": 1,
      "continuation_token": "1245472-1"
    }
  ]
}
```

*Unexpected behavior:* The response still contains "continuation_token": "1245472-1", instead of advancing to the next chunk of events.

# Expected Behavior

When using continuation_token, subsequent requests should return a new continuation_token if there are more events to paginate through.

infrmtcs:
Should fix #2666 
The existing code check if `header.Number == cToken.fromBlock` to know if it's processing the first block and skipping the previously processed events in that case. However, if starting from pending, header.Number = 0.
The alternative solution is to reset the counter after the first block.

### DIFF
--- a/blockchain/event_filter.go
+++ b/blockchain/event_filter.go
@@ -124,9 +124,11 @@ func (e *EventFilter) Events(cToken *ContinuationToken, chunkSize uint64) ([]*Fi
 		}
 	}
 
+	var skippedEvents uint64
 	curBlock := e.fromBlock
 	// skip the blocks that we previously processed for this request
 	if cToken != nil {
+		skippedEvents = cToken.processedEvents
 		curBlock = cToken.fromBlock
 	}
 
@@ -161,7 +163,7 @@ func (e *EventFilter) Events(cToken *ContinuationToken, chunkSize uint64) ([]*Fi
 		}
 
 		var processedEvents uint64
-		matchedEvents, processedEvents, err = e.matcher.AppendBlockEvents(matchedEvents, header, receipts, cToken, chunkSize)
+		matchedEvents, processedEvents, err = e.matcher.AppendBlockEvents(matchedEvents, header, receipts, skippedEvents, chunkSize)
 		if err != nil {
 			if errors.Is(err, errChunkSizeReached) {
 				rToken = &ContinuationToken{fromBlock: curBlock, processedEvents: processedEvents}
@@ -169,6 +171,9 @@ func (e *EventFilter) Events(cToken *ContinuationToken, chunkSize uint64) ([]*Fi
 			}
 			return nil, nil, err
 		}
+
+		// Skipped events are processed, so we can reset the counter
+		skippedEvents = 0
 	}
 
 	if rToken == nil && remainingScannedBlocks == 0 && curBlock <= e.toBlock {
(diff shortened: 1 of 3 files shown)


rodrigo-pino (commented):
Is there any way we can test this behaviour?

infrmtcs:
@rodrigo-pino I used this script to test it:
```
#!/bin/bash

# Start with empty continuation token for first request
CONTINUATION_TOKEN=""

# Loop until we get a null/empty continuation token after first request
while true; do
    # For first request, don't include the continuation token
    if [[ -z "$CONTINUATION_TOKEN" ]]; then
        echo "Making initial request without continuation token"
        RESPONSE=$(curl --silent --location 'localhost:6060' \
            --data '{
            "id": 1,
            "jsonrpc": "2.0",
            "method": "starknet_getEvents",
            "params": [
                {
                "from_block": "pending",
                "to_block": "pending",
                "chunk_size": 1
                }
            ]
            }')
    else
        echo "Using continuation token: $CONTINUATION_TOKEN"
        RESPONSE=$(curl --silent --location 'localhost:6060' \
            --data '{
            "id": 1,
            "jsonrpc": "2.0",
            "method": "starknet_getEvents",
            "params": [
                {
                "from_block": "pending",
                "to_block": "pending",
                "chunk_size": 1,
                "continuation_token": "'$CONTINUATION_TOKEN'"
                }
            ]
            }')
    fi
    
    # Print the response (optional, can be removed if you don't need to see each response)
    echo "Response:"
    echo "$RESPONSE" | jq '.'
    
    # Extract the continuation token from the response
    NEW_TOKEN=$(echo "$RESPONSE" | jq -r '.result.continuation_token')
    
    # Check if we got a null or empty token
    if [[ "$NEW_TOKEN" == "null" || -z "$NEW_TOKEN" ]]; then
        echo "No more events to fetch (continuation token is null or empty)"
        break
    fi
    
    # Update the token for the next iteration
    CONTINUATION_TOKEN="$NEW_TOKEN"
    
    # Optional: Add a small delay between requests to avoid rate limiting
    sleep 0.1
done

echo "Completed fetching all events"
```

rodrigo-pino:
@infrmtcs do you think after we implement https://github.com/NethermindEth/juno/pull/2620 (which I am actively working on) we could also write infrastructure tests for event subscription in Juno?

rodrigo-pino:
I am not worried about this working now, I am more worried about a future change could break it and we could be unaware. Just thinking if we could add a test for it now, or if we can't, what are we missing so that we can write a test for it in the future

infrmtcs:
Ah, I got your point now. I thought you asked for how to reproduce it, because I couldn't too without the script. I can add a unit test for it.

infrmtcs:
I added a test to confirm the behaviour, I also tried the test on main first to confirm that it's failing before moving to the PR.

PR #2674, opened 2025-03-21
refactor(core): update the underlying felt type

rodrigo-pino:


utils/test.go
```diff
@@ -8,14 +8,22 @@ import (
...
+
 	f, err := new(felt.Felt).SetString(hex)
 	require.NoError(t, err)
-	return f
+	x := T(*f)
+	return &x
+}
+
+func HexToFelt(t testing.TB, hex string) *felt.Felt {
+	return HexTo[felt.Felt](t, hex)
```
infrmtcs, line 21: Is this missing t.Helper?

utils/test.go
```diff
@@ -8,14 +8,22 @@ import (
...
 
-func HexToFelt(t testing.TB, hex string) *felt.Felt {
+func HexTo[T ~[4]uint64](t testing.TB, hex string) *T {
 	t.Helper()
+
 	f, err := new(felt.Felt).SetString(hex)
 	require.NoError(t, err)
-	return f
+	x := T(*f)
+	return &x
```
infrmtcs, line 17: Should it be `(*T)(f)`?

infrmtcs (approved):

core/hash/tx_hash.go
```diff
@@ -1,3 +1,13 @@
 package hash
 
+import "github.com/NethermindEth/juno/core/felt"
+
 type TransactionHash Hash
+
+func (h *TransactionHash) AsFelt() *felt.Felt {
+	return (*felt.Felt)(h)
```
hudem1, line 8: shouldn't it be maybe:
```go
return (*Hash)(h).asFelt()
```

Same comment for ClassHash type, and same idea for String().

This would allow that if the Hash.asFelt() method changes, it changes for all the types based on Hash (TransactionHash, ClassHash, etc.).

hudem1: And same for ContractAddress with Address

hudem1 (approved):

utils/test.go
rodrigo-pino: That would be ideal but it seems Go doesn't support that. We can get around with `unsafe.Pointer(...)` but I don't think performance is so critical for this function

core/hash/tx_hash.go
rodrigo-pino: The idea is not have any unnecessary function call since this types will probably be on the main execution path. If the compiler is doing it's work correctly that would still be the case but I would like to have certainty. 

In my opinion this primitive types that will be used everywhere its ok if more is written if only to be extra sure.

hudem1: I see. I saw there was also the directive `//go:inline` but I understand we're not fully guaranteed it will in fact inline the fct

hudem1: Sure, no worries! :)

PR #2680, opened 2025-03-24
fix: Add ErrTraceUnavailable error

hudem1:
rpcv8 starknet_traceTransaction should return [NO_TRACE_AVAILABLE](https://github.com/starkware-libs/starknet-specs/blob/v0.8.0/api/starknet_trace_api_openrpc.json#L36)

### DIFF
--- a/rpc/v8/trace.go
+++ b/rpc/v8/trace.go
@@ -107,6 +107,17 @@ func (h *Handler) TraceTransaction(ctx context.Context, hash felt.Felt) (*Transa
 		return nil, httpHeader, rpccore.ErrTxnHashNotFound
 	}
 
+	// Check if trace available
+	status, statusErr := h.TransactionStatus(ctx, hash)
+	if statusErr != nil {
+		return nil, httpHeader, statusErr
+	}
+	if status.Finality == TxnStatusReceived {
+		return nil, httpHeader, rpccore.ErrTraceUnavailable.CloneWithData("Transaction not executed yet (RECEIVED)")
+	} else if status.Finality == TxnStatusRejected {
+		return nil, httpHeader, rpccore.ErrTraceUnavailable.CloneWithData("Transaction failed (REJECTED)")
+	}
+
 	traceResults, header, traceBlockErr := h.traceBlockTransactions(ctx, block)
 	if traceBlockErr != nil {
 		return nil, header, traceBlockErr
--- a/rpc/rpccore/rpccore.go
+++ b/rpc/rpccore/rpccore.go
@@ -35,6 +35,7 @@ type TraceCacheKey struct {
 }
 
 var (
+	ErrTraceUnavailable                 = &jsonrpc.Error{Code: 10, Message: "No trace available for transaction"}
 	ErrContractNotFound                 = &jsonrpc.Error{Code: 20, Message: "Contract not found"}
 	ErrEntrypointNotFound               = &jsonrpc.Error{Code: 21, Message: "Requested entrypoint does not exist in the contract"}
 	ErrBlockNotFound                    = &jsonrpc.Error{Code: 24, Message: "Block not found"}


rpc/v6/trace.go
```diff
@@ -86,15 +86,24 @@ type OrderedL2toL1Message struct {
 // https://github.com/starkware-libs/starknet-specs/blob/1ae810e0137cc5d175ace4554892a4f43052be56/api/starknet_trace_api_openrpc.json#L11
 func (h *Handler) TraceTransaction(ctx context.Context, hash felt.Felt) (*TransactionTrace, *jsonrpc.Error) {
 	_, blockHash, _, err := h.bcReader.Receipt(&hash)
-	if err != nil {
+	if err != nil && !errors.Is(err, db.ErrKeyNotFound) {
 		return nil, rpccore.ErrTxnHashNotFound
```
hudem1, line 90: Here I think we should change the error we return. Same for v7 & v8.
Should we change in another PR ?

rpc/v6/trace.go
```diff
@@ -86,15 +86,24 @@ type OrderedL2toL1Message struct {
...
+		}
+	}
 
 	var block *core.Block
-	isPendingBlock := blockHash == nil
-	if isPendingBlock {
+	if isPendingBlock := blockHash == nil; isPendingBlock {
 		var pending *sync.Pending
 		pending, err = h.syncReader.Pending()
 		if err != nil {
```
hudem1, line 109: this if should never get executed (dead code) now because if the tx does not exist, then the tx status check would have caught it. Wdyt ? Should we change in another PR ?

rpc/v6/trace.go
rodrigo-pino: Which error should we return?

rpc/v6/trace.go
```diff
@@ -86,15 +86,24 @@ type OrderedL2toL1Message struct {
...
+	// Check if trace is available
+	// Tx might not be in any block (including pending) but in the mempool as received or else rejected
+	if err != nil && errors.Is(err, db.ErrKeyNotFound) {
+		if status, err := h.fetchTxStatusFromFeeder(ctx, &hash); err != nil {
+			return nil, err
+		} else if status.Finality == TxnStatusReceived {
+			return nil, rpccore.ErrTraceUnavailable.CloneWithData("Transaction not executed yet (RECEIVED)")
+		} else if status.Finality == TxnStatusRejected {
+			return nil, rpccore.ErrTraceUnavailable.CloneWithData("Transaction failed (REJECTED)")
+		}
```
rianhughes, line 102: Should the data not be formatted like:
```{"status":"RECIEVED"}```?

```
"data": {
                    "type": "object",
                    "description": "Extra information on why trace is not available. Either it wasn't executed yet (RECEIVED), or the transaction failed (REJECTED)",
                    "properties": {
                        "status": {
                            "type": "string",
                            "enum": [
                                "RECEIVED",
                                "REJECTED"
                            ]
                        }
                    }
                }
                ```

rianhughes: Also, I think we should have a fall back. For example, if the transaction doesn't exist in our db (because we're behind the head of the chain), but we query the feeder, and it turns out the txn status is "accepted on L2/L1" etc.

rpc/v6/trace.go
hudem1: An internal/unexpected error. Because i dont think it makes sense to return an ErrTxnHashNotFound error here

rpc/v6/trace.go
hudem1: Mmh but for the 2nd point, I'm not sure because after, we call `h.traceBlockTransactions(ctx, block)`. That means we have to pass the block in which the tx exists, which we wouldn't have. We could fetch it from the feeder, but i think the biggest issue is if the tx is executed in the vm to generate the trace, we'd have to pass the state reader, see here
```go
executionResults, err := h.vm.Execute(block.Transactions, classes, paidFeesOnL1, &blockInfo, state, network, false, false, false, false)
```

hudem1: Oh maybe you meant have another branch in the if/else to return an error in that case ? smth like "cannot trace it, behind head"

rpc/v8/trace.go
```diff
@@ -82,9 +82,24 @@ func (h *Handler) TraceTransaction(ctx context.Context, hash felt.Felt) (*Transa
...
+		}
+	}
+
 	var block *core.Block
-	isPendingBlock := blockHash == nil
-	if isPendingBlock {
+	if isPendingBlock := blockHash == nil; isPendingBlock {
 		var pending *sync.Pending
 		pending, err = h.syncReader.Pending()
 		if err != nil {
```
hudem1, line 105: I think this if should never get executed (dead code) now because if the tx does not exist, then the tx status check would have caught it. Wdyt ? Do you agree that i remove it ? In addition to removing it, i could also add an assert here to make sure it's not nil, or else fail if it is

rpc/v8/trace_test.go
```diff
@@ -276,6 +282,44 @@ func TestTraceTransaction(t *testing.T) {
 			assert.Equal(t, rpccore.ErrTxnHashNotFound, err)
 			assert.Equal(t, httpHeader.Get(rpc.ExecutionStepsHeader), "0")
 		})
+
+		t.Run("RECEIVED tx is not traceable", func(t *testing.T) { //nolint:dupl
```
rianhughes, line 286: nice

rpc/v8/trace.go
```diff
@@ -82,9 +82,24 @@ func (h *Handler) TraceTransaction(ctx context.Context, hash felt.Felt) (*Transa
...
+		} else if status.Finality == TxnStatusReceived {
+			return nil, httpHeader, rpccore.ErrTraceUnavailable.CloneWithData(struct {
+				Status string `json:"status"`
+			}{Status: "RECEIVED"})
+		} else if status.Finality == TxnStatusRejected {
+			return nil, httpHeader, rpccore.ErrTraceUnavailable.CloneWithData(struct {
+				Status string `json:"status"`
+			}{Status: "REJECTED"})
+		}
+	}
```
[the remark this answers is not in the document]
rianhughes: Otherwise, it LGTM and I'm happy to approve

rpc/v8/trace.go
rianhughes: Yes, nice catch! We don't actually check for the txn in the pending block, maybe lets check the pending block for this txn before entering the `if err != nil && errors.Is(err, db.ErrKeyNotFound) ` flow

PR #2685, opened 2025-03-25
final sequencer PR

rianhughes:
This is the final PR to be merged before Juno can officially function as a sequencer. 

* The main addition is the `builder` pkg, which implements a new service that pops transactions from the mempool, executes them, builds a new block, and then stores it (given the verification checks pass). 
* The RPC pkg is updated so that transactions will be pushed to the `mempool` or the feeder-gateway, depending on which mode the user runs Juno in.
* The `mempool` pkg was updated to contain a reference to the `blockchain`, in place of the `state`, since we can only access the `state` via the `blockchain` type. 
* Some new sequencer related command line flags were added. 
* I also added two `make` commands to start the sequencer with a blank state (kind of useless tbh since we need accounts and fee tokens to be pre-deployed to do anything, but imo it's useful to have), and a state where we deploy accounts and fee tokens (as specified in the `genesis_prefunded_accounst.json` file)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -179,3 +179,30 @@ pathfinder: juno-cached ## Run a node to sync from pathfinder feedernode. P2P us
 
 test-fuzz: ## Run fuzzing script
 	./scripts/fuzz_all.sh
+
+sequencer:
+	./build/juno \
+	--http \
+	--http-port=6060 \
+	--http-host=0.0.0.0 \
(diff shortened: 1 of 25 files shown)


rianhughes:
https://github.com/NethermindEth/juno/compare/feature/sequencer-final-2?expand=1

PR #2686, opened 2025-03-25
perf(blockchain): add latest header cache

weiihann:
This PR improves RPC performance by caching the latest stored block header in `Blockchain` struct. Methods that require the latest block header would access the cache first before hitting the database, reducing disk IO.

### DIFF
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -102,8 +102,6 @@ func New(database db.DB, network *utils.Network) *Blockchain {
 		l1HeadFeed: feed.New[*core.L1Head](),
 	}
 
-	bc.curBlockHeader.Store(nil)
-
 	return bc
 }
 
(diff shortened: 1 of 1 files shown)


blockchain/blockchain.go
```diff
@@ -89,15 +90,21 @@ type Blockchain struct {
...
 func New(database db.DB, network *utils.Network) *Blockchain {
-	return &Blockchain{
+	bc := &Blockchain{
 		database:   database,
 		network:    network,
 		listener:   &SelectiveListener{},
 		l1HeadFeed: feed.New[*core.L1Head](),
 	}
+
+	bc.curBlockHeader.Store(nil)
```
infrmtcs, line 105: I think it's nil by default, so this is not necessary

blockchain/blockchain.go
```diff
@@ -358,6 +375,7 @@ func (b *Blockchain) Store(block *core.Block, blockCommitments *core.BlockCommit
 			return err
 		}
 
+		b.curBlockHeader.Store(block.Header)
```
infrmtcs, line 378: This is before the tx commit, so it can be available to rpc endpoints before the other data is stored to DB. I think this may result in failures.

weiihann: Changed to post-commit

cwkang1998 (approved):
Looks good, just a simple nit.

PR #2687, opened 2025-03-25
refactor: return nil when error has already been checked

threehonor:
Since err != nil has been judged before, nil is returned directly here, which is more obvious, readable and less error-prone.


infrmtcs (approved):

rodrigo-pino:
I don't like PR from bots

PR #2690, opened 2025-03-27
final sequencer PR

rianhughes:
This is the final PR to be merged before Juno can officially function as a sequencer.

* The main addition is the builder pkg. This implements a new service that pops transactions from the mempool, executes them, builds a new block with the resulting traces, and then stores the new block in the db (given the verification checks pass).
* The RPC pkg is updated so that transactions will be pushed to the mempool or the feeder-gateway, depending on which mode the user runs Juno in.
* The mempool pkg was updated to use `blockchain` instead of `state`.
* New command line flags were added.
* I also added two `make` commands to start the sequencer without a genesis state/file, and with a genesis state (using `genesis_prefund_accounts.json`).


rodrigo-pino (commented):
First round of reviews looking at the code in general. There are generally some comments for code in itself (not regarding with correctness) that I think that could be improved. 

In general:
1. Some constructor functions returning by reference instead of by Value 
2. Some methods of type `x` where `x` is nowhere used on that method. They should become functions.
3. Some functions/methods which are one liners are used only once. Functions are used to abstract functionality and/or to avoid repetition. Functions don't come for free and they also add come with code complexity. I've marked two functions which make code a bit harder to read.
4. Some methods that aren't used anywhere. I am not sure why do the exist 🤔 

I'll make another review looking at correctness later

cmd/juno/juno.go
```diff
@@ -182,6 +190,10 @@ const (
 	pluginPathUsage             = "Path to the plugin .so file"
 	logHostUsage                = "The interface on which the log level HTTP server will listen for requests."
 	logPortUsage                = "The port on which the log level HTTP server will listen for requests."
+	seqEnUsage                  = "Enables sequencer mode of operation"
+	seqBlockTimeUsage           = "Time to build a block, in seconds"
+	seqGenesisFileUsage         = "Path to the genesis file"
+	seqDisableFeesUsage         = "Skip charge fee for sequencer execution"
```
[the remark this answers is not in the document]
weiihann: I think it's fine to have the different flags instead of the file-based approach. It does slightly make the code logic complicated, but it's better for user experience. Users don't have to create some configuration file and then set the variables. Documentation is more straightforward. We also support environmental variables directly, so it's easier.

Makefile
```diff
@@ -179,3 +179,30 @@ pathfinder: juno-cached ## Run a node to sync from pathfinder feedernode. P2P us
...
+    --http-port=6060 \
+    --http-host=0.0.0.0 \
+    --db-path=../seq-db-tmp-w-accounts \
+    --log-level=debug \
+    --seq-enable \
+    --seq-block-time=1 \
+	--network sequencer \
+	--disable-l1-verification \
+	--seq-genesis-file "./genesis/genesis_prefund_accounts.json" \
+    --rpc-call-max-steps=4123000
```
weiihann, line 208: There are some flaws with these commands. We should add `juno-cached` in front and also the `db-path` should change.

Generally I'm not in favour of pushing these shortcut commands. The first reason is that we have to maintain it. The second is that it is usually shortcuts for the maintainers, not the users. If users do need to run them, then we should show appropriate documentations. I'd prefer to add it to our internal docs instead.

blockchain/blockchain.go
```diff
@@ -340,7 +342,7 @@ func (b *Blockchain) Store(block *core.Block, blockCommitments *core.BlockCommit
 			return err
 		}
 
-		if err := core.NewState(txn).Update(block.Number, stateUpdate, newClasses); err != nil {
+		if err := core.NewState(txn).Update(block.Number, stateUpdate, newClasses, true); err != nil {
```
weiihann, line 345: this shouldn't be true

node/node.go
```diff
@@ -159,54 +171,82 @@ func New(cfg *Config, version string, logLevel *utils.LogLevel) (*Node, error) {
...
 		}
-		log.Warnw("P2P features enabled. Please note P2P is in experimental stage")
-
-		if !cfg.P2PFeederNode {
-			// Do not start the feeder synchronisation
-			synchronizer = nil
-		}
-		p2pService, err = p2p.New(cfg.P2PAddr, cfg.P2PPublicAddr, version, cfg.P2PPeers, cfg.P2PPrivateKey, cfg.P2PFeederNode,
-			chain, &cfg.Network, log, database)
+		poolDB, err := pebble.NewMem()
```
weiihann, line 197: Is it intended to use an in-memory database? Upon shutdown or crashes the txns will not be persisted

node/node.go
```diff
@@ -159,54 +171,82 @@ func New(cfg *Config, version string, logLevel *utils.LogLevel) (*Node, error) {
...
 		if err != nil {
-			return nil, fmt.Errorf("set up p2p service: %w", err)
+			return nil, err
+		}
+		mempool, mempoolCloser := mempool.New(poolDB, chain, mempoolLimit, log)
+
+		sequencer := builder.New(pKey, new(felt.Felt).SetUint64(1337), chain, nodeVM, //nolint:mnd
```
weiihann, line 203: Not too sure what the 1337 means, is there a special meaning behind it?

node/node.go
```diff
@@ -159,54 +171,82 @@ func New(cfg *Config, version string, logLevel *utils.LogLevel) (*Node, error) {
...
-			synchronizer = nil
-		}
-		p2pService, err = p2p.New(cfg.P2PAddr, cfg.P2PPublicAddr, version, cfg.P2PPeers, cfg.P2PPrivateKey, cfg.P2PFeederNode,
-			chain, &cfg.Network, log, database)
+		poolDB, err := pebble.NewMem()
 		if err != nil {
-			return nil, fmt.Errorf("set up p2p service: %w", err)
+			return nil, err
+		}
+		mempool, mempoolCloser := mempool.New(poolDB, chain, mempoolLimit, log)
```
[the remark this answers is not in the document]
weiihann: Yeah I agree. The mempool should have a public `Close` method. The creator of this mempool object should call `Close` when it closes.

cmd/juno/juno.go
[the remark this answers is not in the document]
rodrigo-pino: The rationale is that I don't want to have four flags for en experimental feature of Juno, almost nobody is going to use it as a sequencer for now. I agree that when we move to a more sequence-centric project this could be re-added

Also, this remind me that this changes should be reflected in both the readme and the docs

Makefile
rodrigo-pino: I agree with these points. Makes sense to move this (not exactly like that) to our docs. We should add an extra page, called sequencing

cmd/juno/juno_test.go
```diff
@@ -276,7 +282,7 @@ db-path: /home/.juno
 network: sepolia
 pprof: true
 `,
-			expectedConfig: &node.Config{
+			expectedConfig: &node.Config{ //nolint:dupl // false trigger (see Pprof,DatabasePath)
```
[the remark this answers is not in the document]
rianhughes: The linter is claiming that the lines are duplicated, however this is not true. For example, the Pprof and DatabasePath fields are different

Makefile
rianhughes: Imo we should have a shortcut command to run Juno in the sequencer mode, it signifcantly increases the user-experience of running Juno as a sequencer, with very little cost to us. We already have a team in Nethermind that is running Juno as a sequencer, and it's very useful for them to have a consistent reference point for how to get Juno up and running quickly, instead of trying to figure things out themselves, or reaching out to the Juno team, it just adds unneccesary friction imo.

node/node.go
rianhughes: Nope this was a mistake, thanks for pointing it out. We should use the persistent DB

node/node.go
rianhughes: Pushed to a const. 

Also, it doesn't have any special meaning. 
https://en.wikipedia.org/wiki/Leet

builder/builder.go
```diff
@@ -0,0 +1,463 @@
...
+	}
+	return b.InitPendingBlock()
+}
+
+func (b *Builder) StartingBlockNumber() (uint64, error) {
+	return 0, nil
+}
+
+func (b *Builder) HighestBlockHeader() *core.Header {
+	return nil
```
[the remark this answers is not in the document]
rianhughes: The Builder service needs to implement the Reader interface 

```
type Reader interface {
    StartingBlockNumber() (uint64, error)
    HighestBlockHeader() *core.Header
    SubscribeNewHeads() NewHeadSubscription
    SubscribeReorg() ReorgSubscription
    SubscribePending() PendingSubscription

    Pending() (*Pending, error)
    PendingBlock() *core.Block
    PendingState() (core.StateReader, func() error, error)
}
```

builder/builder.go
```diff
@@ -0,0 +1,463 @@
...
+}
+
+// Sign returns the builder's signature over data.
+func (b *Builder) Sign(blockHash, stateDiffCommitment *felt.Felt) ([]*felt.Felt, error) {
+	data := crypto.PoseidonArray(blockHash, stateDiffCommitment).Bytes()
+	signatureBytes, err := b.privKey.Sign(data[:], nil)
+	if err != nil {
+		return nil, err
+	}
+	sig := make([]*felt.Felt, 0)
```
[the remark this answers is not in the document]
rianhughes: We can, I'll update it. The original reason is because the core block heade defines signatures as `[][]*felt.Felt`

builder/builder.go
```diff
@@ -0,0 +1,463 @@
...
+	// Update pending block with transaction results
+	b.updatePendingBlock(pending, receipt, txn.Transaction, seqTrace)
+
+	return b.StorePending(pending)
+}
+
+// hasDeclaredClasses checks if the state diff contains declared classes
+func (b *Builder) hasDeclaredClasses(stateDiff *vm.StateDiff) bool {
+	return stateDiff.DeclaredClasses != nil || stateDiff.DeprecatedDeclaredClasses != nil
+}
```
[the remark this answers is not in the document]
rianhughes: I inlined the function. 

I'm also not sure what you mean by this will be true 99% of the time, as this is a check we need to make on a case by case basis. It's also not clear that it will be true 99% of the time, and imo, it actually won't be. Eg most Juno blocks wont declare new classes.

builder/builder.go
```diff
@@ -0,0 +1,463 @@
...
+	}
+	b.pendingBlock.Store(newPending)
+	return nil
+}
+
+// The builder has no reorg logic (centralised sequencer that can't reorg)
+func (b *Builder) SubscribeReorg() sync.ReorgSubscription {
+	return sync.ReorgSubscription{Subscription: b.subReorgFeed.Subscribe()}
+}
+
```
rianhughes, line 456: The builder needs to satisfy
```
type Reader interface {
	StartingBlockNumber() (uint64, error)
	HighestBlockHeader() *core.Header
	SubscribeNewHeads() NewHeadSubscription
	SubscribeReorg() ReorgSubscription
	SubscribePending() PendingSubscription

	Pending() (*Pending, error)
	PendingBlock() *core.Block
	PendingState() (core.StateReader, func() error, error)
}
```

Makefile
weiihann: Anyone who wants to run sequencer will almost certainly have their own configurations. Particularly, the `db-path` will likely differ. So no one will actually run the shortcut.

Regarding reference point, the documentation should be the one. We shouldn't have to maintain 2 separate reference points.

cmd/juno/juno.go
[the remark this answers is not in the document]
weiihann: IMO, as of current stage, there are no differences with both approaches. However, we eventually will use flags-based approach because it's better for UX and also easier for devops deployment. So I'd rather maintain the current approach as it is.

node/node.go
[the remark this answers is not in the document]
rianhughes: Updated so that `Close` now belongs to the `mempool`

builder/builder.go
```diff
@@ -0,0 +1,463 @@
...
+	bc              *blockchain.Blockchain
+	db              db.DB
+	vm              vm.VM
+	log             utils.Logger
+	subNewHeads     *feed.Feed[*core.Block]
+	subPendingBlock *feed.Feed[*core.Block]
+	subReorgFeed    *feed.Feed[*sync.ReorgBlockRange]
+	mempool         *mempool.Pool
+	mempoolCloser   func() error
```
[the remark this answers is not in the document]
rianhughes: Updated so that `Close` now belongs to the `mempool`

Makefile
rianhughes: I removed these commands and added a new sequencer.md doc

builder/builder.go
```diff
@@ -0,0 +1,457 @@
...
+		if pErr := b.ClearPending(); pErr != nil {
+			b.log.Errorw("clearing pending", "err", pErr)
+		}
+	}()
+
+	if err := b.InitPendingBlock(); err != nil {
+		return err
+	}
+
+	doneListen := make(chan struct{})
```
rodrigo-pino, line 136: I am not very sure of this var name,

builder/builder.go
```diff
@@ -0,0 +1,457 @@
...
+			b.finaliseMutex.Lock()
+			err := b.Finalise(b.Sign)
+			b.log.Infof("Finalised new block")
+			b.finaliseMutex.Unlock()
+			if err != nil {
+				return err
+			}
+		}
+	}
+}
```
rodrigo-pino, line 161: Out of curiosity, any reason you chose this approach as a way to signal the end of a Go routine instead of a wait group?

builder/builder.go
```diff
@@ -0,0 +1,463 @@
...
+	header, err := b.bc.HeadsHeader()
+	if err != nil {
+		return err
+	}
+	pendingBlock := core.Block{
+		Header: &core.Header{
+			ParentHash:       header.Hash,
+			Number:           header.Number + 1,
+			SequencerAddress: &b.ownAddress,
+			// Todo: dynamically set fees (follow-up PR)
```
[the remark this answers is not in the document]
rodrigo-pino: Let's remove this Todo then, there is nothing we can do it at the moment until we have a market fee. We will probably get back to it as part of the Consensus development. Leaving it as a Todo now might imply that the current implementation is incomplete while it is not.

builder/builder.go
```diff
@@ -0,0 +1,457 @@
...
+		StateUpdate: &su,
+		NewClasses:  newClasses,
+	}
+	b.pendingBlock.Store(&pending)
+	b.headState, b.headCloser, err = b.bc.HeadState()
+	return err
+}
+
+// Finalise the pending block and initialise the next one
+func (b *Builder) Finalise(signFunc blockchain.BlockSignFunc) error {
```
rodrigo-pino, line 214: call it `sign` only to match it's previous name or call the others `signFunc`. The name is not the problem, is referring to it differently in different contexts (small nit)

builder/builder.go
```diff
@@ -0,0 +1,457 @@
...
+	}
+	b.log.Infow("Finalised block", "number", pending.Block.Number, "hash",
+		pending.Block.Hash.ShortString(), "state", pending.Block.GlobalStateRoot.ShortString())
+
+	if b.plugin != nil {
+		err := b.plugin.NewBlock(pending.Block, pending.StateUpdate, pending.NewClasses)
+		if err != nil {
+			b.log.Errorw("error sending new block to plugin", err)
+		}
+	}
```
rodrigo-pino, line 230: Nice idea of integrating plugins 👍

builder/builder.go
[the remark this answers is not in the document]
rodrigo-pino: And should it be always be nil? Why?

builder/builder.go
```diff
@@ -0,0 +1,457 @@
...
+}
+
+func (b *Builder) getRevealedBlockHash() (*felt.Felt, error) {
+	blockHeight, err := b.bc.Height()
+	if err != nil {
+		return nil, err
+	}
+	const blockHashLag = 10
```
rodrigo-pino, line 327: What is `blockHashLag`?

builder/builder.go
```diff
@@ -0,0 +1,457 @@
...
+	header, err := b.bc.BlockHeaderByNumber(blockHeight - blockHashLag)
+	if err != nil {
+		return nil, err
+	}
+	return header.Hash, nil
+}
+
+// runTxn executes the provided transaction and applies the state changes
+// to the pending state
+func (b *Builder) runTxn(txn *mempool.BroadcastedTransaction, blockHashToBeRevealed *felt.Felt) error {
```
rodrigo-pino, line 341: Why executing transactions one by one, instead of running all of them at the same time through `vm.Execute`?

builder/builder.go
```diff
@@ -0,0 +1,457 @@
...
+	}
+	for {
+		b.finaliseMutex.RLock()
+		userTxn, err := b.mempool.Pop()
+		if err != nil {
+			b.finaliseMutex.RUnlock()
+			return err
+		}
+		b.log.Debugw("running txn", "hash", userTxn.Transaction.Hash().String())
+		if err = b.runTxn(&userTxn, blockHashToBeRevealed); err != nil {
```
rodrigo-pino, line 304: I think you should send everything that's available on the mempool right now, no need to pop one by one

builder/builder.go
```diff
@@ -0,0 +1,457 @@
...
+		&vm.BlockInfo{
+			Header:                pending.Block.Header,
+			BlockHashToBeRevealed: blockHashToBeRevealed,
+		},
+		state,
+		b.bc.Network(),
+		false, false, false, true)
+	if err != nil {
+		return err
+	}
```
rodrigo-pino, line 375: The performance gains come here where everything in this function before and after this block of code gets executed only once instead of `n` where `n` is the amount of transactions in the pool at the moment

builder/init_test.go
rodrigo-pino, line 1: Can you explain this file?

docs/docs/sequencer.md
rodrigo-pino, line 1: 👍

builder/builder.go
rianhughes: It just seemed like the natural thing to do, but I agree executing in batches seems more efficient

builder/init_test.go
rianhughes: Yup, it just registers the custom Juno types with the encoder. We use it throughout the codebase (blockchain_test, node_test, etc)

builder/builder.go
rianhughes: Imo it's cleaner (and maybe more go idiomatic) to use a channel like this when we're just trying to signal a one off event for a single go routine.

adapters/vm2core/vm2core.go
```diff
@@ -87,3 +87,65 @@ func AdaptStateDiff(fromStateDiff *vm.StateDiff) core.StateDiff {
...
 }
+
+func Receipt(fee *felt.Felt, txn core.Transaction,
+	trace *vm.TransactionTrace, txnReceipt *vm.TransactionReceipt,
+) core.TransactionReceipt {
+	// Determine fee unit based on transaction version
+	feeUnit := core.WEI
+	if txn.TxVersion().Is(3) {
+		feeUnit = core.STRK
+	}
```
rodrigo-pino, line 98: Would it be relevant to add a transaction v3? Not to test this code spceifically but test everything with Transaction V3 🤔, or is only this function that's missing tests with v3 transactions?

If that's the case then we can ignore this

adapters/vm2core/vm2core.go
```diff
@@ -87,3 +87,65 @@ func AdaptStateDiff(fromStateDiff *vm.StateDiff) core.StateDiff {
...
+
+	adaptedER := AdaptExecutionResources(trace.TotalExecutionResources(), &txnReceipt.Gas)
+	isReverted := trace.ExecuteInvocation != nil && trace.ExecuteInvocation.FunctionInvocation == nil
+	return core.TransactionReceipt{
+		Fee:                fee,
+		FeeUnit:            feeUnit,
+		Events:             AdaptOrderedEvents(trace.AllEvents()),
+		ExecutionResources: &adaptedER,
+		L1ToL2Message:      nil, // Todo: sequencer currently can't post messages to L1 (follow up PR)
```
rodrigo-pino, line 107: Let's remove the part that says "(follow up PR)". Do we have a check that make sure we don't accept that kind of transactions 🤔

builder/builder.go
```diff
@@ -0,0 +1,469 @@
...
+	"github.com/NethermindEth/juno/sync"
+	"github.com/NethermindEth/juno/utils"
+	"github.com/NethermindEth/juno/vm"
+	"github.com/consensys/gnark-crypto/ecc/stark-curve/ecdsa"
+)
+
+var (
+	_                     service.Service = (*Builder)(nil)
+	_                     sync.Reader     = (*Builder)(nil)
+	NumTxnsToBatchExecute                 = 10
```
rodrigo-pino, line 30: Why through batches and not all the transactions in the mempool? What do you think if we make this number modifiable via flag

builder/builder.go
```diff
@@ -0,0 +1,469 @@
...
+	if err != nil {
+		return err
+	}
+	pendingBlock := core.Block{
+		Header: &core.Header{
+			ParentHash:       header.Hash,
+			Number:           header.Number + 1,
+			SequencerAddress: &b.ownAddress,
+			L1GasPriceETH:    felt.One.Clone(),
+			L1GasPriceSTRK:   felt.One.Clone(),
```
rodrigo-pino, line 188: Why clone it? Are they going to be mutated later on?

builder/builder.go
```diff
@@ -0,0 +1,469 @@
...
+			L1DAMode:         core.Calldata,
+			L1DataGasPrice: &core.GasPrice{
+				PriceInWei: felt.One.Clone(),
+				PriceInFri: felt.One.Clone(),
```
rodrigo-pino, line 192: Same question here? I think we are safe all of them sharing the reference

builder/builder.go
rianhughes: I think we should try and control / cap the number of transactions we try and execute at once. If we tried to execute everything in the mempool at once, we might see massive resource spikes which could cause problems etc. It just feels less controlled.

Regarding making it a flag, we already have a ton of flags and this one doesn't feel very necessary (esp for an experimental feature). We can always add it later if users want it imo

builder/builder.go
rianhughes: Originally I used `new(felt.Felt).SetUint64(1)`, but replaced it with a `clone` because you asked me to in a previous comment haha :) In principle we don't modify this value, but I don't think we should directly reference `felt.One` for safety reasons (and hope it never gets modified throughout the course of the project)

builder/builder.go
rianhughes: I'm personally against using references to felt.One for safety reasons, we can't be certain it won't be modified as the project changes, and there's no real benefit to using the reference as far as I can see

rianhughes: Screenshot from 2025-04-11 15-57-12

adapters/vm2core/vm2core.go
rianhughes: There's currently no way for a user to send these types of transactions to the sequencer. The only way to get l1-handler transactions is by interacting with a core contract on l1, which isn't relevant for the Juno sequencer right now

builder/builder.go
weiihann: Definitely not all transactions in the mempool, that will easily crash the process and becomes a vulnerability. I think we eventually will change the default number of txns from 10 to something else, but that's after we have monitored the system.

builder/builder.go
```diff
@@ -0,0 +1,469 @@
...
+		close(doneListen)
+	}()
+
+	for {
+		select {
+		case <-ctx.Done():
+			<-doneListen
+			return nil
+		case <-time.After(b.blockTime):
+			b.finaliseMutex.Lock()
```
weiihann, line 153: The `finaliseMutex` should be able to be included in the `Finalise` method

builder/builder.go
```diff
@@ -0,0 +1,469 @@
...
+func (b *Builder) listenPool(ctx context.Context) error {
+	for {
+		if err := b.depletePool(ctx); err != nil {
+			if !errors.Is(err, mempool.ErrTxnPoolEmpty) {
+				return err
+			}
+		}
+
+		// push the pending block to the feed
+		b.subPendingBlock.Send(b.PendingBlock())
```
weiihann, line 277: My understanding is that the pending block created after executing the transactions are not the final pending block right? If that's true then we should send the pending block to the feed after we finalize the pending block (i.e. move this to `Finalize()`).

builder/builder.go
rianhughes: +1

builder/builder.go
rianhughes: Yup, I just moved this to the Finalise function

builder/builder.go
rianhughes: I think the current approach is correct. Imo we should push the pending block every time it gets updated from executing transactions. if we only pushed the pending block when finalising it, we might as well just push the latest committed block, which defeats the purpose of the pending block subscription. Eg, imagine pending block times were 5minutes, users should be able to see what the pending block includes throughout this period, not just when we decide to commit it to the chain.

> My understanding is that the pending block created after executing the transactions are not the final pending block right? 

Yes you're right, the pending block is not finalised/committed until we call Finalise(). Up until that point, it can change as we execute transactions from the mempool.

builder/builder.go
rodrigo-pino: Ok 👍

builder/builder.go
```diff
@@ -0,0 +1,470 @@
...
+		return nil, nil, utils.RunAndWrapOnError(txn.Discard, err)
+	}
+
+	return sync.NewPendingState(pending.StateUpdate.StateDiff, pending.NewClasses, b.headState), txn.Discard, nil
+}
+
+func (b *Builder) Run(ctx context.Context) error {
+	defer func() {
+		b.mempool.Close()
+	}()
```
weiihann, line 125: `defer b.mempool.Close()` should be fine

weiihann (approved):

adapters/vm2core/vm2core.go
rodrigo-pino: Not important, let's ignore it for now

mempool/mempool.go
```diff
@@ -69,11 +74,42 @@ func (t *memTxnList) pop() (BroadcastedTransaction, error) {
...
+		numToPop = t.len
+	}
+
+	result := make([]BroadcastedTransaction, 0, numToPop)
+	current := t.head
+
+	for range numToPop {
+		result = append(result, current.Txn)
+		current = current.Next
+	}
```
rodrigo-pino, line 96: We have benchmarked results that this approach is several times slower (in the miliseconds range) than
```go
for i := range numToPop {
   result[i] = current.Txn
}
```

rodrigo-pino (approved):

rianhughes:
The broken test seems unrelated to the changes here https://github.com/NethermindEth/juno/pull/2776

PR #2691, opened 2025-03-27
feat(node): dynamic windows for feeder http requests

MaksymMalicki:
This PR implements a dynamic window for feeder gateway HTTP requests. The problem arises when user has network connection problems, or tries to fetch big blocks like [this one](https://alpha-mainnet.starknet.io/feeder_gateway/get_block?blockNumber=1256131). Currently, Juno allows user to specify the HTTP request timeout for feeder gateway, using the `gw-timeout` flag, or takes 5s timeout by default. This default timeout might be too small for big blocks and cause a loop, in which there are many ineffective request tries.

failed attempts

This new implementation works as follows:

1. The `gw-timeouts` flag was modified to be `string` instead of `duration`. User can pass either a single value, like `"10s"`, or a list of steps `"10s,15s,20s"`. If user provides single value with a trailing comma at the end (like `5s`), don't use dynamic timeouts, but the fixed one.
2. After that, Juno generates list of steps for the timeouts:

- If user specified a single value, generate a list of steps, by applying the multipliers below, until the `timeoutsCount` (30) is met. Multipliers: `1.5x` for durations under `45 seconds` and `1.2x` for durations above `45 seconds`
- If user specified a list of values: cap to the `timeoutsCount` if length greater. If not, use the list as it is, without generating new timeouts.
- If the ascending order is not followed, throw an error.
- During the generation of the steps, all the values are rounded up to integers.
- Default value is fixed 5 seconds (`5s,`)

3. Hold the information about the current timeout in the `Timeouts` struct, which is a part of `Client`. If the request was successful, use the `i-1` time window from the list of steps. If the request failed, try to make the request with the `i+1` time window from the list of steps.
4. The `logHostF` and `logPortF` flags were modified to `httpUpdateHostF` and `httpUpdatePortF`. In the new implementation, user can pass new values of the timeouts during execution, using `HTTP PUT` request to the specified address, using `/feeder/timeouts`, with query params (e.g. `?timeouts=2s` or `?timeouts=2s,3s,4s`). (Implemented [here](https://github.com/NethermindEth/juno/pull/2720)

In the following example, the default timeouts are deliberately set to `gw-timeouts=1s,2s,3s`
good attempts

Here is the update of the timeouts using HTTP shown:
image
http_many


clients/feeder/timeouts_utils.go
```diff
@@ -0,0 +1,128 @@
...
+		} else {
+			count := maxRetries + 1 - len(timeouts)
+			next := getNextTimeout(timeouts[len(timeouts)-1])
+			timeouts := append(timeouts, generateTimeoutsListFromInitial(next, count)...)
+			return timeouts
+		}
+	}
+}
+
+func getNextTimeout(prev time.Duration) time.Duration {
```
rodrigo-pino, line 59: Change name to `increaseDuration`

clients/feeder/timeouts_utils.go
```diff
@@ -0,0 +1,128 @@
...
+
+	timeouts := make(TimeoutsList, 0, len(values))
+	for _, v := range values {
+		d, err := time.ParseDuration(v)
+		if err != nil {
+			return nil, fmt.Errorf("parsing timeouts: %v", err)
+		}
+		if d > maxTimeout {
+			d = maxTimeout
+		}
```
rodrigo-pino, line 87: Why do you have this constraint, If people want to have a timeout of 10000 hours, it's their decision

clients/feeder/feeder.go
```diff
@@ -36,6 +36,8 @@ type Client struct {
 	userAgent  string
 	apiKey     string
 	listener   EventListener
+	timeouts   TimeoutsList
+	curTimeout int
```
rodrigo-pino, line 40: It looks to me that is the job of `TimeoutList` to know what the `curTimeout` is, meaning that `TimeoutsList` should be the one holding the index and not `Client`

clients/feeder/feeder.go
```diff
@@ -249,11 +256,15 @@ func (c *Client) get(ctx context.Context, queryURL string) (io.ReadCloser, error
...
 
+			c.client.Timeout = c.timeouts[c.curTimeout]
 			reqTimer := time.Now()
 			res, err = c.client.Do(req)
 			if err == nil {
 				c.listener.OnResponse(req.URL.Path, res.StatusCode, time.Since(reqTimer))
 				if res.StatusCode == http.StatusOK {
+					if c.curTimeout > 0 {
+						c.curTimeout--
+					}
```
rodrigo-pino, line 267: This new lines should be all logic managed by the `TimeoutList` type

clients/feeder/timeouts_utils.go
```diff
@@ -0,0 +1,128 @@
...
+
+const (
+	maxTimeout        = 2 * time.Minute
+	defaultTimeout    = 5 * time.Second
+	growthFactorFast  = 1.5
+	growthFactorSlow  = 1.2
+	fastGrowThreshold = 45 * time.Second
+)
+
+type TimeoutsList []time.Duration
```
rodrigo-pino, line 20: Nitpick, why not simply call it `Timeouts`?

clients/feeder/timeouts_utils.go
```diff
@@ -0,0 +1,128 @@
...
+		seconds := math.Ceil(float64(prev.Seconds()) * growthFactorSlow)
+		next = time.Duration(seconds) * time.Second
+	}
+	if next > maxTimeout {
+		next = maxTimeout
+	}
+	return next
+}
+
+func ParseTimeouts(values []string) (TimeoutsList, error) {
```
rodrigo-pino, line 74: Thank you for returning a type instead of a reference 👍

node/node.go
```diff
@@ -159,8 +159,17 @@ func New(cfg *Config, version string, logLevel *utils.LogLevel) (*Node, error) {
...
+	timeouts, err := feeder.ParseTimeouts(cfg.GatewayTimeouts)
+	if err != nil {
+		return nil, fmt.Errorf("invalid gateway timeout: %w", err)
+	}
+
+	client := feeder.NewClient(cfg.Network.FeederURL).
+		WithUserAgent(ua).
+		WithLogger(log).
+		WithTimeouts(timeouts).
+		WithAPIKey(cfg.GatewayAPIKey)
```
rodrigo-pino, line 171: Better refactoring, thank you 👍

clients/feeder/feeder.go
```diff
@@ -469,3 +489,14 @@ func findTargetDirectory(targetRelPath string) (string, error) {
...
+// For testing purposes
+func (c *Client) GetCurrentTimeout() int {
+	return c.curTimeout
+}
+
+// For testing purposes
+func (c *Client) SetHTTPTransport(transport http.RoundTripper) *Client {
+	c.client.Transport = transport
+	return c
+}
```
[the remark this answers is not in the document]
rodrigo-pino: I am aware that Juno codes suffers from this, but let's not perpetuate it

clients/feeder/feeder.go
```diff
@@ -262,14 +273,23 @@ func (c *Client) get(ctx context.Context, queryURL string) (io.ReadCloser, error
...
 			wait = c.backoff(wait)
 			if wait > c.maxWait {
 				wait = c.maxWait
 			}
-			c.log.Debugw("Failed query to feeder, retrying...", "req", req.URL.String(), "retryAfter", wait.String(), "err", err)
+			c.log.Debugw("Failed query to feeder, retrying...",
+				"req", req.URL.String(),
+				"retryAfter", wait.String(),
+				"err", err,
+			)
```
rodrigo-pino, line 292: This should be debug only when it is on the fast growing stage, after, it has to be a Warning so everyone is aware

clients/feeder/feeder.go
```diff
@@ -232,6 +235,10 @@ func (c *Client) get(ctx context.Context, queryURL string) (io.ReadCloser, error
 	var res *http.Response
 	var err error
 	wait := time.Duration(0)
+	if c.timeouts == nil {
+		c.timeouts = generateTimeoutsListFromInitial(defaultTimeout, c.maxRetries+1)
```
rodrigo-pino, line 239: I don't understand why `maxRetries` is part of the constructor? How is maxRetries related with the timeouts array?

clients/feeder/feeder.go
```diff
@@ -262,14 +273,23 @@ func (c *Client) get(ctx context.Context, queryURL string) (io.ReadCloser, error
 				res.Body.Close()
 			}
 
+			c.curTimeout++
+			if c.curTimeout >= len(c.timeouts) {
+				c.curTimeout = len(c.timeouts) - 1
+			}
+
```
rodrigo-pino, line 280: This functionality should be part of the `Timeouts` type

clients/feeder/feeder_test.go
```diff
@@ -577,6 +582,83 @@ func TestBackoffFailure(t *testing.T) {
...
+		require.NoError(t, err)
+		require.NotNil(t, block)
+
+		require.Equal(t, 2, client.GetCurrentTimeout())
+	})
+
+	t.Run("fails when max retries is insufficient", func(t *testing.T) {
+		transport := &mockTransport{
+			requestDuration: requestDuration,
+		}
```
rodrigo-pino, line 648: It's good there is a test for max retries, but you have to be aware that maxRetries and Timeouts are not relevant to each other.

clients/feeder/timeouts_utils.go
rodrigo-pino: Also, you have to have a check that number `i` is always bigger than `i-1`. Please add a test for this as well.

clients/feeder/timeouts_utils.go
```diff
@@ -0,0 +1,128 @@
...
+	if len(values) == 0 {
+		return TimeoutsList{defaultTimeout}, nil
+	}
+
+	timeouts := make(TimeoutsList, 0, len(values))
+	for _, v := range values {
+		d, err := time.ParseDuration(v)
+		if err != nil {
+			return nil, fmt.Errorf("parsing timeouts: %v", err)
```
rodrigo-pino, line 83: give index information in the error

rodrigo-pino (commented):
The code looks real good and most of my comments are with code quality and not correctness.

Now, when we discussed this issue I believed we agree on a two PR approach, one PR for the timeout growth and another for setting the timeout window dynamically. Perhaps it was not very clear on my side, but I think it's preferable to keep it. Could you please uncouple the code and make a second PR where you add the functionality to change the timeout manually during runtime

clients/feeder/timeouts_utils.go
```diff
@@ -0,0 +1,128 @@
...
+	maxTimeout        = 2 * time.Minute
+	defaultTimeout    = 5 * time.Second
+	growthFactorFast  = 1.5
+	growthFactorSlow  = 1.2
+	fastGrowThreshold = 45 * time.Second
+)
+
+type TimeoutsList []time.Duration
+
+func generateTimeoutsListFromInitial(initial time.Duration, count int) TimeoutsList {
```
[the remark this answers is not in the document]
rodrigo-pino: Since it is private, I think this is no longer relevant.

Something I forgot to say is that the timeout array when generated it should've a fixed size. It is not related with maxRetires, it can be 10 or 20. Enough for when using the default it gets' to 5 minutes. Maybe of size 20 or size 30.

clients/feeder/feeder.go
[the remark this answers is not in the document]
MaksymMalicki: Removed this, modified tests

clients/feeder/timeouts_utils.go
```diff
@@ -0,0 +1,128 @@
...
+	"time"
+)
+
+const (
+	maxTimeout        = 2 * time.Minute
+	defaultTimeout    = 5 * time.Second
+	growthFactorFast  = 1.5
+	growthFactorSlow  = 1.2
+	fastGrowThreshold = 45 * time.Second
+)
```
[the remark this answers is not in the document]
MaksymMalicki: done

clients/feeder/timeouts_utils.go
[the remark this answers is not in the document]
MaksymMalicki: yes I agree, set to 30. I thought about this, that we need to have sufficient items in the list for all retries to potentially fail with different value. But with the new logging (always >1min), this works

clients/feeder/timeouts_utils.go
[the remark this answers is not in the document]
MaksymMalicki: changed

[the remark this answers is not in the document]
MaksymMalicki: done

clients/feeder/timeouts_utils.go
MaksymMalicki: done

clients/feeder/timeouts_utils.go
MaksymMalicki: Yes agree with the above. Right now the values are being sorted in `getTimeouts` and tested in `TestGetTimeouts`

clients/feeder/feeder.go
MaksymMalicki: Moved to the `Timeouts` struct

clients/feeder/feeder.go
MaksymMalicki: moved

clients/feeder/feeder.go
MaksymMalicki: Yes agree, as mentioned in the comment above, I thought it was a good idea to have a different value for every potential fail.

clients/feeder/feeder.go
MaksymMalicki: moved

clients/feeder/timeouts_utils.go
MaksymMalicki: added

clients/feeder/timeouts_test.go
```diff
@@ -0,0 +1,105 @@
...
+		want       Timeouts
+	}{
+		{
+			name:       "empty input",
+			input:      []time.Duration{},
+			maxRetries: 4,
+			want: Timeouts{
+				curTimeout: 0,
+				timeouts:   []time.Duration{5 * time.Second, 8 * time.Second, 12 * time.Second, 18 * time.Second, 27 * time.Second},
+			},
```
rodrigo-pino, line 68: From this test it seems that the timeout array is still connected with the amount of retries. Max retires means the max amount of continuous retries, it should have 0 influence on the timeouts array.

The timeout array should be of fixed size 30, max retries can be 5 or a 1000, it shouldn't affect it. 

If you have 5 max retries and you are step `i`:
1. You fail 4 times, mean you are now on `i + 4`:
2. You have one succesful trie, you are now on `i + 3`
3. You fail 4 times, mean you are now on `i + 7`
4. and so on

Even if the max retries it's 5 it still possible to use more than that.

rodrigo-pino (commented):
Looks very good, please add as part of this PR the necessary documentation specifying how to use these new changes

clients/feeder/timeouts_test.go
MaksymMalicki: Yess, I understand this. The maxRetries in this test is just an arbitrarily small number, I can manually pass for example 5, so its more readable and clear (that timeouts list is in no way connected to the maxRetries number).

MaksymMalicki: Or just test it against 30 values, I just realized, that after the refactorings the `count` param is used only for tests

rodrigo-pino: I think is not necessary to write the 30 values, if it is not related then all good, I think ti shouldn't be related in the tests either.

rodrigo-pino (commented):
Once documentation is done I'll give the final review

cmd/juno/juno.go
```diff
@@ -174,8 +174,9 @@ const (
 	dbCacheSizeUsage     = "Determines the amount of memory (in megabytes) allocated for caching data in the database."
 	dbMaxHandlesUsage    = "A soft limit on the number of open files that can be used by the DB"
 	gwAPIKeyUsage        = "API key for gateway endpoints to avoid throttling" //nolint: gosec
-	gwTimeoutUsage       = "Timeout for requests made to the gateway"          //nolint: gosec
-	callMaxStepsUsage    = "Maximum number of steps to be executed in starknet_call requests. " +
+	gwTimeoutsUsage      = "Timeouts for requests made to the gateway. Can be specified as a single value or comma-separated list" +
+		"or multiple flags. Values over 2m will be capped. Example: --gw-timeouts=5s,7s,10s"
```
rodrigo-pino: Modify this so that the following information is clear:
1. If only a number is set, after a fail retry the timeout will increase
2. If a list is set, the values of that list will use after a retry
3. To avoid the dynamic windows and setting a single value set it as `5s,` (trailing comma)

cmd/juno/juno.go
```diff
@@ -174,8 +174,9 @@ const (
 	dbCacheSizeUsage     = "Determines the amount of memory (in megabytes) allocated for caching data in the database."
 	dbMaxHandlesUsage    = "A soft limit on the number of open files that can be used by the DB"
 	gwAPIKeyUsage        = "API key for gateway endpoints to avoid throttling" //nolint: gosec
-	gwTimeoutUsage       = "Timeout for requests made to the gateway"          //nolint: gosec
-	callMaxStepsUsage    = "Maximum number of steps to be executed in starknet_call requests. " +
+	gwTimeoutsUsage      = "Timeouts for requests made to the gateway. Can be specified as a single value or comma-separated list" +
+		"Example: --gw-timeouts=5s,7s,10s. The system will automatically adjust timeouts based on request performance."
```
rodrigo-pino, line 178: Modify this so that the following information is clear:
1. If only a number is set, after a fail retry the timeout will increase
2. If a list is set, the values of that list will use after a retry
3. To avoid the dynamic windows and setting a single value set it as `5s,` (trailing comma)

rodrigo-pino: If someone puts a list of `5s, 7s, 10s` then that's the timeout list. If someone puts a list of `2m, 5m, 3h` then that's ok as well and we respect it. There shouldn't be any cap for custom values. 

If someone puts a single value bigger than the max value of our heuristic then we respect it but we don't increase it

clients/feeder/timeouts.go
```diff
@@ -0,0 +1,161 @@
...
+	timeoutsList := timeouts
+
+	if len(timeouts) == 1 {
+		timeoutsList = timeoutsListFromNumber(timeouts[0])
+	}
+	return Timeouts{
+		curTimeout: 0,
+		timeouts:   timeoutsList,
+	}
+}
```
rodrigo-pino, line 105: `getDynamicTimeouts` should only work when passing a single `time.Duration`. Not an array. It is confusing that you can pass an array and nothing is generated dynamically

clients/feeder/timeouts.go
```diff
@@ -0,0 +1,161 @@
...
+
+	hasTrailingComma := len(values) > 0 && values[len(values)-1] == ""
+	if hasTrailingComma {
+		values = values[:len(values)-1]
+	}
+
+	timeouts := make([]time.Duration, 0, len(values))
+	for i, v := range values {
+		if v == "" {
+			continue
```
rodrigo-pino, line 137: This should never happens as far as I am aware.

clients/feeder/timeouts.go
```diff
@@ -0,0 +1,161 @@
...
+		}
+		d, err := time.ParseDuration(v)
+		if err != nil {
+			return nil, false, fmt.Errorf("parsing timeout at index %d: %v", i, err)
```
rodrigo-pino, line 141: suggestion: "parsing timeout parameter number %d: %v"

Since we are writing for users which sometimes are not programmers. The `index` should be also the `index + 1`. 

All in all
```suggestion
			return nil, false, fmt.Errorf("parsing timeout parameter number %d: %v", i + 1, err)
```

clients/feeder/timeouts.go
```diff
@@ -0,0 +1,161 @@
...
+	if len(timeouts) > 1 {
+		for i := 1; i < len(timeouts); i++ {
+			if timeouts[i] <= timeouts[i-1] {
+				return nil, false, fmt.Errorf("timeouts must be in ascending order, got %v <= %v", timeouts[i], timeouts[i-1])
+			}
+		}
+	}
+
+	if len(timeouts) > timeoutsCount {
+		return nil, false, fmt.Errorf("too many timeouts, max is %d", timeoutsCount)
```
rodrigo-pino, line 158: I think we can improve this error message:

"exceeded max amount of allowed timeout parameters. Set %d but max is %d"

clients/feeder/timeouts.go
```diff
@@ -0,0 +1,161 @@
...
+		timeouts = append(timeouts, d)
+	}
+	if len(timeouts) == 1 && hasTrailingComma {
+		return timeouts, true, nil
+	}
+
+	if len(timeouts) > 1 {
+		for i := 1; i < len(timeouts); i++ {
+			if timeouts[i] <= timeouts[i-1] {
+				return nil, false, fmt.Errorf("timeouts must be in ascending order, got %v <= %v", timeouts[i], timeouts[i-1])
```
rodrigo-pino, line 152: "timeout values must be ..."

The word "timeout" itself means something to you and me now, but it probably means something completely different to anyone else and might not make complete sense

clients/feeder/timeouts.go
```diff
@@ -0,0 +1,161 @@
...
+		if err != nil {
+			return nil, false, fmt.Errorf("parsing timeout at index %d: %v", i, err)
+		}
+		timeouts = append(timeouts, d)
+	}
+	if len(timeouts) == 1 && hasTrailingComma {
+		return timeouts, true, nil
+	}
+
+	if len(timeouts) > 1 {
```
rodrigo-pino, line 149: Unnecessary `if` condition, please remove

cmd/juno/juno.go
```diff
@@ -122,7 +122,7 @@ const (
 	defaultCNL2ChainID              = ""
 	defaultCNCoreContractAddressStr = ""
 	defaultCallMaxSteps             = 4_000_000
-	defaultGwTimeout                = 5 * time.Second
+	defaultGwTimeout                = "5s,"
```
rodrigo-pino, line 125: So we are not dynamic by default 🤔 . I was thinking we should. Let's be dynamic for a while

cmd/juno/juno.go
```diff
@@ -174,8 +174,11 @@ const (
 	dbCacheSizeUsage     = "Determines the amount of memory (in megabytes) allocated for caching data in the database."
 	dbMaxHandlesUsage    = "A soft limit on the number of open files that can be used by the DB"
 	gwAPIKeyUsage        = "API key for gateway endpoints to avoid throttling" //nolint: gosec
-	gwTimeoutUsage       = "Timeout for requests made to the gateway"          //nolint: gosec
-	callMaxStepsUsage    = "Maximum number of steps to be executed in starknet_call requests. " +
+	gwTimeoutsUsage      = "Timeouts for requests made to the gateway. Can be specified in three ways:\n" +
+		"- Single value (e.g. '5s'): After each failure, the timeout will increase exponentially\n" +
```
rodrigo-pino, line 178: It won't increase exponentially 😨 , it will increased dynamically, based on a heuristic but not exponentially.

Nonetheless, great explanation!

clients/feeder/feeder.go
```diff
@@ -73,8 +74,12 @@ func (c *Client) WithUserAgent(ua string) *Client {
...
 }
 
-func (c *Client) WithTimeout(t time.Duration) *Client {
-	c.client.Timeout = t
+func (c *Client) WithTimeouts(timeouts []time.Duration, fixed bool) *Client {
+	if fixed {
+		c.timeouts = getFixedTimeouts(timeouts)
+	} else {
+		c.timeouts = getDynamicTimeouts(timeouts)
+	}
```
rodrigo-pino, line 82: In relation with a previous comment, regarding `getDynamicTimeout` based on a single duration.

```suggestion
func (c *Client) WithTimeouts(timeouts []time.Duration, fixed bool) *Client {
	if len(timeouts) > 1 || fixed {
		c.timeouts = getFixedTimeouts(timeouts)
	} else {
		c.timeouts = getDynamicTimeouts(timeouts[0])
	}
```

rodrigo-pino (approved):
Great job!

docs/docs/_config-options.md
```diff
@@ -20,11 +20,13 @@
...
 | `gw-api-key` |  | API key for gateway endpoints to avoid throttling |
-| `gw-timeout` | `5` | Timeout for requests made to the gateway |
+| `gw-timeouts` | `5s` | Timeouts for requests made to the gateway. Can be specified in three ways:\n- Single value (e.g. '5s'): After each failure, the timeout will increase dynamically \n- Comma-separated list (e.g. '5s,10s,20s'): Each value will be used in sequence after failures. - Single value with trailing comma (e.g. '5s,'): Uses a fixed timeout without dynamic adjustment |
 | `http` | `false` | Enables the HTTP RPC server on the default port and interface |
 | `http-host` | `localhost` | The interface on which the HTTP RPC server will listen for requests |
 | `http-port` | `6060` | The port on which the HTTP server will listen for requests |
-| `log-level` | `info` | Options: trace, debug, info, warn, error |
+| `log-host` | `localhost` | The interface on which the log level HTTP server will listen for requests |
+| `log-level` |  | Options: trace, debug, info, warn, error |
+| `log-port` | `0` | The port on which the log level HTTP server will listen for requests |
```
rodrigo-pino, line 29: I think there were some changes here that weren't intentional

rodrigo-pino (approved):

clients/feeder/timeouts.go
```diff
@@ -0,0 +1,151 @@
...
+		timeouts:   timeouts,
+	}
+}
+
+func getDefaultFixedTimeouts() Timeouts {
+	timeouts, _, _ := ParseTimeouts(DefaultTimeouts)
+	return getFixedTimeouts(timeouts)
+}
+
+func ParseTimeouts(value string) ([]time.Duration, bool, error) {
```
rianhughes, line 114: Can we add some comments / doc string indicating what the return types are (mainly the bool tbh)?

MaksymMalicki: added

docs/docs/_config-options.md
MaksymMalicki: removed

clients/feeder/timeouts.go
```diff
@@ -111,6 +111,10 @@ func getDefaultFixedTimeouts() Timeouts {
 	return getFixedTimeouts(timeouts)
 }
 
+// ParseTimeouts parses a comma-separated string of duration values into a slice of time.Duration.
+// Returns:
+// - []time.Duration: parsed timeout values
+// - bool: true if single timeout value with trailing comma, indicating fixed/dynamic timeouts should be used
```
rodrigo-pino, line 117: You don't have to specify the return types in the comments they are already part of the function definition. Also we are returning an error so that should be explained as well

```suggestion
// ParseTimeouts parses a comma-separated string of duration values into a slice of time.Duration.
// Returns:
// - the parsed timeout values
// - if a fixed or dynamic timeouts should be used
// - an error in case the string cannot be parsed
```

rianhughes (approved):
LGTM

Issue #2695, opened 2025-03-28
starknet_estimateFee intermittently fails on pending block with execution error (u64_sub Overflow)


wojciechos:
We're seeing an inconsistent behavior when calling starknet_estimateFee. The same request at first returned a proper fee estimation, but later fails with an execution error.

The request looks like this on mainnet:
```
{
  "method": "starknet_estimateFee",
  "params": {
    "request": [
      {
        "type": "INVOKE",
        "sender_address": "0x39ce856383d6fd8cc26834fc58cf4380f2d6fbcb3d3240f6c893d348375a2e4",
        "calldata": [
          "0x2",
          "0x4718f5a0fc34cc1af16a1cdee98ffb20c31f5cd61d6ab07201858f4287c938d",
          "0x219209e083275171774dab1df80982e9df2096516f06319c5c6d71ae0a8480c",
          "0x3",
          "0x7fb5bcb8525954a60fde4e8fb8220477696ce7117ef264775a1770e23571929",
          "0x2e6da682da760af80",
          "0x0",
          "0x7fb5bcb8525954a60fde4e8fb8220477696ce7117ef264775a1770e23571929",
          "0xc73f681176fc7b3f9693986fd7b14581e8d540519e27400e88b8713932be01",
          "0x3",
          "0x2e6da682da760af80",
          "0x0",
          "0x39ce856383d6fd8cc26834fc58cf4380f2d6fbcb3d3240f6c893d348375a2e4"
        ],
        "version": "0x100000000000000000000000000000003",
        "signature": ["0x1", "0x0", "0x0"],
        "nonce": "0x3",
        "resource_bounds": {
          "l2_gas": { "max_amount": "0x0", "max_price_per_unit": "0x0" },
          "l1_gas": { "max_amount": "0x0", "max_price_per_unit": "0x0" }
        },
        "tip": "0x0",
        "paymaster_data": [],
        "nonce_data_availability_mode": "L1",
        "fee_data_availability_mode": "L1",
        "account_deployment_data": []
      }
    ],
    "block_id": "pending",
    "simulation_flags": []
  },
  "id": 0,
  "jsonrpc": "2.0"
}
```

At first node returned a proper estimation, but later returned this:
```
{
    "jsonrpc": "2.0",
    "error": {
        "code": 41,
        "message": "Transaction execution error",
        "data": {
            "transaction_index": 0,
[...]

PR #2696, opened 2025-03-28
Bug fix: set the pending header correctly 

rianhughes:
closes https://github.com/NethermindEth/juno/issues/2695

Thanks to @infrmtcs for pointing out that the pending block number was not being set correctly. This corrected the estimate fee error (I suspect the fee estimate transaction does some arithmetic on the block number, and having it set to zero resulted in an error). However, we also need to set the rest of the gas price data in the pending header in order to get correct fee estimates.

### DIFF
--- a/sync/sync.go
+++ b/sync/sync.go
@@ -566,11 +566,6 @@ func (s *Synchronizer) fetchAndStorePending(ctx context.Context) error {
 	}
 
 	pendingBlock.Number = head.Number + 1
-	pendingBlock.L1DAMode = head.L1DAMode
-	pendingBlock.L1GasPriceSTRK = head.L1GasPriceSTRK
-	pendingBlock.L1GasPriceETH = head.L1GasPriceETH
-	pendingBlock.L1DataGasPrice = head.L1DataGasPrice
-	pendingBlock.L2GasPrice = head.L2GasPrice
 	newClasses, err := s.fetchUnknownClasses(ctx, pendingStateUpdate)
 	if err != nil {
 		return err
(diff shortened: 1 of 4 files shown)


vm/rust/src/lib.rs
```diff
@@ -759,11 +759,14 @@ fn build_block_context(
 
 #[allow(static_mut_refs)]
 fn get_versioned_constants(version: *const c_char) -> VersionedConstants {
-    let starknet_version = unsafe { CStr::from_ptr(version) }.to_str()
+    let starknet_version = unsafe { CStr::from_ptr(version) }
```
infrmtcs, line 762: Just want to double check, is the logic in this file unchanged?

sync/sync.go
```diff
@@ -565,11 +565,16 @@ func (s *Synchronizer) fetchAndStorePending(ctx context.Context) error {
 		return err
 	}
 
+	pendingBlock.Number = head.Number + 1
+	pendingBlock.L1DAMode = head.L1DAMode
+	pendingBlock.L1GasPriceSTRK = head.L1GasPriceSTRK
+	pendingBlock.L1GasPriceETH = head.L1GasPriceETH
+	pendingBlock.L1DataGasPrice = head.L1DataGasPrice
+	pendingBlock.L2GasPrice = head.L2GasPrice
```
infrmtcs, line 573: Would you please double check this part? I think feeder still return data for these fields for pending block, so we should keep them intact.

vm/rust/src/lib.rs
rianhughes: Yup, `make format` just seemed to change it

infrmtcs (approved):

sync/sync.go
rianhughes: Yup we can remove it

PR #2701, opened 2025-03-29
refactor: optimize code and use modern Go idioms

AnkushinDaniil:


rpc/v6/simulation.go
```diff
@@ -36,7 +36,7 @@ func (s *SimulationFlag) UnmarshalJSON(bytes []byte) (err error) {
 
 type SimulatedTransaction struct {
 	TransactionTrace *TransactionTrace `json:"transaction_trace,omitempty"`
-	FeeEstimation    FeeEstimate       `json:"fee_estimation,omitempty"`
+	FeeEstimation    FeeEstimate       `json:"fee_estimation"`
```
AnkushinDaniil, line 39: Omitempty has no effect on nested struct fields

rodrigo-pino (approved):

rpc/v6/simulation.go
infrmtcs: Is always showing this field the expected behaviour? I'm understanding that Go 1.24 provides a new `omitzero` option for nested struct fields.

infrmtcs: For 2, the docs said it use zero value of the type if `IsZero()` is not available. But in general, I think both approaches are fine to me.

hudem1: Mmh I feel like the specs are not 100% clear because indeed there doesn't seem to be a clear "required" field after the properties (`transaction_trace` & `fee_estimation`) for the `starknet_simulateTransactions` [rpc call](https://github.com/starkware-libs/starknet-specs/blob/v0.6.0/api/starknet_trace_api_openrpc.json#L76), but I still think the `fee_estimation` is required.

Because if we interpret it as not required, then `transaction_trace` too wouldn't be required, there is no additional indication stating that `transaction_trace` is more required than `fee_estimation`.

But then if `transaction_trace` is not required either, then it would really not make any sense i think because the user wouldn't have any result from their rpc call.

rianhughes: @AnkushinDaniil would you mind updating the PR to either make `FeeEstimate` a pointer and use the `omitempty` tag, or use the `omitzero` tag?

@hudem1 I agree that the specs could be made more clear, maybe we can ask the SW guys to add `required` or `not_required` to each method/type etc?  Currently, I don't see any strong reason to change the behaviour of how we handle `starknet_simulateTransactions` right now, but if the SW guys indicate that we should always return `FeeEstimate` then I agree we should remove `omitempty`.

AnkushinDaniil: [Add omitzero tag](https://github.com/NethermindEth/juno/pull/2701/commits/c9949076cbf6e8eb610df4af519c171f9fb2e67e)

infrmtcs (approved):

PR #2702, opened 2025-03-29
build: update code generation tools and generated files

AnkushinDaniil:
https://tip.golang.org/doc/go1.24#tools


Makefile
```diff
@@ -64,7 +64,6 @@ generate-buf: ## Generate protobuf files
 
 generate: ## Generate mocks and code
 	mkdir -p mocks
-	generate-buf
```
infrmtcs, line 67: Should we also remove `install-mockgen`, as it's managed by tool now?

PR #2703, opened 2025-03-30
fix(sn2core): reverse handling of missing L1DataGas resource

### LINKED ISSUE #2699: Incorrect gas_consumed returned when resource_bounds.l2_gas is set
When performing a fee estimation with any value in resource_bounds.l2_gas set to something other than 0x0, the gas_consumed returned is always 0x0.

Example on mainnet:
```
{
    "method": "starknet_estimateFee",
    "params": {
        "request": [
            {
                "type": "INVOKE",
                "sender_address": "0x39ce856383d6fd8cc26834fc58cf4380f2d6fbcb3d3240f6c893d348375a2e4",
                "calldata": [
                    "0x2",
                    "0x4718f5a0fc34cc1af16a1cdee98ffb20c31f5cd61d6ab07201858f4287c938d",
                    "0x219209e083275171774dab1df80982e9df2096516f06319c5c6d71ae0a8480c",
                    "0x3",
                    "0x7fb5bcb8525954a60fde4e8fb8220477696ce7117ef264775a1770e23571929",
                    "0x2e6da682da760af80",
                    "0x0",
                    "0x7fb5bcb8525954a60fde4e8fb8220477696ce7117ef264775a1770e23571929",
                    "0xc73f681176fc7b3f9693986fd7b14581e8d540519e27400e88b8713932be01",
                    "0x3",
                    "0x2e6da682da760af80",
                    "0x0",
                    "0x39ce856383d6fd8cc26834fc58cf4380f2d6fbcb3d3240f6c893d348375a2e4"
                ],
                "version": "0x100000000000000000000000000000003",
                "signature": [
                    "0x1",
                    "0x0",
                    "0x0"
                ],
                "nonce": "0x3",
                "resource_bounds": {
                    "l2_gas": {
                        "max_amount": "0x1",  // when this set to 0x0 gas consumed is returned
                        "max_price_per_unit": "0x0"
                    },
                    "l1_gas": {
                        "max_amount": "0x0",
                        "max_price_per_unit": "0x0"
                    }
                },
                "tip": "0x0",
                "paymaster_data": [],
                "nonce_data_availability_mode": "L1",
                "fee_data_availability_mode": "L1",
                "account_deployment_data": []
            }
        ],
        "block_id": "latest",
        "simulation_flags": ["SKIP_VALIDATE"]
    },
    "id": 0,
    "jsonrpc": "2.0"
}
```

Example result:
```
{
    "jsonrpc": "2.0",
    "result": [
        {
            "gas_consumed": "0x0",
            "gas_price": "0x116ece448574",
            "data_gas_consumed": "0x600",
            "data_gas_price": "0x3cd20dc",
            "overall_fee": "0xec018d9a13b800",
            "unit": "FRI"
        }
    ],
    "id": 0
}
```

AnkushinDaniil:
Fixes #2699

Instead of adding `L1DataGas` with zero values when it's missing but `L2Gas` is non-zero, now we set `L2Gas` to zero when `L1DataGas` is missing. This matches the behavior expected by the `blockifier`, as detailed in the [referenced Rust implementation](https://github.com/starkware-libs/sequencer/blob/6c14c03e2d9d6f72ae05c98442dbaf17c243433e/crates/starknet_api/src/transaction/fields.rs#L426-L458).


adapters/sn2core/sn2core.go
```diff
@@ -194,25 +194,34 @@ func adaptResourceBounds(rb *map[starknet.Resource]starknet.ResourceBounds) map[
...
 	// Blockifier will throw an error if L1DataGas is absent and L2Gas is non-zero.
-	// To avoid that, if L2Gas is non-zero, we set the L1DataGas resource to zero.
-	if l2Gas, ok := coreBounds[core.ResourceL2Gas]; ok && !l2Gas.IsZero() {
-		if _, ok := coreBounds[core.ResourceL1DataGas]; !ok {
-			coreBounds[core.ResourceL1DataGas] = core.ResourceBounds{
-				MaxAmount:       0,
-				MaxPricePerUnit: &felt.Zero,
+	// To avoid that, if L1DataGas is not present, we set L2Gas to zero.
+
+	if l1Gas, ok := (*rb)[starknet.ResourceL1Gas]; ok {
```
infrmtcs, line 203: Is there a case where the original `map[starknet.Resource]starknet.ResourceBounds` has incorrect format? If not, can we just iterate over the source map and set the target map?
If yes, shouldn't the validation be done by the caller?

infrmtcs, line 203: Not because of this PR, but I'm wondering why we need a pointer for `map[starknet.Resource]starknet.ResourceBounds`

AnkushinDaniil: We can iterate and set the target map, but we still need to check if we have `l1DataGas` and after that, depending on this, we set `l2Gas`. So it doesn't make sense to set 2 out of 3 values twice.
The thing is, the validation can't change the `l2Gas' value.
I'm not sure I understand your question correctly, if so, please correct me.

AnkushinDaniil: I guess because `ResourceBounds` is optional in the `Transaction` structure. I tried to remove the pointer and found out that it requires a lot of changes. It should definitely be a separate PR.

infrmtcs: Yeah, I definitely agree, this can be in another PR

infrmtcs: I'm thinking about the source of `map[starknet.Resource]starknet.ResourceBounds`, there should be 2 cases:
- If it's from an existing transaction, the format is supposed to be correct, so no conversion needs to be done here.
- If it's from user's input, e.g, estimate fee, simulate transaction, send a new transaction, etc., the check we're doing here should be done on user input on RPC level, instead of in this adapter.

infrmtcs: Regarding v6/v7 and v8, I assume that v6 and v7 accept both types of transactions, while v8 rejects the "2 bounds" type and only accepts "3 bounds" type.

rianhughes:
Passes with this sepolia request

```
{
    "method": "starknet_estimateFee",
    "params": {
        "request": [
            {
                "type": "INVOKE",
                "sender_address": "0x7c1cbbafca15fec62f943de72793ddd40c0ae92884354e301cdd610f7c90106",
                "calldata": [
                    "0x1",
                    "0x45dd5d8fa19ac764bed57de30f7e07fd194b7bcd4d4759d6779c8c288914b2b",
                    "0x0c844fd57777b0cd7e75c8ea68deec0adf964a6308da7a58de32364b7131cc8",
                    "0x13",
                    "0x4178ba309120c5fa0364516fd1a514055a6b314396f46c30e1a0417c21586",
                    "0x3aaca05",
                    "0xdb4a286c316f4d009066712eaea2e4d9a8c6d2ab14b740a02339c63b5f9004",
                    "0x67e83fe0",
                    "0x102010300000000000000000000000000000000000000000000000000000000",
                    "0x4",
                    "0xe29c69",
                    "0xe2ac9a",
                    "0xe2b1b9",
                    "0xe2b7f9",
                    "0x149fb4831285671001d",
                    "0x0b224de87709",
                    "0x2",
                    "0x15a9c12195d170019bc3e798268052963c7723fd323c2633c8d08b8a3547ff1",
                    "0x385e0f4b69496e49a46125bbcc10483e402c89fde61b1b18e918f04e857f912",
                    "0x319f9f7ce6e6c72e179557e79c57a542508867fa05b505cc32d2380ac72e711",
                    "0x5c33957bcabbb51932c72d5ce2c34786674e2123f378d2cb1c4b94ba9b4fb36",
                    "0x324ed226f838a9a516f75ae6a9f4d4071ebd57d3ef5c3329cd0dc1b92dc28af",
                    "0x7c3d4567dab790adf79c2b899588023fa39cceef8f17beb19864cc73b40a83f"
                ],
                "version": "0x3",
                "signature": [
                    "0x13881822696b7d446dc2d367f9d014459d33e98be5e0711032974394fab5c82",
                    "0x4e589991b210368ff68335e9fdaf9719c8609337a83f625966afe4c286b3a3a"
                ],
                "nonce": "0x8e35",
                "resource_bounds": {
                    "l2_gas": {
                        "max_amount": "0x1",
                        "max_price_per_unit": "0x0"
                    },
                    "l1_gas": {
                        "max_amount": "0x34",
                        "max_price_per_unit": "0x10b374dcb28d"
                    }
                },
                "tip": "0x0",
                "paymaster_data": [],
                "nonce_data_availability_mode": "L1",
                "fee_data_availability_mode": "L1",
                "account_deployment_data": []
            }
        ],
        "block_id": {"block_number":634544},
        "simulation_flags": ["SKIP_VALIDATE"]
    },
    "id": 0,
    "jsonrpc": "2.0"
}
```

rianhughes (approved):

rpc/v7/transaction.go
```diff
@@ -306,7 +306,7 @@ func adaptBroadcastedTransaction(broadcastedTxn *BroadcastedTransaction,
 	if err != nil {
 		return nil, nil, nil, err
 	}
-
+	overrideResourceBounds(txn)
```
infrmtcs, line 309: Should we do it on `BroadcastedTransaction` instead? We can access directly `broadcastedTxn.Transaction.ResourceBounds` without having to type switch, and I think validation should be done as early as possible, so we don't have original `broadcastedTxn.Transaction` and `feederTxn` in wrong format.

infrmtcs (approved):
The rest LGTM, thanks!

rpc/v7/transaction.go
rianhughes: Yup, good point. This is much nicer

wojciechos (approved):

PR #2720, opened 2025-03-31
feat(node): update feeder gateway timeouts by HTTP

MaksymMalicki:


clients/feeder/feeder.go
```diff
@@ -284,12 +284,25 @@ func (c *Client) get(ctx context.Context, queryURL string) (io.ReadCloser, error
...
+			if currentTimeout >= fastGrowThreshold {
+				c.log.Warnw("Failed query to feeder, retrying...",
+					"req", req.URL.String(),
+					"retryAfter", wait.String(),
+					"err", err,
+					"newHTTPTimeout", c.timeouts.GetCurrentTimeout().String(),
+				)
+				c.log.Warnw("Timeouts can be updated via HTTP PUT request",
+					"timeout", currentTimeout.String(),
+					"hint", `Set --http-update-port and --http-update-host flags and make a PUT request to "/feeder/timeouts" with the bigger timeout`)
```
rodrigo-pino, line 297: Changer `bigger` to `specified`. And change `timeout` to `timeouts`. I am assuming they can also set a list there.

rodrigo-pino (approved):
Good!

clients/feeder/feeder.go
MaksymMalicki: yes

clients/feeder/timeouts.go
```diff
@@ -154,3 +155,27 @@ func ParseTimeouts(value string) ([]time.Duration, bool, error) {
...
+			return
+		}
+
+		newTimeouts, fixed, err := ParseTimeouts(timeoutsStr)
+		if err != nil {
+			http.Error(w, err.Error(), http.StatusBadRequest)
+			return
+		}
+
+		client.WithTimeouts(newTimeouts, fixed)
```
infrmtcs, line 176: Is there any case we have race condition here?

infrmtcs: I'm thinking about the case where this happens in the middle of a query loop. In that case there can be a scenario where old `curTimeout` is used while new `timeouts` slice is replaced, then out-of-range error can cause panic.

MaksymMalicki: Not entirely sure about it. `curTimeout` is always set to 0 when using `WithTimeouts`. You might be right. Added mutex just in case 👍

MaksymMalicki: used `atomic.Value` instead of mutex in the end

PR #2729, opened 2025-04-02
ci: add automated draft release workflow with binaries and docker build

wojciechos:
- Adds workflow triggered on `v*` tag pushes  
- Builds and uploads binaries  
- Builds and pushes Docker images 
- Creates a draft GitHub release with auto-generated notes and binaries attached
- Add workflow to re-tag to latest, triggered manually or on pubishe release event

sample result: https://github.com/NethermindEth/juno/releases/tag/untagged-ddf7896ed66ac3a2d0b8

### DIFF
--- a/.github/workflows/docker-image-tag-latest.yml
+++ b/.github/workflows/docker-image-tag-latest.yml
@@ -1,0 +1,37 @@
+name: 'Tag Official Docker Image as Latest'
+
+on:
+  workflow_dispatch: {}
+  release:
+    types: [published]
+
+permissions:
+  contents: read
+  packages: write
+
+jobs:
+  tag_official_image_as_latest:
+    if: github.repository_owner == 'NethermindEth'
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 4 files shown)


.github/workflows/docker-image-tag-latest.yml
```diff
@@ -0,0 +1,48 @@
+name: 'Tag Official Docker Image as Latest'
+
+on:
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: 'Docker image tag to use'
+        required: true
```
derrix060, line 8: not sure if this should be required, since it won't be used when `${{ github.event_name }}" == "release"`

derrix060 (approved):

.github/workflows/docker-image-tag-latest.yml
wojciechos: yeah, actually, i drop the input tag as we should always extract the tag from `${GITHUB_REF#refs/tags/}`

PR #2730, opened 2025-04-02
refactor: Split rules into condition and action

infrmtcs:
Depends on #2728 and #2735

### DIFF
--- a/consensus/tendermint/prevote.go
+++ b/consensus/tendermint/prevote.go
@@ -21,14 +21,23 @@ func (t *Tendermint[V, H, A]) handlePrevote(p Prevote[H, A]) {
 
 	t.messages.addPrevote(p)
 
-	_, prevotesForHR, _ := t.messages.allMessages(p.H, p.R)
+	cachedProposal := t.findMatchingProposal(t.state.r, p.ID)
 
-	t.line28WhenPrevoteIsReceived(p)
+	if cachedProposal != nil && t.uponProposalAndPolkaPrevious(cachedProposal, p.R) {
+		t.doProposalAndPolkaPrevious(cachedProposal)
+	}
 
 	if p.R == t.state.r {
-		t.line34(p, prevotesForHR)
-		t.line44(p, prevotesForHR)
+		if t.uponPolkaAny(p) {
+			t.doPolkaAny(p)
+		}
 
-		t.line36WhenPrevoteIsReceived(p)
+		if t.uponPolkaNil(p) {
+			t.doPolkaNil()
+		}
+	}
+
+	if cachedProposal != nil && t.uponProposalAndPolkaCurrent(cachedProposal) {
+		t.doProposalAndPolkaCurrent(cachedProposal)
 	}
 }
--- a/consensus/tendermint/rule_polka_any.go
+++ b/consensus/tendermint/rule_polka_any.go
@@ -11,11 +11,18 @@ Check the upon condition on line 34:
 	34: upon 2f + 1 {PREVOTE, h_p, round_p, ∗} while step_p = prevote for the first time do
 	35: schedule OnTimeoutPrevote(h_p, round_p) to be executed after timeoutPrevote(round_p)
 */
-func (t *Tendermint[V, H, A]) line34(p Prevote[H, A], prevotesForHR map[A][]Prevote[H, A]) {
-	vals := slices.Collect(maps.Keys(prevotesForHR))
-	if !t.state.timeoutPrevoteScheduled && t.state.s == prevote &&
-		t.validatorSetVotingPower(vals) >= q(t.validators.TotalVotingPower(p.H)) {
-		t.scheduleTimeout(t.timeoutPrevote(p.R), prevote, p.H, p.R)
-		t.state.timeoutPrevoteScheduled = true
-	}
+func (t *Tendermint[V, H, A]) uponPolkaAny(p Prevote[H, A]) bool {
+	prevotes := t.messages.prevotes[t.state.h][t.state.r]
+	vals := slices.Collect(maps.Keys(prevotes))
+
+	isFirstTime := !t.state.timeoutPrevoteScheduled
+
+	hasQuorum := t.validatorSetVotingPower(vals) >= q(t.validators.TotalVotingPower(p.H))
+
+	return p.R == t.state.r && t.state.s == prevote && hasQuorum && isFirstTime
+}
+
+func (t *Tendermint[V, H, A]) doPolkaAny(p Prevote[H, A]) {
+	t.scheduleTimeout(t.timeoutPrevote(p.R), prevote, p.H, p.R)
+	t.state.timeoutPrevoteScheduled = true
 }
--- a/consensus/tendermint/rule_polka_any.go
+++ b/consensus/tendermint/rule_polka_any.go
@@ -19,7 +19,10 @@ func (t *Tendermint[V, H, A]) uponPolkaAny(p Prevote[H, A]) bool {
 
 	hasQuorum := t.validatorSetVotingPower(vals) >= q(t.validators.TotalVotingPower(p.H))
 
-	return p.R == t.state.r && t.state.s == prevote && hasQuorum && isFirstTime
+	return p.R == t.state.r &&
+		t.state.s == prevote &&
+		hasQuorum &&
+		isFirstTime
 }
 
 func (t *Tendermint[V, H, A]) doPolkaAny(p Prevote[H, A]) {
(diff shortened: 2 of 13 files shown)


consensus/tendermint/prevote.go
```diff
@@ -21,14 +21,23 @@ func (t *Tendermint[V, H, A]) handlePrevote(p Prevote[H, A]) {
...
+	if cachedProposal != nil && t.uponProposalAndPolkaPrevious(cachedProposal, p.R) {
+		t.doProposalAndPolkaPrevious(cachedProposal)
+	}
 
 	if p.R == t.state.r {
-		t.line34(p, prevotesForHR)
-		t.line44(p, prevotesForHR)
+		if t.uponPolkaAny(p) {
+			t.doPolkaAny(p)
+		}
```
rodrigo-pino, line 33: Way better with this approach 👌

rodrigo-pino (approved):

consensus/tendermint/rule_skip_round.go
```diff
@@ -29,9 +30,9 @@ func (t *Tendermint[V, H, A]) line55(futureR round) {
...
 	}
 
-	t.futureMessagesMu.Unlock()
+	return t.validatorSetVotingPower(slices.Collect(maps.Keys(vals))) > f(t.validators.TotalVotingPower(t.state.h))
+}
 
-	if t.validatorSetVotingPower(slices.Collect(maps.Keys(vals))) > f(t.validators.TotalVotingPower(t.state.h)) {
-		t.startRound(futureR)
-	}
+func (t *Tendermint[V, H, A]) doSkipRound(futureR round) {
```
rianhughes, line 36: Why split this into a seperate function?

consensus/tendermint/rule_skip_round.go
```diff
@@ -5,13 +5,14 @@ import (
 	"slices"
 )
 
-// line55 assumes the caller has acquired a mutex for accessing future messages.
+// skipRound assumes the caller has acquired a mutex for accessing future messages.
```
rianhughes, line 8: nit: typo, uponSkipRound

consensus/tendermint/rule_proposal_and_precommit_value.go
```diff
@@ -48,21 +17,22 @@ validity of the proposal can be skipped.
...
-		if proposal != nil && hasQuorum {
-			t.blockchain.Commit(t.state.h, *proposal.Value, precommits)
+	// h_p never goes backward, so it's safe to assume that decision_p[h_p] is nil
+	return hasQuorum && isValid
+}
 
-			t.messages.deleteHeightMessages(t.state.h)
-			t.state.h++
-			t.startRound(0)
+func (t *Tendermint[V, H, A]) doProposalAndPrecommitValue(cachedProposal *CachedProposal[V, H, A]) {
```
rianhughes, line 30: Why split this into a seperate function? We could inline in `uponProposalAndPrecommitValue`

consensus/tendermint/rule_skip_round.go
infrmtcs: Context for this PR: From my understanding, Tendermint is a set of rules, each rule has format: `upon Condition do Action`.
I think it's much clearer to keep track the rules if the implementation follow the convention, each rule has 2 functions, 1 for condition and 1 for action

consensus/tendermint/rule_proposal_and_precommit_value.go
rianhughes: Also, what do you think about updating the names to be a bit more descriptive, eg `CommitValue` /etc?

consensus/tendermint/rule_skip_round.go
rianhughes: Okay, makes sense to me

consensus/tendermint/rule_skip_round.go
rianhughes: nit: also could we include a (line 55) comment / some reference to the Tendermint refernece?

consensus/tendermint/tendermint.go
```diff
@@ -391,7 +397,9 @@ func handleFutureRoundMessage[H Hash, A Addr, V Hashable[H], M Message[V, H, A]]
 		addMessage(m)
 		t.futureMessagesMu.Unlock()
 
-		t.line55(mR)
+		if t.uponSkipRound(mR) {
+			t.doSkipRound(mR)
+		}
```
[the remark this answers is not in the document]
infrmtcs: Added a short explanation on why we skip round in the rule

consensus/tendermint/rule_polka_any.go
```diff
@@ -11,11 +11,18 @@ Check the upon condition on line 34:
...
-		t.state.timeoutPrevoteScheduled = true
-	}
+func (t *Tendermint[V, H, A]) uponPolkaAny(p Prevote[H, A]) bool {
+	prevotes := t.messages.prevotes[t.state.h][t.state.r]
+	vals := slices.Collect(maps.Keys(prevotes))
+
+	isFirstTime := !t.state.timeoutPrevoteScheduled
+
+	hasQuorum := t.validatorSetVotingPower(vals) >= q(t.validators.TotalVotingPower(p.H))
+
```
rianhughes, line 21: Nice

consensus/tendermint/rule_polka_any.go
```diff
@@ -11,11 +11,18 @@ Check the upon condition on line 34:
...
+	return p.R == t.state.r && t.state.s == prevote && hasQuorum && isFirstTime
```
rianhughes, line 22:
```suggestion
	return p.R == t.state.r &&
		t.state.s == prevote &&
		hasQuorum &&
		isFirstTime
```

consensus/tendermint/rule_proposal_and_polka_current.go
```diff
@@ -12,77 +12,21 @@ Check upon condition on line 36:
...
-
-			t.state.validValue = p.Value
-			t.state.validRound = cr
-			t.state.lockedValueAndOrValidValueSet = true
-		}
-	}
+func (t *Tendermint[V, H, A]) uponProposalAndPolkaCurrent(cachedProposal *CachedProposal[V, H, A]) bool {
+	hasQuorum := t.checkQuorumPrevotesGivenProposalVID(t.state.r, *cachedProposal.ID)
+	firstTime := !t.state.lockedValueAndOrValidValueSet
+	return hasQuorum && cachedProposal.Valid && t.state.s >= prevote && firstTime
```
rianhughes, line 19:
```suggestion
	return hasQuorum &&
		cachedProposal.Valid &&
		t.state.s >= prevote &&
		firstTime
```

consensus/tendermint/prevote.go
```diff
@@ -21,14 +21,23 @@ func (t *Tendermint[V, H, A]) handlePrevote(p Prevote[H, A]) {
 
 	t.messages.addPrevote(p)
 
-	_, prevotesForHR, _ := t.messages.allMessages(p.H, p.R)
+	cachedProposal := t.findMatchingProposal(t.state.r, p.ID)
 
-	t.line28WhenPrevoteIsReceived(p)
```
rianhughes, line 26: I'm confused. In the original code we called `line28WhenPrevoteIsReceived(prevote)` which acted on prevote information. Now we call `uponProposalAndPolkaPrevious(cachedProposal, prevote.Round)` which seems to act primarily on the cachedProposal. Are we certain these are the same?

consensus/tendermint/rule_proposal_and_polka_current.go
```diff
@@ -12,77 +12,21 @@ Check upon condition on line 36:
...
-	43: validRound_p ← round_p
-
-Fetching the relevant proposal implies the sender of the proposal was the proposer for that
-height and round. Also, since only the proposals with valid value are added to the message set, the
-validity of the proposal can be skipped.
-
-Calculating quorum of prevotes is more resource intensive than checking other condition on line 36,
-therefore, it is checked in a subsequent if statement.
-*/
-func (t *Tendermint[V, H, A]) line36WhenPrevoteIsReceived(p Prevote[H, A]) {
```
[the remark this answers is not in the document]
rianhughes: Basically, should we not have a `Prevotes` version of `uponProposalAndPolkaPrevious`? Similarly for `uponProposalAndPolkaCurrent`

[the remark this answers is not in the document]
rianhughes: eg, At the surface level it looks like we used to check that the prevote ID was non nil and the proposal was non-nil, but these checks are no longer present

rianhughes (approved):

consensus/tendermint/prevote.go
infrmtcs: Resolved, as explained below

PR #2741, opened 2025-04-07
feat(l1): exponential retry for chainID req

weiihann:
Fixes https://github.com/NethermindEth/juno/issues/1385

### DIFF
--- a/l1/l1.go
+++ b/l1/l1.go
@@ -17,6 +17,10 @@ import (
 	"github.com/ethereum/go-ethereum/event"
 )
 
+const (
+	maxRetryDelay = 60000 // 1 minute in ms
+)
+
 //go:generate mockgen -destination=../mocks/mock_subscriber.go -package=mocks github.com/NethermindEth/juno/l1 Subscriber
 type Subscriber interface {
 	FinalisedHeight(ctx context.Context) (uint64, error)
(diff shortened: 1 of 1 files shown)


l1/l1.go
```diff
@@ -86,19 +90,43 @@ func (c *Client) subscribeToUpdates(ctx context.Context, updateChan chan *contra
...
+				return fmt.Errorf("mismatched L1 and L2 networks: got L1 chain ID %s, expected %s for L2 network %s",
+					actualChainID.String(), expectedChainID.String(), c.network.String())
+			}
+
+			c.log.Debugw("Failed to check chain ID", "error", err)
 
-	// NOTE: for now we return an error. If we want to support users who fork
-	// Starknet to create a "custom" Starknet network, we will need to log a warning instead.
-	return fmt.Errorf("mismatched L1 and L2 networks: L2 network %s; is the L1 node on the correct network?", c.network.String())
+			timer := time.NewTimer(retryDelay)
```
cwkang1998, line 117: If possible would be great to have a test for the backoff mechanism.

cwkang1998 (commented):
Curious if we should also have a maximum retry amount rather than having the node retry forever?

cwkang1998 (approved):

PR #2743, opened 2025-04-07
refactor: Remove unnecessary dependencies on message data for some rules

infrmtcs:
Rationale: Among 8 rules, only `precommit_any` and `skip_round` are not requiring messages to be in current round. We can remove the message from other rule arguments and check the condition against the current height and round instead.

### DIFF
--- a/consensus/tendermint/rule_polka_any.go
+++ b/consensus/tendermint/rule_polka_any.go
@@ -11,21 +11,20 @@ Check the upon condition on line 34:
 	34: upon 2f + 1 {PREVOTE, h_p, round_p, ∗} while step_p = prevote for the first time do
 	35: schedule OnTimeoutPrevote(h_p, round_p) to be executed after timeoutPrevote(round_p)
 */
-func (t *Tendermint[V, H, A]) uponPolkaAny(p Prevote[H, A]) bool {
+func (t *Tendermint[V, H, A]) uponPolkaAny() bool {
 	prevotes := t.messages.prevotes[t.state.height][t.state.round]
 	vals := slices.Collect(maps.Keys(prevotes))
 
 	isFirstTime := !t.state.timeoutPrevoteScheduled
 
-	hasQuorum := t.validatorSetVotingPower(vals) >= q(t.validators.TotalVotingPower(p.Height))
+	hasQuorum := t.validatorSetVotingPower(vals) >= q(t.validators.TotalVotingPower(t.state.height))
 
-	return p.Round == t.state.round &&
-		t.state.step == prevote &&
+	return t.state.step == prevote &&
 		hasQuorum &&
 		isFirstTime
 }
 
-func (t *Tendermint[V, H, A]) doPolkaAny(p Prevote[H, A]) {
-	t.scheduleTimeout(t.timeoutPrevote(p.Round), prevote, p.Height, p.Round)
+func (t *Tendermint[V, H, A]) doPolkaAny() {
+	t.scheduleTimeout(t.timeoutPrevote(t.state.round), prevote, t.state.height, t.state.round)
 	t.state.timeoutPrevoteScheduled = true
 }
--- a/consensus/tendermint/rule_polka_nil.go
+++ b/consensus/tendermint/rule_polka_nil.go
@@ -9,7 +9,7 @@ Check the upon condition on line 44:
 
 Line 36 and 44 for a round are mutually exclusive.
 */
-func (t *Tendermint[V, H, A]) uponPolkaNil(p Prevote[H, A]) bool {
+func (t *Tendermint[V, H, A]) uponPolkaNil() bool {
 	prevotes := t.messages.prevotes[t.state.height][t.state.round]
 
 	var vals []A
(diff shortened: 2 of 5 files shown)


consensus/tendermint/rule_precommit_any.go
```diff
@@ -11,20 +11,18 @@ Check the upon condition on line 47:
...
 	48: schedule OnTimeoutPrecommit(h_p , round_p) to be executed after timeoutPrecommit(round_p)
 */
-func (t *Tendermint[V, H, A]) uponPrecommitAny(p Precommit[H, A]) bool {
+func (t *Tendermint[V, H, A]) uponPrecommitAny() bool {
 	precommits := t.messages.precommits[t.state.height][t.state.round]
 	vals := slices.Collect(maps.Keys(precommits))
 
 	isFirstTime := !t.state.timeoutPrecommitScheduled
 
-	hasQuorum := t.validatorSetVotingPower(vals) >= q(t.validators.TotalVotingPower(p.Height))
```
rianhughes, line 20: Why is the precommit we receive from a peer guaranteed to have the same height as the state as we see it? Similar for prevotes

consensus/tendermint/rule_polka_any.go
```diff
@@ -11,21 +11,20 @@ Check the upon condition on line 34:
...
 	35: schedule OnTimeoutPrevote(h_p, round_p) to be executed after timeoutPrevote(round_p)
 */
-func (t *Tendermint[V, H, A]) uponPolkaAny(p Prevote[H, A]) bool {
+func (t *Tendermint[V, H, A]) uponPolkaAny() bool {
 	prevotes := t.messages.prevotes[t.state.height][t.state.round]
 	vals := slices.Collect(maps.Keys(prevotes))
 
 	isFirstTime := !t.state.timeoutPrevoteScheduled
 
-	hasQuorum := t.validatorSetVotingPower(vals) >= q(t.validators.TotalVotingPower(p.Height))
```
rianhughes, line 20: Why is the height guaranteed to be the same in the prevote and the step? I see we check the rounds are equal before calling this function

consensus/tendermint/rule_polka_nil.go
```diff
@@ -22,7 +22,7 @@ func (t *Tendermint[V, H, A]) uponPolkaNil(p Prevote[H, A]) bool {
 	}
 
 	// TODO: refactor this
-	hasQuorum := t.validatorSetVotingPower(vals) >= q(t.validators.TotalVotingPower(p.Height))
```
rianhughes, line 25: similar question here about whye the heights are the same

consensus/tendermint/rule_precommit_any.go
```diff
@@ -11,20 +11,18 @@ Check the upon condition on line 47:
 	47: upon 2f + 1 {PRECOMMIT, h_p, round_p, ∗} for the first time do
 	48: schedule OnTimeoutPrecommit(h_p , round_p) to be executed after timeoutPrecommit(round_p)
 */
-func (t *Tendermint[V, H, A]) uponPrecommitAny(p Precommit[H, A]) bool {
```
rianhughes, line 14: and why the heights and rounds are the same here (we don't perform any checks directly before calling this as far as I can see)

consensus/tendermint/rule_polka_any.go
infrmtcs: `h_p` and `round_p` denote the current height and round in the state of the state machine, so I believe if we strictly follow the algorithm, we should only check against the current height and round.
In the worst case when an event in different height/round is received, nothing actually happens.

infrmtcs: Although I think a possible argument against this change is we have to pay the cost to check in that cast, I think that the cost to check for quorum should be instead optimized so that it's not a big cost, and even if we want to optimize it, it can be done on the higher layer.

rianhughes: I think this changes the behaviour of the Tendermint algorithm, and I'm not sure if we gain much from deviating from it tbh (eg we don't save any function calls as far as I can see). For example, previously we would check if the incoming prevote(h,r) has reached quorum, and (potentially), schedule a timeout for the Prevote-round. But now we check if our states round has reached quorum and (potentially) schedule a timeout for that. At the surface level these seem like different behaviours. 

Let me know if I'm missing something, but currently I don't think we should push ahead with this change.

infrmtcs: To clarify, what I meant regarding optimization is that I "reverted" the optimization in the current implementation. 
Old:
- If message is from other round: skip the check
- If message is from current round: do the check
New:
- Regardless whether new message is from current round, or no new message at all: do the check

rianhughes (approved):

consensus/tendermint/rule_polka_any.go
rianhughes: Just posting an update here, it turns out we constrain the messages height and round to the current states height and round by the time we get to these functions. This PR primarily simplifies the function signatures

rodrigo-pino (approved):

PR #2745, opened 2025-04-07
refactor: Accept 1 message from a sender per height, round and type

infrmtcs:
Each sender is expected to send a message for each height, round and step.

### DIFF
--- a/consensus/tendermint/tendermint.go
+++ b/consensus/tendermint/tendermint.go
@@ -263,40 +263,34 @@ func (t *Tendermint[V, H, A]) startRound(r round) {
 func (t *Tendermint[V, H, A]) processFutureMessages(h height, r round) {
 	proposals, prevotes, precommits := t.futureMessages.allMessages(h, r)
 	if len(proposals) > 0 {
-		for _, addrProposals := range proposals {
-			for _, proposal := range addrProposals {
-				select {
-				case <-t.quit:
-					return
-				default:
-					t.handleProposal(proposal)
-				}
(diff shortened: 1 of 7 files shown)


consensus/tendermint/tendermint.go
```diff
@@ -438,36 +430,24 @@ func (t *Tendermint[V, H, A]) checkQuorumPrevotesGivenProposalVID(r round, vID H
...
-			}
+	for addr, p := range prevotes {
+		if *p.ID == vID {
+			vals = append(vals, addr)
 		}
 	}
 	return t.validatorSetVotingPower(vals) >= q(t.validators.TotalVotingPower(t.state.height))
 }
 
 // TODO: Improve performance. Current complexity is O(n).
```
MaksymMalicki, line 441: nit: is this comment still valid?

MaksymMalicki (approved):

rianhughes (approved):
Just adding some additional context. 

Currently this PR means that we will only store the first msg we receive from a peer (for a given `h,r,a`). This means we forget / ignore any other msgs we may receive from them, which means we can't handle things like slashing. The plan is that we will have another component that receives all such msgs and handles slashing etc, and the `Tendermint.state` should only receive a single valid msg (which has a signature check filter)

consensus/tendermint/tendermint.go
infrmtcs: You're right, I'll remove it in the next PR

PR #2754, opened 2025-04-09
feat(node): hash scheme for TrieDB

MaksymMalicki:
#2810 

This PR introduces a hash scheme for the TrieDB, in the package `core/trie2/triedb/hashdb`. In the hash scheme, nodes are differentiated not only by their `Path` in the trie, but also their unique hash. Nodes are encoded in the following manner:  

ClassTrie :  
[1 byte prefix][1 byte node-type][8 byte from path][32 byte hash]. 

ContractTrie :  
[1 byte prefix][1 byte node-type][8 byte from path][32 byte hash]. 

StorageTrie of a Contract :  
[1 byte prefix][32 bytes owner][1 byte node-type][8 byte from path][32 byte hash]. 

Hash: [Pedersen(path, value) + length] if length > 0 else [value].  

The package is integrated with the `trie2` package. It utilizes the `TrieDB` interface for reading, iterating and comiting the nodes. Similarly to the `PathDB`, `HashDB` has 2 cache layers - clean (for unmodified nodes, read from the disk) and dirty (for modified/created nodes, which will be written to the disk in the future). For the clean cache the `VictoriaMetrics``fastcache` package was used, the dirty cache however is a very simple structure, which holds three maps, each for dirty nodes from different tries (class, contract, storage). Committing simply writes all the dirty nodes to the current `Batch` and flushes it once it hits optimal size (arbitrary).


core/trie2/trieutils/accessors.go
```diff
@@ -0,0 +1,155 @@
...
+//
+// StorageTrie of a Contract :
+// [1 byte prefix][1 section byte][32 bytes owner][1 byte node-type][8 byte from path][1 path length byte][32 byte hash]
+//
+// Section:
+// 0 if state and path length is <= 5.
+// 1 if state and path length is > 5.
+// 2 if storage.
+//
+// Hash: [Pedersen(path, value) + length] if length > 0 else [value].
```
weiihann, line 108: I think we should store `node-type` after the prefix. For example:
```
// [1 byte prefix][1 byte node-type]...
```

core/trie2/trieutils/accessors.go
```diff
@@ -0,0 +1,155 @@
...
+
+	if prefix == db.ContractTrieStorage {
+		section = 2
+	} else {
+		if path.Len() <= shortPathLength {
+			section = 0
+		} else {
+			section = 1
+		}
+	}
```
weiihann, line 132: The section part seems like a duplication for `ContractTrieStorage`, because we already have a unique prefix byte for contract storage, and now adding `section = 2` which is also unique for `ContractTrieStorage`.

The [original rationale](https://github.com/NethermindEth/nethermind/pull/6331) of top level node doesn't really fit what we have in Starknet because on Ethereum, the average trie depth is around 7 or 8 IIRC, but on Starknet the height is fixed at 251. I'd suggest removing the `section` byte completely, because it seems like an optimization that should be done after we have thorough analysis on the trie node access patterns.

core/trie2/trieutils/accessors.go
```diff
@@ -0,0 +1,155 @@
...
+	} else {
+		nodeType = nonLeaf.Bytes()
+	}
+
+	key := make([]byte, 0, len(prefixBytes)+len(ownerBytes)+len(nodeType)+len(pathBytes))
+	key = append(key, prefixBytes...)
+	key = append(key, section)
+	key = append(key, ownerBytes...)
+	key = append(key, nodeType...)
+	key = append(key, pathBytes[0:pathSignificantBytes]...)
```
weiihann, line 150: Perhaps `EncodedBytes` is a bit misleading. It returns the `currently used bytes` + `length of the bit array`. So in a case where the path is `<64 bits`, the first 7 bytes are the active bytes and the last byte is the length.

I'd suggest making `BitArray.activeBytes` a public method and used it instead. Also, if the length of active bytes is less than 8 bytes, then you need to do padding (add additional zero bytes).

core/trie2/triedb/hashdb/cache.go
```diff
@@ -0,0 +1,50 @@
...
+type DirtyCache interface {
+	Get(key []byte) (cachedNode, bool)
+	Set(key []byte, value cachedNode) bool
+	Remove(key []byte) bool
+	Hits() uint64
+	HitRate() float64
+	Len() int
+	GetOldest() (key []byte, value cachedNode, ok bool)
+	RemoveOldest() bool
+}
```
weiihann, line 32: On a second thought I don't think we need to use the interfaces for `CleanCache` and `DirtyCache`. As of right now, we use LRU for dirty cache and `fastcache` for clean cache. Adding the interfaces seem to complicates things further.

core/trie2/triedb/hashdb/database.go
```diff
@@ -0,0 +1,408 @@
...
+func (d *Database) insert(bucket db.Bucket, owner felt.Felt, path trieutils.Path, hash felt.Felt, blob []byte, isLeaf bool) {
+	key := trieutils.NodeKeyByHash(bucket, owner, path, hash, isLeaf)
+	_, found := d.dirtyCache.Get(key)
+	if found {
+		return
+	}
+	d.dirtyCache.Set(key, cachedNode{
+		blob: blob,
+	})
+	d.dirtyCacheSize += len(blob)
```
weiihann, line 87: size of the key should also be considered in the size

core/trie2/triedb/hashdb/database.go
```diff
@@ -0,0 +1,408 @@
...
+			return blob, nil
+		}
+	}
+
+	if d.dirtyCache != nil {
+		node, found := d.dirtyCache.Get(key)
+		if found {
+			return node.blob, nil
+		}
+	}
```
weiihann, line 104: I think we are guaranteed to have the `cleanCache` and `dirtyCache`, so they won't be `nil` right?

core/trie2/triedb/hashdb/database.go
```diff
@@ -0,0 +1,408 @@
...
+		blob = value
+		return nil
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	if d.cleanCache != nil {
+		d.cleanCache.Set(key, blob)
+	}
```
weiihann, line 117: Same as the above, `nil` check seems unnecessary

core/trie2/triedb/hashdb/database.go
```diff
@@ -0,0 +1,408 @@
...
+
+	return blob, nil
+}
+
+func (d *Database) remove(bucket db.Bucket, owner felt.Felt, path trieutils.Path, hash felt.Felt, blob []byte, isLeaf bool) {
+	key := trieutils.NodeKeyByHash(bucket, owner, path, hash, isLeaf)
+	d.dirtyCache.Remove(key)
+	d.dirtyCacheSize -= len(blob)
```
weiihann, line 125: add key size

core/trie2/triedb/hashdb/database.go
```diff
@@ -0,0 +1,408 @@
...
+			return err
+		}
+
+		if batch.Size() > idealBatchSize {
+			if err := batch.Write(); err != nil {
+				d.log.Errorw("Failed to write flush list to disk", "error", err)
+				return err
+			}
+			batch.Reset()
+		}
```
weiihann, line 156: For the batch flushing strategy, it seems quite hard to handle crashes and recovery. Because any of nodes in the trie could potentially be missing, so it doesn't seem possible to handle it.

Maybe we just force flush on every state update. It's not performant, but I think that's fine. I'd prefer stability for archive node.

core/trie2/triedb/hashdb/database.go
```diff
@@ -0,0 +1,408 @@
...
+}
+
+func decodeRoots(val []byte) (*felt.Felt, *felt.Felt, error) {
+	if len(val) != 2*felt.Bytes {
+		return nil, nil, fmt.Errorf("invalid value length: got %d, want %d", len(val), 2*felt.Bytes)
+	}
+	var classRoot, contractRoot felt.Felt
+	classRoot.SetBytes(val[:felt.Bytes])
+	contractRoot.SetBytes(val[felt.Bytes:])
+	return &classRoot, &contractRoot, nil
```
weiihann, line 359: should return value instead of reference

core/trie2/triedb/hashdb/database.go
```diff
@@ -0,0 +1,408 @@
...
+	default:
+		return nil, fmt.Errorf("unknown node type: %d", nodeType)
+	}
+
+	return children, nil
+}
+
+func (d *Database) getRootsForStateCommitment(stateComm felt.Felt) (*felt.Felt, *felt.Felt, error) {
+	bytes := stateComm.Bytes()
+	key := db.StateHashToRoots.Key(bytes[:])
```
weiihann, line 328: Please add this to the `schema.go` file instead

core/trie2/triedb/hashdb/database.go
```diff
@@ -0,0 +1,408 @@
...
+		children, err := d.getNodeChildren(root.blob)
+		if err != nil {
+			return fmt.Errorf("failed to get node children: %w", err)
+		}
+		for i, childKey := range children {
+			childPath := new(trieutils.Path).AppendBit(&path, uint8(i))
+			if err := d.commit(batch, bucket, *childPath, childKey); err != nil {
+				return err
+			}
+		}
```
weiihann, line 252: I think we are complicating things here. We don't really need to traverse through all the child nodes and then commit them one by one. In the dirty cache and clean cache, we can simply maintain separate the maps for each trie type (i.e. class trie, contract trie, storage trie). Flushing is as simple as iterating through all the items in the maps and putting into the database.

Unless I'm missing something here

core/trie2/triedb/hashdb/cache.go
weiihann: I commented somewhere below but I'm gonna repeat here:

In the dirty cache, we can simply maintain separate the maps for each trie type (i.e. class trie, contract trie, storage trie). Flushing is as simple as iterating through all the items in the maps and putting into the database.

weiihann:
I'm gonna stop here for now. There are some implementation details that can be improved, but let's finalize the overall architecture first. A few items to take note:
- Flushing strategy - prefer stability to performance (flush on every state update)
- Trie commit strategy - it should be as simple as adding the trie nodes to maps (i.e. `map[hash]node`) instead of using the dereferencing strategy
- The data structure used for dirty cache and clean cache
- Trie node key format

core/trie2/trieutils/accessors.go
MaksymMalicki: Yes, I agree with that. Actually it's older part of the code which I didn't revisit in a long time

core/trie2/trieutils/accessors.go
MaksymMalicki: referenced below

core/trie2/trieutils/accessors.go
MaksymMalicki: done, thanks for clarification

core/trie2/triedb/hashdb/cache.go
MaksymMalicki: yes, the interfaces were added, to make measurements of performance of differerent cache types easier. We can drop the idea for now

core/trie2/triedb/hashdb/database.go
MaksymMalicki: yes, this should be enforced

core/trie2/triedb/hashdb/database.go
MaksymMalicki: removed

core/trie2/triedb/hashdb/database.go
MaksymMalicki: facts

core/trie2/triedb/hashdb/database.go
MaksymMalicki: thats right

core/trie2/triedb/hashdb/database.go
MaksymMalicki: done

core/trie2/triedb/hashdb/database.go
MaksymMalicki: done

core/trie2/triedb/hashdb/clean_cache.go
```diff
@@ -0,0 +1,32 @@
...
+)
+
+type CleanCache struct {
+	cache *fastcache.Cache
+}
+
+func NewCleanCache(size int) *CleanCache {
+	if size <= 0 {
+		return nil
+	}
```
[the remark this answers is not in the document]
MaksymMalicki: done

core/trie2/triedb/hashdb/clean_cache.go
```diff
@@ -0,0 +1,32 @@
...
+	return &CleanCache{
+		cache: fastcache.New(size),
+	}
```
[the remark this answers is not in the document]
MaksymMalicki: done

core/trie2/triedb/hashdb/database.go
```diff
@@ -0,0 +1,202 @@
...
+	config Config
+
+	cleanCache *CleanCache
+	dirtyCache *DirtyCache
+
+	lock sync.RWMutex
+	log  utils.SimpleLogger
+}
+
+func New(disk db.KeyValueStore, config *Config) *Database {
```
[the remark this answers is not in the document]
MaksymMalicki: its meant to be a long lived object, created only once as well at the node startup

core/trie2/triedb/hashdb/database.go
```diff
@@ -0,0 +1,202 @@
...
+	if config == nil {
+		config = &Config{
+			CleanCacheSize: 1024 * 1024 * 64,
+		}
```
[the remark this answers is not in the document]
MaksymMalicki: added

core/trie2/triedb/hashdb/database.go
```diff
@@ -0,0 +1,202 @@
...
+
+func (d *Database) insert(owner *felt.Felt, path *trieutils.Path, hash *felt.Felt, isClass bool, node trienode.TrieNode) {
+	_, found := d.dirtyCache.getNode(owner, path, hash, isClass)
+	if found {
+		return
+	}
+	d.dirtyCache.putNode(owner, path, hash, isClass, node)
+}
+
+func (d *Database) node(bucket db.Bucket, owner *felt.Felt, path *trieutils.Path, hash *felt.Felt, isLeaf bool) ([]byte, error) {
```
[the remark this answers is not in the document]
MaksymMalicki: it could maybe be `readNode`? more verbose name

core/trie2/trie.go
```diff
@@ -46,8 +46,18 @@ type Trie struct {
...
+// - height: the height of the trie, 251 for contract and class trie
+// - hashFn: the hash function to use
+// - nodeDB: database interface, which provides methods to read and write trie nodes
+// - rootHash: the root hash of the trie (provided only for hash scheme of nodeDB, needed to properly recreate the trie)
+func New(id trieutils.TrieID,
+	height uint8,
+	hashFn crypto.HashFn,
+	nodeDB database.NodeDatabase,
+	rootHash *felt.Felt,
+) (*Trie, error) {
```
rodrigo-pino, line 60: I am not fully convinced here, I don't like the idea of having optional arguments. I would prefer to have two constructors one for hash db and another for path db. Is that what we are doing here or am I missing something?

core/trie2/triedb/hashdb/clean_cache.go
```diff
@@ -0,0 +1,29 @@
...
+
+type CleanCache struct {
+	cache *fastcache.Cache
+}
+
+func NewCleanCache(size uint64) CleanCache {
+	return CleanCache{
+		cache: fastcache.New(int(size)),
+	}
+}
```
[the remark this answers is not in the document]
MaksymMalicki: done

core/trie2/trie.go
MaksymMalicki: Yes exactly. Okey I agree, it isn't the cleanest. Added `NewWithRootHash` constructor, will use it later during the hashdb integration

core/trie2/trie.go
```diff
@@ -66,7 +74,42 @@ func New(id trieutils.TrieID, height uint8, hashFn crypto.HashFn, nodeDB databas
...
+		return nil, err
+	}
+	tr.root = root
+
+	return tr, nil
+}
+
+// Similar to New, creates a new trie, but additionally takes one more argument:
+// - rootHash: the root hash of the trie (provided only for hash scheme of trieDB, needed to properly recreate the trie)
+func NewWithRootHash(id trieutils.TrieID,
```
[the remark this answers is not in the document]
MaksymMalicki: done

core/trie2/trie.go
```diff
@@ -66,7 +75,43 @@ func New(id trieutils.TrieID, height uint8, hashFn crypto.HashFn, nodeDB databas
...
+		return nil, err
+	}
+	tr.root = root
+
+	return tr, nil
+}
+
+// Similar to New, creates a new trie, but additionally takes one more argument:
+// - rootHash: the root hash of the trie (provided only for hash scheme of trieDB, needed to properly recreate the trie)
+func NewWithRootHash(
```
rodrodros: So this create a new trie with Hash Scheme? Why not call it `NewWithHashScheme`?

core/trie2/trienode/node_enc.go
```diff
@@ -157,7 +157,7 @@ func decodeBinaryNode(blob []byte, hash *felt.Felt, pathLen, maxPathLen uint8) (
 	}
 
 	binary := &BinaryNode{}
-	if !hash.IsZero() {
+	if hash != nil && !hash.IsZero() {
```
rodrodros: Why `hash` would be `nil`. I am fine with the check but is that an expected behaviour or a safety check?

core/trie2/trienode/node_enc.go
```diff
@@ -177,7 +177,7 @@ func decodeEdgeNode(blob []byte, hash *felt.Felt, pathLen, maxPathLen uint8) (*E
 	}
 
 	edge := &EdgeNode{Path: &trieutils.Path{}}
-	if !hash.IsZero() {
+	if hash != nil && !hash.IsZero() {
```
rodrodros: Marking this part of the code because however the above comment is resolved will impact this as well

core/trie2/trie.go
MaksymMalicki: So I would say this name would be missleading. We will be creating a single, long lived DB object with given scheme, and this method will be used. However its internals don't directly depend on the state. In the scope of this package, it just reads the root from the DB using the information about its hash, and recreates the trie. Maybe a good rename option would be `NewFromRootHash`

core/trie2/trienode/node_enc.go
MaksymMalicki: Should never be nil. I don't think this check is necessary in this case, it used to be a protection in some past scenario - which turned out to be a bug. I would rather error out the case when the hash is nil

core/trie2/trienode/node_enc.go
```diff
@@ -156,8 +156,12 @@ func decodeBinaryNode(blob []byte, hash *felt.Felt, pathLen, maxPathLen uint8) (
 		return nil, fmt.Errorf("invalid binary node size: %d", len(blob))
 	}
 
+	if hash == nil {
+		return nil, fmt.Errorf("hash is nil")
+	}
```
rodrodros: Let's remove this check, we will panic if it is the case, which should never be the case. If we were to be consistent every reference argument should then be checked which we will not do

core/trie2/trienode/node_enc.go
```diff
@@ -176,8 +180,12 @@ func decodeEdgeNode(blob []byte, hash *felt.Felt, pathLen, maxPathLen uint8) (*E
 		return nil, fmt.Errorf("invalid edge node size: %d", len(blob))
 	}
 
+	if hash == nil {
+		return nil, fmt.Errorf("hash is nil")
+	}
+
```
rodrodros: mark

rodrigo-pino:
Bypassing merge rules and going ahead

PR #2755, opened 2025-04-10
refactor: Use a loop to run rules until no rules available

infrmtcs:


consensus/tendermint/tendermint_test.go
```diff
@@ -162,7 +162,7 @@ func TestStartRound(t *testing.T) {
 		assert.Contains(t, algo.messages.proposals[expectedHeight][expectedRound], *nodeAddr)
 		assert.Equal(t, expectedProposalMsg, algo.messages.proposals[expectedHeight][expectedRound][*nodeAddr])
 
-		assert.Equal(t, propose, algo.state.step)
+		assert.Equal(t, prevote, algo.state.step)
```
infrmtcs, line 165: This is because in this test, the node is the proposer, so when the rules automatically apply, it should move to prevote.

consensus/tendermint/tendermint.go
```diff
@@ -357,43 +291,26 @@ func q(totalVotingPower votingPower) votingPower {
...
-			return false
-		}
-
-		addMessage()
-
-		if t.uponSkipRound(header.Round) {
-			t.doSkipRound(header.Round)
-		}
-		return false
+func (t *Tendermint[V, H, A]) preprocessMessage(header MessageHeader[A], addMessage func()) bool {
```
infrmtcs, line 294: This requires rewrite in order to move the `skipRound` rule outside of the function.

MaksymMalicki (approved):

consensus/tendermint/tendermint.go
```diff
@@ -357,43 +291,26 @@ func q(totalVotingPower votingPower) votingPower {
...
-	return true
 }
 
-func (t *Tendermint[V, H, A]) preprocessMessage(header MessageHeader[A], addMessage func()) bool {
-	return header.Height >= t.state.height &&
-		t.handleFutureHeightMessage(header, addMessage) &&
-		t.handleFutureRoundMessage(header, addMessage)
+func isCurrentOrNearFuture[T ~uint | ~int](header, current, maxFuture T) bool {
+	return header >= current && header <= current+maxFuture
 }
```
rianhughes, line 314: preprocessMessage is a bit hard to understand initially. Can we improve the readability by doing something like the following? 

```suggestion
// preprocessMessage calls `addMessage` if the `msgHeader` is within `maxFutureHeight` of the current height
// and `maxFutureRound` of the current round, otherwise returns false.
func (t *Tendermint[V, H, A]) preprocessMessage(msgHeader MessageHeader[A], addMessage func()) bool {
	var currentRound round
	msgIsCurrentHeight := msgHeader.Height == t.state.height
	if msgIsCurrentHeight {
		currentRound = t.state.round
	}

	switch {
	case !isCurrentOrNearFuture(msgHeader.Height, t.state.height, maxFutureHeight):
		return false
	case !isCurrentOrNearFuture(msgHeader.Round, currentRound, maxFutureRound):
		return false
	default:
		addMessage()
		return msgIsCurrentHeight
	}
}

func isCurrentOrNearFuture[T ~uint | ~int](msgHeader, currentState, maxFuture T) bool {
	return msgHeader >= currentState && msgHeader <= currentState+maxFuture
}
```

consensus/tendermint/process.go
```diff
@@ -0,0 +1,76 @@
+package tendermint
+
+func (t *Tendermint[V, H, A]) processMessage(header MessageHeader[A], addMessage func()) {
+	if !t.preprocessMessage(header, addMessage) {
+		return
+	}
+
+	t.processLoop(&header.Round)
+}
```
rianhughes, line 9: I don't think we should ignore messages in the current height but previous rounds. I think Prevote and Precommit msgs need to be handled. Eg Previously these may trigger

```
if cachedProposal != nil && t.uponProposalAndPolkaPrevious(cachedProposal, p.Round) {
		t.doProposalAndPolkaPrevious(cachedProposal)
	}
```

and (I think)

```
if t.uponProposalAndPolkaPrevious(cachedProposal, p.ValidRound) {
		t.doProposalAndPolkaPrevious(cachedProposal)
	}
```    

but this is no longer the case from what I can see

rianhughes: Also, eg if we are at height 3 and rond 4, we would also ignore all msgs with rounds less than 4, if they come from a nearby future height? I don't think this is correct either. Althought maybe I'm getting something wrong here

consensus/tendermint/tendermint.go
rianhughes: My suggested changes aren't the cleanest sorry

consensus/tendermint/process.go
rianhughes: Let me double check this actually

infrmtcs: This is a great catch! The previous commit wasn't handle the past round properly, so I modified the preprocess function to fix it.

consensus/tendermint/tendermint.go
```diff
@@ -357,43 +291,29 @@ func q(totalVotingPower votingPower) votingPower {
...
-		}
-
-		if header.Round > maxFutureRound {
-			return false
-		}
-
-		addMessage()
 
+	switch {
+	case header.Height < t.state.height || header.Height > t.state.height+maxFutureHeight:
```
rianhughes, line 309: I wonder if this slows down the time it takes to catch up? Eg imagine the node is at block 0, and the consensus head is at block 100. Currently we can only jump forward 5 blocks at a time. However, we should be able to jump straight to block 100 in principle.

consensus/tendermint/process.go
```diff
@@ -0,0 +1,76 @@
...
+	cachedProposal := t.findProposal(t.state.round)
+
+	var roundCachedProposal *CachedProposal[V, H, A]
+	if recentlyReceivedRound != nil {
+		roundCachedProposal = t.findProposal(*recentlyReceivedRound)
+	}
+
+	switch {
+	// Line 22
+	case cachedProposal != nil && t.uponFirstProposal(cachedProposal):
```
rianhughes, line 40: Should this be `cachedProposal == nil &&`?

consensus/tendermint/tendermint.go
infrmtcs: The current implementation doesn't cover it yet, it's called "state sync". I think it should be a separate feature and type of message to request blocks in batch with signature proof from another peer, instead of going through the main algorithm.
Initially, the `maxFutureHeight` and `maxFutureRound` were introduced to prevent spamming, but I also think that it's maybe not a good way to do so.

consensus/tendermint/rule_proposal_and_polka_previous.go
```diff
@@ -11,11 +11,10 @@ Check the upon condition on line 28:
...
-func (t *Tendermint[V, H, A]) uponProposalAndPolkaPrevious(cachedProposal *CachedProposal[V, H, A], prevoteRound round) bool {
+func (t *Tendermint[V, H, A]) uponProposalAndPolkaPrevious(cachedProposal *CachedProposal[V, H, A]) bool {
 	vr := cachedProposal.ValidRound
 	hasQuorum := t.checkQuorumPrevotesGivenProposalVID(vr, *cachedProposal.ID)
-	return cachedProposal.ValidRound == prevoteRound &&
-		hasQuorum &&
+	return hasQuorum &&
 		t.state.step == propose &&
 		vr >= 0 &&
 		vr < t.state.round
```
rianhughes, line 20: Hey, my question is actually related to `doProposalAndPolkaPrevious` but github won't let me make a comment that.

What do you think about moving the `t.state.step = prevote` line from `t.sendPrevote(votedID)`, and placing it after whereever we call `t.sendPrevote(votedID)`? Currently sendPrevote() also changes the states step field, which doesn't feel as clean / readable? I was trying to see where we change this step, but couldn't find it my "clicking through the code" and had to explicitly search for it

rianhughes: We can do it in a seperate PR as it's not related to the changes here, but just wanted to bring it up before I forget

rianhughes: similar for `sendPrecommit`

consensus/tendermint/process.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+func (t *Tendermint[V, H, A]) processLoop(recentlyReceivedRound *round) {
+	shouldContinue := t.process(recentlyReceivedRound)
+	for shouldContinue {
+		shouldContinue = t.process(recentlyReceivedRound)
+	}
+}
+
+func (t *Tendermint[V, H, A]) process(recentlyReceivedRound *round) bool {
```
rianhughes, line 30: Nice, I really like this. It makes reasoning about Tendermint much easier. 

Eg if we receive a quorum of `prevote`s before the `proposal`. We won't update the state until we get the `proposal`, and then we immediately check (in the next loop) that we can call `doProposalAndPolkaCurrent` to handle the quorum of `prevote`s, etc

consensus/tendermint/rule_proposal_and_polka_previous.go
infrmtcs: Originally I grouped them into `sendX` functions because I noticed that there are some actions that always be done when the algorithm needs to broadcast a message:
- Set `step` to X
- Store the message
- Push X to a broadcast channel
The rational behind this was to avoid forgetting to set the step when broadcast the message, as each of `sendX` is called multiple times, although I think your argument makes sense too.

consensus/tendermint/tendermint.go
infrmtcs: Yeah, I also think in general it should be the scope of the P2P layer instead of the core algorithm, we can even switch to throttling the messages from talkative peers instead of dropping their messages.

consensus/tendermint/rule_proposal_and_polka_previous.go
rianhughes: Yup, that makes sense too, I was thinking the same thing. Although does make it a bit harder to understand since it's easy to forget (or not realise in the first place) that `sendX` also updates the state. Imo I think it's worth splitting this logic, but I'll leave it up to you.

rianhughes (approved):

PR #2761, opened 2025-04-15
refactor: Separate driver and state machine in Tendermint

infrmtcs:
This PR mainly does 2 things:
- Separate the `Tendermint` struct into 2 structs, one is `Tendermint` (I keep the name to reduce the diff) for the pure execution of the state machine itself, and the other is `Driver` which is in charge of the concurrent operations.
- The state machine functions now returns `Action` objects which represent the concurrent actions, including broadcast message and schedule timeout.
Minor changes:
- Rename some functions to better readability.

### DIFF
--- a/consensus/tendermint/action.go
+++ b/consensus/tendermint/action.go
@@ -1,56 +1,21 @@
 package tendermint
 
-import (
-	"time"
-)
-
 type Action[V Hashable[H], H Hash, A Addr] interface {
-	Execute(d *Driver[V, H, A])
+	isTendermintAction()
 }
 
 type BroadcastProposal[V Hashable[H], H Hash, A Addr] Proposal[V, H, A]
 
-type BroadcastPrevote[V Hashable[H], H Hash, A Addr] Prevote[H, A]
+type BroadcastPrevote[H Hash, A Addr] Prevote[H, A]
 
-type BroadcastPrecommit[V Hashable[H], H Hash, A Addr] Precommit[H, A]
+type BroadcastPrecommit[H Hash, A Addr] Precommit[H, A]
 
-type ScheduleTimeout[V Hashable[H], H Hash, A Addr] timeout
+type ScheduleTimeout timeout
 
-func (a *BroadcastProposal[V, H, A]) Execute(d *Driver[V, H, A]) {
-	d.broadcasters.ProposalBroadcaster.Broadcast(Proposal[V, H, A](*a))
-}
+func (a *BroadcastProposal[V, H, A]) isTendermintAction() {}
 
-func (a *BroadcastPrevote[V, H, A]) Execute(d *Driver[V, H, A]) {
-	d.broadcasters.PrevoteBroadcaster.Broadcast(Prevote[H, A](*a))
-}
+func (a *BroadcastPrevote[H, A]) isTendermintAction() {}
 
-func (a *BroadcastPrecommit[V, H, A]) Execute(d *Driver[V, H, A]) {
-	d.broadcasters.PrecommitBroadcaster.Broadcast(Precommit[H, A](*a))
-}
+func (a *BroadcastPrecommit[H, A]) isTendermintAction() {}
 
-func (a *ScheduleTimeout[V, H, A]) Execute(d *Driver[V, H, A]) {
-	var duration time.Duration
-	switch a.s {
-	case propose:
-		duration = d.timeoutPropose(a.r)
-	case prevote:
-		duration = d.timeoutPrevote(a.r)
-	case precommit:
-		duration = d.timeoutPrecommit(a.r)
-	default:
-		return
-	}
-
-	d.scheduledTms[timeout(*a)] = time.AfterFunc(duration, func() {
-		select {
-		case <-d.quit:
-		case d.timeoutsCh <- timeout(*a):
-		}
-	})
-}
-
-func (d *Driver[V, H, A]) execute(actions []Action[V, H, A]) {
-	for _, action := range actions {
-		action.Execute(d)
-	}
-}
+func (a *ScheduleTimeout) isTendermintAction() {}
--- a/consensus/tendermint/action.go
+++ b/consensus/tendermint/action.go
@@ -1,0 +1,56 @@
+package tendermint
+
+import (
+	"time"
+)
+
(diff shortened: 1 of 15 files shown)


consensus/tendermint/action.go
```diff
@@ -0,0 +1,56 @@
...
+
+type Action[V Hashable[H], H Hash, A Addr] interface {
+	Execute(d *Driver[V, H, A])
+}
+
+type BroadcastProposal[V Hashable[H], H Hash, A Addr] Proposal[V, H, A]
+
+type BroadcastPrevote[V Hashable[H], H Hash, A Addr] Prevote[H, A]
+
+type BroadcastPrecommit[V Hashable[H], H Hash, A Addr] Precommit[H, A]
```
rodrigo-pino, line 15: Why do we define generic types `V` for these type which don't use it explicitly? Is it necessary or do you think it's good for legibility/extensibility?

consensus/tendermint/action.go
```diff
@@ -0,0 +1,56 @@
...
+
+type ScheduleTimeout[V Hashable[H], H Hash, A Addr] timeout
+
+func (a *BroadcastProposal[V, H, A]) Execute(d *Driver[V, H, A]) {
+	d.broadcasters.ProposalBroadcaster.Broadcast(Proposal[V, H, A](*a))
```
rodrigo-pino, line 20: Is the copy made in this method intentional?

We are re interpreting `a` as a different type and copying it, why not use a reference instead?

rodrigo-pino (approved):
Liked the separation. Some small comments

consensus/tendermint/action.go
infrmtcs: The main reason is that currently `Action` requires all of `V, H, A` to determint the type arguments of `Driver` in `Execute` function.

infrmtcs: Pushed the fix

consensus/tendermint/action.go
infrmtcs: Is it ok if I address this in another PR? It needs some changes in tests, which will conflict with my next PR.

consensus/tendermint/rule_commit_value.go
```diff
@@ -26,12 +26,12 @@ func (t *Tendermint[V, H, A]) uponCommitValue(cachedProposal *CachedProposal[V,
 	return hasQuorum && isValid
 }
 
-func (t *Tendermint[V, H, A]) doCommitValue(cachedProposal *CachedProposal[V, H, A]) {
+func (t *Tendermint[V, H, A]) doCommitValue(cachedProposal *CachedProposal[V, H, A]) Action[V, H, A] {
 	// TODO: Optimise this
 	precommits, _ := t.checkForQuorumPrecommit(cachedProposal.Round, *cachedProposal.ID)
 	t.blockchain.Commit(t.state.height, *cachedProposal.Value, precommits)
```
rianhughes, line 32: Will the Tendermint State type interact with the blockchain/sequencer service directly, or will it also have to go through the driver? If so, then we might have to return an addtional action here (although maybe we can worry about that when we implement that logic)

consensus/tendermint/rule_first_proposal.go
```diff
@@ -25,5 +25,5 @@ func (t *Tendermint[V, H, A]) doFirstProposal(cachedProposal *CachedProposal[V,
 	if shouldVoteForValue {
 		votedID = cachedProposal.ID
 	}
-	t.sendPrevote(votedID)
+	return t.setStepAndSendPrevote(votedID)
```
rianhughes, line 28: Just an observation, but it looks like the logic inside this function is being pushed to the Driver as an Action, whereas we actually just want to push the BroadcastPrevote to the Driver. I think we could restructure this slightly to improve the readability, but we don't have to do it in this PR

consensus/tendermint/tendermint.go
```diff
@@ -187,49 +202,77 @@ type CachedProposal[V Hashable[H], H Hash, A Addr] struct {
...
 
-func (t *Tendermint[V, H, A]) Stop() {
-	close(t.quit)
-	t.wg.Wait()
-	for _, tm := range t.scheduledTms {
-		tm.Stop()
+func (d *Driver[V, H, A]) execute(actions []Action[V, H, A]) {
+	for _, action := range actions {
+		switch action := action.(type) {
+		case *BroadcastProposal[V, H, A]:
```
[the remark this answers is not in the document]
rianhughes: Actually, maybe it's not neccessary

consensus/tendermint/process.go
```diff
@@ -1,33 +1,58 @@
...
+}
+
+func (t *Tendermint[V, H, A]) processPrevote(p Prevote[H, A]) []Action[V, H, A] {
+	return t.processMessage(p.MessageHeader, func() { t.messages.addPrevote(p) })
+}
+
+func (t *Tendermint[V, H, A]) processPrecommit(p Precommit[H, A]) []Action[V, H, A] {
+	return t.processMessage(p.MessageHeader, func() { t.messages.addPrecommit(p) })
+}
+
```
rianhughes, line 18: Why is the motivation for creating these functions, we can just inline them right?

infrmtcs: My main motivation is because this is modifying the internal state of the state machine via `t.messages`, which should be inaccessible from the driver. We'll need it when we separate them into 2 different packages.

consensus/tendermint/rule_first_proposal.go
infrmtcs: I got your point. Probably we can separate it into 2 lines, 1 for getting the action struct and 1 for return the action.

consensus/tendermint/rule_commit_value.go
infrmtcs: Agree with your point. This is also one of my concern, however I wasn't sure about how the `Commit` function implementation looks like for now, so it's easier to refactor once I have a better understanding.

consensus/tendermint/tendermint.go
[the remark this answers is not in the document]
infrmtcs: I'm wondering if there exists a linter that can double check if the type switch is exhaustive. My intuition is to avoid having a default case, so it's better to fail the check when a new possible case is added instead of silently falling back to default case.

Issue #2765, opened 2025-04-16
Failing trace transaction


rodrodros:
Unexpected error occurred in Sepolia endpoint for method:

```json
[{"jsonrpc":"2.0","method":"starknet_traceBlockTransactions","params":[{"block_number":693414}],"id":1}]
```
Error:
```json
[
    {
        "jsonrpc": "2.0",
        "error": {
            "code": 63,
            "message": "An unexpected error occurred",
            "data": "execute transaction #3: {\"class_hash\":\"0x3957f9f5a1cbfe918cedc2015c85200ca51a5f7506ecb6de98a5207b759bf8a\",\"contract_address\":\"0x71d44556dfdb51b55871700bbf9a23210b4eeef739b520896cebf29d05f97f0\",\"error\":{\"class_hash\":\"0x3957f9f5a1cbfe918cedc2015c85200ca51a5f7506ecb6de98a5207b759bf8a\",\"contract_address\":\"0x71d44556dfdb51b55871700bbf9a23210b4eeef739b520896cebf29d05f97f0\",\"error\":{\"class_hash\":\"0xb0033df5e7911aee0e0edff293d9e89bcdf9169cf768fbe5bad8b0911a088a\",\"contract_address\":\"0x3f32e152b9637c31bfcf73e434f78591067a01ba070505ff6ee195642c9acfb\",\"error\":\"\\\"Attestation is out of window\\\"\",\"selector\":\"0x37446750a403c1b4014436073cf8d08ceadc5b156ac1c8b7b0ca41a0c9c1c54\"},\"selector\":\"0x15d40a3d6ca2ac30f4031e42be28da9b056fef9bb7357ac5e85627ee876e5ad\"},\"selector\":\"0x15d40a3d6ca2ac30f4031e42be28da9b056fef9bb7357ac5e85627ee876e5ad\"}"
        },
        "id": 1
    }
]
```

PR #2766, opened 2025-04-16
rpc fix: include err_on_revert flag in binary search execution flow

rianhughes:
closes https://github.com/NethermindEth/juno/issues/2765

Basically, blockifier should never return an error when tracing a block. Transactions can revert, but this is not an error and is handled in the `Ok()` flow. For normal execution (ie non binary search), we have a flag `err_on_revert` that specifies whether we should classify these reverts as errors or not. When tracing blocks this flag is set to `false`. However, this was not propagated to the binary search execution flow. This PR includes this flag in that binary search flow.

The only change is that `simulate_execution` now accepts `error_on_revert: bool`, which is used in the second `Ok()` flow
`if tx_info.is_reverted() && error_on_revert`. Everything else should be from format changes


rianhughes:
We actually shouldn't be able to enter the binary search flow when getting traces. The next commit updates the code to reflect this change

rodrigo-pino:
Just for context, I am approving because this is an error fix but all this use of parameters in the execute function seem excesive to me. I think a good approach would be to have wrappers. Like `executeEstimateFee` and then that function call `execute` with the right params and so on. This is very hard to read and mantain. For the (near) future

MaksymMalicki (approved):

PR #2767, opened 2025-04-16
docs: small docs change on call-max-steps usage

MaksymMalicki:


wojciechos:
Let's also update it on [v0.14.0 ](https://github.com/NethermindEth/juno/blob/main/docs/versioned_docs/version-0.14.0/_config-options.md)which is displayed as latest.

wojciechos (approved):

PR #2771, opened 2025-04-17
fix(rpc): temporary fix for out-of-gas estimation

MaksymMalicki:
This PR is a temporary fix for the `Out of gas` error, which appears during calling `starknet_estimateFee`. This PR is based on [this](https://github.com/NethermindEth/juno/pull/2638/files#diff-fa1da12fdf0a62002aeafaa6fda37721c6dd59b1ac5490d4f1bef9eb46615ce1).

The problem:
Currently, when the user calls `starknet_estimateFee`, with a CURL like this:
```
curl --location https://free-rpc.nethermind.io/integration-sepolia-juno/rpc/v0_8 --header 'Content-Type: application/json'  \
  --data '{
    "jsonrpc":"2.0",
    "method":"starknet_estimateFee",
    "params":
        {
            "block_id": {"block_number": 71202},
            "simulation_flags": ["SKIP_VALIDATE"],
            "request": [
                {"signature": [], "version": "0x3", "tip": "0x0", "nonce_data_availability_mode": "L1", "fee_data_availability_mode": "L1", "paymaster_data": [], "resource_bounds": {"l1_gas": {"max_amount": "0x0", "max_price_per_unit": "0x0"}, "l2_gas": {"max_amount": "0x0", "max_price_per_unit": "0x0"}, "l1_data_gas": {"max_amount": "0x0", "max_price_per_unit": "0x0"}}, "calldata": ["0x1", "0x4718f5a0fc34cc1af16a1cdee98ffb20c31f5cd61d6ab07201858f4287c938d", "0x83afd3f4caedc6eebf44246fe54e38c95e3179a5ec9ea81740eca5b482d12e", "0x3", "0x492fa2fc8f8be26888785b5b548c5630cdc743d15cdd1385d48e92d715b0dad", "0x1", "0x0"], "sender_address": "0x745d525a3582e91299d8d7c71730ffc4b1f191f5b219d800334bc0edad0983b", "nonce": "0x709b", "account_deployment_data": [], "type": "INVOKE"}
            ]
        },
    "id":1
}'
```
Juno fails with `Out of gas` error, which is not correct behavior. User should be able to specify any resource bounds for the estimation. The estimation was failing, because `l2_gas_limit > initial_gas_limit` and we were not checking for the `charge_fee` execution flag, which enforces fee limits and ensures, that the transaction will revert if insufficient balance/gas, which is not necessary for the fee estimation.


vm/rust/src/execution.rs
```diff
@@ -324,39 +331,92 @@ fn set_l2_gas_limit(
...
+
+            if balance > max_fee_without_l2_gas.0.into() {
+                // Calculate the maximum amount of L2 gas that can be bought with the balance.
+                let max_l2_gas = (balance - max_fee_without_l2_gas.0)
+                    / initial_resource_bounds
+                        .l2_gas
+                        .max_price_per_unit
+                        .0
+                        .max(1u64.into());
+                Ok(u64::try_from(max_l2_gas).unwrap_or(u64::MAX).into())
```
rodrigo-pino, line 408: I am unsure if we should go for u64::Max, we can be exposing ourselves to overflow errors during the execution. I am still not sure how could it happened but I remember some communication about it in the Shared Node meetings. Could we set `GasAmount::Max` here? Does it makes sense?

MaksymMalicki: GasAmount::Max is u64:MAX underneath

rianhughes:
LGTM, just want to test it locally

rianhughes (approved):

vm/rust/src/execution.rs
rodrigo-pino: Any reason why not use `GasAmount::Max` then?

rodrigo-pino (approved):

vm/rust/src/execution.rs
MaksymMalicki: just type casting, let me change it

PR #2777, opened 2025-04-19
fix: nil resource bounds incorrectly map to struct

cwkang1998:
When `l1_data_gas` is nil from `core.ResourceBounds` which is a map, the adaptation to `rpcv8.ResourceBound` is incorrect, which also causes the marshalling to json to be incorrect.

This PR fixes that via adding an additional check for the `core.ResourceBounds` map during adaptation and at the same time add custom json marshal to handle cases where `l1_data_gas` is nil.


rodrigo-pino (approved):

weiihann (approved):
LGTM

PR #2781, opened 2025-04-22
test: Tendermint state machine unit tests

infrmtcs:
This PR implemented:
- A "DSL" for writing scenarios for tests asserting a single node behaviour. This includes sending messages, trigger timeouts and asserting state and output actions.
- Increase the test coverage for the state machine to near 100%.


consensus/tendermint/tendermint_test.go
```diff
@@ -83,159 +81,23 @@ func (v *validators) addValidator(addr felt.Felt) {
...
-		PrecommitBroadcaster: newSenderAndReceiver[Precommit[felt.Felt, felt.Felt], value, felt.Felt, felt.Felt](),
-	}
+func setupStateMachine(t *testing.T, selectedNode int) *Tendermint[value, felt.Felt, felt.Felt] {
+	t.Helper()
+	app, chain, vals := newApp(), newChain(), newVals()
 
-	return listeners, broadcasters
-}
-
-func TestStartRound(t *testing.T) {
```
infrmtcs, line 125: These tests are moved to separate files

consensus/tendermint/propose_test.go
```diff
@@ -1,544 +0,0 @@
...
-	"slices"
-	"testing"
-	"time"
-
-	"github.com/NethermindEth/juno/core/felt"
-	"github.com/NethermindEth/juno/utils"
-	"github.com/stretchr/testify/assert"
-)
-
-func TestPropose(t *testing.T) {
```
infrmtcs, line 13: These tests are moved to separate files

consensus/tendermint/round_utils_test.go
```diff
@@ -0,0 +1,167 @@
...
+		testing:        t.testing,
+		extractMessage: nil,
+		actions:        t.stateMachine.processTimeout(timeout{s: s, h: t.height, r: t.round}),
+	}
+}
+
+func (t testRoundUtils) start() actionAsserter[any] {
+	return actionAsserter[any]{
+		testing:        t.testing,
+		extractMessage: nil,
```
rodrigo-pino, line 123: Suggestion `extractMessage` to `message`

rodrigo-pino (approved):
LG!

consensus/tendermint/tendermint_test.go
```diff
@@ -83,159 +81,23 @@ func (v *validators) addValidator(addr felt.Felt) {
...
-
-		assert.Equal(t, expectedPrevoteMsg, prevoteMsg)
-		assert.Contains(t, algo.stateMachine.messages.prevotes[expectedHeight][expectedRound], *nodeAddr)
-		assert.Equal(t, expectedPrevoteMsg, algo.stateMachine.messages.prevotes[expectedHeight][expectedRound][*nodeAddr])
-
-		assert.Equal(t, prevote, algo.stateMachine.state.step)
-		assert.Equal(t, expectedHeight, algo.stateMachine.state.height)
-		assert.Equal(t, expectedRound, algo.stateMachine.state.round)
-	})
+	return New(*nodeAddr, app, chain, vals)
```
rianhughes, line 100: Why does the `New` function return by reference and not value? Is it too heavy or something?

consensus/tendermint/tendermint_test.go
```diff
@@ -83,159 +81,23 @@ func (v *validators) addValidator(addr felt.Felt) {
...
-		PrevoteListener:   newSenderAndReceiver[Prevote[felt.Felt, felt.Felt], value, felt.Felt, felt.Felt](),
-		PrecommitListener: newSenderAndReceiver[Precommit[felt.Felt, felt.Felt], value, felt.Felt, felt.Felt](),
-	}
-
-	broadcasters := Broadcasters[value, felt.Felt, felt.Felt]{
-		ProposalBroadcaster:  newSenderAndReceiver[Proposal[value, felt.Felt, felt.Felt], value, felt.Felt, felt.Felt](),
-		PrevoteBroadcaster:   newSenderAndReceiver[Prevote[felt.Felt, felt.Felt], value, felt.Felt, felt.Felt](),
-		PrecommitBroadcaster: newSenderAndReceiver[Precommit[felt.Felt, felt.Felt], value, felt.Felt, felt.Felt](),
-	}
+func setupStateMachine(t *testing.T, selectedNode int) *Tendermint[value, felt.Felt, felt.Felt] {
```
rianhughes, line 88: nit: `setupStateMachine(t *testing.T, numValidators, thisValidator int)`

consensus/tendermint/tendermint_test.go
infrmtcs: `Tendermint` struct represents the state of the state machine, and its methods mutate the state, so it needs to be a reference.

consensus/tendermint/tendermint_test.go
```diff
@@ -83,159 +81,23 @@ func (v *validators) addValidator(addr felt.Felt) {
...
-		assert.Equal(t, expectedPrevoteMsg, prevoteMsg)
-		assert.Contains(t, algo.stateMachine.messages.prevotes[expectedHeight][expectedRound], *nodeAddr)
-		assert.Equal(t, expectedPrevoteMsg, algo.stateMachine.messages.prevotes[expectedHeight][expectedRound][*nodeAddr])
-
-		assert.Equal(t, prevote, algo.stateMachine.state.step)
-		assert.Equal(t, expectedHeight, algo.stateMachine.state.height)
-		assert.Equal(t, expectedRound, algo.stateMachine.state.round)
-	})
+	return New(*nodeAddr, app, chain, vals)
 }
```
rianhughes, line 101:
```suggestion
func setupStateMachine(t *testing.T, numValidators, thisNode int) *Tendermint[value, felt.Felt, felt.Felt] {
	t.Helper()
	app, chain, vals := newApp(), newChain(), newVals()

	for i := 0; i< numValidators; i++  {
		vals.addValidator(*getVal(i))
	}

	thisNodeAddr := getVal(thisNode)

	return New(*thisNodeAddr, app, chain, vals)
}
```

consensus/tendermint/round_utils_test.go
```diff
@@ -0,0 +1,167 @@
...
+	"github.com/NethermindEth/juno/core/felt"
+	"github.com/NethermindEth/juno/utils"
+	"github.com/stretchr/testify/assert"
+)
+
+type testRoundUtils struct {
+	testing      *testing.T
+	stateMachine *Tendermint[value, felt.Felt, felt.Felt]
+	height       height
+	round        round
```
rianhughes, line 15: Can we add a note about what height and round represent here? eg the "true"/"global" height and round of the entire network?

consensus/tendermint/start_test.go
```diff
@@ -0,0 +1,32 @@
...
+
+		currentRound := getTestRound(t, stateMachine, 0, 0)
+
+		val := value(1)
+		currentRound.start().expectActions(
+			currentRound.assertBroadcastProposal(val, -1),
+			currentRound.assertBroadcastPrevote(utils.HeapPtr(val)),
+		)
+
+		assertState(t, stateMachine, 0, 0, prevote)
```
rianhughes, line 21:
```suggestion
		expectedHeight,expectedRound,expectedStep:=height(0),round(0),prevote
		assertState(t, stateMachine, expectedHeight, expectedRound, expectedStep)
```

consensus/tendermint/round_utils_test.go
```diff
@@ -0,0 +1,167 @@
...
+
+	return actionAsserter[Precommit[felt.Felt, felt.Felt]]{
+		testing:      t.testing,
+		inputMessage: precommit,
+		actions:      actions,
+	}
+}
+
+func (t testRoundUtils) processTimeout(s step) actionAsserter[any] {
+	return actionAsserter[any]{
```
rianhughes, line 113: nit: all the methods should be grouped together

consensus/tendermint/round_utils_test.go
```diff
@@ -0,0 +1,167 @@
...
+	actions      []Action[value, felt.Felt, felt.Felt]
+}
+
+func (a actionAsserter[T]) expectActions(expected ...Action[value, felt.Felt, felt.Felt]) actionAsserter[T] {
+	a.testing.Helper()
+	assert.ElementsMatch(a.testing, expected, a.actions)
+	return a
+}
+
+func getTestRound(t *testing.T, stateMachine *Tendermint[value, felt.Felt, felt.Felt], h height, r round) testRoundUtils {
```
rianhughes, line 36: nit:
```suggestion
func newTestRound(t *testing.T, stateMachine *Tendermint[value, felt.Felt, felt.Felt], h height, r round) testRoundUtils {
```

consensus/tendermint/round_utils_test.go
```diff
@@ -0,0 +1,167 @@
...
+	testing      *testing.T
+	stateMachine *Tendermint[value, felt.Felt, felt.Felt]
+	header       MessageHeader[felt.Felt]
+}
+
+type actionAsserter[T any] struct {
+	testing      *testing.T
+	inputMessage T
+	actions      []Action[value, felt.Felt, felt.Felt]
+}
```
rianhughes, line 28: I'm finding this testing logic a bit difficult to understand tbh. Would you mind writing a description / pseudo code highlight what these types are (eg testMessageHeader, actionAsserter, testRoundUtils, etc), and how they are to be used?

rianhughes: I'm finding this testing logic a bit difficult to understand tbh. Would you mind writing a description / pseudo code highlight what these types are (eg testMessageHeader, actionAsserter, testRoundUtils, etc), and how they are to be used?

rianhughes: For example, we 1. define the (proposer/) node, the set of validators, 2. the height and round of the network (testRoundUtils), 3. then we trigger `t.stateMachine.processStart()` via testRoundUtils.start(), which 4. generates a set of actions. And finally we 5. ensure that the generated actions match what we expect by calling `expectActions()`

consensus/tendermint/start_test.go
```diff
@@ -0,0 +1,32 @@
...
+
+func TestStartRound(t *testing.T) {
+	t.Run("node is the proposer", func(t *testing.T) {
+		stateMachine := setupStateMachine(t, 0)
+
+		currentRound := getTestRound(t, stateMachine, 0, 0)
+
+		val := value(1)
+		currentRound.start().expectActions(
+			currentRound.assertBroadcastProposal(val, -1),
```
rianhughes, line 17: What do you think about splitting this into something like:
```
geneartedActions := currentRound.start()
generatedActions.assertExpected(
    currentRound.assertBroadcastProposal(val, -1),
	currentRound.assertBroadcastPrevote(utils.HeapPtr(val)),
)
```

rianhughes: The current method chaining approach isn't very idiomatic for Golang, and imo (althouhg I may be biased) the above suggestion is easier to read/understand

infrmtcs: My intention is to make it like a normal sentence. For example,
> In current round, validator 1 propose committed value with valid round -1. The receiver expects to send a prevote for that value in current round.
Translates to
```
currentRound.validator(1).proposal(committedValue, -1).expectActions(currentRound.assertBroadcastPrevote(&committedValue))
```

infrmtcs: My original version was similar to what you proposed, but I felt that there was too many noise/implementation detail due to the boilerplate, which makes it harder to see the scenario as a whole, especially for cases with long steps (15-20 messages)

consensus/tendermint/tendermint_test.go
rianhughes: Github formatted this strangely, the suggestion was this

```
func setupStateMachine(t *testing.T, numValidators, thisNode int) *Tendermint[value, felt.Felt, felt.Felt] {
	t.Helper()
	app, chain, vals := newApp(), newChain(), newVals()
	for i := 0; i< numValidators; i++  {
		vals.addValidator(*getVal(i))
	}
	thisNodeAddr := getVal(thisNode)
	return New(*thisNodeAddr, app, chain, vals)

```

consensus/tendermint/round_utils_test.go
```diff
@@ -0,0 +1,167 @@
...
+	"github.com/NethermindEth/juno/utils"
+	"github.com/stretchr/testify/assert"
+)
+
+type testRoundUtils struct {
+	testing      *testing.T
+	stateMachine *Tendermint[value, felt.Felt, felt.Felt]
+	height       height
+	round        round
+}
```
rianhughes, line 16: What do you think about renaming `testRoundUtils` to
```
type testNetworkState struct {
    testing      *testing.T
    stateMachine *Tendermint[value, felt.Felt, felt.Felt] // this nodes state
    height       height // network height
    round        round // network round
}
```
Imo the current naming convention isn't clear, as the round could refer to round as seen by any of the nodes, or the quorum round.

rianhughes: Then we the test could look something like this
```
t.Run("node is the proposer", func(t *testing.T) {
		stateMachine := setupStateMachine(t, 0)
		stateMachineContext := newStateMachineContext(t, stateMachine, 0, 0)

		val := value(1)
		stateMachineContext.start().expectActions(
			stateMachineContext.assertBroadcastProposal(val, -1),
			stateMachineContext.assertBroadcastPrevote(utils.HeapPtr(val)),
		)
		assertState(t, stateMachine, 0, 0, prevote)
	})
	```

File: consensus/tendermint/round_utils_test.go [hunk 14]
```diff
@@ -0,0 +1,167 @@
...
+
+func (t testRoundUtils) processTimeout(s step) actionAsserter[any] {
+	return actionAsserter[any]{
+		testing:      t.testing,
+		inputMessage: nil,
+		actions:      t.stateMachine.processTimeout(timeout{s: s, h: t.height, r: t.round}),
+	}
+}
+
+func (t testRoundUtils) start() actionAsserter[any] {
```
rianhughes, line 120:
```suggestion

// start triggers the start of a new round in the state machine using t.round,
// and returns an actionAsserter containing the resulting actions for testing.
func (t testRoundUtils) start() actionAsserter[any] {
```

consensus/tendermint/start_test.go
```diff
@@ -0,0 +1,32 @@
...
+		assertState(t, stateMachine, 0, 0, prevote)
+	})
+
+	t.Run("node is not the proposer: schedule timeoutPropose", func(t *testing.T) {
+		stateMachine := setupStateMachine(t, 3)
+		currentRound := getTestRound(t, stateMachine, 0, 0)
+
+		currentRound.start().expectActions(currentRound.scheduleTimeoutAction(propose))
+
+		assertState(t, stateMachine, 0, 0, propose)
```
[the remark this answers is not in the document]
infrmtcs: How about something like:
```
assertState(t, stateMachine).height(0).round(0).step(prevote)
```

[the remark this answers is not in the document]
infrmtcs: Or another option is to explicitly:
```suggestion
		assertState(t, stateMachine, height(0), round(0), propose)
```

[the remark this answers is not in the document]
infrmtcs: I'm trying with my second approach, let me know if you think it make sense.

consensus/tendermint/incoming_message_builder_test.go
```diff
@@ -0,0 +1,70 @@
...
+// incomingMessageBuilder is a helper struct to build incoming messages received from other nodes for the state machine.
+// Each method builds a message, processes it, assert if it's stored in the state machine, and returns an actionAsserter to assert the result actions.
+type incomingMessageBuilder struct {
+	testing      *testing.T
+	stateMachine *Tendermint[value, felt.Felt, felt.Felt]
+	header       MessageHeader[felt.Felt]
+}
+
+func (t incomingMessageBuilder) proposal(val value, validRound round) actionAsserter[Proposal[value, felt.Felt, felt.Felt]] {
+	t.testing.Helper()
```
rianhughes, line 18:
```suggestion
// proposal builds and processes a Proposal message, asserts it's stored in the state machine,
// and returns an actionAsserter to check resulting actions.
func (t incomingMessageBuilder) proposal(val value, validRound round) actionAsserter[Proposal[value, felt.Felt, felt.Felt]] {
	t.testing.Helper()
```

consensus/tendermint/incoming_message_builder_test.go
```diff
@@ -0,0 +1,70 @@
...
+
+	return actionAsserter[Proposal[value, felt.Felt, felt.Felt]]{
+		testing:      t.testing,
+		stateMachine: t.stateMachine,
+		inputMessage: proposal,
+		actions:      actions,
+	}
+}
+
+func (t incomingMessageBuilder) prevote(val *value) actionAsserter[Prevote[felt.Felt, felt.Felt]] {
```
rianhughes, line 36:
```suggestion
// prevote builds and processes a Prevote message, asserts it's stored in the state machine,
// and returns an actionAsserter to check resulting actions.
func (t incomingMessageBuilder) prevote(val *value) actionAsserter[Prevote[felt.Felt, felt.Felt]] {
```

consensus/tendermint/incoming_message_builder_test.go
```diff
@@ -0,0 +1,70 @@
...
+
+	return actionAsserter[Prevote[felt.Felt, felt.Felt]]{
+		testing:      t.testing,
+		stateMachine: t.stateMachine,
+		inputMessage: prevote,
+		actions:      actions,
+	}
+}
+
+func (t incomingMessageBuilder) precommit(val *value) actionAsserter[Precommit[felt.Felt, felt.Felt]] {
```
rianhughes, line 54:
```suggestion
// precommit builds and processes a Precommit message, asserts it's stored in the state machine,
// and returns an actionAsserter to check resulting actions.
func (t incomingMessageBuilder) precommit(val *value) actionAsserter[Precommit[felt.Felt, felt.Felt]] {
```

consensus/tendermint/assert_utils_test.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+// assertPrevote asserts that the prevote message is in the state machine, except when the state machine advanced to the next height.
+func assertPrevote(t *testing.T, stateMachine *Tendermint[value, felt.Felt, felt.Felt], expectedMsg Prevote[felt.Felt, felt.Felt]) {
+	t.Helper()
+	// New height will discard the previous height messages.
+	if stateMachine.state.height != expectedMsg.Height {
+		return
+	}
+	assertMessage(t, stateMachine.messages.prevotes, expectedMsg.MessageHeader, expectedMsg)
```
rianhughes, line 43:
```suggestion
// assertPrecommit asserts teh Precomit was stored in the state machine
func assertPrecommit(t *testing.T, stateMachine *Tendermint[value, felt.Felt, felt.Felt], expectedMsg Precommit[felt.Felt, felt.Felt]) {
```

consensus/tendermint/assert_utils_test.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+func assertMessage[T Message[value, felt.Felt, felt.Felt]](t *testing.T, messages map[height]map[round]map[felt.Felt]T, expectedMsgHeader MessageHeader[felt.Felt], expectedMsg T) {
+	t.Helper()
+	assert.Contains(t, messages, expectedMsgHeader.Height, "height not found")
+	assert.Contains(t, messages[expectedMsgHeader.Height], expectedMsgHeader.Round, "round not found")
+	assert.Contains(t, messages[expectedMsgHeader.Height][expectedMsgHeader.Round], expectedMsgHeader.Sender, "sender not found")
+	assert.Equal(t, expectedMsg, messages[expectedMsgHeader.Height][expectedMsgHeader.Round][expectedMsgHeader.Sender], "message not equal")
+}
+
```
rianhughes, line 25:
```suggestion
// assertProposal asserts the proposal was stored in the state machine
func assertProposal(t *testing.T, stateMachine *Tendermint[value, felt.Felt, felt.Felt], expectedMsg Proposal[value, felt.Felt, felt.Felt]) {
```

consensus/tendermint/round_utils_test.go
infrmtcs: Added doc string with sample usages.

consensus/tendermint/round_utils_test.go
infrmtcs: Note: Renamed to `stateMachineContext`, however I'll keep `currentRound` at the call site, to differentiate between incoming messages with different height/round.

consensus/tendermint/rule_skip_round_test.go
```diff
@@ -0,0 +1,51 @@
...
+		futureRound := newTestRound(t, stateMachine, expectedHeight, rPrime)
+
+		stateMachine.processStart(0)
+		futureRound.validator(1).prevote(&rPrimeVal)
+		futureRound.validator(0).proposal(rPrimeVal, -1)
+
+		// The step is not propose because the proposal which is received in round r' leads to consensus
+		// engine broadcasting prevote to the proposal which changes the step from propose to prevote.
+		assertState(t, stateMachine, expectedHeight, rPrime, prevote)
+	})
```
rianhughes, line 19: The stateMachine starts out with height 0 and round 0. It recieves a proposal for round 4. And a prevote for that proposal at round 4. And this test asserts that we expect the state machine to jump to round 4 and progress the step to prevote.

I just need to double check that we expect the round to jump ahead like this. I would have expected that the state machine would need to process messages between 0 and 4 in order to jump ahead.

rianhughes: Yup, we can jump ahead because we received `f+1` (2 here) messags from `rPrime`

```
55: upon f + 1 〈∗, hp , round, ∗, ∗〉 with round > roundp do
56: StartRound(round)
```

rianhughes: So we jump ahead to round 4 because of line 55. Then we process the proposal to send a prevote, and update the step to `prevote`. At this point we don't haven't received a quorum of prevotes, so can't progress futher. And the timeouts aren't triggered, so we don't progress to a new round.

consensus/tendermint/rule_skip_round_test.go
```diff
@@ -0,0 +1,51 @@
+package tendermint
+
+import "testing"
+
+func TestSkipRound(t *testing.T) {
```
rianhughes, line 5: Just a suggestion to further improve readability

rianhughes, line 5: Just a suggestion to further improve readability
```suggestion
func TestSkipRound(t *testing.T) {
	thisNode := 3
	peer1:=0
	peer2:=1
	
```

consensus/tendermint/rule_skip_round_test.go
```diff
@@ -0,0 +1,51 @@
...
+	})
+
+	t.Run("Line 55 (Precommit): Start round r' when f+1 future round messages are received from round r'", func(t *testing.T) {
+		stateMachine := setupStateMachine(t, 4, 3)
+		expectedHeight := height(0)
+		rPrime := round(4)
+		round4Value := value(10)
+		futureRound := newTestRound(t, stateMachine, expectedHeight, rPrime)
+
+		stateMachine.processStart(0)
```
rianhughes, line 43: Also, just another suggestion 
```suggestion
		stateMachine.processStart(round(0))
```

consensus/tendermint/rule_skip_round_test.go
rianhughes:
```
func TestSkipRound(t *testing.T) {
	thisNode := 3
	peer1:=0
	peer2:=1

	expectedHeight := height(0)
	rPrime, rPrimeVal := round(4), value(10)
	round4Value := value(10)

	t.Run("Line 55 (Proposal): Start round r' when f+1 future round messages are received from round r'", func(t *testing.T) {
		stateMachine := setupStateMachine(t, 4, thisNode)
	
		futureRound := newTestRound(t, stateMachine, expectedHeight, rPrime)

		stateMachine.processStart(round(0))
		futureRound.validator(peer1).prevote(&rPrimeVal)
		futureRound.validator(peer2).proposal(rPrimeVal, -1)

		// The step is not propose because the proposal which is received in round r' leads to consensus
		// engine broadcasting prevote to the proposal which changes the step from propose to prevote.
		assertState(t, stateMachine, expectedHeight, rPrime, prevote)
	})
```

infrmtcs: I think `thisNode` makes sense.
Use `peer1` for validator 0 may cause confusion, while `peer0` is redundant.

infrmtcs: By confusion, I meant switching between 0-based and 1-based index

infrmtcs: How about
```

  futureRound.peer(0).prevote(&rPrimeVal)
        futureRound.peer(1).proposal(rPrimeVal, -1

```
Another approach is to replace 0-based index by 1-based index

consensus/tendermint/rule_commit_value_test.go
```diff
@@ -0,0 +1,84 @@
...
+
+		currentRound.start().expectActions(
+			currentRound.action().scheduleTimeout(propose),
+		)
+		val0Precommit := currentRound.validator(0).precommit(&committedValue).inputMessage
+		val1Precommit := currentRound.validator(1).precommit(&committedValue).inputMessage
+		val2Precommit := currentRound.validator(2).precommit(&committedValue).expectActions(
+			currentRound.action().scheduleTimeout(precommit),
+		).inputMessage
+		assert.True(t, stateMachine.state.timeoutPrecommitScheduled)
```
rianhughes, line 27: Just commenting to clarify my understanding, at this point we trigger
```
47: upon 2f + 1 〈PRECOMMIT, hp , roundp, ∗〉 for the first time do
48: schedule OnT imeoutP recommit(hp , roundp) to be executed after timeoutP recommit(roundp )
```

rianhughes: After receiving the proposal we trigger line 22 (cast Prevote), then line 49, and finally line 11 (scheduleTimeout).

I think there is an assumption here that we aren't the proposer on the newRound

consensus/tendermint/rule_commit_value_test.go
```diff
@@ -0,0 +1,84 @@
...
+		currentRound := newTestRound(t, stateMachine, 0, 0)
+		nextRound := newTestRound(t, stateMachine, 1, 0)
+		committedValue := value(10)
+
+		currentRound.start().expectActions(
+			currentRound.action().scheduleTimeout(propose),
+		)
+
+		val0Precommit := currentRound.validator(0).precommit(&committedValue).inputMessage
+		currentRound.validator(0).proposal(committedValue, -1)
```
rianhughes, line 62: Do we not expect to broadcast a Prevote here? Why don't we check the resulting action?

consensus/tendermint/rule_first_proposal_test.go
```diff
@@ -0,0 +1,124 @@
...
+		expectedValue := value(42)
+
+		// Simulate the previous round such that it times out in the precommit step
+		// Validator 0 is a faulty proposer, proposing an invalid value to validator 1
+		previousRound.start()
+		previousRound.validator(0).proposal(expectedValue, -1)
+		previousRound.validator(0).prevote(utils.HeapPtr(expectedValue))
+		previousRound.validator(1).prevote(nil)
+		previousRound.validator(2).prevote(utils.HeapPtr(expectedValue))
+		assertState(t, stateMachine, height(0), round(0), precommit)
```
rianhughes, line 40: Should we not also scheudle a timeout?
```
34: upon 2f + 1 〈PREVOTE, hp , roundp, ∗〉 while stepp = prevote for the first time do
35: schedule OnT imeoutP revote(hp , roundp) to be executed after timeoutP revote(roundp )
```

rianhughes: We also broadcast a precommit
```
44: upon 2f + 1 〈PREVOTE, hp , roundp, nil〉 while stepp = prevote do
45: broadcast 〈PRECOMMIT, hp , roundp, nil〉
46: stepp ← precommit
```

rianhughes: Also, we progress to the `precommut` step because of this
```
61: Function OnT imeoutP revote(height, round) :
62: if height = hp ∧ round = roundp ∧ stepp = prevote then
63: broadcast 〈PRECOMMIT, hp , roundp, nil〉
64: stepp ← precommit
```
and not this
```
44: upon 2f + 1 〈PREVOTE, hp , roundp, nil〉 while stepp = prevote do
45: broadcast 〈PRECOMMIT, hp , roundp, nil〉
46: stepp ← precommit
```
right?

rianhughes: I don't think we should enter line 36 because we never received a quorum of Prevotes for the same value

```
36: upon 〈PROPOSAL, hp , roundp, v, ∗〉 from proposer(hp , roundp) AND 2f + 1 〈PREVOTE, hp, roundp , id(v)〉 while
valid(v) ∧ stepp ≥ prevote for the first time do
37: if stepp = prevote then
38: lockedV aluep ← v
39: lockedRoundp ← roundp
40: broadcast 〈PRECOMMIT, hp , roundp, id(v))〉
41: stepp ← precommit
42: validV aluep ← v
43: validRoundp ← roundp
```

consensus/tendermint/action_asserter_test.go
```diff
@@ -0,0 +1,52 @@
...
+			// If the post state is not in new height and a value is committed, assert the valid and locked tuples.
+			if a.stateMachine.state.height == action.Height && action.ID != nil {
+				assert.Equal(a.testing, a.stateMachine.state.lockedRound, action.Round)
+				assert.Equal(a.testing, getHash(a.stateMachine.state.lockedValue), action.ID)
+				assert.Equal(a.testing, a.stateMachine.state.validRound, action.Round)
+				assert.Equal(a.testing, getHash(a.stateMachine.state.validValue), action.ID)
+			}
+			assertPrecommit(a.testing, a.stateMachine, Precommit[felt.Felt, felt.Felt](*action))
+		case *ScheduleTimeout:
+		}
```
rianhughes, line 42: Should we not check timeouts as well?

consensus/tendermint/rule_commit_value_test.go
infrmtcs: It's correct that the node will broadcast prevote there, but the objective of this test is to test the `commit_value` rule, so I don't check the action here.

consensus/tendermint/action_asserter_test.go
infrmtcs: I should have added a comment there. The action to schedule timeout doesn't change the state, but only the driver, so there is nothing to assert here. Unlike the broadcast message action, which always stores the broadcasted message to `messages` in state machine.

consensus/tendermint/rule_first_proposal_test.go
infrmtcs: > Should we not also scheudle a timeout?

Same as a comment above, the objective of this test is not to check the timeout, so sometimes I may skip asserting the expected actions.

infrmtcs: > Also, we progress to the precommut step because of this
> I don't think we should enter line 36 because we never received a quorum of Prevotes for the same value

There is a quorum because validator 0, 2 and 3 (current node) cast prevote. Maybe I can make it a bit easier to understand by adding expectActions prevote for the proposal line, but the logic is correct.

consensus/tendermint/rule_polka_nil_test.go
```diff
@@ -0,0 +1,104 @@
...
+		// Receive a proposal and move to prevote step
+		currentRound.validator(0).proposal(value(42), -1)
+
+		// Receive 2 prevotes
+		currentRound.validator(0).prevote(nil)
+		currentRound.validator(1).prevote(nil)
+
+		// This validator votes for a value instead of nil
+		val := value(42)
+		currentRound.validator(2).prevote(&val)
```
rianhughes, line 99: This should also trigger a timeout right?
```
34: upon 2f + 1 〈PREVOTE, hp , roundp, ∗〉 while stepp = prevote for the first time do
35: schedule OnT imeoutP revote(hp , roundp) to be executed after timeoutP revote(roundp 
```

consensus/tendermint/timeout_test.go
```diff
@@ -0,0 +1,93 @@
...
+	t.Run("OnTimeoutPrecommit: move to next round", func(t *testing.T) {
+		stateMachine := setupStateMachine(t, 4, 3)
+
+		currentRound := newTestRound(t, stateMachine, 0, 0)
+		nextRound := newTestRound(t, stateMachine, 0, 1)
+
+		currentRound.start()
+		currentRound.validator(0).precommit(utils.HeapPtr(value(10)))
+		currentRound.validator(1).precommit(nil)
+		currentRound.validator(2).precommit(nil).expectActions(currentRound.action().scheduleTimeout(precommit))
```
rianhughes, line 46: I understand that the current tests focus on the timeouts specifically, and that we ignore messages here (but test elsewhere). But on first glance it looks like the tests are incomplete. I think if we are going to go with the current approach we should have a readme explaining the current strucutre of the tests

consensus/tendermint/rule_precommit_any_test.go
```diff
@@ -0,0 +1,91 @@
...
+		currentRound := newTestRound(t, stateMachine, 0, 0)
+
+		// Initialise state
+		currentRound.start()
+
+		// Set up to reach precommit step
+		// Validator 0 is a faulty proposer
+		currentRound.validator(0).proposal(value(42), -1)
+		currentRound.validator(0).prevote(utils.HeapPtr(value(42)))
+		currentRound.validator(1).prevote(utils.HeapPtr(value(42)))
```
rianhughes, line 79: The proposer should never send a prevote, so in principle the Driver layer should block this case from happening right?

rianhughes: We also shouldn't trigger any actions at this point (namely line 34 or line 44), so it should be as if we never received the prevote from the proposer?

infrmtcs: > The proposer should never send a prevote, so in principle the Driver layer should block this case from happening right?

My current understanding is that proposer will also send prevote for its own proposal. Otherwise, 2f + 1 prevote quorum will never reach if the f faulty nodes stay silent.

rianhughes: Ah, actually I'm wrong, the propser does send out prevotes and precommits

rianhughes: Can we add an assertion for the resulting actions here? As far as I understand, we should schedule the timeout, broadcast our vote, update the locks, and update the step to `precommit`

consensus/tendermint/rule_polka_nil_test.go
infrmtcs: No, it shouldn't because 3 prevotes were gathered from the previous 2 prevotes + the node's own prevote.
That said, it's not the objective of the test, so I also don't assert there.

consensus/tendermint/rule_skip_round_test.go
rianhughes: > I think thisNode makes sense.
> Use peer1 for validator 0 may cause confusion, while peer0 is redundant.

What about just using `peer0` and `peer1`?

rianhughes: It just seems like an easier update than renaming validator to peer. Plus I think validator is a better name. I don't feel strongly either way, so I'll let you make the decision :)

consensus/tendermint/rule_polka_nil_test.go
rianhughes: Are you sure we don't enter line 34 at some point, eg after `currentRound.validator(1).prevote(nil)`? I agree that given how we are currently structuring tests that there is no need to assert here, just asking to confirm I don't misunderstand anything 

```
// Receive a proposal and move to prevote step
		currentRound.validator(0).proposal(value(42), -1)

		// Receive 2 prevotes
		currentRound.validator(0).prevote(nil)
		currentRound.validator(1).prevote(nil)

		// This validator votes for a value instead of nil
		val := value(42)
		currentRound.validator(2).prevote(&val)
```

infrmtcs: Yeah, you're correct, that line is the one that schedules the timeout, not the original commented line. I don't assert the timeout there just because the test is not targeting timeout.

infrmtcs: To clarify, it's true that that this scenario will eventually move to `precommit` step when prevote timeout triggers, but the objective of this test is to ensure that not enough 2f + 1 nil precommits which results in no immediate state change and action.

consensus/tendermint/rule_proposal_and_polka_current_test.go
```diff
@@ -0,0 +1,98 @@
...
+		currentRound.validator(3).prevote(&committedValue)
+
+		currentRound.processTimeout(prevote).expectActions(
+			currentRound.action().broadcastPrecommit(nil),
+		)
+
+		// Receives proposal and prevote from validator 0, record valid value and round
+		currentRound.validator(0).proposal(committedValue, -1)
+		currentRound.validator(0).prevote(&committedValue)
+		assertState(t, stateMachine, height(0), round(0), precommit)
```
rianhughes, line 57: I think something that would be very helpful when reading tests, is adding a small comment indicating when and which "line" we enter

rianhughes, line 57: I think something that would be very helpful when reading tests, is adding a small comment indicating when and which "line" we enter

```suggestion
		currentRound.validator(3).prevote(&committedValue) // line34: schedule timeout

		currentRound.processTimeout(prevote).expectActions(
			currentRound.action().broadcastPrecommit(nil),
		)

		// Receives proposal and prevote from validator 0, record valid value and round
		currentRound.validator(0).proposal(committedValue, -1)
		currentRound.validator(0).prevote(&committedValue) // line 36
		assertState(t, stateMachine, height(0), round(0), precommit)
```

consensus/tendermint/rule_proposal_and_polka_current_test.go
```diff
@@ -0,0 +1,98 @@
...
+		currentRound.validator(2).prevote(&committedValue)
+		currentRound.validator(3).prevote(&committedValue)
+
+		currentRound.processTimeout(prevote).expectActions(
+			currentRound.action().broadcastPrecommit(nil),
+		)
+
+		// Receives proposal and prevote from validator 0, record valid value and round
+		currentRound.validator(0).proposal(committedValue, -1)
+		currentRound.validator(0).prevote(&committedValue)
```
rianhughes, line 56:
```suggestion
		currentRound.validator(0).prevote(&committedValue) // line 36: dont broadcast commit.
```

consensus/tendermint/rule_proposal_and_polka_current_test.go
```diff
@@ -0,0 +1,98 @@
...
+
+		// Initialise the round
+		currentRound.start()
+
+		// Prevotes are collected but not proposal
+		currentRound.processTimeout(propose).expectActions(
+			currentRound.action().broadcastPrevote(nil),
+		)
+		currentRound.validator(2).prevote(&committedValue)
+		currentRound.validator(3).prevote(&committedValue)
```
rianhughes, line 48:
```suggestion
		currentRound.validator(3).prevote(&committedValue) // line34: schedule timeout
```

consensus/tendermint/rule_proposal_and_polka_current_test.go
```diff
@@ -0,0 +1,98 @@
...
+
+		// Use valid value on next round
+		currentRound.validator(2).precommit(&committedValue)
+		currentRound.validator(3).precommit(&committedValue).expectActions(
+			currentRound.action().scheduleTimeout(precommit),
+		)
+		currentRound.processTimeout(precommit).expectActions(
+			nextRound.action().broadcastProposal(committedValue, 0),
+			nextRound.action().broadcastPrevote(&committedValue),
+		)
```
rianhughes, line 69:
```suggestion
		assertState(t, stateMachine, height(0), round(1), prevote)
```

consensus/tendermint/rule_proposal_and_polka_current_test.go
```diff
@@ -0,0 +1,98 @@
...
+
+		committedValue := value(42)
+
+		// Initialise the round
+		currentRound.start()
+
+		// 2 prevotes are collected
+		currentRound.validator(0).proposal(committedValue, -1).expectActions(
+			currentRound.action().broadcastPrevote(&committedValue),
+		)
```
rianhughes, line 84: Do you think adding additional state assertions is too verbose?

```suggestion
		currentRound.validator(0).proposal(committedValue, -1).expectActions( // line 22
			currentRound.action().broadcastPrevote(&committedValue),
		)
		assertState(t, stateMachine, height(0), round(0), prevote)
```

consensus/tendermint/rule_proposal_and_polka_current_test.go
```diff
@@ -0,0 +1,98 @@
...
+		currentRound := newTestRound(t, stateMachine, 0, 0)
+
+		committedValue := value(42)
+
+		// Initialise the round
+		currentRound.start()
+
+		// 2 prevotes are collected
+		currentRound.validator(0).proposal(committedValue, -1).expectActions(
+			currentRound.action().broadcastPrevote(&committedValue),
```
rianhughes, line 83: What do you think about adding additional state assertions? Too Verbose? I feel like it makes it easier to check we are performing all the actions required by the tendermint algorithm, but it does add more code..

```suggestion
		currentRound.validator(0).proposal(committedValue, -1).expectActions( // line 22
			currentRound.action().broadcastPrevote(&committedValue),
		)
		assertState(t, stateMachine, height(0), round(0), prevote)
```

consensus/tendermint/rule_proposal_and_polka_previous_test.go
```diff
@@ -0,0 +1,97 @@
...
+		// In this test, we're validator 1. Validator 0 is faulty.
+		stateMachine := setupStateMachine(t, 4, 3)
+
+		firstRound := newTestRound(t, stateMachine, 0, 0)
+		secondRound := newTestRound(t, stateMachine, 0, 1)
+
+		wrongValue := value(42)
+		correctValue := value(43)
+
+		// In the first round, validator 0 "tricked" us with invalid value.
```
rianhughes, line 20:
```suggestion
		// In the first round, validator 0 sent us a different (valid) value form all the other peers
```

consensus/tendermint/rule_proposal_and_polka_previous_test.go
```diff
@@ -0,0 +1,97 @@
...
+		firstRound.validator(0).proposal(wrongValue, -1).expectActions(
+			firstRound.action().broadcastPrevote(&wrongValue),
+		)
+		firstRound.validator(0).prevote(&correctValue)
+		firstRound.validator(1).prevote(&correctValue)
+		firstRound.validator(2).prevote(&correctValue)
+		firstRound.validator(1).precommit(&correctValue)
+		firstRound.validator(2).precommit(&correctValue)
+		firstRound.processTimeout(precommit)
+
```
rianhughes, line 31: Does this test assume that we only ever register and trigger time outs? And that when the new round starts we requested and processed the valid value (that we never saw) from round(0)?

rianhughes: I guess once we prevote nil, we must precommit nil. Even in the case we see a quorum of prevotes for a valid value.

consensus/tendermint/rule_proposal_and_polka_previous_test.go
```diff
@@ -0,0 +1,97 @@
...
+		firstRound := newTestRound(t, stateMachine, 0, 0)
+		secondRound := newTestRound(t, stateMachine, 0, 1)
+		thirdRound := newTestRound(t, stateMachine, 0, 2)
+
+		wrongValue := value(42)
+		correctValue := value(43)
+
+		// In the first round, validator 0 "tricked" us with invalid value.
+		firstRound.start()
+		firstRound.validator(0).proposal(wrongValue, -1).expectActions(
```
rianhughes, line 52:
```suggestion
		firstRound.validator(0).proposal(wrongValue, -1).expectActions( // line 22
```

consensus/tendermint/rule_proposal_and_polka_previous_test.go
```diff
@@ -0,0 +1,97 @@
...
+			firstRound.action().broadcastPrevote(&wrongValue),
+		)
+		firstRound.validator(0).prevote(&wrongValue)
+		// Validator 2 receives an invalid proposal
+		firstRound.validator(2).prevote(nil)
+		firstRound.validator(1).prevote(&wrongValue).expectActions(
```
rianhughes, line 58:
```suggestion
		firstRound.validator(1).prevote(&wrongValue).expectActions( // line 28
```

consensus/tendermint/rule_proposal_and_polka_previous_test.go
```diff
@@ -0,0 +1,97 @@
...
+			firstRound.action().broadcastPrecommit(&wrongValue),
```
rianhughes, line 59:
```suggestion
		assertState(t, stateMachine, height(0), round(0), prevote)
```

consensus/tendermint/rule_proposal_and_polka_previous_test.go
rianhughes, line 52:
```suggestion
		assertState(t, stateMachine, height(0), round(0), prevote)
```

consensus/tendermint/rule_proposal_and_polka_previous_test.go
```diff
@@ -0,0 +1,97 @@
...
+		correctValue := value(43)
+
+		// In the first round, validator 0 "tricked" us with invalid value.
+		firstRound.start()
+		firstRound.validator(0).proposal(wrongValue, -1).expectActions(
+			firstRound.action().broadcastPrevote(&wrongValue),
+		)
+		firstRound.validator(0).prevote(&wrongValue)
+		// Validator 2 receives an invalid proposal
+		firstRound.validator(2).prevote(nil)
```
rianhughes, line 57: This is line 36 right? And we also lock the value?

consensus/tendermint/rule_proposal_and_polka_previous_test.go
```diff
@@ -0,0 +1,97 @@
...
+		firstRound.validator(1).prevote(&wrongValue).expectActions(
+			firstRound.action().broadcastPrecommit(&wrongValue),
+		)
```
rianhughes, line 60:
```suggestion
		firstRound.validator(1).prevote(&wrongValue).expectActions( // line 36
			firstRound.action().broadcastPrecommit(&wrongValue),
		)
		assert.Equal(t, stateMachine.state.lockedValue, &wrongValue)
		assertState(t, stateMachine, height(0), round(0), precommit)
```

consensus/tendermint/rule_proposal_and_polka_previous_test.go
```diff
@@ -0,0 +1,97 @@
...
+		// Validator 0 sends prevote nil to validator 1, so validator 1 will precommit nil.
+		firstRound.validator(1).precommit(nil)
+		firstRound.validator(2).precommit(nil)
```
rianhughes, line 63: This is faulty behaviour by validator(1) right? Validators should never precommit for a different value form the one they prevoted for (in the same round)

consensus/tendermint/rule_proposal_and_polka_previous_test.go
```diff
@@ -0,0 +1,97 @@
...
+		assertState(t, stateMachine, height(0), round(1), propose)
+
+		secondRound.validator(1).proposal(correctValue, 0).expectActions(
+			secondRound.action().broadcastPrevote(&correctValue),
+		)
+	})
+
+	t.Run("Line 28: locked value change", func(t *testing.T) {
+		// In this test, we're validator 1. Validator 0 is faulty.
+		stateMachine := setupStateMachine(t, 4, 3)
```
rianhughes, line 41: This is incorrect, we're validator 3 right?

consensus/tendermint/rule_proposal_and_polka_previous_test.go
```diff
@@ -0,0 +1,97 @@
...
+
+		firstRound := newTestRound(t, stateMachine, 0, 0)
+		secondRound := newTestRound(t, stateMachine, 0, 1)
+		thirdRound := newTestRound(t, stateMachine, 0, 2)
+
+		wrongValue := value(42)
```
rianhughes, line 47:
```suggestion
		// In the first round, validator 0 sent us a different (valid) value form all the other peers
		firstRound.start()
		firstRound.validator(0).proposal(wrongValue, -1).expectActions( // line 22
			firstRound.action().broadcastPrevote(&wrongValue),
		)
		assertState(t, stateMachine, height(0), round(0), prevote)
```

consensus/tendermint/rule_proposal_and_polka_previous_test.go
rianhughes, line 52: validator(1) should not prevote on the wrong value, right? It should be prevote(&correctValue)??

consensus/tendermint/rule_proposal_and_polka_previous_test.go
```diff
@@ -0,0 +1,97 @@
...
+
+		assertState(t, stateMachine, height(0), round(1), propose)
+
+		secondRound.validator(1).proposal(correctValue, 0).expectActions(
+			secondRound.action().broadcastPrevote(&correctValue),
+		)
+	})
+
+	t.Run("Line 28: locked value change", func(t *testing.T) {
+		// In this test, we're validator 1. Validator 0 is faulty.
```
rianhughes, line 40: I'm finidng this test difficult to follow. Can we add a description at the top, something like this?

```suggestion
	t.Run("Line 28: locked value change", func(t *testing.T) {
		// In this test, we're validator 1. Validator 0 is faulty.
		// Round 1: We lock on the valid value. The other validators do not lock on it,
		//          because they receive a different valid value.
		// Round 2: Validator 1 now proposes the different value from the one we locked on. We reject it.
```

rianhughes (changes requested):
Overall, the PR is really great, very nice work!

The DSL is very concise and effective, although requires getting used to a little bit (to be expected I guess). The tests themselves all seem correct (I didn't find a single error!), and the coverage is great.

My only comments are related to: 
1) making the tests more readable, and 
2) making it easier to understand / get used to the DSL. 

I think we can significantly improve the readability of the tests by 
1) adding a comment whenever we trigger a new rule (eg //line 22, etc), and 
2) including the state assertions whenever we update the state. I agree that it's not always the purpose of the test, but it's a simple addition that really makes it easy to track the changes in state throughout the test, which imo is particularly important when the tests get more complicated.

As for the DSL, I think adding a small doc/readme or something could work. And also highlight that for a given test, we don't assert all the changes for brevity. But that these assertions are make in other tests (eg we don't assert that we register timeouts in all the tests, because we have a specific set of tests for that, namely `timeout_tests.go`)

Oh, and thanks for putting up with all of my nit-picky comments!!

consensus/tendermint/rule_proposal_and_polka_current_test.go
infrmtcs: Agree, for line 34 comment, I think I'll add an expectActions there, so it can be clearer.

consensus/tendermint/rule_proposal_and_polka_current_test.go
infrmtcs: Agree, but I think this is duplicate with the comment above

consensus/tendermint/rule_proposal_and_polka_previous_test.go
infrmtcs: In this particular test, I decided to rename `wrongValue` and `correctValue` to `firstValue` and `secondValue`, because both are correct, but we have 2 different values just because of disrupt from validator 0.

consensus/tendermint/rule_proposal_and_polka_previous_test.go
infrmtcs: Validator 0 send prevote nil to validator 1, so validator 1 doesn't have enough quorum to precommit the prevoted value, so timeout happens and it has to precommit nil.

consensus/tendermint/rule_skip_round_test.go
infrmtcs: I'd keep `validator(0)` for now, because I think it's more concise than `validator(peer0)`.

consensus/tendermint/rule_proposal_and_polka_previous_test.go
rianhughes: Ah right, it hit its timeout. Makes sense, thanks

rianhughes (approved):

PR #2785, opened 2025-04-25
refactor: Make some necessary types and functions public

infrmtcs:
Although the number of changes is big, the changes are minor:
- Move the `Driver` struct and its functions to a separate `driver` package. The tests are added in the next PR.
- Rename `Tendermint` struct to `stateMachine` and add `StateMachine` interface to list the exported functions from `stateMachine`.
- Rename `timeout`, `step`, `height`, `round` and `votingPower` to `Timeout`, `Step`, `Height`, `Round` and `VotingPower` to be able to use in `driver` package.
- Rename `propose`, `prevote` and `precommit` steps to `StepPropose`, `StepPrevote` and `StepPrecommit`. The `Step` prefix is to differentiate with the message types.

### DIFF
--- a/consensus/driver/driver.go
+++ b/consensus/driver/driver.go
@@ -1,0 +1,127 @@
+package driver
+
+import (
+	"sync"
+	"time"
+
+	"github.com/NethermindEth/juno/consensus/tendermint"
+)
+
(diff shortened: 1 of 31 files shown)


infrmtcs:
Close in favor of #2783 because code coverage will complain when we separate the code without unit tests (which is covered by #2783).

PR #2786, opened 2025-04-25
test: Unit tests for Tendermint driver

infrmtcs:
This PR consists of 2 commits:

### First commit
Although the number of changes is big, the changes are minor:

- Move the Driver struct and its functions to a separate driver package. The tests are added in the next PR.
- Rename Tendermint struct to stateMachine and add StateMachine interface to list the exported functions from stateMachine.
- Rename timeout, step, height, round and votingPower to Timeout, Step, Height, Round and VotingPower to be able to use in driver package.
- Rename propose, prevote and precommit steps to StepPropose, StepPrevote and StepPrecommit. The Step prefix is to differentiate with the message types.

### Second commit:
This add unit tests for `driver` package, using a mock `stateMachine`. The reason why mock is used here is:
- The behaviour of every rule in the core algorithm is covered by the tests in `tendermint` package.
- The expected consensus that a cluster of nodes should reach will be covered by integ tests, in PR #2790.
- The tests in this PR mainly to ensure that the broadcasted message returned by actions will be sent, the timeout will be scheduled and the driver will forward all received messages to the state machine.


consensus/driver/driver_test.go
```diff
@@ -0,0 +1,223 @@
...
+}
+
+func (m *mockBroadcaster[M]) Broadcast(msg M) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	m.broadcastedMessages = append(m.broadcastedMessages, msg)
+	m.wg.Done()
+}
+
+func (m *mockBroadcaster[M]) SendMsg(addr felt.Felt, msg M) {
```
rodrigo-pino, line 57: Empty implementation intentional?

Also, receive address by reference, although since it is a test is not that important. The only thing that worries me in this scenario is making the code inconsistent by at copies being by reference and in other instances as by value.

consensus/driver/driver_test.go
```diff
@@ -0,0 +1,223 @@
...
+}
+
+const (
+	actionCount = 10
+)
+
+func registerActions(expectedBroadcast *expectedBroadcast) []tendermint.Action[value, felt.Felt, felt.Felt] {
+	actions := make([]tendermint.Action[value, felt.Felt, felt.Felt], actionCount)
+	for i := range actionCount {
+		switch rand.Int() % 3 {
```
rodrigo-pino, line 129: Should we use a seed for deterministic randomization?

rodrigo-pino (approved):

consensus/driver/driver_test.go
infrmtcs: Agree, I'll update the code

consensus/driver/driver_test.go
infrmtcs: I'm wondering if we can remove this `SendMsg`, because the algorithm only needs to broadcast message, while direct communication between peers is handled inside the libp2p library.

consensus/driver/driver_test.go
```diff
@@ -0,0 +1,223 @@
...
+		Round:  tendermint.Round(rand.Int()),
+	}
+}
+
+const (
+	actionCount = 10
+)
+
+func registerActions(expectedBroadcast *expectedBroadcast) []tendermint.Action[value, felt.Felt, felt.Felt] {
+	actions := make([]tendermint.Action[value, felt.Felt, felt.Felt], actionCount)
```
rianhughes, line 127:
```suggestion
// registerActions generates a fixed number of randomized Tendermint actions.
// Each action is a broadcast proposal, prevote, or precommit, and is also
// recorded in the corresponding field of expectedBroadcast.
func registerActions(expectedBroadcast *expectedBroadcast) []tendermint.Action[value, felt.Felt, felt.Felt] {
	actions := make([]tendermint.Action[value, felt.Felt, felt.Felt], actionCount)
```

consensus/driver/driver_test.go
```diff
@@ -0,0 +1,223 @@
...
+		}
+	}
+	return actions
+}
+
+func toAction(timeout tendermint.Timeout) tendermint.Action[value, felt.Felt, felt.Felt] {
+	return utils.HeapPtr(tendermint.ScheduleTimeout(timeout))
+}
+
+func requireBroadcast[M tendermint.Message[value, felt.Felt, felt.Felt]](
```
rianhughes, line 151: nit: what do you think about?
```suggestion
func incBroadcasterWaitGroup[M tendermint.Message[value, felt.Felt, felt.Felt]](
```

consensus/driver/driver_test.go
```diff
@@ -0,0 +1,223 @@
...
+	expectedBroadcast []M,
+	broadcaster driver.Broadcaster[M, value, felt.Felt, felt.Felt],
+) {
+	broadcaster.(*mockBroadcaster[M]).wg.Add(len(expectedBroadcast))
+}
+
+func assertBroadcast[M tendermint.Message[value, felt.Felt, felt.Felt]](
```
rianhughes, line 158: nit: feel free to ignore
```suggestion
func waitAndAssertBroadcaster[M tendermint.Message[value, felt.Felt, felt.Felt]](
```

consensus/driver/driver_test.go
rianhughes: Yup, that works as well

consensus/driver/driver_test.go
```diff
@@ -0,0 +1,223 @@
...
+		Step:   step,
+		Round:  tendermint.Round(rand.Int()),
+	}
+}
+
+const (
+	actionCount = 10
+)
+
+func registerActions(expectedBroadcast *expectedBroadcast) []tendermint.Action[value, felt.Felt, felt.Felt] {
```
rianhughes, line 126: nit
```suggestion
func genRandomActions(expectedBroadcast *expectedBroadcast) []tendermint.Action[value, felt.Felt, felt.Felt] {
```

consensus/driver/driver_test.go
```diff
@@ -0,0 +1,223 @@
...
+	inputPrevoteMsg := getRandPrevote()
+	inputPrecommitMsg := getRandPrecommit()
+
+	stateMachine.EXPECT().ProcessStart(tendermint.Round(0)).Return(append(registerActions(expectedBroadcast), toAction(inputTimeoutProposal)))
+	stateMachine.EXPECT().ProcessProposal(inputProposalMsg).Return(append(registerActions(expectedBroadcast), toAction(inputTimeoutPrevote)))
+	stateMachine.EXPECT().ProcessPrevote(inputPrevoteMsg).Return(append(registerActions(expectedBroadcast), toAction(inputTimeoutPrecommit)))
+	stateMachine.EXPECT().ProcessPrecommit(inputPrecommitMsg).Return(nil)
+	stateMachine.EXPECT().ProcessTimeout(inputTimeoutProposal).Return(registerActions(expectedBroadcast))
+	stateMachine.EXPECT().ProcessTimeout(inputTimeoutPrevote).Return(registerActions(expectedBroadcast))
+	stateMachine.EXPECT().ProcessTimeout(inputTimeoutPrecommit).Return(registerActions(expectedBroadcast))
```
rianhughes, line 200: Is this correct?
```suggestion
	// We don't braodcast messages because the input msgs are very unlikely to be for height(0) and round(0)
	stateMachine.EXPECT().ProcessTimeout(inputTimeoutProposal).Return(genRandomActions(expectedBroadcast))
	stateMachine.EXPECT().ProcessTimeout(inputTimeoutPrevote).Return(genRandomActions(expectedBroadcast))
```

consensus/driver/driver_test.go
infrmtcs: Actually it's something like `generateAndRegisterRandomActions`

consensus/driver/driver_test.go
infrmtcs: It actually doesn't have any special reason, I just want to also test the case where there are no actions returned.

consensus/driver/driver_test.go
rianhughes: Yup, I was thinking about suggesting that, but it's a bit long and makes the 
```
stateMachine.EXPECT().ProcessStart(tendermint.Round(0)).Return(...
```
lines a bit harder to read

rianhughes: Maybe `genAndRegRandomActions`??

consensus/driver/driver_test.go
infrmtcs: The line `stateMachine.EXPECT().ProcessTimeout(inputTimeoutProposal)` ensures that the timeout (proposal in this example) is scheduled.
The part `.Return(registerActions(expectedBroadcast))` ensures that the timeout is triggered, because we assert that the registered actions are received.

consensus/driver/driver_test.go
```diff
@@ -0,0 +1,223 @@
...
+		prevoteCh <- inputPrevoteMsg
+	}()
+	go func() {
+		precommitCh <- inputPrecommitMsg
+	}()
+
+	assertBroadcast(t, expectedBroadcast.proposals, broadcasters.ProposalBroadcaster)
+	assertBroadcast(t, expectedBroadcast.prevotes, broadcasters.PrevoteBroadcaster)
+	assertBroadcast(t, expectedBroadcast.precommits, broadcasters.PrecommitBroadcaster)
+
```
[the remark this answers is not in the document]
infrmtcs: The line stateMachine.EXPECT().ProcessTimeout(inputTimeoutProposal) ensures that the timeout (proposal in this example) is scheduled.
The part .Return(registerActions(expectedBroadcast)) ensures that the timeout is triggered, because we assert that the registered actions are received.

consensus/driver/driver.go
```diff
@@ -0,0 +1,127 @@
...
+		// For the time being it can be assumed the signature is correct.
+
+		for {
+			select {
+			case <-d.quit:
+				return
+			case tm := <-d.timeoutsCh:
+				// Handling of timeouts is priorities over messages
+				delete(d.scheduledTms, tm)
+				actions = d.stateMachine.ProcessTimeout(tm)
```
rianhughes, line 88: Would it be safer to delete after we process the timeout??

```
d.execute(d.stateMachine.ProcessTimeout(tm))
delete(d.scheduledTms, tm)
                ```

File: consensus/driver/driver.go [hunk 10]
```diff
@@ -0,0 +1,127 @@
...
+		getTimeout:   getTimeout,
+		listeners:    listeners,
+		broadcasters: broadcasters,
+		scheduledTms: make(map[tendermint.Timeout]*time.Timer),
+		timeoutsCh:   make(chan tendermint.Timeout),
+		quit:         make(chan struct{}),
+	}
+}
+
+func (d *Driver[V, H, A]) Start() {
```
rianhughes, line 70:
```suggestion
// The Driver is responsible for listening to messages from the network
// and passing them into the stateMachine. The stateMachine processes
// these messages and returns a set of actions to be executed by the Driver.
// The Driver executes these actions (namely broadcasting messages
// and triggering scheduled timeouts).
func (d *Driver[V, H, A]) Start() {
```

rianhughes:
LGTM up to the above comments

consensus/driver/driver_test.go
rodrigo-pino: It makes sense to me

consensus/driver/driver_test.go
rodrigo-pino: I think it's fine to go for the longer name, I think it's the lesser eveil. `generateAndRegisterRandomActions`

consensus/driver/driver.go
rodrigo-pino: I think is a good idea. I like the description you have provided Rian, it's very good. It's general enough so that anyone looking at the implementation can understand what's happening in the background without being specific enough that needs to be updating if we change implementations detail later

consensus/driver/driver.go
infrmtcs: I think it's not safer, because once we receive the "tock", we don't need to keep track of the timeout anymore.

rianhughes (approved):

PR #2787, opened 2025-04-28
feat: add pebble/v2 support

weiihann:
### Description
PebbleDB V2 will have performance improvements, but is not backward compatible with PebbleDB V1. Some strategies:

**1. Force users to do node resync**
Pros: Don't have to handle migrations, completely remove pebbleDB V1
Cons: Bad ux 

**2. Migration from V1 to V2**
Pros: Users don't have to do resync
Cons: migration time will be long, users need to have sufficient disk space, need to write the migration codes

**3. Introduce the standalone `pebblev2` package first, remove `pebble` package when we do state refactor**
Pros: Don't need to handle migrations, aligns nicely with the state and trie refactor roadmap
Cons: Depends on whether we can ship state refactor :)

Personally I lean towards option 3. We are not in a hurry to use PebbleDB V2 now. Since we are already doing state refactor, we can integrate pebble V2 as well. The integration is as simple as, removing `pebble` package and renaming `pebblev2` to `pebble`.

In this PR, the `pebblev2` package is a direct copy of the `pebble` package. Merge this only after https://github.com/NethermindEth/juno/pull/2661 is merged.

### Side note
The indirect dependency of `github.com/cockroachdb/swiss` can't be used in go1.24. So I temporarily replaced it with the commit that supports go1.24. Remove the replacement once there is a newer release.


go.mod
```diff
@@ -197,3 +201,5 @@ require (
 	lukechampine.com/blake3 v1.4.0 // indirect
 	rsc.io/tmplfunc v0.0.3 // indirect
 )
+
+replace github.com/cockroachdb/swiss v0.0.0-20240612210725-f4de07ae6964 => github.com/cockroachdb/swiss v0.0.0-20250327203710-2932b022f6df
```
infrmtcs, line 205: Can we have a TODO + explanation here?

infrmtcs (approved):

rianhughes (approved):

PR #2790, opened 2025-04-30
test: Integration tests for cluster of Tendermint drivers

infrmtcs:
This PR setup minimal tests to ensure that the nodes will eventually reach consensus. The setup configurations currently are simple and can be improved in the future:
- 10 nodes should reach consensus in 200 blocks.
- 200 nodes should reach consensus in 10 blocks.

For each of the setup, we're testing only 2 cases:
- All nodes are honest (happy path).
- All faulty nodes are silent.

### DIFF
--- a/consensus/integ/application.go
+++ b/consensus/integ/application.go
@@ -14,10 +14,10 @@ func (v value) Hash() felt.Felt {
 
 type application struct{}
 
-func (a *application) Value() value {
+func (a application) Value() value {
 	return value(rand.Int()) //nolint:gosec // The whole package is for testing purpose only, so it's safe to use weak random.
 }
 
-func (a *application) Valid(v value) bool {
+func (a application) Valid(v value) bool {
 	return v >= 0
 }
--- a/consensus/integ/application.go
+++ b/consensus/integ/application.go
@@ -1,0 +1,23 @@
+package integ
+
+import (
+	"math/rand"
+
+	"github.com/NethermindEth/juno/core/felt"
+)
+
+type value int64
+
+func (v value) Hash() felt.Felt {
+	return *new(felt.Felt).SetUint64(uint64(v))
+}
+
+type application struct{}
+
+func (a *application) Value() value {
+	return value(rand.Int()) //nolint:gosec // The whole package is for testing purpose only, so it's safe to use weak random.
+}
+
+func (a *application) Valid(v value) bool {
+	return v >= 0
+}
--- a/consensus/integ/blockchain.go
+++ b/consensus/integ/blockchain.go
@@ -7,17 +7,17 @@ import (
 
 type blockchain struct {
 	height   types.Height
-	nodeAddr felt.Felt
+	nodeAddr *felt.Felt
 	commits  chan commit
 }
 
 type commit struct {
-	nodeAddr felt.Felt
+	nodeAddr *felt.Felt
 	height   types.Height
 	value    value
 }
 
-func newBlockchain(commits chan commit, nodeAddr felt.Felt) *blockchain {
+func newBlockchain(commits chan commit, nodeAddr *felt.Felt) *blockchain {
 	return &blockchain{
 		height:   0,
 		nodeAddr: nodeAddr,
--- a/consensus/integ/blockchain.go
+++ b/consensus/integ/blockchain.go
@@ -1,0 +1,39 @@
+package integ
+
+import (
+	"github.com/NethermindEth/juno/consensus/types"
+	"github.com/NethermindEth/juno/core/felt"
+)
+
+type blockchain struct {
+	height   types.Height
+	nodeAddr felt.Felt
+	commits  chan commit
+}
+
+type commit struct {
+	nodeAddr felt.Felt
+	height   types.Height
+	value    value
+}
+
+func newBlockchain(commits chan commit, nodeAddr felt.Felt) *blockchain {
+	return &blockchain{
+		height:   0,
+		nodeAddr: nodeAddr,
+		commits:  commits,
+	}
+}
+
+func (b *blockchain) Height() types.Height {
+	return b.height
+}
+
+func (b *blockchain) Commit(height types.Height, value value, precommits []types.Precommit[felt.Felt, felt.Felt]) {
+	b.height = max(b.height, height)
+	b.commits <- commit{
+		nodeAddr: b.nodeAddr,
+		height:   height,
+		value:    value,
+	}
+}
--- a/consensus/integ/broadcaster.go
+++ b/consensus/integ/broadcaster.go
@@ -6,15 +6,15 @@ import (
 )
 
 type broadcaster[M types.Message[value, felt.Felt, felt.Felt]] struct {
-	addr      felt.Felt
+	addr      *felt.Felt
 	listeners map[felt.Felt]listener[M]
 }
 
 func newBroadcaster[M types.Message[value, felt.Felt, felt.Felt]](
-	addr felt.Felt,
+	addr *felt.Felt,
 	listeners map[felt.Felt]listener[M],
-) *broadcaster[M] {
-	return &broadcaster[M]{
+) broadcaster[M] {
+	return broadcaster[M]{
 		addr:      addr,
 		listeners: listeners,
 	}
@@ -22,7 +22,7 @@ func newBroadcaster[M types.Message[value, felt.Felt, felt.Felt]](
 
 func (b broadcaster[M]) Broadcast(msg M) {
 	for addr, listener := range b.listeners {
-		if addr != b.addr {
+		if !addr.Equal(b.addr) {
 			go func() {
 				listener.ch <- msg
 			}()
--- a/consensus/integ/broadcaster.go
+++ b/consensus/integ/broadcaster.go
@@ -1,0 +1,31 @@
+package integ
+
+import (
+	"github.com/NethermindEth/juno/consensus/types"
+	"github.com/NethermindEth/juno/core/felt"
+)
+
+type broadcaster[M types.Message[value, felt.Felt, felt.Felt]] struct {
+	addr      felt.Felt
+	listeners map[felt.Felt]listener[M]
+}
+
+func newBroadcaster[M types.Message[value, felt.Felt, felt.Felt]](
+	addr felt.Felt,
+	listeners map[felt.Felt]listener[M],
+) *broadcaster[M] {
+	return &broadcaster[M]{
+		addr:      addr,
+		listeners: listeners,
+	}
+}
+
+func (b broadcaster[M]) Broadcast(msg M) {
+	for addr, listener := range b.listeners {
+		if addr != b.addr {
+			go func() {
+				listener.ch <- msg
+			}()
+		}
+	}
+}
--- a/consensus/integ/integ_test.go
+++ b/consensus/integ/integ_test.go
@@ -43,10 +43,10 @@ func runTest(t *testing.T, cfg testConfig) {
 		testDB, err := pebble.New(dbPath)
 		require.NoError(t, err)
 
-		nodeAddr := allNodes.addr[i]
+		nodeAddr := &allNodes.addr[i]
 
 		stateMachine := tendermint.New(
-			nodeAddr,
+			*nodeAddr,
 			&application{},
 			newBlockchain(commits, nodeAddr),
 			validators,
@@ -75,7 +75,7 @@ func runTest(t *testing.T, cfg testConfig) {
 			continue
 		}
 
-		nodeIdx, ok := allNodes.index[commit.nodeAddr]
+		nodeIdx, ok := allNodes.index[*commit.nodeAddr]
 		assert.True(t, ok)
 
 		// Ignore faulty nodes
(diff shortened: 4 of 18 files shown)


consensus/tendermint/rule_commit_value.go
```diff
@@ -33,5 +33,9 @@ func (t *stateMachine[V, H, A]) doCommitValue(cachedProposal *CachedProposal[V,
 
 	t.messages.deleteHeightMessages(t.state.height)
 	t.state.height++
+	t.state.lockedRound = -1
```
infrmtcs, line 36: This integ test caught this error.

consensus/integ/integ_test.go
```diff
@@ -0,0 +1,144 @@
...
+	timeoutStepFactor  time.Duration
+	timeoutRoundFactor time.Duration
+	targetHeight       types.Height
+}
+
+func getTimeoutFn(cfg testConfig) func(types.Step, types.Round) time.Duration {
+	return func(step types.Step, round types.Round) time.Duration {
+		return cfg.timeoutBase + time.Duration(step)*cfg.timeoutStepFactor + time.Duration(round)*cfg.timeoutRoundFactor
+	}
+}
```
rodrigo-pino, line 29: I was going to write a whole comment here saying that because the fields of `cfg` are copied value, no closure would happen over `cfg` so passing it by value (in `getTimeoutFn`) looking for guarantees of immutability inside the delegate wasn't necessary and we could pass by reference (i.e. `cfg *testConfig`) but I think I am wrong. Passing by value for the `getTimeoutFn` makes sense.

consensus/integ/network.go
```diff
@@ -0,0 +1,45 @@
...
+	for _, addr := range allNodes.addr {
+		n.proposals[addr] = newListener[types.Proposal[value, felt.Felt, felt.Felt]](buffer)
+		n.prevotes[addr] = newListener[types.Prevote[felt.Felt, felt.Felt]](buffer)
+		n.precommits[addr] = newListener[types.Precommit[felt.Felt, felt.Felt]](buffer)
+	}
+
+	return &n
+}
+
+func (n network) getListeners(addr felt.Felt) driver.Listeners[value, felt.Felt, felt.Felt] {
```
rodrigo-pino, line 31: Pass `addr` by reference?

consensus/integ/network.go
```diff
@@ -0,0 +1,45 @@
...
+	return driver.Listeners[value, felt.Felt, felt.Felt]{
+		ProposalListener:  n.proposals[addr],
+		PrevoteListener:   n.prevotes[addr],
+		PrecommitListener: n.precommits[addr],
+	}
+}
+
+func (n network) getBroadcasters(addr felt.Felt) driver.Broadcasters[value, felt.Felt, felt.Felt] {
```
rodrigo-pino, line 39: pass `addr` by reference?

consensus/integ/broadcaster.go
```diff
@@ -0,0 +1,31 @@
...
+) *broadcaster[M] {
+	return &broadcaster[M]{
+		addr:      addr,
+		listeners: listeners,
+	}
+}
+
+func (b broadcaster[M]) Broadcast(msg M) {
+	for addr, listener := range b.listeners {
+		if addr != b.addr {
```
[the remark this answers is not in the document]
infrmtcs: The current behaviour doesn't expect to receive the node's own message. Another approach to solve the issue is to accept that possibility.

consensus/integ/application.go
```diff
@@ -0,0 +1,23 @@
+package integ
```
rianhughes, line 1: nit: can we rename this package `integration`?

consensus/integ/application.go
```diff
@@ -0,0 +1,23 @@
...
+
+type value int64
+
+func (v value) Hash() felt.Felt {
+	return *new(felt.Felt).SetUint64(uint64(v))
+}
+
+type application struct{}
+
+func (a application) Value() value {
```
rianhughes, line 17:
```suggestion
// Value() returns both valid and invalid values, with equal likelihood.
func (a application) Value() value {
```

infrmtcs: The current behaviour is only returning valid values. We're testing only happy case (all honest) and simple Byzantine case (1 node is always silent).

consensus/integ/broadcaster.go
```diff
@@ -0,0 +1,31 @@
...
+	return broadcaster[M]{
+		addr:      addr,
+		listeners: listeners,
+	}
+}
+
+func (b broadcaster[M]) Broadcast(msg M) {
+	for addr, listener := range b.listeners {
+		if !addr.Equal(b.addr) {
+			go func() {
```
rianhughes, line 26:
```suggestion
func (b broadcaster[M]) Broadcast(msg M) {
	for addr, listener := range b.listeners {
		// Don't send message to ourself
		if !addr.Equal(b.addr) {
			go func() {
```

consensus/integ/broadcaster.go
```diff
@@ -0,0 +1,31 @@
+package integ
+
+import (
+	"github.com/NethermindEth/juno/consensus/types"
+	"github.com/NethermindEth/juno/core/felt"
+)
+
+type broadcaster[M types.Message[value, felt.Felt, felt.Felt]] struct {
+	addr      *felt.Felt
```
rianhughes, line 9:
```suggestion
	addr      *felt.Felt // our address
```

consensus/integ/blockchain.go
```diff
@@ -0,0 +1,39 @@
...
+		nodeAddr: nodeAddr,
+		commits:  commits,
+	}
+}
+
+func (b *blockchain) Height() types.Height {
+	return b.height
+}
+
+func (b *blockchain) Commit(height types.Height, value value, precommits []types.Precommit[felt.Felt, felt.Felt]) {
```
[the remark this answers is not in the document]
infrmtcs: For the signature, I think we can wait until the specs specifying how the signature is created and the field storing it.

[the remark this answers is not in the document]
infrmtcs: For the precommits, the test cannot asserts the exact set of nodes sending precommits in happy case, so that's why we don't need to use it here.

consensus/integ/validator.go
```diff
@@ -0,0 +1,36 @@
...
+}
+
+func (v validators) ValidatorVotingPower(addr felt.Felt) types.VotingPower {
+	return types.VotingPower(1)
+}
+
+func (v validators) Proposer(height types.Height, round types.Round) felt.Felt {
+	idx := (int(height)*31 + int(round)*17) % len(v.allocations)
+	return v.allocations[idx]
+}
```
rianhughes, line 36: Why 31 and 17?

consensus/integ/validator.go
```diff
@@ -0,0 +1,36 @@
...
+import (
+	"math/rand"
+	"slices"
+
+	"github.com/NethermindEth/juno/consensus/types"
+	"github.com/NethermindEth/juno/core/felt"
+)
+
+type validators struct {
+	allocations []felt.Felt
```
rianhughes, line 12: `allocations` is just the validator address right? Why not just call it `address`?

consensus/integ/broadcaster.go
```diff
@@ -0,0 +1,31 @@
...
+	listeners map[felt.Felt]listener[M]
+}
+
+func newBroadcaster[M types.Message[value, felt.Felt, felt.Felt]](
+	addr *felt.Felt,
+	listeners map[felt.Felt]listener[M],
+) broadcaster[M] {
+	return broadcaster[M]{
```
rianhughes, line 17: Can we add a doc string here please? Why does each validator need to be aware of all other listener channels? Just so we can relay messages to them?

consensus/integ/integ_test.go
```diff
@@ -0,0 +1,144 @@
...
+
+type testConfig struct {
+	nodeCount          int
+	silentFaultyNodes  bool
+	buffer             int
+	timeoutBase        time.Duration
+	timeoutStepFactor  time.Duration
+	timeoutRoundFactor time.Duration
+	targetHeight       types.Height
+}
```
rianhughes, line 23: Would you mind adding a comment for each of the fields? 🙏

consensus/integ/integ_test.go
```diff
@@ -0,0 +1,144 @@
...
+		cfg := cfg
+		cfg.silentFaultyNodes = false
+		runTest(t, cfg)
+	})
+
+	t.Run("silent faulty nodes", func(t *testing.T) {
+		cfg := cfg
+		cfg.silentFaultyNodes = true
+		runTest(t, cfg)
+	})
```
rianhughes, line 119: `silentFaultyNodes` doesn't do anything, so these are the same tests.

rianhughes:
Generally looks good! Just some comments about making it a bit more readable.

consensus/integ/validator.go
infrmtcs: It's just a random function, with prime coefficients so that for each height, the order of proposers is different.

consensus/integ/network.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/NethermindEth/juno/consensus/driver"
+	"github.com/NethermindEth/juno/consensus/types"
+	"github.com/NethermindEth/juno/core/felt"
+)
+
+type network struct {
+	proposals  map[felt.Felt]listener[types.Proposal[value, felt.Felt, felt.Felt]]
+	prevotes   map[felt.Felt]listener[types.Prevote[felt.Felt, felt.Felt]]
+	precommits map[felt.Felt]listener[types.Precommit[felt.Felt, felt.Felt]]
+}
```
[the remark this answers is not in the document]
infrmtcs: I'd do this in another PR. I'd create a small package named `starknet` and put the type aliases there, so we can change the type easily if needed.

consensus/integ/application.go
infrmtcs: I renamed it to `integtest`, so show the intention clearly that this is only for testing.

consensus/integ/validator.go
infrmtcs: Added an explanation

consensus/integ/validator.go
infrmtcs: Replaced by named type

consensus/integ/broadcaster.go
infrmtcs: Added an explanation

consensus/integ/integ_test.go
infrmtcs: Added comments

rodrigo-pino (approved):

consensus/integ/broadcaster.go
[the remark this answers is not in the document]
rodrigo-pino: How does `b.addr` ends up inside `b.listeners`? Is that intended?

consensus/tendermint/tendermint.go
```diff
@@ -163,29 +158,15 @@ func q(totalVotingPower types.VotingPower) types.VotingPower {
...
-	// If the height is a future height, the round is considered to be 0, as the height hasn't started yet.
-	if isCurrentHeight {
-		currentRoundOfHeaderHeight = t.state.round
-	}
-
-	switch {
-	case header.Height < t.state.height || header.Height > t.state.height+maxFutureHeight:
-		return false
-	case header.Round < 0 || header.Round > currentRoundOfHeaderHeight+maxFutureRound:
+	if header.Height < t.state.height || header.Round < 0 {
```
rianhughes, line 164: I don't think header.Round should ever be less than 0 (unless there's a bug in our implementation)

https://github.com/starknet-io/starknet-p2p-specs/blob/main/p2p/proto/consensus/consensus.proto#L33C5-L33C43

consensus/integtest/integ_test.go
```diff
@@ -0,0 +1,151 @@
...
+		}
+
+		nodeIdx, ok := allNodes.index[*commit.nodeAddr]
+		assert.True(t, ok)
+
+		// Ignore faulty nodes
+		if nodeIdx >= honestNodeCount {
+			continue
+		}
+
```
rianhughes, line 92: I think we can remove this. The faulty nodes never run, so never `commit`, and the honest nodes are labelled 0 to N

rianhughes (approved):

consensus/integtest/integ_test.go
infrmtcs: I think in the near future we can start having non trivial tests. E.g, drop 50% of the meesages

consensus/integtest/integ_test.go
```diff
@@ -0,0 +1,151 @@
...
+			// Subsequent committer, check the value
+			assert.Equal(t, *value, commit.value)
+		} else {
+			// First committer, store the value
+			committedValues[commit.height] = &commit.value
+		}
+
+		if commit.height == cfg.targetHeight-1 {
+			finished++
+		}
```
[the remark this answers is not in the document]
infrmtcs: If the driver stop before it can broadcast to another node, that node can stucks forever, so I decided not to stop here but let them run.

PR #2791, opened 2025-05-02
refactor: remove use of `copier` pkg

### LINKED ISSUE #2772: Remove the use of copier package in rpc
We currently are using the copier package only in rpc, and there's no reason to use it as it introduces hidden dependencies and might break stuff unexpectedly. We should remove and just directly copy the structs.

cwkang1998:
This PR removes the use of `copier` pkg from all rpc package and instead replace it with the use of the adapt functions.

Closes #2772.


infrmtcs (approved):

weiihann (approved):
LGTM

PR #2795, opened 2025-05-05
feature: implement DB for consensus WAL

rianhughes:
## Implement DB for Tendermint Consensus Logic & WAL Storage

**Description:**

This PR implements a dedicated database layer (`TMDB`) for the Tendermint consensus logic. This is a crucial step towards preventing equivocation in case of node failures and reboots by enabling the persistent storage and retrieval of Write-Ahead Log (WAL) entries (consensus messages and timeouts).

**Key Features & Changes:**
    *   Introduces `TMDB` type to manage consensus state persistence.
    *   Utilizes `pebble.Batch` for efficient writes, accumulating changes before committing. 
    *   Note: Custom marshal functions were implemented for the `timeout` type in order to keep the fields private.

**Important Note on Batching Writes**

 * Writes are batched. In the case where the node crashes, these writes may be lost. This introduces a risk of equivocation etc when restarting the node after a crash. This was done purely for performance purposes (writing to disk is expensive). We may just want to commit any WAL-msg to disk immediately for safety reasons. The current implementation is flexible, we can batch commits, or commit every message immediately. 

**Next Steps:**

*   Utilize this DB implementation to replay messages upon restart.


consensus/tendermint/tendermint.go
```diff
@@ -299,6 +307,38 @@ type timeout struct {
 	r round
 }
 
+// MarshalCBOR implements the cbor.Marshaler interface.
+func (t timeout) MarshalCBOR() ([]byte, error) {
```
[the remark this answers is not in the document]
rianhughes: I can remove this, but I'll need to export the timeout fields

consensus/tendermint/messages.go
```diff
@@ -26,13 +32,57 @@ type (
...
 type MessageHeader[A Addr] struct {
-	Height height
-	Round  round
-	Sender A
+	Height height `cbor:"height"`
+	Round  round  `cbor:"round"`
+	Sender A      `cbor:"sender"`
+}
+
+func (v *Vote[H, A]) MarshalCBOR() ([]byte, error) {
```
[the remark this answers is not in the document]
rianhughes: Whoops, should have been removed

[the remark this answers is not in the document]
rianhughes: Done

consensus/tendermint/tendermint.go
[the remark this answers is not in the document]
infrmtcs: Actually my current PR is exporting Timeout and other types. The rationale is because I need to separate driver and state machine in different pkg to limit the ability of driver to modify state machine state.

consensus/tendermint/tmdb.go
```diff
@@ -0,0 +1,353 @@
...
+		return "precommit"
+	case MessageTypeTimeout:
+		return "timeout"
+	default:
+		return "unknown"
+	}
+}
+
+type WALMsg[V Hashable[H], H Hash, A Addr] struct {
+	Msg     any
```
infrmtcs, line 71: Is it possible to use a different type than any?

consensus/tendermint/tmdb.go
```diff
@@ -0,0 +1,353 @@
...
+// We use a Batch (db.IndexedBatch) to accumulate writes before committing them to the DB.
+// This reduces expensive disk I/O. Reads check the batch first.
+// WARNING: If the process crashes before Commit(), buffered messages are lost,
+// risking consensus issues like equivocation (missing votes, double-signing).
+
+const NumBytesForHeight = 4
+
+type wrappedMsg struct {
+	Type string          `cbor:"type"`
+	Data cbor.RawMessage `cbor:"data"` // cbor serialised Msg or Timeout
```
infrmtcs, line 40: Is it possible to use a different type than this?

consensus/tendermint/tendermint.go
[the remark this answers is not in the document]
rianhughes: Cool, in that case it makes sense to remove these methods

vm/rust/src/execution.rs
```diff
@@ -1,5 +1,6 @@
 use crate::error::ExecutionError;
 use crate::juno_state_reader::JunoStateReader;
+use anyhow::{anyhow, Error};
```
infrmtcs, line 3: Outside of this PR, but I think we should run a Rust formatter in our CI to make sure Rust code are also well formatted

consensus/tendermint/tmdb.go
```diff
@@ -0,0 +1,353 @@
...
+	// Get NumMsgsAtHeight and increment
+	numMsgsAtHeight, err := s.GetNumMsgsAtHeight(height)
+	if err != nil {
+		if errors.Is(err, db.ErrKeyNotFound) {
+			numMsgsAtHeight = 0 // first time storing a msg
+		} else {
+			return fmt.Errorf("setWALEntry: failed to get numMsgsAtHeight: %w", err)
+		}
+	}
+	numMsgsAtHeight++
```
infrmtcs, line 232: I think it's maybe clearer if we change this to something like
```
nextMsgId := numMsgsAtHeight + 1
```

consensus/tendermint/tendermint.go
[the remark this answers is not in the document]
rianhughes: Done

consensus/tendermint/tmdb.go
rianhughes: I was thinking about implementing defining an interface and passing it in here, but I feel like it would result in a lot of boiler plate

consensus/tendermint/tmdb.go
rianhughes: I was thinking about defining an interface for this, but I feel like it would result in unnecessary boiler plate code

consensus/tendermint/tmdb.go
rianhughes: Ideally we would use 

```
type WALMsg[V Hashable[H], H Hash, A Addr] struct {
	Msg     Message[V, H, A]
	Timeout *timeout
}
```

but Golang doesn't seem to like using Unions of types to constrain fields

consensus/tendermint/tmdb.go
rianhughes: For the Data field? Maybe we could implement an interface and use that?

consensus/tendermint/tmdb.go
infrmtcs: Yeah, I'm also thinking about creating an interface there. What are the different purposes of `wrappedMsg` and `WALMsg`?

consensus/tendermint/tmdb.go
```diff
@@ -0,0 +1,353 @@
...
+		return "unknown"
+	}
+}
+
+type WALMsg[V Hashable[H], H Hash, A Addr] struct {
+	Msg     any
+	Timeout *timeout
+}
+
+type TMDB struct {
```
infrmtcs, line 75: I think it's a good practice to have a `TMDB` interface and a `tmdb` struct as the implementation.

infrmtcs: And I think we can separate this into a separate package, so we can control the access of internal fields better.

consensus/tendermint/tmdb.go
infrmtcs: Yeah, I meant the `Data` field, also think that we should use an interface here if possible.

consensus/tendermint/tendermint.go
```diff
@@ -294,17 +301,17 @@ func (t *Tendermint[V, H, A]) startRound(r round) Action[V, H, A] {
 }
 
 type timeout struct {
-	s step
-	h height
-	r round
+	S step   `cbor:"s"`
+	H height `cbor:"h"`
+	R round  `cbor:"r"`
```
infrmtcs, line 306:
```suggestion
	_ struct{} `cbor:",toarray"`
	S step
	H height
	R round
```
The default encoding mode is using a map with the field name as key. I think it would be much more efficient to use the array mode instead. In that case, it's not required to annotate the fields, as the field name doesn't matter anymore.

infrmtcs: Sample
```
package main

import (
	"encoding/hex"
	"fmt"

	"github.com/fxamacker/cbor/v2"
)

type A1 struct {
	X int
	Y int
	Z string
}

type B1 struct {
	A1
	U int
	V string
}

type A2 struct {
	_ struct{} `cbor:",toarray"`
	X int
	Y int
	Z string
}

type B2 struct {
	_ struct{} `cbor:",toarray"`
	A2
	U int
	V string
}

func main() {
	b1 := B1{A1: A1{X: 1, Y: 2, Z: "3"}, U: 4, V: "5"}
	b2 := B2{A2: A2{X: 1, Y: 2, Z: "3"}, U: 4, V: "5"}
	res1, _ := cbor.Marshal(b1)
	res2, _ := cbor.Marshal(b2)
	fmt.Println(hex.EncodeToString(res1))
	fmt.Println(hex.EncodeToString(res2))
}
```

```
a5615801615902615a613361550461566135
8501026133046135
```

consensus/tendermint/tendermint.go
```diff
@@ -116,6 +117,8 @@ type Broadcasters[V Hashable[H], H Hash, A Addr] struct {
 }
 
 type Driver[V Hashable[H], H Hash, A Addr] struct {
+	tmdb TMDB
```
infrmtcs, line 120: I'm wondering if we can do this instead:
```suggestion
	db DB
```

consensus/tendermint/tmdb.go
rianhughes: Yup, good point. I re-factored to introduce the interface separation

consensus/tendermint/tmdb.go
rianhughes: `wrappedMsg` contains additional fields to help with decoding (eg the type we endoed, `prevote`, `timeout` etc, and if we need to timestamp msgs in the future). `WALMsg` was just meant to return either a `Message` or `Timeout`. However, it has been removed in the latest refactor

consensus/tendermint/tendermint.go
rianhughes: Yup, nice one. Updated

consensus/tendermint/tendermint.go
rianhughes: Updated

db/buckets_consensus.go
```diff
@@ -0,0 +1,21 @@
...
+// man's bucket alternative.
+const (
+	WALEntry      BucketConsensus = iota // key: WAL_prefix + Height + MsgIndex. Val: Encoded Tendermint consensus message.
+	WALEntryCount                        // Key: WAL_count_prefix + Height. Val: Counter (number of WAL entries stored at this height)
+)
+
+// Key flattens a prefix and series of byte arrays into a single []byte.
+func (b BucketConsensus) Key(key ...[]byte) []byte {
+	return append([]byte{byte(b)}, slices.Concat(key...)...)
+}
```
weiihann, line 21: All new buckets must strictly be put into `buckets.go`. Each bucket is splitted uniquely by `byte`, for example:
- `StateTrie` - 0
- `Peer` - 1
- ...

Doing it in separate file will overwrite existing buckets and cause db corruption.

infrmtcs: We intend to use a separate file for the consensus, so this can be treated as a separate DB.

consensus/tendermint/tmdb.go
```diff
@@ -0,0 +1,394 @@
...
+	// SetWALEntry schedules the storage of a WAL message in the batch.
+	SetWALEntry(entry IsWALMsg, height height) error
+	// DeleteWALMsgs schedules the deletion of all WAL messages for a specific height in the batch.
+	DeleteWALMsgs(height height) error
+}
+
+// TMDB provides database access for Tendermint consensus state.
+type TMDB[V Hashable[H], H Hash, A Addr] struct {
+	db    db.KeyValueStore
+	batch db.IndexedBatch
```
weiihann, line 88: `IndexedBatch` should be avoided at all cost. All the use cases that I see is that you need an in-memory cache. You can use `"github.com/VictoriaMetrics/fastcache"` to write the key-value pair. To persist the batch, in `Commit` you create `pebble.Batch` and write the key-value pairs from the cache and write to the DB.

consensus/tendermint/tmdb.go
```diff
@@ -0,0 +1,394 @@
...
+	// GetWALMsgs retrieves all WAL messages (consensus messages and timeouts) stored for a given height from the database.
+	GetWALMsgs(height height) ([]IsWALMsg, error)
+	// SetWALEntry schedules the storage of a WAL message in the batch.
+	SetWALEntry(entry IsWALMsg, height height) error
+	// DeleteWALMsgs schedules the deletion of all WAL messages for a specific height in the batch.
+	DeleteWALMsgs(height height) error
+}
+
+// TMDB provides database access for Tendermint consensus state.
+type TMDB[V Hashable[H], H Hash, A Addr] struct {
```
infrmtcs, line 86: I think TM isn't a good abbreviation for Tendermint
```suggestion
type TendermintDB[V Hashable[H], H Hash, A Addr] struct {
```

consensus/tendermint/tmdb.go
```diff
@@ -0,0 +1,394 @@
...
+}
+
+// MessageType represents the type of message stored in the WAL.
+type MessageType uint8
+
+const (
+	MessageTypeProposal MessageType = iota
+	MessageTypePrevote
+	MessageTypePrecommit
+	MessageTypeTimeout
```
infrmtcs, line 51: I think it's not necessary to store timeout because:
- Even if we don't have timeout information on restart, the algorithm can still reschedule the timeout immediately if it's still in the same height and round. If it has moved to a new height and round, timeout won't do anything so no need to reschedule.
- The `Timeout` struct doesn't have any information about when it's set or when the deadline is, so it doesn't have enough information to schedule the original deadline. That said, I think it's not necessary to do so.

consensus/tendermint/tmdb.go
infrmtcs: > All the use cases that I see is that you need an in-memory cache. You can use "github.com/VictoriaMetrics/fastcache" to write the key-value pair.

Regarding this part, we actually need the database to at least store the prevotes and precommits broadcasted by our node, otherwise the node can accidentally have dishonest behaviour.

That said, let us know the best practices to use pebbledb in this case, I'm understanding that Rian wants to improve the performance of DB operations here.

db/buckets_consensus.go
infrmtcs: I think probably we should put this and the `tmdb.go` file in a separate package in `consensus/db` to show the intention clearer.

consensus/tendermint/tmdb.go
```diff
@@ -0,0 +1,394 @@
...
+		// If adding the count key deletion fails, return the error.
+		return fmt.Errorf("DeleteWALMsgs: failed to add count key deletion for height %d to batch: %w", height, err)
+	}
+
+	// All delete operations successfully added to the batch.
+	return nil
+}
+
+// setNumMsgsAtHeight stores the total message count for a height in the batch.
+func (s *TMDB[V, H, A]) setNumMsgsAtHeight(height height, walIter uint32) error {
```
infrmtcs, line 181: Can we wrap `uint32` type as something like `walIter` and use it instead?

db/buckets_consensus.go
weiihann: What's the reason of using a separate DB instead of putting it into the same one?

consensus/tendermint/tmdb.go
weiihann: You can think about `IndexedBatch` as a combination of an in-memory cache plus the underlying disk database. But the write performance is worse than a regular `Batch`, so you might as well use a custom-built in-memory cache to buffer the changes, and use a regular `Batch` to flush the buffer to the database.

consensus/tendermint/tmdb.go
```diff
@@ -0,0 +1,394 @@
...
+	heightBytes := heightToBytes(height)
+
+	// Delete individual message keys. Message iterators are 1-based.
+	for i := uint32(1); i <= numMsgs; i++ {
+		iterBytes := encodeNumMsgsAtHeight(i) // Get the bytes for the iterator number
+		msgKey := db.WALEntry.Key(heightBytes, iterBytes)
+		if err := s.batch.Delete(msgKey); err != nil {
+			return fmt.Errorf("DeleteWALMsgs: failed to add deletion for message key %d/%d to batch: %w", height, i, err)
+		}
+	}
```
infrmtcs, line 167: I think we should use this underlying instead:
https://pkg.go.dev/github.com/cockroachdb/pebble#Batch.DeleteRange

consensus/tendermint/tmdb.go
```diff
@@ -0,0 +1,394 @@
...
+	return s.batch.Put(key, val)
+}
+
+// SetWALEntry implements TMDBInterface.
+func (s *TMDB[V, H, A]) SetWALEntry(entry IsWALMsg, height height) error {
+	var msgType MessageType
+	var msgDataInner []byte
+	var err error
+
+	switch m := entry.(type) {
```
infrmtcs, line 194: Removing the `Marshal` and `Unmarshal` functions for most types is great because most of them are having good default behaviour. However, for the sum types, the library doesn't support any default behaviour, so I think it's fine to have `Marshal` and `Unmarshal` functions for them.
I think it's better to implement those functions in the same expected function signature as the library instead of putting them here.

consensus/tendermint/tmdb.go
infrmtcs: Got it, the writes are supposed to be sequential anyway, so we should avoid indexed batch and not pay the cost of supporting read.

consensus/tendermint/tmdb.go
```diff
@@ -0,0 +1,394 @@
...
+		return "unknown"
+	}
+}
+
+// TMDBInterface defines the methods for interacting with the Tendermint WAL database.
+type TMDBInterface[V Hashable[H], H Hash, A Addr] interface {
+	// CommitBatch writes the accumulated batch operations to the underlying database.
+	CommitBatch() error
+	// GetWALCount retrieves the number of WAL messages stored for a given height.
+	GetWALCount(height height) (uint32, error)
```
infrmtcs, line 76: I think this can be private, because probably the driver doesn't need to know this information.

consensus/tendermint/tmdb.go
```diff
@@ -0,0 +1,394 @@
...
+		return "timeout"
+	default:
+		return "unknown"
+	}
+}
+
+// TMDBInterface defines the methods for interacting with the Tendermint WAL database.
+type TMDBInterface[V Hashable[H], H Hash, A Addr] interface {
+	// CommitBatch writes the accumulated batch operations to the underlying database.
+	CommitBatch() error
```
infrmtcs, line 74: I think this can be private, because the implementation itself should determint when to call `CommitBatch` instead of letting `Driver` to decides. For example, the implementation can detected if the written message is from our own node, then call `CommitBatch` immediately.

db/buckets_consensus.go
```diff
@@ -0,0 +1,21 @@
...
+//
+//go:generate go run github.com/dmarkham/enumer -type=BucketConsensus -output=buckets_consensus_enumer.go
+type BucketConsensus byte
+
+// Pebble does not support buckets to differentiate between groups of
+// keys like Bolt or MDBX does. We use a global prefix list as a poor
+// man's bucket alternative.
+const (
+	WALEntry      BucketConsensus = iota // key: WAL_prefix + Height + MsgIndex. Val: Encoded Tendermint consensus message.
+	WALEntryCount                        // Key: WAL_count_prefix + Height. Val: Counter (number of WAL entries stored at this height)
```
infrmtcs, line 15: I'm wondering if we actually need this. The writes are sequential anyway, so we can keep track of this in memory instead.

db/buckets_consensus.go
weiihann: Separating the execution and consensus makes sense

consensus/tendermint/tmdb.go
rianhughes: Actually, I think I might be able to replace `IndexedBatch` with `Batch`, and just handle the `WALEntryCount` explicitly in memory. I'll try this approach

consensus/tendermint/tmdb.go
rianhughes: I originally implemented this using `DeleteRange` but it turned out to be a bit uglier than the current approach (I think it had something to do with the end range,  but I can try again)

db/buckets_consensus.go
rianhughes: Do you mean remove `WALEntryCount` from the DB, and just handle it in memory by scanning the number of `WALEntry`s at startup? I like this approach

consensus/tendermint/tmdb.go
rianhughes: `IndexedBatch` has been replaced by `Batch`. The counter is now stored in memory so we don't need to read from the batch / cache

consensus/tendermint/tmdb.go
rianhughes: Removed

db/buckets_consensus.go
rianhughes: The counter is now stored in memory so we don't need to read from the batch / cache. `WALEntryCount` has been removed

db/buckets_consensus.go
rianhughes: I've moved this file to `db/consensus`

consensus/tendermint/tmdb.go
rianhughes: Updated

consensus/tendermint/tmdb.go
rianhughes: I simplified the marshalling logic to remove this switch statement

consensus/tendermint/tmdb.go
rianhughes: I updated the `wrappedMsg` type to

```
type walEntry struct {
	Type  MessageType `cbor:"type"`
	Entry IsWALMsg    `cbor:"data"` // cbor serialised Msg or Timeout
}
```

consensus/tendermint/tmdb.go
rianhughes: I've updated the code to use `DeleteRange` (now that we store the counter in memory it's much cleaner than before)

consensus/tendermint/tendermint.go
```diff
@@ -294,17 +301,21 @@ func (t *Tendermint[V, H, A]) startRound(r round) Action[V, H, A] {
 }
 
 type timeout struct {
-	s step
-	h height
-	r round
+	S step   `cbor:"s"`
+	H height `cbor:"h"`
+	R round  `cbor:"r"`
```
infrmtcs, line 306:
```suggestion
	Step step `cbor:"s"`
	Height height `cbor:"h"`
	Round round  `cbor:"r"`
```
to reduce conflict with #2786

consensus/tendermint/tmdb.go
```diff
@@ -0,0 +1,311 @@
...
+	// GetWALMsgs retrieves all WAL messages (consensus messages and timeouts) stored for a given height from the database.
+	GetWALMsgs(height height) ([]walEntry, error)
+	// SetWALEntry schedules the storage of a WAL message in the batch.
+	SetWALEntry(entry IsWALMsg, height height) error
+	// DeleteWALMsgs schedules the deletion of all WAL messages for a specific height in the batch.
+	DeleteWALMsgs(height height) error
+}
+
+// TendermintDB provides database access for Tendermint consensus state.
+type TendermintDB[V Hashable[H], H Hash, A Addr] struct {
```
infrmtcs, line 85:
```suggestion
type tendermintDB[V Hashable[H], H Hash, A Addr] struct {
```
to hide this struct

consensus/tendermint/tmdb.go
```diff
@@ -0,0 +1,311 @@
...
+		return "precommit"
+	case MessageTypeTimeout:
+		return "timeout"
+	default:
+		return "unknown"
+	}
+}
+
+// TMDBInterface defines the methods for interacting with the Tendermint WAL database.
+type TMDBInterface[V Hashable[H], H Hash, A Addr] interface {
```
infrmtcs, line 73:
```suggestion
type TendermintDB[V Hashable[H], H Hash, A Addr] interface {
```

consensus/tendermint/tmdb.go
```diff
@@ -0,0 +1,311 @@
...
+// We use a Batch (db.IndexedBatch) to accumulate writes before committing them to the DB.
+// This reduces expensive disk I/O. Reads check the batch first.
+// WARNING: If the process crashes before Commit(), buffered messages are lost,
+// risking consensus issues like equivocation (missing votes, double-signing).
+
+const NumBytesForHeight = 4
+
+type walIter uint32
+
+type walEntry struct {
```
infrmtcs, line 40: We should use
```suggestion
type walEntry struct {
_ struct{} `cbor:",toarray"`
```

rianhughes: I removed this before because it was causing problems with marhsalling, but it seems to be working now, so I reintroduced it

consensus/tendermint/tmdb.go
rianhughes: Just adding some comments here for visibility incase anyone else visits this message. Basically, it's not clear to me that we can drop the timeouts given 1) both [tendermint/tendermint](https://github.com/tendermint/tendermint/blob/main/consensus/replay.go#L82-L84) and [malachite](https://github.com/informalsystems/malachite/blob/main/code/crates/engine/src/consensus.rs#L773) explicitly store and process timeouts, and 2) there seems to be situations where replaying the broadcasted msgs isn't enough to recover the state (eg line 34 - where we receive a quorum of prevotes but not msgs is broadcasted)

consensus/tendermint/tmdb.go
```diff
@@ -0,0 +1,311 @@
...
+
+// tendermintDB provides database access for Tendermint consensus state.
+type tendermintDB[V Hashable[H], H Hash, A Addr] struct {
+	db       db.KeyValueStore
+	batch    db.Batch
+	walCount map[height]walIter
+}
+
+// NewTMDB creates a new TMDB instance implementing the TMDBInterface.
+func NewTMDB[V Hashable[H], H Hash, A Addr](db db.KeyValueStore, h height) TendermintDB[V, H, A] {
```
infrmtcs, line 92:
```suggestion
// NewTendermintDB creates a new tendermintDB instance implementing the TendermintDB interface.
func NewTendermintDB[V Hashable[H], H Hash, A Addr](db db.KeyValueStore, h height) TendermintDB[V, H, A] {
```

consensus/tendermint/tmdb.go
```diff
@@ -0,0 +1,311 @@
...
+//
+// We use two primary key prefixes based on db.BucketConsensus:
+//
+// 1. db.WALEntryCount: Stores the total number of messages (proposals, votes, timeouts)
+//    for a given height.
+//    - Key: db.WALEntryCount + <heightBytes>
+//    - Value: <countBytes>
+//    - Acts as a 1-based index for the individual message keys.
+//
+// 2. db.WALEntry: Stores the actual CBOR-encoded wrapped messages or timeouts.
```
infrmtcs, line 23:
```suggestion
// 1. db.WALEntry: Stores the actual CBOR-encoded wrapped messages or timeouts.
```

consensus/tendermint/tmdb.go
```diff
@@ -0,0 +1,311 @@
...
+	"fmt"
+
+	tmdb "github.com/NethermindEth/juno/consensus/db"
+	db "github.com/NethermindEth/juno/db"
+	"github.com/fxamacker/cbor/v2"
+)
+
+// Example DB layout for storing Tendermint WAL messages per height:
+//
+// We use two primary key prefixes based on db.BucketConsensus:
```
infrmtcs, line 15:
```suggestion
// We use a single primary key prefix based on db.BucketConsensus:
```

consensus/tendermint/tmdb.go
```diff
@@ -0,0 +1,311 @@
...
+//    - Value: CBOR-encoded wrapped Msg/Timeout
+//    - <iterBytes> ranges from 1 to the count stored under db.WALEntryCount.
+//
+// Both <heightBytes> and <iterBytes> are 4-byte big-endian uint32 representations.
+//
+// Batching:
+// We use a Batch (db.IndexedBatch) to accumulate writes before committing them to the DB.
+// This reduces expensive disk I/O. Reads check the batch first.
+// WARNING: If the process crashes before Commit(), buffered messages are lost,
+// risking consensus issues like equivocation (missing votes, double-signing).
```
infrmtcs, line 34: We need to update this part too

consensus/tendermint/tmdb.go
infrmtcs: This makes sense, I think we just need to rename it to `MessageTypeTimeoutTriggered` or `MessageTypeTimeoutProcessed` to differentiate it with the schedule timeout event.

consensus/tendermint/tmdb.go
infrmtcs: Agree, we can do it in the next PR.

consensus/tendermint/tmdb.go
infrmtcs: I meant we don't use `db.IndexedBatch` anymore.

consensus/tendermint/tmdb.go
```diff
@@ -0,0 +1,311 @@
...
+	})
+	if err != nil {
+		return nil, fmt.Errorf("scanWALRaw: db view error: %w", err)
+	}
+	return rawEntries, nil
+}
+
+// decodeWALMessageData decodes the inner CBOR message data based on the provided type string.
+// It assumes the outer wrapper has already been decoded.
+func decodeWALMessageData[V Hashable[H], H Hash, A Addr](msgType MessageType, data cbor.RawMessage) (walEntry, error) {
```
infrmtcs, line 276: I think this is basically the `UnmarshalCBOR` function in the `cbor.Unmarshaler` interface. I think we should convert this function to `UnmarshalCBOR` and use `cbor.Unmarshal` directly at the call site.

consensus/tendermint/tmdb.go
```diff
@@ -0,0 +1,311 @@
...
+		return "precommit"
+	case MessageTypeTimeout:
+		return "timeout"
+	default:
+		return "unknown"
+	}
+}
+
+// TendermintDB defines the methods for interacting with the Tendermint WAL database.
+type TendermintDB[V Hashable[H], H Hash, A Addr] interface {
```
infrmtcs, line 73: Do you think we should also move this file to `consensus/db`? My idea is to separate the code in packages: `driver`, `tendermint` (for state machine) and `db`, so it's easier to control the access and also easier to navigate inside the `tendermint` folder, which is currently having quite a lot of things.

consensus/tendermint/tmdb.go
```diff
@@ -0,0 +1,311 @@
...
+	heightBytes := make([]byte, NumBytesForHeight)
+	binary.BigEndian.PutUint32(heightBytes, uint32(height))
+	return heightBytes
+}
+
+func encodeNumMsgsAtHeight(numMsgsAtHeight walIter) []byte {
+	numMsgsAtHeightBytes := make([]byte, NumBytesForHeight)
+	binary.BigEndian.PutUint32(numMsgsAtHeightBytes, uint32(numMsgsAtHeight))
+	return numMsgsAtHeightBytes
+}
```
infrmtcs, line 311: Personally I think that we should have `Marshaler` and `Unmarshaler inferfaces for DB keys and the functions should be `MarshalKey` and `UnmarshalKey`. Not in this PR though.
For now, I think let's just have a consistent naming between these 2 functions.

consensus/tendermint/tmdb.go
```diff
@@ -0,0 +1,311 @@
...
+		defer iter.Close()
+
+		if !iter.Seek(startKey) {
+			// Changed error message slightly for clarity
+			return fmt.Errorf("failed to seek to start key when scanning WAL msgs")
+		}
+		if !iter.Valid() {
+			// It's possible there are no entries, return empty slice and no error
+			return nil
+		}
```
infrmtcs, line 250: I think we can remove this

consensus/tendermint/tmdb.go
rianhughes: Yup, in that case I think it makes sense to put this in the `/consensus/db/` pkg

consensus/tendermint/tmdb.go
rianhughes: I created an `unmarsahalCBOR` method for `walEntry` and renamed `decodeWALMessageData`

rianhughes: I'd rather not inline `decodeWALMessageData` though, I think the code is more readable with it

consensus/tendermint/tmdb.go
rianhughes: I see that you refactored a lot of these types in https://github.com/NethermindEth/juno/pull/2786. It might be easier and cleaner to refactor the `db` as part of / after that PR.

consensus/tendermint/tmdb.go
```diff
@@ -0,0 +1,283 @@
...
+	return walMsgs, nil
+}
+
+// scanWALRaw iterates over raw WAL entries in the database for a given height.
+func (s *tendermintDB[V, H, A]) scanWALRaw(height height) ([][]byte, error) {
+	startKey := tmdb.WALEntry.Key(encodeNumMsgsAtHeight(walIter(height)))
+	rawEntries := [][]byte{}
+
+	err := s.db.View(func(snap db.Snapshot) error {
+		iter, err := snap.NewIterator(startKey, false)
```
infrmtcs, line 240:
```suggestion
		iter, err := snap.NewIterator(startKey, true)
```

consensus/tendermint/tmdb.go
```diff
@@ -0,0 +1,283 @@
...
+			return nil, err
+		}
+		walMsgs[i] = walMsg
+	}
+	return walMsgs, nil
+}
+
+// scanWALRaw iterates over raw WAL entries in the database for a given height.
+func (s *tendermintDB[V, H, A]) scanWALRaw(height height) ([][]byte, error) {
+	startKey := tmdb.WALEntry.Key(encodeNumMsgsAtHeight(walIter(height)))
```
infrmtcs, line 236: I think this should be:
```suggestion
	prefix := tmdb.WALEntry.Key(encodeHeight(height))
```
because the bucket says:
```
	WALEntry BucketConsensus = iota // key: WAL_prefix + Height + MsgIndex. Val: Encoded Tendermint consensus message.
```

consensus/tendermint/tmdb.go
```diff
@@ -0,0 +1,283 @@
...
+		iter, err := snap.NewIterator(startKey, false)
+		if err != nil {
+			return err
+		}
+		defer iter.Close()
+
+		if !iter.Seek(startKey) {
+			// Changed error message slightly for clarity
+			return fmt.Errorf("failed to seek to start key when scanning WAL msgs")
+		}
```
infrmtcs, line 249: If we set the upper bound above, I think we can remove this because:
- The for loop won't run any iteration because the first key is already after the upper bound.
- I don't think this should fail when there is no messages, because it can be a possible normal case.

consensus/tendermint/tmdb.go
```diff
@@ -0,0 +1,283 @@
...
+	wrappedEntry, err := cbor.Marshal(wrapper)
+	if err != nil {
+		return fmt.Errorf("SetWALEntry: marshal wrapper failed: %w", err)
+	}
+
+	numMsgsAtHeight, ok := s.walCount[height]
+	if !ok {
+		numMsgsAtHeight = 0
+	}
+	nextNumMsgsAtHeight := numMsgsAtHeight + 1
```
infrmtcs, line 203: Personally I think it's more natural to have 0-based index here.

consensus/tendermint/tmdb.go
```diff
@@ -0,0 +1,283 @@
...
+	nextNumMsgsAtHeightBytes := encodeNumMsgsAtHeight(nextNumMsgsAtHeight)
+
+	msgKey := tmdb.WALEntry.Key(encodeHeight(height), nextNumMsgsAtHeightBytes)
```
infrmtcs, line 206: nit:
```suggestion
	msgKey := tmdb.WALEntry.Key(encodeHeight(height), encodeNumMsgsAtHeight(nextNumMsgsAtHeight))
```
or
```suggestion
	heightBytes := encodeHeight(height)
	nextNumMsgsAtHeightBytes := encodeNumMsgsAtHeight(nextNumMsgsAtHeight)

	msgKey := tmdb.WALEntry.Key(heightBytes, nextNumMsgsAtHeightBytes)
```

consensus/tendermint/tmdb.go
```diff
@@ -0,0 +1,283 @@
...
+	rawEntries, err := s.scanWALRaw(height)
+	if err != nil {
+		return nil, fmt.Errorf("GetWALMsgs: failed to scan raw WAL entries for height %d: %w", height, err)
+	}
+
+	walMsgs := make([]walEntry[V, H, A], len(rawEntries))
+
+	for i, rawEntry := range rawEntries {
+		var walMsg walEntry[V, H, A]
+		if err := cbor.Unmarshal(rawEntry, &walMsg); err != nil {
```
infrmtcs, line 226: I think this is slightly more efficient
```suggestion
		if err := cbor.Unmarshal(rawEntry, &walMsgs[i]); err != nil {
```

consensus/tendermint/tmdb.go
```diff
@@ -0,0 +1,283 @@
...
+			if !bytes.HasPrefix(key, startKey) {
+				break // Stop if we moved past keys for this height
+			}
+			v, err := iter.Value()
+			if err != nil {
+				return fmt.Errorf("scanWALRaw: failed to get value: %w", err)
+			}
+			valueCopy := make([]byte, len(v))
+			copy(valueCopy, v)
+			rawEntries = append(rawEntries, valueCopy)
```
infrmtcs, line 263: Why don't we unmarshal it here? I think it will save us `rawEntries` allocation.

consensus/tendermint/tmdb.go
```diff
@@ -0,0 +1,283 @@
...
+	if err := s.batch.Put(msgKey, wrappedEntry); err != nil {
+		return fmt.Errorf("writeWALEntryToBatch: failed to set MsgsAtHeight: %w", err)
+	}
+
+	s.walCount[height] = nextNumMsgsAtHeight
+	return nil
+}
+
+// GetWALMsgs implements TMDBInterface.
+func (s *tendermintDB[V, H, A]) GetWALMsgs(height height) ([]walEntry[V, H, A], error) {
```
infrmtcs, line 216: I think not this PR, but we should consider returning `iter.Seq` instead, so we can avoid allocating a slice.

consensus/tendermint/tmdb.go
rianhughes: Yup, now that the rest of the code has been simplified we can inline the get functions

consensus/tendermint/tmdb.go
rianhughes: `s.walCount[height]` needs to be one indexed, and using different bases for it and the key-suffix is confusing imo

consensus/tendermint/timeout.go
```diff
@@ -1,5 +1,16 @@
 package tendermint
 
+type timeout struct {
+	Step   step     `cbor:"s"`
+	Height height   `cbor:"h"`
+	Round  round    `cbor:"r"`
```
rodrigo-pino, line 6: I believe this `cbor` tags will be ignored since we have the `toarray`. I think it's safe to remove them

consensus/tendermint/tmdb.go
```diff
@@ -0,0 +1,251 @@
...
+	"fmt"
+
+	tmdb "github.com/NethermindEth/juno/consensus/db"
+	db "github.com/NethermindEth/juno/db"
+	"github.com/fxamacker/cbor/v2"
+)
+
+const NumBytesForHeight = 4
+
+type walIter uint32
```
rodrigo-pino, line 15: Could you add a comment explaining this internal type? I am not sure I follow it's purpose

consensus/tendermint/tmdb.go
rodrigo-pino, line 1: Why not call this field just `db`? It's implicit that it is tendemint database due to being part of the package

consensus/tendermint/tmdb.go
```diff
@@ -0,0 +1,251 @@
...
+	case MessageTypeProposal:
+		return "proposal"
+	case MessageTypePrevote:
+		return "prevote"
+	case MessageTypePrecommit:
+		return "precommit"
+	case MessageTypeTimeout:
+		return "timeout"
+	default:
+		return "unknown"
```
rodrigo-pino, line 85: Is it possible for the `default` flag to be used at any point. What do you think of panicking directly here?

consensus/tendermint/tmdb.go
```diff
@@ -0,0 +1,251 @@
...
+			if !bytes.HasPrefix(key, prefix) {
+				break
+			}
+			count++
+		}
+		return nil
+	})
+	if err != nil {
+		// Failing to retrieve the WAL msgs can result in a loss of funds (consensus slashing)
+		panic(fmt.Sprintf("getWALCount: failed to scan WAL entries for height %d: %v", height, err))
```
rodrigo-pino, line 160: Should we panic here. Shouldn't we gracefully handle the error, in case we want to implement a recovery mechanism later for example?

consensus/tendermint/tmdb_test.go
```diff
@@ -0,0 +1,159 @@
...
+	testHeight := height(2000)
+	testRound := round(5)
+	sender1 := *new(felt.Felt).SetUint64(101)
+	var val1 value = 99
+	valHash1 := val1.Hash()
+
+	// Add some messages for the test height using SetWALEntry
+	proposal := Proposal[value, felt.Felt, felt.Felt]{
+		MessageHeader: MessageHeader[felt.Felt]{Height: testHeight, Round: testRound, Sender: sender1},
+		Value:         utils.HeapPtr(val1),
```
rodrigo-pino, line 132: why not `&val1`?

rodrigo-pino (commented):
It looks good to me overall. I will give a final review once all comments has been addressed

consensus/tendermint/tmdb.go
infrmtcs: Don't forget to remove this :smile:

consensus/tendermint/tmdb.go
infrmtcs: Yeah, I agree that we should use a consistent base. Shouldn't we use 0-based index for both `walCount` and the key suffix? I'm thinking about something similar to transaction nonce: the next nonce is the current transaction count, in this case the next `walIter` is the current `walCount`.

consensus/tendermint/tmdb_test.go
```diff
@@ -0,0 +1,159 @@
...
+	"github.com/NethermindEth/juno/core/felt"
+	"github.com/NethermindEth/juno/db/pebble"
+	"github.com/NethermindEth/juno/utils"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// Test helper to get a TMDBInterface instance
+func newTestTMDB(t *testing.T) TendermintDB[value, felt.Felt, felt.Felt] {
+	dbPath := t.TempDir()
```
infrmtcs, line 16:
```suggestion
func newTestTMDB(t *testing.T) TendermintDB[value, felt.Felt, felt.Felt] {
	t.Helper()
	dbPath := t.TempDir()
```

consensus/tendermint/tmdb_test.go
```diff
@@ -0,0 +1,159 @@
...
+	// Assert Messages by Type
+	var (
+		proposalFound  *Proposal[value, felt.Felt, felt.Felt]
+		prevoteFound   *Prevote[felt.Felt, felt.Felt]
+		precommitFound *Precommit[felt.Felt, felt.Felt]
+		timeoutFound   *timeout
+	)
+
+	for _, entry := range retrievedEntries {
+		fmt.Println(entry)
```
infrmtcs, line 83: Remove

consensus/tendermint/tmdb_test.go
```diff
@@ -0,0 +1,159 @@
...
+		Height: testHeight,
+		Round:  testRound,
+		Step:   testStep,
+	}
+
+	// Store Messages using SetWALEntry
+	require.NoError(t, tmState.SetWALEntry(proposalMessage, testHeight))
+	require.NoError(t, tmState.SetWALEntry(prevoteMessage, testHeight))
+	require.NoError(t, tmState.SetWALEntry(precommitMessage, testHeight))
+	require.NoError(t, tmState.SetWALEntry(timeoutEvent, testHeight))
```
infrmtcs, line 64: All of these events has the `height`, so I think it's better to get the height from them instead of having an explicit argument for height in `SetWALEntry`.

consensus/tendermint/tmdb_test.go
```diff
@@ -0,0 +1,159 @@
...
+
+	// Commit the Batch
+	require.NoError(t, tmState.CommitBatch())
```
infrmtcs, line 67: I think we can group this using `t.Run`, then we can remove the `TestCommitBatch` above because that test doesn't commit any write.

consensus/tendermint/tmdb_test.go
```diff
@@ -0,0 +1,159 @@
...
+			found := (entry.Entry).(Precommit[felt.Felt, felt.Felt])
+			precommitFound = &found
+		case MessageTypeTimeout:
+			require.Nil(t, timeoutFound, "Found multiple timeouts")
+			msg := (entry.Entry).(timeout)
+			timeoutFound = &msg
+		default:
+			t.Fatalf("Found unexpected message type in WAL: %T", entry)
+		}
+	}
```
infrmtcs, line 105: Why don't we just assert retrievedEntries equals a slice of elements?

consensus/tendermint/tmdb_test.go
infrmtcs, line 16: I think this file should be shared between multiple tests, so we can write tests to assert that the writes are persistent

consensus/tendermint/tmdb_test.go
infrmtcs: I think we can have these subtests in this order:
- Write entries
- Commit batch and get entries (to check the written entries)
- Reload the db and get entries
- Delete entries
- Commit batch and get entries (to check if the entries are deleted)

consensus/tendermint/tmdb.go
rianhughes: I think using a zero index for a variable that counts the number entries is kind of confusing. Eg `s.walCount[height]=0` indicating that there is a single entry is misleading

consensus/tendermint/tmdb.go
rianhughes: If a message somehow gets stored that isn't one of the `MessageTypeProposal`,..`MessageTypeTimeout`, then we can entry this flow. I think it's worth handling explicitly, and I don't think we should panic here, but we will panic if an error is returned in `GetWALMsgs`

consensus/tendermint/tmdb.go
rianhughes: I think we should panic if we can't retrieve and replay the WAL messages given it could create a loss of funds for the user.

consensus/tendermint/tmdb.go
rianhughes: I renamed the var and added a comment

consensus/tendermint/tmdb_test.go
rianhughes: Yup, looks like we can safely remove the height arguement since we'll only every store msgs at the current height

consensus/tendermint/tmdb.go
rianhughes: Whoops, looks like it managed to sneak it's way back in

consensus/tendermint/tmdb_test.go
rianhughes: That works as well, updated

consensus/tendermint/tmdb_test.go
rianhughes: Updated

rianhughes:
Note: the snapshot logic was incorrect as it didn't implement a `close()` function, and so would leak. I had to update the snapshot interface and implement the closer. I don't think we need to support snapshots since we never really need to get a stable view of a dynamic state

consensus/tendermint/tmdb.go
infrmtcs: I think this one is also missing

infrmtcs: Probably the reason is because we can't provide UnmarshalCBOR for an interface.

consensus/tendermint/db_test.go
```diff
@@ -0,0 +1,100 @@
...
+		retrieved, err := tmState.GetWALMsgs(testHeight)
+		require.NoError(t, err)
+		require.ElementsMatch(t, expectedEntries, retrieved)
+	})
+
+	t.Run("Reload the db and get entries", func(t *testing.T) {
+		tmState, _ = reopenTestTMDB(t, db, dbPath, testHeight)
+		retrieved, err := tmState.GetWALMsgs(testHeight)
+		require.NoError(t, err)
+		require.ElementsMatch(t, expectedEntries, retrieved)
```
infrmtcs, line 88: I think it should be
```suggestion
		require.Equal(t, expectedEntries, retrieved)
```
Because we should assert that the elements are returned in the same order as the write order.

consensus/tendermint/tmdb.go
infrmtcs: How about something like:
```
	nextNumMsgsAtHeight := s.walCount[height]
	s.walCount[height]++
```

infrmtcs (approved):

rianhughes:
https://github.com/NethermindEth/juno/pull/2815

PR #2798, opened 2025-05-06
feat: Codegen proto from official p2p specs repo

infrmtcs:
- This PR configures `buf` to generate Go code directly from protobuf specs in https://github.com/starknet-io/starknet-p2p-specs with commit hash pinned, so we don't have to commit the proto files.
- This PR also creates a `gen` package containing aliases to facilitates the changes from old types to new types. Next step would be removing the `gen` package and use the generated type directly.

Notes:
- 95% of the changes are generated.
- Everything inside `p2p/proto` is generated by buf generate command (it's already in our Makefile, nothing needs to change in the build)
- Every existing code in `p2p/gen` is generated, I removed and write some "adapters" to minimize the changes.
- We need to re-generate the code because the current code is using a vendored version of old protobuf definitions. I changed the directory from `p2p/gen` to `p2p/proto` to reflect the structure inside the specs repo, which help reduce the friction.


adapters/core2p2p/receipt.go
```diff
@@ -128,12 +127,11 @@ func AdaptExecutionResources(er *core.ExecutionResources) *gen.Receipt_Execution
 			MulMod:       uint32(er.BuiltinInstanceCounter.MulMod),
 			RangeCheck96: uint32(er.BuiltinInstanceCounter.RangeCheck96),
 		},
-		Steps:          uint32(er.Steps),
-		MemoryHoles:    uint32(er.MemoryHoles),
-		L1Gas:          AdaptFelt(l1Gas),
-		L1DataGas:      AdaptFelt(l1DataGas),
-		TotalL1Gas:     AdaptFelt(totalL1Gas),
-		TotalL1DataGas: AdaptFelt(totalL1DataGas),
```
infrmtcs, line 136: Remove because `total_l1_data_gas = l1_data_gas` and `total_l1_data_gas` doesn't exist in specs.

adapters/core2p2p/transaction.go
```diff
@@ -86,17 +86,18 @@ func AdaptTransaction(transaction core.Transaction) *gen.Transaction {
 		case tx.Version.Is(3):
 			specTx.Txn = &gen.Transaction_DeclareV3_{
 				DeclareV3: &gen.Transaction_DeclareV3{
-					Sender:                    AdaptAddress(tx.SenderAddress),
-					Signature:                 AdaptAccountSignature(tx.Signature()),
-					ClassHash:                 AdaptHash(tx.ClassHash),
```
infrmtcs, line 91: `class_hash` was removed in this PR: https://github.com/starknet-io/starknet-p2p-specs/pull/54

buf.yaml
```diff
@@ -1,12 +0,0 @@
-version: v2
```
infrmtcs, line 1: Remove because we don't commit the proto here anymore.

p2p/gen/class.go
```diff
@@ -0,0 +1,12 @@
+package gen
```
infrmtcs, line 1: This package is an adapter to create aliases for types in new specs with so the call site doesn't change. Next step would be removing this package and use the new types directly.

rodrigo-pino: Can we add a todo in the code or an issue that reflect this intention, unless it's going to be part of an immediate follow up PR

rodrigo-pino (approved):
LGTM!

p2p/gen/class.go
infrmtcs: #2824 will be the follow-up PR, so I won't add a TODO here.

rianhughes (approved):

PR #2801, opened 2025-05-06
fix: mempool.LoadFromDB incorrect restore mempool from db

### LINKED ISSUE #2792: Bug: [Mempool] LoadFromDB incorrect restore from db
Bug in `mempool.LoadFromDB` function, while loading from DB function discards the head and appends the last element twice when mempool has more than one transaction.

In the following code when there is 2 tx in the mempool;

In first iteration head will be stored in the `newMemPoolTxn` then if `head.NextHash != nil`, `newMemPoolTxn` will be overridden by head.next, thus skipping head. In the second iteration of the loop
`curTxn.NextHash != nil` will be false and `newMemPoolTxn` will be `head.next` and `head.next` will be appended second time.
 
```go
// LoadFromDB restores the in-memory transaction pool from the database
func (p *Pool) LoadFromDB() error { 
	headVal, err := GetHeadValue(p.db)
	if err != nil {
			return nil
		}
		return err
	}

	currentHash := &headVal
	for currentHash != nil {
		curTxn, err := GetTxn(p.db, currentHash)
		if err != nil {
			return err
		}
		newMemPoolTxn := &memPoolTxn{Txn: curTxn.Txn}
		if curTxn.NextHash != nil {
			nextDBTxn, err := GetTxn(p.db, curTxn.NextHash)
			if err != nil {
				return err
			}
			newMemPoolTxn = &memPoolTxn{Txn: nextDBTxn.Txn}
		}
		p.memTxnList.push(newMemPoolTxn)
		currentHash = curTxn.NextHash
	}

	return nil
}
```

Potential fix

```go
// LoadFromDB restores the in-memory transaction pool from the database
func (p *Pool) LoadFromDB() error { 
	headVal, err := GetHeadValue(p.db)
	if err != nil {
			return nil
		}
		return err
	}

	currentHash := &headVal
	for currentHash != nil {
		curTxn, err := GetTxn(p.db, currentHash)
		if err != nil {
			return err
		}
		newMemPoolTxn := &memPoolTxn{Txn: curTxn.Txn}
		p.memTxnList.push(newMemPoolTxn)
		currentHash = curTxn.NextHash
	}

	return nil
}
```

EgeCaner:
This PR addresses the incorrect mempool restoration from DB.
Enhanced `TestRestoreMempool` test with deep equality check for `expected transactions` and `restored transactions`

Close #2792

### DIFF
--- a/mempool/mempool.go
+++ b/mempool/mempool.go
@@ -167,15 +167,7 @@ func (p *Pool) LoadFromDB() error {
 		if err != nil {
 			return err
 		}
-		newMemPoolTxn := &memPoolTxn{Txn: curTxn.Txn}
-		if curTxn.NextHash != nil {
-			nextDBTxn, err := GetTxn(p.db, curTxn.NextHash)
-			if err != nil {
-				return err
-			}
-			newMemPoolTxn = &memPoolTxn{Txn: nextDBTxn.Txn}
-		}
-		p.memTxnList.push(newMemPoolTxn)
+		p.memTxnList.push(&memPoolTxn{Txn: curTxn.Txn})
 		currentHash = curTxn.NextHash
 	}
 
--- a/mempool/mempool_test.go
+++ b/mempool/mempool_test.go
@@ -176,7 +176,7 @@ func TestRestoreMempool(t *testing.T) {
 	restoredTransactions, err := poolRestored.PopBatch(lenRestored)
 	require.NoError(t, err)
 	for i, txn := range restoredTransactions {
-		require.Equal(t, txn.Transaction.Hash(), expectedTxs[i].Transaction.Hash())
+		require.Equal(t, expectedTxs[i].Transaction.Hash(), txn.Transaction.Hash())
 	}
 	lenDB, err = poolRestored.LenDB()
 	require.NoError(t, err)
--- a/mempool/mempool_test.go
+++ b/mempool/mempool_test.go
@@ -135,18 +135,21 @@ func TestRestoreMempool(t *testing.T) {
 	require.Equal(t, 0, pool.Len())
 
 	// push multiple transactions to empty mempool (1,2,3)
+	transactions := make([]mempool.BroadcastedTransaction, 3)
 	for i := uint64(1); i < 4; i++ {
 		senderAddress := new(felt.Felt).SetUint64(i)
 		chain.EXPECT().HeadState().Return(state, func() error { return nil }, nil)
 		state.EXPECT().ContractNonce(senderAddress).Return(new(felt.Felt).SetUint64(0), nil)
-		require.NoError(t, pool.Push(&mempool.BroadcastedTransaction{
+		transaction := mempool.BroadcastedTransaction{
 			Transaction: &core.InvokeTransaction{
 				TransactionHash: new(felt.Felt).SetUint64(i),
 				Version:         new(core.TransactionVersion).SetUint64(1),
 				SenderAddress:   senderAddress,
 				Nonce:           new(felt.Felt).SetUint64(0),
 			},
-		}))
+		}
+		require.NoError(t, pool.Push(&transaction))
+		transactions[i-1] = transaction
 		require.Equal(t, int(i), pool.Len())
 	}
 	// check the db has stored the transactions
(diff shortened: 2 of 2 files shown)


mempool/mempool_test.go
```diff
@@ -135,18 +135,21 @@ func TestRestoreMempool(t *testing.T) {
 	require.Equal(t, 0, pool.Len())
 
 	// push multiple transactions to empty mempool (1,2,3)
+	transactions := make([]mempool.BroadcastedTransaction, 3)
 	for i := uint64(1); i < 4; i++ {
 		senderAddress := new(felt.Felt).SetUint64(i)
 		chain.EXPECT().HeadState().Return(state, func() error { return nil }, nil)
 		state.EXPECT().ContractNonce(senderAddress).Return(new(felt.Felt).SetUint64(0), nil)
-		require.NoError(t, pool.Push(&mempool.BroadcastedTransaction{
+		transaction := mempool.BroadcastedTransaction{
```
rodrigo-pino, line 143: Nitpick: Could this one be called `tx`? The difference of just an `s` between two vars with such a long name makes it a bit hard to differentiate

mempool/mempool_test.go
```diff
@@ -135,18 +135,21 @@ func TestRestoreMempool(t *testing.T) {
 	require.Equal(t, 0, pool.Len())
 
 	// push multiple transactions to empty mempool (1,2,3)
+	transactions := make([]mempool.BroadcastedTransaction, 3)
```
rodrigo-pino, line 138: Also a nitpick, this could be named `expectedTxs` which let the reader know this value is going to be used for test comparison

mempool/mempool_test.go
```diff
@@ -166,17 +169,20 @@ func TestRestoreMempool(t *testing.T) {
 	lenDB, err = poolRestored.LenDB()
 	require.NoError(t, err)
 	require.Equal(t, 3, lenDB)
-	require.Equal(t, 3, poolRestored.Len())
+	lenRestored := poolRestored.Len()
+	require.Equal(t, 3, lenRestored)
```
rodrigo-pino, line 173: I believe this equality might be redundant. You are already checking the length with `transactions` below which is created with size three. You could make it a static array if you want to guarantee it will never grow.

mempool/mempool.go
```diff
@@ -167,15 +167,7 @@ func (p *Pool) LoadFromDB() error {
...
-		newMemPoolTxn := &memPoolTxn{Txn: curTxn.Txn}
-		if curTxn.NextHash != nil {
-			nextDBTxn, err := GetTxn(p.db, curTxn.NextHash)
-			if err != nil {
-				return err
-			}
-			newMemPoolTxn = &memPoolTxn{Txn: nextDBTxn.Txn}
-		}
-		p.memTxnList.push(newMemPoolTxn)
+		p.memTxnList.push(&memPoolTxn{Txn: curTxn.Txn})
```
rodrigo-pino, line 170: Just to be clear I understand the deleted code was a bug because the first transaction from the DB was always being skipped

rodrigo-pino (approved):
Overall, it looks good to me. Approving now and leaving @rianhughes to have the finaly say

mempool/mempool.go
rianhughes: The bug was that 1. we skipped the first transaction and 2. stored the last transaction twice

rianhughes (approved):

mempool/mempool_test.go
```diff
@@ -166,17 +169,19 @@ func TestRestoreMempool(t *testing.T) {
...
 
-	// Remove transactions
-	_, err = poolRestored.Pop()
-	require.NoError(t, err)
-	_, err = poolRestored.Pop()
+	// Pop transactions
+	restoredTransactions, err := poolRestored.PopBatch(lenRestored)
 	require.NoError(t, err)
+	for i, txn := range restoredTransactions {
+		require.Equal(t, txn.Transaction.Hash(), expectedTxs[i].Transaction.Hash())
```
rodrigo-pino, line 179: Final nitpick, the expected value should be on the 2nd parameter

rodrigo-pino (approved):

PR #2817, opened 2025-05-12
refactor: push common consensus types to types sub-pkg

rianhughes:
This PR re-factors the existing `consensus` pkg by pushing common types to the `types` pkg. There is no change in logic. Some fields were exported (mainly message fields, and renamed the `Height` function to `GetHeight` to prevent name collision with field).


consensus/types/timeout.go
```diff
@@ -0,0 +1,16 @@
...
+	_      struct{} `cbor:",toarray"`
+}
+
+func (t Timeout) MsgType() MessageType {
+	return MessageTypeTimeout
+}
+
+func (t Timeout) GetHeight() Height {
+	return t.Height
+}
```
rodrigo-pino, line 16: Why having a public field and a getter at the same time 🤔. Is it to implement an interface?

rodrigo-pino (approved):
Quick approval. Leaving the final say to @infrmtcs

consensus/types/timeout.go
rianhughes: Yup, there's an `IsWalMsg` interface which requires a method to get the height

consensus/tendermint/tendermint.go
```diff
@@ -273,3 +233,24 @@ func (t *stateMachine[V, H, A]) findProposal(r Round) *CachedProposal[V, H, A] {
...
+	}
+	return nil
+}
+
+func (t *stateMachine[V, H, A]) onTimeoutPrecommit(h types.Height, r types.Round) Action[V, H, A] {
+	if t.state.height == h && t.state.round == r {
+		return t.startRound(r + 1)
+	}
+	return nil
+}
```
infrmtcs, line 256: I think these lines should still be in a separate `timeout.go` file. Currently `tendermint.go` only contains mostly utils functions.

consensus/tendermint/rule_proposal_and_polka_current_test.go
```diff
@@ -16,19 +17,19 @@ func TestProposalAndPolkaCurrent(t *testing.T) {
 		// Initialise the round
 		currentRound.start()
 
-		// Prevotes are collected
+		// types.Prevotes are collected
```
infrmtcs, line 20: nit:
```suggestion
		// Prevotes are collected
```

infrmtcs (approved):

PR #2828, opened 2025-05-14
feature: replay consensus wal messages

rianhughes:
This PR implements the consensus replay mechanism. 

I followed malachites implementation to determine things like, when we should write a wal msg, and when we should commit the wal batch to disk. I also had to move the `action` type to the `types` pkg to resolve an import cycle error. Howeever, I left the types that implement this interface in the `tendermint` pkg.

Note: there is a single todo that I would like to address before merging this (weird behaviour in one of the tests), but I think we can start reviewing it in the meantime.


consensus/db/db.go
```diff
@@ -60,11 +60,26 @@ func (w *walEntry[V, H, A]) UnmarshalCBOR(data []byte) error {
...
+//
+// We call Delete when we start a new height and commit a block
 type TendermintDB[V types.Hashable[H], H types.Hash, A types.Addr] interface {
-	// CommitBatch writes the accumulated batch operations to the underlying database.
-	CommitBatch() error
+	// FlushWAL writes the accumulated batch operations to the underlying database.
+	FlushWAL() error
 	// GetWALMsgs retrieves all WAL messages (consensus messages and timeouts) stored for a given height from the database.
-	GetWALMsgs(height types.Height) ([]walEntry[V, H, A], error)
+	GetWALMsgs(height types.Height) ([]WalEntry[V, H, A], error)
```
infrmtcs, line 82: I'm wondering if we need to access `WalEntry` from other packages. What do you think if we keep it private and extract the message from `walEntry`?
```suggestion
	GetWALMsgs(height types.Height) ([]Message[V, H, A], error)
```

infrmtcs: Nevermind, I forgot the timeout case.

consensus/db/db.go
```diff
@@ -60,11 +60,26 @@ func (w *walEntry[V, H, A]) UnmarshalCBOR(data []byte) error {
...
+// 1. We start a new round
+// 2. Right before we propose a block
+// 3. When consensus is reached
+//
+// We call Delete when we start a new height and commit a block
 type TendermintDB[V types.Hashable[H], H types.Hash, A types.Addr] interface {
-	// CommitBatch writes the accumulated batch operations to the underlying database.
-	CommitBatch() error
+	// FlushWAL writes the accumulated batch operations to the underlying database.
+	FlushWAL() error
```
infrmtcs, line 80: nit: these are methods of `TendermintDB`, so I think `Flush` and `GetEntries` should be enough.

consensus/tendermint/action.go
```diff
@@ -14,10 +10,10 @@ type BroadcastPrecommit[H types.Hash, A types.Addr] types.Precommit[H, A]
 
 type ScheduleTimeout types.Timeout
 
-func (a *BroadcastProposal[V, H, A]) isTendermintAction() {}
+func (a *BroadcastProposal[V, H, A]) IsTendermintAction() {}
```
infrmtcs, line 13: We should move all of them. `isTendermintAction` needs to be private in order to seal the interface.

consensus/types/action.go
```diff
@@ -0,0 +1,21 @@
+package types
+
+type Action[V Hashable[H], H Hash, A Addr] interface {
+	IsTendermintAction()
```
infrmtcs, line 4: We should make it private again.

consensus/tendermint/state_machine_context_test.go
```diff
@@ -11,7 +11,7 @@ import (
 // Sample usages:
 //
 //	// Create a new state machine with 4 validators and this node is the validator with index 3 (0-based).
-//	sm := setupStateMachine(t, 4, 3)
+//	sm := setupStateMachine(t, 4, 3,true)
```
infrmtcs, line 14:
```suggestion
//	sm := setupStateMachine(t, 4, 3)
```

consensus/tendermint/action.go
```diff
@@ -2,22 +2,10 @@ package tendermint
 
 import "github.com/NethermindEth/juno/consensus/types"
 
-type Action[V types.Hashable[H], H types.Hash, A types.Addr] interface {
```
infrmtcs, line 5: We should remove the whole file

consensus/db/db.go
```diff
@@ -60,15 +60,32 @@ func (w *walEntry[V, H, A]) UnmarshalCBOR(data []byte) error {
 }
 
 // TendermintDB defines the methods for interacting with the Tendermint WAL database.
+//
+// The purpose of the WAL is to record any event that may result in a state change.
+// These events fall into the following categories:
+// 1. Incoming messages. We do not need to store outgoing messages.
```
infrmtcs, line 66: We actually do need to store outgoing messages, otherwise the node can double vote.

infrmtcs: Ah, maybe I get what you mean, because the state machine should be deterministic, so it will always result in the same vote, except for proposal where it calls the `Value()` function which can be non-deterministic.

consensus/db/db.go
```diff
@@ -60,15 +60,32 @@ func (w *walEntry[V, H, A]) UnmarshalCBOR(data []byte) error {
...
+// 2. When we propose a value.
+// 3. When a timeout is triggered (not when it is scheduled).
+// The purpose of the WAL is to allow the node to recover the state it was in before the crash.
+// No new messages should be broadcast during replay.
+//
+// We commit the WAL to disk when:
+// 1. We start a new round
+// 2. Right before we propose a block
+// 3. When consensus is reached
```
infrmtcs, line 75: I think these can be summarized as:
```suggestion
// 2. Right before we broadcast a message
```

consensus/tendermint/tendermint_test.go
```diff
@@ -98,8 +101,13 @@ func setupStateMachine(
 	}
 
 	thisNodeAddr := getVal(thisValidator)
-
-	return New(*thisNodeAddr, app, chain, vals).(*stateMachine[value, felt.Felt, felt.Felt])
+	ctrl := gomock.NewController(t)
```
infrmtcs, line 104: I think we should avoid using mock, but since it's not trivial, and we should move the DB usage to driver anyway, so I think it's fine for now.

consensus/tendermint/rule_commit_value.go
```diff
@@ -26,11 +28,19 @@ func (t *stateMachine[V, H, A]) uponCommitValue(cachedProposal *CachedProposal[V
...
+func (t *stateMachine[V, H, A]) doCommitValue(cachedProposal *CachedProposal[V, H, A]) types.Action[V, H, A] {
+	if err := t.db.Flush(); err != nil {
+		t.log.Fatalf("failed to flush WAL during commit", "height", cachedProposal.Height, "round", cachedProposal.Round, "err", err)
+	}
+
 	// TODO: Optimise this
 	precommits, _ := t.checkForQuorumPrecommit(cachedProposal.Round, *cachedProposal.ID)
 	t.blockchain.Commit(t.state.height, *cachedProposal.Value, precommits)
 
+	if err := t.db.DeleteWALEntries(t.state.height); err != nil {
```
infrmtcs, line 40: Maybe not this PR, but we should reconsider deleting entries, or at least keep the precommits, in order to create the proof of commit.
https://github.com/starknet-io/starknet-p2p-specs/blob/main/p2p/proto/sync/header.proto#L31

consensus/tendermint/tendermint.go
```diff
@@ -214,3 +234,51 @@ func (t *stateMachine[V, H, A]) findProposal(r types.Round) *CachedProposal[V, H
...
+// from persistent storage and applies them to the internal state. This is used for
+// recovering consensus state after a crash or restart.
+//
+// ReplayWAL must not trigger any external effects such as broadcasting messages or
+// scheduling timeouts. It is strictly a state recovery mechanism.
+//
+// Panics if the replaying the messages fails for whatever reason.
+func (t *stateMachine[V, H, A]) ReplayWAL() {
+	height := t.blockchain.Height()
+	walEntries, err := t.db.GetWALEntries(height)
```
infrmtcs, line 248: The current implementation only replays once at startup, and on the current height. This means any received messages in future height won't replay.
I think we should instead replay `t.state.height` and do it every time we start a new height.

consensus/tendermint/tendermint_test.go
rianhughes: I agree

consensus/tendermint/rule_commit_value.go
rianhughes: Like they'll be a field in the block header right?

consensus/tendermint/tendermint.go
```diff
@@ -214,3 +234,51 @@ func (t *stateMachine[V, H, A]) findProposal(r types.Round) *CachedProposal[V, H
...
 }
+
+// ReplayWAL replays all WAL (Write-Ahead Log) messages for the current block height
+// from persistent storage and applies them to the internal state. This is used for
+// recovering consensus state after a crash or restart.
+//
+// ReplayWAL must not trigger any external effects such as broadcasting messages or
+// scheduling timeouts. It is strictly a state recovery mechanism.
+//
+// Panics if the replaying the messages fails for whatever reason.
```
rodrigo-pino, line 245: Is it ok to panic and why? This means we have a corrupted WAL, how do users recuperate from that. We should at least give some instruction, otherwise, each time the node is restarted it will panic forever

consensus/db/db.go
```diff
@@ -60,15 +60,31 @@ func (w *walEntry[V, H, A]) UnmarshalCBOR(data []byte) error {
...
 
 // TendermintDB defines the methods for interacting with the Tendermint WAL database.
+//
+// The purpose of the WAL is to record any event that may result in a state change.
+// These events fall into the following categories:
+// 1. Incoming messages. We do not need to store outgoing messages.
+// 2. When we propose a value.
+// 3. When a timeout is triggered (not when it is scheduled).
+// The purpose of the WAL is to allow the node to recover the state it was in before the crash.
+// No new messages should be broadcast during replay.
```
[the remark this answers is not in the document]
infrmtcs: I'm also shocked now: https://www.merriam-webster.com/grammar/is-it-broadcast-or-broadcasted-forecast-or-forecasted-usage

consensus/db/db.go
```diff
@@ -60,15 +60,31 @@ func (w *walEntry[V, H, A]) UnmarshalCBOR(data []byte) error {
...
-	// CommitBatch writes the accumulated batch operations to the underlying database.
-	CommitBatch() error
-	// GetWALMsgs retrieves all WAL messages (consensus messages and timeouts) stored for a given height from the database.
-	GetWALMsgs(height types.Height) ([]walEntry[V, H, A], error)
+	// Flush writes the accumulated batch operations to the underlying database.
+	Flush() error
+	// GetWALEntries retrieves all WAL messages (consensus messages and timeouts) stored for a given height from the database.
+	GetWALEntries(height types.Height) ([]WalEntry[V, H, A], error)
 	// SetWALEntry schedules the storage of a WAL message in the batch.
 	SetWALEntry(entry types.Message[V, H, A]) error
```
infrmtcs, line 85: I just realize another issue, `walCount` is not set properly for future height after restart.

consensus/tendermint/wal_test.go
```diff
@@ -0,0 +1,186 @@
...
+			Height: types.Height(0),
+			Round:  types.Round(0),
+			Sender: *proposerAddr,
+		},
+		ValidRound: types.Round(-1),
+		Value:      &proposedValue,
+	}
+
+	t.Run("ReplayWAL: replay on empty db", func(t *testing.T) {
+		mockDB := mocks.NewMockTendermintDB[value, felt.Felt, felt.Felt](ctrl)
```
infrmtcs, line 57: I think in this specific case, we should test against a real DB. The test cases can be in a simple format:
Do something -> Crash -> Restart -> Assert if the new state and the actions are the same as the old ones.

consensus/tendermint/tendermint.go
rianhughes: Yes, we should panic if we can't recover the state from before the crash. The reason is that it can result in equivocation and loss of funds which should be avoided. As for how to address this case, it ill likely depend on a case by case basis. One solution would be to delete the WAL db and wait for the network to progress a block or two and fallback to the sync mechanism so we never broadcast a message for the old height

infrmtcs (approved):
The comments can be addressed in another PR.

consensus/db/db.go
rianhughes: I've restricted the wal to only store the msgs/timeouts at the current height. All msgs from future/past heights are (safely) ignored.

consensus/tendermint/tendermint.go
rianhughes: I've restricted the wal to only store the msgs/timeouts at the current height. All msgs from future/past heights are (safely) ignored.

rodrigo-pino (approved):

PR #2829, opened 2025-05-14
refactor: update block id

rodrigo-pino:
Make interacting with Block ID safer by making it's data private and only interacting with it via getter functions.

Also, refactored the fields inside. The struct grows in size but is capable of holding all the required information inside itself. No external references.

We can make it smaller by using reference and `unsafe` code but I am not sure if we should go there (specifically for this type)


rpc/v8/block.go
```diff
@@ -3,56 +3,125 @@ package rpcv8
...
+	}
+}
+
 // https://github.com/starkware-libs/starknet-specs/blob/a789ccc3432c57777beceaa53a34a7ae2f25fda0/api/starknet_api_openrpc.json#L814
 type BlockID struct {
-	Pending bool
-	Latest  bool
-	Hash    *felt.Felt
-	Number  uint64
+	typeID blockIDType
```
infrmtcs, line 53: Why don't we have 4 types implementing `BlockIdentifier`? Something like
```
type PendingBlock struct{}
type LatestBlock struct{}
type BlockByHash felt.Felt
type BlockByNumber uint64
```

infrmtcs: Some rationale:
- We don't allocate `data` if not necessary.
- `Hash()` or `Number()` are unsafe functions, so we need to make sure that we only call them in correct cases. If we can instead extract them directly from the switch cases, I think it's safer. Something like:
```
switch blockId := blockId.(type) {
case BlockByHash:
  hash := felt.Felt(blockId)
...
```

infrmtcs: If we keep the current `BlockID` implementation, I think we can remove `BlockIdentifier`, because there is only 1 implementation and we don't hide it anyway.

rpc/v8/block.go
```diff
@@ -3,56 +3,125 @@ package rpcv8
...
 
 	"github.com/NethermindEth/juno/core"
 	"github.com/NethermindEth/juno/core/felt"
 	"github.com/NethermindEth/juno/jsonrpc"
 	rpcv6 "github.com/NethermindEth/juno/rpc/v6"
 )
 
+var _ BlockIdentifier = (*BlockID)(nil)
+
+type BlockIdentifier interface {
```
infrmtcs, line 16: We can consider remove this and use `BlockID` directly

PR #2832, opened 2025-05-16
test

rianhughes:
do not merge.

This is a test branch to identify the root cause of the broken CICD


rianhughes:
Tests broke because of rust dep

PR #2838, opened 2025-05-19
refactor: create a package for real starknet types used in tendermint

infrmtcs:
This `consensus/starknet` package defines the aliases for concrete types to be used in production.

### DIFF
--- a/consensus/starknet/starknet.go
+++ b/consensus/starknet/starknet.go
@@ -1,0 +1,33 @@
+package starknet
+
+import (
+	"github.com/NethermindEth/juno/consensus/types"
+	"github.com/NethermindEth/juno/core/address"
+	"github.com/NethermindEth/juno/core/felt"
+	"github.com/NethermindEth/juno/core/hash"
+)
+
+// TODO: Replace with the actual value type.
+type Value uint64
+
+func (v Value) Hash() hash.Hash {
+	return hash.Hash(*new(felt.Felt).SetUint64(uint64(v)))
+}
+
+type (
+	Address = address.Address
+	Hash    = hash.Hash
+
+	Message       = types.Message[Value, Hash, Address]
+	Proposal      = types.Proposal[Value, Hash, Address]
+	Prevote       = types.Prevote[Hash, Address]
+	Precommit     = types.Precommit[Hash, Address]
+	Vote          = types.Vote[Hash, Address]
+	MessageHeader = types.MessageHeader[Address]
+	Messages      = types.Messages[Value, Hash, Address]
+
+	Action             = types.Action[Value, Hash, Address]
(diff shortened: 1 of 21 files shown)


consensus/types/types.go
```diff
@@ -1,14 +1,11 @@
...
-
 type Addr interface {
-	// Ethereum Addresses are 20 bytes
-	~[20]byte | felt.Felt
+	~[4]uint64
 }
 
 type Hash interface {
-	~[32]byte | felt.Felt
+	~[4]uint64
```
rianhughes, line 8: Also, what do you think about replacing the interfaces with an `Alias`? It should simplify the function definitions, and I'm not sure what we gain by using the interface instead of an alias?

consensus/types/types.go
```diff
@@ -1,14 +1,11 @@
 package types
 
-import "github.com/NethermindEth/juno/core/felt"
-
 type Addr interface {
-	// Ethereum Addresses are 20 bytes
-	~[20]byte | felt.Felt
+	~[4]uint64
```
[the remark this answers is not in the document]
infrmtcs: This interface is only for the consensus packages, so I don't think it affects other parts of the code base.

consensus/types/types.go
infrmtcs: I think for now, the only requirement for Hash is `comparable`, because it needs to be able to be map key. I think the original intention of this interface is to be able to use different implementation of hash if necessary.

consensus/starknet/starknet.go
```diff
@@ -0,0 +1,33 @@
...
+	"github.com/NethermindEth/juno/core/hash"
+)
+
+// TODO: Replace with the actual value type.
+type Value uint64
+
+func (v Value) Hash() hash.Hash {
+	return hash.Hash(*new(felt.Felt).SetUint64(uint64(v)))
+}
+
```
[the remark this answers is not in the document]
infrmtcs: I think it will still be 1 single struct, which is the end result of the state machine.
https://github.com/NethermindEth/juno/pull/2839/files#diff-3311dd8d967035669c7e84986837bfa7492f240b682937e81b1ee8cbf7e11b3aR65

rianhughes (approved):

consensus/types/types.go
rodrigo-pino: Why not using the Hash types from the general `type` package?. Why making it an interface and not `type Hash felt.Felt`? It makes conversion one way or another very easy

infrmtcs: This interface is for the constraint of the type argument of the generic types. The concrete type is still the same Hash type you mentioned.

rodrigo-pino (approved):

PR #2839, opened 2025-05-19
feat: Add scaffolding for proposal state machine

infrmtcs:
```mermaid
stateDiagram-v2
    InitialState --> ReceivingBlockInfoState: OnInit(ProposalInit)
    ReceivingBlockInfoState --> ReceivingHashState: OnEmptyBlock(ProposalCommitment)
    ReceivingBlockInfoState --> ReceivingTransactionsState: OnBlockInfo(BlockInfo)
    ReceivingTransactionsState --> ReceivingTransactionsState: OnTransactions([]ConsensusTransaction)
    ReceivingTransactionsState --> ReceivingHashState: OnCommitment(ProposalCommitment)
    ReceivingHashState --> FinState: OnFin(ProposalFin)
```

### DIFF
--- a/consensus/p2p/validator/state_machine_test.go
+++ b/consensus/p2p/validator/state_machine_test.go
@@ -1,0 +1,245 @@
+package validator_test
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/NethermindEth/juno/consensus/p2p/validator"
+	"github.com/NethermindEth/juno/consensus/starknet"
+	"github.com/NethermindEth/juno/consensus/types"
+	"github.com/NethermindEth/juno/core/felt"
+	"github.com/starknet-io/starknet-p2pspecs/p2p/proto/common"
+	"github.com/starknet-io/starknet-p2pspecs/p2p/proto/consensus/consensus"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+type validTransitionTestStep struct {
+	event     *consensus.ProposalPart
+	postState validator.ProposalStateMachine
+}
+
+func TestProposalStateMachine_ValidTranstions(t *testing.T) {
+	height := uint64(1337)
+	round := uint32(1338)
+	validRound := types.Round(-1)
+	proposer := validator.GetRandomAddress(t)
+
+	expectedHeader := &starknet.MessageHeader{
+		Height: types.Height(height),
+		Round:  types.Round(round),
+		Sender: starknet.Address(*new(felt.Felt).SetBytes(proposer.Elements)),
+	}
+	t.Run("valid proposal stream", func(t *testing.T) {
+		value := uint64(1339)
+		starknetValue := starknet.Value(value)
+		expectedHash := &common.Hash{Elements: validator.ToBytes(*new(felt.Felt).SetUint64(value))}
+		expectedProposal := &starknet.Proposal{
+			MessageHeader: *expectedHeader,
+			ValidRound:    validRound,
+			Value:         &starknetValue,
+		}
+
+		runTestSteps(t, []validTransitionTestStep{
+			{
+				event: &consensus.ProposalPart{
+					Messages: &consensus.ProposalPart_Init{Init: &consensus.ProposalInit{
+						BlockNumber: height,
+						Round:       round,
+						Proposer:    proposer,
+					}},
+				},
+				postState: &validator.AwaitingBlockInfoOrCommitmentState{
+					Header:     expectedHeader,
+					ValidRound: validRound,
+				},
+			},
+			{
+				event: &consensus.ProposalPart{
+					Messages: &consensus.ProposalPart_BlockInfo{BlockInfo: &consensus.BlockInfo{
+						BlockNumber: height,
+						Timestamp:   value,
+					}},
+				},
+				postState: &validator.ReceivingTransactionsState{
+					Header:     expectedHeader,
+					ValidRound: validRound,
+					Value:      &starknetValue,
+				},
+			},
+			{
+				event: &consensus.ProposalPart{
+					Messages: &consensus.ProposalPart_Transactions{Transactions: &consensus.TransactionBatch{
+						Transactions: validator.GetRandomTransactions(t, 4),
+					}},
+				},
+				postState: &validator.ReceivingTransactionsState{
+					Header:     expectedHeader,
+					ValidRound: validRound,
+					Value:      &starknetValue,
+				},
+			},
+			{
+				event: &consensus.ProposalPart{
+					Messages: &consensus.ProposalPart_Transactions{Transactions: &consensus.TransactionBatch{
+						Transactions: validator.GetRandomTransactions(t, 5),
+					}},
+				},
+				postState: &validator.ReceivingTransactionsState{
+					Header:     expectedHeader,
+					ValidRound: validRound,
+					Value:      &starknetValue,
+				},
+			},
+			{
+				event: &consensus.ProposalPart{
+					Messages: &consensus.ProposalPart_Commitment{
+						Commitment: &consensus.ProposalCommitment{
+							BlockNumber: height,
+							Timestamp:   value,
+						},
+					},
+				},
+				postState: &validator.AwaitingProposalFinState{
+					Header:     expectedHeader,
+					ValidRound: validRound,
+					Value:      &starknetValue,
+				},
+			},
+			{
+				event: &consensus.ProposalPart{
+					Messages: &consensus.ProposalPart_Fin{
+						Fin: &consensus.ProposalFin{
+							ProposalCommitment: expectedHash,
+						},
+					},
+				},
+				postState: (*validator.FinState)(expectedProposal),
+			},
+		})
+	})
+
+	t.Run("empty proposal stream", func(t *testing.T) {
+		expectedProposal := &starknet.Proposal{
+			MessageHeader: *expectedHeader,
+			ValidRound:    validRound,
+		}
+		runTestSteps(t, []validTransitionTestStep{
+			{
+				event: &consensus.ProposalPart{
+					Messages: &consensus.ProposalPart_Init{Init: &consensus.ProposalInit{
+						BlockNumber: height,
+						Round:       round,
+						Proposer:    proposer,
+					}},
+				},
+				postState: &validator.AwaitingBlockInfoOrCommitmentState{
+					Header:     expectedHeader,
+					ValidRound: validRound,
+				},
+			},
+			{
+				event: &consensus.ProposalPart{
+					Messages: &consensus.ProposalPart_Commitment{
+						Commitment: &consensus.ProposalCommitment{
+							BlockNumber: 1337,
+							Timestamp:   1338,
+						},
+					},
+				},
+				postState: &validator.AwaitingProposalFinState{
+					Header:     expectedHeader,
+					ValidRound: validRound,
+				},
+			},
+			{
+				event: &consensus.ProposalPart{
+					Messages: &consensus.ProposalPart_Fin{
+						Fin: &consensus.ProposalFin{},
+					},
+				},
+				postState: (*validator.FinState)(expectedProposal),
+			},
+		})
+	})
+}
+
+func runTestSteps(t *testing.T, steps []validTransitionTestStep) {
+	var err error
+	transition := validator.NewTransition()
+	var state validator.ProposalStateMachine = &validator.InitialState{}
+
+	for _, step := range steps {
+		t.Run(fmt.Sprintf("apply %T to get %T", step.event.Messages, step.postState), func(t *testing.T) {
+			state, err = state.OnEvent(t.Context(), transition, step.event)
+			require.NoError(t, err)
+			assert.Equal(t, step.postState, state)
+		})
+	}
+}
+
+type invalidTransitionTestCase struct {
+	state  validator.ProposalStateMachine
+	events []*consensus.ProposalPart
+}
+
+func TestProposalStateMachine_InvalidTransitions(t *testing.T) {
+	steps := []invalidTransitionTestCase{
+		{
+			state: &validator.InitialState{},
+			events: []*consensus.ProposalPart{
+				{Messages: &consensus.ProposalPart_BlockInfo{}},
+				{Messages: &consensus.ProposalPart_Commitment{}},
+				{Messages: &consensus.ProposalPart_Transactions{}},
+				{Messages: &consensus.ProposalPart_Fin{}},
+			},
+		},
+		{
+			state: &validator.AwaitingBlockInfoOrCommitmentState{},
+			events: []*consensus.ProposalPart{
+				{Messages: &consensus.ProposalPart_Init{}},
+				{Messages: &consensus.ProposalPart_Transactions{}},
+				{Messages: &consensus.ProposalPart_Fin{}},
+			},
+		},
+		{
+			state: &validator.ReceivingTransactionsState{},
+			events: []*consensus.ProposalPart{
+				{Messages: &consensus.ProposalPart_Init{}},
+				{Messages: &consensus.ProposalPart_BlockInfo{}},
+				{Messages: &consensus.ProposalPart_Fin{}},
+			},
+		},
+		{
+			state: &validator.AwaitingProposalFinState{},
+			events: []*consensus.ProposalPart{
+				{Messages: &consensus.ProposalPart_Init{}},
+				{Messages: &consensus.ProposalPart_BlockInfo{}},
+				{Messages: &consensus.ProposalPart_Transactions{}},
+				{Messages: &consensus.ProposalPart_Commitment{}},
+			},
+		},
+		{
+			state: &validator.FinState{},
+			events: []*consensus.ProposalPart{
+				{Messages: &consensus.ProposalPart_Init{}},
+				{Messages: &consensus.ProposalPart_BlockInfo{}},
+				{Messages: &consensus.ProposalPart_Transactions{}},
+				{Messages: &consensus.ProposalPart_Commitment{}},
+				{Messages: &consensus.ProposalPart_Fin{}},
+			},
+		},
+	}
+
+	transition := validator.NewTransition()
+	for _, step := range steps {
+		t.Run(fmt.Sprintf("State %T", step.state), func(t *testing.T) {
+			for _, event := range step.events {
+				t.Run(fmt.Sprintf("Event %T", event.Messages), func(t *testing.T) {
+					_, err := step.state.OnEvent(t.Context(), transition, event)
+					require.Error(t, err)
+				})
+			}
+		})
+	}
+}
--- a/consensus/p2p/validator/transition.go
+++ b/consensus/p2p/validator/transition.go
@@ -1,0 +1,161 @@
+package validator
+
+import (
+	"context"
+	"errors"
+
+	"github.com/NethermindEth/juno/consensus/starknet"
+	"github.com/NethermindEth/juno/consensus/types"
+	"github.com/NethermindEth/juno/core/felt"
+	"github.com/NethermindEth/juno/utils"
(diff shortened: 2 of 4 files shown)


consensus/p2p/proposal/state_machine.go
```diff
@@ -0,0 +1,69 @@
...
+
+func (s *initialState) onEvent(part *consensus.ProposalPart) (stateMachine, error) {
+	switch part.GetMessages().(type) {
+	case *consensus.ProposalPart_Init:
+		return &receivingBlockInfoState{}, nil
+	default:
+		return nil, errInvalidMessage
+	}
+}
+
```
rianhughes, line 27: If the proposal is empty we move directly to ProposalCommitment

rianhughes: Ah, I see you handle that in receivingBlockInfoState. Imo we should separate blockInfo from commitment

infrmtcs: Upon receiving Init, we actually don't know whether the next message is `BlockInfo` or `ProposalCommitment`, so we can't separate these 2 states.
I think probably the confusion is because the naming is based on the happy path (block info).

consensus/p2p/proposal/state_machine.go
```diff
@@ -0,0 +1,69 @@
...
+	"errors"
+
+	"github.com/NethermindEth/juno/consensus/starknet"
+	"github.com/starknet-io/starknet-p2pspecs/p2p/proto/consensus/consensus"
+)
+
+// TODO: better error handling
+var errInvalidMessage = errors.New("invalid message")
+
+type stateMachine interface {
```
rianhughes, line 13: Also, can we rename this, just to make conversations about the stateMachine easier?

consensus/p2p/proposal/proposal_stream.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+func (s *proposalStream) OnStreamMessage(streamMessage *consensus.StreamMessage) (*starknet.Proposal, error) {
+	if s.nextSequenceNumber != streamMessage.SequenceNumber {
+		s.messages[streamMessage.SequenceNumber] = streamMessage
+		return nil, nil
+	}
+
+	exists := true
+	for exists {
```
rodrigo-pino, line 34: Why this approach instead of `break`/`return` when the execution path that is the one `exists` to true. Nothing against this one. 

Since we are having `exists` as a "top" level variable maybe we can give it a more detailed name such as `streamNextMessage`.

rodrigo-pino (commented):
Looks good for me, let me know as it progresses more reviews are welcomed

consensus/p2p/proposal/transition.go
```diff
@@ -0,0 +1,132 @@
...
+type Transition interface {
+	InitialState() *InitialState
+	OnInit(*InitialState, *consensus.ProposalInit) (*ReceivingBlockInfoState, error)
+	OnEmptyBlock(*ReceivingBlockInfoState, *consensus.ProposalCommitment) (*ReceivingHashState, error)
+	OnBlockInfo(*ReceivingBlockInfoState, *consensus.BlockInfo) (*ReceivingTransactionsState, error)
+	OnTransactions(*ReceivingTransactionsState, []*consensus.ConsensusTransaction) (*ReceivingTransactionsState, error)
+	OnCommitment(*ReceivingTransactionsState, *consensus.ProposalCommitment) (*ReceivingHashState, error)
+	OnFin(*ReceivingHashState, *consensus.ProposalFin) (*FinState, error)
+}
+
```
rianhughes, line 28: Imo we should decouple the Validator interface from this state machine. Maybe Transition could wrap the Validator interface, and adapt the types etc. What do you think?

rianhughes: Although, maybe we could remove the Transition interface completely, and just inline the functions in `func (s *InitialState) OnEvent(...)` etc

rianhughes:
```
type Validator[V types.Hashable[H], H types.Hash, A types.Addr] interface {
    // ProposalInit initialises the pending state according to the proposal
    ProposalInit(pInit types.ProposalInit) error

    // BlockInfo sets the pending header according to the proposal header
    BlockInfo(blockInfo types.BlockInfo)

    // TransactionBatch executes the provided transactions, and stores the result in the pending state
    TransactionBatch(txn types.Transaction) error

    // ProposalCommitment checks the set of proposed commitments against those generated locally.
    ProposalCommitment(proposalCommitment types.ProposalCommitment) error

    // ProposalFin compares the provided commitment with that generated locally
    ProposalFin(proposalCommitment types.ProposalFin) error
}
```

consensus/p2p/proposal/state_machine.go
```diff
@@ -0,0 +1,81 @@
...
+	switch part := part.GetMessages().(type) {
+	case *consensus.ProposalPart_BlockInfo:
+		return transition.OnBlockInfo(s, part.BlockInfo)
+	case *consensus.ProposalPart_Commitment:
+		return transition.OnEmptyBlock(s, part.Commitment)
+	default:
+		return nil, errInvalidMessage
+	}
+}
+
```
rianhughes, line 44: I find the current code to kind of confusing and unclear tbh. 

For example, `ReceivingBlockInfoState` suggests we are receiving a BlockInfo msg, which isn't necessarily the case. Similarly with `ReceivingTransactionsState` which suggests we are recieving a transaction, which also isn't the case. I think we should restructure the code to increase the clarity (eg splitting the SM into two distinct flows, or using a table or something, idk)

infrmtcs: How about renaming them to something like `AfterInitialState` and `AfterBlockInfo`

consensus/p2p/proposal/dispatcher.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/starknet-io/starknet-p2pspecs/p2p/proto/consensus/consensus"
+)
+
+type ProposalDispatcher[V types.Hashable[H], H types.Hash, A types.Addr] interface {
+	ProposalInit(types.Proposal[V, H, A]) (*consensus.ProposalInit, error)
+	ProposalBlockInfo(types.Proposal[V, H, A]) (*consensus.BlockInfo, error)
+	ProposalTransactions(types.Proposal[V, H, A]) ([]consensus.ConsensusTransaction, error)
+	ProposalCommitment(types.Proposal[V, H, A]) (*consensus.ProposalCommitment, error)
+	ProposalFin(types.Proposal[V, H, A]) (*consensus.ProposalFin, error)
+}
```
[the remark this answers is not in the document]
rianhughes: Also nit: proposer.go seems a more natural name that dispatcher.go. Feel free to ignore this though

consensus/p2p/proposal/broadcast.go
```diff
@@ -0,0 +1,185 @@
...
+		seqNo++
+		if !yield(fromBlockInfo(dispatcher, streamId, seqNo, &proposal)) {
+			return
+		}
+
+		seqNo++
+		if !yield(fromTransactions(dispatcher, streamId, seqNo, &proposal)) {
+			return
+		}
+
```
rianhughes, line 36: In the current design, are we able to call this more than once? SAfaik, Starknet will require us to stream txns multiple times, not all txns in one go (eg we can stream on set of txns at time t, and another set ad time t+a)

consensus/p2p/proposal/broadcast.go
```diff
@@ -0,0 +1,185 @@
...
+import (
+	"iter"
+
+	"github.com/NethermindEth/juno/consensus/types"
+	"github.com/starknet-io/starknet-p2pspecs/p2p/proto/common"
+	"github.com/starknet-io/starknet-p2pspecs/p2p/proto/consensus/consensus"
+	"google.golang.org/protobuf/proto"
+)
+
+func Broadcast[V types.Hashable[H], H types.Hash, A types.Addr](
```
rianhughes, line 12: Also, this is specific to broadcasting the propsal right, not broadcasting general msgs? can we rename it (eg `ProposalBroadcaster`) or add a doc string?

consensus/p2p/proposal/broadcast.go
```diff
@@ -0,0 +1,185 @@
...
+	}
+
+	return sendProposalPart(streamId, seqNo, &consensus.ProposalPart{
+		Messages: &consensus.ProposalPart_Transactions{
+			Transactions: &consensus.TransactionBatch{
+				Transactions: txBatch,
+			},
+		},
+	})
+}
```
rianhughes, line 126: We should double check this, but as I mentioned above I think Starknet might require us to send multipl distinct txn batches. In which case, this logic isn't correct. However, if I'm wrong and we send all the block txns in one go, it's correct.

consensus/p2p/proposal/state_machine.go
```diff
@@ -0,0 +1,81 @@
...
+func (s *ReceivingBlockInfoState) OnEvent(transition Transition, part *consensus.ProposalPart) (ProposalStateMachine, error) {
+	switch part := part.GetMessages().(type) {
+	case *consensus.ProposalPart_BlockInfo:
+		return transition.OnBlockInfo(s, part.BlockInfo)
+	case *consensus.ProposalPart_Commitment:
+		return transition.OnEmptyBlock(s, part.Commitment)
+	default:
+		return nil, errInvalidMessage
+	}
```
rianhughes, line 42:
```suggestion
// AwaitingProposalContentState handles the transition after ProposalInit,
// accepting either a BlockInfo (for full proposals) or a Commitment (for empty blocks).
type AwaitingProposalContentState struct {
	Header     *starknet.MessageHeader
	ValidRound types.Round
}

func (s *AwaitingProposalContentState) OnEvent(transition Transition, part *consensus.ProposalPart) (ProposalStateMachine, error) {
	switch part := part.GetMessages().(type) {
	case *consensus.ProposalPart_BlockInfo:
		return transition.OnBlockInfo(s, part.BlockInfo)
	case *consensus.ProposalPart_Commitment:
		return transition.OnEmptyBlock(s, part.Commitment)
	default:
		return nil, errInvalidMessage
	}
```

consensus/p2p/proposal/state_machine.go
```diff
@@ -0,0 +1,81 @@
...
+)
+
+// TODO: better error handling
+var errInvalidMessage = errors.New("invalid message")
+
+type ProposalStateMachine interface {
+	OnEvent(Transition, *consensus.ProposalPart) (ProposalStateMachine, error)
+}
+
+type InitialState struct{}
```
rianhughes, line 18:
```suggestion
// InitialState handles the first step in the proposal flow,
// accepting only a ProposalInit message to begin the process.
type InitialState struct{}
```

consensus/p2p/proposal/state_machine.go
```diff
@@ -0,0 +1,81 @@
...
+	case *consensus.ProposalPart_BlockInfo:
+		return transition.OnBlockInfo(s, part.BlockInfo)
+	case *consensus.ProposalPart_Commitment:
+		return transition.OnEmptyBlock(s, part.Commitment)
+	default:
+		return nil, errInvalidMessage
+	}
+}
+
+type ReceivingTransactionsState struct {
```
rianhughes, line 45:
```suggestion
// ReceivingTransactionsState handles the flow where transaction batches are received,
// continuing until a ProposalCommitment message is received to conclude the input.
type ReceivingTransactionsState struct {
```

consensus/p2p/proposal/state_machine.go
```diff
@@ -0,0 +1,81 @@
...
+	Header     *starknet.MessageHeader
+	ValidRound types.Round
+}
+
+func (s *ReceivingBlockInfoState) OnEvent(transition Transition, part *consensus.ProposalPart) (ProposalStateMachine, error) {
+	switch part := part.GetMessages().(type) {
+	case *consensus.ProposalPart_BlockInfo:
+		return transition.OnBlockInfo(s, part.BlockInfo)
+	case *consensus.ProposalPart_Commitment:
+		return transition.OnEmptyBlock(s, part.Commitment)
```
rianhughes, line 39: Why don't we call `transition.OnCommitment(s, part.Commitment)` here?

consensus/p2p/proposal/transition.go
```diff
@@ -0,0 +1,132 @@
...
+var (
+	errNilProposer           = errors.New("proposer is nil")
+	errNilProposalCommitment = errors.New("proposal commitment is nil")
+	errHashMismatch          = errors.New("hash mismatch")
+)
+
+type Transition interface {
+	InitialState() *InitialState
+	OnInit(*InitialState, *consensus.ProposalInit) (*ReceivingBlockInfoState, error)
+	OnEmptyBlock(*ReceivingBlockInfoState, *consensus.ProposalCommitment) (*ReceivingHashState, error)
```
rianhughes, line 22:
```suggestion
	OnPropsalInit(*InitialState, *consensus.ProposalInit) (*ReceivingBlockInfoState, error)
```

rianhughes:
Finished the current review. My main comments are:
1. It looks like the Broadcaster expects all txns to be sent at once (let me know if I'm wrong here)
2. We need tests
3. Renaming various structs etc, just to help with clarity

consensus/p2p/proposal/state_machine.go
```diff
@@ -0,0 +1,81 @@
...
+	Value      *starknet.Value
+}
+
+func (s *ReceivingTransactionsState) OnEvent(transition Transition, part *consensus.ProposalPart) (ProposalStateMachine, error) {
+	switch part := part.GetMessages().(type) {
+	case *consensus.ProposalPart_Transactions:
+		return transition.OnTransactions(s, part.Transactions.Transactions)
+	case *consensus.ProposalPart_Commitment:
+		return transition.OnCommitment(s, part.Commitment)
+	default:
```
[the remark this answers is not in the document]
infrmtcs: Agree, but it's the broadcaster's problem. I updated the broadcaster.

consensus/p2p/proposer/proposer.go
```diff
@@ -0,0 +1,195 @@
...
+				}
+				txBatch = txBatch[:0]
+			}
+		}
+
+		if len(txBatch) > 0 {
+			if !yield(seqNoCounter.fromTransactions(txBatch)) {
+				return
+			}
+		}
```
rianhughes, line 63: Just a suggestion to avoid the append overhead

```suggestion
		txBatch := make([]*consensus.ConsensusTransaction, txBatchSize)
		i := 0
		for tx := range txStream {
			txBatch[i] = tx
			i++

			if i == txBatchSize {
				if !yield(seqNoCounter.fromTransactions(txBatch[:i])) {
					return
				}
				i = 0
			}
		}

		if i > 0 {
			if !yield(seqNoCounter.fromTransactions(txBatch[:i])) {
				return
			}
		}
```

rianhughes:
Looks good to me, up to tests

consensus/p2p/proposer/proposer.go
infrmtcs: I'm curious about this too, so I wrote a short benchmark to compare the performance between append with pre-allocated capacity and using an int as index:

```
package main

import (
	"iter"
	"math/rand"
	"testing"
)

const (
	maxSize    = 100000
	actualSize = maxSize / 2
)

type item struct {
	x, y, z int
}

func stream() iter.Seq[item] {
	return func(yield func(item) bool) {
		for range actualSize {
			if !yield(item{x: rand.Int(), y: rand.Int(), z: rand.Int()}) {
				return
			}
		}
	}
}

// Memory allocation benchmark
func BenchmarkDirectAssignmentMem(b *testing.B) {
	b.ReportAllocs()
	for range b.N {
		slice := make([]item, maxSize)
		stream := stream()
		j := 0
		for item := range stream {
			slice[j] = item
			j++
		}
	}
}

func BenchmarkAppendStrategyMem(b *testing.B) {
	b.ReportAllocs()
	for range b.N {
		slice := make([]item, 0, maxSize)
		stream := stream()
		for item := range stream {
			slice = append(slice, item)
		}
	}
}

func BenchmarkGrowingSliceMem(b *testing.B) {
	b.ReportAllocs()
	for range b.N {
		var slice []item
		stream := stream()
		for item := range stream {
			slice = append(slice, item)
		}
	}
}
```

image

infrmtcs: That said, I think direct assignment can get slight performance gain because it doesn't have to check bound, while append has better clarity. I don't have strong opinion here, so let me know what you think.

rianhughes: I think we've been trying to go for the direct assignment over the append approach recently, given it's more efficient

consensus/p2p/validator/transition.go
```diff
@@ -0,0 +1,166 @@
...
+		context.Context,
+		*ReceivingTransactionsState,
+		*consensus.ProposalCommitment,
+	) (*AwaitingProposalFinState, error)
+	OnProposalFin(
+		context.Context,
+		*AwaitingProposalFinState,
+		*consensus.ProposalFin,
+	) (*FinState, error)
+}
```
rodrodros, line 50: Seeing the implementation of these functions, they all create a type and return a reference to them. Why not return them by value and avoid the heap allocation

consensus/p2p/validator/transition.go
```diff
@@ -0,0 +1,166 @@
...
+
+// TODO: Implement this function properly
+func (t *transition) OnProposalFin(
+	ctx context.Context,
+	state *AwaitingProposalFinState,
+	fin *consensus.ProposalFin,
+) (*FinState, error) {
+	var actual *starknet.Hash
+	if fin.ProposalCommitment != nil {
+		actual = utils.HeapPtr(starknet.Hash(felt.FromBytes(fin.ProposalCommitment.Elements)))
```
rodrodros: I would avoid to use this method since it means a heap allocation.

I think we can keep the code intention but by changing the code instruction there is no need for it. For example by changing the checks on `expected` and `actual` down in lines 151 & 152 to checks on `state.Value` and `fin.ProposalCommitment`. 

Then calculate `Expected` and `Actual` at the end and compare them

consensus/p2p/validator/transition.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+
+	var expected *starknet.Hash
+	if state.Value != nil {
+		expected = utils.HeapPtr(state.Value.Hash())
+	}
+	if expected == nil || actual == nil {
+		if expected != nil {
+			return nil, errHashMismatch
+		}
```
rodrodros: It's intended that if `actual` is the nil one no error is thrown?

infrmtcs: Agree, fixed

consensus/p2p/validator/transition.go
infrmtcs: Updated, let me know if this version matches what you're expecting.

consensus/p2p/validator/transition.go
infrmtcs: The fields of the different states are different
=> we need to store the state as an interface (`StateMachine`) inside `proposalStream`
=> escape analysis will probably still allocate it to heap anyway
So I think this doesn't avoid heap allocation, but risk copying the struct.

rodrodros: I see, your explanation makes sense 👍

rodrigo-pino:
@infrmtcs please add an issue that represents adding test for this and please link it to either of the main consensus issues

infrmtcs:
@rodrigo-pino @rianhughes I added tests to this PR

consensus/p2p/validator/state_machine_test.go
```diff
@@ -0,0 +1,245 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+type validTransitionTestStep struct {
+	event     *consensus.ProposalPart
+	postState validator.ProposalStateMachine
+}
+
+func TestProposalStateMachine_ValidTranstions(t *testing.T) {
```
rianhughes, line 22: Nice

PR #2841, opened 2025-05-19
feat: Add adapters for vote

infrmtcs:


consensus/p2p/adapters/vote.go
```diff
@@ -0,0 +1,66 @@
...
+	"github.com/starknet-io/starknet-p2pspecs/p2p/proto/common"
+	"github.com/starknet-io/starknet-p2pspecs/p2p/proto/consensus/consensus"
+)
+
+type VoteAdapter[H types.Hash, A types.Addr] interface {
+	ToVote(*consensus.Vote) (*types.Vote[H, A], error)
+	FromVote(*types.Vote[H, A], consensus.Vote_VoteType) (*consensus.Vote, error)
+	FromPrevote(*types.Prevote[H, A]) (*consensus.Vote, error)
+	FromPrecommit(*types.Precommit[H, A]) (*consensus.Vote, error)
+}
```
rodrigo-pino, line 21: All the types being returned here are created inside the function, what do you think of returning them as value types instead?

consensus/p2p/vote/vote.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+func (a starknetVoteAdapter) FromVote(vote *starknet.Vote, voteType consensus.Vote_VoteType) (consensus.Vote, error) {
+	sender := vote.Sender.AsFelt().Bytes()
+
+	var id *common.Hash
+	if vote.ID != nil {
+		bytes := vote.ID.AsFelt().Bytes()
+		id = &common.Hash{Elements: bytes[:]}
+	}
```
rianhughes, line 53:
```suggestion
	// This is optional since a vote can be NIL.
	var id *common.Hash
	if vote.ID != nil {
		bytes := vote.ID.AsFelt().Bytes()
		id = &common.Hash{Elements: bytes[:]}
	}
```

consensus/p2p/vote/vote.go
```diff
@@ -0,0 +1,62 @@
...
+	"github.com/NethermindEth/juno/consensus/types"
+	"github.com/NethermindEth/juno/core/address"
+	"github.com/NethermindEth/juno/core/felt"
+	"github.com/NethermindEth/juno/core/hash"
+	"github.com/NethermindEth/juno/utils"
+	"github.com/starknet-io/starknet-p2pspecs/p2p/proto/common"
+	"github.com/starknet-io/starknet-p2pspecs/p2p/proto/consensus/consensus"
+)
+
+type VoteAdapter[H types.Hash, A types.Addr] interface {
```
rianhughes, line 16: Why use an interface for this? We normally Junos adaptors are just functions (eg `func AdaptBlock(response *starknet.Block, sig *starknet.Signature) (*core.Block, error)`)

rianhughes:
Looks good

consensus/p2p/vote/vote.go
rianhughes: It's not clear to me that we need this abstraction atm, and I think it introduces a small overhead, but it could save us from problems later on (eg if we need multiple adaptors for whatever reason)

rianhughes (approved):

PR #2842, opened 2025-05-19
feat: Implement P2P components

infrmtcs:
```mermaid
stateDiagram-v2
    classDef merged fill:#e1bee7,stroke:#9c27b0,stroke-width:2px,color:#4a148c
    classDef ready fill:#c8e6c9,stroke:#4caf50,stroke-width:2px,color:#1b5e20
    classDef pending fill:#fff9c4,stroke:#ffeb3b,stroke-width:2px,color:#f57f17

    state "feat: Add scaffolding for proposer sending messages to P2P (#2856)" as dispatcher
    state "feat: Implement proposer component (#2947)" as proposer
    state "feat: Implement P2P components (#2842)" as p2p

    dispatcher --> p2p
    proposer --> p2p

    class merged
    class proposer ready
    class dispatcher,p2p pending
```
```mermaid
stateDiagram-v2
    classDef merged fill:#e1bee7,stroke:#9c27b0,stroke-width:2px,color:#4a148c
    classDef ready fill:#c8e6c9,stroke:#4caf50,stroke-width:2px,color:#1b5e20
    classDef pending fill:#fff9c4,stroke:#ffeb3b,stroke-width:2px,color:#f57f17

    state "🟣 Merged" as legendMerged
    state "🟢 Ready for Review (Good Test Coverage)" as legendReady
    state "🟡 Implementation Complete (Pending Tests)" as legendPending

    class legendMerged merged
    class legendReady ready
    class legendPending pending
```


consensus/integtest/p2p.go
```diff
@@ -0,0 +1,139 @@
...
+
+func connect(t *testing.T, mutex *sync.Mutex, thisNodeConfig, otherNodeConfig networkNodeConfig) {
+	mutex.Lock()
+	defer mutex.Unlock()
+
+	peer := peer.AddrInfo{ID: otherNodeConfig.host.ID(), Addrs: otherNodeConfig.host.Addrs()}
+	require.NoError(t, thisNodeConfig.host.Connect(t.Context(), peer))
+}
+
+func (c networkConfig) setup(t *testing.T) {
```
rodrodros, line 44: Should we add t.Helper()?

consensus/integtest/p2p.go
```diff
@@ -0,0 +1,139 @@
...
+}
+
+func (c networkNodeConfig) isConnected(to int) bool {
+	_, ok := c.adjacentNodes[to]
+	return ok
+}
+
+type networkConfig []networkNodeConfig
+
+func connect(t *testing.T, mutex *sync.Mutex, thisNodeConfig, otherNodeConfig networkNodeConfig) {
```
rodrodros, line 34: Should we add t.Helper()?

consensus/integtest/p2p.go
```diff
@@ -0,0 +1,139 @@
...
+		require.NoError(t, err)
+
+		host, err := libp2p.New(
+			libp2p.ListenAddrs(addr),
+			// libp2p.Identity(prvKey),
+			// libp2p.UserAgent(makeAgentName(version)),
+			// // Use address factory to add the public address to the list of
+			// // addresses that the node will advertise.
+			// libp2p.AddrsFactory(addressFactory),
+			// If we know the public ip, enable the relay service.
```
rodrodros, line 81: Are the commented lines of code still (or will) be relevant?

consensus/p2p/validator/proposal_stream_demux.go
```diff
@@ -201,6 +201,11 @@ func (t *proposalStreamDemux) processStreamMessage(ctx context.Context, pubsubMe
 // onFirstMessage is called when the first message is received for a stream.
 func (t *proposalStreamDemux) onFirstMessage(ctx context.Context, streamID streamID, message *consensus.StreamMessage) error {
 	stream := t.getStream(streamID)
+	if stream.started {
+		// If the stream is already started, then we don't need to start it again
+		return nil
+	}
```
rodrodros, line 207: Why can this function be called twice. Shouldn't it be only called once by design?

consensus/proposer/proposer.go
```diff
@@ -224,18 +231,27 @@ func (p *proposer[V, H]) runTransactions(transaction []mempool.BroadcastedTransa
...
 
 	value := p.toValue(buildResult.Pending.Block.Hash)
 	p.proposalStore.Store(value.Hash(), &buildResult)
 
 	responseChannel <- value
-	close(responseChannel)
+}
 
-	return nil
+func ask[D, R any](channel chan request[D, R], data D) R {
```
rodrodros: Let's add a small description to this simple function

rodrigo-pino (commented):
Revew

consensus/integtest/p2p.go
infrmtcs: We may need these options in the future, but after I make a more intensive study on those settings.

consensus/p2p/validator/proposal_stream_demux.go
infrmtcs: There are multiple cases where a node can receive the first message twice:
- A correct proposer rebroadcasts the first part. This is an expected behaviour.
- A faulty proposer equivocates the first part.

rianhughes: In the equivocation case, this means we are tied to the first proposal part they send us right? But we can still receive the correct proposal (assuming one was sent out to the rest of the network), via other peers?

```
Receivers identify streams based on (sender_id, stream_id), so distinct senders need not coordinate IDs.
message ConsensusStreamId {
    uint64 block_number = 1;
    uint32 round        = 2;
    uint64 nonce        = 3;
}
```

infrmtcs: Yes, that's correct. We can still receive the alternative first part, but currently we're not processing or store it. This is part of the scenario where I'm having a long discussion with malachite and starkware.

consensus/p2p/p2p.go
```diff
@@ -0,0 +1,171 @@
...
+	"github.com/libp2p/go-libp2p/core/host"
+	"github.com/sourcegraph/conc"
+)
+
+type topicName string
+
+const (
+	proposalTopicName topicName = "consensus_proposals"
+	voteTopicName     topicName = "consensus_votes"
+	gossipSubHistory            = 60
```
rianhughes, line 28: Why change the default to this?

infrmtcs: I'll add a comment to this. Basically each node will keep track of messages received in the last `params.HistoryLength` heartbeats and broadcast messages in the last `params.HistoryGossip` heartbeats.
Each heartbeat is 1 second long, and the default value is 3 and 5, so it's keep track of messages from the last 5s and broadcast the ones from the last 3s.
I increase them because of the cases in integration tests where the P2P network takes more than 3s to broadcast to the next node, when it's heavily overloaded.

rianhughes: This means that the node will rebroadcast a lot more messages than it would in the default cause, which uses more resources. Are we sure we want to do this in production? If we just need to modify it for tests, maybe we can override the default value (eg make it a field of the `p2p` type, and implement `setGossipParams()`)?

infrmtcs: No, it doesn't rebroadcast the whole messages, but it's used for exchanging the message IDs with other peers via IHAVE and IWANT messages. So A will send IHAVE to B, containing the message IDs that it receives in the last `HistoryGossip` seconds. B check if it has that message ID in the last `HistoryLength` seconds. If B doesn't have, B send IWANT to A telling A that B want A to send that message.

rodrigo-pino (approved):

PR #2844, opened 2025-05-19
refactor: Remove precommits argument from Commit function

infrmtcs:
As discussed, the precommits used as proof should be stored in the DB, so we don't need to pass it in `Commit` function.


infrmtcs:
PR closed accidentally, reopen now.

rianhughes (approved):

PR #2847, opened 2025-05-20
Gateway: URL split

EgeCaner:
Updated Feeder URL, gateway and feeder still uses the same api-key, needs to be clarified
Changed supported starknet version to 0.14.0

Tested on sepolia-integration w/o api-key, my nodes syncs

### DIFF
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -53,7 +53,7 @@ type Reader interface {
 
 var (
 	ErrParentDoesNotMatchHead = errors.New("block's parent hash does not match head block hash")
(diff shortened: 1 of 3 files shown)


rianhughes (changes requested):
It looks like SN have only updated the `sepolia-integration` endpoints so far, so we can't update the other networks just yet. Also, it looks like the old `sepolia-integration` endpoint has been deprecated so we need to update this endpoint (as you've done).

EgeCaner:
@rianhughes I updated the code to roll back to previous urls except for `sepolia-integration`

rodrigo-pino (approved):

rianhughes (approved):

PR #2849, opened 2025-05-20
Cache for submitted txs

EgeCaner:
Submitted transactions cache implementation to be able to return `RECEIVED` finality status when tx is forwarded to sequencer but not included in a block yet.

- Submitted transactions considered valid for a period (Entry-Time-to-Live).
- With each `Add` operation cache is flushed.
- Flush mechanism removes expired entries form cache.
- If cache is full with valid entries, attempting to append new tx will remove least-recently-used entry.
- Cache parameters can be adjusted via node config(`size` and `entryTTL`).
- Cache size defaulted to 1024, which is the default value for mempoolLimit and entryTTL is defautled to 5 minutes.


rpc/v8/transaction_test.go
```diff
@@ -1433,6 +1434,118 @@ func TestTransactionStatus(t *testing.T) {
...
+				txnToAdd.TransactionHash = txnHash
+				mockReader := mocks.NewMockReader(mockCtrl)
+				mockSyncReader := mocks.NewMockSyncReader(mockCtrl)
+				mockReader.EXPECT().TransactionByHash(txnHash).Return(nil, db.ErrKeyNotFound)
+				mockSyncReader.EXPECT().PendingBlock().Return(nil)
+				handler := rpc.New(mockReader, mockSyncReader, nil, "", log).
+					WithFeeder(client).
+					WithGateway(gw).
+					WithSubmittedTransactionsCache(submittedTransactionCache)
+				/// Test Gateway server wont add this tx, gateway should return txnNotFound upon querying transaction status
```
EgeCaner, line 1483: Does it makes sense to use, test gateway relying on it will return `txnNotFound` considering it wont store txn somewhere? Can gateway be mocked as well?

rpc/v8/transaction_test.go
```diff
@@ -1433,6 +1434,118 @@ func TestTransactionStatus(t *testing.T) {
...
+				require.Nil(t, err)
+				require.Equal(t, rpc.TxnStatusReceived, status.Finality)
+			})
+
+			t.Run("transaction not found in db and feeder, found in cache but expired", func(t *testing.T) {
+				gw := gateway.NewTestClient(t)
+				cacheEntryTimeOut := time.Second
+				submittedTransactionCache := rpccore.NewSubmittedTransactionsCache(5, cacheEntryTimeOut)
+
+				txnToAdd := &core.InvokeTransaction{
```
EgeCaner, line 1499: Should this test entry belongs here or should be placed 'tests' mapping in top of this test?

rodrigo-pino (commented):
There is a missing delete case which is when a transaction is queried and it does exist on the gw, then it should be removed from the cache (it already has a status). I don't see this reflected in the code, let me know if I am missing something.

rpc/v6/transaction.go
```diff
@@ -703,6 +717,11 @@ func (h *Handler) TransactionStatus(ctx context.Context, hash felt.Felt) (*Trans
 			status.Finality = TxnStatusAcceptedOnL2
 		case starknet.Received:
 			status.Finality = TxnStatusReceived
+		case starknet.NotReceived:
+			if h.submittedTransactionsCache == nil || !h.submittedTransactionsCache.Contains(hash) {
+				return nil, rpccore.ErrTxnHashNotFound
+			}
+			status.Finality = TxnStatusReceived
```
rianhughes, line 724: Are we sure the FGWs `get_transaction` now returns `NotReceived`?

I thought the flow would be like this;
1. User sends Tcn to Juno node
2. Juno node sends it to the fgw, and if the fgw accepts it, we store the hash in our cache
3. User querys the same Juno node for the transaction status. It is not in the pending block yet, so we enter the `case rpccore.ErrTxnHashNotFound` above. We query the fgW again, and it returns "txn hash not found" (because it is in their mempool but not in the pending block just yet), so we check our cache, find it, and then return "Received"

node/node.go
```diff
@@ -244,7 +248,16 @@ func New(cfg *Config, version string, logLevel *utils.LogLevel) (*Node, error) {
...
+
+		submittedTransactionsCache := rpccore.NewSubmittedTransactionsCache(
+			int(cfg.SubmittedTransactionsCacheSize),
+			cfg.SubmittedTransactionsCacheEntryTTL,
+		)
+
+		rpcHandler = rpc.New(chain, syncReader, throttledVM, version, log, &cfg.Network).
+			WithGateway(gatewayClient).
+			WithFeeder(client).
+			WithSubmittedTransactionsCache(submittedTransactionsCache)
```
infrmtcs, line 260: Maybe this is not a problem of this PR, but we're mixing between a constructor and builder pattern here.

rpc/v8/transaction.go
```diff
@@ -606,11 +606,25 @@ func (h *Handler) TransactionReceiptByHash(hash felt.Felt) (*TransactionReceipt,
 
 // AddTransaction relays a transaction to the gateway, or to the sequencer if enabled
 func (h *Handler) AddTransaction(ctx context.Context, tx BroadcastedTransaction) (*AddTxResponse, *jsonrpc.Error) { //nolint:gocritic
+	var (
+		res *AddTxResponse
+		err *jsonrpc.Error
+	)
 	if h.memPool != nil {
```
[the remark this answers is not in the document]
rianhughes: It's not critical, the sequencer can do what it wants.

rpc/rpccore/submitted_transaction_cache_test.go
```diff
@@ -0,0 +1,110 @@
...
+	"time"
+
+	"github.com/NethermindEth/juno/core/felt"
+	"github.com/NethermindEth/juno/rpc/rpccore"
+	"github.com/stretchr/testify/require"
+)
+
+func TestAdd(t *testing.T) {
+	capacity := 5
+	entryTTL := time.Second
```
rianhughes, line 14: nit: I think we can reduce this to a Microsecond, so the tests can finish faster

rpc/rpccore/submitted_transaction_cache_test.go
```diff
@@ -0,0 +1,110 @@
...
+			txnHash := new(felt.Felt).SetUint64(uint64(i))
+			cache.Add(*txnHash)
+		}
+
+		txnHash = *new(felt.Felt).SetUint64(uint64(capacity))
+		// Expected to remove the least-recently-used entry.
+		cache.Add(txnHash)
+		require.True(t, cache.Contains(txnHash))
+		require.False(t, cache.Contains(felt.Zero))
+	})
```
rianhughes, line 57: nit: readability regarding felt.Zero
```suggestion
	t.Run("Inserting to full cache with non-expired txs", func(t *testing.T) {
		cache := rpccore.NewSubmittedTransactionsCache(capacity, entryTTL)
		firstTxnHash := felt.Zero
		cache.Add(firstTxnHash)
		require.True(t, cache.Contains(firstTxnHash))
		// Populates the cache to cap.
		for i := 1; i < capacity; i++ {
			txnHash := new(felt.Felt).SetUint64(uint64(i))
			cache.Add(*txnHash)
		}

		txnHash := *new(felt.Felt).SetUint64(uint64(capacity))
		// Expected to remove the least-recently-used entry.
		cache.Add(txnHash)
		require.True(t, cache.Contains(txnHash))
		require.False(t, cache.Contains(firstTxnHash))
	})
```

rpc/v8/transaction_test.go
```diff
@@ -1433,6 +1434,118 @@ func TestTransactionStatus(t *testing.T) {
...
+				require.Nil(t, err)
+				require.Equal(t, rpc.TxnStatusReceived, status.Finality)
+			})
+
+			t.Run("transaction not found in db and feeder, found in cache but expired", func(t *testing.T) {
+				gw := gateway.NewTestClient(t)
+				cacheEntryTimeOut := time.Second
+				submittedTransactionCache := rpccore.NewSubmittedTransactionsCache(5, cacheEntryTimeOut)
+
+				txnToAdd := &core.InvokeTransaction{ //nolint:dupl
```
rianhughes, line 1499: Regarding the `//nolint:dupl`, a lot of the code in this and the previous test is duplicated. I think we could seperate these out into their own `TestRPCCahche` and cut out the duplciated code

rpc/rpccore/submitted_transaction_cache.go
```diff
@@ -0,0 +1,72 @@
...
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	c.flush()
+}
+
+// Removes all entries that have exceeded their lifespan.
+func (c *SubmittedTransactionsCache) flush() {
+	expiredKeys := make([]felt.Felt, 0)
+	for _, k := range c.cache.Keys() {
```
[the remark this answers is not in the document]
EgeCaner: I guess it wont gonna work cuz what is meant by the oldest in those functions is the least recently used item, and in `Contains` function of `SubmittedTransactionsCache` I used `BasicLRU.Get()` which updates items recency, thus entry inserted at time `t` can be front of the item inserted at time `t=1` in order (LRU) list if first item is queried after insertion of the second. Briefly valid items can appear earlier than expired ones in this iteration and if we exit upon valid entry, we will leave the expired ones in cache. Please let me know if Im missing something

rpc/v6/transaction_test.go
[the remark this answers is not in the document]
EgeCaner: I thought we are propogating such changes to all rpc versions thus I also mirrored the tests. Is it unnecessary?

rpc/v8/transaction_test.go
EgeCaner: Makes sense. Regarding the duplicate code in scope of this PR:
The config tests in `juno_test` contain a lot of duplicated code for the expected configs. The linter didn’t complain about this previously, but after I introduced two more fields to node.Config, I think it crossed some threshold and started flagging the duplicate code. That’s why I also added //nolint:dupl to silence those warnings as well.

Do you recommend reducing the duplication by extracting those expected configs, or is it fine to keep it as is?

rpc/v8/transaction.go
```diff
@@ -606,11 +606,25 @@ func (h *Handler) TransactionReceiptByHash(hash felt.Felt) (*TransactionReceipt,
...
-		return h.addToMempool(&tx)
+		res, err = h.addToMempool(&tx)
 	} else {
-		return h.pushToFeederGateway(ctx, tx)
+		res, err = h.pushToFeederGateway(ctx, tx)
+	}
+
+	if err != nil {
+		return res, err
```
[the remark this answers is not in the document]
EgeCaner: Changed as suggested

node/node.go
rodrigo-pino: Do you find something inherently wrong with it?

The way I see it that we have now I think is:
1. This hybrid approach where we have mandatory fields and optional fields are set through the builder pattern
2. Do you think we should go full constructor or full build pattern? I think it makes code harder to read, but happy to hear your thoughts

rpc/rpccore/submitted_transaction_cache.go
```diff
@@ -0,0 +1,81 @@
...
+	return true
+}
+
+// Remove deletes the entry for txnHash from the cache if it exists.
+// It returns true if the entry was present and removed, or false otherwise.
+func (c *SubmittedTransactionsCache) Remove(txnHash felt.Felt) bool {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	return c.cache.Remove(txnHash)
```
rodrigo-pino, line 57: We added the code but no test yet it seems 👀

rpc/rpccore/submitted_transaction_cache.go
```diff
@@ -0,0 +1,81 @@
...
+func (c *SubmittedTransactionsCache) Add(txnHash felt.Felt) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	c.flush()
+	c.cache.Add(txnHash, time.Now())
+}
+
+// Returns true if the entry exists in the cache and has not exceeded its lifespan; otherwise, returns false.
+func (c *SubmittedTransactionsCache) Contains(txnHash felt.Felt) bool {
```
rodrigo-pino, line 34: Pass txnHash by reference please

rpc/rpccore/submitted_transaction_cache.go
```diff
@@ -0,0 +1,81 @@
...
+		c.cache.Remove(txnHash)
+		return false
+	}
+
+	return true
+}
+
+// Remove deletes the entry for txnHash from the cache if it exists.
+// It returns true if the entry was present and removed, or false otherwise.
+func (c *SubmittedTransactionsCache) Remove(txnHash felt.Felt) bool {
```
rodrigo-pino, line 53: pass txnHash by reference please

rpc/v8/transaction_test.go
EgeCaner: @rianhughes extracted rpc cache cases into its own test

rodrigo-pino (approved):

rpc/rpccore/submitted_transaction_cache.go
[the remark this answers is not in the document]
infrmtcs: Yeah you're right. We evict the cache by both LRU and TTL policy, the current approach seems inevitable. But I think we can consider whether it's worth to do the Flush, but instead delete on retrieval if it's older than TTL. I don't have a strong opinion though, so feel free to ignore this.

node/node.go
infrmtcs: I think there are 3 kinds of fields here:
- Mandatory fields like `chain` or `syncReader`. Without them, nothing can work.
- Optional fields like `submittedTransactionsCache`. With the current implementation, everything works fine with or without it.
- Fields like gateway and feeder are unclear to me. They're currently optional by builder pattern, but at the same time we throw 500 if they're missing in some particular endpoints. At the same time, `vm` is mandatory, but it's similar to gateway and feeder because it's only used in some particular endpoints like trace and simulate.

infrmtcs: I don't have strong opinion here, because probably the gain is probably not much, but I think:
- If we prefer builder pattern, there should be a `Build` function to maintain immutability. Reason: we can call `With...` 20 lines below, which make it error prone and hard to debug/reason.
- Another approach which is more idiomatic in my opinion is Option pattern. In short, Option is a function that mutate the implementation type (private struct). Options are used as variadic arguments of the constructor, which return the interface type (public interface).

infrmtcs (approved):

PR #2851, opened 2025-05-22
feature: implement the consensus Validator interface

rianhughes:
This PR implements the Validator interface for the consensus pkg. It is required to process the correctness of a proposed block. 

Notes:
* I currently use placeholder types (`tmp.go`). These will be replaced once we autogenerate types from the spec, and implement the adaptors. It should be a pretty quick/easy change.
* I added a cmd-line flag to specify the protocol version that the builder supports when executing transactions.
* For the tests I used a custom chain so that we can execute transactions, instead of just mocking. Currently we can't execute transactions from `mainnet/sepolia` because 1) the first few blocks are no longer supported by our current blockifier version, and 2) we can't execute blocks at the head because we can't access the state.
* When verifying a block, we do not check against all fields in the `ProposalCommit` since these fields are either not present, or will be dropped soon. We should return to this once we have the final spec.


blockchain/blockchain.go
```diff
@@ -507,9 +535,53 @@ func (b *Blockchain) Finalise(
...
+	// Simulate without commit
+	txn := b.database.NewIndexedBatch()
+	if err := simulate(txn); err != nil {
+		return nil, nil, nil, nil, err
+	}
+	txn.Reset()
+	return newBlock, newSU, newCommitments, concatCount, nil
+}
+
+// Finalise appends the block to the chain, assuming the checks pass
```
rodrigo-pino, line 560: I would suggest:
```suggestion
// Finalise checks the block correctness and appends it to the chain
```
Or something similar, I think it's easier to read

consensus/validator/validator.go
```diff
@@ -0,0 +1,181 @@
...
+	}
+	return a.builder.Finalise(nil)
+}
+
+// Todo: there are fields in ProposalCommitment that we don't check against. Some of these fields
+// will be dropped in the finalised spec, so I don't think we should worry about them until then
+//  1. Some fields we can't get / compute: VersionConstantCommitment, NextL2GasPriceFRI
+//  2. The gas prices. Currently the spec sets eth gas prices, but in v1, these will be dropped
+//     for fri prices.
+func compareProposalCommitment( //nolint:gocyclo // disable linter because this function has a simple pattern
```
rodrigo-pino, line 123: 👍

builder/builder.go
```diff
@@ -54,8 +56,16 @@ type Builder struct {
...
+func New(
+	privKey *ecdsa.PrivateKey,
+	ownAddr *felt.Felt,
+	bc *blockchain.Blockchain,
+	vm vm.VM,
+	blockTime time.Duration,
+	mempool *mempool.Pool,
+	log utils.Logger,
+	disableFees bool,
+	database db.KeyValueStore,
```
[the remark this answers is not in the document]
rianhughes: 🤝

consensus/validator/validator.go
```diff
@@ -0,0 +1,198 @@
...
+	"github.com/NethermindEth/juno/core"
+	"github.com/NethermindEth/juno/core/felt"
+	"github.com/NethermindEth/juno/mempool"
+)
+
+const defaultTxnPoolSize int = 1024
+
+// Validator is used to validate new proposals. There are two potential flows, and functions must be called in order:
+// Flow 1) Non-empty proposal - ProposalInit(), BlockInfo(), TransactionBatch(), ProposalCommitment(), ProposalFin()
+// Flow 1) Empty proposal - ProposalInit(), ProposalCommitment(), ProposalFin()
```
infrmtcs, line 18:
```suggestion
// Flow 2) Empty proposal - ProposalInit(), ProposalCommitment(), ProposalFin()
```

consensus/validator/validator.go
```diff
@@ -0,0 +1,198 @@
...
+//  1. Some fields we can't get / compute: VersionConstantCommitment, NextL2GasPriceFRI
+//  2. The gas prices. Currently the spec sets eth gas prices, but in v1, these will be dropped
+//     for fri prices.
+func compareProposalCommitment( //nolint:gocyclo // disable linter because this function has a simple pattern
+	p *types.ProposalCommitment,
+	h *core.Header,
+	c *core.BlockCommitments,
+	concatCount *felt.Felt,
+) error {
+	compareProposalField := func(name string, a, b *felt.Felt) error {
```
infrmtcs, line 146:
```suggestion
	compareFeltField := func(name string, a, b *felt.Felt) error {
```

infrmtcs: And I think we can move this function definition outside

consensus/validator/validator.go
```diff
@@ -0,0 +1,198 @@
...
+			return nil
+		}
+		return fmt.Errorf("%s commitment mismatch: proposal=%s commitments=%s", name, a, b)
+	}
+
+	if p.BlockNumber != h.Number {
+		return fmt.Errorf("block number mismatch: proposal=%d header=%d", p.BlockNumber, h.Number)
+	}
+
+	if !p.ParentCommitment.Equal(h.ParentHash) {
```
infrmtcs, line 160: We can't assume that `p.ParentCommitment` is non-nil. I think we can use the compare function for Felt above instead.

consensus/validator/validator.go
```diff
@@ -0,0 +1,198 @@
...
+		return fmt.Errorf("parent hash mismatch: proposal=%s header=%s", p.ParentCommitment.String(), h.ParentHash.String())
+	}
+
+	if !p.Builder.Equal(h.SequencerAddress) {
+		return fmt.Errorf("builder mismatch: proposal=%s header=%s", p.Builder.String(), h.SequencerAddress.String())
+	}
+
+	if p.Timestamp > h.Timestamp {
```
infrmtcs, line 168: Why do we have this comparison? I'm wondering if there is any rule for the timestamp.

consensus/validator/validator.go
```diff
@@ -0,0 +1,198 @@
...
+		return fmt.Errorf("invalid timestamp: proposal timestamp (%d) is later than header timestamp (%d)", p.Timestamp, h.Timestamp)
+	}
+
+	if !p.ProtocolVersion.Equal(blockchain.SupportedStarknetVersion) {
+		return fmt.Errorf("protocol version mismatch: proposal=%s header=%s", p.ProtocolVersion, h.ProtocolVersion)
+	}
```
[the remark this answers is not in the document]
rianhughes: Afaik there isn't any guarantee that the Starknet versions don't implement breaking changes, so we will probably have to update the node every time. Eg if there is a change in the blockifier version, or some commitment function, etc

consensus/validator/validator.go
rianhughes: Why? It's not marked as option in the spec

https://github.com/starknet-io/starknet-p2p-specs/blob/main/p2p/proto/consensus/consensus.proto#L87

consensus/validator/validator.go
```diff
@@ -0,0 +1,198 @@
...
+	c *core.BlockCommitments,
+	concatCount *felt.Felt,
+) error {
+	compareProposalField := func(name string, a, b *felt.Felt) error {
+		if a.IsZero() && b == nil {
+			return nil
+		}
+		if b != nil && a.Equal(b) {
+			return nil
+		}
```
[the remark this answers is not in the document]
rianhughes: Yeah, it's not ideal. 

The reason I did it like this is because the  fields in `ProposalCommitment` are value types, whereas our `core.BlockCommitments` fields are reference types.

In the empty block flow - the `ProposalCommitment` field is set to zero, whereas the `core.BlockCommitments` are nil. This is the first check.

In the non-empty block flow - `ProposalCommitment` field is non-zero, and `core.BlockCommitments` fields are non-nil. This is the second check.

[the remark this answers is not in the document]
rianhughes: Maybe I could explicitly set the `core.BlockCommitments` fields to zero in the empty block flow though??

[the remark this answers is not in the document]
infrmtcs: > The reason I did it like this is because the fields in ProposalCommitment are value types (felt.Felt), whereas our core.BlockCommitments fields are pointers/reference types (*felt.Felt).

I think probably this is because we're using the temporary types. If we switch to the generated protobuf type, the field is optional, so it will align with `core.BlockCommitments`.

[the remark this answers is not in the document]
infrmtcs: > In the empty block flow - the ProposalCommitment field is set to zero, whereas the core.BlockCommitments are nil. This is the first check.
In the non-empty block flow - ProposalCommitment field is non-zero, and core.BlockCommitments fields are non-nil. This is the second check.

I think this is a part of my rationale behind the state machine. The assertion we need in 2 cases are different. In empty block case, we should assert that the field is nil, it can never be non-nil. In the non-empty flow, it can never be nil.

consensus/validator/validator.go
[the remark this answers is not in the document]
rianhughes: Yup, it's a good point, but I don't see any way around this, since we must build blocks with the same protocol version as the builder (eg it makes it into the block hash, and affects blockifier execution).

consensus/validator/validator.go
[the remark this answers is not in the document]
infrmtcs: Yeah, I agree that an honest proposer must send non-nil value, but since protobuf 3 only allows optional fields, a faulty proposer may send nil here, so we must check that case.

consensus/validator/validator.go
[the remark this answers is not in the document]
infrmtcs: As far as I understand, blockifier generally supports all versions up to the latest version as of blockifier release date. I'm thinking that because blockifier does switch starknet version to apply different checks with different versioned constants.
I think instead of taking a protocol version from CLI and check if it's equal to the incoming block protocol version, we can check if incoming block protocol version <= blockifier latest version.

consensus/validator/validator.go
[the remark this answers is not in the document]
rianhughes: Given that proto3 treats Hash and Addr (and other message types) as optional, but the spec requires it (eg we shouldn't consider a ProposalInit with an absent proposer field as valid), we will need to validate the messages.

Imo, this should be done before calling Validator, for safety and the avoid using resources.

consensus/validator/validator.go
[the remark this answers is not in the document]
rianhughes: Block hash:
https://github.com/NethermindEth/juno/blob/main/core/block.go#L243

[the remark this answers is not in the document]
rianhughes: Updated

builder/builder.go
```diff
@@ -177,6 +200,7 @@ func (b *Builder) InitPendingBlock() error {
 			ParentHash:       header.Hash,
 			Number:           header.Number + 1,
 			SequencerAddress: &b.ownAddress,
+			ProtocolVersion:  b.protocolVersion.String(),
```
infrmtcs, line 203: Why isn't this extracted from the proposal commitment?

rianhughes: >I'm wondering if we can remove this protocol-version from CLI. I think which version is expected should be determined by the network (not 100% sure about this for now) instead of by command line argument.

It's also useful for the sequencer.

consensus/validator/init_test.go
```diff
@@ -0,0 +1,5 @@
+package validator
```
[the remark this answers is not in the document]
rianhughes: Nope, it's how we register the custom Juno types with the encoder. Every sub-pkg uses this, eg vm, rpc, sync, blockchain, etc

blockchain/blockchain.go
```diff
@@ -507,9 +535,53 @@ func (b *Blockchain) Finalise(
...
+		block, err := core.GetBlockByNumber(txn, block.Number)
+		if err != nil {
+			return err
+		}
+		su, err := core.GetStateUpdateByBlockNum(txn, block.Number)
+		if err != nil {
+			return err
+		}
+		newBlock = block
+		newSU = su
```
infrmtcs:
```suggestion
		if newBlock, err = core.GetBlockByNumber(txn, block.Number); err != nil {
			return err
		}

		if newSU, err = core.GetStateUpdateByBlockNum(txn, block.Number); err != nil {
			return err
		}
```

blockchain/blockchain.go
```diff
@@ -481,23 +481,51 @@ func (b *Blockchain) revertHead(txn db.IndexedBatch) error {
...
 	sign BlockSignFunc,
-) error {
-	return b.database.Update(func(txn db.IndexedBatch) error {
+) (*core.Block, *core.StateUpdate, *core.BlockCommitments, *felt.Felt, error) {
+	var newBlock *core.Block
+	var newSU *core.StateUpdate
+	var newCommitments *core.BlockCommitments
+	var concatCount *felt.Felt
+
+	simulate := func(txn db.IndexedBatch) error {
```
infrmtcs: Why do we need to put the simulate function here?

blockchain/blockchain.go
```diff
@@ -481,23 +481,51 @@ func (b *Blockchain) revertHead(txn db.IndexedBatch) error {
...
+		stateUpdate.BlockHash = blockHash
+
+		// Starknet consensus requires zero values for empty blocks
+		if block.TransactionCount == 0 {
+			newCommitments = &core.BlockCommitments{
+				TransactionCommitment: new(felt.Felt).SetUint64(0),
+				EventCommitment:       new(felt.Felt).SetUint64(0),
+				ReceiptCommitment:     new(felt.Felt).SetUint64(0),
+				StateDiffCommitment:   new(felt.Felt).SetUint64(0),
+			}
```
infrmtcs: Why isn't this default value be returned by `core.BlockHash`?

blockchain/blockchain.go
```diff
@@ -507,9 +535,53 @@ func (b *Blockchain) Finalise(
...
+) error {
+	finaliseFn := func(txn db.IndexedBatch) error {
+		if err := b.updateStateRoots(txn, block, stateUpdate, newClasses); err != nil {
+			return err
+		}
+		commitments, err := b.calculateBlockHash(block, stateUpdate)
+		if err != nil {
+			return err
+		}
+		if err := b.signBlock(block, stateUpdate, sign); err != nil {
```
infrmtcs, line 600: I think this isn't applicable in this case. Blocks built by other nodes are signed by them, we can't do anything here.

blockchain/blockchain.go
```diff
@@ -481,23 +481,51 @@ func (b *Blockchain) revertHead(txn db.IndexedBatch) error {
...
+// Simulate returns what the new completed header and state update would be if the
+// provided block was added to the chain.
+func (b *Blockchain) Simulate(
 	block *core.Block,
 	stateUpdate *core.StateUpdate,
 	newClasses map[felt.Felt]core.Class,
 	sign BlockSignFunc,
-) error {
-	return b.database.Update(func(txn db.IndexedBatch) error {
+) (*core.Block, *core.StateUpdate, *core.BlockCommitments, *felt.Felt, error) {
```
infrmtcs: I think 4 return outputs should be put into a struct.

blockchain/blockchain.go
infrmtcs: Now I see the check for `sign == nil` inside `signBlock`.

builder/builder.go
```diff
@@ -80,6 +93,10 @@ func (b *Builder) WithPlugin(junoPlugin plugin.JunoPlugin) *Builder {
 	return b
 }
 
+func (b *Builder) Network() *utils.Network {
```
infrmtcs: Why do we need this function? Is it only for test?

blockchain/blockchain.go
rianhughes: I think it's neater

blockchain/blockchain.go
rianhughes: The default value depends on the block version. Sometimes we don't compute the commitments and they are set to `nil`. Other times, we will explicitly set the commitment to `nil` if the block is empty, and other times, we will set it to `zero`.

Although, now that I think about it, this should be handled in validator, since it's only relevant for consensus, not regular starknet.

builder/builder.go
rianhughes: We don't define the latest version anywhere. We do have a "const" that tracks what the latest version we are aware of is, but this would require setting a new variable.

builder/builder.go
rianhughes: This can be removed

builder/builder.go
rianhughes: We don't currently have a const that tracks the current supported version, only a variable that tracks the latest known version. We could introduce a variable for this. But it would mean coupling the Sequencer and Consensus logic, which I don't think we should do.

infrmtcs: If we assume that starknet will follow the same practice as other chains, I believe there will be a config file or something similar that states the activation block for each version. It should be an independent package, and other packages including both sequencer and consensus can depend on.

consensus/validator/validator.go
```diff
@@ -0,0 +1,198 @@
...
+		a.txnPool[i] = types.Transaction{}
+	}
+	a.latestExecutedTxn = 0
+
+	return a.builder.ProposalInit(pInit)
+}
+
+// BlockInfo sets the pending header according to BlockInfo
+func (a *validator[V, H, A]) BlockInfo(blockInfo *types.BlockInfo) {
+	a.builder.SetBlockInfo(blockInfo)
```
infrmtcs: I think we also need to validate the fields in block info. It can be a TODO though.

consensus/validator/validator.go
```diff
@@ -0,0 +1,198 @@
...
+// ProposalFin executes the provided transactions, and stores the result in the pending state
+func (a *validator[V, H, A]) ProposalFin(proposalFin types.ProposalFin) error {
+	pendingBlock := a.builder.PendingBlock()
+
+	proposerCommitmentFelt := felt.Felt(proposalFin)
+	if !proposerCommitmentFelt.Equal(pendingBlock.Hash) {
+		return fmt.Errorf("proposal fin: commitments do not match: expected=%s actual=%s",
+			proposerCommitmentFelt.String(), pendingBlock.Hash.String())
+	}
+	return a.builder.Finalise(nil)
```
infrmtcs: I'm understanding that this call will write the block to the database. I think it should be instead sent to the proposal channel (#2839).

consensus/validator/validator.go
```diff
@@ -0,0 +1,198 @@
...
+			Transaction:   txns[start].Transaction,
+			DeclaredClass: txns[start].Class,
+			PaidFeeOnL1:   txns[start].PaidFeeOnL1,
+		}
+	}
+
+	if err := a.builder.ExecuteTxns(txnsToExecute); err != nil {
+		return err
+	}
+	a.latestExecutedTxn = end + 1
```
infrmtcs: It's actually next transaction instead of latest executed transaction.

consensus/validator/validator.go
```diff
@@ -0,0 +1,198 @@
...
+	// ProposalCommitment checks the set of proposed commitments against those generated locally.
+	ProposalCommitment(proposalCommitment *types.ProposalCommitment) error
+
+	// ProposalFin compares the provided commitment with that generated locally
+	ProposalFin(proposalFin types.ProposalFin) error
+}
+
+type validator[V types.Hashable[H], H types.Hash, A types.Addr] struct {
+	builder           *builder.Builder    // Builder manages the pending block and state
+	txnPool           []types.Transaction // Txns may not arrive in order
```
infrmtcs: For validator, transactions should arrive in order, because we already have the sequence number for each proposal part. Part re-ordering to match the sequence number should be handled by the P2P layer.

consensus/validator/validator.go
```diff
@@ -0,0 +1,198 @@
...
+
+	if !p.ParentCommitment.Equal(h.ParentHash) {
+		return fmt.Errorf("parent hash mismatch: proposal=%s header=%s", p.ParentCommitment.String(), h.ParentHash.String())
+	}
+
+	if err := compareFeltField("proposer address", &p.Builder, h.SequencerAddress); err != nil {
+		return err
+	}
+
+	if p.Timestamp > h.Timestamp {
```
infrmtcs, line 144: According to the specs, timestamp is actually included in both `BlockInfo` and `ProposalCommitment`. I think we should run 2 different validation here:
- `BlockInfo`: Check if timestamp is valid (e.g, later than the parent block timestamp).
- `ProposalCommitment`: Check if it matches the `timestamp` in `BlockInfo`.

blockchain/blockchain.go
rianhughes: This is now handled in the consensus layer

consensus/validator/validator.go
rianhughes: Yup, we'll need to do that. Although it should be done before calling validator imo, and will also require using the real types, not the current tmp ones

consensus/validator/validator.go
rianhughes: I think we only call this when consensus has been reached, so we are able to commit the block at this point

```
// Identifies the content proposed (and executed). Consensus is reached on the value contained here.
message ProposalFin {
    Hash proposal_commitment = 1;
}
```

consensus/validator/validator.go
rianhughes: Given the txns are guaranteed to arrive in order, we can remove the txn pool entirely.

consensus/validator/validator.go
rianhughes: Given the txns are guaranteed to arrive in order, we can remove the txn pool entirely, along with this counter

consensus/validator/validator.go
rianhughes: I think we should ask the SN guys about this since the specs aren't clear, it can probably be an Todo for now since I don't think it's critical for the MVP

blockchain/blockchain.go
rodrodros: What would be your suggested location @infrmtcs?

blockchain/blockchain.go
```diff
@@ -481,35 +481,104 @@ func (b *Blockchain) revertHead(txn db.IndexedBatch) error {
...
+	stateUpdate *core.StateUpdate,
+	newClasses map[felt.Felt]core.Class,
+	sign BlockSignFunc,
+) (SimulateResult, error) {
+	var newBlock *core.Block
+	var newSU *core.StateUpdate
+	var newCommitments *core.BlockCommitments
+	var concatCount *felt.Felt
+
+	simulate := func(txn db.IndexedBatch) error {
```
rodrodros: Why are you using `db.IndexedBatch` when it's recommended against it by it's description (cc @MaksymMalicki )

rianhughes: If we want to Update the DB, we need to call 
```
Update(func(IndexedBatch) error) error
```
which currently only accepts `IndexedBatch`. This is what we currently use in `blockchain.Store()`. We should update it, but it's outside of the scope of this PR

consensus/validator/validator_test.go
```diff
@@ -0,0 +1,379 @@
...
+
+	blockNumber := uint64(100)
+
+	p := &types.ProposalCommitment{
+		BlockNumber:           blockNumber,
+		ParentCommitment:      *utils.HexToFelt(t, "111"),
+		Builder:               *proposer,
+		Timestamp:             1000,
+		ProtocolVersion:       *blockchain.SupportedStarknetVersion,
+		ConcatenatedCounts:    *utils.HexToFelt(t, "1"),
```
rodrodros: Don't we have a constructor that uses a number?

blockchain/blockchain.go
infrmtcs: I think top-level functions are slightly more performant than closures. I wrote a simple benchmark and compared these 2, top-level functions are 5% faster.
Although the differences don't matter that much, but in this case I think write a top-level function for simulate is even clearer. Something like `func simulate(txn db.IndexedBatch, block *core.Block,...) (..., error)`

consensus/validator/validator_test.go
rianhughes: Don't think so, but let me know otherwise

infrmtcs: I think you can use `new(felt.Felt).SetUint64(1)`

blockchain/blockchain.go
```diff
@@ -481,35 +481,105 @@ func (b *Blockchain) revertHead(txn db.IndexedBatch) error {
...
 	stateUpdate *core.StateUpdate,
 	newClasses map[felt.Felt]core.Class,
 	sign BlockSignFunc,
 ) error {
-	return b.database.Update(func(txn db.IndexedBatch) error {
+	finaliseFn := func(txn db.IndexedBatch) error {
 		if err := b.updateStateRoots(txn, block, stateUpdate, newClasses); err != nil {
 			return err
 		}
-
```
infrmtcs, line 521: nit: Why do we remove the blank lines here. In my opinion, the blank lines help group code into blocks. For example, the `calculateBlockHash` line and `if err != nil` check can be considered as a block. Just personal opinion, feel free to ignore though.

blockchain/blockchain.go
```diff
@@ -481,35 +481,105 @@ func (b *Blockchain) revertHead(txn db.IndexedBatch) error {
...
+}
+
+type SimulateResult struct {
+	Block            *core.Block
+	StateUpdate      *core.StateUpdate
+	BlockCommitments *core.BlockCommitments
+	ConcatCount      *felt.Felt
+}
+
+// Finalise checks the block correctness and appends it to the chain
```
infrmtcs, line 585: I think the existing behaviour can run everything in one go because we don't have to conduct the consensus algorithm. Now because we have `Simulate` function, so `Finalise` function should reuse the result of `Simulate` instead of re-calculating from scratch.

blockchain/blockchain.go
```diff
@@ -481,35 +481,105 @@ func (b *Blockchain) revertHead(txn db.IndexedBatch) error {
...
+			return err
+		}
+
+		if newBlock, err = core.GetBlockByNumber(txn, block.Number); err != nil {
+			return err
+		}
+
+		if newSU, err = core.GetStateUpdateByBlockNum(txn, block.Number); err != nil {
+			return err
+		}
```
infrmtcs: Why do we need to load them from database if we're creating them here anyway?

blockchain/blockchain.go
```diff
@@ -481,35 +481,105 @@ func (b *Blockchain) revertHead(txn db.IndexedBatch) error {
...
+			stateUpdate.StateDiff.Length(),
+			block.L1DAMode)
+
+		if err := b.signBlock(block, stateUpdate, sign); err != nil {
+			return err
+		}
+
+		if err := b.storeBlockData(txn, block, stateUpdate, newCommitments); err != nil {
+			return err
+		}
```
infrmtcs: Also, do we need to write to the database?

blockchain/blockchain.go
```diff
@@ -481,35 +481,105 @@ func (b *Blockchain) revertHead(txn db.IndexedBatch) error {
...
+
+		if newSU, err = core.GetStateUpdateByBlockNum(txn, block.Number); err != nil {
+			return err
+		}
+		return nil
+	}
+
+	// Simulate without commit
+	txn := b.database.NewIndexedBatch()
```
infrmtcs: If we can remove the lines writing the data above, we can actually use a read snapshot instead.

builder/builder.go
```diff
@@ -463,3 +500,83 @@ func (b *Builder) SubscribeNewHeads() sync.NewHeadSubscription {
...
+
+// ExecutePending updates the pending block and state-update
+func (b *Builder) ExecutePending() (*core.BlockCommitments, *felt.Felt, error) {
+	pending, err := b.Pending()
+	if err != nil {
+		return nil, nil, err
+	}
+	simulateResult, err := b.bc.Simulate(pending.Block, pending.StateUpdate, pending.NewClasses, nil)
+	pending.Block = simulateResult.Block
+	pending.StateUpdate = simulateResult.StateUpdate
```
infrmtcs: We're passing `pending.Block` and `pending.StateUpdate` as pointers and mutating them in `bc.Simulate` anyway, so we don't need to reassign here.

infrmtcs: Additionally:
- If we assume that `ExecutePending` and other functions doesn't run concurrently and assume the caller's responsibility, we don't need `atomic.Pointer` to guard.
- If we assume that they can run concurrently, `atomic.Pointer` can only guard multiple goroutines reading and writing the pointer, but it can't guard the reads and writes of the fields, so race condition can still happens.

builder/builder.go
```diff
@@ -463,3 +500,83 @@ func (b *Builder) SubscribeNewHeads() sync.NewHeadSubscription {
...
+	su := core.StateUpdate{
+		StateDiff: &emptyStateDiff,
+	}
+	pending := sync.Pending{
+		Block:       &pendingBlock,
+		StateUpdate: &su,
+		NewClasses:  newClasses,
+	}
+	b.pendingBlock.Store(&pending)
+	b.headState, b.headCloser, err = b.bc.HeadState()
```
infrmtcs: I assume that multiple `Builder` instance can run at the same time (because we can have multiple valid proposals for different rounds).
We need to check `ChainHeight` here, because race condition can happen with this sequence:
- `b.bc.HeadsHeader()` got block N.
- Block N+1 is committed.
- `b.bc.HeadState()` got block N+1.

consensus/validator/validator.go
infrmtcs: If we only run this when consensus has been reached, then that means we can prevote/precommit it and then refuse to commit because hash doesn't match. I believe it should be part of the validation, and it helps to confirm if the proposal implementation is correct where hash needs to match.

consensus/validator/validator_test.go
```diff
@@ -0,0 +1,379 @@
...
+		StateDiffCommitment:   *commitment,
+		TransactionCommitment: *commitment,
+		EventCommitment:       *commitment,
+		ReceiptCommitment:     *commitment,
```
infrmtcs: nit: Can we use different values for each of the fields? I think it's easy to make a mistake like `if a.StateDiffCommitment != b.TransactionCommitment`.

consensus/validator/validator_test.go
```diff
@@ -0,0 +1,379 @@
...
+		L1DAMode:              0,
+	}
+
+	h := &core.Header{
+		Number:           blockNumber,
+		ParentHash:       utils.HexToFelt(t, "111"),
+		SequencerAddress: proposer,
+		Timestamp:        1000,
+		ProtocolVersion:  blockchain.SupportedStarknetVersion.String(),
+		L1DAMode:         0,
```
infrmtcs: nit: Can we use either `core.Blob` or `core.CallData` here?

consensus/validator/validator_test.go
rianhughes: Ah, right, I thought he meant some function in utils, my bad!

blockchain/blockchain.go
rianhughes: Cool, I'll update it then. Interesting that the top-level is faster

rianhughes: Apparently closures are not inlined, and may cause heap allocation, whereas top-level functions may be inlined. I would have expected the closure to be inlined..

consensus/validator/validator_test.go
```diff
@@ -0,0 +1,379 @@
...
+		other := &felt.Felt{}
+		other.SetUint64(99)
+		p.TransactionCommitment = *other
+		require.Error(t, compareProposalCommitment(p, h, c, concatCount))
+	})
+
+	t.Run("MismatchedEventCommitment", func(t *testing.T) {
+		other := &felt.Felt{}
+		other.SetUint64(99)
+		p.EventCommitment = *other
```
infrmtcs, line 366: The tests are modifying the fields on the same var. This means even if we make a mistake and forget to check for `ReceiptCommitment`, the next test cannot catch it, because we already modify `p.EventCommitment` in this test which cause `compareProposalCommitment` to fail regardless the value of `p.ReceiptCommitment`

blockchain/blockchain.go
rianhughes: I ended up inlining simulate directly, it turned out to be cleaner than moving it to a top-level function

consensus/validator/validator_test.go
infrmtcs: I think easiest approach would be declaring `p := types.ProposalCommitment{...}` on top and set `p := p` at the beginning of each test.

consensus/validator/validator_test.go
rianhughes: Updated

blockchain/blockchain.go
rianhughes: No real reason, probably thought it was a little neater, happy to have the additional lines though

blockchain/blockchain.go
rianhughes: I've removed these

blockchain/blockchain.go
rianhughes: We don't, I was just being an idiot. I've updated the code

blockchain/blockchain.go
rianhughes: How? We need to update the tries (/call `updateStateRoots`) which requires passing in an IndexedBatch txn

builder/builder.go
rianhughes: > We're passing pending.Block and pending.StateUpdate as pointers and mutating them in bc.Simulate anyway, so we don't need to reassign here.

Yup, I've removed these lines

consensus/validator/validator_test.go
rianhughes: Sure, updated

consensus/validator/validator_test.go
rianhughes: Sure, updated

consensus/validator/validator_test.go
rianhughes: Yup, good catch, updated the code

blockchain/blockchain.go
rianhughes: Yup, we don't need to re-compute the state etc. Ive introduced a new function that just stores the pending block

rianhughes:
@infrmtcs  I've created a new issue to keep track of the issues related to handling multiple concurrent proposals

builder/builder.go
rianhughes: I've created a new issue to keep track of the issues related to handling multiple concurrent proposals. I think it would be neater to handle this separately at this point

https://github.com/NethermindEth/juno/issues/2868

builder/builder.go
rianhughes: I've created a new issue to keep track of the issues related to handling multiple concurrent proposals. I think it would be neater to handle this separately at this point

https://github.com/NethermindEth/juno/issues/2868

consensus/validator/validator.go
rianhughes: Left a todo for this, to unblock while we wait for SN

consensus/validator/validator.go
rianhughes: The comment left on the specs is confusing / misleading. We decided to update PropsalFin() to just ensure the blockhashes match.

consensus/validator/validator.go
rodrodros: Issue please, no to-dos

rodrigo-pino (approved):

infrmtcs (approved):

PR #2854, opened 2025-05-26
feature: implement consensus Proposer interface

rianhughes:


rpc/v8/transaction.go
```diff
@@ -217,29 +217,17 @@ type ResourceBoundsMap struct {
...
-		L2Gas *ResourceBounds `json:"l2_gas"`
-	}
-
-	temp := tempResourceBoundsMap{
-		L1Gas: r.L1Gas,
-		L2Gas: r.L2Gas,
-	}
-
-	// Check if L1DataGas is nil, if it is, remove it from the struct/map
+	// Check if L1DataGas is nil, if it is, provide default values
```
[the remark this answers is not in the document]
rianhughes: Yes, this looks strange..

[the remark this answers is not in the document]
rianhughes: Oh, looks like merging main fixed this..

consensus/proposer/proposer.go
```diff
@@ -0,0 +1,229 @@
...
+		if !p.builder.MempoolIsEmpty() {
+			break
+		}
+
+		select {
+		case <-ctx.Done():
+			return types.BlockInfo{}, false
+		case <-ticker.C:
+		}
+	}
```
infrmtcs, line 124: Why do we need this? I think it's better if we can preemptively build the block info even before receiving transactions.

rianhughes: Yup, that would be ideal. However, imagine the case where block times are 10 minutes, and when we start building the block at minute 0 the mempool is closed. There are two scenarios:
1. Over the next ten minutes we don't receive any txns. Then we will propose an empty block.
2. At minute 9 we receive a txn, and we can build and propose a non-empty block. 

The problem is, between minutes 0 and 9 we don't know if the block will be empty or not, so can't send the blockInfo to our peers at this point. 

My proposed solution to this is to check if txns are in the mempool. If there are, then we can propose a non-empty block. If there aren't, then we should check periodically until either 1) we get a txn, and then continue with the non-empty block flow, or 2) we hit our timeout, and propose an empty block.

rianhughes: We could in principle just check at time 0 whether we can txns in the mempool or not, and then decide whether to build an empty block or not. But I think that's bad for UX, since it increases the amount of time on average that users need to wait to have their txn included in a block.

rianhughes:
I opened a fresh branch to solve an annoying rebase. https://github.com/NethermindEth/juno/pull/2873

PR #2855, opened 2025-05-27
refactor: remove the Addr and Hash generics in consensus

rianhughes:
### Summary

The consensus implementation currently defines `Hash` and `Addr` as generic interfaces to support flexibility across protocols. This was originally intended to allow the consensus logic to work with any hash and address types that can be represented as `[4]uint`.

However, in the context of the Starknet protocol, both the address and hash types are explicitly defined as `felt.Felt`, and this is not expected to change. As a result, the use of generics here is unnecessary and introduces avoidable complexity.

---

### Motivation

- **Unnecessary Boilerplate:** Since `Hash` and `Addr` are used pervasively across the consensus package, requiring them to be passed explicitly as generic parameters to every struct and function results in significant boilerplate and reduces readability.
  
- **Interface Overhead:** Defining these as interfaces requires explicit type declarations, type assertions, and additional complexity that brings no real benefit in this fixed-type context.

- **Improved Ergonomics:** By redefining `Hash` and `Addr` as aliases (e.g., `type Addr felt.Felt`), we significantly reduce syntactic noise and make the code more straightforward and maintainable.

---

### Change

This PR:
- Replaces the generic `Hash` and `Addr` interfaces with type aliases to `felt.Felt`.
- Updates the affected types and function signatures to remove unnecessary generic parameters.

This change simplifies the consensus implementation while preserving all intended functionality under the assumption of fixed `felt.Felt`-based types.

---

### Next (this can be done in a separate PR)

Investigate whether the `Hashable` interface can also be removed by explicitly defining the concrete value types used in the Starknet protocol (e.g., `Block`, `StateUpdate`, etc.). 

Since these types are known and fixed within the protocol, relying on an interface abstraction may be unnecessary. Replacing `Hashable` with concrete types could further simplify the codebase and eliminate another layer of indirection.

### DIFF
--- a/consensus/starknet/starknet.go
+++ b/consensus/starknet/starknet.go
@@ -4,30 +4,23 @@ import (
 	"github.com/NethermindEth/juno/consensus/types"
 	"github.com/NethermindEth/juno/core/address"
 	"github.com/NethermindEth/juno/core/felt"
-	"github.com/NethermindEth/juno/core/hash"
 )
 
 // TODO: Replace with the actual value type.
 type Value uint64
 
-func (v Value) Hash() hash.Hash {
-	return hash.Hash(*new(felt.Felt).SetUint64(uint64(v)))
+func (v Value) Hash() types.Hash {
+	return types.Hash(*new(felt.Felt).SetUint64(uint64(v)))
 }
 
 type (
 	Address = address.Address
-	Hash    = hash.Hash
+	Hash    = types.Hash
 
-	Message       = types.Message[Value, Hash, Address]
-	Proposal      = types.Proposal[Value, Hash, Address]
-	Prevote       = types.Prevote[Hash, Address]
-	Precommit     = types.Precommit[Hash, Address]
-	Vote          = types.Vote[Hash, Address]
-	MessageHeader = types.MessageHeader[Address]
-	Messages      = types.Messages[Value, Hash, Address]
+	Message  = types.Message[Value]
+	Proposal = types.Proposal[Value]
+	Messages = types.Messages[Value]
 
-	Action             = types.Action[Value, Hash, Address]
-	BroadcastProposal  = types.BroadcastProposal[Value, Hash, Address]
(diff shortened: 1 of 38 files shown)


consensus/types/types.go
```diff
@@ -1,14 +1,20 @@
 package types
 
-type Addr interface {
-	~[4]uint64
+import "github.com/NethermindEth/juno/core/felt"
+
+// Hashable's Hash() is used as ID()
+type Hash felt.Felt
+
+type Addr felt.Felt
```
infrmtcs, line 8: Why don't we use existing types in `core` instead? That would reduce one layer of conversion.

consensus/starknet/starknet.go
```diff
@@ -4,30 +4,23 @@ import (
...
-func (v Value) Hash() hash.Hash {
-	return hash.Hash(*new(felt.Felt).SetUint64(uint64(v)))
+func (v Value) Hash() types.Hash {
+	return types.Hash(*new(felt.Felt).SetUint64(uint64(v)))
 }
 
 type (
 	Address = address.Address
-	Hash    = hash.Hash
+	Hash    = types.Hash
```
infrmtcs, line 18: Why don't we use existing hash type from `core` instead?

consensus/types/types.go
rianhughes: Yup, also happy to do that

consensus/starknet/starknet.go
rianhughes: Yup, also happy to do that

infrmtcs:
> I just don't think it's worth it, particularly given Starknet is using their own custom Tendermint logic, not the vanilla implementation.

As far as I understand, the only changes in Starknet are not in the core part of the algorithm including state machine and driver, but instead in the communication part, namely proposal stream and re-broadcast.

rianhughes:
To do: investigate if we can remove the interfaces, but keep the Tendermint-Starknet abstraction layer

PR #2856, opened 2025-05-27
feat: Add scaffolding for proposer sending messages to P2P

infrmtcs:


consensus/p2p/proposer/proposer_dispatcher.go
```diff
@@ -0,0 +1,188 @@
...
+		if err != nil {
+			yield(nil, err)
+			return
+		}
+
+		for txBatch := range slices.Chunk(txStream, txBatchSize) {
+			if !yield(d.fromTransactions(txBatch)) {
+				return
+			}
+		}
```
rianhughes, line 70: At this point we have built the entire block right? So why bother breaking the txns into chunks?

infrmtcs: We may want to send out transactions in chunk even if we have everything. Imagine the case where there are 1000 transactions, we want to gradually broadcast them instead of building a big message of 1k transactions and then the receiver has to process 1k transactions at a time.

rodrigo-pino (approved):

PR #2857, opened 2025-05-28
refactor: replace for i := range b.N with for b.Loop() and remove calls to b.StopTimer, b.StartTimer, and b.ResetTimer

tongjicoder:
replace "for i := range b.N" or "for range b.N" in a benchmark with "for b.Loop()", and remove any preceding calls to b.StopTimer, b.StartTimer, and b.ResetTimer.


tongjicoder:
> Bot


Sorry to bother, but I'm not a bot.

PR #2860, opened 2025-05-28
refactor: Add Commit action

infrmtcs:
Problem: Proposal demux fails when receiving newer block proposals before committing older blocks (when node lags behind network).

Solution:
- This PR makes Commit an Action to enable Driver notifications for block commits
- Next step is change the P2P PR to block execution until parent commits.

Another reason for this is to gradually move the side effects out of the state machine.

### DIFF
--- a/consensus/driver/driver_test.go
+++ b/consensus/driver/driver_test.go
@@ -6,6 +6,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/NethermindEth/juno/consensus/db"
 	"github.com/NethermindEth/juno/consensus/driver"
 	"github.com/NethermindEth/juno/consensus/mocks"
 	"github.com/NethermindEth/juno/consensus/p2p"
@@ -13,9 +14,10 @@ import (
 	"github.com/NethermindEth/juno/consensus/types"
 	"github.com/NethermindEth/juno/core/felt"
 	"github.com/NethermindEth/juno/core/hash"
-	"github.com/NethermindEth/juno/db/memory"
+	"github.com/NethermindEth/juno/db/pebble"
 	"github.com/NethermindEth/juno/utils"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"go.uber.org/mock/gomock"
 )
 
@@ -23,6 +25,8 @@ type (
 	broadcaster[M starknet.Message] = p2p.Broadcaster[M, starknet.Value, starknet.Hash, starknet.Address]
 	listeners                       = p2p.Listeners[starknet.Value, starknet.Hash, starknet.Address]
 	broadcasters                    = p2p.Broadcasters[starknet.Value, starknet.Hash, starknet.Address]
+	tendermintDB                    = db.TendermintDB[starknet.Value, starknet.Hash, starknet.Address]
+	blockchain                      = driver.Blockchain[starknet.Value, starknet.Hash]
 )
 
 const (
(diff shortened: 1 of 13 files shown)


consensus/driver/driver_test.go
```diff
@@ -191,7 +221,18 @@ func TestDriver(t *testing.T) {
 
 	stateMachine := mocks.NewMockStateMachine[starknet.Value, hash.Hash, starknet.Address](ctrl)
 	stateMachine.EXPECT().ReplayWAL().AnyTimes().Return() // ignore WAL replay logic here
-	driver := driver.New(memory.New(), stateMachine, mockListeners(proposalCh, prevoteCh, precommitCh), broadcasters, mockTimeoutFn)
+
+	commitAction := (*starknet.Commit)(utils.HeapPtr(getRandProposal(random)))
```
rodrodros: Not that it is important but why not
```go
commitAction := (starknet.Commit)(getRandProposal(random))
```

And then use a reference to `&commitAction`

consensus/tendermint/process.go
```diff
@@ -71,16 +71,19 @@ func (t *stateMachine[V, H, A]) processLoop(action types.Action[V, H, A], recent
...
 		actions = append(actions, action)
-		action = t.process(recentlyReceivedRound)
+		action, commit = t.process(recentlyReceivedRound)
 	}
 
 	return actions
 }
 
-func (t *stateMachine[V, H, A]) process(recentlyReceivedRound *types.Round) types.Action[V, H, A] {
+func (t *stateMachine[V, H, A]) process(recentlyReceivedRound *types.Round) (types.Action[V, H, A], *types.Commit[V, H, A]) {
```
rodrodros: What is the rationale behind this change?

rodrodros: Answered my self, missed PR description

rodrigo-pino (approved):

consensus/tendermint/process.go
infrmtcs: There are a couple of reasons:
- Primary problem: Other components need to be notified when the algorithm commits. The P2P component can receive proposal for block N+1, but it should only execute the proposal once block N is committed and written to the database.
- Secondary problem: Side effects are currently done inside the state machine, including writing to consensus DB (via Flush and SetWALEntry) and main DB (via blockchain.Commit). In my opinion, they should be done outside of the state machine, in the Driver instead, to keep the state machine pure.
- Considered alternative: Primary problem can still be solved by checking if height is different pre-execution and post-execution, but secondary problem is not addressed in this case.

consensus/driver/driver_test.go
infrmtcs: Agree, let me update this.

consensus/tendermint/process.go
```diff
@@ -71,16 +71,19 @@ func (t *stateMachine[V, H, A]) processLoop(action types.Action[V, H, A], recent
...
 	cachedProposal := t.findProposal(t.state.round)
```
rianhughes: Seperating the Commit, which is also an action, from the other actions is a bit strange / confusing. Why did you decide to do this?

It seems we could massively simplify this liogic by appending the commit action to the other actions in doCommitValue(). Eg

```
// simpified action handling
action := t.process(recentlyReceivedRound)
	for action != nil {
		actions = append(actions, action)
		action = t.process(recentlyReceivedRound)
	}

// simplified signature
func (t *stateMachine[V, H, A]) process(recentlyReceivedRound *types.Round) (types.Action[V, H, A]) {
    ...
}
```

infrmtcs: After separating it, rule_commit_value is the only rule that can return more than 1 action. So in order to accommodate this change, process needs to either return a slice of 2 elements, or 2 pointer/interface for action and optionally a commit.

rianhughes: Ah yes, sorry the process function would need to return a slice. Given commit is also an action, I feel like this is cleaner than the current approach. What do you think?

eg
```
// simpified action handling
processedActions := t.process(recentlyReceivedRound)
actions = append(actions, processedActions...)
	for len(processedActions) != 0 {
		processedActions = t.process(recentlyReceivedRound)
	}

// simplified signature
func (t *stateMachine[V, H, A]) process(recentlyReceivedRound *types.Round) ([]types.Action[V, H, A]) {
    ...
}
```

infrmtcs: Yeah, I also think so, but I think return 1 pointer + 1 interface would be more performant than allocating a slice of 1-2 elements

rianhughes (approved):

PR #2864, opened 2025-06-02
feat(state): new state package + state clean cache

MaksymMalicki:


core/class.go
```diff
@@ -255,3 +258,31 @@ func VerifyClassHashes(classes map[felt.Felt]Class) error {
...
+	size := 8 + len(classEnc)
+	buf := make([]byte, size)
+	binary.BigEndian.PutUint64(buf[:8], d.At)
+	copy(buf[8:], classEnc)
+
+	return buf, nil
+}
+
+func (d *DeclaredClass) UnmarshalBinary(data []byte) error {
+	if len(data) < 8 { //nolint:mnd
```
rodrodros: Add a reason for the nolint, might as well use a named constant directly

core/state/cache.go
```diff
@@ -0,0 +1,145 @@
...
+	links      map[felt.Felt]felt.Felt  // child -> parent
+	oldestRoot felt.Felt                // root of the oldest layer in the cache
+}
+
+func newStateCache() *stateCache {
+	return &stateCache{
+		diffs: make(map[felt.Felt]*diffCache),
+		links: make(map[felt.Felt]felt.Felt),
+	}
+}
```
rodrodros: I think it's safe to return by value here?

core/state/cache.go
```diff
@@ -0,0 +1,145 @@
...
+
+func (c *stateCache) PushLayer(stateRoot, parentRoot *felt.Felt, diff *diffCache) {
```
rodrodros, line 43: Could you add a small description to this function please. Looks good otherwise 👍

core/state/cache.go
```diff
@@ -0,0 +1,145 @@
...
+
+type diffCache struct {
+	storageDiffs      map[felt.Felt]map[felt.Felt]*felt.Felt // addr -> {key -> value, ...}
+	nonces            map[felt.Felt]*felt.Felt               // addr -> nonce
+	deployedContracts map[felt.Felt]*felt.Felt               // addr -> class hash
+}
+
+type stateCache struct {
+	diffs      map[felt.Felt]*diffCache // state root -> state diff
+	links      map[felt.Felt]felt.Felt  // child -> parent
```
rodrodros: Am I getting this correctly, does this `links` represents a tree of the form

`ch1 -> ch2 -> ch3 -> parent`, where `ch1` is child of `ch2` and so on?

If that's the case I strongly believe a map is not the right data structure here. It looks that a linked list with two pointers might be a better fit for the job. _If I got it correctly_, then I think the best is to implement our own, didn't find any convincing package.

The pros from this is that we can now do all the cache modification in constant times, the cons is that all getter methods need to iterate through the list. To solve this, then we can add another map to the mix to be (key, ptr to linked listnode). Then we have O(1) (instead of O(128)) op for pushing/popping a layer and O(log(128)) for getter operations (stays the same).

The space cost is 128 additional pointers (64 bits each) to the current implementation. I think we are fine. 

Let me know what you think.

core/state/cache.go
```diff
@@ -0,0 +1,145 @@
...
+	oldestRoot felt.Felt                // root of the oldest layer in the cache
```
rodrodros: It also means you don't have to keep this value since it will be the first element of the list

core/state/cache.go
```diff
@@ -0,0 +1,145 @@
...
+)
+
+type diffCache struct {
+	storageDiffs      map[felt.Felt]map[felt.Felt]*felt.Felt // addr -> {key -> value, ...}
+	nonces            map[felt.Felt]*felt.Felt               // addr -> nonce
+	deployedContracts map[felt.Felt]*felt.Felt               // addr -> class hash
+}
+
+type stateCache struct {
+	diffs      map[felt.Felt]*diffCache // state root -> state diff
```
rodrodros: And this value can be stored alongside each node

core/state/cache.go
```diff
@@ -0,0 +1,145 @@
...
+	links      map[felt.Felt]felt.Felt  // child -> parent
+	oldestRoot felt.Felt                // root of the oldest layer in the cache
+}
+
+func newStateCache() *stateCache {
+	return &stateCache{
+		diffs: make(map[felt.Felt]*diffCache),
+		links: make(map[felt.Felt]felt.Felt),
```
rodrodros: Also, for future references just as slices can be created with a fixed size, so can maps. Not something necessary to apply now, since we might change the implementation

```suggestion
		diffs: make(map[felt.Felt]*diffCache, constant),
		links: make(map[felt.Felt]felt.Felt, constant),
```

core/state/contract.go
```diff
@@ -0,0 +1,96 @@
...
+
+	"github.com/NethermindEth/juno/core/crypto"
+	"github.com/NethermindEth/juno/core/felt"
+)
+
+const (
+	contractEncSize          = 3*felt.Bytes + 8 // nonce + class hash + storage root + deploy height
+	contractEmptyRootEncSize = 2*felt.Bytes + 8 // nonce + class hash + deploy height
+)
+
```
rodrodros, line 15: 👍

core/state/contract.go
```diff
@@ -0,0 +1,96 @@
...
+type stateContract struct {
+	Nonce        felt.Felt // Contract's nonce
+	ClassHash    felt.Felt // Hash of the contract's class
+	DeployHeight uint64    // Block height at which the contract is deployed
+	StorageRoot  felt.Felt // Root hash of the contract's storage
+}
```
rodrodros: 👍 

(Should it be DeployedHeight (?)) small nitpick

core/state/contract.go
```diff
@@ -0,0 +1,96 @@
...
+
+func newContractDeployed(classHash felt.Felt, deployHeight uint64) *stateContract {
+	return &stateContract{
+		Nonce:        felt.Zero,
+		ClassHash:    classHash,
+		StorageRoot:  felt.Zero,
+		DeployHeight: deployHeight,
+	}
+}
```
rodrodros: I think we can return by value here

core/state/database.go
```diff
@@ -0,0 +1,41 @@
...
+	ClassTrieHeight           = 251
+	ContractTrieHeight        = 251
+	ContractStorageTrieHeight = 251
+)
+
+type StateDB struct {
+	disk       db.KeyValueStore
+	triedb     *triedb.Database
+	stateCache *stateCache
+}
```
rodrodros, line 22: Why are these fields pointers? Not saying they shouldn't, just trying to understand the bigger picture

core/state/database.go
```diff
@@ -0,0 +1,41 @@
...
+
+func NewStateDB(disk db.KeyValueStore, triedb *triedb.Database) *StateDB {
+	return &StateDB{disk: disk, triedb: triedb, stateCache: newStateCache()}
+}
```
rodrodros: `StateDB` is meant to be a long persistent object used across runtimes? If true, then I think we should return by reference 👍 like right now

core/state/history.go
```diff
@@ -0,0 +1,101 @@
+package state
+
+import (
+	"github.com/NethermindEth/juno/core"
+	"github.com/NethermindEth/juno/core/felt"
+	"github.com/NethermindEth/juno/core/trie2"
+	"github.com/NethermindEth/juno/db"
+)
+
+var _ StateReader = &StateHistory{}
```
rodrodros: Is this to make sure it implements an interface?

```
var _ StateReader = (*StateHistory)(nil)
```

You should use this so it doesn't use any memory at runtime and the check is still made at compile time

core/state/history.go
```diff
@@ -0,0 +1,101 @@
...
+
+type StateHistory struct {
+	blockNum uint64
+	state    *State
+}
```
rodrodros, line 16: Everything has been amazingly explained so far. Can we add a bit to this struct as well

core/state/history.go
```diff
@@ -0,0 +1,101 @@
...
+
+func NewStateHistory(blockNum uint64, stateRoot felt.Felt, db *StateDB) (*StateHistory, error) {
```
rodrodros: It looks to me that `*StateHistory` can be returned by value, also `stateRoot` can be passed by reference.

core/state/history.go
```diff
@@ -0,0 +1,101 @@
...
+		return nil, err
+	}
+
+	return &StateHistory{
+		blockNum: blockNum,
+		state:    state,
+	}, nil
+}
+
+func (s *StateHistory) ContractClassHash(addr felt.Felt) (felt.Felt, error) {
```
rodrodros: `addr` by ref, also this functions seems to be untested

core/state/history.go
```diff
@@ -0,0 +1,101 @@
...
+		return felt.Felt{}, err
+	}
+	ret, err := s.state.ContractClassHashAt(addr, s.blockNum)
+	if err != nil {
+		return felt.Felt{}, err
+	}
+	return ret, nil
+}
+
+func (s *StateHistory) ContractNonce(addr felt.Felt) (felt.Felt, error) {
```
rodrodros: ref, also this functions seems to be untested

core/state/history.go
```diff
@@ -0,0 +1,101 @@
...
+		return felt.Felt{}, err
+	}
+	ret, err := s.state.ContractNonceAt(addr, s.blockNum)
+	if err != nil {
+		return felt.Felt{}, err
+	}
+	return ret, nil
+}
+
+func (s *StateHistory) ContractStorage(addr, key felt.Felt) (felt.Felt, error) {
```
rodrodros: ref, also this functions seems to be untested

core/state/history.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+	ret, err := s.state.ContractStorageAt(addr, key, s.blockNum)
+	if err != nil {
+		return felt.Felt{}, err
+	}
+	return ret, nil
+}
+
+// Checks if the contract is deployed at the given block number.
+func (s *StateHistory) checkDeployed(addr felt.Felt) error {
```
rodrodros: ref, also this functions seems to be untested

core/state/history.go
```diff
@@ -0,0 +1,101 @@
...
+	isDeployed, err := s.state.ContractDeployedAt(addr, s.blockNum)
+	if err != nil {
+		return err
+	}
+
+	if !isDeployed {
+		// TODO(weiihann): previously this was db.ErrKeyNotFound
+		// remember to handle it in the rpc
```
rodrodros, line 72: Does this todo makes sense? What should we handle in the RPC

core/state/history.go
```diff
@@ -0,0 +1,101 @@
...
+		return ErrContractNotDeployed
+	}
+
+	return nil
+}
+
+func (s *StateHistory) Class(classHash felt.Felt) (*core.DeclaredClass, error) {
```
rodrodros: ref, also this functions seems to be untested

core/state/object.go
```diff
@@ -0,0 +1,118 @@
...
+	state *State // the state this object belongs to
+
+	addr     felt.Felt      // address of the contract
+	contract *stateContract // contract info
+
+	dirtyStorage Storage     // storage changes
+	storageTrie  *trie2.Trie // storage trie
+}
+
+func newStateObject(state *State, addr felt.Felt, contract *stateContract) *stateObject {
```
rodrodros: `addr` by ref and I believe `stateObject` can be returned by value. It's just a reference type

core/state/object.go
```diff
@@ -0,0 +1,118 @@
...
+	if err != nil {
+		return felt.Zero, err
+	}
+
+	// TODO(weiihann): there must be a better way to do this
+	path := tr.FeltToPath(key)
+	v, err := trieutils.GetNodeByPath(s.state.db.disk, db.ContractTrieStorage, &s.addr, &path, true)
+	if err != nil {
+		return felt.Zero, err
+	}
```
rodrodros: Thoughts?

core/state/object.go
```diff
@@ -0,0 +1,118 @@
...
+	keys := maps.Keys(s.dirtyStorage)
+	slices.SortFunc(keys, func(a, b felt.Felt) int {
+		return b.Cmp(&a)
+	})
+
+	for _, key := range keys {
+		val := s.dirtyStorage[key]
+		if err := tr.Update(&key, val); err != nil {
+			return nil, err
+		}
```
rodrodros, line 107: Why does `keys` need to be sorted? 

This function is not the most efficient. When does it get called? With what frequency?

core/state/state.go
```diff
@@ -0,0 +1,785 @@
...
+type StateReader interface {
+	ContractReader
+	ClassReader
+	TrieProvider
+}
+
+type ContractReader interface {
+	ContractClassHash(addr felt.Felt) (felt.Felt, error)
+	ContractNonce(addr felt.Felt) (felt.Felt, error)
+	ContractStorage(addr, key felt.Felt) (felt.Felt, error)
```
rodrodros: All of this interface method should receive by reference

core/state/state.go
```diff
@@ -0,0 +1,785 @@
...
+}
+
+type ClassReader interface {
+	Class(classHash felt.Felt) (*core.DeclaredClass, error)
```
rodrodros: reference

core/state/state.go
```diff
@@ -0,0 +1,785 @@
...
+}
+
+type TrieProvider interface {
+	ClassTrie() (*trie2.Trie, error)
+	ContractTrie() (*trie2.Trie, error)
+	ContractStorageTrie(addr felt.Felt) (*trie2.Trie, error)
+}
```
rodrodros, line 54: addr by reference

core/state/state.go
```diff
@@ -0,0 +1,785 @@
...
+type State struct {
+	initRoot     felt.Felt
+	db           *StateDB
+	contractTrie *trie2.Trie
+	classTrie    *trie2.Trie
+
+	stateObjects map[felt.Felt]*stateObject
+}
+
+func New(stateRoot felt.Felt, db *StateDB) (*State, error) {
```
rodrodros: reference

core/state/state.go
```diff
@@ -0,0 +1,785 @@
...
+	}
+
+	return &State{
+		initRoot:     stateRoot,
+		db:           db,
+		contractTrie: contractTrie,
+		classTrie:    classTrie,
+		stateObjects: make(map[felt.Felt]*stateObject),
+	}, nil
+}
```
rodrodros: This is a type that contains reference to other types, why are we returning it by reference and not by value?

core/class.go
MaksymMalicki: moved to a named constant

core/state/cache.go
MaksymMalicki: Yes, I was thinking about the linked list as well, I like the idea. This was supposed to mock a simpler version of the layertree. I will be implementing it.

core/state/contract.go
MaksymMalicki: yes, agree - changed

core/state/cache.go
MaksymMalicki: yes!

core/state/cache.go
MaksymMalicki: yes, removed

core/state/cache.go
MaksymMalicki: returning by value

core/state/cache.go
MaksymMalicki: added

core/state/contract.go
MaksymMalicki: returning by value

core/state/database.go
MaksymMalicki: each of those fields is meant to be a long-lived object, also possibly quite large in size. They are meant to be created only once at the start of the node

core/state/database.go
MaksymMalicki: Yes, exactly. It is created only on node start and is shared across runtimes

core/state/history.go
MaksymMalicki: done

core/state/history.go
MaksymMalicki: done

core/state/history.go
MaksymMalicki: added

core/state/history.go
MaksymMalicki: used ref and added tests

core/state/history.go
MaksymMalicki: same as above

core/state/history.go
MaksymMalicki: same as above

core/state/history.go
MaksymMalicki: same as above, added tests handling different scenarios of this func

core/state/history.go
MaksymMalicki: same as above

core/state/object.go
MaksymMalicki: done

core/state/state.go
MaksymMalicki: done

core/state/state.go
MaksymMalicki: done

core/state/state.go
MaksymMalicki: done

core/state/state.go
MaksymMalicki: done

core/state/cache.go
MaksymMalicki: added

core/state/history.go
MaksymMalicki: in the previous version of `func (s *stateSnapshot) checkDeployed(addr *felt.Felt) error` in the `state_snapshot.go` we were returning the `db.ErrKeyNotFound`. In the new version in `state/history.go`, we return `ErrContractNotDeployed`. However, wherever I look in the RPC package, we are only checking if `err != nil` - on first glance we don't need to change anything. I would however like to keep this comment here, and remove it once I'm 100% sure this is not the case during the integration.

core/state/object.go
MaksymMalicki: From what I understand, by sorting keys, we ensure that the same set of updates always produces the same trie structure, we want the trie structure to be deterministic, but why in descending order - I'm not yet sure.

This function gets called a lot, in every state commit, so in each Update/Revert, but it gets called concurently

core/state/cache.go
```diff
@@ -0,0 +1,139 @@
...
+)
+
+type diffCache struct {
+	storageDiffs      map[felt.Felt]map[felt.Felt]*felt.Felt // addr -> {key -> value, ...}
+	nonces            map[felt.Felt]*felt.Felt               // addr -> nonce
+	deployedContracts map[felt.Felt]*felt.Felt               // addr -> class hash
+}
+
+type cacheNode struct {
+	root   *felt.Felt
```
rodrodros: Why the root is a reference? Is it shared?

core/state/cache.go
```diff
@@ -0,0 +1,139 @@
...
+	parent *cacheNode
+	child  *cacheNode
+	diff   *diffCache
```
rodrodros: Is `diffCache` optional? Is it shared? Why is it a reference?

core/state/cache.go
```diff
@@ -0,0 +1,139 @@
...
+import (
+	"fmt"
+
+	"github.com/NethermindEth/juno/core/felt"
+)
+
+const (
+	// DefaultMaxLayers is the default maximum number of layers to keep in the cache
+	DefaultMaxLayers = 128
+)
```
rodrodros:
```suggestion
const DefaultMaxLayers = 128
```

core/state/cache.go
```diff
@@ -0,0 +1,139 @@
...
+
+	// Evict if over capacity
+	for len(c.rootMap) > DefaultMaxLayers {
+		evict := c.oldestNode
+		c.oldestNode = evict.child
+		if c.oldestNode != nil {
+			c.oldestNode.parent = nil
+		}
+		delete(c.rootMap, *evict.root)
+	}
```
rodrodros: If we push one by one, is it possible to get a number of layers greater than `DefaultMaxLayers + 1`? Trying to understand why is it a `for` and not just an `if`.

core/state/object.go
```diff
@@ -0,0 +1,118 @@
...
+func newStateObject(state *State, addr *felt.Felt, contract *stateContract) *stateObject {
+	return &stateObject{
+		state:        state,
+		addr:         *addr,
+		contract:     contract,
+		dirtyStorage: make(Storage),
+	}
+}
+
+func (s *stateObject) setClassHash(classHash felt.Felt) {
```
rodrodros: Let's receive by reference and then copy

core/state/object.go
```diff
@@ -0,0 +1,118 @@
...
+	s.contract.ClassHash = classHash
+}
+
+func (s *stateObject) setNonce(nonce felt.Felt) {
```
rodrodros: same idea

core/state/object.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+func newStateObject(state *State, addr *felt.Felt, contract *stateContract) *stateObject {
+	return &stateObject{
+		state:        state,
+		addr:         *addr,
+		contract:     contract,
+		dirtyStorage: make(Storage),
+	}
+}
```
rodrodros: Why not returning by value here?

core/state/cache.go
MaksymMalicki: right, nope

core/trie2/triedb/pathdb/database.go
```diff
@@ -19,8 +19,13 @@ const (
...
-	CleanCacheSize  int // Maximum size (in bytes) for caching clean nodes
-	WriteBufferSize int // Maximum size (in bytes) for buffering writes before flushing
+	CleanCacheSize  uint64 // Maximum size (in bytes) for caching clean nodes
+	WriteBufferSize int    // Maximum size (in bytes) for buffering writes before flushing
+}
+
+var DefaultConfig = &Config{
+	CleanCacheSize:  16 * utils.Megabyte,
+	WriteBufferSize: 64 * utils.Megabyte,
 }
```
rodrodros: Why are we rolling this back?

core/trie2/triedb/pathdb/database.go
```diff
@@ -36,10 +41,7 @@ type Database struct {
 // which is 16MB for clean cache and 64MB for dirty cache.
 func New(disk db.KeyValueStore, config *Config) (*Database, error) {
 	if config == nil {
-		config = &Config{
-			CleanCacheSize:  16 * utils.Megabyte,
-			WriteBufferSize: 64 * utils.Megabyte,
-		}
+		config = DefaultConfig
```
rodrodros: Didn't we stablish we were going to do it like this

core/state/accessors.go
```diff
@@ -0,0 +1,133 @@
...
+func GetStateObject(r db.KeyValueReader, state *State, addr *felt.Felt) (*stateObject, error) {
+	contract, err := GetContract(r, addr)
+	if err != nil {
+		return nil, err
+	}
+
+	return newStateObject(state, addr, &contract), nil
+}
+
+// Wrapper around getContract which checks if a contract is deployed
```
rodrodros: I think we can improve this comment, I understand what you meant but it is not very clear. 

One suggestion:
```suggestion
// Get a contract instance from the database. If it doesn't exist returns a contract not deployed error
```

core/state/state.go
```diff
@@ -0,0 +1,788 @@
...
+		dc, err := s.Class(hash)
+		if err != nil {
+			return err
+		}
+
+		if dc.At != blockNum {
+			continue
+		}
+
+		if dc.Class.Version() == 1 {
```
rodrodros: Please use a type assertion here. I want to remove this method from the interface later on, it is not used almost anywhere

core/state/accessors.go
MaksymMalicki: done

core/trie2/triedb/pathdb/database.go
MaksymMalicki: Okey I thought for a moment, that using the DefaultConfig during the database creation would be cleaner (when no config provided, default to pathDB with default config). I removed the `DefaultConfig` again and modified the database creation, so that if the nil `Config` is passed to the triedb `New()`, a pathDB with nil `PathConfig` will be created - which will default to the hardcoded config.

core/state/state.go
MaksymMalicki: done

core/trie2/triedb/pathdb/database.go
MaksymMalicki: removed, comment below

core/state/object.go
MaksymMalicki: Generally those are meant to be mutable. The `GetStateObject` accessor would still return a reference, but this could reduce the number of heap allocations in the `state.Update()`.

core/state/object.go
MaksymMalicki: Regarding this, the goal here is to hit the cache before reading the node from the DB. When writing this TODO, Han probably intended to read from cache right here, but I decided to move the clean cache reading to the beginning of the `State` getters (like `func (s *State) ContractStorage(addr, key *felt.Felt) (felt.Felt, error)`). Removed this TODO, as it's no longer relevant.

Issue #2853, opened 2025-05-23
Split Builder and Sequencer


rianhughes:
Currently the consensus pkgs `Validator` and `Proposer` use `Builder`, but the sequencing logic also lives inside of `Builder`. We should separate the sequencing logic into its own pkg. This isn't critical, but should be done at some point.

PR #2865, opened 2025-06-03
maintenance: seperate builder and sequencer

rianhughes:
closes https://github.com/NethermindEth/juno/issues/2853

### DIFF
--- a/sequencer/sequencer_test.go
+++ b/sequencer/sequencer_test.go
@@ -205,10 +205,11 @@ func TestRunOnce(t *testing.T) {
 	require.NotEmpty(t, block.Transactions)
 	require.NotEmpty(t, block.Receipts)
 }
+
 func TestHelpers(t *testing.T) {
 	seqAddr := utils.HexToFelt(t, "0xDEADBEEF")
 	blockTime := 100 * time.Millisecond
-	seq, _, _, _ := getGenesisSequencer(t, blockTime, seqAddr)
+	seq, _, _, _ := getGenesisSequencer(t, blockTime, seqAddr) //nolint:dogsled
 
 	require.NoError(t, seq.RunOnce())
 
--- a/vm/rust/src/execution.rs
+++ b/vm/rust/src/execution.rs
@@ -407,7 +407,11 @@ where
             // DEPLOY_ACCOUNT: Normal logic calculates max gas based on account balance, but account doesn't exist yet (balance = 0).
             // This would set gas limit to zero, causing immediate "out of gas". Use max gas limit from versioned constants instead.
             if is_deploy_account_transaction(tx) {
-                let max_sierra_gas_limit = block_context.versioned_constants().os_constants.validate_max_sierra_gas.0;
+                let max_sierra_gas_limit = block_context
+                    .versioned_constants()
+                    .os_constants
+                    .validate_max_sierra_gas
+                    .0;
                 return Ok(GasAmount::from(max_sierra_gas_limit));
             }
 


node/node.go
```diff
@@ -196,13 +197,14 @@ func New(cfg *Config, version string, logLevel *utils.LogLevel) (*Node, error) {
 			return nil, kErr
 		}
 		mempool := mempool.New(database, chain, mempoolLimit, log)
-		sequencer := builder.New(pKey, new(felt.Felt).SetUint64(sequencerAddress), chain, nodeVM,
-			time.Second*time.Duration(cfg.SeqBlockTime), mempool, log, cfg.SeqDisableFees, database)
-		sequencer.WithPlugin(junoPlugin)
-		chain.WithPendingBlockFn(sequencer.PendingBlock)
-		rpcHandler = rpc.New(chain, &sequencer, throttledVM, version, log, &cfg.Network)
+		builder := builder.New(chain, nodeVM, log, cfg.SeqDisableFees)
+		seq := sequencer.New(&builder, mempool, *new(felt.Felt).SetUint64(sequencerAddress),
```
rodrodros: Why is the sequencer address a uint64? Will it always be in this region?

sequencer/init_test.go
```diff
@@ -1,4 +1,4 @@
-package builder_test
+package sequencer
```
rodrodros: sequencer_test?

sequencer/sequencer.go
```diff
@@ -0,0 +1,254 @@
...
+	subNewHeads     *feed.Feed[*core.Block]
+	subPendingBlock *feed.Feed[*core.Block]
+	subReorgFeed    *feed.Feed[*sync.ReorgBlockRange]
+	plugin          plugin.JunoPlugin
+}
+
+func New(
+	builder *builder.Builder,
+	mempool *mempool.Pool,
+	sequencerAddress felt.Felt,
```
rodrodros: sequencer address by reference

sequencer/sequencer.go
```diff
@@ -0,0 +1,254 @@
...
+	if err := s.builder.ClearPending(); err != nil {
+		s.log.Errorw("clearing pending", "err", err)
+	}
+
+	if err := s.builder.InitPendingBlock(s.sequencerAddress); err != nil {
+		return err
+	}
+
+	ctx, cancel := context.WithCancel(context.Background())
+	cancel()
```
rodrodros: You want to call `cancel` or `defer cancel`? If it was the second I wonder why it was not caught by tests

sequencer/sequencer.go
```diff
@@ -0,0 +1,254 @@
...
+	blockHashToBeRevealed, err := s.builder.GetRevealedBlockHash()
+	if err != nil {
+		return err
+	}
+	for {
+		userTxns, err := s.mempool.PopBatch(NumTxnsToBatchExecute)
+		if err != nil {
+			return err
+		}
+		fmt.Println("  userTxns", userTxns)
```
rodrodros: don't forget to delete this println

sequencer/sequencer.go
```diff
@@ -0,0 +1,254 @@
...
+		s.log.Debugw("running txns", userTxns)
+		if err = s.builder.RunTxns(userTxns, blockHashToBeRevealed); err != nil {
+			fmt.Println("  userTxns err", err)
```
rodrodros: and this one

utils/sign_test.go
```diff
@@ -0,0 +1,19 @@
...
+	"github.com/consensys/gnark-crypto/ecc/stark-curve/ecdsa"
+	"github.com/stretchr/testify/require"
+)
+
+func TestSign(t *testing.T) {
+	pKrivKey, err := ecdsa.GenerateKey(rand.Reader)
+	require.NoError(t, err)
+	_, err = utils.Sign(pKrivKey, new(felt.Felt), new(felt.Felt))
+	require.NoError(t, err)
+	// We don't check the signature since the private key generation is not deterministic.
```
infrmtcs, line 19: Probably because we don't have verify function implemented, but we should test the round trip:
- Verify with the same public key should pass
- Verify with another public key should fail

utils/sign.go
```diff
@@ -0,0 +1,28 @@
...
+import (
+	"github.com/NethermindEth/juno/core/crypto"
+	"github.com/NethermindEth/juno/core/felt"
+	"github.com/consensys/gnark-crypto/ecc/stark-curve/ecdsa"
+)
+
+type BlockSignFunc func(privateKey *ecdsa.PrivateKey, blockHash, stateDiffCommitment *felt.Felt) ([]*felt.Felt, error)
+
+// Sign returns the builder's signature over data.
+func Sign(privKey *ecdsa.PrivateKey, blockHash, stateDiffCommitment *felt.Felt) ([]*felt.Felt, error) {
```
infrmtcs: I'm wondering if we can curry the function to

```suggestion
type BlockSignFunc func(blockHash, stateDiffCommitment *felt.Felt) ([]*felt.Felt, error)

// Sign returns the builder's signature over data.
func Sign(privKey *ecdsa.PrivateKey) BlockSignFunc {
```
The reason is that there is only 1 private key, and whether it's nil determines whether the sign function is nil. We can then pass the sign function around without having to pass the private key.

infrmtcs: Another way is
```suggestion
type BlockSigner interface {
	func Sign(blockHash, stateDiffCommitment *felt.Felt) ([]*felt.Felt, error)
}

type blockSigner ecdsa.PrivateKey

// Sign returns the builder's signature over data.
func (s *blockSigner) Sign(blockHash, stateDiffCommitment *felt.Felt) ([]*felt.Felt, error) {
```

node/node.go
rianhughes: In principle the sequencer address can be any felt, although for now there's no particular need to set it to anything specfic.

sequencer/init_test.go
rianhughes: I don't think it matters, at least it works either way

sequencer/sequencer.go
rianhughes: I want to cancel. It allows us to exit `listenPool` after executing a batch of txns

sequencer/sequencer.go
rianhughes: Whoops, thanks!

sequencer/sequencer.go
rianhughes: Whoops, thanks!

utils/sign.go
rianhughes: I've implemented this, but tbh I find it's overly abstract and find the previous approach to be clearer. Wdyt @rodrigo-pino ?

utils/sign_test.go
rianhughes: Yeah we should implement a Verify function, but this is outside the scope of this PR

infrmtcs: Agree, I just saw the old test. This PR refactors code, so probably we should address it in another PR.
This was my concern because I think even a simple `return nil, nil` will pass the test.

builder/builder.go
```diff
@@ -1,95 +1,65 @@
...
+		blockchain:  bc,
+		disableFees: disableFees,
+		vm:          vm,
 	}
 }
 
-func (b *Builder) WithPlugin(junoPlugin plugin.JunoPlugin) *Builder {
-	b.plugin = junoPlugin
-	return b
+func (b *Builder) Finalise(pending *sync.Pending, signer utils.BlockSignFunc, privateKey *ecdsa.PrivateKey) error {
```
infrmtcs, line 49: Why does this receive `pending` as argument instead of getting from `b.pendingBlock`? Otherwise I think this doesn't use anything from `(b *Builder)` so we can consider calling from `blockchain` directly by a caller. I'm thinking so because `blockchain` is a singleton anyway.

builder/builder.go
```diff
@@ -167,16 +98,16 @@ func (b *Builder) ClearPending() error {
 	return nil
 }
 
-func (b *Builder) InitPendingBlock() error {
-	header, err := b.bc.HeadsHeader()
+func (b *Builder) InitPendingBlock(ownAddress *felt.Felt) error {
```
rodrodros: Why is it called `ownAddress`? Perhaps `sequencerAddress`?

sequencer/sequencer.go
```diff
@@ -0,0 +1,260 @@
...
+		subReorgFeed:     feed.New[*sync.ReorgBlockRange](),
+	}
+}
+
+func (s *Sequencer) WithPlugin(junoPlugin plugin.JunoPlugin) *Sequencer {
+	s.plugin = junoPlugin
+	return s
+}
+
+func (s *Sequencer) Run(ctx context.Context) error {
```
rodrodros, line 70: It seems this entire function it's untested. Should it have some tests?

sequencer/sequencer.go
```diff
@@ -0,0 +1,260 @@
...
+			if err := s.builder.InitPendingBlock(s.sequencerAddress); err != nil {
+				return err
+			}
+			s.mu.Unlock()
+		}
+	}
+}
+
+// Execute a single block. Useful for tests.
+func (s *Sequencer) RunOnce() error {
```
rodrodros: Why don't we have this function in the testing package since it's goal is to be using during tests mainly?

infrmtcs: Probably because the test file is using `package sequencer_test`. I think we can either:
- Switch from black-box to white-box by using `package sequencer`.
- Create a separate `_test.go` file but with `package sequencer`.

utils/sign.go
rianhughes: Closing this point. Imo is is a bit more abstract, but it does solve passing the key and signer separately, which is a bit ugly and could cause errors

sequencer/sequencer.go
rianhughes: I'm not sure I understand. Both `TestBuildEmptyBlocks` and `TestPrefundedAccounts` test `Run`

sequencer/sequencer.go
rianhughes: I'm not sure I understand, we have a test for this function `TestRunOnce`. Also, RunOnce will only be used by tests in other pkgs, so I think it should stay in `sequencer_test.go`

builder/builder.go
rianhughes: Updated

builder/builder.go
rianhughes: The sequencer needs to load the `pending` type before calling `Finalise` anyway, to update the plugin and subscribers. I figured there's no point in loading it twice.

sequencer/sequencer.go
infrmtcs: Based on the docstring, Rodrigo and I think that this function is only for tests. But in this function, I see that it accesses some private fields and method like `s.sequencerAddress`, `s.listenPool` and `s.privKey`. That means moving this to `sequencer_test.go` is not possible because `sequencer_test.go` is using `package sequencer_test` which is not allowed to access private fields in `package sequencer`.

infrmtcs: So I suggest that we should still move this function to `_test.go` file if it's only for test, but to address the issue above, there are 2 approaches in my previous comments.

rianhughes: After discussing this point with @infrmtcs, I decided to move `RunOnce` to a separate file under the `sequencer` pkg

sequencer/sequencer_utils.go
```diff
@@ -0,0 +1,52 @@
+package sequencer
```
infrmtcs: This file name should end with `_test.go` to ensure that it's only accessible in tests

builder/builder.go
infrmtcs: I meant we can remove this `Finalise` function, because `sequencer` should already had enough inputs and can call `blockchain.Finalise` directly.

sequencer/sequencer_utils.go
rianhughes: Updated

builder/builder.go
rianhughes: Closing after internal discussion

PR #2866, opened 2025-06-03
feat: event query acceleration

EgeCaner:
Aggregated Bloom Filter implementation to accelerate getEvents endpoint for large block ranges.

- Aggregated bloom cache implementation
- Running event filter implementation to aggregate blooms before making it persistent
   - Add new blocks bloom 
   - Reorg handling
   - Lazy loads itself, reconstruct the filter if needed
 - Matched Block Iterator
   - Iterates over non-pending blocks
   - Loads aggregated filter and running filter on demand
   - Next() moves to next candidate block
 
New keys added to DB
- To persist aggregated filters
- To persist running filter (running filter should be persisted, on node shutdown) otherwise persisted as aggregated filter when reach capacity


core/aggregated_bloom_filter.go
```diff
@@ -0,0 +1,288 @@
...
+// Select rows 1 and 4 (Idx1 & Idx4):
+// Idx1:    1    0    1   ...   0
+// Idx4:    1    0    1   ...   1
+//
+//		       --------------------------
+//		AND:    1    0    1   ...   0
+//
+//	After AND: Resulting vector is 1 0 1 ... 0
+//
+//	This means Block 0 and Block 2 are possible matches for this key. Block numbers are
```
rodrodros: Being extra nitpicky but I think there is an unintended indentation in this comments

core/aggregated_bloom_filter.go
```diff
@@ -0,0 +1,288 @@
...
+var (
+	ErrAggregatedBloomFilterBlockOutOfRange error = errors.New("block number is not within range")
+	ErrBloomFilterSizeMismatch              error = errors.New("bloom filter len mismatch")
+	ErrMatchesBufferNil                     error = errors.New("matches buffer must not be nil")
+	ErrMatchesBufferSizeMismatch            error = errors.New("matches buffer size mismatch")
+)
+
+// NewAggregatedFilter creates a new AggregatedBloomFilter starting from the specified block number.
+// It initialises the bitmap array with empty bitsets of size AggregateBloomBlockRangeLen.
+func NewAggregatedFilter(fromBlock uint64) *AggregatedBloomFilter {
```
rodrodros: I think we can return by value here. At it's core it is a reference type (it holds a pointer and metadata) and only three words in size

core/aggregated_bloom_filter.go
```diff
@@ -0,0 +1,288 @@
...
+// Using this method, you can quickly identify candidate blocks for a key, improving
+// the performance of large-range event queries.
+type AggregatedBloomFilter struct {
+	bitmap    []bitset.BitSet
+	fromBlock uint64
+	toBlock   uint64
+}
+
+const (
+	AggregateBloomBlockRangeLen uint64 = 8192
```
rodrodros: If this constant is meant to be the same as `EventBloomLength` this constant should equal the other (from what I understand they should). This way we avoid a change in one affect the other.

core/aggregated_bloom_filter.go
```diff
@@ -0,0 +1,288 @@
...
+	ErrMatchesBufferNil                     error = errors.New("matches buffer must not be nil")
+	ErrMatchesBufferSizeMismatch            error = errors.New("matches buffer size mismatch")
+)
+
+// NewAggregatedFilter creates a new AggregatedBloomFilter starting from the specified block number.
+// It initialises the bitmap array with empty bitsets of size AggregateBloomBlockRangeLen.
+func NewAggregatedFilter(fromBlock uint64) *AggregatedBloomFilter {
+	bitmap := make([]bitset.BitSet, EventsBloomLength)
+	for i := range bitmap {
+		bitmap[i] = *bitset.New(uint(AggregateBloomBlockRangeLen))
```
rodrodros: Here `bitset.New` allocates to the heap. Then it's copied to `bitmap[i]` which is also another position in the heap. Which means we have 8096 independent heap allocations meant to be deleted after this `for` loop ends. 

I would suggest to create them by value directly (even if not as beautiful) but to avoid the heap allocation. Bassed on the bitset interface, I would suggest something as
```
bitmap[i] = bitset.Bitset{}
bitmap[i].Set(...args)
```

Also, I am happy to be proved that I am mistaken if the compiler is smart enough to realize, but I don't think it is.

A similar suggestion following this idea, we can have a custom constructor for bitset.Bitset which creates as a value type. It would also use the `AggregatedBloomBlockRangeLen` parameter by default since it is the code intention.
```
func makeBitset() bitset.Bitset {
   b := bitset.Bitset{}
   b.Set(AgregatedBloomBlockRangeLen)
   return b
}
```

core/aggregated_bloom_filter.go
```diff
@@ -0,0 +1,288 @@
...
+// If filter is nil, no-op.
+// Returns an error if the block number is out of range or if the bloom filter size doesn't match.
+func (f *AggregatedBloomFilter) Insert(filter *bloom.BloomFilter, blockNumber uint64) error {
+	if f.fromBlock > blockNumber || f.toBlock < blockNumber {
+		return ErrAggregatedBloomFilterBlockOutOfRange
+	}
+
+	if filter == nil {
+		return nil
+	}
```
rodrodros, line 119: Is there a usefulness for what happen if the filter is nil. Why would we call `insert` with a nil filter?

core/aggregated_bloom_filter.go
```diff
@@ -0,0 +1,288 @@
...
+	for index := range EventsBloomLength {
+		f.bitmap[index].Clear(uint(relativeBlockNumber))
+	}
+
+	return nil
+}
+
+// BlocksForKeys returns a bitset indicating which blocks within the range might contain
+// the given keys. If no keys are provided, returns a bitset with all bits set.
+func (f *AggregatedBloomFilter) BlocksForKeys(keys [][]byte) *bitset.BitSet {
```
rodrodros, line 155: I would return a value type here as well. `blockMatches` is a "reference" type by definition. Two words, one which is a pointer and the other the length. 

I would use `makeBitset` in this function for both `blockMatches ` and `innerMatches`

core/aggregated_bloom_filter.go
```diff
@@ -0,0 +1,288 @@
...
+			innerMatches.InPlaceIntersection(&row)
+		}
+		out.InPlaceUnion(innerMatches)
+	}
+
+	return nil
+}
+
+// Copy creates a deep copy of the AggregatedBloomFilter.
+func (f *AggregatedBloomFilter) Copy() *AggregatedBloomFilter {
```
rodrodros: Return by value

rodrigo-pino (commented):
Amazing work and thank you for all the commentary explaining. Very useful!

First round of comments, will continue reviewing

blockchain/aggregated_bloom_filter_cache.go
```diff
@@ -0,0 +1,255 @@
...
+	return &MatchedBlockIterator{
+		rangeStart:         fromBlock,
+		rangeEnd:           toBlock,
+		maxScanned:         maxScanned,
+		cache:              c,
+		runningFilter:      runningFilter,
+		matcher:            matcher,
+		currentWindowStart: windowStart,
+	}, nil
+}
```
rodrodros: I would say to return a value type here as well

blockchain/aggregated_bloom_filter_cache.go
```diff
@@ -0,0 +1,255 @@
...
+type MatchedBlockIterator struct {
+	currentBits        *bitset.BitSet // Current candidate blocks bitset to iterate
+	nextIndex          uint64         // Next bit index to test and possibly yield
+	rangeStart         uint64         // starting block number of the filter range
+	currentWindowStart uint64         // absolute block start of currently loaded window
+	rangeEnd           uint64         // total number of blocks in filter range
+	done               bool           // iteration finished flag
+
+	maxScanned   uint // max number of blocks to iterate (0 = unlimited)
+	scannedCount uint // number of blocks yielded so far
```
rodrodros: Why not uint64 🤔 ?

migration/migration.go
```diff
@@ -793,3 +794,29 @@ func migrateCairo1CompiledClass2(txn db.KeyValueWriter, key, value []byte, _ *ut
 
 	return txn.Put(key, value)
 }
+
+func reconstructAggregatedBloomFilters(txn db.IndexedBatch, network *utils.Network) error {
```
rianhughes, line 798: We should use `processBlock` to speed this up, see `calculateL1MsgHashes2`

rianhughes: Currently we loop through the entire chain at a time, but we can take advantage of worker pools. Also, if the block range used in filters is fixed (eg 100 blocks), then it might be worth handling each range of blocks in a single process

migration/migration.go
```diff
@@ -793,3 +794,29 @@ func migrateCairo1CompiledClass2(txn db.KeyValueWriter, key, value []byte, _ *ut
...
+			}
+			return err
+		}
+
+		if err := filter.Insert(header.EventsBloom, header.Number); err != nil {
+			return err
+		}
+
+		nextBlock := blockNumber + 1
+		if blockNumber > 0 && nextBlock%core.AggregateBloomBlockRangeLen == 0 {
```
rianhughes: I think `blockNumber > 0` can be removed since this is always true

rianhughes: nit: it would be nice to add a small comment here, eg // If BoomFilter is full, write to DB, and reset

migration/migration.go
```diff
@@ -793,3 +794,29 @@ func migrateCairo1CompiledClass2(txn db.KeyValueWriter, key, value []byte, _ *ut
...
+			if err = core.WriteAggregatedBloomFilter(txn, filter); err != nil {
+				return err
+			}
+			filter = core.NewAggregatedFilter(nextBlock)
+		}
+	}
+}
```
rianhughes: I think we can remove the nextBlock variabl, by moving the last block up front, eg 

```suggestion
func reconstructAggregatedBloomFilters(txn db.IndexedBatch, network *utils.Network) error {
	filter := core.NewAggregatedFilter(0)
	for blockNumber := uint64(0); ; blockNumber++ {
		// If bloomFilter reaches capacity, write to DB, and reset
		if blockNumber%core.AggregateBloomBlockRangeLen == 0 {
			if err := core.WriteAggregatedBloomFilter(txn, filter); err != nil {
				return err
			}
			filter = core.NewAggregatedFilter(blockNumber)
		}

		header, err := core.GetBlockHeaderByNumber(txn, blockNumber)
		if err != nil {
			if errors.Is(err, db.ErrKeyNotFound) {
				rf := core.NewRunningEventFilterHot(nil, filter, blockNumber)
				return core.WriteRunningEventFilter(txn, rf)
			}
			return err
		}

		if err := filter.Insert(header.EventsBloom, header.Number); err != nil {
			return err
		}

	}
}
```

migration/migration.go
```diff
@@ -793,3 +794,29 @@ func migrateCairo1CompiledClass2(txn db.KeyValueWriter, key, value []byte, _ *ut
...
 	return txn.Put(key, value)
 }
+
+func reconstructAggregatedBloomFilters(txn db.IndexedBatch, network *utils.Network) error {
+	filter := core.NewAggregatedFilter(0)
+	for blockNumber := uint64(0); ; blockNumber++ {
+		header, err := core.GetBlockHeaderByNumber(txn, blockNumber)
+		if err != nil {
+			if errors.Is(err, db.ErrKeyNotFound) {
+				rf := core.NewRunningEventFilterHot(nil, filter, blockNumber)
```
rianhughes: rf := core.NewRunningEventFilterHot(nil, filter, blockNumber)

    why not pass in the txn here?

migration/migration.go
```diff
@@ -71,6 +71,7 @@ var defaultMigrations = []Migration{
 	NewBucketMigrator(db.Class, migrateCairo1CompiledClass2).WithBatchSize(1_000),              //nolint:mnd
 	MigrationFunc(calculateL1MsgHashes2),
 	MigrationFunc(removePendingBlock),
+	MigrationFunc(reconstructAggregatedBloomFilters),
```
rianhughes, line 74: Any estimates about how long this migration will take? I think a lot of the work can be split into jobs and run in parallel, which should speed it up

MaksymMalicki:
From the DB perspective, looks good. This PR uses the `db.IndexedBatch`, like the rest of the project, but this technical debt will be obsolete in the near future. LGTM

core/aggregated_bloom_filter.go
EgeCaner: I guess naming led to confusion, let me explain the difference, 

`EventBloomLength` - is the length of the bloom filter
`AggregateBloomBlockRangeLen` - is the number of blocks/blooms to aggregate per filter

Our bitmatrix has `EventBloomLength` rows and `AggregateBloomBlockRangeLen` columns

Would such naming as`NumBlocksPerFilter` makes more sense?

core/running_event_filter.go
```diff
@@ -0,0 +1,357 @@
...
+
+func NewRunningEventFilterLazy(txn db.KeyValueStore) *RunningEventFilter {
+	return &RunningEventFilter{txn: txn, lazyInit: true}
+}
+
+// Insert adds a bloom filter for a single block, updating the internal
+// aggregated filter. If the current window is full, it will be persisted using
+// WriteAggregatedBloomFilter and a new window will be started.
+// This implementation assumes blocks are not missed.
+// Returns an error if the block cannot be inserted.
```
rianhughes, line 64: Nice

core/aggregated_bloom_filter.go
EgeCaner: Not sure that we should ever call it with nil filter. I added this check cuz `builder_test.TestBuildTwoEmptyBlocks` ends up passing nil value to this function

core/aggregated_bloom_filter.go
EgeCaner: I benchmarked it, and you were right. I updated the code as you suggested, and the allocations per operation have been halved.

```
BenchmarkNewAggregatedBloomFilter_makeBitset-14             1080           1089018 ns/op         8650864 B/op       8195 allocs/op
BenchmarkNewAggregatedBloomFilter-14                        1028           1118595 ns/op         8912976 B/op      16387 allocs/op
```

migration/migration.go
EgeCaner: txn here is `db.IndexedBatch` what `NewRunningEventFilterHot` expects is `db.KeyValueStore`

rianhughes:
Not a blocking comment: but it would be interesting to know what sort of performance benefits this provides (eg measure before and after this PR is merged).

blockchain/aggregated_bloom_filter_cache.go
```diff
@@ -0,0 +1,255 @@
...
+// filter is cached. Used as the lookup key for bloom filter caches.
+type EventFiltersCacheKey struct {
+	fromBlock uint64
+	toBlock   uint64
+}
+
+// AggregatedBloomFilterCache stores and manages LRU-cached aggregated bloom filters
+// for block ranges, supporting fallback loading and bulk insertion.
+// It is safe for concurrent use.
+type AggregatedBloomFilterCache struct {
```
rianhughes, line 29: It looks like AggregatedBloomFilterCache and RunningEventFilter implement a lot of the same functions. I think it would be worth defining an BloomFilter interface that each of these satisfy. It would also help us mock their behaviour in other tests that don't care about this logic.

core/running_event_filter.go
```diff
@@ -0,0 +1,357 @@
...
+		}
+		f.inner = lastStoredFilter
+	}
+
+	f.next = curBlock
+	return f.inner.clear(curBlock)
+}
+
+// Persist writes the current state of the RunningEventFilter to persistent storage.
+func (f *RunningEventFilter) Persist() error {
```
rianhughes: nit
```suggestion
// Writes the current state of the RunningEventFilter to persistent storage.
func (f *RunningEventFilter) Write() error {
```

core/aggregated_bloom_filter.go
rodrodros: Mm, I think the second name is clearer, yes

core/aggregated_bloom_filter.go
rodrodros: If it doesn't make sense to call with `nil` then we shouldn't have that check. Everyone calling the function should first check that they are not using a nil filter

blockchain/aggregated_bloom_filter_cache.go
```diff
@@ -0,0 +1,255 @@
...
+}
+
+// SetMany inserts multiple aggregated bloom filters into the cache.
+// Each filter is keyed by its block range.
+func (c *AggregatedBloomFilterCache) SetMany(filters []*core.AggregatedBloomFilter) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	for _, filter := range filters {
+		c.cache.Add(EventFiltersCacheKey{fromBlock: filter.FromBlock(), toBlock: filter.ToBlock()}, filter)
```
rodrodros: long line, please refactor

core/aggregated_bloom_filter.go
EgeCaner: Is it possible for node to receive empty block or block with nil bloom filter, such as in the `builder_test.TestBuildTwoEmptyBlocks` test? If not we can remove the check, however if value is nil just skipping the block is fine since no bits to set in bloom.

core/accessors.go
```diff
@@ -529,3 +529,46 @@ func GetClassAndContractRootByStateCommitment(r db.KeyValueReader, stateCommitme
 
 	return val, nil
 }
+
+func GetAggregatedBloomFilter(r db.KeyValueReader, fromBlock, toBLock uint64) (*AggregatedBloomFilter, error) {
```
rodrodros: Return filter by value

core/accessors.go
```diff
@@ -529,3 +529,46 @@ func GetClassAndContractRootByStateCommitment(r db.KeyValueReader, stateCommitme
...
+func WriteAggregatedBloomFilter(w db.KeyValueWriter, filter *AggregatedBloomFilter) error {
+	enc, err := encoder.Marshal(filter)
+	if err != nil {
+		return err
+	}
+	return w.Put(db.AggregatedBloomFilterKey(filter.FromBlock(), filter.ToBlock()), enc)
+}
+
+func GetRunningEventFilter(r db.KeyValueReader) (*RunningEventFilter, error) {
+	var filter RunningEventFilter
```
rodrodros, line 555: Return filter by value

rodrodros: My mistake, `RunningEventFilter` uses mutex so to avoid copy errors. It's better to return by reference here

migration/migration.go
```diff
@@ -793,3 +794,78 @@ func migrateCairo1CompiledClass2(txn db.KeyValueWriter, key, value []byte, _ *ut
...
+	var writeMu sync.Mutex
+
+	// Last block number to fit into full aggregated filter
+	lastBlockToStore := chainHeight - (chainHeight % core.NumBlocksPerFilter) - 1
+
+	/// Step by window size and construct and write full aggregated filters to db
+	for start := uint64(0); start < lastBlockToStore; start += windowLen {
+		// Capture loop vars
+		rangeStart := start
+		end := start + windowLen - 1
```
rianhughes: nit: we should be consistent with naming, eg rangeEnd

blockchain/aggregated_bloom_filter_cache.go
```diff
@@ -0,0 +1,261 @@
...
+		cache: *lru.NewCache[EventFiltersCacheKey, *core.AggregatedBloomFilter](size),
+	}
+}
+
+// WithFallback sets a fallback fetch function to be used if a requested
+// AggregatedBloomFilter is not found in the cache. The provided function must
+// return a filter matching the queried range, or an error.
+func (c *AggregatedBloomFilterCache) WithFallback(fallback func(EventFiltersCacheKey) (core.AggregatedBloomFilter, error)) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
```
rianhughes: I don't think we need the lock here, since we only set the fallback at the initialization stage

rianhughes (approved):

rodrigo-pino (approved):

blockchain/aggregated_bloom_filter_cache.go
```diff
@@ -0,0 +1,258 @@
...
+// AggregatedBloomFilter is not found in the cache. The provided function must
+// return a filter matching the queried range, or an error.
+func (c *AggregatedBloomFilterCache) WithFallback(fallback func(EventFiltersCacheKey) (core.AggregatedBloomFilter, error)) {
+	c.fallbackFunc = fallback
+}
+
+// Reset clears the entire bloom filter cache, removing all stored filters.
+func (c *AggregatedBloomFilterCache) Reset() {
+	c.mu.Lock()
+	defer c.mu.Unlock()
```
infrmtcs: I think we don't need this lock because `lru.Cache` itself is using lock

infrmtcs: Otherwise, if we need to guard a batch of operation, probably `lru.BasicLRU` is more suitable

EgeCaner: I removed the lock, there is no batch operation to guard

PR #2871, opened 2025-06-09
Fix/0.14.0 beta

EgeCaner:
Includes Minor Changes
- Updates feeder urls
- SubmittedTransactionsCache uses txHash from gateway response
- SubmittedTransactionsCache test uses mock gateway instead of gateway test client
- Updates `validator_test` expected commitments to correct values for starknet version v0.14.0

### DIFF
--- a/consensus/validator/validator_test.go
+++ b/consensus/validator/validator_test.go
@@ -132,7 +132,7 @@ func TestEmptyProposal(t *testing.T) {
 	// Step 3: ProposalFin
 	// Note: this commitment depends on the SupportedStarknetVersion, so block1Hash test should be updated whenever
 	// we update SupportedStarknetVersion
-	block1Hash, err := new(felt.Felt).SetString("0x27a852d180c77e0b7bc2b3e9b635b625db3ae48f2c469c93930c1889ae2d09f")
+	block1Hash, err := new(felt.Felt).SetString("0x9979aca51e836c8fd0f40a777980210120998a236a23d74615219f557a09f1")
 	require.NoError(t, err)
 	proposalFin := types.ProposalFin(*block1Hash)
 	require.NoError(t, validator.ProposalFin(proposalFin))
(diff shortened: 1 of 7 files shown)


consensus/validator/validator_test.go
```diff
@@ -132,7 +132,7 @@ func TestEmptyProposal(t *testing.T) {
 	// Step 3: ProposalFin
 	// Note: this commitment depends on the SupportedStarknetVersion, so block1Hash test should be updated whenever
 	// we update SupportedStarknetVersion
-	block1Hash, err := new(felt.Felt).SetString("0x27a852d180c77e0b7bc2b3e9b635b625db3ae48f2c469c93930c1889ae2d09f")
+	block1Hash, err := new(felt.Felt).SetString("0x9979aca51e836c8fd0f40a777980210120998a236a23d74615219f557a09f1")
```
rodrodros, line 135: Why are we using a different block hash?

rpc/v6/transaction_test.go
```diff
@@ -1686,28 +1686,41 @@ func TestSubmittedTransactionsCache(t *testing.T) {
...
-	txnHash, err := core.TransactionHash(txnToAdd, &network)
-	require.NoError(t, err)
-	txnToAdd.TransactionHash = txnHash
 	broadcastedTxn := &rpc.BroadcastedTransaction{Transaction: *rpc.AdaptTransaction(txnToAdd)}
 
-	// transaction no† found in db and feeder but found in cache
+	var gatewayResponse struct {
+		TransactionHash *felt.Felt `json:"transaction_hash"`
+		ContractAddress *felt.Felt `json:"address"`
+		ClassHash       *felt.Felt `json:"class_hash"`
```
rodrodros: Are these two other fields required? If they are nil they don't get marshalled right, so we might as well delete them from here? 

Did you add them for readability?

consensus/validator/validator_test.go
rianhughes, line 135: hmmm, these shouldn't need to change

rianhughes: Ah, the supported starknet version is different for the base branch

rianhughes (approved):

rpc/v6/transaction_test.go
EgeCaner: You are right, they are not required. I deleted them

consensus/validator/validator_test.go
EgeCaner: For protocol version greater or equal to 0.13.4 we use different block hash function ([ref](https://github.com/NethermindEth/juno/blob/179cf382e2f3cdc48b5031c7b17524eef2416956/core/block.go#L149)), thus yield to different hash

PR #2873, opened 2025-06-10
feature: implement proposer interface for consensus

rianhughes:
This PR implements the Proposer interface for the consensus layer. The proposer is responsible for generating block proposals, handling both empty and non-empty blocks depending on whether the mempool receives transactions within a given time window.

Notes:

    Proposal timing: The caller is responsible for deciding:
    * When to stop waiting for transactions and proceed with an empty block.
    * When to stop executing transactions (in the non-empty block flow) and continue with proposal finalisation.

    Gas strategy: This implementation reuses gas values from the previous block. Dynamic gas fee computation is out of scope of this PR, and probably requires SNs input.

    Code reuse: Some logic is shared with the Validator PR. Either PR can be merged first without conflict.


blockchain/blockchain.go
```diff
@@ -641,3 +641,58 @@ func (b *Blockchain) StoreGenesis(
 
 	return b.Finalise(block, stateUpdate, newClasses, nil)
 }
+
+type SimulateResult struct {
+	Block            *core.Block
+	StateUpdate      *core.StateUpdate
+	BlockCommitments *core.BlockCommitments
+	ConcatCount      *felt.Felt
```
rodrodros: I think ConcatCount should be a value type from SimulateResult since it is not shared with other type

consensus/proposer/proposer.go
```diff
@@ -0,0 +1,231 @@
...
+
+// Proposer defines the interface for constructing a block proposal in two possible flows:
+//
+// 1. **Empty Block Flow**
+//    - Used when no transactions are available.
+//    - Sequence: ProposalInit() → ProposalCommitment() → ProposalFin()
+//
+// 2. **Non-Empty Block Flow**
+//    - Used when transactions are present in the mempool.
+//    - Sequence: ProposalInit() → BlockInfo(ctx) → Txns(ctx) → ProposalCommitment() → ProposalFin()
```
rodrodros, line 27: Thank you for the explanation 👍

core/block.go
```diff
@@ -402,7 +402,7 @@ func UnmarshalBlockNumber(val []byte) uint64 {
 	return binary.BigEndian.Uint64(val)
 }
 
-func concatCounts(txCount, eventCount, stateDiffLen uint64, l1Mode L1DAMode) *felt.Felt {
+func ConcatCounts(txCount, eventCount, stateDiffLen uint64, l1Mode L1DAMode) *felt.Felt {
```
rodrodros: Oh this function returns a pointer, it should return by value. How big is the part of the code that gets affected by changing this to return by value. If it's just a couple of line I would hope we can make it return by value. Otherwise we open an issue with it

blockchain/blockchain.go
```diff
@@ -641,3 +641,58 @@ func (b *Blockchain) StoreGenesis(
 
 	return b.Finalise(block, stateUpdate, newClasses, nil)
 }
+
+type SimulateResult struct {
```
infrmtcs: I guess we need to rebase on the validator PR, because it looks like the changes in this file is very similar but different from the changes in validator PR.

rianhughes: I just merged the validator changes (now on main) into this PR

infrmtcs: Should we discard this change? I think the rebase result moves the code downward but doesn't change it.

core/block.go
rianhughes: It's not a big change, I updated the function to return it by value

builder/builder.go
```diff
@@ -270,6 +279,23 @@ func (b *Builder) storePending(newPending *sync.Pending) error {
...
 
+func (b *Builder) ExecuteTxns(txns []mempool.BroadcastedTransaction) (uint64, error) {
+	b.log.Debugw("calling ExecuteTxns")
+	blockHashToBeRevealed, err := b.GetRevealedBlockHash()
+	if err != nil {
+		return 0, err
+	}
+
+	l2gasConsumed := uint64(0)
+	l2gasConsumed, err = b.RunTxns(txns, blockHashToBeRevealed)
```
infrmtcs:
```suggestion
	l2gasConsumed, err := b.RunTxns(txns, blockHashToBeRevealed)
```

blockchain/blockchain.go
rianhughes: Updated

builder/builder.go
```diff
@@ -270,6 +279,23 @@ func (b *Builder) storePending(newPending *sync.Pending) error {
 	return nil
 }
 
+func (b *Builder) ExecuteTxns(txns []mempool.BroadcastedTransaction) (uint64, error) {
```
infrmtcs: I think move this back to the original position would reduce the diff and make it easier to review.

builder/builder.go
```diff
@@ -154,11 +159,11 @@ func (b *Builder) GetRevealedBlockHash() (*felt.Felt, error) {
 
 // RunTxns executes the provided transaction and applies the state changes
 // to the pending state
-func (b *Builder) RunTxns(txns []mempool.BroadcastedTransaction, blockHashToBeRevealed *felt.Felt) error {
+func (b *Builder) RunTxns(txns []mempool.BroadcastedTransaction, blockHashToBeRevealed *felt.Felt) (uint64, error) {
```
infrmtcs: nit: 
```suggestion
func (b *Builder) RunTxns(txns []mempool.BroadcastedTransaction, blockHashToBeRevealed *felt.Felt) (l2GasConsumed uint64, err error) {
```
Although it doesn't change any behaviour, but I think in case the return value is unclear (it's hard to guess what `uint64` here represents), we should use named return as a documentation method.

builder/builder.go
rianhughes: Updated

builder/builder.go
rianhughes: Moved

blockchain/blockchain.go
infrmtcs: In the case we change something here, moving it back to original position would reduce diff a lot and make it easier to review.

rianhughes: Yup, sorry about that. I moved `simulate` to reduce the diff, but I think `SimulateResult` should be moved to be beside `Simulate`, it's previous location didn't really make sense. SimulateResult is quite small, so shouldn't really cause problems with the review

builder/builder.go
```diff
@@ -309,6 +318,22 @@ func (b *Builder) ProposalInit(pInit *types.ProposalInit) error {
 	return err
 }
 
+func (b *Builder) ExecuteTxns(txns []mempool.BroadcastedTransaction) (uint64, error) {
```
infrmtcs: Can we move this back to the original position?

builder/builder.go
```diff
@@ -217,7 +222,11 @@ func (b *Builder) RunTxns(txns []mempool.BroadcastedTransaction, blockHashToBeRe
 	// Update pending block with transaction results
 	updatePendingBlock(pending, receipts, coreTxns, mergedStateDiff)
 
-	return b.storePending(pending)
+	l2GasConsumed := uint64(1) // Todo: should be 0?? Blockifer seems to return 0..
```
infrmtcs: Why do we start with 1 instead of 0?

builder/builder.go
rianhughes: Moved

builder/builder.go
rianhughes: Updated

builder/builder.go
rianhughes: It should be zero, I forget why I set it to 1. updated to 0

consensus/proposer/proposer.go
```diff
@@ -0,0 +1,237 @@
...
+
+	"github.com/Masterminds/semver/v3"
+	"github.com/NethermindEth/juno/builder"
+	"github.com/NethermindEth/juno/consensus/types"
+	"github.com/NethermindEth/juno/core"
+	"github.com/NethermindEth/juno/core/felt"
+	"github.com/NethermindEth/juno/mempool"
+)
+
+var NumTxnsToBatchExecute = 10
```
infrmtcs:
```suggestion
const NumTxnsToBatchExecute = 10 // TODO: Make this configurable
```

consensus/proposer/proposer.go
```diff
@@ -0,0 +1,237 @@
...
+
+	return out
+}
+
+func (p *proposer) ProposalCommitment() (types.ProposalCommitment, error) {
+	commitments, concatCommitment, err := p.builder.ExecutePending()
+	if err != nil {
+		return types.ProposalCommitment{}, err
+	}
+	pending, err := p.builder.Pending()
```
infrmtcs, line 197: Super nit:
```suggestion

	pending, err := p.builder.Pending()
```

consensus/proposer/proposer.go
```diff
@@ -0,0 +1,237 @@
...
+				out <- TxnBatchResult{Err: err}
+				return
+			}
+
+			l2GasConsumed, err := p.builder.ExecuteTxns(txns)
+			if err != nil {
+				out <- TxnBatchResult{Err: err}
+				return
+			}
+			p.l2GasConsumed += l2GasConsumed
```
infrmtcs: Why don't we put this inside builder instead? I think it's much clearer and ensures the sync between added transaction and the calculation of this value.

consensus/proposer/proposer.go
```diff
@@ -0,0 +1,237 @@
...
+					time.Sleep(100 * time.Millisecond) //nolint:mnd
+					continue
+				}
+				out <- TxnBatchResult{Err: err}
+				return
+			}
+
+			l2GasConsumed, err := p.builder.ExecuteTxns(txns)
+			if err != nil {
+				out <- TxnBatchResult{Err: err}
```
infrmtcs: Should select with ctx

consensus/proposer/proposer.go
```diff
@@ -0,0 +1,237 @@
...
+			default:
+			}
+
+			txns, err := p.mempool.PopBatch(NumTxnsToBatchExecute)
+			if err != nil {
+				if errors.Is(err, mempool.ErrTxnPoolEmpty) {
+					time.Sleep(100 * time.Millisecond) //nolint:mnd
+					continue
+				}
+				out <- TxnBatchResult{Err: err}
```
infrmtcs: Should select with ctx

consensus/proposer/proposer.go
```diff
@@ -0,0 +1,237 @@
...
+
+	// Blocks until the proposal commitment is ready or context is cancelled.
+	ProposalCommitment() (types.ProposalCommitment, error)
+
+	ProposalFin() (types.ProposalFin, error)
+}
+
+type TxnBatchResult struct {
+	Txns []types.Transaction
+	Err  error
```
infrmtcs: Why do we need to send error to p2p component? I think we can log error somewhere, but p2p part shouldn't "care" about why a transaction fails to execute.

consensus/proposer/proposer.go
```diff
@@ -0,0 +1,237 @@
...
+// If the context is cancelled or times out first, it returns a zero-value BlockInfo
+// and false. Without a context deadline, this method may block indefinitely.
+// The boolean return indicates whether the block is non-empty.
+func (p *proposer) BlockInfo(ctx context.Context) (types.BlockInfo, bool) {
+	// Check if there are any txns to create a non-empty block, with
+	// a timeout set by the context
+	ticker := time.NewTicker(100 * time.Millisecond) //nolint:mnd
+	defer ticker.Stop()
+	for {
+		if p.mempool.Len() != 0 {
```
infrmtcs, line 124: Probably out of scope of this PR, but we should communicate with mempool via a channel instead.

consensus/proposer/proposer.go
```diff
@@ -0,0 +1,237 @@
...
+	go func() {
+		defer close(out)
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			default:
+			}
+
+			txns, err := p.mempool.PopBatch(NumTxnsToBatchExecute)
```
infrmtcs, line 162: I think a rare race condition may happen. This is inside a spawned goroutine and we currently don't wait when it stops, so technically 2 concurrent calls to this may happen, when the previous run is stopping while the next run is already created

consensus/proposer/proposer.go
rianhughes: Yup, good point. I've moved it

consensus/proposer/proposer.go
rianhughes: True, we could just log. Although I'm wondering if any problems will arise if we indicate we will propose an non-empty block, but the proposer fails to return any (eg error in popping them from the mempool)..

What do you think, should we just log? I thought it might have been useful to indicate to the p2p layer that the proposer is experiencing errors

consensus/proposer/proposer.go
rianhughes: If we decide to log then this will be deleted.

consensus/proposer/proposer.go
rianhughes: If we decide to log then this will be deleted.

consensus/proposer/proposer.go
rianhughes: Damn, I pushed to the wrong PR :/

rianhughes: Okay updated

consensus/proposer/proposer.go
rianhughes: Yeah, if the caller proceeds with the proposal flow without cancelling the context then we can get a bad race condition. 

Maybe I can add an additional comment to highlight this in the functions doc-string (it is in the pkg documentation but maybe this isn't clear enough?). Another approach is to force the go-routine to shut down in `ProposalCommitment`, but that's kind of ugly imo. What do you think?

rianhughes: Actually, this can be prevented if the caller ranges on the `out` channel. This means the caller only proceeds when the channel closes, and the go-routine is shutdown 

```
//                      for res := range proposer.Txns(ctx) {
//                      ...
//			}
```

Omor12-2: 0x1A36d12Bb29562146104a4819d95AbBcBc7C1d12

My Eth receiving account..

consensus/proposer/proposer.go
infrmtcs: I think even if p2p component is notified, it can't do anything anyway, so it doesn't need to know. I think the simplest solution for now is just log here.

consensus/proposer/proposer.go
infrmtcs: Let's log and delete this.

consensus/proposer/proposer.go
infrmtcs: Let's log and delete this.

consensus/proposer/proposer.go
rianhughes: Makes sense. I've updated the code so that errors are logged, and not returned to the caller

consensus/proposer/proposer.go
rianhughes: Makes sense. I've updated the code so that errors are logged, and not returned to the caller

consensus/proposer/proposer.go
rianhughes: Makes sense. I've updated the code so that errors are logged, and not returned to the caller

infrmtcs (approved):

rodrigo-pino (approved):

PR #2874, opened 2025-06-10
feat: Add general buffered topic subscription

infrmtcs:


consensus/p2p/buffered/buffered_topic_subscription.go
```diff
@@ -0,0 +1,44 @@
...
+	pubsub "github.com/libp2p/go-libp2p-pubsub"
+)
+
+type TopicSubscription struct {
+	log        utils.Logger
+	bufferSize int
+	callback   func(context.Context, *pubsub.Message)
+}
+
+func New(log utils.Logger, bufferSize int, callback func(context.Context, *pubsub.Message)) TopicSubscription {
```
rodrodros: nitpick, shouldn't `New` be called `NewTopicSubscription`, thinking the use of this type outside the package where we will do

```
x := buffered.New(...)
```

to 

```
x := buffered.NewTopicSubscription(...)
```
If this is the only type meant to be inside the package, I think it's ok to leave the original constructor name `New`

consensus/p2p/buffered/buffered_topic_subscription.go
```diff
@@ -0,0 +1,44 @@
...
+	defer sub.Cancel()
+
+	for {
+		msg, err := sub.Next(ctx)
+		if errors.Is(err, context.Canceled) {
+			return
+		}
+		if err != nil {
+			b.log.Errorw("unable to receive message", "error", err)
+			continue
```
rodrodros: Another nitpick is to put the error `is` check inside the `if err != nil` block of code

rodrigo-pino (approved):

consensus/p2p/buffered/buffered_topic_subscription.go
infrmtcs: Although it's the only type, I agree that `buffered.New` isn't clear, so I renamed to `NewTopicSubscription`.

consensus/p2p/buffered/buffered_topic_subscription.go
infrmtcs: Done

rianhughes (approved):

PR #2875, opened 2025-06-10
feat: Add proposal stream with ability to reorder messages

infrmtcs:


rodrigo-pino (approved):

consensus/p2p/validator/proposal_stream.go
```diff
@@ -0,0 +1,149 @@
...
+		transition:         transition,
+		stateMachine:       &InitialState{},
+		nextSequenceNumber: 0,
+	}
+}
+
+func (s *proposalStream) start(ctx context.Context, firstMessage *consensus.StreamMessage) (types.Height, error) {
+	content := firstMessage.GetContent()
+	if content == nil {
+		return 0, fmt.Errorf("first message is not a proposal part")
```
rianhughes: nit
```suggestion
		return 0, fmt.Errorf("first message has empty content")
```

consensus/p2p/validator/proposal_stream.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+
+	if err := s.onProposalPart(ctx, content); err != nil {
+		return 0, err
+	}
+
```
rianhughes, line 53: nit
```suggestion
	// The state machine should have progressed from InitialState to AwaitingBlockInfoOrCommitmentState
```

consensus/p2p/validator/proposal_stream.go
```diff
@@ -0,0 +1,149 @@
...
+				return fmt.Errorf("stream does not end with proposal fin")
+			}
+		default:
+			return fmt.Errorf("unknown message type")
+		}
+	}
+	return nil
+}
+
+func (s *proposalStream) onProposalPart(ctx context.Context, messageContent []byte) error {
```
rianhughes: nit
```suggestion
func (s *proposalStream) processProposalPart(ctx context.Context, messageContent []byte) error {
```

consensus/p2p/validator/proposal_stream.go
```diff
@@ -0,0 +1,149 @@
...
+	case s.input <- streamMessage:
+	}
+}
+
+//nolint:unused // TODO: This is used in the next PR and this comment should be removed there.
+func (s *proposalStream) close() {
+	close(s.input)
+}
+
+func (s *proposalStream) processMessage(ctx context.Context, nextMessage *consensus.StreamMessage) error {
```
rianhughes: super nit:

```suggestion
func (s *proposalStream) processMessages(ctx context.Context, nextMessage *consensus.StreamMessage) error {
```

consensus/p2p/validator/proposal_stream.go
```diff
@@ -0,0 +1,149 @@
...
+	"fmt"
+
+	"github.com/NethermindEth/juno/consensus/starknet"
+	"github.com/NethermindEth/juno/consensus/types"
+	"github.com/NethermindEth/juno/utils"
+	"github.com/starknet-io/starknet-p2pspecs/p2p/proto/consensus/consensus"
+	"google.golang.org/protobuf/proto"
+)
+
+type proposalStream struct {
```
rianhughes, line 14:
```suggestion
// proposalStream receives and processes parts of Starknet proposals streamed from peers.
// Each part is delivered via the input channel, then ordered and validated.
// Once a complete and valid proposal is assembled, it is sent to the caller via the outputs channel.
type proposalStream struct {
```

rianhughes:
```suggestion
// proposalStream receives and processes parts of Starknet proposals streamed from peers.
// Each part is delivered via the input channel, then ordered and validated.
// Once a complete and valid proposal is assembled, it is sent to the caller via the outputs channel.
type proposalStream struct {
```

consensus/p2p/validator/proposal_stream_test.go
```diff
@@ -0,0 +1,368 @@
...
+	"github.com/NethermindEth/juno/utils"
+	"github.com/starknet-io/starknet-p2pspecs/p2p/proto/common"
+	"github.com/starknet-io/starknet-p2pspecs/p2p/proto/consensus/consensus"
+	"github.com/starknet-io/starknet-p2pspecs/p2p/proto/transaction"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"google.golang.org/protobuf/proto"
+)
+
+func TestProposalStream_Start(t *testing.T) {
```
rianhughes, line 18: Nice tests

consensus/p2p/validator/proposal_stream_test.go
```diff
@@ -0,0 +1,368 @@
...
+
+			if step.expectedStorage {
+				require.Contains(t, stream.messages, step.message.SequenceNumber, "message not stored")
+				assert.Equal(t, step.message, stream.messages[step.message.SequenceNumber], "message not stored correctly")
+				assert.Equal(t, oldStateMachine, stream.stateMachine, "state machine changed")
+				assert.Equal(t, oldNextSequenceNumber, stream.nextSequenceNumber, "next sequence number changed")
+				return
+			}
+
+			require.NotContains(t, stream.messages, step.message.SequenceNumber, "message stored")
```
rianhughes, line 295: I'm confused. I thought we store the step.message in the stream.messages field?

```
if s.nextSequenceNumber != nextMessage.SequenceNumber {
		s.messages[nextMessage.SequenceNumber] = nextMessage
		return nil
	}
```

rianhughes: I'm thinking about the case where we receive a message early. Eg ProposalInit, then ProposalTxn. At this point we should have stored ProposalTxn, sine we didn't recieve ProposalBlockInfo yet

rianhughes: Like this should only be true when we process messages in order (since we don't need to store them)

rianhughes: Ah, I see we return in the step.expectedStorage block

consensus/p2p/validator/proposal_stream_test.go
```diff
@@ -0,0 +1,368 @@
...
+			require.NoError(t, err)
+
+			if step.expectedStorage {
+				require.Contains(t, stream.messages, step.message.SequenceNumber, "message not stored")
+				assert.Equal(t, step.message, stream.messages[step.message.SequenceNumber], "message not stored correctly")
+				assert.Equal(t, oldStateMachine, stream.stateMachine, "state machine changed")
+				assert.Equal(t, oldNextSequenceNumber, stream.nextSequenceNumber, "next sequence number changed")
+				return
+			}
+
```
rianhughes, line 294: Just the additional comments 
```suggestion
			// The message arrived early, so we can't process it. It is stored for processing later.
			if step.expectedStorage {
				require.Contains(t, stream.messages, step.message.SequenceNumber, "message not stored")
				assert.Equal(t, step.message, stream.messages[step.message.SequenceNumber], "message not stored correctly")
				assert.Equal(t, oldStateMachine, stream.stateMachine, "state machine changed")
				assert.Equal(t, oldNextSequenceNumber, stream.nextSequenceNumber, "next sequence number changed")
				return
			}

			// We were able to process the message and progress the state machine etc.
```

rianhughes:
LGTM, if we could add the doc string comment to `proposalStream` and the two comments in the `testProposalStreamProcessMessage` function that would be great. I'll leave the other nits up to you

consensus/p2p/validator/proposal_stream_test.go
infrmtcs: I think a clearer approach would be:
```
type expectedResult interface {
...
}

type storeAndReturn struct {
...
}

type process struct {
...
}
```

What do you think?

infrmtcs: What do you think about this:
https://github.com/NethermindEth/juno/pull/2878/files

consensus/p2p/validator/proposal_stream.go
infrmtcs: Done

consensus/p2p/validator/proposal_stream.go
infrmtcs: Done

consensus/p2p/validator/proposal_stream.go
infrmtcs: Done

consensus/p2p/validator/proposal_stream.go
infrmtcs: Done

consensus/p2p/validator/proposal_stream.go
infrmtcs: Done

consensus/p2p/validator/proposal_stream_test.go
rianhughes: Just commented on the PR. I like that things are grouped, that's much nicer. I think using functions would have been a little clearer than the interface, but don't mind too much. If we could add the comment to the interface that would be great, happy to merge otherwise

PR #2876, opened 2025-06-10
feat: Add proposal stream demux with ability to process multiple streams

infrmtcs:


rodrigo-pino (approved):

consensus/p2p/validator/proposal_stream_demux.go
```diff
@@ -0,0 +1,224 @@
...
+	log                 utils.Logger
+	transition          Transition
+	bufferSizeConfig    *config.BufferSizes
+	commitNotifier      <-chan types.Height
+	currentHeight       types.Height
+	streams             map[streamID]*proposalStream
+	streamHeights       map[types.Height][]streamID
+	outputs             chan starknet.Proposal
+	cancelCurrentHeight context.CancelFunc
+	currentHeightPool   *pool.ContextPool
```
rianhughes, line 43: nit

```suggestion
	currentHeightCancel context.CancelFunc
	currentHeightCtxPool   *pool.ContextPool
```

consensus/p2p/validator/proposal_stream_demux.go
```diff
@@ -0,0 +1,224 @@
...
+}
```
rianhughes: nit
```suggestion
	log                  utils.Logger
	transition           Transition
	bufferSizeConfig     *config.BufferSizes
	commitNotifier       <-chan types.Height
	streams              map[streamID]*proposalStream
	streamHeights        map[types.Height][]streamID
	proposalOutputs      chan starknet.Proposal
	currentHeight        types.Height
	currentHeightCancel  context.CancelFunc
	currentHeightCtxPool *pool.ContextPool
}
```

consensus/p2p/validator/proposal_stream_demux.go
```diff
@@ -0,0 +1,224 @@
...
+	Listen() <-chan types.Proposal[V, H, A]
+}
+
+// proposalStreamDemux is a demux for the proposal streams.
+// It is responsible for:
+// - Start the streams when a new height starts.
+// - Demultiplex the incoming messages to the streams.
+// - Cancel the current height streams when a new commit is received.
+// Although the methods must be called sequentially and the streams are created, loaded, started, and stopped sequentially,
+// the streams run concurrently.
```
rianhughes: Can we add a line or two just describing the motivation for this, or an example of what it allows us to do? I think it would help anyone who comes along later and is trying to understand why we do this

consensus/p2p/validator/proposal_stream_demux.go
```diff
@@ -0,0 +1,224 @@
...
+		t.streams[streamID].close()
+		delete(t.streams, streamID)
+	}
+
+	// Finally delete the list of streams for the old height
+	delete(t.streamHeights, t.currentHeight)
+}
+
+// createPool creates a new pool for the current height
+func (t *proposalStreamDemux) createPool(ctx context.Context) {
```
rianhughes: nit
```suggestion
func (t *proposalStreamDemux) createCtxPool(ctx context.Context) {
```

consensus/p2p/validator/proposal_stream_demux.go
```diff
@@ -0,0 +1,239 @@
...
+		stream := t.streams[streamID]
+		t.currentHeightCtxPool.Go(func(ctx context.Context) error {
+			stream.loop(ctx)
+			return nil
+		})
+	}
+
+	return nil
+}
+
```
rianhughes, line 177: nit: feel free to ignore
```suggestion
	// If the msg is for this height, then start a go-routine
	// that will process this particular proposal in a loop
	if message.SequenceNumber == 0 {
		return t.onFirstMessage(ctx, streamID, &message)
	}

	// Forward the proposal onto the go-routine started above
	return t.onSubsequentMessage(ctx, streamID, &message)
}
```

consensus/p2p/validator/proposal_stream_demux.go
```diff
@@ -0,0 +1,239 @@
...
+func NewProposalStreamDemux(
+	log utils.Logger,
+	transition Transition,
+	bufferSizeConfig *config.BufferSizes,
+	commitNotifier <-chan types.Height,
+	currentHeight types.Height,
+) ProposalStreamDemux[starknet.Value, starknet.Hash, starknet.Address] {
+	return &proposalStreamDemux{
```
rianhughes, line 68: We should be careful about overwhemling the node. Eg processing 104 proposals may cause a degredation in the nodes performance

rianhughes (approved):

consensus/p2p/validator/proposal_stream_demux.go
infrmtcs: Agree, and I think spam prevention is a big topic that we definitely need to address after having the MVP.

PR #2877, opened 2025-06-10
feature: implemet p2p2consensus adapters for proposal types

rianhughes:
closes https://github.com/NethermindEth/juno/issues/2890

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -302,32 +302,17 @@ func (b *Builder) ProposalInit(pInit *types.ProposalInit) error {
 			ProtocolVersion: blockchain.SupportedStarknetVersion.String(),
 			// Todo: once the spec is finalised, handle these fields (if they still exist)
 			// OldStateRoot, VersionConstantCommitment, NextL2GasPriceFRI
-			L1GasPriceETH: header.L1GasPriceETH,
+			// Note: we use the header values by default, since the proposer only
+			// sends over a subset of the gas prices (eg for L1DataGasPrice it
+			// only sends the L1DataGasPriceWEI, but not the price in FRI, but
+			// we need both for the block hash)
+			L1GasPriceETH:  header.L1GasPriceETH,
+			L1GasPriceSTRK: header.L1GasPriceSTRK,
+			L1DataGasPrice: header.L1DataGasPrice,
+			L2GasPrice:     header.L2GasPrice,
 		},
 		Transactions: []core.Transaction{},
 		Receipts:     []*core.TransactionReceipt{},
-	}
-
-	// Old blocks do not have these values, but we need them
-	// when we transition to a new block hash function
-	if pendingBlock.L1GasPriceSTRK == nil {
-		pendingBlock.L1GasPriceSTRK = new(felt.Felt).SetUint64(1)
-	}
-	if pendingBlock.L1GasPriceETH == nil {
-		pendingBlock.L1GasPriceETH = new(felt.Felt).SetUint64(1)
-	}
-
-	if pendingBlock.L1DataGasPrice == nil {
-		pendingBlock.L1DataGasPrice = &core.GasPrice{
-			PriceInWei: new(felt.Felt).SetUint64(1),
-			PriceInFri: new(felt.Felt).SetUint64(1),
-		}
-	}
-	if pendingBlock.L2GasPrice == nil {
-		pendingBlock.L2GasPrice = &core.GasPrice{
-			PriceInWei: new(felt.Felt).SetUint64(1),
-			PriceInFri: new(felt.Felt).SetUint64(1),
-		}
 	}
 
 	newClasses := make(map[felt.Felt]core.Class)
@@ -350,9 +335,9 @@ func (b *Builder) SetBlockInfo(blockInfo *types.BlockInfo) {
 	pending.Block.Header.Number = blockInfo.BlockNumber
 	pending.Block.Header.SequencerAddress = &blockInfo.Builder
 	pending.Block.Header.Timestamp = blockInfo.Timestamp
-	pending.Block.Header.L2GasPrice = &core.GasPrice{PriceInFri: &blockInfo.L2GasPriceFRI}
+	pending.Block.Header.L2GasPrice.PriceInFri = &blockInfo.L2GasPriceFRI
 	pending.Block.Header.L1GasPriceETH = &blockInfo.L1GasPriceWEI
-	pending.Block.Header.L1DataGasPrice = &core.GasPrice{PriceInWei: &blockInfo.L1DataGasPriceWEI}
+	pending.Block.Header.L1DataGasPrice.PriceInWei = &blockInfo.L1DataGasPriceWEI
 	pending.Block.Header.L1DAMode = blockInfo.L1DAMode
 	b.pendingBlock.Store(pending)
 }
--- a/consensus/p2p/validator/proposal_stream.go
+++ b/consensus/p2p/validator/proposal_stream.go
@@ -61,7 +61,6 @@ func (s *proposalStream) start(ctx context.Context, firstMessage *consensus.Stre
 	}
 }
 
-//nolint:unused // TODO: This is used in the next PR and this comment should be removed there.
 func (s *proposalStream) loop(ctx context.Context) {
 	for {
 		select {
@@ -76,7 +75,6 @@ func (s *proposalStream) loop(ctx context.Context) {
 	}
 }
 
-//nolint:unused // TODO: This is used in the next PR and this comment should be removed there.
 func (s *proposalStream) enqueueMessage(ctx context.Context, streamMessage *consensus.StreamMessage) {
 	select {
 	case <-ctx.Done():
@@ -85,7 +83,6 @@ func (s *proposalStream) enqueueMessage(ctx context.Context, streamMessage *cons
 	}
 }
 
-//nolint:unused // TODO: This is used in the next PR and this comment should be removed there.
 func (s *proposalStream) close() {
 	close(s.input)
 }
--- a/consensus/p2p/validator/proposal_stream_demux.go
+++ b/consensus/p2p/validator/proposal_stream_demux.go
@@ -1,0 +1,242 @@
+package validator
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/NethermindEth/juno/consensus/p2p/buffered"
+	"github.com/NethermindEth/juno/consensus/p2p/config"
+	"github.com/NethermindEth/juno/consensus/starknet"
+	"github.com/NethermindEth/juno/consensus/types"
+	"github.com/NethermindEth/juno/utils"
+	pubsub "github.com/libp2p/go-libp2p-pubsub"
+	"github.com/sourcegraph/conc"
+	"github.com/sourcegraph/conc/pool"
+	"github.com/starknet-io/starknet-p2pspecs/p2p/proto/consensus/consensus"
+	"google.golang.org/protobuf/proto"
+)
+
+type streamID string
+
+type ProposalStreamDemux[V types.Hashable[H], H types.Hash, A types.Addr] interface {
+	Loop(context.Context, *pubsub.Topic)
+	Listen() <-chan types.Proposal[V, H, A]
+}
+
+// proposalStreamDemux is a demux for the proposal streams.
+//
+// Motivation:
+// In a consensus protocol, validators need to efficiently handle multiple concurrent proposal streams
+// from different proposers across various consensus rounds and heights. Without proper demultiplexing,
+// the system would struggle with:
+// 1. Concurrent processing of proposals from multiple sources
+// 2. Resource management and cleanup when heights advance
+// 3. Proper isolation between different proposal streams
+//
+// This component solves these challenges by:
+// - Initialising dedicated streams per proposal, running them concurrently and independently
+// - Automatically managing stream lifecycle based on consensus height progression
+// - Enabling concurrent proposal processing while maintaining safety guarantees
+// - Centralising proposal routing logic to simplify downstream consumers
+//
+// It is responsible for:
+// - Start the streams when a new height starts.
+// - Demultiplex the incoming messages to the streams.
+// - Cancel the current height streams when a new commit is received.
+// Although the methods must be called sequentially and the streams are created, loaded, started, and stopped sequentially,
+// the streams run concurrently.
+type proposalStreamDemux struct {
+	log                  utils.Logger
+	transition           Transition
+	bufferSizeConfig     *config.BufferSizes
+	commitNotifier       <-chan types.Height
+	streams              map[streamID]*proposalStream
+	streamHeights        map[types.Height][]streamID
+	outputs              chan starknet.Proposal
+	currentHeight        types.Height
+	currentHeightCancel  context.CancelFunc
+	currentHeightCtxPool *pool.ContextPool
+}
+
+func NewProposalStreamDemux(
+	log utils.Logger,
+	transition Transition,
+	bufferSizeConfig *config.BufferSizes,
+	commitNotifier <-chan types.Height,
+	currentHeight types.Height,
+) ProposalStreamDemux[starknet.Value, starknet.Hash, starknet.Address] {
+	return &proposalStreamDemux{
+		log:              log,
+		transition:       transition,
+		bufferSizeConfig: bufferSizeConfig,
+		commitNotifier:   commitNotifier,
+		streams:          make(map[streamID]*proposalStream),
+		streamHeights:    make(map[types.Height][]streamID),
+		outputs:          make(chan starknet.Proposal, bufferSizeConfig.ProposalOutputs),
+		currentHeight:    currentHeight,
+	}
+}
+
+func (t *proposalStreamDemux) Loop(ctx context.Context, topic *pubsub.Topic) {
+	defer close(t.outputs)
+
+	messages := make(chan *pubsub.Message, t.bufferSizeConfig.ProposalDemux)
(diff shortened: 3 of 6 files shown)


adapters/p2p2consensus/class.go
```diff
@@ -0,0 +1,89 @@
...
+	"github.com/NethermindEth/juno/adapters/sn2core"
+	"github.com/NethermindEth/juno/core"
+	"github.com/NethermindEth/juno/core/crypto"
+	"github.com/NethermindEth/juno/starknet"
+	"github.com/NethermindEth/juno/starknet/compiler"
+	"github.com/NethermindEth/juno/utils"
+	"github.com/starknet-io/starknet-p2pspecs/p2p/proto/class"
+)
+
+func AdaptClass(cls *class.Cairo1Class) core.Class {
```
rodrodros: Why returning a pointer instead of an actual value, I don't understand the use of the interface for a return type if the only returnable type is a Cairo1Class

rianhughes: Returning a `*core.Cairo1Class` also works. However, it overly couples the adapter to this type imo. Also if SN adds additional classes then we'll need to implement more changes (they would probably be minimal - but still). If you feel strongly about this then let me know and I'll update it

adapters/p2p2consensus/proposal.go
```diff
@@ -0,0 +1,83 @@
+package p2p2consensus
+
+import (
+	"math/big"
+
+	"github.com/Masterminds/semver/v3"
+	consensus "github.com/NethermindEth/juno/consensus/types"
+	"github.com/NethermindEth/juno/core"
+	"github.com/NethermindEth/juno/core/felt"
+	p2pconsensus "github.com/NethermindEth/juno/p2p/proto/consensus/consensus"
```
infrmtcs:
```suggestion
	p2pconsensus "github.com/starknet-io/starknet-p2pspecs/p2p/proto/consensus/consensus"
```

I know this doesn't look good, but it's how the generated code is importing its dependencies, so we need to follow it for now. Probably I'll try again to find a way allow using `NethermindEth/juno` instead.

adapters/p2p2consensus/felt.go
```diff
@@ -0,0 +1,48 @@
+package p2p2consensus
```
infrmtcs: Why do we need to copy the existing file? I think we should have something like `p2p2core` for these functions to be reusable between sync and consensus.

infrmtcs: I just noticed that we already have `p2p2core`, but it also contains the contents of (supposedly) `p2p2sync`. `p2p2consensus` should depends on that for common functions instead.

adapters/p2p2consensus/class.go
```diff
@@ -0,0 +1,89 @@
+package p2p2consensus
```
infrmtcs: Same comment as the one in `felt.go`. We should reuse this instead of copying the code.

consensus/types/proposal.go
```diff
@@ -0,0 +1,61 @@
+package types
```
infrmtcs: nit: I think this file should be in `consensus/p2p/types` instead of `consensus/types`, because it's only for the conversion back and forth between p2p and the rest of consensus.

adapters/p2p2consensus/transaction.go
```diff
@@ -0,0 +1,178 @@
...
+			ResourceBounds: map[core.Resource]core.ResourceBounds{
+				core.ResourceL1Gas:     adaptResourceLimits(tx.Common.ResourceBounds.L1Gas),
+				core.ResourceL2Gas:     adaptResourceLimits(tx.Common.ResourceBounds.L2Gas),
+				core.ResourceL1DataGas: adaptResourceLimits(tx.Common.ResourceBounds.L1DataGas),
+			},
+			PaymasterData:         utils.Map(tx.Common.PaymasterData, AdaptFelt),
+			AccountDeploymentData: utils.Map(tx.Common.AccountDeploymentData, AdaptFelt),
+			NonceDAMode:           nDAMode,
+			FeeDAMode:             fDAMode,
+		}
```
infrmtcs: These lines are actually shared between sync protocol and consensus protocol. We should avoid duplicating the code, but instead create common functions to convert between the same pairs of types.

adapters/p2p2consensus/proposal.go
```diff
@@ -0,0 +1,83 @@
...
+		EthToStrkRate:     *U128ToFelt(msg.EthToStrkRate),
+		L1DAMode:          core.L1DAMode(msg.L1DaMode),
+	}
+}
+
+func AdaptProposalCommitment(msg *p2pconsensus.ProposalCommitment) consensus.ProposalCommitment {
+	return consensus.ProposalCommitment{
+		BlockNumber: msg.BlockNumber,
+		Builder:     *new(felt.Felt).SetBytes(msg.Builder.Elements),
+
```
infrmtcs, line 37: nit: feel free to ignore, but I think removing blank lines between fields aligns the values, which help readability a bit.

adapters/p2p2consensus/felt.go
infrmtcs: Simplest solution would be `p2p2consensus` depends on `p2p2core` and we create an issue to track refactoring sync adapters to `p2p2sync`. It can be addressed when we integrate the sync protocol.

adapters/p2p2consensus/proposal.go
rianhughes: I'll update it, woudl you mind creating an issue to track this?

adapters/p2p2consensus/class.go
rianhughes: The declare txns in consensus (`Cairo1Class`) sends a different definition of class than the one used in p2p2core (`*class.Class, class.Class_Cairo1`). p2p2core also handles both classes whereas we only need to handle one here. Maybe we could wrap the type used here in order to use that function, but it just seems a bit forced /messy

adapters/p2p2consensus/transaction.go
rianhughes: There is a deviation between the transaction types used in consensus and those used in p2p2core. Here we only consider V3 txns, whereas p2p2core allows txn types. I think reusing the same logic for both could cause issues.

consensus/types/proposal.go
rianhughes: Currently `validator` and `proposal` use these types, which exist outside `consensus/p2p.` If we end up moving `validator` and `proposal` into `consensus/p2p` then we can consider moving them then, but it doesn't seem appropriate just yet, since they may remain in their current place

adapters/p2p2consensus/felt.go
rianhughes: I deleted this file

rianhughes: >Simplest solution would be p2p2consensus depends on p2p2core and we create an issue to track refactoring sync adapters to p2p2sync. It can be addressed when we integrate the sync protocol.

Just a quick note, currently the p2psync and p2pconsensus use different types, so it's not obvious we should merge them. But I think we can explore this in a separate issue

adapters/p2p2consensus/class.go
rodrodros: I think we should keep it without the interface, if we get a new very different class I would say we add a different method completely as well. And have each type be adapted by their specific method, and I think makes the code easier to read as well (hypothetically of course).

adapters/p2p2consensus/class.go
```diff
@@ -0,0 +1,90 @@
...
+	program := utils.Map(cls.Program, p2p2core.AdaptFelt)
+	compiled, err := createCompiledClass(cls)
+	if err != nil {
+		panic(err)
+	}
+
+	adaptEP := func(points []*class.SierraEntryPoint) []core.SierraEntryPoint {
+		// usage of NonNilSlice is essential because relevant core class fields are non nil
+		return utils.Map(utils.NonNilSlice(points), adaptSierra)
+	}
```
rodrodros: I would prefer we don't use this method. Functional programming is not the strong points of Go and this code is more inefficient than your classic go `for` loops.

```golang
adaptEp := make([]core.SierraEntryPoint, len(points))
count := 0
for i := range points {
    if points[i] != nil {
       adaptEp[count] = adaptSierra(points[i])
       count++
    }
}
```

adapters/p2p2consensus/class.go
```diff
@@ -0,0 +1,90 @@
...
+		}{
+			Constructor: adaptEP(entryPoints.Constructors),
+			External:    adaptEP(entryPoints.Externals),
+			L1Handler:   adaptEP(entryPoints.L1Handlers),
+		},
+		Program:         program,
+		ProgramHash:     crypto.PoseidonArray(program...),
+		SemanticVersion: cls.ContractClassVersion,
+		Compiled:        compiled,
+	}
```
rodrodros: After we don't use the interface as a return type, let's return by value here

adapters/p2p2consensus/class.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+func adaptSierra(e *class.SierraEntryPoint) core.SierraEntryPoint {
+	return core.SierraEntryPoint{
+		Index:    e.Index,
+		Selector: p2p2core.AdaptFelt(e.Selector),
+	}
```
rodrodros: Also, I think this function is so simple and was added for the sake of functional programmign that I advocate for removing it completely (unless it's used else where)

adapters/p2p2consensus/class.go
```diff
@@ -0,0 +1,90 @@
...
+	def := &starknet.SierraDefinition{
+		Abi: cairo1.Abi,
+		EntryPoints: starknet.SierraEntryPoints{
+			// WARNING: usage of utils.NonNilSlice is essential, otherwise compilation will finish with errors
+			// todo move NonNilSlice to Compile ?
+			Constructor: utils.Map(utils.NonNilSlice(ep.Constructors), adapt),
+			External:    utils.Map(utils.NonNilSlice(ep.Externals), adapt),
+			L1Handler:   utils.Map(utils.NonNilSlice(ep.L1Handlers), adapt),
+		},
+		Program: utils.Map(cairo1.Program, p2p2core.AdaptFelt),
```
rodrodros: I know writing code this way is cooler but it doesn't apply to Go strengths. I think we should apply the same for loop mentioned in an earlier comment. If it's a pattern repeated across the code we can wrap it in a function and use generics to apply it in different scenarios (if possible, if not we are doom to have a bit of repetition). The worst part is that the `adapt` function cannot be even inlined. Suddently we have three function calls for `n` entry points (thinking on the previous comment) where there should've been none

This comments that are in the critical code path, meaning, they will be being executed at all times, should be written as efficient as possible (without being too hacky).

adapters/p2p2consensus/class.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+	ep := cairo1.EntryPoints
+	def := &starknet.SierraDefinition{
+		Abi: cairo1.Abi,
+		EntryPoints: starknet.SierraEntryPoints{
+			// WARNING: usage of utils.NonNilSlice is essential, otherwise compilation will finish with errors
+			// todo move NonNilSlice to Compile ?
+			Constructor: utils.Map(utils.NonNilSlice(ep.Constructors), adapt),
+			External:    utils.Map(utils.NonNilSlice(ep.Externals), adapt),
+			L1Handler:   utils.Map(utils.NonNilSlice(ep.L1Handlers), adapt),
```
rodrodros: Same to this. I know this functional pattern is present in other parts of the code but I wouldn't like to replicate it anymore. If we were in Haskell or Rust I wouldn't be making this type of comments. If you are going for a function, define it locally (not as lambda) so the compiler can inline it if considered

adapters/p2p2consensus/class.go
```diff
@@ -0,0 +1,90 @@
...
+		Program: utils.Map(cairo1.Program, p2p2core.AdaptFelt),
+		Version: cairo1.ContractClassVersion,
+	}
+
+	compiledClass, err := compiler.Compile(def)
+	if err != nil {
+		return nil, err
+	}
+
+	return sn2core.AdaptCompiledClass(compiledClass)
```
rodrodros: I would like that we return by value here, but it seems it is not easy. Will it affect a large portion of the code? If that's the case can you add an issue for it

adapters/p2p2consensus/proposal.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/NethermindEth/juno/utils"
+	common "github.com/starknet-io/starknet-p2pspecs/p2p/proto/common"
+	p2pconsensus "github.com/starknet-io/starknet-p2pspecs/p2p/proto/consensus/consensus"
+)
+
+func U128ToFelt(u *common.Uint128) *felt.Felt {
+	lowBig := new(big.Int).SetUint64(u.Low)
+	highBig := new(big.Int).SetUint64(u.High)
+	highBig.Lsh(highBig, 64) //nolint:mnd
+	return new(felt.Felt).SetBigInt(highBig.Or(highBig, lowBig))
```
rodrodros: Big int is slow and expensive to use, let's find a smarter way to implement it. Also, the felt returned should be a value type. Do you disagree these kind of types in this context should be returned by value?

adapters/p2p2consensus/proposal.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+func AdaptProposalInit(msg *p2pconsensus.ProposalInit) consensus.ProposalInit {
+	return consensus.ProposalInit{
+		BlockNum: msg.BlockNumber,
+		Proposer: *new(felt.Felt).SetBytes(msg.Proposer.Elements),
+	}
+}
+
+func AdaptBlockInfo(msg *p2pconsensus.BlockInfo) *consensus.BlockInfo {
```
rodrodros: Return by value

adapters/p2p2consensus/proposal.go
rodrodros: Note to self: investigate `common.Uint128`

adapters/p2p2consensus/class.go
rianhughes: This file has been deleted

adapters/p2p2consensus/class.go
rianhughes: This file has been deleted. I introduced the common functions as discussed

adapters/p2p2consensus/transaction.go
rianhughes: I introduced the common functions as discussed

adapters/p2p2consensus/class.go
rianhughes: This file has been deleted.

adapters/p2p2consensus/class.go
rianhughes: This file has been deleted.

adapters/p2p2consensus/class.go
rianhughes: This file has been deleted.

adapters/p2p2consensus/class.go
rianhughes: This file has been deleted.

adapters/p2p2consensus/class.go
rianhughes: This file has been deleted.

adapters/p2p2consensus/proposal.go
rianhughes: Updated

adapters/p2p2consensus/class.go
rianhughes: This file has been deleted.

adapters/p2p2consensus/proposal.go
rianhughes: It turns out we already have a Uin128ToFelt adapter. I've deleted this code, and reused the existing code

rianhughes: As for updating the function to return by value, this has larger knock-on effects in the code, since we specify most felts by reference, and some are optional etc. I think its outside the scope of this PR

adapters/p2p2consensus/proposal.go
infrmtcs: Thanks, I created #2900 to track this.

infrmtcs (approved):

adapters/p2p2consensus/proposal.go
infrmtcs: This is interesting that, I think this implementation should be faster than the existing `AdaptUint128` function, because:
1. `AdaptUint128` uses `felt.Felt.SetBytes()`, which underlying also use `SetBigInt()`
2. `AdaptUint128` uses an array of 16 byte as the intermediate representation, while this use an array of 2 uint64 directly (on 64-bit machine).

infrmtcs: That said, the compiler optimization is unpredictable, so we'll need benchmark to compare the 2 implementations.

adapters/p2p2core/transaction.go
```diff
@@ -11,7 +11,123 @@ import (
...
+		TransactionHash:    AdaptHash(txnHash),
+		ContractAddress:    AdaptAddress(tx.Address),
+		EntryPointSelector: AdaptFelt(tx.EntryPointSelector),
+		Nonce:              AdaptFelt(tx.Nonce),
+		CallData:           utils.Map(tx.Calldata, AdaptFelt),
+		Version:            txVersion(0),
+	}
+}
+
+//nolint:funlen
```
rodrodros, line 130: Is this linter warning still required?

rianhughes: The linter complains without it, so yes

PR #2880, opened 2025-06-11
fix: setting header values for new block hash functions

rianhughes:
To compute the block hash for `0.13.4` we need to set the `fee`s in the block header. Normally we get these from the `FGW`. However, for `consensus`, the validator never receives these values (in the empty block flow - we never receive `BlockInfo`) so we can't compute the block hash. We also can't use the values from the old header since they do not always exist (eg we added fees to the header in 0.13.4 that don't exist in old versions).

This PR is a short-term fix.

Longer term, we need the consensus spec to be updated, so that the fields required to compute the block hash are always sent to the validator (if move fields from BlockInfo to ProposalInit).

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -294,6 +294,28 @@ func (b *Builder) ProposalInit(pInit *types.ProposalInit) error {
 		Receipts:     []*core.TransactionReceipt{},
 	}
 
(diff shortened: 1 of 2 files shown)


EgeCaner (approved):
LGTM

Issue #2894, opened 2025-06-12
Tendermint: update tests to use latest starknet protocol version


rianhughes:
In Tendermint, the proposer only sends a subset of the gas fees we need to compute the block hash. We should use the header values by default. This prevents validator_test.go tests from crashing when we will update supportedStarknetVersion to 0.14.0.

Also, I updated the empty flow test to use the same custom blockchain as the non-empty flow test. There's no need to have both.

Also also, we require tendermint to run on the latest starknet versions, not old blocks. The reason is the old blocks don't have the values we need to compute the blockhash function (namely the fee prices)

PR #2895, opened 2025-06-12
Fix: update Tendermint tests and validator logic

rianhughes:
closes https://github.com/NethermindEth/juno/issues/2894

In Tendermint, the proposer only sends a subset of the gas fees we need to compute the block hash. We should use the header values by default. This prevents validator_test.go tests from crashing when we will update supportedStarknetVersion to 0.14.0.

Also, I updated the empty flow test to use the same custom blockchain as the non-empty flow test. There's no need to have both.

Also also, as discussed with @infrmtcs  we require tendermint to run on the latest starknet versions, not old versions. The reason is the old blocks don't have the values we need to compute the blockhash function (namely the fee prices)

### DIFF
--- a/consensus/p2p/buffered/buffered_topic_subscription.go
+++ b/consensus/p2p/buffered/buffered_topic_subscription.go
@@ -1,0 +1,45 @@
+package buffered
+
+import (
+	"context"
+	"errors"
+
+	"github.com/NethermindEth/juno/utils"
+	pubsub "github.com/libp2p/go-libp2p-pubsub"
+)
+
+type TopicSubscription struct {
+	log        utils.Logger
+	bufferSize int
+	callback   func(context.Context, *pubsub.Message)
+}
+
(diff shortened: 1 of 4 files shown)


rodrigo-pino (approved):
Approving to unblock

consensus/tendermint/tendermint.go
```diff
@@ -8,8 +8,14 @@ import (
 	"github.com/NethermindEth/juno/utils"
 )
 
+// H 3 R3 , we propose
+// Call Value() V - this blocks out consensus progress
+// The network commits H3 R2 while we generate a proposal
+// We could cancel the block builder, and move to the new height
+// Msg times : delta + maxTime
+
```
infrmtcs: Don't forget to remove our discussion draft :smile:

infrmtcs (approved):
LGTM, just need to remove the unrelated comments

consensus/tendermint/tendermint.go
rianhughes: Lol nice catch!

rianhughes: Done

PR #2897, opened 2025-06-12
feat: Implement vote broadcasters

infrmtcs:


consensus/p2p/buffered/proto_broadcaster.go
```diff
@@ -0,0 +1,51 @@
...
+			return
+		case msg := <-b.ch:
+			msgBytes, err := proto.Marshal(msg)
+			if err != nil {
+				b.log.Errorw("unable to marshal message", "error", err)
+				continue
+			}
+
+			for {
+				if err := topic.Publish(ctx, msgBytes, pubsub.WithReadiness(pubsub.MinTopicSize(1))); err != nil && !errors.Is(err, context.Canceled) {
```
rodrodros: It looks like a super long line, I think it can be slight readability improvement by separating the error declaration

consensus/p2p/vote/vote_listeners.go
```diff
@@ -0,0 +1,75 @@
...
+	"google.golang.org/protobuf/proto"
+)
+
+type voteListener[M types.Message[V, H, A], V types.Hashable[H], H types.Hash, A types.Addr] chan M
+
+func newListener[M types.Message[V, H, A], V types.Hashable[H], H types.Hash, A types.Addr](bufferSize int) voteListener[M, V, H, A] {
+	return voteListener[M, V, H, A](make(chan M, bufferSize))
+}
+
+func (l voteListener[M, V, H, A]) Listen() <-chan M {
```
rodrodros, line 21: method should take a pointer (`l *voteListener[...]` instead of `l voteListener[...]`), otherwise everytime there is a method call there will be a copy will be performed. Is this intentional?

consensus/p2p/vote/vote_listeners.go
```diff
@@ -0,0 +1,75 @@
...
+	return l
+}
+
+func (l voteListener[M, V, H, A]) Receive(ctx context.Context, message M) {
```
rodrodros, line 25: same

consensus/p2p/vote/vote_listeners.go
```diff
@@ -0,0 +1,75 @@
...
+	log               utils.Logger
+	PrevoteListener   voteListener[types.Prevote[H, A], V, H, A]
+	PrecommitListener voteListener[types.Precommit[H, A], V, H, A]
+}
+
+func NewVoteListeners[V types.Hashable[H], H types.Hash, A types.Addr](
+	log utils.Logger,
+	voteAdapter VoteAdapter[H, A],
+	bufferSizeConfig *config.BufferSizes,
+) *voteListeners[V, H, A] {
```
rodrodros: I don't see any reason not to return by value here. Is this intentional?

consensus/p2p/vote/vote_broadcasters_listeners_test.go
```diff
@@ -0,0 +1,218 @@
...
+		}
+	}()
+
+	// Self broadcast invalid votes and make sure they are not processed
+	go func() {
+		for range invalidCount {
+			vote := &consensus.Vote{}
+			voteBytes, err := proto.Marshal(vote)
+			require.NoError(t, err)
+			_ = destination.topic.Publish(t.Context(), voteBytes)
```
rodrodros: Maybe `vote` variable creation is unnecessary, and we can create directly on
```
voteBytes, err := proto.Marshal(&consensus.Vote{})
```

consensus/p2p/vote/vote_listeners.go
infrmtcs: `voteListener` is only a thin wrapper of a channel to comply with the requirement of `Listener` interface, which requires `Listen() <-chan M`, so it should be fine to copy.

consensus/p2p/vote/vote_listeners.go
infrmtcs: Same as `Listen()`

PR #2902, opened 2025-06-17
refactor: concentrate commitment computation to one method

infrmtcs:
Make some methods in `Builder` internal and replace by a public method to compute `ProposalCommitment`. This is to make sure that the computations in validator and proposer must follow a single function.

### DIFF
--- a/builder/builder.go
+++ b/builder/builder.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"sync/atomic"
 
+	"github.com/Masterminds/semver/v3"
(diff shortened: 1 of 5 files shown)


rodrigo-pino (approved):

builder/builder.go
```diff
@@ -110,6 +112,12 @@ func (b *Builder) InitPendingBlock(sequencerAddress *felt.Felt) error {
 	if err != nil {
 		return err
 	}
+
```
rianhughes, line 115: Nice improvement to getRevealedBlockHash

rianhughes (approved):

PR #2905, opened 2025-06-17
feat(state): integrate new state package with the hash scheme of trieDB

MaksymMalicki:
Resolves [#2812](https://github.com/NethermindEth/juno/issues/2812)

### DIFF
--- a/rpc/v6/transaction.go
+++ b/rpc/v6/transaction.go
@@ -494,7 +494,7 @@ func (h *Handler) TransactionByBlockIDAndIndex(id BlockID, txIndex int) (*Transa
 			return nil, rpccore.ErrBlockNotFound
 		}
 
-		if uint64(txIndex) > pending.GetBlock().TransactionCount {
+		if uint64(txIndex) >= pending.GetBlock().TransactionCount {
 			return nil, rpccore.ErrInvalidTxIndex
 		}
 
--- a/rpc/v7/transaction.go
+++ b/rpc/v7/transaction.go
@@ -402,7 +402,7 @@ func (h *Handler) TransactionByBlockIDAndIndex(id BlockID, txIndex int) (*Transa
 			return nil, rpccore.ErrBlockNotFound
 		}
 
-		if uint64(txIndex) > pending.GetBlock().TransactionCount {
+		if uint64(txIndex) >= pending.GetBlock().TransactionCount {
 			return nil, rpccore.ErrInvalidTxIndex
 		}
 
--- a/rpc/v8/transaction.go
+++ b/rpc/v8/transaction.go
@@ -514,7 +514,7 @@ func (h *Handler) TransactionByBlockIDAndIndex(
 			return nil, rpccore.ErrBlockNotFound
 		}
 
-		if uint64(txIndex) > pending.GetBlock().TransactionCount {
+		if uint64(txIndex) >= pending.GetBlock().TransactionCount {
 			return nil, rpccore.ErrInvalidTxIndex
 		}
 
--- a/rpc/v9/transaction.go
+++ b/rpc/v9/transaction.go
@@ -528,7 +528,7 @@ func (h *Handler) TransactionByBlockIDAndIndex(
 			return nil, rpccore.ErrBlockNotFound
 		}
 
-		if uint64(txIndex) > pending.GetBlock().TransactionCount {
+		if uint64(txIndex) >= pending.GetBlock().TransactionCount {
 			return nil, rpccore.ErrInvalidTxIndex
 		}
 


core/accessors.go
```diff
@@ -552,3 +552,54 @@ func WriteRunningEventFilter(w db.KeyValueWriter, filter *RunningEventFilter) er
...
+}
+
+// This is used to get the contract storage root for a given state commitment and contract commitment,
+// needed to properly initialise the trie or recover the cache, if the triedb is in hash scheme
+func GetContractStorageRoot(r db.KeyValueReader, stateCommitment, contractCommitment *felt.Felt) ([]byte, error) {
+	var val []byte
+	err := r.Get(db.ContractStorageRootKey(stateCommitment, contractCommitment), func(data []byte) error {
+		val = data
+		return nil
+	})
```
rodrodros, line 595: For me it's so crazy that we have to use this pattern for such a simple operation, instead of directly having something like

```go
val, err := r.GetRaw(db.ContractStorageRootKey(stateCommitment, contractCommintment))
```

Can I get some thoughts why is that not ideal or it doesn't make sense. 

If I am not mistaken I've seen this approach several times across the code. Why are we using this one specifically?

core/state/database.go
```diff
@@ -27,16 +31,106 @@ func NewStateDB(disk db.KeyValueStore, triedb *triedb.Database) *StateDB {
 }
 
 // Opens a class trie for the given state root
+//
+//nolint:dupl
```
rodrodros: Add a reason per lint ignore, or fix it.


Both this function and the other one (`ContractTrie`), specifically on the `triedb.HashDB` case, code seems to be similar. 

Having duplicated code means it is twice the maintenance, is there an issue I am missing that is not allowing to wrap the repeated code in a common function?

core/state/state.go
```diff
@@ -778,6 +815,41 @@ func (s *State) compareContracts(a, b felt.Felt) int {
 	return len(contractB.dirtyStorage) - len(contractA.dirtyStorage)
 }
 
+// TODO(maksym): this could be cached
```
rodrodros: Could you expand on this todo

MaksymMalicki: Actually, this was the thought I had when writing this. I was thinking about putting those into a cache, and flushing later with the accumulated trie modifications, once they hit certain buffer size. But this idea was dropped. In the hash scheme (which is only used for archive node purposes), we are flushing the changes on every block, to ensure the data integrity for the archive node providers. Caching only this information won't make a big difference and is prone to node dirty node shutdowns - I will remove this todo

core/state/state.go
```diff
@@ -64,11 +64,13 @@ type State struct {
 }
 
 func New(stateRoot *felt.Felt, db *StateDB) (*State, error) {
+	fmt.Println("contract trie")
 	contractTrie, err := db.ContractTrie(stateRoot)
 	if err != nil {
 		return nil, err
 	}
 
+	fmt.Println("class trie")
```
rodrodros: Oops

core/state/state.go
rodrodros: Is no longer necessary?

core/state/state.go
```diff
@@ -778,6 +817,40 @@ func (s *State) compareContracts(a, b felt.Felt) int {
 	return len(contractB.dirtyStorage) - len(contractA.dirtyStorage)
 }
 
+func (s *State) persistTrieRoots(
+	stateCommitment, parentRoot, classRoot, contractRoot *felt.Felt,
+	storageRoots map[felt.Felt]felt.Felt,
+	revert bool,
+) error {
```
rodrodros, line 828: Can we give a small description to this function

core/trie2/triedb/database.go
```diff
@@ -12,6 +12,13 @@ import (
 	"github.com/NethermindEth/juno/db"
 )
 
+type Scheme string
+
+const (
+	PathDB Scheme = "pathdb"
+	HashDB Scheme = "hashdb"
+)
+
```
rodrodros: Why strings vs nums?

MaksymMalicki: no particular reason outside of explicit readability. but yeah, changing to nums sounds better

core/state/state.go
MaksymMalicki: my bad

core/state/state.go
MaksymMalicki: added comment, explaining the function

core/accessors.go
MaksymMalicki: During the review of the original PR, I asked the same question: https://github.com/NethermindEth/juno/pull/2661#discussion_r2012005225

core/trie2/triedb/database.go
rodrodros: It's way faster comparison wise and takes less space memory wise, let's use nums 👍

core/accessors.go
rodrodros: I am not sure that answer is fully correct, it would be odd to me that a callback which uses an external variable (hence applying a closure) would be more efficient, specially in memory. Let me see if I can back it up with some numbers

PR #2912, opened 2025-06-18
feat(state): Integrate new state package with Juno

MaksymMalicki:


adapters/p2p2core/state.go
```diff
@@ -53,16 +54,16 @@ func AdaptStateDiff(reader core.StateReader, contractDiffs []*state.ContractDiff
 				classHash: diff.ClassHash,
 			}
 
-			var stateClassHash *felt.Felt
+			var stateClassHash felt.Felt
 			if reader == nil {
 				// zero block
-				stateClassHash = &felt.Zero
+				stateClassHash = felt.Zero
 			} else {
```
rodrodros, line 61:
```go
stateClassHash := felt.Zero
if reader != nil {
  ...
}
```

PR #2913, opened 2025-06-19
tendermint: connect validators

rianhughes:
This PR connects the builder-validator with the p2p-validator.

I have renamed the p2p/validator to p2p/transition for two reasons, 1: it prevents a pkg name conflict, and 2: the transition pkg inherits the builder-validator, but is mostly concerned with transitioning the "p2p state machine".

I have skipped on of the transition tests. This test executes a non-empty block flow. The txn it tries to execute returns an "argent-invalid-tx-version" error. I think the class we are useing might not support v3 txns (although I need to confirm this), which would require updating the genesis logic to include a more recent class, and hand crafting a txn to make it pass. If so, it would be cleaner to do this work in a separate PR.

In the transition logic, we are populating the value before we can actually know it, I left some todos indicating we should remove this. I want to get confirmation with @infrmtcs before I remove this. It also changes the tests, so it might be cleaner to open a small PR to address this there.

The genesis block header now includes fields required to compute the block hash for later starknet versions


consensus/p2p/statemachine/proposal_stream.go
```diff
@@ -1,4 +1,4 @@
-package validator
+package statemachine
```
infrmtcs, line 1: Why do we rename this to `statemachine`? `validator` is much clearer, because this state machine is only for `validator`. Besides that, this only increases the diff and increase resolve conflict time and review time.

rianhughes: Copying my reasons from the PR description

I have renamed the p2p/validator to p2p/transition for two reasons, 1: it prevents a pkg name conflict, and 2: the transition pkg inherits the builder-validator, but is mostly concerned with transitioning the "p2p state machine".

rianhughes: >this only increases the diff 

The diff is pretty minor, no? It's a single line at the top of the file, nothing else

rianhughes: I don't mind what we call this pkg tbh, but I don't think we should have two pkgs named `validator`, especially when one `validator` imports the other `validator`. And I don't think it makes sense to merge the two pkgs

infrmtcs: > The diff is pretty minor, no? It's a single line at the top of the file, nothing else

It creates noise going back and forth between files.

blockchain/blockchain.go
```diff
@@ -727,8 +727,16 @@ func (b *Blockchain) StoreGenesis(
...
+			L1GasPriceETH:    new(felt.Felt).SetUint64(1),
+			L1GasPriceSTRK:   new(felt.Felt).SetUint64(1),
+			L1DataGasPrice: &core.GasPrice{
+				PriceInWei: new(felt.Felt).SetUint64(1),
+				PriceInFri: new(felt.Felt).SetUint64(1),
+			},
+			L2GasPrice: &core.GasPrice{
+				PriceInWei: new(felt.Felt).SetUint64(1),
+				PriceInFri: new(felt.Felt).SetUint64(1),
+			},
```
infrmtcs, line 739: Using the same test number will never catch errors like accidentally setting `L1GasPriceETH` to `L1GasPriceSTRK`

consensus/p2p/statemachine/proposal_stream_demux_test.go
```diff
@@ -55,8 +57,15 @@ func TestProposalStreamDemux(t *testing.T) {
...
-	demux := NewProposalStreamDemux(logger, NewTransition(), &config.DefaultBufferSizes, commitNotifier, block1)
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+	mockedValidator := mocks.NewMockValidator[value, felt.Felt, felt.Felt](ctrl)
+	mockedValidator.EXPECT().ProposalInit(gomock.Any()).AnyTimes().Return(nil)
+	mockedValidator.EXPECT().BlockInfo(gomock.Any()).AnyTimes()
+	mockedValidator.EXPECT().TransactionBatch(gomock.Any()).AnyTimes().Return(nil)
+	mockedValidator.EXPECT().ProposalCommitment(gomock.Any()).AnyTimes().Return(nil)
+	mockedValidator.EXPECT().ProposalFin(gomock.Any()).AnyTimes().Return(nil)
+	demux := NewProposalStreamDemux(logger, NewTransition[value, felt.Felt, felt.Felt](mockedValidator), &config.DefaultBufferSizes, commitNotifier, block1)
```
infrmtcs, line 68: Why do we use mock here?

consensus/p2p/statemachine/proposal_stream_demux_test.go
```diff
@@ -176,24 +185,32 @@ func createProposal(t *testing.T, height types.Height) ([][]byte, starknet.Propo
...
 			BlockInfo: &consensus.BlockInfo{
-				BlockNumber: uint64(height),
-				Timestamp:   timestamp,
+				Builder:           proposerAddress,
+				BlockNumber:       uint64(height),
+				Timestamp:         value,
+				L2GasPriceFri:     someUint128,
+				L1GasPriceWei:     someUint128,
+				L1DataGasPriceWei: someUint128,
+				EthToStrkRate:     someUint128,
```
infrmtcs, line 212: Using the same test number will never catch errors like accidentally setting `L1GasPriceWei` to `L1DataGasPriceWei`

consensus/p2p/statemachine/proposal_stream_test.go
```diff
@@ -78,7 +87,16 @@ func buildMessage(t *testing.T, sequenceNumber uint64, proposalPart *consensus.P
...
+
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+	mockedValidator := mocks.NewMockValidator[value, felt.Felt, felt.Felt](ctrl)
+	mockedValidator.EXPECT().ProposalInit(gomock.Any()).AnyTimes().Return(nil)
+	mockedValidator.EXPECT().BlockInfo(gomock.Any()).AnyTimes()
+	mockedValidator.EXPECT().TransactionBatch(gomock.Any()).AnyTimes().Return(nil)
+	mockedValidator.EXPECT().ProposalCommitment(gomock.Any()).AnyTimes().Return(nil)
+	mockedValidator.EXPECT().ProposalFin(gomock.Any()).AnyTimes().Return(nil)
+	stream := newSingleProposalStream(utils.NewNopZapLogger(), NewTransition[value, felt.Felt, felt.Felt](mockedValidator), 0, nil)
```
infrmtcs, line 99: Why do we use mock here?

consensus/p2p/statemachine/proposal_stream_test.go
```diff
@@ -324,8 +356,16 @@ func TestProposalStream_ProcessMessage(t *testing.T) {
...
-	stream := newSingleProposalStream(utils.NewNopZapLogger(), NewTransition(), 0, outputs)
-	_, err := stream.start(t.Context(), buildMessage(t, 1, &consensus.ProposalPart{
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+	mockedValidator := mocks.NewMockValidator[value, felt.Felt, felt.Felt](ctrl)
+	mockedValidator.EXPECT().ProposalInit(gomock.Any()).AnyTimes().Return(nil)
+	mockedValidator.EXPECT().BlockInfo(gomock.Any()).AnyTimes()
+	mockedValidator.EXPECT().TransactionBatch(gomock.Any()).AnyTimes().Return(nil)
+	mockedValidator.EXPECT().ProposalCommitment(gomock.Any()).AnyTimes().Return(nil)
+	mockedValidator.EXPECT().ProposalFin(gomock.Any()).AnyTimes().Return(nil)
```
infrmtcs, line 366: Why do we use mock here?

consensus/p2p/statemachine/proposal_stream_test.go
```diff
@@ -180,8 +203,14 @@ func TestProposalStream_ProcessMessage(t *testing.T) {
...
 					Messages: &consensus.ProposalPart_BlockInfo{BlockInfo: &consensus.BlockInfo{
-						BlockNumber: uint64(height),
-						Timestamp:   timestamp,
+						Builder:           sender,
+						BlockNumber:       uint64(height),
+						Timestamp:         valueUint,
+						L2GasPriceFri:     someUint128,
+						L1GasPriceWei:     someUint128,
+						L1DataGasPriceWei: someUint128,
+						EthToStrkRate:     someUint128,
```
infrmtcs, line 212: Use different numbers

consensus/p2p/statemachine/validator_utils_test.go
```diff
@@ -0,0 +1,136 @@
...
+		StateDiffCommitment:       someHash,
+		TransactionCommitment:     someHash,
+		EventCommitment:           someHash,
+		ReceiptCommitment:         someHash,
+		ConcatenatedCounts:        someFelt,
+		L1GasPriceFri:             someU128,
+		L1DataGasPriceFri:         someU128,
+		L2GasPriceFri:             someU128,
+		L2GasUsed:                 someU128,
+		NextL2GasPriceFri:         someU128,
```
infrmtcs, line 133: Use different, random numbers

consensus/p2p/statemachine/transition_test.go
```diff
@@ -0,0 +1,246 @@
...
+				},
+				L1DataGas: &transaction.ResourceLimits{
+					MaxAmount:       hexToCommonFelt252(t, "0x1"),
+					MaxPricePerUnit: hexToCommonFelt252(t, "0x1"),
+				},
+				L2Gas: &transaction.ResourceLimits{
+					MaxAmount:       hexToCommonFelt252(t, "0x1"),
+					MaxPricePerUnit: hexToCommonFelt252(t, "0x1"),
+				},
+			},
```
infrmtcs, line 223: Please use different, random numbers

consensus/p2p/statemachine/transition_test.go
```diff
@@ -0,0 +1,246 @@
...
+			StateDiffCommitment:       hexToCommonHash(t, "0x49973925542c74a9d9ff0efaa98c61e1225d0aedb708092433cbbb20836d30a"),
+			TransactionCommitment:     hexToCommonHash(t, "0x0"), // Todo: This should NOT be 0??
+			EventCommitment:           hexToCommonHash(t, "0x0"), // Todo: This should NOT be 0??
+			ReceiptCommitment:         hexToCommonHash(t, "0x0"), // Todo: This should NOT be 0??
+			ConcatenatedCounts:        hexToCommonFelt252(t, "0x8000000000000000"),
+			L1GasPriceFri:             someU128, // Todo: update when we support this
+			L1DataGasPriceFri:         someU128, // Todo: update when we support this
+			L2GasPriceFri:             someU128, // Todo: update when we support this
+			L2GasUsed:                 someU128, // Todo: update when we support this
+			NextL2GasPriceFri:         someU128, // Todo: update when we support this
```
infrmtcs, line 168: Please use different, random numbers

consensus/p2p/statemachine/transition.go
```diff
@@ -94,68 +97,95 @@ func (t *transition) OnEmptyBlockCommitment(
...
-// TODO: Implement this function properly
-func (t *transition) OnTransactions(
+func (t *transition[V, H, A]) OnTransactions(
 	ctx context.Context,
 	state *ReceivingTransactionsState,
 	transactions []*consensus.ConsensusTransaction,
 ) (*ReceivingTransactionsState, error) {
+	txns := make([]types.Transaction, len(transactions))
+	for i := range transactions {
+		txn, class, err := p2p2consensus.AdaptTransaction(transactions[i])
```
infrmtcs, line 126:
```suggestion
		txns[i].Transaction, txns[i].Class, err = p2p2consensus.AdaptTransaction(transactions[i])
```
to avoid copying

consensus/p2p/statemachine/transition.go
```diff
@@ -94,68 +97,95 @@ func (t *transition) OnEmptyBlockCommitment(
...
+	if err != nil {
+		return nil, err
+	}
+
 	return &ReceivingTransactionsState{
 		Header:     state.Header,
 		ValidRound: state.ValidRound,
-		Value:      state.Value,
+		// Todo: remove, OnTransactions can't know value / block hash yet..
+		Value: state.Value,
```
infrmtcs, line 145: Let's remove it. The state should keep only what it knows.

consensus/p2p/statemachine/transition.go
```diff
@@ -49,14 +45,19 @@ type Transition interface {
 	) (*FinState, error)
 }
 
-type transition struct{}
+type transition[V types.Hashable[H], H types.Hash, A types.Addr] struct {
+	validator validator.Validator[V, H, A]
```
infrmtcs, line 49: This is important:
Transition shouldn't have the state, but only the necessary dependencies to calculate the transitions. States should be in the states like `InitialState`,....
Why:
- `Transition` is supposed to be shared between multiple proposal stream. Putting the states here mean it's a mutable state, which means there must be a new instance created for every proposal stream.
- Putting everything inside `Transition` defeats its purpose. Firstly, we can't differentiate between what's known and what's not. For example, `InitialState` or `AwaitingBlockInfoOrCommitmentState` shouldn't know anything about the block, so it shouldn't have access to `Transition`. In case of empty proposal, we should instead create the result block directly instead of going through validator/builder flows.

consensus/p2p/statemachine/transition.go
```diff
@@ -94,68 +97,95 @@ func (t *transition) OnEmptyBlockCommitment(
...
+	if err = t.validator.ProposalCommitment(&adaptedCommitment); err != nil {
+		return nil, err
+	}
+
 	return &AwaitingProposalFinState{
 		Header:     state.Header,
 		ValidRound: state.ValidRound,
-		Value:      state.Value,
+		// Todo: remove, we aren't populating with the value/block hash we generate..
+		Value: state.Value,
```
infrmtcs, line 167: We should populate it here, because at this point, we already have the full `ProposalCommitment`, so we can calculate the hash. When `ProposalFin` arrives, what we need is to compare the calculated hash and the received hash.

consensus/p2p/statemachine/transition.go
```diff
@@ -94,68 +97,95 @@ func (t *transition) OnEmptyBlockCommitment(
...
+	err := t.validator.TransactionBatch(txns)
+	if err != nil {
+		return nil, err
+	}
+
 	return &ReceivingTransactionsState{
 		Header:     state.Header,
 		ValidRound: state.ValidRound,
-		Value:      state.Value,
+		// Todo: remove, OnTransactions can't know value / block hash yet..
```
infrmtcs, line 144: Agree to all of these comments, except the one in `OnProposalCommitment`

adapters/p2p2consensus/transaction.go
```diff
@@ -5,32 +5,32 @@ import (
 
 	"github.com/NethermindEth/juno/adapters/p2p2core"
 	"github.com/NethermindEth/juno/core"
-	"github.com/NethermindEth/juno/utils"
 	p2pconsensus "github.com/starknet-io/starknet-p2pspecs/p2p/proto/consensus/consensus"
 )
 
-func AdaptTransaction(t *p2pconsensus.ConsensusTransaction, network *utils.Network) (core.Transaction, core.Class) {
+func AdaptTransaction(t *p2pconsensus.ConsensusTransaction) (core.Transaction, core.Class, error) {
```
infrmtcs, line 11: This is from the previous PR, but why don't we return `consensus.Transaction` instead?

adapters/p2p2consensus/validation.go
```diff
@@ -0,0 +1,112 @@
...
+		return errors.New("proposer must be set")
+	}
+	return nil
+}
+
+func validateBlockInfo(b *p2pconsensus.BlockInfo) error {
+	if b == nil {
+		return errors.New("block info is nil")
+	}
+	if b.Builder == nil {
```
infrmtcs, line 23: I wasn't aware of this, but I've just realized that `Felt.SetBytes` use `Element.SetBytes`, which will implictly calculate `felt % q` with q is a big constant when len(felt) > q. This means an invalid input can still be accepted and derives another number. I believe we should reject any possible invalid value instead.

infrmtcs: Another option is write a short function `Felt.SetBytesCanonical` which uses `Element.SetBytesCanonical`. This function return error when felt validation fails, so we don't have to check anything except checking non nil (which is what we're already doing).

adapters/p2p2consensus/validation.go
```diff
@@ -0,0 +1,112 @@
...
+	if tx.Txn == nil {
+		return errors.New("txn must be set to a valid variant")
+	}
+	if tx.TransactionHash == nil {
+		return errors.New("transaction_hash must be set")
+	}
+	return nil
+}
+
+func validateProposalCommitment(p *p2pconsensus.ProposalCommitment) error { //nolint:gocyclo // simple, repetative code
```
infrmtcs, line 55: We also need to validate `ProtocolVersion`. Currently we're using `semver.MustParse`, which will panic and cause the node crash because of invalid messages from attackers.
We should run `semver.NewVersion` instead and return error if fail.

adapters/p2p2consensus/transaction.go
```diff
@@ -5,32 +5,32 @@ import (
...
-
 	switch t.Txn.(type) {
 	case *p2pconsensus.ConsensusTransaction_DeclareV3:
 		tx := t.GetDeclareV3()
 		// Todo: we pass in CompiledClassHash here, but in sync we pass in ClassHash.
 		// Are we expected to compile the class hash here???
 		class := p2p2core.AdaptCairo1Class(tx.Class)
-		return p2p2core.AdaptDeclareV3TxnCommon(tx.Common, tx.Common.CompiledClassHash, t.TransactionHash), &class
-
+		return p2p2core.AdaptDeclareV3TxnCommon(tx.Common, tx.Common.CompiledClassHash, t.TransactionHash), &class, nil
```
infrmtcs, line 24: We need a validate function for each transaction type, called inside each of these cases.

PR #2916, opened 2025-06-19
tendermint: adapters for validator

rianhughes:
Validators ensure that we don't try and access fields inside nil types

### DIFF
--- a/adapters/p2p2consensus/proposal.go
+++ b/adapters/p2p2consensus/proposal.go
@@ -10,6 +10,15 @@ import (
 )
 
 func AdaptProposalInit(msg *p2pconsensus.ProposalInit) (consensus.ProposalInit, error) {
+	if err := validateProposalInit(msg); err != nil {
+		return consensus.ProposalInit{}, err
+	}
+
+	var proposer felt.Felt
+	if err := proposer.SetBytesCanonical(msg.Proposer.Elements); err != nil {
+		return consensus.ProposalInit{}, err
+	}
+
 	var validRound consensus.Round = -1
 	if msg.ValidRound != nil {
 		validRound = consensus.Round(*msg.ValidRound)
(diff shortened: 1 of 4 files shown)


adapters/p2p2consensus/proposal.go
```diff
@@ -6,69 +6,137 @@ import (
...
 }
 
-func AdaptProposalCommitment(msg *p2pconsensus.ProposalCommitment) consensus.ProposalCommitment {
+func AdaptProposalCommitment(msg *p2pconsensus.ProposalCommitment) (consensus.ProposalCommitment, error) {
+	if err := validateProposalCommitment(msg); err != nil {
+		return consensus.ProposalCommitment{}, err
+	}
+
+	var builder felt.Felt
+	if err := builder.SetBytesCanonical(msg.Builder.Elements); err != nil {
```
infrmtcs, line 74: nit:
```suggestion
	if err := builder.SetBytesCanonical(msg.Builder.GetElements()); err != nil {
```
So we don't need a separate `validateProposalCommitment` function

core/felt/felt.go
```diff
@@ -92,6 +92,11 @@ func (z *Felt) SetBytes(e []byte) *Felt {
 	return z
 }
 
+// SetBytesCanonical forwards the call to underlying field element implementation
+func (z *Felt) SetBytesCanonical(e []byte) error {
```
infrmtcs, line 96: nit: If we have a `FromBytesCanonical` function similar to `FromBytes`, we can simplify the call site.

adapters/p2p2consensus/proposal.go
```diff
@@ -6,69 +6,137 @@ import (
...
+
+	snVersion, err := semver.NewVersion(msg.ProtocolVersion)
+	if err != nil {
+		return consensus.ProposalCommitment{}, err
+	}
+
 	return consensus.ProposalCommitment{
 		BlockNumber: msg.BlockNumber,
-		Builder:     *new(felt.Felt).SetBytes(msg.Builder.Elements),
+		Builder:     builder,
```
infrmtcs: nit: Would you mind remove the space between the fields, so we can align the field values?

infrmtcs (approved):
All comments are nits, the rest LGTM

adapters/p2p2consensus/proposal.go
rianhughes: I might keep the validate function for now just because it allows us to distinguish between cases where the field is either 1. nil, or 2. non-nil but not the correct number of bytes.

rianhughes: If we don't need this then I agree using `GetElements()` will simplify the code quite a bit

core/felt/felt.go
rianhughes: `SetBytesCanonical` checks that the sender must send [32] bytes, otherwise it returns an error. `FromBytesCanonical` allows any number of bytes. I think it's safer to have `SetBytesCanonical`

adapters/p2p2consensus/proposal.go
rianhughes: Done

rodrigo-pino (approved):

PR #2920, opened 2025-06-24
feat: Implement simple proposer store

infrmtcs:


consensus/proposal/proposal_store.go
```diff
@@ -0,0 +1,39 @@
...
+	Store(key H, value *builder.Builder)
+}
+
+type proposalStore[H types.Hash] struct {
+	syncmap.Map
+}
+
+func NewProposalStore[H types.Hash]() ProposalStore[H] {
+	return &proposalStore[H]{}
+}
```
rodrodros: Why are we creating an interface for a single type?

rodrigo-pino (approved):

rianhughes (approved):

PR #2921, opened 2025-06-24
refactor: Create a BuildResult types for data that needs to be stored

infrmtcs:
- This create a `BuildResult` type containing the output for `Builder`. This `BuildResult` only contains data and no dependencies like VM and Blockchain, so it's more suitable to be stored in proposal store (in the next PR).
- This also remove the `atomic.Pointer`. The current code isn't enough to guard concurrency, and it's better to let caller control the concurrency in this case, because we need concurrency in validator case, but we don't in proposer case.

### DIFF
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -507,8 +507,6 @@ func (b *Blockchain) revertHead(txn db.IndexedBatch) error {
 }
 
 type SimulateResult struct {
-	Block            *core.Block
-	StateUpdate      *core.StateUpdate
 	BlockCommitments *core.BlockCommitments
 	ConcatCount      felt.Felt
 }
@@ -521,8 +519,6 @@ func (b *Blockchain) Simulate(
 	newClasses map[felt.Felt]core.Class,
 	sign utils.BlockSignFunc,
 ) (SimulateResult, error) {
-	var newCommitments *core.BlockCommitments
-
 	// Simulate without commit
 	txn := b.database.NewIndexedBatch()
 	defer txn.Reset()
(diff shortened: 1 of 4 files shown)


builder/builder.go
```diff
@@ -35,7 +39,7 @@ type Builder struct {
 
 	// Builder state
 	// TODO: move to a builder state struct
-	pendingBlock      atomic.Pointer[sync.Pending]
+	pendingBlock      *sync.Pending
```
rodrodros, line 42: What's the reason for this change?

blockchain/blockchain.go
```diff
@@ -521,66 +519,36 @@ func (b *Blockchain) Simulate(
...
 	if err := b.updateStateRoots(txn, block, stateUpdate, newClasses); err != nil {
 		return SimulateResult{}, err
 	}
-	blockHash, commitments, err := core.BlockHash(
-		block,
-		stateUpdate.StateDiff,
-		b.network,
-		block.SequencerAddress)
+
+	commitments, err := b.updateBlockHash(block, stateUpdate)
```
rodrodros, line 530: 👍

builder/builder.go
infrmtcs: - Currently, this struct is returning the stored `Pending` pointer to the outside via `Pending()` and `PendingBlock()`. This means the outsider can read/write the pointer value without be guarded by `atomic.Pointer`.
- The current pattern is: read the pointer, mutate the value referenced by the pointer, store the same pointer to `atomic.Pointer`. So basically `atomic.Pointer` doesn't really guard anything.

infrmtcs: In general, there are 2 approaches:
1. (The current approach) The struct methods assume that it's not thread safe, and it's the responsibility of the caller. This means remove all `atomic.Pointer` to avoid confusion.
2. Add a `RWMutex` to guard every single method.

I chose (1) because the need for concurrency in 2 cases validator and proposer are different, and it's more "idiomatic" by using channel to provide concurrency.

builder/builder.go
```diff
@@ -329,32 +332,30 @@ func (b *Builder) ProposalInit(pInit *types.ProposalInit) error {
...
+	b.pendingBlock.Block.Header.SequencerAddress = &blockInfo.Builder
+	b.pendingBlock.Block.Header.Timestamp = blockInfo.Timestamp
+	b.pendingBlock.Block.Header.L2GasPrice.PriceInFri = &blockInfo.L2GasPriceFRI
+	b.pendingBlock.Block.Header.L1GasPriceETH = &blockInfo.L1GasPriceWEI
+	b.pendingBlock.Block.Header.L1DataGasPrice.PriceInWei = &blockInfo.L1DataGasPriceWEI
+	b.pendingBlock.Block.Header.L1DAMode = blockInfo.L1DAMode
 }
 
-func (b *Builder) ProposalCommitment() (types.ProposalCommitment, error) {
+func (b *Builder) Finish() (BuildResult, error) {
```
rianhughes, line 350: Why rename from `ProposalCommitment()`? Imo the old name indicates that this logic is tied to consensus, whereas `Finish()` is more vague. `BuildResult` is also tied to consensus, so I think it would be worth updating this name to tie it to consensus more directly

builder/builder.go
```diff
@@ -24,6 +23,11 @@ var (
 	ErrPendingParentHash  = errors.New("pending block parent hash does not match chain head")
 )
 
+type BuildResult struct {
```
rianhughes, line 26: Imo we should tie this name to the `Finish()` function

builder/builder.go
infrmtcs: The problem is that the old name implies that this calculates the `ProposalCommitment`, but in fact it runs `Simulate` and modifies the internal state.

infrmtcs: If we don't want to tie `BuildResult` to consensus, we can replace `ProposalCommitment`  by `simulatedResult` and `concatCounts`, then have another function inside `consensus` to convert from `BuildResult` to `ProposalCommitment`.

infrmtcs: Actually it can be a clearer approach, so `builder` doesn't need to know anything about consensus.

rianhughes (approved):

Issue #2919, opened 2025-06-20
Invalid HELP comment format in metrics


AntoTG:
The Prometheus metrics exposed by Juno are not following the proper `# HELP` line syntax, causing parsing failures when using [parse-prometheus-text-format](https://github.com/yunyu/parse-prometheus-text-format) library:

```
InvalidLineError: Encountered invalid line: # HELP blockchain_reads
Root Cause
```

Current format: `# HELP blockchain_reads (missing description)`
Expected format: `# HELP <metric_name> <description_text>`

According to the Prometheus exposition format specification, `# HELP` lines must include both the metric name and a description text.

Juno version affected: v0.14.6

thiagodeev:
Hi @AntoTG! Thanks for report.
I took a look at it and it's not a Juno error at all. The Prometheus exposition format specification says "_If the token is `HELP`, at least one more token is expected, which is the **metric name**. All remaining tokens are considered the docstring for that metric name._" ([ref](https://prometheus.io/docs/instrumenting/exposition_formats/#comments-help-text-and-type-information)). And the `prometheus/client_golang@1.21.0 `(the one we use) says this regarding the `Help` param ([ref](https://pkg.go.dev/github.com/prometheus/client_golang@v1.21.0/prometheus#Opts)): "_It is mandatory to set Name to a non-empty string. All other fields are optional and can safely be left at their zero value, although it is strongly encouraged to set a Help string._"
It seems the real issue is with the [parse-prometheus-text-format](https://github.com/yunyu/parse-prometheus-text-format) library you're using (last updated 5 years ago), which appears not to support empty HELP descriptions (related [issue](https://github.com/yunyu/parse-prometheus-text-format/issues/24) in their repo).

But, to help you and improve clarity, we're implementing the metrics descriptions in PR [2925](https://github.com/NethermindEth/juno/pull/2925). Please take a look and let me know if this solves your problem.

PR #2925, opened 2025-06-25
Enhance metrics: Add help descriptions for various Prometheus metrics

thiagodeev:
solves https://github.com/NethermindEth/juno/issues/2919


node/metrics.go
```diff
@@ -259,6 +275,7 @@ func makeFeederMetrics() feeder.EventListener {
 		Namespace: "feeder",
 		Subsystem: "client",
 		Name:      "request_latency",
+		Help:      "Feeder gateway request latency in seconds",
```
wojciechos: what do you say to change it to?  `Feeder client request latency in seconds`

PR #2926, opened 2025-06-25
update genesis to execute v3 transactions

rianhughes:
This PR updates the class used in the genesis state, so that it accepts v3 transactions.


consensus/proposer/proposer_test.go
```diff
@@ -24,31 +24,6 @@ import (
 // This allows us to execute real txns in the tests.
 func getCustomBuilder(t *testing.T, seqAddr *felt.Felt) (*builder.Builder, *rpc.Handler, *mempool.Pool) {
 	t.Helper()
-
```
rianhughes, line 27: removed since it isn't needed

consensus/validator/validator_test.go
```diff
@@ -36,32 +35,6 @@ func (v value) Hash() felt.Felt {
 // in these tests.
 func getCustomBC(t *testing.T, seqAddr *felt.Felt) (*builder.Builder, *blockchain.Blockchain, *core.Header) {
 	t.Helper()
-
```
rianhughes, line 39: removed since it isn't needed

genesis/genesis_prefund_accounts.json
```diff
@@ -20,29 +20,63 @@
 				"0x0"
 			]
 		},
+		"0x04718f5a0fc34cc1af16a1cdee98ffb20c31f5cd61d6ab07201858f4287c938d": {
```
rianhughes, line 23: We now deploy both the eth and strk contracts

PR #2935, opened 2025-06-27
feat: Connect validator block execution

infrmtcs:
- This PR merges `Validator` in `consensus/validator` into `Transition` in `consensus/p2p/validator`. Now the flow is connected from the demux to the block execution.
- Test fixtures are loaded from real blocks on Sepolia-Integration.


consensus/p2p/validator/transition.go
```diff
@@ -3,159 +3,296 @@ package validator
...
+	if err := compareProposalCommitment(&computed, receivedCommitment); err != nil {
+		return builder.BuildResult{}, err
+	}
+
+	return buildResult, nil
+}
+
+// Todo: the validator interface assumes that the msgs are prevalidated before it is called.
+// This is an issue because proto3 may leave messages empty, whereas starknet requires them
+// to be present
```
rodrodros, line 226: Is this Todo still relevant and should it be above `compareFeltField`?

consensus/p2p/validator/transition.go
```diff
@@ -3,159 +3,296 @@ package validator
...
+func (t *transition) finishAndValidateBlock(
+	buildState *builder.BuildState,
+	receivedCommitment *types.ProposalCommitment,
+) (builder.BuildResult, error) {
+	buildResult, err := t.builder.Finish(buildState)
+	if err != nil {
+		return builder.BuildResult{}, err
+	}
+
+	computed, err := buildResult.ProposalCommitment()
```
rianhughes: nit: `computedCommitment`

consensus/p2p/validator/proposal_stream_demux_test.go
```diff
@@ -25,12 +30,11 @@ const (
 	logLevel      = zapcore.DebugLevel
 	topicName     = "proposal-stream-demux-test"
 	block0        = block1 - 1
-	block1        = 1337
+	block1        = 1164619
 	block2        = block1 + 1
 	block3        = block2 + 1
 	delay         = 10 * time.Millisecond
```
rianhughes: Can we use the real block numbers here just so it's more clear why we store blocks 1164607, 1164610, 1164611, 1164617, 1164620, 1164621. Or at least add a comment.

infrmtcs: The test scenario is that:
- The next block to be executed is block 1
- Previous block 0 will not be executed
- Block 2 will be stored and executed after committing the current block 1
- Block 3, will be stored and executed after committing block 2, even if received in the current block 1.

However, even put the real number here, it only justifies the consecutive sequence of blocks from 1164618 to 1164621. The reason why we need the other blocks is:
- 1164617 is the head block for 1164618
- 1164607 -> 1164610 are the revealed block during execution of 1164618 -> 1164621
These are the silent dependencies not specified anywhere in the tests.

consensus/p2p/validator/proposal_stream_test.go
```diff
@@ -149,61 +152,43 @@ type step struct {
...
-	timestamp := uint64(1340)
-	value := starknet.Value(felt.FromUint64(timestamp))
-	sender := GetRandomAddress(t)
-	expectedHeader := &starknet.MessageHeader{
-		Height: height,
-		Round:  round,
-		Sender: starknet.Address(*new(felt.Felt).SetBytes(sender.Elements)),
+	testCase := TestCase{
+		Height:       1164618,
+		Round:        1339,
```
rianhughes, line 157: That's a lot of rounds haha

consensus/p2p/validator/proposal_stream_demux_test.go
rianhughes: Adding the above as a comment in the code might be useful.

But my comment was more related to the fact that it's not clear why (eg) `1164607.json` is in the codebase, unless you are very familiar with this specific test. By adding a `// uses 1164607.json` comment it will help other devs understand where this file is being used

rianhughes:
Apart form the small comments above this PR LGMT

consensus/p2p/validator/transition.go
infrmtcs: Yes, the TODO was because this function is expecting that `received` is non-nil`, and the caller needs to handle nil (possible, because of protobuf) before calling this function.

PR #2938, opened 2025-06-30
RPCv9 PreConfirmed block polling, and integration with RPCs

EgeCaner:
This PR includes:
- Introduces PreConfirmed block type
- PreConfirmed block polling from feeder
- Transition logic from polling `pending` blocks to `pre_confirmed` blocks
- Accessors for pending data for RPCs, to support queries on `pending` for RPC versions <= 9 when network upgrades to `0.14.0`
- When networks starts to fetch pre_confirmed blocks, RPC version less than 9, serves empty `pending` block derived from latest, and for state of `latest` block as `StateReader` and `StateWriter`, when blockID == `pending` 
 - Introduces RPCv9 statuses `PRE_CONFIRMED` and `CANDIDATE`.
 - RPCv9 when `starknet_version` < `0.14.0`, serves data from `pending` block and return `ACCEPTED_ON_L2`, and will start returning `PRE_CONFIRMED` and `CANDIDATE` statuses, when polling `pre_confirmed`  block.
 - RPCv9 re-implements methods which RPCv8 relies on RPCv6, to support new `blockStatus` and to be able to access `pre_confirmed` data


adapters/sn2core/sn2core.go
```diff
@@ -397,12 +411,96 @@ func AdaptStateUpdate(response *starknet.StateUpdate) (*core.StateUpdate, error)
...
+	for _, stateDiff := range response.TransactionStateDiffs {
+		if stateDiff == nil {
+			break
+		}
+
+		if adaptedStateDiff, err = AdaptStateDiff(stateDiff); err != nil {
+			return nil, err
+		}
+		txStateDiffs = append(txStateDiffs, adaptedStateDiff)
+	}
```
rodrodros: Let's use indexing over append since it is several times faster (i.e. `slice[i] = v` instead of `slice = append(slice, v)`)

rpc/rpccore/rpccore.go
```diff
@@ -72,6 +72,7 @@ var (
 	ErrTooManyAddressesInFilter        = &jsonrpc.Error{Code: 67, Message: "Too many addresses in filter sender_address filter"}
 	ErrTooManyBlocksBack               = &jsonrpc.Error{Code: 68, Message: fmt.Sprintf("Cannot go back more than %v blocks", MaxBlocksBack)}
 	ErrCallOnPending                   = &jsonrpc.Error{Code: 69, Message: "This method does not support being called on the pending block"}
+	ErrCallOnPreConfirmed              = &jsonrpc.Error{Code: 70, Message: "This method does not support being called on the pre_confirmed block"} //nolint:lll
```
rodrodros: remove lint ignore

rpc/v9/trace.go
```diff
@@ -194,7 +193,7 @@ func (h *Handler) traceBlockTransactions(ctx context.Context, block *core.Block)
 		headState       core.StateReader
 		headStateCloser blockchain.StateCloser
 	)
-	if isPending {
+	if isPreConfirmed {
 		headState, headStateCloser, err = h.syncReader.PendingState()
 	} else {
 		headState, headStateCloser, err = h.bcReader.HeadState()
```
rodrodros, line 199: Is this correct?

rpc/v9/pending_data_wrapper.go
rodrodros, line 1: Are the `pending_data_wrapper` scripts similar in any fashion, is there a possibility to re-use them?

EgeCaner: For RPC version < 9, they are all same but wrappers wraps the syncReader.PendingData thus need to access to the handler or needs to recieve syncReader as parameter to be able to reuse.  We can reuse them expect for RPCv9 since its behaviour is different. If can access both stateReader and RPC version, we can have one pending_data_wrapper shared among all RPCs, but Im not sure such interface will be better than now

rpc/v9/trace.go
EgeCaner: You meant using `h.syncReader.PendingState()`? If so yes it should be correct. PreConfirmedBlock and PendingBlock uses same type `PendingState`.

```go

type PendingState struct {
	stateDiff  *core.StateDiff
	newClasses map[felt.Felt]core.Class
	head       core.StateReader
}
```

adapters/sn2core/sn2core.go
EgeCaner: I didnt used indexing approach here because the size of the actual size of `txStateDiffs` is not know upfront, it can be at max `len(response.TransactionsStateDiff)` but `response.TransactionsStateDiff` will have null entries if there is candidate transactions. Here I use append to derive the size of expected transactions and candidate transactions from there, and used in indexing approach for other fields

rodrodros: Right, makes sense 💯

adapters/sn2core/sn2core_test.go
```diff
@@ -623,3 +623,112 @@ func TestAdaptCompiledClass(t *testing.T) {
 	require.NoError(t, err)
 	assert.Nil(t, result)
 }
+
+func TestAdaptPreConfirmed(t *testing.T) { //nolint:gocyclo
```
rodrodros: Let's add a reason to the //nolint directives. Usually we do, if it makes sense to have the directive, let's write the reason why. If there is no reason why, then it is best to not have the directive and refactor the code.

In this case I think the directive ignore is not wrong since the function makes sense as is

clients/feeder/feeder.go
```diff
@@ -127,7 +127,7 @@ func NewTestClient(t testing.TB, network *utils.Network) *Client {
 	return c
 }
 
-func newTestServer(t testing.TB, network *utils.Network) *httptest.Server {
+func newTestServer(t testing.TB, network *utils.Network) *httptest.Server { //nolint:gocyclo
```
rodrodros: Let's add a reason please

core/pending.go
```diff
@@ -0,0 +1,112 @@
...
+
+type PendingData struct {
+	data PendingDataInterface
+}
+
+func NewPendingData(data PendingDataInterface) *PendingData {
+	return &PendingData{
+		data: data,
+	}
+}
```
rodrodros: Return by value please, it holds a reference so it is similar to a pointer

core/pending.go
```diff
@@ -0,0 +1,112 @@
...
+		StateUpdate:           stateUpdate,
+		NewClasses:            make(map[felt.Felt]Class),
+		TransactionStateDiffs: transactionStateDiffs,
+		CandidateTxs:          candidateTxs,
+	}
+}
+
+func (p *PreConfirmed) AsPendingData() *PendingData {
+	return &PendingData{data: p}
+}
```
rodrodros: I believe you should return by value, it just holds a pointer. `PendingData` represents a pointer to `PreConfirmed` by returning a pointer to `PendingData` you are returning a pointer to a pointer (basically, `**PreConfirmed`). Is that the intention?

core/pending.go
```diff
@@ -0,0 +1,112 @@
...
+
+func NewPreConfirmed(block *Block, stateUpdate *StateUpdate, transactionStateDiffs []*StateDiff, candidateTxs []Transaction) *PreConfirmed {
+	return &PreConfirmed{
+		Block:                 block,
+		StateUpdate:           stateUpdate,
+		NewClasses:            make(map[felt.Felt]Class),
+		TransactionStateDiffs: transactionStateDiffs,
+		CandidateTxs:          candidateTxs,
+	}
+}
```
rodrodros: Return by value, why returning a pointer? 

An object that is created inside a function should be returned unless it is unsafe for it to be copy by value. Usually these are objects which hold concurrent structures such as mutexes. This is neither, do you think we should return by reference here?

core/version.go
```diff
@@ -9,6 +9,7 @@ import (
 var (
 	Ver0_13_2 = semver.MustParse("0.13.2")
 	Ver0_13_4 = semver.MustParse("0.13.4")
+	Ver0_14_0 = semver.MustParse("0.14.0")
```
rodrodros, line 12: @infrmtcs I saw something similar in one of your PRs signaling it here when the time comes of merging main with 0.15.0

rpc/v6/pending_data_wrapper.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/NethermindEth/juno/sync"
+)
+
+func (h *Handler) PendingData() (*core.PendingData, error) {
+	pending, err := h.syncReader.PendingData()
+	if err != nil {
+		return nil, err
+	}
+	// If pending network is polling pending block and running on < 0.14.0
+	if pending.Variant() == core.PendingBlockVariant {
```
rodrodros, line 17: Small nitpick: I don't think an enum is necessary here (referring to the variants). I think we can go away with a bool. Thinking that this variant will exist just during this transition period at most. It will never increase and will disappear eventually

rpc/v6/pending_data_wrapper.go
```diff
@@ -0,0 +1,87 @@
...
+		return pending, nil
+	} else {
+		// If pre_confirmed, network is polling pre_confirmed block and running on >= 0.14.0
+		latestHeader, err := h.bcReader.HeadsHeader()
+		if err != nil {
+			return nil, err
+		}
+
+		return emptyPendingForParent(latestHeader).AsPendingData(), nil
```
rodrodros: I am not very sure what we are returning here. Maybe I am missing something, when asking for pending in 0.14.0 shouldn't we return `latest`?

rpc/v6/pending_data_wrapper_test.go
```diff
@@ -0,0 +1,99 @@
...
+		)
+		pending, err := handler.PendingData()
+		require.NoError(t, err)
+		require.Equal(t, expectedPending, &sync.Pending{
+			Block:       pending.GetBlock(),
+			StateUpdate: pending.GetStateUpdate(),
+			NewClasses:  pending.GetNewClasses(),
+		})
+	})
+	t.Run("Returns empty pending block when starknet version >= 0.14.0", func(t *testing.T) {
```
rodrodros, line 49: We should return latest, meaning the last block included in the blockchain (if pre-confirmed will become block `n` we should return `n-1`.

rpc/v7/pending_data_wrapper.go
rodrodros, line 1: This file seems similar to v6 (and probably v8) why not just have one and every one pointing at it. I know we have a similar comment talking about this. 

If it is not possible (or it doesn't make sense because is too much effort) I am happy with the repetition

rpc/v8/pending_data_wrapper.go
```diff
@@ -0,0 +1,87 @@
+package rpcv8
```
rodrodros, line 1: Similar comment

rpc/v9/class.go
```diff
@@ -52,3 +55,104 @@ func adaptDeclaredClass(declaredClass json.RawMessage) (core.Class, error) {
...
+
+/****************************************************
+		Class Handlers
+*****************************************************/
+
+// Class gets the contract class definition in the given block associated with the given hash
+//
+// It follows the specification defined here:
+// https://github.com/starkware-libs/starknet-specs/blob/9377851884da5c81f757b6ae0ed47e84f9e7c058/api/starknet_api_openrpc.json#L410
+func (h *Handler) Class(id BlockID, classHash felt.Felt) (*rpcv6.Class, *jsonrpc.Error) {
```
rodrodros: `id` and `classHash` by reference

rpc/v9/events.go
```diff
@@ -50,3 +52,78 @@ func setEventFilterRange(filter blockchain.EventFilterer, from, to *BlockID, lat
...
+
+/****************************************************
+		Events Handlers
+*****************************************************/
+
+// Events gets the events matching a filter
+//
+// It follows the specification defined here:
+// https://github.com/starkware-libs/starknet-specs/blob/9377851884da5c81f757b6ae0ed47e84f9e7c058/api/starknet_api_openrpc.json#L813
+func (h *Handler) Events(args EventsArg) (*rpcv6.EventsChunk, *jsonrpc.Error) {
```
rodrodros: nit, should be `EventsArg` or `EventArgs` or `EventsArgs`?

rpc/v9/events.go
```diff
@@ -50,3 +52,78 @@ func setEventFilterRange(filter blockchain.EventFilterer, from, to *BlockID, lat
...
+			BlockNumber:     blockNumber,
+			BlockHash:       fEvent.BlockHash,
+			TransactionHash: fEvent.TransactionHash,
+			Event: &rpcv6.Event{
+				From: fEvent.From,
+				Keys: fEvent.Keys,
+				Data: fEvent.Data,
+			},
+		})
+	}
```
rodrodros: I believe int his case we can apply the indexing instead of the appending?

rpc/v9/events.go
```diff
@@ -50,3 +52,78 @@ func setEventFilterRange(filter blockchain.EventFilterer, from, to *BlockID, lat
...
+
+	cTokenStr := ""
+	if cToken != nil {
+		cTokenStr = cToken.String()
+	}
+	return &rpcv6.EventsChunk{Events: emittedEvents, ContinuationToken: cTokenStr}, nil
```
rodrodros: This is also something we can return by value. If it only affects RPC v9, then let's return by value here

rpc/v9/handlers_test.go
```diff
@@ -76,7 +76,7 @@ func TestThrottledVMError(t *testing.T) {
 	t.Run("trace", func(t *testing.T) {
 		blockHash := utils.HexToFelt(t, "0x0001")
 		header := &core.Header{
-			// hash is not set because it's pending block
+			// hash is not set because it's pre_confirmed block
```
rodrodros, line 79: 👍

rpc/v9/state_update.go
```diff
@@ -0,0 +1,113 @@
...
+
+/****************************************************
+		StateUpdate Handlers
+*****************************************************/
+
+// StateUpdate returns the state update identified by the given BlockID.
+//
+// It follows the specification defined here:
+// https://github.com/starkware-libs/starknet-specs/blob/9377851884da5c81f757b6ae0ed47e84f9e7c058/api/starknet_api_openrpc.json#L136
+func (h *Handler) StateUpdate(id BlockID) (*rpcv6.StateUpdate, *jsonrpc.Error) {
```
rodrodros: `id` by reference

rpc/v9/state_update.go
```diff
@@ -0,0 +1,113 @@
...
+			return nil, rpccore.ErrBlockNotFound
+		}
+		return nil, rpccore.ErrInternal.CloneWithData(err)
+	}
+
+	nonces := make([]rpcv6.Nonce, 0, len(update.StateDiff.Nonces))
+	for addr, nonce := range update.StateDiff.Nonces {
+		nonces = append(nonces, rpcv6.Nonce{ContractAddress: addr, Nonce: *nonce})
+	}
+
```
rodrodros: indexing

rpc/v9/state_update.go
```diff
@@ -0,0 +1,113 @@
...
+	storageDiffs := make([]rpcv6.StorageDiff, 0, len(update.StateDiff.StorageDiffs))
+	for addr, diffs := range update.StateDiff.StorageDiffs {
+		entries := make([]rpcv6.Entry, 0, len(diffs))
+		for key, value := range diffs {
+			entries = append(entries, rpcv6.Entry{
+				Key:   key,
+				Value: *value,
+			})
+		}
```
rodrodros: Indexing

rpc/v9/state_update.go
```diff
@@ -0,0 +1,113 @@
...
+		})
+	}
+
+	declaredClasses := make([]rpcv6.DeclaredClass, 0, len(update.StateDiff.DeclaredV1Classes))
+	for classHash, compiledClassHash := range update.StateDiff.DeclaredV1Classes {
+		declaredClasses = append(declaredClasses, rpcv6.DeclaredClass{
+			ClassHash:         classHash,
+			CompiledClassHash: *compiledClassHash,
+		})
+	}
```
rodrodros: indexing

rpc/v9/state_update.go
```diff
@@ -0,0 +1,113 @@
...
+
+	replacedClasses := make([]rpcv6.ReplacedClass, 0, len(update.StateDiff.ReplacedClasses))
+	for addr, classHash := range update.StateDiff.ReplacedClasses {
+		replacedClasses = append(replacedClasses, rpcv6.ReplacedClass{
+			ClassHash:       *classHash,
+			ContractAddress: addr,
+		})
+	}
```
rodrodros: indexing

rpc/v9/state_update.go
```diff
@@ -0,0 +1,113 @@
...
+		StateDiff: &rpcv6.StateDiff{
+			DeprecatedDeclaredClasses: update.StateDiff.DeclaredV0Classes,
+			DeclaredClasses:           declaredClasses,
+			ReplacedClasses:           replacedClasses,
+			Nonces:                    nonces,
+			StorageDiffs:              storageDiffs,
+			DeployedContracts:         deployedContracts,
+		},
+	}, nil
+}
```
rodrodros: Return by value

rpc/v9/state_update.go
```diff
@@ -0,0 +1,113 @@
...
+
+func (h *Handler) stateUpdateByID(id BlockID) (*core.StateUpdate, error) {
```
rodrodros: `id` by reference

rpc/v9/subscriptions.go
```diff
@@ -74,7 +75,7 @@ func (b *SubscriptionBlockID) UnmarshalJSON(data []byte) error {
 		return err
 	}
 
-	if blockID.IsPending() {
+	if blockID.IsPreConfirmed() {
 		return errors.New("subscription block id cannot be pending")
```
rodrodros: "subscription block id cannot be pre-confirmed"

rpc/v9/subscriptions.go
```diff
@@ -475,13 +497,29 @@ func (h *Handler) SubscribePendingTxs(ctx context.Context, getDetails *bool, sen
...
+					return nil
+				}
+				return err
 			}
-			return nil
+			switch v := pendingData.Variant(); v {
+			case core.PendingBlockVariant:
+				return h.onPendingBlock(id, w, getDetails, senderAddr, pendingData.GetBlock(), &lastParentHash, sentTxHashes)
+			case core.PreConfirmedBlockVariant:
+				return h.onPreConfirmedBlock(id, w, getDetails, senderAddr, pendingData.GetBlock(), &lastBlockNumber, sentTxHashes)
```
rodrodros: Both return statements are a bit long, can you refactor

rpc/v9/subscriptions.go
```diff
@@ -475,13 +497,29 @@ func (h *Handler) SubscribePendingTxs(ctx context.Context, getDetails *bool, sen
...
+
+			default:
+				panic(fmt.Errorf("unknown PendingDataVariant %v", v))
+			}
+		},
+		onPreConfirmed: func(ctx context.Context, id string, _ *subscription, preConfirmed *core.PreConfirmed) error {
+			return h.onPreConfirmedBlock(id, w, getDetails, senderAddr, preConfirmed.Block, &lastBlockNumber, sentTxHashes)
```
rodrodros: long line

rpc/v9/subscriptions.go
```diff
@@ -526,6 +564,42 @@ func (h *Handler) onPendingBlock(
 	return nil
 }
 
+// If getDetails is true, response will contain the transaction details.
+// If getDetails is false, response will only contain the transaction hashes.
+func (h *Handler) onPreConfirmedBlock(
+	id string,
+	w jsonrpc.Conn,
+	getDetails *bool,
```
rodrodros: Why a pointer to a bool instead of the actual value?

rpc/v9/transaction.go
```diff
@@ -113,15 +121,19 @@ func (es TxnExecutionStatus) MarshalText() ([]byte, error) {
 	}
 }
 
+// https://github.com/starkware-libs/starknet-specs/blob/9377851884da5c81f757b6ae0ed47e84f9e7c058/api/starknet_api_openrpc.json#L3134
 type TxnFinalityStatus uint8
 
 const (
-	TxnAcceptedOnL2 TxnFinalityStatus = iota + 3
+	TxnPreConfirmed TxnFinalityStatus = iota + 4
```
rodrodros, line 129: Why `iota + 4`?

rpc/v9/transaction.go
```diff
@@ -933,6 +965,12 @@ func adaptTransactionStatus(txStatus *starknet.TransactionStatus) (*TransactionS
 		status.Finality = TxnStatusAcceptedOnL2
 	case starknet.Received:
 		status.Finality = TxnStatusReceived
+		return &status, nil
+	case starknet.PreConfirmed:
+		status.Finality = TxnStatusPreConfirmed
+	case starknet.Candidate:
+		status.Finality = TxnStatusCandidate
+		return &status, nil
```
rodrodros: Return status by value, signaling it here. Even if this is not part of the code, I would like the RPC v9 to help standarize the code in the pass by reference and return by value (in the places it makes sense)

sync/pending.go
```diff
@@ -18,6 +18,46 @@ type Pending struct {
 	NewClasses  map[felt.Felt]core.Class
 }
 
+func NewPending(block *core.Block, stateUpdate *core.StateUpdate, newClasses map[felt.Felt]core.Class) *Pending {
+	return &Pending{
+		Block:       block,
+		StateUpdate: stateUpdate,
+		NewClasses:  newClasses,
+	}
+}
```
rodrodros: Return by value? If it's just a few lines of change, if it spans too much outside this PR then we leave it for later

sync/sync.go
```diff
@@ -331,8 +347,19 @@ func (s *Synchronizer) verifierTask(ctx context.Context, block *core.Block, stat
...
 
 					if head != nil {
-						if err := s.storeEmptyPending(head); err != nil {
-							s.log.Errorw("Failed to store empty pending block", "number", block.Number)
+						blockVer, err := core.ParseBlockVersion(head.ProtocolVersion)
+						if err == nil {
+							if blockVer.GreaterThanEqual(core.Ver0_14_0) {
+								if err := s.storeEmptyPreConfirmed(head); err != nil {
+									s.log.Errorw("Failed to store empty pre_confirmed block", "number", block.Number)
+								}
```
rodrodros, line 355: Can you help me understand this part of the code. What's the purpose of storing an empty pre-confirmed/pending block?

rpc/v9/events.go
EgeCaner: Not sure tho in RPCv6 it was `EventsArg` and I used it like that. IMO 'EventArgs' sounds better.

rpc/v9/transaction.go
EgeCaner: The following is TxnStatus enum, iota + 4 is used to map finality status to TxnStatus correctly with TxnStatus(finaliyStatus)
```go
const (
	TxnStatusReceived TxnStatus = iota + 1
	TxnStatusRejected
	TxnStatusCandidate
	TxnStatusPreConfirmed
	TxnStatusAcceptedOnL2
	TxnStatusAcceptedOnL1
)
```

sync/sync.go
EgeCaner: As far as I understood, on reorg it creates an empty pending state for new tip, and on each block it refreshes the pending data with new tip

rpc/v9/events.go
rodrodros: Let's have the updated naming for v9

rpc/v9/transaction.go
rodrodros: Got it, makes sense. Can we add a small comment explaining this

core/pending.go
EgeCaner: Addressed

core/pending.go
EgeCaner: Addressed

core/pending.go
EgeCaner: Addressed

rpc/v9/class.go
EgeCaner: Addressed

rpc/v9/events.go
EgeCaner: Addressed

rpc/v9/state_update.go
EgeCaner: adressed

rpc/v9/state_update.go
EgeCaner: addressed

rpc/v9/state_update.go
EgeCaner: addressed

rpc/v9/state_update.go
EgeCaner: addressed

rpc/v9/state_update.go
EgeCaner: addressed

rpc/v9/state_update.go
EgeCaner: addressed

rpc/v9/state_update.go
EgeCaner: addressed

rpc/v9/events.go
EgeCaner: Addressed

rpc/v9/events.go
EgeCaner: Addresed

rpc/v6/pending_data_wrapper.go
EgeCaner: At the moment we are returning, empty pending block for latest header. Im not sure this is the expected behaviour however I dont grasp what is the upside to serve data from latest block when blockID == pending, such as receipts, transactions, events.  Does not feel natural, serving empty pending block for latest know header with pending block semantics, feels more natural to me. Lets keep it like this and clarify the expected behavior ,if this is not the intended behavior, we can tweak pending data wrapper to return latest block instead

rpc/v6/pending_data_wrapper.go
EgeCaner: I agree that these variants are temporary and probably disappear soon. By boolean, you meant something like `IsPreconfirmed` or `IsPending`. I would prefer such indicator to be coupled with concrete type rather than having as field PendingData and do extra book keeping and to eliminate any inconsistency concern between flag and actual data.
```go
type PendingData struct {
 data Interface
 IsPending bool
}
```

Would you prefer replacing `Variant()` function with `IsPreConfirmed()` or `IsPending()` or both. Or having such boolean flag in `PendingData`

rpc/v9/subscriptions.go
EgeCaner: Switched to using value

rpc/v9/subscriptions.go
EgeCaner: addressed

rpc/v9/subscriptions.go
EgeCaner: addressed

sync/pending.go
EgeCaner: addressed

rpc/v9/transaction.go
EgeCaner: Addressed

rpc/v9/subscriptions.go
EgeCaner: addressed

rpc/v7/pending_data_wrapper.go
EgeCaner: We need to access to sync.Reader to access pending data, each handler has its own sync.Reader field. We can make [rpcv6, rpcv8] use the same pending data wrapper, but each function has to receive sync.Reader as parameter, since rpcv8.Handler does not have access to rpcv6.Handler. Or similarly we can have one PendingDataWrapper in rpccore or other shared space, which receives sync.Reader and RPC version. IMO interaction with PendingData is simpler this way, without arguments, each RPC has its own wrapper and does not need to pass sync.Reader or rpc version. sync.PendingData() can be tweaked to receive rpc version and respond based on that but not sure it is ideal to propogate rpc specific logic to sync.

clients/feeder/feeder.go
EgeCaner: refactored, extracted resolving Dir and QueryArg logic to its own function

adapters/sn2core/sn2core_test.go
EgeCaner: refactored

clients/feeder/feeder.go
rodrodros: Amazing!!!

rpc/v7/pending_data_wrapper.go
rodrodros: I agree with you that the current way. Thank you for the detailed explanation!

rpc/v6/pending_data_wrapper.go
rodrodros: Let's leave the enums, it was just a nitpick. I was thinking on `IsPreconfirmed` and it being false meant it was Pending but to be honest I don't think we should put time in this since I also like the enum approach

adapters/sn2core/sn2core.go
```diff
@@ -357,35 +357,49 @@ func AdaptCairo0Class(response *starknet.Cairo0Definition) (core.Class, error) {
...
+		return nil, err
+	}
+
+	return &core.StateUpdate{
+		BlockHash: response.BlockHash,
+		NewRoot:   response.NewRoot,
+		OldRoot:   response.OldRoot,
+		StateDiff: stateDiff,
+	}, nil
+}
```
rodrodros, line 371: Just a note that I think we should return by value here but haven't mention it because changes will be very extensive.

sync/sync.go
```diff
@@ -511,6 +571,36 @@ func (s *Synchronizer) pollPending(ctx context.Context, sem chan struct{}) {
 	}
 }
 
+func (s *Synchronizer) pollPreConfirmed(ctx context.Context, sem chan struct{}) {
+	if s.pendingPollInterval == time.Duration(0) {
+		return
+	}
```
rodrodros, line 579: Can you help me understand why does this check exist?

EgeCaner: Im not sure we need such check either. It was present in the polling pending logic so I copied from there. Seems like when Interval is 0. We stop polling pending data, default value for pending is 5 second atm. Preconfirmed block should use shorter interval, we can either introduce new config param or decrease the duration for pendingInterval. But as far as I understood for this value to be zero, it has to be set to zero intentionally, If we should allow pollingInterval to be 0 we can remove this check

sync/sync.go
```diff
@@ -409,7 +447,7 @@ func (s *Synchronizer) syncBlocks(syncCtx context.Context) {
 
 	go s.pollLatest(syncCtx, latestSem)
 	pendingSem := make(chan struct{}, 1)
-	go s.pollPending(syncCtx, pendingSem)
+	go s.pollPendingData(syncCtx, pendingSem)
```
rodrodros, line 450: This is a go-routine that just spawns two go-routines, I think it can be a normal call

sync/sync.go
```diff
@@ -482,17 +520,32 @@ func (s *Synchronizer) revertHead(forkBlock *core.Block) {
 	s.listener.OnReorg(head.Number)
 }
 
+func (s *Synchronizer) pollPendingData(ctx context.Context, sem chan struct{}) {
+	if s.pendingPollInterval == time.Duration(0) {
+		return
+	}
+
+	go s.pollPending(ctx, sem)
+	s.log.Infow("Detected block version 0.14.0; switching to polling mode for pre_confirmed blocks")
```
rodrodros, line 530: Aren't we always pending for pre-confirmed. I am a bit confused by this log. Perhaps it should be behind a conditional or am I missing something

sync/sync.go
```diff
@@ -482,17 +520,32 @@ func (s *Synchronizer) revertHead(forkBlock *core.Block) {
 	s.listener.OnReorg(head.Number)
 }
 
+func (s *Synchronizer) pollPendingData(ctx context.Context, sem chan struct{}) {
+	if s.pendingPollInterval == time.Duration(0) {
+		return
+	}
```
rodrodros, line 527: I think we can leave a `todo` here to investigate why this is here

sync/sync.go
rodrodros: Let's add a todo and check it later

sync/sync.go
EgeCaner: My bad this, I should have call these `poolPending` and `poolPreConfirmed` sequentially. polling should start with polling pending block and if the head or block we fetched by polling pending is version 0.14.0 block,. pollPending should stop polling and execution should continue from pollPreConfirmed, just pushed a fix

sync/sync.go
EgeCaner: My bad, this should spawn goroutine and `pollPendingData` should start polling Pending then when network upgraded to 0.14.0 it should switch polling preConfirmed, with the first received 0.14.0 block. Just pushed a fix

sync/sync.go
wojciechos: > Preconfirmed block should use shorter interval, we can either introduce new config param or decrease the duration for pendingInterval

regarding pendingInterval - let’s set default pendingInterval to 1 s - pretty sure Leo brought it up, and Pathfinder's already upgraded it to 1s

EgeCaner: I was about to introduce new config param as `preConfirmedPollInterval`. Should we do that or just set `pendingPollInteval` to lower number? Polling pending as frequent as we need poll pre_confirmed might be wasteful till network switches to pre_confirmed block. WDYT? @wojciechos

EgeCaner: Lets merge this, I will introduce the new config param in follow up PR

PR #2941, opened 2025-06-30
feat: rpcv9 l1 handler traces return revert reason

EgeCaner:
In starknet version `0.14.0`, reverted L1Handler transactions will be included in blocks

This PR add support to return revert reason for L1Handler traces for rpc version >= 9, and for rpc version < 9 returns default, L1Handler Trace when query hits reverted L1Handler transaction. This is due to previous rpc methods cannot describe revert reason.
 

Depends on #2938

### DIFF
--- a/rpc/v6/trace_test.go
+++ b/rpc/v6/trace_test.go
@@ -906,32 +906,60 @@ func TestAdaptVMTransactionTrace(t *testing.T) {
 	})
 
 	t.Run("successfully adapt L1_HANDLER tx from vm", func(t *testing.T) {
-		vmTrace := vm.TransactionTrace{
-			Type:                  vm.TxnL1Handler,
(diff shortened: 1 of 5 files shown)


core/pending.go
```diff
@@ -0,0 +1,112 @@
...
+	GetBlock() *Block
+	GetHeader() *Header
+	GetTransactions() []Transaction
+	GetStateUpdate() *StateUpdate
+	GetNewClasses() map[felt.Felt]Class
+	GetCandidateTransaction() []Transaction
+	Variant() PendingDataVariant
+}
+
+type PendingData struct {
```
infrmtcs: Why do we need a struct to wrap the interface?

EgeCaner: We store `PendingState` in `atomic.Pointer`, to be able to store interface inside `atomic.Pointer` I introduced this Wrapper struct

PR #2947, opened 2025-07-01
feat: Implement proposer component

infrmtcs:


consensus/proposer/proposer.go
```diff
@@ -2,103 +2,162 @@ package proposer
...
 	}
 }
 
-func (p *proposer) getDefaultBuildParams() builder.BuildParams {
-	return builder.BuildParams{
-		Builder:           *p.sequencerAddress,
+func (p *proposer[V, H]) Run(ctx context.Context) error {
+	p.init()
+
+	receiver := p.transactionReceiver
```
rodrodros, line 78: Nitpick: why `receiver` get's an alias?

consensus/proposer/proposer.go
```diff
@@ -2,103 +2,162 @@ package proposer
...
+		}
+		break
+	}
+
+	p.buildState.Store(buildState)
+}
+
+func (p *proposer[V, H]) getBuildParams() *builder.BuildParams {
+	return &builder.BuildParams{
+		Builder:           felt.Felt(p.nodeAddress),
```
rodrodros: This wouldn't need to be a reference return but seeing the rest of the code, make it return by value here will have no effect later on since those other types will hold references to it. Mainly, everything in core.Header. 

I think core.Header shouldn't actually hold that many references but it goes beyond the scope of this PR.

consensus/proposer/proposer.go
```diff
@@ -2,103 +2,162 @@ package proposer
...
+		select {
+		case <-ctx.Done():
+			return nil
+
+		case transactions := <-p.commitTrigger:
+			// Initialise the state back to pending block
+			p.init()
+			// Re-run the previous transactions if any
+			if err := p.rerunTransactions(transactions); err != nil {
+				p.log.Errorw("Fail to rerun transactions", "error", err)
```
rodrodros: nit: "Fail to re-run transactions"

consensus/proposer/proposer.go
```diff
@@ -108,118 +167,55 @@ func (p *proposer) getDefaultBuildParams() builder.BuildParams {
 	}
 }
 
-func (p *proposer) ProposalInit() (types.ProposalInit, error) {
-	err := p.buildState.ClearPending()
-	if err != nil {
-		return types.ProposalInit{}, err
+func (p *proposer[V, H]) rerunTransactions(transactions []core.Transaction) error {
```
rodrodros: I think `transaction` variable can have a better name, since this are transactions that are not meant to be re-run. Now, a better name is hard to find for me. Suggestions: `except`, `ignore`, `alreadyRunned`, `toDelete`. Open to others. 

I would also add a small description to the function even if it is meant for internal use

consensus/proposer/proposer.go
```diff
@@ -2,103 +2,162 @@ package proposer
...
-	builder          *builder.Builder
-	buildState       *builder.BuildState
-	mempool          *mempool.Pool
-	log              utils.Logger
+type proposer[V types.Hashable[H], H types.Hash] struct {
+	log                 utils.Logger
+	builder             *builder.Builder
+	proposalStore       *proposal.ProposalStore[H]
+	nodeAddress         starknet.Address
+	toValue             func(*felt.Felt) V
```
rodrodros: Why `toValue` needs to be a function that is passed on creation?

consensus/proposer/proposer.go
```diff
@@ -2,103 +2,162 @@ package proposer
...
+
+func (p *proposer[V, H]) Valid(value V) bool {
+	return p.proposalStore.Get(value.Hash()) != nil
+}
+
+func (p *proposer[V, H]) Value() V {
+	responseChannel := make(chan V)
+	p.valueTrigger <- responseChannel
+	return <-responseChannel
+}
```
rodrodros, line 142: Interesting design 👍

consensus/proposer/proposer.go
infrmtcs: When the node propose a block, it needs to stop processing more transactions, in order to calculate the block hash.
However, the proposer shouldn't stop accepting more transactions, in order to avoid blocking the transaction sender, e.g RPC endpoint.
`p.transactionReceiver` plays the role of a buffer. When we want to process transactions, we set `receiver` to it and select read from `receiver`. When we don't want to process transactions, we set `receiver` to nil, so the select will never go to that case.

consensus/proposer/proposer.go
infrmtcs: Because V is a type parameter unknown until the struct initialization.

builder/state.go
```diff
@@ -26,3 +30,60 @@ func (b *BuildState) ClearPending() error {
...
+
+// Assumptions we make to avoid deep copying some fields and types:
+// - This function is only called before running `Finish`
+// - *felt.Felt is immutable
+// - *GasPrice is immutable
+// - Signatures and EventsBloom are not set before `Finish` is called
+func (b *BuildState) Clone() BuildState {
+	return BuildState{
+		Pending:           clonePending(b.Pending),
+		RevealedBlockHash: b.RevealedBlockHash, // Safe to reuse an immutable value
```
rianhughes, line 42: Why do we use a reference type here and not `felt.Felt`?

rianhughes: Or why not just call `Clone()` on the felt?

infrmtcs: Mainly to avoid copying 32 bytes if not necessary

builder/state.go
```diff
@@ -26,3 +30,60 @@ func (b *BuildState) ClearPending() error {
...
+	}
+}
+
+func cloneBlock(block *core.Block) *core.Block {
+	return &core.Block{
+		Header:       utils.HeapPtr(*block.Header),
+		Transactions: slices.Clone(block.Transactions),
+		Receipts:     slices.Clone(block.Receipts),
+	}
+}
```
rianhughes, line 61: Since the core transaction and receipts tend to use `*felt.Felt` over `felt.Felt`, these clone operations will perform a shallow copy, right? Is this intended (I thought we wanted a deep copy)?

builder/state.go
```diff
@@ -26,3 +30,60 @@ func (b *BuildState) ClearPending() error {
 
 	return nil
 }
+
+// Assumptions we make to avoid deep copying some fields and types:
+// - This function is only called before running `Finish`
```
rianhughes, line 35: Can you explain why we need to call this before Finish? And why we need these assumptions?

builder/state.go
infrmtcs: Please see the assumption above on line 36:
```
// - *felt.Felt is immutable
```

builder/state.go
infrmtcs: Firstly, `Finish` writes the block hash, which is expected to be null in a pending block. This is the current behaviour of feeder gateway and some of the current RPC endpoints are relying on this.
Secondly, the 4th condition of these explain the fields that are set after calling `Finish`.

consensus/proposer/proposer.go
```diff
@@ -2,103 +2,163 @@ package proposer
...
+
+func (p *proposer[V, H]) init() {
+	var buildState *builder.BuildState
+	var err error
+	for {
+		buildParams := p.getBuildParams()
+		if buildState, err = p.builder.InitPendingBlock(&buildParams); err != nil {
+			p.log.Errorw("Fail to reinitialize proposer", "error", err)
+			time.Sleep(time.Second)
```
rianhughes: nit: can we use a const here and place it at the top? Eg `retryTimeForFailedInit` or something?

consensus/proposer/proposer.go
```diff
@@ -2,103 +2,163 @@ package proposer
...
+	builder             *builder.Builder
+	proposalStore       *proposal.ProposalStore[H]
+	nodeAddress         starknet.Address
+	toValue             func(*felt.Felt) V
+	buildState          atomic.Pointer[builder.BuildState]
+	transactions        map[H]mempool.BroadcastedTransaction
+	commitTrigger       chan []core.Transaction
+	valueTrigger        chan chan<- V
+	transactionReceiver chan []mempool.BroadcastedTransaction
 }
```
rianhughes: What are these? Can we place a small comment for context?

consensus/proposer/proposer.go
```diff
@@ -2,103 +2,163 @@ package proposer
...
-	mempool          *mempool.Pool
-	log              utils.Logger
+type proposer[V types.Hashable[H], H types.Hash] struct {
+	log                 utils.Logger
+	builder             *builder.Builder
+	proposalStore       *proposal.ProposalStore[H]
+	nodeAddress         starknet.Address
+	toValue             func(*felt.Felt) V
+	buildState          atomic.Pointer[builder.BuildState]
+	transactions        map[H]mempool.BroadcastedTransaction
```
rianhughes: Do you mind if we rename `transactions` to `seenTransactions`? The current name conflicts with the inputs to `receiveTransactions` which makes it a bit confusing

consensus/proposer/proposer.go
```diff
@@ -108,118 +168,55 @@ func (p *proposer) getDefaultBuildParams() builder.BuildParams {
...
-		BlockNumber:   pBlock.Number,
-		Builder:       *pBlock.SequencerAddress,
-		Timestamp:     uint64(time.Now().Unix()),
-		L2GasPriceFRI: *pBlock.L2GasPrice.PriceInFri,
-		L1GasPriceWEI: *pBlock.L1GasPriceETH,
-		EthToStrkRate: *new(felt.Felt).SetUint64(1), // TODO: Implement this properly
-		L1DAMode:      core.Blob,
-	}, true
+	return p.runTransactions(transactions)
 }
```
rianhughes:
```suggestion
func (p *proposer[V, H]) receiveTransactions(transactions []mempool.BroadcastedTransaction) error {
	transactions = slices.DeleteFunc(transactions, func(tx mempool.BroadcastedTransaction) bool {
		key := H(*tx.Transaction.Hash())
		if _, ok := p.seenTransactions[key]; ok {
			// we already processed this txn, so drop it
			return true
		}
		// register this txn in case we see it again
		p.seenTransactions[key] = tx
		return false
	})

	return p.runTransactions(transactions)
}
```

consensus/proposer/proposer.go
```diff
@@ -2,103 +2,163 @@ package proposer
...
-	buildState       *builder.BuildState
-	mempool          *mempool.Pool
-	log              utils.Logger
+type proposer[V types.Hashable[H], H types.Hash] struct {
+	log                 utils.Logger
+	builder             *builder.Builder
+	proposalStore       *proposal.ProposalStore[H]
+	nodeAddress         starknet.Address
+	toValue             func(*felt.Felt) V
+	buildState          atomic.Pointer[builder.BuildState]
```
rianhughes: Becuase 

```
func (p *proposer[V, H]) Pending() *sync.Pending {
    return p.buildState.Load().Pending
}
```

returns the pointer to the `buildState`, we can't guarantee that the caller won't modify it. This means we can get data races etc. Ideally we would make a deepcopy, but this is maybe a bit too heavy / error prone?? 

If we continue with the current approach we should really leave a comment indicating that the caller to this function must not modify the fields since it's not thread-safe

consensus/proposer/proposer.go
infrmtcs: The diff looks strange. Do you mean adding the comments or something else?

consensus/proposer/proposer.go
```diff
@@ -108,118 +168,55 @@ func (p *proposer) getDefaultBuildParams() builder.BuildParams {
...
+	buildState := p.buildState.Load().Clone()
+	err := p.builder.RunTxns(&buildState, transaction)
+	if err != nil {
+		return err
+	}
 
-	return out
+	p.buildState.Store(&buildState)
+	return nil
 }
```
rianhughes: Becuause `Clone()` doesn't perform a deep copy this operation isn't thread safe, right? Specifically, the fields in buildState that are modified in `RunTxns() may have data races.
`
At a quick galnce, I think these fields are the `Transactions`, `Receipts`, `TxnCount`, `EventCount`, `StateDiff` and `L2GasConsumed`

consensus/proposer/proposer.go
```diff
@@ -108,118 +168,55 @@ func (p *proposer) getDefaultBuildParams() builder.BuildParams {
...
 
-func (p *proposer) ProposalCommitment() (types.ProposalCommitment, error) {
-	buildResult, err := p.builder.Finish(p.buildState)
+func (p *proposer[V, H]) finish(responseChannel chan<- V) error {
+	buildState := p.buildState.Load().Clone()
+	buildResult, err := p.builder.Finish(&buildState)
 	if err != nil {
-		return types.ProposalCommitment{}, err
+		return err
 	}
```
rianhughes, line 232: Similar data-race question as in runTransactions

consensus/proposer/proposer.go
infrmtcs: `TxnCount`, `EventCount` and `L2GasConsumed` are int, they're copied

`Transactions`, `Receipts` and `StateDiff` are containers, they're cloned to new containers but the key value are immutable, so they're not deep copy

consensus/proposer/proposer.go
infrmtcs: Agree with this. This is exactly my assumption, so I'll add a comment here

consensus/proposer/proposer.go
rianhughes: The diff does look weird. Yeah I just meant to add these comments

```
// we already processed this txn, so drop it
			return true
```

and

```
// register this txn in case we see it again
		p.seenTransactions[key] = tx
```

I also had a comment about the `seenTransactions` field but I think that was addressed in the meantime.

rianhughes:
I think we should really keep track of the fact our pending block does not have proper read-write controls. I've created an issue to track this here https://github.com/NethermindEth/juno/issues/2957

builder/state.go
```diff
@@ -26,3 +30,60 @@ func (b *BuildState) ClearPending() error {
...
+// - *GasPrice is immutable
+// - Signatures and EventsBloom are not set before `Finish` is called
+func (b *BuildState) Clone() BuildState {
+	return BuildState{
+		Pending:           clonePending(b.Pending),
+		RevealedBlockHash: b.RevealedBlockHash, // Safe to reuse an immutable value
+		L2GasConsumed:     b.L2GasConsumed,     // Value, safe to shallow copy
+	}
+}
+
```
rianhughes, line 46: Given Clone() doesn't provide any read-write protections over the underlying type, and the current implementation isn't thread-safe, I think this should be removed. It also incurs overhead, and requires the caller to understand and respect the assumptions which may be problematic in practice. Imo, we can address the issue of making this threadsafe in a seperate issue.

rianhughes (approved):

PR #2950, opened 2025-07-02
feat: starknet 0.14.0 provide `RECEIVED` status from cache

EgeCaner:
As of Starknet version 0.14.0, the Feeder Gateway no longer returns the RECEIVED status.

This PR introduces an LRU cache with time-based expiration to support tracking the RECEIVED status.

Cache entries expire after their time-to-live (TTL) elapses and are considered invalid.
Expired entries are evicted first.
If all entries are still valid, eviction is based on the least-recently-used (LRU) policy.


rpc/v6/transaction.go
```diff
@@ -695,28 +711,25 @@ func (h *Handler) TransactionStatus(ctx context.Context, hash felt.Felt) (*Trans
...
-			status.Finality = TxnStatusAcceptedOnL1
-		case starknet.AcceptedOnL2:
-			status.Finality = TxnStatusAcceptedOnL2
-		case starknet.Received:
-			status.Finality = TxnStatusReceived
-		default:
-			return nil, rpccore.ErrTxnHashNotFound
+		if h.submittedTransactionsCache != nil {
+			switch txStatus.FinalityStatus {
+			case starknet.NotReceived:
```
rodrodros, line 716: I am thinking that the transaction cache only makes sense on rpc v9. If there will be two versions of Juno, then v6, v7 and v8 during 0.13.5 will have their behavior affected by it. Ideally, this behaviour only starts happening on the four rpcs after 0.14.0 🤔

EgeCaner: It should be fine I believe, lets go over what was the behavior before and what is the new behavior, please correct me if I am wrong.

Pre - starknet v0.14.0 Feeder GW was returning `RECEIVED`, `ACCEPTED_ON_L2` or `ACCEPTED_ON_L1` as Finality status

We are receiving and serving the `RECEIVED` status from Feeder Gateway.

In version starknet v0.14.0;
>Endpoint `get_transaction_status`
In v0.14.0, the finality status property of the `get_transaction_status` response can only be `ACCEPTED_ON_L2` or `ACCEPTED_ON_L1`. Calling `get_transaction_status` on a transaction that is still in the mempool and hasn’t entered a block yet will return an error `Transaction hash not found`. For a wider discussion on how the nodes will support status `RECEIVED` and other statuses, see section RPC 0.9 below.
 
Feeder Gateway no longer returns `RECEIVED` status, node serve `RECEIVED` status from cache.

We query the submitted transaction cache in the following case:
- Transaction is not found on DB
- Transaction is not found on Pending Data
- Then we query the FGW, if FGW return `NOT_RECEIVED` then we will query the cache, and respond based on local view.

While FGW still provides the `RECEIVED` status we will use the response, and once it stops serving the `RECEIVED` status we will fallback into cache. Only inconsistency I would expect is the following;
-  Pre 0.14.0 when FGW returns `RECEIVED` status, our node relays user tx to GW, added to cache, then upon querying the status, FGW returns `NOT_RECEIVED` somehow, but we will respond with `RECEIVED` if txs is still in the cache. Im not sure the likelyhood of this case, where GW returns success for `add_Transaction` then respond with `NOT_RECEIVED` when we it should be responding with `RECEIVED`.

rodrodros: >  * Pre 0.14.0 when FGW returns RECEIVED status, our node relays user tx to GW, added to cache, then upon querying the status, FGW returns NOT_RECEIVED somehow, but we will respond with RECEIVED if txs is still in the cache. Im not sure the likelyhood of this case, where GW returns success for add_Transaction then respond with NOT_RECEIVED when we it should be responding with RECEIVED.

I am worried on that the Sequencer not actually receiving the transaction and we saying it was received. When there is high workload this can happen, although the risk is low. On further thought I am ok having this and I don't think they will be any issues

EgeCaner: I understand the concern, this can happen in high load. We add the tx hash to cache if only sequencer returns with success. But after sequencer returning success, it might drop the TX from mempool which makes `RECEIVED` status we returned incorrect. Post 0.14.0 it is okay since we wont have a way to ensure tx is in the mempool, but on < 0.14.0, if sequencer says `NOT_RECEIVED` it is should be not received and and `RECEIVED` status we serve will be incorrect for `entryTTL`

rodrodros: Thanks for the detailed write-up as well!

cmd/juno/juno_test.go
```diff
@@ -147,7 +84,40 @@ func TestConfigPrecedence(t *testing.T) {
 				"--cn-unverifiable-range", "0,10",
 				"--cn-core-contract-address", "0xc662c410C0ECf747543f5bA90660f6ABeBD9C8c4",
 			},
-			expectedConfig: &expectedConfig1,
+			expectedConfig: &node.Config{ //nolint:dupl
```
rianhughes: why re-introduce `//nolint:dupl`? I think previous version avoided this with `expectedConfig1` and `expectedConfig2`

cmd/juno/juno.go
```diff
@@ -37,104 +37,108 @@ Juno is a Go implementation of a Starknet full-node client created by Nethermind
...
+	defaultGwTimeout                          = "5s"
+	defaultCorsEnable                         = false
+	defaultVersionedConstantsFile             = ""
+	defaultPluginPath                         = ""
+	defaultSeqEn                              = false
+	defaultSeqBlockTime                       = 60
+	defaultSeqGenesisFile                     = ""
+	defaultSeqDisableFees                     = false
+	defaultHTTPUpdatePort                     = 0
+	defaultSubmittedTransactionsCacheSize     = 1024
```
rianhughes: The max TPS every recorded on Starknet was 992 (according to Voyager). I wonder if we should bump this to 2048 just to prevent potential issues if/when we see some very high load in the future. What do you think? It's just a thought, I haven't thought too deeply about it

rpc/v6/transaction.go
```diff
@@ -677,6 +691,8 @@ func (h *Handler) pushToFeederGateway(ctx context.Context, tx BroadcastedTransac
 	}, nil
 }
 
+var errTransactionNotFound = errors.New("transaction not found")
```
rianhughes: nit: variables like this should go at the top of the file

rianhughes (commented):
Up to my small comments the PR LGTM

cmd/juno/juno_test.go
EgeCaner: Rebase mistake, I will correct it

cmd/juno/juno.go
EgeCaner: Very good point! I agree with you. I set this value cuz it was the default capacity our mempool, but we can use higher value here

EgeCaner: default size is set to 10k

rianhughes (approved):

PR #2951, opened 2025-07-02
feat: 0.14.0 gateway url split

EgeCaner:
Two distinct URLs for GW and FGW

   - feeder will be at feeder.{chain_name}.starknet.io/feeder_gateway
   - gateway will be at {chain_name}.starknet.io/gateway

   > While the chain is on v0.13.5, both URLs will point to their respective 0.13.5 locations so that users can upgrade their nodes to a 0.14.0-compatible version while the chain hasn’t upgraded yet.


utils/network.go
```diff
@@ -44,7 +44,7 @@ var (
 	// The docs states the addresses for each network: https://docs.starknet.io/tools/important-addresses/
 	Mainnet = Network{
 		Name:                "mainnet",
-		FeederURL:           "https://alpha-mainnet.starknet.io/feeder_gateway/",
+		FeederURL:           "https://feeder.alpha-mainnet.starknet.io/feeder_gateway/",
```
rodrodros, line 47: This won't break mainnet right?

rodrigo-pino (approved):

utils/network.go
EgeCaner: Both previous and new urls are responding at the moment

rianhughes (approved):

PR #2952, opened 2025-07-02
feat: 0.14.0 preconfirmed block polling

EgeCaner:
- PreConfirmedBlock polling mechanism which automatically switches to fetching pending blocks to preconfirmed block as network transitioned to 0.14.0
  - Transition logic works as following;
     -  Node start by polling pending block and if it encounters block with version `0.14.0` or higher, then switches to polling pre_confirmed blocks. 
     - Pending and PreConfirmed types are abstracted as PendingData.
     - RPCs acts different based on the concrete type of the pending data
- Intoduces abstractions over pending data for backward compatibility.
- Introduces wrappers around pending data to handle behaviour of previous RPCs when network is at 0.14.0
- Introduces `PreConfirmedBlockPollInterval` config parameter

Depends on #2954


core/pending.go
```diff
@@ -4,6 +4,61 @@ import (
...
+	GetTransactions() []Transaction
+	GetStateUpdate() *StateUpdate
+	GetNewClasses() map[felt.Felt]Class
+	GetCandidateTransaction() []Transaction
+	Variant() PendingDataVariant
+}
+
+type PendingData struct {
+	data PendingDataInterface
+}
```
rodrodros: Could you give me some context on why are we doing this design, a `type T = interface I`, why not just remove the middle layer and replace wherever you have type `T` interface `I`?'

Where `T = PendingData` and `I = PendingDataInterface`?

core/pending.go
```diff
@@ -22,3 +77,35 @@ func NewPreConfirmed(block *Block, stateUpdate *StateUpdate, transactionStateDif
 		CandidateTxs:          candidateTxs,
 	}
 }
+
+func (p *PreConfirmed) AsPendingData() PendingData {
+	return PendingData{data: p}
+}
```
rodrodros: Related with this 👀

node/node_test.go
```diff
@@ -74,7 +74,7 @@ func TestNetworkVerificationOnNonEmptyDB(t *testing.T) {
 			chain := blockchain.New(database, &network)
 			ctx, cancel := context.WithCancel(t.Context())
 			dataSource := sync.NewFeederGatewayDataSource(chain, adaptfeeder.New(feeder.NewTestClient(t, &network)))
-			syncer := sync.New(chain, dataSource, log, 0, false, database).WithListener(&sync.SelectiveListener{OnSyncStepDoneCb: func(op string, _ uint64, _ time.Duration) {
+			syncer := sync.New(chain, dataSource, log, 0, 0, false, database).WithListener(&sync.SelectiveListener{OnSyncStepDoneCb: func(op string, _ uint64, _ time.Duration) {
```
rodrodros: This is a bit of a long line (also I am aware that it is old code), could we refactor it

core/pending.go
EgeCaner: Only reason why I wrapped the interface is, we are storing the `PendingData` in `atomic.Pointer`. Atomic pointer does not work with interfaces, thus I wrapped the interface to have concrete type.

infrmtcs: Actually `atomic.Pointer` can work with interface, it's just not straightforward:
```
type I interface{}

type T struct {
	x int
}

func main() {
	ptr := atomic.Pointer[I]{}
	ptr.Store(utils.HeapPtr[I](T{x: 1}))
}
```

I still think we should drop this struct type and use interface type directly.

rpc/v8/pending_data_wrapper.go
```diff
@@ -0,0 +1,90 @@
...
+
+func (h *Handler) PendingData() (*core.PendingData, error) {
+	pending, err := h.syncReader.PendingData()
+	if err != nil && !errors.Is(err, sync.ErrPendingBlockNotFound) {
+		return nil, err
+	}
+	// If pending network is polling pending block and running on < 0.14.0
+	if err == nil && pending.Variant() == core.PendingBlockVariant {
+		return pending, nil
+	}
```
infrmtcs, line 20: It's unclear to me what we expect in the `sync.ErrPendingBlockNotFound` case. I think maybe we can do this instead:
```suggestion
	if err != nil {
	        if errors.Is(err, sync.ErrPendingBlockNotFound) {
	        	...
	        } else {
	        	...
	        }
	}
	// If pending network is polling pending block and running on < 0.14.0
	if pending.Variant() == core.PendingBlockVariant {
		return pending, nil
	}
```

infrmtcs: If err is non nil, I think the returned `pending` is expected to be zero, and should be undefined behaviour, so we shouldn't call `pending.Variant()`

rpc/v6/block_test.go
```diff
@@ -612,38 +703,71 @@ func TestBlockWithReceipts(t *testing.T) {
...
-		header := resp.BlockHeader
-
-		var txsWithReceipt []rpc.TransactionWithReceipt
+		var expectedTxsWithReceipt []rpc.TransactionWithReceipt
 		for i, tx := range block0.Transactions {
 			receipt := block0.Receipts[i]
 			adaptedTx := rpc.AdaptTransaction(tx)
 
-			txsWithReceipt = append(txsWithReceipt, rpc.TransactionWithReceipt{
+			expectedTxsWithReceipt = append(expectedTxsWithReceipt, rpc.TransactionWithReceipt{
```
infrmtcs:
```suggestion
			expectedTxsWithReceipt[i] = rpc.TransactionWithReceipt{
```

sync/sync.go
```diff
@@ -122,14 +134,17 @@ type Synchronizer struct {
 	newHeads            *feed.Feed[*core.Block]
 	reorgFeed           *feed.Feed[*ReorgBlockRange]
 	pendingFeed         *feed.Feed[*core.Block]
+	preConfirmedFeed    *feed.Feed[*core.PreConfirmed]
```
infrmtcs: Why don't we share the same feed with `pendingFeed`? I'm understanding that preconfirmed block will supersede pending block, and we're having a common interface shared between them anyway.

EgeCaner: I initially preferred to have separate feeds because when network is at `0.14.0` RPC versions < 9, we should stop serving pending data in subscription since there will be no more `pending` blocks thus no data to serve from it. Having separate feeds was straightforward for this cause, once network switches to fetching `pre_confirmed` block, we wont be sending entries to `PendingFeed` anymore. If I unify the feeds, I might have to additional checks for backwards compatibility, or I might just let the program run as before and rely on it will end up seeing empty pending data(might end up doing redundant work since we know response will always empty), never triggering becomes a shortcut in this sense. However at the moment, Im trying to clarify, if we should serve `pendingTxs` from `pre_confirmed` block on `starknet_subscribeTransactionStatus` for RPC versions < 9. If so Im gonna need to bring `preConfirmedFeed` to previous RPCs as well, then it might make lot of sense to unify the feed and we probably gonna end up here.

rpc/v8/pending_data_wrapper.go
EgeCaner: I will try to make the code more clear. Behavour here should be:
- If there is valid `Pending` block, use it directly
- If `PendingData`  doesn't exist or not valid anymore (`sync.ErrPendingBlockNotFound`)  or variant is `PreConfirmed` on rpc < 9, we return empty pending block, derived from the latest.
- If we get unexpected error we propagate the error.

So essentially there is 3 branches: 
- case where we got unexpected error
- case where we got valid `Pending` block
- case where we got either `sync.ErrPendingBlockNotFound` or valid `Preconfirmed`.

core/pending.go
EgeCaner: I will try as you suggested @infrmtcs

sync/sync.go
infrmtcs: I was wondering this because we're sharing `pendingData` here for both pending and preconfirmed block, but have separate feeds for each of them.

core/pending.go
```diff
@@ -4,6 +4,22 @@ import (
...
 )
 
+type PendingDataVariant uint8
+
+const (
+	PendingBlockVariant PendingDataVariant = iota + 1
+	PreConfirmedBlockVariant
+)
+
+type PendingDataInterface interface {
```
infrmtcs:
```suggestion
type PendingData interface {
```

sync/sync.go
infrmtcs: I think what you said makes sense, it's just I think that by introducing `core.PendingDataInterface` and logic around it in `pending_data_wrapper.go`, I'm understanding that we've already paid the cost of backward compatibility here. That's why I think we can be more consistent here, and either approach is equally fine to me:
- Separate pending and preconfirmed logic, separate `atomic.Pointer` and `feed.Feed`.
- Unify them with an interface, unify the atomic pointers, unify the feeds.

I can be terribly wrong because I may miss something here, so let me know if it's the case.

.github/workflows/starknet-go-tests.yml
```diff
@@ -61,4 +61,8 @@ jobs:
       - name: Run RPC tests
         run: |
           cd rpc
-          go test -timeout 1200s -v -env testnet .
\ No newline at end of file
+          # `TestSubscribePendingTransactions` is ignored for starknet version > 0.14.0.
+          #  In starknet version 0.14.0 `pending` block is replaced by 'pre_confirmed' block, 
+          #  thus there will be no more pending transcations with `pending` semantics.
+          TESTS_TO_SKIP="TestSubscribePendingTransactions"
```
infrmtcs, line 67: Should use `env` and `working-directory` fields for step instead.

infrmtcs: Moved this comment to the PR fixing this.

sync/sync.go
```diff
@@ -290,8 +307,19 @@ func (s *Synchronizer) verifierTask(ctx context.Context, block *core.Block, stat
...
+					if err := s.storeEmptyPreConfirmed(block.Header); err != nil {
+						s.log.Errorw("Failed to store empty pre_confirmed block", "number", block.Number)
+					}
+				} else {
+					if err := s.storeEmptyPending(block.Header); err != nil {
+						s.log.Errorw("Failed to store empty pending block", "number", block.Number)
+					}
+				}
+			} else {
+				s.log.Errorw("Failed to parse block version", "err", err)
```
infrmtcs: These lines are duplicate with the above. Should we have a common function for them?

sync/sync.go
```diff
@@ -458,6 +504,36 @@ func (s *Synchronizer) pollPending(ctx context.Context, sem chan struct{}) {
 	}
 }
 
+func (s *Synchronizer) pollPreConfirmed(ctx context.Context, sem chan struct{}) {
+	if s.preConfirmedPollInterval == time.Duration(0) {
+		return
+	}
```
infrmtcs: I think this is dangerous. This may silently stop working without notifying users.

infrmtcs: Yeah, probably we should fix in the `pollPending` too, probably out of scope of this PR.

sync/sync.go
```diff
@@ -429,17 +457,28 @@ func (s *Synchronizer) revertHead(forkBlock *core.Block) {
 	s.listener.OnReorg(head.Number)
 }
 
+func (s *Synchronizer) pollPendingData(ctx context.Context, sem chan struct{}) {
+	s.pollPending(ctx, sem)
```
infrmtcs, line 439: I think we should return error from `pollPending` and check if it's non-nil. Otherwise `pollPreconfirmed` may starts even `pollPending` stops because of error.

sync/sync.go
```diff
@@ -449,6 +488,13 @@ func (s *Synchronizer) pollPending(ctx context.Context, sem chan struct{}) {
 					}()
 					err := s.fetchAndStorePending(ctx)
 					if err != nil {
+						if errors.Is(err, ErrMustSwitchPollingPreConfirmed) {
+							select {
+							case switchCh <- struct{}{}:
```
infrmtcs: Why don't we just simply return here?

sync/sync.go
EgeCaner: The following is the usage description for the parameter 

> "Sets how frequently pending block will be updated (0s will disable fetching of pending block)."

Default values are non-zero and one should explicitly set the polling intervals to 0 to disable pending/pre_confirmed polling. It is a feature I guess, not sure who would want to disable it tho. Would you recommend removing this check or adding a log here to at least inform the users? or any other recommendation

sync/sync.go
EgeCaner: Inside `pollPending`, on any error, we log the error and continue fetching, except `ErrMustSwitchPollingPreConfirmed`. Is it possible for `pollPending` to stop with unexpected error and continue to `pollPreConfirmed`?

sync/sync.go
EgeCaner: This is inside another goroutine, returning in the goroutine alone does not return from the outer loop. To terminate the outer loop I used channel to signal. Let me know if I am mistaken or if this can be handled better

infrmtcs: Ah ok, I missed that part. I think a clearer way is use `WithCancel` on the original ctx, use the new ctx and and call cancel here.

sync/sync.go
infrmtcs: I think adding a log is great! User can know it's not syncing because of this parameter, and it also makes the code easier to read.

EgeCaner: There is a case where Pending Interval is 0 and PreConfirmed Interval is non-zero, when network is not yet at 0.14.0. This will result in attempting to poll pre_confirmed blocks and consistently getting errors. This should be a rare case tho. I wonder you opinion, if should we rely on users setting parameters correctly, or should we handle this case internally?

infrmtcs: This is a tricky question. I think there can be several approaches to handle this:
1. If we merge the pending and preconfirmed interval flag, then it's pretty simple, either we poll both or none.
2. Another way is to merge these 2 polling functions and also 2 `fetchAndStore...` functions, because structurally they're the same, the only difference is how to fetch the pending/preconfirmed block. Then we can move the interval check to after fetching head, and decide whether to fetch based on the argument corresponding to whether the head is pending or preconfirmed.

EgeCaner: Merging the flags seems like the simpler way. I prefer to keep functions separate in order to easily remove the codes related to pending when we no longer require to support it.

sync/sync.go
EgeCaner: Addressed as suggested

sync/sync.go
rodrodros: I don't think not polling on 0 is an error. It is true that miss use of the flag might cause issues in Juno.

The idea of having two flags were discussed previously since it meant we could have two flags name, each with their dedicated functionality and would make updating the node easier. Also, with the current 30s blocks the polling time doesn't need to be as frequent as with the 6s, I am not sure if we should put both in one. 

Also, this problem would only exist during this time period. Why don't we disable syncing is either is 0 (or if pending is 0)

EgeCaner: Updated as if either is 0 we disable fetching pending data.

sync/sync.go
```diff
@@ -559,44 +653,95 @@ func (s *Synchronizer) StorePending(p *Pending) error {
...
+		// pre_confirmed
+		expectedOldRoot := &felt.Zero
+		expectedBlockNumber := uint64(0)
+		if head, err := s.blockchain.HeadsHeader(); err == nil {
+			expectedOldRoot = head.GlobalStateRoot
+			expectedBlockNumber = head.Number + 1
+		}
+
+		if p.GetStateUpdate().OldRoot.Equal(expectedOldRoot) &&
+			p.GetBlock().Number == expectedBlockNumber {
```
infrmtcs, line 715: Why don't we also compare parent hash?

sync/sync.go
```diff
@@ -559,44 +653,95 @@ func (s *Synchronizer) StorePending(p *Pending) error {
...
+		return err
 	}
-	if p.Block.ParentHash.Equal(expectedParentHash) {
-		return p, nil
+
+	h, err := s.blockchain.HeadsHeader()
+	if err != nil && !errors.Is(err, db.ErrKeyNotFound) {
+		return err
+	} else if err == nil {
+		p.StateUpdate.OldRoot = h.GlobalStateRoot
```
infrmtcs: Why do we need to update this?

infrmtcs (approved):

sync/sync.go
EgeCaner: `PreConfirmed` block does not have `parentHash` and not required by spec thus we didnt assigned one. But has `blockNumber` and `oldRoot` is required by the spec [ref](https://github.com/starkware-libs/starknet-specs/blob/9264a4eec76ccb7cbc24624bb13d243518cda6e6/api/starknet_api_openrpc.json#L1406) for oldRoot. [Ref](https://github.com/starkware-libs/starknet-specs/blob/9264a4eec76ccb7cbc24624bb13d243518cda6e6/api/starknet_api_openrpc.json#L1627) for PreConfirmed header

sync/sync.go
EgeCaner: `OldRoot` has to be set from local view, we do not receive this from FGW. But spec requires us to have one. We have access to this information in sync package, but along the execution other components does not have access to state. I can either pass OldRoot as parameter to `PreConfirmedBlockByNumber(...)` to encapsulate the logic or keep as is but adding a comment explaining why

EgeCaner: Looks like new FeederDataSource also has access to blockchain state, I moved the assignment inside DataSource

rodrigo-pino (approved):

sync/sync.go
```diff
@@ -511,12 +574,43 @@ func (s *Synchronizer) fetchAndStorePending(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
-	pending.Block.Number = head.Number + 1
```
EgeCaner, line 514: @rodrigo-pino this might be related to issues with staking tool

PR #2953, opened 2025-07-02
Rianhughes/tendermint catchup p2p sync

rianhughes:
This PR implements (what seems to be) a very simple and effective sync mechanism for the consensus protocol.

The main idea is that the p2p sync process runs in the background, and notifies the consensus algorithm
1. whenever it is behind the chain (so that we can stop consensus - we shouldn't waste resources processing outdated heights)
2. has committed a block to the chain (so that the state machine knows which height to start from).


infrmtcs:
Close because we take the other approach.

PR #2954, opened 2025-07-02
feat: 0.14.0 preconfirmed block

EgeCaner:
This PR introduces 
- Preconfirmed Block definition
- Feeder client function to fetch preconfirmed blocks
- Adapters for preconfirmed block

### DIFF
--- a/adapters/sn2core/sn2core.go
+++ b/adapters/sn2core/sn2core.go
@@ -357,26 +357,40 @@ func AdaptCairo0Class(response *starknet.Cairo0Definition) (core.Class, error) {
 }
 
 func AdaptStateUpdate(response *starknet.StateUpdate) (*core.StateUpdate, error) {
-	stateDiff := new(core.StateDiff)
-	stateDiff.DeclaredV0Classes = response.StateDiff.OldDeclaredContracts
+	stateDiff, err := AdaptStateDiff(&response.StateDiff)
+	if err != nil {
+		return nil, err
+	}
 
-	stateDiff.DeclaredV1Classes = make(map[felt.Felt]*felt.Felt, len(response.StateDiff.DeclaredClasses))
-	for _, declaredV1Class := range response.StateDiff.DeclaredClasses {
+	return &core.StateUpdate{
+		BlockHash: response.BlockHash,
+		NewRoot:   response.NewRoot,
+		OldRoot:   response.OldRoot,
+		StateDiff: stateDiff,
+	}, nil
+}
+
+func AdaptStateDiff(response *starknet.StateDiff) (*core.StateDiff, error) {
+	stateDiff := new(core.StateDiff)
+	stateDiff.DeclaredV0Classes = response.OldDeclaredContracts
+
+	stateDiff.DeclaredV1Classes = make(map[felt.Felt]*felt.Felt, len(response.DeclaredClasses))
+	for _, declaredV1Class := range response.DeclaredClasses {
 		stateDiff.DeclaredV1Classes[*declaredV1Class.ClassHash] = declaredV1Class.CompiledClassHash
 	}
 
-	stateDiff.ReplacedClasses = make(map[felt.Felt]*felt.Felt, len(response.StateDiff.ReplacedClasses))
-	for _, replacedClass := range response.StateDiff.ReplacedClasses {
+	stateDiff.ReplacedClasses = make(map[felt.Felt]*felt.Felt, len(response.ReplacedClasses))
+	for _, replacedClass := range response.ReplacedClasses {
 		stateDiff.ReplacedClasses[*replacedClass.Address] = replacedClass.ClassHash
 	}
 
-	stateDiff.DeployedContracts = make(map[felt.Felt]*felt.Felt, len(response.StateDiff.DeployedContracts))
-	for _, deployedContract := range response.StateDiff.DeployedContracts {
+	stateDiff.DeployedContracts = make(map[felt.Felt]*felt.Felt, len(response.DeployedContracts))
+	for _, deployedContract := range response.DeployedContracts {
 		stateDiff.DeployedContracts[*deployedContract.Address] = deployedContract.ClassHash
 	}
 
-	stateDiff.Nonces = make(map[felt.Felt]*felt.Felt, len(response.StateDiff.Nonces))
-	for addrStr, nonce := range response.StateDiff.Nonces {
+	stateDiff.Nonces = make(map[felt.Felt]*felt.Felt, len(response.Nonces))
+	for addrStr, nonce := range response.Nonces {
 		addr, err := new(felt.Felt).SetString(addrStr)
 		if err != nil {
 			return nil, err
@@ -384,8 +398,8 @@ func AdaptStateUpdate(response *starknet.StateUpdate) (*core.StateUpdate, error)
 		stateDiff.Nonces[*addr] = nonce
 	}
 
-	stateDiff.StorageDiffs = make(map[felt.Felt]map[felt.Felt]*felt.Felt, len(response.StateDiff.StorageDiffs))
-	for addrStr, diffs := range response.StateDiff.StorageDiffs {
+	stateDiff.StorageDiffs = make(map[felt.Felt]map[felt.Felt]*felt.Felt, len(response.StorageDiffs))
+	for addrStr, diffs := range response.StorageDiffs {
 		addr, err := new(felt.Felt).SetString(addrStr)
 		if err != nil {
 			return nil, err
@@ -397,12 +411,96 @@ func AdaptStateUpdate(response *starknet.StateUpdate) (*core.StateUpdate, error)
 		}
 	}
 
-	return &core.StateUpdate{
-		BlockHash: response.BlockHash,
-		NewRoot:   response.NewRoot,
-		OldRoot:   response.OldRoot,
-		StateDiff: stateDiff,
-	}, nil
+	return stateDiff, nil
+}
+
+func AdaptPreConfirmedBlock(response *starknet.PreConfirmedBlock, number uint64) (core.PreConfirmed, error) {
+	if response == nil {
+		return core.PreConfirmed{}, errors.New("nil preconfirmed block")
+	}
+
+	var adaptedStateDiff *core.StateDiff
+	var err error
+
+	txStateDiffs := make([]*core.StateDiff, 0, len(response.TransactionStateDiffs))
+	for _, stateDiff := range response.TransactionStateDiffs {
+		if stateDiff == nil {
+			break
+		}
+
+		if adaptedStateDiff, err = AdaptStateDiff(stateDiff); err != nil {
+			return core.PreConfirmed{}, err
+		}
+		txStateDiffs = append(txStateDiffs, adaptedStateDiff)
+	}
+
+	lastPreconfirmedIndex := len(txStateDiffs) - 1
+
+	txns := make([]core.Transaction, lastPreconfirmedIndex+1)
+	for i, txn := range response.Transactions[:lastPreconfirmedIndex+1] {
+		adaptedTxn, err := AdaptTransaction(txn)
+		if err != nil {
+			return core.PreConfirmed{}, err
+		}
+		txns[i] = adaptedTxn
+	}
+
+	candidateTxs := make([]core.Transaction, len(response.Transactions)-lastPreconfirmedIndex-1)
+	for i, txn := range response.Transactions[lastPreconfirmedIndex+1:] {
+		adaptedTxn, err := AdaptTransaction(txn)
+		if err != nil {
+			return core.PreConfirmed{}, err
+		}
+		candidateTxs[i] = adaptedTxn
+	}
+
+	receipts := make([]*core.TransactionReceipt, lastPreconfirmedIndex+1)
+	eventCount := uint64(0)
+	for i, receipt := range response.Receipts[:lastPreconfirmedIndex+1] {
+		receipts[i] = AdaptTransactionReceipt(receipt)
+		eventCount += uint64(len(receipt.Events))
+	}
+
+	// Squash per-tx state updates
+	stateDiff := core.EmptyStateDiff()
+	for _, txStateDiff := range txStateDiffs {
+		stateDiff.Merge(txStateDiff)
+	}
+
+	stateUpdate := core.StateUpdate{
+		BlockHash: nil,
+		NewRoot:   nil,
+		// Must be set to previous global state root, when have access to latest header
+		OldRoot:   nil,
+		StateDiff: &stateDiff,
+	}
+
+	adaptedBlock := &core.Block{
+		// https://github.com/starkware-libs/starknet-specs/blob/9377851884da5c81f757b6ae0ed47e84f9e7c058/api/starknet_api_openrpc.json#L1636
+		Header: &core.Header{
+			Number:           number,
+			Timestamp:        response.Timestamp,
+			ProtocolVersion:  response.Version,
+			SequencerAddress: response.SequencerAddress,
+			L1GasPriceETH:    response.L1GasPriceETH(),
+			L1GasPriceSTRK:   response.L1GasPriceSTRK(),
+			L1DAMode:         core.L1DAMode(response.L1DAMode),
+			L1DataGasPrice:   (*core.GasPrice)(response.L1DataGasPrice),
+			L2GasPrice:       (*core.GasPrice)(response.L2GasPrice),
+			// Not required in spec but useful
+			TransactionCount: uint64(len(txns)),
+			EventCount:       eventCount,
+			EventsBloom:      core.EventsBloom(receipts),
+			// Following fields are nil for pre_confirmed block
+			Hash:            nil,
+			ParentHash:      nil,
+			GlobalStateRoot: nil,
+			Signatures:      nil,
+		},
+		Transactions: txns,
+		Receipts:     receipts,
+	}
+	return core.NewPreConfirmed(adaptedBlock, &stateUpdate, txStateDiffs, candidateTxs), nil
 }
 
 func safeFeltToUint64(f *felt.Felt) uint64 {
--- a/adapters/sn2core/sn2core_test.go
+++ b/adapters/sn2core/sn2core_test.go
@@ -623,3 +623,148 @@ func TestAdaptCompiledClass(t *testing.T) {
 	require.NoError(t, err)
 	assert.Nil(t, result)
 }
+
+func TestAdaptPreConfirmed(t *testing.T) {
+	n := utils.HeapPtr(utils.SepoliaIntegration)
+	client := feeder.NewTestClient(t, n)
+	emptyPreConfirmedBlock := uint64(1201960)
+	blockWithCandidates := uint64(1204672)
+	blockWithNoCandidates := uint64(1204673)
+	blockFullOfCandidates := uint64(1204674)
+
+	type preConfirmedTest struct {
+		description string
+		blockNumber uint64
+	}
+
+	tests := []preConfirmedTest{
+		{
+			description: "PreConfirmedBlock when there is no candidates",
+			blockNumber: blockWithNoCandidates,
+		},
+		{
+			description: "PreConfirmedBlock with candidates",
+			blockNumber: blockWithCandidates,
+		},
+		{
+			description: "PreConfirmedBlock when full of candidates",
+			blockNumber: blockFullOfCandidates,
+		},
+		{
+			description: "PreConfirmedBlock when empty",
+			blockNumber: emptyPreConfirmedBlock,
+		},
+	}
+
+	for _, test := range tests {
+		response, err := client.PreConfirmedBlock(t.Context(), strconv.FormatUint(test.blockNumber, 10))
+		require.NoError(t, err)
+
+		expectedEventCount, expectedPreConfirmedTxCount := countEventsAndTxs(response.Receipts)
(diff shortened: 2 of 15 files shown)


starknet/pre_confirmed_block.go
```diff
@@ -0,0 +1,33 @@
...
+	SequencerAddress *felt.Felt            `json:"sequencer_address"`
+	L1GasPrice       *GasPrice             `json:"l1_gas_price"`
+	L2GasPrice       *GasPrice             `json:"l2_gas_price"`
+	L1DAMode         L1DAMode              `json:"l1_da_mode"`
+	L1DataGasPrice   *GasPrice             `json:"l1_data_gas_price"`
+
+	TransactionStateDiffs []*StateDiff `json:"transaction_state_diffs"`
+}
+
+func (b *PreConfirmedBlock) L1GasPriceETH() *felt.Felt {
```
infrmtcs: It's a bit strange to me that we need these functions for `L1GasPrice` but not `L2GasPrice` and `L1DataGasPrice`.

starknet/pre_confirmed_block.go
```diff
@@ -0,0 +1,33 @@
+package starknet
+
+import "github.com/NethermindEth/juno/core/felt"
+
+// Block object returned by the feeder in JSON format for "get_preconfirmed_block" endpoint
+type PreConfirmedBlock struct {
+	Status           string                `json:"status"`
+	Transactions     []*Transaction        `json:"transactions"`
```
infrmtcs: Why is this `[]*Transaction` instead of `[]Transaction`? Is it because the transaction is optional, or because of how some library or existing code work?

starknetdata/feeder/feeder.go
```diff
@@ -172,3 +172,23 @@ func (f *Feeder) StateUpdatePendingWithBlock(ctx context.Context) (*core.StateUp
...
+
+// PreConfirmedWithBlockByNumber gets both pending state update and pending block from the feeder,
+// then adapts them to the core.PreConfirmed and list of transaction hashes types respectively
+func (f *Feeder) PreConfirmedBlockByNumber(ctx context.Context, blockNumber uint64) (*core.PreConfirmed, error) {
+	response, err := f.client.PreConfirmedBlock(ctx, strconv.FormatUint(blockNumber, 10))
+	if err != nil {
+		return nil, err
+	}
+
+	if response.Status != "PRE_CONFIRMED" {
```
infrmtcs: nit: Why don't we put this status in `AdaptPreConfirmedBlock`? Feel free to ignore this comment though.

starknet/pre_confirmed_block.go
EgeCaner: I agree! I checked it again, looks like we dont need these too. I will remove

starknet/pre_confirmed_block.go
EgeCaner: Good catch! I used the type we used in sn block type [here](https://github.com/NethermindEth/juno/blob/0e14bb9695a84481a575b3a6388425c9bb12a1e3/starknet/block.go#L21). I tried to replace it with `[]Transaction`, it works well. Lets use `[]Transaction`

clients/feeder/feeder.go
```diff
@@ -195,6 +166,61 @@ func newTestServer(t testing.TB, network *utils.Network) *httptest.Server {
 	}))
 }
 
+func resolveDirAndQueryArg(t testing.TB, path string, queryMap url.Values) (string, string, error) {
```
rodrodros: I am not sure if this is the right place to put this function and not a `*_test.go` file?

clients/feeder/feeder.go
```diff
@@ -140,39 +140,10 @@ func newTestServer(t testing.TB, network *utils.Network) *httptest.Server {
...
-			dir = "public_key"
-			queryArg = "pk"
-			queryMap[queryArg] = []string{queryArg}
-		case strings.HasSuffix(r.URL.Path, "get_signature"):
-			dir = "signature"
-			queryArg = blockNumberArg
-		case strings.HasSuffix(r.URL.Path, "get_block_traces"):
-			dir = "traces"
-			queryArg = "blockHash"
+		dir, queryArg, err := resolveDirAndQueryArg(t, r.URL.Path, queryMap)
```
rodrodros: But I see the function in this code is also here. Is there a good reason for this testing functions (which I am aware were already existing) to be in this script rather on a dedicated test file?

adapters/sn2core/sn2core_test.go
```diff
@@ -623,3 +623,148 @@ func TestAdaptCompiledClass(t *testing.T) {
 	require.NoError(t, err)
 	assert.Nil(t, result)
 }
+
+func TestAdaptPreConfirmed(t *testing.T) {
+	n := utils.HeapPtr(utils.SepoliaIntegration)
```
rodrodros: This is a bit unnecessary, it can be set as:

```
n := &utils.SepoliaIntegration
```

adapters/sn2core/sn2core_test.go
```diff
@@ -623,3 +623,148 @@ func TestAdaptCompiledClass(t *testing.T) {
...
+}
+
+func assertCandidateTxs(
+	t *testing.T,
+	transactions []starknet.Transaction,
+	candidateTxs []core.Transaction,
+	offset int,
+) {
+	t.Helper()
+	if len(candidateTxs) > 0 {
```
rodrodros: This `if` is not necessary, if length is 0 then the for is not executed by default

adapters/sn2core/sn2core_test.go
```diff
@@ -623,3 +623,148 @@ func TestAdaptCompiledClass(t *testing.T) {
...
+	}
+}
+
+func countEventsAndTxs(receipts []*starknet.TransactionReceipt) (uint64, int) {
+	var evCount uint64
+	txCount := 0
+	for _, r := range receipts {
+		if r == nil {
+			break
+		}
```
rodrodros, line 733: Why having a nil receipt implies breaking the for instead of continuing.

I know this is a testing framework. Does it means our receipt slices are of the form `[r1, r2, .., rk, nil, nil, ..., nil]`?

adapters/sn2core/sn2core_test.go
```diff
@@ -623,3 +623,148 @@ func TestAdaptCompiledClass(t *testing.T) {
...
+	if response.L2GasPrice != nil && response.L2GasPrice.PriceInFri != nil {
+		assert.Equal(t, response.L2GasPrice.PriceInFri, block.L2GasPrice.PriceInFri)
+	}
+	if response.L2GasPrice != nil && response.L2GasPrice.PriceInWei != nil {
+		assert.Equal(t, response.L2GasPrice.PriceInWei, block.L2GasPrice.PriceInWei)
+	}
+}
+
+func countEventsAndTxs(receipts []*starknet.TransactionReceipt) (uint64, int) {
+	var evCount uint64
```
rodrodros, line 728: Should this function have a `t.helper()`?

core/pending.go
```diff
@@ -0,0 +1,25 @@
+package core
+
+import (
+	"github.com/NethermindEth/juno/core/felt"
+)
+
+type PreConfirmed struct {
+	Block       *Block
```
rodrodros, line 8: Is block a reference because it is meant to be shared?

core/pending.go
```diff
@@ -0,0 +1,25 @@
...
+	StateUpdate *StateUpdate
+	// Does not fetch unknown classes. but we keep it for sequencer
+	NewClasses map[felt.Felt]Class
+	// We kept this field for sequencer, else non-sequencer node should use StateUpdate
+	TransactionStateDiffs []*StateDiff
```
rodrodros: This should be on the description of the `PreConfirmed` type. 

Can you explain to me why a PreConfirmed should have optional fields depending on how you use it. I think this can be prone to bugs later on since it is up to the programmer use it correctly. 

We should either have two PreConfirmed types or define them using generics if possible

core/pending.go
```diff
@@ -0,0 +1,25 @@
...
+
+func NewPreConfirmed(block *Block, stateUpdate *StateUpdate, transactionStateDiffs []*StateDiff, candidateTxs []Transaction) PreConfirmed {
+	return PreConfirmed{
+		Block:                 block,
+		StateUpdate:           stateUpdate,
+		NewClasses:            make(map[felt.Felt]Class),
+		TransactionStateDiffs: transactionStateDiffs,
+		CandidateTxs:          candidateTxs,
+	}
+}
```
rodrodros, line 24: If having two types is very difficult to code or it doesn't make sense, we should have at least two constructors, one for each modality. 

I am not fan of optional parameters on ctors. What if I initialize it with StateUpdate and TransactionStateDiffs?

starknetdata/feeder/feeder.go
```diff
@@ -172,3 +172,19 @@ func (f *Feeder) StateUpdatePendingWithBlock(ctx context.Context) (*core.StateUp
...
+		return nil, err
+	}
+
+	adaptedPreConfirmed, err := sn2core.AdaptPreConfirmedBlock(response, blockNumber)
+	if err != nil {
+		return nil, err
+	}
+
+	return &adaptedPreConfirmed, nil
+}
```
rodrodros: Return by value please

adapters/sn2core/sn2core_test.go
EgeCaner: Preconfirmed block returns `Transactions`, `Receipts`, `PerTxStateDiffs`. All lists has equal length, even though it might contain `CANDIDATE` transactions. For candidate transactions state diff and receipt will be `null`.  So lets say Transactions list has len `m` and for some `n` <= `m`, where `n` is the index where candidate txs start, and order is always first pre_confirmed then candidate transactions. Then [n m] state diffs and receipts will be `null` thus upon first occurance of nil entry we can stop processing pre_confirmed and start processing candidate transactions

clients/feeder/feeder.go
EgeCaner: Sounds better to extract test client into its own file

core/pending.go
EgeCaner: For per tx state diff, we are now using them for single tx traces. But for `NewClasses` since we are not able to fetch them they will always be empty for non=-sequencer node

adapters/sn2core/sn2core.go
```diff
@@ -397,12 +411,102 @@ func AdaptStateUpdate(response *starknet.StateUpdate) (*core.StateUpdate, error)
...
+
+		if adaptedStateDiff, err = AdaptStateDiff(stateDiff); err != nil {
+			return core.PreConfirmed{}, err
+		}
+		txStateDiffs = append(txStateDiffs, adaptedStateDiff)
+	}
+
+	lastPreconfirmedIndex := len(txStateDiffs) - 1
+
+	txns := make([]core.Transaction, lastPreconfirmedIndex+1)
```
infrmtcs:
```suggestion
	preconfirmedTxCount := len(txStateDiffs)

	txns := make([]core.Transaction, preconfirmedTxCount)
```

adapters/sn2core/sn2core.go
```diff
@@ -397,12 +411,102 @@ func AdaptStateUpdate(response *starknet.StateUpdate) (*core.StateUpdate, error)
...
+	for i := range lastPreconfirmedIndex + 1 {
```
infrmtcs:
```suggestion
	for i := range preconfirmedTxCount {
```

adapters/sn2core/sn2core.go
```diff
@@ -397,12 +411,102 @@ func AdaptStateUpdate(response *starknet.StateUpdate) (*core.StateUpdate, error)
...
+		adaptedTxn, err := AdaptTransaction(&response.Transactions[i])
+		if err != nil {
+			return core.PreConfirmed{}, err
+		}
+		txns[i] = adaptedTxn
+	}
+
+	rawTxCount := len(response.Transactions)
+	candidateTxs := make([]core.Transaction, rawTxCount-lastPreconfirmedIndex-1)
```
infrmtcs:
```suggestion
	candidateTxs := make([]core.Transaction, rawTxCount-preconfirmedTxCount)
```

adapters/sn2core/sn2core.go
```diff
@@ -397,12 +411,102 @@ func AdaptStateUpdate(response *starknet.StateUpdate) (*core.StateUpdate, error)
...
+
+	for i := range rawTxCount - lastPreconfirmedIndex - 1 {
```
infrmtcs:
```suggestion
	for i := range rawTxCount - preconfirmedTxCount {
```

adapters/sn2core/sn2core.go
```diff
@@ -397,12 +411,102 @@ func AdaptStateUpdate(response *starknet.StateUpdate) (*core.StateUpdate, error)
...
+		adaptedTxn, err := AdaptTransaction(&response.Transactions[i+lastPreconfirmedIndex+1])
```
infrmtcs:
```suggestion
		adaptedTxn, err := AdaptTransaction(&response.Transactions[preconfirmedTxCount+i])
```

adapters/sn2core/sn2core.go
```diff
@@ -397,12 +411,102 @@ func AdaptStateUpdate(response *starknet.StateUpdate) (*core.StateUpdate, error)
...
+		if err != nil {
+			return core.PreConfirmed{}, err
+		}
+		candidateTxs[i] = adaptedTxn
+	}
+
+	receipts := make([]*core.TransactionReceipt, lastPreconfirmedIndex+1)
+	eventCount := uint64(0)
+	for i, receipt := range response.Receipts[:lastPreconfirmedIndex+1] {
```
infrmtcs:
```suggestion
	receipts := make([]*core.TransactionReceipt, preconfirmedTxCount)
	eventCount := uint64(0)
	for i, receipt := range response.Receipts[:preconfirmedTxCount] {
```

adapters/sn2core/sn2core.go
```diff
@@ -397,12 +411,102 @@ func AdaptStateUpdate(response *starknet.StateUpdate) (*core.StateUpdate, error)
...
+
+	lastPreconfirmedIndex := len(txStateDiffs) - 1
+
+	txns := make([]core.Transaction, lastPreconfirmedIndex+1)
+	for i := range lastPreconfirmedIndex + 1 {
+		adaptedTxn, err := AdaptTransaction(&response.Transactions[i])
+		if err != nil {
+			return core.PreConfirmed{}, err
+		}
+		txns[i] = adaptedTxn
```
infrmtcs: nit:

```suggestion
		txns[i], err = AdaptTransaction(&response.Transactions[i])
		if err != nil {
			return core.PreConfirmed{}, err
		}
```

or

```suggestion
		if txns[i], err = AdaptTransaction(&response.Transactions[i]); err != nil {
			return core.PreConfirmed{}, err
		}
```

adapters/sn2core/sn2core.go
infrmtcs: This is nit, but I think it will help to avoid coding errors.

adapters/sn2core/sn2core.go
```diff
@@ -397,12 +411,102 @@ func AdaptStateUpdate(response *starknet.StateUpdate) (*core.StateUpdate, error)
...
+			L2GasPrice:       (*core.GasPrice)(response.L2GasPrice),
+			// Not required in spec but useful
+			TransactionCount: uint64(len(txns)),
+			EventCount:       eventCount,
+			EventsBloom:      core.EventsBloom(receipts),
+			// Following fields are nil for pre_confirmed block
+			Hash:            nil,
+			ParentHash:      nil,
+			GlobalStateRoot: nil,
+			Signatures:      nil,
```
infrmtcs: It's great if we can sort these fields in the same order as the type definition. It would help us to double check if we're missing any field, although I think ideally it should be checked by a linter instead.

infrmtcs: I think in this case, probably we can sort only the non-nil fields and leave the nil fields at the end, I like what you're doing here too.

adapters/sn2core/sn2core_test.go
```diff
@@ -623,3 +623,148 @@ func TestAdaptCompiledClass(t *testing.T) {
...
+		},
+		{
+			description: "PreConfirmedBlock when full of candidates",
+			blockNumber: blockFullOfCandidates,
+		},
+		{
+			description: "PreConfirmedBlock when empty",
+			blockNumber: emptyPreConfirmedBlock,
+		},
+	}
```
infrmtcs, line 657: I'm always a big fan of table tests :tada:

adapters/sn2core/sn2core_test.go
```diff
@@ -623,3 +623,148 @@ func TestAdaptCompiledClass(t *testing.T) {
...
+	if response.L1DataGasPrice != nil && response.L1DataGasPrice.PriceInWei != nil {
+		assert.Equal(t, response.L1DataGasPrice.PriceInWei, block.L1DataGasPrice.PriceInWei)
+	}
+	if response.L2GasPrice != nil && response.L2GasPrice.PriceInFri != nil {
+		assert.Equal(t, response.L2GasPrice.PriceInFri, block.L2GasPrice.PriceInFri)
+	}
+	if response.L2GasPrice != nil && response.L2GasPrice.PriceInWei != nil {
+		assert.Equal(t, response.L2GasPrice.PriceInWei, block.L2GasPrice.PriceInWei)
+	}
```
infrmtcs: Why don't we assert them directly? Is it possible that either one of them is nil while the other one isn't?

adapters/sn2core/sn2core_test.go
```diff
@@ -623,3 +623,148 @@ func TestAdaptCompiledClass(t *testing.T) {
...
+func assertPreConfirmedBlockReceipts(
+	t *testing.T,
+	expectedReceipts []*starknet.TransactionReceipt,
+	blockReceipts []*core.TransactionReceipt,
+	expectedPreConfirmedTxCount int,
+) {
+	t.Helper()
+	if assert.Equal(t, expectedPreConfirmedTxCount, len(blockReceipts)) {
+		for i := range expectedPreConfirmedTxCount {
+			r := expectedReceipts[i]
```
infrmtcs: nit: I think this could help readability a bit.

```suggestion
		for i, actualReceipt := range blockReceipts {
			expectedReceipt := expectedReceipts[i]
```
or 
```suggestion
		for i, actual := range blockReceipts {
			expected := expectedReceipts[i]
```

infrmtcs (approved):
The rest LGTM!

adapters/sn2core/sn2core_test.go
EgeCaner: Not sure, it does need access to it. Does such pure functions need `t.Helper()`?

adapters/sn2core/sn2core_test.go
EgeCaner: changed as suggested

adapters/sn2core/sn2core.go
EgeCaner: Non-nil fields ordered exactly as in `core.Header` except nil fields in between, they moved to bottom

adapters/sn2core/sn2core.go
EgeCaner: Changed as in the first one

adapters/sn2core/sn2core.go
EgeCaner: changed as suggested

adapters/sn2core/sn2core.go
EgeCaner: changed as suggested

adapters/sn2core/sn2core.go
EgeCaner: changed as suggested

adapters/sn2core/sn2core.go
EgeCaner: changed as suggested

adapters/sn2core/sn2core.go
EgeCaner: changed as suggested

adapters/sn2core/sn2core.go
EgeCaner: changed as suggested

starknetdata/feeder/feeder.go
EgeCaner: changed as suggested

adapters/sn2core/sn2core_test.go
EgeCaner: changed as suggested

adapters/sn2core/sn2core_test.go
rodrodros: Mmm, I think it doesn't

rodrigo-pino (approved):

core/pending.go
EgeCaner: Lets address this, when we integrate PreConfirmed block into block building process and see how does it fit. At the moment we utilize all the fields, except `NewClasses` (beside passing to `PendingState` as an empty entry).

core/pending.go
EgeCaner: Pending `Block`data should be considered immutable and read-only after creation. We store it in atomic.Pointer and pending data should be replaced as a whole on mutations to it. `Pending` implementation we had before was using reference for a block, thats why I used the same here, to easily integrate with the existing code. While integrating new `PendingData` into RPCs, I will check if we can convert it to value type without much effort. I will do it in this [PR](https://github.com/NethermindEth/juno/pull/2952)

PR #2958, opened 2025-07-07
fix: Tendermint Commit should be executed eagerly

infrmtcs:


consensus/driver/driver.go
```diff
@@ -105,7 +108,7 @@ func (d *Driver[V, H, A]) Stop() {
 	}
 }
 
-func (d *Driver[V, H, A]) execute(actions []types.Action[V, H, A]) {
+func (d *Driver[V, H, A]) execute(actions []types.Action[V, H, A]) (isCommitted bool) {
```
rodrodros, line 113: Can we add a description to this internal function

consensus/tendermint/process.go
rodrodros, line 1: Looks better 👍

rodrigo-pino (approved):

consensus/tendermint/rule_commit_value.go
```diff
@@ -28,12 +28,13 @@ func (t *stateMachine[V, H, A]) uponCommitValue(cachedProposal *CachedProposal[V
...
-func (t *stateMachine[V, H, A]) doCommitValue() types.Action[V, H, A] {
+func (t *stateMachine[V, H, A]) doCommitValue(cachedProposal *CachedProposal[V, H, A]) types.Action[V, H, A] {
 	t.messages.DeleteHeightMessages(t.state.height)
 	t.state.height++
 	t.state.lockedRound = -1
 	t.state.lockedValue = nil
 	t.state.validRound = -1
 	t.state.validValue = nil
-	return t.startRound(0)
+	t.resetState(0)
```
rianhughes, line 38: Do we need to call resetState here? After we commit the block, we will vall ProcessStart(0) which will call resetState(0)

infrmtcs: Technically no, but I think the state looks weird when we move to new height with old round. This is what I see when writing the tests.

PR #2960, opened 2025-07-07
feat: Implement commit listener and consensus data source

infrmtcs:


consensus/consensus.go
```diff
@@ -0,0 +1,125 @@
...
+	Driver         *driver.Driver[starknet.Value, starknet.Hash, starknet.Address]
+	CommitListener driver.CommitListener[starknet.Value, starknet.Hash, starknet.Address]
+}
+
+func Init(
+	logger *utils.ZapLogger,
+	database db.KeyValueStore,
+	blockchain *blockchain.Blockchain,
+	vm vm.VM,
+	nodeAddress starknet.Address,
```
rodrodros: Address by reference?

consensus/datasource/consensus_data_source.go
```diff
@@ -0,0 +1,84 @@
...
+	"github.com/NethermindEth/juno/service"
+	"github.com/NethermindEth/juno/sync"
+)
+
+const maxCommitHistory = 1024 // TODO: make this configurable
+
+type DataSourceService interface {
+	service.Service
+	sync.DataSource
+}
```
rodrodros: Some comments explaining this interface 🙏

consensus/integtest/p2p.go
```diff
@@ -31,115 +24,84 @@ func (c networkNodeConfig) isConnected(to int) bool {
 
 type networkConfig []networkNodeConfig
 
-func connect(t *testing.T, mutex *sync.Mutex, thisNodeConfig, otherNodeConfig networkNodeConfig) {
+func (c networkConfig) setup(t *testing.T, configFn networkConfigFn) {
```
rodrodros, line 29: `networkConfig` by reference, otherwise every time this method is called they will be a stack copy of the `c` param. Is this approach intentional?

consensus/integtest/p2p.go
```diff
@@ -31,115 +24,84 @@ func (c networkNodeConfig) isConnected(to int) bool {
...
 }
 
 func (c networkConfig) add(from, to int) {
 	c[from].adjacentNodes[to] = struct{}{}
 }
 
-func initEmptyNetworkConfig(t *testing.T, n int) networkConfig {
-	t.Helper()
+//nolint:gosec // The whole package is for testing purpose only, so it's safe to use weak random.
+func (c networkConfig) addRandomDirection(from, to int) bool {
```
rodrodros, line 61: Similar observation to the previous comment

rodrodros: Thank you for the lint ignore explanation

consensus/integtest/p2p.go
infrmtcs: `networkConfig` is just a type definition of a slice, so it's safe to use value here.

consensus/datasource/consensus_data_source.go
infrmtcs: I just want to hide the implementation type (`consensusDataSource`), so the `New` function will return an interface. However, this return type needs to be both:
- `service.Service`: to call `Run(context.Context) error` together with other services
- `sync.DataSource`: to be used in `Synchronizer`
Probably a way to avoid this is to return the implementation type directly.

consensus/consensus.go
```diff
@@ -0,0 +1,125 @@
...
+	currentHeight := types.Height(chainHeight + 1)
+
+	tendermintDB := consensusDB.NewTendermintDB[starknet.Value, starknet.Hash, starknet.Address](database, currentHeight)
+
+	executor := builder.NewExecutor(blockchain, vm, logger, false, true) // TODO: We're currently skipping signature validation
+	builder := builder.New(blockchain, executor)
+
+	proposalStore := proposal.ProposalStore[starknet.Hash]{}
+	proposer := proposer.New(logger, &builder, &proposalStore, *nodeAddress, toValue)
+	stateMachine := tendermint.New(tendermintDB, logger, *nodeAddress, proposer, validators, currentHeight)
```
rodrodros, line 64: Should nodeAddress be passed by reference here as well? In both `proposer.New` and `tendermint.New`

consensus/consensus.go
```diff
@@ -0,0 +1,125 @@
...
+	p2p := p2p.New(host, logger, &builder, &proposalStore, currentHeight, &config.DefaultBufferSizes)
+
+	commitListener := driver.NewCommitListener(logger, &proposalStore, proposer, p2p)
+	driver := driver.New(logger, tendermintDB, stateMachine, commitListener, p2p, timeoutFn)
+
+	return ConsensusServices{
+		Host:           host,
+		Proposer:       proposer,
+		P2P:            p2p,
+		Driver:         &driver,
```
rodrodros, line 94: Why hold a reference to the `driver` instead of the `driver` itself. Not saying it should be done either way, just curious

consensus/datasource/consensus_data_source.go
```diff
@@ -0,0 +1,78 @@
...
+) *consensusDataSource[V, H, A] {
+	return &consensusDataSource[V, H, A]{
+		commitListener: commitListener,
+		proposer:       proposer,
+		cache:          syncmap.Map{},
+		latest:         atomic.Uint64{},
+	}
+}
+
+func (c *consensusDataSource[V, H, A]) Run(ctx context.Context) error {
```
rodrodros, line 40: Does this method returns an error to satisfy an interface? Otherwise I think we can return nothing

consensus/datasource/consensus_data_source.go
```diff
@@ -0,0 +1,78 @@
...
+
+func New[V types.Hashable[H], H types.Hash, A types.Addr](
+	commitListener driver.CommitListener[V, H, A],
+	proposer proposer.Proposer[V, H],
+	p2p p2p.P2P[V, H, A],
+) *consensusDataSource[V, H, A] {
+	return &consensusDataSource[V, H, A]{
+		commitListener: commitListener,
+		proposer:       proposer,
+		cache:          syncmap.Map{},
```
rodrodros, line 35: This `Map` type is optimized for "ever growing" cache and when writing/overwriting disjoint sets of key at the same time and multiple reads.

Given the use of `consensusDataSource`, it is meant to write a key, and delete another key. The safety mechanisms come out of the need for reading later on, we want to avoid race condition.

It looks that this data type we are looking for here is a queue (of size maxCommitHistory) with constant look up time. What would be the best implementation for this 🤔

consensus/driver/commit_listener.go
```diff
@@ -0,0 +1,76 @@
...
+	"github.com/NethermindEth/juno/consensus/types"
+	"github.com/NethermindEth/juno/sync"
+	"github.com/NethermindEth/juno/utils"
+)
+
+type CommitListener[V types.Hashable[H], H types.Hash, A types.Addr] interface {
+	// Commit is called by Tendermint when a block has been decided on and can be committed to the DB.
+	Commit(context.Context, types.Height, V)
+	Listen() <-chan sync.CommittedBlock
+}
```
rodrodros: Please a description for `CommitListener`

consensus/consensus.go
infrmtcs: The other 3 are interfaces, so I can use them directly. For driver, it's a struct, to typically I'll use reference first for correctness before switching to value for optimization.

consensus/datasource/consensus_data_source.go
infrmtcs: Ideally, we don't need any of these data sources, and we can use the channel directly. Specifically, the `Synchronizer` should select a channel to get newly committed block. However, the current structure there is using a polling manner, where `Synchronizer` run separate goroutines to poll the feeder gateway and store them in an atomic pointer, and the loop just fetch from the atomic pointer directly.

consensus/datasource/consensus_data_source.go
infrmtcs: Yes, currently the data source implementation for feeder gateway can return error.

consensus/consensus.go
infrmtcs: In fact I tried that, but in order to convert all of them down the line, I'll need to change the fields in `MessageHeader`. I think there will be perf gain there, and but the cost of changing is higher while it's outside the scope of this PR, so I think we should have a separate PR to track this specifically and do it after this PR to avoid conflicts.

consensus/datasource/consensus_data_source.go
rodrodros: Are then plans to materialize this idea?

infrmtcs: Yes, I also plan to work on this after this PR, but it needs to change more on how Synchronizer works.

rodrigo-pino (approved):

consensus/driver/commit_listener.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/NethermindEth/juno/consensus/p2p"
+	"github.com/NethermindEth/juno/consensus/proposal"
+	"github.com/NethermindEth/juno/consensus/proposer"
+	"github.com/NethermindEth/juno/consensus/types"
+	"github.com/NethermindEth/juno/sync"
+	"github.com/NethermindEth/juno/utils"
+)
+
+// CommitListener is a component that is used to notify different components that a new committed block is available.
+type CommitListener[V types.Hashable[H], H types.Hash, A types.Addr] interface {
```
rianhughes, line 15: nit: `CommitListener` isn't the best name imo because it can also be used to commit a block. Maybe something like `BlockCommitter`?

sync/data_source.go
```diff
@@ -16,6 +16,7 @@ type CommittedBlock struct {
 	Block       *core.Block
 	StateUpdate *core.StateUpdate
 	NewClasses  map[felt.Felt]core.Class
+	Persisted   chan struct{}
```
rianhughes: What is this for? Can we add a doc string here?

consensus/datasource/consensus_data_source.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+
+	return committedBlock.Block, nil
+}
+
+func (c *consensusDataSource[V, H, A]) BlockPending(ctx context.Context) (sync.Pending, error) {
+	return sync.Pending{}, errors.New("not implemented") // TODO: Revise this
+}
+
+func (c *consensusDataSource[V, H, A]) PreConfirmedBlockByNumber(ctx context.Context, blockNumber uint64) (core.PreConfirmed, error) {
```
rianhughes, line 76: `blockNumber` is a dead arguement

consensus/datasource/consensus_data_source.go
```diff
@@ -0,0 +1,78 @@
...
+func (c *consensusDataSource[V, H, A]) BlockLatest(ctx context.Context) (*core.Block, error) {
+	committedBlock, err := c.BlockByNumber(ctx, c.latest.Load())
+	if err != nil {
+		return nil, err
+	}
+
+	return committedBlock.Block, nil
+}
+
+func (c *consensusDataSource[V, H, A]) BlockPending(ctx context.Context) (sync.Pending, error) {
```
rianhughes, line 72: ctx is a dead argument

consensus/datasource/consensus_data_source.go
```diff
@@ -0,0 +1,78 @@
...
+	return sync.Pending{}, errors.New("not implemented") // TODO: Revise this
+}
+
+func (c *consensusDataSource[V, H, A]) PreConfirmedBlockByNumber(ctx context.Context, blockNumber uint64) (core.PreConfirmed, error) {
+	return *c.proposer.Preconfirmed(), nil
+}
```
rianhughes: will `blockNumber` ever be used? Currently it's misleading

consensus/datasource/consensus_data_source.go
```diff
@@ -0,0 +1,78 @@
...
+	"github.com/NethermindEth/juno/core"
+	"github.com/NethermindEth/juno/sync"
+)
+
+const maxCommitHistory = 1024 // TODO: make this configurable
+
+type consensusDataSource[V types.Hashable[H], H types.Hash, A types.Addr] struct {
+	commitListener driver.CommitListener[V, H, A]
+	proposer       proposer.Proposer[V, H]
+	cache          syncmap.Map
```
rianhughes, line 23: I think we could use a circular buffer here given we only store maxCommitHistory blocks at a time. It should be much faster than syncMap

consensus/driver/commit_listener.go
```diff
@@ -0,0 +1,79 @@
...
+	case <-ctx.Done():
+		return
+	case b.commits <- committedBlock:
+	}
+
+	select {
+	case <-ctx.Done():
+		return
+	case <-committedBlock.Persisted:
+	}
```
rianhughes, line 71: I'm guessing Persisted will be closed after blockchain.Store is called??

consensus/datasource/consensus_data_source.go
```diff
@@ -0,0 +1,78 @@
...
+	"github.com/NethermindEth/juno/consensus/p2p"
+	"github.com/NethermindEth/juno/consensus/proposer"
+	"github.com/NethermindEth/juno/consensus/types"
+	"github.com/NethermindEth/juno/core"
+	"github.com/NethermindEth/juno/sync"
+)
+
+const maxCommitHistory = 1024 // TODO: make this configurable
+
+type consensusDataSource[V types.Hashable[H], H types.Hash, A types.Addr] struct {
```
rianhughes, line 20: This type isn't used anywhere at the moment, what is its purpose? Will it have tests?

infrmtcs: `New` function returns it, which is called by `consensus.Init`. The integ test is covering this.

consensus/driver/commit_listener.go
infrmtcs: Yes, we're doing that in `sync.go`.

consensus/driver/commit_listener.go
infrmtcs: IMO, `BlockCommitter` is more misleading. We're not committing the block here, it's the driver and state machine's decision. We're also not persisting the block here. We're only listening to the commits and forward to appropriate channels, including the `Synchronizer` to persist the block, `p2p` to reset the validator streams and `proposer` to reset the block.

sync/data_source.go
infrmtcs: Added comment

consensus/datasource/consensus_data_source.go
infrmtcs: Added check for head block number

rianhughes (approved):

consensus/datasource/consensus_data_source.go
infrmtcs: The problem is that `Synchronizer` is currently actively query the blocks from feeder gateway, instead of listening to a channel returning them. That's why we can't guarantee that a block will be queried exactly once, so circular buffer may not work. The proper solution in my opinion is to replace the current mechanism in `Synchronizer` to use channel instead, but it's outside the scope of this PR, which doesn't do anything outside of consensus.

PR #2962, opened 2025-07-08
Rianhughes/tendermint sync

rianhughes:
This PR implements the sync service for consensus. Its purpose is to sync to the chain head, then switch off. It is not a mechanism to catchup to the chain head if we fall behind (which should be addressed separately).

The service asks P2P for the next block. It then queries peers for the precommits associated with this block (assuming they won't be in the header), builds the proposal, and sends all of this to the `Driver`. The `Driver` should then commit it (by triggering `line 49`). The sync service is stopped whenever the state machine sees a quorum of `prevotes` (earliest possible indication we are at the chain head), and sends a signal to the sync service to shut down. 

Note: sync requires the precommits to be exposed. Currently they are not. To push the block through the state machine, we may have to forge them. Ie create a {H,R,sender, ID} for each sender for a given block.

Note: The p2p logic has little to no tests (eg no test for the `Run()` function for the `p2p.Service` type). To get around this I implemented a new interface (`WithBlockCh`), until we implement the p2p tests. This should probably be done next given it's a core part of the nodes functionality.

### DIFF
--- a/consensus/sync/sync.go
+++ b/consensus/sync/sync.go
@@ -46,6 +46,8 @@ func New[V types.Hashable[H], H types.Hash, A types.Addr](
 
 func (s *Sync[V, H, A]) Run(originalCtx context.Context) {
 	ctx, cancel := context.WithCancel(originalCtx)
+	defer cancel()
+
 	go func() {
 		err := s.syncService.Run(ctx)
 		if err != nil {
@@ -57,7 +59,6 @@ func (s *Sync[V, H, A]) Run(originalCtx context.Context) {
 	for {
 		select {
 		case <-ctx.Done():
-			cancel()
 			return
 		case committedBlock := <-s.syncService.Listen():
 			msgV := s.toValue(committedBlock.Block.Hash)
@@ -90,7 +91,6 @@ func (s *Sync[V, H, A]) Run(originalCtx context.Context) {
 			for _, precommit := range precommits {
 				select {
 				case <-ctx.Done():
-					cancel()
 					return
 				case s.driverPrecommitCh <- precommit:
 				}
@@ -108,7 +108,6 @@ func (s *Sync[V, H, A]) Run(originalCtx context.Context) {
 
 			select {
 			case <-ctx.Done():
-				cancel()
 				return
 			case s.driverProposalCh <- proposal:
 			}


consensus/tendermint/process.go
```diff
@@ -88,6 +88,14 @@ func (t *stateMachine[V, H, A]) process(
 		roundCachedProposal = t.findProposal(*recentlyReceivedRound)
 	}
 
+	if t.isSyncing {
+		// Stop syncing when we receive a quorum of prevotes
+		if t.uponPolkaAny() || t.uponPolkaNil() {
```
infrmtcs: Why do we stop syncing if we receive a quorum of prevotes?

consensus/driver/driver.go
```diff
@@ -131,6 +134,12 @@ func (d *Driver[V, H, A]) execute(actions []types.Action[V, H, A]) {
 			if err := d.db.DeleteWALEntries(action.Height); err != nil {
 				d.log.Errorw("failed to delete WAL messages during commit", "height", action.Height, "round", action.Round, "err", err)
 			}
+		case *types.StopSync:
+			if d.stopSyncCh != nil {
+				d.stopSyncCh <- struct{}{}
+				close(d.stopSyncCh)
```
infrmtcs: nit:
```suggestion
				close(d.stopSyncCh)
```

consensus/sync/sync.go
```diff
@@ -0,0 +1,117 @@
...
+		case <-s.stopSyncCh:
+			syncCancel()
+			return
+		case committedBlock := <-s.blockCh:
+			msgHeader := types.MessageHeader[A]{
+				Height: types.Height(committedBlock.Block.Number),
+				Round:  -1, // Todo: placeholder until round is placed in the spec
+			}
+			msgV := s.toValue(committedBlock.Block.Hash)
+			msgH := s.toHash(committedBlock.Block.Hash)
```
infrmtcs:
```suggestion
			msgH := msgV.Hash()
```
So we can remove `toHash`

consensus/sync/sync.go
```diff
@@ -0,0 +1,117 @@
...
+		stopSyncCh:        stopSyncCh,
+		toValue:           toValue,
+		toHash:            toHash,
+		proposalStore:     proposalStore,
+		blockCh:           blockCh,
+	}
+}
+
+func (s *Sync[V, H, A]) Run(ctx context.Context) {
+	syncCtx, syncCancel := context.WithCancel(ctx)
```
infrmtcs: nit: It's easy to make a mistake here using `ctx` in subsequent lines, so I think we should either

```suggestion
	ctx, syncCancel := context.WithCancel(originalCtx)
```
or
```suggestion
	var syncCancel context.CancelFunc
	ctx, syncCancel = context.WithCancel(ctx)
```

consensus/sync/sync.go
```diff
@@ -0,0 +1,117 @@
...
+					NewClasses:  committedBlock.NewClasses,
+				},
+				SimulateResult: &blockchain.SimulateResult{
+					BlockCommitments: committedBlock.Commitments,
+					ConcatCount:      concatCommitments,
+				},
+				// Todo: this needs added to the spec.
+				L2GasConsumed: 1,
+			}
+			s.proposalStore.Store(msgH, &buildResult)
```
infrmtcs, line 85: `buildResult` must be written to `proposalStore` first, otherwise there can be a race condition where the driver decides to commit quickly and cannot find the proposal in `proposalStore`, because it's not written yet. We did this similarly in proposal stream demux.

consensus/sync/sync.go
```diff
@@ -0,0 +1,117 @@
...
+			msgHeader := types.MessageHeader[A]{
+				Height: types.Height(committedBlock.Block.Number),
+				Round:  -1, // Todo: placeholder until round is placed in the spec
+			}
+			msgV := s.toValue(committedBlock.Block.Hash)
+			msgH := s.toHash(committedBlock.Block.Hash)
+
+			precommits := s.getPrecommits(types.Height(committedBlock.Block.Number))
+			for _, precommit := range precommits {
+				s.driverPrecommitCh <- precommit
```
infrmtcs: I think we should select with `context.Context` if possible.

consensus/sync/sync.go
```diff
@@ -0,0 +1,117 @@
...
+		case <-syncCtx.Done():
+			syncCancel()
+			return
+		case <-s.stopSyncCh:
+			syncCancel()
+			return
+		case committedBlock := <-s.blockCh:
+			msgHeader := types.MessageHeader[A]{
+				Height: types.Height(committedBlock.Block.Number),
+				Round:  -1, // Todo: placeholder until round is placed in the spec
```
infrmtcs: We should make this a const instead. This is because it's our convention and not specified in specs.

consensus/sync/sync.go
```diff
@@ -0,0 +1,117 @@
...
+	driverProposalCh  chan types.Proposal[V, H, A]
+	driverPrecommitCh chan types.Precommit[H, A]
+	// Todo: for now we can forge the precommit votes of our peers
+	// In practice, this information needs to be exposed to peers.
+	getPrecommits func(types.Height) []types.Precommit[H, A]
+	stopSyncCh    <-chan struct{}
+	toValue       func(*felt.Felt) V
+	toHash        func(*felt.Felt) H
+	proposalStore *proposal.ProposalStore[H]
+	blockCh       chan p2pSync.BlockBody
```
infrmtcs:
```suggestion
	blockCh       <-chan p2pSync.BlockBody
```

consensus/sync/sync.go
```diff
@@ -0,0 +1,117 @@
...
+					Block:       committedBlock.Block,
+					StateUpdate: committedBlock.StateUpdate,
+					NewClasses:  committedBlock.NewClasses,
+				},
+				SimulateResult: &blockchain.SimulateResult{
+					BlockCommitments: committedBlock.Commitments,
+					ConcatCount:      concatCommitments,
+				},
+				// Todo: this needs added to the spec.
+				L2GasConsumed: 1,
```
infrmtcs, line 83: I think we should also check with Starkware to understand whether the node is expected to validate the block or blindly trust it as long as there are 2f+1 commits.

consensus/sync/sync.go
```diff
@@ -0,0 +1,117 @@
...
+			}
+			msgV := s.toValue(committedBlock.Block.Hash)
+			msgH := s.toHash(committedBlock.Block.Hash)
+
+			precommits := s.getPrecommits(types.Height(committedBlock.Block.Number))
+			for _, precommit := range precommits {
+				s.driverPrecommitCh <- precommit
+			}
+
+			msgHeader.Sender = committedBlock.Block.SequencerAddress.Bits()
```
infrmtcs: nit: I think this should be inline in the declaration of `msgHeader` instead, or we can declare a variable for this first and use that variable in `msgHeader` declaration.

p2p/p2p.go
```diff
@@ -35,6 +36,14 @@ const (
 	clientName                = "juno"
 )
 
+// Todo: this interface allows us to mock the P2P service until we implement additional tests / test infrastructure
+type WithBlockCh interface {
+	service.Service
+	WithBlockCh(blockCh chan p2pSync.BlockBody)
```
infrmtcs:
```suggestion
	Listen() <-chan p2pSync.BlockBody
```
This is because ideally, the write side should be the one "owns" the channel, because a write to a closed channel can panic while a read doesn't. The existing code already write data to a channel, so we can expose it instead of forwarding it again to another channel.

infrmtcs: To do this, we can:
- Initialize the channel in `New`
- Return this channel in `Listen`
- Modify `pipeline.Bridge` utils to accept a channel as an argument instead of initializing it inside.

p2p/sync/sync.go
```diff
@@ -131,18 +135,22 @@ func (s *Service) processBlock(ctx context.Context, blockNumber uint64) error {
...
+				return fmt.Errorf("failed to store block: %w", err)
+			}
 
-		s.log.Infow("Stored Block", "number", b.block.Number, "hash", b.block.Hash.ShortString(),
-			"root", b.block.GlobalStateRoot.ShortString())
-		s.listener.OnSyncStepDone(junoSync.OpStore, b.block.Number, time.Since(storeTimer))
+			s.log.Infow("Stored Block", "number", b.Block.Number, "hash", b.Block.Hash.ShortString(),
+				"root", b.Block.GlobalStateRoot.ShortString())
+			s.listener.OnSyncStepDone(junoSync.OpStore, b.Block.Number, time.Since(storeTimer))
+		}
```
infrmtcs: We need to connect the `Synchronizer` here, because this is missing necessary logic when committing a block, e.g setting feed, setting atomic pointers. This can be done in another PR though.

consensus/tendermint/process.go
rianhughes: We should stop syncing when we receive a quorum of prevotes at our current height. Eg

t=0
Our node is at height 0
Network is at height 100
Msgs are at height 100 (+-1). 
We don't see a quorum of prevotes at the current height 0, so we don't stop syncing.

t=t'
Our node is at height 110
Network is at height 100
Msgs are at 110. We are at height 110. 
We see a quorum of prevotes at our height 100, so we should stop syncing.

consensus/sync/sync.go
rianhughes: To cancel early?

infrmtcs: Any "naked" blocking operation is a source of deadlock preventing graceful shutdown.

consensus/tendermint/process.go
infrmtcs: Doesn't this mean we can be blocked forever if not receiving the precommits? I'm thinking about the case:
- Receive prevote quorum at height 100.
- Lost internet connection for 20 seconds.
- Connectivity restored, network moved to height 110 and don't broadcast the precommits for 100 anymore.

infrmtcs: Another case is:
- We're at height 1, network is at height 1000000.
- Attackers send a prevote quorum for height 1.
- This is triggered, which blocks the sync at height 1 forever.

consensus/sync/sync.go
```diff
@@ -0,0 +1,116 @@
...
+		toValue:           toValue,
+		proposalStore:     proposalStore,
+		blockCh:           blockCh,
+	}
+}
+
+func (s *Sync[V, H, A]) Run(originalCtx context.Context) {
+	ctx, cancel := context.WithCancel(originalCtx)
+	go func() {
+		s.syncService.WithBlockCh(s.blockCh)
```
infrmtcs: This can result in race condition, because it's possible to be called after the `syncService` starts receiving blocks.

consensus/sync/sync.go
rianhughes: Yup, good point. I've implemented this

consensus/sync/sync.go
rianhughes: The sync mechanism has been updated making this comment outdated

p2p/p2p.go
rianhughes: I've updated the interface, but want to clean one thing up (stop Listen from blocking the select case in sync.go)

consensus/sync/sync.go
rianhughes: I've asked the Starknet guys, just waiting on a response

consensus/sync/sync.go
rianhughes: This comment is outdated given the changes related to Listen()

consensus/tendermint/process.go
rianhughes: Yup, good point. I've removed this, along with the stopchannel. I think there is an easier solution where the sync service might be able to shut itself down (if it can detect a block was added to the chain that it didn't know about - but I'm still exploring this)

consensus/sync/sync.go
```diff
@@ -0,0 +1,132 @@
...
+	blockCh chan p2pSync.BlockBody,
+) Sync[V, H, A] {
+	return Sync[V, H, A]{
+		syncService:       syncService,
+		driverProposalCh:  driverProposalCh,
+		driverPrecommitCh: driverPrecommitCh,
+		getPrecommits:     getPrecommits,
+		toValue:           toValue,
+		proposalStore:     proposalStore,
+	}
```
rodrodros: `blockCh` is passed but not used?

consensus/sync/sync.go
```diff
@@ -0,0 +1,132 @@
...
+			return
+		}
+	}()
+
+	// Todo: this is kind of ugly, but it unblocks the listen() function in the select-case below
+	forwardCh := make(chan p2pSync.BlockBody)
+	go func() {
+		for {
+			inner := s.syncService.Listen()
+			for blk := range inner {
```
rodrodros: nit: `blk` to `block`

consensus/sync/sync.go
```diff
@@ -0,0 +1,132 @@
...
+			cancel()
+			return
+		}
+	}()
+
+	// Todo: this is kind of ugly, but it unblocks the listen() function in the select-case below
+	forwardCh := make(chan p2pSync.BlockBody)
+	go func() {
+		for {
+			inner := s.syncService.Listen()
```
rodrodros: What does `inner` means?

consensus/sync/sync.go
```diff
@@ -0,0 +1,132 @@
...
+			for blk := range inner {
+				forwardCh <- blk
+			}
+		}
+	}()
+
+	for {
+		select {
+		case <-ctx.Done():
+			cancel()
```
rodrodros: Why calling `cancel()` manually in each execution branch instead of deferring it?

p2p/sync/sync.go
```diff
@@ -130,19 +137,26 @@ func (s *Service) processBlock(ctx context.Context, blockNumber uint64) error {
...
-			return fmt.Errorf("failed to process block: %w", b.err)
-		}
+	if s.listeningToBlock {
+		// Note: if this service calls blockchain.Height()
+		// before the listener stores the block, we may
+		// end up requesting the same block again..
+		s.ListenerBlockCh <- blocksCh
+	} else {
+		for b := range blocksCh {
+			if b.Err != nil {
```
rodrodros: Why not return right at then end of `if` execution path and remove the `else` block. It makes clear that the first conditions means the end of an execution path. Also subjective, feel free to ignore

p2p/sync/sync.go
```diff
@@ -76,16 +79,20 @@ func (s *Service) Run(ctx context.Context) {
...
 		}
-
 		cancelIteration()
 	}
 }
 
+func (s *Service) SetListener() {
+	s.ListenerBlockCh = make(chan (<-chan BlockBody), 1)
+	s.listeningToBlock = true
+}
```
infrmtcs: We should avoid this. This would cause race condition when we don't know exactly when this is called.

p2p/sync/sync.go
```diff
@@ -40,6 +40,9 @@ type Service struct {
 	blockchain *blockchain.Blockchain
 	listener   junoSync.EventListener
 	log        utils.SimpleLogger
+
+	listeningToBlock bool
+	ListenerBlockCh  chan (<-chan BlockBody)
```
infrmtcs:
```suggestion
	blockCh  chan BlockBody
```
Then `Listen` function can return this `blockCh`. The read side (`consensus/p2p/sync.go`) can use it.

rianhughes: I tried this approach before but ran into problems, I'll try it again

consensus/sync/sync.go
rianhughes: I've updated this code so that the proposal is stored first

consensus/sync/sync.go
rianhughes: Removed

p2p/p2p.go
rianhughes: Updated

consensus/sync/sync.go
rianhughes: The code was updated, and this comment is no longer relevant

consensus/sync/sync.go
rianhughes: The code was updated, and this comment is no longer relevant

p2p/sync/sync.go
rianhughes: The code was updated, and this comment is no longer relevant

p2p/sync/sync.go
rianhughes: The code was updated, and this comment is no longer relevant

p2p/sync/sync.go
rianhughes: The code was updated, and this comment is no longer relevant

consensus/sync/sync.go
rianhughes: `defer cancel()` also works, I updated the code.

PR #2965, opened 2025-07-08
fix: when pending data is invalid return empty pending/pre_confirmed for latest

EgeCaner:
It is possible for receiving `ErrPendingBlockNotFound`, even though pending data is not nil. Accessor for pending data checks if pending data complies with semantics if not returns `ErrPendingBlockNotFound`. In between pending data is fetched and latest header is fetched there is chance that latest is updated by new block thus invalidates the pending data we fetched. In such cases we will return empty pending data for latest header

### DIFF
--- a/rpc/v9/subscriptions_test.go
+++ b/rpc/v9/subscriptions_test.go
@@ -354,11 +354,10 @@ func TestSubscribeTxnStatus(t *testing.T) {
 		mockChain.EXPECT().TransactionByHash(txHash).Return(nil, db.ErrKeyNotFound)
 		mockSyncer.EXPECT().PendingData().Return(nil, sync.ErrPendingBlockNotFound).Times(2)
 		mockChain.EXPECT().HeadsHeader().Return(nil, db.ErrKeyNotFound).Times(2)
-		fmt.Print("HEre check 0\n")
+
 		id, conn := createTestTxStatusWebsocket(t, handler, txHash)
-		fmt.Print("HEre check 0.1\n")
+
 		assertNextTxnStatus(t, conn, id, txHash, TxnStatusReceived, UnknownExecution, "")
-		fmt.Print("HEre check 0.2\n")
 		// Candidate Status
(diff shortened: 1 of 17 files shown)


rpc/v9/subscriptions_test.go
```diff
@@ -370,12 +371,11 @@ func TestSubscribeTxnStatus(t *testing.T) {
...
 			&pendingData,
 			nil,
-		).Times(2)
-		mockSyncer.EXPECT().PendingBlock().Return(nil)
+		).Times(4)
 		handler.preConfirmedFeed.Send(&core.PreConfirmed{
 			Block: &core.Block{Header: &core.Header{}},
 		})
-
+		fmt.Print("HEre check 1\n")
```
rodrodros: Rogue print

rpc/v9/subscriptions_test.go
```diff
@@ -353,11 +352,13 @@ func TestSubscribeTxnStatus(t *testing.T) {
 		require.Nil(t, addErr)
 		txHash := addRes.TransactionHash
 		mockChain.EXPECT().TransactionByHash(txHash).Return(nil, db.ErrKeyNotFound)
-		mockSyncer.EXPECT().PendingData().Return(nil, sync.ErrPreConfirmedBlockNotFound)
-		mockSyncer.EXPECT().PendingBlock().Return(nil)
+		mockSyncer.EXPECT().PendingData().Return(nil, sync.ErrPendingBlockNotFound).Times(2)
+		mockChain.EXPECT().HeadsHeader().Return(nil, db.ErrKeyNotFound).Times(2)
+		fmt.Print("HEre check 0\n")
```
rodrodros: Rogue print

rodrigo-pino (approved):

rpc/v9/subscriptions_test.go
EgeCaner: removed

rpc/v9/subscriptions_test.go
EgeCaner: removed

wojciechos (approved):

PR #2966, opened 2025-07-08
tests: Add extra tests to check the case when a rule return nil action

infrmtcs:


consensus/tendermint/rule_proposal_and_polka_current_test.go
```diff
@@ -32,6 +32,43 @@ func TestProposalAndPolkaCurrent(t *testing.T) {
 		assertState(t, stateMachine, types.Height(0), types.Round(0), types.StepPrecommit)
 	})
 
+	t.Run("Line 36: lock and broadcast precommit when step is precommit", func(t *testing.T) {
```
rianhughes, line 35: Why did you add this test? I'm all for it. But it seems like you were concerned that line 36 would block line 49?

rianhughes (approved):

consensus/tendermint/rule_proposal_and_polka_current_test.go
infrmtcs: Yes, that's correct. I want to make sure that if line 36 return nil action, line 49 can still trigger.

PR #2973, opened 2025-07-11
Fix: GetNonce method in RPCv09 not working with 'pre_confirmed'

thiagodeev:
- Fix the issue with the GetNonce method not working with 'pre_confirmed' block tag
- Add a user-friendly warning when a user passes a wrong block tag


wojciechos:
PR to v0.14.0-beta here: https://github.com/NethermindEth/juno/pull/2975

PR #2978, opened 2025-07-14
refactor: Turn mempool.Pool into an interface

infrmtcs:
To prepare to make `consensus/proposer.Proposer` a drop-in replacement for mempool when consensus mode is used.

### DIFF
--- a/consensus/proposer/proposer.go
+++ b/consensus/proposer/proposer.go
@@ -37,8 +37,9 @@ type (
 type Proposer[V types.Hashable[H], H types.Hash] interface {
 	service.Service
 	tendermint.Application[V, H]
+	mempool.Pool
 	OnCommit(context.Context, types.Height, V)
-	Submit(transactions []mempool.BroadcastedTransaction)
+	Submit(context.Context, []mempool.BroadcastedTransaction)
 	Pending() *sync.Pending
 }
 
@@ -138,8 +139,16 @@ func (p *proposer[V, H]) Value() V {
 	return ask(p.valueTrigger, struct{}{}, p.lastValue)
 }
 
-func (p *proposer[V, H]) Submit(transactions []mempool.BroadcastedTransaction) {
-	p.transactionReceiver <- transactions
+func (p *proposer[V, H]) Submit(ctx context.Context, transactions []mempool.BroadcastedTransaction) {
+	select {
+	case <-ctx.Done():
+	case p.transactionReceiver <- transactions:
+	}
+}
+
+func (p *proposer[V, H]) Push(ctx context.Context, transaction *mempool.BroadcastedTransaction) error {
+	p.Submit(ctx, []mempool.BroadcastedTransaction{*transaction})
+	return nil
 }
 
 // Return the pending block currently guarded by the atomic pointer. The implementation assumes that
--- a/mempool/mempool.go
+++ b/mempool/mempool.go
@@ -1,6 +1,7 @@
 package mempool
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"sync"
@@ -11,6 +12,10 @@ import (
 	"github.com/NethermindEth/juno/db"
 	"github.com/NethermindEth/juno/utils"
 )
+
+type Pool interface {
+	Push(context.Context, *BroadcastedTransaction) error
+}
 
 var (
 	ErrTxnPoolFull  = errors.New("transaction pool is full")
@@ -104,9 +109,9 @@ func (t *memTxnList) popBatch(numToPop int) ([]BroadcastedTransaction, error) {
 	return result, nil
 }
 
-// Pool represents a blockchain mempool, managing transactions using both an
+// SequencerMempool represents a blockchain mempool, managing transactions using both an
 // in-memory and persistent database.
-type Pool struct {
+type SequencerMempool struct {
 	log         utils.SimpleLogger
 	bc          blockchain.Reader
 	db          db.KeyValueStore // to store the persistent mempool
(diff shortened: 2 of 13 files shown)


mempool/mempool.go
```diff
@@ -228,7 +233,7 @@ func (p *Pool) writeToDB(userTxn *BroadcastedTransaction) error {
 }
 
 // Push queues a transaction to the pool
-func (p *Pool) Push(userTxn *BroadcastedTransaction) error {
+func (p *SequencerMempool) Push(ctx context.Context, userTxn *BroadcastedTransaction) error {
```
rianhughes, line 236: Why do you require a ctx here if its not being used?

mempool/mempool.go
```diff
@@ -12,6 +13,10 @@ import (
 	"github.com/NethermindEth/juno/utils"
 )
 
+type Pool interface {
+	Push(context.Context, *BroadcastedTransaction) error
+}
```
rianhughes, line 18: Is this the final interface definition? I imagine we would need Pop() and Remove() functions etc

mempool/mempool.go
infrmtcs: It should have been used, currently the consensus version is using it to select `ctx.Done()`.

mempool/mempool.go
infrmtcs: There are 2 sides of a mempool.
- The RPC endpoints only want to push a transaction to a mempool and nothing else, so they only expect the interface to have a Push function.
- On the other hand, the block producers don't have the same interface, because of the differences in implementation.
  + Sequencer has `Pop` and `Remove`
  + `Proposer` from consensus don't use it, but instead run a loop to re-insert the transactions to the next block if it's not used by the previous block.
  + Feeder gateway, although we haven't converted it to use this interface yet, doesn't need to do anything, because it's the responsibility of a remote server.

mempool/mempool.go
infrmtcs: The current implementation of the sequencer version is not in the scope of this PR. In fact, I think the ideal version of this function should accept a slice of transactions, so the node can bundle transactions in the future if it wants.

consensus/proposer/proposer.go
```diff
@@ -37,8 +37,9 @@ type (
 type Proposer[V types.Hashable[H], H types.Hash] interface {
 	service.Service
 	tendermint.Application[V, H]
+	mempool.Pool
```
rianhughes, line 40: Why did you add the mempool interface to the proposer? This results in the Proposer presenting two seperate entrypoints to submit transactions, and it's not clear which one should be used.

Also, if the mempool interface grows (eg to include, Pop, Remove, etc), then it will force Proposer to implement these.

consensus/proposer/proposer.go
```diff
@@ -138,8 +139,16 @@ func (p *proposer[V, H]) Value() V {
...
+	select {
+	case <-ctx.Done():
+	case p.transactionReceiver <- transactions:
+	}
+}
+
+func (p *proposer[V, H]) Push(ctx context.Context, transaction *mempool.BroadcastedTransaction) error {
+	p.Submit(ctx, []mempool.BroadcastedTransaction{*transaction})
+	return nil
 }
```
rianhughes, line 153: This doesn't seem to provide any benefit over just calling Submit directly

infrmtcs: This is because `Push` should accept multiple transactions (see this comment: https://github.com/NethermindEth/juno/pull/2978#discussion_r2215111416).
However, because of the current implementation of sequencer mempool doesn't allow it, `Push` can only accept 1 transaction at a time. In the future, we should replace `Push` by `Submit` when sequencer mempool supports it.

consensus/proposer/proposer.go
infrmtcs: Regarding `Submit` vs `Push`, please see my comment here: 
https://github.com/NethermindEth/juno/pull/2978#discussion_r2215146895

Regarding `Pop` and `Remove`, please see my comment here: 
https://github.com/NethermindEth/juno/pull/2978#discussion_r2215103742

consensus/proposer/proposer.go
rianhughes: > we should replace Push by Submit when sequencer mempool supports it.

It would be nice if we could do this now, or if not, create and issue to track it

mempool/mempool.go
rianhughes: Good point. It seems different pks require different functionality for the mempool. Do you think it would be cleaner to implement separate interfaces for these?


```
type RPCPool interface {
	Push(context.Context, *BroadcastedTransaction) error
}
type SequencerPool interface {
	RPCPool
	Pop()
	Remove()
	//..
}
type CosnensusPool interface {
	Push(context.Context, []*BroadcastedTransaction) error
	Reinsert()
	//...
}
```

consensus/proposer/proposer.go
rianhughes: Closing this comment as it overlaps with [this](https://github.com/NethermindEth/juno/pull/2978#discussion_r2222314033)

mempool/mempool.go
rianhughes: If we create a separate interface for the Sequencer and Consensus pkgs (as mentioned below), then we can avoid requiring Push() to accept a dead parameter. 

If we are going to keep this parameter here, we should track it in an issue imo

mempool/mempool.go
infrmtcs: I thought about this, but each of them has a different caller, so we can use the struct type for now. Although interface can help to hide the struct implementation, I don't think it's necessary in this case.

mempool/mempool.go
infrmtcs: I think we should make the Sequencer mempool select context and accept the transaction list, it's just outside the scope of this PR.

infrmtcs: I'll create an issue to track it.

rianhughes (approved):

PR #2979, opened 2025-07-14
fix: `starknet_getEvents` return empty response when `from_block` > `to_block`

EgeCaner:
After introducing `AggregatedBloomFilter` logic, `MatchedBlockIterator` was raising an error on initialization if `from_block` is greater than `to_block`. This PR adjust the iterator to return empty response when `from_block` > `to_block`.

### DIFF
--- a/blockchain/aggregated_bloom_filter_cache.go
+++ b/blockchain/aggregated_bloom_filter_cache.go
@@ -84,7 +84,6 @@ type MatchedBlockIterator struct {
 }
 
 var (
-	ErrInvalidBlockRange                = errors.New("fromBlock > toBlock")
 	ErrMaxScannedBlockLimitExceed       = errors.New("max scanned blocks exceeded")
 	ErrAggregatedBloomFilterFallbackNil = errors.New("aggregated bloom filter does not have fallback")
 	ErrFetchedFilterBoundsMismatch      = errors.New("fetched filter bounds mismatch")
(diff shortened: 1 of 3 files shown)


blockchain/aggregated_bloom_filter_cache.go
```diff
@@ -102,10 +101,6 @@ func (c *AggregatedBloomFilterCache) NewMatchedBlockIterator(
 	matcher *EventMatcher,
 	runningFilter *core.RunningEventFilter,
 ) (MatchedBlockIterator, error) {
-	if fromBlock > toBlock {
-		return MatchedBlockIterator{}, ErrInvalidBlockRange
-	}
-
```
rodrodros, line 108: Why are we moving away from this approach?

EgeCaner: This piece of logic was resulting in responding with error to user when we receive a query where `from_block` is greater than `to_block`. We were aiming to provide empty response for such case. We can either keep raising this error and handle the situation in caller, or we can just return already exhausted iterator. My approach was returning exhausted iterator which results in empty response. However handling this is caller is straight forward too. Would you prefer to keep raising the error?

rodrodros: I don't have an opinion on what is best. Happy to go with your approach

PR #2982, opened 2025-07-14
tests: implement p2p sync tests

rianhughes:
This PR implements a test for the P2P sync pkg. The test involves syncing a single empty block from a peer.


p2p/sync/sync_test.go
```diff
@@ -0,0 +1,188 @@
...
+
+	stateDiffsIter := NewStateDiffsResponseSeq()
+	client.EXPECT().RequestStateDiffs(gomock.Any(), &state.StateDiffsRequest{Iteration: iter}).Return(stateDiffsIter, nil)
+
+	ctx, cancel := context.WithCancel(t.Context())
+	go func() {
+		time.Sleep(10 * time.Millisecond)
+		cancel()
+	}()
+	sync.Run(ctx)
```
infrmtcs, line 66: Error is not asserted

p2p/sync/sync_test.go
```diff
@@ -0,0 +1,188 @@
...
+
+func TestSyncEmptyBlock(t *testing.T) {
+	mockCtrl := gomock.NewController(t)
+	t.Cleanup(mockCtrl.Finish)
+	testDB := memory.New()
+	network := &utils.Mainnet
+	bc := blockchain.New(testDB, network)
+	host := mocks.NewMockHost(mockCtrl)
+	sync := New(bc, host, &utils.Sepolia, utils.NewNopZapLogger())
+	client := mocks.NewMockClient(mockCtrl)
```
infrmtcs, line 34: I'm strongly against using mock here because by using mock, it means there are a lot of things untested:
- Non empty blocks.
- p2p initialization is done correctly, given correct configuration (because host is mocked)
- Server side: Query from database
- Server side, client side: Block streaming
- Server side, client side: Data serialization/deserialization with protobuf

I don't mean to insist that we need to write the tests for all of these above in this PR. What I meant is that instead of using mock here, we can at least have a simple test not using mock but cover most of the cases above. The idea is:
- Start 2 p2p hosts, connect to each other via libp2p. Examples can be found in consensus tests.
- One host plays the role of server. Store N blocks to its database.
- The other host plays the role of client. Run the sync protocol and validate whether it can fetch all N blocks from the server host.
- Assertions can be done easily but throughly, by verifying whether the stored blocks in client and server databases match exactly.

p2p/sync/sync_test.go
```diff
@@ -0,0 +1,188 @@
...
+
+	eventsIter := NewEventsResponseSeq()
+	client.EXPECT().RequestEvents(gomock.Any(), &event.EventsRequest{Iteration: iter}).Return(eventsIter, nil)
+
+	stateDiffsIter := NewStateDiffsResponseSeq()
+	client.EXPECT().RequestStateDiffs(gomock.Any(), &state.StateDiffsRequest{Iteration: iter}).Return(stateDiffsIter, nil)
+
+	ctx, cancel := context.WithCancel(t.Context())
+	go func() {
+		time.Sleep(10 * time.Millisecond)
```
infrmtcs, line 63: We should use `eventually` with the assertions below instead.

rodrigo-pino:
We will re-do this PR avoiding the use of Mocs

PR #2985, opened 2025-07-15
fix: pending state access `PendingBlockNotFound`

EgeCaner:
sync.PendingData() returns `sync.ErrPendingBlockNotFound` either when the data is `nil` or when the cached pending data is outdated and no longer represents the current pending/pre-confirmed state. Juno updates the pending data after fetching each block, but it’s possible to encounter validation failures if PendingData is updated after a read yet before fetching the latest header for validation. This issue has been addressed before for PendingData access in wrappers, but this was missed in `PendingState` accesses. With these changes, we will use latest state if our pending data is invalid

### DIFF
--- a/rpc/v9/pending_data_wrapper.go
+++ b/rpc/v9/pending_data_wrapper.go
@@ -48,6 +48,18 @@ func (h *Handler) PendingBlock() *core.Block {
 	return pending.GetBlock()
 }
 
+func (h *Handler) PendingState() (core.StateReader, func() error, error) {
+	state, closer, err := h.syncReader.PendingState()
+	if err != nil {
+		if errors.Is(err, sync.ErrPendingBlockNotFound) {
+			return h.bcReader.HeadState()
+		}
+		return nil, nil, err
+	}
+
+	return state, closer, nil
+}
+
 func (h *Handler) PendingBlockFinalityStatus() TxnFinalityStatus {
 	pending, err := h.PendingData()
 	if err != nil {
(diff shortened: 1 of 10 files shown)


rpc/v9/pending_data_wrapper.go
```diff
@@ -48,6 +48,18 @@ func (h *Handler) PendingBlock() *core.Block {
...
+	state, closer, err := h.syncReader.PendingState()
+	if err != nil {
+		if errors.Is(err, sync.ErrPendingBlockNotFound) {
+			return h.bcReader.HeadState()
+		}
+		return nil, nil, err
+	}
+
+	return state, closer, nil
+}
```
rodrodros, line 61: No need to check for the pending data variant in v9?

Let's say someone would deploy this branch to mainnet (currently on 0.13.6) and try to use RPC v9, will it work as expected?

rodrigo-pino (approved):

rpc/v9/pending_data_wrapper.go
EgeCaner: In RPCv9, when starknet version < 0.14.0, we are serving the data from pending block and for state we use the state after executing the pending block. While we are still having pending block it should be okay to serve it in RPCv9 since it has harder finality status `ACCEPTED_ON_L2` than pre-confirmed block with finality status `PRE_CONFIRMED`.

EgeCaner: So when network is at < `0.14.0`. Calling `estimate_fee` or `get_nonce` on `RPCv9` with `pre_confirmed` ID will be equavalent to calling the same methods in `RPCv8` with `pending` ID

PR #2986, opened 2025-07-16
fix: Fix test genesis constructor args and VM serialization

infrmtcs:


vm/transaction.go
```diff
@@ -133,8 +133,8 @@ func adaptTransaction(txn core.Transaction) *Transaction {
 		if tx.Version.Uint64() == 3 {
 			tx.ResourceBounds = utils.HeapPtr(adaptResourceBounds(t.ResourceBounds))
 			tx.Tip = new(felt.Felt).SetUint64(t.Tip)
-			tx.PaymasterData = &t.PaymasterData
-			tx.AccountDeploymentData = &t.AccountDeploymentData
+			tx.PaymasterData = nilToZero(t.PaymasterData)
```
infrmtcs, line 136: If `PaymasterData` and `AccountDeploymentData` is `nil`, the marshaled JSON will set these 2 fields as explicit null, which is not unmarshallable on the rust side. This fix is to ensure that they're replaced by empty slices if nil.

genesis/genesis_prefund_accounts.json
```diff
@@ -37,22 +37,19 @@
 		"0x101": {
 			"class_hash": "0x61dac032f228abef9c6626f995015233097ae253a7f72d68552db02f2971b8f",
 			"constructor_args": [
-				"0x16d03d341717ab11083a481f53278d8e54f610af815cbdab4035b2df283fcc0",
-				"0x0"
+				"0x16d03d341717ab11083a481f53278d8e54f610af815cbdab4035b2df283fcc0"
```
infrmtcs, line 40: This class is OZ account contract, which only accepts an argument that is the public key. Adding the second element here makes the contract failing signature verification.

vm/transaction.go
```diff
@@ -208,3 +208,10 @@ func adaptResourceBounds(rb map[core.Resource]core.ResourceBounds) map[Resource]
 	}
 	return rpcResourceBounds
 }
+
+func nilToZero(v []*felt.Felt) *[]*felt.Felt {
+	if v == nil {
+		return &[]*felt.Felt{}
+	}
+	return &v
+}
```
rodrodros, line 217: I think we had this utility somewhere already?

infrmtcs: The other one is also a private function and converting from `*felt.Felt` to `*felt.Felt`, which is slightly different from what we want to do here.

rianhughes (approved):

PR #2988, opened 2025-07-16
ignore starknet go 'TestSubscribePendingTransactions'

EgeCaner:


.github/workflows/starknet-go-tests.yml
```diff
@@ -61,4 +61,4 @@ jobs:
       - name: Run RPC tests
         run: |
           cd rpc
-          go test -timeout 1200s -v -env testnet .
\ No newline at end of file
+          go test -skip 'TestSubscribePendingTransactions' -timeout 1200s -v -env testnet .
```
rodrodros: Can we add a comment explaining why we are skipping this test

EgeCaner: Sure I will! I was trying to test something, but realized this should be tested on top of polling PR.I also pushed the changes there and after finalizing, I will update both PRs.

rodrodros: Thank you!

.github/workflows/starknet-go-tests.yml
```diff
@@ -61,4 +61,8 @@ jobs:
       - name: Run RPC tests
         run: |
           cd rpc
-          go test -timeout 1200s -v -env testnet .
\ No newline at end of file
+          # `TestSubscribePendingTransactions` is ignored for starknet version > 0.14.0.
+          #  In starknet version 0.14.0 `pending` block is replaced by 'pre_confirmed' block, 
+          #  thus there will be no more pending transcations with `pending` semantics.
+          TESTS_TO_SKIP="TestSubscribePendingTransactions"
```
infrmtcs: Should use env and working-directory fields for step instead.

.github/workflows/starknet-js-tests.yml
```diff
@@ -43,7 +43,7 @@ jobs:
           fi
       
       - name: Run tests (${{ inputs.test_mode }} mode)
-        run: npm test -- ${{ steps.tests.outputs.tests }} --testNamePattern="^(?!.*(getSyncingStats|traceTransaction)).*$"
+        run: npm test -- ${{ steps.tests.outputs.tests }} --testNamePattern="^(?!.*(getSyncingStats|traceTransaction|waitForBlock pending)).*$"
```
infrmtcs, line 49: It's great if you can copy the reason from PR description to a comment here explaining why we ignore `waitForBlock`.

infrmtcs (approved):
LGTM!

PR #2989, opened 2025-07-17
feat(rpc): update cache transaction performance

rianhughes:
This PR updates the txn cache logic. Benchmark tests suggests a significant speed up (stats given below). 

Some benchmark results are (more details [here](https://github.com/NethermindEth/juno/compare/main...rianhughes/cache-v1#diff-a64bc115b86a094c59ed21a5ab206adf73b62eeded460e1d1585eeb6c1faeda9R12-R15)):
```
BenchmarkCacheOriginal-24        1   2776591293 ns/op    989802040 B/op    21610 allocs/op
BenchmarkCache-24              513      2388190 ns/op            0 B/op        0 allocs/op
```

The TLDR for why the previous solution was so inefficient is because every read resulted in a new array of len(keys) being allocated under the hood. This is an O(n) write operation every read. The locking mechanism was also not ideal (single lock,  reads and writes lock each other, etc).

See the doc string for the `TxnCache` type for more details on the new approach. Deletes are separated from reads and writes, and are made in place. We never allocate new memory, unless the map needs re-sized, which can be avoided if we specify the appropriate size up front. Each bucket has it's own Read-Write lock.


rpc/rpccore/txn_cache_test.go
```diff
@@ -0,0 +1,73 @@
...
+		k := *new(felt.Felt).SetUint64(uint64(i))
+		cache.Add(k)
+	}
+
+	// Verify all keys are present
+	for i := 0; i < nKeys; i++ {
+		k := *new(felt.Felt).SetUint64(uint64(i))
+		require.Truef(t, cache.Contains(k),
+			"expected key %d to be present after bulk insert", i)
+	}
```
rodrodros: Nitpick: another formatting either the
```go
x := f(
     arg1, arg2, ..., argn
);
```

or 
```go
x := f(
    arg1,
    arg2,
    ...,
    argn,
)
```

rpc/rpccore/txn_cache.go
```diff
@@ -0,0 +1,107 @@
...
+
+	"github.com/NethermindEth/juno/core/felt"
+)
+
+const (
+	mapCapacityHint = 1024  // Assuming 1024 TPS
+	NumTimeBuckets  = 5 + 1 // TTL is 5s. 1 additional bucket for outdated entries.
+)
+
+type TxnCache struct {
```
rodrigo-pino, line 17: Some description here of what's the idea behind this. I am looking forward to understanding it

rpc/rpccore/txn_cache.go
```diff
@@ -0,0 +1,107 @@
...
+	return cache
+}
+
+// Call this _before_ Run.
+func (c *TxnCache) WithTicker(tickC <-chan time.Time) *TxnCache {
+	c.tickC = tickC
+	return c
+}
+
+// Call this _before_ Run.
```
rodrodros: unintended comment

rodrigo-pino (commented):
Quick scan and light comments. Will come back when is marked as ready for review

rpc/rpccore/txn_cache.go
```diff
@@ -0,0 +1,127 @@
...
+)
+
+// TxnCache provides a time-bucketed cache for tracking transaction entries with TTL-based eviction.
+//
+// The cache divides time into 3 fixed-size buckets (NumTimeBuckets), each covering one TTL interval.
+// For example, with TTL = 5 minutes, and current time = 12 minutes:
+//
+//   - Bucket 1 stores entries received between 0–5 min (expired).
+//   - Bucket 2 stores entries from 5–10 min (mix of valid and expired).
+//   - Bucket 3 stores entries from 10–15 min (valid).
```
rodrodros:
```suggestion
//   - Bucket 1 stores entries received between 0–5 min (valid).
//   - Bucket 2 stores entries from 5–10 min (mix of valid and expired).
//   - Bucket 3 stores entries from 10–15 min (expired).
```

It's like this right?

rpc/rpccore/txn_cache.go
```diff
@@ -0,0 +1,127 @@
...
+//
+// On each tick (i.e., every TTL interval), the active bucket index advances modulo 3.
+// For example, after a tick at time = 15 min:
+//   - Bucket 1 (previously 0–5 min) now stores entries received between 15–20 min, where all entries are now valid.
+//   - Bucket 2 still stores entries received between 5–10 min, but all entries are now completely expired.
+//   - Bucket 3 still stores entries received between 10–15 min, but the entries can be either valid or expired.
```
rodrodros: This contradicts the previous statement. Why bucket 1 would have 15-20 min old transaction which are all valid?

Why bucket 2 stores completely expired transactions but in the previous bullet points it is said they are a mixed of expired and valid.

Similarly, why bucket 3 holds a mix of expired and valid, but they should be all expired

rpc/rpccore/txn_cache.go
```diff
@@ -0,0 +1,127 @@
...
+//
+// The cache behaviour is:
+//   - `Add()` inserts an entry into the current active bucket (latest time window).
+//   - `Contains()` checks the two most recent buckets (excluding the oldest).
+//   - On each tick, the evictor deletes the oldest bucket in-place to reclaim memory.
+//
```
rodrodros: I believe now I get a better sense. I believe you can improve the previous explanations, something along the lines of: "This cache holds a rotating list of buckets each representing a 5 minute time slot. After each time tick a different bucket will activate, if we were before in bucket `i` we will move to bucket `i + 1 % 3`. It will get deleted and prepared to hold all the transaction sent this way"

Just a draft but the basic idea is this right?

rpc/rpccore/txn_cache.go
```diff
@@ -0,0 +1,127 @@
...
+// This design enables efficient TTL-based eviction without scanning the entire cache,
+// ensuring high-throughput insertion and lookup with bounded memory usage.
+type TxnCache struct {
```
rodrodros: Please rename it to `TransactionCache`. I am against the name reduction because sometimes it is called `Tx` and others `Txn` making it inconsistent

rpc/rpccore/txn_cache.go
```diff
@@ -0,0 +1,127 @@
...
+	return (cur + 1) % NumTimeBuckets
+}
+
+func (c *TxnCache) incrementTimeSlot() {
+	old := atomic.LoadUint32(&c.curTimeBucket)
+	next := (old + 1) % NumTimeBuckets
+	atomic.StoreUint32(&c.curTimeBucket, next)
+}
+
+func (c *TxnCache) Add(key felt.Felt) {
```
rodrodros: `key` by reference

rpc/rpccore/txn_cache.go
```diff
@@ -0,0 +1,127 @@
...
+	if c.Contains(key) {
+		return
+	}
+	timeSlot := c.getCurTimeSlot()
+	c.locks[timeSlot].Lock()
+	c.buckets[timeSlot][key] = time.Now()
+	c.locks[timeSlot].Unlock()
+}
+
+func (c *TxnCache) Contains(key felt.Felt) bool {
```
rodrodros: `key` by reference

rpc/rpccore/txn_cache.go
```diff
@@ -0,0 +1,127 @@
...
+// Call this _before_ Run. Should only be used for tests.
+func (c *TxnCache) WithTicker(tickC <-chan time.Time) *TxnCache {
+	c.tickC = tickC
+	return c
+}
+
+func (c *TxnCache) Run(ctx context.Context) error {
+	c.evictor(ctx)
+	return nil
+}
```
rodrodros: This function should have an empty return type

rpc/rpccore/txn_cache.go
```diff
@@ -0,0 +1,127 @@
...
+
+func (c *TxnCache) incrementTimeSlot() {
+	old := atomic.LoadUint32(&c.curTimeBucket)
+	next := (old + 1) % NumTimeBuckets
+	atomic.StoreUint32(&c.curTimeBucket, next)
+}
+
+func (c *TxnCache) Add(key felt.Felt) {
+	if c.Contains(key) {
+		return
```
rodrodros: Shouldn't we update the time slot here? 

I don't like silently not doing anything when we re-add, we should either: 
1. update the time stamp
2. panic since it is not an expected behaviour

I believe it should be 1

rpc/rpccore/txn_cache.go
rodrodros, line 1: I see this type has no clear mechanism for deleting entries, I think it is not hard to implement either. 

One of the things the cache should do is to remove a transaction that we have a known status for, like `PRE_CONFIRMED`

rpc/rpccore/txn_cache_test.go
```diff
@@ -0,0 +1,71 @@
...
+func TestBulkSetContains(t *testing.T) {
+	const nKeys = 2048
+
+	const ttl = time.Second
+
+	cache := rpccore.NewTxnCache(ttl, 1024)
+	go func() {
+		err := cache.Run(t.Context())
+		require.NoError(t, err)
+	}()
```
rodrodros, line 22: You can call cache.Run at the end after `Add` and `Contain` Also I don't like to depend time for running.

I think this test can refactored to:
1. Execute the `Add`
2. Check the `Contains`
3. Execute the `Run`
4. Send an artificial time to trigger the eviction
5. Wait 100 miliseconds or something (it might be unavoidable)
6. Check that eviction was succesfull.

rpc/rpccore/txn_cache_test.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+	time.Sleep(ttl)
+	for i := range nKeys {
+		k := *new(felt.Felt).SetUint64(uint64(i))
+		require.Falsef(
+			t,
+			cache.Contains(k),
+			"expected key %d to be deleted after ttl was hit",
+			i)
+	}
```
rodrodros: I would prefer this to be a separate test. Also I would like avoiding the time.Sleep(), is better to send a time duration to trigger the eviction. Otherwise this tests look very flaky to me and dependent of computer hardware (or load)

rpc/rpccore/txn_cache_test.go
rodrodros, line 1: Finally, I would like for it to have another set of tests where we check for contains for transaction that are in the mixed bucket of valid and expired transactions

rpc/rpccore/txn_cache.go
```diff
@@ -0,0 +1,127 @@
...
+	cache.tickC = time.NewTicker(ttl).C
+	cache.ttl = ttl
+	return cache
+}
+
+// Call this _before_ Run. Should only be used for tests.
+func (c *TxnCache) WithTicker(tickC <-chan time.Time) *TxnCache {
+	c.tickC = tickC
+	return c
+}
```
rodrodros: I am against the idea of having production code just for the sake of testing. Also, was thinking on how to do it, but I cannot find a solution that is objectively better than this one, so we might as well live with this one :( 

The only one I am thinking is passing `tickC` as part of the constructor too. I would think I've slightly prefer that. I am happy to hear your thoughs

rpc/rpccore/txn_cache.go
rianhughes: Yes, this is correct. I'll try and make the description clearer

rpc/rpccore/txn_cache.go
rianhughes: The evictor routine is responsible for evicting entries. 

I decided against deleting living entries, since it adds unnecessary pressure to the cache. If the FGW updates its txn status to something other than `NOT_RECEIVED`, then we can bypass the cache entirely. That entry will eventually get deleted (assuming we actually have it). This saves us from having to perform a `read` and `delete` operation (again for txns we may not even have), and also saves lock contention for txns that actually require the cache (new entries, and those that have txn status `NOT_RECEIVED`).

rpc/rpccore/txn_cache.go
rianhughes: The TxnCache is a service, so must implement the service interface

```
type Service interface {
	// Run starts the service and returns an error. Services should not return non-critical errors,
	// instead they should log and retry. Services should be blocking.
	Run(ctx context.Context) error
}
```

rpc/rpccore/txn_cache.go
rianhughes: > Shouldn't we update the time slot here?

Do you mean update `curTimeBucket`? Why would we do this?

> I don't like silently not doing anything when we re-add, we should either:

Okay, I've updated the code so that any call to Add will insert the entry into the latest time bucket. If the entry already exists it will be updated. If it is in the partial time bucket, then that entry will not be updated, but added to the latest time bucket. Contains will search the latest time bucket first, ensuring users get the latest data. I'm against deleting entries outside of the evictor loop.

rpc/rpccore/txn_cache.go
rodrodros: Thank you, I appreciate the examples above as well. I would think we should positioned them below the general idea explanation

rpc/rpccore/txn_cache.go
rodrodros, line 1: nitpick: I am not sure if the functions follow an ordering here, which if it exists, I am happy to know the rationale behind them.

I would like to propose to sort the methods in the following order:
1. Constructor
2. Public Methods
3. Private Methods

rpc/rpccore/txn_cache_test.go
rianhughes: I've updated the test so that it doesn't require sleeping for the ttl

rpc/rpccore/txn_cache.go
rodrodros: I see, it makes sense, thank you

rpc/rpccore/txn_cache_test.go
rianhughes: The sleeping for `ttl` has been removed. Let me know if you still want this in a separate test, I'm not sure what benefit it would provide over the current approach

rpc/rpccore/txn_cache_test.go
rianhughes: This will explicitly depend on the time btw, because of the `time.Since(insertTime) <= c.ttl` check in Contains(). But I can add this

rpc/rpccore/txn_cache.go
rianhughes: I personally don't mind this function, I think it serves its purpose quite well. 

Passing in both the `ttl` and `ticker` as constructor arguments can cause problems if they are not of the same `time.Duration`, and I'm not sure h ow to get around this.

We could export the ticker, but I think that's worse than having the above method.

rpc/rpccore/txn_cache.go
rianhughes: Updated

rpc/rpccore/txn_cache.go
rianhughes: Updated

rpc/rpccore/txn_cache.go
rianhughes: I think it's my own weird ordering hah. But I updated the ordering to follow your structure

rpc/rpccore/txn_cache.go
rianhughes: >15-20 min old transaction

They are entries received *between* minute 15 and minute 20 (with the current time = 15), not 15-20min in the past.

I've updated the description to be more concise, hopefully it causes less confusion.

rpc/rpccore/txn_cache.go
rianhughes: I've updated the description, let me know what you think

rpc/rpccore/txn_cache.go
thiagodeev: We could create a test constructor, something like `NewTestTransactionCache` (just as we do in the `feeder.NewTestClient`), that accepts a custom ticker as parameter and set it to `c.tickC`. That way we would avoid the need of this method for testing

```go
func NewTestTransactionCache(ttl time.Duration, cacheSizeHint uint, tickC <-chan time.Time) *TransactionCache {
	cache := &TransactionCache{}
	for b := range NumTimeBuckets {
		cache.buckets[b] = make(map[felt.Felt]time.Time, cacheSizeHint)
	}
	atomic.StoreUint32(&cache.curTimeBucket, 0)
	cache.tickC = tickC
	cache.ttl = ttl
	return cache
}
```

rpc/v6/transaction.go
```diff
@@ -712,13 +712,10 @@ func (h *Handler) TransactionStatus(ctx context.Context, hash felt.Felt) (*Trans
 		}
 
 		if h.submittedTransactionsCache != nil {
-			switch txStatus.FinalityStatus {
-			case starknet.NotReceived:
+			if txStatus.FinalityStatus == starknet.NotReceived {
```
thiagodeev: Checking the finalityStatus first, since the cache is only necessary if it's `NotReceived`. That way, we can avoid the second `if`.
WDYT?
If it makes sense, we could do the same with the other occurrences

``` go
if txStatus.FinalityStatus == starknet.NotReceived && h.submittedTransactionsCache != nil {
  if h.submittedTransactionsCache.Contains(&hash) {
	  txStatus.FinalityStatus = starknet.Received
  }
}
```

rpc/rpccore/txn_cache.go
```diff
@@ -0,0 +1,111 @@
...
+// - Bucket (curTimeBucket - 1) % 3: the previous bucket — entries may be valid or expired, depending on timestamp.
+// - Bucket (curTimeBucket    ) % 3: the current bucket — new entries are added here.
+// - Bucket (curTimeBucket + 1) % 3: the next bucket — cleared by the evictor-routine before reuse.
+//
+// On every tick (i.e., TTL duration), the index advances and the evictor clears the oldest bucket in-place.
+type TransactionCache struct {
+	ttl           time.Duration
+	buckets       [NumTimeBuckets]map[felt.Felt]time.Time // map[txn-hash]entry-time
+	locks         [NumTimeBuckets]sync.RWMutex
+	curTimeBucket uint32
```
thiagodeev: If the intention is to use it atomically, we could use the explicit `atomic.Uint32` type.
This will make things more obvious and the code cleaner.

``` diff
type TransactionCache struct {
	ttl           time.Duration
	buckets       [NumTimeBuckets]map[felt.Felt]time.Time // map[txn-hash]entry-time
	locks         [NumTimeBuckets]sync.RWMutex
-       curTimeBucket uint32
+	curTimeBucket atomic.Uint32
	tickC         <-chan time.Time
}

// ...

func (c *TransactionCache) getExpiredTimeSlot() uint32 {
-       cur := atomic.LoadUint32(&c.curTimeBucket)
+	cur := c.curTimeBucket.Load()
	return (cur + 1) % NumTimeBuckets
}

func (c *TransactionCache) incrementTimeSlot() {
-       old := atomic.LoadUint32(&c.curTimeBucket)
+	old := c.curTimeBucket.Load()
	next := (old + 1) % NumTimeBuckets
-       atomic.StoreUint32(&c.curTimeBucket, next)
+	c.curTimeBucket.Store(next)
}

```

rpc/v6/transaction.go
rianhughes: Yes, it's neater to collapse both if statements into one. I'll update the code

rpc/rpccore/txn_cache.go
rianhughes: Yes, nice suggestion. I've updated the code

rpc/v6/transaction.go
rianhughes: Given we don't allow the cache to be nil, it probably makes sense to check against the finality status first to save work. I'll update the code

thiagodeev (approved):

PR #2990, opened 2025-07-17
feat: blockifier upgrade

EgeCaner:
This PR includes:
- Blockifier update for starknet 0.14.0
- Fix: RPC version >= 8 traces to return inner execution resources
- Converting L2 gas to L1 gas on fee estimation for L1Handler txs
- Fundamental changes for supporting L1Handler traces to be revertable
  - Backwards compatibility changes to address the case where reverted L1Handler trace appears in a query from RPC version < 9.


vm/rust/src/jsonrpc.rs
```diff
@@ -50,7 +52,7 @@ pub struct TransactionTrace {
     #[serde(skip_serializing_if = "Option::is_none")]
     constructor_invocation: Option<FunctionInvocation>,
     #[serde(skip_serializing_if = "Option::is_none")]
-    function_invocation: Option<FunctionInvocation>,
+    function_invocation: Option<ExecuteInvocation>,
```
rodrodros, line 55: `ExecuteInvocation = Result<FunctionInvocation>`, I think the name is a bit confusing and maybe we can change it in the future. Just making a note of this here. Maybe `FunctionInvocationResult` or another one more original but still relatable

rodrigo-pino (approved):
Positive changes to the package and no real comment to add. Great job!

vm/rust/src/jsonrpc.rs
rodrodros: Issue at https://github.com/NethermindEth/juno/issues/2991

PR #2995, opened 2025-07-18
feat: rpcv9 init

EgeCaner:
This PR contains base for the RPCv9, currently most of rpcv9 code is copy of rpcv8, to facilitate the review process and make the changes clear, lets get this merge first then changes should come top of this to easily identify what has been changed from previous rpcs. This is copy of rpcv8 with minimal changes, mostly renaming to rpcv9, but rpcv8 not contain some methods we need for rpcv9 such as `Events` `Nonce` etc... they are added to rpcv9 handler but commented at the moment.

Also includes [changes](https://github.com/NethermindEth/juno/pull/2995/commits/cdd03a859497df1eb6cdf08618a08c2e72e1f1c3) regarding `L1_Handler` traces to return one of function invocation or revert reason.

### DIFF
--- a/rpc/v9/adapters.go
+++ b/rpc/v9/adapters.go
@@ -27,7 +27,7 @@ func AdaptVMTransactionTrace(trace *vm.TransactionTrace) TransactionTrace {
 
 	var constructorInvocation *FunctionInvocation
 	var executeInvocation *ExecuteInvocation
-	var functionInvocation *FunctionInvocation
+	var functionInvocation *ExecuteInvocation
 
 	switch trace.Type {
 	case vm.TxnDeployAccount, vm.TxnDeploy:
@@ -40,7 +40,7 @@ func AdaptVMTransactionTrace(trace *vm.TransactionTrace) TransactionTrace {
 		}
 	case vm.TxnL1Handler:
 		if trace.FunctionInvocation != nil {
-			functionInvocation = utils.HeapPtr(adaptVMFunctionInvocation(trace.FunctionInvocation))
+			functionInvocation = utils.HeapPtr(adaptVMExecuteInvocation(trace.FunctionInvocation))
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


rpc/handlers.go
```diff
@@ -96,11 +107,206 @@ func (h *Handler) Run(ctx context.Context) error {
...
+		{
+			Name:    "starknet_addDeclareTransaction",
+			Params:  []jsonrpc.Parameter{{Name: "declare_transaction"}},
+			Handler: h.rpcv9Handler.AddTransaction,
+		},
+		// {
+		// 	Name:    "starknet_getEvents",
+		// 	Params:  []jsonrpc.Parameter{{Name: "filter"}},
+		// 	Handler: h.rpcv9Handler.Events,
+		// },
```
infrmtcs, line 212: Why do we disable this and some of the above?

rpc/v9/adapters.go
```diff
@@ -0,0 +1,276 @@
...
+			constructorInvocation = utils.HeapPtr(adaptVMFunctionInvocation(trace.ConstructorInvocation))
+		}
+	case vm.TxnInvoke:
+		if trace.ExecuteInvocation != nil {
+			executeInvocation = utils.HeapPtr(adaptVMExecuteInvocation(trace.ExecuteInvocation))
+		}
+	case vm.TxnL1Handler:
+		if trace.FunctionInvocation != nil {
+			functionInvocation = utils.HeapPtr(adaptVMExecuteInvocation(trace.FunctionInvocation))
+		}
```
infrmtcs, line 44: Does this logic change from v8?

infrmtcs: I found the changes in the specs.

rpc/v9/block.go
```diff
@@ -0,0 +1,331 @@
...
+	pending blockIDType = iota + 1
+	latest
+	hash
+	number
+)
+
+func (b *blockIDType) String() string {
+	switch *b {
+	case pending:
+		return "pending"
```
infrmtcs, line 26: (can be next PR) Don't forget to rename this to `preconfirmed`.

infrmtcs (approved):

rpc/v9/block.go
rodrodros: I believe this is just a naive copy of rpc v8 with the changes of rpc v9 in a separate one so they are easier to spot

rpc/handlers.go
EgeCaner: I copied the rpcv9 from rpcv8 and rpcv8 were using rpcv6 under the hood for the commented out methods. Due to new BlockTag `PreConfirmed`, we have to re implement these methods in rpcv9. In the next PR I will push their implementation

PR #2998, opened 2025-07-21
feat: rpcv9 new block and transaction statuses

EgeCaner:
This PR introduces new Block and TransactionStatuses introduced in RPCv9.

Also partially implements `ws` endpoints. For transaction status subscription, we return the new statuses but for the rest of the methods we do not use data from `pre_confirmed` block

### DIFF
--- a/rpc/v9/class.go
+++ b/rpc/v9/class.go
@@ -7,6 +7,9 @@ import (
 	"github.com/NethermindEth/juno/adapters/sn2core"
 	"github.com/NethermindEth/juno/core"
 	"github.com/NethermindEth/juno/core/felt"
+	"github.com/NethermindEth/juno/jsonrpc"
+	"github.com/NethermindEth/juno/rpc/rpccore"
+	rpcv6 "github.com/NethermindEth/juno/rpc/v6"
 	"github.com/NethermindEth/juno/starknet"
 	"github.com/NethermindEth/juno/starknet/compiler"
 	"github.com/NethermindEth/juno/utils"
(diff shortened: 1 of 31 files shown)


rpc/v9/class.go
```diff
@@ -52,3 +55,104 @@ func adaptDeclaredClass(declaredClass json.RawMessage) (core.Class, error) {
...
+			Program: c.Program,
+			EntryPoints: rpcv6.EntryPoints{
+				// Note that utils.Map returns nil if provided slice is nil
+				// but this is not the case here, because we rely on sn2core adapters that will set it to empty slice
+				// if it's nil. In the API spec these fields are required.
+				Constructor: utils.Map(c.Constructors, adaptEntryPoint),
+				External:    utils.Map(c.Externals, adaptEntryPoint),
+				L1Handler:   utils.Map(c.L1Handlers, adaptEntryPoint),
+			},
+		}
```
rodrodros: I know this pattern appears in Juno other places, but I wanted to ask to not perpetuate it since functional programming is not memory friendly in Go (the compiler doesn't apply optimization for this type of function calls). 

Can we define a private function (not a lambda) called
```go
func adaptCairo0EntryPoint(arg []core.EntryPoints) []rpcv6.EntryPoint
```

Also, on a closer inspection you realize `rpcv6.EntryPoint` is a type meant to have two roles, both a Cairo Zero and modern Cairo entry point, which I think is not ideal. Can be done in a follow up but it would be a good readability improvement in my book

rpc/v9/class.go
```diff
@@ -52,3 +55,104 @@ func adaptDeclaredClass(declaredClass json.RawMessage) (core.Class, error) {
...
+			Abi:                  c.Abi,
+			SierraProgram:        c.Program,
+			ContractClassVersion: c.SemanticVersion,
+			EntryPoints: rpcv6.EntryPoints{
+				// Note that utils.Map returns nil if provided slice is nil
+				// but this is not the case here, because we rely on sn2core adapters that will set it to empty slice
+				// if it's nil. In the API spec these fields are required.
+				Constructor: utils.Map(c.EntryPoints.Constructor, adaptEntryPoint),
+				External:    utils.Map(c.EntryPoints.External, adaptEntryPoint),
+				L1Handler:   utils.Map(c.EntryPoints.L1Handler, adaptEntryPoint),
```
rodrodros: Let's also have a dedicated function for these here

rpc/v9/subscriptions.go
```diff
@@ -538,6 +551,44 @@ func (h *Handler) onPendingBlock(
...
+// If getDetails is false, response will only contain the transaction hashes.
+//
+//nolint:unused // retained for future use in 0.9.0-rc3, currently unused
+func (h *Handler) onPreConfirmedBlock(
+	id string,
+	w jsonrpc.Conn,
+	getDetails bool,
+	senderAddr []felt.Felt,
+	preConfirmed *core.Block,
+	lastBlockNumber *uint64,
```
rodrodros, line 548: I am not sure if I like the idea of passing an argument (i.e. `lastBlockNumber`) by reference and then modifying it. Also given the function I am also not sure the best approach is  to pass it by value and include it in the return form. 

Is it possible to have another design 🤔  (as long as it is not a headache to change it). Otherwise let's keep it for now but then add it to the function description that this function modifies certain inputs (like `lastBlockNumber`).

rpc/v9/class.go
rodrodros: Added an issue for the latter https://github.com/NethermindEth/juno/issues/3001

rpc/v9/subscriptions.go
EgeCaner: I can return the last block number from the function and handle in caller

EgeCaner: This function is unused in the moment. High likely there will be 2 new methods for `pre_confirmed` block;
- `starknet_subscribePreConfirmedTransactions`
- `starknet_subscribePreConfirmedEvents`

We can handle this issue while implementing those, we will have clarity in requirements once spec is finalized

rodrodros: Ok, let's leave it as is for now, and go back to it later

rpc/v9/nonce.go
```diff
@@ -0,0 +1,30 @@
...
+
+/****************************************************
+		Nonce Handler
+*****************************************************/
+
+// Nonce returns the nonce associated with the given address in the given block number
+//
+// It follows the specification defined here:
+// https://github.com/starkware-libs/starknet-specs/blob/25533f8b7999219120a4a654709d47dc9376d640/api/starknet_api_openrpc.json#L851
+func (h *Handler) Nonce(id *BlockID, address felt.Felt) (*felt.Felt, *jsonrpc.Error) {
```
rodrodros: `address` by reference please

rodrigo-pino (approved):
Looks good

rpc/v9/transaction_test.go
```diff
@@ -1949,12 +2020,194 @@ func TestSubmittedTransactionsCache(t *testing.T) {
...
 		require.Equal(t, rpccore.ErrTxnHashNotFound, err)
-		require.Nil(t, status)
+		require.Empty(t, status)
 	})
 }
+
+func getTestPreConfirmed(t *testing.T) core.PreConfirmed {
+	t.Helper()
+
+	preConfirmedJSONStr := `{
```
infrmtcs: Is it possible to put this in testdata folder?

rpc/v9/subscriptions.go
```diff
@@ -254,16 +258,22 @@ func (h *Handler) SubscribeTransactionStatus(ctx context.Context, txHash *felt.F
 	var err error
 
 	subscriber := subscriber{
-		onStart: func(ctx context.Context, id string, sub *subscription, _ any) error {
+		onStart: func(
+			ctx context.Context, id string, sub *subscription, _ any,
```
infrmtcs: nit: feel free to ignore this, personally I think we should use either the old style or this style:

```suggestion
			ctx context.Context,
			id string,
			sub *subscription,
			_ any,
```

rpc/v9/subscriptions.go
```diff
@@ -492,6 +504,7 @@ func (h *Handler) SubscribePendingTxs(ctx context.Context, getDetails *bool, sen
 			return nil
 		},
 		onPendingData: func(ctx context.Context, id string, _ *subscription, pending core.PendingData) error {
+			// We should only return `ACCEPTED_ON_L2` txs in this stream.
 			if pending == nil || pending.Variant() != core.PendingBlockVariant {
```
infrmtcs, line 492: Do we need this check anymore? I saw a similar check removed above.

rpc/v9/subscriptions.go
```diff
@@ -471,7 +483,7 @@ func (h *Handler) SubscribeNewHeads(ctx context.Context, blockID *SubscriptionBl
 // SubscribePendingTxs creates a WebSocket stream which will fire events when a new pending transaction is added.
 // The getDetails flag controls if the response will contain the transaction details or just the transaction hashes.
 // The senderAddr flag is used to filter the transactions by sender address.
-func (h *Handler) SubscribePendingTxs(ctx context.Context, getDetails *bool, senderAddr []felt.Felt) (SubscriptionID, *jsonrpc.Error) {
+func (h *Handler) SubscribePendingTxs(ctx context.Context, getDetails bool, senderAddr []felt.Felt) (SubscriptionID, *jsonrpc.Error) {
```
infrmtcs, line 470: `getDetails` is optional, so please verify if it works if input doesn't contain this field.

rpc/v9/subscriptions.go
EgeCaner: Yes, for `pre_confirmed` block we will have, new methods for preConfirmedTx subscription and for pre_confirmed events. We still have these subscription endpoints but they should not serve data from pre_confirmed block, to have clean stream of `ACCEPTED_ON_L2` data

rpc/v9/nonce.go
```diff
@@ -0,0 +1,30 @@
...
+
+/****************************************************
+		Nonce Handler
+*****************************************************/
+
+// Nonce returns the nonce associated with the given address in the given block number
+//
+// It follows the specification defined here:
+// https://github.com/starkware-libs/starknet-specs/blob/25533f8b7999219120a4a654709d47dc9376d640/api/starknet_api_openrpc.json#L851
+func (h *Handler) Nonce(id *BlockID, address *felt.Felt) (*felt.Felt, *jsonrpc.Error) {
```
infrmtcs, line 17: Please be careful here, `address` is required, so turning it to reference will accept nil here.

infrmtcs: cc @rodrigo-pino

rpc/v9/subscriptions.go
infrmtcs: I'm understanding that `*lastBlockNumber` will be the same as `preConfirmed.Number` after the call, so probably the caller can use `preConfirmed.Number` directly instead.
The only thing we need `lastBlockNumber` here is to clear the `sendTxHashes` map.

rpc/v9/nonce.go
EgeCaner: Im getting this response when I omit the address field, looks like it is fine. returns same error for previous rpcs
```json
{
    "jsonrpc": "2.0",
    "method": "starknet_getNonce",
    "params": {
        "block_id": "latest"
    },
    "id": 1
}
```
response
```json
{
    "jsonrpc": "2.0",
    "error": {
        "code": -32602,
        "message": "Invalid Params",
        "data": "missing non-optional param: contract_address"
    },
    "id": 1
}
```

rpc/v9/subscriptions.go
EgeCaner: I check its defaulted to false when not provided

PR #3011, opened 2025-07-24
refactor: Implement vote counter

### LINKED ISSUE #2715: [Tendermint] Optimize O(n) operations

infrmtcs:
Close #2715


consensus/integtest/node.go
```diff
@@ -18,7 +18,7 @@ func (n nodes) TotalVotingPower(height types.Height) types.VotingPower {
 	return types.VotingPower(len(n))
 }
 
-func (n nodes) ValidatorVotingPower(addr starknet.Address) types.VotingPower {
+func (n nodes) ValidatorVotingPower(height types.Height, addr starknet.Address) types.VotingPower {
 	return types.VotingPower(1)
 }
```
rodrodros: Can voting power change through heights I guess 🤔 .

Since this values are meant to be read only, what do you think of passing them as references

consensus/tendermint/broadcast.go
```diff
@@ -20,7 +20,7 @@ func (t *stateMachine[V, H, A]) sendProposal(value *V) types.Action[V, H, A] {
 		t.log.Fatalf("Failed to store propsal in WAL")
 	}
 
-	t.messages.AddProposal(proposalMessage)
+	t.voteCounter.AddProposal(&proposalMessage)
```
rodrodros, line 23: What's the idea behind this name change. Out of curiosity

consensus/tendermint/rule_polka_nil.go
```diff
@@ -12,17 +15,7 @@ Check the upon condition on line 44:
 Line 36 and 44 for a round are mutually exclusive.
 */
 func (t *stateMachine[V, H, A]) uponPolkaNil() bool {
```
rodrodros, line 17: nit: is this `t` a leftover when this was called Tendermint. What do you think of giving it an update name. It's the most nitpicks of nitpicks and to do on a separate PR. Otherwise, happy leaving the `t`

consensus/tendermint/tendermint_test.go
```diff
@@ -41,7 +40,7 @@ func (v *validators) TotalVotingPower(h types.Height) types.VotingPower {
 	return v.totalVotingPower
 }
 
-func (v *validators) ValidatorVotingPower(validatorAddr starknet.Address) types.VotingPower {
+func (v *validators) ValidatorVotingPower(h types.Height, validatorAddr starknet.Address) types.VotingPower {
```
rodrodros: It's a test package so less prio, but `validatorAddress` by ref

consensus/votecounter/ballot.go
```diff
@@ -0,0 +1,41 @@
...
+import "github.com/NethermindEth/juno/consensus/types"
+
+type VoteType uint8
+
+const (
+	Prevote VoteType = iota
+	Precommit
+)
+
+type ballot []bool
```
rodrodros: Are ballots a dynamic slice, or can their size be known before hand. Looking down the initialized with size two, so maybe they can by `type ballot [2]bool`?

consensus/votecounter/ballot.go
```diff
@@ -0,0 +1,41 @@
...
+
+type ballotSet[A types.Addr] struct {
+	ballots     map[A]ballot
+	perVoteType []types.VotingPower
```
rodrodros: The same applies here for `perVoteType`. By giving static sizes we give the compiler and the runtime enough knowledge to perform optimizations in both stages. In this case, I believe quite relevant too

consensus/votecounter/round_data.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+	return true
+}
+
+func (r *roundData[V, H, A]) addVote(vote *types.Vote[H, A], votingPower types.VotingPower, voteType VoteType) bool {
+	var perVote *ballotSet[A]
+	if vote.ID != nil {
+		if perVote = r.perIDVotes[*vote.ID]; perVote == nil {
+			perVote = utils.HeapPtr(newBallotSet[A]())
+			r.perIDVotes[*vote.ID] = perVote
```
rodrodros, line 46: What's the rationale behind storing a pointer to `perVote` instead of `perVote` itself? You can still keep the pointer, assuming not a reference on the map, you can do:

```golang
r.perIDVotes[*vote.ID] = newBallotSet[A]()
perVote = &r.perIDVotes[*vote.ID]
```

consensus/votecounter/round_data.go
```diff
@@ -0,0 +1,80 @@
...
+	r.proposal = proposal
+
+	proposerBallot, ok := r.allVotes.ballots[proposal.Sender]
+	if proposerNotVoted := !ok || (!proposerBallot[Prevote] && !proposerBallot[Precommit]); proposerNotVoted {
+		r.uncountedProposerPower = votingPower
+	}
+	return true
+}
+
+func (r *roundData[V, H, A]) addVote(vote *types.Vote[H, A], votingPower types.VotingPower, voteType VoteType) bool {
```
rodrodros, line 41: This function only returns `true`, is there a point in it having a explicit return type (perhaps to satisfy an interface?)

consensus/votecounter/vote_counter.go
```diff
@@ -0,0 +1,194 @@
...
+}
+
+func getOrCreateRoundData[V types.Hashable[H], H types.Hash, A types.Addr](
+	roundMap map[types.Round]*roundData[V, H, A],
+	round types.Round,
+) *roundData[V, H, A] {
+	entry, ok := roundMap[round]
+	if !ok {
+		entry = utils.HeapPtr(newRoundData[V, H, A]())
+		roundMap[round] = entry
```
rodrodros, line 103: Similar comment to the one earlier regarding storing a reference in the map instead of the actual value. Marking it here to analize this case as well

consensus/integtest/node.go
infrmtcs: height is an uint64, so I don't think we can gain any performance by using reference.
But we can consider changing addr, as it's a felt.

consensus/votecounter/round_data.go
infrmtcs: I don't think the second line is possible, as map is not addressable.

consensus/votecounter/round_data.go
infrmtcs: Good catch! I forgot to use the return result from `perVote.add`.

consensus/votecounter/vote_counter.go
infrmtcs: Same as above, I don't think the second line is possible, as map is not addressable.

consensus/tendermint/rule_polka_nil.go
infrmtcs: Agree, I also wanted to do this

consensus/tendermint/broadcast.go
infrmtcs: Originally I planned to do this gradually, by adding functions to `voteCounter` one by one and remove the corresponding from `messages`, so I need to keep both `voteCounter` and `messages` at the same time. 
However, it turned out that removing `messages` is easier than I thought.

consensus/votecounter/round_data.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+func (r *roundData[V, H, A]) setProposal(proposal *types.Proposal[V, H, A], votingPower types.VotingPower) bool {
+	if r.proposal != nil {
+		return false
+	}
+	r.proposal = proposal
+
+	proposerBallot, ok := r.allVotes.ballots[proposal.Sender]
+	if proposerNotVoted := !ok || (!proposerBallot[Prevote] && !proposerBallot[Precommit]); proposerNotVoted {
```
rodrodros, line 35: 👍

consensus/votecounter/vote_counter_test_utils.go
```diff
@@ -0,0 +1,42 @@
...
+	t.Helper()
+	// New height will discard the previous height messages.
+	if v.currentHeight > expectedMsg.Height {
+		return
+	}
+	assert.Contains(t, v.roundData, expectedMsg.Round)
+	assert.Equal(t, v.roundData[expectedMsg.Round].proposal, &expectedMsg)
+}
+
+func (v *VoteCounter[V, H, A]) AssertEmpty(t *testing.T) {
```
rodrodros, line 39: nit: add `t.Helper()`

rodrigo-pino (approved):

PR #3016, opened 2025-07-29
CI: Wait for node to be synced before running the tests

brbrr:
Adds an additional check to CI tests, that make sure the node is proprely synced before running the tests.


.github/workflow-scripts/verify_deployment.sh
```diff
@@ -9,11 +9,14 @@ SLEEP_DURATION=10
 
 for i in $(seq 1 $MAX_ATTEMPTS); do
   DEPLOYED_VERSION=$(curl -s -X POST "$URL" -H "Content-Type: application/json" -d '{"method": "juno_version", "jsonrpc": "2.0", "id": 0}' | jq -r '.result')
+  IS_SYNCING=$(curl -s -X POST "$URL" -H "Content-Type: application/json" -d '{"method": "starknet_syncing", "jsonrpc": "2.0", "id": 0}' | jq -r '.result')
```
wojciechos: What do you think about using the `/ready` endpoint for this? starknet_syncing as false doesn't always mean the node is synced

brbrr: good idea. I've implemented it.

rodrigo-pino (approved):

wojciechos (approved):

PR #3017, opened 2025-07-31
refactor(state): common interface and adapters for trie and state

MaksymMalicki:


cmd/juno/juno.go
```diff
@@ -209,6 +210,7 @@ const (
 	httpUpdatePortUsage                = "The port on which the log level and gateway timeouts HTTP server will listen for requests."
 	submittedTransactionsCacheSize     = "Maximum number of entries in the submitted transactions cache"
 	submittedTransactionsCacheEntryTTL = "Time-to-live for each entry in the submitted transactions cache"
+	newStateUsage                      = "Use the new state package implementation instead of the legacy one"
```
rodrodros:
```suggestion
	newStateUsage                      = "EXPERIMENTAL: Use the new state package implementation instead of the legacy one"
```

core/state/commontrie/trie.go
rodrodros, line 1: Let's make naming a bit more descriptive and represent where we are heading

Let's call old `Trie`: `DeprecatedTrie` 
And the `Trie2`: `Trie`. 

If you want, this changes can be limited to the new code; that is, keep the names `Trie` and `Trie2` but everything that's a common interface such as it's adapters, let's call them `DeprecatedTrieAdapter` and `TrieAdapater`.

core/state/commontrie/trie.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+// TrieAdapter wraps trie.Trie to implement commontrie.CommonTrie
+type TrieAdapter struct {
+	Trie *trie.Trie
+}
+
+func NewTrieAdapter(t *trie.Trie) *TrieAdapter {
+	return &TrieAdapter{Trie: t}
+}
```
rodrodros: I believe the `TrieAdapter`struct can be improved. Currently, it holds a reference, and in the constructor it is returned as a reference, making it need to go through two pointer to get the same value. 

We can do the following thing
```golang
type TrieAdapter trie.Trie

func NewTrieAdapter(t *trie.Trie) *TrieAdapter {
   return (*TrieAdapter)(t)
}
```

core/state/commontrie/trie.go
```diff
@@ -0,0 +1,71 @@
...
+
+func (ta *TrieAdapter) HashFn() crypto.HashFn {
+	return ta.Trie.HashFn()
+}
+
+// Trie2Adapter wraps trie2.Trie to implement commontrie.CommonTrie
+type Trie2Adapter struct {
+	Trie *trie2.Trie
+}
```
rodrodros: Similar comment to the one immediately above

core/state/state.go
```diff
@@ -166,6 +166,7 @@ func (s *State) Update(
 	blockNum uint64,
 	update *core.StateUpdate,
 	declaredClasses map[felt.Felt]core.Class,
+	skipVerifyNewRoot bool,
```
rodrodros, line 182: Please add a comment that this variable is here to satisfy the State interface

rodrigo-pino (commented):
It looks good overall!

I think some other changes might be required, perhaps in a different PR, but I think all the current code should be prefixed with `Deprecated`. For example, `Trie` should be now `DeprecatedTrie` and `State` should be `DeprecatedState`. 

Their package names should be updated as well. It will be clear for all of us programming in the node that we are using deprecated structures

cmd/juno/juno.go
MaksymMalicki: done

core/state/commontrie/trie.go
MaksymMalicki: Yes, I kept them in the commonstate and commontrie packages. Also renamed CoreStateAdapter -> DeprecatedStateAdapter and CoreStateReaderAdapter -> DeprecatedStateReaderAdapter

core/state/commontrie/trie.go
MaksymMalicki: Nice idea, thanks

core/state/commontrie/trie.go
MaksymMalicki: done

core/state/state.go
MaksymMalicki: key good catch, this is actually used, for some reason the whole file was outdated

core/state/commonstate/state.go
```diff
@@ -0,0 +1,333 @@
...
+
+	ClassTrie() (commontrie.Trie, error)
+	ContractTrie() (commontrie.Trie, error)
+	ContractStorageTrie(addr *felt.Felt) (commontrie.Trie, error)
+}
+
+// DeprecatedStateAdapter wraps core.State to implement CommonState
+type DeprecatedStateAdapter struct {
+	*core.State
+}
```
rodrodros: What do you think of moving this deprecated files to their own script inside the package. For example for this one "deprecated_state.go".

core/state/commonstate/state.go
```diff
@@ -0,0 +1,333 @@
...
+	return commontrie.NewTrieAdapter(t), nil
+}
+
+type DeprecatedStateReaderAdapter struct {
+	core.StateReader
+}
+
+func NewDeprecatedStateReaderAdapter(s core.StateReader) *DeprecatedStateReaderAdapter {
+	return &DeprecatedStateReaderAdapter{StateReader: s}
+}
```
rodrodros: And this other (to be) deprecated type can go there as well. I believe it will make both codes easier to read

core/state/commonstate/state_test.go
```diff
@@ -0,0 +1,59 @@
...
+	if err != nil {
+		panic(err)
+	}
+	stateDB := state.NewStateDB(memDB, db)
+	state, err := state.New(&felt.Zero, stateDB)
+	require.NoError(t, err)
+
+	stateAdapter := NewStateAdapter(state)
+	assert.NotNil(t, stateAdapter)
+}
```
rodrodros: Remember to update the test names to match the new type names (e.g. `TestCoreStateAdapter` to `TestDeprecatedStateAdapter`)

core/state/commontrie/trie.go
```diff
@@ -0,0 +1,68 @@
...
+)
+
+type Trie interface {
+	Update(key, value *felt.Felt) error
+	Get(key *felt.Felt) (felt.Felt, error)
+	Hash() (felt.Felt, error)
+	HashFn() crypto.HashFn
+}
+
+type DeprecatedTrieAdapter trie.Trie
```
rodrodros: I think we could have also a "deprecated_trie.go" file inside the module. I think this type separation per script improves readability and will make the act of removing them later on easier.

core/state/commonstate/state.go
MaksymMalicki: yeah, this could be cleaner

core/state/commonstate/state.go
MaksymMalicki: agreed

core/state/commonstate/state_test.go
MaksymMalicki: done

core/state/commontrie/trie.go
MaksymMalicki: agreed

rodrigo-pino:
I see that we have a trie and a trie2 package, this is not very good naming in my opinion, 

Let's have a single package named `trie` with two submodules `trie` and `deprecated`.  That is instead of having:
* `trie`
* `trie2`

We have:
* `trie/trie` <- which is `trie2` at the time of this comment
* `trie/deprecated`  <- which is `trie` at the time of this comment

This can be done in a follow up PR.

core/state/commonstate/deprecated_state.go
```diff
@@ -0,0 +1,164 @@
...
+	"github.com/NethermindEth/juno/core"
+	"github.com/NethermindEth/juno/core/felt"
+	"github.com/NethermindEth/juno/core/state/commontrie"
+)
+
+// DeprecatedStateAdapter wraps core.State to implement CommonState
+type DeprecatedStateAdapter struct {
+	*core.State
+}
+
```
rodrodros: Why is this type defined like this and not like:
```golang
type DeprecatedStateAdapter core.State
```

core/state/commonstate/deprecated_state.go
```diff
@@ -0,0 +1,164 @@
...
+func NewDeprecatedStateAdapter(s *core.State) *DeprecatedStateAdapter {
+	return &DeprecatedStateAdapter{State: s}
+}
```
rodrodros: The return type here would be as simple as
```golang
func NewDeprecatedStateAdapter(s *core.State) *DeprecatedStateAdapter {
	return (*DeprecatedStateAdapter)(s)
}
```
No allocation, just changing how the pointer is interpreted

core/state/commonstate/state.go
```diff
@@ -0,0 +1,176 @@
...
+
+// StateAdapter wraps state.State to implement CommonState
+type StateAdapter struct {
+	*state.State
+}
+
+func NewStateAdapter(s *state.State) *StateAdapter {
+	return &StateAdapter{State: s}
+}
+
```
rodrodros: Similar comment here

core/state/commontrie/trie.go
```diff
@@ -0,0 +1,36 @@
...
+	Get(key *felt.Felt) (felt.Felt, error)
+	Hash() (felt.Felt, error)
+	HashFn() crypto.HashFn
+}
+
+type TrieAdapter trie2.Trie
+
+func NewTrieAdapter(t *trie2.Trie) *TrieAdapter {
+	return (*TrieAdapter)(t)
+}
```
rodrodros: Ah it was done like that here! Why not above there too?

core/state/commonstate/state.go
MaksymMalicki: addressed

core/state/commonstate/deprecated_state.go
MaksymMalicki: addressed

core/state/commonstate/deprecated_state.go
MaksymMalicki: addressed

core/trie2/trieutils/accessors.go
```diff
@@ -78,7 +78,7 @@ func WritePersistedStateID(w db.KeyValueWriter, id uint64) error {
 func ReadTrieJournal(r db.KeyValueReader) ([]byte, error) {
 	var journal []byte
 	if err := r.Get(db.TrieJournal.Key(), func(value []byte) error {
-		journal = value
+		journal = append([]byte(nil), value...)
```
infrmtcs: I think `slices.Clone` is clearer

core/state/commonstate/deprecated_state.go
```diff
@@ -0,0 +1,175 @@
...
+import (
+	"github.com/NethermindEth/juno/core"
+	"github.com/NethermindEth/juno/core/felt"
+	"github.com/NethermindEth/juno/core/state/commontrie"
+)
+
+// DeprecatedStateAdapter wraps core.State to implement CommonState
+type DeprecatedStateAdapter core.State
+
+func NewDeprecatedStateAdapter(s *core.State) *DeprecatedStateAdapter {
```
infrmtcs: nit: Not important, but I think we're not allocating anything new here, so probably it's not really "New". I think use the type casting at call site is not too long.

core/state/commonstate/deprecated_state.go
```diff
@@ -0,0 +1,175 @@
+package commonstate
+
+import (
+	"github.com/NethermindEth/juno/core"
+	"github.com/NethermindEth/juno/core/felt"
+	"github.com/NethermindEth/juno/core/state/commontrie"
+)
+
+// DeprecatedStateAdapter wraps core.State to implement CommonState
+type DeprecatedStateAdapter core.State
```
infrmtcs: Instead of this, I'm wondering if we can refactor the old state to return `felt.Felt` instead of `*felt.Felt`, so we don't need this adapter.

MaksymMalicki: The idea behind this whole package was to have the common behavior for old and the new state, but keep the internals unchanged. I'm open for discussion here

core/trie2/trieutils/accessors.go
MaksymMalicki: added

core/state/commonstate/deprecated_state.go
MaksymMalicki: Refactored the state anyways

core/state/commonstate/deprecated_state.go
MaksymMalicki: removed, added casting

blockchain/blockchain.go
```diff
@@ -61,9 +64,21 @@ type Blockchain struct {
 	l1HeadFeed    *feed.Feed[*core.L1Head]
 	cachedFilters *AggregatedBloomFilterCache
 	runningFilter *core.RunningEventFilter
+	stateFactory  *commonstate.StateFactory
 }
 
 func New(database db.KeyValueStore, network *utils.Network) *Blockchain {
+	trieDB, err := triedb.New(database, nil) // TODO: handle hashdb
```
MaksymMalicki: I will move this part to the followup PR

core/state/commontrie/deprecated_trie.go
```diff
@@ -0,0 +1,26 @@
...
+import (
+	"github.com/NethermindEth/juno/core/crypto"
+	"github.com/NethermindEth/juno/core/felt"
+	"github.com/NethermindEth/juno/core/trie"
+)
+
+type DeprecatedTrieAdapter trie.Trie
+
+func (dta *DeprecatedTrieAdapter) Update(key, value *felt.Felt) error {
+	_, err := (*trie.Trie)(dta).Put(key, value)
```
infrmtcs: I think the only reason why we need this `DeprecatedTrieAdapter` instead of using `trie.Trie` directly is this function. If we can unify the interface here, we can remove `DeprecatedTrieAdapter`. 
Additionally, this is also the only reason why we need `DeprecatedStateReaderAdapter`. We can also remove it if we don't have to wrap the returned trie with `DeprecatedTrieAdapter`

core/state/commontrie/trie.go
```diff
@@ -0,0 +1,32 @@
...
+	return (*trie2.Trie)(ta).Get(key)
+}
+
+func (ta *TrieAdapter) Hash() (felt.Felt, error) {
+	return (*trie2.Trie)(ta).Hash(), nil
+}
+
+func (ta *TrieAdapter) HashFn() crypto.HashFn {
+	return (*trie2.Trie)(ta).HashFn()
+}
```
infrmtcs: We should remove this, because `trie2.Trie` itself conforms to the `Trie` interface.

core/state/commontrie/trie_test.go
```diff
@@ -0,0 +1,41 @@
...
+import (
+	"testing"
+
+	"github.com/NethermindEth/juno/core/felt"
+	"github.com/NethermindEth/juno/core/trie2"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestTrieAdapter(t *testing.T) {
```
infrmtcs: This comment is for another PR. To guarantee that the old and the new trie will produce the same behaviour, we should write a single set of tests that depends on this interface, and run them with both the old and the new trie. 

The reason why I suggest to avoid writing 2 separate test suites for 2 tries is because of the scenarios where a single typo in 1 of the 2 tests can make both tests passed even though they produce different behaviours.

core/state/state_test_utils/new_state_flag.go
infrmtcs, line 1: nit: Folder name should be `statetestutils` to match the package name.

core/state/state_test_utils/new_state_flag.go
```diff
@@ -0,0 +1,21 @@
...
+	"sync"
+)
+
+var (
+	useNewState bool
+	once        sync.Once
+)
+
+func UseNewState() bool {
+	once.Do(func() {
```
infrmtcs: Why do we need `sync.Once` here? We can just replace this by something like `var UseNewState = func() { ... }()`

core/state/state.go
```diff
@@ -25,6 +25,7 @@ var (
 	noClassContractsClassHash = felt.Zero
 	noClassContracts          = map[felt.Felt]struct{}{
 		*new(felt.Felt).SetUint64(1): {},
+		*new(felt.Felt).SetUint64(2): {},
```
infrmtcs: nit: We can use the new constructors

infrmtcs: And why do we need this? I understand the existing code is already using `1` here, but I think we should make it clear what are `1` and `2` here by creating constants or variables for them.

core/state/state.go
```diff
@@ -217,8 +219,15 @@ func (s *State) Update(
 	}
 
 	// Check if the new commitment matches the one in state diff
-	if !newComm.Equal(update.NewRoot) {
-		return fmt.Errorf("state commitment mismatch: %v (expected) != %v (actual)", update.NewRoot, &newComm)
+	// The following check isn't relevant for the centralised Juno sequencer
+	if !skipVerifyNewRoot {
```
infrmtcs, line 235: Why do we need this flag? For correctness, fail fast can help to avoid cascading failures. For example, if the centralized sequencer updated to a new version while our node didn't, we should notify users instead of silently not checking the commitment.

core/state/history_test.go
```diff
@@ -19,12 +19,6 @@ func TestNewStateHistory(t *testing.T) {
 		assert.Equal(t, uint64(0), history.blockNum)
 		assert.NotNil(t, history.state)
 	})
-
-	t.Run("invalid state root", func(t *testing.T) {
```
infrmtcs, line 23: Why do we remove this test?

core/state/commonstate/deprecated_state_test.go
```diff
@@ -0,0 +1,27 @@
...
+}
+
+func TestDeprecatedStateReaderAdapter(t *testing.T) {
+	testDB := memory.New()
+	txn := testDB.NewIndexedBatch()
+
+	state := core.NewState(txn)
+	deprecatedStateAdapter := NewDeprecatedStateReaderAdapter(state)
+	assert.NotNil(t, deprecatedStateAdapter)
+}
```
infrmtcs: Another reason why I advocate removing the adapter is because of missing tests. We have several groups of functions sharing the same function signature. For example, `ContractClassHash`, `ContractNonce` and `ContractStorage` all share the same signature `func(addr *felt.Felt) (felt.Felt, error)`. The same thing applies to `...At` and `...Trie` variants. Without tests, there is no way we can guarantee that the adapter's `ContractClassHash` calls underlying `ContractClassHash` instead of `ContractNonce`.

core/trie2/triedb/pathdb/buffer.go
```diff
@@ -58,10 +65,17 @@ func (b *buffer) flush(kvs db.KeyValueStore, cleans *cleanCache, id uint64) erro
 		)
 	}
 
-	batch := kvs.NewBatchWithSize(b.nodes.dbSize())
+	dbSize := b.nodes.dbSize()
+
+	batch := kvs.NewBatchWithSize(dbSize)
+	if batch == nil {
+		return fmt.Errorf("failed to create batch")
+	}
```
infrmtcs: When is this possible? I'm wondering if we can check for error here instead of checking for nil.

core/trie2/triedb/pathdb/disklayer.go
```diff
@@ -77,7 +82,7 @@ func (dl *diskLayer) node(id trieutils.TrieID, owner *felt.Felt, path *trieutils
 	}
 
 	// Finally, read from disk
-	blob, err := trieutils.GetNodeByPath(dl.db.disk, id.Bucket(), owner, path, isClass)
+	blob, err := trieutils.GetNodeByPath(dl.db.disk, id.Bucket(), owner, path, isLeaf)
```
infrmtcs, line 85: This kind of error looks quite dangerous to me. Is there a way we can prevent this?

core/trie2/triedb/pathdb/difflayer.go
```diff
@@ -38,10 +40,13 @@ func (dl *diffLayer) node(id trieutils.TrieID, owner *felt.Felt, path *trieutils
 	isClass := id.Type() == trieutils.Class
 	n, ok := dl.nodes.node(owner, path, isClass)
 	if ok {
+		if _, deleted := n.(*trienode.DeletedNode); deleted {
+			return nil, db.ErrKeyNotFound
+		}
```
infrmtcs, line 45: Why do we return error in this specific case? I checked the other usages of `Blob()` and it seems that we accept a nil for deleted node.

core/state/state_test_utils/new_state_flag.go
```diff
@@ -0,0 +1,21 @@
...
+		val := os.Getenv("USE_NEW_STATE")
```
infrmtcs: I think we should use
```suggestion
		val := os.Getenv("JUNO_NEW_STATE")
```
So that we can reuse the same env flag used by cobra declared in the next PR.

infrmtcs: FYI, if we declare the flag `--new-state`, we'll automatically accept `JUNO_NEW_STATE` as an alternative.

core/state/commontrie/trie.go
MaksymMalicki: The Hash() method from `trie2` would have to be refactored, so that it returns a `nil` error. I can do it

core/state/commontrie/deprecated_trie.go
MaksymMalicki: agreed, after refactors its possible to do it

core/state/state_test_utils/new_state_flag.go
MaksymMalicki: changed

core/trie2/triedb/pathdb/buffer.go
MaksymMalicki: Actually it's never returned as nil. Tbh this was done a long time ago and I don't really remember why I did this, but after checking it up, this check is not nessecary. In the flow of the batch creation, I don't see any place where the batch is returned as nil. For `memory`/`remote` we always return some non-nil struct. For `pebble` we internally use the `newBatchWithSize`, which just caps the batch data size to the desired size

core/trie2/triedb/pathdb/disklayer.go
MaksymMalicki: This was kinda hard to spot, I'm not sure if there is any way to enforce this particular boolean here

core/state/state.go
MaksymMalicki: This was basically copied from the old state to mimic the behaviour. From what @rianhughes explained to me, the state update needs to be populated with the old and new state roots. The FGW normally calculates this for us, so we just verify it. However, when running Juno as a sequencer, we need to generate these fields ourselves, which is done in `updateStateRoots()`. If we set `skipVerifyNewRoot` to true (which is the case for the sequencer), then we only require the old state root

core/state/state_test_utils/new_state_flag.go
MaksymMalicki: Yup, done

core/state/state_test_utils/new_state_flag.go
MaksymMalicki: thanks, name changed

core/trie2/triedb/pathdb/difflayer.go
MaksymMalicki: Oooh, this was done a long time ago as a fix for syncing. If I remember correctly, we don't want to be able to read the nodes marked for deletion from the cache layers

core/state/state.go
MaksymMalicki: Used new constructors and consts, this state package was implemented before the 0x2 system contract was introduced, hence it had to be added.

core/state/history_test.go
MaksymMalicki: Hmmmm, I think this was done for the compatibility with the pathDB, but now that we are using rawDB it can be brought back

core/state/commonstate/state.go
```diff
@@ -0,0 +1,108 @@
...
+		blockNum uint64,
+		update *core.StateUpdate,
+		declaredClasses map[felt.Felt]core.Class,
+		skipVerifyNewRoot bool,
+	) error
+	Revert(blockNum uint64, update *core.StateUpdate) error
+	Commitment() (felt.Felt, error)
+}
+
+type StateReader interface {
```
infrmtcs, line 32: Is this going to replace the old `StateReader` or is this a different thing? Just asking question to understand the plan, because I think it's also good idea to replace in another PR if it's the case.

rpc/v8/storage.go
```diff
@@ -206,26 +207,40 @@ func (h *Handler) isBlockSupported(blockID *BlockID, chainHeight uint64) *jsonrp
 	return nil
 }
 
-func getClassProof(tr *trie.Trie, classes []felt.Felt) ([]*HashToNode, error) {
+func getClassProof(tr commontrie.Trie, classes []felt.Felt) ([]*HashToNode, error) {
+	t, ok := tr.(*trie.Trie)
```
infrmtcs, line 213: Can we leave a TODO here? I think eventually the RPC packages shouldn't care about which trie implementation is being used and the output format should be the same, but we can do it step by step.

infrmtcs: Please also put the TODO in other functions with similar changes and v9 too

core/state/state.go
```diff
@@ -12,19 +12,26 @@ import (
...
+)
+
 var (
 	stateVersion0             = new(felt.Felt).SetBytes([]byte(`STARKNET_STATE_V0`))
 	leafVersion0              = new(felt.Felt).SetBytes([]byte(`CONTRACT_CLASS_LEAF_V0`))
 	noClassContractsClassHash = felt.Zero
 	noClassContracts          = map[felt.Felt]struct{}{
-		*new(felt.Felt).SetUint64(1): {},
+		*felt.NewFromUint64[felt.Felt](systemContract1Addr): {},
+		*felt.NewFromUint64[felt.Felt](systemContract2Addr): {},
```
EgeCaner:
```suggestion
		felt.FromUint64[felt.Felt](systemContract1Addr): {},
		felt.FromUint64[felt.Felt](systemContract2Addr): {},
```

core/state/commonstate/state.go
MaksymMalicki: yes! it will be done in a separate PR

rpc/v8/storage.go
MaksymMalicki: done

core/state/state.go
MaksymMalicki: thanks!

PR #3024, opened 2025-08-01
Feat (Tendermint): implement sync mechanism

rianhughes:


consensus/sync/sync.go
```diff
@@ -0,0 +1,115 @@
...
+					Block:       committedBlock.Block,
+					StateUpdate: committedBlock.StateUpdate,
+					NewClasses:  committedBlock.NewClasses,
+				},
+				SimulateResult: &blockchain.SimulateResult{
+					BlockCommitments: committedBlock.Commitments,
+					ConcatCount:      concatCommitments,
+				},
+				// Todo: this needs added to the spec.
+				L2GasConsumed: 1,
```
rianhughes: Re-highlighting this 

https://github.com/NethermindEth/juno/pull/2962#discussion_r2209440199

utils/pipeline/pipeline.go
```diff
@@ -62,14 +62,11 @@ func FanIn[T any](ctx context.Context, channels ...<-chan T) <-chan T {
...
 
-func Bridge[T any](ctx context.Context, chanCh <-chan <-chan T) <-chan T {
-	out := make(chan T)
+func Bridge[T any](ctx context.Context, out chan T, chanCh <-chan <-chan T) {
 	if chanCh == nil {
-		close(out)
-		return out
+		return
 	}
 	go func() {
```
infrmtcs: I think we should remove `go` here and let the caller decide how to run it. In this way, the caller can know when this function stop in order to close the channel properly and avoid panic.

utils/pipeline/pipeline_test.go
```diff
@@ -108,10 +108,18 @@ func TestPipeline(t *testing.T) {
 			}
 		}()
 
+		out := make(chan int)
+		go Bridge[int](ctx, out, chOfInts) // Blocking writes into `out`
```
infrmtcs, line 113: We either need to remove `go` here or inside `Bridge`, otherwise we're creating new goroutine twice.

p2p/p2p.go
```diff
@@ -35,6 +36,13 @@ const (
 	clientName                = "juno"
 )
 
+type BlockListener interface {
+	service.Service
+	Listen() <-chan p2pSync.BlockBody
+}
+
+var _ BlockListener = (*Service)(nil)
```
infrmtcs: nit: Probably we can remove this, because we already include `Service` in `BlockListener` definition.

consensus/sync/sync.go
```diff
@@ -0,0 +1,115 @@
...
+
+const syncRoundPlaceHolder = -1 // Todo: We use this value until the round is added to the spec
+
+type Sync[V types.Hashable[H], H types.Hash, A types.Addr] struct {
+	syncService       p2p.BlockListener
+	driverProposalCh  chan types.Proposal[V, H, A]
+	driverPrecommitCh chan types.Precommit[H, A]
+	// Todo: for now we can forge the precommit votes of our peers
+	// In practice, this information needs to be exposed by peers.
+	getPrecommits func(types.Height) []types.Precommit[H, A]
```
infrmtcs: I think this should accept `BlockBody` instead

consensus/sync/mock_p2p_sync_test.go
```diff
@@ -0,0 +1,47 @@
...
+func newMockP2PSyncService(syncReceiveCh chan sync.BlockBody) mockP2PSyncService {
+	return mockP2PSyncService{
+		syncReceiveCh:   syncReceiveCh,
+		blockListenerCh: make(chan sync.BlockBody),
+	}
+}
+
+func (m *mockP2PSyncService) shouldTriggerErr(triggerErr bool) {
+	m.triggerErr = triggerErr
+}
```
rodrodros: Isn't superfluos that the triggerErr value is passed, shouldn't it be true just by the fact this function is being called 🤔. Looking at it's use later on it's only called once and with `true`

consensus/sync/sync.go
```diff
@@ -0,0 +1,115 @@
...
+func (s *Sync[V, H, A]) Run(originalCtx context.Context) {
+	ctx, cancel := context.WithCancel(originalCtx)
+	defer cancel()
+
+	go func() {
+		defer cancel()
+		err := s.syncService.Run(ctx)
+		if err != nil {
+			return
+		}
```
rodrodros: Why is the error completely ignored?

consensus/sync/sync_test.go
```diff
@@ -0,0 +1,178 @@
...
+}
+
+func newMockBlockchain(t *testing.T) blockchain {
+	return &mockBlockchain{
+		t: t,
+	}
+}
+
+func mockTimeoutFn(step types.Step, round types.Round) time.Duration {
+	return 10 * time.Second
```
rodrodros, line 56: Why such a long timeout?

consensus/sync/sync_test.go
```diff
@@ -0,0 +1,178 @@
...
+	go func() {
+		mockP2PSyncService.recieveBlockOverP2P(block0)
+	}()
+
+	go func() {
+		time.Sleep(2 * time.Second)
+		cancel()
+	}()
+	consensusSyncService.Run(ctx)                     // Driver should trigger stopSyncCh and shut this service down
+	require.NotEmpty(t, proposalStore.Get(valueHash)) // Ensure the Driver sees the correct proposal
```
rodrodros: I understand what we are trying to do here but I am wary of the approach. I was hoping there was a way we could test this without depending on time. 

For example, wait on that the proposal is sent through the proposal channel. Once the proposal is sent through the `s.driverProposalCH`

consensus/sync/sync.go
```diff
@@ -0,0 +1,115 @@
...
+					NewClasses:  committedBlock.NewClasses,
+				},
+				SimulateResult: &blockchain.SimulateResult{
+					BlockCommitments: committedBlock.Commitments,
+					ConcatCount:      concatCommitments,
+				},
+				// Todo: this needs added to the spec.
+				L2GasConsumed: 1,
+			}
+			s.proposalStore.Store(msgH, &buildResult)
```
rodrodros, line 72: I see that tests check correctness up until this point

consensus/sync/sync.go
```diff
@@ -0,0 +1,115 @@
...
+				Value:      &msgV,
+			}
+
+			select {
+			case <-ctx.Done():
+				return
+			case s.driverProposalCh <- proposal:
+			}
+		}
+	}
```
rodrodros: Is the follow up code tested anywhere?

consensus/sync/sync.go
rodrodros: Looking more at the code, I am more convinced this error should be propagated

consensus/sync/sync_test.go
```diff
@@ -0,0 +1,178 @@
...
+	mockInCh := make(chan sync.BlockBody)
+	mockP2PSyncService := newMockP2PSyncService(mockInCh)
+	proposalStore := proposal.ProposalStore[starknet.Hash]{}
+
+	consensusSyncService := consensusSync.New(&mockP2PSyncService, proposalCh, precommitCh, getPrecommits, toValue, &proposalStore)
+	cancel()
+	consensusSyncService.Run(ctx)
+
+	mockP2PSyncService.shouldTriggerErr(true)
+	consensusSyncService.Run(t.Context())
```
rodrodros: `Run` should return an error and here we should check each time for the kind of error we are expecting. 

Additionally, I think this error(s) should be logged somewhere, in the context of production code.

p2p/sync/sync.go
```diff
@@ -122,28 +128,14 @@ func (s *Service) processBlock(ctx context.Context, blockNumber uint64) error {
 		return fmt.Errorf("failed to get state diffs: %w", err)
 	}
 
-	blocksCh := pipeline.Bridge(ctx, s.processSpecBlockParts(ctx, blockNumber, pipeline.FanIn(ctx,
+	pipeline.Bridge(ctx, s.blockCh, s.processSpecBlockParts(ctx, blockNumber, pipeline.FanIn(ctx,
 		pipeline.Stage(ctx, headersAndSigsCh, specBlockPartsFunc[specBlockHeaderAndSigs]),
 		pipeline.Stage(ctx, classesCh, specBlockPartsFunc[specClasses]),
 		pipeline.Stage(ctx, stateDiffsCh, specBlockPartsFunc[specContractDiffs]),
 		pipeline.Stage(ctx, txsCh, specBlockPartsFunc[specTxWithReceipts]),
 		pipeline.Stage(ctx, eventsCh, specBlockPartsFunc[specEvents]),
```
rodrodros: Let's use the 
```
f(
  a1,
  a2,
  ...,
  an
)
```

formatting here

p2p/sync/sync.go
```diff
@@ -339,7 +331,7 @@ func (s *Service) adaptAndSanityCheckBlock(
 				coreC := p2p2core.AdaptClass(cls)
 				h, err := coreC.Hash()
 				if err != nil {
-					bodyCh <- blockBody{err: fmt.Errorf("class hash calculation error: %v", err)}
+					bodyCh <- BlockBody{Err: fmt.Errorf("class hash calculation error: %v", err)}
```
rodrodros, line 340: I know this is old code but I believe this general code style is not good and I am using this comment to open up a discussion and hear what you guys @rianhughes and @infrmtcs think. 

I believe a dedicated channel for errors is requried in this case of scenarios and not having a type represent multiple things...

utils/pipeline/pipeline_test.go
```diff
@@ -21,9 +21,9 @@ func TestPipeline(t *testing.T) {
 			_, open := <-FanIn[int](t.Context(), nil, nil, nil)
 			assert.False(t, open)
 		})
-		t.Run("Bridge", func(t *testing.T) {
-			_, open := <-Bridge[int](t.Context(), nil)
-			assert.False(t, open)
+		t.Run("Bridge doesn't call out", func(t *testing.T) {
```
rodrodros, line 24: 👍

consensus/sync/sync.go
infrmtcs: For now, we can compute this field ourselves by data from this field: https://github.com/starknet-io/starknet-p2p-specs/blob/main/p2p/proto/sync/receipt.proto#L50

consensus/sync/sync.go
infrmtcs: I think we should instead remove this block and start it together with other services in our init code.

consensus/sync/sync_test.go
```diff
@@ -0,0 +1,178 @@
...
+		time.Sleep(2 * time.Second)
+		cancel()
+	}()
+	consensusSyncService.Run(ctx)                     // Driver should trigger stopSyncCh and shut this service down
+	require.NotEmpty(t, proposalStore.Get(valueHash)) // Ensure the Driver sees the correct proposal
+}
+
+func TestShutdownOnError(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
```
infrmtcs: Can we remove this?

consensus/sync/mock_p2p_sync_test.go
```diff
@@ -0,0 +1,47 @@
...
+import (
+	"context"
+	"errors"
+
+	"github.com/NethermindEth/juno/p2p/sync"
+)
+
+type mockP2PSyncService struct {
+	syncReceiveCh   chan sync.BlockBody // blocks received over p2p go here
+	blockListenerCh chan sync.BlockBody // Listener will get this
```
infrmtcs: Why do we need 2 channels for this mock service?

consensus/sync/sync_test.go
```diff
@@ -0,0 +1,178 @@
...
+		stateMachine,
+		newMockBlockchain(t),
+		p2p,
+		newMockProposer(),
+		mockTimeoutFn,
+	)
+
+	go func() {
+		require.NoError(t, driver.Run(ctx))
+	}()
```
infrmtcs: We don't need these components, because the component under test depends on the channel, not the driver or any of these components. We can simply create 2 channels, drain them and assert the elements.

consensus/sync/sync.go
```diff
@@ -0,0 +1,115 @@
...
+	driverPrecommitCh chan types.Precommit[H, A]
+	// Todo: for now we can forge the precommit votes of our peers
+	// In practice, this information needs to be exposed by peers.
+	getPrecommits func(types.Height) []types.Precommit[H, A]
+	toValue       func(*felt.Felt) V
+	proposalStore *proposal.ProposalStore[H]
+}
+
+func New[V types.Hashable[H], H types.Hash, A types.Addr](
+	syncService p2p.BlockListener,
```
infrmtcs: If we replace this by a channel directly, it can decouple this component with the BlockListener and make it more flexible and easier to test.

utils/pipeline/pipeline.go
rianhughes: I've removed go from the Bridge function

utils/pipeline/pipeline_test.go
rianhughes: go has been removed from Bridge, so we should only create a single routine

p2p/p2p.go
rianhughes: It also constrains on the 
```
Listen() <-chan p2pSync.BlockBody
```
function

consensus/sync/sync.go
rianhughes: Why? Currently block body doesn't contain any `precommits`, and it's not clear why we should request more information than we need. The `getPrecommits()` function should be used until we actually start storing the precommits in the block body imo

consensus/sync/mock_p2p_sync_test.go
rianhughes: Yeah true, it could be simplified given it's never called with `false`. Updated the code

consensus/sync/sync_test.go
rianhughes: The main flow being tested is that we can actually sync blocks via the Driver, and I don't want to accidentally trigger the timeouts

consensus/votecounter/vote_counter.go
```diff
@@ -111,10 +112,11 @@ func (v *VoteCounter[V, H, A]) AddProposal(proposal *types.Proposal[V, H, A]) bo
 		return false
 	}
 
+	fmt.Println(" [][]  v.validators.Proposer()")
 	if expectedProposer := v.validators.Proposer(proposal.Height, proposal.Round); proposal.Sender != expectedProposer {
+		fmt.Println(" [][] weird bug!!!", proposal.Sender, expectedProposer)
```
infrmtcs: Remove

consensus/votecounter/vote_counter.go
```diff
@@ -111,10 +112,11 @@ func (v *VoteCounter[V, H, A]) AddProposal(proposal *types.Proposal[V, H, A]) bo
 		return false
 	}
 
+	fmt.Println(" [][]  v.validators.Proposer()")
```
infrmtcs: Remove

consensus/tendermint/rule_precommit_any.go
```diff
@@ -15,7 +15,6 @@ func (s *stateMachine[V, H, A]) uponPrecommitAny() bool {
 	isFirstTime := !s.state.timeoutPrecommitScheduled
 
 	hasQuorum := s.voteCounter.HasQuorumForAny(s.state.round, votecounter.Precommit)
-
```
infrmtcs: Remove

consensus/tendermint/rule_commit_value.go
```diff
@@ -26,7 +28,7 @@ func (s *stateMachine[V, H, A]) uponCommitValue(cachedProposal *CachedProposal[V
 
 	// This is checked here instead of inside execution, because it's the only case in execution in this rule
 	isValid := cachedProposal.Valid
-
+	fmt.Println("hasQuorum && isValid", hasQuorum && isValid, isValid, hasQuorum)
```
infrmtcs: Remove

consensus/tendermint/process.go
```diff
@@ -114,7 +119,7 @@ func (s *stateMachine[V, H, A]) process(recentlyReceivedRound *types.Round) (act
 	case s.uponPrecommitAny():
 		return s.doPrecommitAny(), true
 
-	// Line 49
+		// Line 49
```
infrmtcs: Remove

consensus/tendermint/process.go
```diff
@@ -89,6 +91,9 @@ func (s *stateMachine[V, H, A]) process(recentlyReceivedRound *types.Round) (act
 		roundCachedProposal = s.findProposal(*recentlyReceivedRound)
 	}
 
+	if recentlyReceivedRound != nil {
```
infrmtcs: Remove

consensus/sync/sync.go
infrmtcs: Firstly, real code won't be able to extract the precommits just from the height.
Secondly,
> The getPrecommits() function should be used until we actually start storing the precommits in the block body imo

This is assuming that the precommits are stored in the format that is exactly the same as `Precommit[starknet.Value, starknet.Hash, starknet.Address]`. Otherwise, we still need this `getPrecommits` function to extract the precommits from the downloaded block.

consensus/votecounter/ballot.go
```diff
@@ -38,7 +40,6 @@ func (b *ballotSet[A]) add(addr *A, addrPower types.VotingPower, voteType VoteTy
 	ballot := b.ballots[*addr]
 	ballot[voteType] = true
 	b.ballots[*addr] = ballot
-
```
infrmtcs: The blank line is intentional, to separate between `ballots` update (3 lines) and `perVoteType` update

consensus/votecounter/vote_counter.go
```diff
@@ -150,7 +148,6 @@ func (v *VoteCounter[V, H, A]) HasQuorumForVote(round types.Round, voteType Vote
 	if !ok {
 		return false
 	}
-
```
infrmtcs: The blank line is intentional, to separate "get round data" (map lookup + ok check) and condition.

consensus/votecounter/round_data.go
```diff
@@ -66,7 +66,6 @@ func (r *roundData[V, H, A]) countVote(voteType VoteType, id *H) types.VotingPow
 	} else {
 		perVote = &r.nilVotes
 	}
-
 	return perVote.perVoteType[voteType]
```
infrmtcs: The blank line is intentional, to separate between selecting ballots and extracting the aggregate per `voteType`.

p2p/p2p.go
infrmtcs: OK, my bad, this is `Service` defined in this file, not `service.Service`

consensus/votecounter/vote_counter.go
```diff
@@ -111,10 +111,9 @@ func (v *VoteCounter[V, H, A]) AddProposal(proposal *types.Proposal[V, H, A]) bo
 		return false
 	}
 
-	if expectedProposer := v.validators.Proposer(proposal.Height, proposal.Round); proposal.Sender != expectedProposer {
+	if expectedProposer := v.validators.Proposer(proposal.Height, proposal.Round); types.AddrCmp(proposal.Sender, expectedProposer) {
```
infrmtcs, line 114: This is strange. Both `proposal.Sender` and `expectedProposer` types are `A` which is a value type because of `~[4]uint64`. The concrete type comparison should work.

consensus/tendermint/rule_commit_value.go
```diff
@@ -26,7 +26,6 @@ func (s *stateMachine[V, H, A]) uponCommitValue(cachedProposal *CachedProposal[V
 
 	// This is checked here instead of inside execution, because it's the only case in execution in this rule
 	isValid := cachedProposal.Valid
-
```
infrmtcs: Please undo this

consensus/sync/sync.go
rianhughes: I've updated the test so that we not longer mock the state machine, and check that the `Commit` function is called. This tests the rest of this function

consensus/sync/sync.go
rianhughes: I've updated the test so that we not longer mock the state machine, and check that the Commit function is called. This tests the rest of this function

consensus/sync/sync_test.go
rianhughes: I've updated the code so that Run returns an error. I didn't add a logger since the returned error is logged in `StartService()`

p2p/sync/sync.go
rianhughes: Done

consensus/votecounter/vote_counter.go
rianhughes: I forgot to make the PR a draft while making changes, sorry. These have been removed though

consensus/votecounter/vote_counter.go
rianhughes: I forgot to make the PR a draft while making changes, sorry. These have been removed though

consensus/tendermint/rule_commit_value.go
rianhughes: I forgot to make the PR a draft while making changes, sorry. These have been removed though

consensus/tendermint/process.go
rianhughes: Done

consensus/votecounter/vote_counter.go
rianhughes: It is strange. It seems the endianess gets flipped after returning from `Proposer` for some reason

consensus/sync/sync.go
rianhughes: This block has been removed, now we return the error to the caller (which is logged if non-nil)

consensus/sync/sync_test.go
rianhughes: I removed the time.Sleep code. After the `Driver` calls `Commit()`, I cancel the context

consensus/sync/sync.go
rianhughes: > Firstly, real code won't be able to extract the precommits just from the height.

Why? I thought they were going to be stored in the block body, in which case we should be able to just load the block at a given height and extract them.

> This is assuming that the precommits are stored in the format that is exactly the same as

Sure, but couldn't we call some adapters in the `getPrecommits` function? Eg` types.Height -> uint -> core.Block -> core.Precommits -> types.Precommits`

rianhughes: What exactly do you mean by accept BlockBody btw? Eg what would you change this to?

```
	getPrecommits func(types.Height) []types.Precommit[H, A]
```

consensus/sync/mock_p2p_sync_test.go
rianhughes: It was conceptually very similar to how the p2p sync service used to work. Namely we would receive a complete block into one channel, then send that result into another Listener channel. I've updated the code to just use a single channel here though

consensus/sync/sync.go
infrmtcs: > Why? I thought they were going to be stored in the block body, in which case we should be able to just load the block at a given height and extract them.

Where will you load the block? Firstly, it's not stored in the database or anywhere else yet. Secondly, why do we need to get it from somewhere else if we already have access to it at call site?

> Sure, but couldn't we call some adapters in the getPrecommits function? Eg types.Height -> uint -> core.Block -> core.Precommits -> types.Precommits

All of these arguments are supposed to be extracted from `BlockBody`. Then why don't we just pass `BlockBody` to the function directly?

infrmtcs:
```
	getPrecommits func(*sync.BlockBody) []types.Precommit[H, A]
```

rianhughes: > getPrecommits func(*sync.BlockBody) []types.Precommit[H, A]

Ahh, right. Makes sense, I'll update it now

rianhughes: Updated

consensus/sync/sync.go
rianhughes: Yeah this is much nicer, I'll update it now

rianhughes: I've updated this to

```
type Sync[V types.Hashable[H], H types.Hash, A types.Addr] struct {
	blockListener     <-chan sync.BlockBody 
```

which also simplified `Run` quite a bit

consensus/sync/sync_test.go
rianhughes: True, this would have been much simpler. However in the current tests the `driver` and `state machine` aren't mocked, so the test is more comprehensive compared to just draining the input channels.

consensus/tendermint/rule_commit_value.go
rianhughes: Done

consensus/votecounter/round_data.go
rianhughes: Done

consensus/votecounter/vote_counter.go
rianhughes: Done

consensus/votecounter/ballot.go
rianhughes: Done

consensus/tendermint/process.go
rianhughes: Done

consensus/tendermint/rule_precommit_any.go
rianhughes: Done

consensus/sync/sync.go
rianhughes: I've created a method for a block that returns the cumulative sum of the receipts `ExecutionResources.TotalGasConsumed.L2Gas`

p2p/sync/sync.go
rianhughes: Imo moving the error to a separate channel adds unnecessary complexity, and it's not clear what benefits we gain in terms of functionality or readability.

PR #3026, opened 2025-08-02
refactor(juno): use state and trie common interface throughout the project

MaksymMalicki:
1. This PR should be reviewed after [#3017](https://github.com/NethermindEth/juno/pull/3017).

2. It integrates the common interfaces across the project, including in utilities and tests.

3. Some parts of the logic weren’t straightforward to unify due to the different ways the old and new states interact with PebbleDB. The old state is tightly coupled with `IndexedBatch` for reads and writes, while the new state is designed to read using raw DB/`snapshots` and write using regular PebbleDB `Batch`. As a result, the following methods were "duplicated" and tailored to the way the respective `state` and `trie` functionalities are meant to interact with the DB:

- `blockchain.Store`

- `blockchain.GetReverseStateDiff`

- `blockchain.Finalise`

The adapting had to be also done for the prove logic inside of `rpc/v8/storage.go`, `rpc/v8/storage_test.go`, `rpc/v9/storage.go`, `rpc/v9/storage_test.go`

4. A new method for writing a TrieDB journal on clean node shutdown (for the new state/trie using the PathDB scheme) was added at the end of the blockchain package.

5. The experimental --new-state boolean CLI flag is now fully integrated. It can be used via the command line and in the `dbcmd` utilities.

6. The new state package includes some minor bug fixes discovered during previous work ([#2912](https://github.com/NethermindEth/juno/pull/2912/files)) while debugging sync, RPC, and E2E tests.

7. A new command was added to the `Makefile`, to run the unit tests strictly with the new state and trie implementations in the future CI/CD:
```
test-new-state: clean-testcache rustdeps ## Run tests with new state
	USE_NEW_STATE=true go test $(GO_TAGS) -v ./...
```

8. Breaking changes:

- `blockchain.L1Head()` and `blockchain.L1HandlerTxnHash(msgHash *common.Hash)` now return by value instead of reference

Next PR to review: [#3035](https://github.com/NethermindEth/juno/pull/3035)

### DIFF
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -6,6 +6,9 @@ import (
 
 	"github.com/NethermindEth/juno/core"
 	"github.com/NethermindEth/juno/core/felt"
+	"github.com/NethermindEth/juno/core/state"
+	"github.com/NethermindEth/juno/core/state/commonstate"
+	"github.com/NethermindEth/juno/core/trie2/triedb"
 	"github.com/NethermindEth/juno/db"
 	"github.com/NethermindEth/juno/feed"
 	"github.com/NethermindEth/juno/utils"
@@ -112,14 +115,6 @@ func (b *Blockchain) WithListener(listener EventListener) *Blockchain {
 
 func (b *Blockchain) Network() *utils.Network {
 	return b.network
-}
-
-// StateCommitment returns the latest block state commitment.
-// If blockchain is empty zero felt is returned.
-func (b *Blockchain) StateCommitment() (felt.Felt, error) {
-	b.listener.OnRead("StateCommitment")
-	batch := b.database.NewIndexedBatch() // this is a hack because we don't need to write to the db
-	return core.NewState(batch).Root()
 }
 
 // Height returns the latest block height. If blockchain is empty nil is returned.
--- a/blockchain/blockchain_test.go
+++ b/blockchain/blockchain_test.go
@@ -252,10 +252,6 @@ func TestStore(t *testing.T) {
 		require.NoError(t, err)
 		assert.Equal(t, block0, headBlock)
 
-		root, err := chain.StateCommitment()
-		require.NoError(t, err)
-		assert.Equal(t, stateUpdate0.NewRoot, &root)
-
 		got0Block, err := chain.BlockByNumber(0)
 		require.NoError(t, err)
 		assert.Equal(t, block0, got0Block)
@@ -279,10 +275,6 @@ func TestStore(t *testing.T) {
 		headBlock, err := chain.Head()
 		require.NoError(t, err)
 		assert.Equal(t, block1, headBlock)
-
-		root, err := chain.StateCommitment()
-		require.NoError(t, err)
-		assert.Equal(t, stateUpdate1.NewRoot, &root)
 
 		got1Block, err := chain.BlockByNumber(1)
 		require.NoError(t, err)
--- a/cmd/juno/juno.go
+++ b/cmd/juno/juno.go
@@ -149,6 +149,7 @@ const (
 	defaultSubmittedTransactionsCacheEntryTTL = 5 * time.Minute
 	defaultNewState                           = false
 	defaultDisableRPCBatchRequests            = false
+	newStateF                                 = "new-state"
 
 	configFlagUsage                       = "The YAML configuration file."
 	logLevelFlagUsage                     = "Options: trace, debug, info, warn, error."
@@ -220,6 +221,7 @@ const (
 	submittedTransactionsCacheSize     = "Maximum number of entries in the submitted transactions cache"
 	submittedTransactionsCacheEntryTTL = "Time-to-live for each entry in the submitted transactions cache"
 	disableRPCBatchRequestsUsage       = "Disables handling of batched RPC requests."
+	newStateUsage                      = "EXPERIMENTAL: Use the new state package implementation"
 )
 
 var Version string
--- a/node/node.go
+++ b/node/node.go
@@ -118,6 +118,8 @@ type Config struct {
 	HTTPUpdatePort uint16 `mapstructure:"http-update-port"`
 
 	ForbidRPCBatchRequests bool `mapstructure:"disable-rpc-batch-requests"`
+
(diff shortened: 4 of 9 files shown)


adapters/p2p2core/state.go
```diff
@@ -53,16 +54,16 @@ func AdaptStateDiff(reader core.StateReader, contractDiffs []*state.ContractDiff
 				classHash: diff.ClassHash,
 			}
 
-			var stateClassHash *felt.Felt
+			var stateClassHash felt.Felt
```
rodrodros: This body can be changed to 
```
stateClassHash := felt.Zero{}
if ready != nil {
 // origina code from else body
}
```

Actual code is below this comment but is not letting me mark it for some reason

blockchain/blockchain.go
```diff
@@ -87,21 +89,23 @@ var _ Reader = (*Blockchain)(nil)
 type Blockchain struct {
 	network       *utils.Network
 	database      db.KeyValueStore
+	trieDB        *triedb.Database
+	StateDB       *state.StateDB // TODO(weiihann): not sure if it's a good idea to expose this
```
rodrodros: 👀 why is it exposed and why it shouldn't ?

blockchain/blockchain.go
```diff
@@ -215,10 +213,9 @@ func (b *Blockchain) StateUpdateByHash(hash *felt.Felt) (*core.StateUpdate, erro
 	return core.GetStateUpdateByHash(b.database, hash)
 }
 
-func (b *Blockchain) L1HandlerTxnHash(msgHash *common.Hash) (*felt.Felt, error) {
+func (b *Blockchain) L1HandlerTxnHash(msgHash *common.Hash) (felt.Felt, error) {
 	b.listener.OnRead("L1HandlerTxnHash")
-	txnHash, err := core.GetL1HandlerTxnHashByMsgHash(b.database, msgHash.Bytes())
-	return &txnHash, err // TODO: return felt value
+	return core.GetL1HandlerTxnHashByMsgHash(b.database, msgHash.Bytes()) // TODO: return felt value
```
rodrodros: Remove `todo`

core/state/history_test.go
```diff
@@ -20,12 +20,6 @@ func TestNewStateHistory(t *testing.T) {
 		assert.Equal(t, uint64(0), history.blockNum)
 		assert.NotNil(t, history.state)
 	})
-
-	t.Run("invalid state root", func(t *testing.T) {
-		invalidRoot := new(felt.Felt).SetUint64(999) // Non-existent root
-		_, err := NewStateHistory(1, invalidRoot, stateDB)
-		assert.Error(t, err)
-	})
 }
```
rodrodros, line 22: Why are we removing this test?

blockchain/blockchain.go
MaksymMalicki: ahh I just copied it as a marker. tbh, after giving it some thought, I don't see any pros or cons of exposing this. We don't need it outside of the blockchain package, so lets keep it private

blockchain/blockchain.go
MaksymMalicki: done

core/state/history_test.go
MaksymMalicki: Because I modified the state history constructor. Previously, the state was created using `state.New()`. All the state related initializations had to be done, including fetching the tries. Since the new state in PathDB can only fetch the tries from the last 128 cached layers and the disk layer by the state commitment hash, this test is invalid. All the methods from the state object, required for reading the historical state could be technically static, as they only read from the DB - doing all the state initializations and safety checks is a bug

blockchain/blockchain.go
```diff
@@ -87,21 +89,23 @@ var _ Reader = (*Blockchain)(nil)
...
-func New(database db.KeyValueStore, network *utils.Network) *Blockchain {
+func New(database db.KeyValueStore, network *utils.Network, stateVersion bool) *Blockchain {
 	trieDB, err := triedb.New(database, nil) // TODO: handle hashdb
 	if err != nil {
 		panic(err)
 	}
 	stateDB := state.NewStateDB(database, trieDB)
 
-	stateFactory, err := commonstate.NewStateFactory(true, trieDB, stateDB)
+	StateFactory, err := commonstate.NewStateFactory(stateVersion, trieDB, stateDB)
```
rodrodros: lower case var name (`StateFactory` to `stateFactory`)

blockchain/blockchain.go
rodrodros, line 1: We should apply the `deprecated` prefix the same as we did in as in #3017

blockchain/blockchain.go
```diff
@@ -272,13 +268,29 @@ func (b *Blockchain) SetL1Head(update *core.L1Head) error {
 // Store takes a block and state update and performs sanity checks before putting in the database.
 func (b *Blockchain) Store(block *core.Block, blockCommitments *core.BlockCommitments,
 	stateUpdate *core.StateUpdate, newClasses map[felt.Felt]core.Class,
+) error {
+	// old state
+	// TODO(maksymmalick): remove this once we have a new state implementation
+	if !b.StateFactory.UseNewState {
+		return b.storeCoreState(block, blockCommitments, stateUpdate, newClasses)
```
rodrodros: This should be called `deprecatedStore`, also I don't think the "core" descriptor shouldn't be use in function naming, is there any `state` which is not "core"?

blockchain/blockchain.go
```diff
@@ -595,28 +779,46 @@ func (b *Blockchain) updateStateRoots(
 	block *core.Block,
 	stateUpdate *core.StateUpdate,
 	newClasses map[felt.Felt]core.Class,
+	flushChanges bool,
```
rodrodros, line 809: Why does the blockchain can decide when the state flushes its changes, shouldn't this logic be self contained within the State package 🤔. It feels like incorrect design but I might be wrong as well

adapters/p2p2core/state.go
MaksymMalicki: done

blockchain/blockchain.go
MaksymMalicki: my bad

PR #3030, opened 2025-08-04
chore: update Juno dependencies

rodrigo-pino:


rodrigo-pino:
Replaced by https://github.com/NethermindEth/juno/pull/3061

PR #3044, opened 2025-08-08
Tendermint: implement a sync mechanism

rianhughes:
This PR is a continuation of https://github.com/NethermindEth/juno/pull/3024. Creating a fresh PR was easier than rebaseing on main.


consensus/sync/sync_test.go
```diff
@@ -0,0 +1,171 @@
...
+		utils.NewNopZapLogger(),
+		newTendermintDB(t),
+		stateMachine,
+		mockCommitListener,
+		p2p,
+		mockTimeoutFn,
+	)
+
+	go func() {
+		require.NoError(t, driver.Run(ctx))
```
infrmtcs, line 100: We don't need the driver here anymore, because the subject under test writes data to the channel, so it doesn't need to know about driver.

p2p/sync/sync.go
```diff
@@ -122,28 +128,20 @@ func (s *Service) processBlock(ctx context.Context, blockNumber uint64) error {
...
+		s.processSpecBlockParts(
+			ctx,
+			blockNumber,
+			pipeline.FanIn(ctx,
+				pipeline.Stage(ctx, headersAndSigsCh, specBlockPartsFunc[specBlockHeaderAndSigs]),
+				pipeline.Stage(ctx, classesCh, specBlockPartsFunc[specClasses]),
+				pipeline.Stage(ctx, stateDiffsCh, specBlockPartsFunc[specContractDiffs]),
+				pipeline.Stage(ctx, txsCh, specBlockPartsFunc[specTxWithReceipts]),
+				pipeline.Stage(ctx, eventsCh, specBlockPartsFunc[specEvents]),
+			)))
```
infrmtcs: nit: I think we can separate the closing brackets into separate lines

consensus/votecounter/vote_counter.go
```diff
@@ -111,7 +111,7 @@ func (v *VoteCounter[V, H, A]) AddProposal(proposal *types.Proposal[V, H, A]) bo
 		return false
 	}
 
-	if expectedProposer := v.validators.Proposer(proposal.Height, proposal.Round); proposal.Sender != expectedProposer {
+	if expectedProposer := v.validators.Proposer(proposal.Height, proposal.Round); types.AddrCmp(proposal.Sender, expectedProposer) {
```
infrmtcs: What happens if we revert this change? Can you describe a case where the original code doesn't work?

consensus/sync/mock_p2p_sync_test.go
```diff
@@ -0,0 +1,28 @@
...
+
+type mockP2PSyncService struct {
+	blockCh chan sync.BlockBody
+}
+
+func newMockP2PSyncService(blockCh chan sync.BlockBody) mockP2PSyncService {
+	return mockP2PSyncService{blockCh: blockCh}
+}
+
+func (m *mockP2PSyncService) recieveBlockOverP2P(block sync.BlockBody) {
```
infrmtcs: nit:
```suggestion
func (m *mockP2PSyncService) receiveBlockOverP2P(block sync.BlockBody) {
```

consensus/sync/sync_test.go
```diff
@@ -0,0 +1,171 @@
...
+	stateMachine := tendermint.New(tendermintDB, logger, nodeAddr, mockApp, allNodes, types.Height(0))
+
+	proposalCh := make(chan starknet.Proposal)
+	prevoteCh := make(chan starknet.Prevote)
+	precommitCh := make(chan starknet.Precommit)
+	p2p := newMockP2P(proposalCh, prevoteCh, precommitCh)
+
+	driver := driver.New(
+		utils.NewNopZapLogger(),
+		newTendermintDB(t),
```
infrmtcs: Why do we need to create TendermintDB twice here, one is using mock and one is using in-memory? I think we can remove the mock here.

utils/pipeline/pipeline_test.go
```diff
@@ -40,8 +40,9 @@ func TestPipeline(t *testing.T) {
 			assert.False(t, open)
 		})
 		t.Run("Bridge", func(t *testing.T) {
-			_, open := <-Bridge[int](ctx, nil)
-			assert.False(t, open)
+			out := make(chan int)
+			chOfInts := make(chan (<-chan int))
+			go Bridge[int](ctx, out, chOfInts) // non-blocking
```
infrmtcs: What's the purpose of this test if we don't assert anything here?

consensus/sync/sync.go
```diff
@@ -0,0 +1,100 @@
...
+					return nil
+				case s.driverPrecommitCh <- precommit:
+				}
+			}
+
+			proposal := types.Proposal[V, H, A]{
+				MessageHeader: types.MessageHeader[A]{
+					Height: types.Height(committedBlock.Block.Number),
+					Round:  syncRoundPlaceHolder,
+					Sender: A(committedBlock.Block.SequencerAddress.Bits()),
```
infrmtcs:
```suggestion
					Sender: A(committedBlock.Block.SequencerAddress),
```
This is the implementation of the `Bits()` function:

```
// Bits provides access to z by returning its value as a little-endian [4]uint64 array.
// Bits is intended to support implementation of missing low-level Element
// functionality outside this package; it should be avoided otherwise.
func (z *Element) Bits() [4]uint64 {
	_z := *z
	fromMont(&_z)
	return _z
}
```

consensus/votecounter/vote_counter.go
infrmtcs: Probably this is the reason: https://github.com/NethermindEth/juno/pull/3044/files#r2262328559

consensus/sync/node_test.go
```diff
@@ -0,0 +1,30 @@
...
+	return types.VotingPower(len(n))
+}
+
+func (n nodes) ValidatorVotingPower(height types.Height, addr *starknet.Address) types.VotingPower {
+	return types.VotingPower(1)
+}
+
+// Randomised proposer selection, with prime coefficients so that for each height, the order of proposers is different.
+func (n nodes) Proposer(height types.Height, round types.Round) starknet.Address {
+	f := new(felt.Felt).SetUint64(5)
```
infrmtcs: This is not randomised. I think we either update the comment or properly use random here.

consensus/sync/sync_test.go
```diff
@@ -0,0 +1,171 @@
...
+	mockP2PSyncService := newMockP2PSyncService(mockInCh)
+
+	consensusSyncService := consensusSync.New(mockP2PSyncService.Listen(), proposalCh, precommitCh, getPrecommits, toValue, &proposalStore)
+
+	block0 := getCommittedBlock()
+	block0Hash := block0.Block.Hash
+	valueHash := toValue(block0Hash).Hash()
+	go func() {
+		mockP2PSyncService.recieveBlockOverP2P(block0)
+	}()
```
infrmtcs:
```suggestion
	consensusSyncService := consensusSync.New(inputBlock, outputProposal, outputPrecommits, getPrecommits, toValue, &proposalStore)

	block0 := getCommittedBlock()
	block0Hash := block0.Block.Hash
	valueHash := toValue(block0Hash).Hash()
	inputBlock <- block0
```
and make `mockInCh` a channel with buffer size 1.

consensus/votecounter/vote_counter.go
rianhughes:
```
if expectedProposer := v.validators.Proposer(proposal.Height, proposal.Round); proposal.Sender != expectedProposer {
		fmt.Println(" proposal.Sender", proposal.Sender)
		fmt.Println(" expectedProposer", expectedProposer)
		...
		}
```

prints

```
proposal.Sender [5 0 0 0]
 expectedProposer [18446744073709551457 18446744073709551615 18446744073709551615 576460752303420784]
 ```
 
The `expectedProposer` is in Montgomery form, whereas the `proposal.Sender` is in the raw / non-Montgomery form. We need a function that compares them in the same form.

File: consensus/sync/sync_test.go [hunk 10]
```diff
@@ -0,0 +1,171 @@
...
+func TestSync(t *testing.T) {
+	ctrl := gomock.NewController(t)
+
+	ctx, cancel := context.WithCancel(t.Context())
+
+	nodeAddr := starknet.Address(felt.FromUint64(123))
+	logger := utils.NewNopZapLogger()
+	tendermintDB := newDB(t)
+	proposalStore := proposal.ProposalStore[starknet.Hash]{}
+	allNodes := newNodes(4) // Proposer set to 5
```
infrmtcs: What is the purpose of having 4 validators but setting the proposer to 5, which is outside of the validator set?

consensus/votecounter/vote_counter.go
rianhughes: > Probably this is the reason: https://github.com/NethermindEth/juno/pull/3044/files#r2262328559

Yes, this is the source of the problem. I cam update it, although I wonder if it's safer to leave the `AddrCmp` function in

utils/pipeline/pipeline_test.go
infrmtcs: If we want to assert that the function is running and waiting for input, we can actually do something like this:
```suggestion
			go func() {
						Bridge[int](ctx, out, chOfInts)
						select {
						case t.Context().Done():
						default:
									assert.FailNow(...)
						}
			}()
```
Although personally I'm not sure if this is needed, compared to just simply remove this test.

consensus/sync/sync_test.go
rianhughes: Yup, I've removed the duplicate

consensus/sync/sync.go
rianhughes: Ah, given they suggest it should be avoided, I've removed its use from the code

utils/pipeline/pipeline_test.go
rianhughes: It ensures the Bridge function returns if the context is cancelled

consensus/sync/sync_test.go
rianhughes: The suggested approach is more concise, but I prefer the current approach as it also helps describe how the service is used in practice, ie it forms as a type of documentation

consensus/sync/sync_test.go
```diff
@@ -0,0 +1,159 @@
...
+}
+
+func TestSync(t *testing.T) {
+	ctrl := gomock.NewController(t)
+
+	ctx, cancel := context.WithCancel(t.Context())
+
+	nodeAddr := starknet.Address(felt.FromUint64(123))
+	logger := utils.NewNopZapLogger()
+	tendermintDB := newDB(t)
```
infrmtcs: Why don't we just use the in memory version as suggested in the previous comment? It's simpler, shorter and reflects the behaviour better than the mock version?

consensus/sync/node_test.go
rianhughes: I've reverted to the randomised version

consensus/sync/sync_test.go
rianhughes: I was investigating a bug from before, but I've reverted to the randomised version

consensus/votecounter/vote_counter.go
infrmtcs: I don't think it's safer, but open more room for error because:
1. The original comparison does exactly what it's expected to do while it's simple, easy to read compare to going through another function `addrToFelt`.
2. The original implementation using `Bits` prints the different values because the implementation provides the wrong values. In fact, the test is supposed to fail in this case, because the values don't match.
3. I'm trying to print out as you're doing above and the test still passed even if I add back `Bits()`. So the problem here is about the tests, not the comparison.

infrmtcs: Screenshot From 2025-08-08 16-16-14
Screenshot From 2025-08-08 16-16-51

consensus/sync/sync_test.go
rianhughes: I'm not sure if it's significantly shorter and simpler, and it is slower, but I agree it reflects the behaviour better than the mock. I've swapped out the mock for the real DB

consensus/votecounter/vote_counter.go
infrmtcs: Found the issue. Inside `AddrCmp`:
```
return af.Cmp(&bf) == 0
```
which means it return `true` if they're the same. However, in the call site, we need to check if the addresses are different in order to return early. This means we need NOT `AddrCmp` instead.

This is the exact reason why I think if `!=` does the job, use `!=` instead.

utils/pipeline/pipeline_test.go
infrmtcs: Try running this:
```
func TestInfiniteLoop(t *testing.T) {
	go func() {
		for {
			time.Sleep(1 * time.Second)
		}
	}()
}
```

The test still pass even if the function never return.

consensus/votecounter/vote_counter.go
rianhughes: Actually, it seems to be causing problems with the vote counter tests for some reason, so I've removed it

infrmtcs (approved):
LGTM up to the remaining 2 comments

p2p/sync/sync.go
rianhughes: Updated

consensus/sync/sync_test.go
rianhughes: mockP2PSyncService has been removed

utils/pipeline/pipeline_test.go
rianhughes: Ah yeah, the Bridge function probably shouldn't run in a separate routine. I've updated the test so it doesn't run in a separate routine. And confirmed it blocks if the cancel is not called.

PR #3045, opened 2025-08-11
refactor: Test utils for p2p transaction builder

infrmtcs:


adapters/testutils/transaction_test_utils.go
```diff
@@ -0,0 +1,352 @@
...
+	"github.com/starknet-io/starknet-p2pspecs/p2p/proto/common"
+	"github.com/starknet-io/starknet-p2pspecs/p2p/proto/transaction"
+	"github.com/stretchr/testify/require"
+)
+
+const maxTransactionSize = 10
+
+type toCoreType[C any] func(core.Transaction, core.Class, *felt.Felt) C
+
+type toP2PType[P, I any] func(I, *common.Hash) P
```
rodrodros, line 23: I am missing what does generic `I` means 🤔

adapters/testutils/transaction_test_utils.go
```diff
@@ -0,0 +1,352 @@
...
+	p2pTransactions := make([]P, len(factories))
+
+	for i := range factories {
+		consensusTransactions[i], p2pTransactions[i] = factories[i](t)
+	}
+
+	return consensusTransactions, p2pTransactions
+}
+
+func getRandomFelt(t *testing.T) (felt.Felt, []byte) {
```
rodrodros, line 47: Nit. Shouldn't we add the `t.Helper()` call at the beginning of each of these functions?

infrmtcs: I added `t.Helper()` to all functions. Basically they're helper functions, so shouldn't appear in the test trace.

adapters/testutils/transaction_test_utils.go
infrmtcs: The compiled definition of protobuf looks roughly like this:
```
type MempoolTransaction struct {
    Txn struct {
        DeclareV3 struct {
            DeclareV3 *DeclareV3WithClass
        }
        DeployAccountV3 struct {
            DeployAccountV3 *DeployAccountV3
        }
        InvokeV3 struct {
            InvokeV3 *InvokeV3
        }
    }
    TransactionHash *common.Hash
}

type ConsensusTransaction struct {
    Txn struct {
        DeclareV3 struct {
            DeclareV3 *DeclareV3WithClass
        }
        DeployAccountV3 struct {
            DeployAccountV3 *DeployAccountV3
        }
        InvokeV3 struct {
            InvokeV3 *InvokeV3
        }
        L1Handler struct {
            L1Handler *L1HandlerV0
        }
    }
    TransactionHash *common.Hash
}
```
The most inner types (`DeclareV3WithClass`, `DeployAccountV3`, `InvokeV3`, `L1HandlerV0`, `common.Hash`) are all shared types. However, all of the outer types are separate types defined in 2 different package. 
These util functions exist to solve this problem: given the inner transaction (e.g, `InvokeV3`) and the hash, return the `ConsensusTransaction` or `MempoolTransaction`. So we don't have to repeat the logic to generate the inner transaction twice, while we only have to implement the wrapper to build `ConsensusTransaction` and `MempoolTransaction`.

PR #3046, opened 2025-08-11
feat: Implement Mempool P2P protocol

infrmtcs:


adapters/mempool2p2p/transaction.go
```diff
@@ -0,0 +1,64 @@
...
+	"github.com/NethermindEth/juno/adapters/core2p2p"
+	"github.com/NethermindEth/juno/core"
+	"github.com/NethermindEth/juno/core/felt"
+	"github.com/NethermindEth/juno/mempool"
+	"github.com/starknet-io/starknet-p2pspecs/p2p/proto/common"
+	mempooltransaction "github.com/starknet-io/starknet-p2pspecs/p2p/proto/mempool/transaction"
+	"github.com/starknet-io/starknet-p2pspecs/p2p/proto/transaction"
+)
+
+func toHash(felt felt.Felt) *common.Hash {
```
rodrodros: `felt` by ref 👀

adapters/mempool2p2p/transaction.go
```diff
@@ -0,0 +1,64 @@
...
+	feltBytes := felt.Bytes()
+	return &common.Hash{Elements: feltBytes[:]}
```
rodrodros, line 18: Return by value 👀 ?

mempool/p2p/transaction_listener.go
```diff
@@ -0,0 +1,39 @@
...
+		transaction, err := p2p2mempool.AdaptTransaction(&p2pTransaction)
+		if err != nil {
+			log.Errorw("unable to convert transaction message to transaction", "error", err)
+			return
+		}
+
+		if err := pool.Push(ctx, transaction); err != nil {
+			log.Errorw("unable to push transaction to mempool", "error", err)
+		}
+	}
```
rodrodros: Is it ok not to propagate the error? 

I would expect for the caller to decide how to react to what happens on `onMessage`?

mempool/p2p/transaction_broadcaster.go
```diff
@@ -0,0 +1,38 @@
...
+}
+
+func (b *transactionBroadcaster) Broadcast(ctx context.Context, message *mempool.BroadcastedTransaction) {
+	msg, err := mempool2p2p.AdaptTransaction(message)
+	if err != nil {
+		b.log.Errorw("unable to convert transaction", "error", err)
+		return
+	}
+
+	b.ProtoBroadcaster.Broadcast(ctx, msg)
```
rodrodros, line 37: I am not sure we should silently fail here as well. Why not return the error?

adapters/p2p2mempool/transaction.go
```diff
@@ -0,0 +1,71 @@
...
+	if t.Txn == nil {
+		return errors.New("txn must be set to a valid variant")
+	}
+	if t.TransactionHash == nil {
+		return errors.New("transaction_hash must be set")
+	}
+	return nil
+}
+
+func AdaptTransaction(t *mempooltransaction.MempoolTransaction) (*mempool.BroadcastedTransaction, error) {
```
rodrodros: Why not return by value here?

adapters/mempool2p2p/transaction.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+func AdaptTransaction(tx *mempool.BroadcastedTransaction) (*mempooltransaction.MempoolTransaction, error) {
```
rodrodros, line 21: Return by value?

adapters/mempool2p2p/transaction.go
infrmtcs: This is from protobuf, which contains a lock and must not be copied.

adapters/mempool2p2p/transaction.go
infrmtcs: This is from protobuf, which contains a lock and must not be copied.

adapters/p2p2mempool/transaction.go
infrmtcs: Done

adapters/mempool2p2p/transaction.go
infrmtcs: Done, I also updated the original code from `consensus2p2p` and `p2p2consensus` packages.

mempool/p2p/transaction_listener.go
infrmtcs: The caller of this `onMessage` callback is the run loop inside `TopicSubscription`. This component's job is to subscribe to the topic and then repeat indefinitely: receive message -> deserialize it -> call the callback. Any failure shouldn't block the loop running. An example is the case where a malicious attacker send a malformed message, the test is also covering this case. This means the caller can't do anything effectively except logging the error, so I think we don't need to propagate the error here.

mempool/p2p/transaction_broadcaster.go
infrmtcs: This is a great question. 
With the current state of the code, this function only select `ctx.Done()` and enqueue `msg` to a channel, which is read by a run loop to broadcast to libp2p. There is not error to be returned here.
However, thinking more about it, I think probably the run loop is not optimal, because `Publish` function from `libp2p` already has synchronization with mutex there, so probably using channel here is not optimal and reduces concurrency. I think (in another PR), I will remove the run loop and call `Publish` directly, so we'll return the error here and all of the other call sites.

adapters/p2p2mempool/transaction.go
```diff
@@ -0,0 +1,71 @@
...
+	if t == nil {
+		return errors.New("mempool transaction is nil")
+	}
+	// TODO: consider further validation here
+	if t.Txn == nil {
+		return errors.New("txn must be set to a valid variant")
+	}
+	if t.TransactionHash == nil {
+		return errors.New("transaction_hash must be set")
+	}
```
rodrodros, line 23: We should added check here that verifies that the transaction data matches the hash 🤔

adapters/p2p2mempool/transaction.go
```diff
@@ -0,0 +1,71 @@
...
+	case *mempooltransaction.MempoolTransaction_DeclareV3:
+		tx := t.GetDeclareV3()
+		// Todo: we pass in CompiledClassHash here, but in sync we pass in ClassHash.
+		// Are we expected to compile the class hash here???
+		class := p2p2core.AdaptCairo1Class(tx.Class)
+		classHash, err := class.Hash()
+		classHashBytes := classHash.Bytes()
+		if err != nil {
+			return mempool.BroadcastedTransaction{}, err
+		}
```
rodrodros, line 45: nit: the error check should be done immediately after the statement 
```suggestion
		classHash, err := class.Hash()
		if err != nil {
			return mempool.BroadcastedTransaction{}, err
		}
		classHashBytes := classHash.Bytes()
```

mempool/p2p/p2p_broadcasters_listeners_test.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+func TestMempoolBroadcastersAndListeners(t *testing.T) {
+	logger, err := utils.NewZapLogger(utils.NewLogLevel(logLevel), true)
+	require.NoError(t, err)
+
+	transactions := make([][]mempool.BroadcastedTransaction, nodeCount)
+	iter.ForEach(transactions, func(txs *[]mempool.BroadcastedTransaction) {
+		*txs = getRandomTransactions(t)
+	})
```
rodrodros, line 50: nit: why an iterator over the traditional for loop?

mempool/p2p/p2p_broadcasters_listeners_test.go
```diff
@@ -0,0 +1,117 @@
...
+	for node := range nodeCount {
+		for index, tx := range transactions[node] {
+			txSet[tx.Transaction.Hash().String()] = origin{Source: node, Index: index}
+		}
+	}
+
+	transactionWait := conc.NewWaitGroup()
+	peerWait := conc.NewWaitGroup()
+	for index, node := range nodes {
+		logger := &utils.ZapLogger{SugaredLogger: logger.Named(fmt.Sprint(index))}
```
rodrodros, line 64: Is the logger re-initialized in each iteration intentionally?

mempool/p2p/p2p_broadcasters_listeners_test.go
infrmtcs: This iter is running the closure concurrently on all elements of the slice.

mempool/p2p/p2p_broadcasters_listeners_test.go
infrmtcs: By using `Named` with the index, the log will display the node index right beside the log level, which makes it easy to run grep to filter the log by node.

adapters/p2p2mempool/transaction.go
infrmtcs: The changes get quite big, so I made a separate PR #3068

rodrigo-pino (approved):

PR #3051, opened 2025-08-12
feat: rpcv9 ws endpoints

EgeCaner:
This PR implements, rpc 0.9.0 ws endpoints defined [here](https://github.com/starkware-libs/starknet-specs/tree/v0.9.0)

Fixes pending block event access in rpcv8 which is an issue started after bounding out of upper bound block number to latest.

**Notice**: does not support received tx stream on `starknet_subscribeNewTransactions`, this will be enabled in follow up PR

### DIFF
--- a/blockchain/event_matcher.go
+++ b/blockchain/event_matcher.go
@@ -34,7 +34,7 @@ func makeKeysMaps(filterKeys [][]felt.Felt) []map[felt.Felt]struct{} {
 	return filterKeysMaps
 }
 
-func (e *EventMatcher) matchesEventKeys(eventKeys []*felt.Felt) bool {
+func (e *EventMatcher) MatchesEventKeys(eventKeys []*felt.Felt) bool {
 	// short circuit if event doest have enough keys
 	for i := len(eventKeys); i < len(e.keysMap); i++ {
 		if len(e.keysMap[i]) > 0 {
@@ -148,7 +148,8 @@ func (e *EventMatcher) AppendBlockEvents(matchedEventsSofar []*FilteredEvent, he
 		for i, event := range receipt.Events {
 			var blockNumber *uint64
 			// if header.Hash == nil it's a pending block
-			if header.Hash != nil {
+			// if header.Hash == nil and header.ParentHash is nil preconfirmed block
+			if header.Hash != nil || header.ParentHash == nil {
 				blockNumber = &header.Number
 			}
 
@@ -164,7 +165,7 @@ func (e *EventMatcher) AppendBlockEvents(matchedEventsSofar []*FilteredEvent, he
 				continue
 			}
 
-			if !e.matchesEventKeys(event.Keys) {
+			if !e.MatchesEventKeys(event.Keys) {
 				processedEvents++
 				continue
 			}
--- a/rpc/v8/subscriptions.go
+++ b/rpc/v8/subscriptions.go
@@ -202,11 +202,13 @@ func (h *Handler) SubscribeEvents(
 
 	nextBlock := headHeader.Number + 1
 	eventsPreviouslySent := make(map[SentEvent]struct{})
-
+	pendingID := &BlockID{typeID: pending}
 	subscriber := subscriber{
 		onStart: func(ctx context.Context, id string, _ *subscription, _ any) error {
+			fromB := BlockIDFromNumber(requestedHeader.Number)
+			toB := BlockIDFromNumber(headHeader.Number)
 			return h.processEvents(
-				ctx, w, id, requestedHeader.Number, headHeader.Number, fromAddr, keys, nil,
+				ctx, w, id, &fromB, &toB, fromAddr, keys, nil, headHeader.Number,
 			)
 		},
 		onReorg: func(
@@ -219,8 +221,10 @@ func (h *Handler) SubscribeEvents(
 			return nil
 		},
 		onNewHead: func(ctx context.Context, id string, _ *subscription, head *core.Block) error {
+			fromB := BlockIDFromNumber(nextBlock)
+			toB := BlockIDFromNumber(head.Number)
 			err := h.processEvents(
-				ctx, w, id, nextBlock, head.Number, fromAddr, keys, eventsPreviouslySent,
+				ctx, w, id, &fromB, &toB, fromAddr, keys, eventsPreviouslySent, head.Number,
 			)
 			if err != nil {
 				return err
@@ -234,7 +238,15 @@ func (h *Handler) SubscribeEvents(
 			}
 
 			return h.processEvents(
-				ctx, w, id, nextBlock, nextBlock, fromAddr, keys, eventsPreviouslySent,
+				ctx,
+				w,
+				id,
+				pendingID,
+				pendingID,
+				fromAddr,
+				keys,
+				eventsPreviouslySent,
+				pending.GetBlock().Number-1,
 			)
 		},
 	}
@@ -350,8 +362,16 @@ func (h *Handler) checkTxStatus(
 	return status.Finality, nil
 }
 
-func (h *Handler) processEvents(ctx context.Context, w jsonrpc.Conn, id string, from, to uint64, fromAddr *felt.Felt,
-	keys [][]felt.Felt, eventsPreviouslySent map[SentEvent]struct{},
+// processEvents queries database for events and stream filtered events.
+func (h *Handler) processEvents(
+	ctx context.Context,
+	w jsonrpc.Conn,
+	id string,
+	from, to *BlockID,
+	fromAddr *felt.Felt,
+	keys [][]felt.Felt,
+	eventsPreviouslySent map[SentEvent]struct{},
+	height uint64,
 ) error {
 	filter, err := h.bcReader.EventFilter(fromAddr, keys, h.PendingBlock)
 	if err != nil {
@@ -361,9 +381,7 @@ func (h *Handler) processEvents(ctx context.Context, w jsonrpc.Conn, id string, 
 
 	defer h.callAndLogErr(filter.Close, "Error closing event filter in events subscription")
 
-	fromBlockID := BlockIDFromNumber(from)
-	toBlockID := BlockIDFromNumber(to)
-	err = setEventFilterRange(filter, &fromBlockID, &toBlockID, to)
+	err = setEventFilterRange(filter, from, to, height)
 	if err != nil {
 		h.log.Warnw("Error setting event filter range", "err", err)
 		return err
--- a/rpc/v8/subscriptions.go
+++ b/rpc/v8/subscriptions.go
@@ -375,40 +375,34 @@ func (h *Handler) processEvents(
 ) error {
 	filter, err := h.bcReader.EventFilter(fromAddr, keys, h.PendingBlock)
 	if err != nil {
-		h.log.Warnw("Error creating event filter", "err", err)
 		return err
 	}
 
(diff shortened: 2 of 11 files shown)


rpc/v9/subscriptions.go
```diff
@@ -204,50 +218,274 @@ func (h *Handler) SubscribeEvents(
 	if rpcErr != nil {
 		return "", rpcErr
 	}
+	// default to ACCEPTED_ON_L2
+	if finalityStatus == nil {
+		finalityStatus = utils.HeapPtr(TxnFinalityStatusWithoutL1(TxnAcceptedOnL2))
```
rodrodros, line 223: Does this work?
```suggestion
		finalityStatus = &TxnFinalityStatusWithoutL1(TxnAcceptedOnL2)
```

rpc/v9/subscriptions.go
```diff
@@ -204,50 +218,274 @@ func (h *Handler) SubscribeEvents(
...
+	height uint64,
+	finalityStatus TxnFinalityStatusWithoutL1,
+) error {
+	filter, err := h.bcReader.EventFilter(fromAddr, keys, h.PendingBlock)
+	if err != nil {
+		h.log.Warnw("Error creating event filter", "err", err)
+		return err
+	}
+
+	defer h.callAndLogErr(filter.Close, "Error closing event filter in events subscription")
```
rodrodros: Nit. Should this error message should start with upper caps, also, should it start with the word "Error" 🤔. Based on the code of `callAndLogErr` I think we should do:
```suggestion
	defer h.callAndLogErr(filter.Close, "closing event filter in events subscription")
```

rpc/v9/subscriptions.go
```diff
@@ -204,50 +218,274 @@ func (h *Handler) SubscribeEvents(
...
+	from, to *BlockID,
+	fromAddr *felt.Felt,
+	keys [][]felt.Felt,
+	eventsPreviouslySent map[SentEvent]TxnFinalityStatusWithoutL1,
+	height uint64,
+	finalityStatus TxnFinalityStatusWithoutL1,
+) error {
+	filter, err := h.bcReader.EventFilter(fromAddr, keys, h.PendingBlock)
+	if err != nil {
+		h.log.Warnw("Error creating event filter", "err", err)
```
rodrodros: Nit. On the same thought process as the previous comment, I think a more appropiate error message "cannot create event filter"

rpc/v9/subscriptions.go
```diff
@@ -204,50 +218,274 @@ func (h *Handler) SubscribeEvents(
...
+		if err != nil {
+			h.log.Warnw("Error filtering events", "err", err)
+			return err
+		}
+
+		err = sendEvents(ctx, w, filteredEvents, eventsPreviouslySent, id, finalityStatus)
+		if err != nil {
+			h.log.Warnw("Error sending events", "err", err)
+			return err
+		}
```
rodrodros: Nit. Similar comments for this warning logs.

Why do we log them if we are returning them 🤔, by returning doesn't that mean they will also be logged else were. 

Isn't it better to log the error outside this function? I am not convinced they should be logged here

rpc/v9/subscriptions.go
```diff
@@ -204,50 +218,274 @@ func (h *Handler) SubscribeEvents(
...
+				BlockNumber:     blockNumber,
+				BlockHash:       block.Hash,
+				TransactionHash: receipt.TransactionHash,
+				EventIndex:      i,
+				Event:           event,
+			}
+
+			if err := sendEventWithoutDuplicate(w, &event, eventsPreviouslySent, id, finalityStatus); err != nil {
+				return err
+			}
```
rodrodros, line 405: For example here, we are just returning the error without any explicit logging

rpc/v9/subscriptions.go
```diff
@@ -295,7 +533,7 @@ func (h *Handler) SubscribeTransactionStatus(ctx context.Context, txHash *felt.F
 
 // If the error is transaction not found that means the transaction has not been submitted to the feeder gateway,
 // therefore, we need to wait for a specified time and at regular interval check if the transaction has been found.
-// If the transaction is found during the timout expiry, then we continue to keep track of its status otherwise the
+// If the transaction is found during the timeout expiry, then we continue to keep track of its status otherwise the
```
rodrodros, line 530: 👍

rpc/v9/subscriptions.go
```diff
@@ -335,7 +573,7 @@ func (h *Handler) checkTxStatus(
 		return lastStatus, errorTxnHashNotFound{*txHash}
 	}
 
-	if status.Finality <= lastStatus {
+	if status.Finality == lastStatus {
```
rodrodros, line 570: How does this comparison operator changes the flow of the function 🤔

rpc/v9/transaction.go
```diff
@@ -924,10 +924,10 @@ func AdaptReceipt(receipt *core.TransactionReceipt, txn core.Transaction, finali
 	}
 
 	var receiptBlockNumber *uint64
-	// TODO(Ege): case for preconfirmed blocks: they don't have blockHash
-	// but they do have block number, so far we were not returning blocknumber for pending block
-	// pending block didnt had blocknumber, clarify whether we need to return number or not
-	if blockHash != nil {
+
+	// Do not return block number for pending block
+	// Return block number for canonical blocks and pending block
```
rodrodros: I am confused by these two comments, they seem contradictory regarding pending block number

rpc/v9/subscriptions.go
EgeCaner: I agree, it seems unnecessary to log here, I will update. I reused this part from previous implementation.

rpc/v9/transaction.go
EgeCaner: My bad second one should be pre_confirmed, I will correct

rpc/v9/subscriptions.go
EgeCaner: In tx status subscription, we keep track of sent status for given txnHash. Previous behavior was sending hardest finality status we ever seen once due to `actualStatus <= lastStatus` check, and sent statuses only increase in finality guarantees. `actualStatus == lastStatus` check allow status regression `CANDIDATE` -> `RECEIVED`, and sends status to subscriber if its changed either way

rpc/v9/subscriptions.go
EgeCaner: >invalid operation: cannot take address of TxnFinalityStatusWithoutL1(TxnAcceptedOnL2) (constant 4 of uint8 type TxnFinalityStatusWithoutL1)

:(

rpc/v9/subscriptions.go
EgeCaner: addressed

rpc/v9/subscriptions.go
EgeCaner: I removed the logs

rpc/v9/transaction.go
EgeCaner: corrected as 
>// Do not return block number for pending block
	// Return block number for canonical blocks and pre_confirmed block

blockchain/event_matcher.go
```diff
@@ -148,7 +148,8 @@ func (e *EventMatcher) AppendBlockEvents(matchedEventsSofar []*FilteredEvent, he
 		for i, event := range receipt.Events {
 			var blockNumber *uint64
 			// if header.Hash == nil it's a pending block
-			if header.Hash != nil {
+			// if header.Hash == nil and header.ParentHash is nil preconfirmed block
```
infrmtcs, line 151: Do we need the `header.ParentHash == nil` condition? In both case `header.Hash` is nil.

EgeCaner: Pending block has parentHash and pre_confirmed does not have. I added the `parentHash` check to differentiate `pending` block from `pre_confirmed`. We dont assign block number for `pending` block but we should assign for `pre_confirmed` block. 

if header.Hash is nil it is either pending or pre_confirmed and if header.Hash is nil and parent hash is nil, it should be pre_confirmed block.Not sure genesis is an exception tho

rpc/v9/subscriptions.go
rodrodros: Ah gotcha, looking deep, it is dumb design and requires some design changes to fix correctly

rpc/v8/subscriptions.go
```diff
@@ -202,11 +202,13 @@ func (h *Handler) SubscribeEvents(
 
 	nextBlock := headHeader.Number + 1
 	eventsPreviouslySent := make(map[SentEvent]struct{})
-
+	pendingID := &BlockID{typeID: pending}
```
rodrodros: The only reason we don't have a constructor for this is that at the time it was unnecessary. I think we can create a PendingBlockID constructor (which returns by value) and use it here.

EgeCaner: I added constructor for pending and preconfirmed blockID

rpc/v9/subscriptions.go
```diff
@@ -718,16 +726,402 @@ func (h *Handler) Unsubscribe(ctx context.Context, id string) (bool, *jsonrpc.Er
...
+		onPendingData: func(ctx context.Context, id string, sub *subscription, pending core.PendingData) error {
+			if pending == nil {
+				return nil
+			}
+			block := pending.GetBlock()
+			switch pending.Variant() {
+			case core.PendingBlockVariant:
+				if !slices.Contains(finalityStatuses, TxnFinalityStatusWithoutL1(TxnAcceptedOnL2)) {
+					return nil
+				}
```
infrmtcs, line 816: We can move this outside of the switch to avoid duplicate

rpc/v9/subscriptions.go
```diff
@@ -718,16 +726,402 @@ func (h *Handler) Unsubscribe(ctx context.Context, id string) (bool, *jsonrpc.Er
...
+
+				return processBlockReceipts(
+					id,
+					w,
+					senderAddress,
+					block,
+					receiptsPreviouslySent,
+					TxnFinalityStatusWithoutL1(TxnAcceptedOnL2),
+				)
```
infrmtcs: We can move this outside of the switch to avoid duplicate

rpc/v9/subscriptions.go
```diff
@@ -204,50 +218,268 @@ func (h *Handler) SubscribeEvents(
...
+	finalityStatus TxnFinalityStatusWithoutL1,
+) error {
+	if isMatch := eventMatcher.TestBloom(block.EventsBloom); !isMatch {
+		return nil
+	}
+
+	var blockNumber *uint64
+	// if header.Hash == nil and parentHash != nil it's a pending block
+	// if header.Hash == nil and parentHash == nil it's a pre_confirmed block
+	if block.Hash != nil || block.ParentHash == nil {
```
infrmtcs: nit:
```suggestion
	if isNotPending := block.Hash != nil || block.ParentHash == nil; isNotPending {
```
feel free to ignore

rpc/v9/subscriptions.go
EgeCaner: This check belongs to this case. for pending we check if user requested `ACCEPTED_ON_L2` and for pre_confirmed we check if user requested `PRE_CONFIRMED` then stream accordingly

rpc/v9/subscriptions.go
EgeCaner: I moved it outside switch

PR #3054, opened 2025-08-12
fix(rpcv8/v9): use snapshot in the HeadState, to retrieve tries consistently

MaksymMalicki:


rpc/v8/storage.go
```diff
@@ -67,6 +67,8 @@ func (h *Handler) StorageProof(
 		return nil, rpccore.ErrInternal.CloneWithData(err)
 	}
 
+	defer h.callAndLogErr(closer, "Error closing state reader in getStorageProof")
```
EgeCaner: Error messages start with lowercase, I guess same should apply here

rpc/v9/storage.go
```diff
@@ -67,6 +67,8 @@ func (h *Handler) StorageProof(
 		return nil, rpccore.ErrInternal.CloneWithData(err)
 	}
 
+	defer h.callAndLogErr(closer, "Error closing state reader in getStorageProof")
```
EgeCaner: Error messages start with lowercase, I guess same should apply here

sync/sync.go
```diff
@@ -762,7 +762,7 @@ func (s *Synchronizer) PendingState() (core.StateReader, func() error, error) {
 		return nil, nil, err
 	}
 
-	return NewPendingState(pending.GetStateUpdate().StateDiff, pending.GetNewClasses(), core.NewState(txn)), noop, nil
+	return NewPendingState(pending.GetStateUpdate().StateDiff, pending.GetNewClasses(), core.NewState(txn, nil)), noop, nil
```
EgeCaner: Should we also pass db.snapshot into this. Will passing nil, result in state progress if there is new block while operating on pending state?

sync/sync.go
```diff
@@ -786,7 +786,7 @@ func (s *Synchronizer) PendingStateBeforeIndex(index int) (core.StateReader, fun
 		stateDiff.Merge(txStateDiff)
 	}
 
-	return NewPendingState(&stateDiff, pending.GetNewClasses(), core.NewState(txn)), noop, nil
+	return NewPendingState(&stateDiff, pending.GetNewClasses(), core.NewState(txn, nil)), noop, nil
```
EgeCaner: Same concern as `PendingState`

core/trie/storage.go
```diff
@@ -30,14 +30,16 @@ func getBuffer() *bytes.Buffer {
...
-func NewStorage(txn db.IndexedBatch, prefix []byte) *Storage {
+func NewStorage(txn db.IndexedBatch, snapshot db.Snapshot, prefix []byte) *Storage {
 	return &Storage{
-		txn:    txn,
-		prefix: prefix,
+		txn:      txn,
+		prefix:   prefix,
+		snapshot: snapshot,
 	}
 }
```
rodrodros: I believe we have discuss this already during the daily: we will use the constructor pattern for mandatory parameters. Parameters with which the type wouldn't function and use the builder pattern for optional parameters.

Given that snapshot is not fundamental for the `Storage` functioning, I believe we should add a `func (s *Storage) WithSnapshot(...) (*Storage)` method and use it on the Storage when it makes sense

rodrodros: I think we can remove the snapshot from this constructor since it is optional

sync/sync.go
```diff
@@ -786,7 +790,11 @@ func (s *Synchronizer) PendingStateBeforeIndex(index int) (core.StateReader, fun
 		stateDiff.Merge(txStateDiff)
 	}
 
-	return NewPendingState(&stateDiff, pending.GetNewClasses(), core.NewState(txn)), noop, nil
+	state := core.NewState(txn)
+	snapshot := s.db.NewSnapshot()
+	state = state.WithSnapshot(snapshot)
```
rodrodros: Lets improve the legibility a bit
```suggestion
    snapshot := s.db.NewSnapshot()
	state := core.NewState(txn).WithSnapshot(snapshot)
```

sync/sync.go
rodrodros, line 1: There is a linter error here about `var noop` which is safe to delete

core/state.go
```diff
@@ -61,7 +62,16 @@ func NewState(txn db.IndexedBatch) *State {
 	}
 }
 
+func (s *State) WithSnapshot(snapshot db.Snapshot) *State {
```
infrmtcs, line 65: Why don't we put this in constructor instead?

infrmtcs: Just see the other comment, let's discuss there

core/trie/storage.go
```diff
@@ -79,10 +85,18 @@ func (t *Storage) Get(key *BitArray) (*Node, error) {
...
+		err = t.snapshot.Get(buffer.Bytes(), func(val []byte) error {
+			node = nodePool.Get().(*Node)
+			return node.UnmarshalBinary(val)
+		})
+	} else {
+		err = t.txn.Get(buffer.Bytes(), func(val []byte) error {
+			node = nodePool.Get().(*Node)
+			return node.UnmarshalBinary(val)
+		})
+	}
```
infrmtcs, line 99: Let me know if I misunderstand something here, but this means that if `snapshot` is configured, we don't read from `txn`. This means any write done in `txn` can't be retrieved, which means `PendingState` will be incorrect.

PR #3061, opened 2025-08-15
chore: update all Juno dependencies

rodrodros:


brbrr (approved):
After looking through all the files, I haven't spotted anything suspicious, except maybe changed API in cmd/juno/dbcmd.go, but I guess that's expected.

.golangci.yaml
```diff
@@ -1,174 +1,182 @@
...
+    gocritic:
+      disabled-checks:
+        - sloppyReassign
+        - dupImport # https://github.com/go-critic/go-critic/issues/845
+        - ifElseChain
+        - octalLiteral
+        - whyNoLint
+        - unnamedResult
+        # note(rdr): when lib is called x and then you do x = x.New(). Widely used in the repo. 
+        #            Should we change it?
```
EgeCaner, line 64: As I see, this PR already fixes some importShadow issues. I think not having import shadows is better for readability. If it requires minimal effort, let's do it or schedule it for later if we decide so.

rodrodros: Yes, I initially tried doing some changes and some made sense but in the greater scope I think it was hurting readability. Let's leave it disabled then 👌

PR #3068, opened 2025-08-19
feat: Add transaction hash and class hash validation in adapters

infrmtcs:


adapters/p2p2consensus/transaction.go
```diff
@@ -1,66 +1,56 @@
...
-	}
+func AdaptTransaction(t *p2pconsensus.ConsensusTransaction, network *utils.Network) (consensus.Transaction, error) {
 	if err := validateConsensusTransaction(t); err != nil {
 		return consensus.Transaction{}, err
 	}
 
+	var (
+		tx          core.Transaction
+		class       core.Class
+		paidFeeOnL1 *felt.Felt
```
rodrodros, line 22: Why not initialize it here as zero instead of nil

adapters/testutils/transaction_test_utils.go
```diff
@@ -136,66 +136,65 @@ func getSampleClass(t *testing.T) (felt.Felt, *core.Cairo1Class) {
 	return classHash, cairo1Class
 }
 
-func (b *TransactionBuilder[C, P]) GetTestDeclareTransaction(t *testing.T) (C, P) {
+func getTransactionHash(t *testing.T, tx core.Transaction, network *utils.Network) (*felt.Felt, *common.Hash) {
```
rodrodros, line 139: t.Helper() 👀

adapters/p2p2consensus/transaction.go
infrmtcs: - Inside blockifier, the `L1Handler` transaction expect a `paid_fee_on_l1` representing the fee. However, it's not asserting anything except that this is a non-zero number ([source](https://github.com/starkware-libs/sequencer/blob/main/crates/blockifier/src/transaction/l1_handler_transaction.rs#L98-L99))
- In Juno rust code, we're expecting that if the block has N transactions and M of them are L1 handler transactions, the rust code will expect a vector of M elements for `paid_fees_on_l1`.
- That's why this field is expected to be either 1 for L1 handlers or 0 for other cases. 0 is an invalid value.

adapters/p2p2consensus/transaction.go
```diff
@@ -1,66 +1,56 @@
...
 )
 
-func AdaptTransaction(t *p2pconsensus.ConsensusTransaction) (consensus.Transaction, error) {
-	if t == nil {
-		return consensus.Transaction{}, errors.New("transaction is nil")
-	}
+func AdaptTransaction(t *p2pconsensus.ConsensusTransaction, network *utils.Network) (consensus.Transaction, error) {
 	if err := validateConsensusTransaction(t); err != nil {
 		return consensus.Transaction{}, err
 	}
```
rodrodros, line 17: I realized that this might not be ideal, but should this validate call here makes it confusing to me. 

For example, is this the right place to call validate, when this validate the only thing it does is to make sure all the respective methods of p2pConsensus are non nil.

Shouldn't this all checks be done upon receive of the message. It seems wrong to me that they are in adapters. Adapters should assume correctness of their inputs (not the validity of its data).

I cannot find where in the code this function is supposed to be called so I cannot propose a better architecture but opening the discussion here.

infrmtcs: While I agree that we should remove this `validateConsensusTransaction` (reason below), I don't think we can simply assume that the adapter input is correct and require the caller to assert it. One of the big reason is that the whether the data is correct can only be known during the conversion itself.

A great example is `SetBytesCanonical` to convert from `[]byte` to `felt.Felt`. What this function does are:
- Check if the slice has 64 elements.
- Convert the slice into `[4]uint64`.
- Check if the array is smaller than a modulus constant.

So practically it does both validation and conversion together. If we want to separate them, where validation is done by caller while conversion is done by adapter, we essentially need to do the conversion twice.

Then regarding `validateConsensusTransaction`, the check for non-nil `TransactionHash` should be done when converting the `TransactionHash`. The check for non-nil `Txn` is actually already done by the switch in this function, covered by the default case, so we can also remove it. Only the check for non-nil `t` can be relaxed, because t will be non-nil if `proto.Unmarshal` succeeds.

infrmtcs: Regarding why transaction hash validation is done here, but not at the caller after the adapter, my rationale is that: The validation done in the adapter should be self-contained. For example, the hash can be calculated by only the transaction fields. In contrast, some other validations like gas prices or calldata can't be done without the VM, or the context of last block, so they aren't included in the adapter. This is debatable though.

PR #3069, opened 2025-08-19
feat: mpmc broadcast

EgeCaner:
## Problem
Juno's internal message delivery system currently uses the Feed type (in feed/feed.go). It broadcasts events to all subscribers in a non-blocking manner by either dropping messages when subscriber buffers are full or overwriting existing values. This approach can cause slow subscribers to miss events. Without proper missing-event notifications and handling, these gaps may lead to subtle and unexpected bugs or inconsistencies.

## Proposed Solution
To guarantee that all messages reach every subscriber, two naive approaches exist:

- Block senders until all subscribers receive the message.
- Use unbounded buffers (any bounded buffer can get full theoratically and will create backpreasure).

Both approaches have significant downsides:

- Blocking senders risks halting critical processes like syncing.
- Unbounded buffers may grow without limit, leading to memory bloat.

### Possible Solutions with bounded buffers.

By having sufficiently large buffered channels act like a message queue per subscriber and terminating the subscriber when subscriber is slow and its queue is full. Simple to implement, and alligns with GOs

> Don't communicate by sharing memory; share memory by communicating.

However per-subscriber buffered channels scales worse due to more memory usage compared to shared buffer. And disconnecting slow subscribers might not be desirable.

###  Bounded Buffer with Lag Notification
This PR implements a multi-producer, multi-consumer broadcast mechanism built on an array-backed shared ring buffer.

Key properties:
- Ring buffer holds messages shared across subscribers.
- Each subscriber maintains its own cursor into the ring.
- Subscribers consume at their own pace, asynchronous from producers and peers.
- If a subscriber falls behind and the ring buffer wraps around (overwrites unread messages), it is notified about lag.
- Lag notifications allow subscribers to detect missed events and respond gracefully.
- Each subscription spawns a dedicated goroutine that reads from the ring and forwards messages (or lag notices) onto the user-facing channel. This isolates complexity, improves user experience, and supports multiplexing.
Benefits:
- Non-blocking sends without unbounded buffer growth.
- Slow subscribers are safely handled with explicit lag awareness.
- Memory usage is shared and bounded efficiently.

###  Possible optimizations

Multiple readers holding slot.lock creates backpresurre if producer tries to write to that slot.
On seq mismatch on ring, reader acquires tail lock creates backpressure.
Subsribe unsubscribe acquire subMu, create backpressure on send
Further optimize to reduce lock contention and backpressure, potentially by leveraging lock-free or atomic-based data structures. Might use something else to notify (herd wakeup) waiting goroutines rather than channels

### DIFF
--- a/broadcast/errors.go
+++ b/broadcast/errors.go
@@ -10,7 +10,8 @@ var (
 	ErrClosed = errors.New("broadcast channel closed")
 	// Returned when the requested sequence is ahead of the latest published (not yet available).
 	ErrFutureSeq = errors.New("requested sequence is not published yet")
-	// Returned when the requested sequence is 0. Which is indicates unitialized sequence in this system.
+	// Returned when the requested sequence is 0. Which is indicates unitialized sequence in this
+	// system.
 	ErrInvalidSequence = errors.New("invalid sequence: 0")
 	ErrNoEvent         = errors.New("no event present")
 	ErrNoLag           = errors.New("no lag info present")
@@ -25,5 +26,9 @@ type LaggedError struct {
 }
 
 func (e *LaggedError) Error() string {
-	return fmt.Sprintf("subscriber lagged: missed seq=%d, next available seq=%d", e.MissedSeq, e.NextSeq)
+	return fmt.Sprintf(
+		"subscriber lagged: missed seq=%d, next available seq=%d",
+		e.MissedSeq,
+		e.NextSeq,
+	)
 }


broadcast/broadcast.go
```diff
@@ -0,0 +1,283 @@
...
+		ring: newRingBuffer[T](capacity),
+		subs: make(map[uint64]*Subscription[T]),
+	}
+}
+
+// Send publishes msg to the ring and wakes subscribers.
+//   - Returns ErrClosed if closed is already set (a concurrent Close after this check may still allow this Push).
+//   - Push is serialised via ring’s tailMu.
+//   - Wakes subscribers by iterating subs under subMu.RLock and attempting non-blocking sends on their notifyC.
+func (b *Broadcast[T]) Send(msg T) error {
```
rodrodros: I think here you should get a reference to T and pass it to `Push` as a reference as well. Then `Push` can write it as a value. For small types it is the same but think of huge structs such as the one we have in Juno would make these function silently sub-optimal

broadcast/broadcast.go
```diff
@@ -0,0 +1,283 @@
...
+// Unsubscribe removes this subscriber from the broadcast (if still present) and
+// closes its control channels.
+//   - single-run via CAS.
+//   - Deletes from subs under subMu.Lock, then closes done (to stop run).
+func (sub *Subscription[T]) Unsubscribe() {
+	if !sub.isClosed.CompareAndSwap(false, true) {
+		return
+	}
+
+	if sub.bcast != nil {
```
rodrodros: Is it possible for `sub.bcast` to be nil in real world usage 🤔

broadcast/broadcast.go
```diff
@@ -0,0 +1,283 @@
...
+}
+
+// Unsubscribe removes this subscriber from the broadcast (if still present) and
+// closes its control channels.
+//   - single-run via CAS.
+//   - Deletes from subs under subMu.Lock, then closes done (to stop run).
+func (sub *Subscription[T]) Unsubscribe() {
+	if !sub.isClosed.CompareAndSwap(false, true) {
+		return
+	}
```
rodrodros: Should unsubscribing of an already closed subscription silently do nothing or return an error. I think the second might be more appropriate 🤔

broadcast/ring_buffer.go
```diff
@@ -0,0 +1,154 @@
...
+
+// LaggedError indicates the requested sequence was overwritten by newer writes.
+// - MissedSeq: the requested sequence that was lost.
+// - NextSeq: the oldest sequence still available (resume point).
+type LaggedError struct {
+	MissedSeq uint64 // The sequence the subscriber attempted to read
+	NextSeq   uint64 // The oldest available sequence in the buffer (where subscriber resumes)
+}
+
+func (e LaggedError) Error() string {
```
rodrodros: receive e as a pointer: `e *LaggedError`

broadcast/broadcast.go
EgeCaner: Not really since we always create subscription from broadcast, I guess it should always be non-nil

broadcast/broadcast.go
EgeCaner: I'm not sure, there are multiple paths were Unsubscribe is called, internally or by consumer. If we wont gonna avoid multiple unsubscribe paths, idempotent behavior is better since in such cases error would force caller to branch or log noise for something that isn't actually exceptional. Example of internal unsubscribe paths are Unsubscribing to cleanup resources and terminate connection upon unexpected error, or broadcast closes then after draining the buffer, forwarder goroutine unsubscribes and exits since no more values will be sent and received. More like canceling same context twice does not yield in error or panic since subsequent calls are no-op.

broadcast/broadcast.go
EgeCaner: Would having
```go
broadcast[T = BigStruct].Send(val *T)
... same for ring buffer...
```
is optimal or different than following, beside giving the guarantee that we will only operate with pointers
```go
broadcast[T = *BigStruct].Send(val T)
... same for ring buffer...
```

broadcast/broadcast.go
```diff
@@ -0,0 +1,283 @@
...
+//
+// - Acquires subMu to add to the subs map.
+// - Reads ring.tail under ring's tail lock to set starting sequence.
+// - Spawns a goroutine to deliver messages to sub.out.
+func (b *Broadcast[T]) Subscribe() *Subscription[T] {
+	b.ring.tailMu.RLock()
+	next := b.ring.tail
+	b.ring.tailMu.RUnlock()
+
+	subID := b.nextSubID.Add(1)
```
infrmtcs: Why don't we also use `subMu` to guard this?

broadcast/ring_buffer.go
```diff
@@ -0,0 +1,154 @@
...
+type slot[T any] struct {
+	seq  uint64
+	mu   sync.RWMutex
+	data T
+}
+
+// ringBuffer supports MPMC access with per-slot locks plus a global tail lock.
+// - buffer: power-of-two-sized array of slots (masking index).
+// - tailMu: guards tail updates and consistent reads of tail.
+// - tail: next sequence to assign (1-based; newest = tail-1).
```
infrmtcs: Why don't we use 0-based instead? I think it would make calculating the index simpler, avoid unnessary minus 1

broadcast/ring_buffer.go
```diff
@@ -0,0 +1,154 @@
...
+		return val, nil
+	}
+	slot.mu.RUnlock()
+
+	// Requested seq is ahead of newest -> not ready yet.
+	return zero, ErrFutureSeq
+}
+
+// nextPowerOfTwo computes the next power-of-two >= x, returning 1 for x=0.
+func nextPowerOfTwo(x uint64) uint64 {
```
infrmtcs: We can replace this by https://pkg.go.dev/math/bits#Len64

broadcast/broadcast.go
```diff
@@ -0,0 +1,283 @@
...
+		return LaggedInfo{}, errors.New("no lag info present")
+	}
+	return e.lag, nil
+}
+
+// LaggedInfo is a transport-friendly struct to deliver lag information to subscribers.
+type LaggedInfo struct {
+	MissedSeq uint64
+	NextSeq   uint64
+}
```
infrmtcs: Should we replace this by `LaggedError`?

broadcast/broadcast.go
EgeCaner: broadcast.nextSubID is atomic.Uint64, we use it as atomic counter to reserve subIDs, atomic.Add, atomically increments the value and returns the incremented value, it guarantees that each call to add returns a different number so it should be safe to use without locks. We can turn `nextSubID` to uint64 and access under subMu, if it is preferred

infrmtcs: Yeah, I think if we have to use the mutex anyway, they reducing an atomic value help improve performance a bit.

broadcast/broadcast.go
```diff
@@ -0,0 +1,283 @@
...
+	done chan struct{}      // to shutdown goroutine
+}
+
+// EventOrLag is a tagged union:
+//   - If isLag is false, event holds a regular message.
+//   - If isLag is true, lag holds the lag notification.
+type EventOrLag[T any] struct {
+	event T          // For a regular message
+	lag   LaggedInfo // For a lag notification
+	isLag bool
```
infrmtcs: Instead of sum type, why don't we return both? Something like event + optional lagged info

broadcast/broadcast.go
```diff
@@ -0,0 +1,283 @@
...
+
+	out  chan EventOrLag[T] // user-facing channel for messages
+	done chan struct{}      // to shutdown goroutine
+}
+
+// EventOrLag is a tagged union:
+//   - If isLag is false, event holds a regular message.
+//   - If isLag is true, lag holds the lag notification.
+type EventOrLag[T any] struct {
+	event T          // For a regular message
```
infrmtcs: I think this is debatable, but by sharing ring buffer, we may also want to avoid copying large data. Maybe `*T` is preferrable here.

infrmtcs: Thinking more about this, the problem is the buffer can be overwritten, so sharing the pointer maybe unsafe.

broadcast/broadcast.go
EgeCaner: I agree, combining atomics with locks adds unnecessary complexity without clear benefits. We should choose either mutexes or a fully lock-free approach for the subscribers map. For simplicity, I plan to stick with mutexes. However, if you believe a lock-free strategy would better reduce lock contention during heavy subscribe/unsubscribe activity under a hot producer, we can considering it as well.

broadcast/broadcast.go
EgeCaner: Makes sense, having two types serving same purpose seems redundant

broadcast/ring_buffer.go
EgeCaner: 1 - based indexing is for empty case, to avoid read uninitialized data when sub.seq == 0 and uninitialized slot.seq == 0. I agree can be handled better, I guess simplest would be initializing slot.seq as uint64.MAX or (buffer_cap + 1) to indicate slot is uninitialized and guarantee reads will not read uninitialized state. This would avoid having subtraction cost on index calculation, considering uninitialized state wont live long it makes sense not reflect it cost to whole runtime

broadcast/broadcast.go
EgeCaner: Can you elaborate this bit more? Im not sure I get the idea

broadcast/ring_buffer.go
EgeCaner: Good suggestion, I didn't know about this. Even though current implementation is a well known pattern ([ref](https://graphics.stanford.edu/~seander/bithacks.html#RoundUpPowerOf2)). Relying on standard library is preferred to custom implementation I guess :) 

I benchmarked `nextPowerOfTwoWithLen64` impl below and compared with current the one, they are almost same same but so far nextPowerOfTwoWithLen64 was always faster even though difference is negligable.  I will update it to use Len64

```go
// nextPowerOfTwoWithLen64 calculates next power of two >= x using Len64.
func nextPowerOfTwoWithLen64(x uint64) uint64 {
	if x == 0 {
		return 1
	}
	return 1 << uint(bits.Len64(x-1))
}

func BenchmarkNextPowerOfTwo(b *testing.B) {
	var result uint64
	for i := range b.N {
		x := uint64(i % 100000)
		result = nextPowerOfTwo(x)
	}
	_ = result
}

func BenchmarkNextPowerOfTwoWithLen64(b *testing.B) {
	var res uint64
	for i := range b.N {
		x := uint64(i % 100000)
		res = nextPowerOfTwoWithLen64(x)
	}
	_ = res
}
```

```bash
BenchmarkNextPowerOfTwo-14              1000000000               0.2557 ns/op          0 B/op          0 allocs/op
BenchmarkNextPowerOfTwoWithLen64-14     1000000000               0.2492 ns/op          0 B/op          0 allocs/op
```

broadcast/broadcast.go
EgeCaner: I removed this branching

broadcast/ring_buffer.go
EgeCaner: Updated as suggested

broadcast/broadcast.go
EgeCaner: Updated as discussed, nextSubID is now accessed under subMu

broadcast/ring_buffer.go
EgeCaner: Switched to using math/bits.Len64 based nextPowerOfTwo

broadcast/broadcast.go
EgeCaner: Removed the `LaggedInfo` type

broadcast/broadcast.go
rodrodros: Ok, let's leave it as no op 👍

broadcast/broadcast.go
rodrodros: I think the first one is better because the user can have a large type that it is only used by reference in specific scenarios (no unnecessary value copies until the last part). 

In the second example if the user defines the type as a pointer, we will be passing a pointer to a pointer (`**T`) which is not that bad. 

I believe the first one offer a a bit more of flexibility degree

broadcast/broadcast.go
rodrodros, line 1: I think organization can be improved a bit here. It is organized as follows:
1. `Broadcast` type definition
2. `Subscription` type definition
3. `EventOrLag` type definition
4. `EventOrLag` functionality
5. `Broadcast` functionality
6. `Subscription` functionality

It's a bit non consistent or at least I don't find the ordering straight forward. I would suggest to either sort them like:

**types next to their definition (from general to specific)**

1. `Broadcast` type definition and functionality
2. `Subscription` type definition and functionality
3. `EventOrLag` type definition and functionality

**types next to their definition inverse (personally prefered)**

1. `EventOrLag` type definition and functionality
2. `Subscription` type definition and functionality
3. `Broadcast` type definition and functionality

I believe this last one makes the code easier to read so that when you arrive to the final type everything makes sense, but this is also very subjective.

I am also happy to use any other ordering different from these suggestion, just that it has some sense to it.

broadcast/broadcast.go
```diff
@@ -0,0 +1,277 @@
...
+type Subscription[T any] struct {
+	bcast    *Broadcast[T]
+	seq      uint64
+	notifyC  chan struct{}
+	isClosed atomic.Bool
+	id       uint64
+
+	out  chan EventOrLag[T] // user-facing channel for messages
+	done chan struct{}      // to shutdown goroutine
+}
```
rodrodros: If you look at Juno code overall, when having go routines many of them auxiliate themselves from the "context" object to cancel it's execution and clean themselves up. Nonetheless, here we are using a `done` channel to shutdown the go-routine. I don't think they are mutually exclusive either. The question is: "Should either/both of `Subscription` and `Broadcast` have a context object that guarantees that they clean themselves successfully?"

broadcast/broadcast.go
```diff
@@ -0,0 +1,277 @@
...
+	return e.event, nil
+}
+
+// Lag returns the lag info or an error if none.
+func (e EventOrLag[T]) Lag() (LaggedError, error) {
+	if !e.isLag {
+		return LaggedError{}, errors.New("no lag info present")
+	}
+	return e.lag, nil
+}
```
rodrodros: This method should receive the "self" parameter by pointer: `e *EventOrLag`). 

The difference between defining:
```golang
func (t T) m1 (...) {
    // method body
}
```

and

```golang
func (t *T) m2 (...) {
    // method body
}
```

is that the first one, `m1`, per each call is going to make a copy of `t`. If `t` is small, it doesn't matter but if `t` is a huge struct, then it will be copied each time.

In the case of the second one,`m2`, there will be also a copy but only of the pointer/reference which is always one word in size. 

In this specific example per call on `m1` to `EventOrLag[T]` we are copying: `sizeOf(T) + sizeOf(laggedError) + sizeOf(bool)` which is `sizeOf(T) + 2 + 1` which assuming the smallest `T` possible it is `1 + 2 + 1`, which is still bigger than just `1`. 

If unsure, benchmark yourself with a huge T and see the results (e.g. `T = [1000]uint64`)

broadcast/broadcast.go
```diff
@@ -0,0 +1,277 @@
...
+//   - bcast: back-reference to the Broadcast.
+//   - seq: next sequence to read.
+//   - notifyC: 1-buffered wakeup channel (coalesced notifications).
+//   - isClosed: ensures Unsubscribe runs only once.
+//   - id: subscription id
+//   - out: user-facing channel for events or lag notifications.
+//   - done: closed by Unsubscribe; run observes it to abort blocking sends
+type Subscription[T any] struct {
+	bcast    *Broadcast[T]
+	seq      uint64
```
rodrodros: Nit: rename `seq` to `nextSeq` (if applied remember to update the description on top as well)

broadcast/broadcast.go
```diff
@@ -0,0 +1,277 @@
...
+//   - single-run via CAS.
+//   - Deletes from subs under subMu.Lock, then closes done (to stop run).
+func (sub *Subscription[T]) Unsubscribe() {
+	if !sub.isClosed.CompareAndSwap(false, true) {
+		return
+	}
+
+	sub.bcast.subMu.Lock()
+	delete(sub.bcast.subs, sub.id)
+	sub.bcast.subMu.Unlock()
```
rodrodros: I think this three lines should be a `Broadcast` method. Rationale is that we want to abstract away what happens in `Broadcast`.

broadcast/ring_buffer.go
```diff
@@ -0,0 +1,147 @@
...
+	"fmt"
+	"math/bits"
+	"sync"
+)
+
+// Returned when the requested sequence is ahead of the latest published (not yet available).
+var (
+	ErrFutureSeq       = errors.New("requested seq. not published yet")
+	ErrInvalidSequence = errors.New("invalid sequence: 0")
+)
```
rodrodros: I would move this to an error.go file since they are used in all the package

broadcast/ring_buffer.go
```diff
@@ -0,0 +1,147 @@
...
+// - MissedSeq: the requested sequence that was lost.
+// - NextSeq: the oldest sequence still available (resume point).
+type LaggedError struct {
+	MissedSeq uint64 // The sequence the subscriber attempted to read
+	NextSeq   uint64 // The oldest available sequence in the buffer (where subscriber resumes)
+}
+
+func (e *LaggedError) Error() string {
+	return fmt.Sprintf("subscriber lagged: missed seq=%d, next available seq=%d", e.MissedSeq, e.NextSeq)
+}
```
rodrodros: I believe this could fit in `error.go` as well

rodrigo-pino (commented):
Second review

broadcast/broadcast.go
EgeCaner: I completely agree with the proposed reorganization and I would also prefer the inverse order to build knowledge progressively in a natural read order

broadcast/broadcast.go
EgeCaner: Yes context might improve the system. Let me explore how it goes with drain-on-close semantics(when broadcast closed before subscriber) and how it will change it API.

broadcast/broadcast_test.go
```diff
@@ -0,0 +1,644 @@
...
+				}
+
+				wg.Wait()
+				b.StopTimer()
+			})
+		}
+	}
+}
+
+func BenchmarkSubsribe(b *testing.B) {
```
rodrodros: Let's have another file `broadcast_benchmark_test.go`. Rationale is that they have their own space and it is inherently obvious by scanning the code that this piece of software includes benchmarks. Let me know what you think!

broadcast/errors.go
```diff
@@ -0,0 +1,29 @@
...
+import (
+	"errors"
+	"fmt"
+)
+
+var (
+	// Returned when the broadcast is closed and no more sends are possible.
+	ErrClosed = errors.New("broadcast channel closed")
+	// Returned when the requested sequence is ahead of the latest published (not yet available).
+	ErrFutureSeq = errors.New("requested seq. not published yet")
```
rodrodros: I think it is a good idea to use full words in error messages. Let's change "seq." to "sequence" 🙏

broadcast/broadcast.go
```diff
@@ -0,0 +1,272 @@
...
+// Readers are synchronised with writers to avoid tearing data/seq pairs
+func (s *slot[T]) read() (T, uint64) {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+	return s.data, s.seq
+}
+
+// write updates the slot's data and sequence under an exclusive write lock.
+// Writers are serialised per-slot; they are also globally serialized by Broadcast.Send.
+func (s *slot[T]) write(data T, seq uint64) {
```
rodrodros: Write should receive by reference as we discussed in a previous comment

broadcast/broadcast.go
```diff
@@ -0,0 +1,272 @@
...
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	s.data = data
```
rodrodros: It's ok to perform a copy here. The difference is that we go from copying the value of `T` more than once to just once

broadcast/broadcast.go
```diff
@@ -0,0 +1,272 @@
...
+}
+
+// Send publishes msg to the ring and wakes subscribers.
+// Behaviour:
+// - Returns ErrClosed if the Broadcast context is done at entry.
+// - Serialises all producers via b.mu; Send is effectively single-writer.
+// - Writes the message and its sequence into the computed slot.
+// - Increments tail to point to the next sequence to be written.
+// - Wakes all waiters by closing the current notify channel and swapping in a new one.
+func (b *Broadcast[T]) Send(msg T) error {
```
rodrodros: Send should receive by pointer.  Explain in the function description that we take by pointer but we promise immutability. It's just for performance reason

broadcast/broadcast.go
```diff
@@ -0,0 +1,272 @@
...
+		idx := tail & b.mask
+
+		b.buffer[idx].write(msg, tail)
+
+		b.tail.Add(1)
+
+		// Wake all waiters by closing the current notify channel and swapping a new one.
+		notify := b.loadNotify()
+		close(notify)
+		b.notifyChan.Store(make(chan struct{}))
```
rodrodros: Currently running benchmark but creating a channel on send might be too expensive memory wise 🤔. Let me see if I can confirm it and if there is a better way of communication a new send

broadcast/broadcast.go
EgeCaner: Converted to receive *T

broadcast/broadcast_test.go
EgeCaner: Moved to its own file

broadcast/errors.go
EgeCaner: adjusted the message

broadcast/broadcast.go
EgeCaner: No longer relevant, code has changed

broadcast/broadcast.go
EgeCaner: Adjusted as suggested

broadcast/broadcast.go
EgeCaner: Switched to using sync.Cond

broadcast/broadcast.go
EgeCaner: With sync.Cond we need to wake the subscriber if it is waiting. We cannot multiplex sync.Cond thus we need to wake goroutines on Close and Unsubscribe, for this reason we cannot cancel from outside via context. Closing and Unsubscribe is done via their explicit methods

broadcast/broadcast.go
```diff
@@ -0,0 +1,301 @@
...
+		// Already closed
+	default:
+		close(sub.done)
+		// wake if subscriber is waiting
+		bcast := sub.bcast
+		idx := sub.nextSeq.Load() & bcast.mask
+		slot := &bcast.buffer[idx]
+		slot.cond.L.Lock()
+		slot.cond.Broadcast()
+		slot.cond.L.Unlock()
```
infrmtcs, line 163: I think the docs don't require to hold L before Broadcast.

```suggestion
		slot.cond.Broadcast()
```

broadcast/broadcast.go
```diff
@@ -0,0 +1,301 @@
...
+	return sub.out
+}
+
+// Unsubscribe terminates the subscribers delivery goroutine.
+func (sub *Subscription[T]) Unsubscribe() {
+	select {
+	case <-sub.done:
+		// Already closed
+	default:
+		close(sub.done)
```
infrmtcs: I assume that we need to select `<-sub.done` because `Unsubscribe` needs to be idempotent. This can cause race condition in case where 2 goroutines go to default case at the same time.

broadcast/broadcast.go
infrmtcs: OK, I now understand that we want to acquire the underlying lock, not related to this broadcast.

broadcast/broadcast.go
```diff
@@ -0,0 +1,301 @@
...
+// New constructs a Broadcast with a ring of at least the given capacity.
+// - The actual capacity is rounded up to the next power of two.
+// - Each slot’s condition variable is initialised to use the slot’s mutex.
+// - The initial tail is 0 (no messages published).
+func New[T any](capacity uint64) *Broadcast[T] {
+	capacity = nextPowerOfTwo(capacity)
+	rb := make([]slot[T], capacity)
+	for i := range rb {
+		slot := &rb[i]
+		slot.cond = *sync.NewCond(&slot.mu)
```
infrmtcs, line 219: We're not using `mu` directly anywhere else, so probably we can remove `mu` from `slot` and also remove this initialization loop.

broadcast/broadcast.go
EgeCaner: Good catch! I will switch to sync.Once or CAS on closed flag

broadcast/broadcast.go
EgeCaner: As I know sync.Cond uses Locker we provide in construction, so slot.mu should be used internally when we do slot.cond.L.{Lock/Unlock}.

infrmtcs: Yeah, what I meant is without the loop, the zero value of `sync.Cond` and `sync.Mutex` is valid, so we can use `slot.cond.L` directly without this initialization and remove `slot.mu` to avoid confusion.

feed/feed_benchmark_test.go
infrmtcs, line 1: Can we reuse the new broadcast test? Because it seems that both old `Feed` and new `Broadcast` follow the same interface. If we can reuse the same test, we don't have to worry about the case where updating one test but not the other.

broadcast/broadcast.go
```diff
@@ -0,0 +1,307 @@
...
+//
+// Lifecycle:
+//   - Created by Broadcast.Subscribe, which spawns a delivery goroutine (run).
+//   - Call Unsubscribe to stop; it requests the goroutine to exit promptly and
+//     wakes it if blocked, then the goroutine closes out upon return.
+//   - If Broadcast.Close is called, subscriptions will only exit once
+//     they can observe the closed context in their run loop, after draining the channel.
+type Subscription[T any] struct {
+	bcast     *Broadcast[T]
+	nextSeq   atomic.Uint64
```
infrmtcs, line 70: Can revert this back to `uint64`? I think it's only used in the `run` loop, except in cleanup code. If we do the cleanup properly, then I think we can void having another atomic value here.

broadcast/broadcast.go
```diff
@@ -0,0 +1,307 @@
...
+	out       chan EventOrLag[T] // user-facing channel for messages
+	unsubOnce sync.Once
```
infrmtcs, line 72: nit: We can use `sync.OnceFunc` instead.
```suggestion
	unsub func()
```

broadcast/broadcast.go
infrmtcs: I forgot that `L` is an interface, not a struct, so it doesn't have valid zero value.
We still need this loop, although I still think we can remove `slot.mu`

PR #3071, opened 2025-08-19
feat: add reorg notification to receipt sub and tx sub

EgeCaner:


rpc/v9/subscriptions.go
```diff
@@ -968,8 +971,8 @@ func (h *Handler) SubscribeNewTransactions(
 	lastBlockNumber := uint64(0)
 
 	subscriber := subscriber{
-		onStart: func(ctx context.Context, id string, _ *subscription, _ any) error {
```
rodrodros, line 971: Why can we dismiss this one?

rodrigo-pino (approved):
Looks good so approving.

So far this for transactions subscriptions, should this apply to event subscriptions as well?

rpc/v9/subscriptions.go
EgeCaner: `onStart` was a no-op, so removing its callback makes the code clearer without changing behavior. Since receipt subscriptions stream only live events, on incoming heads, received txs and preconfirmed blocks—not historical ones—`onStart` isn’t needed.

EgeCaner:
> So far this for transactions subscriptions, should this apply to event subscriptions as well?

Event subscription receives SubscriptionBlockID as starting point of event stream, and most recent block consumer can start receiving is `latest`. For event subscription `onStart` from `SubscriptionBlockID` to most recent (either head or pre_confirmed based on user input (finality_status)) is streamed then live events

PR #3072, opened 2025-08-19
fix: support 'ACCEPTED_ON_L1' status on 'starknet_subscribeEvents'

EgeCaner:
`starknet_SubscribeEvents` does not allow `ACCEPTED_ON_L1` as finality parameter. However upon querying historical block by block number as starting point of stream, some events might be `ACCEPTED_ON_L1` and this case was unhandled resulting in returning `ACCEPTED_ON_L2` instead.

### DIFF
--- a/rpc/v9/subscriptions.go
+++ b/rpc/v9/subscriptions.go
@@ -423,7 +423,7 @@ func sendEvents(
 				finalityStatus = TxnPreConfirmed
 			case *event.BlockNumber <= l1Head:
 				finalityStatus = TxnAcceptedOnL1
-			default: // Canonical block not finalized on L1
+			default: // Canonical block not finalised on L1
 				finalityStatus = TxnAcceptedOnL2
 			}
 
--- a/rpc/v9/subscriptions.go
+++ b/rpc/v9/subscriptions.go
@@ -180,7 +180,7 @@ func (h *Handler) subscribe(
 
 type SubscriptionEmittedEvent struct {
 	rpcv6.EmittedEvent
-	FinalityStatus TxnFinalityStatusWithoutL1 `json:"finality_status"`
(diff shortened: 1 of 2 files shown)


rpc/v9/subscriptions.go
```diff
@@ -307,16 +299,16 @@ func (h *Handler) SubscribeEvents(
 }
 
 // processEvents queries database for events and stream filtered events.
-func (h *Handler) processEvents(
+func (h *Handler) processHistoricalEvents(
```
rodrodros: Why the name change?
Nit: update the function description as well

rpc/v9/subscriptions.go
```diff
@@ -409,14 +401,32 @@ func processBlockEvents(
...
-func sendEvents(ctx context.Context, w jsonrpc.Conn, events []*blockchain.FilteredEvent,
-	eventsPreviouslySent map[SentEvent]TxnFinalityStatusWithoutL1, id string, finalityStatus TxnFinalityStatusWithoutL1,
+func sendEvents(
+	ctx context.Context,
+	w jsonrpc.Conn,
+	events []*blockchain.FilteredEvent,
+	eventsPreviouslySent map[SentEvent]TxnFinalityStatus,
+	id string,
+	height uint64,
+	l1Head uint64,
```
rodrodros, line 411: 👍

rpc/v9/subscriptions.go
EgeCaner: IMO `processHistoricalEvents` clearly indicates handling of historical data, distinguishing it from processing live events. `processEvents` sounds more generic

EgeCaner: Corrected the function description

infrmtcs (approved):

PR #3085, opened 2025-08-27
refactor: Reading consensus WAL return iterator to avoid allocation

infrmtcs:


rodrigo-pino (approved):
It looks good. 

I am not sure exactly how more performant it will end up being though.

PR #3088, opened 2025-08-28
feat: pre_latest

EgeCaner:


sync/data_source.go
```diff
@@ -140,12 +140,5 @@ func (f *feederGatewayDataSource) PreConfirmedBlockByNumber(ctx context.Context,
 		return core.PreConfirmed{}, err
 	}
 
-	h, err := f.blockchain.HeadsHeader()
-	if err != nil && !errors.Is(err, db.ErrKeyNotFound) {
-		return core.PreConfirmed{}, err
-	} else if err == nil {
-		preConfirmed.StateUpdate.OldRoot = h.GlobalStateRoot
-	}
-
```
rodrodros, line 149: Why is this part deleted?

sync/pending.go
```diff
@@ -58,6 +58,70 @@ func (p *Pending) Variant() core.PendingDataVariant {
 	return core.PendingBlockVariant
 }
 
+func (p *Pending) GetPreLatest() *Pending {
+	return nil
+}
+
+type PendingDataWPreLatest interface {
```
rodrodros: Not sure why not just have the `PreLatestData` interface?

```golang
type PreLatestData interface {
   GetPreLatest() *Pending
}
```

Then you can have the union type (if justified), but I am not sure so far in the code

sync/data_source.go
EgeCaner: Previously we are fetching PreConfirmed with number = latest +1, so latest.GlobalStateRoot was preConfirmed.OldRoot. But with preLatest, we are either constructing pre_confirmed top of latest or preLatest (latest+1) and prelatest does not have commitments such as GlobalStateRoot

sync/sync.go
```diff
@@ -134,7 +135,7 @@ type Synchronizer struct {
 	log      utils.SimpleLogger
 	listener EventListener
 
-	pendingData              atomic.Pointer[core.PendingData]
+	pendingData              atomic.Pointer[PendingDataWPreLatest]
```
rodrodros: In here, 
```golang
pendingData atomic.Pointer[PendingData | PreLatestData]
```

A bit longer, but type-wise makes more sense to me. Opinionated, so feel free to ignore.

sync/sync.go
```diff
@@ -451,7 +452,12 @@ func (s *Synchronizer) pollPendingData(ctx context.Context, sem chan struct{}) {
 		return
 	}
 	s.pollPending(ctx, sem)
-	s.pollPreConfirmed(ctx, sem)
+	select {
+	case <-ctx.Done():
+		return
+	default:
+		s.pollPreConfirmed(ctx, sem)
+	}
```
rodrodros: I am unsure if this `select` here is the right choice. What's the point of it?

sync/sync.go
```diff
@@ -495,33 +501,144 @@ func (s *Synchronizer) pollPending(ctx context.Context, sem chan struct{}) {
...
+
+	newHeadSub := s.newHeads.SubscribeKeepLast()
+	defer newHeadSub.Unsubscribe()
+
+	seenBlockHashes := make(map[felt.Felt]*Pending)
+
+	var subCtx context.Context
+	var cancel context.CancelFunc = func() {}
+	sem := make(chan struct{}, 1)
```
rodrodros: Please, a better name than `sem`

sync/sync.go
```diff
@@ -602,28 +719,50 @@ func (s *Synchronizer) fetchAndStorePending(ctx context.Context) error {
 	return s.StorePending(&pending)
 }
 
-func (s *Synchronizer) fetchAndStorePreConfirmed(ctx context.Context) error {
-	highestBlockHeader := s.highestBlockHeader.Load()
-	if highestBlockHeader == nil {
-		return nil
-	}
+func (s *Synchronizer) fetchAndStorePreConfirmed(ctx context.Context, sem <-chan struct{}, preLatest *Pending) {
```
rodrodros: long. line

sync/sync.go
EgeCaner: When starknet version <= 0.14.0, if `pollPending` is cancelled from parent context, before exiting the the `pollPendingData` we were unnecessarily triggering `s.pollPreConfirmed`, which eventually exit since it starts with cancelled context. I added this to see sync_tests ever start polling pre_confirmed in coverage, entering here on exit makes the coverage bit noisy. I guess it is also fine to keep it as before.

sync/sync.go
```diff
@@ -495,33 +501,144 @@ func (s *Synchronizer) pollPending(ctx context.Context, sem chan struct{}) {
 	}
 }
 
+func (s *Synchronizer) pollPreLatest(ctx context.Context, preLatestChan chan *Pending) {
+	if s.preConfirmedPollInterval == time.Duration(0) {
+		s.log.Infow("Pre-latest block polling is disabled")
+		return
```
rodrodros: Shouldn't pre-latest use the pending block interval?

EgeCaner: default pending block interval is 5 second and configured to be used when network is not at 0.14.0. With faster block times, we should poll pre-latest faster than block time. I used pre-confirmed interval for this reason, not sure the best but maybe we can introduce another parameter? So far my observation on sepolia, I was using 500 ms as pre-confirmed interval, I frequently get "no pending block" due to early polling. I guess we can use 1-2s for prelatest and 0.5 sec for pre_confirmed

sync/sync.go
rodrodros: I see, I agree that perhaps a better handling, this so far seems a bit obscure to me. 
I think we can leave it as is but let's have some comments there plase

sync/sync.go
rodrodros: I think we can keep it for now and see what happens after 👍

sync/sync.go
rodrodros: Just realized is in other places, older than this PR, maybe there is a convention I am missing?

EgeCaner: Yeap, as far as I understood channels are used as semaphore and ensures at most one goroutine actively fetches, also they are used in cancellation of polling goroutines. IMO for pollPending and pollLatest we can remove them and use waitGroups

sync/sync.go
EgeCaner: I introduced `PreLatestData` interface, but this didnt worked
```go
pendingData atomic.Pointer[PendingData | PreLatestData]
```
 I had to introduced Union interface `PendingDataWithPreLatest`

rodrodros: Ah I see 👍. Thanks

sync/pending.go
```diff
@@ -58,6 +67,33 @@ func (p *Pending) Variant() core.PendingDataVariant {
...
 }
 
+func (p *Pending) GetPreLatest() *Pending {
+	return nil
+}
+
+type PreConfirmedWithPreLatest struct {
+	*core.PreConfirmed
+	PreLatest *Pending
+}
```
infrmtcs: I think this is quite confusing. For example, this means `preConfirmedWithPreLatest.PreLatest.GetPreLatest() == nil`

sync/data_source.go
infrmtcs: Would you mind also revisit the logic in `consensusDataSource`?

sync/sync.go
```diff
@@ -880,7 +1076,28 @@ func (s *Synchronizer) storeEmptyPreConfirmed(latestHeader *core.Header) error {
 		CandidateTxs:          make([]core.Transaction, 0),
 	}
 
-	s.pendingData.Store(utils.HeapPtr[core.PendingData](&preConfirmed))
+	if existingPending, err := s.PendingData(); err == nil {
+		existingPendingB := existingPending.GetBlock()
+		if existingPendingB.Number == preConfirmed.GetBlock().Number {
+			// do not overwrite if pre_confirmed is present for this head
```
infrmtcs: What happens if `s.pendingData` is written between read and write?

sync/sync.go
```diff
@@ -495,33 +504,144 @@ func (s *Synchronizer) pollPending(ctx context.Context, sem chan struct{}) {
...
+		StateDiff: &emptyStateDiff,
+	}
+	pendingFromLatest := NewPending(head, &stateUpdate, make(map[felt.Felt]core.Class))
+	select {
+	case <-ctx.Done():
+		return
+	case preLatestChan <- &pendingFromLatest:
+	}
+
+	preLatestPollTicker := time.NewTicker(s.preConfirmedPollInterval)
```
infrmtcs: This looks strange to me

sync/sync.go
```diff
@@ -495,33 +504,144 @@ func (s *Synchronizer) pollPending(ctx context.Context, sem chan struct{}) {
 	}
 }
 
+func (s *Synchronizer) pollPreLatest(ctx context.Context, preLatestChan chan *Pending) {
+	if s.preConfirmedPollInterval == time.Duration(0) {
```
infrmtcs: Why do we use preconfirmed poll interval for prelatest?

infrmtcs: Saw the other comment

sync/sync.go
infrmtcs: Saw the other comment

sync/sync.go
infrmtcs: I think we need to utilize `CompareAndSwap` if we don't want to use channels here.

sync/sync.go
```diff
@@ -495,33 +504,144 @@ func (s *Synchronizer) pollPending(ctx context.Context, sem chan struct{}) {
...
 
-	preConfirmedPollTicker := time.NewTicker(s.preConfirmedPollInterval)
+	preLatestChan := make(chan *Pending)
+	defer close(preLatestChan)
+	var wg stdsync.WaitGroup
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		s.pollPreLatest(ctx, preLatestChan)
+	}()
```
infrmtcs: We can use `wg.Go` instead.

sync/sync.go
```diff
@@ -495,33 +504,144 @@ func (s *Synchronizer) pollPending(ctx context.Context, sem chan struct{}) {
...
 func (s *Synchronizer) pollPreConfirmed(ctx context.Context, sem chan struct{}) {
 	if s.preConfirmedPollInterval == time.Duration(0) {
 		s.log.Infow("Pre-confirmed block polling is disabled")
 		return
 	}
 
-	preConfirmedPollTicker := time.NewTicker(s.preConfirmedPollInterval)
+	preLatestChan := make(chan *Pending)
+	defer close(preLatestChan)
+	var wg stdsync.WaitGroup
```
infrmtcs:
```suggestion
	var wg stdsync.WaitGroup
	defer wg.Wait()
```

sync/sync.go
```diff
@@ -602,28 +722,54 @@ func (s *Synchronizer) fetchAndStorePending(ctx context.Context) error {
...
-	preConfirmedBlock, err := s.dataSource.PreConfirmedBlockByNumber(ctx, highestBlockHeader.Number+1)
-	if err != nil {
-		return err
+		select {
+		case <-ctx.Done():
+			preConfirmedPollTicker.Stop()
+			return
+		case <-preConfirmedPollTicker.C:
+			continue
+		}
```
infrmtcs: I think this can be simpler:

```suggestion
		select {
		case <-ctx.Done():
			preConfirmedPollTicker.Stop()
			return
		case preLatest := <-preLatestChan:
			// Update new prelatest
		case <-preConfirmedPollTicker.C:
		}
		// Store logic
```

sync/pending.go
EgeCaner: Introduced `PreLatest` type.

```go
type PreLatest Pending
```

sync/data_source.go
EgeCaner: @infrmtcs I couldnt found similar logic there regarding the old root. If you knew the certain place, can you point it?

sync/pending.go
```diff
@@ -58,6 +58,44 @@ func (p *Pending) Variant() core.PendingDataVariant {
...
+
+func (p *Pending) GetPreLatest() *PreLatest {
+	return nil
+}
+
+type PreConfirmedWithPreLatest struct {
+	*core.PreConfirmed
+	PreLatest *PreLatest
+}
```
rodrodros: I was thinking in the objective of this type, does it makes sense to have both a `PreConfirmed` and a `PreConfirmedWithPrelatest`?

Would it makes sense to just have
```golang
type PreConfirmed struct {
	Block       *Block
	StateUpdate *StateUpdate
    PreLatest   *PreLatest
	// Node does not fetch unknown classes. but we keep it for sequencer
	NewClasses            map[felt.Felt]Class
	TransactionStateDiffs []*StateDiff
	CandidateTxs          []Transaction
}
```
`NewPreconfirmed` is only being used inside `NewPreconfirmedWithLatest` (outside of tests), which makes me more prone to this change. 

Also, why is `PreConfirmed` part of core but `PreConfirmedWithPreLatest` part of the sync package?


----

On another topic, I I think the last three fields of `sync. PreConfirmed` can be ditched, and we can have another type specifically defined in the block builder (but for another time)
```golang
// this time defined in the sequence package
type PreConfirmed struct {
    core.PreConfirmed
	NewClasses            map[felt.Felt]Class
	TransactionStateDiffs []*StateDiff
	CandidateTxs          []Transactio
}
```
It doesn't have to be defined in that way; the core idea is to have type separations. I don't feel comfortable with structs that have optional fields that make sense to be filled or not, depending on what they are being used for.

sync/pending.go
```diff
@@ -58,6 +58,44 @@ func (p *Pending) Variant() core.PendingDataVariant {
...
+type PreLatest Pending
+
+type PreLatestData interface {
+	GetPreLatest() *PreLatest
+}
+
+type PendingDataWithPreLatest interface {
+	core.PendingData
+	PreLatestData
+}
```
rodrodros: It also makes me question if we should also have separate interface 🤔

sync/sync.go
```diff
@@ -444,117 +447,102 @@ func (s *Synchronizer) revertHead(forkBlock *core.Block) {
...
-		case <-ticker.C:
-			poll()
+		case <-pendingPollTicker.C:
+			wg.Wait()
+			wg.Go(func() {
+				err := s.fetchAndStorePending(ctx)
+				if err != nil {
+					if errors.Is(err, ErrMustSwitchPollingPreConfirmed) {
+						s.log.Infow(
+							"Detected block version 0.14.0; switching to polling mode for pre_confirmed blocks",
```
rodrodros: Suggestion: "Detected block version > 0.14.0; polling for pre_confirmed blocks"

cmd/juno/juno.go
```diff
@@ -108,8 +108,8 @@ const (
 	defaultPprof                              = false
 	defaultPprofPort                          = 6062
 	defaultColour                             = true
-	defaultPendingPollInterval                = 5 * time.Second
-	defaultPreConfirmedPollInterval           = time.Second
+	defaultPendingPollInterval                = time.Second
+	defaultPreConfirmedPollInterval           = 500 * time.Millisecond
```
rodrodros, line 112: I believe we might have to update our docs here as well:
juno.nethermind.io/configuring#configuration-options

sync/pending.go
EgeCaner: We can have single `PreConfirmed` type which includes `PreLatest`. I introduce the `PreConfirmedWithPrelatest` because `PreLatest` is `sync.Pending` and importing `sync.Pending` from core leds to cyclic dependency, but I agree two types are not necessary at all, I just need to move `sync.Pending` to core. If we will utilize `PreLatest` in RPC, we can also add `GetPreLatest` in to `PendingData` and keep having single interface. 

Regarding this
```go
// this time defined in the sequence package
type PreConfirmed struct {
    core.PreConfirmed
	NewClasses            map[felt.Felt]Class
	TransactionStateDiffs []*StateDiff
	CandidateTxs          []Transactio
}
```

Looks like those 3 fields are redundant since core.Preconfirmed already has those.

sync/pending.go
EgeCaner: I will move sync.Pending into core and include `GetPreLatest` into `PendingData`and keep having single interface, this will also allow us to use PreLatest data in RPC incase we need

docs/docs/_config-options.md
rodrodros, line 1: It is not optimal but in the meantime, you have to update also "./docs/versioned_docs/version-0.15.0"

This initial structure doesn't make sense since the latest docs are always the same. Also an improvement for the near future

core/pending.go
```diff
@@ -69,6 +147,35 @@ func (p *PreConfirmed) GetTransactionStateDiffs() []*StateDiff {
...
 
+func (p *PreConfirmed) GetPreLatest() *PreLatest {
+	return p.PreLatest
+}
+
 func (p *PreConfirmed) Variant() PendingDataVariant {
 	return PreConfirmedBlockVariant
 }
+
+func (p *PreConfirmed) Validate(parent *Header) bool {
```
rodrodros, line 163: Thinking in abstract here, but shouldn't `PreConfirmed` know who its parent is on creation? How can we create an invalid `PreConfirmed ` or how it's validity could change that we have to call for validation.

sync/sync.go
```diff
@@ -444,117 +447,101 @@ func (s *Synchronizer) revertHead(forkBlock *core.Block) {
...
-		return
-	}
+func (s *Synchronizer) SubscribePendingData() PendingDataSubscription {
+	return PendingDataSubscription{s.pendingDataFeed.Subscribe()}
+}
 
-	preConfirmedPollTicker := time.NewTicker(s.preConfirmedPollInterval)
+func (s *Synchronizer) pollLatest(ctx context.Context) {
+	ticker := time.NewTicker(time.Minute)
+	fetchWq := &stdsync.WaitGroup{}
```
rodrodros: nitpick: intentionally `fetchWq` or should it be `fetchWg`?

core/pending.go
EgeCaner: This function ensures that when storing or reading a pre_confirmed or pending block, the current pending data is properly validated as a valid successor of the latest block in the database. Although a similar check existed before, it has been refactored into this method to encapsulate the validation logic within the pending data type itself.

This validation is necessary both at the time of storing and reading because the pending data can become invalid in certain scenarios—for example, if a new head is committed to the database before pending data is stored, the pending data may no longer be the correct successor, or when accessing pending data, a new block might have been committed but the pending data not yet updated.

sync/sync.go
EgeCaner: should be `fetchWg`

core/pending.go
rodrodros, line 1: Can we add for every type that implements the `PendingData` interface the initial runtime check:

```golang
var _ MyInterface = (*MyType)(nil)
```

sync/sync.go
```diff
@@ -602,55 +589,233 @@ func (s *Synchronizer) fetchAndStorePending(ctx context.Context) error {
...
+	defer newHeadSub.Unsubscribe()
 
-func (s *Synchronizer) SubscribeReorg() ReorgSubscription {
-	return ReorgSubscription{s.reorgFeed.Subscribe()}
+	preLatestChan := make(chan *core.PreLatest)
+	defer close(preLatestChan)
+	var preLatestWg stdsync.WaitGroup
+	preLatestWg.Go(func() { s.pollPreLatest(ctx, preLatestChan) })
+
+	mostRecentPredecessor := int64(-1)
```
rodrodros: Do we need it to be -1 for when we are syncing from block 0? Is this something that can happen in a real usage of Juno (independent of network).

sync/sync.go
```diff
@@ -602,55 +589,233 @@ func (s *Synchronizer) fetchAndStorePending(ctx context.Context) error {
...
+}
 
-	// not at the tip of the chain yet, no need to poll preconfirmed
-	if highestBlockHeader.Number > head.Number {
-		return nil
+// fetchPreLatest fetches pre-latest block for given head and forwards the value to preLatestChan.
+//
+// Consumer of preLatestChan will receive each preLatest once. It is possible preLatest block will skipped,
+// incase latest for the pre-latest is already published.
+// Incase fetched pre-latest belongs to future block, it will be cached and served later.
```
rodrigo-pino, line 634: I think it should be "in case" instead of "incase". Minor nitpick

rodrodros:
```suggestion
// fetchPreLatest fetches the prelatest block for a given head and forwards the value to preLatestChan.
// Consumers of preLatestChan will receive each pre-latest once. It is possible for a pre-latest block to 
// be skipped if:
//  - the pre-latest has already become the latest block of the chain.
//  - the pre-latest is part of the future of the chain. In this case, it will be cached and served later.
```

sync/sync.go
EgeCaner: Yes this is for when we are syncing for pre_confirmed for head 0. Otherwise this check will skip due to head == 0, mostRecentPredecessor == 0. Not sure how beneficial it is in real usage, it is only useful on fresh chain.

```go
			if int64(head.Number) <= mostRecentPredecessor {
				// already got more recent data for this head
				continue
			}
```

EgeCaner: Switched to tracking nextPreConfirmedNumber instead which starts from 1 to eliminate type casting and -1.

sync/sync.go
```diff
@@ -627,55 +614,233 @@ func (s *Synchronizer) fetchAndStorePending(ctx context.Context) error {
...
+		preConfirmed, err := s.dataSource.PreConfirmedBlockByNumber(ctx, preConfirmedNumber)
+		if err == nil {
+			preConfirmed.WithPreLatest(preLatest)
+			isChanged, err := s.StorePreConfirmed(&preConfirmed)
+			if err == nil && isChanged {
+				s.log.Tracew(
+					"pre_confirmed block stored",
+					"number", preConfirmed.Block.Number,
+					"txCount", preConfirmed.Block.TransactionCount,
+				)
```
infrmtcs, line 882: This will print a lot

sync/sync.go
```diff
@@ -627,55 +614,233 @@ func (s *Synchronizer) fetchAndStorePending(ctx context.Context) error {
...
-		return nil
+// fetchPreLatest fetches the prelatest block for a given head and forwards the value to preLatestChan.
+// Consumers of preLatestChan will receive each pre-latest once. It is possible for a pre-latest block to
+// be skipped if:
+//   - the pre-latest has already become the latest block of the chain.
+//   - the pre-latest is part of the future of the chain. In this case, it will be cached and served later.
+func (s *Synchronizer) fetchPreLatest(
+	ctx context.Context,
+	head *core.Block,
+	seenBlockHashes map[felt.Felt]*core.PreLatest,
```
infrmtcs, line 723: Personally, I think that the name implies that the key is the hash of the block, but in fact it's parent block hash.

core/state_update.go
```diff
@@ -45,7 +45,8 @@ func (d *StateDiff) Merge(incoming *StateDiff) {
 			if oldAddrStorage, exists := oldMap[addr]; exists {
 				maps.Copy(oldAddrStorage, newAddrStorage)
 			} else {
-				oldMap[addr] = newAddrStorage
+				oldMap[addr] = make(map[felt.Felt]*felt.Felt, len(newAddrStorage))
+				maps.Copy(oldMap[addr], newAddrStorage)
```
rodrodros: I am curious why creating and copying vs just copying? Was there a bug? Why was it undetected before?

sync/sync.go
```diff
@@ -627,55 +614,233 @@ func (s *Synchronizer) fetchAndStorePending(ctx context.Context) error {
 	return s.StorePending(&pending)
 }
 
-func (s *Synchronizer) fetchAndStorePreConfirmed(ctx context.Context) error {
-	highestBlockHeader := s.highestBlockHeader.Load()
-	if highestBlockHeader == nil {
-		return nil
+// pollPreLatest fetches preLatest block for each new head and forwards the prelatest to given chanel.
+//
+// Consumer of preLatestChan never receives the same preLatest twice.
```
rodrodros, line 682: On the same idea as the previous comment:
```suggestion
// pollPreLatest fetches the pre-latest block for each new head and forwards it to a given 
// channel (`preLatesChan`).
// Consumers of `preLatestChan` will never receive the same pre-latest block twice. 
```

sync/sync.go
```diff
@@ -627,55 +614,233 @@ func (s *Synchronizer) fetchAndStorePending(ctx context.Context) error {
...
-	if err != nil {
-		return err
+	newHeadSub := s.newHeads.SubscribeKeepLast()
+	defer newHeadSub.Unsubscribe()
+	// When fetching pre-latest it is possible that pre-latest we fetched
+	// belongs to l2 finalised block node not fetched yet.
+	// In such scenario we cache the pre-latest for the future block.
+	// Upon receiving the corresponding head for the cached pre-latest,
+	// it will be removed from cache. However in case of reorg this key might not get evicted,
+	// and incase of reorg, this goroutine must be restarted.
```
rodrodros, line 697: Same idea some grammar fix, feel free to ignore

```suggestion
// When fetching the pre-latest block, it is possible that the pre-latest we fetched
// belongs to an L2 finalized block we haven't fetched yet.
// In such a scenario, we cache the pre-latest for the future. Upon receiving the
// corresponding head for the cached pre-latest, it will be removed from the cache.
// Nonetheless, in case of re-org, this no longer holds true and the go-routine should
// be restarted
```

sync/sync.go
```diff
@@ -627,55 +614,233 @@ func (s *Synchronizer) fetchAndStorePending(ctx context.Context) error {
...
+		select {
+		case <-ctx.Done():
+			cancel()
+			fetchWg.Wait()
+			return
+		case head := <-newHeadSub.Recv():
+			cancel()
+			fetchWg.Wait()
+			subCtx, cancel = context.WithCancel(ctx)
+			fetchWg.Go(func() { s.fetchPreLatest(subCtx, head, seenBlockHashes, preLatestChan) })
```
rodrodros: Why make it a go-routine with a wait group instead of a single function call in the same thread?. 

In the meantime, while it is executing and we get `ctx.Done()`, it will stop inside `s.fetchPreLatest(...) at some point either way, right?

The way I see it `fetchPreLatest` at every blocking step there is a check for `ctx.Done()`

sync/sync.go
rodrodros: Is there a mechanism to start the go-routine 🤔 (note for me)

sync/sync.go
rodrodros: Mmm, random thought, but perhaps not good, I want to share it nonetheless. What do you think of
```golang
type parentBlockHash = felt.Felt

// and then ...

cache map[parentBlockHash]*core.PreLatest
```

Making it self-described by its typed definition, otherwise the old reliable `seenParentBlockHashes`, but I believe that name is not super expressive. The most expressive name would be `parentBlockHashesToPreLatest` (or even longer `parentBlockHashesToPreLatestCache), which is also verbose.  

Not sure what would be the best, to be honest. I would try my first suggestion, but I don't know how it will hold long-term.

sync/sync.go
EgeCaner: Yes, I can remove it. Also when prelatest and pre_confirmed fetching is cancelled via context or pending is not ready yet we log messages (DEBUG level) but they are frequent.

>DEBUG	sync/sync.go:707	Error while trying to poll pre_latest block	{"err": "no pending block"}
>DEBUG	sync/sync.go:829	Error while trying to poll pre_confirmed block	{"err": "context canceled"}

Do you think will they create panic, or fine?

sync/sync.go
EgeCaner: I agree, let me rename it to something more expressive

core/state_update.go
EgeCaner: oldMap has the type `map[felt.Felt]map[felt.Felt]*felt.Felt`. In the branch where `oldMap[key]` does not exist, attempting to copy directly into a nil map causes a panic.

The previous implementation had this problematic line:

```go
oldMap[addr] = newAddrStorage
```

This assigned the `newAddrStorage` map reference directly to `oldMap[addr]` without making a copy. As a result, multiple consecutive calls to `StateDiff.Merge` that modify the same address would overwrite the original `newAddrStorage` map.

EgeCaner: As far as I know this was only affecting single tx traces on `pre_confirmed` where per-tx state diff is utilized (which some entries may be corrupted due to Merge), else we are using aggregated state diff.

infrmtcs: nit: use `maps.Clone` is cleaner here

PR #3091, opened 2025-08-29
refactor: Use function from the standard library

zhoujiaweii:
`maps.Copy` has been moved to the standard library and can provide better performance with the new version.


rodrigo-pino:
@zhoujiaweii can you provide the source for this improvement

PR #3100, opened 2025-09-04
feat: Reorg detection

infrmtcs:


sync/reorg_test.go
```diff
@@ -0,0 +1,335 @@
...
+			committedBlock.StateUpdate,
+			committedBlock.NewClasses,
+		)
+		require.NoError(t, err)
+	}
+
+	dataSource.setBlocks(b.blocks)
+}
+
+func getBuilder(log utils.Logger, bc *blockchain.Blockchain) *builder.Builder {
```
rodrodros: nit `getBuilder` to `newTestBuilder`?

sync/reorg_test.go
```diff
@@ -0,0 +1,335 @@
...
+	executor := builder.NewExecutor(bc, vm.New(false, log), log, false, true)
+	builder := builder.New(bc, executor)
+	return &builder
+}
+
+func getBuildParams(sequencer *felt.Felt) builder.BuildParams {
```
rodrodros: nit `getBuildParams` -> `defaultBuildParams` or `newBuildParams` since it is a constructor and not a getter

sync/sync.go
```diff
@@ -196,6 +196,16 @@ func (s *Synchronizer) fetcherTask(ctx context.Context, height uint64, verifiers
...
 			committedBlock, err := s.dataSource.BlockByNumber(ctx, height)
 			if err != nil {
+				lastPossiblyValidHeight := s.isReverting(ctx, height)
+				if lastPossiblyValidHeight != nil {
+					return func() {
+						verifiers.Go(func() stream.Callback {
+							return func() {
+								s.revertTask(ctx, *lastPossiblyValidHeight, resetStreams)
+							}
+						})
```
rodrodros: Returns a function that calls a go routine with a function that returns a function that calls a function. 

Partially as a note to myself, is there a way around this?

sync/sync.go
```diff
@@ -208,6 +218,47 @@ func (s *Synchronizer) fetcherTask(ctx context.Context, height uint64, verifiers
 	}
 }
 
+func (s *Synchronizer) isReverting(ctx context.Context, nextHeight uint64) (lastPossiblyValidHeight *uint64) {
```
rodrodros: I prefer that instead of returning a pointer, we would return (uint64, bool) following the  `val, ok := f()` pattern in this instance. This way memory allocated stays the same [uint64 ptr + uint64] vs [uint64 + bool]. The only difference the latter will not use the heap.

sync/sync.go
```diff
@@ -208,6 +218,47 @@ func (s *Synchronizer) fetcherTask(ctx context.Context, height uint64, verifiers
...
+		if localHead, err = s.blockchain.BlockHeaderByNumber(localHeight); err != nil {
+			return nil
+		}
+	}
+
+	if *remoteHead.Hash == *localHead.Hash {
+		return nil
+	}
+
+	return utils.HeapPtr(remoteHeight - 1)
```
rodrodros: Why use the `return utils.HeapPtr(x)`  instead of `return &x`?

infrmtcs: `remoteHeight - 1` is an expression, so I think the outcome would be the same

rodrodros: Yes, not discussing the output but the semantics. I've seen through some of your PRs that you would choose to use HeapPtr than assigning a reference. Curious of what was the reason behind it.

I'm not a fan of the function call because if not I lined it will incur in an extra call cost

sync/sync.go
infrmtcs: Updated

sync/reorg_test.go
infrmtcs: Updated

sync/reorg_test.go
infrmtcs: Updated

sync/sync.go
infrmtcs: `fetchers` and `verifiers` are `stream.Stream`. A `stream.Stream` enqueue `func() func()` items which runs 2 things:
- The body of the outer `func()` runs concurrently.
- The returned `func()` runs sequentially in the queue order.

We're doing 3 things here: `fetcherTask` and `verifierTask` can be done concurrently, while `storeTask` needs to be sequential. The current code achieves this by:
- `verifierTask` runs the verifier logic and return the `storeTask`.
- In order to run `storeTask` in the correct order, `verifierTask` must be enqueued in the correct order
- Therefore `fetcherTask` runs the fetcher logic and return a `func()` to enqueue `verifierTask`.

In the reorg case:
- There is nothing to verify, while we still want to execute `revertTask` and `storeTask` sequentially.
- Therefore, we still return a callback, which enqueue a task that runs nothing concurrently and return a `revertTask`.

sync/sync.go
```diff
@@ -208,6 +217,50 @@ func (s *Synchronizer) fetcherTask(ctx context.Context, height uint64, verifiers
...
+		return 0, false
+	}
+	remoteHeight := remoteHead.Number
+
+	// If a newer block is available, revert will be handled in storeTask
+	if remoteHeight > localHeight {
+		return 0, false
+	}
+
+	// If the latest block is at the same height as the head, compare their hashes
```
rodrodros, line 248: This `if` comment at line 248, I believe it is best placed below, at line 256

sync/sync.go
```diff
@@ -345,26 +400,47 @@ func (s *Synchronizer) storeTask(
...
 
-	if head != nil {
-		s.storeEmptyPendingData(head)
+			// Double check to avoid reverting the head if the hash is the same
+			if *remoteHeader.Hash == *localHeader.Hash {
+				break
+			}
+
+			// Terminate the loop if the parent hash is the same
+			shouldContinue = *remoteHeader.ParentHash != *localHeader.ParentHash
```
rodrodros, line 431: Thinking on an extreme case, can block 0 parent hash be nil (I guess it can since we are dealing with references 🤔 )

rodrigo-pino (commented):
All looks good in sync.go, want to give a final look at reorg_test.go

sync/sync.go
infrmtcs: I queried feeder gateway on block 0 and got parent block hash = 0x0, so I think it's alway non-nil

sync/sync.go
```diff
@@ -196,6 +196,15 @@ func (s *Synchronizer) fetcherTask(ctx context.Context, height uint64, verifiers
 		default:
 			committedBlock, err := s.dataSource.BlockByNumber(ctx, height)
 			if err != nil {
+				if lastPossiblyValidHeight, isReorg := s.isReverting(ctx, height); isReorg {
```
EgeCaner, line 199: Im not sure we should trigger this mechanism for all errors. I guess triggering this check only for `StarknetErrorCode.BLOCK_NOT_FOUND` error should be sufficient. This error is returned in the response body with HTTP status code 400.

```json
{"code": "StarknetErrorCode.BLOCK_NOT_FOUND", "message": "Block number 2000000 was not found."}
```

However our feeder.Client .get method returns http status as error([ref](https://github.com/NethermindEth/juno/blob/8d2ccb333fe113e5dff6e2c51ebc355fa4947fd8/clients/feeder/feeder.go#L147)), thus we cannot differentiate HTTP 400 for other reason from `BLOCK_NOT_FOUND`. Maybe we can limit the scope to status code 400 as quick solution? WDYT?

EgeCaner: Idealy we should map errors to Feeder Errors as we do in gateway client ([ref](https://github.com/NethermindEth/juno/blob/8d2ccb333fe113e5dff6e2c51ebc355fa4947fd8/clients/gateway/gateway.go#L132))

sync/sync.go
```diff
@@ -345,26 +400,47 @@ func (s *Synchronizer) storeTask(
...
+			// Double check to avoid reverting the head if the hash is the same
+			if *remoteHeader.Hash == *localHeader.Hash {
+				break
+			}
+
+			// Terminate the loop if the parent hash is the same
+			shouldContinue = *remoteHeader.ParentHash != *localHeader.ParentHash
+		}
+
+		// Actuallly revert the head and restart the sync process
```
EgeCaner, line 434:
```suggestion
		// Actually revert the head and restart the sync process
```

PR #3103, opened 2025-09-05
fix(rpc): relay to sequencer tracing blocks from 1943705 to 1952704

rodrodros:


rpc/v8/trace.go
```diff
@@ -119,70 +114,59 @@ func (h *Handler) TraceBlockTransactions(
...
-			return trace, httpHeader, nil
+		// Check if the trace should be provided by the feeder gateway
+		blockVer, err := core.ParseBlockVersion(block.ProtocolVersion)
+		if err != nil {
+			return nil,
+				defaultExecutionHeader(),
+				rpccore.ErrUnexpectedError.CloneWithData(err.Error())
+		}
+		// For older cairo versions we rely on the sequencer for retrocompatibility
+		fetchFromFeederGW := blockVer.LessThan(core.Ver0_13_2)
```
EgeCaner: As I see here we were not falling back to FGW for version 0.13.1.1  with less than 0.13.2  check we will fallback to FGW. Should it be handled explicitly?

PR #3109, opened 2025-09-12
feat: Rewrite WAL usage and avoid missing entries

infrmtcs:
**Problem 1: WAL writes violate state machine purity**
- WAL writes were scattered between state machine and driver
- State machine should be pure and side-effect free, but was writing directly to DB

**Problem 2: Replay behavior diverges from runtime**
- Future messages only update vote counters without triggering rules during runtime
- During replay, rules are triggered every time, causing inconsistent behavior

**Solutions:**
- Introduce `WriteWAL` action to separate concerns: state machine decides *what* to write, driver handles *how* to write
- Add `WALStart` entry type to mark height boundaries, ensuring replay follows actual execution path
- Clean up redundant mock code in updated tests

This ensures state machine purity and consistent replay behavior across different execution contexts.


consensus/db/db.go
```diff
@@ -33,8 +33,6 @@ type walMsgCount uint32
 // 2. Right before we broadcast a message
 //
 // We call Delete when we start a new height and commit a block
-//
-//go:generate mockgen -destination=../mocks/mock_db.go -package=mocks github.com/NethermindEth/juno/consensus/db TendermintDB
```
rodrodros, line 37: Why don't we need a mock anymore?

consensus/starknet/starknet.go
```diff
@@ -25,6 +25,7 @@ type (
 	MessageHeader = types.MessageHeader[Address]
 
 	Action             = actions.Action[Value, Hash, Address]
+	WriteWAL           = actions.WriteWAL[Value, Hash, Address]
```
rodrodros, line 28: Why not at the bottom? Following alphabetic placing

consensus/tendermint/process.go
```diff
@@ -7,82 +7,99 @@ import (
...
-func (s *stateMachine[V, H, A]) processLoop(action actions.Action[V, H, A], recentlyReceivedRound *types.Round) []actions.Action[V, H, A] {
-	actions, shouldContinue := []actions.Action[V, H, A]{}, true
-	if action != nil {
-		actions = append(actions, action)
+func (s *stateMachine[V, H, A]) ProcessWAL(walEntry wal.Entry[V, H, A]) []actions.Action[V, H, A] {
+	switch walEntry := walEntry.(type) {
+	case *wal.WALStart:
+		return s.ProcessStart(0)
+	case *wal.WALProposal[V, H, A]:
+		return s.ProcessProposal((*types.Proposal[V, H, A])(walEntry))
```
rodrodros, line 79: Huge nitpick: the same way we have `wal.Entry`, couldn't we have also `wal.Start` and `wal.Proposal`, and so on. 

Tried the changes locally, and no issues. Also very subjective, but I think it's a bit better. Feel free to ignore

consensus/driver/driver.go
```diff
@@ -109,37 +137,55 @@ func (d *Driver[V, H, A]) Run(ctx context.Context) error {
 // It returns true if a commit action was executed. This is to notify the caller to start a new height with round 0.
```
rodrodros, line 138: Please add to the description how does the `isReplaying` flag affects the execution of this method

consensus/driver/driver.go
```diff
@@ -109,37 +137,55 @@ func (d *Driver[V, H, A]) Run(ctx context.Context) error {
...
-				}
-			})
-		case *actions.Commit[V, H, A]:
-			if err := d.db.Flush(); err != nil {
-				d.log.Fatalf("failed to flush WAL during commit", "height", action.Height, "round", action.Round, "err", err)
-			}
+			d.setTimeout(ctx, types.Timeout(*action))
 
+		case *actions.Commit[V, H, A]:
 			d.log.Debugw("Committing", "height", action.Height, "round", action.Round)
```
rodrodros: Just something to add that I've just noticed by accident, `action.Height` and `action.Round` escape both to heap. This is an artifact of `Debugw` and no way around that I can see right now :(

consensus/driver/driver.go
```diff
@@ -109,37 +137,55 @@ func (d *Driver[V, H, A]) Run(ctx context.Context) error {
...
+}
+
+func (d *Driver[V, H, A]) setTimeout(ctx context.Context, timeout types.Timeout) {
+	d.scheduledTms[timeout] = time.AfterFunc(d.getTimeout(timeout.Step, timeout.Round), func() {
+		select {
+		case <-ctx.Done():
+		case d.timeoutsCh <- timeout:
+		}
+	})
 }
```
rodrodros, line 201: Don't know if this was intentional, but by passing `timeout` by value, we avoid a heap allocation because the compiler is smart enough to realize that it doesn't mutate after this (or whatever optimization wizardry) 👍

infrmtcs: I mainly think that `types.Timeout` is small because it's 2 64-bit numbers and an enum, so even copying is still cheap.

consensus/starknet/starknet.go
infrmtcs: Although it looks like alphabetic order, my intention was actually:
- First is the interface type (`Action`)
- Then the specialized types (if needed, because `ScheduleTimeout` doesn't have type parameters) in definition order.

The order of definitions is following the order of "expected occurrence time":
- Firstly `WriteWAL`, which may happen in the previous height, and also happens first during its height before the node can make any other action.
- Then the action a node may take, namely broadcast and schedule timeout.
- Lastly `Commit`, which is the last action of a height.

consensus/db/db.go
infrmtcs: `TendermintDB` is basically a wrapper of `db.KeyValueStore`. Instead of mocking it, the simpler and better way in my opinion is to initialize it with a `memory.Database`.
All of the other tests are using this method, except `wal_test.go`. This PR removes the dependency of `StateMachine` on database, so we can finally remove the mock code.

consensus/driver/driver.go
infrmtcs: I believe it's because we're using `SugaredLogger`, which is less performant than the original `Logger`. This version accept `any` as arguments which results in heap allocation. In order to use `Logger`, we'll need to convert the log into something like:
```suggestion
			d.log.Debug("Committing", zap.Uint("height", uint(action.Height)), zap.Int("round", int(action.Round)))
```
Personally, I think this still look good and is more structured, as it's clearer which one is field name (`height`, `round`) and which one is data.

consensus/driver/driver.go
infrmtcs: Updated.

consensus/tendermint/process.go
infrmtcs: Updated.

consensus/tendermint/process.go
```diff
@@ -7,82 +7,99 @@ import (
...
+	case *wal.Proposal[V, H, A]:
+		return s.ProcessProposal((*types.Proposal[V, H, A])(walEntry))
+	case *wal.Prevote[H, A]:
+		return s.ProcessPrevote((*types.Prevote[H, A])(walEntry))
+	case *wal.Precommit[H, A]:
+		return s.ProcessPrecommit((*types.Precommit[H, A])(walEntry))
+	case *wal.Timeout:
+		return s.ProcessTimeout(types.Timeout(*walEntry))
+	default:
+		return nil
```
rodrodros: Shouldn't we panic or something here since it is a case we should never hit. I believe it's better to fail laudly than some nil pointer exception later

consensus/driver/driver.go
```diff
@@ -101,51 +126,72 @@ func (d *Driver[V, H, A]) Run(ctx context.Context) error {
...
-					zap.Error(err))
+				return true, fmt.Errorf("failed to delete WAL messages during commit: %w", err)
 			}
 
-			return true
+			return true, nil
 		}
 	}
-	return false
+	return false, nil
```
rodrodros, line 191: On a similar note, I believe we should panic on unexpected runtime paths, or in this case where we return an error, return a new error

rodrodros (approved):

PR #3110, opened 2025-09-12
feat: `pre latest` with centralized writes

EgeCaner:
Alternative to #3088


sync/pending_polling.go
```diff
@@ -0,0 +1,672 @@
...
+			}
+		}
+	}
+}
+
+// handleHeadInPreConfirmed processes a new head during the pre_confirmed phase.
+// It computes nextHeight = head.Number + 1, clears any staged pre_latest if the
+// head catches up to the current target, and stores an empty pre_confirmed when
+// advancing. Returns updated (targetNum, stagedPreLatest).
+func (s *Synchronizer) handleHeadInPreConfirmed(
```
infrmtcs:
```suggestion
func (s *Synchronizer) handleHeadInPreConfirmedPhase(
```

sync/pending_polling.go
```diff
@@ -0,0 +1,672 @@
...
+			if !ok {
+				return
+			}
+
+			targetNum, stagedPreLatest = s.handleHeadInPreConfirmed(head, targetNum, stagedPreLatest)
+		case pl, ok := <-preLatestCh:
+			if !ok {
+				preLatestCh = nil
+				continue
+			}
```
infrmtcs: I think `preLatestCh` and `preConfirmedCh` are created inside this function, so this function should manage its lifecycle. Probably we don't need to check if it's closed here, and instead we can close it in the Done case if possible.

sync/pending_polling.go
```diff
@@ -0,0 +1,672 @@
...
+}
+
+// handleHeadInPreConfirmed processes a new head during the pre_confirmed phase.
+// It computes nextHeight = head.Number + 1, clears any staged pre_latest if the
+// head catches up to the current target, and stores an empty pre_confirmed when
+// advancing. Returns updated (targetNum, stagedPreLatest).
+func (s *Synchronizer) handleHeadInPreConfirmed(
+	head *core.Block,
+	targetNum uint64,
+	stagedPreLatest *core.PreLatest,
```
infrmtcs, line 276: It looks like we're not modifying `stagedPreLatest` here.

sync/pending_polling.go
```diff
@@ -0,0 +1,672 @@
...
+	return targetNum, stagedPreLatest
+}
+
+// handlePreLatest processes an incoming pre_latest during the pre_confirmed phase.
+// If it raises the target, it stages the attachment and stores a baseline with it.
+// Returns updated (targetNum, stagedPreLatest).
+func (s *Synchronizer) handlePreLatest(
+	pl *core.PreLatest,
+	targetNum uint64,
+	stagedPreLatest *core.PreLatest,
```
infrmtcs, line 301: It looks like we're not using or modifying `stagedPreLatest` here.

sync/pending_polling.go
infrmtcs: Probably the idea here is to either return the existing one of the new one.

sync/pending_polling.go
infrmtcs: Probably the idea here is to either return the existing one of the new one.

sync/pending_polling.go
```diff
@@ -0,0 +1,672 @@
...
+// we clear the staged attachment and also clear the stored baseline’s attachment (if present).
+func (s *Synchronizer) runPreConfirmedPhase(ctx context.Context, headsSub *feed.Subscription[*core.Block]) {
+	preLatestCh := make(chan *core.PreLatest, 1)
+	preConfirmedCh := make(chan *core.PreConfirmed, 1)
+
+	go s.pollPreLatest(ctx, preLatestCh)
+	go s.pollPreConfirmed(ctx, preConfirmedCh)
+
+	var (
+		targetNum       uint64
```
infrmtcs: I think `targetNum` is a confusing name, whether it's target of head, prelatest or preconfirmed.

sync/pending_polling.go
```diff
@@ -0,0 +1,672 @@
...
+	if s.pendingPollInterval == 0 || s.preConfirmedPollInterval == 0 {
+		s.log.Infow("Pending data polling is disabled")
+		return
+	}
+
+	headsSub := s.newHeads.SubscribeKeepLast()
+	defer headsSub.Unsubscribe()
+
+	// Phase 1: pending path
+	switched := s.runPendingPhase(ctx, headsSub)
```
infrmtcs, line 392:
```suggestion
	switched, err := s.runPendingPhase(ctx, headsSub)
```

sync/pending_polling.go
```diff
@@ -0,0 +1,672 @@
...
+
+	pc.WithPreLatest(stagedPreLatest)
+	changed, err := s.StorePreConfirmed(pc)
+	if err != nil {
+		s.log.Debugw("Error while trying to store pre_confirmed block", "err", err)
+		return
+	}
+	if changed {
+		s.pendingDataFeed.Send(pc)
+	}
```
infrmtcs, line 334: Why isn't this done inside `StorePreConfirmed`?

sync/pending_polling.go
```diff
@@ -0,0 +1,672 @@
...
+// head catches up to the current target, and stores an empty pre_confirmed when
+// advancing. Returns updated (targetNum, stagedPreLatest).
+func (s *Synchronizer) handleHeadInPreConfirmed(
+	head *core.Block,
+	targetNum uint64,
+	stagedPreLatest *core.PreLatest,
+) (uint64, *core.PreLatest) {
+	next := head.Number + 1
+	if next == targetNum && stagedPreLatest != nil {
+		s.UpdatePreLatestAttachment(targetNum, nil)
```
infrmtcs: If new head is the same as current prelatest, don't we preserve the prelatest?

sync/pending_polling.go
```diff
@@ -0,0 +1,672 @@
...
+// If it raises the target, it stages the attachment and stores a baseline with it.
+// Returns updated (targetNum, stagedPreLatest).
+func (s *Synchronizer) handlePreLatest(
+	pl *core.PreLatest,
+	targetNum uint64,
+	stagedPreLatest *core.PreLatest,
+) (uint64, *core.PreLatest) {
+	next := pl.Block.Number + 1
+
+	if next > targetNum {
```
infrmtcs: There can be a case where pending version is replaced by latest version.

sync/pending_polling.go
EgeCaner: This is for the following case, when latest is N, prelatest N+1, preConfirmed N+2, then when head for n +1 arrives our pre_confirmed is still valid but since prelatest is in DB now as latest we can remove it from preConfirmed.

core/pending.go
```diff
@@ -64,29 +73,65 @@ func (p *Pending) GetTransactionStateDiffs() []*StateDiff {
...
 	}
 }
 
+func (p *PreConfirmed) WithNewClasses(newClasses map[felt.Felt]Class) {
+	p.NewClasses = newClasses
+}
+
+func (p *PreConfirmed) WithPreLatest(preLatest *PreLatest) {
+	p.PreLatest = preLatest
+}
```
rodrodros: I believe that these setter methods should return `p`, so they can be chained. Eg:

`p.WithNewClasses(...).WithPreLatests(...)`

Even if we don't use it in this case, to keep it consistent with the project style

core/pending.go
```diff
@@ -64,29 +73,65 @@ func (p *Pending) GetTransactionStateDiffs() []*StateDiff {
...
+
+func (p *PreConfirmed) Clone() *PreConfirmed {
```
rodrodros: I would argue it should be called `Copy()` instead of `Clone()` since in my experience, cloning means doing a full copy (duplicating the entire type), not just its references. 

Also, I am not sure what the benefit of making a shallow copy is.

sync/pending_polling.go
EgeCaner: renamed as `targetPreConfirmedNum`

core/pending.go
EgeCaner: Addressed

sync/pending_polling.go
EgeCaner: Removed chanel close handling for channels managed in same scope

sync/pending_polling.go
EgeCaner: renamed

sync/pending_polling.go
EgeCaner: `runPendingPhase` does not exit with error, all errors are gracefully handled, skipped. It only exits when, context is cancelled or switching to preConfirmed polling. Would it still worth adding?

sync/pending_polling.go
EgeCaner: Atm pendingData writes can still race with `sync.StoreTask` storing empty pending, this result in success/failure of CAS operation and many branches if I want to broadcast inside store method. Also IMO broadcasting outside of StoreFunc sounds better to me, store should only store, broadcast after should be optional. I will do same for pending, this will also avoid broadcasting store empty operations

core/pending.go
EgeCaner: I introduced this method to assign prelatest to nil without modifying the original ptr. To clear prelatest when we have preconfirmed N +2 and when we receive N+1 as a head. I will rename or maybe inline it and remove this function

sync/pending_polling.go
EgeCaner: Can you elaborate it bit more?

sync/pending_polling.go
EgeCaner: Did the same for pending

core/pending.go
EgeCaner: Renamed as `Copy`

core/pending.go
```diff
@@ -64,29 +73,67 @@ func (p *Pending) GetTransactionStateDiffs() []*StateDiff {
...
+	return p
+}
+
+func (p *PreConfirmed) Copy() *PreConfirmed {
+	if p == nil {
+		return nil
+	}
+	cp := *p // shallow copy of the struct
+	return &cp
+}
```
rodrodros: I don't agree with having a nil-safe method. It doesn't go with the project style (or it shouldn't). Points are:
1. The expected behaviour when doing (`type.Method()`) is for the call to fail unless explicitly stated. Hiding this fact can subtly hide bugs (not the case here).
2. Looking at the code where Copy() is being done, after the call, there is an immediate call that will cause a nil reference exception. 

I think having the Copy method is good, but let's make it panic on call if using a nil parameter. Suggestion in a comment below. 

I think nil-safe methods are passable in certain scenarios, but they are very specific, and I am not convinced this is one of them.

sync/pending_polling.go
```diff
@@ -0,0 +1,673 @@
...
+	if pc.PreLatest == preLatest {
+		// No change.
+		return false
+	}
+
+	// Clone and update attachment on the clone.
+	next := pc.Copy()
+	next.WithPreLatest(preLatest)
+
+	return s.pendingData.CompareAndSwap(curPtr, utils.HeapPtr[core.PendingData](next))
```
rodrodros: Keeping the Copy() method but not making it nil safe

```golang
var next *PreConfirmed
if pc != nil {
   next = pc.Copy()
}
```

rodrodros, line 464: Perhaps I am missing something, but I think `utils.HeapPtr` is not necessary

sync/pending_polling.go
```diff
@@ -0,0 +1,673 @@
...
+	}
+
+	if pc.PreLatest == preLatest {
+		// No change.
+		return false
+	}
+
+	// Clone and update attachment on the clone.
+	next := pc.Copy()
+	next.WithPreLatest(preLatest)
```
rodrodros, line 462: Also the nil safety is redundant, since `next.WithPrelatest(...)` will fail if `next` is nil. I believe all of this should be wrapped inside an `if` in that case (and if that was the intention).

core/pending.go
```diff
@@ -115,6 +162,35 @@ func (p *PreConfirmed) GetTransactionStateDiffs() []*StateDiff {
 	return p.TransactionStateDiffs
 }
 
+func (p *PreConfirmed) GetPreLatest() *PreLatest {
+	return p.PreLatest
+}
```
rodrodros, line 164: I think we don't need this getter. Did you add it for future proofing? I think it might be to keep it consistent with other parts of the code, but I believe this is not good Go programming style. 

I think if `PreLatest` is public, then it shouldn't have a getter in the first place. If we want to keep the getter, then it should be private and the getter should be called `PreLatest` (it is not very go-ish having `GetSmth` methods). 

We can leave it here if the rest of the Sync package is like this.

sync/pending_polling.go
```diff
@@ -0,0 +1,673 @@
...
+// handleTickerPreLatest polls a pre-latest once and either:
+//   - emits it to out and returns true when if delivered,
+//   - caches it by its ParentHash for a future head and returns false,
+//   - returns false on errors or context cancellation.
+//
+// Caller should invoke only when at tip and not yet delivered for the current head.
+func (s *Synchronizer) handleTickerPreLatest(
+	ctx context.Context,
+	currentHead *core.Block,
+	seenByParent *lru.BasicLRU[felt.Felt, *core.PreLatest],
```
rodrodros, line 32: This cache is known to be very, very inefficient if used wrong. Just a note that we are making the most of it.

sync/pending_polling.go
infrmtcs: The reason I'm asking is because we're checking `ctx.Err()` below.

sync/pending_polling.go
infrmtcs: I'm thinking about the case where latest = N and prelatest = N + 1 and prelatest is in "pending" format. Then a prelatest = N + 1 in "latest" format arrives.

sync/pending_polling.go
```diff
@@ -0,0 +1,673 @@
...
+		// Try to update attachment to match p.PreLatest (can be nil to clear)
+		_ = s.UpdatePreLatestAttachment(incomingB.Number, p.PreLatest)
+		return false, nil
+	}
+
+	// Otherwise, swap in the better pre_confirmed.
+	return s.pendingData.CompareAndSwap(
+		existingPtr,
+		utils.HeapPtr[core.PendingData](p),
+	), nil
```
infrmtcs: I think if it's easier to keep track if we can reduce the number of returns. For example, something like:
```
	shouldPreserveCurrentPreConfirmed := existingPtr != nil && *existingPtr != nil && !(*existingPtr).Validate(head) && ...
	if shouldPreserveCurrentPreConfirmed {
		_ = s.UpdatePreLatestAttachment(incomingB.Number, p.PreLatest)
		return false, nil
	}
	return s.pendingData.CompareAndSwap(
		existingPtr,
		utils.HeapPtr[core.PendingData](p),
	), nil
```

sync/pending_polling.go
```diff
@@ -0,0 +1,673 @@
...
+				select {
+				case <-ctx.Done():
+					return
+				case out <- pl:
+					deliveredForHead = true
+				}
+			}
+
+		case <-ticker.C:
+			if !shouldPoll {
```
infrmtcs, line 116: nit: Should we move `shouldPoll` definition here?

core/pending.go
EgeCaner: We have this getter even though `PreLatest` is public, because in most places we interact with PendingData via interface not the concrete types, specifically in RPC.

EgeCaner: We can make `PreLatest` and other fields private, then have getters/setters...  would it be more idiomatic?

sync/pending_polling.go
EgeCaner: Unfortunately I cannot use `next` or `&next` here, they are `*core.PreConfirmed` and `**core.PreConfirmed` respectively and what we need here is `*core.PendingData`

sync/pending_polling.go
EgeCaner: I used ctx.Err() to determine if parent context is cancelled but switched should be enough, if `runPendingPhase` exit due to context cancellation it will return false. ctx.Err() check seems redundant, I will remove

EgeCaner:
Closing in favor of #3120  and #3124

PR #3120, opened 2025-09-17
refactor: sync pending-data

EgeCaner:


sync/pending_polling.go
```diff
@@ -0,0 +1,475 @@
...
+			return false, err
+		}
+		head = nil
+	}
+
+	if !p.Validate(head) {
+		return false, fmt.Errorf("store pending: %w", blockchain.ErrParentDoesNotMatchHead)
+	}
+
+	existingPtr := s.pendingData.Load()
```
rodrodros, line 81: In what instances is the data stored at `s.pendingData` is nil?

sync/pending_polling.go
```diff
@@ -0,0 +1,475 @@
...
+// makeStateDiffForEmptyBlock constructs a minimal state diff for an empty block.
+// It optionally writes a historical block hash mapping when blockNumber >= blockHashLag.
+func makeStateDiffForEmptyBlock(bc blockchain.Reader, blockNumber uint64) (*core.StateDiff, error) {
+	stateDiff := &core.StateDiff{
+		StorageDiffs:      make(map[felt.Felt]map[felt.Felt]*felt.Felt),
+		Nonces:            make(map[felt.Felt]*felt.Felt),
+		DeployedContracts: make(map[felt.Felt]*felt.Felt),
+		DeclaredV0Classes: make([]*felt.Felt, 0),
+		DeclaredV1Classes: make(map[felt.Felt]*felt.Felt),
+		ReplacedClasses:   make(map[felt.Felt]*felt.Felt),
```
rodrodros: Please give size 0 to the maps as well, except for StorageDiffs, which needs size one :)

sync/pending_polling.go
EgeCaner: pendingData is only nil at startup, with the first fetched block it will be assigned to emptyPendingData, then it will always be empty place holder or actual pending/pre_confirmed

sync/pending_polling.go
```diff
@@ -0,0 +1,475 @@
...
+		if err := s.storeEmptyPending(lastHeader); err != nil {
+			s.log.Errorw("Failed to store empty pending block", "number", lastHeader.Number)
+		}
+	}
+}
+
+// pollPreConfirmed polls for the current target pre_confirmed number at a fixed interval.
+// The target is read from s.targetPreConfirmedNum and only polled while at tip.
+// On success, the pre_confirmed is forwarded to out.
+func (s *Synchronizer) pollPreConfirmed(ctx context.Context, out chan<- *core.PreConfirmed) {
```
rodrodros: super nitpick: perhaps I am missing something but I think `pollPreConfirmed` should be bellow the method for "pending". Thinking just below `runPendingPhase`

sync/sync.go
```diff
@@ -137,6 +137,7 @@ type Synchronizer struct {
 	pendingData              atomic.Pointer[core.PendingData]
 	pendingPollInterval      time.Duration
 	preConfirmedPollInterval time.Duration
+	targetPreConfirmedNum    atomic.Uint64
```
rodrodros: Let's add a better name or a description to this field because, based on a first look, I am at a loss for what its intended use is. 

After some reading, I believe this number is meant for us to know when the next time we should poll for a preconfirmed is.

I think though, it doesn't need to be a property and can be encapsulated in the `runningPreConfirmedPhase` 🤔

sync/pending_polling.go
```diff
@@ -0,0 +1,475 @@
...
+	}
+}
+
+// runPreConfirmedPhase coordinates baselines and final storage.
+func (s *Synchronizer) runPreConfirmedPhase(ctx context.Context, headsSub *feed.Subscription[*core.Block]) {
+	preConfirmedCh := make(chan *core.PreConfirmed, 1)
+
+	go s.pollPreConfirmed(ctx, preConfirmedCh)
+
+	var targetPreConfirmedNum uint64
```
rodrodros: I think you can create the atomic point for targetPreConfirmedNum up here and pass it to both the go routine while using it in this function as well.

sync/pending_polling_test.go
```diff
@@ -0,0 +1,484 @@
...
+			written, err := s.StorePreConfirmed(preConfirmed)
+			require.Error(t, err)
+			require.False(t, written)
+		})
+	})
+
+	t.Run("returns error if ProtocolVersion unsupported", func(t *testing.T) {
+		pc := &core.PreConfirmed{
+			Block: &core.Block{
+				Header: &core.Header{Number: 1, ProtocolVersion: "1.9.0"},
```
rodrodros: Let's change the `ProtocolVersion` fixed string for a dynamic one that never fails (`SupportedStarknetVersion.IncMajor().String()`)

rodrigo-pino (approved):

sync/pending_polling.go
```diff
@@ -0,0 +1,471 @@
...
+
+	if !p.Validate(head) {
+		return false, fmt.Errorf("store pending: %w", blockchain.ErrParentDoesNotMatchHead)
+	}
+
+	existingPtr := s.pendingData.Load()
+
+	shouldPreserveCurrentPending := existingPtr != nil && *existingPtr != nil &&
+		(*existingPtr).Validate(head) &&
+		(*existingPtr).GetBlock().TransactionCount >= p.Block.TransactionCount
```
infrmtcs: nit: We can have a separate function, as it looks the same as the logic in preconfirmed. Actually most of the logic are similar too. Can be in a different PR though.

EgeCaner: I agree, store functions are almost same for pending/pre_confirmed, except pre_confirmed's prelatest swap. But we can definitely extract shouldPreserve logic into helper function.

PR #3124, opened 2025-09-18
feat: pre latest polling

EgeCaner:


core/pending.go
```diff
@@ -154,5 +177,15 @@ func (p *PreConfirmed) Validate(parent *Header) bool {
...
+		return false
+	}
+
+	if p.Block.Number == p.PreLatest.Block.Number+1 &&
+		p.PreLatest.Block.ParentHash.Equal(parent.Hash) {
+		// preconfirmed is latest + 2
+		return true
+	}
+
 	return false
```
rodrodros: nitpick: this can be a single return statement:
```golang
if <cond> {
  return true
}
return false
```

to 
```golang
return <cond>
```

sync/data_source.go
```diff
@@ -140,12 +140,5 @@ func (f *feederGatewayDataSource) PreConfirmedBlockByNumber(ctx context.Context,
 		return core.PreConfirmed{}, err
 	}
 
-	h, err := f.blockchain.HeadsHeader()
-	if err != nil && !errors.Is(err, db.ErrKeyNotFound) {
-		return core.PreConfirmed{}, err
-	} else if err == nil {
-		preConfirmed.StateUpdate.OldRoot = h.GlobalStateRoot
-	}
-
```
rodrodros, line 149: Can you remind me why is that this disappeared?

infrmtcs (approved):

sync/data_source.go
EgeCaner: With the notion of `pre_latest`, preconfirmed wont be always based from db Head, and when prelatest is the base for `pre_confirmed` we wont be able to assign OldRoot.

rodrigo-pino (approved):

PR #3125, opened 2025-09-18
feat(core): add new types for core felt based types

rodrigo-pino:


core/felt/ctors.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+// FromUint64 creates a new Felt based type given an uint64
+func FromUint64[F FeltLike](num uint64) F {
+	f := fp.NewElement(num)
+	return F(f)
+}
+
+// FromInt64 creates a new Felt based type given an int64
+func FromInt64[F FeltLike](num int64) F {
```
infrmtcs: Do we have practical usages of this? I think if not we can remove it to avoid wrong usage.

core/felt/ctors.go
```diff
@@ -0,0 +1,98 @@
...
+
+// Interface that represents all types that implement Felt as a sublying type
+// E.g: Hash and Address
+type FeltLike interface {
+	~[4]uint64
+}
+
+// New crates a new Felt based type given any integer. Use this function with
+// care since it forces a heap allocation. For efficient code use `FromUint64`
+func New[F FeltLike, N constraints.Integer](num N) *F {
```
infrmtcs: I think we should also remove this, to avoid confusion and unnecessary check to compare with 0.

core/felt/ctors.go
rodrodros: What could be the wrong usage?

The reason the `FromInt<smth>` is to make it easy to add when dealing with those types. If I remove this function I would have to add this logic into `FromInt` which then is kind of messy. I could make this function private and call it `FromNeg` since its only use is for negative numbers but I think that might be confusing as well.

core/felt/ctors.go
rodrodros: I'm not sure why we should remove it, there are many instances when we crate felt by pointers, this function just makes it easier. 

Do you think is better to have a `NewFromUint64` and `NewFromInt64` and so on? I think this is specially useful when writing test code.

core/felt/ctors.go
infrmtcs: I'm trying to search for its usage in the current code, but it seems that we don't use it anywhere except in the `consensus/mocks.go`, which is only for testing purpose, and it's using `int` because of convenience, not for negative integer. So technically I think we're not converting negative integer to felt anywhere.

core/felt/ctors.go
infrmtcs: I think `NewFromUint64` and `NewFromInt64` are better, because don't have to check for `num >= 0`.

core/felt/ctors.go
```diff
@@ -0,0 +1,98 @@
...
+func FromInt64[F FeltLike](num int64) F {
+	if num >= 0 {
+		return FromUint64[F](uint64(num))
+	}
+
+	// To get the negative felt value of `num` we
+	// subtract Abs(num) from zero
+	posNum := uint64(num * -1)
+	value := FromUint64[Felt](posNum)
+	value.Sub(&Zero, &value)
```
infrmtcs:
```suggestion
	value.Neg(&value)
```

core/felt/ctors.go
infrmtcs: Combined with my other comment below, I think we only need `NewFromUint64`

core/felt/ctors.go
infrmtcs: Or I think we should use the `SetInt64` function from `fp.Element`

core/felt/ctors.go
infrmtcs: Or I think we should use the `SetInt64` function from fp.Element

consensus/votecounter/vote_counter_test.go
```diff
@@ -112,7 +112,7 @@ func TestVoteCounter(t *testing.T) {
 				case *proposalTestCase:
 					t.Run("Try with invalid proposer", func(t *testing.T) {
 						proposal := testCase.proposal
-						proposal.Sender = starknet.Address(felt.FromUint64(uint64(len(tests))))
+						proposal.Sender = starknet.Address(felt.FromUint64[felt.Felt](uint64(len(tests))))
```
infrmtcs:
```suggestion
						proposal.Sender = felt.FromUint64[starknet.Address](uint64(len(tests)))
```

consensus/votecounter/validator_test.go
```diff
@@ -14,7 +14,7 @@ type mockValidator struct {
 
 func newMockValidator(proposerIndex uint64, votingPower map[starknet.Address]types.VotingPower) *mockValidator {
 	return &mockValidator{
-		proposer:    utils.HeapPtr(starknet.Address(felt.FromUint64(proposerIndex))),
+		proposer:    utils.HeapPtr(starknet.Address(felt.FromUint64[felt.Felt](proposerIndex))),
```
infrmtcs:
```suggestion
		proposer:    utils.HeapPtr(felt.FromUint64[starknet.Address](proposerIndex)))
```

consensus/votecounter/round_data_test.go
```diff
@@ -141,10 +141,10 @@ func (v voter) propose(idIndex uint64) *proposalTestCase {
 			MessageHeader: types.MessageHeader[starknet.Address]{
 				Height: testHeight,
 				Round:  testRound,
-				Sender: starknet.Address(felt.FromUint64(uint64(v))),
+				Sender: starknet.Address(felt.FromUint64[felt.Felt](uint64(v))),
 			},
 			ValidRound: -1,
-			Value:      utils.HeapPtr(starknet.Value(felt.FromUint64(idIndex))),
+			Value:      utils.HeapPtr(starknet.Value(felt.FromUint64[felt.Felt](idIndex))),
```
infrmtcs:
```suggestion
			Value:      utils.HeapPtr(felt.FromUint64[starknet.Value](idIndex)),
```

infrmtcs: Or if we have `NewFromUint64`
```suggestion
			Value:      felt.NewFromUint64[starknet.Value](idIndex),
```

consensus/p2p/vote/vote_test.go
```diff
@@ -131,15 +129,15 @@ func TestStarknetVoteAdapter_RoundTrip(t *testing.T) {
 				MessageHeader: starknet.MessageHeader{
 					Height: types.Height(200),
 					Round:  types.Round(10),
-					Sender: address.Address(felt.FromUint64(999)),
+					Sender: felt.Address(felt.FromUint64[felt.Felt](999)),
```
infrmtcs:
```suggestion
					Sender: felt.FromUint64[felt.Address](999),
```

sync/pending.go
```diff
@@ -5,8 +5,8 @@ import (
 	"fmt"
 
 	"github.com/NethermindEth/juno/core"
-	"github.com/NethermindEth/juno/core/felt"
 	"github.com/NethermindEth/juno/core/trie"
+	"github.com/NethermindEth/juno/core/felt"
```
infrmtcs, line 8: I think running `golangci-lint` should revert this change and other similar changes

PR #3126, opened 2025-09-19
refactor: vm hardcoded gas token

EgeCaner:
Previously, gas tokens were hardcoded within the VM code, limiting Juno’s ability to support custom gas tokens for different chains. This PR is a foundational step towards enabling custom gas token support. While it does not add full support for custom gas tokens yet, it includes the necessary refactoring to make that future enhancement possible.

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -211,8 +211,11 @@ func New(cfg *Config, version string, logLevel *utils.LogLevel) (*Node, error) {
 			return nil, kErr
 		}
 		feeTokens := utils.DefaultFeeTokenAddresses
-		chainInfo := vm.NewChainInfo(cfg.Network.L2ChainID, &feeTokens)
-		nodeVM = vm.New(chainInfo, false, log)
+		chainInfo := vm.ChainInfo{
+			ChainID:           cfg.Network.L2ChainID,
+			FeeTokenAddresses: feeTokens,
+		}
+		nodeVM = vm.New(&chainInfo, false, log)
 		throttledVM = NewThrottledVM(nodeVM, cfg.MaxVMs, int32(cfg.MaxVMQueue))
 		mempool := mempool.New(database, chain, mempoolLimit, log)
 		executor := builder.NewExecutor(chain, nodeVM, log, cfg.SeqDisableFees, false)
@@ -246,8 +249,11 @@ func New(cfg *Config, version string, logLevel *utils.LogLevel) (*Node, error) {
 				return nil, fmt.Errorf("failed to fetch fee token addresses for custom network: %w", err)
 			}
 		}
-		chainInfo := vm.NewChainInfo(cfg.Network.L2ChainID, &feeTokens)
-		nodeVM = vm.New(chainInfo, false, log)
+		chainInfo := vm.ChainInfo{
+			ChainID:           cfg.Network.L2ChainID,
+			FeeTokenAddresses: feeTokens,
+		}
+		nodeVM = vm.New(&chainInfo, false, log)
 		throttledVM = NewThrottledVM(nodeVM, cfg.MaxVMs, int32(cfg.MaxVMQueue))
 		feederGatewayDataSource := sync.NewFeederGatewayDataSource(chain, adaptfeeder.New(client))
 		synchronizer = sync.New(
@@ -517,9 +523,12 @@ func (n *Node) Run(ctx context.Context) {
 			return
 		}
 		feeTokens := utils.DefaultFeeTokenAddresses
-		chainInfo := vm.NewChainInfo(n.cfg.Network.L2ChainID, &feeTokens)
+		chainInfo := vm.ChainInfo{
+			ChainID:           n.cfg.Network.L2ChainID,
+			FeeTokenAddresses: feeTokens,
+		}
 		if err = buildGenesis(n.cfg.SeqGenesisFile, n.blockchain,
-			vm.New(chainInfo, false, n.log), uint64(n.cfg.RPCCallMaxSteps)); err != nil {
+			vm.New(&chainInfo, false, n.log), uint64(n.cfg.RPCCallMaxSteps)); err != nil {
 			n.log.Errorw("Error building genesis state", "err", err)
 			return
 		}
--- a/vm/vm.go
+++ b/vm/vm.go
@@ -185,16 +185,10 @@ type CallInfo struct {
 	Selector        *felt.Felt
 	Calldata        []felt.Felt
 }
+
 type ChainInfo struct {
 	ChainID           string
 	FeeTokenAddresses starknet.FeeTokenAddresses
-}
-
-func NewChainInfo(chainID string, feeTokenAddresses *starknet.FeeTokenAddresses) *ChainInfo {
-	return &ChainInfo{
-		ChainID:           chainID,
-		FeeTokenAddresses: *feeTokenAddresses,
-	}
 }
 
 type BlockInfo struct {
--- a/consensus/consensus_test.go
+++ b/consensus/consensus_test.go
@@ -72,8 +72,11 @@ func loadGenesis(t *testing.T, log *utils.ZapLogger) (core.StateDiff, map[felt.F
 	}
 
 	feeTokens := utils.DefaultFeeTokenAddresses
-	chainInfo := vm.NewChainInfo(network.L2ChainID, &feeTokens)
-	diff, classes, err := genesis.GenesisStateDiff(genesisConfig, vm.New(chainInfo, false, log), &network, 40000000)
+	chainInfo := vm.ChainInfo{
+		ChainID:           network.L2ChainID,
+		FeeTokenAddresses: feeTokens,
+	}
+	diff, classes, err := genesis.GenesisStateDiff(genesisConfig, vm.New(&chainInfo, false, log), &network, 40000000)
 	require.NoError(t, err)
 
 	return diff, classes
@@ -98,8 +101,11 @@ func initNode(
 	bc := getBlockchain(t, genesisDiff, genesisClasses)
 
 	feeTokens := utils.DefaultFeeTokenAddresses
-	chainInfo := vm.NewChainInfo(network.L2ChainID, &feeTokens)
-	vm := vm.New(chainInfo, false, logger)
+	chainInfo := vm.ChainInfo{
+		ChainID:           network.L2ChainID,
+		FeeTokenAddresses: feeTokens,
+	}
+	vm := vm.New(&chainInfo, false, logger)
 
 	services, err := consensus.Init(
 		node.Host,
--- a/consensus/p2p/validator/transition_test.go
+++ b/consensus/p2p/validator/transition_test.go
@@ -44,12 +44,15 @@ func getBuilder(t *testing.T, seqAddr *felt.Felt) (*builder.Builder, *core.Heade
 	}
 
 	feeTokens := utils.DefaultFeeTokenAddresses
-	chainInfo := vm.NewChainInfo(network.L2ChainID, &feeTokens)
-	diff, classes, err := genesis.GenesisStateDiff(genesisConfig, vm.New(chainInfo, false, log), bc.Network(), 40000000)
+	chainInfo := vm.ChainInfo{
+		ChainID:           network.L2ChainID,
+		FeeTokenAddresses: feeTokens,
+	}
+	diff, classes, err := genesis.GenesisStateDiff(genesisConfig, vm.New(&chainInfo, false, log), bc.Network(), 40000000)
 	require.NoError(t, err)
 	require.NoError(t, bc.StoreGenesis(&diff, classes))
 	blockTime := 100 * time.Millisecond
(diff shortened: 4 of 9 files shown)


vm/rust/src/lib.rs
```diff
@@ -106,6 +103,14 @@ pub struct CallInfo {
     pub len_calldata: usize,
 }
 
+#[repr(C)]
+#[derive(Clone, Copy)]
+pub struct ChainContext {
+    pub chain_id: *const c_char,
+    pub eth_fee_token_address: [c_uchar; 32],
+    pub strk_fee_token_address: [c_uchar; 32],
+}
```
rodrodros, line 113: There is a type `ChainInfo` defined in Rust, which is identical to this type. Why don't we recreate it in Go instead of having a new one.

vm/vm.go
```diff
@@ -168,6 +185,28 @@ type CallInfo struct {
...
+	if strkFeeTokenAddress == nil {
+		strkFeeTokenAddress, _ = new(felt.Felt).SetString("0x04718f5a0fc34cc1af16a1cdee98ffb20c31f5cd61d6ab07201858f4287c938d")
+	}
+
+	return &ChainContext{
+		ChainID:             chainID,
+		EthFeeTokenAddress:  ethFeeTokenAddress,
+		StrkFeeTokenAddress: strkFeeTokenAddress,
+	}
+}
```
rodrodros: This type definition and its constructor go against everything we have been trying to target, which is reducing unnecessary heap allocations. 

Starting from `ChainContext`, which takes two `*felt.Felt`? Are they optional or meant to be shared across different types? I believe not, so they should be values.

Then the constructor, which returns a pointer to `ChainContext`; why? The type is created in this function, and it has no mutexes; it should return by value to avoid the heap allocation.

vm/vm.go
```diff
@@ -168,6 +185,28 @@ type CallInfo struct {
...
+
+func NewChainContext(chainID string, ethFeeTokenAddress, strkFeeTokenAddress *felt.Felt) *ChainContext {
+	if ethFeeTokenAddress == nil {
+		ethFeeTokenAddress, _ = new(felt.Felt).SetString("0x049d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7")
+	}
+
+	if strkFeeTokenAddress == nil {
+		strkFeeTokenAddress, _ = new(felt.Felt).SetString("0x04718f5a0fc34cc1af16a1cdee98ffb20c31f5cd61d6ab07201858f4287c938d")
+	}
```
rodrodros: ~~Of all the ways to create a felt, this is the most expensive. Please initialize the values with the default Montgomery form directly, as the one seen in `felt.One`.~~

Edit: I've checked, and it is only called once during the main workflow. It is ok having this felt initialization

builder/executor.go
```diff
@@ -84,8 +84,12 @@ func (e *executor) RunTxns(state *BuildState, txns []mempool.BroadcastedTransact
...
 		},
 		stateWriter,
-		e.blockchain.Network(),
-		e.disableFees, e.skipValidate, false, true, false)
+		e.disableFees,
+		e.skipValidate,
+		false,
+		true,
+		false,
+	)
```
rodrodros, line 92: 👍

core/contract_addresses.go
```diff
@@ -0,0 +1,9 @@
+package core
+
+import "github.com/NethermindEth/juno/core/felt"
+
+type ContractAddresses struct {
+	CoreContractAddress *felt.Felt
+	StrkL2TokenAddress  *felt.Felt
+	EthL2TokenAddress   *felt.Felt
+}
```
rodrodros: 👀

rpc/v6/trace.go
```diff
@@ -192,15 +192,25 @@ func (h *Handler) traceBlockTransactions(ctx context.Context, block *core.Block,
...
+		classes,
+		paidFeesOnL1,
+		&blockInfo,
+		state,
+		false,
+		false,
+		false,
+		false,
+		false,
+	)
```
rodrodros, line 213: 👍

rpc/v6/trace.go
```diff
@@ -290,7 +300,12 @@ func (h *Handler) Call(funcCall *FunctionCall, id *BlockID) ([]*felt.Felt, *json
 	}, &vm.BlockInfo{
 		Header:                header,
 		BlockHashToBeRevealed: blockHashToBeRevealed,
-	}, state, h.bcReader.Network(), h.callMaxSteps, sierraVersion, false, false)
+	}, state,
+		h.callMaxSteps,
+		sierraVersion,
+		false,
+		false,
+	)
```
rodrodros, line 308: 👍

vm/vm.go
```diff
@@ -168,6 +185,28 @@ type CallInfo struct {
 	Calldata        []felt.Felt
 }
 
+type ChainContext struct {
+	ChainID             string
+	EthFeeTokenAddress  *felt.Felt
+	StrkFeeTokenAddress *felt.Felt
+}
+
+func NewChainContext(chainID string, ethFeeTokenAddress, strkFeeTokenAddress *felt.Felt) *ChainContext {
```
rodrodros: I don't think having a constructor with optional parameters is good for readability. It seems you need several constructors. I cannot make a tangible suggestion since I am unaware of how the custom integration looks like

core/contract_addresses.go
EgeCaner: Ahh this slipped out to this PR, should not be here, its from experimental follow up PR

builder/executor.go
infrmtcs: Not related to this PR, but I think we should consider moving these bool flags into a struct, so they're named by struct field names and we can also use default value if we want. Another option is to use functional options, although it can be overkill in this case.

clients/feeder/feeder.go
```diff
@@ -375,3 +375,18 @@ func (c *Client) PreConfirmedBlock(ctx context.Context, blockNumber string) (*st
 	}
 	return preConfirmedBlock, nil
 }
+
+func (c *Client) FeeTokenAddresses(ctx context.Context) (*starknet.FeeTokenAddresses, error) {
```
rodrodros: Return by value please

builder/executor.go
rodrodros: Opened issue https://github.com/NethermindEth/juno/issues/3127

node/node.go
```diff
@@ -199,11 +197,22 @@ func New(cfg *Config, version string, logLevel *utils.LogLevel) (*Node, error) {
 		services = append(services, plugin.NewService(junoPlugin))
 	}
 
+	var nodeVM vm.VM
+	var throttledVM *ThrottledVM
+
 	if cfg.Sequencer {
+		// Sequencer mode only supports known networks and uses default fee tokens
+		if !slices.Contains(utils.KnownNetworkNames(), cfg.Network.Name) {
+			return nil, fmt.Errorf("custom networks not supported in sequencer mode")
```
rodrodros: I would add the "yet" at the end: "not supported yet". Rationale is that Juno first time as a sequencer will probably start from a custom chain.

starknet/fee_token_addresses.go
```diff
@@ -0,0 +1,11 @@
+package starknet
+
+import "github.com/NethermindEth/juno/core/felt"
+
+// FeeTokenAddresses holds the L2 token contract addresses for STRK and ETH as returned by the
+// feeder gateway's "get_contract_addresses" endpoint. These addresses are used to
+// identify the respective fee tokens on the network.
+type FeeTokenAddresses struct {
+	StrkL2TokenAddress *felt.Felt `json:"strk_l2_token_address"`
+	EthL2TokenAddress  *felt.Felt `json:"eth_l2_token_address"`
```
rodrodros: No need for these to be reference type since they are not optional. WDYT?

utils/network.go
```diff
@@ -167,7 +167,7 @@ func (n *Network) L2ChainIDFelt() *felt.Felt {
 	return new(felt.Felt).SetBytes([]byte(n.L2ChainID))
 }
 
-func knownNetworkNames() []string {
+func KnownNetworkNames() []string {
```
infrmtcs: nit: not related to this PR, but this should be a var instead

vm/vm.go
```diff
@@ -168,6 +187,60 @@ type CallInfo struct {
...
+		return FeeTokenAddresses{}, err
+	}
+
+	return FeeTokenAddresses{
+		EthFeeTokenAddress:  feeTokenAddresses.EthL2TokenAddress,
+		StrkFeeTokenAddress: feeTokenAddresses.StrkL2TokenAddress,
+	}, nil
+}
+
+func DefaultFeeTokenAddresses() FeeTokenAddresses {
```
infrmtcs: Can we convert this to var? If yes, I think we can also replace the [4]uint64 by a function call to the hex numbers, as it's only computed once.

vm/vm.go
```diff
@@ -168,6 +187,60 @@ type CallInfo struct {
 	Calldata        []felt.Felt
 }
 
+type FeeTokenAddresses struct {
+	EthFeeTokenAddress  felt.Felt
+	StrkFeeTokenAddress felt.Felt
+}
+
+func FeeTokenAddressesFromGateway(feederGateway *feeder.Client) (FeeTokenAddresses, error) {
```
infrmtcs: This should take a context instead.

infrmtcs: Besides that, can we cache it during node startup? I'm wondering if it ever changes.

vm/vm.go
```diff
@@ -168,6 +187,60 @@ type CallInfo struct {
 	Calldata        []felt.Felt
 }
 
+type FeeTokenAddresses struct {
+	EthFeeTokenAddress  felt.Felt
+	StrkFeeTokenAddress felt.Felt
+}
```
infrmtcs: A bit opinionated here, but I think we should:
1. Either use the `starknet.FeeTokenAddresses` directly or write a small adapter to convert between 2 types. 
2. Move `FeeTokenAddressesFromGateway` the calculation to another place, because `vm` shouldn't call feeder gateway.

vm/rust/src/lib.rs
```diff
@@ -779,24 +780,21 @@ fn build_block_context(
...
-            )
-            .unwrap(),
-            strk_fee_token_address: ContractAddress::try_from(
-                StarkHash::from_hex(
-                    "0x04718f5a0fc34cc1af16a1cdee98ffb20c31f5cd61d6ab07201858f4287c938d",
-                )
-                .unwrap(),
-            )
-            .unwrap(),
+            eth_fee_token_address: ContractAddress(
```
infrmtcs: We can use `ContractAddress::try_from` instead.

infrmtcs (commented):
The rest LGTM

vm/vm.go
EgeCaner: We are only making a call at startup for custom networks

vm/vm.go
```diff
@@ -167,6 +185,17 @@ type CallInfo struct {
 	Selector        *felt.Felt
 	Calldata        []felt.Felt
 }
+type ChainInfo struct {
```
infrmtcs, line 189: nit: blank line above

utils/network.go
```diff
@@ -7,12 +7,41 @@ import (
...
 
+var KnownNetworkNames = []string{
+	Mainnet.String(),
+	Sepolia.String(),
+	SepoliaIntegration.String(),
+	Sequencer.String(),
+}
+
+// 0x049d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7
+var DefaultEthFeeTokenAddress = felt.Felt([4]uint64{
```
infrmtcs, line 23: Can we do something like 
```suggestion
var DefaultEthFeeTokenAddress = felt.FromUint64(0x049d36570d4e46f48e99674bd3fcc84644ddd6b96f7c741b1562b82f9e004dc7)
```
so we don't need the comment above

vm/vm.go
```diff
@@ -167,6 +185,17 @@ type CallInfo struct {
...
+	ChainID           string
+	FeeTokenAddresses starknet.FeeTokenAddresses
+}
+
+func NewChainInfo(chainID string, feeTokenAddresses *starknet.FeeTokenAddresses) *ChainInfo {
+	return &ChainInfo{
+		ChainID:           chainID,
+		FeeTokenAddresses: *feeTokenAddresses,
+	}
+}
```
infrmtcs: nit: I think this is the same as the default constructor, so we can either remove this, or make all fields private.

infrmtcs (approved):
The rest LGTM, thanks!

utils/network.go
EgeCaner: We have `SetString` but it also returns an error

vm/vm.go
EgeCaner: I will remove

utils/network.go
rodrodros: We can have a create string that never errors 🤔. We can solve it after the type PR gets merged

rodrigo-pino (approved):

PR #3128, opened 2025-09-20
fix: add missing number check in shouldPreservePending

EgeCaner:


infrmtcs (approved):

sync/pending_polling.go
```diff
@@ -55,6 +55,10 @@ func shouldPreservePendingData(
 		return false
 	}
 
+	if existingPending.GetBlock().Number != incomingPending.GetBlock().Number {
+		return false
+	}
```
rodrodros: Do we have to accept any block number regardless of transaction. It makes sense to me in the case of higher number, but in the case of a lower number, should we do it as well?

EgeCaner: We should accept lower block numbers to support reorgs, other option is when restarting sync on reorg, we can assign pending data to nil. But following will work too, I will add this line to `HandlePreConfirmed`. 
```go
	if pc.Block.Number != targetPreConfirmedNum.Load() {
		return
	}
```

This will ensure we dont overwrite new values with old ones unless there is a reorg. `targetPreConfirmedNum` is in scope of `runPreConfirmedPhase` and when there is reorg we restart the goroutine thus `targetPreConfirmedNum` will be reinitialize.

EgeCaner: Reconsidering number for reorg, I guess we can enforce increasing numbers here, since `pending.Validate` will fail when reorg.

sync/pending_polling.go
```diff
@@ -55,11 +55,17 @@ func shouldPreservePendingData(
...
+	if incomingB.Number > existingB.Number {
 		return false
 	}
 
-	return true
+	if incomingB.Number < existingB.Number {
+		return true
+	}
+
+	return existingB.TransactionCount >= incomingB.TransactionCount
```
rodrodros: What do you think of  removing the ifs and just returning the following. Just a nitpick though.
```go
return (incomingB.Number < existingB.Number) || 
  (incomingB.Number == existingB.Number && incomingB.TransactionCount <= existingB.TransactionCount)
```

On a more serious question, what happens during a re-org, does this mean we will never accept an older pending during a re-org?

EgeCaner: Incase of reorg, existingpPending.Validate(head) should fail and we should not fall into block number logic. 

Following condition is to avoid pre_confirmed HEAD +1 to overwrite HEAD + 2 since both are valid for head
```go
	if incomingB.Number < existingB.Number {
		return true
	}
```

PR #3132, opened 2025-09-23
Handled panic case in p2p2core

ndatta-nethermind:


adapters/p2p2core/class.go
```diff
@@ -14,13 +15,13 @@ import (
...
 
-func AdaptCairo1Class(cairo1 *class.Cairo1Class) core.Cairo1Class {
+func AdaptCairo1Class(cairo1 *class.Cairo1Class) (core.Cairo1Class, error) {
 	abiHash := crypto.StarknetKeccak([]byte(cairo1.Abi))
 
 	program := utils.Map(cairo1.Program, AdaptFelt)
 	compiled, err := createCompiledClass(cairo1)
 	if err != nil {
-		panic(err)
+		return core.Cairo1Class{}, errors.New("Invalid format data from")
```
infrmtcs: We should wrap the error by using `fmt.Errorf` and `%w`, you can check this for more info: https://go.dev/blog/go1.13-errors#wrapping-errors-with-w

adapters/p2p2core/class.go
```diff
@@ -67,10 +68,10 @@ func AdaptClass(cls *class.Class) core.Class {
...
 			Program:      cairo0.Program,
-		}
+		}, nil
 	case *class.Class_Cairo1:
-		cairoClass := AdaptCairo1Class(cls.Cairo1)
-		return &cairoClass
+		cairoClass, err := AdaptCairo1Class(cls.Cairo1)
+		return &cairoClass, err
 	default:
 		panic(fmt.Errorf("unsupported class %T", cls))
```
infrmtcs: Remove `panic` here

p2p/sync/sync.go
```diff
@@ -336,7 +336,12 @@ func (s *Service) adaptAndSanityCheckBlock(
 
 			newClasses := make(map[felt.Felt]core.Class)
 			for _, cls := range classes {
-				coreC := p2p2core.AdaptClass(cls)
+				coreC, err := p2p2core.AdaptClass(cls)
+				if err != nil {
+					bodyCh <- BlockBody{Err: fmt.Errorf("%v", err)}
```
infrmtcs: We should either remove `fmt.Errorf` and use `err` directly, or wrap it with extra message and use `%w`.

utils/slices.go
```diff
@@ -21,6 +21,23 @@ func Map[T1, T2 any](slice []T1, f func(T1) T2) []T2 {
 	return result
 }
 
+func MapWithErrors[T1, T2 any](slice []T1, f func(T1) (T2, error)) []T2 {
```
infrmtcs: I think the code in this file is legacy. If we want to create this function, we should return `iter.Seq2` instead. You can see more details here: https://antonz.org/go-1-23/#iterators

adapters/p2p2core/class.go
infrmtcs: And I think the message can be something more specific, like "Invalid compiled class"

adapters/p2p2core/class.go
```diff
@@ -14,13 +14,13 @@ import (
...
 
-func AdaptCairo1Class(cairo1 *class.Cairo1Class) core.Cairo1Class {
+func AdaptCairo1Class(cairo1 *class.Cairo1Class) (core.Cairo1Class, error) {
 	abiHash := crypto.StarknetKeccak([]byte(cairo1.Abi))
 
 	program := utils.Map(cairo1.Program, AdaptFelt)
 	compiled, err := createCompiledClass(cairo1)
 	if err != nil {
-		panic(err)
+		return core.Cairo1Class{}, fmt.Errorf("Invalid format data from: %w", err)
```
infrmtcs: Please also check this comment: https://github.com/NethermindEth/juno/pull/3132#discussion_r2371608009

adapters/p2p2core/class.go
```diff
@@ -67,12 +67,12 @@ func AdaptClass(cls *class.Class) core.Class {
...
-		}
+		}, nil
 	case *class.Class_Cairo1:
-		cairoClass := AdaptCairo1Class(cls.Cairo1)
-		return &cairoClass
+		cairoClass, err := AdaptCairo1Class(cls.Cairo1)
+		return &cairoClass, err
 	default:
-		panic(fmt.Errorf("unsupported class %T", cls))
+		return nil, fmt.Errorf("format string %T", cls)
```
infrmtcs: Please keep the original error message

utils/slices.go
```diff
@@ -21,6 +22,17 @@ func Map[T1, T2 any](slice []T1, f func(T1) T2) []T2 {
 	return result
 }
 
+func MapWithErrors[T1, T2 any](slice []T1, f func(T1) (T2, error)) iter.Seq2[T2, error] {
```
infrmtcs: Please add some tests for this function

infrmtcs: If the `T1` is big, calling `f(e)` will copy e. We should use reference instead.

p2p/sync/sync.go
```diff
@@ -364,7 +369,11 @@ func (s *Service) adaptAndSanityCheckBlock(
 				}
 			}()
 
-			stateDiff := p2p2core.AdaptStateDiff(stateReader, contractDiffs, classes)
+			stateDiff, err := p2p2core.AdaptStateDiff(stateReader, contractDiffs, classes)
+			if err != nil {
+				bodyCh <- BlockBody{Err: fmt.Errorf("%w", err)}
```
infrmtcs: Please add extra message here, otherwise use `err` instead of `fmt.Errorf("%w", err)`

p2p/sync/sync.go
```diff
@@ -336,10 +336,15 @@ func (s *Service) adaptAndSanityCheckBlock(
 
 			newClasses := make(map[felt.Felt]core.Class)
 			for _, cls := range classes {
-				coreC := p2p2core.AdaptClass(cls)
+				coreC, err := p2p2core.AdaptClass(cls)
+				if err != nil {
+					bodyCh <- BlockBody{Err: fmt.Errorf("%w", err)}
```
infrmtcs: Please add extra message here, otherwise use `err` instead of `fmt.Errorf("%w", err)`

infrmtcs: In this case, we should add extra message.

p2p/sync/sync.go
infrmtcs: In this case, we should add extra message.

adapters/p2p2core/state.go
```diff
@@ -64,7 +67,7 @@ func AdaptStateDiff(reader core.StateReader, contractDiffs []*state.ContractDiff
 					if errors.Is(err, db.ErrKeyNotFound) {
 						stateClassHash = &felt.Zero
 					} else {
-						panic(err)
+						return nil, fmt.Errorf("%w", err)
```
infrmtcs: Please add extra message here

adapters/p2p2core/state.go
```diff
@@ -13,16 +13,19 @@ import (
...
 
-	for _, class := range utils.Map(classes, AdaptClass) {
+	for class, err := range utils.MapWithErrors(classes, AdaptClass) {
+		if err != nil {
+			return nil, fmt.Errorf("%w", err)
+		}
 		h, err := class.Hash()
 		if err != nil {
-			panic(fmt.Errorf("unexpected error: %v when calculating class hash", err))
+			return nil, fmt.Errorf("%w", err)
```
infrmtcs: Using %w is good, but please keep the original message.

adapters/p2p2core/state.go
```diff
@@ -13,16 +13,19 @@ import (
 	"github.com/starknet-io/starknet-p2pspecs/p2p/proto/sync/state"
 )
 
-func AdaptStateDiff(reader core.StateReader, contractDiffs []*state.ContractDiff, classes []*class.Class) *core.StateDiff {
+func AdaptStateDiff(reader core.StateReader, contractDiffs []*state.ContractDiff, classes []*class.Class) (*core.StateDiff, error) {
```
rodrodros: You have a linter error, the proposed formatting solution is either:
```golang
func AdaptStateDiff(
   <list of input args>
) (<list of output args>
```
or 
```golang
func AdaptStateDif(
   <arg1>
   <arg2>
   ...
   <argn>
) (<list of output args>)
```

utils/slices.go
rodrodros: @ndatta-nethermind go allows arguments to be passed by reference or by value. By copying by value `T`, you copy the exact bytes in memory of `T`, but by copying by reference, you only copy a pointer that points to those bytes in memory.

For example. This is a copy by value from `a` to `b`
```golang
a := 55
b := a

fmt.Println(a) // prints 55
fmt.Println(b) // prints 55

b = b +1

fmt.PrintLn(a) // prints 55
fmt.PrintLn(b) // prints 56 (cause b is in a different place in memory)
```

And this is a copy by reference from `a` to `b`:
```golang
a := 55
b := &a

fmt.Println(a) // prints 55
fmt.Println(*b) // prints 55

*b = *b + 1

fmt.Println(a) // prints 56
fmt.Println(*b) // prints 56
```

As you can see, the difference between the examples is that in the first one, `b` is of type `int` while in the second one, `b` is of type `*int` (which means a pointer to an integer).

rodrodros: You can test quick things in Go with https://goplay.tools/

ndatta-nethermind: Got it. But, the problem in `MapWithErrors` method is that, the caller method passes an array of pointers here. So, each element is a pointer and thus the argument pass in the function call `f(e)` is also a reference. I did implement reference based `MapWithErrors`, but then the caller method fails invoke this method. We can discuss in details in a call or slack channel.

adapters/p2p2core/state.go
ndatta-nethermind: I use `make format` to format the `go` and `rust` code. Is there any such tool that will lint automatically we use?

rodrodros: Regarding long lines the go formatter makes no decision and leaves it up to the programmer. It has to be done manually

utils/slices.go
rodrodros: The caller passes an array of elements with type `T`.  While the slice (go idiom for array) is a reference type, the elements it holds are not.

An slice of `T` is:
```
x := make([]T, len)
```

While a slice of pointers to `T` or a slice of `*T` is:
```
x := make([]*T, len)
```

Why does the caller fail to invoke the method? Which error does it return?

ndatta-nethermind: This is the error I get:

`in call to utils.MapWithErrors, type func(cls *class.Class) (core.Class, error) of AdaptClass does not match inferred type func(**class.Class) (T2, error) for func(*T1) (T2, error)compiler[CannotInferTypeArgs](https://pkg.go.dev/golang.org/x/tools/internal/typesinternal#CannotInferTypeArgs)
func AdaptClass(cls *class.Class) (core.Class, error)
[p2p2core.AdaptClass on pkg.go.dev](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html)`

I suspect it because, the  `classes` variable in `utils.MapWithErrors(classes, AdaptClass)` is of type `[]*class.Class`. When we execute `res, err := f(&e)` in 


```
func MapWithErrors[T1, T2 any](slice []T1, f func(*T1) (T2, error)) iter.Seq2[T2, error] {
	return func(yield func(T2, error) bool) {
		for _, e := range slice {
			res, err := f(&e)
			if !yield(res, err) {
				return
			}
		}
	}
}
```


AdaptClass method get `**calss.Class` type when it's expecting `*class.Class` type.

infrmtcs: In this case, the original version is actually good enough, because `T1` is a pointer, which is small.

utils/slices_test.go
```diff
@@ -22,6 +24,66 @@ func TestMap(t *testing.T) {
...
+		var input []int
+		seq := MapWithErrors(input, func(i int) (string, error) {
+			return strconv.Itoa(i), nil
+		})
+
+		count := 0
+		seq(func(_ string, _ error) bool {
+			count++
+			return true
+		})
```
infrmtcs: Use `for` instead. The purpose of `iter.Seq` and `iter.Seq2` is for for-range loop, so we should also use for-range loop in tests.

utils/slices.go
infrmtcs: Although I think that in this specific case, why don't we just for range over `classes` and call `AdaptClass` instead.

adapters/p2p2core/state.go
```diff
@@ -13,16 +13,19 @@ import (
...
 
-	for _, class := range utils.Map(classes, AdaptClass) {
+	for class, err := range utils.MapWithErrors(classes, AdaptClass) {
+		if err != nil {
+			return nil, fmt.Errorf("unsupported class: %w", err)
+		}
 		h, err := class.Hash()
 		if err != nil {
-			panic(fmt.Errorf("unexpected error: %v when calculating class hash", err))
+			return nil, fmt.Errorf("unexpected error: %w when calculating class hash", err)
```
infrmtcs: I understand it's legacy, but let's do this instead:

```suggestion
			return nil, fmt.Errorf("unexpected error when calculating class hash: %w", err)
```

p2p/sync/sync.go
```diff
@@ -364,7 +369,11 @@ func (s *Service) adaptAndSanityCheckBlock(
 				}
 			}()
 
-			stateDiff := p2p2core.AdaptStateDiff(stateReader, contractDiffs, classes)
+			stateDiff, err := p2p2core.AdaptStateDiff(stateReader, contractDiffs, classes)
+			if err != nil {
+				bodyCh <- BlockBody{Err: fmt.Errorf("unsupported class: %w", err)}
```
infrmtcs:
```suggestion
				bodyCh <- BlockBody{Err: fmt.Errorf("failed to adapt state diff: %w", err)}
```

p2p/sync/sync.go
```diff
@@ -336,10 +336,15 @@ func (s *Service) adaptAndSanityCheckBlock(
 
 			newClasses := make(map[felt.Felt]core.Class)
 			for _, cls := range classes {
-				coreC := p2p2core.AdaptClass(cls)
+				coreC, err := p2p2core.AdaptClass(cls)
+				if err != nil {
+					bodyCh <- BlockBody{Err: fmt.Errorf("unsupported class: %w", err)}
```
infrmtcs:
```suggestion
					bodyCh <- BlockBody{Err: fmt.Errorf("failed to adapt class: %w", err)}
```
because it's not necessary unsupported class, but can be wrong format or something else.

utils/slices_test.go
ndatta-nethermind: Removed this section of the code.

PR #3137, opened 2025-09-25
chore: Use new felt functions in consensus

infrmtcs:


consensus/driver/driver_test.go
```diff
@@ -48,29 +48,29 @@ func getRandMessageHeader(random *rand.Rand) starknet.MessageHeader {
...
-		ID:            utils.HeapPtr(felt.Hash(felt.FromUint64[felt.Felt](random.Uint64()))),
+		ID:            felt.NewFromUint64[starknet.Hash](random.Uint64()),
 	}
 }
 
 func getRandPrecommit(random *rand.Rand) starknet.Precommit {
 	return starknet.Precommit{
 		MessageHeader: getRandMessageHeader(random),
-		ID:            utils.HeapPtr(felt.Hash(felt.FromUint64[felt.Felt](random.Uint64()))),
+		ID:            felt.NewFromUint64[starknet.Hash](random.Uint64()),
```
rodrodros, line 73: I think we can add a new constructor `felt.UnsafeNewRandom[FeltLike]()`. What do you think?

consensus/proposal/proposal_store_test.go
```diff
@@ -58,17 +52,17 @@ func TestProposalStore_StoreAndGet(t *testing.T) {
...
 		{
 			name:  "store and get multiple values",
-			key:   createTestHash(2),
+			key:   felt.FromUint64[starknet.Hash](2),
 			value: createTestBuildResult(),
 		},
 		{
 			name:  "store with zero hash key",
-			key:   starknet.Hash{},
+			key:   felt.FromUint64[starknet.Hash](0),
```
rodrodros, line 65: What do you think of adding a default constructor `felt.Default[FeltLike]()` which returns by value?

rodrigo-pino (approved):

consensus/proposal/proposal_store_test.go
infrmtcs: In terms of constructors, I think less is more. I think it's clearer to explicitly see 0 as an argument of the constructor, instead of guessing what the default is (0, 1, mod, etc.)

consensus/driver/driver_test.go
infrmtcs: Yeah, I think it's a good idea. I checked the implementation, `err` can only happens when we cannot read more bytes from `rand.Reader`, in that case we can't do anything but panic anyway.

PR #3142, opened 2025-09-26
feat: add more ctors for felt-like types (and remove unnecessary utilities)

rodrigo-pino:


core/felt/ctors.go
```diff
@@ -72,3 +79,18 @@ func UnsafeFromString[F FeltLike](value string) F {
...
+
+// Creates a new random Felt based type, errors if `rand.Reader` errors
+func Random[F FeltLike]() (F, error) {
+	f, err := new(Felt).SetRandom()
+	return F(*f), err
+}
+
+// Creates a new random Felt based type, panics if `rand.Reader` errors
+func UnsafeRandom[F FeltLike]() F {
+	f, err := new(Felt).SetRandom()
```
EgeCaner: Can this randomly error? If so, we might want to avoid this function due to its non-deterministic behavior.
Typically, the caller must ensure that an `Unsafe` function executes correctly.
For example, with `UnsafeFromString`, we ensure provided strings fit within a single felt.
However, for `UnsafeRandom`, we have no such control, which makes it risky since errors could occur unpredictably.

rodrodros: I agree, I was thinking the use case of this one targeted at test where the value.

Maybe we can add the parameter t, so it is clear it is meant only for testing 🤔

EgeCaner: [Docs](https://pkg.go.dev/crypto/rand#Read) says rand.Reader never returns error. Error might be there to comply with the interface.

rodrodros: Since we don't support old Linux, I will remove UnsafeRandom, it doesn't makes sense

rodrodros: Great catch!!!

PR #3144, opened 2025-09-30
feat: pending data lookup methods

EgeCaner:


core/pending.go
```diff
@@ -89,6 +98,29 @@ func (p *Pending) Validate(parent *Header) bool {
...
+		}
+	}
+
+	return nil, ErrTransactionNotFound
+}
+
+func (p *Pending) ReceiptByHash(
+	hash *felt.Felt,
+) (*TransactionReceipt, *felt.Felt, uint64, error) {
+	pendingBlock := p.GetBlock()
```
infrmtcs: We can use `p.Block` directly here

core/pending.go
```diff
@@ -89,6 +98,29 @@ func (p *Pending) Validate(parent *Header) bool {
...
+	for _, receipt := range pendingBlock.Receipts {
+		if receipt.TransactionHash.Equal(hash) {
+			return receipt, pendingBlock.ParentHash, pendingBlock.Number, nil
+		}
+	}
```
infrmtcs:
```suggestion
	for _, receipt := range p.Block.Receipts {
		if receipt.TransactionHash.Equal(hash) {
			return receipt, p.Block.ParentHash, p.Block.Number, nil
		}
	}
```

core/pending.go
rodrodros: Personally, the existence of a `p.GetBlock` and also a `Block` property makes it a bit confusing to me. This is more or less a nitpick, but generally, I think we shouldn't have `GetSmth` methods but instead just `Smth` method. In this case, I think we should have just a `Block` method for the interface and have the property private. 

There is no performance loss since this will be inlined by the compiler.  

Also, out of the scope of this PR, but just bringing these thoughts out

core/pending.go
```diff
@@ -185,3 +217,48 @@ func (p *PreConfirmed) Validate(parent *Header) bool {
...
+	if preLatest := p.GetPreLatest(); preLatest != nil {
+		for _, receipt := range preLatest.Block.Receipts {
+			if receipt.TransactionHash.Equal(hash) {
+				return receipt, preLatest.Block.Header.ParentHash, preLatest.Block.Number, nil
+			}
+		}
+	}
+
+	preConfirmedBlock := p.GetBlock()
+	for _, receipt := range preConfirmedBlock.Receipts {
```
rodrodros: In a similar fashion to Dat's comment above, `p.Block` can be used directly here

sync/sync.go
```diff
@@ -600,6 +600,14 @@ func (s *Synchronizer) PendingData() (core.PendingData, error) {
 
 	p := *ptr
 	if p.Validate(head) {
+		// If pending data is PreConfirmed and it's latest + 1, return it without pre_latest
+		if p.Variant() == core.PreConfirmedBlockVariant && head != nil {
+			preConfirmed := p.(*core.PreConfirmed)
+			if preConfirmed.Block.Number == head.Number+1 && preConfirmed.PreLatest != nil {
```
rodrodros: I am not sure the code should be updated, but I am sharing this second way of doing this in case you didn't know already:
```golang
if preConfirmed, ok := p(*core.Preconfirmed); ok && head != nil. {
    if preconfirmed.Block.Number == ... {
       // smth smth
   }
}
```

I think we can discuss which way is better, but I find it as legible, and a quick question to ChatGPT says that the type assertion is faster.

rodrodros: Can you add the meaning of what this `if` condition represents, either as a comment or a variable? 

Is this the case when pending has pre-latest attached, but pre-latest is already latest?

rodrigo-pino (approved):

core/pending.go
EgeCaner: I agree that this getter is not very idiomatic Go. I also believe the fields should be private; currently, they’re public mainly to allow direct writes where setter methods are not used(Dont have explicit setter rn). Let me review the code to see if we can make these fields private with minimal changes.

PR #3145, opened 2025-09-30
feat: prelatest tx status and receipt

EgeCaner:
Integrates `PreLatest` block in following queries:
- `starknet_getTransactionStatus`
- `starknet_getMessagesStatus`  Utilizes same mechanism with get transaction status 
- `starknet_getTransactionByHash`
- `starknet_getTransactionReceipt`

Indirectlly supports transaction status subscription since it also uses same transaction status mechanism, but wont be triggered on prelatest yet, support for prelatest in subscription will be included in follow up PR

Depends on #3144


rodrigo-pino (approved):

rpc/v9/transaction.go
```diff
@@ -552,62 +540,43 @@ func (h *Handler) TransactionByBlockIDAndIndex(
...
+			if err == nil {
+				status := TxnPreConfirmed
+				if parentHash != nil {
+					// preLatest block
+					status = TxnAcceptedOnL2
+				}
+				return AdaptReceipt(receipt, txn, status, nil, blockNumber), nil
+			}
+		}
+	}
```
infrmtcs: I'm wondering if we can move this logic into a separate function, because the condition is quite nested. Something like a `extractPendingReceipt` function and inside that, we return nil, err everytime err != nil.

PR #3146, opened 2025-10-01
feat: Integrate Sync P2P protocol to consensus

infrmtcs:
Depends #3172


consensus/driver/driver.go
```diff
@@ -124,13 +139,26 @@ func (d *Driver[V, H, A]) listen(ctx context.Context) error {
...
+				}
 			}
 
 			isCommitted, err = d.execute(ctx, false, actions)
 			if err != nil {
 				return err
 			}
 		}
+
+		d.syncCurrentHeight(ctx)
```
rodrodros, line 161: Is the idea to trigger sync (or more check if we are out of sync, and sync if we are) after every message. Why should we get out of sync while tracking consensus?

infrmtcs: Yes, the idea is "check if we are out of sync, and sync if we are". We can be out of sync when network disruption happens, e.g disconnected from internet for 1 minute.

PR #3147, opened 2025-10-01
refactor: pending state access

EgeCaner:
# Key Changes

## Core Package Changes
- Enhanced `PendingData` interface: Added `PendingState()` and `PendingStateBeforeIndex()` methods.
- Moved pending state logic: Relocated from `sync/pending.go` to `core/pending_state.go`.

## Sync Package Changes
- New helper functions: Created `sync/pending_state_helpers.go` with:
  - `ResolvePendingDataBaseState()` – Determines appropriate base state for pending data.
  - `PendingState()` – Convenience function combining base state resolution with pending state creation.
  - `GetPendingStateBeforeIndexWithBaseState()` – Handles pending state before index creation.
- Refactored `sync.go`: Uses new helper functions for cleaner separation of concerns.
- Removed `PendingStateBeforeIndex()` from `sync.Reader`.  
  (Also planning to remove `PendingState()` as well. Now with `PendingData` and `blockchain.Reader`, pending state can be created anywhere.)

PR #3149, opened 2025-10-01
feat: trace single on prelatest

EgeCaner:
Depends on #3147 #3193

### DIFF
--- a/rpc/v9/trace_test.go
+++ b/rpc/v9/trace_test.go
@@ -583,9 +583,9 @@ func TestTraceTransaction(t *testing.T) {
 			Transactions: []core.Transaction{tx},
 		}
 
-		declaredClass := &core.DeclaredClass{
+		declaredClass := &core.DeclaredClassDefinition{
 			At:    3002,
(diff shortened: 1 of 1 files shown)


rpc/v9/trace_test.go
```diff
@@ -569,6 +569,119 @@ func TestTraceTransaction(t *testing.T) {
...
+		headState := mocks.NewMockStateHistoryReader(mockCtrl)
+		mockReader.EXPECT().StateAtBlockHash(preLatest.Block.ParentHash).
+			Return(headState, nopCloser, nil)
+
+		innerExecutionResources := `{
+			"l1_gas": 1,
+			"l2_gas": 2
+		}`
+
+		vmTraceJSON := fmt.Sprintf(`{
```
rodrodros: mmmmmmmm, what if we put this line of code directly in JSON, like test data for the RPC package 🤔 , also the source of it

rodrigo-pino (approved):
Loooks goood, one small comment

PR #3150, opened 2025-10-01
feat: prelatest in websocket

EgeCaner:
Prelatest block integration into websocket API. Explicity following methods:

- `starknet_subscribeTransactionStatus`
- `starknet_subscribeEvents`
- `starknet_subscribeNewTransactionReceipts`
- `starknet_subscribeNewTransactions`

Depends on #3145


rpc/v9/subscriptions.go
```diff
@@ -269,6 +277,22 @@ func (h *Handler) SubscribeEvents(
 				TxnAcceptedOnL2,
 			)
 		},
+		onPreLatest: func(
+			ctx context.Context,
+			id string, _ *subscription,
```
rodrodros: Let's have one parameter per line

rodrigo-pino (approved):

rpc/v9/subscriptions.go
```diff
@@ -517,6 +542,19 @@ func (h *Handler) SubscribeTransactionStatus(ctx context.Context, txHash *felt.F
...
+			id string,
+			sub *subscription,
+			preLatest *core.PreLatest,
+		) error {
+			if lastStatus < TxnStatusAcceptedOnL2 {
+				if lastStatus, err = h.checkTxStatus(ctx, sub, id, txHash, lastStatus); err != nil {
+					return err
+				}
+			}
+			return nil
```
infrmtcs: nit: Is it possible to create a function for this? I think it's the same as `onPendingData`

infrmtcs: And actually it's quite similar to `onL1Head`

rodrigo-pino:
Depends on #3179

PR #3152, opened 2025-10-03
feat: Implement tracker to synchronize WaitGroup Add() before Wait()

infrmtcs:


utils/tracker/tracker_test.go
```diff
@@ -0,0 +1,104 @@
...
+	runTests(t, func(ctx context.Context) waitGroup {
+		return NewSemaphoreWrapper(ctx)
+	})
+}
+
+func TestWaitGroupFailure(t *testing.T) {
+	t.Skip("This test is skipped because it's supposed to fail to demonstrate the issue")
+	runTests(t, func(context.Context) waitGroup {
+		return &waitGroupWrapper{}
+	})
```
rodrodros: Nit: if we went this length to test why we need a tracker, wouldn't it be possible to expect a panic when running these ones?

utils/tracker/wrapper_test.go
```diff
@@ -0,0 +1,43 @@
+package tracker_test
+
+import (
+	"context"
+	"sync"
+
+	"golang.org/x/sync/semaphore"
+)
+
+const highThreshold = 1000000
```
rodrodros: nit: rename to `semaphoreThreshold`

utils/tracker/service_test.go
```diff
@@ -0,0 +1,96 @@
...
+
+	// Simulate a small delay to create async boundary to increase the chances of race condition
+	time.Sleep(asyncBoundaryDelay)
+	return true
+}
+
+func (s *service) run(ctx context.Context) {
+	<-ctx.Done()
+	s.tracker.Wait()
+}
```
rodrodros, line 96: Don't you think it might be better to set `isShutdown` to true at the end of this function so the handling of the field is all done in the same file making it a bit easier to follow the test logic

utils/tracker/tracker.go
```diff
@@ -0,0 +1,37 @@
...
+)
+
+// Inspired by net/http
+// See https://cs.opensource.google/go/go/+/refs/tags/go1.25.1:src/net/http/server.go;l=3604
+type Tracker struct {
+	mu   sync.Mutex
+	wg   sync.WaitGroup
+	done bool
+}
+
```
rodrodros, line 25: Let's add an explanation of why Tracker needs to exist since it is not clear at all without context.

My suggestion, but feel free to change it or improve it:
```golang
// we need to use the tracker as a replacement for a `sync.WaitGroup` to properly
// handle the case when a Go routines starts after `Wait` was called. For example:
//
// Goroutine A (server shutdown)
// wg.Wait()  // waits for all work to finish
//
// Goroutine B (handling request) - might run AFTER Wait() starts
// wg.Add(1)  // PANIC or undefined behavior
//
// With a tracker `false` is returned instead and Goroutine B stops the tasks
```

utils/tracker/tracker_test.go
```diff
@@ -0,0 +1,104 @@
...
+	delayed int,
+	expected int,
+	tracker func(context.Context) waitGroup,
+) {
+	t.Helper()
+
+	ctx, cancel := context.WithCancel(t.Context())
+	isShutdown := atomic.Bool{}
+	success := atomic.Uint32{}
+	service := NewService(tracker(ctx), &isShutdown, nonDelayed, delayed)
```
rodrodros: `isShutdown` should then be completely internal to `service`

utils/tracker/tracker_test.go
infrmtcs: Done, probably in the past I didn't know the best way to assert panics.

utils/tracker/wrapper_test.go
infrmtcs: Updated

utils/tracker/service_test.go
infrmtcs: Agree, updated

utils/tracker/tracker.go
infrmtcs: Done

utils/tracker/tracker_test.go
infrmtcs: Agree, updated

PR #3153, opened 2025-10-05
refactor: return errors instead of panicking in p2p2core adapters

ndatta-nethermind:


adapters/p2p2consensus/transaction.go
```diff
@@ -29,9 +29,15 @@ func AdaptTransaction(t *p2pconsensus.ConsensusTransaction, network *utils.Netwo
...
-		tx = p2p2core.AdaptDeployAccountV3TxnCommon(t.GetDeployAccountV3(), t.TransactionHash)
+		tx, err = p2p2core.AdaptDeployAccountV3TxnCommon(t.GetDeployAccountV3(), t.TransactionHash)
+		if err != nil {
+			return consensus.Transaction{}, err
+		}
 	case *p2pconsensus.ConsensusTransaction_InvokeV3:
-		tx = p2p2core.AdaptInvokeV3TxnCommon(t.GetInvokeV3(), t.TransactionHash)
+		tx, err = p2p2core.AdaptInvokeV3TxnCommon(t.GetInvokeV3(), t.TransactionHash)
+		if err != nil {
+			return consensus.Transaction{}, nil
```
infrmtcs:
```suggestion
			return consensus.Transaction{}, err
```

adapters/p2p2core/transaction.go
```diff
@@ -34,14 +37,14 @@ func AdaptDeclareV3WithClass(
 	return declareCommon, &class, nil
 }
 
-func AdaptDeclareV3TxnCommon(tx *transaction.DeclareV3Common, classHash *felt.Felt, txnHash *common.Hash) *core.DeclareTransaction {
+func AdaptDeclareV3TxnCommon(tx *transaction.DeclareV3Common, classHash *felt.Felt, txnHash *common.Hash) (*core.DeclareTransaction, error) {
```
infrmtcs: Please fix the lint error

adapters/p2p2core/transaction.go
```diff
@@ -34,14 +37,14 @@ func AdaptDeclareV3WithClass(
...
 	nDAMode, err := adaptVolitionDomain(tx.NonceDataAvailabilityMode)
 	if err != nil {
-		panic(fmt.Sprintf("Failed to convert Nonce DA mode: %v to uint32", tx.NonceDataAvailabilityMode))
+		return nil, fmt.Errorf("failed to convert Nonce DA mode: %v to uint32", tx.NonceDataAvailabilityMode)
```
infrmtcs: Please wrap with the returned `err`

adapters/p2p2core/transaction.go
```diff
@@ -34,14 +37,14 @@ func AdaptDeclareV3WithClass(
...
 	}
 	fDAMode, err := adaptVolitionDomain(tx.FeeDataAvailabilityMode)
 	if err != nil {
-		panic(fmt.Sprintf("Failed to convert Fee DA mode: %v to uint32", tx.FeeDataAvailabilityMode))
+		return nil, fmt.Errorf("failed to convert Fee DA mode: %v to uint32", tx.FeeDataAvailabilityMode)
```
infrmtcs: Please wrap with the returned `err`

adapters/p2p2core/transaction.go
```diff
@@ -63,18 +66,18 @@ func AdaptDeclareV3TxnCommon(tx *transaction.DeclareV3Common, classHash *felt.Fe
...
-	return declareTx
+	return declareTx, nil
 }
 
-func AdaptDeployAccountV3TxnCommon(tx *transaction.DeployAccountV3, txnHash *common.Hash) *core.DeployAccountTransaction {
+func AdaptDeployAccountV3TxnCommon(tx *transaction.DeployAccountV3, txnHash *common.Hash) (*core.DeployAccountTransaction, error) {
 	nDAMode, err := adaptVolitionDomain(tx.NonceDataAvailabilityMode)
 	if err != nil {
-		panic(fmt.Sprintf("Failed to convert Nonce DA mode: %v to uint32", tx.NonceDataAvailabilityMode))
+		return nil, fmt.Errorf("failed to convert Nonce DA mode: %v to uint32", tx.NonceDataAvailabilityMode)
```
infrmtcs: Please wrap with the returned `err`

adapters/p2p2core/transaction.go
```diff
@@ -63,18 +66,18 @@ func AdaptDeclareV3TxnCommon(tx *transaction.DeclareV3Common, classHash *felt.Fe
...
 	}
 
 	fDAMode, err := adaptVolitionDomain(tx.FeeDataAvailabilityMode)
 	if err != nil {
-		panic(fmt.Sprintf("Failed to convert Fee DA mode: %v to uint32", tx.FeeDataAvailabilityMode))
+		return nil, fmt.Errorf("failed to convert Fee DA mode: %v to uint32", tx.FeeDataAvailabilityMode)
```
infrmtcs: Please wrap with the returned `err`

adapters/p2p2core/transaction.go
```diff
@@ -102,18 +105,18 @@ func AdaptDeployAccountV3TxnCommon(tx *transaction.DeployAccountV3, txnHash *com
...
-	}
+	}, nil
 }
 
-func AdaptInvokeV3TxnCommon(tx *transaction.InvokeV3, txnHash *common.Hash) *core.InvokeTransaction {
+func AdaptInvokeV3TxnCommon(tx *transaction.InvokeV3, txnHash *common.Hash) (*core.InvokeTransaction, error) {
 	nDAMode, err := adaptVolitionDomain(tx.NonceDataAvailabilityMode)
 	if err != nil {
-		panic(fmt.Sprintf("Failed to convert Nonce DA mode: %v to uint32", tx.NonceDataAvailabilityMode))
+		return nil, fmt.Errorf("failed to convert Nonce DA mode: %v to uint32", tx.NonceDataAvailabilityMode)
```
infrmtcs: Please wrap with the returned `err`

adapters/p2p2core/transaction.go
```diff
@@ -102,18 +105,18 @@ func AdaptDeployAccountV3TxnCommon(tx *transaction.DeployAccountV3, txnHash *com
...
 	}
 
 	fDAMode, err := adaptVolitionDomain(tx.FeeDataAvailabilityMode)
 	if err != nil {
-		panic(fmt.Sprintf("Failed to convert Fee DA mode: %v to uint32", tx.FeeDataAvailabilityMode))
+		return nil, fmt.Errorf("failed to convert Fee DA mode: %v to uint32", tx.FeeDataAvailabilityMode)
```
infrmtcs: Please wrap with the returned `err`

p2p/sync/sync.go
```diff
@@ -318,7 +318,10 @@ func (s *BlockFetcher) adaptAndSanityCheckBlock(
 
 			var coreTxs []core.Transaction
 			for _, tx := range txs {
-				coreTxs = append(coreTxs, p2p2core.AdaptTransaction(tx, s.network))
+				coreTx, err := p2p2core.AdaptTransaction(tx, s.network)
+				if err == nil {
+					coreTxs = append(coreTxs, coreTx)
+				}
```
infrmtcs: We should fail fast, because missing a transaction results in a completely different and invalid block.

adapters/p2p2mempool/transaction.go
```diff
@@ -42,9 +42,15 @@ func AdaptTransaction(t *mempooltransaction.MempoolTransaction, network *utils.N
...
+		tx, err = p2p2core.AdaptDeployAccountV3TxnCommon(t.GetDeployAccountV3(), t.TransactionHash)
+		if err != nil {
+			return mempool.BroadcastedTransaction{}, err
+		}
 	case *mempooltransaction.MempoolTransaction_InvokeV3:
-		tx = p2p2core.AdaptInvokeV3TxnCommon(t.GetInvokeV3(), t.TransactionHash)
+		tx, err = p2p2core.AdaptInvokeV3TxnCommon(t.GetInvokeV3(), t.TransactionHash)
+		if err != nil {
+			return mempool.BroadcastedTransaction{}, err
+		}
```
rodrodros: Since every case checks for an error now, I believe the error check can be written once outside the switch case statement

adapters/p2p2core/transaction.go
```diff
@@ -151,9 +182,12 @@ func AdaptL1Handler(tx *transaction.L1HandlerV0, txnHash *common.Hash) *core.L1H
...
 //nolint:funlen
-func AdaptTransaction(t *synctransaction.TransactionInBlock, network *utils.Network) core.Transaction {
+func AdaptTransaction(
+	t *synctransaction.TransactionInBlock,
+	network *utils.Network,
+) (core.Transaction, error) {
 	if t == nil {
-		return nil
+		return nil, nil
 	}
```
rodrodros: Let's remove this check. I don't think it is a good idea to silently do nothing when we receive a nil pointer in general. This affects the code in `sync.go` so it will need updating to handle this change in the condition.

p2p/sync/sync.go
```diff
@@ -318,7 +318,12 @@ func (s *BlockFetcher) adaptAndSanityCheckBlock(
 
 			var coreTxs []core.Transaction
```
rodrodros: I am aware it is part of old code, but please pre-allocate the `coreTxs` to have the same length as `txs`. This is more efficient since the runtime will be aware of the size of the transaction

adapters/p2p2core/transaction.go
```diff
@@ -151,9 +182,12 @@ func AdaptL1Handler(tx *transaction.L1HandlerV0, txnHash *common.Hash) *core.L1H
 }
 
 //nolint:funlen
-func AdaptTransaction(t *synctransaction.TransactionInBlock, network *utils.Network) core.Transaction {
+func AdaptTransaction(
+	t *synctransaction.TransactionInBlock,
+	network *utils.Network,
+) (core.Transaction, error) {
```
rodrodros, line 188: Let's break apart this function since it has gotten quite big. Let's have a specific call per case branch. I think each non-trivial execution branch should have its own adapter function defined. For example, "declare v0", "declare v1", and "declare v2" should have their own adapters; "declare v3" doesn't need it. 

Ideally an adapter for every execution branch whose code is not trivial

infrmtcs: I totally agree, but I think this should be outside of the scope of this PR to avoid the noise. We can open a separate PR to handle this.

rodrodros: Makes sense, let's make it a follow up PR 👌  cc @ndatta-nethermind

adapters/p2p2core/block.go
```diff
@@ -48,20 +56,20 @@ func AdaptBlockHeader(h *header.SignedBlockHeader, eventsBloom *bloom.BloomFilte
...
 	switch da {
 	case common.L1DataAvailabilityMode_Calldata:
-		return core.Calldata
+		return core.Calldata, nil
 	case common.L1DataAvailabilityMode_Blob:
-		return core.Blob
+		return core.Blob, nil
 	default:
-		panic(fmt.Errorf("unsupported DA mode %v", da))
+		return 0, fmt.Errorf("unsupported DA mode: %v", da)
```
infrmtcs, line 73:
```suggestion
		return 0, fmt.Errorf("unsupported DA mode: %w", da)
```

infrmtcs (commented):
The rest LGTM, please also address Rodrigo's comments, thanks!

adapters/p2p2core/transaction.go
ndatta-nethermind: If we remove this check, won't the program crash at the following statement?
```
switch t.Txn.(type)
```

adapters/p2p2core/transaction.go
ndatta-nethermind: Created this issue: https://github.com/NethermindEth/juno/issues/3160
Let's discuss tomorrow.

adapters/p2p2core/block.go
ndatta-nethermind: Isn't %w used for error unwrapping?

infrmtcs: My bad, %v is correct.

adapters/p2p2core/transaction.go
infrmtcs: I think what @rodrigo-pino means is in our convention, it's the responsibility of the caller to send the non-nil interface, so the statement failing means a programming error.

ndatta-nethermind: Removed the check.

rodrigo-pino:
@ndatta-nethermind the target is 60%, you can check this info in the failed CI checks or in the .github action workflow that covers codecov. 

Note that this is not a hard limit and it we can bypass it if it doesn't make sense. 

What we are doing now is changing the behaviour of a few functions to return an error instead of panicking. It would be nice to have tests that check this new failing behaviour

rodrigo-pino:
I've just realized that none of the adapters have tests 😱

PR #3155, opened 2025-10-06
feat: Customimze max gas for starknet_call

infrmtcs:


vm/rust/src/lib.rs
```diff
@@ -167,14 +166,6 @@ pub extern "C" fn cairoVMCall(
         }
     }
 
-    let version_constants = get_versioned_constants(block_info.version);
-    let sierra_version_str = unsafe { CStr::from_ptr(sierra_version) }.to_str().unwrap();
-    let sierra_version = SierraVersion::from_str(sierra_version_str).unwrap();
-    let initial_gas: u64 = if sierra_version < version_constants.min_sierra_version_for_sierra_gas {
```
infrmtcs, line 173: This check is already done by the lib, so we don't have to do this.

PR #3164, opened 2025-10-09
refactor(state): make state methods return felts by value

MaksymMalicki:


adapters/p2p2core/state.go
```diff
@@ -61,16 +61,16 @@ func AdaptStateDiff(
 				classHash: diff.ClassHash,
 			}
 
-			var stateClassHash *felt.Felt
+			var stateClassHash felt.Felt
 			if reader == nil {
 				// zero block
-				stateClassHash = &felt.Zero
+				stateClassHash = felt.Zero
```
infrmtcs: nit: Actually we don't need this, because it's already zero

adapters/p2p2core/state.go
```diff
@@ -61,16 +61,16 @@ func AdaptStateDiff(
...
 			} else {
 				var err error
 				stateClassHash, err = reader.ContractClassHash(address)
 				if err != nil {
 					if errors.Is(err, db.ErrKeyNotFound) {
-						stateClassHash = &felt.Zero
+						stateClassHash = felt.Zero
```
infrmtcs: nit: Actually we don't need this, because it's already zero

core/contract.go
```diff
@@ -163,21 +164,22 @@ func (c *ContractUpdater) UpdateStorage(diff map[felt.Felt]*felt.Felt, cb OnValu
...
-func ContractStorage(addr, key *felt.Felt, txn db.IndexedBatch) (*felt.Felt, error) {
+func ContractStorage(addr, key *felt.Felt, txn db.IndexedBatch) (felt.Felt, error) {
 	cStorage, err := storage(addr, txn)
 	if err != nil {
-		return nil, err
+		return felt.Felt{}, err
 	}
-	return cStorage.Get(key)
+	storage, err := cStorage.Get(key)
+	return *storage, err
```
infrmtcs: I think storage is likely to be `nil` in error case, so we should check first.

core/contract.go
```diff
@@ -163,21 +164,22 @@ func (c *ContractUpdater) UpdateStorage(diff map[felt.Felt]*felt.Felt, cb OnValu
...
 // ContractClassHash returns hash of the class that the contract at the given address instantiates.
-func ContractClassHash(addr *felt.Felt, txn db.IndexedBatch) (*felt.Felt, error) {
+func ContractClassHash(addr *felt.Felt, txn db.IndexedBatch) (felt.Felt, error) {
 	classHash, err := GetContractClassHash(txn, addr)
 	if err != nil {
-		return nil, err
+		return felt.Felt{}, err
 	}
-	return &classHash, nil // TODO: this should return a value
+	return classHash, nil
```
infrmtcs:
```suggestion
	return GetContractClassHash(txn, addr)
```

core/contract.go
```diff
@@ -130,12 +130,13 @@ func (c *ContractUpdater) UpdateNonce(nonce *felt.Felt) error {
...
-func ContractRoot(addr *felt.Felt, txn db.IndexedBatch) (*felt.Felt, error) {
+func ContractRoot(addr *felt.Felt, txn db.IndexedBatch) (felt.Felt, error) {
 	cStorage, err := storage(addr, txn)
 	if err != nil {
-		return nil, err
+		return felt.Felt{}, err
 	}
-	return cStorage.Root()
+	root, err := cStorage.Root()
+	return *root, err
```
infrmtcs, line 138: I think storage is likely to be `nil` in error case, so we should check first.

core/contract.go
```diff
@@ -115,12 +115,12 @@ func (c *ContractUpdater) Purge() error {
...
 // Only account contracts can have a non-zero nonce.
-func ContractNonce(addr *felt.Felt, txn db.IndexedBatch) (*felt.Felt, error) {
+func ContractNonce(addr *felt.Felt, txn db.IndexedBatch) (felt.Felt, error) {
 	nonce, err := GetContractNonce(txn, addr)
 	if err != nil {
-		return nil, err
+		return felt.Felt{}, err
 	}
-	return &nonce, nil // TODO: this should return a value
+	return nonce, nil
```
infrmtcs:
```suggestion
	return GetContractNonce(txn, addr)
```

core/history.go
```diff
@@ -93,14 +94,14 @@ func (h *history) DeleteContractNonceLog(contractAddress *felt.Felt, height uint
...
 	key := db.ContractNonceHistoryKey(contractAddress)
 	value, err := h.valueAt(key, height)
 	if err != nil {
-		return nil, err
+		return felt.Felt{}, err
 	}
-
-	return new(felt.Felt).SetBytes(value), nil
+	contractNonce := new(felt.Felt).SetBytes(value)
+	return *contractNonce, nil
```
infrmtcs:
```suggestion
	return felt.FromBytes[felt.Felt](value), nil
```

core/history.go
```diff
@@ -75,14 +75,15 @@ func (h *history) DeleteContractStorageLog(contractAddress, storageLocation *fel
...
 	key := db.ContractStorageHistoryKey(contractAddress, storageLocation)
 	value, err := h.valueAt(key, height)
 	if err != nil {
-		return nil, err
+		return felt.Felt{}, err
 	}
 
-	return new(felt.Felt).SetBytes(value), nil
+	contractStorage := new(felt.Felt).SetBytes(value)
+	return *contractStorage, nil
```
infrmtcs:
```suggestion
	return felt.FromBytes[felt.Felt](value), nil
```

core/history.go
```diff
@@ -111,12 +112,13 @@ func (h *history) DeleteContractClassHashLog(contractAddress *felt.Felt, height
...
 	key := db.ContractClassHashHistoryKey(contractAddress)
 	value, err := h.valueAt(key, height)
 	if err != nil {
-		return nil, err
+		return felt.Felt{}, err
 	}
 
-	return new(felt.Felt).SetBytes(value), nil
+	contractClassHash := new(felt.Felt).SetBytes(value)
+	return *contractClassHash, nil
```
infrmtcs:
```suggestion
	return felt.FromBytes[felt.Felt](value), nil
```

sync/pending.go
```diff
@@ -114,7 +114,7 @@ func (p *PendingStateWriter) IncrementNonce(contractAddress *felt.Felt) error {
 	if err != nil {
 		return fmt.Errorf("get contract nonce: %v", err)
 	}
-	p.stateDiff.Nonces[*contractAddress] = currentNonce.Add(currentNonce, feltOne)
+	p.stateDiff.Nonces[*contractAddress] = currentNonce.Add(&currentNonce, feltOne)
```
infrmtcs, line 117: I understand that this is not from your changes, but I think our team should double check the usages of mutable functions like `Add`, because it may mutate a shared variable which can cause unpredictable behaviours.

core/contract.go
MaksymMalicki: true, actually I think we can go even further and make the trie return by value immediately, but maybe this can be included in a separate PR for the trie

adapters/p2p2core/state.go
MaksymMalicki: addressed

adapters/p2p2core/state.go
MaksymMalicki: addressed

core/contract.go
MaksymMalicki: addressed

core/contract.go
MaksymMalicki: addressed, same here

core/contract.go
MaksymMalicki: addressed

core/history.go
MaksymMalicki: addressed

core/history.go
MaksymMalicki: addressed

sync/pending.go
MaksymMalicki: agreed

core/state_test.go
```diff
@@ -393,7 +393,7 @@ func TestRevert(t *testing.T) {
 		require.NoError(t, state.Revert(2, replaceStateUpdate))
 		classHash, sErr := state.ContractClassHash(new(felt.Felt).Set(&su1FirstDeployedAddress))
 		require.NoError(t, sErr)
-		assert.Equal(t, su1.StateDiff.DeployedContracts[*new(felt.Felt).Set(&su1FirstDeployedAddress)], classHash)
+		assert.Equal(t, su1.StateDiff.DeployedContracts[*new(felt.Felt).Set(&su1FirstDeployedAddress)], &classHash)
 	})
```
rodrodros: Old code, but I believe it can be updated directly to
```golang
	assert.Equal(t, su1.StateDiff.DeployedContracts[su1FirstDeployedAddress], &classHash)
```

sync/pending.go
infrmtcs: cc @rodrigo-pino What do you think?

PR #3168, opened 2025-10-10
chore: update docker env vars in CI

brbrr:


.github/workflows/docker-image-build-push.yml
```diff
@@ -100,7 +100,8 @@ jobs:
         run: docker system prune -af
 
   create_and_push_official_image:
-    needs: [setup, build_and_push_docker_image_amd, build_and_push_docker_image_arm]
+    needs:
+      [setup, build_and_push_docker_image_amd, build_and_push_docker_image_arm]
```
infrmtcs: nit: I think if we have to break lines anyway, then this is clearer:

```suggestion
      - setup
      - build_and_push_docker_image_amd
      - build_and_push_docker_image_arm
```

PR #3179, opened 2025-10-14
feat: prelatest in get events

EgeCaner:
Adds supports for fetching events from `pre-latest` block for `[Canonical, PreConfirmed]` queries.
[PreConfirmed, PreConfirmed] will only return Preconfirmed events, not includes prelatest.

### DIFF
--- a/blockchain/event_filter.go
+++ b/blockchain/event_filter.go
@@ -17,7 +17,7 @@ var errChunkSizeReached = errors.New("chunk size reached")
 type EventFilterer interface {
 	io.Closer
 
-	Events(cToken *ContinuationToken, chunkSize uint64) ([]*FilteredEvent, ContinuationToken, error)
+	Events(cToken *ContinuationToken, chunkSize uint64) ([]FilteredEvent, ContinuationToken, error)
 	SetRangeEndBlockByNumber(filterRange EventFilterRange, blockNumber uint64) error
 	SetRangeEndBlockByHash(filterRange EventFilterRange, blockHash *felt.Felt) error
 	SetRangeEndBlockToL1Head(filterRange EventFilterRange) error
@@ -140,8 +140,8 @@ type FilteredEvent struct {
 func (e *EventFilter) Events(
 	cToken *ContinuationToken,
 	chunkSize uint64,
-) ([]*FilteredEvent, ContinuationToken, error) {
-	var matchedEvents []*FilteredEvent
+) ([]FilteredEvent, ContinuationToken, error) {
+	var matchedEvents []FilteredEvent
 
 	latest, err := core.GetChainHeight(e.txn)
 	if err != nil {
@@ -193,12 +193,12 @@ func (e *EventFilter) Events(
 }
 
 func (e *EventFilter) canonicalEvents(
-	matchedEvents []*FilteredEvent,
+	matchedEvents []FilteredEvent,
 	fromBlock,
 	toBlock,
 	skippedEvents,
 	chunkSize uint64,
-) ([]*FilteredEvent, ContinuationToken, error) {
+) ([]FilteredEvent, ContinuationToken, error) {
 	matchedBlockIter, err := e.cachedFilters.NewMatchedBlockIterator(
 		fromBlock,
 		toBlock,
@@ -275,11 +275,11 @@ func (e *EventFilter) canonicalEvents(
 // Returns events from pre-latest block and pre-confirmed block
 // Support access to pre-confirmed block in isolation when fromBlock > preLatest.Block.Number
 func (e *EventFilter) pendingEvents(
-	matchedEvents []*FilteredEvent,
+	matchedEvents []FilteredEvent,
 	fromBlock,
 	skippedEvents,
 	chunkSize uint64,
-) ([]*FilteredEvent, ContinuationToken, error) {
+) ([]FilteredEvent, ContinuationToken, error) {
 	pendingData, err := e.pendingDataFn()
 	if err != nil {
 		if errors.Is(err, core.ErrPendingDataNotFound) {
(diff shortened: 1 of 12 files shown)


blockchain/event_filter.go
```diff
@@ -126,33 +133,46 @@ type FilteredEvent struct {
...
+			skippedEvents,
+			chunkSize,
+		)
 	}
 
-	var rToken *ContinuationToken
-	// [canonicalBlock, pending]
-	if curBlock <= latest {
-		matchedEvents, rToken, err = e.canonicalEvents(matchedEvents, curBlock, latest, skippedEvents, chunkSize)
+	// Case [canonicalBlock, pending]
```
rodrodros: Since pending is no longer a thing, I would change this to `[canonical, latest]`  (or `[canonical, pre-latest]`, or `[canonical, pre-confirmed]`). Not sure exactly if it is any of those.

blockchain/event_filter.go
```diff
@@ -163,30 +183,8 @@ func (e *EventFilter) Events(cToken *ContinuationToken, chunkSize uint64) ([]*Fi
 		// Skipped events are processed, so we can reset the counter
 		skippedEvents = 0
 	}
-
 	// Case [canonicalBlock, pending] || [pending, pending]
```
rodrodros: Similar comment here, since `pending` should no longer be a thing anymore. Unless, I am not understanding it right, and this `pending` refers to `pendingData` which is then fine.

P.D.: I am also aware that we still need to support pending, but ..., I think the concept should slowly fade out of the code

blockchain/event_filter.go
```diff
@@ -260,3 +265,107 @@ func (e *EventFilter) canonicalEvents(
...
+// pendingEvents processes pending events
+// Returns events from pre-latest block and pre-confirmed block
+// Support access to pre-confirmed block in isolation when fromBlock > preLatest.Block.Number
+func (e *EventFilter) pendingEvents(
+	matchedEvents []*FilteredEvent,
+	fromBlock, skippedEvents, chunkSize uint64,
+) ([]*FilteredEvent, *ContinuationToken, error) {
+	pendingData, pendingDataErr := e.pendingDataFn()
+	if pendingDataErr != nil {
+		return matchedEvents, nil, nil //nolint:nilerr // ignore error return matches so far
```
rodrodros: Please add why it is safe to ignore the error 🙏

blockchain/event_filter.go
```diff
@@ -260,3 +265,107 @@ func (e *EventFilter) canonicalEvents(
...
+	// Process pre-confirmed block
+	return e.processPreConfirmedBlock(matchedEvents, pendingData, skippedEvents, chunkSize)
+}
+
+// processPreLatestBlock processes pre-latest block events
+func (e *EventFilter) processPreLatestBlock(
+	matchedEvents []*FilteredEvent,
+	preLatest *core.Block,
+	skippedEvents, chunkSize uint64,
+) ([]*FilteredEvent, *ContinuationToken, error) {
```
rodrodros: Any chance we can make `*ContinuationToken` to be returned by value instead: `ContinuationToken`?

blockchain/event_filter.go
```diff
@@ -260,3 +265,107 @@ func (e *EventFilter) canonicalEvents(
...
+	}
+	// Process pre-confirmed block
+	return e.processPreConfirmedBlock(matchedEvents, pendingData, skippedEvents, chunkSize)
+}
+
+// processPreLatestBlock processes pre-latest block events
+func (e *EventFilter) processPreLatestBlock(
+	matchedEvents []*FilteredEvent,
+	preLatest *core.Block,
+	skippedEvents, chunkSize uint64,
```
rodrodros: one line per argument please. You can still do

```golang
skippedEvents,
chunSize uint64,
```
if you'd like. It looks better to me as well. 

Let's avoid two arguments per line since it messes with what I think we programmers expect for function parameters to be located.

blockchain/event_matcher.go
```diff
@@ -140,16 +140,22 @@ func (e *EventMatcher) getCandidateBlocksForFilterInto(filter *core.AggregatedBl
 	return nil
 }
 
-func (e *EventMatcher) AppendBlockEvents(matchedEventsSofar []*FilteredEvent, header *core.Header, receipts []*core.TransactionReceipt,
-	skippedEvents uint64, chunkSize uint64,
+func (e *EventMatcher) AppendBlockEvents(
+	matchedEventsSofar []*FilteredEvent,
```
rodrodros: I was looking at the logic of `matchedEvents` being a slice of pointers and I see no reason for it to be that. 

If it were a map, it would be understandable (since you cannot ref values inside it) but it being a slice, I don't understand the need for double pointer. 

Maybe I am missing something, what do you think of making `matchedEventsSoFar` a `[]FilteredEvent` slice. 

In a follow-up PR (and not necessarily done by you)

blockchain/event_filter.go
EgeCaner: Pending refers to pending data but lets use `[canonical, pre-confirmed]`.

blockchain/event_filter.go
EgeCaner: Previously we were using `sync.PendingBlock()` which was returning nil when there is an error. This was resulting in returning collected events even if there is an error reading pending block. We pass `PendingDataFn` from pending data wrappers and they already handle out of date pending data and throws error for db reads etc..  I guess we can return error here or gracefully handle error if error is `PendingBlockNotFound`, else return error.

blockchain/event_matcher.go
EgeCaner: I opened the followup [PR](https://github.com/NethermindEth/juno/pull/3182) for it

blockchain/event_filter.go
EgeCaner: Updated as follows
>gracefully handle error if error is PendingBlockNotFound, else return error.

blockchain/event_matcher.go
rodrodros: Accidentally merged it into here 🤦

blockchain/event_filter.go
```diff
@@ -70,19 +70,26 @@ func (e *EventFilter) WithLimit(limit uint) *EventFilter {
...
+	blockNumber uint64,
+) error {
+	switch filterRange {
+	case EventFilterFrom:
 		e.fromBlock = blockNumber
-	} else if filterRange == EventFilterTo {
+	case EventFilterTo:
 		e.toBlock = blockNumber
-	} else {
+	default:
```
rodrodros, line 82: 👍

PR #3182, opened 2025-10-16
refactor: events return types

EgeCaner:


rpc/v8/subscriptions_test.go
rodrodros, line 1136: Just realized that there is an `rpc.Event` and a `core.Event` which have the same definition. Something for another time as well.

PR #3187, opened 2025-10-18
refactor(juno): integrate common interface with rawDB as trieDB engine

MaksymMalicki:
1. This PR should be reviewed after #3360  
2. It integrates the common interfaces across the project, including in utilities and tests. This PR uses `rawDB` - direct pebbleDB reads/writes - as a trieDB engine.
3. Some adapting had to be also done for the prove logic inside of `storage.go` and `storage_test.go` to support both `trie` and `trie2` packages.
4. The experimental --new-state boolean CLI flag is now fully integrated. It can be used via the command line and in the `dbcmd` `Revert` utility.
5. A new command was added to the `Makefile`, to run the unit tests strictly with the new state and trie implementations in the future CI/CD:
```
test-new-state: clean-testcache rustdeps ## Run tests with new state
	JUNO_NEW_STATE=true go test $(GO_TAGS) -v ./...
```


consensus/p2p/validator/fixtures_test.go
```diff
@@ -108,7 +109,9 @@ func BuildTestFixture(
 
 	executor.RegisterBuildResult(&buildResult)
 
-	builder := builder.New(blockchain.New(database, testCase.Network), executor)
+	builder := builder.New(
+		blockchain.New(database, testCase.Network, statetestutils.UseNewState()), executor,
```
infrmtcs: nit:
```suggestion
		blockchain.New(database, testCase.Network, statetestutils.UseNewState()),
		executor,
```

blockchain/blockchain_test.go
```diff
@@ -270,18 +268,14 @@ func TestStore(t *testing.T) {
...
 		require.NoError(t, chain.Store(block0, &emptyCommitments, stateUpdate0, nil))
 		require.NoError(t, chain.Store(block1, &emptyCommitments, stateUpdate1, nil))
 
 		headBlock, err := chain.Head()
 		require.NoError(t, err)
 		assert.Equal(t, block1, headBlock)
 
-		root, err := chain.StateCommitment()
-		require.NoError(t, err)
-		assert.Equal(t, stateUpdate1.NewRoot, &root)
```
infrmtcs: Why do we remove this test?

blockchain/blockchain_test.go
```diff
@@ -243,17 +245,13 @@ func TestStore(t *testing.T) {
...
+		chain := blockchain.New(memory.New(), &utils.Mainnet, statetestutils.UseNewState())
 		require.NoError(t, chain.Store(block0, &emptyCommitments, stateUpdate0, nil))
 
 		headBlock, err := chain.Head()
 		require.NoError(t, err)
 		assert.Equal(t, block0, headBlock)
 
-		root, err := chain.StateCommitment()
-		require.NoError(t, err)
-		assert.Equal(t, stateUpdate0.NewRoot, &root)
```
infrmtcs: Why do we remove this test?

blockchain/blockchain_test.go
```diff
@@ -724,3 +724,55 @@ func TestSubscribeL1Head(t *testing.T) {
...
+		block, err := gw.BlockByNumber(context.Background(), uint64(i))
+		if err != nil {
+			return nil, nil, err
+		}
+		blocks[i] = block
+	}
+	return suList, blocks, nil
+}
+
+func BenchmarkBlockchainStore(b *testing.B) {
```
infrmtcs: We should write this in a separate file, because this file is already quite long.

blockchain/blockchain_test.go
```diff
@@ -724,3 +724,55 @@ func TestSubscribeL1Head(t *testing.T) {
...
+
+func fetchStateUpdatesAndBlocks(samples int) ([]*core.StateUpdate, []*core.Block, error) {
+	client := feeder.NewClient(utils.Mainnet.FeederURL)
+	gw := adaptfeeder.New(client)
+
+	suList := make([]*core.StateUpdate, samples)
+	blocks := make([]*core.Block, samples)
+	for i := range samples {
+		fmt.Println("fetching", i)
+		su, err := gw.StateUpdate(context.Background(), uint64(i))
```
infrmtcs: Use test context instead

blockchain/blockchain_test.go
```diff
@@ -724,3 +724,55 @@ func TestSubscribeL1Head(t *testing.T) {
...
+		if err != nil {
+			return nil, nil, err
+		}
+		suList[i] = su
+		block, err := gw.BlockByNumber(context.Background(), uint64(i))
```
infrmtcs: Use test context instead

blockchain/blockchain_test.go
```diff
@@ -724,3 +724,55 @@ func TestSubscribeL1Head(t *testing.T) {
...
+		for b.Loop() {
+			b.StopTimer()
+			chain := blockchain.New(memory.New(), &utils.Mainnet, false)
+			b.StartTimer()
+
+			for j := range samples {
+				require.NoError(b, chain.Store(blocks[j], &emptyCommitments, stateUpdates[j], nil))
+			}
+		}
+	})
```
infrmtcs: We should write a for loop for `useNewState`. It's not only about less code, but also about guarantee that the structure of the 2 tests are the same.

blockchain/blockchain_test.go
```diff
@@ -724,3 +724,55 @@ func TestSubscribeL1Head(t *testing.T) {
 	require.True(t, ok)
 	assert.Equal(t, l1Head, got)
 }
+
+func fetchStateUpdatesAndBlocks(samples int) ([]*core.StateUpdate, []*core.Block, error) {
+	client := feeder.NewClient(utils.Mainnet.FeederURL)
```
infrmtcs: nit: We should have a definition `var network = utils.Mainnet` and use `network.FeederURL` here, because:
1. Guarantee that the used network here and in the test functions are the same.
2. It's easy to change the test network if necessary.

blockchain/blockchain_test.go
```diff
@@ -724,3 +724,55 @@ func TestSubscribeL1Head(t *testing.T) {
...
+	gw := adaptfeeder.New(client)
+
+	suList := make([]*core.StateUpdate, samples)
+	blocks := make([]*core.Block, samples)
+	for i := range samples {
+		fmt.Println("fetching", i)
```
infrmtcs: We should either remove it, use zap Logger or logging functions from testing.

blockchain/blockchain_test.go
```diff
@@ -724,3 +724,55 @@ func TestSubscribeL1Head(t *testing.T) {
...
 	assert.Equal(t, l1Head, got)
 }
+
+func fetchStateUpdatesAndBlocks(samples int) ([]*core.StateUpdate, []*core.Block, error) {
+	client := feeder.NewClient(utils.Mainnet.FeederURL)
+	gw := adaptfeeder.New(client)
+
+	suList := make([]*core.StateUpdate, samples)
+	blocks := make([]*core.Block, samples)
+	for i := range samples {
```
infrmtcs: Setup time can take time. If it's the case, we can consider using `pool` or `iter` to query concurrently with rate limit.

blockchain/blockchain.go
```diff
@@ -703,3 +967,20 @@ func (b *Blockchain) StoreGenesis(
...
+		if err != nil {
+			return err
+		}
+
+		stateUpdate, err := b.StateUpdateByNumber(head.Number)
+		if err != nil {
+			return err
+		}
+
+		return b.trieDB.Journal(stateUpdate.NewRoot)
```
infrmtcs: Question: What happens if synchronizer writes a new block in the middle of this process?

blockchain/blockchain.go
```diff
@@ -633,35 +897,35 @@ func (b *Blockchain) signBlock(
 
 // storeBlockData persists all block-related data to the database
 func (b *Blockchain) storeBlockData(
-	txn db.IndexedBatch,
+	w db.KeyValueWriter,
```
infrmtcs: I think we should create a separate small PR to replace this `db.IndexedBatch` with either `db.KeyValueWriter` or `db.KeyValueReader`. Because `db.IndexedBatch` implements both of them, I think it should be quite easy to do.

consensus/p2p/validator/fixtures_test.go
MaksymMalicki: thanks!

blockchain/blockchain_test.go
MaksymMalicki: actually removed this benchmark, its not used anymore and not very precise

blockchain/blockchain_test.go
MaksymMalicki: removed entirely

blockchain/blockchain_test.go
MaksymMalicki: removed entirely

blockchain/blockchain.go
MaksymMalicki: good idea!

blockchain/blockchain_test.go
MaksymMalicki: well, this method from the `blockchain` package is only used here in the tests. I think replacing it with the direct check on the state Root is a better idea than having a utility in the main package just for tests

blockchain/blockchain_test.go
MaksymMalicki: Same here

core/common_state.go
```diff
@@ -17,11 +17,11 @@ type CommonState interface {
 		blockNumber uint64,
 	) (felt.CasmClassHash, error)
 
-	Update(
-		blockNum uint64,
+	Update(blockNum uint64,
```
EgeCaner:
```suggestion
	Update(
	     blockNum uint64,
```

blockchain/blockchain_test.go
MaksymMalicki: Actualy added a test helper function for this

blockchain/statebackend/deprecated.go
```diff
@@ -14,23 +12,6 @@ type deprecatedStateBackend struct {
 	baseState
 }
 
-func (b *deprecatedStateBackend) StateCommitment() (felt.Felt, error) {
```
MaksymMalicki, line 18: This method for bot `stateBackend` and `deprecatedStateBacked` was only used in the tests. If needed, can be added in the future, for now I removed it and swapped with a test helper called `chainStateCommitment` in the `blockchain_test.go`

cmd/juno/dbcmd.go
```diff
@@ -162,14 +178,27 @@ func dbRevert(cmd *cobra.Command, args []string) error {
...
+		return err
+	}
+
 	database, err := openDB(dbPath)
 	if err != nil {
 		return err
 	}
 	defer database.Close()
 
+	if err = blockchain.ValidateStateVersion(database, newState); err != nil {
```
MaksymMalicki, line 177: validation required before reverting

core/receipt.go
```diff
@@ -71,15 +71,15 @@ func messagesSentHash(messages []*L2ToL1Message) felt.Felt {
...
+		trie2.RunOnTempTriePoseidon,
 		func(receipt *TransactionReceipt) felt.Felt {
 			return receipt.hash()
 		},
 	)
 }
 
 type (
-	onTempTrieFunc     func(uint8, func(*trie.Trie) error) error
+	onTempTrieFunc     func(uint8, func(*trie2.Trie) error) error
```
MaksymMalicki, line 82: those are static functions, used solely for the new Hash calculation. `trie2` is more performant for the Hashes, so we can swap those

core/state/history.go
```diff
@@ -76,9 +76,10 @@ func (s *stateHistory) checkDeployed(addr *felt.Felt) error {
 	}
 
 	if !isDeployed {
-		// TODO(weiihann): previously this was db.ErrKeyNotFound
-		// remember to handle it in the rpc
-		return ErrContractNotDeployed
+		// TODO: return a dedicated error like ErrContractNotFound
+		// once deprecated state is removed. Using db.ErrKeyNotFound
+		// for backward compatibility.
+		return db.ErrKeyNotFound
```
MaksymMalicki, line 82: I've decided to make the state fully backward compatible for now. Returning different errors for the state history lookups would force us to do a lot of `oldStateError || newStateError` checks. The errors can be changed to more descriptive ones after the old state is fully deprecated. Hence this TODO

core/state/state.go
```diff
@@ -275,7 +274,11 @@ func (s *State) Update(
 		}
 	}
 
-	return s.flush(blockNum, &stateUpdate, dirtyClasses, true)
+	// TODO: Remove this guard once the State and StateReader are split
```
MaksymMalicki: Handled in the #3563

vm/vm_test.go
```diff
@@ -39,6 +47,7 @@ func TestCallDeprecatedCairo(t *testing.T) {
 	}, map[felt.Felt]core.ClassDefinition{
 		*classHash: simpleClass,
 	}, false))
+	require.NoError(t, batch.Write())
```
MaksymMalicki, line 51: The new state can't read its own writes - database is used for reads. The changes applied to the batch by `state.Update` need to be applied to the DB.

node/node.go
```diff
@@ -206,6 +208,11 @@ func New(cfg *Config, version string, logLevel *log.Level) (*Node, error) {
 	if cfg.Metrics {
 		opts = append(opts, blockchain.WithListener(makeBlockchainMetrics()))
 	}
+	if cfg.NewState {
+		opts = append(opts, blockchain.WithNewState(
+			cfg.NewState,
+		))
+	}
```
EgeCaner: Maybe we can add option regardless since option receives bool, following would provide the same outcome 
```suggestion
		opts = append(opts, blockchain.WithNewState(
			cfg.NewState,
		))
```

Let me know what do you think

node/node.go
EgeCaner:
```suggestion
	opts := make([]blockchain.Option, 0, 2)
```

vm/vm_test.go
```diff
@@ -387,3 +424,24 @@ func TestSetVersionedConstants(t *testing.T) {
...
+		deprecatedState := deprecatedstate.New(txn)
+		return deprecatedState, nil
+	}
+	triedb := triedb.New(testDB, nil)
+	stateDB := state.NewStateDB(testDB, triedb)
+	state, err := state.New(stateRoot, stateDB, batch)
+	if err != nil {
+		return nil, err
+	}
+	return state, nil
```
EgeCaner:
```suggestion
	return state.New(stateRoot, stateDB, batch)
```

rpc/v9/storage_test.go
```diff
@@ -418,21 +474,22 @@ func TestStorageProof(t *testing.T) {
...
-		verifyIf(t, &trieRoot, key, value, proof.ClassesProof, tempTrie.HashFn())
-		verifyIf(t, &trieRoot, key2, value2, proof.ClassesProof, tempTrie.HashFn())
+		verifyIf(t, &trieRoot, key, value, proof.ClassesProof, classTrie.HashFn())
+		verifyIf(t, &trieRoot, key2, value2, proof.ClassesProof, classTrie.HashFn())
 	})
 	t.Run("storage trie address does not exist in a trie", func(t *testing.T) {
 		mockReader.EXPECT().BlockHeaderByNumber(blockNumber).
 			Return(headBlock.Header, nil)
 		mockState.EXPECT().ContractClassHash(noSuchKey).Return(felt.Zero, db.ErrKeyNotFound).Times(1)
+		mockState.EXPECT().ContractNonce(noSuchKey).Return(felt.Zero, db.ErrKeyNotFound).Times(0)
```
EgeCaner: Why times 0?

node/node.go
MaksymMalicki: Agreed

rpc/v9/storage_test.go
MaksymMalicki: could be removed, old regression

EgeCaner (approved):

PR #3190, opened 2025-10-21
fix: rpcv9 websocket notification deduplication

EgeCaner:
In the previous deduplication mechanism, sent notifications were being cleared whenever the pending data changed to a different block. This approach worked fine with a single pending block-to-cache scenario. However, with the introduction of the prelatest feature, this mechanism became insufficient. Consider the following scenario:

- Preconfirmed N arrives: notifications are sent and cached.
- Prelatest N arrives: notifications are sent and cached.
- Preconfirmed N + 1 arrives: since preconfirmed data changed, the cache is cleared, causing loss of notifications related to the prelatest cache.

This PR introduces a new deduplication mechanism that takes the block number into account, preventing unintended cache clearance and preserving all relevant notifications.


rpc/v9/subscription_cache.go
rodrodros, line 1: As a further thought, I think subscription-related logic should have its own package. For another moment.

rpc/v9/subscriptions.go
```diff
@@ -216,10 +216,9 @@ func (h *Handler) SubscribeEvents(
 
 	lenKeys := len(keys)
 	for _, k := range keys {
-		lenKeys += len(k)
-	}
-	if lenKeys > rpccore.MaxEventFilterKeys {
-		return "", rpccore.ErrTooManyKeysInFilter
+		if lenKeys += len(k); lenKeys > rpccore.MaxEventFilterKeys {
```
rodrodros, line 219: I don't mind the approach change, but take into account, this is not objectively better than the previous one. 

It's just different, which makes me a bit curious about the reason for the change?

rpc/v9/subscriptions.go
```diff
@@ -422,7 +422,14 @@ func processBlockEvents(
...
 
-			if err := sendEventWithoutDuplicate(w, &event, eventsPreviouslySent, id, finalityStatus); err != nil {
+			if err := sendEventWithoutDuplicate(
+				w,
+				&event,
+				sentCache,
+				id,
+				finalityStatus,
+				&block.Number,
+			); err != nil {
```
rodrodros: I think if it comes to this is better to have the error call outside the `if` and then just have the
```golang
if err != nil { ... }
```

rpc/v9/subscriptions.go
```diff
@@ -1168,32 +1177,56 @@ func processCandidateTransactions(
...
 		}
 	}
 	return nil
 }
 
+// sendTransactionWithoutDuplicate is a helper function that handles transaction deduplication
+// and sends the transaction if it hasn't been sent before with the same finality status
+func sendTransactionWithoutDuplicate(
+	w jsonrpc.Conn,
+	sentCache *SubscriptionDeduplicationCache[felt.Hash, TxnFinalityStatusWithoutL1],
```
rodrodros: What kind of `hash` is this, Transaction Hash, then I think we should create that specific type in the `Hash` file

rpc/v9/dedup_cache.go
```diff
@@ -0,0 +1,113 @@
...
+// SubscriptionDeduplicationCache is a minimal cache for subscription notifications deduplication.
+// Stores notifications by block number and key, with automatic eviction.
+//
+// The cache is designed to hold at most 3 blocks:
+// - up to 2 non-committed blocks (pre-latest and pre-confirmed)
+// - and 1 committed block.
+// The finalised block will always have the smallest block number.
+// When the cache reaches capacity, it evicts the slot with the lowest block number, which always
+// corresponds to the oldest finalised block.
+type SubscriptionDeduplicationCache[K comparable, V comparable] struct {
```
rodrodros: I think it can be called just `SubscriptionCache`, the Deduplication comes as part of its explanation

rpc/v9/dedup_cache.go
```diff
@@ -0,0 +1,113 @@
...
+	slotIdx := c.findSlotToUse()
+
+	clear(c.slots[slotIdx].data)
+	c.slots[slotIdx].blockNum = blockNum
+	c.slots[slotIdx].data[key] = value
+}
+
+// ShouldSend returns true if the notification should be sent.
+// Returns false if already cached for this block and key.
+func (c *SubscriptionDeduplicationCache[K, V]) ShouldSend(blockNum uint64, key K, value V) bool {
```
rodrodros: This should receive the key and value by reference, given the use we are giving it to them. Small nitpick, since keys are `SentEvent` (5 words) and `felt.Hash` (4 words)

rpc/v9/subscription_cache.go
```diff
@@ -0,0 +1,113 @@
...
+func (c *SubscriptionDeduplicationCache[K, V]) ShouldSend(blockNum uint64, key K, value V) bool {
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+
+	if idx := c.findSlotIndex(blockNum); idx >= 0 {
+		if existingValue, exists := c.slots[idx].data[key]; exists {
+			return existingValue != value
+		}
+	}
+	return true
```
rodrodros, line 69: I am confused right now. Why should it notify only when it cannot find the value?

rpc/v9/dedup_cache.go
```diff
@@ -0,0 +1,113 @@
...
+	for i := range c.slots {
+		if len(c.slots[i].data) == 0 {
+			return i
+		}
+	}
+
+	// All slots are in use; evict the one with the lowest block number
+	lowestIdx := 0
+	lowestBlockNum := c.slots[0].blockNum
+	for i := 1; i < BlockCacheCapacity; i++ {
```
rodrodros: Why it doesn't start from zero?

rpc/v9/subscriptions.go
```diff
@@ -475,26 +483,21 @@ func sendEvents(
...
 	event *blockchain.FilteredEvent,
-	eventsPreviouslySent map[SentEvent]TxnFinalityStatus,
+	sentCache *SubscriptionDeduplicationCache[SentEvent, TxnFinalityStatus],
 	id string,
 	finalityStatus TxnFinalityStatus,
+	blockNum *uint64,
 ) error {
-	if eventsPreviouslySent != nil {
+	// Only use cache for deduplication if we have a block number
+	if blockNum != nil {
```
rodrodros, line 494: When there is no block number 🤔 

Only with a pending block?

If that's the case, then I think it is a good idea to add a `todo` (in a standard format, i.e., "todo: remove check when we drop support for Starknet < 0.14.0"

rpc/v9/subscriptions.go
```diff
@@ -963,25 +953,46 @@ func processBlockReceipts(
...
+			sentCache,
+			block.Number,
+			sentReceipt,
+			finalityStatus,
+			id,
+			adaptedReceipt,
+		); err != nil {
 			return err
 		}
 	}
```
rodrodros: This function (`sendReceiptWithoutDuplicate`) is only abstracting away three lines of code, and it is being used only in one place. We can remove it and inline it's content here

rpc/v9/subscriptions.go
EgeCaner: Sorry for the confusion. The reason I made this change is,  linter was complaining about the function length, I made the change only to save one line to silence it. I can revert this and use nolint instead. The actual problem of function size getting larger will be solved in broader refactor PR.

rpc/v9/subscriptions.go
EgeCaner: I didnt realize we have TransactionHash, I will use that instead

rodrodros: If it doesn't exist, then feel free to create, the it should be 1 more line of code

rpc/v9/subscriptions.go
rodrodros: hahaha ok, let's keep it then. Smart ways of silencing the linter :p

rpc/v9/subscription_cache.go
EgeCaner: If found it means we already sent this event. We only track last sent status to allow status regression for pre_confirmed.

rpc/v9/dedup_cache.go
EgeCaner: It tries to find the lowest block number in cache. Since we know list is not empty, we can use index 0 as initial lowest

I guess alternative should be something like following

```go
       lowestIdx := -1
       lowestBlockNum := ^uint64(0)
       for i := range c.slots {
```

rodrodros: Ah makes sense what you have there. Didn't read the code right the first time 🤦

rpc/v9/subscription_cache.go
```diff
@@ -0,0 +1,113 @@
...
+// SubscriptionCache is a minimal cache for subscription notifications deduplication.
+// Stores notifications by block number and key, with automatic eviction.
+//
+// The cache is designed to hold at most 3 blocks:
+// - up to 2 non-committed blocks (pre-latest and pre-confirmed)
+// - and 1 committed block.
+// The finalised block will always have the smallest block number.
+// When the cache reaches capacity, it evicts the slot with the lowest block number, which always
+// corresponds to the oldest finalised block.
+type SubscriptionCache[K comparable, V comparable] struct {
```
rodrodros, line 18: The explanation is sound, thank you!

Can you please expand on the "why", something similar to the PR description 🙏

rodrigo-pino (approved):

PR #3193, opened 2025-10-22
refactor: rpcv9 tracing

EgeCaner:


rpc/v9/trace.go
```diff
@@ -62,308 +61,444 @@ type FunctionInvocation struct {
...
+			class, stateErr := state.Class(tx.ClassHash)
+			if stateErr != nil {
+				return nil, nil, jsonrpc.Err(jsonrpc.InternalError, stateErr.Error())
+			}
+			declaredClasses = append(declaredClasses, class.Class)
+		case *core.L1HandlerTransaction:
+			var fee felt.Felt
+			l1HandlerFees = append(l1HandlerFees, fee.SetUint64(1))
+		}
 	}
```
rodrodros, line 278: Use `felt.One`

rpc/v9/trace.go
```diff
@@ -62,308 +61,444 @@ type FunctionInvocation struct {
...
-		}
+	return h.traceBlockWithVM(block)
+}
+
+// shouldFetchTracesFromFeederGateway determines if
+// traces for a block should be fetched from the feeder gateway.
+func (h *Handler) shouldFetchTracesFromFeederGateway(block *core.Block) (bool, error) {
+	blockVer, err := core.ParseBlockVersion(block.ProtocolVersion)
+	if err != nil {
+		return false, err
```
rodrodros: I think this is better refactored as a function rather than a method, due to the scarce use of `h`

rpc/v9/trace.go
```diff
@@ -62,308 +61,443 @@ type FunctionInvocation struct {
...
-			}
-			h.blockTraceCache.Add(rpccore.TraceCacheKey{BlockHash: *block.Hash}, traces)
-			return traces, defaultExecutionHeader(), nil
-		}
+	return h.traceBlockWithVM(block)
+}
+
+// shouldFetchTracesFromFeederGateway determines if
+// traces for a block should be fetched from the feeder gateway.
+func shouldFetchTracesFromFeederGateway(block *core.Block, network *utils.Network) (bool, error) {
```
rodrodros: Maybe we should move it to after the methods from `h`?

PR #3194, opened 2025-10-22
Feat: added gzip encoding support in rpc server

ndatta-nethermind:


jsonrpc/http_test.go
```diff
@@ -74,3 +78,118 @@ func TestHTTP(t *testing.T) {
...
+
+	t.Run("success: gzip encoded request & response", func(t *testing.T) {
+		var buf bytes.Buffer
+		gz := gzip.NewWriter(&buf)
+		if _, err := gz.Write([]byte(msg)); err != nil {
+			panic(err)
+		}
+		if err := gz.Close(); err != nil {
+			panic(err)
+		}
```
rodrodros: Please do not panic, use `req.NoError`

jsonrpc/http.go
```diff
@@ -57,9 +63,49 @@ func (h *HTTP) ServeHTTP(writer http.ResponseWriter, req *http.Request) {
...
+		gz, err := gzip.NewReader(req.Body)
+		if err != nil {
+			http.Error(writer, "invalid gzip", http.StatusBadRequest)
+			return false
+		}
+		defer gz.Close()
+		req.Body = gz
+	}
+	return true
+}
```
rodrodros: I've read that the http package performs automatic decompression if the data is encoded in gzip. Do we need this function?

jsonrpc/http.go
```diff
@@ -57,9 +63,49 @@ func (h *HTTP) ServeHTTP(writer http.ResponseWriter, req *http.Request) {
...
+		if err := gw.Close(); err != nil {
+			// Handle gzip close error
+			http.Error(writer, "gzip close error", http.StatusInternalServerError)
+			return true
+		}
+
+		return true
+	}
+	return false
+}
```
rodrodros: This function refactoring can be improved; it doesn't blend nicely with the code. 

Please integrate it into the parent function. 

I think a better structure is to have a function that compress the data only. In the parent function, the function that calls this one, put the `if need to compress` and compress the data. 

All the other logic should still use the one from the parent function

jsonrpc/http_test.go
```diff
@@ -74,3 +78,118 @@ func TestHTTP(t *testing.T) {
...
+		req, err := http.NewRequestWithContext(ctx, http.MethodPost, srv.URL, &buf)
+		require.NoError(t, err)
+		req.Header.Set("Content-Encoding", "gzip")
+		req.Header.Set("Content-Type", "application/json") // still JSON payload, just compressed
+		req.Header.Set("Accept-Encoding", "gzip")
+		resp, err := client.Do(req)
+		require.NoError(t, err)
+		defer resp.Body.Close()
```
infrmtcs: Please create a function for these repetitive lines.

jsonrpc/http_test.go
```diff
@@ -74,3 +78,118 @@ func TestHTTP(t *testing.T) {
...
+
+		require.Equal(t, http.StatusOK, resp.StatusCode)
+		require.Equal(t, "gzip", resp.Header.Get("Content-Encoding"))
+		var compressedBody bytes.Buffer
+		_, err = io.Copy(&compressedBody, resp.Body)
```
infrmtcs: Why do we need to copy?

jsonrpc/http_test.go
```diff
@@ -74,3 +78,118 @@ func TestHTTP(t *testing.T) {
...
+			req.Header.Set("Accept-Encoding", "gzip")
+			resp, err := client.Do(req)
+			require.NoError(t, err)
+			defer resp.Body.Close()
+
+			require.Equal(t, http.StatusBadRequest, resp.StatusCode)
+		})
+}
+
+func randomAlphaNumeric(n int) string {
```
infrmtcs: You can replace this with `rand.Text()` and `rand.Read()` from `crypto/rand`

jsonrpc/http_test.go
```diff
@@ -74,3 +78,118 @@ func TestHTTP(t *testing.T) {
...
+		require.NoError(t, err)
+
+		gzr, err := gzip.NewReader(&compressedBody)
+		require.NoError(t, err)
+		defer gzr.Close()
+
+		decompressedBody, err := io.ReadAll(gzr)
+		require.NoError(t, err)
+		t.Logf("Decompressed: %s\n", decompressedBody)
+		assert.Contains(t, string(decompressedBody), expected)
```
infrmtcs:
```suggestion
		assert.Equal(t, string(decompressedBody), expected)
```

jsonrpc/http_test.go
```diff
@@ -74,3 +78,118 @@ func TestHTTP(t *testing.T) {
...
+			panic(err)
+		}
+		if err := gz.Close(); err != nil {
+			panic(err)
+		}
+		req, err := http.NewRequestWithContext(ctx, http.MethodPost, srv.URL, &buf)
+		require.NoError(t, err)
+		req.Header.Set("Content-Encoding", "gzip")
+		req.Header.Set("Content-Type", "application/json") // still JSON payload, just compressed
+		req.Header.Set("Accept-Encoding", "gzip")
```
infrmtcs: Please test with the standard header:
```suggestion
		req.Header.Set("Accept-Encoding", "gzip, deflate, br")
```

infrmtcs: The reason is to guarantee that the header check is correct in this case.

jsonrpc/http.go
ndatta-nethermind: yes, we need this function. The automatic decompression happens on client side only. Also, it happens iff "Accpet-Encoding" header is not set automatically. But, on server side, the whole compression and decompression is manually handled.

For reference: https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/internal/accept-encoding

jsonrpc/http.go
ndatta-nethermind: Refactored.

jsonrpc/http_test.go
ndatta-nethermind: Refactored.

jsonrpc/http_test.go
ndatta-nethermind: Removed the copy operation.

jsonrpc/http_test.go
ndatta-nethermind: I did not find any `rand.Text()` method.
`rand.Read()` method is deprecated.

jsonrpc/http.go
```diff
@@ -57,9 +67,28 @@ func (h *HTTP) ServeHTTP(writer http.ResponseWriter, req *http.Request) {
 		writer.WriteHeader(http.StatusInternalServerError)
 	}
 	if resp != nil {
+		if strings.Contains(req.Header.Get("Accept-Encoding"), "gzip") {
+			encodeDataInGzip(writer, resp)
+			return
+		}
```
infrmtcs, line 73:
```suggestion
		var ioWriter io.Writer = writer
		if strings.Contains(req.Header.Get("Accept-Encoding"), "gzip") {
			writer.Header().Set("Content-Encoding", "gzip")
			gw := gzip.NewWriter(writer)
			defer func() {
				if err := gw.Close(); err != nil {
					http.Error(writer, "gzip close error", http.StatusInternalServerError)
				}
			}()
			ioWriter = gw
		}
```

jsonrpc/http_test.go
infrmtcs: The function is available in `crypto/rand` as mentioned in the comment, not `math/rand`

jsonrpc/http_test.go
```diff
@@ -74,3 +78,107 @@ func TestHTTP(t *testing.T) {
...
+	for i := range b {
+		b[i] = letters[r.Intn(len(letters))]
+	}
+	return string(b)
+}
+
+func setHeaderAndProcessRequest(
+	client *http.Client,
+	t *testing.T,
+	req *http.Request,
```
infrmtcs: Why don't we create `req` inside this function?

jsonrpc/http_test.go
```diff
@@ -74,3 +78,107 @@ func TestHTTP(t *testing.T) {
...
+		verifyResponse(resp, t, expected)
+	})
+
+	t.Run("success: gzip encoded request & response", func(t *testing.T) {
+		var buf bytes.Buffer
+		gz := gzip.NewWriter(&buf)
+		_, err := gz.Write([]byte(msg))
+		require.NoError(t, err)
+		err = gz.Close()
+		require.NoError(t, err)
```
infrmtcs: nit:

```suggestion
		require.NoError(t, gz.Close())
```

jsonrpc/http_test.go
infrmtcs: We can also put `t.Cleanup(func() { req.Body.Close() })`

jsonrpc/http_test.go
```diff
@@ -74,3 +78,107 @@ func TestHTTP(t *testing.T) {
...
+	method := jsonrpc.Method{
+		Name: "echo",
+		Handler: func(msg string) (string, *jsonrpc.Error) {
+			return msg, nil
+		},
+		Params: []jsonrpc.Parameter{{Name: "msg"}},
+	}
+	listener := CountingEventListener{}
+	log := utils.NewNopZapLogger()
+	rpc := jsonrpc.NewServer(1, log).WithListener(&listener)
```
infrmtcs: If we don't use the listener:
```suggestion
	rpc := jsonrpc.NewServer(1, log)
```

jsonrpc/http_test.go
```diff
@@ -74,3 +78,107 @@ func TestHTTP(t *testing.T) {
...
+	require.NoError(t, rpc.RegisterMethods(method))
+
+	ctx, cancel := context.WithCancel(t.Context())
+	defer cancel()
+
+	srv := httptest.NewServer(jsonrpc.NewHTTP(rpc, log))
```
infrmtcs, line 105: According to the doc:
```
// NewServer starts and returns a new [Server].
// The caller should call Close when finished, to shut it down.
```
we need:

```suggestion
	srv := httptest.NewServer(jsonrpc.NewHTTP(rpc, log))
	t.Cleanup(srv.Close)
```

jsonrpc/http_test.go
```diff
@@ -74,3 +78,107 @@ func TestHTTP(t *testing.T) {
...
+		},
+		Params: []jsonrpc.Parameter{{Name: "msg"}},
+	}
+	listener := CountingEventListener{}
+	log := utils.NewNopZapLogger()
+	rpc := jsonrpc.NewServer(1, log).WithListener(&listener)
+	require.NoError(t, rpc.RegisterMethods(method))
+
+	ctx, cancel := context.WithCancel(t.Context())
+	defer cancel()
```
infrmtcs: We don't need this. `t.Context()` is automatically cancelled during test cleanup.

jsonrpc/http_test.go
```diff
@@ -74,3 +78,107 @@ func TestHTTP(t *testing.T) {
...
+}
+
+func setHeaderAndProcessRequest(
+	client *http.Client,
+	t *testing.T,
+	req *http.Request,
+) *http.Response {
+	req.Header.Set("Content-Encoding", "gzip")
+	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set("Accept-Encoding", "gzip, deflate, br")
```
infrmtcs: If we want to customize this between tests, we can add a `extraHeaders map[string]string` as a function argument.

jsonrpc/http.go
```diff
@@ -45,6 +47,14 @@ func (h *HTTP) ServeHTTP(writer http.ResponseWriter, req *http.Request) {
...
 	}
 
+	if req.Header.Get("Content-Encoding") == "gzip" {
+		gz, err := gzip.NewReader(req.Body)
+		if err != nil {
+			http.Error(writer, "invalid gzip", http.StatusBadRequest)
+			return
+		}
+		req.Body = gz
+	}
```
infrmtcs: I strongly think we should reconsider the support for gzip in request body. This can be the source of some bigger problems:
- Decompression bomb: Attacker can try sending a huge amount of data gzipped.
- It's hard for the outer layer to inspect, filter or cache the requests if necessary.

infrmtcs: cc @rodrigo-pino

jsonrpc/http_test.go
ndatta-nethermind: Fixed.

jsonrpc/http_test.go
ndatta-nethermind: Added.

jsonrpc/http_test.go
```diff
@@ -74,3 +76,92 @@ func TestHTTP(t *testing.T) {
...
+		verifyResponse(resp, t, expected)
+	})
+
+	t.Run("success: gzip encoded request & response", func(t *testing.T) {
+		var buf bytes.Buffer
+		gz := gzip.NewWriter(&buf)
+		_, err := gz.Write([]byte(msg))
+		require.NoError(t, err)
+		require.NoError(t, gz.Close())
+		headers["Content-Encoding"] = "gzip"
```
infrmtcs: We should clone `headers` first, to make sure a test can't affect other tests.

jsonrpc/http_test.go
```diff
@@ -74,3 +76,92 @@ func TestHTTP(t *testing.T) {
...
+		resp := setHeaderAndProcessRequest(ctx, client, headers, &buf, t, srv)
+		defer resp.Body.Close()
```
infrmtcs: We don't need this, because we already have `t.Cleanup` inside `setHeaderAndProcessRequest`

jsonrpc/http_test.go
```diff
@@ -74,3 +76,92 @@ func TestHTTP(t *testing.T) {
...
+func setHeaderAndProcessRequest(
+	ctx context.Context,
+	client *http.Client,
+	headers map[string]string,
+	msg io.Reader,
+	t *testing.T,
+	srv *httptest.Server,
+) *http.Response {
+	req, err := http.NewRequestWithContext(
+		ctx,
```
infrmtcs:
```suggestion
		t.Context(),
```

jsonrpc/http_test.go
```diff
@@ -74,3 +76,92 @@ func TestHTTP(t *testing.T) {
...
+		defer resp.Body.Close()
+		verifyResponse(resp, t, expected)
+	})
+
+	t.Run("failed: request is not gzip encoded but set header as gzip encoded",
+		func(t *testing.T) {
+			resp := setHeaderAndProcessRequest(ctx, client, headers, bytes.NewReader([]byte(msg)), t, srv)
+			defer resp.Body.Close()
+			require.Equal(t, http.StatusBadRequest, resp.StatusCode)
+		})
```
infrmtcs:
```suggestion
	t.Run("failed: request is not gzip encoded but set header as gzip encoded",	func(t *testing.T) {
			resp := setHeaderAndProcessRequest(ctx, client, headers, bytes.NewReader([]byte(msg)), t, srv)
			require.Equal(t, http.StatusBadRequest, resp.StatusCode)
		},
	)
```

jsonrpc/http.go
rodrodros: Agree! @ndatta-nethermind please remove support for recieving requests encoded with gzip 🙏

infrmtcs: @ndatta-nethermind Would you please double check this? I see the comment resolved but the code wasn't removed.

jsonrpc/http_test.go
```diff
@@ -74,3 +89,69 @@ func TestHTTP(t *testing.T) {
...
+
+	srv := httptest.NewServer(jsonrpc.NewHTTP(rpc, log))
+	t.Cleanup(srv.Close)
+	client := new(http.Client)
+
+	payload := rand.Text()
+	msg := fmt.Sprintf(`{"jsonrpc":"2.0", "method":"echo", "params":[%q], "id":1}`, payload)
+	expected := fmt.Sprintf(`{"jsonrpc":"2.0","result":%q,"id":1}`, payload)
+	commonHeaders := map[string]string{
+		"Accept-Encoding": "gzip, deflate, br",
```
rodrodros, line 113: Why do we need `deflate` and `br` as well, isn't `gzip` enough?

rodrigo-pino (approved):

jsonrpc/http_test.go
ndatta-nethermind: It's to check if server returns only `gzip` skipping others as the server only support it.

PR #3206, opened 2025-10-24
refactor(trie): renames to unify the trie interfaces

MaksymMalicki:
To unify the interfaces of new and old state, this PR:
- Renames `trie` method from `Root` -> `Hash`
- Makes the `Hash` method from `trie2` return a nil error

### DIFF
--- a/core/crypto/blake2s/utils.go
+++ b/core/crypto/blake2s/utils.go
@@ -40,9 +40,13 @@ func encodeFeltsToUint32s(felts ...*felt.Felt) []uint32 {
 		} else {
 			start := len(encoding)
 
-			for i := range len(fb) {
-				encoding = append(encoding, uint32(fb[i]>>32), uint32(fb[i]))
-			}
+			encoding = append(
+				encoding,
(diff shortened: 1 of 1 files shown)


core/state/state.go
```diff
@@ -154,8 +154,8 @@ func (s *State) ContractStorageTrie(addr *felt.Felt) (*trie2.Trie, error) {
 
 // Returns the state commitment
 func (s *State) Commitment() (felt.Felt, error) {
-	contractRoot := s.contractTrie.Hash()
-	classRoot := s.classTrie.Hash()
+	contractRoot, _ := s.contractTrie.Hash()
+	classRoot, _ := s.classTrie.Hash()
```
infrmtcs: We should check for error here, because it's not always guaranteed to be non-nil, depending on the trie version.

infrmtcs (approved):
I only have 1 comment here, the rest LGTM!

core/state/state.go
MaksymMalicki: yes!

PR #3212, opened 2025-10-24
feat: add workflow to test buf difs

rodrodros:
I think I have an error somewhere because it should've failed either at the first or second commit error.

I think we can add a check for `go generate` code as well 🤔

### DIFF
--- a/.github/workflows/generated-code-matches.yml
+++ b/.github/workflows/generated-code-matches.yml
@@ -1,7 +1,17 @@
-name: Check generated protobufs
+name: Verify generated protobufs
 on:
-  pull_request:
   push:
+    tags:
+      - v*
+    branches:
+      - main
+  pull_request:
+permissions:
+  contents: read
+  pull-requests: read
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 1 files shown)


.github/workflows/generated-code-matches.yml
```diff
@@ -0,0 +1,37 @@
...
+      - v*
+    branches:
+      - main
+  pull_request:
+permissions:
+  contents: read
+  pull-requests: read
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
```
infrmtcs, line 14: We probably don't need this.

.github/workflows/generated-code-matches.yml
```diff
@@ -0,0 +1,37 @@
...
+      - name: Set up buf
+        uses: bufbuild/buf-setup-action@v1
+        with:
+          version: latest
+      - name: Generate protobuf files
+        run: |
+          make generate-buf
+      - name: Check for uncommitted changes
+        run: |
+          if ! git diff --exit-code; then
```
infrmtcs, line 34: I think writing a new uncommitted file won't fail this.

PR #3224, opened 2025-10-30
Refactor: Refactored adapters/core2sn/class.go

ndatta-nethermind:


adapters/core2sn/class.go
```diff
@@ -24,41 +32,53 @@ func AdaptCasmClass(coreCasmClass *core.CasmClass) starknet.CasmClass {
...
+	for index, ep := range coreCasmClass.L1Handler {
+		feederCompiledClass.EntryPoints.L1Handler[index] = AdaptCasmEntryPoint(&ep)
+	}
+	feederCompiledClass.EntryPoints.Constructor = make(
+		[]starknet.CompiledEntryPoint,
+		len(coreCasmClass.Constructor),
 	)
+	for index, ep := range coreCasmClass.Constructor {
+		feederCompiledClass.EntryPoints.Constructor[index] = AdaptCasmEntryPoint(&ep)
+	}
```
rodrodros, line 63: Let's separate each array creation and iteration over it with a single whiteline.

Currently it is 
```
// make the array 1
// iterate over it
// make the array 2
// iterate over it
// and so on ...
```

Let's make it
```
// make the array1
// iterate over it

// make the array2
// iterate over it

// and so on ...
```

rodrodros: Additionally, let's change the for loops to perform no copies, when we do:

```golang
for index, ep := range array {
   feederCompiledClass.EntryPoints.Constructor[index] = AdaptCasmEntryPoint(&ep)
}
```

We are making a copy of `ep` for every iteration, we can avoid doing this copy by

```golang
for index, ep := range. array {
   feederCompiledClass.EntryPoints.COnstructor[index] = AdaptCasmEntryPoint(&array[index])
}
```

adapters/core2sn/class.go
```diff
@@ -24,41 +32,53 @@ func AdaptCasmClass(coreCasmClass *core.CasmClass) starknet.CasmClass {
...
+		constructors[index] = AdaptSierraEntryPoint(&ep)
+	}
+	external := make([]starknet.SierraEntryPoint, len(class.EntryPoints.External))
+	for index, ep := range class.EntryPoints.External {
+		external[index] = AdaptSierraEntryPoint(&ep)
+	}
+	handlers := make([]starknet.SierraEntryPoint, len(class.EntryPoints.L1Handler))
+	for index, ep := range class.EntryPoints.L1Handler {
+		handlers[index] = AdaptSierraEntryPoint(&ep)
 	}
```
rodrodros, line 89: Let's apply the same comment as before where we don't copy `ep` and address it directly in the code (i.e. instead of `&ep`, `&array[index]`)

rodrigo-pino (commented):
Looks good, some minor comments

Issue #3204, opened 2025-10-24
Refactor p2p2core adapters


rodrodros:
In "adapters/p2p2core/class.go"

Please refactor `AdaptClass` (by renaming it to `AdaptClassDefinition` and:
1. change the `adaptEP` lambda function into a normal function (I think a similar function already exists which is called `adaptEntryPoint`.
2. remove the use of `utils.map` and use a for-loop. 

Rename `createCompiledClass` and name it `compileToCasm`:
1. Change the return type from a reference type to a value type (*core.CasmClass` to `core.CasmClass`)
2. Add a `todo` on top which asks: "should this be a method of `class.Cairo1Class`"
3. In order to get this change correct you will have to update `sn2coreAdaptCompiledClass`. 

In "adapters/p2p2core/transaction.go", change the return types of `AdaptDeclareV3WithClass` to return by value.

PR #3225, opened 2025-10-30
Refactor: Refactored p2p2core adapater

ndatta-nethermind:


adapters/p2p2core/class.go
```diff
@@ -76,23 +67,42 @@ func AdaptClass(cls *class.Class) (core.ClassDefinition, error) {
...
+
 func adaptEntryPoint(e *class.EntryPoint) core.DeprecatedEntryPoint {
 	return core.DeprecatedEntryPoint{
 		Selector: AdaptFelt(e.Selector),
 		Offset:   new(felt.Felt).SetUint64(e.Offset),
 	}
 }
 
-func createCompiledClass(cairo1 *class.Cairo1Class) (*core.CasmClass, error) {
+// todo should this be a method of class.Cairo1Class
```
rodrodros: Is this `todo` a leftover?

starknetdata/feeder/feeder_test.go
```diff
@@ -215,7 +215,7 @@ func TestClassV1(t *testing.T) {
 		if test.hasCompiledClass {
 			assert.NotNil(t, adaptedResponse.Compiled)
 		} else {
-			assert.Nil(t, adaptedResponse.Compiled)
+			assert.Empty(t, adaptedResponse.Compiled)
```
infrmtcs, line 218: Why do we change the behaviour?

migration/migration_pkg_test.go
```diff
@@ -349,7 +349,7 @@ func TestMigrateCairo1CompiledClass(t *testing.T) {
 		if test.checkCompiledExists {
 			assert.NotNil(t, actualClass.Compiled)
 		} else {
-			assert.Nil(t, actualClass.Compiled)
+			assert.Empty(t, actualClass.Compiled)
```
infrmtcs, line 348: Why do we change the behaviour?

adapters/sn2core/sn2core_test.go
```diff
@@ -629,7 +629,7 @@ func TestClassV1(t *testing.T) {
 func TestAdaptCompiledClass(t *testing.T) {
 	result, err := sn2core.AdaptCompiledClass(nil)
 	require.NoError(t, err)
-	assert.Nil(t, result)
+	assert.Empty(t, result)
```
infrmtcs, line 643: Why do we change the behaviour?

adapters/p2p2core/class.go
```diff
@@ -76,23 +67,42 @@ func AdaptClass(cls *class.Class) (core.ClassDefinition, error) {
...
+
+func compileToCasm(cairo1 *class.Cairo1Class) (core.CasmClass, error) {
 	if cairo1 == nil {
-		return nil, nil
+		//nolint:exhaustruct // intentionally returning an empty CasmClass
+		return core.CasmClass{}, nil
```
infrmtcs: I'm wondering if this is correct because we instantiate a `core.CasmClass` in the Cairo 0 case, where `*class.Cairo1Class` is `nil`.

adapters/sn2core/sn2core.go
```diff
@@ -323,9 +323,10 @@ func AdaptSierraClass(
 	}, nil
 }
 
-func AdaptCompiledClass(compiledClass *starknet.CasmClass) (*core.CasmClass, error) {
+func AdaptCompiledClass(compiledClass *starknet.CasmClass) (core.CasmClass, error) {
 	if compiledClass == nil {
-		return nil, nil
+		//nolint:exhaustruct // intentionally returning an empty CasmClass
+		return core.CasmClass{}, nil
```
infrmtcs, line 351: I'm wondering if this is correct because we instantiate a `core.CasmClass` in the Cairo 0 case, where `*class.Cairo1Class` is `nil`.

migration/migration.go
```diff
@@ -808,7 +808,7 @@ func migrateCairo1CompiledClass2(txn db.KeyValueWriter, key, value []byte, _ *ut
 			Program:         class.Class.Program,
 			ProgramHash:     class.Class.ProgramHash,
 			SemanticVersion: class.Class.SemanticVersion,
-			Compiled:        casmClass,
+			Compiled:        &casmClass,
```
infrmtcs, line 808: This means we have a zero `core.CasmClass` instead of `nil` in deprecated case.

starknetdata/feeder/feeder_test.go
ndatta-nethermind: This is the reason: https://github.com/NethermindEth/juno/pull/3225#discussion_r2488528069

migration/migration_pkg_test.go
ndatta-nethermind: This is the reason: https://github.com/NethermindEth/juno/pull/3225#discussion_r2488528069

adapters/sn2core/sn2core_test.go
ndatta-nethermind: This is the reason: https://github.com/NethermindEth/juno/pull/3225#discussion_r2488528069

migration/migration_pkg_test.go
infrmtcs: Why do we change the behaviour?

adapters/sn2core/sn2core_test.go
infrmtcs: Why do we change the behaviour?

starknetdata/feeder/feeder_test.go
infrmtcs: Why do we change the behaviour?

ndatta-nethermind: This was the requirement: https://github.com/NethermindEth/juno/issues/3204

adapters/p2p2core/class.go
ndatta-nethermind: This was a requirement: 
`Add a todo on top which asks: "should this be a method of class.Cairo1Class"`

starknetdata/feeder/feeder_test.go
infrmtcs: Here is the screenshot of the issue:
image

None of the above is saying that we should set `Compiled` to `&CasmClass{}` instead of `nil` if the declared class is Cairo V0.

adapters/p2p2core/class.go
```diff
@@ -76,23 +67,42 @@ func AdaptClass(cls *class.Class) (core.ClassDefinition, error) {
...
 
+func adaptSierraEntryPoints(points []*class.SierraEntryPoint) []core.SierraEntryPoint {
+	sierraEntryPoints := make([]core.SierraEntryPoint, len(points))
+	for index := range points {
+		sierraEntryPoints[index] = adaptSierra(points[index])
+	}
+	return sierraEntryPoints
+}
+
 func adaptSierra(e *class.SierraEntryPoint) core.SierraEntryPoint {
```
EgeCaner: What do you think about renaming the `adaptSierra` to `adaptSierraEntryPoint`?

adapters/p2p2core/class.go
```diff
@@ -76,23 +67,42 @@ func AdaptClass(cls *class.Class) (core.ClassDefinition, error) {
...
+// todo should this be a method of class.Cairo1Class
+
+func compileToCasm(cairo1 *class.Cairo1Class) (core.CasmClass, error) {
 	if cairo1 == nil {
-		return nil, nil
+		//nolint:exhaustruct // intentionally returning an empty CasmClass
+		return core.CasmClass{}, nil
 	}
 
 	adapt := func(ep *class.SierraEntryPoint) starknet.SierraEntryPoint {
```
EgeCaner, line 102: Cant we reuse `adaptSierra` or `adaptSierraEntryPoints` here?

adapters/p2p2core/transaction.go
```diff
@@ -27,10 +27,16 @@ func AdaptDeclareV3WithClass(
 
 	declareCommon, err := AdaptDeclareV3TxnCommon(tx.Common, classHash, txnHash)
 	if err != nil {
-		return nil, nil, fmt.Errorf("failed to adapt declare v3 transaction common: %w", err)
+		return nil,
+			nil,
+			fmt.Errorf("failed to adapt declare v3 transaction common: %w", err)
 	}
 	if *class.Compiled.Hash() != *declareCommon.CompiledClassHash {
```
EgeCaner: I guess we cache the output of `class.Compiled.Hash()`, we dont want to calculate CASM hash twice in error case, it is expensive. Also we can use `!class.Compiled.Hash().Equal(declareCommon.CompiledClassHash)` here for equality check

migration/migration.go
EgeCaner: Yeah, looks like we should have nil instead

starknetdata/feeder/feeder_test.go
ndatta-nethermind: It's asking to return value type instead of reference type in `CompileToCasm` method into `adapters/p2p2core/class.go`.

adapters/p2p2core/class.go
ndatta-nethermind: No. `adaptSierra` return `core.SierraEntrypoint`.

starknetdata/feeder/feeder_test.go
infrmtcs: - If Cairo V0: Set `Compiled` to `nil`
- If Cairo V1: Set `Compiled` to `&casmClass`

adapters/p2p2core/class.go
```diff
@@ -76,23 +67,42 @@ func AdaptClass(cls *class.Class) (core.ClassDefinition, error) {
...
+}
+
+func adaptSierraEntryPoint(e *class.SierraEntryPoint) core.SierraEntryPoint {
 	return core.SierraEntryPoint{
 		Index:    e.Index,
 		Selector: AdaptFelt(e.Selector),
 	}
 }
 
+func adaptEntryPoints(points []*class.EntryPoint) []core.DeprecatedEntryPoint {
```
infrmtcs:
```suggestion
func adaptDeprecatedEntryPoints(points []*class.EntryPoint) []core.DeprecatedEntryPoint {
```

adapters/sn2core/sn2core.go
infrmtcs: It seems that now we're only calling this function when `compiledClass` is non-nil, so we can throw an error here instead.

adapters/p2p2core/class.go
infrmtcs: It seems that now we're only calling this function when `compiledClass` is non-nil, so we can throw an error here instead.

starknetdata/feeder/feeder_test.go
rodrodros: @infrmtcs  but if we are adapting Sierra classes, Cairo 0 wouldn't be adapted to it. Where do we adapt Cairo zero and set the Casm to nil?

.golangci.yaml
```diff
@@ -46,6 +46,8 @@ linters:
               desc: use github.com/NethermindEth/juno/utils for logging
     dupl:
       threshold: 100
+    exhaustruct:
+      allow-empty-returns: true
```
infrmtcs, line 50: Please revert

PR #3227, opened 2025-10-30
refactor(vm): use protobuf compiled class marshalling

MaksymMalicki:


rodrigo-pino:
Closing for now, since we have shift our attention to other parts, and the benefits of this changes won't be super visible

PR #3229, opened 2025-10-31
feat: sync 0.14.1 blocks

EgeCaner:
Assumes same is class is not declared and migrated in the same block

TODO: Pre-compute blake hashes in sync, clear once class is migrated, on reorg repopulate precomputed blake hashes table. Will require migration

### DIFF
--- a/core/state.go
+++ b/core/state.go
@@ -592,11 +592,12 @@ func (s *State) Revert(blockNumber uint64, update *StateUpdate) error {
 		return fmt.Errorf("verify state update root: %v", err)
 	}
 
-	if err = s.removeDeclaredClasses(
+	err = s.removeDeclaredClasses(
 		blockNumber,
 		update.StateDiff.DeclaredV0Classes,
 		update.StateDiff.DeclaredV1Classes,
-	); err != nil {
+	)
+	if err != nil {
 		return fmt.Errorf("remove declared classes: %v", err)
 	}
 


core/state_update.go
```diff
@@ -233,11 +233,13 @@ func declaredClassesDigest(
 	}
 	slices.SortFunc(keys, func(a felt.Felt, b felt.Felt) int { return a.Cmp(&b) })
 
-	// todo(rdr): Is it more performant to do multiple calls to `diget.Update`, or
-	// pre-store all the values and make a single call to `digest.Update`?
```
rodrodros, line 237: Miss-delete, or is better this way?

core/version.go
```diff
@@ -12,7 +12,8 @@ var (
 	Ver0_13_2 = semver.MustParse("0.13.2")
 	Ver0_13_4 = semver.MustParse("0.13.4")
 	Ver0_14_0 = semver.MustParse("0.14.0")
-	LatestVer = Ver0_14_0
+	Ver0_14_1 = semver.MustParse("0.14.1")
```
rodrodros: I am worried if this is really relevant, and might break stuff for current Sepolia and Mainnet. Since 0.14.1 has no breaking changes, just extra fields, do we need to update this?

core/felt/hash.go
rodrodros, line 1: 👍

core/version.go
EgeCaner: Node was syncing fine without it, I guess I can remove

core/state_update.go
EgeCaner: This is inside `digest.Update` , I guess we only have call overhead. WDYT?

```go
func (d *PoseidonDigest) Update(elems ...*felt.Felt) Digest {
	for idx := range elems {
		if d.lastElem == nil {
			d.lastElem = new(felt.Felt).Set(elems[idx])
		} else {
			d.state[0].Add(&d.state[0], d.lastElem)
			d.state[1].Add(&d.state[1], elems[idx])
			HadesPermutation(d.state[:])
			d.lastElem = nil
		}
	}
	return d
}
```

db/schema.go
```diff
@@ -175,3 +175,8 @@ func uint64ToBytes(num uint64) [8]byte {
 func StateHashToTrieRootsKey(stateCommitment *felt.Felt) []byte {
 	return StateHashToTrieRoots.Key(stateCommitment.Marshal())
 }
+
+func ClassHashToCompiledClassHashV1Key(blockNum uint64, classHash *felt.Felt) []byte {
```
rodrodros: Let's update the naming to

```golang
SierraClassHashToDeprecatedCasmHash(blokcNum uint64, classHash *felt.SierraClassHash) []byte {
```

What do you think?

db/buckets.go
```diff
@@ -38,16 +38,17 @@ const (
...
+	ContractTrieContract           // ContractTrieContract + nodetype + path + pathlength -> Trie Node
+	ContractTrieStorage            // ContractTrieStorage + nodetype + path + pathlength -> Trie Node
+	Contract                       // Contract + ContractAddr -> Contract
+	StateHashToTrieRoots           // StateHash -> ClassRootHash + ContractRootHash
+	StateID                        // StateID + root hash -> state id
+	PersistedStateID               // PersistedStateID -> state id
+	TrieJournal                    // TrieJournal -> journal
+	AggregatedBloomFilters         // maps block range to AggregatedBloomFilter
+	RunningEventFilter             // aggregated filter not full yet
+	ClassHashToCompiledClassHashV1 // SierraClassHash to CompiledClassHashV1 (starknet v0.14.1)
```
rodrodros: Update the naming of this constant as well please

core/state.go
```diff
@@ -785,3 +839,30 @@ func (s *State) performStateDeletions(blockNumber uint64, diff *StateDiff) error
 
 	return nil
 }
+
+func (s *State) revertMigratedCompiledClasses(
+	blockNumber uint64,
+	migrateCompiledClasses map[felt.SierraClassHash]felt.CasmClassHash,
```
rodrodros: migratedCasmClasses

core/state.go
```diff
@@ -785,3 +839,30 @@ func (s *State) performStateDeletions(blockNumber uint64, diff *StateDiff) error
...
+) error {
+	classesTrie, classesCloser, err := s.classesTrie()
+	if err != nil {
+		return err
+	}
+
+	for classHash := range migrateCompiledClasses {
+		classHashFelt := (*felt.Felt)(&classHash)
+		compiledCasmHashV1, err := GetCompiledClassHashV1(s.txn, blockNumber, classHashFelt)
```
rodrodros: deprecatedCasmHash

core/state.go
```diff
@@ -550,6 +558,47 @@ func (s *State) updateDeclaredClassesTrie(
...
+	var l1HeadNumber uint64
+	if len(migrateCompiledClasses) > 0 {
+		l1Head, err := GetL1Head(s.txn)
+		if err != nil {
+			return err
+		}
+		l1HeadNumber = l1Head.BlockNumber
+	}
+
+	for classHash, compiledClassHash := range migrateCompiledClasses {
```
rodrodros: `compiledClassHash` to `casmClassHash` please

core/state.go
```diff
@@ -531,8 +537,10 @@ func calculateContractCommitment(storageRoot, classHash, nonce *felt.Felt) *felt
 }
 
 func (s *State) updateDeclaredClassesTrie(
+	blockNumber uint64,
 	declaredClasses map[felt.Felt]*felt.Felt,
 	classDefinitions map[felt.Felt]ClassDefinition,
+	migrateCompiledClasses map[felt.SierraClassHash]felt.CasmClassHash,
```
rodrodros: `migratedCasmClasses` please 🙏

core/accessors.go
rodrodros, line 1: Let's update this accessors to with the `Sierra` and `Casm` notation 🙏

blockchain/blockchain.go
```diff
@@ -227,7 +227,12 @@ func (b *Blockchain) L1Head() (core.L1Head, error) {
 
 func (b *Blockchain) SetL1Head(update *core.L1Head) error {
 	b.l1HeadFeed.Send(update)
-	return core.WriteL1Head(b.database, update)
+	err := core.WriteL1Head(b.database, update)
+	if err != nil {
+		return err
+	}
+
+	return core.DeleteDeprecatedCasmHashUpToBlockNum(b.database, update.BlockNumber)
```
rodrodros: Love this long but descriptive name 🫶

core/felt/hash.go
```diff
@@ -22,12 +22,28 @@ func (h *SierraClassHash) String() string {
...
 	return (*ClassHash)(h).String()
 }
 
+func (h *CasmClassHash) UnmarshalJSON(data []byte) error {
+	return (*Felt)(h).UnmarshalJSON(data)
+}
+
+func (h *CasmClassHash) MarshalJSON() ([]byte, error) {
+	return (*Felt)(h).MarshalJSON()
+}
```
rodrodros, line 61: I am thinking that maybe the should call the underlying implementation of `ClassHash`, since they are "children" of it

core/accessors.go
```diff
@@ -552,3 +552,71 @@ func WriteRunningEventFilter(w db.KeyValueWriter, filter *RunningEventFilter) er
 
 	return w.Put(db.RunningEventFilter.Key(), enc)
 }
+
+// GetDeprecatedCasmHash retrieves the deprecated casm hash for
+// a given class hash and block number.
+func GetDeprecatedCasmHash(
+	r db.KeyValueReader,
+	blockNum uint64,
+	classHash *felt.Felt,
```
rodrodros: What do you think if we start using the new types from now on, otherwise we would never do it :( 

I think classHash should be `*felt.SierraClassHash`

core/accessors.go
```diff
@@ -552,3 +552,71 @@ func WriteRunningEventFilter(w db.KeyValueWriter, filter *RunningEventFilter) er
...
+) (felt.Felt, error) {
+	var compiledClassHash felt.Felt
```
rodrodros: And I think we could do the same here, `var casmHash felt.CasmClassHash`

core/accessors.go
```diff
@@ -552,3 +552,71 @@ func WriteRunningEventFilter(w db.KeyValueWriter, filter *RunningEventFilter) er
...
+func WriteDeprecatedCasmHash(
+	w db.KeyValueWriter,
+	blockNum uint64,
+	classHash,
+	compiledClassHash *felt.Felt,
+) error {
+	return w.Put(
+		db.SierraClassHashToDeprecatedCasmHashKey(blockNum, classHash),
+		compiledClassHash.Marshal(),
+	)
```
rodrodros: Similar comment here. I know there is a conversion to felt necessarily after, but look at it as multiple phase program where we:

1. Introduce the new types
2. We start them using gradually, initially there is a lot of casting to avoid intensive refactoring. This casting can be annoying, but if we don't start doing it little by little we increase the code that uses `felt.Felt`.
3. Eventually already written methods that used `felt.Felt` will have the right type associated, when this happens we remove the castings that we added in step 2
4. There will be almost no `felt.Felt` usage.

I will start adding this type of comments in any open PR so we all start applying this.

core/accessors.go
```diff
@@ -552,3 +552,71 @@ func WriteRunningEventFilter(w db.KeyValueWriter, filter *RunningEventFilter) er
...
+}
+
+func DeleteDeprecatedCasmHash(
+	w db.KeyValueWriter,
+	blockNum uint64,
+	classHash *felt.Felt,
```
rodrodros: Please use `*felt.SierraClassHash` here

core/accessors.go
```diff
@@ -552,3 +552,71 @@ func WriteRunningEventFilter(w db.KeyValueWriter, filter *RunningEventFilter) er
...
+	return w.DeleteRange(start, end)
+}
+
+// DeleteDeprecatedCasmHashFromBlockNum deletes all deprecated casm hash mappings
+// where blockNum >= given fromBlock. Useful for pruning data from a specific block onwards.
+// This should be called when you want to remove all data from a certain block number to the end.
+func DeleteDeprecatedCasmHashFromBlockNum(w db.KeyValueRangeDeleter, fromBlock uint64) error {
+	start := db.SierraClassHashToDeprecatedCasmHash.Key(MarshalBlockNumber(fromBlock))
+
+	maxBlockNum := ^uint64(0) // max uint64
```
rodrodros: Just in case you didn't know, there is a `math.MaxUint64` constant, but I am happy with this cool initialization as well

core/state_update.go
```diff
@@ -233,11 +233,13 @@ func declaredClassesDigest(
...
-	sortedDeclaredV1ClassHashes := sortedFeltKeys(declaredClasses)
-	for _, classHash := range sortedDeclaredV1ClassHashes {
-		digest.Update(&classHash, declaredClasses[classHash])
+	for _, classHash := range keys {
+		if casmHash, ok := declaredClasses[classHash]; ok {
+			digest.Update(&classHash, casmHash)
+		} else {
+			casmHash := migratedClasses[felt.SierraClassHash(classHash)]
+			digest.Update(&classHash, (*felt.Felt)(&casmHash))
+		}
```
rodrodros, line 242: 👍

core/accessors.go
EgeCaner: I agree we should start using them, let me update

core/state.go
```diff
@@ -577,10 +592,19 @@ func (s *State) Revert(blockNumber uint64, update *StateUpdate) error {
 		return fmt.Errorf("verify state update root: %v", err)
 	}
 
-	if err = s.removeDeclaredClasses(blockNumber, update.StateDiff.DeclaredV0Classes, update.StateDiff.DeclaredV1Classes); err != nil {
+	if err = s.removeDeclaredClasses(
+		blockNumber,
+		update.StateDiff.DeclaredV0Classes,
+		update.StateDiff.DeclaredV1Classes,
+	); err != nil {
```
rodrodros: nitpick. I am aware it's old code but let's not make complex expressions, such as this assign (not easy to read) part of the `if` statement

PR #3230, opened 2025-10-31
refactor(vm/rust): break jsonrpc.rs into ffi_types/... module

rodrodros:


vm/rust/src/ffi_type/transaction_trace.rs
```diff
@@ -0,0 +1,174 @@
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    fee_transfer_invocation: Option<FunctionInvocation>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    constructor_invocation: Option<FunctionInvocation>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    function_invocation: Option<ExecuteInvocation>,
+    r#type: TransactionType,
+    state_diff: StateDiff,
+}
+pub fn new_transaction_trace(
```
infrmtcs, line 62: Maybe in another PR, I think we should put this in `impl TransactionTrace` and rename it to `new`

rodrodros: Yes, I 100% agree. First I just want to move things around a bit and then move on to those changes

vm/rust/src/lib.rs
```diff
@@ -1,10 +1,18 @@
...
+
 mod juno_state_reader;
 
-use crate::juno_state_reader::{ptr_to_felt, BlockHeight, JunoStateReader};
+use crate::{
+    ffi_type::{
+        state_diff::StateDiff,
+        transaction_receipt::TransactionReceipt,
+        transaction_trace::{new_transaction_trace, TransactionTrace},
+    },
```
infrmtcs, line 13: I think it would be nicer and less redundant if we could do:
```suggestion
    ffi_type::{StateDiff, TransactionReceipt, TransactionTrace},
```
In order to do this, we can re-export the types in `lib.rs` in `ffi_type` by `pub use`.

infrmtcs (approved):

vm/rust/src/lib.rs
rodrodros: I initially did it, but then reverted it because I don't think it is good the long term. This is a library meant only for us and having specific types exported (done right!) it's a bit of extra maintenance and it can go wrong and cause more issues later if not taken care the right way.

I think is better to pay the price of redundancy and be extra clear and since is the default behaviour it cannot go wrong. 

If we were developing a library for users other than us I agree that we should put that extra effort into exporting the right types.

PR #3231, opened 2025-10-31
refactor(vm/rust): split the error module

rodrodros:


vm/rust/src/error/stack.rs
```diff
@@ -6,6 +6,31 @@ use blockifier::transaction::errors::TransactionExecutionError;
...
+
+impl From<Cairo1RevertFrame> for Frame {
+    fn from(value: Cairo1RevertFrame) -> Self {
+        Self::CallFrame(CallFrame {
+            storage_address: *value.contract_address.0,
+            class_hash: value.class_hash.unwrap_or_default().0,
+            selector: Some(value.selector.0),
+        })
+    }
+}
```
infrmtcs, line 30: `impl` should go after type declaration.

vm/rust/src/error/stack.rs
```diff
@@ -6,6 +6,31 @@ use blockifier::transaction::errors::TransactionExecutionError;
...
+
+#[derive(Clone, Debug, PartialEq)]
+pub enum Frame {
+    CallFrame(CallFrame),
+    StringFrame(String),
+}
+
+struct Frames(pub Vec<Frame>);
```
infrmtcs: type declaration should go right before its `impl`

vm/rust/src/error/juno.rs
```diff
@@ -0,0 +1,23 @@
+pub struct JunoError {
```
infrmtcs, line 1: nit: Personally I think this name is more generic than the `VMError` name.

rodrodros: I agree that Juno might not be as expressive, the idea was to represent it as the error, Juno receives from the blockifier.  Definitely something to improve upon.

PR #3232, opened 2025-11-03
Refactor: Refactored sn2core adapter

ndatta-nethermind:


adapters/sn2core/sn2core.go
```diff
@@ -19,19 +19,19 @@ func AdaptBlock(response *starknet.Block, sig *starknet.Signature) (*core.Block,
 	}
 
 	txns := make([]core.Transaction, len(response.Transactions))
-	for i, txn := range response.Transactions {
+	for i := range response.Transactions {
 		var err error
-		txns[i], err = AdaptTransaction(txn)
+		txns[i], err = AdaptTransaction(response.Transactions[i])
```
infrmtcs, line 24: This doesn't really optimize anything.

adapters/sn2core/sn2core.go
```diff
@@ -19,19 +19,19 @@ func AdaptBlock(response *starknet.Block, sig *starknet.Signature) (*core.Block,
...
 		}
 	}
 
 	receipts := make([]*core.TransactionReceipt, len(response.Receipts))
 	eventCount := uint64(0)
-	for i, receipt := range response.Receipts {
-		receipts[i] = AdaptTransactionReceipt(receipt)
-		eventCount += uint64(len(receipt.Events))
+	for i := range response.Receipts {
+		receipts[i] = AdaptTransactionReceipt(response.Receipts[i])
```
infrmtcs, line 33: This doesn't really optimize anything.

adapters/sn2core/sn2core.go
```diff
@@ -270,6 +270,32 @@ func AdaptDeployAccountTransaction(t *starknet.Transaction) *core.DeployAccountT
 	}
 }
 
+func adaptSierraEntrypoint(ep *starknet.SierraEntryPoint) core.SierraEntryPoint {
+	return core.SierraEntryPoint{Index: ep.Index, Selector: ep.Selector}
```
infrmtcs: nit: There are more than 1 field, so I think splitting lines is easier to read.

adapters/sn2core/sn2core.go
```diff
@@ -389,13 +405,12 @@ func AdaptStateUpdate(response *starknet.StateUpdate) (*core.StateUpdate, error)
...
-		StateDiff: stateDiff,
+		StateDiff: &stateDiff,
 	}, nil
 }
 
-// todo(rdr): return `core.StateDiff` by value
-func AdaptStateDiff(response *starknet.StateDiff) (*core.StateDiff, error) {
-	stateDiff := new(core.StateDiff)
+func AdaptStateDiff(response *starknet.StateDiff) (core.StateDiff, error) {
+	stateDiff := core.StateDiff{} //nolint:exhaustruct // Fields are initiated later
```
infrmtcs:
```suggestion
	var stateDiff core.StateDiff
```

adapters/sn2core/sn2core.go
ndatta-nethermind: Yes. Rewritten in this format to stick with index-based iteration. It optimizes when the `response.Transaction` is a value type.

adapters/sn2core/sn2core.go
ndatta-nethermind: Yes. Rewritten in this format to stick with index-based iteration. It optimizes when the `response.Receipts` is a value type.

adapters/sn2core/sn2core.go
```diff
@@ -472,7 +490,7 @@ func AdaptPreConfirmedBlock(
 		if adaptedStateDiff, err = AdaptStateDiff(stateDiff); err != nil {
 			return core.PreConfirmed{}, err
 		}
-		txStateDiffs = append(txStateDiffs, adaptedStateDiff)
+		txStateDiffs = append(txStateDiffs, &adaptedStateDiff)
```
infrmtcs: This is incorrect and will be pretty hard to debug because all elements in `txStateDiffs` will be the same, which is the pointer to the `adaptedStateDiff` declared above. Any mutation changing `adaptedStateDiff` such as line 490 will change the common variable.

infrmtcs: @EgeCaner I think we should assert the content of `adapted.TransactionStateDiffs` instead of just the size.
https://github.com/NethermindEth/juno/blob/main/adapters/sn2core/sn2core_test.go#L681

core/class.go
```diff
@@ -71,11 +71,7 @@ type SierraClass struct {
 	AbiHash *felt.Felt
 	// TODO: will implement this on a follow up PR commit to avoid the migration
 	// EntryPoints     SierraEntryPointsByType
-	EntryPoints struct {
-		Constructor []SierraEntryPoint
-		External    []SierraEntryPoint
-		L1Handler   []SierraEntryPoint
-	}
+	EntryPoints SierraEntryPointsByType
```
infrmtcs, line 72: Please remove the comments in the 2 lines above

adapters/sn2core/sn2core.go
EgeCaner: @infrmtcs I agree test should be more robust

EgeCaner: @infrmtcs test is getting updated in this [PR](https://github.com/NethermindEth/juno/pull/3233), it will now asserts for its content

adapters/sn2core/sn2core.go
```diff
@@ -398,9 +417,8 @@ func AdaptStateUpdate(response *starknet.StateUpdate) (*core.StateUpdate, error)
 	}, nil
 }
 
-// todo(rdr): return `core.StateDiff` by value
```
infrmtcs, line 401: I'm understanding that this means the return type of the function should be `(core.StateDiff, error)` instead.

ndatta-nethermind: Fixed it.,

adapters/sn2core/sn2core.go
rodrodros: It is true it doesn't because it is a reference type. I think is ok to leave the code this way nonetheless

adapters/sn2core/sn2core.go
rodrodros: It is true it doesn't because it is a reference type. I think is ok to leave the code this way as well

PR #3233, opened 2025-11-03
fix(sn2core): Assume random order of preconfirmed txns

brbrr:


adapters/sn2core/sn2core.go
```diff
@@ -448,6 +448,10 @@ func AdaptStateDiff(response *starknet.StateDiff) (*core.StateDiff, error) {
 	return stateDiff, nil
 }
 
+func IsCandidateTx(response *starknet.PreConfirmedBlock, id int) bool {
+	return response.TransactionStateDiffs[id] == nil || response.Receipts[id] == nil
```
rodrodros, line 455: Can you add a small comment why CandidateTxs in Starknet are identified by having either of the fields nil

adapters/sn2core/sn2core.go
```diff
@@ -460,46 +464,49 @@ func AdaptPreConfirmedBlock(
...
-	for _, stateDiff := range response.TransactionStateDiffs {
-		if stateDiff == nil {
-			break
-		}
-
-		if adaptedStateDiff, err = AdaptStateDiff(stateDiff); err != nil {
-			return core.PreConfirmed{}, err
-		}
-		txStateDiffs = append(txStateDiffs, adaptedStateDiff)
+	if len(response.Transactions) != len(response.TransactionStateDiffs) || len(response.Transactions) != len(response.Receipts) {
```
rodrodros: Valid linter error here. To be inline with go style please create a boolean var before as

```golang
boolVar := <condition>
if boolVar {
   ... 
}
```

This is good because the boolean var name helps the reader identify its intention and at the same time we avoid the long line or having 
```golang
  if smth || other_something || and || so || on || lalala &&
          then || indentation || which is || not && go < friendly {
   ...
```

adapters/sn2core/sn2core.go
```diff
@@ -460,46 +464,49 @@ func AdaptPreConfirmedBlock(
...
 	eventCount := uint64(0)
-	for i, receipt := range response.Receipts[:preConfirmedTxCount] {
-		receipts[i] = AdaptTransactionReceipt(receipt)
-		eventCount += uint64(len(receipt.Events))
+	candidateTxs := make([]core.Transaction, candidateCount)
+
+	var err error
+	preIdx := 0
+	candIdx := 0
+
```
rodrodros: Nitpick: usually if this above vars (lines 485, 476 & 487) are meant to interact with the `for` below the should be no white space here

adapters/sn2core/sn2core.go
```diff
@@ -460,46 +464,49 @@ func AdaptPreConfirmedBlock(
...
+	for i := range len(response.Transactions) {
+		if !IsCandidateTx(response, i) {
+			txns[preIdx], err = AdaptTransaction(&response.Transactions[i])
+			if err != nil {
+				return core.PreConfirmed{}, err
+			}
+
+			if txStateDiffs[preIdx], err = AdaptStateDiff(response.TransactionStateDiffs[i]); err != nil {
```
rodrodros: Nitpick: Please separate the variable assignation from the condition of the `if` statement. Even if the notation is valid, having complex (where complex is subjective) expressions here hurt readability. I think is best to use when you have things that only return an error and they are not too long and easy to read.

adapters/sn2core/sn2core_test.go
```diff
@@ -670,18 +675,29 @@ func TestAdaptPreConfirmed(t *testing.T) {
...
 		expectedEventCount, expectedPreConfirmedTxCount := countEventsAndTxs(response.Receipts)
 		expectedCandidateCount := len(response.Transactions) - expectedPreConfirmedTxCount
+		expectedReceipts := getPreconfirmedReceipts(response.Receipts)
 
 		adapted, err := sn2core.AdaptPreConfirmedBlock(response, test.blockNumber)
 		require.NoError(t, err)
 
 		assertPreConfirmedBlockBasics(t, &adapted, test.blockNumber, response, expectedPreConfirmedTxCount, expectedCandidateCount, expectedEventCount)
-		assertPreConfirmedBlockReceipts(t, response.Receipts, adapted.Block.Receipts, expectedPreConfirmedTxCount)
+		assertPreConfirmedBlockReceipts(t, expectedReceipts, adapted.Block.Receipts, expectedPreConfirmedTxCount)
```
rodrodros: Linter warning here

rodrigo-pino (commented):
Looks ver good! Some minor comments!

adapters/sn2core/sn2core.go
brbrr: tbh, I'm not sure why, except it's by design/according to spec 🤷

EgeCaner: For context CANDIDATE transaction is a candidate to be executed soon but not executed yet thus cannot have receipt and state diff. For CANDIDATE transactions both should be always nil. For Preconfirmed all should be present, any other combination should be invalid such as there is receipt but no state diff or vice versa.

adapters/sn2core/sn2core_test.go
```diff
@@ -711,16 +741,35 @@ func assertPreConfirmedBlockBasics(
...
+		}
+	}
+}
+
+func assertStateDiffs(
+	t *testing.T,
+	response *starknet.PreConfirmedBlock,
+	stateDiffs []*core.StateDiff,
+) {
+	preID := 0
```
EgeCaner, line 768:
```suggestion
    t.Helper()

	preID := 0
```

adapters/sn2core/sn2core_test.go
```diff
@@ -711,16 +741,35 @@ func assertPreConfirmedBlockBasics(
...
 	candidateTxs []core.Transaction,
-	offset int,
 ) {
 	t.Helper()
 
-	for i, txn := range candidateTxs {
-		adaptedTx, err := sn2core.AdaptTransaction(&transactions[offset+i])
-		require.NoError(t, err)
-		assert.Equal(t, adaptedTx, txn)
+	candID := 0
```
EgeCaner, line 749: Should we also check if all transcations in candidateTxs is covered? Previously we were iterating over candidateTxs list thus ensuring that all of them are covered

adapters/sn2core/sn2core_test.go
EgeCaner, line 768: Should we also check if all stateDiffs are covered by asserting the PreID reaches end of list?

adapters/sn2core/sn2core_test.go
brbrr: Good points! Addressed

adapters/sn2core/sn2core.go
```diff
@@ -460,46 +464,52 @@ func AdaptPreConfirmedBlock(
...
-
-		if adaptedStateDiff, err = AdaptStateDiff(stateDiff); err != nil {
-			return core.PreConfirmed{}, err
-		}
-		txStateDiffs = append(txStateDiffs, adaptedStateDiff)
-	}
-
-	preConfirmedTxCount := len(txStateDiffs)
+	isInvalidPayloadSizes := len(response.Transactions) != len(response.TransactionStateDiffs) ||
+		len(response.Transactions) != len(response.Receipts)
```
rodrodros: nitpick: remove whiteline between var declaration and `if`

PR #3234, opened 2025-11-03
Refactor: Rename compiled to casm in sierraclass

ndatta-nethermind:


adapters/sn2core/sn2core.go
```diff
@@ -300,7 +300,7 @@ func AdaptSierraClass(
 		Abi:     response.Abi,
 		AbiHash: crypto.StarknetKeccak([]byte(response.Abi)),
 
-		Compiled: coreCompiledClass,
+		Casm: coreCompiledClass,
```
brbrr: note: we might want to rename the variables related to the types as well, as it might be a bit confusing long term

brbrr (approved):
Aside from a small nitpick it looks good!

core/class.go
```diff
@@ -80,7 +80,7 @@ type SierraClass struct {
 	ProgramHash *felt.Felt
 	// TODO: Remove this semantic version on a follow up PR. Let's put Sierra version instead
 	SemanticVersion string
-	Compiled        *CasmClass
+	Casm            *CasmClass
```
infrmtcs, line 83: Please do not rename this field:
- `SierraClass` implements `ClassDefinition`.
- `ClassDefinition` is the type of a field in `DeclaredClassDefinition`
- `DeclaredClassDefinition` is stored in bucket 4 (`Class`) in database.
=> This type is stored in the database, while the current CBOR encoding also serializes the field name, so renaming it cause backward incompatibility.

rodrodros: @ndatta-nethermind please add a note here specifying this. Restore all the other changes in this PR and leave just that

ndatta-nethermind: > Please do not rename this field:
> 
> * `SierraClass` implements `ClassDefinition`.
> * `ClassDefinition` is the type of a field in `DeclaredClassDefinition`
> * `DeclaredClassDefinition` is stored in bucket 4 (`Class`) in database.
>   => This type is stored in the database, while the current CBOR encoding also serializes the field name, so renaming it cause backward incompatibility.

Closing the PR for the reason mentioned above.

PR #3237, opened 2025-11-04
refactor(vm/rust): split lib.rs into multiple modules

rodrodros:


vm/rust/src/entrypoint/execute/process/utils.rs
```diff
@@ -0,0 +1,89 @@
...
+    // L2 gas accounting is enabled if:
+    // 1. The gas computation mode requires all gas vectors.
+    // 2. The sender contract's tracked resource is Sierra gas.
+    Ok(
+        matches!(gas_computation_mode, GasVectorComputationMode::All)
+            && sender_tracked_resource == TrackedResource::SierraGas,
+    )
+}
+
+pub fn determine_gas_vector_mode(transaction: &Transaction) -> GasVectorComputationMode {
```
EgeCaner, line 69: This is also implemented in `/execute/utils.rs` do we need both?

vm/rust/src/entrypoint/execute/utils.rs
```diff
@@ -0,0 +1,248 @@
...
+        tx_hash,
+        class_info,
+        paid_fee_on_l1,
+        None,
+        execution_flags,
+    )
+    .map_err(|err| format!("failed to create transaction from api: {:?}", err))
+}
+
+pub fn determine_gas_vector_mode(transaction: &Transaction) -> GasVectorComputationMode {
```
EgeCaner: This is also implemented in /execute/process/utils.rs do we need both?

EgeCaner (approved):
LGTM!

PR #3251, opened 2025-11-10
chore: Fix linter

infrmtcs:
`.golangci.yaml` is enforcing line length 140 while `.golangci_diff.yaml` is enforcing line length 100.

We have a line (containing an URL) with length 120, so:
- Not adding `nolint` will be complained by `.golangci_diff.yaml`
- Adding `nolint` will be complained by `.golangci.yaml`

The temporary workaround is to get this forced merged to `main`, because `.golangci.yaml` always checks while `.golangci_diff.yaml` only checks PR changes.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -77,13 +77,6 @@ linters:
     govet:
       enable:
         - nilness
-      settings:
-        printf:
-          funcs:
-            - (github.com/golangci/golangci-lint/pkg/logutils.Log).Infof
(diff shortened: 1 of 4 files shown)


.golangci.yaml
```diff
@@ -141,45 +132,11 @@ linters:
       - linters:
           - lll
         source: '^//go:generate '
-      - path: pkg/golinters/errcheck.go
```
infrmtcs, line 144: These are unrelated settings pasted from golangci-lint repo config file.

MaksymMalicki (approved):

PR #3253, opened 2025-11-10
refactor(state): simplify the state in `core` package

MaksymMalicki:
This PR:
- moves the logic (the `Read...At` methods) from `history.go` into the `state.go`.
- For the historical CRUDs, the accessors pattern is used, instead of direct, unstandardised reads and writes to the batch (previously using `logDBKey`).
- some minor refactors querying the chain height using `BlockchainReader` instead of `StateReader`


core/accessors.go
rodrodros, line 1: This file will probably fail due to linting issues. 

Is it possible to use the new types (address, class hash and so on ...) over this accessors?

core/history.go
```diff
@@ -1,128 +1,108 @@
...
-	return binary.BigEndian.AppendUint64(key, height)
-}
-
-func (h *history) logOldValue(key, value []byte, height uint64) error {
-	return h.txn.Put(logDBKey(key, height), value)
+func NewStateHistory(state StateHistoryReader, blockNumber uint64) StateReader {
+	return &stateHistory{
+		blockNumber: blockNumber,
+		state:       state,
+	}
```
rodrodros: nitpick: this could be returned by value, it is just a `uint64` and a `*ptr`

rodrigo-pino (commented):
Some comments

core/state_test.go
```diff
@@ -321,6 +321,96 @@ func TestStateHistory(t *testing.T) {
 		})
 }
 
+func TestHistory(t *testing.T) {
+	testDB := memory.New()
+	txn := testDB.NewIndexedBatch()
+
+	state := core.NewState(txn)
+	contractAddress := new(felt.Felt).SetUint64(123)
```
EgeCaner: nit: Can we switch to initializing felts with felt constructors? felt.FromUint64 etc..

core/state.go
```diff
@@ -788,29 +783,105 @@ func (s *State) performStateDeletions(blockNumber uint64, diff *StateDiff) error
...
 			return err
 		}
 	}
 
 	return nil
 }
 
+func historyDBKey(key []byte, height uint64) []byte {
+	return binary.BigEndian.AppendUint64(key, height)
+}
```
EgeCaner: Should this reside in db/schema.go?

core/state.go
```diff
@@ -788,29 +783,105 @@ func (s *State) performStateDeletions(blockNumber uint64, diff *StateDiff) error
...
+
+func (s *State) valueAt(key []byte, height uint64) ([]byte, error) {
```
EgeCaner, line 855: Can we make this a function in accessors?

core/state.go
MaksymMalicki: fair point, removed this entirely and unified the accessors between the states even more

core/state_test.go
MaksymMalicki: changed

core/state.go
MaksymMalicki: Why? I think it makes more sense here, since its strictly coupled with the `Contract..At` logic

core/history.go
MaksymMalicki: changed

core/accessors.go
MaksymMalicki: updated to pass the linter

db/schema.go
```diff
@@ -119,17 +119,17 @@ func ContractKey(addr *felt.Felt) []byte {
 	return Contract.Key(addr.Marshal())
 }
 
-func ContractHistoryNonceKey(addr *felt.Felt, blockNum uint64) []byte {
+func ContractNonceHistoryAtBlockKey(addr *felt.Felt, blockNum uint64) []byte {
```
infrmtcs, line 122: nit: Do we need to rename this?

core/history.go
```diff
@@ -1,128 +1,108 @@
...
 	"github.com/NethermindEth/juno/db"
-	"github.com/NethermindEth/juno/utils"
 )
 
-var ErrCheckHeadState = errors.New("check head state")
+var ErrHistoricalTrieNotSupported = errors.New("cannot support historical trie")
 
-type history struct {
-	txn db.IndexedBatch
+type stateHistory struct {
```
infrmtcs: Initially, it looks like you're removing state snapshot and use state history instead. But instead, I think you're actually removing state history, replace it with state snapshot implementation and rename to state history. Please next time, break this into 2 PRs:
- Remove state history, replace by the existing state snapshot
- Rename state snapshot to state history

We can actually achieve the same outcomes with much simpler diff.

core/state/history.go
```diff
@@ -7,27 +7,27 @@ import (
 	"github.com/NethermindEth/juno/db"
 )
 
-var _ StateReader = (*StateHistory)(nil)
+var _ StateReader = (*stateHistory)(nil)
 
 // StateHistory represents a snapshot of the blockchain state at a specific block number.
-type StateHistory struct {
+type stateHistory struct {
```
infrmtcs, line 13: Can we remove this completely? Because:
- This is currently unused
- We're already having another state history
- The implementation is not correct, because it needs to handle `ErrCheckHeadState`

core/state.go
infrmtcs: We may, just probably in another PR. Except the error type, this is basically an implementation of finding the key preceding a given key, which is quite generic in my opinion.

infrmtcs: However, this can be done much better, because the current abstraction `NewIterator` is pretty bad. It doesn't provide a lot features that are provided by pebble:
- Missing `Last()`, which prevent reverse iteration
- Customized range instead of by prefix only, like (key, 0) -> (key, height - 1), which is what we need in this case.
- `SeekLT`, which is exactly what we need here instead of `SeekGE` used by `it.Seek`. This is why we have to check `if seekedHeight < height ... else if seekedHeight == height ...`

core/state/history.go
MaksymMalicki: I'd say we shouldn't remove it now. The internal handling of the historical reads is slightly different. I have to revisit the Iterator abstraction, and maybe unify the reads in a separate PR

core/state.go
MaksymMalicki: fair point, I'll revisit this logic in a separate PR

db/schema.go
MaksymMalicki: This allows us to do two things at once:
1. Unify the schema for accessors for the new and the old state
2. Get rid of `func logDBKey(key []byte, height uint64)` in the history, which internally was doing exactly this (appending the block number to the key)

core/accessors.go
MaksymMalicki: Regarding the new types, I think its a good idea, but it should be done in a separate PR. Would it be a good moment for a general `core` package refactor to use the proper types? 🤔

core/history.go
MaksymMalicki: done!

infrmtcs (approved):

PR #3262, opened 2025-11-12
feat: casm hash v2

EgeCaner:


core/class_test.go
```diff
@@ -267,8 +281,14 @@ func TestSegmentedBytecodeHash(t *testing.T) {
 			},
 		},
 	}
+
+	hasher := core.NewCasmHasher(core.HashVersionV1)
```
infrmtcs: Do we need to add tests for V2?

infrmtcs (approved):

core/crypto/blake2s/hash.go
```diff
@@ -45,3 +47,34 @@ func Blake2sArray[F felt.FeltLike](feltLikes ...*F) felt.Hash {
...
+	encoding := encodeFeltsToBytes(elems...)
+	_, err := d.hasher.Write(encoding)
+	if err != nil {
+		panic(err)
+	}
+	return d
+}
+
+func (d *Blake2sDigest) Finish() felt.Felt {
+	result := make([]byte, 0, 32)
```
infrmtcs, line 75:
```suggestion
	var result []byte
```
to avoid unnecessary allocation.

infrmtcs: I read the wrong docs :(

PR #3264, opened 2025-11-14
feat: pre compute casm class hash v2

EgeCaner:
Introduces new bucket `ClassHashToCasmHashV2` to serve as a cache for reading casm hashes in new state reader API `get_compiled_class_hash_v2`. While syncing new blocks Juno computes and stores blake2s casm hash of every declared class for blocks pre starknet v0.14.1 and stores blake2s hash directly from state diff if block is >= starknet v0.14.1.

## This PR includes migration

### DIFF
--- a/migration/migration.go
+++ b/migration/migration.go
@@ -954,50 +954,42 @@ func calculateCasmClassHashesV2(txn db.IndexedBatch, network *utils.Network) err
 		classHash := felt.FromBytes[felt.Felt](classHashBytes)
 		sierraClassHash := felt.SierraClassHash(classHash)
 
-		// Check if already has CASM hash V2
-		writeMu.Lock()
-		_, err := core.GetCasmClassHashV2(txn, &sierraClassHash)
-		writeMu.Unlock()
-		if err == nil {
-			continue // Already exists, skip
-		}
-		if !errors.Is(err, db.ErrKeyNotFound) {
-			return err
-		}
-
-		// Read class value
 		value, err := it.Value()
 		if err != nil {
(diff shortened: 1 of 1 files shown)


migration/migration.go
```diff
@@ -922,3 +923,82 @@ func startMigrationStatusServer(log utils.SimpleLogger, host string, port uint16
...
+		if len(classHashBytes) != felt.Bytes {
+			continue // Skip invalid keys
+		}
+		classHash := felt.FromBytes[felt.Felt](classHashBytes)
+		sierraClassHash := felt.SierraClassHash(classHash)
+
+		// Check if already has CASM hash V2
+		writeMu.Lock()
+		_, err := core.GetCasmClassHashV2(txn, &sierraClassHash)
+		writeMu.Unlock()
```
infrmtcs: I think if we use snapshot for read instead, we don't need this mutex.

infrmtcs: Another solution is run an iterator to load all class hashes from database first, build a map.

migration/migration.go
```diff
@@ -922,3 +923,82 @@ func startMigrationStatusServer(log utils.SimpleLogger, host string, port uint16
...
+		// Read class value
+		value, err := it.Value()
+		if err != nil {
+			return err
+		}
+
+		var declaredClass core.DeclaredClassDefinition
+		if err := encoder.Unmarshal(value, &declaredClass); err != nil {
+			continue // Skip classes that can't be unmarshaled
+		}
```
infrmtcs: This can also be done concurrently.

PR #3265, opened 2025-11-17
feat: blockifier upgrade and state reader api

EgeCaner:


core/pending_state.go
```diff
@@ -78,6 +78,21 @@ func (p *PendingState) Class(classHash *felt.Felt) (*DeclaredClassDefinition, er
 	return p.head.Class(classHash)
 }
 
+func (p *PendingState) CompiledClassHash(classHash *felt.Felt) (felt.Felt, error) {
```
rodrodros: Let's use `CasmClassHash`. I think the entry point should be called `*felt.SierraClassHash` and the return type should be `felt.CasmClassHash`

core/pending_state.go
```diff
@@ -78,6 +78,21 @@ func (p *PendingState) Class(classHash *felt.Felt) (*DeclaredClassDefinition, er
...
+	if casmHash, found := p.stateDiff.DeclaredV1Classes[*classHash]; found {
+		return *casmHash, nil
+	}
+	return p.head.CompiledClassHash(classHash)
+}
+
+func (p *PendingState) CompiledClassHashV2(classHash *felt.Felt) (felt.Felt, error) {
```
rodrodros: And here `CompiledCasmHashV2`? What do you think? 

Also, I would add of applying the same thoughts of the previous comment to the input and output of this function. What do you think?

core/state.go
```diff
@@ -367,6 +368,34 @@ func (s *State) Class(classHash *felt.Felt) (*DeclaredClassDefinition, error) {
 	return class, err
 }
 
+func (s *State) CompiledClassHash(classHash *felt.Felt) (felt.Felt, error) {
```
rodrodros: Same comment idea for this function

core/state.go
```diff
@@ -367,6 +368,34 @@ func (s *State) Class(classHash *felt.Felt) (*DeclaredClassDefinition, error) {
...
+	}
+
+	if casmHash.IsZero() {
+		return felt.Felt{}, errors.New("casm hash not found")
+	}
+
+	return casmHash, nil
+}
+
+func (s *State) CompiledClassHashV2(classHash *felt.Felt) (felt.Felt, error) {
```
rodrodros: And this function.

The goal is to have any new code we write use the right types instead of the generic `felt.Felt`

starknet/class.go
```diff
@@ -88,7 +88,7 @@ type CasmClass struct {
 	Hints                  json.RawMessage `json:"hints"`
 	PythonicHints          json.RawMessage `json:"pythonic_hints"`
 	CompilerVersion        string          `json:"compiler_version"`
-	BytecodeSegmentLengths SegmentLengths  `json:"bytecode_segment_lengths"`
+	BytecodeSegmentLengths *SegmentLengths `json:"bytecode_segment_lengths,omitempty"`
```
rodrodros, line 91: Why make it a pointer, as far as I can see, it is not `nil` checked against anything in the code?

rodrigo-pino (commented):
Looks good! Some comments

starknet/class.go
EgeCaner: As far as understood this field is optional, see `0x1338d85d3e579f6944ba06c005238d145920afeb32f94e3a1e234d21e1e9292.go`, and in blockifier side we use this `RunnableCompiledClass::try_from(class.contract_class).unwrap();` which internally fallbacks to using default values for `BytecodeSegmentLengths` for classes that does not contain the field. We were initializing it to zero value which makes internal assertion to fail. To solve this we should either implement the falling back to default value ourselves, but Im not sure side effects of it. Just passing the correct type rust expects which is `Option::None` would trigger the correct fallback logic.

rodrodros: Makes sense

core/state.go
```diff
@@ -367,6 +368,41 @@ func (s *State) Class(classHash *felt.Felt) (*DeclaredClassDefinition, error) {
...
+}
+
+func (s *State) CompiledClassHashV2(
+	classHash *felt.SierraClassHash,
+) (felt.CasmClassHash, error) {
+	casmHash, err := GetCasmClassHashV2(s.txn, classHash)
+	if err != nil {
+		return felt.CasmClassHash{}, err
+	}
+	return casmHash, nil
```
infrmtcs:
```suggestion
	return GetCasmClassHashV2(s.txn, classHash)
```

core/pending_state.go
```diff
@@ -78,6 +78,25 @@ func (p *PendingState) Class(classHash *felt.Felt) (*DeclaredClassDefinition, er
...
+	classHash *felt.SierraClassHash,
+) (felt.CasmClassHash, error) {
+	classHashFelt := felt.Felt(*classHash)
+	if casmHash, found := p.stateDiff.DeclaredV1Classes[classHashFelt]; found {
+		return felt.CasmClassHash(*casmHash), nil
+	}
+	return p.head.CompiledClassHash(classHash)
+}
+
+func (p *PendingState) CompiledClassHashV2(
```
rodrodros, line 91: What do you think of renaming `CompiledClassHash` to `CasmClassHash`?

EgeCaner: IMO, the current naming aligns better with the blockifier’s StateReader API, `get_compiled_class_hash(..)` [ref](https://github.com/starkware-libs/sequencer/blob/c2e2dff39f7105ef3adb622e7e38cf9a1ebbd996/crates/blockifier/src/state/state_api.rs#L46).

vm/rust/src/entrypoint/execute/utils.rs
```diff
@@ -47,15 +47,13 @@ pub fn transaction_from_api(
...
+                "Unsupported deploy transaction in the traced block (transaction_hash={tx_hash})",
             ))
         }
         StarknetApiTransaction::Declare(_) if class_info.is_none() => {
             return Err(format!(
-                "Declare transaction must be created with a ContractClass (transaction_hash={})",
-                tx_hash,
-            ))
+            "Declare transaction must be created with a ContractClass (transaction_hash={tx_hash})",
+        ))
```
infrmtcs, line 56: nit: Indentation looks strange

adapters/core2sn/class.go
```diff
@@ -28,9 +28,12 @@ func AdaptCasmClass(coreCasmClass *core.CasmClass) starknet.CasmClass {
 	feederCompiledClass.CompilerVersion = coreCasmClass.CompilerVersion
 	feederCompiledClass.Hints = coreCasmClass.Hints
 	feederCompiledClass.Prime = utils.ToHex(coreCasmClass.Prime)
-	feederCompiledClass.BytecodeSegmentLengths = AdaptSegmentLengths(
-		coreCasmClass.BytecodeSegmentLengths,
-	)
+
+	if coreCasmClass.BytecodeSegmentLengths.Length != 0 ||
```
infrmtcs, line 32: If `BytecodeSegmentLengths` can be `nil`, is it correct to store it as value instead of pointer?

PR #3280, opened 2025-11-20
fix: event matcher

### LINKED ISSUE #1922: Encounter potential bug while using get events with complex filter
Using https://free-rpc.nethermind.io/sepolia-juno endpoint

Encountered issue that if query event and filter by complex component key node will fire for me events that not satisfy filter.
I didn't manage to reproduce this behaviour over https://starknet-sepolia.public.blastapi.io/rpc/v0_7 so i presume the issue might be in node. Maybe once it finds block with satisfied events it just iterate over superset in this block not honouring original filter

Steps to reproduce:
https://sepolia.starkscan.co/tx/0x564f7fb5078f2f1ff708fbf41828d5f46d4494c2cde08274dc3aa32d65b0aca (we are interested in this block)

Filter (for indexed filed called taker)
```
    [
      [
        '0x22ea134d4126804c60797e633195f8c9aa5fd6d1567e299f4961d0e96f373ee'
      ],
      [
        '0x2e0a012a863e6b614014d113e7285b06e30d2999e42e6e03ba2ef6158b0a8f1'
      ],
      [],
      [
        '0x03765D53fdd82921fDDc888c04072fb618E6AC40eCAbc83084df6848a037a079'
      ]
    ]

```
Example of the event (https://sepolia.voyager.online/event/77850_310_10)

If i query, then it would additional add to response  events, emitted in this https://sepolia.starkscan.co/tx/0x564f7fb5078f2f1ff708fbf41828d5f46d4494c2cde08274dc3aa32d65b0aca transaction where taker of events is `0x059f71591a53fe7e7614292f36a6688e5290e1f73ad59be58c64b61c39aac23f` while we are filtering by 
`0x03765D53fdd82921fDDc888c04072fb618E6AC40eCAbc83084df6848a037a079`

EgeCaner:
# Problem

The `EventMatcher.MatchesEventKeys` function had a bug when encountering an empty filter position (which means "match any value") in the middle of a filter array. The code used break instead of continue, causing the matching loop to exit early and skip validation of subsequent filter positions.

Additionnaly `[[1]]` and `[[1], []]` were equavalent before. Now they represent 2 different queries
   - `[[1]]` Find all events with first key is 1.
   - `[[1], []]` Find all events with at least 2 keys and first key is 1.
 

# Impact
Example of incorrect behavior:
- Filter: [["V1"], [], ["V3"]]
- Event keys: ["V1", "anything", "WRONG"]

Expected: false (position 2 should be "V3" but is "WRONG")
Actual: true (loop stopped at position 1, never checked position 2)
This caused false positives where events that should not match the filter were incorrectly identified as matches.

Closes #1922

### DIFF
--- a/blockchain/event_matcher_test.go
+++ b/blockchain/event_matcher_test.go
@@ -102,7 +102,7 @@ func TestEventMatcher_MatchesEventKeys(t *testing.T) {
 			filterKeys: [][]felt.Felt{
 				{felt.FromUint64[felt.Felt](1)},
 				{felt.FromUint64[felt.Felt](2)},
-				{felt.FromUint64[felt.Felt](3)},
+				{},
 			},
 			eventKeys: []*felt.Felt{
 				felt.NewFromUint64[felt.Felt](1),
@@ -167,12 +167,6 @@ func TestEventMatcher_MatchesEventKeys(t *testing.T) {
 			filterKeys: [][]felt.Felt{{felt.FromUint64[felt.Felt](1)}},
 			eventKeys:  []*felt.Felt{},
 			expected:   false,
-		},
-		{
-			name:       "trailing empty filter positions are trimmed",
-			filterKeys: [][]felt.Felt{{felt.FromUint64[felt.Felt](1)}, {}, {}},
-			eventKeys:  []*felt.Felt{felt.NewFromUint64[felt.Felt](1)},
-			expected:   true,
 		},
 		{
 			name: "complex filter with OR logic",


blockchain/event_matcher.go
```diff
@@ -22,10 +22,23 @@ func NewEventMatcher(contractAddress *felt.Felt, keys [][]felt.Felt) EventMatche
...
+	for i := len(filterKeys) - 1; i >= 0; i-- {
+		if len(filterKeys[i]) > 0 {
+			lastNonEmpty = i
+			break
+		}
+	}
+
+	if lastNonEmpty < 0 {
+		return []map[felt.Felt]struct{}{}
+	}
```
infrmtcs: In my opinion, we should remove this, because `[[1]]` and `[[1], []]` will be implicitly the same, while they're representing 2 different queries:
- Find all events with first key is 1.
- Find all events with at least 2 keys and first key is 1.

Even though I'd agree that the second query is "rare", I think removing this avoid users having to guess the meaning.

PR #3282, opened 2025-11-20
feature(trie): create `rawDB` trieDB engine

MaksymMalicki:
It introduces rawDB engine for the `trie2`, which is used by the new state. In this PR:
- The new package with rawDB is created. The structure of the package is quite simple and follows the pattern of `pathdb` and `hashdb` both in structure and interfaces.
- `Contract`, `Class` and `Contract Storages` are written directly to the pebbleDB and read directly from it. The writing is done in the `Update()` method, which takes the nodesets generated by `trie2.Commit()`, flattens them and writes them to the DB. For reading, the `readNode()` method was introduced.
- For the new State, Juno defaults to `rawdb` as a `triedb` engine

Next PR to review: #3283

### DIFF
--- a/core/trie2/triedb/database/db.go
+++ b/core/trie2/triedb/database/db.go
@@ -9,6 +9,14 @@ import (
 	"github.com/NethermindEth/juno/db"
 )
 
+type TrieDBScheme uint8
+
+const (
+	RawScheme TrieDBScheme = iota + 1
+	PathScheme
+	HashScheme
+)
+
 // Represents a reader for trie nodes
 type NodeReader interface {
 	Node(owner *felt.Felt, path *trieutils.Path, hash *felt.Felt, isLeaf bool) ([]byte, error)
@@ -37,5 +45,5 @@ type TrieDB interface {
 		mergeClassNodes,
 		mergeContractNodes *trienode.MergeNodeSet,
 	) error
-	Scheme() string
+	Scheme() TrieDBScheme
 }
--- a/core/trie2/triedb/hashdb/database.go
+++ b/core/trie2/triedb/hashdb/database.go
@@ -13,8 +13,6 @@ import (
 	"github.com/NethermindEth/juno/utils"
 )
 
-const HashScheme string = "hash"
-
 var _ database.TrieDB = (*Database)(nil)
 
 type Config struct {
@@ -259,6 +257,6 @@ func (d *Database) GetTrieRootNodes(classRootHash, contractRootHash *felt.Felt) 
 	return classRootNode, contractRootNode, nil
 }
 
-func (d *Database) Scheme() string {
-	return HashScheme
+func (d *Database) Scheme() database.TrieDBScheme {
+	return database.HashScheme
 }
--- a/core/trie2/triedb/rawdb/database.go
+++ b/core/trie2/triedb/rawdb/database.go
@@ -11,8 +11,6 @@ import (
 	"github.com/NethermindEth/juno/utils"
 )
 
-const RawScheme string = "raw"
-
 var _ database.TrieDB = (*Database)(nil)
 
 type Config struct{}
@@ -137,6 +135,6 @@ func (d *Database) Close() error {
 	return nil
 }
 
-func (d *Database) Scheme() string {
-	return RawScheme
+func (d *Database) Scheme() database.TrieDBScheme {
+	return database.RawScheme
 }
--- a/core/trie2/triedb/pathdb/database.go
+++ b/core/trie2/triedb/pathdb/database.go
@@ -12,9 +12,8 @@ import (
 )
 
 const (
-	maxDiffLayers                  = 128 // TODO(weiihann): might want to make this configurable
-	contractClassTrieHeight        = 251
-	PathScheme              string = "path"
+	maxDiffLayers           = 128 // TODO(weiihann): might want to make this configurable
+	contractClassTrieHeight = 251
 )
 
 var _ database.TrieDB = (*Database)(nil)
@@ -107,6 +106,6 @@ func (d *Database) NewIterator(id trieutils.TrieID) (db.Iterator, error) {
 	return d.disk.NewIterator(prefix, true)
 }
 
-func (d *Database) Scheme() string {
-	return PathScheme
+func (d *Database) Scheme() database.TrieDBScheme {
+	return database.PathScheme
 }


core/trie2/triedb/rawdb/database.go
```diff
@@ -0,0 +1,134 @@
...
+func New(disk db.KeyValueStore) *Database {
+	return &Database{
+		disk: disk,
+		log:  utils.NewNopZapLogger(),
+	}
+}
+
+func (d *Database) readNode(
+	id trieutils.TrieID,
+	owner *felt.Felt,
```
rodrodros, line 34: Is it an owner a `*felt.Felt` or can we find a more descriptive type?

core/trie2/triedb/rawdb/reader.go
```diff
@@ -0,0 +1,27 @@
...
+
+type reader struct {
+	id trieutils.TrieID
+	d  *Database
+}
+
+func (r *reader) Node(
+	owner *felt.Felt,
+	path *trieutils.Path,
+	hash *felt.Felt,
```
rodrodros, line 19: Are this type on their more descriptive form? Because this is satisfying an interface, then please leave a todo over the interface definition to update the types to more descriptive ones.

core/trie2/triedb/rawdb/database.go
MaksymMalicki: actually, it can be replaced by `felt.Address`. I opened a PR for this [here](https://github.com/NethermindEth/juno/pull/3295/files#diff-c39282f209956ca4c9edc1f47b8f8bd864d8c0280acaf87ab8ec1931e50f573f)

core/trie2/triedb/rawdb/reader.go
MaksymMalicki: Actually I think this is a good moment to include this refactor after this PR gets merged. I will include it in the followup PRs

core/state/history_test.go
```diff
@@ -20,10 +20,12 @@ func TestNewStateHistory(t *testing.T) {
 		assert.NotNil(t, history.state)
 	})
 
-	t.Run("invalid state root", func(t *testing.T) {
+	t.Run("bigger state root", func(t *testing.T) {
```
infrmtcs: Why is this `bigger` while it's `invalidRoot` below?

core/trie2/triedb/rawdb/database.go
```diff
@@ -0,0 +1,134 @@
...
+
+	if mergedClassNodes != nil {
+		classNodes, _ = mergedClassNodes.Flatten()
+	}
+	if mergedContractNodes != nil {
+		contractNodes, contractStorageNodes = mergedContractNodes.Flatten()
+	}
+
+	for path, n := range classNodes {
+		if err := d.updateNode(batch, db.ClassTrie, &felt.Zero, &path, n); err != nil {
```
EgeCaner: nit: IMO for this sort of long function calls following pattern is more readable
```go
err := call()
if err != nil{
    return err
}
```

core/trie2/triedb/rawdb/database.go
```diff
@@ -0,0 +1,134 @@
...
+			return err
+		}
+	}
+
+	for path, n := range contractNodes {
+		if err := d.updateNode(batch, db.ContractTrieContract, &felt.Zero, &path, n); err != nil {
```
EgeCaner: same comment as above

core/trie2/triedb/rawdb/database.go
```diff
@@ -0,0 +1,134 @@
...
+			return err
+		}
+	}
+
+	for owner, nodes := range contractStorageNodes {
+		for path, n := range nodes {
+			if err := d.updateNode(batch, db.ContractTrieStorage, &owner, &path, n); err != nil {
```
EgeCaner: same as above

core/trie2/triedb/rawdb/database.go
```diff
@@ -0,0 +1,134 @@
...
+
+func (d *Database) updateNode(
+	batch db.KeyValueWriter,
+	bucket db.Bucket,
+	owner *felt.Felt,
+	path *trieutils.Path,
+	n trienode.TrieNode,
+) error {
+	if _, deleted := n.(*trienode.DeletedNode); deleted {
+		if err := trieutils.DeleteNodeByPath(batch, bucket, owner, path, n.IsLeaf()); err != nil {
```
EgeCaner: same as above

core/trie2/triedb/rawdb/database.go
```diff
@@ -0,0 +1,134 @@
...
+			return err
+		}
+	} else {
+		if err := trieutils.WriteNodeByPath(
```
EgeCaner: same as above

core/trie2/triedb/database.go
```diff
@@ -30,19 +33,18 @@ type Database struct {
 func New(disk db.KeyValueStore, config *Config) (*Database, error) {
```
EgeCaner: Can we return Database interface rather than Database struct and then directly interact with underlying db through interface, avoiding switches in other functions to choose which db we are working with?

EgeCaner (commented):
Looks good! Some nitpicks

core/trie2/triedb/database.go
MaksymMalicki: Yeah, agreed

core/trie2/triedb/rawdb/database.go
MaksymMalicki: done

core/trie2/triedb/rawdb/database.go
MaksymMalicki: done

core/trie2/triedb/rawdb/database.go
MaksymMalicki: done

core/trie2/triedb/rawdb/database.go
MaksymMalicki: done

core/trie2/triedb/rawdb/database.go
MaksymMalicki: I like it actually. I was just doing that out of dev convenience, but for the usage it can be unhandy, I agree

core/trie2/triedb/database/db.go
```diff
@@ -29,4 +30,8 @@ type TrieDB interface {
 	io.Closer
 
 	Commit(stateComm *felt.Felt) error
+	Update(
```
EgeCaner, line 41: Nit: One line per parameter

core/state/history_test.go
MaksymMalicki: yeah, I turned it back to invalid with a TODO, that this test only applies to the `pathDB` scheme turned on

core/trie2/triedb/database/db.go
MaksymMalicki: done

core/trie2/triedb/rawdb/database.go
```diff
@@ -0,0 +1,139 @@
...
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (d *Database) Close() error {
+	return nil
+}
```
rodrodros, line 136: Please add a comment explaining that this is to satisfy the `io.Closer` interface

core/trie2/triedb/rawdb/database.go
```diff
@@ -0,0 +1,139 @@
...
+)
+
+type Config struct{}
+
+type Database struct {
+	disk db.KeyValueStore
+
+	lock sync.RWMutex
+	log  utils.SimpleLogger
+}
```
rodrodros, line 23: Since `Database` is implementing an interface, I believe it is a good programming practice to add a compilation check for it. It helps because:
1. We have a compilation check on this same file
2. Everyone reading this code will know that this DB is intentionally following a certain(s) interface(s).

```golang
var _ database.TrieDB = (*Database)(nil)
```

core/trie2/triedb/rawdb/database.go
```diff
@@ -0,0 +1,139 @@
...
+		oBytes := owner.Bytes()
+		key = append(key, oBytes[:]...)
+	}
+
+	return d.disk.NewIterator(key, true)
+}
+
+func (d *Database) Commit(_ *felt.Felt) error {
+	return nil
+}
```
rodrodros: Please add a comment explaining why `Commit` remains unimplemented

core/trie2/trieutils/accessors.go
```diff
@@ -130,7 +130,7 @@ func GetNodeByHash(r db.KeyValueReader, bucket db.Bucket, owner *felt.Felt, path
 	var res []byte
 	if err := r.Get(nodeKeyByHash(bucket, owner, path, hash, isLeaf),
 		func(value []byte) error {
-			res = value
+			res = slices.Clone(value)
```
rodrodros, line 133: Please a new test to assert the new behaviour

core/trie2/trieutils/accessors.go
```diff
@@ -13,7 +13,7 @@ func GetNodeByPath(r db.KeyValueReader, bucket db.Bucket, owner *felt.Felt, path
 	var res []byte
 	if err := r.Get(nodeKeyByPath(bucket, owner, path, isLeaf),
 		func(value []byte) error {
-			res = value
+			res = slices.Clone(value)
```
rodrodros, line 16: Here as well

core/trie2/triedb/rawdb/database.go
rodrodros, line 1: Nothing of the rawdb has unit tests, I think we should add them

core/trie2/triedb/database.go
EgeCaner: If we are not doing this at the moment, and decided to plan it for the later, please open an issue so we wont forget.

EgeCaner: let me repost the idea here: 

As I see Database struct is just a wrapper to handle creation and interaction with different trieDBs. I was wondering what if we return lets say Database interface instead of struct and remove struct all together. I guess all tries implement NodeReader, Update, Close, NewIterator and Commit. This leaves Journal and Scheme, other than that we are just routing call to underlying DB directly. What if we just make Underlying DBs to implement Scheme() which returns pathdb etc.. and Journal. Then New function creates and returns Database interface then from there consumer will directly interact with DB through interface. This allow us to remove `Database` struct and replace it with interface implemented by the underlying trie then we can avoid switches and remove all the code in that file except `New`.

core/trie2/triedb/database.go
```diff
@@ -15,34 +16,35 @@ import (
...
+	RawConfig  *rawdb.Config
 }
 
 type Database struct {
 	triedb database.TrieDB
 	config *Config
 }
 
-func New(disk db.KeyValueStore, config *Config) (*Database, error) {
+func New(disk db.KeyValueStore, config *Config) (database.TrieDB, error) {
```
EgeCaner, line 19: Is this correct?  We are returning `TrieDB` here and I suppose you cannot call `Schema` and `Journal` on it. I guess we either need `Database` back in here or interface comment I mentioned

MaksymMalicki: There is an option to refactor this, to get rid of the Database entirely. I would however like to do it after the write atomicity PR.

core/trie2/triedb/rawdb/database.go
MaksymMalicki: added tests

core/trie2/triedb/rawdb/database.go
MaksymMalicki: added comment

core/trie2/triedb/rawdb/database.go
MaksymMalicki: added compilation check, thanks

core/trie2/triedb/rawdb/database.go
MaksymMalicki: added comment

core/state/history_test.go
infrmtcs: Should we also rename the test back to `invalid`?

core/trie2/trieutils/accessors.go
infrmtcs: Let's create an issue to track this. WE need to do this everywhere we use the returned byte slice outside of the closure.

core/state/history_test.go
MaksymMalicki: yep

core/trie2/triedb/rawdb/database_test.go
```diff
@@ -0,0 +1,274 @@
...
+	"github.com/NethermindEth/juno/db"
+	"github.com/NethermindEth/juno/db/memory"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+var (
+	leaf1Hash = new(felt.Felt).SetUint64(201)
+	leaf2Hash = new(felt.Felt).SetUint64(202)
+	rootHash  = new(felt.Felt).SetUint64(100)
```
EgeCaner: Use felt constructors

core/trie2/triedb/rawdb/database_test.go
```diff
@@ -0,0 +1,274 @@
...
+	t.Run("Update with all node types", func(t *testing.T) {
+		memDB := memory.New()
+		database := New(memDB)
+
+		contractHash := new(felt.Felt).SetUint64(210)
+		contractPath := trieutils.NewBitArray(1, 0x01)
+		contractNode := trienode.NewLeaf(*contractHash, []byte{7, 8, 9})
+
+		contractOwner := new(felt.Felt).SetUint64(123)
+		storageHash := new(felt.Felt).SetUint64(220)
```
EgeCaner: Use felt constructors here

core/trie2/triedb/rawdb/types.go
```diff
@@ -0,0 +1,13 @@
...
+import (
+	"github.com/NethermindEth/juno/core/felt"
+	"github.com/NethermindEth/juno/core/trie2/trienode"
+	"github.com/NethermindEth/juno/core/trie2/trieutils"
+)
+
+type (
+	classNodesMap           = map[trieutils.Path]trienode.TrieNode
+	contractNodesMap        = map[trieutils.Path]trienode.TrieNode
+	contractStorageNodesMap = map[felt.Felt]map[trieutils.Path]trienode.TrieNode
```
infrmtcs, line 12: Should it be:
```suggestion
	contractStorageNodesMap = map[felt.Address]map[trieutils.Path]trienode.TrieNode
```

MaksymMalicki: I decided to include this change in a separate PR https://github.com/NethermindEth/juno/pull/3295

core/trie2/triedb/rawdb/database_test.go
```diff
@@ -0,0 +1,274 @@
...
+
+		err := database.Update(&felt.Zero, &felt.Zero, 1, createMergeNodeSet(basicClassNodes), nil)
+		require.NoError(t, err)
+
+		classID := trieutils.NewClassTrieID(felt.Zero)
+		verifyNode(t, database, classID, &rootPath, rootNode)
+		verifyNode(t, database, classID, &leaf1Path, leaf1Node)
+		verifyNode(t, database, classID, &leaf2Path, leaf2Node)
+
+		newLeafHash := new(felt.Felt).SetUint64(203)
```
EgeCaner: felt ctors

core/trie2/triedb/rawdb/database.go
```diff
@@ -0,0 +1,144 @@
...
+			owner,
+			path,
+			n.IsLeaf(),
+			n.Blob(),
+		)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
```
infrmtcs: nit:
```suggestion
	if _, deleted := n.(*trienode.DeletedNode); deleted {
		return trieutils.DeleteNodeByPath(batch, bucket, owner, path, n.IsLeaf())
	}
	return trieutils.WriteNodeByPath(
		batch,
		bucket,
		owner,
		path,
		n.IsLeaf(),
		n.Blob(),
	)
```

core/trie2/triedb/rawdb/database_test.go
```diff
@@ -0,0 +1,274 @@
+package rawdb
```
infrmtcs, line 1: Can be in another PR, but can we use the same test for all 3 implementations?

core/trie2/triedb/database.go
infrmtcs: I agree with this, we can remove this abstraction layer.

core/trie2/trieutils/accessors.go
infrmtcs: I created an issue here: https://github.com/NethermindEth/juno/issues/3318

core/trie2/triedb/rawdb/database_test.go
MaksymMalicki: done

core/trie2/triedb/rawdb/database.go
rodrodros: Based on Ege's previous comment, https://github.com/NethermindEth/juno/pull/3282#discussion_r2572122964, shouldn't refactor this logic?

core/trie2/triedb/rawdb/database_test.go
MaksymMalicki: done

core/trie2/triedb/rawdb/database_test.go
MaksymMalicki: done

core/trie2/triedb/rawdb/database.go
MaksymMalicki: nice, thanks

core/trie2/triedb/rawdb/database.go
MaksymMalicki: yes I removed the `Database` structure and return the interface `database.TrieDB` interface with underlying implementation from the `New` method

core/trie2/triedb/database.go
MaksymMalicki: its done, I removed the `Database` struct and I'm returning the `database.TrieDB` interface with underlying triedb implementation

MaksymMalicki: thanks!

core/trie2/triedb/rawdb/database_test.go
MaksymMalicki: I'll create a different PR for this, this one covers a lot so far. It's an interesting idea, I'll have to think if its possible, because the underlying triedb implementations don't have the same behavior

core/trie2/triedb/hashdb/database.go
```diff
@@ -13,6 +13,10 @@ import (
 	"github.com/NethermindEth/juno/utils"
 )
 
+const HashScheme string = "hash"
```
rodrodros: Why does `Scheme()` returns a string instead of an enum?

The advantage of enums over strings is that comparisons are much cheaper. Comparing one number vs `n` chars.

core/trie2/triedb/rawdb/database_test.go
rodrodros: @MaksymMalicki maybe I am missing something but shouldn't the all the trieDB implementations have the same behaviour?

Shouldn't for all three implementations return the same output given the same input? You can write the same set of test for the `database.Database` interface and then do a full set of test with the first trie implementation, with the second trie and the third one.

infrmtcs (approved):

PR #3285, opened 2025-11-20
fix: CI skip declare test cases with old casm hash

EgeCaner:
Since sepolia upgraded to v0.14.1, rpc test cases that are declaring transactions with poseidon class hash will constantly fails. This PR skips them, however, they would be useful again when starknet.rs releases version which uses blake2s hash for declared classes

### DIFF
--- a/.github/workflows/starknet-rs-tests.yml
+++ b/.github/workflows/starknet-rs-tests.yml
@@ -26,7 +26,9 @@ jobs:
       - name: Run jsonrpc tests
         run: |
           cd starknet-providers && cargo test jsonrpc -- --skip jsonrpc_get_block_with_receipts
-          cd ../starknet-accounts && cargo test jsonrpc -- \
+          cd ../starknet-accounts
+          # TODO: Once SDK is updated, remove ignore for declare tests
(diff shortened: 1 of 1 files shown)


rodrigo-pino (approved):

.github/workflows/starknet-rs-tests.yml
```diff
@@ -26,7 +26,10 @@ jobs:
       - name: Run jsonrpc tests
         run: |
           cd starknet-providers && cargo test jsonrpc -- --skip jsonrpc_get_block_with_receipts
-          cd ../starknet-accounts && cargo test jsonrpc -- --skip can_declare_cairo0_contract_with_jsonrpc
+          cd ../starknet-accounts && cargo test jsonrpc -- \
+            --skip can_declare_cairo0_contract_with_jsonrpc \
+            --skip can_estimate_declare_v3_fee_with_jsonrpc \
+            --skip can_declare_cairo1_contract_v3_with_jsonrpc
```
infrmtcs: Can we leave a TODO here or create an issue to track it, so we don't forget about this.

PR #3286, opened 2025-11-20
fix use tip only for v3 txs

EgeCaner:
Fixes a bug where we attempt to get tip from non v3 transaction which results in panic


rodrigo-pino (commented):
Is there any possible way we can add a test for this?

PR #3288, opened 2025-11-21
fix: Unwind Rust panic on FFI boundary

infrmtcs:
According to [Rustonomicon](https://doc.rust-lang.org/nomicon/unwinding.html):
> Rust's unwinding strategy is not specified to be fundamentally compatible with any other language's unwinding. As such, unwinding into Rust from another language, or unwinding into another language from Rust is Undefined Behavior. You must absolutely catch any panics at the FFI boundary! What you do at that point is up to you, but something must be done. If you fail to do this, at best, your application will crash and burn. At worst, your application won't crash and burn, and will proceed with completely clobbered state.

This PR does 2 things:
- `panic::catch_unwind` in the top-level function at FFI boundary.
- The current error handling behaviour on the Go side requires the error to be a JSON, so any string error must be wrap by double quotes.

### DIFF
--- a/vm/rust/src/entrypoint/call/call.rs
+++ b/vm/rust/src/entrypoint/call/call.rs
@@ -13,7 +13,6 @@ use blockifier::{
     transaction::objects::{DeprecatedTransactionInfo, TransactionInfo},
 };
 use once_cell::sync::Lazy;
-use serde_json::json;
 use starknet_api::{
     contract_class::EntryPointType,
     core::{ClassHash, ContractAddress},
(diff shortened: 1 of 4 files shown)


EgeCaner (approved):
LGTM! Can you also remove this unused import 

> warning: unused import: `serde_json::json`
  --> src/entrypoint/execute/execute.rs:21:5
   |
21 | use serde_json::json;
   |     ^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(unused_imports)]` on by default

PR #3290, opened 2025-11-21
refactor(trie): Separate the deprecated`trie` reader from the `trie`

MaksymMalicki:
This PR aims to separate the `TrieReader` from the `Trie`, and utilise snapshot instead of IndexedBatch for reads. This way, we get a clean, stable view into the tries which can be served in RPC endpoints.

### DIFF
--- a/core/trie/trie.go
+++ b/core/trie/trie.go
@@ -17,7 +17,7 @@ import (
 
 const globalTrieHeight = 251 // TODO(weiihann): this is declared in core also, should be moved to a common place
 
-// TODO(maksym): try to make this interface private after common trie is integrated
+// TODO(maksym): try to make this struct private after common trie is integrated, rename to reader
 type TrieReader struct {
 	height uint8
 	// TODO(maksym): Trie mutates rootKey field, and the idea behind TrieReader is to be immutable.
@@ -34,7 +34,7 @@ func NewTrieReaderPedersen(
 	r db.KeyValueReader,
 	prefix []byte,
 	height uint8,
-) (*TrieReader, error) {
+) (TrieReader, error) {
 	return newTrieReader(r, prefix, height, crypto.Pedersen)
 }
 
@@ -42,7 +42,7 @@ func NewTrieReaderPoseidon(
 	r db.KeyValueReader,
 	prefix []byte,
 	height uint8,
-) (*TrieReader, error) {
+) (TrieReader, error) {
 	return newTrieReader(r, prefix, height, crypto.Poseidon)
 }
 
@@ -51,22 +51,24 @@ func newTrieReader(
 	prefix []byte,
 	height uint8,
 	hash crypto.HashFn,
-) (*TrieReader, error) {
+) (TrieReader, error) {
 	if height > felt.Bits {
-		return nil, fmt.Errorf("max trie height is %d, got: %d", felt.Bits, height)
+		return TrieReader{}, fmt.Errorf("max trie height is %d, got: %d", felt.Bits, height)
 	}
 
 	// maxKey is 2^height - 1
-	maxKey := new(felt.Felt).Exp(new(felt.Felt).SetUint64(2), new(big.Int).SetUint64(uint64(height)))
-	maxKey.Sub(maxKey, new(felt.Felt).SetUint64(1))
+	x := felt.NewFromUint64[felt.Felt](2)
+	y := new(big.Int).SetUint64(uint64(height))
+	maxKey := x.Exp(x, y)
+	maxKey.Sub(maxKey, &felt.One)
 
 	readStorage := NewReadStorage(r, prefix)
 	rootKey, err := readStorage.RootKey()
 	if err != nil && !errors.Is(err, db.ErrKeyNotFound) {
-		return nil, err
+		return TrieReader{}, err
 	}
 
-	return &TrieReader{
+	return TrieReader{
 		readStorage: readStorage,
 		height:      height,
 		rootKey:     rootKey,
@@ -93,7 +95,7 @@ func newTrieReader(
 //
 // [specification]: https://docs.starknet.io/architecture-and-concepts/network-architecture/starknet-state/#merkle_patricia_trie
 type Trie struct {
-	*TrieReader
+	TrieReader
 	storage        *Storage
 	dirtyNodes     []*BitArray
 	rootKeyIsDirty bool


db/database.go
```diff
@@ -15,6 +15,7 @@ type KeyValueReader interface {
 	//		return nil
 	//	})
 	Get(key []byte, cb func(value []byte) error) error
+	Iterable
```
infrmtcs: While I think `Iterable` is over abstraction, we should do either of these, depending on the relationship between Reader and Iterable:
- If Reader === Iterable: Remove `Iterable` and move its functions to `KeyValueReader`
- If Iterable is a subset of Reader: Iterable interface should contain `KeyValueReader` instead
- If Iterable and Reader are independent (current assumption): If we need to provide both, create another interface to provide both of them instead.

core/trie/trie.go
```diff
@@ -63,31 +120,40 @@ func newTrie(storage *Storage, height uint8, hash crypto.HashFn) (*Trie, error)
 	maxKey := new(felt.Felt).Exp(new(felt.Felt).SetUint64(2), new(big.Int).SetUint64(uint64(height)))
 	maxKey.Sub(maxKey, new(felt.Felt).SetUint64(1))
 
+	storage := NewStorage(txn, prefix)
```
infrmtcs, line 120: nit:
```suggestion
	readStorage := NewReadStorage(txn, prefix)
	storage := NewStorage(txn, prefix)
```

core/trie/trie.go
```diff
@@ -63,31 +120,40 @@ func newTrie(storage *Storage, height uint8, hash crypto.HashFn) (*Trie, error)
...
 
+	readOnlyStorage := NewReadStorage(storage.txn, storage.prefix)
 	return &Trie{
+		TrieReader: &TrieReader{
+			readStorage: readOnlyStorage,
+			height:      height,
+			rootKey:     rootKey,
+			maxKey:      maxKey,
+			hash:        hash,
+		},
```
infrmtcs: We should use `newTrieReader` instead of this.

core/trie/trie.go
```diff
@@ -11,11 +11,67 @@ import (
 	"github.com/NethermindEth/juno/core/crypto"
 	"github.com/NethermindEth/juno/core/felt"
 	"github.com/NethermindEth/juno/db"
+	"github.com/NethermindEth/juno/db/memory"
 	"github.com/NethermindEth/juno/utils"
 )
 
 const globalTrieHeight = 251 // TODO(weiihann): this is declared in core also, should be moved to a common place
 
+type TrieReader struct {
```
infrmtcs, line 21: Can be in the next PR, but I strongly believe that we should make this private if possible. Main reason is:
- There are some functions provided by `TrieReader` that are superseded by another version in `Trie`, for example `Hash()`.
- It's dangerous and easy to make a mistake to do `trie.TrieReader.Hash()` instead of `trie.Hash()`, therefore we should hide `trieReader` if possible.

infrmtcs (approved):
The rest LGTM, thanks!

core/trie/trie.go
```diff
@@ -11,11 +11,67 @@ import (
...
+	height      uint8
+	rootKey     *BitArray
```
infrmtcs, line 25: Actually we're mutating `rootKey`, so it's contradicts with reader

core/trie/trie.go
MaksymMalicki: done

db/database.go
MaksymMalicki: I think the Iterable was never meant to be used outside of the KeyValueReader. Personally it makes much sense to move it there. So I opt for point 1

core/trie/trie.go
MaksymMalicki: yes, hiding this interface will be much simpler once the `commontrie` interface is integrated into Juno.

core/trie/trie.go
```diff
@@ -63,31 +120,33 @@ func newTrie(storage *Storage, height uint8, hash crypto.HashFn) (*Trie, error)
 	maxKey := new(felt.Felt).Exp(new(felt.Felt).SetUint64(2), new(big.Int).SetUint64(uint64(height)))
 	maxKey.Sub(maxKey, new(felt.Felt).SetUint64(1))
 
-	rootKey, err := storage.RootKey()
-	if err != nil && !errors.Is(err, db.ErrKeyNotFound) {
+	storage := NewStorage(txn, prefix)
+
+	trieReader, err := newTrieReader(storage.txn, prefix, height, hash)
```
infrmtcs, line 122: Can we also remove the code above to validate `height` and calculate `maxKey`, because they're already covered by `newTrieReader`?

core/trie/trie.go
infrmtcs: Let's put a TODO here and address in another PR.

core/trie/trie.go
infrmtcs: I agree, let's do it after integrating `commontrie`.

core/trie/trie.go
MaksymMalicki: removed

core/trie/trie.go
MaksymMalicki: added TODO

core/trie/trie.go
```diff
@@ -11,11 +11,70 @@ import (
...
+	height uint8,
+	hash crypto.HashFn,
+) (*TrieReader, error) {
+	if height > felt.Bits {
+		return nil, fmt.Errorf("max trie height is %d, got: %d", felt.Bits, height)
+	}
+
+	// maxKey is 2^height - 1
+	maxKey := new(felt.Felt).Exp(new(felt.Felt).SetUint64(2), new(big.Int).SetUint64(uint64(height)))
+	maxKey.Sub(maxKey, new(felt.Felt).SetUint64(1))
```
rodrodros: Please use the constructors for felt here

core/trie/trie.go
```diff
@@ -11,11 +11,70 @@ import (
 	"github.com/NethermindEth/juno/core/crypto"
 	"github.com/NethermindEth/juno/core/felt"
 	"github.com/NethermindEth/juno/db"
+	"github.com/NethermindEth/juno/db/memory"
 	"github.com/NethermindEth/juno/utils"
 )
 
 const globalTrieHeight = 251 // TODO(weiihann): this is declared in core also, should be moved to a common place
 
+// TODO(maksym): try to make this interface private after common trie is integrated
```
rodrodros: nitpick: I believe this comment should refer to a `struct` (make the struct private)

core/trie/trie.go
```diff
@@ -11,11 +11,70 @@ import (
...
+	}
+
+	return &TrieReader{
+		readStorage: readStorage,
+		height:      height,
+		rootKey:     rootKey,
+		maxKey:      maxKey,
+		hash:        hash,
+	}, nil
+}
```
rodrodros: Should this type be returned by reference? The answer is yes if:
- it's shared by different go-routines
- it has concurrency control mechanisms (such as a mutex) 

Otherwise, I think the answer is maybe. Let me see how it is used

core/trie/trie.go
```diff
@@ -34,60 +93,55 @@ const globalTrieHeight = 251 // TODO(weiihann): this is declared in core also, s
 //
 // [specification]: https://docs.starknet.io/architecture-and-concepts/network-architecture/starknet-state/#merkle_patricia_trie
 type Trie struct {
-	height  uint8
-	rootKey *BitArray
-	maxKey  *felt.Felt
-	storage *Storage
-	hash    crypto.HashFn
-
+	*TrieReader
```
rodrodros: Why a pointer to `TrieReader`? and not `TrieReader`?

core/trie/trie.go
rodrodros: Additionally, if we make this private, I think the best name for it is `Reader`

rodrigo-pino (approved):
Looks good, just some nitpicks

core/trie/trie.go
MaksymMalicki: done

core/trie/trie.go
MaksymMalicki: renamed

core/trie/trie.go
MaksymMalicki: Its not used by different goroutines. It's not a big structure as well, so we may return it by value

core/trie/trie.go
MaksymMalicki: same here

PR #3291, opened 2025-11-21
refactor(state): split the old state into state reader and state writer structs

MaksymMalicki:
This PR introduces the following change. The current implementation of the `State` gets split into two structs:
- `StateReader` - a structure, which sole purpose is reading the state. It's fed with `db.KeyValueReader` to enforce read-only mode for the state. Eventually we want to feed it only with the `db.Snapshot`. It is going to be mainly used during the RPC requests.
- `State` - a structure which purpose is reading and writing. The `StateReader` is a part of it. It is used during the synchronisation and execution. Currently it runs using an `IndexedBatch`, but eventually we want to feed it with an `IndexedBatch`-like structure, which will contain a write layer and `db.Snapshot` for reads.

### DIFF
--- a/core/contract.go
+++ b/core/contract.go
@@ -196,5 +196,9 @@ func storage(addr *felt.Felt, txn db.IndexedBatch) (*trie.Trie, error) {
 
 func storageReader(addr *felt.Felt, txn db.KeyValueReader) (*trie.TrieReader, error) {
 	addrBytes := addr.Marshal()
-	return trie.NewTrieReaderPedersen(txn, db.ContractStorage.Key(addrBytes), ContractStorageTrieHeight)
+	return trie.NewTrieReaderPedersen(
+		txn,
+		db.ContractStorage.Key(addrBytes),
+		ContractStorageTrieHeight,
+	)
 }
--- a/core/trie/proof.go
+++ b/core/trie/proof.go
@@ -349,7 +349,11 @@ func isEdge(parentKey *BitArray, sNode StorageNode) bool {
 // storageNodeToProofNode converts a StorageNode to the ProofNode(s).
 // Juno's Trie has nodes that are Binary AND Edge, whereas the protocol requires nodes that are Binary XOR Edge.
 // We need to convert the former to the latter for proof generation.
-func storageNodeToProofNode(tri *TrieReader, parentKey *BitArray, sNode StorageNode) (*Edge, *Binary, error) {
+func storageNodeToProofNode(
+	tri *TrieReader,
+	parentKey *BitArray,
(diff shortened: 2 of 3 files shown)


core/pending_state.go
```diff
@@ -136,7 +134,7 @@ func (p *PendingStateWriter) IncrementNonce(contractAddress *felt.Felt) error {
 	if err != nil {
 		return fmt.Errorf("get contract nonce: %v", err)
 	}
-	p.stateDiff.Nonces[*contractAddress] = currentNonce.Add(&currentNonce, feltOne)
+	p.stateDiff.Nonces[*contractAddress] = currentNonce.Add(&currentNonce, &felt.One)
```
infrmtcs: If we're unsure if the felt is eventually modified, we should consider copying it.

core/state.go
```diff
@@ -12,59 +11,47 @@ import (
...
-	stateVersion      = felt.NewFromBytes[felt.Felt]([]byte(`STARKNET_STATE_V0`))
-	leafVersion       = felt.NewFromBytes[felt.Felt]([]byte(`CONTRACT_CLASS_LEAF_V0`))
-	ErrCheckHeadState = errors.New("check head state")
+	stateVersion             = felt.NewFromBytes[felt.Felt]([]byte(`STARKNET_STATE_V0`))
+	leafVersion              = felt.NewFromBytes[felt.Felt]([]byte(`CONTRACT_CLASS_LEAF_V0`))
+	ErrCheckHeadState        = errors.New("check head state")
+	systemContractsClassHash = felt.NewFromUint64[felt.Felt](0)
+	systemContracts          = map[felt.Felt]struct{}{
+		*felt.NewFromUint64[felt.Felt](1): {},
+		*felt.NewFromUint64[felt.Felt](2): {},
```
infrmtcs:
```suggestion
		felt.FromUint64[felt.Felt](1): {},
		felt.FromUint64[felt.Felt](2): {},
```

core/state.go
```diff
@@ -30,45 +27,28 @@ var (
...
 }
 
-// putNewContract creates a contract storage instance in the state and stores the relation between contract address and class hash to be
+// putNewContract creates a contract storage instance in the state and stores
+// the relation between contract address and class hash to be
 // queried later with [GetContractClass].
-func (s *State) putNewContract(stateTrie *trie.Trie, addr, classHash *felt.Felt, blockNumber uint64) error {
+func (s *State) putNewContract(
+	contractTrie *trie.Trie,
+	addr, classHash *felt.Felt,
```
rodrodros, line 49: nit: 1 param per line

PR #3294, opened 2025-11-21
fix(rpc): Fetch only header for StorageProof

### LINKED ISSUE #3259: `starknet_getStorageProof` doesn't need to load transactions and receipts via `bcReader.GetBlockByNumber`
Using `bcReader.GetBlockByNumber` means we load all transactions and receipts from the database, while we actually only need `head.Hash` here.

https://github.com/NethermindEth/juno/blob/v0.15.10/rpc/v9/storage.go#L79

brbrr:
fixes #3259

### DIFF
--- a/rpc/v8/storage_test.go
+++ b/rpc/v8/storage_test.go
@@ -246,9 +246,12 @@ func TestStorageProof(t *testing.T) {
 	handler := rpc.New(mockReader, nil, nil, log)
 
 	t.Run("global roots are filled", func(t *testing.T) {
-		proof, rpcErr := handler.StorageProof(&blockLatest, nil, nil, nil)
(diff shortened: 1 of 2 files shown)


rpc/v9/storage.go
brbrr, line 76: Not sure what to do about this TODO item - is it still relevant?

rodrigo-pino (approved):

rpc/v9/storage.go
infrmtcs: Yeah, it's still relevant. The problem is functions in `bcReader` doesn't use the same snapshot, which can cause race condition. Ideally, we should create a snapshot first before reading from it.

infrmtcs (approved):

PR #3296, opened 2025-11-24
chore(rpc): Fetch only block number for TransactionByBlockIDAndIndex

### LINKED ISSUE #3257: `starknet_getTransactionByBlockIdAndIndex` doesn't have to query block header
Only `header.Number` is used, while the block number can be extracted either from the input in block number case or from `BlockHeaderNumbersByHash` bucket in block hash case.

https://github.com/NethermindEth/juno/blob/v0.15.10/rpc/v9/transaction.go#L552-L557

brbrr:
fixes #3257


blockchain/blockchain.go
```diff
@@ -150,6 +152,16 @@ func (b *Blockchain) BlockHeaderByNumber(number uint64) (*core.Header, error) {
...
 }
 
+func (b *Blockchain) BlockNumberByHash(hash *felt.Felt) (uint64, error) {
+	b.listener.OnRead("BlockNumberByHash")
+	blockNum, err := core.GetBlockHeaderNumberByHash(b.database, hash)
+	if err != nil {
+		return 0, err
+	}
+
+	return blockNum, nil
```
infrmtcs: nit:

```suggestion
	return core.GetBlockHeaderNumberByHash(b.database, hash)
```

rpc/v9/transaction.go
```diff
@@ -550,12 +550,24 @@ func (h *Handler) TransactionByBlockIDAndIndex(
...
+	case number:
+		blockNumber = blockID.Number()
+	case l1Accepted, preConfirmed, latest:
+		return nil, rpccore.ErrBlockNotFound
+	default:
+		panic("unknown block type id")
 	}
 
-	txn, err := h.bcReader.TransactionByBlockNumberAndIndex(header.Number, uint64(txIndex))
+	txn, err := h.bcReader.TransactionByBlockNumberAndIndex(blockNumber, uint64(txIndex))
```
infrmtcs, line 579: We're having 4 versions in v6, v7, v8 and v9, please also fix for the other versions.

rpc/v9/transaction.go
```diff
@@ -550,12 +550,24 @@ func (h *Handler) TransactionByBlockIDAndIndex(
 		return AdaptTransaction(pending.GetBlock().Transactions[txIndex]), nil
 	}
 
-	header, rpcErr := h.blockHeaderByID(blockID)
-	if rpcErr != nil {
-		return nil, rpcErr
+	var blockNumber uint64
+
+	switch blockID.Type() {
+	case hash:
```
infrmtcs: This syntax looks strange

rpc/v9/transaction.go
```diff
@@ -550,12 +550,24 @@ func (h *Handler) TransactionByBlockIDAndIndex(
...
+		var err error
+		blockNumber, err = h.bcReader.BlockNumberByHash(blockID.Hash())
+		if err != nil {
+			return nil, rpccore.ErrBlockNotFound
+		}
+	case number:
+		blockNumber = blockID.Number()
+	case l1Accepted, preConfirmed, latest:
```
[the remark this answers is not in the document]
infrmtcs: `l1Accepted` should be like this, but use `ChainHeight()` for `latest` and `PendingData()` for `preconfirmed`.

rpc/v9/transaction.go
brbrr: yup, I'll update other versions, once the v9 implementation is good enough :)

rpc/v9/transaction.go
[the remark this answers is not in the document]
brbrr: preConfirmed case is covered above the switch case. Also, it seems that we could follow the same logic for latest (assuming `h.bcReader.Head()` is better than following default code path)

rpc/v6/transaction.go
```diff
@@ -484,11 +484,15 @@ func (h *Handler) TransactionByHash(hash felt.Felt) (*Transaction, *jsonrpc.Erro
 //
 // It follows the specification defined here:
 // https://github.com/starkware-libs/starknet-specs/blob/master/api/starknet_api_openrpc.json#L184
-func (h *Handler) TransactionByBlockIDAndIndex(id BlockID, txIndex int) (*Transaction, *jsonrpc.Error) {
+func (h *Handler) TransactionByBlockIDAndIndex(
+	id BlockID, txIndex int,
```
infrmtcs: Can we split these args into multiple lines?

rpc/v9/transaction.go
rodrodros: No need v8, v9 and v10 only please

infrmtcs (approved):

rpc/v9/transaction.go
brbrr: @rodrodros should I revert the changes for v6 and v7 then?

MaksymMalicki (approved):

rpc/v9/transaction.go
rodrodros: No need, if they already done, let's leave them there. Just for the future, optimizations should be done in priority order of recency: v10, v9 and v8.

PR #3301, opened 2025-11-25
refactor: Unify Marshal and MarshalBinary function signatures

infrmtcs:


db/schema.go
```diff
@@ -47,13 +47,17 @@ type BlockNumIndexKey struct {
 	Index  uint64
 }
 
-func (b *BlockNumIndexKey) MarshalBinary() []byte {
+func (b BlockNumIndexKey) Marshal() []byte {
```
MaksymMalicki, line 50: Asking for some context, why do we need to change those signatures in particular? I'm asking, because I see many other accessors, for instance in the `core/state` package, where `MarshalBinary` is still used. Why not change it everywhere? Is it outside of the scope for this PR?

infrmtcs: The intention wasn't to remove `MarshalBinary`, but rather to standardize the method signatures: all `Marshal` methods should return `[]byte`, while all `MarshalBinary` methods should return `([]byte, error)`. This standardization allows me to define interfaces for both `Marshal` and `MarshalBinary`.

MaksymMalicki: okok, thanks for the answer

PR #3302, opened 2025-11-25
feat: Typed prefixed bucket

infrmtcs:


db/typed/prefix/bucket.go
```diff
@@ -0,0 +1,86 @@
...
+		}
+
+		isError := false
+		defer func() {
+			if !isError {
+				if err := it.Close(); err != nil {
+					yield(Entry[V]{}, err)
+				}
+			}
+		}()
```
rodrodros: The goal of this function is to, given that no error has happened till now during iteration, close the iterator and check no errors are given. 

In every error execution flow, you do check for the error for `it.Close`, leaving this defer only for the case where there is no error, which is only one defined from line 81 to 83. Why not remove this defer and this `isError` variable and write this check directly there?

db/typed/prefix/bucket_test.go
```diff
@@ -0,0 +1,204 @@
...
+)
+
+const (
+	layerKeyCount = 50
+	testPerLayer  = 5
+)
+
+type testKey struct {
+	number uint64
+	hash   felt.Felt
```
rodrodros: nitpick: felt.Hash? 👀

db/typed/prefix/bucket_test.go
rodrodros, line 1: Thank you for this very thorough tests!

db/typed/prefix/bucket_test.go
infrmtcs: Done

PR #3304, opened 2025-11-25
feat: Combined layout for transactions

infrmtcs:


blockchain/blockchain.go
```diff
@@ -255,11 +249,13 @@ func (b *Blockchain) Store(
...
-			if err := core.WriteTxAndReceipt(txn, block.Number, uint64(i), tx,
-				block.Receipts[i]); err != nil {
-				return err
-			}
+		if err := core.WriteTransactionsAndReceipts(
+			txn,
+			block.Number,
+			block.Transactions,
+			block.Receipts,
+		); err != nil {
```
rodrodros: Let's separate the assign from the if please

blockchain/blockchain.go
```diff
@@ -724,11 +720,13 @@ func (b *Blockchain) storeBlockData(
...
-	for i, tx := range block.Transactions {
-		if err := core.WriteTxAndReceipt(txn, block.Number, uint64(i), tx, block.Receipts[i]); err != nil {
-			return err
-		}
+	if err := core.WriteTransactionsAndReceipts(
+		txn,
+		block.Number,
+		block.Transactions,
+		block.Receipts,
+	); err != nil {
```
rodrodros: Here as well 🙏

core/accessors.go
```diff
@@ -425,57 +410,71 @@ func GetBlockByNumber(r db.KeyValueReader, blockNum uint64) (*Block, error) {
...
+	if err := BlockTransactionsBucket.Put(w, blockNumber, &blockTransactions); err != nil {
 		return err
 	}
 
-	if err := TransactionsByBlockNumberAndIndexBucket.Put(w, key, &tx); err != nil {
-		return err
+	return nil
+}
+
+func GetTxByBlockNumberAndIndex(
```
rodrodros: nitpick: some methods have been rewritten to use `Transaction` full name instead of `Tx`.

It's a very samll issue, but sometimes we use `Tx`, `Txn` or `Transaction`, as a way to standardize it to `Transaction`

core/accessors.go
```diff
@@ -246,48 +246,40 @@ func DeleteBlockHeaderByNumber(w db.KeyValueWriter, number uint64) error {
...
-			return err
-		}
+	if err := BlockTransactionsBucket.Delete(batch, blockNum); err != nil {
+		return err
+	}
 
-		txHash := (*felt.TransactionHash)(txn.Hash())
+	for _, tx := range blockTransactions.Transactions {
+		txHash := (*felt.TransactionHash)(tx.Hash())
 		if err := TransactionBlockNumbersAndIndicesByHashBucket.Delete(batch, txHash); err != nil {
```
rodrodros: nitpick: I know this line is outside the scope of the PR but could you split the assignment from the if

core/accessors.go
```diff
@@ -246,48 +246,40 @@ func DeleteBlockHeaderByNumber(w db.KeyValueWriter, number uint64) error {
...
 			return err
 		}
 
-		if l1handler, ok := txn.(*L1HandlerTransaction); ok {
+		if l1handler, ok := tx.(*L1HandlerTransaction); ok {
 			if err := DeleteL1HandlerTxnHashByMsgHash(batch, l1handler.MessageHash()); err != nil {
```
rodrodros: Similarly here

core/block_transaction.go
```diff
@@ -0,0 +1,6 @@
+package core
+
+type BlockTransactions struct {
+	Transactions []Transaction         `cbor:"1,keyasint,omitempty"`
+	Receipts     []*TransactionReceipt `cbor:"2,keyasint,omitempty"`
+}
```
rodrodros, line 115: Let's add a description to this, I believe it will help future Juno memebers

migration/blocktransactions/blocktransactions.go
```diff
@@ -0,0 +1,291 @@
...
+		}
+		if has {
+			return blockTransactions{}, false, nil
+		}
+	}
+
+	return blockTransactions{
+		Transactions: transactions,
+		Receipts:     receipts,
+	}, true, nil
```
rodrigo-pino: small nitpick: I believe the boolean here is redundant since `blockTransaction{}` is equivalent to  `false`.

migration/blocktransactions/blocktransactions.go
```diff
@@ -0,0 +1,291 @@
...
+
+		if !hasFirstBlock {
+			logger.Infow("no starting block found, exiting")
+			return nil, clearOldBuckets(database)
+		}
+
+		chainHeight, err := core.GetChainHeight(database)
+		if err != nil {
+			return nil, err
+		}
```
rodrodros: Will this value change across iterations?

migration/blocktransactions/blocktransactions.go
```diff
@@ -0,0 +1,291 @@
...
+}
+
+func (BlockTransactionsMigration) Migrate(
+	ctx context.Context,
+	database db.KeyValueStore,
+	network *utils.Network,
+	logger utils.SimpleLogger,
+) ([]byte, error) {
+	for {
+		firstBlock, hasFirstBlock, err := getFirstBlock(database)
```
rodrodros: nitpick: I think `getFirstBlock` is a bit more readable if it is called `getFirstBlockToMigrate`

and `hasFirstBlock` and `shouldMigrate`

migration/blocktransactions/blocktransactions.go
```diff
@@ -0,0 +1,291 @@
...
+		if err != nil {
+			return nil, err
+		}
+
+		if !hasFirstBlock {
+			logger.Infow("no starting block found, exiting")
```
rodrodros: nitpick: We can give a more reassuring and informative message here such as: "all transactions and receipts migrated" or something along those lines. I also think this probably should be a debug log and not something to appear by default from now on every time someone starts a node.

migration/blocktransactions/blocktransactions.go
```diff
@@ -0,0 +1,291 @@
...
+	if err != nil {
+		return 0, false, err
+	}
+
+	if !hasReceipts && !hasTransactions {
+		return 0, false, nil
+	}
+
+	minBlock := min(transactions, receipts)
+	minBlock -= minBlock % batchSize
```
rodrodros: nitpick: parentization would help a bit here in my opinion

As I understand it, here you want to start through the oldest non-migrated block transactions or receipts, in case the migration was interrupted (I assume). I don't understand very well the logic though.

If the first block with transactions to migrate is `t = 5` and the first one with receipts to migrate is `r = 10` with batch size `b = 2` then:
1. The minimum is `min = min(t, r)` -> `min = min(10, 5)` ->  `min = 5`
2. The result is `res = min - (min % 2)`
3. Which then becomes -> `res = 5 - (5 % 2)` -> `res = 5 - 1` -> `res = 4`. 
4. The block we should then start from is block `4` and not `5` 🤔 – what am I missing.

Additionally, if either transactions and receipts is zero, min block will be zero as well 🤔

migration/blocktransactions/blocktransactions.go
```diff
@@ -0,0 +1,291 @@
...
+					return
+				}
+
+				counter.log(batch, batchSize)
+
+				debugLog(logger, "wrote block", startBlock, endBlock)
+
+				cleanup.CloseBatch(batch)
+			}
+		})
```
rodrodros: What's the behaviour difference between this approach and 
```golang
s.Go(func() stream.Callback {
			endBlock := min(startBlock+batchSize-1, chainHeight)
			batch, err := createBatch(logger, database, startBlock, endBlock)

            if err != nil {
				cleanup.AddError(err)
				return
			}

			debugLog(logger, "writing block", startBlock, endBlock)
			if err := batch.Write(); err != nil {
				cleanup.AddError(err)
				return
			}

			counter.log(batch, batchSize)

			debugLog(logger, "wrote block", startBlock, endBlock)

			cleanup.CloseBatch(batch)
			return func() { }				
		})
```

Trying to understand why using a Callback instead of a normal worker pool

rodrodros: The callbacks run sync, I missed that. By running the write to db synchronously we optimize performance.

migration/blocktransactions/blocktransactions.go
infrmtcs: After successful execution, this migration function will never execute again, due to the `SchemaVersion` check.

core/block_transaction.go
```diff
@@ -0,0 +1,96 @@
...
+	)
+}
+
+func (b *BlockTransactions) Transactions() indexed.LazySlice[Transaction] {
+	end := len(b.Data)
+	if len(b.Indexes.Receipts) > 0 {
+		end = b.Indexes.Receipts[0]
+	}
+
+	return indexed.NewLazySlice[Transaction](b.Indexes.Transactions, b.Data[:end])
```
rodrodros, line 91: What's the reasoning behind `b.Data[:end]`? It looks similar to just passing `b.Data`.

infrmtcs: First half of the data is for transactions, second half is for receipts. We're utilizing end of data as the end for the last item, so for transactions case, we need to slice at the beginning of the first receipt

migration/blocktransactions/blocktransactions.go
infrmtcs: It's mainly to "normalize" the batch start and end to be multiples of `batchSize`, which is currently `10`. The main benefits are to guarantee that multiple runs use consistent batches. For example, we always run batch 0 -> 9, 10 -> 19, 20 -> 29,... but never get batch 5 -> 14. This is to avoid unexpected batch that isn't created during the experiments, which may  create unexpected spike in resource usage.

infrmtcs: Now I remember one more strict reason why we need this. Imagine that the actual next block to migrate is block 20, but block 20 doesn't have any transaction nor receipt. The the first batch will be 21 -> 30 instead of 20 -> 29.
Before rounds of benchmark, I chose batch of 100, which may this approach okay-ish because it only fails in the case where there are 100 consecutive empty blocks. However, now the batch size is 10, which means probably I'll need a better approach, since 10 consecutive empty blocks will fail.

core/accessors.go
infrmtcs: Is there any particular reason why we favor breaking it into 2 lines? IMO it's a good feature to scope the declaration of err, which avoid (possibly) wrong reuse of `err`.

migration/blocktransactions/blocktransactions.go
infrmtcs: Updated, addressed in #3348

migration/blocktransactions/blocktransactions.go
infrmtcs: Updated in #3348

core/block_transaction.go
infrmtcs: Added description in #3348

core/accessors.go
infrmtcs: Updated

MaksymMalicki (approved):

PR #3311, opened 2025-11-25
feat: Upgrade to pebble v2

infrmtcs:


db/pebblev2/upgrade.go
```diff
@@ -1,30 +1,49 @@
...
 // Because both versions are supported by pebble v1, we use pebble v1 to peek at the database and
 // then check the version. If the version is older than version 13, we upgrade the database to
 // version 13 by opening it with pebble v1 using version 13, and then closing it.
-func upgradeFormatIfNeeded(path string) {
+func upgradeFormatIfNeeded(path string) (pebblev2.FormatMajorVersion, error) {
 	desc, err := pebble.Peek(path, vfs.Default)
-	if err != nil {
-		return
+	if isNotV1 := err != nil || !desc.Exists; isNotV1 {
+		fmt.Println("Error peeking database:", err)
```
rodrodros: Is this print intentional? This message will confuse users

infrmtcs: No it wasn't, I removed the print now.

db/pebblev2/db.go
```diff
@@ -28,10 +29,14 @@ type DB struct {
 
 // New opens a new database at the given path with default options
 func New(path string, options ...Option) (db.KeyValueStore, error) {
-	upgradeFormatIfNeeded(path)
+	version, err := upgradeFormatIfNeeded(path)
+	if err != nil {
+		return nil, err
+	}
 
+	fmt.Println("Opening database with version:", version)
```
MaksymMalicki: Is this print intentional or should this be on the log level?

node/node.go
```diff
@@ -146,11 +146,11 @@ func New(cfg *Config, version string, logLevel *utils.LogLevel) (*Node, error) {
 	if dbIsRemote {
 		database, err = remote.New(cfg.RemoteDB, context.TODO(), log, grpc.WithTransportCredentials(insecure.NewCredentials()))
 	} else {
-		database, err = pebble.New(
+		database, err = pebblev2.New(
```
MaksymMalicki, line 149: So do I understand correctly, that after this PR Juno will use `pebblev2` by default, and every user, which is not using `v2` version already will be migrated accordingly by `upgradeFormatIfNeeded`?

db/pebblev2/db.go
infrmtcs: You're right, removed it now

node/node.go
infrmtcs: That's correct. `upgradeFormatIfNeeded` will detect the current format major version. If it's version 1, we use pebble v1 to upgrade to version 13 which is supported by both v1 and v2, and the reopen the database by v2. If it's already version 13, we'll skip upgrade and open with version 13. If the database doesn't exist for a new instance, we'll open with version 13.

PR #3313, opened 2025-11-26
feat: rpc v0.10.0

EgeCaner:
## Added

Updates rpc block header to include commitments.

## Changed

rpc version to 0.10.0

For blocks which does not have L2GasPrice etc... rpcv10 will now defaults resource prices to 0x1, only for block header. This matches the behaviour of Feeder and Pathfinder

## Note
lint-diff will fail for urls, supressing them with nolint will result in general ci run to fail due to conflict in linter configs. Favoring general run over lint-diff will allow us to keep CI green after merging.

### DIFF
--- a/rpc/v10/handlers.go
+++ b/rpc/v10/handlers.go
@@ -173,5 +173,5 @@ func (h *Handler) Run(ctx context.Context) error {
 }
 
 func (h *Handler) SpecVersion() (string, *jsonrpc.Error) {
-	return "0.10.0-rc2", nil
+	return "0.10.0", nil
 }
--- a/rpc/v10/handlers_test.go
+++ b/rpc/v10/handlers_test.go
@@ -22,7 +22,7 @@ func TestSpecVersion(t *testing.T) {
 	handler := rpcv10.New(nil, nil, nil, nil)
 	version, rpcErr := handler.SpecVersion()
 	require.Nil(t, rpcErr)
-	require.Equal(t, "0.10.0-rc2", version)
+	require.Equal(t, "0.10.0", version)
 }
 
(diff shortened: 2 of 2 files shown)


rpc/v10/block.go
```diff
@@ -0,0 +1,306 @@
...
+	L1DAMode              *rpcv6.L1DAMode      `json:"l1_da_mode,omitempty"`
+	StarknetVersion       string               `json:"starknet_version"`
+	L2GasPrice            *rpcv6.ResourcePrice `json:"l2_gas_price"`
+	TransactionCommitment *felt.Hash           `json:"transaction_commitment,omitempty"`
+	EventCommitment       *felt.Hash           `json:"event_commitment,omitempty"`
+	ReceiptCommitment     *felt.Hash           `json:"receipt_commitment,omitempty"`
+	StateDiffCommitment   *felt.Hash           `json:"state_diff_commitment,omitempty"`
+	EventCount            *uint64              `json:"event_count,omitempty"`
+	TransactionCount      *uint64              `json:"transaction_count,omitempty"`
+	StateDiffLength       *uint64              `json:"state_diff_length,omitempty"`
```
rodrodros, line 36: Why do we set this one as pointers, if they are required by the spec?

EgeCaner: There is 2 different kind of block header in spec. One is BLOCK_HEADER and the other is PRE_CONFIRMED_BLOCK_HEADER. These new fields will be present for BLOCK_HEADER which is committed block header, and they will be missing in PRE_CONFIRMED_BLOCK_HEADER. Since we use the same block header type for both we omit those fields for PRE_CONFIRMED_BLOCK_HEADER

PR #3317, opened 2025-12-01
chore(vm): update the blockifier to v0.16.0-rc.2

rodrodros:


vm/rust/src/entrypoint/utils.rs
```diff
@@ -17,6 +17,7 @@ use starknet_api::{
         GasPrices, NonzeroGasPrice, StarknetVersion,
     },
     core::{ChainId, ContractAddress},
+    versioned_constants_logic::VersionedConstantsTrait,
```
infrmtcs, line 20: Do we need this import?

rodrodros: Yes, we have to import the trait to have access `VersionConstant::latest_constant` and `VersionConstant::get` constructors

PR #3322, opened 2025-12-04
fix: getBlockWithTx* fetch block header and txns separately

### LINKED ISSUE #3261: `starknet_getBlockWithTxHashes` and `starknet_getBlockWithTxs` don't need to load receipts
Both are using `blockByID` which loads all receipts while these endpoints don't need anything from the receipts.

https://github.com/NethermindEth/juno/blob/v0.15.10/rpc/v9/block.go#L244
https://github.com/NethermindEth/juno/blob/v0.15.10/rpc/v9/block.go#L322

brbrr:
fixes #3261

### DIFF
--- a/rpc/v10/block_test.go
+++ b/rpc/v10/block_test.go
@@ -1017,7 +1017,9 @@ func TestBlockWithTxsWithResponseFlags(t *testing.T) {
 
 	blockID := rpcv9.BlockIDFromNumber(block.Header.Number)
 	mockReader.EXPECT().BlockHeaderByNumber(block.Header.Number).Return(block.Header, nil).AnyTimes()
-	mockReader.EXPECT().TransactionsByBlockNumber(block.Header.Number).Return(block.Transactions, nil).AnyTimes()
+	mockReader.EXPECT().TransactionsByBlockNumber(
+		block.Header.Number,
+	).Return(block.Transactions, nil).AnyTimes()
 	mockReader.EXPECT().Network().Return(network).AnyTimes()
 	mockReader.EXPECT().L1Head().Return(core.L1Head{}, nil).AnyTimes()
 
--- a/rpc/v9/block_test.go
+++ b/rpc/v9/block_test.go
@@ -365,6 +365,7 @@ func TestBlockWithTxHashes(t *testing.T) {
 	})
 }
 
+//nolint:dupl // similar to TestBlockWithTxs_TxnsFetchError
 func TestBlockWithTxHashes_TxnsFetchError(t *testing.T) {
 	blockNumber := uint64(123)
 	header := &core.Header{Number: blockNumber}
@@ -606,6 +607,7 @@ func TestBlockWithTxs(t *testing.T) {
 	})
 }
 
+//nolint:dupl // similar to TestBlockWithTxHashes_TxnsFetchError
 func TestBlockWithTxs_TxnsFetchError(t *testing.T) {
 	blockNumber := uint64(123)
 	header := &core.Header{Number: blockNumber}


rpc/v8/block.go
```diff
@@ -231,29 +242,40 @@ func (h *Handler) BlockWithReceipts(id *BlockID) (*BlockWithReceipts, *jsonrpc.E
...
 	return &BlockWithTxs{
 		Status:       status,
-		BlockHeader:  adaptBlockHeader(block.Header),
+		BlockHeader:  adaptBlockHeader(header),
 		Transactions: txs,
 	}, nil
 }
 
-func (h *Handler) blockStatus(id *BlockID, block *core.Block) (rpcv6.BlockStatus, *jsonrpc.Error) {
+func (h *Handler) blockStatus(id *BlockID, blockNumber uint64) (rpcv6.BlockStatus, *jsonrpc.Error) {
```
MaksymMalicki, line 278: nice!

rpc/v10/block.go
```diff
@@ -62,34 +62,45 @@ type BlockWithReceipts struct {
 // It follows the specification defined here:
 // https://github.com/starkware-libs/starknet-specs/blob/cce1563eff702c87590bad3a48382d2febf1f7d9/api/starknet_api_openrpc.json#L25
 func (h *Handler) BlockWithTxHashes(id *rpcv9.BlockID) (*BlockWithTxHashes, *jsonrpc.Error) {
-	block, rpcErr := h.blockByID(id)
+	header, rpcErr := h.blockHeaderByID(id)
```
MaksymMalicki, line 71: question, this PR basically makes a trade-off of reducing fetched data size (get rid of unnecessary receipts field) for the tradeoff of doing 2 DB reads instead of 1 right?

rpc/v8/block.go
```diff
@@ -168,24 +168,35 @@ type BlockWithReceipts struct {
 // It follows the specification defined here:
 // https://github.com/starkware-libs/starknet-specs/blob/a789ccc3432c57777beceaa53a34a7ae2f25fda0/api/starknet_api_openrpc.json#L11
 func (h *Handler) BlockWithTxHashes(id *BlockID) (*BlockWithTxHashes, *jsonrpc.Error) {
```
MaksymMalicki, line 170: btw why don't we do it for v6 and v7 as well? very small refactor, and I don't think we have a concrete plan to deprecate those anytime soon

rpc/v10/helpers.go
```diff
@@ -66,6 +66,30 @@ func (h *Handler) blockByID(blockID *rpcv9.BlockID) (*core.Block, *jsonrpc.Error
...
+		if err != nil {
+			if errors.Is(err, core.ErrPendingDataNotFound) {
+				return nil, rpccore.ErrBlockNotFound
+			}
+			return nil, rpccore.ErrInternal.CloneWithData(err)
+		}
+		txns := pending.GetTransactions()
+		return txns, nil
+	default:
+		txns, err := h.bcReader.TransactionsByBlockNumber(blockID.Number())
```
MaksymMalicki, line 83: btw, I don't think we have this method anymore. Since we have the new typed buckets, we fetch the transactions by block number and index. I think this will require a new `helper` function, where we can pass the number of transactions from header

rpc/v8/block.go
brbrr: I'm not sure it worth it, and Rodrigo suggested against it: https://github.com/NethermindEth/juno/pull/3296#discussion_r2657671909

rpc/v10/block.go
brbrr: hm, from my understanding, blockByID does 3 DB reads (header, txns, receipts), where this PR makes only 2 DB reads. I'm not sure how indexed batches work (they're used in `blockByID`), and potentially we might be performing a single database operation, but still, we are pulling unnecessary receipt data, which I assume will incur some overhead anyway.

rpc/v10/helpers.go
brbrr: I think it's there, as I can find it locally and the CI is not complaining 🤷

blockchain/blockchain.go
```diff
@@ -32,13 +32,16 @@ type Reader interface {
...
 	BlockNumberByHash(hash *felt.Felt) (uint64, error)
-
-	TransactionByHash(hash *felt.Felt) (transaction core.Transaction, err error)
-	TransactionByBlockNumberAndIndex(blockNumber, index uint64) (transaction core.Transaction, err error)
 	BlockNumberAndIndexByTxHash(
 		hash *felt.TransactionHash,
 	) (blockNumber uint64, index uint64, err error)
 
+	TransactionByHash(hash *felt.Felt) (transaction core.Transaction, err error)
+	TransactionByBlockNumberAndIndex(
```
MaksymMalicki, line 40: could we increase the codecov by adding some tests for this method?

PR #3334, opened 2025-12-15
fix: Remove redundant Close() call which results in panic

infrmtcs:


db/typed/prefix/bucket_test.go
```diff
@@ -266,43 +308,74 @@ func runReadTests(
...
+func populateDatabase(
+	t *testing.T,
+	database db.KeyValueWriter,
+	content entryMap,
+	layerKeyCount int,
+) {
+	t.Helper()
+	for range layerKeyCount {
+		blockNumber := rand.Uint64()
+		content[blockNumber] = make(map[felt.Address]map[string]testEntry)
```
EgeCaner: Should we create the map with size layerKeyCount?

db/typed/prefix/bucket_test.go
```diff
@@ -266,43 +308,74 @@ func runReadTests(
...
 
-	t.Run("Populate database", func(t *testing.T) {
 		for range layerKeyCount {
-			blockNumber := rand.Uint64()
-			content[blockNumber] = make(map[felt.Address]map[string]testEntry)
+			address := felt.Random[felt.Address]()
+			content[blockNumber][address] = make(map[string]testEntry)
```
EgeCaner: Should we create the map with layerKeyCount?

EgeCaner (approved):
Looks good!

db/typed/prefix/bucket_test.go
```diff
@@ -266,43 +308,74 @@ func runReadTests(
...
+				content[blockNumber][address][slot] = entry
+				require.NoError(t, bucket.Put(database, key, &value))
 			}
 		}
+	}
+}
+
+func insertInvalidValue(t *testing.T, database db.KeyValueWriter) testKey {
+	maxFeltBigInt := new(big.Int).Sub(fp.Modulus(), new(big.Int).SetUint64(1))
+	maxFelt := new(felt.Felt).SetBigInt(maxFeltBigInt)
```
rodrodros: Nitpick: this operation is similar to `felt.Zero minus felt.One`. I think it is easier to read since it doesn't require access to `fp` or `big` libraries

```golang
maxFlet := new(felt.Felt).Sub(felt.Zero, felt.One)
```

db/typed/prefix/bucket_test.go
```diff
@@ -402,14 +475,61 @@ func TestPrefixedBucket(t *testing.T) {
...
+				insertInvalidValue(t, database)
+			}
+			wg := sync.WaitGroup{}
+			t.Cleanup(wg.Wait)
+			for range 32 {
+				wg.Go(func() {
+					for range 10000 {
+						testFullScan(t, database, content, expectErr)
+					}
+				})
```
rodrodros, line 529: Why are we spawning 32 go routines that scan the db ten thousand times?

db/typed/prefix/bucket_test.go
infrmtcs: Updated

db/typed/prefix/bucket_test.go
infrmtcs: Updated

db/typed/prefix/bucket_test.go
infrmtcs: This is the exact test case that can reproduce panic error happening when we don't have the fix. The idea is to have enough concurrent iterators so that the second redundant `Close()` call will affects another concurrent iterator.

db/typed/prefix/bucket_test.go
infrmtcs: Updated

PR #3335, opened 2025-12-18
feat(cmd): DB verification with cmd tool

MaksymMalicki:
This PR introduces a new cmd tool, which can be used for the DB corruption verification on the Merkle Trie side. For now its just an empty command. To use the tool, hit:
`juno verify --db-path <path>`


rodrigo-pino:
I don't know if the PR plans to go live like this, but I strongly suggest it gets divided in two:
- Adding the new cmd overall command
- Adding the trie verification

cmd/juno/verify/trie.go
```diff
@@ -0,0 +1,103 @@
...
+		if len(tries) == 0 {
+			hasContractStorage = true
+		}
+
+		if !hasContractStorage {
+			return fmt.Errorf("--address flag can only be used with --type contract-storage")
+		}
+
+		var addr felt.Felt
+		_, err := (&addr).SetString(contractAddrStr)
```
thiagodeev: Nitpick: just `addr.SetString(...)` should be enough, right?

verify/trie/types.go
```diff
@@ -0,0 +1,29 @@
...
+	ContractTrie        TrieType = "contract"
+	ClassTrie           TrieType = "class"
+	ContractStorageTrie TrieType = "contract-storage"
+)
+
+type TrieInfo struct {
+	Name       string
+	Prefix     []byte
+	HashFunc   trie.NewTrieFunc
+	HashFn     crypto.HashFn
```
thiagodeev: Nitpick: Both have the "same" name. Wouldn't be better to give them more descriptive names?
Some ideas: TrieHashF/TrieHash, CryptoHashF/CryptoHash...

The way it is it's confusing to use: trieInfo.HashFunc x trieInfo.HashFn

Note: I realized `HashFunc` is not being used anywhere. Should we remove it?

cmd/juno/verify/verify.go
```diff
@@ -0,0 +1,82 @@
...
+	ctx := cmd.Context()
+
+	verifiers := []Verifier{
+		trie.NewTrieVerifier(database, logger, nil, nil),
+	}
+
+	for _, v := range verifiers {
+		if err := v.Run(ctx); err != nil {
+			return fmt.Errorf("%s verification failed: %w", v.Name(), err)
+		}
```
thiagodeev: This is the final destiny of every error returned by the verifiers, right? 
If so, I think the current "verification failed" message is not good for some cases.
For errors like "Failed to get stored root hash for..." makes sense, since it means a failure, but what about the error "node corruption detected at key %s: stored hash=%s..."? It's not a failure, it's a corruption successfully detected by the tool.
So, we can think of a message that says something like "An error was found while performing X verification: ..."

verify/trie/trie_verifier.go
```diff
@@ -0,0 +1,261 @@
...
+
+func (v *TrieVerifier) verifyTrieWithLogging(ctx context.Context, trieInfo TrieInfo) error {
+	err := v.verifyTrie(ctx, trieInfo)
+	if err != nil {
+		if errors.Is(err, context.Canceled) {
+			v.logger.Info("Verification stopped", zap.String("trie", trieInfo.Name))
+			return err
+		}
+		v.logger.Error(fmt.Sprintf("%s verification failed: %v", trieInfo.Name, err))
+		return fmt.Errorf("%s verification failed: %w", trieInfo.Name, err)
```
thiagodeev: Isn't this the same message we have in verifyAll() ([ref](https://github.com/NethermindEth/juno/blob/73cb507d9efde3a8846b7ced01c899cb18a34726/cmd/juno/verify/verify.go#L63))? It seems we are logging every error twice, and also using the same message.

WDYT about logging only in the top-level verifyAll() func? Anyway, we need to remove the duplication

verify/trie/trie_verifier.go
```diff
@@ -0,0 +1,261 @@
...
+	return contractAddresses
+}
+
+func (v *TrieVerifier) verifyTrieWithLogging(ctx context.Context, trieInfo TrieInfo) error {
+	err := v.verifyTrie(ctx, trieInfo)
+	if err != nil {
+		if errors.Is(err, context.Canceled) {
+			v.logger.Info("Verification stopped", zap.String("trie", trieInfo.Name))
+			return err
+		}
```
thiagodeev: Here, we verify if error is `context.Canceled`, if so, we log BUT we return it, only for it be verified again in the `TrieVerifier.Run()` method in one of the three `if typeSet[...] {...}` cases. What about unifying this?
Note: the "Verification stopped" log is not being emitted for the `if typeSet[ContractStorageTrie]` case since it uses the `verifyContractStorageTries()` func, which returns `case <-ctx.Done(): return ctx.Err()` without logging anything.

What about this?

```go
func (v *TrieVerifier) Run(ctx context.Context) error {
  ...
  var err error //a single error variable for all cases

  // this could also be a switch case if you want
	if typeSet[ContractTrie] {
		...
    // just store the error in the common variable, no other action inside the condition
		err = v.verifyTrieWithLogging(...)
	}

  if typeSet[ClassTrie] {
		...
		err = v.verifyTrieWithLogging(...)
	}

  if typeSet[ContractStorageTrie] {
		...
		err = v.verifyContractStorageTries(...)
	}

  // a single point of check, dealing with all the cases at once (also dealing with
  // the `typeSet[ContractStorageTrie]` case, which is currently not logging)
  if err != nil{
    if errors.Is(err, context.Canceled) {
      v.logger.Info("Verification stopped", zap.String("trie", trieInfo.Name))
      return nil
    }
    return err
  }

  v.logger.Info("=== Trie verification completed successfully ===")
	return nil
}

func (v *TrieVerifier) verifyTrieWithLogging(...) error {
  ...
	if err != nil {
		if errors.Is(err, context.Canceled) {
      // do not log here
			return err
		}
		...
	}
	...
}

```

verify/trie/trie_verifier.go
```diff
@@ -0,0 +1,261 @@
...
+func (v *TrieVerifier) verifyTrie(ctx context.Context, trieInfo TrieInfo) error {
+	v.logger.Info(fmt.Sprintf("=== Starting %s verification ===", trieInfo.Name))
+	expectedRoot := felt.Zero
+	err := v.database.View(func(snap db.Snapshot) error {
+		reader, err := trieInfo.ReaderFunc(snap, trieInfo.Prefix, trieInfo.Height)
+		if err != nil {
+			return err
+		}
+		if reader.RootKey() == nil {
+			expectedRoot = felt.Zero
```
thiagodeev: Nitpick: There's no previous change, so it's already zero. Only `return nil` is sufficient

verify/trie/trie_verifier.go
```diff
@@ -0,0 +1,261 @@
...
+			return nil
+		}
+		expectedRoot, err = reader.Hash()
+		return err
+	})
+	if err != nil {
+		v.logger.Error(fmt.Sprintf("Failed to get stored root hash for %s: %v", trieInfo.Name, err))
```
thiagodeev: One more example about duplicate logs, because we are both log and returning the error. `verifyTrie()`, `verifyTrieWithLogging()`, and `verifyAll()` will all log the same error. Three logs.

`verifyTrie()`&emsp;&emsp;&emsp;&emsp;&emsp;   &nbsp;: "Failed to X".
`verifyTrieWithLogging()`: "Y failed: Failed to X"
`verifyAll()`&emsp;&emsp;&emsp;&emsp;&emsp;&emsp;&nbsp;: "Z failed: Y failed: Failed to X"

We need to fix it through the code :)

verify/trie/trie_verifier.go
```diff
@@ -0,0 +1,261 @@
...
+
+			if nodeKey.Len() == StarknetTrieHeight {
+				contractAddr := nodeKey.Felt()
+				contractAddresses = append(contractAddresses, contractAddr)
+			}
+		}
+		return nil
+	})
+	if err != nil {
+		return nil
```
thiagodeev: What about at least logging the error here, so we know what happened?

verify/trie/trie_verifier.go
```diff
@@ -0,0 +1,261 @@
...
+		v.logger.Info("Found contracts to verify",
+			zap.Int("count", len(contractAddresses)))
+	}
+
+	for i, contractAddress := range contractAddresses {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		default:
+		}
```
thiagodeev: For the case you don't know, here's something I learned recently :). For this case, this code could also be replaced with (no need to change, just FYI):

```go
if err := ctx.Err(); err != nil {
    return err
}
```

verify/trie/trie_core.go
```diff
@@ -0,0 +1,117 @@
...
+	ctx context.Context,
+	reader *trie.ReadStorage,
+	key *trie.BitArray,
+	parentKey *trie.BitArray,
+	height uint8,
+	hashFn crypto.HashFn,
+) (*felt.Felt, error) {
+	select {
+	case <-ctx.Done():
+		return nil, fmt.Errorf("verification cancelled: %w", ctx.Err())
```
thiagodeev: No need to add a new error message here since the current logic will just check if the error is context.Canceled and logs a new message, like [here](https://github.com/NethermindEth/juno/blob/73cb507d9efde3a8846b7ced01c899cb18a34726/verify/trie/trie_verifier.go#L150-L152). We can just return the error (or change the logic)

cmd/juno/verify/trie.go
```diff
@@ -0,0 +1,103 @@
...
+		Long:          `Verify trie integrity by rebuilding tries and comparing root hashes.`,
+		RunE:          runTrieVerify,
+		SilenceUsage:  true,
+		SilenceErrors: true,
+	}
+
+	cmd.Flags().StringSlice(
+		verifyTrieType,
+		nil,
+		"Trie types to verify (contract, class, contract-storage)."+
```
thiagodeev, line 31:
```suggestion
		"Trie types to verify (contract, class, contract-storage). "+
```

verify/trie/trie_verifier.go
```diff
@@ -0,0 +1,261 @@
...
+		addrBytes := contractAddress.Marshal()
+		prefix := db.ContractStorage.Key(addrBytes)
+		trieInfo := TrieInfo{
+			Name:     fmt.Sprintf("ContractStorage[%s]", contractAddress.String()),
+			Prefix:   db.ContractStorage.Key(addrBytes),
```
thiagodeev:
```suggestion
			Prefix:   prefix,
```

verify/trie/trie_verifier.go
```diff
@@ -0,0 +1,261 @@
...
+		v.logger.Info("Trie is empty (zero root)", zap.String("trie", trieInfo.Name))
+		return nil
+	}
+
+	v.logger.Info("Starting verification",
+		zap.String("trie", trieInfo.Name),
+		zap.String("expectedRoot", expectedRoot.String()))
+	storageReader := trie.NewReadStorage(v.database, trieInfo.Prefix)
+
+	err = VerifyTrie(ctx, storageReader, StarknetTrieHeight, trieInfo.HashFn, &expectedRoot)
```
thiagodeev: Nice catch from Claude:

  The TrieInfo struct has a Height field which is used for the reader creation (line 164), but the actual verification call ignores it and uses the constant directly. This makes trieInfo.Height partially dead and introduces an inconsistency - if someone sets a different height, the reader would use it but the verifier wouldn't. Should be trieInfo.Height.

cmd/juno/verify/trie.go
```diff
@@ -0,0 +1,103 @@
...
+	if err != nil {
+		return err
+	}
+
+	var tries []verifytrie.TrieType
+	if len(trieTypes) > 0 {
+		tries = make([]verifytrie.TrieType, len(trieTypes))
+		for i, t := range trieTypes {
+			tries[i] = verifytrie.TrieType(t)
+		}
```
thiagodeev: Found by Claude:

  User input is blindly cast to TrieType. Running juno verify trie --type foo would silently succeed doing nothing, since foo won't match any typeSet entry in Run(). Should validate against known types.

verify/trie/trie_verifier_test.go
thiagodeev, line 1: Unfortunately, due to my lack of understanding of the state and DB, I can't help very much with the test assertions, but here's the return from Claude Code's review. Take a look to see if it makes sense or not:

  

> The tests are decent structurally but have significant gaps in coverage.
>                                                                                                                                                                                               
>   verify/trie/trie_test.go   
>                                                                                                                                                                                               
>   What's covered: Valid state trie, valid class trie, corrupted trie, multiple trie types, empty trie. Good baseline.
> 
>   Issues:
> 
>   1. Weak success assertion in ValidStateTrie (lines 58-62)
> 
>   The last block after the verifier passes:
>   reader, err := trie.NewTrieReaderPedersen(testDB, prefix, StarknetTrieHeight)
>   storedHash, err := reader.Hash()
>   assert.True(t, rootHash.Equal(&storedHash))
>   This tests that the trie library stores hashes correctly, not the verifier. The verifier already passed on line 56. It's out of scope for this test.
> 
>   2. No test for ContractStorageTrie
> 
>   This is the most complex verification path (verifyContractStorageTries → collectContractAddresses → per-contract loop), and it has zero test coverage. It's the only path that uses the
>   iterator-based address collection, the --address filter, and the closure-based ReaderFunc. All the existing tests only cover ContractTrie and ClassTrie.
> 
>   3. No test for context cancellation
> 
>   There's a lot of context.Canceled handling code (5+ checks across the verifier), but no test verifying that cancelling the context actually stops verification cleanly.
> 
>   4. No test for collectContractAddresses
> 
>   This function has its own non-trivial logic (DB iteration, BitArray unmarshalling, filtering by height, error swallowing). No direct or indirect test.
> 
>   5. No test for root hash mismatch vs node corruption
> 
>   VerifyTrie has two distinct failure modes: "node corruption detected" (internal node hash doesn't match children) and "root hash mismatch" (final root doesn't match expected). The
>   corrupted trie test only triggers the first one. A test that corrupts the stored root hash (without touching nodes) would exercise the other path.
> 
>   6. EmptyTrie test has unused variable (line 216)
> 
>   trieStorage is created but never used - the if testTrie.RootKey() != nil block is skipped for empty tries. Not a bug but dead code in the test.
> 
>   cmd/juno/verify/trie_test.go
> 
>   What's covered: --address flag validation with 3 cases. Good table-driven structure.
> 
>   Issues:
> 
>   1. Success cases don't actually assert success (lines 74-77)
> 
>   } else if err != nil {
>       addrFlagErr := "--address flag can only be used with --type contract-storage"
>       assert.NotContains(t, err.Error(), addrFlagErr)
>   }
> 
>   For the two "should succeed" cases, the test only asserts "if there's an error, it shouldn't be the address-flag error." Any other error is silently accepted. This means these cases will
>   pass even if the command fails for completely unrelated reasons. The test is really only validating the negative case properly.
> 
>   I understand why - the DB is empty so the verifier will hit some downstream error. But this should at least be documented with a comment, or the test should be structured to only exercise
>   the flag validation logic (e.g., by mocking or isolating that code path).
> 
>   2. Missing test cases:
>   - Invalid --address format (e.g., --address notahex) - should test SetString parsing failure
>   - Invalid --type values (e.g., --type foo) - currently silently accepted with no error
>   - verifyAll (running juno verify without subcommand)
>   - openDB with non-existent path
> 
>   Summary
> 
>   The tests have good structure and cover the happy path well, but they're mostly happy-path with a single corruption scenario. The most critical gaps are: no ContractStorageTrie coverage
>   (the most complex path), no context cancellation tests (despite extensive cancellation handling code), and weak assertions in the "should succeed" CLI cases. The codecov report showing
>   ~59% patch coverage aligns with these gaps.
>

thiagodeev (changes requested):
Hey. Nice work!

I've made a few comments. Some of them are just nitpicks and can be easily ignored, but I believe the rest are relevant.
Unfortunately, due to my lack of knowledge of the state and DB, I couldn't help very much with the verification logic, only with general logic things.

verify/trie/types.go
MaksymMalicki: thanks! didn't notice it

verify/trie/trie_verifier.go
MaksymMalicki: agreed 👍

cmd/juno/verify/verify.go
MaksymMalicki: I changed this message just to say "Verification stopped". Whatever the error is (if its an actual error or mismatch detected), the verifier got stopped and the error explains the reason. I think its appropriate

verify/trie/trie_verifier.go
MaksymMalicki: I've added a top-level function called `handleResults`, which handles the outputs of the verification. This should delegate most of the logging to the top-level, enable to get rid of `verifyTrieWithLogging`

verify/trie/trie_verifier_test.go
MaksymMalicki: Thanks for this feedback, more in-depth tests suite is always spot on

rodrodros:
No resources to review and merge

PR #3338, opened 2025-12-22
feat(cmd): trie verifier in the cmd

MaksymMalicki:
This PR introduces the first verification command: trie verifier.
To use it, hit `juno verify trie --db-path <path>`. When no other flag is specified, this verifier runs with default config and scans all the contract, class and contracts storage tries residing in the DB. If the user wants to scan a particular group of tries, they can pass the `--type` flag, pointing to which tries should be verified.
For example:
`juno verify trie --db-path <path> --type contract-storage --type class` will omit the contract trie verification and will verify class and contract storage tries only.

### DIFF
--- a/cmd/juno/verify/trie.go
+++ b/cmd/juno/verify/trie.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"slices"
 	"sync"
 	"time"
 
@@ -19,6 +20,7 @@ import (
 const (
 	starknetTrieHeight  = 251
 	concurrencyMaxDepth = 8
+	verifyContractAddr  = "address"
 )
 
 type TrieType string
@@ -40,7 +42,13 @@ func verifyTrieCmd() *cobra.Command {
 	cmd.Flags().StringSlice(
 		verifyTrieType,
 		nil,
-		"Trie types to verify (state, class, contract). Can be specified multiple times. Empty = all.",
+		"Trie types to verify (contract, class, contract-storage). Can be specified multiple times. Empty = all.",
+	)
+
+	cmd.Flags().String(
+		verifyContractAddr,
+		"",
+		"Contract address to verify (only used with --type contract-storage). If not specified, all contract storage tries are verified.",
 	)
 
 	return cmd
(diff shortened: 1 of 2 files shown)


cmd/juno/verify/trie.go
```diff
@@ -0,0 +1,498 @@
...
+type TrieInfo struct {
+	Name       string
+	prefix     []byte
+	HashFunc   trie.NewTrieFunc
+	HashFn     crypto.HashFn
+	ReaderFunc func(db.KeyValueReader, []byte, uint8) (trie.TrieReader, error)
+	Height     uint8
+}
+
+func (v *TrieVerifier) Run(ctx context.Context, cfg Config) error {
```
rodrodros, line 153: I'm not sure I understand the need of the `Config` interface. What's the rationale behind it?

cmd/juno/verify/trie.go
```diff
@@ -0,0 +1,534 @@
...
+		if node.Right.IsEmpty() {
+			zero := felt.Zero
+			rightHash = &zero
+		} else {
+			h, err := verifyNode(ctx, reader, node.Right, key, height, hashFn)
+			if err != nil {
+				return nil, err
+			}
+			rightHash = h
+		}
```
infrmtcs, line 507: Can we write a common function for these logic? It can also be used above too.

cmd/juno/verify/trie.go
```diff
@@ -0,0 +1,534 @@
...
+	}
+
+	useConcurrency := key.Len() <= concurrencyMaxDepth
+	var leftHash, rightHash *felt.Felt
+	var leftErr, rightErr error
+
+	if useConcurrency {
+		var wg sync.WaitGroup
+		wg.Add(1)
+		go func() {
```
infrmtcs, line 457: nit: Use `wg.Go` instead.

cmd/juno/verify/trie.go
infrmtcs: In fact, I think this function can be refactored to something simpler. For example, it takes the key and the expected hash, then load the node to check if the hash matches expected hash and matches hash(left, right). Then enqueue left and right to continue processing. Recursively verifying node can be resource intensive.

cmd/juno/verify/trie.go
infrmtcs, line 1: I think the `cmd` package should contain only the logic for arg parsing. The logic should be in a separate `verify` package outside of `cmd`.

infrmtcs: I also think we can try splitting the file into multiple small files, to make it easier to follow.

PR #3339, opened 2025-12-23
feat: Enable zstd level 1 compression

infrmtcs:


db/pebblev2/option.go
```diff
@@ -51,3 +51,14 @@ func WithCompression(compression *block.CompressionProfile) Option {
 		return nil
 	}
 }
+
+var Zstd1 = func() block.CompressionProfile {
+	zstd1 := block.BalancedCompression.ValueBlocks
```
EgeCaner, line 56: Is this only to take `compression.ZstdLevel1` from the predefined compressionProfile? If so can we directly use compression.ZstdLevel1? Is it missing the registry or something?

db/pebblev2/upgrade_test.go
```diff
@@ -95,7 +94,7 @@ func runScenario(t *testing.T, scenario scenario) {
 	if scenario.v2 {
 		name += "v2 -> "
 	}
-	name += "v2zstd"
+	name += "v2good"
```
EgeCaner, line 97: Where this name is coming from?

EgeCaner (commented):
Looks good, dropped some questions

db/pebblev2/upgrade_test.go
infrmtcs: My bad, this PR was created by modifying my "good compression" experiment branch.

PR #3341, opened 2025-12-23
feat: Add Close() to batch and call on Write()

infrmtcs:


db/pebble/batch.go
```diff
@@ -110,10 +110,19 @@ func (b *batch) Write() error {
...
 		return pebble.ErrClosed
 	}
-	return b.batch.Commit(pebble.Sync)
+
+	if err := b.batch.Commit(pebble.Sync); err != nil {
+		return err
+	}
+
+	return b.batch.Close()
 }
```
rodrodros, line 114: Could you please add a new a test asserting the new behaviour of Write

infrmtcs: I can add more assertion to check that a subsequent `Close()` call will throw an error, but there isn't anything else to check, because the close status is not exposed by pebble.

infrmtcs:
> Do we never use batch again after writing?

For pebble docs (I removed the unrelated parts):
```
	// [...] A Batch and its
	// underlying batchInternal.data byte slice may be reused. There are two
	// mechanisms for reuse:
	//
	// 1. The caller may explicitly call [Batch.Reset] to reset the batch to be
	//    empty (while retaining the underlying repr's buffer).
	// 2. The caller may call [Batch.Close], passing ownership off to Pebble,
	//    which may reuse the batch's memory to service new callers to
	//    [DB.NewBatch].
	//
	// There's a complication to reuse: When WAL failover is configured, the
	// Pebble commit pipeline may retain a pointer to the batch.data beyond the
	// return of [Batch.Commit]. The user of the Batch may commit their batch
	// and call Close or Reset before the commit pipeline is finished reading
	// the data slice. Recycling immediately would cause a data race.
	//
	// [...]
	//
	// When the committer of a batch indicates intent to recycle a Batch through
	// calling [Batch.Reset] or [Batch.Close], the lifecycle atomic is read. If
	// an outstanding reference remains, it's unsafe to reuse Batch.data yet. In
	// [Batch.Reset] the caller wants to reuse the [Batch] immediately, so we
	// discard b.data to recycle the struct but not the underlying byte slice.
	// [..]
```

So the main difference between these 2 mechanism is: (1) reuse the batch immediately, while (2) reuse it somewhere else. However, both doesn't guarantee that the underlying data is reused, if the commit pipeline decides to retain its ownership, or the size of the batch is larger than a threshold that pebble decides to let GC to collect it instead of keep it. In my opinion, we don't need (1) because (2) can do the job anyway. So here are the expected behaviours:
- If we want to write: call `Write()` and don't use the batch anymore. Reusability is done by `NewBatch`.
- If we don't want to write, but want to reuse the batch: For now we can either call `Reset()` or `Close()` then create a new batch, but I think we should deprecate `Reset()`.
- If we don't want to write: `Close()`

db/pebble/batch.go
rodrodros: > I can add more assertion to check that a subsequent Close() call will throw an error

I think this will be a good addition.

db/syncbatch.go
```diff
@@ -71,3 +71,9 @@ func (s *SyncBatch) Size() int {
 	defer s.lock.RUnlock()
 	return s.batch.Size()
 }
+
+func (s *SyncBatch) Close() error {
+	s.lock.RLock()
```
EgeCaner: Is `RLock` is the correct lock to acquire here?

infrmtcs: My bad 👍 It should be Lock()

db/pebble/batch.go
infrmtcs: Actually it's unexpected behaviour to call Close twice, because it the database doesn't need to retain anything, the first `Close()` call immediately recycle the batch to be used in future `NewBatch` call, so we can't really test it here.

db/syncbatch.go
infrmtcs: Updated

db/testutil.go
```diff
@@ -712,3 +708,80 @@ func TestKeyValueStoreSuite(t *testing.T, newDB func() KeyValueStore) {
 		})
 	})
 }
+
+func assertIteratorErrorsAfterClose(t *testing.T, it Iterator) {
+	t.Helper()
+	require.NoError(t, it.Close(), "iterator close should fail after close")
```
EgeCaner: Error message does not match with assertion.

PR #3342, opened 2025-12-23
feat: Add UncopiedValue() to iterators

infrmtcs:


db/pebblev2/iterator.go
```diff
@@ -36,6 +36,16 @@ func (i *iterator) Value() ([]byte, error) {
...
 
+// DO NOT USE this if you don't unmarshal the value immediately.
+// See [db.Iterator] for more details.
+func (i *iterator) UncopiedValue() ([]byte, error) {
+	val, err := i.iter.ValueAndErr()
+	if err != nil || val == nil {
+		return nil, err
+	}
+	return val, nil
+}
```
rodrodros: Why not?

```suggestion
return i.iter.ValueAndErr()
```

The output of the function doesn't change

db/iterator.go
```diff
@@ -28,6 +28,16 @@ type Iterator interface {
...
 	Value() ([]byte, error)
 
+	// DO NOT USE this if you don't unmarshal the value immediately.
+	//
+	// UncopiedValue returns the value at the current position without copying it to a new slice.
+	// The returned byte slice is invalid after the next call to [Next], [Prev], or [Seek].
+	// Therefore it's the caller's responsibility to copy the value to not use the byte slice
+	// after its iteration or copy it to a new slice (e.g, in []byte case).
+	// This function DOES NOT copy the value to a new slice. This is useful when the typed bucket
+	// calls unmarshal the value immediately.
```
rodrodros: I think the clarity of this text can be improved a bit. 

A suggestion:

```suggestion
    // DO NOT USE this unless the value is consumed immediately.
    //
    // UncopiedValue returns the value at the current position without copying it.
    // The returned slice is invalidated by the next call to [Next], [Prev], or [Seek].
    // Callers must copy the value if it needs to outlive the current iteration.
    //
    // This is intended for immediate unmarshalling to avoid an extra allocation.
```

db/pebblev2/iterator.go
infrmtcs: Updated

db/iterator.go
infrmtcs: Updated

EgeCaner (approved):

PR #3343, opened 2025-12-23
feat: Add indexed LazySlice

infrmtcs:


core/indexed/lazy_slice.go
```diff
@@ -0,0 +1,46 @@
...
+
+import (
+	"github.com/NethermindEth/juno/db"
+	"github.com/NethermindEth/juno/encoder"
+)
+
+type LazySlice[T any] struct {
+	indexes []int
+	data    []byte
+}
```
rodrodros, line 20: A small description will help other understand this data structure easily

core/indexed/lazy_slice.go
```diff
@@ -0,0 +1,46 @@
...
+func (l LazySlice[T]) All() ([]T, error) {
+	var err error
+	items := make([]T, len(l.indexes))
+	for i := range l.indexes {
+		if items[i], err = l.Get(i); err != nil {
+			return nil, err
+		}
+	}
+	return items, nil
+}
```
rodrodros, line 57: I wanted to see if there was a way for us to inline the `Get` call by dividing it between `Get` and `unsafeGet` to avoid rechecking the same bounds. 

It seems there is no way to do it because there is no split that is simple enough for that. 

What I would propose but I leave it up to your discretion is writing similar code from the `Get` method in the `All` method, avoiding the extra function call as well as the continuous bound checks.

core/indexed/lazy_slice.go
```diff
@@ -0,0 +1,46 @@
...
+
+	if index < 0 || index >= len(l.indexes) {
+		return value, db.ErrKeyNotFound
+	}
+
+	start := l.indexes[index]
+	end := len(l.data)
+	if index < len(l.indexes)-1 {
+		end = l.indexes[index+1]
+	}
```
rodrodros, line 34: I would remove this `if` by having instead of `indexes` of size `n`, have it of size `n + 1`, so that `indexes[n] == len(l.data)`. 

I believe this solution to cost one bit more memory but will make the code easier to read.

infrmtcs: len(indexes) is the number of transactions/receipts in the block, so I think adding one more element causes more confusion than this if.

core/indexed/lazy_slice.go
infrmtcs: Yeah, I think the second way sounds good, we can also avoid the condition for computation for end too.

core/indexed/lazy_slice.go
rodrodros: In the context that is being used it might not make too much sense. The other thing that I see to avoid having this if is growing by the size of `T` in bytes. It can be stored as part of the struct. You can have the size of T, which is the difference between `l.data[i + 1] - l.data[i]`.

infrmtcs: Items are stored contiguously in one byte slice, with an `indexes` slice marking each item’s start offset.

infrmtcs: In my opinion, using next item start as current item end is as implicit as using end of data as end of the last item. If we don't store the end of all items, why do we store the end of the last item?

core/indexed/indexed.go
```diff
@@ -0,0 +1,40 @@
...
+		Buffer:  &buf,
+		Encoder: encoder.NewEncoder(&buf),
+	}
+}
+
+func Write(
+	writer BufferedEncoder,
+	items iter.Seq2[any, error],
+) ([]int, error) {
+	indexes := make([]int, 0)
```
rodrodros, line 27: Is there anyway for us to know the size of the elements of the Buffer in advance, so that we can pre-plan and build the `indexes` array with the right size. 

Maybe this is not true because same type of elements can have different sizes when serialized and hence is not possible.

core/indexed/lazy_slice.go
rodrodros: Mmm, I guess this goes more to perspective, you can also hold the end of all items and never the beginning, since the beginning is always zero. 
You can avoid the sum altogether by knowing that `indexes[i]` represents the ends of `i` and not the beginning.

rodrodros: I think with this approach above you can remove the need for the if while carrying the same semantic meaning

infrmtcs: Firstly, even if it's possible, doing this (in my opinion) has the same readability.
```suggestion
	start := 0
	if index > 0 {
		start = l.indexes[index-1]
	}
```
Secondly, the receipts doesn't start from 0, but starts after the end of transactions.

rodrodros: For the above I was thinking in the `All` scenario, for `Get` it doesn't apply. 

The goal is removing was checking if the `if` could be removed but it seems it can't. 

> Secondly, the receipts doesn't start from 0, but starts after the end of transactions.

I see

infrmtcs:
```
transactions:  [00    02          06]
                ↓     ↓           ↓
                ----- ----------- -----------
data:          [00|01|02|03|04|05|06|07|08|09|10|11|12|13|14|15|16|17|18|19]
                                              ----- -------- --------------
                                              ↑     ↑        ↑              ↑
receipts:                                    [10    12       15]            20 (len(data))
```

infrmtcs: Regarding adding another item to `indexes`, we'll have to do these in order to do it:
- Bound check needs to compare `index + 1< len(indexes)`
- For range indexes needs to break in the last iterations, or we use `for i := 0; i + 1 < len(indexes); i++`
- Remember to write `nil` to `indexes` everywhere we construct it.

rodrodros (approved):

core/indexed/lazy_slice.go
infrmtcs: Added description with an example

core/indexed/indexed.go
infrmtcs: We don't know the length of `items` in advance based on the input, so probably not possible.

core/indexed/lazy_slice.go
infrmtcs: Updated by adding `get`.

PR #3348, opened 2025-12-26
feat: Add bucket for BlockTransactions

infrmtcs:


migration/blocktransactions/blocktransactions.go
```diff
@@ -0,0 +1,391 @@
...
+	err = core.ReceiptsByBlockNumberAndIndexBucket.Prefix().DeleteRange(
+		database,
+		startBlock,
+		endBlock+1,
+	)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
EgeCaner:
```suggestion
	return core.ReceiptsByBlockNumberAndIndexBucket.Prefix().DeleteRange(
		database,
		startBlock,
		endBlock+1,
	)
```

core/block_transaction.go
```diff
@@ -0,0 +1,115 @@
...
+) (BlockTransactions, error) {
+	return NewBlockTransactionsFromIterators(
+		wrapWithNilError(transactions),
+		wrapWithNilError(receipts),
+	)
+}
+
+func (b *BlockTransactions) Transactions() indexed.LazySlice[Transaction] {
+	end := len(b.Data)
+	if len(b.Indexes.Receipts) > 0 {
```
EgeCaner, line 87: Should it be allowed to txs and receipts counts to mismatch or one being absent? Isn't this is an unexpected state?

migration/blocktransactions/blocktransactions.go
```diff
@@ -0,0 +1,393 @@
...
+
+	fetchedTxCount := len(blockTransactions.Indexes.Transactions)
+	fetchedReceiptCount := len(blockTransactions.Indexes.Receipts)
+
+	if fetchedTxCount == 0 || fetchedReceiptCount == 0 {
+		has, err := core.BlockTransactionsBucket.Has(database, blockNumber)
+		if err != nil {
+			return err
+		}
+		// ALready migrated
```
EgeCaner:
```suggestion
		// Already migrated
```

migration/blocktransactions/blocktransactions.go
```diff
@@ -0,0 +1,393 @@
...
+	return func(yield func(cbor.RawMessage, error) bool) {
+		for item, err := range seq {
+			if !yield(cbor.RawMessage(item.Value), err) {
+				return
+			}
+		}
+	}
+}
+
+func readBlock(
```
EgeCaner: As far as I understood this function does more than reading the data. It reads the data from old buckets, transforms the data and writes to batch. IMO we can rename it to better reflect what it does.

migration/blocktransactions/blocktransactions.go
```diff
@@ -0,0 +1,393 @@
...
+		var blockNumIndex db.BlockNumIndexKey
+		if err := blockNumIndex.UnmarshalBinary(item.Key[1:]); err != nil {
+			return 0, false, err
+		}
+		return blockNumIndex.Number, true, nil
+	}
+	return 0, false, nil
+}
+
+func getFirstBlockToMigrate(
```
EgeCaner: Shouldn't txs and receipts should always belong to same block number, if there is a mismatch, isn't this is an unexpected state since we cant move txs and receipts partially for a block.

migration/blocktransactions/blocktransactions.go
```diff
@@ -0,0 +1,393 @@
...
+			return 0, false, err
+		}
+
+		var blockNumIndex db.BlockNumIndexKey
+		if err := blockNumIndex.UnmarshalBinary(item.Key[1:]); err != nil {
+			return 0, false, err
+		}
+		return blockNumIndex.Number, true, nil
+	}
+	return 0, false, nil
```
EgeCaner: How about consuming iter manually here to avoid loop, since as far as I understood loop iterates only once, and loop is there for calling `next` once

```suggestion
	next, stop := iter.Pull2(items)
	defer stop()
	item, err, ok := next()
	if !ok {
		return 0, false, nil
	}

	if err != nil {
		return 0, false, err
	}

	var blockNumIndex db.BlockNumIndexKey
	if err := blockNumIndex.UnmarshalBinary(item.Key[1:]); err != nil {
		return 0, false, err
	}
	return blockNumIndex.Number, true, nil
```

migration/blocktransactions/blocktransactions.go
```diff
@@ -0,0 +1,393 @@
...
+				endBlock := min(startBlock+batchSize-1, chainHeight)
+
+				txCount, err := createBatch(logger, database, startBlock, endBlock, batch)
+				if err != nil {
+					taskCh <- writeTask{err: err}
+					return
+				}
+
+				totalTxCount += txCount
+				if batch.Size() >= targetBatchByteSize {
```
EgeCaner: If Im not mistaken `batch.Size` does not accounts for `DeleteRange` queries

migration/blocktransactions/blocktransactions.go
```diff
@@ -0,0 +1,393 @@
...
+	return minBlock, true, nil
+}
+
+func clearOldBuckets(database db.KeyValueStore) error {
+	err := core.TransactionsByBlockNumberAndIndexBucket.Prefix().DeletePrefix(database)
+	if err != nil {
+		return err
+	}
+
+	err = core.ReceiptsByBlockNumberAndIndexBucket.Prefix().DeletePrefix(database)
```
EgeCaner: I guess we can directly return here

migration/blocktransactions/blocktransactions.go
```diff
@@ -0,0 +1,393 @@
...
+		}
+
+		firstBlock, shouldMigrate, err := getFirstBlockToMigrate(database)
+		if err != nil {
+			return nil, err
+		}
+
+		if !shouldMigrate {
+			logger.Infow("no starting block found, exiting")
+			return nil, clearOldBuckets(database)
```
EgeCaner: If `!shouldMigrate`  is true, shouldnt this mean old bucket are empty, why do we need to clear them

migration/blocktransactions/blocktransactions.go
infrmtcs: Updated

migration/blocktransactions/blocktransactions.go
infrmtcs: Updated

migration/blocktransactions/blocktransactions.go
infrmtcs: Updated

migration/blocktransactions/blocktransactions.go
infrmtcs: I renamed it to `ingestOldData`

migration/blocktransactions/blocktransactions.go
infrmtcs: Updated

migration/blocktransactions/blocktransactions.go
infrmtcs: That's correct, but we actually don't need the exact size, but rather an estimation to terminate early. On the other hand, we should replace our own `size` counter by using pebble batch `Len()`, but it's outside the scope of this PR.

migration/blocktransactions/blocktransactions.go
infrmtcs: Actually I tried to "hint" the DB to do compaction. I need to revisit whether this is effective though.

core/block_transaction.go
rodrodros: I believe the question should be more, do we need to check that data we stored initially is correct when reading from it, or do we assume our code was correctly stored and wasn't tampered with?

EgeCaner: Yes, if every tx has corresponding receipt, we can just check if end is greater than zero for non-empty check and assume receipts[0] is valid, but I with second thought, current approach seems fine

PR #3355, opened 2025-12-30
feat: Implement migration script for combined layout

infrmtcs:


migration/blocktransactions/check_status.go
```diff
@@ -0,0 +1,65 @@
...
+	)
+	if err != nil {
+		return 0, false, err
+	}
+
+	if !hasReceipts && !hasTransactions {
+		return 0, false, nil
+	}
+
+	minBlock := min(transactions, receipts)
```
EgeCaner, line 32: How we will proceed if`transactions` != `receipts`? Seems like this will fail`validateCount`and we can exit early here since `transactions` != `receipts` should hint DB corruption.

migration/blocktransactions/blocktransactions.go
```diff
@@ -0,0 +1,155 @@
...
+		}
+
+		firstBlock, shouldMigrate, err := getFirstBlockToMigrate(database)
+		if err != nil {
+			return shouldRerun, err
+		}
+
+		if !shouldMigrate {
+			logger.Infow("no starting block found, exiting")
+			return shouldNotRerun, clearOldBuckets(database)
```
EgeCaner, line 146: Does `clearOldBuckets` had positive effect? If so can you add brief comment why we call it while `!shouldMigrate` is hinting buckets are already empty

migration/blocktransactions/batch_provider.go
```diff
@@ -0,0 +1,39 @@
...
+		totalBlockCount: 0,
+	}
+}
+
+func (b *batchProvider) put() {
+	b.sem <- struct{}{}
+}
+
+func (b *batchProvider) close() {
+	close(b.sem)
```
EgeCaner: Not sure if this is an issue, but I guess after `close` is called, `get` will always immediately returns successfully.

migration/blocktransactions/blocktransactions_test.go
```diff
@@ -0,0 +1,269 @@
...
+			t.Context(),
+			database,
+			&utils.Sepolia,
+			utils.NewNopZapLogger(),
+		)
+		require.NoError(t, err)
+		require.Nil(t, state)
+	})
+
+	t.Run("Generate data", func(t *testing.T) {
```
EgeCaner, line 50: Should it be helper function instead? I am not sure if it is testing something.

migration/blocktransactions/blocktransactions_test.go
```diff
@@ -0,0 +1,269 @@
...
+func runTestBlockTransactionsMigration(
+	t *testing.T,
+	counts []int,
+	allTransactions [][]core.Transaction,
+	allReceipts [][]*core.TransactionReceipt,
+	shouldCancelRuns []bool,
+) {
+	dir := t.TempDir()
+
+	t.Run("Init", func(t *testing.T) {
```
EgeCaner, line 88: Should this also be helper function instead? Looks more like a fixture rather than test

migration/blocktransactions/blocktransactions_test.go
```diff
@@ -0,0 +1,269 @@
...
+						batch,
+						key,
+						allReceipts[blockNumber][i],
+					),
+				)
+			}
+		})
+	})
+
+	t.Run("Read old data", func(t *testing.T) {
```
EgeCaner, line 138: This looks like a post verification for the fixture above. IMO extracting these into functions will make this function easier to read.

migration/blocktransactions/blocktransactions_test.go
infrmtcs: You're right that it's not testing anything, it's just a step that takes substantial amount of time, so I want to put a `t.Run` there so it can show some log. An alternative is to use `t.Log` instead.
The reason why I want to generate a lot of data is to make sure that we need to run multiple batches sequentially in order to migrate the whole database.

migration/blocktransactions/blocktransactions_test.go
infrmtcs: Yeah, probably I'll remove Init

PR #3283, opened 2025-11-20
refactor: use a new unified Trie and State interfaces

MaksymMalicki:
This PR should be reviewed after #3282 . It aims to:
- Move the `State` and `Trie` interfaces from `commonstate` and `commontrie` packages to `core` package and rename them to `CommonState` and `CommonTrie`. This way the circular imports in `pending_state.go` are omitted and there is no naming clash.
- Use the `CommonState` and `CommonTrie` in Juno. Right now both old and new implementations of State and Trie use these interfaces.  It is possible to utilise the new interfaces in Juno, but always feed the old implementations to them. The CMD flag, specifying which set of implementations to use will be introduced in the next PR.
- Implement the missing `CompiledClassHash()` and `CompiledClassHashV2()` methods for the new `State`
- Rename `MockStateHistoryReader` and all its methods to `MockCommonState`

The next PR to review: #3360

PR #3360, opened 2026-01-05
refactor(state): remove DB writes from the new state to make writes atomic

MaksymMalicki:
This PR is to be reviewed after: #3283 

Up to this point, the new state was writing to the database internally. This caused a problem with write atomicity, since 3 separate batches were used to write processed block information in `blockchain.Store`.
The state internally:
- Wrote the trie nodes to the trieDB in one batch and
- wrote the remainder of the state update information in second batch;
- in parallel!
- The remainder of the block information was then written in yet another batch in `blockchain.store()`.

If there was an error in any of these flushes to disk (batch.Write()) data corruption was prone to happen easily. 

Now, the new State was modified to follow the flow of the old State in the DB communication.
Only one batch is used to write all the `blockchain.Store()` information, as the new `State` ingests a batch, writes the information to it sequentially. Note that parallel writes of `trieDB` data and the remainder of the state update information is no longer possible, as pebbleDB Batch is not thread safe, but the batch is not committed to the DB.

Next PR to review: #3187


core/trie2/triedb/rawdb/database.go
```diff
@@ -83,12 +83,11 @@ func (d *Database) Update(
 	blockNum uint64,
 	mergedClassNodes *trienode.MergeNodeSet,
 	mergedContractNodes *trienode.MergeNodeSet,
+	batch db.Batch,
```
EgeCaner, line 66: Since caller controls the batch, cache can expose uncommitted data: nodes are added to cache before batch.Write() completes, after the lock is released, till batch is committed, any concurrent reads will serve stale entries from cache instead of the true disk state. Also batch.write can fail or caller might not commit the batch due any other reason, leaving cache dirty.

core/trie2/triedb/rawdb/database.go
```diff
@@ -112,6 +112,10 @@ func (d *Database) updateNode(
 	path *trieutils.Path,
 	n trienode.TrieNode,
 ) error {
+	if batch == nil {
```
EgeCaner: IMO we shouldn't allow calling with nil batch, otherwise calling `database.Update` is redundant. Caller shouldn't call if they don't want to write

core/state/state.go
EgeCaner: Should we update the cache if batch is nil (dry-run)?

EgeCaner: Cache can contain uncommitted data, but data is accessed by the state root, reading the data should be safe I guess, but does it breaks the layered approach or anything else?

EgeCaner (commented):
Dropped couple comments

core/trie2/triedb/rawdb/database.go
MaksymMalicki: The clean cache idea is halted for now

core/state/statefactory/state_factory.go
```diff
@@ -73,7 +74,7 @@ func (sf *StateFactory) EmptyState() (core.StateReader, error) {
 		emptyState := core.NewDeprecatedState(txn)
 		return emptyState, nil
 	}
-	state, err := state.New(&felt.Zero, sf.stateDB)
+	state, err := state.New(&felt.Zero, sf.stateDB, nil)
 	if err != nil {
 		return nil, err
 	}
```
rodrodros: What's the point of this constructor that never gets called?

core/state/state.go
```diff
@@ -497,74 +500,64 @@ func (s *State) flush(
...
+	if err := s.db.triedb.Update(
+		(*felt.StateRootHash)(&update.curComm),
+		(*felt.StateRootHash)(&update.prevComm),
+		blockNum,
+		update.classNodes,
+		update.contractNodes,
+		s.batch,
+	); err != nil {
+		return err
+	}
```
rodrodros: gostyle: err outside of the if statement please. Unnecessary coupling

core/state/history.go
```diff
@@ -15,7 +15,7 @@ type stateHistory struct {
 }
 
 func NewStateHistory(blockNum uint64, stateRoot *felt.Felt, db *StateDB) (stateHistory, error) {
-	state, err := New(stateRoot, db)
+	state, err := New(stateRoot, db, nil)
```
rodrodros: I've seen multiple instances of the `New` constructor used with a `nil` argument. This is a major flag that something is not right.

Why would you use a constructor with an optional parameter that has no use, or if it has use, it is unknown to the reader. 

For now, I am guessing, that passing a batch means that the new state will write to disk or not. If that's the case then why not just having a new constructor called `NewReader` or something like that with the appropriate description, instead of populating every instance of `New` (except 1 or 2) with `nil`.

core/trie2/triedb/rawdb/database.go
rodrodros: This check is redundant, looking at this function callers this nil guard will never be true.

Additionally, on the topic of "nil guards" inside functions, I don't think it is a good idea to have a safe output when an incorrect parameter is passed, it should error / panic – specially if this an internal code API. 

If there is going to be a guard at all, it's better for it to be outside the function call. Improving readability through explicit code. Users don't need to check the function to see what happens when the value is nil. (I agree that there exceptions to these, but I don't consider this one)

core/trie2/triedb/database.go
```diff
@@ -10,6 +10,12 @@ import (
 	"github.com/NethermindEth/juno/db"
 )
 
+const (
+	PathScheme string = "path"
+	HashScheme string = "hash"
+	RawScheme  string = "raw"
+)
+
```
rodrodros: Why does this constant exists?

Why are they strings and not formal enums?

core/trie2/triedb/pathdb/database.go
```diff
@@ -75,6 +75,7 @@ func (d *Database) Update(
 	blockNum uint64,
 	mergeClassNodes,
 	mergeContractNodes *trienode.MergeNodeSet,
+	_ db.Batch, // todo(maksym): added for interface compatibility, not used
```
rodrodros: Why is this a `todo`? Is there plans to act on it on the future?

core/trie2/triedb/hashdb/database.go
```diff
@@ -192,6 +192,7 @@ func (d *Database) Update(
 	blockNum uint64,
 	mergedClassNodes *trienode.MergeNodeSet,
 	mergedContractNodes *trienode.MergeNodeSet,
+	_ db.Batch, // todo(maksym): added for interface compatibility, not used
```
rodrodros: Similar question

core/trie2/triedb/pathdb/database.go
rodrodros: A todo should leave an action to be done, not a note, this is more like a normal comment.

MaksymMalicki: yea, removed todo

core/trie2/triedb/rawdb/database.go
MaksymMalicki: Actually, I think this guard should error if batch is nil

core/state/state.go
MaksymMalicki: removed the cache. brings low benefit and unnecessary complication and inconsistencies now that the caller is responsible for the DB communication

core/state/history.go
MaksymMalicki: I think the state shouldn't internally differentiate if the batch is nil or not. That is not its responsibility. The responsibility of the state is to process the state update / reverse update and serve proofs/leaves data/historical leaves data. It's the responsibility of the caller to either flush it or not. Hence I decided, that we should enforce the condition, that the batch is never `nil` in the state. If the caller decides not to flush the data, then the batch can simply be discarded.

core/state/statefactory/state_factory.go
MaksymMalicki: I imagined its going to be used widely in the follow-up PR. But turns out it can be removed cleanly and replaced with just the `StateFactory.NewState` since I enforce the batch in the new state

core/state/history.go
```diff
@@ -15,7 +15,8 @@ type stateHistory struct {
 }
 
 func NewStateHistory(blockNum uint64, stateRoot *felt.Felt, db *StateDB) (stateHistory, error) {
-	state, err := New(stateRoot, db)
+	batch := db.disk.NewBatch()
+	state, err := New(stateRoot, db, batch)
```
rodrodros: A batch that is never used seems like bad coding practice to me. I am not convinced that this approach is good either

core/state/state.go
```diff
@@ -43,9 +43,14 @@ type State struct {
...
 
 	stateObjects map[felt.Felt]*stateObject
+
+	batch db.Batch
 }
 
-func New(stateRoot *felt.Felt, db *StateDB) (*State, error) {
+func New(stateRoot *felt.Felt, db *StateDB, batch db.Batch) (*State, error) {
+	if batch == nil {
+		return nil, errors.New("batch is nil")
```
rodrodros: Improve error description: "cannot create state, nil Batch received". 

I'm not sure it makes sense to have a non nil batch since there are uses for this type, which don't require a db.Batch

core/trie2/triedb/rawdb/database.go
```diff
@@ -112,6 +113,10 @@ func (d *Database) updateNode(
...
+		return errors.New("updateNode called with nil batch")
+	}
```
rodrodros: Remove the guard please, this is internal API. I believe we should trust our code

core/state/history.go
MaksymMalicki: created a `NewStateReader` constructor, which doesn't take the batch

core/state/state.go
MaksymMalicki: created a NewStateReader constructor, which doesn't take the batch

core/trie2/triedb/rawdb/database.go
MaksymMalicki: removed the guard

core/state/state.go
```diff
@@ -63,14 +68,31 @@ func New(stateRoot *felt.Felt, db *StateDB) (*State, error) {
...
 		classTrie:    classTrie,
 		stateObjects: make(map[felt.Felt]*stateObject),
+		batch:        batch,
 	}, nil
 }
 
-func (s *State) ContractClassHash(addr *felt.Felt) (felt.Felt, error) {
-	if classHash := s.db.stateCache.getReplacedClass(&s.initRoot, addr); classHash != nil {
-		return *classHash, nil
+func NewStateReader(stateRoot *felt.Felt, db *StateDB) (*State, error) {
```
rodrodros, line 80: Description for this functions plesae

core/trie2/triedb/database.go
```diff
@@ -10,6 +10,12 @@ import (
 	"github.com/NethermindEth/juno/db"
 )
 
+const (
+	PathScheme = iota + 1
+	HashScheme
+	RawScheme
+)
```
rodrodros: Question stands, why does this constant exists? They seem to be not being used anywhere?

core/state/history_test.go
```diff
@@ -143,9 +143,11 @@ func TestStateHistoryClassOperations(t *testing.T) {
 		NewRoot:   &felt.Zero,
 		StateDiff: &core.StateDiff{},
 	}
-	state, err := New(&felt.Zero, stateDB)
+	batch := stateDB.disk.NewBatch()
+	state, err := New(&felt.Zero, stateDB, batch)
 	require.NoError(t, err)
 	err = state.Update(&core.Header{Number: 0}, stateUpdate, classes, false)
+	require.NoError(t, batch.Write())
```
rodrodros, line 150: Why does this require a state and not a reader?

state.test
rodrodros, line 1: i don't think this file should be commited

MaksymMalicki: my bad

core/state/history_test.go
MaksymMalicki: I'm not sure I understand - we need a writable state to utilise the `state.Update`, the Update ran on `StateReader` without a batch would fail. Technically I could spawn a new reader to calculate the commitment after the Update, but it seems like an overkill to do this while (just) preparing the `memDB` for history tests.

PR #3361, opened 2026-01-06
fix: Properly track db application layer metrics

infrmtcs:


node/metrics.go
```diff
@@ -54,20 +54,6 @@ func makeDBMetrics() db.EventListener {
 		Namespace: "db",
 		Name:      "commit_latency",
 		Help:      "Database transaction commit latency in microseconds",
```
EgeCaner: As I see below `commitLatency.Observe(duration.Seconds())` now rather than microseconds? Should we update this description as well?

infrmtcs: You're right, I'll update the description.

PR #3362, opened 2026-01-06
fix: Track metrics for iterator

infrmtcs:


db/pebble/iterator.go
```diff
@@ -49,27 +53,27 @@ func (i *iterator) UncopiedValue() ([]byte, error) {
...
 	return i.iter.Prev()
 }
 
 func (i *iterator) Next() bool {
 	if !i.positioned {
-		i.positioned = true
-		return i.iter.First()
+		return i.First()
 	}
 	return i.iter.Next()
```
EgeCaner, line 74: Do we need to measure `i.iter.Next` and `i.iter.Prev` as well?

db/pebblev2/iterator.go
```diff
@@ -49,27 +53,27 @@ func (i *iterator) UncopiedValue() ([]byte, error) {
...
 	return i.iter.Prev()
 }
 
 func (i *iterator) Next() bool {
 	if !i.positioned {
-		i.positioned = true
-		return i.iter.First()
+		return i.First()
 	}
 	return i.iter.Next()
```
EgeCaner, line 74: Same here

db/pebble/iterator.go
infrmtcs: I aimed to track only once per iteration, so I avoided tracking Next and Prev for now.

PR #3367, opened 2026-01-08
fix: l1 handler message hash to txn hash mapping

EgeCaner:
Due to a bug, only the first L1Handler transaction in a block was written to MsgHash to L2TxnHash mapping, resulting in consecutive ones to be missing, leading to to error in `starknet_getMessageStatus`.

Includes migration

### DIFF
--- a/migration/l1handlermapping/ingestor.go
+++ b/migration/l1handlermapping/ingestor.go
@@ -14,14 +14,14 @@ type ingestor struct {
 	database        db.KeyValueReader
 	batchSemaphore  semaphore.ResourceSemaphore[db.Batch]
 	batches         []db.Batch
-	progressTracker *progresslogger.BlockNumberProgressTracker
+	progressTracker *progresslogger.BlockProgressTracker
 }
 
 func newIngestor(
 	database db.KeyValueReader,
 	batchSemaphore semaphore.ResourceSemaphore[db.Batch],
 	maxWorkers int,
-	progressTracker *progresslogger.BlockNumberProgressTracker,
+	progressTracker *progresslogger.BlockProgressTracker,
 ) *ingestor {
 	batches := make([]db.Batch, maxWorkers)
 	for i := range batches {
--- a/migration/l1handlermapping/migrator.go
+++ b/migration/l1handlermapping/migrator.go
@@ -53,13 +53,10 @@ func (m *Migrator) Migrate(
 		return nil, err
 	}
 
-	startFrom := m.startFrom
-	startTime := time.Now()
-
-	if startFrom > 0 {
+	if m.startFrom > 0 {
 		log.Infow("Resuming L1 handler message hash migration",
 			"chain_height", chainHeight,
-			"from_block", startFrom,
+			"from_block", m.startFrom,
 		)
 	} else {
 		log.Infow("Starting L1 handler message hash migration",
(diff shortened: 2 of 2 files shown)


migration/migration.go
rodrodros, line 1: is it possible to have tests 👀

EgeCaner: Sure, let me see what I can do

thiagodeev (approved):
LGTM.

By generating a benchmark with cursor, it seems the approach with a buffered channel is at least 15% faster. It's a one-line change, WDYT about it?

Image

migration/migration.go
```diff
@@ -993,3 +994,70 @@ func calculateCasmClassHashesV2(txn db.IndexedBatch, network *utils.Network) err
...
+		for blockNum := range chainHeight + 1 {
+			blockNumbers <- blockNum
+		}
+		close(blockNumbers)
+	}()
+
+	var writeMu sync.Mutex
+	numWorkers := runtime.GOMAXPROCS(0)
+	workerPool := pool.New().WithErrors().WithMaxGoroutines(numWorkers)
+	batch := database.NewBatch()
```
rodrodros: How long does the migration take. Is it super quick, no need, but if not, what do you think of adding some logging to inform users. I leave it completely at your discretion. 

Also, is this operation light enough to work under the 8gb of ram and 4 core minimum requirements of Juno? I think yes, but asking just to be safe. An intuitive answer is good enough in this case in my opinion – let's not put more time into this PR than strictly required

rodrodros (approved):

migration/migration.go
```diff
@@ -993,3 +994,70 @@ func calculateCasmClassHashesV2(txn db.IndexedBatch, network *utils.Network) err
...
+			return nil, nil // Empty DB, nothing to migrate
+		}
+		return nil, err
+	}
+
+	// Channel to distribute block numbers to workers
+	blockNumbers := make(chan uint64)
+	go func() {
+		for blockNum := range chainHeight + 1 {
+			blockNumbers <- blockNum
```
infrmtcs: I think we should select context here, in case user want to cancel.

migration/migration.go
EgeCaner: Migration was taking ~1 min 15 sec in my machine, let me run it in environment similar to minimum requirements

infrmtcs: I think how long it takes also matters because we're writing everything in a single batch. If user kills the process in the middle, all progress will be lost, and user has to suffer the same cost of migrating them again.

infrmtcs: @EgeCaner Is it sepolia or mainnet?

EgeCaner: Im not sure if we should allow to cancel. In case of cancel and re-run we would do the all operations again, migration wont start from where it was cancelled. To be able to remain from where we left, we need to track additional state, by taking a look at the table we are writing L1MessageHash to L2 txn hash we cannot tell were we left?

infrmtcs (approved):

migration/migration.go
EgeCaner: >Is it sepolia or mainnet?

It was Sepolia

migration/migration.go
```diff
@@ -993,3 +994,70 @@ func calculateCasmClassHashesV2(txn db.IndexedBatch, network *utils.Network) err
 
 	return workerPool.Wait()
 }
+
+// recalculateL1HandlerMsgHashes recalculates L1Handler message hash to txn hash mappings.
+// Needed because calculateL1MsgHashes2 ran with a buggy WriteL1HandlerMsgHashes.
+// Functionally same as calculateL1MsgHashes2, but optimised for concurrent reads.
+type recalculateL1HandlerMsgHashesToTxnHashes struct{}
```
infrmtcs: nit: Is it possible to move this into a separate file or package? I think this file is getting too long now.

EgeCaner: Sure

migration/migration.go
EgeCaner: I will add some logs as well, on Mainnet it will take lot more than on sepolia

PR #3369, opened 2026-01-09
chore: fix lint errors

rodrodros:
Fixing linting errors

### DIFF
--- a/core/trie/bitarray.go
+++ b/core/trie/bitarray.go
@@ -676,7 +676,7 @@ func (b *BitArray) Copy() BitArray {
 func (b *BitArray) EncodedString() string {
 	bt := b.Bytes()
 
-	// 1 byte to accomoaate the length and the other 32 for the data
+	// 1 byte to accommodate the length and the other 32 for the data
 	res := make([]byte, 1+32)
 	res[0] = b.len
 	copy(res[1:33], bt[0:32])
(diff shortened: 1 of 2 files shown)


core/trie/bitarray.go
```diff
@@ -674,10 +674,12 @@ func (b *BitArray) Copy() BitArray {
 
 // Returns the encoded string representation of the bit array.
 func (b *BitArray) EncodedString() string {
-	var res []byte
 	bt := b.Bytes()
-	res = append(res, b.len)
-	res = append(res, bt[:]...)
+
+	// 1 byte to accomoaate the length and the other 32 for the data
```
EgeCaner:
```suggestion
	// 1 byte to accommodate the length and the other 32 for the data
```

core/trie/trie_test.go
```diff
@@ -162,23 +162,21 @@ func TestPutZero(t *testing.T) {
 	require.NoError(t, trie.RunOnTempTriePedersen(251, func(tempTrie *trie.Trie) error {
 		emptyRoot, err := tempTrie.Hash()
 		require.NoError(t, err)
-		var roots []*felt.Felt
-		var keys []*felt.Felt
 
 		// put random 64 keys and record roots
-		for range 64 {
+		const amount = 64
```
MaksymMalicki: turbo small nit: I'd rename `amount` to something like `numKeys`

rodrodros:
> We can initialize the slices with size and assign by index, but I'm not sure if it worth doing since all those are test code and difference should be negligible.

There were some cases where it was not trivial so I decided to leave it as it is (unless trivial) in the test code. I did change all production code that would benefit from this.

PR #3370, opened 2026-01-09
feat: optimize some methods of the trie package

rodrodros:
Some small optimization to the trie methods

### DIFF
--- a/core/trie2/trieutils/accessors_test.go
+++ b/core/trie2/trieutils/accessors_test.go
@@ -48,16 +48,22 @@ func owners() []ownerCase {
 	// they're all real Sepolia addresses
 	return []ownerCase{
 		{
-			name:  "one leading zero",
-			owner: felt.UnsafeFromString[felt.Address]("0x012b66312a967681ca775a20bb3445883b82477888c76790091c0b59593c5f9e"),
+			name: "one leading zero",
+			owner: felt.UnsafeFromString[felt.Address](
+				"0x012b66312a967681ca775a20bb3445883b82477888c76790091c0b59593c5f9e",
+			),
 		},
 		{
-			name:  "two leading zeros",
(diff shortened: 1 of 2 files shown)


core/trie2/trieutils/bitarray.go
```diff
@@ -448,24 +447,33 @@ func (b *BitArray) IsEmpty() bool {
 // Serialises the BitArray into a bytes buffer in the following format:
 // - First few bytes: the necessary bytes included in big endian order
 // - Last byte: length of the bit array (0-255)
-// Example:
+//
+// The returned error is always nil
+// Example: Err
```
MaksymMalicki: I don't understand this comment, what does `Example: Err` mean?

core/trie2/trieutils/accessors.go
```diff
@@ -198,50 +206,50 @@ func nodeKeyByHash(
...
+	dst = dst[len(prefixBytes):]
+
+	copy(dst, ownerBytes)
+	dst = dst[len(ownerBytes):]
+
+	dst[0] = nodeType
+	dst = dst[nodeTypeSize:]
+
+	pathBytes := path.Bytes()
+	copy(dst, pathBytes[0:pathSize])
```
MaksymMalicki: doesn't `path.Bytes()` return big-endian? What happens, when the `pathSize` is not 32?

core/trie2/trieutils/accessors.go
```diff
@@ -118,29 +118,37 @@ func WriteTrieJournal(w db.KeyValueWriter, journal []byte) error {
...
-	var (
-		prefixBytes = prefix.Key()
-		ownerBytes  []byte
-		nodeType    []byte
-		pathBytes   = path.EncodedBytes()
-	)
+	prefixBytes := prefix.Key()
 
+	const ownerByteSize = 32
+	var ownerBytes [32]byte
```
MaksymMalicki: Just marking, that this breaks the key format, by always appending 32 bytes of the owner address, even when the owner is zero. For the `class` tries and `contract` tries, the owner bytes should not be appended at all. The fix is in #3426

PR #3372, opened 2026-01-13
feat: Compaction concurrency configuration

infrmtcs:


db/pebblev2/option.go
```diff
@@ -51,3 +54,58 @@ func WithCompression(compression *block.CompressionProfile) Option {
...
+	}
+}
+
+func parseCompactionConcurrency(s string) (lower, upper int, err error) {
+	parts := strings.Split(s, ",")
+	switch len(parts) {
+	case 1:
+		upper, err = strconv.Atoi(strings.TrimSpace(parts[0]))
+		if err != nil {
+			return 0, 0, fmt.Errorf("invalid compaction concurrency value: %w", err)
```
rodrodros: I think we should specify this error to be the upper value as well 🤔

infrmtcs: Updated

PR #3376, opened 2026-01-15
feat: Add resource semaphore tool to migration

infrmtcs:


migration/semaphore/resource_semaphore.go
```diff
@@ -0,0 +1,40 @@
...
+type ResourceSemaphore[T any] struct {
+	sem *semaphore.Weighted
+	fn  func() T
+}
+
+func New[T any](concurrency int, fn func() T) ResourceSemaphore[T] {
+	sem := make(chan struct{}, concurrency)
+	for range concurrency {
+		sem <- struct{}{}
+	}
```
EgeCaner: I guess we no longer need this

migration/semaphore/resource_semaphore.go
```diff
@@ -0,0 +1,40 @@
+package semaphore
+
+import (
+	"context"
+
+	"golang.org/x/sync/semaphore"
+)
+
+// ResourceSemaphore provides batches to the ingestor and committer, rate limited using a semaphore.
```
EgeCaner: Comments looks specific to original place it has been used first

EgeCaner (commented):
Overall looks good, thank you. I left comments about leftover code from the previous implementation.

migration/semaphore/resource_semaphore.go
infrmtcs: Removed the redundant code

migration/semaphore/resource_semaphore.go
infrmtcs: Updated

rodrodros (approved):

PR #3379, opened 2026-01-15
fix: state reader CompiledCasmHash API

EgeCaner:


blockchain/blockchain.go
```diff
@@ -334,57 +340,111 @@ func (b *Blockchain) Store(
 	)
 }
 
-// storeCasmClassHashesV2ForBlock stores CASM class hashes V2 based on the block version.
-// For versions < 0.14.1, it computes hashes from class definitions.
-// For versions >= 0.14.1, it uses pre-computed hashes from the state update.
-func storeCasmClassHashesV2ForBlock(
+// storeCasmHashMetadata stores CASM hash metadata for declared and migrated classes.
```
rodrodros, line 343: Please give a small description of what `declared` and `migrated` stand for: `declared` mean declared clases post 0.14.1 and `migrated` means declared classes pre 0.14.1 which had their casm hash migrated. 

I think it is important because this context will be lost in future Starknet version

core/state/history.go
```diff
@@ -89,6 +89,16 @@ func (s *stateHistory) Class(classHash *felt.Felt) (*core.DeclaredClassDefinitio
 	return declaredClass, nil
 }
 
+func (s *stateHistory) CompiledClassHash(
```
rodrodros, line 92: Can we call this `CasmClassHash`

core/class.go
```diff
@@ -353,3 +354,191 @@ func (d *DeclaredClassDefinition) UnmarshalBinary(data []byte) error {
 	d.At = binary.BigEndian.Uint64(data[:8])
 	return encoder.Unmarshal(data[8:], &d.Class)
 }
+
+type ClassCasmHashMetadata struct {
+	declaredAt uint64
+
```
rodrodros: nit: Intentional white line?

core/class.go
```diff
@@ -353,3 +354,191 @@ func (d *DeclaredClassDefinition) UnmarshalBinary(data []byte) error {
...
+// CasmHash returns the CASM hash for a class at most recent height.
+func (c *ClassCasmHashMetadata) CasmHash() felt.CasmClassHash {
+	// If declared with V2 directly (no V1), return V2
+	if c.casmHashV1 == nil {
+		return c.casmHashV2
+	}
+	// If migrated, always return V2
+	if c.migratedAt != nil {
+		return c.casmHashV2
+	}
```
rodrodros: Small nit, but since we are using `c.CasmHashV1 != nil` to identify when a class was declared with v2 and `c.MigratedAt != nil` to identify when they were migrated, I would suggest we wrap them in methods. Since they are small they will be inlined.

core/class.go
```diff
@@ -353,3 +354,191 @@ func (d *DeclaredClassDefinition) UnmarshalBinary(data []byte) error {
 	d.At = binary.BigEndian.Uint64(data[:8])
 	return encoder.Unmarshal(data[8:], &d.Class)
 }
+
+type ClassCasmHashMetadata struct {
+	declaredAt uint64
+
+	casmHashV2 felt.CasmClassHash // always there either set or pre_computed
+	migratedAt *uint64
```
rodrodros: nit: To avoid the pointer you can have a `int64` but this is an hyperoptimization and probably will affect the legibility of the code more than any performance gain

core/class.go
```diff
@@ -353,3 +354,191 @@ func (d *DeclaredClassDefinition) UnmarshalBinary(data []byte) error {
...
+	return c.casmHashV2
+}
+
+func (c *ClassCasmHashMetadata) MarshalBinary() ([]byte, error) {
+	// Structure:
+	// declaredAt (8) + casmHashV2 (32) +
+	// migratedAt flag (1) + migratedAt (8 if present) +
+	// casmHashV1 flag (1) + casmHashV1 (32 if present)
+	hasMigratedAt := c.migratedAt != nil
+	hasCasmHashV1 := c.casmHashV1 != nil
```
rodrodros, line 485: 👍

core/class_test.go
```diff
@@ -349,3 +350,131 @@ func TestSierraVersion(t *testing.T) {
...
+				hash, err := record.CasmHashAt(150)
+				require.NoError(t, err)
+				assert.Equal(t, v2Hash, hash)
+			})
+
+			t.Run("after migration", func(t *testing.T) {
+				hash, err := record.CasmHashAt(200)
+				require.NoError(t, err)
+				assert.Equal(t, v2Hash, hash)
+			})
```
rodrodros, line 433: I think we should a test as well unmigrating and checking that the hash revers back to v1

core/class_test.go
```diff
@@ -349,3 +350,131 @@ func TestSierraVersion(t *testing.T) {
...
+
+		t.Run("after declaration", func(t *testing.T) {
+			hash, err := record.CasmHashAt(declaredAt + 1)
+			require.NoError(t, err)
+			require.Equal(t, v1Hash, hash)
+		})
+	})
+
+	t.Run("NewCasmHashMetadataDeclaredV2", func(t *testing.T) {
+		record := core.NewCasmHashMetadataDeclaredV2(declaredAt, &v2Hash)
```
rodrodros, line 383: We should add a test for unmigrating here expecting an error

core/class_test.go
```diff
@@ -349,3 +350,131 @@ func TestSierraVersion(t *testing.T) {
...
+			})
+		})
+	})
+
+	t.Run("MarshalBinary and UnmarshalBinary", func(t *testing.T) {
+		original := core.NewCasmHashMetadataDeclaredV1(declaredAt, &v1Hash, &v2Hash)
+		require.NoError(t, original.Migrate(150))
+
+		data, err := original.MarshalBinary()
+		require.NoError(t, err)
```
rodrodros, line 589: I believe you should assert the length of the data here

core/class_test.go
```diff
@@ -349,3 +350,131 @@ func TestSierraVersion(t *testing.T) {
...
+		hash, err = unmarshaled.CasmHashAt(150)
+		require.NoError(t, err)
+		assert.Equal(t, v2Hash, hash) // At migration
+	})
+
+	t.Run("MarshalBinary and UnmarshalBinary for V2-only", func(t *testing.T) {
+		original := core.NewCasmHashMetadataDeclaredV2(declaredAt, &v2Hash)
+
+		data, err := original.MarshalBinary()
+		require.NoError(t, err)
```
rodrodros, line 616: And here as well.

migration/casmhashmetadata/casmhashmetadata.go
```diff
@@ -0,0 +1,243 @@
...
+		m.startFrom = binary.BigEndian.Uint64(intermediateState[:8])
+	}
+	return nil
+}
+
+func (m *CasmHashMetadataMigration) Migrate(
+	ctx context.Context,
+	database db.KeyValueStore,
+	network *utils.Network,
+	log utils.SimpleLogger, //nolint:staticcheck,nolintlint
```
rodrodros: 👀 why not `utils.StructuredLogger` instead?

migration/casmhashmetadata/casmhashmetadata.go
```diff
@@ -0,0 +1,243 @@
...
+			return nil, nil // empty db nothing to process
+		}
+		return nil, err
+	}
+
+	if m.startFrom > 0 {
+		log.Infow("Class CASM hash metadata migration started",
+			"resuming_from", m.startFrom,
+			"total_blocks", chainHeight+1,
+		)
```
rodrodros: With structured logger is the same, the only difference is
```
log.Info("Class CASM hash metadata migration started",
    zap.Uint("resuming_from", m.startFrom),
    zap.Uint("total_blocks", chainHeight + 1,
)
```

migration/casmhashmetadata/casmhashmetadata.go
```diff
@@ -0,0 +1,243 @@
...
+		}
+	}()
+
+	p := pipeline.New(blockNumbers, maxWorkers, processor)
+	return p.Wait()
+}
+
+func firstBlockWithProtocolVersionGreaterThanOrEqual(
+	database db.KeyValueStore, version *semver.Version, rangeEnd uint64,
+) uint64 {
```
rodrodros: 👌

migration/casmhashmetadata/casmhashmetadata.go
```diff
@@ -0,0 +1,243 @@
...
+
+// startProgressLogger starts a goroutine that logs migration progress every 30 seconds.
+// It returns a function to stop the logger.
+func startProgressLogger(
+	ctx context.Context,
+	log utils.SimpleLogger, //nolint:staticcheck,nolintlint
+	startTime time.Time,
+	blocksProcessed *atomic.Uint64,
+	totalBlocks uint64,
+) func() {
```
rodrodros: Ah here you log the progress, ok!

rodrodros: Not for this PR, but I think a good to do is generalize this for the new migration system, so that every migration can benefit from it

migration/casmhashmetadata/casmhashmetadata.go
```diff
@@ -0,0 +1,243 @@
...
+					percentage = float64(current) / float64(totalBlocks) * 100
+				}
+				remaining := totalBlocks - current
+
+				log.Infow("Class CASM hash metadata migration progress",
+					"blocks_processed", current,
+					"blocks_remaining", remaining,
+					"percentage", fmt.Sprintf("%.2f%%", percentage),
+					"time_elapsed", time.Since(startTime).Seconds(),
+				)
```
rodrodros: Contrary to the 2 logs using a simple logger before, in this case it is important that Structured logger is used to avoid the allocations every 30 secs.

migration/casmhashmetadata/casmhashmetadata.go
EgeCaner: We comply with the Migration interface it enforces deprecated logger

```go
type Migration interface {
	Before(intermediateState []byte) error
	// Migration should return intermediate state whenever it requests new txn or detects cancelled ctx.
	Migrate(context.Context, db.KeyValueStore, *utils.Network, utils.SimpleLogger) ([]byte, error)
}
```

core/state/history.go
EgeCaner: `CompiledClassHash` matches with blockifiers state reader api `get_compiled_class_hash` [ref](https://github.com/starkware-libs/sequencer/blob/5c7ab512f5db4b2661ebeb7682cc32e9055c4a39/crates/blockifier/src/state/state_api.rs#L46). I prefer to keep StateReader api consistent with blockifier.

core/class.go
EgeCaner: addressed now using value field, it is safe because we cannot migrate at 0 class, (cant migrate in the same block declared). Thus we can use it as indicator.

migration/casmhashmetadata/casmhashmetadata.go
EgeCaner: replied above

migration/casmhashmetadata/casmhashmetadata.go
EgeCaner: Replied above

blockchain/blockchain.go
EgeCaner: Done

core/class.go
EgeCaner: Done

core/class_test.go
EgeCaner: Added more tests

core/class_test.go
EgeCaner: Done

core/class_test.go
EgeCaner: Done

migration/casmhashmetadata/casmhashmetadata.go
EgeCaner: Added time based logger

PR #3380, opened 2026-01-19
feat(rpcv10): `response_flags` for the `rpcv10`; `proof_facts`, `proof` for invoke v3 txs

MaksymMalicki:
This PR introduces support for new optional fields for transactions:
- `proof_facts` for invoke v3 transactions
- `proof` and `proof_facts` for broadcasted invoke v3 transactions, which are passed to the gateway write API

The invoke v3 transactions will contain `proof_facts` optional field, when the `INCLUDE_PROOF_FACTS` flag is included in the `response_flags` new parameter in `starknet_getBlockWithTxs`, `starknet_getBlockWithReceipts`, `starknet_getTransactionByHash`, `starknet_getTransactionByBlockIdAndIndex`. If the flag is not present, the transactions are processed in the same way as in previous versions of the RPC.

### DIFF
--- a/rpc/v10/transaction_test.go
+++ b/rpc/v10/transaction_test.go
@@ -1910,7 +1910,11 @@ func TestAdaptBroadcastedTransactionValidation(t *testing.T) {
 			broadcastedTxn,
 			network,
 		)
-		require.NoError(t, err, "AdaptBroadcastedTransaction should succeed for valid INVOKE v3 with proof and proof_facts")
+		require.NoError(
+			t,
+			err,
+			"AdaptBroadcastedTransaction should succeed for valid INVOKE v3 with proof and proof_facts",
+		)
 	})
 }
 


rpc/v10/transaction.go
```diff
@@ -2,17 +2,325 @@ package rpcv10
...
 	"github.com/NethermindEth/juno/starknet"
+	"github.com/NethermindEth/juno/utils"
 )
 
+type Transaction struct {
+	rpcv9.Transaction
+	ProofFacts *[]*felt.Felt `json:"proof_facts,omitempty"`
+}
+
+func AdaptTransaction(t core.Transaction, includeProofFacts bool) *Transaction {
```
rodrodros: Return transaction by value please

rpc/v10/transaction.go
```diff
@@ -2,17 +2,325 @@ package rpcv10
...
+	tx := &Transaction{
+		Transaction: *v9Tx,
+	}
+
+	if includeProofFacts {
+		if invokeTx, ok := t.(*core.InvokeTransaction); ok {
+			if invokeTx.Version != nil && invokeTx.Version.Is(3) && invokeTx.ProofFacts != nil {
+				tx.ProofFacts = &invokeTx.ProofFacts
+			}
+		}
```
rodrodros, line 39: Can an `invokeTx` have a nil field?

rpc/v10/transaction.go
rodrodros: Please don't use single letter, use `coreTx` or `transaction`

rpc/v10/transaction.go
```diff
@@ -2,17 +2,325 @@ package rpcv10
...
+	}
+
+	return txn, declaredClass, paidFeeOnL1, nil
+}
+
+type AddTxResponse struct {
+	TransactionHash *felt.Felt `json:"transaction_hash"`
+	ContractAddress *felt.Felt `json:"contract_address,omitempty"`
+	ClassHash       *felt.Felt `json:"class_hash,omitempty"`
```
rodrodros: Is it possible to use the new types `felt.TransactionHash`, `felt.Address` and `felt.ClassHash`?

rodrodros: Should `TransactionHash` be a pointer given that it is not optional. I believe having the value directly fits better

rpc/v10/transaction.go
```diff
@@ -2,17 +2,325 @@ package rpcv10
...
+	respJSON, err := h.gatewayClient.AddTransaction(ctx, txJSON)
+	if err != nil {
+		return AddTxResponse{}, makeJSONErrorFromGatewayError(err)
+	}
+
+	var gatewayResponse struct {
+		TransactionHash *felt.Felt `json:"transaction_hash"`
+		ContractAddress *felt.Felt `json:"address"`
+		ClassHash       *felt.Felt `json:"class_hash"`
+	}
```
rodrodros, line 272: Why pointer to felts, are any of these fields optional? Could they be the right felt types?

rpc/v10/transaction.go
```diff
@@ -2,17 +2,325 @@ package rpcv10
...
+
+	return AddTxResponse{
+		TransactionHash: gatewayResponse.TransactionHash,
+		ContractAddress: gatewayResponse.ContractAddress,
+		ClassHash:       gatewayResponse.ClassHash,
+	}, nil
+}
+
+func adaptToFeederResourceBounds(rb *rpcv9.ResourceBoundsMap) *map[starknet.Resource]starknet.ResourceBounds { //nolint:gocritic
```
rodrodros: Remove the nolint

rpc/v10/transaction.go
```diff
@@ -2,17 +2,325 @@ package rpcv10
...
+	}
+
+	return &feederResourceBounds
+}
+
+func adaptToFeederDAMode(mode *rpcv9.DataAvailabilityMode) *starknet.DataAvailabilityMode {
+	if mode == nil {
+		return nil
+	}
+	return utils.HeapPtr(starknet.DataAvailabilityMode(*mode))
```
rodrodros: Remove this call to heap pointer here please. It's ok to return the reference given the way `starknet.Transaction` is defined

rpc/v10/transaction.go
```diff
@@ -2,17 +2,325 @@ package rpcv10
...
+}
+
+func adaptRPCTxToFeederTx(rpcTx *rpcv9.Transaction) *starknet.Transaction {
```
rodrodros: Return by value please

rpc/v10/transaction.go
```diff
@@ -2,17 +2,325 @@ package rpcv10
...
+		ResourceBounds:        adaptToFeederResourceBounds(rpcTx.ResourceBounds),
+		Tip:                   rpcTx.Tip,
+		NonceDAMode:           adaptToFeederDAMode(rpcTx.NonceDAMode),
+		FeeDAMode:             adaptToFeederDAMode(rpcTx.FeeDAMode),
+		AccountDeploymentData: rpcTx.AccountDeploymentData,
+		PaymasterData:         rpcTx.PaymasterData,
+	}
+}
+
+func makeJSONErrorFromGatewayError(err error) *jsonrpc.Error { //nolint:gocyclo
```
rodrodros: 👁️

rpc/v10/transaction.go
```diff
@@ -69,6 +377,95 @@ func (h *Handler) TransactionStatus(
...
+/****************************************************
+		Transaction Handlers
+*****************************************************/
+
+// TransactionByHash returns the details of a transaction identified by the given hash.
+//
+// It follows the specification defined here:
+// https://github.com/starkware-libs/starknet-specs/blob/0bf403bfafbfbe0eaa52103a9c7df545bec8f73b/api/starknet_api_openrpc.json#L315
+func (h *Handler) TransactionByHash(hash *felt.Felt, responseFlags *ResponseFlags) (*Transaction, *jsonrpc.Error) {
+	includeProofFacts := responseFlags != nil && responseFlags.IncludeProofFacts
```
rodrodros: I'm not convinced `responseFlag` should be considered different than nil. I think it is better to assume it will always be different than nil 🤔

rpc/v10/transaction.go
MaksymMalicki: done

MaksymMalicki: changed

rpc/v10/transaction.go
MaksymMalicki: done

rpc/v10/transaction.go
MaksymMalicki: done

rpc/v10/transaction.go
MaksymMalicki: removed return by reference

rpc/v10/transaction.go
MaksymMalicki: yea, now that I think about it its a good idea

rpc/v10/transaction.go
MaksymMalicki: done

rpc/v10/transaction.go
MaksymMalicki: changed to proper felt types. Outside of TransactionHash, the rest of the fields are optional

rpc/v10/transaction.go
MaksymMalicki: yes!

rpc/v10/transaction.go
MaksymMalicki: removed the check

rpc/v10/response_flags.go
```diff
@@ -0,0 +1,30 @@
+package rpcv10
+
+import (
+	"encoding/json"
+	"fmt"
+)
+
+type ResponseFlags struct {
+	IncludeProofFacts bool `json:"include_proof_facts"`
```
EgeCaner: I guess json tag can be removed since there is a custom marshaller

rpc/v10/transaction.go
```diff
@@ -2,17 +2,365 @@ package rpcv10
...
+	"github.com/NethermindEth/juno/mempool"
 	"github.com/NethermindEth/juno/rpc/rpccore"
 	rpcv9 "github.com/NethermindEth/juno/rpc/v9"
 	"github.com/NethermindEth/juno/starknet"
+	"github.com/NethermindEth/juno/utils"
 )
 
+type Transaction struct {
+	rpcv9.Transaction
+	ProofFacts *[]*felt.Felt `json:"proof_facts,omitempty"`
```
EgeCaner: Do we need extra indirection here?
```suggestion
	ProofFacts []*felt.Felt `json:"proof_facts,omitempty"`
```

rpc/v10/transaction.go
```diff
@@ -2,17 +2,365 @@ package rpcv10
...
+		}
+	}
+
+	return tx
+}
+
+type BroadcastedTransaction struct {
+	rpcv9.BroadcastedTransaction
+	Proof      []uint64     `json:"proof,omitempty"`
+	ProofFacts []*felt.Felt `json:"proof_facts,omitempty"`
```
EgeCaner: Do we need the extra indirection here?

rpc/v10/block_test.go
```diff
@@ -814,3 +817,201 @@ func nilToOne(f *felt.Felt) *felt.Felt {
 	}
 	return f
 }
+
+func TestBlockWithTxsWithResponseFlags(t *testing.T) {
```
EgeCaner: Can you convert this test as test case for table test above? I guess you can use the same helper functions by slightly extending them with response flags.

rpc/v10/transaction_test.go
```diff
@@ -601,3 +574,292 @@ func TestSubmittedTransactionsCache(t *testing.T) {
...
+			ProofFacts: []*felt.Felt{proofFact},
+		}
+
+		_, _, _, err := rpcv10.AdaptBroadcastedTransaction(broadcastedTxn, network)
+		require.Error(t, err)
+		require.Contains(t, err.Error(), "proof_facts can only be included in invoke v3 transactions")
+	})
+}
+
+func createBaseInvokeTransactionV3() core.InvokeTransaction {
```
EgeCaner, line 1921: Check adapters/testutils/transaction_test_utils.go and adapters/testutils/synctransaction_test_utils.go. There should be helpers to create random invoke V3 txs

rpc/v10/transaction.go
```diff
@@ -2,17 +2,365 @@ package rpcv10
...
+		*starknet.Transaction
+		ContractClass json.RawMessage `json:"contract_class,omitempty"`
+		Proof         []uint64        `json:"proof,omitempty"`
+		ProofFacts    []*felt.Felt    `json:"proof_facts,omitempty"`
+	}{
+		Transaction:   &feederTx,
+		ContractClass: v9Tx.ContractClass,
+	}
+
+	if v9Tx.Transaction.Type == rpcv9.TxnInvoke && isVersion3(v9Tx.Transaction.Version) {
```
EgeCaner: Can we move this to `adaptRpcTxToFeederTx`

rpc/v10/transaction.go
```diff
@@ -2,17 +2,365 @@ package rpcv10
...
+}
+
+func adaptToFeederDAMode(mode *rpcv9.DataAvailabilityMode) starknet.DataAvailabilityMode {
+	if mode == nil {
+		return 0
+	}
+	return starknet.DataAvailabilityMode(*mode)
+}
+
+func adaptRPCTxToFeederTx(rpcTx *rpcv9.Transaction) starknet.Transaction {
```
EgeCaner: Why not  receive rpcv10.Transaction and also handle proof here

EgeCaner (commented):
First round of review

rpc/v10/transaction_test.go
MaksymMalicki: Its this particular transaction, which is also used in many places in rpcv9, and used for tests for example in `TestSubmittedTransactionsCache`. I don't think it brings any overhead, as its just abstracted away from `TestSubmittedTransactionsCache` and used in one more place

rpc/v10/block.go
```diff
@@ -98,7 +104,12 @@ func (h *Handler) BlockWithTxHashes(id *rpcv9.BlockID) (*BlockWithTxHashes, *jso
 //
 // It follows the specification defined here:
 // https://github.com/starkware-libs/starknet-specs/blob/cce1563eff702c87590bad3a48382d2febf1f7d9/api/starknet_api_openrpc.json#L99
-func (h *Handler) BlockWithReceipts(id *rpcv9.BlockID) (*BlockWithReceipts, *jsonrpc.Error) {
+func (h *Handler) BlockWithReceipts(
+	id *rpcv9.BlockID,
+	responseFlags ResponseFlags,
```
EgeCaner:
```suggestion
	responseFlags *ResponseFlags,
```

rpc/v10/block_test.go
```diff
@@ -251,20 +284,22 @@ func assertBlockWithTxs(
...
+	includeProofFacts bool,
 ) {
 	t.Helper()
 	require.Equal(t, len(expectedTransactions), len(actualTransactions))
 	for i, expectedTransaction := range expectedTransactions {
 		require.Equal(t, expectedTransaction.Hash(), actualTransactions[i].Hash)
-		adaptedTransaction := rpcv9.AdaptTransaction(expectedTransaction)
-		require.Equal(t, adaptedTransaction, actualTransactions[i])
+		adaptedTransaction := rpcv10.AdaptTransaction(expectedTransaction, includeProofFacts)
+		require.Equal(t, adaptedTransaction.Transaction, actualTransactions[i].Transaction)
```
EgeCaner, line 301: I believe we should be checking the proof facts here as well

rpc/v10/block_test.go
```diff
@@ -284,21 +319,21 @@ func assertTransactionsWithReceiptsEq(
 	t *testing.T,
 	expectedBlock *core.Block,
 	expectedTxnFinalityStatus rpcv9.TxnFinalityStatus,
-	actual []rpcv9.TransactionWithReceipt,
+	actual []rpcv10.TransactionWithReceipt,
```
EgeCaner, line 322:
```suggestion
	actual []rpcv10.TransactionWithReceipt,
	includeProofFacts bool,
```

or 

```suggestion
	actual []rpcv10.TransactionWithReceipt,
	responseFlags ResponseFlags,
```

then necessary checks

rpc/v10/block_test.go
```diff
@@ -814,3 +856,107 @@ func nilToOne(f *felt.Felt) *felt.Felt {
...
+			for _, tx := range blockWithReceipts.Transactions {
+				require.Nil(
+					t,
+					tx.Transaction.ProofFacts,
+					"proof_facts should not be included when flag is not set",
+				)
+			}
+		})
+	})
+}
```
EgeCaner: Lets also move this to table tests

rpc/v10/block_test.go
```diff
@@ -15,20 +15,22 @@ import (
...
 	rpcv6 "github.com/NethermindEth/juno/rpc/v6"
 	rpcv9 "github.com/NethermindEth/juno/rpc/v9"
+	adaptfeeder "github.com/NethermindEth/juno/starknetdata/feeder"
 	"github.com/NethermindEth/juno/utils"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"go.uber.org/mock/gomock"
 )
 
 type blockTagTestCase struct {
```
EgeCaner:
```suggestion
type blockTestCase struct {
```

rpc/v10/response_flags.go
```diff
@@ -0,0 +1,30 @@
...
+	IncludeProofFacts bool
+}
+
+func (r *ResponseFlags) UnmarshalJSON(data []byte) error {
+	var flags []string
+	if err := json.Unmarshal(data, &flags); err != nil {
+		return err
+	}
+
+	r.IncludeProofFacts = false
```
EgeCaner: If it would be easy can you reflect https://github.com/NethermindEth/juno/pull/3390#discussion_r2768125381 in this PR, else it is fine

rpc/v10/transaction.go
```diff
@@ -2,18 +2,378 @@ package rpcv10
...
+		ContractClass: contractClass,
+	}
+	if tx.Transaction.Type == rpcv9.TxnInvoke && isVersion3(tx.Transaction.Version) {
+		payload.Proof = tx.Proof
+		payload.ProofFacts = tx.ProofFacts
+	}
+	return payload
+}
+
+func rpcv9TxToStarknetTx(rpcTx *rpcv9.Transaction) starknet.Transaction {
```
EgeCaner: If change here (return by val) is not propagating to many places, we can move this to RPC v9, make it public and reuse here.

rpc/v10/transaction.go
```diff
@@ -2,18 +2,378 @@ package rpcv10
...
+		Tip:                   rpcTx.Tip,
+		NonceDAMode:           nonceDAModePtr,
+		FeeDAMode:             feeDAModePtr,
+		AccountDeploymentData: rpcTx.AccountDeploymentData,
+		PaymasterData:         rpcTx.PaymasterData,
+	}
+}
+
+//nolint:gocyclo // complex error mapping logic with many cases
+func makeJSONErrorFromGatewayError(err error) jsonrpc.Error {
```
EgeCaner: May reuse this from rpcv9 by making it public if it is the same implementation

rpc/v10/transaction_test.go
EgeCaner: Previously rpcv10 was not implementing `AddTransaction` and forwarding the request to rpcv9, thus we used rpcv9 here to send transaction. We can remove rpcv9 and directly use rpcv10 instead

EgeCaner: same as above

rpc/v10/transaction_test.go
```diff
@@ -601,3 +574,291 @@ func TestSubmittedTransactionsCache(t *testing.T) {
...
+
+	t.Run("WithoutResponseFlag", func(t *testing.T) {
+		tx, rpcErr := handler.TransactionByHash(txnHash, rpcv10.ResponseFlags{})
+		require.Nil(t, rpcErr)
+		require.NotNil(t, tx)
+		require.Nil(t, tx.ProofFacts)
+	})
+}
+
+func TestTransactionByBlockIDAndIndexWithResponseFlags(t *testing.T) {
```
EgeCaner: We need broader tests for other cases since rpcv10 now implements the endpoint, can you copy relative test from rpcv9 in to rpcv10

rpc/v10/transaction_test.go
```diff
@@ -601,3 +574,291 @@ func TestSubmittedTransactionsCache(t *testing.T) {
...
+			Times(1)
+
+		h := handler.WithGateway(mockGateway)
+		got, rpcErr := h.AddTransaction(t.Context(), broadcastedTxn)
+		require.Nil(t, rpcErr)
+		require.NotZero(t, got.TransactionHash)
+	})
+}
+
+func TestTransactionByHashWithResponseFlags(t *testing.T) {
```
EgeCaner: We need broader tests for other cases since rpcv10 now implements the endpoint, can you copy relative test from rpcv9 in to rpcv10

rpc/v10/transaction_test.go
```diff
@@ -601,3 +574,291 @@ func TestSubmittedTransactionsCache(t *testing.T) {
 		require.Empty(t, status)
 	})
 }
+
+func TestAddTransactionWithProofAndProofFacts(t *testing.T) {
```
EgeCaner: We need broader tests for other cases since rpcv10 now implements the endpoint, can you copy relative test from rpcv9 in to rpcv10

adapters/p2p2core/transaction.go
```diff
@@ -335,6 +337,7 @@ func AdaptInvokeV3TxnCommon(
 		NonceDAMode:           nDAMode,
 		FeeDAMode:             fDAMode,
 		AccountDeploymentData: nil, // todo(kirill) recheck
+		ProofFacts:            nil,
```
thiagodeev, line 344: Question: Is this correct?
I saw that the InvokeV3 from the p2p pkg wasn't updated to contain the new ProofFacts fields, so we can't pass it. I don't know the current status of the p2p feature; should we do something about it or not?

rpc/v10/transaction.go
```diff
@@ -2,18 +2,253 @@ package rpcv10
...
+	Proof      []uint64    `json:"proof,omitempty"`
+	ProofFacts []felt.Felt `json:"proof_facts,omitempty"`
+}
+
+func AdaptBroadcastedTransaction(
+	broadcastedTxn *BroadcastedTransaction,
+	network *utils.Network,
+) (core.Transaction, core.ClassDefinition, *felt.Felt, error) {
+	isInvokeV3 := broadcastedTxn.Transaction.Type == rpcv9.TxnInvoke &&
+		broadcastedTxn.Transaction.Version.Cmp(felt.NewFromUint64[felt.Felt](3)) == 0
```
thiagodeev: By comparing the tx version solely against the number 3, we are excluding the query bit version, which is `0x100000000000000000000000000000003`. Since this function is used for both the SimulateTxn and EstimateFee methods at the end, it'll be impossible to call these methods with an invokev3 tx that contains the new `proof` or `proof_facts` fields, and has the query bit tx version. Tested locally

rpc/v10/transaction.go
```diff
@@ -2,18 +2,253 @@ package rpcv10
...
+			return nil, nil, nil, fmt.Errorf("proof_facts can only be included in invoke v3 transactions")
+		}
+		if len(broadcastedTxn.Proof) > 0 {
+			return nil, nil, nil, fmt.Errorf("proof can only be included in invoke v3 transactions")
+		}
+	}
+
+	txn, declaredClass, paidFeeOnL1, err := rpcv9.AdaptBroadcastedTransaction(
+		&broadcastedTxn.BroadcastedTransaction,
+		network,
```
thiagodeev: Even though the new `BroadcastedTransaction` type includes the new `proof_facts` field, we're not passing it to this function.
This rpcv9 function call is also responsible for calculating the tx hash through the `core.TransactionHash` function ([here](https://github.com/NethermindEth/juno/blob/7a236950c9f68e863ffe3aaab088b36f944c40a4/rpc/v9/transaction.go#L394)). Although the `core.TransactionHash` function was refactored to include the new `proof_facts` field in the invoke hash calculation ([ref](https://github.com/NethermindEth/juno/blob/7a236950c9f68e863ffe3aaab088b36f944c40a4/core/transaction.go#L491)), since we're not passing it, it'll be nil/len(0), thus it'll be ignored, thus generating a wrong tx hash for invokeV3 txs that contain the new `proof_facts` field.

thiagodeev (changes requested):
Hey. Overall, good job!

I've found two bugs while reviewing.
This is only a partial review. I still need to finish reviewing the test files, but I thought it would be good to submit this review right now, so that you can fix them while I finish.

These bugs show us that we need to improve the new tests, since they passed without catching them

rpc/v10/transaction_test.go
```diff
@@ -37,6 +44,770 @@ func loadBlockFromFeederTestdata(
...
+
+			res, rpcErr := handler.TransactionByHash(hash, test.responseFlags)
+			require.Nil(t, rpcErr)
+
+			resJSON, err := json.Marshal(res)
+			require.NoError(t, err)
+			resMap := make(map[string]any)
+			require.NoError(t, json.Unmarshal(resJSON, &resMap))
+
+			assert.Equal(t, expectedMap, resMap, string(resJSON))
```
thiagodeev: No changes required, just to let you know, for future uses, we can do it this way. It's cleaner and effective :)

```suggestion
			res, rpcErr := handler.TransactionByHash(hash, test.responseFlags)
			require.Nil(t, rpcErr)

			resJSON, err := json.Marshal(res)
			require.NoError(t, err)

			assert.JSONEq(t, test.expected, string(resJSON))
```

MaksymMalicki: changed this anyways, thanks!

adapters/p2p2core/transaction.go
MaksymMalicki: let me leave a comment here

rpc/v10/transaction.go
```diff
@@ -2,18 +2,324 @@ package rpcv10
...
+}
+
+func AdaptBroadcastedTransaction(
+	ctx context.Context,
+	compiler compiler.Compiler,
+	broadcastedTxn *BroadcastedTransaction,
+	network *utils.Network,
+) (core.Transaction, core.ClassDefinition, *felt.Felt, error) {
+	isInvokeV3 := broadcastedTxn.Transaction.Type == rpcv9.TxnInvoke &&
+		isVersion3(broadcastedTxn.Transaction.Version)
```
thiagodeev: This validation is still wrong WRT the query bit version. The problem lies in the `isVersion3` function, which only compares the version agains an variable representing the `0x3` version. Since the v3 query-bit is `0x100000000000000000000000000000003` instead of `0x3`, the `isInvokeV3` variable will be false, and inside the `validateNonV3Transaction`, this will trigger the "proof can only be included in invoke v3 transactions" error.
So, we are still unable to estimate/simulate an invokeV3 tx containing the query-bit version + the new fields ('proof' AND/OR 'proof_facts').
I noticed that we still don't have any test case covering this. Please add it to the `TestAdaptBroadcastedTransactionValidation`.

I remember in the huddle we had in Slack, you asked me something related to this, I think it was about the correctness of the `isVersion3` function. If that's the case, appologies; I thought you're talking about the `core.TransactionVersion.Is()` method. I believe this method would handle both normal and query-bit version when comparing.

HOWEVERRRRR,I realize we don't need this validation (WRT tx version). While diving into the code, I noticed we're already validating the transaction version with the `validate: ...` json tags ([ref](https://github.com/NethermindEth/juno/blob/95a1967c94d76edb17475da137791c0efa8f1f95/rpc/v9/transaction.go#L277), [ref](https://github.com/NethermindEth/juno/blob/95a1967c94d76edb17475da137791c0efa8f1f95/validator/validator.go#L32), notice the `validate:"required,version_0x3` tag), so we can assume that every tx that came here will be an invokeV3 tx (but we need my pr #3419 to be merged first, since there's a current bug in the estimateFee, SimulateTxns methods)

rpc/v10/transaction.go
```diff
@@ -2,18 +2,324 @@ package rpcv10
...
+
+	if err := validateNonV3Transaction(broadcastedTxn, isInvokeV3); err != nil {
+		return nil, nil, nil, err
+	}
```
thiagodeev: Since we have the `validate:...` json tags, IMO, we can remove this custom logic and use them instead. I've made a code suggestion in this review that implements a validation using them. Tested locally, and it works. Take it if you want (but we need my pr #3419 to be merged first)

rpc/v10/transaction.go
```diff
@@ -2,18 +2,324 @@ package rpcv10
...
+
+	return tx
+}
+
+type BroadcastedTransaction struct {
+	rpcv9.BroadcastedTransaction
+	Proof      []uint64    `json:"proof,omitempty"`
+	ProofFacts []felt.Felt `json:"proof_facts,omitempty"`
+}
```
thiagodeev, line 49: Here's the validation using the json tags.
https://pkg.go.dev/github.com/go-playground/validator/v10@v10.27.0#hdr-Excluded_Unless

```suggestion
type BroadcastedTransaction struct {
	rpcv9.BroadcastedTransaction
	Proof      []uint64    `json:"proof,omitempty" validate:"excluded_unless=Type INVOKE"`
	ProofFacts []felt.Felt `json:"proof_facts,omitempty" validate:"excluded_unless=Type INVOKE"`
}
```

rpc/v10/transaction.go
```diff
@@ -2,18 +2,324 @@ package rpcv10
...
+
+func AdaptBroadcastedTransaction(
+	ctx context.Context,
+	compiler compiler.Compiler,
+	broadcastedTxn *BroadcastedTransaction,
+	network *utils.Network,
+) (core.Transaction, core.ClassDefinition, *felt.Felt, error) {
```
thiagodeev, line 56: Question: at the end, we are returning a `core.Transaction` that doesn't contain the new `proof` field for the invoke broadcast tx, so this value will not be sent to the sequencer (only the new invoke v3 `proof_facts` field will be included). Is this correct?

adapters/p2p2core/transaction.go
thiagodeev: What about this?

MaksymMalicki: actually removed it for now

rpc/v10/transaction.go
MaksymMalicki: actually both for gateway and mempool, I added the proof field to the payload and not the transaction itself. I don't want to add a proof field to the `InvokeV3Transaction` which the Broadcasted Transaction gets adapted into, because the `proof` field is used only in the write API

rpc/v10/transaction_test.go
```diff
@@ -580,24 +1781,171 @@ func TestSubmittedTransactionsCache(t *testing.T) {
...
+
+		jsonBytes, err := json.Marshal(broadcastedTxn)
+		require.NoError(t, err)
+		var unmarshaledTxn rpcv10.BroadcastedTransaction
+		err = json.Unmarshal(jsonBytes, &unmarshaledTxn)
+		require.NoError(t, err)
+
+		validate := validator.Validator()
+		err = validate.Struct(unmarshaledTxn)
+		require.Error(t, err, "validation should fail because proof_facts are excluded unless Type is INVOKE")
```
thiagodeev: Why not use the original struct directly? I didn't notice any difference in the test result.
Also applicable to the other test cases, which do the same thing

```suggestion
		validate := validator.Validator()
		err := validate.Struct(broadcastedTxn)
		require.Error(t, err, "validation should fail because proof_facts are excluded unless Type is INVOKE")
```

rpc/v10/transaction.go
```diff
@@ -2,18 +2,307 @@ package rpcv10
...
+	tx := Transaction{
+		Transaction: *v9Tx,
+	}
+
+	if includeProofFacts {
+		if invokeTx, ok := coreTx.(*core.InvokeTransaction); ok {
+			if invokeTx.Version.Is(3) && invokeTx.ProofFacts != nil {
+				tx.ProofFacts = invokeTx.ProofFacts
+			}
+		}
```
thiagodeev, line 39: While reviewing PR #3390, I realize one thing: since `ProofFacts` is `omitempty`, does that mean that, even though `includeProofFacts` is true, the final json will not include the `proof_facts` if there are no proof facts?
IMO, this is not correct. We should do as we do for the other required fields if they're empty: return an empty value (e.g: we always return [] for the `account_deployment_data` field). We should return `proof_facts: []` instead of omitting it when the user specifies the `includeProofFacts` flag.

WDYT?

thiagodeev: Update: it was shown to me that the decision not to return the field when there's no proof comes from Starkware

PR #3384, opened 2026-01-23
feat(rpcv10): initial reads for `starknet_traceBlockTransactions` and `starknet_simulateTransactions`

MaksymMalicki:
This PR introduces support for the new `initial_reads` optional field for the `starknet_traceBlockTransactions` and `starknet_simulateTransactions`.
Changes:
- adds a new `RETURN_INITIAL_READS` flag to the simulation flags. If the `RETURN_INITIAL_READS` is included in the `simulation_flags` of `starknet_simulateTransactions`, the `initial_reads` field will be added to the response. 
- adds a new `trace_flags` to `starknet_traceBlockTransactions`. If the `RETURN_INITIAL_READS` flag is included, the `initial_reads` field will be added to the response.
- For both `starknet_traceBlockTransactions` and `starknet_simulateTransactions`, if the `RETURN_INITIAL_READS` was not included in the flags set, the response looks the same as in previous versions of the RPC (list of transaction simulations / transaction traces)
- If the `RETURN_INITIAL_READS` was passed to `starknet_traceBlockTransactions`, the response format is different. It contains two fields: `traces` (list of transaction traces for this block) and `initial_reads` (aggregated initial reads for all the traced transactions)
- If the `RETURN_INITIAL_READS` was passed to `starknet_simulateTransactions`, the response format is different. It contains two fields: `simulated_transactions` (list of transaction simulations) and `initial_reads` (aggregated initial reads for all the simulated transactions)

NOTE:
If the `initial_reads` are requested, but they don't exist (i. e. older blocks, where the traces are fetched from the gateway), they are not included in the response.

This PR required bumping the blockifier to [newer](https://github.com/starkware-libs/sequencer/commit/be54aeea510ddfe1190c742e902acbd8103ff941) version.

### DIFF
--- a/rpc/v10/simulation.go
+++ b/rpc/v10/simulation.go
@@ -60,6 +60,33 @@ func (t *TraceFlag) UnmarshalJSON(bytes []byte) (err error) {
 	return err
 }
 
+type EstimateFlag int
+
+const (
+	EstimateSkipValidateFlag EstimateFlag = iota + 1
+)
+
+func (e *EstimateFlag) UnmarshalJSON(bytes []byte) (err error) {
+	switch flag := string(bytes); flag {
+	case `"SKIP_VALIDATE"`:
+		*e = EstimateSkipValidateFlag
+	default:
+		err = fmt.Errorf("unknown estimate flag %q", flag)
+	}
+
+	return err
+}
+
+// ToSimulationFlag converts an EstimateFlag to the corresponding SimulationFlag.
+func (e EstimateFlag) ToSimulationFlag() (SimulationFlag, error) {
+	switch e {
+	case EstimateSkipValidateFlag:
+		return SkipValidateFlag, nil
+	default:
+		return 0, fmt.Errorf("unknown estimate flag %v", e)
+	}
+}
+
 type SimulatedTransaction struct {
 	TransactionTrace *TransactionTrace `json:"transaction_trace,omitempty"`
 	FeeEstimation    rpcv9.FeeEstimate `json:"fee_estimation,omitzero"`
--- a/rpc/v10/trace.go
+++ b/rpc/v10/trace.go
@@ -481,9 +481,13 @@ func (h *Handler) traceBlockTransactions(
 			},
 		)
 
+		var initialReads *InitialReads
+		if returnInitialReads {
+			initialReads = &InitialReads{}
+		}
 		return TraceBlockTransactionsResponse{
 			Traces:       traces,
-			InitialReads: &InitialReads{},
+			InitialReads: initialReads,
 		}, defaultExecutionHeader(), nil
 	}
 
--- a/rpc/v10/estimate_fee.go
+++ b/rpc/v10/estimate_fee.go
@@ -19,9 +19,18 @@ import (
 */
 func (h *Handler) EstimateFee(
 	broadcastedTxns BroadcastedTransactionInputs,
-	simulationFlags []SimulationFlag,
+	estimateFlags []EstimateFlag,
 	id *rpcv9.BlockID,
 ) ([]rpcv9.FeeEstimate, http.Header, *jsonrpc.Error) {
+	simulationFlags := make([]SimulationFlag, 0, len(estimateFlags)+1)
+	for _, flag := range estimateFlags {
+		simulationFlag, err := flag.ToSimulationFlag()
+		if err != nil {
+			return nil, nil, jsonrpc.Err(jsonrpc.InvalidParams, err.Error())
+		}
+		simulationFlags = append(simulationFlags, simulationFlag)
+	}
+
 	txnResults, httpHeader, err := h.simulateTransactions(
 		id,
 		broadcastedTxns.Data,
--- a/rpc/v10/estimate_fee_test.go
+++ b/rpc/v10/estimate_fee_test.go
@@ -57,7 +57,7 @@ func TestEstimateFee(t *testing.T) {
 
 		_, httpHeader, err := handler.EstimateFee(
 			rpcv10.BroadcastedTransactionInputs{},
-			[]rpcv10.SimulationFlag{},
+			[]rpcv10.EstimateFlag{},
 			&blockID,
 		)
 		require.Nil(t, err)
@@ -87,8 +87,8 @@ func TestEstimateFee(t *testing.T) {
 
 		_, httpHeader, err := handler.EstimateFee(
 			rpcv10.BroadcastedTransactionInputs{},
-			[]rpcv10.SimulationFlag{
-				rpcv10.SkipValidateFlag,
+			[]rpcv10.EstimateFlag{
+				rpcv10.EstimateSkipValidateFlag,
 			},
 			&blockID,
 		)
@@ -113,7 +113,7 @@ func TestEstimateFee(t *testing.T) {
 
 		_, httpHeader, err := handler.EstimateFee(
 			rpcv10.BroadcastedTransactionInputs{},
-			[]rpcv10.SimulationFlag{rpcv10.SkipValidateFlag},
+			[]rpcv10.EstimateFlag{rpcv10.EstimateSkipValidateFlag},
 			&blockID,
 		)
 		require.Equal(t, rpccore.ErrTransactionExecutionError.CloneWithData(
@@ -143,7 +143,7 @@ func TestEstimateFee(t *testing.T) {
 		}
 		_, _, err := handler.EstimateFee(
 			rpcv10.BroadcastedTransactionInputs{Data: []rpcv9.BroadcastedTransaction{invalidTx}},
-			[]rpcv10.SimulationFlag{},
+			[]rpcv10.EstimateFlag{},
 			&blockID,
 		)
 		expectedErr := &jsonrpc.Error{


vm/vm.go
```diff
@@ -103,6 +105,7 @@ type callContext struct {
 	gasConsumed     []core.GasConsumed
 	executionSteps  uint64
 	receipts        []json.RawMessage
+	initialReads    *InitialReads
```
infrmtcs: Should we use `json.RawMessage` here and defer the unmarshal to the outer layer?

vm/rust/src/entrypoint/execute/execute.rs
```diff
@@ -185,7 +188,40 @@ pub fn cairo_vm_execute(
...
+                .map_err(|err| {
+                    JunoError::tx_non_execution_error(
+                        format!("failed to get initial reads: {err:?}"),
+                        txn_index,
+                    )
+                })?;
+            
+            match &mut aggregated_initial_reads {
+                Some(aggregated) => {
+                    aggregated.storage.extend(state_maps.storage);
```
EgeCaner: If I'm not mistaken extend overwrites existing values with newer ones. We should be preserving existing one

EgeCaner: If I'm not mistaken extend overwrites existing values with newer ones. We should be preserving existing one. Alternatively calling get_initial_reads on parent state  after loop should work I guess

rpc/v6/simulation.go
```diff
@@ -20,6 +20,7 @@ type SimulationFlag int
 const (
 	SkipValidateFlag SimulationFlag = iota + 1
 	SkipFeeChargeFlag
+	ReturnInitialReadsFlag
```
EgeCaner: Lets keep rpcv6.SimulationFlag unchanged and declare SimulationFlag in rpcv10. This change belongs to rpcv10 and shouldnt affect previous versions

rpc/v10/trace.go
```diff
@@ -428,39 +439,41 @@ func (h *Handler) traceInPreConfirmedBlock(
...
 func (h *Handler) traceBlockTransactions(
-	ctx context.Context, block *core.Block,
-) ([]TracedBlockTransaction, http.Header, *jsonrpc.Error) {
+	ctx context.Context, block *core.Block, returnInitialReads bool,
+) (TraceBlockTransactionsResponse, http.Header, *jsonrpc.Error) {
 	// Check if it was already traced
 	traces, hit := h.blockTraceCache.Get(rpccore.TraceCacheKey{BlockHash: *block.Hash})
-	if hit {
-		return traces, defaultExecutionHeader(), nil
+	if hit && !returnInitialReads {
```
EgeCaner: We can extend the TraceCacheKey with initial reads flag and can cache traces with/without initial reads

rpc/v10/simulation.go
```diff
@@ -24,11 +25,100 @@ type SimulatedTransaction struct {
...
+}
+
+func (r SimulateTransactionsResponse) MarshalJSON() ([]byte, error) {
+	if r.InitialReads == nil {
+		return json.Marshal(r.SimulatedTransactions)
+	}
+	type simulateTransactionsResponse struct {
+		SimulatedTransactions []SimulatedTransaction `json:"simulated_transactions"`
+		InitialReads          *InitialReads          `json:"initial_reads"`
+	}
```
EgeCaner: Nit: We can move the the json tags to original type and create an alias here

```go

type simulateTransactionsResponse SimulateTransactionsResponse
```

rpc/v10/simulation.go
```diff
@@ -24,11 +25,100 @@ type SimulatedTransaction struct {
...
+}
+
+func (r TraceBlockTransactionsResponse) MarshalJSON() ([]byte, error) {
+	if r.InitialReads == nil {
+		return json.Marshal(r.Traces)
+	}
+	type traceBlockTransactionsResponse struct {
+		Traces       []TracedBlockTransaction `json:"traces"`
+		InitialReads *InitialReads            `json:"initial_reads"`
+	}
```
EgeCaner: Nit: Same as above

rpc/v10/simulation_test.go
```diff
@@ -317,3 +318,180 @@ func TestSimulateTransactionsShouldErrorWithoutSenderAddressOrResourceBounds(t *
 		})
 	}
 }
+
+// initialReadsTestCases returns the shared test cases for testing RETURN_INITIAL_READS flag.
+// These test cases are used by both TestSimulateTransactionsWithReturnInitialReads
+// and TestTraceBlockTransactionsWithReturnInitialReads since they test the same
+// initial reads behaviour.
+func initialReadsTestCases() []struct {
```
EgeCaner: lets extract this into type

rpc/v10/simulation_test.go
```diff
@@ -317,3 +318,180 @@ func TestSimulateTransactionsShouldErrorWithoutSenderAddressOrResourceBounds(t *
...
+			simulationFlags      []rpcv6.SimulationFlag
+			initialReads         *vm.InitialReads
+			expectedInitialReads *rpcv10.InitialReads
+		}{
+			name:                 tc.name,
+			simulationFlags:      tc.flags,
+			initialReads:         tc.initialReads,
+			expectedInitialReads: tc.expectedInitialReads,
+		}
+	}
```
EgeCaner: Do we need this, isn't testCases already the same list we are constructing?

rpc/v10/trace.go
```diff
@@ -130,18 +130,19 @@ func (h *Handler) Call(
 //
 //nolint:lll // URL exceeds line limit but should remain intact for reference
 func (h *Handler) TraceBlockTransactions(
-	ctx context.Context, id *rpcv9.BlockID,
-) ([]TracedBlockTransaction, http.Header, *jsonrpc.Error) {
+	ctx context.Context, id *rpcv9.BlockID, traceFlags []SimulationFlag,
```
EgeCaner: I guess we need a separate type for `TraceFlag`, otherwise this will accept other flags `SKIP_FEE_CHARGE` etc...

EgeCaner (approved):
Lets have type for`TraceFlag`, rest looks good to me. Thank you!

vm/rust/src/entrypoint/utils.rs
```diff
@@ -96,6 +96,12 @@ pub fn build_block_context(
         constants.invoke_tx_max_n_steps = max_steps as u32;
     }
 
+    let starknet_version = unsafe { CStr::from_ptr(block_info.version) }
+        .to_str()
+        .ok()
+        .and_then(|version_str| StarknetVersion::try_from(version_str).ok())
+        .unwrap_or_else(|| StarknetVersion::default());
```
thiagodeev: nitpick:
```suggestion
        .unwrap_or_default();
```

vm/rust/src/entrypoint/execute/execute.rs
```diff
@@ -185,7 +187,23 @@ pub fn cairo_vm_execute(
...
+            .map_err(|err| {
+                JunoError::tx_non_execution_error(
+                    format!("failed to get initial reads: {err:?}"),
+                    0,
+                )
+            })?;
+        let initial_reads_ffi: InitialReads = initial_reads.into();
+        append_initial_reads(reader_handle, &initial_reads_ffi, &mut writer_buffer)
+            .map_err(|err| JunoError::tx_non_execution_error(err, 0))?;
+    }
```
thiagodeev, line 203: Using 'tx_non_execution_error' will make the final error be a `vm.TransactionExecutionError` with index 0. Isn't it better to return it as `JunoError::block_error` instead in both cases?

rpc/v10/simulation.go
```diff
@@ -19,16 +20,122 @@ import (
...
+
+func NewTraceBlockTransactionsResponse(
+	traces []TracedBlockTransaction,
+	initialReads *InitialReads,
+) TraceBlockTransactionsResponse {
+	return TraceBlockTransactionsResponse{
+		Traces:       traces,
+		InitialReads: initialReads,
+	}
+}
```
thiagodeev: The new `NewTraceBlockTransactionsResponse` and `NewSimulateTransactionsResponse` functions bring no value IMO; they only return a struct literal with the provided values. Why not just use the struct literal instead in the code? Is there any specific reason to have them?

rpc/v10/simulation.go
```diff
@@ -19,16 +20,122 @@ import (
...
+
+type ClassHashEntry struct {
+	ContractAddress *felt.Address   `json:"contract_address"`
+	ClassHash       *felt.ClassHash `json:"class_hash"`
+}
+
+type DeclaredContractEntry struct {
+	ClassHash  *felt.ClassHash `json:"class_hash"`
+	IsDeclared bool            `json:"is_declared"`
+}
```
thiagodeev, line 154: These InitialReads field types from the rpc pkg are different from the ones in the vm pkg. Here, they are all accepting a pointer to Felt, and in the vm pkg, they're all accepting a concrete felt type ([ref](https://github.com/NethermindEth/juno/blob/ea3a51c975dc915462dd25aa94209bcbcc6d4563/vm/trace.go#L58-L77)). Is there any specific reason for that? Wouldn't it be better to use only one way?

rpc/v10/trace.go
thiagodeev: Indeed. The same is true for the `EstimateFee` method

rpc/v10/trace.go
```diff
@@ -428,39 +438,52 @@ func (h *Handler) traceInPreConfirmedBlock(
...
+			return TraceBlockTransactionsResponse{}, defaultExecutionHeader(), err
 		}
-		h.blockTraceCache.Add(rpccore.TraceCacheKey{BlockHash: *block.Hash}, traces)
-		return traces, defaultExecutionHeader(), nil
+		// Feeder gateway doesn't provide initial reads, so cache without them
+		h.blockTraceCache.Add(rpccore.TraceCacheKey{BlockHash: *block.Hash}, TraceBlockTransactionsResponse{
+			Traces:       traces,
+			InitialReads: nil,
+		})
+		return NewTraceBlockTransactionsResponse(traces, nil), defaultExecutionHeader(), nil
```
thiagodeev: The spec says ([ref](https://github.com/starkware-libs/starknet-specs/blob/b57ccf8fcbc32f89a42f088ff9c8b93115bfd40a/api/starknet_trace_api_openrpc.json#L216)):

> **Required** when RETURN_INITIAL_READS is present in trace_flags. Returns an **empty object** instead of INITIAL_READS when the execution trace for the referenced block is inconsistent with the canonical block trace.

The way the code is implemented, if we need to fetch from the feeder, even if the `returnInitialReads` flag is true, it will return `nil` (since the feeder doesn't provide initial reads), resulting in the `"initial_reads": null` json, which is incorrect based on the spec.

I'm not sure though if we should return an empty struct like `"initial_reads": {}` or if we can return the empty value of the initial reads (`"initial_reads": {"storage": [], "nonces": []...}`)

thiagodeev (changes requested):
Hey. Nice work!

I made a few comments. Some of them are okay to ignore, but there are some points that I believe must be resolved before merging

rpc/v10/simulation.go
MaksymMalicki: thanks!

rpc/v10/trace.go
```diff
@@ -466,11 +473,18 @@ func (h *Handler) traceBlockTransactions(
...
+			},
+		)
+
+		return TraceBlockTransactionsResponse{
 			Traces:       traces,
-			InitialReads: nil,
-		})
-		return NewTraceBlockTransactionsResponse(traces, nil), defaultExecutionHeader(), nil
+			InitialReads: &InitialReads{},
+		}, defaultExecutionHeader(), nil
```
thiagodeev, line 491: Since we're not checking the `returnInitialReads` flag here, this will make the 'initialReads' field appear in the final JSON even without setting the flag (next calls for the same trace will work normally due to the logic at the top of the function WRT the cache, but for every first call it will fail)

PR #3385, opened 2026-01-23
feat: new migration package

EgeCaner:
# New Migration System

Introduces a registry-based migration system to replace the deprecated migration pipeline.

## Features

- **Registry-based registration**: Centralized migration registration with support for mandatory and optional migrations
- **Resumable migrations**: Two-phase interface (`Before`/`Migrate`) with intermediate state persistence
- **Schema versioning**: Bitset-based version tracking (uint64, up to 64 migrations)
- **Safety validations**: Prevents version downgrades and opt-out of previously enabled migrations
- **Progress tracking**: Structured logging with progress indicators and optional HTTP status server

## Architecture

- `migration.Registry`: Registers migrations and builds target version
- `migration.Migration`: Interface with `Before()` for state restoration and `Migrate()` for execution
- `migration.MigrationRunner`: Executes migrations with validation and state management
- `migration.SchemaVersion`: Bitset-based version tracking

## Integration

Runs alongside the deprecated system: deprecated migrations execute first, then new migrations.

## Example

```go
registry := migration.NewRegistry().
    With(&casmhashmetadata.Migrator{}).
    WithOptional(&l1handlermapping.Migrator{}, cfg.P2P)

runner, err := migration.NewRunner(registry.Entries(), registry.TargetVersion(), db, &network, log)
return runner.Run(ctx)
```

## Notice

Following packages moved from old migration package to new one with minimal changes (only the logger)

- pipeline/
- semaphore/
- progresslogger/
- l1handlermapping/
- casmhashmetadata/

### DIFF
--- a/cmd/juno/dbcmd.go
+++ b/cmd/juno/dbcmd.go
@@ -12,6 +12,7 @@ import (
 	"github.com/NethermindEth/juno/db"
 	"github.com/NethermindEth/juno/db/pebblev2"
 	"github.com/NethermindEth/juno/deprecatedmigration" //nolint:staticcheck,nolintlint // deprecated package will be removed later
+	"github.com/NethermindEth/juno/migration"
 	"github.com/NethermindEth/juno/utils"
 	"github.com/olekukonko/tablewriter"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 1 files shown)


cmd/juno/dbcmd.go
```diff
@@ -100,7 +100,7 @@ func dbInfo(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("failed to get the state update: %v", err)
 	}
 
-	schemaMeta, err := migration.SchemaMetadata(utils.NewNopZapLogger(), database)
+	schemaMeta, err := deprecatedmigration.SchemaMetadata(utils.NewNopZapLogger(), database)
```
EgeCaner: What should we display here as DB version, new one? Or should new one continue from where previous left

migration/registry.go
```diff
@@ -0,0 +1,66 @@
...
+// Returns the registry for method chaining.
+func (r *Registry) WithOptional(m Migration, enabled bool) *Registry {
+	if r.nextIndex >= maxMigrations {
+		panic("exceeded maximum number of 64 migrations")
+	}
+	r.entries[r.nextIndex] = m
+
+	if enabled {
+		r.targetVersion.Set(r.nextIndex)
+	}
```
rodrodros, line 47: What happen if I:
1. Run a migration with optional enabled, the migration finishes, I shut down the node, and restart it without the optional flag.
2. Harder case, what if I stopped the optional lmigration by half, and then restart it without the flag?

EgeCaner: See runner.validateNoOptOut [ref](https://github.com/NethermindEth/juno/pull/3385/files#diff-06d6b6f2392d0ffcb3a644d0e0ba1d26bf771e7e475844f1a5f87d3209d31c66R103). Before running any migration we save the current target which contain optional preferences. We compare current target build by registry accounting for optionals with LastTarget from db to check if there is any optout attempt. If any previously enabled migration is disabled it will return error. This requires users to run with flags that they been enabled before in consecutive runs till it becomes mandatory. This happens even though we havent start running optional migration at all due to saving target schema to LastTarget at the very beginning. LastTarget can be updated before running per migration, this will allow user to start with enabled and if they cancel before reaching to that migration step, they will be able to disable.

PR #3388, opened 2026-01-26
fix: add context to pipeline to cancel generator on error

EgeCaner:


migration/casmhashmetadata/migrator.go
```diff
@@ -174,8 +177,9 @@ func migrateRange(
 ) (uint64, error) {
 	blockNumbers := make(chan uint64)
 
-	ingestorPipeline := pipeline.New(blockNumbers, maxWorkers, processor)
+	ingestorPipeline := pipeline.New(ctx, blockNumbers, maxWorkers, processor)
 	committerPipeline := pipeline.New(
+		ingestorPipeline.Context(),
```
infrmtcs, line 182: I think this pattern is easy to misuse and we have to ensure the cycle of context. Let me think if there is a better way to structure this.

PR #3398, opened 2026-01-29
chore(ci): Build and publish official docker image when release is published

brbrr:
Currently, we are publishing "official" tagged docker image when draft release is created. That's suboptimal, since when the image is available for users it's not yet "released". 

This PR updates the workflow as follows:
- on tag, we build non-official Docker image, publish it, and label it as `latest` (for our test nodes to pick up), also we create draft release and build binaries (as before)
- on release publish, we build official Docker image, publish it and set it as `latest`.

this way, any intermidiate releases (`.rc.X` etc) will not be available in official docker repo, and reduce user's confusion


rodrodros (approved):

.github/workflows/prepare-draft-release.yml
```diff
@@ -20,17 +20,30 @@ jobs:
...
+        with:
+          username: ${{ secrets.DOCKER_HUB_USERNAME }}
+          password: ${{ secrets.DOCKER_HUB_PASSWORD }}
+
+      - name: Tag non-official image as latest
+        run: docker buildx imagetools create --tag nethermindeth/juno:latest nethermindeth/juno:${{ needs.docker-image.outputs.tag }}
+
+      - name: Clean up Docker config
+        if: always()
+        run: rm -f ${HOME}/.docker/config.json
```
derrix060, line 46: why is this necessary?

PR #3404, opened 2026-01-30
migration: better error message on optout attempt

EgeCaner:
Now returned error is in the following format

>Error while running migrations	{"err": "create migration runner: cannot opt out of previously enabled migrations: [--transaction-combined-layout] (set these flags to enable)"}

### DIFF
--- a/cmd/juno/juno.go
+++ b/cmd/juno/juno.go
@@ -101,7 +101,7 @@ const (
 	dbCompactionConcurrencyF            = "db-compaction-concurrency"
 	dbMemtableSizeF                     = "db-memtable-size"
 	dbCompressionF                      = "db-compression"
-	transactionCombinedLayoutF          = "transaction-combined-layout"
+	transactionCombinedLayoutF          = node.FlagTransactionCombinedLayout
 
 	defaultConfig                             = ""
 	defaultHost                               = "localhost"
--- a/migration/runner.go
+++ b/migration/runner.go
@@ -61,28 +61,16 @@ type MigrationRunner struct {
 // NewRunner creates a migration runner that executes pending migrations.
 //
 // Parameters:
-//   - entries: All registered migrations
-//   - targetVersion: End state to reach after running all migrations
+//   - registry: Migration registry containing all migrations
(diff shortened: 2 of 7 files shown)


migration/runner.go
```diff
@@ -212,10 +207,30 @@ func validateNoVersionDowngrade(current, target SchemaVersion) error {
...
+			flagList[i] = fmt.Sprintf("--migration-%d", idx)
+		}
+		i++
 	}
-	return nil
+
+	return fmt.Errorf(
+		"cannot opt out of previously enabled migrations: %v (set these flags to enable)",
+		flagList,
+	)
```
rodrodros, line 235: 👌

cmd/juno/juno.go
```diff
@@ -101,7 +101,7 @@ const (
 	dbCompactionConcurrencyF            = "db-compaction-concurrency"
 	dbMemtableSizeF                     = "db-memtable-size"
 	dbCompressionF                      = "db-compression"
-	transactionCombinedLayoutF          = "transaction-combined-layout"
+	transactionCombinedLayoutF          = node.FlagTransactionCombinedLayout
```
rodrodros, line 104: Do we need to set this. Looking at the code, but it doesn't seem the global variable is being used anywhere but here?

EgeCaner: I kept it to have exhaustive list of all flags in this file. I can directly use `node.FlagTransactionCombinedLayout` instead. Would you prefer that?

rodrodros: Sorry, I mean the other, `node.FlagTransactionCombinedLayout`, which is only used as an assign in this constant. Why not leave the previous change here?

EgeCaner: I declared the flag in node package to use while registering the migration.

EgeCaner: [See](https://github.com/NethermindEth/juno/blob/293b4b306d397ec4d1df84acb225acfb07ec66d8/node/migration.go#L22)

rodrodros (approved):

PR #3413, opened 2026-02-09
feat: add db-memtable-count configuration option

rodrodros:
## Summary

- Add `--db-memtable-count` CLI flag (default: 2) that maps to Pebble's `MemTableStopWritesThreshold`, controlling how many memtables can queue before stalling writes
- Wire the option through `cmd/juno` → `node.Config` → `pebblev2.WithMemtableCount()`
- Add tests and documentation (config options table + performance tuning guide)

## Test plan

- [x] `go test ./cmd/juno/...` — config precedence tests pass with new field
- [x] `go test ./db/pebblev2/...` — option test verifies `MemTableStopWritesThreshold` is set
- [x] `go test ./node/...` — node package tests pass
- [x] `make lint-diff` — no lint issues

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/db/pebblev2/option_test.go
+++ b/db/pebblev2/option_test.go
@@ -33,9 +33,9 @@ func TestOptions(t *testing.T) {
 		require.NoError(t, option(&opt))
 	}
 
-	assert.Equal(t, opt.Cache.MaxSize(), int64(testCacheSizeMB*1024*1024))
-	assert.Equal(t, opt.MemTableSize, uint64(testMemtableSizeMB*1024*1024))
-	assert.Equal(t, opt.MaxOpenFiles, testMaxOpenFiles)
+	assert.Equal(t, int64(testCacheSizeMB*1024*1024), opt.Cache.MaxSize())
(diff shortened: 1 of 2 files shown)


db/pebblev2/option_test.go
```diff
@@ -34,6 +36,7 @@ func TestOptions(t *testing.T) {
 	assert.Equal(t, opt.Cache.MaxSize(), int64(testCacheSizeMB*1024*1024))
 	assert.Equal(t, opt.MemTableSize, uint64(testMemtableSizeMB*1024*1024))
 	assert.Equal(t, opt.MaxOpenFiles, testMaxOpenFiles)
+	assert.Equal(t, testMemtableCount, opt.MemTableStopWritesThreshold)
```
infrmtcs, line 39: My bad, this line is correct, but the lines above should also put `opt.*` as the last argument.

brbrr (approved):
There is a lint error, aside from that - LGTM 🚢

PR #3414, opened 2026-02-10
refactor(compiler): add Compiler interface and thread through call sites

rodrodros:
Update the architecture to, instead of having a `Compile` function inside the compiler package, having a `Compiler` interface with a `Compile` method. Juno aspects that require compilation were updated to receive this interface:
1. RPC
2. P2P
3. Sequencer

Among the architectural changes is that `Compile` will now receive a context, this is meant to make the compilation process "stop-able" in the future when the context gets cancelled. 

**Note** that there shouldn't be any behaviour change after this commit, since functionality remains the same, the only thing that is changing is how we call the `Compile` function/method.

### DIFF
--- a/rpc/v6/estimate_fee.go
+++ b/rpc/v6/estimate_fee.go
@@ -1,6 +1,7 @@
 package rpcv6
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 
@@ -40,12 +41,13 @@ type FeeEstimate struct {
 *****************************************************/
 
 func (h *Handler) EstimateFee(
+	ctx context.Context,
 	broadcastedTxns BroadcastedTransactionInputs,
 	simulationFlags []SimulationFlag,
 	id BlockID,
 ) ([]FeeEstimate, *jsonrpc.Error) {
 	result, err := h.simulateTransactions(
-		id,
+		ctx, id,
 		broadcastedTxns.Data,
 		append(simulationFlags, SkipFeeChargeFlag),
 		true,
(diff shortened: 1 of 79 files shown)


rpc/v9/estimate_fee.go
```diff
@@ -271,12 +272,13 @@ curl --location 'http://localhost:6060/rpc/v0_8' \
...
 
 func (h *Handler) EstimateFee(
+	ctx context.Context,
 	broadcastedTxns BroadcastedTransactionInputs,
 	simulationFlags []rpcv6.SimulationFlag,
 	id *BlockID,
 ) ([]FeeEstimate, http.Header, *jsonrpc.Error) {
 	txnResults, httpHeader, err := h.simulateTransactions(
-		id,
+		ctx, id,
```
EgeCaner: Nit: one line per param

rpc/v8/simulation.go
```diff
@@ -139,7 +149,10 @@ func prepareTransactions(transactions []BroadcastedTransaction, network *utils.N
 			return nil, nil, nil, jsonrpc.Err(jsonrpc.InvalidParams, "resource_bounds is required for this transaction type")
 		}
 
-		txn, declaredClass, paidFeeOnL1, aErr := AdaptBroadcastedTransaction(&transactions[idx], network)
+		txn, declaredClass, paidFeeOnL1, aErr := AdaptBroadcastedTransaction(
+			ctx, h.compiler,
```
EgeCaner: Nit: One line per param

rpc/v6/estimate_fee.go
```diff
@@ -60,21 +62,27 @@ func (h *Handler) EstimateFee(
...
 type estimateFeeHandler func(
+	ctx context.Context,
 	broadcastedTxns BroadcastedTransactionInputs,
 	simulationFlags []SimulationFlag,
 	id BlockID,
 ) ([]FeeEstimate, *jsonrpc.Error)
 
-func (h *Handler) estimateMessageFee(msg MsgFromL1, id BlockID, f estimateFeeHandler) (*FeeEstimate, *jsonrpc.Error) { //nolint:gocritic
+func (h *Handler) estimateMessageFee(
+	ctx context.Context, msg MsgFromL1, id BlockID, //nolint:gocritic // MsgFromL1 passed by value
```
EgeCaner, line 83: Nit: one line per param

rpc/v6/estimate_fee.go
```diff
@@ -40,12 +41,13 @@ type FeeEstimate struct {
...
 
 func (h *Handler) EstimateFee(
+	ctx context.Context,
 	broadcastedTxns BroadcastedTransactionInputs,
 	simulationFlags []SimulationFlag,
 	id BlockID,
 ) ([]FeeEstimate, *jsonrpc.Error) {
 	result, err := h.simulateTransactions(
-		id,
+		ctx, id,
```
EgeCaner, line 50: Nit: one line per param

rpc/v6/estimate_fee.go
rodrodros: Will ignore RPC v6 since we plan to drop support for them in the very near future

EgeCaner (approved):

PR #3416, opened 2026-02-11
fix: Fix some flaky tests

infrmtcs:


consensus/proposer/proposer_test.go
```diff
@@ -30,7 +30,7 @@ import (
 const (
 	maxConcurrentRead  = 8
 	delayPerRead       = 100 * time.Millisecond
-	waitPerTransaction = 1 * time.Second
+	waitPerTransaction = 3 * time.Second
```
infrmtcs, line 33: This is because CI is slower, so VM can take more time to execute a transaction. This value is the upper bound for assert eventually before fail.

db/typed/prefix/bucket_test.go
```diff
@@ -171,7 +172,6 @@ func randomDeletedRange[K comparable, V any](
 				keys = append(keys, key)
 			}
 		}
-		maxDeletions := layerKeyCount / deleteTests // delete at most half of the entries
```
infrmtcs, line 174: This is because in the worst case, we will run into infinite loop, because no entries are left.

PR #3419, opened 2026-02-13
Fix: add `validate:"dive"` tag for the `rpccore.LimitSlice` type

thiagodeev:
There's a bug in Juno for the RPC methods that are using the `rpccore.LimitSlice` as input (`SimulateTransactions` and  `EstimateFee` for rpcv7, 8, 9, and 10).

https://github.com/NethermindEth/juno/blob/95a1967c94d76edb17475da137791c0efa8f1f95/rpc/v9/transaction.go#L274-L295
Juno currently uses the `"validate"` json tag to validate the tx fields when receiving them from users' requests. It works normally for the `Add...Transaction` methods, but it's currently not working for the mentioned methods above.
The issue is that the Validate tool doesn't work automatically with types inside a slice, unless it's told to do so.
Before, the slice of txs sent by the user was handled automatically by our jsonrpc server ([ref](https://github.com/NethermindEth/juno/blob/95a1967c94d76edb17475da137791c0efa8f1f95/jsonrpc/server.go#L649)), so the validation was working. But with the introduction of the `rpccore.LimitSlice` type, which is not a slice but a struct containing a slice field, since the tool doesn't validate types inside slices out-of-the-box, the validation is not occurring, making the node panic with `nil pointer dereference` when a required field is not specified, instead of returning an error.

The solution is straightforward: just add the `validate:"dive"` tag.

ref: https://pkg.go.dev/github.com/go-playground/validator/v10@v10.27.0#hdr-Dive

### DIFF
--- a/rpc/rpccore/limit_slice_test.go
+++ b/rpc/rpccore/limit_slice_test.go
@@ -60,18 +60,22 @@ func TestLazySlice(t *testing.T) {
 
 	// This test ensures that the validation logic works for the values inside the Data slice.
 	t.Run("ValidateRequiredFields", func(t *testing.T) {
-		type BroadcastedTxLimitSlice = rpccore.LimitSlice[
-			rpcv9.BroadcastedTransaction,
+		// Random type with validation tags
+		type RandType struct {
+			A int `validate:"required"`
+			B int `validate:"gt=5"`
+		}
+
+		type RandLimitSlice = rpccore.LimitSlice[
+			RandType,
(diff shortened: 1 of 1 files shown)


infrmtcs (commented):
Is it possible to write a test case to confirm the expected behaviour? Thanks

MaksymMalicki (approved):

rodrodros:
@thiagodeev the test should test limit slice directly, not a type which uses limit slice.

PR #3426, opened 2026-02-18
refactor: fix and optimize methods of the trie package

thiagodeev:
This PR further enhances the improvements implemented by PR #3370, adding extensive tests, benchmarks, and fixing some bugs.

Bugs fixed in the new implementation of the `EncodedString` method:
- string(b.Bytes()[:]) always pads to 32 bytes instead of using only the active bytes, as the old implementation does. This would lead to a different result than the previous implementation for the cases when not all the 32 bytes were being used. (e.g.: for `BitArray{len: 10, words: [4]uint64{0x03FF}}`, old will produce `[0x03, 0xFF, 0x0A]`, while new would be `[0x00, 0x00, 0x00, 0x00, 0x00, 0x00, ..., 0x03, 0xFF, 0x0A]`
- string(b.len): When you convert an integer to a string in Go using string(n), it interprets n as a Unicode code point (rune) and produces the UTF-8 encoding of that character. So, in the case len >= 128, string(b.len) emits a 2-byte UTF-8 sequence instead of a literal single byte (e.g. û instead of \xfb for len=251). This would lead to a different result than the previous implementation.

Bug fixed in the new implementation of the `nodeKeyByPath` function:
-  Always writing 32 zero bytes for zero-owner keys:   For `ClassTrie` and `ContractTrieContract`, the owner is always the zero address. The new implementation allocated and wrote 32 zero bytes unconditionally, even for empty owners, making every `ClassTrie` and `ContractTrieContract` key 32 bytes longer than expected. This would have broken all DB lookups for class trie and contract trie nodes, since keys written by the old code would no longer match keys computed by the new code.

Bug fixed in the new implementation of the `nodeKeyByHash` function:
- `path.len` (bit count) used as a byte count: `pathSignificantBytes` is a byte count, while `path.len` is a bit count for the BitArray type. In the function body, there was an operation being done with the two values (`pathSize := max(pathSignificantBytes, path.len)`), which was wrong since they serve different purposes. This was leading to a wrong final key calculation, and even panicking sometimes.

#### Benchmarks (some notes at the end):
- `nodeKeyByPath()`: new implementation is ~22% **faster**. (see notes)
New ~57.3 ns/op and 2 allocs, old ~73.7 ns/op and 3 allocs.
- `nodeKeyByHash()`: new implementation is ~22% **faster**.
New ~57.6 ns/op and 1 alloc, old ~73.5 ns/op and 2 allocs.
- `BitArray.Write()`: new implementation is ~10% **faster**.
New ~12 ns/op, old ~13.8 ns/op, both 0 allocs.
- `BitArray.EncodedBytes()`: new implementation is ~40% **faster**. (see notes)
New ~23 ns/op, old ~38.7 ns/op, both 1 alloc.
- `BitArray.EncodedString()`: new implementation is ~20% **slower**. (see notes)
New ~32 ns/op, old ~25 ns/op, both 1 alloc.

#### The new change in the `BitArray.EncodedBytes()` method:
The benchmark result described above for the `EncodedBytes()` method was taking into account the implementation from the PR #3370, but this PR brings a new optimization in the code. By creating a new local helper function, we now avoid copying the entire [32]byte array in the `EncodedBytes()` method, making it more "cheaper", and it is now inlined by the compiler. This makes the current implementation, in some specific circumstances, ~20x faster than the old one, and ~10x faster than the one implemented by the previous PR, with zero allocation.
Fortunately, this is the case for the `nodeKeyByPath` and `EncodedString()` (it is now calling `EncodedBytes()` internally) functions. The `EncodedBytes()`  call is now being inlined by the compiler, and the updated benchmarks are:

- `nodeKeyByPath()`: ~37.3 ns/op and only 1 alloc.
This is ~50% faster than the old one, and ~35% faster than the one in the previous PR.

- `BitArray.EncodedString()`: ~15 ns/op and 0 allocs. New is ~50% **faster**

<br>
<br>
Most of the tests and benchmarks between the old and new versions can be found in this commit hash of this PR: c255f55a7498e53f1838fc92854b815c83d9393f

### DIFF
--- a/core/trie2/trieutils/accessors.go
+++ b/core/trie2/trieutils/accessors.go
@@ -138,10 +138,7 @@ func nodeKeyByPath(prefix db.Bucket, owner *felt.Address, path *Path, isLeaf boo
 		nodeTypeSize = 1
 	)
 
-	key := make([]byte, prefixSize+
-		ownerLen+
-		nodeTypeSize+
-		len(pathBytes))
+	key := make([]byte, prefixSize+ownerLen+nodeTypeSize+len(pathBytes))
 	dst := key
 
 	dst[0] = byte(prefix)
@@ -233,11 +230,7 @@ func nodeKeyByHash(
 		hashSize             = 32
 	)
 
-	key := make([]byte, prefixSize+
-		ownerLen+
-		nodeTypeSize+
-		pathSignificantBytes+
-		hashSize)
+	key := make([]byte, prefixSize+ownerLen+nodeTypeSize+pathSignificantBytes+hashSize)
 	dst := key
 
 	dst[0] = byte(prefix)


core/trie2/trieutils/accessors_test.go
```diff
@@ -0,0 +1,310 @@
...
+		{name: "ContractTrieStorage", prefix: db.ContractTrieStorage},
+	}
+}
+
+func owners() []ownerCase {
+	// they're all real Sepolia addresses
+	return []ownerCase{
+		{
+			name:  "one leading zero",
+			owner: *felt.NewUnsafeFromString[felt.Address]("0x012b66312a967681ca775a20bb3445883b82477888c76790091c0b59593c5f9e"),
```
MaksymMalicki: in all of those 3 cases you could just use `felt.UnsafeFromString[felt.Address]`

core/trie2/trieutils/accessors.go
```diff
@@ -198,57 +204,56 @@ func WriteNodeByHash(
...
-	nodeTypeSize := 1
 	if isLeaf {
 		nodeType = leaf.Byte()
 	} else {
 		nodeType = nonLeaf.Byte()
 	}
 
-	hashBytes := hash.Bytes()
-
+	const prefixSize = 1
```
MaksymMalicki: nit: could be groupped

core/trie2/trieutils/accessors.go
```diff
@@ -136,14 +133,23 @@ func nodeKeyByPath(prefix db.Bucket, owner *felt.Address, path *Path, isLeaf boo
 
 	pathBytes := path.EncodedBytes()
 
-	key := make([]byte, len(prefixBytes)+ownerByteSize+nodeTypeSize+len(pathBytes))
+	const prefixSize = 1
```
MaksymMalicki: nit: could be groupped

core/trie2/trieutils/accessors.go
```diff
@@ -198,57 +204,56 @@ func WriteNodeByHash(
...
 
 	dst[0] = nodeType
 	dst = dst[nodeTypeSize:]
 
-	pathBytes := path.Bytes()
-	copy(dst, pathBytes[0:pathSize])
-	dst = dst[pathSize:]
+	bytes := path.Bytes()
+	activeBytes := bytes[path.inactiveBytes():]
+	copy(dst[:pathSignificantBytes], activeBytes)
```
MaksymMalicki, line 250: just to confirm, in here only the first 8 bytes of the activeBytes will be copied right?

core/trie2/trieutils/accessors_test.go
```diff
@@ -0,0 +1,310 @@
...
+				// [1 byte prefix][1 byte node-type][path]
+				wantLen := 1 + 1 + len(pathEncoded)
+				assert.Equal(t, wantLen, len(key), "wrong key length for zero owner")
+				assert.Equal(t, nodeTypeByte, key[1], "second byte must be node type")
+				assert.Equal(t, pathEncoded, key[2:], "path must follow node type")
+			} else {
+				// StorageTrie of a Contract :
+				// [1 byte prefix][32 bytes owner][1 byte node-type][path]
+				ownerBytes := tc.owner.Bytes()
+				wantLen := 1 + 32 + 1 + len(pathEncoded)
```
MaksymMalicki: nit: can we place the magic numbers as consts?

core/trie2/trieutils/bitarray.go
```diff
@@ -445,37 +455,49 @@ func (b *BitArray) IsEmpty() bool {
...
-	// we could avoid it by passing the return parameter as an arg or returning the [33]byte
 	return encoding[b.inactiveBytes():]
 }
 
+// Serialises the BitArray into a given slice in the following format:
+//   - The first 32 bytes: all bytes in big endian order (including inactive bytes)
+//   - Last byte: length of the bit array (0-255)
+//
+// The slice must be at least 33 bytes long.
+func (b *BitArray) EncodeBytesTo(dst []byte) {
```
MaksymMalicki, line 496: dead code, any plans to use it in the future?

core/trie2/trieutils/bitarray.go
```diff
@@ -51,13 +51,23 @@ func (b *BitArray) Len() uint8 {
...
+	return res
+}
 
-	binary.BigEndian.PutUint64(res[0:8], b.words[3])
-	binary.BigEndian.PutUint64(res[8:16], b.words[2])
-	binary.BigEndian.PutUint64(res[16:24], b.words[1])
-	binary.BigEndian.PutUint64(res[24:32], b.words[0])
+// Writes the bytes representation of the bit array to a given slice in big endian format.
+// The slice must be at least 32 bytes long.
+func (b *BitArray) BytesTo(dst []byte) {
```
MaksymMalicki, line 60: dead code, any plans to use it in the future?

core/trie2/trieutils/bitarray.go
thiagodeev: Indeed. They're not used.

> any plans to use it in the future?
Yes and no. I've created these methods as alternatives to the current ones in scenarios of high performance. I didn't want to refactor the code to use them within this PR; it would be out of scope.

IMO, we could let them be there; we could use them when creating new functions using the bitarray that requires high performance, and in a future PR, refactoring the code to use them. WDYT?

Also, one thing to note is that currently we already have more than 5 unused methods of the BitArray type (IsBitSetFromLSB, IsBitSet...)

core/trie2/trieutils/bitarray.go
thiagodeev: same as the above comment

core/trie2/trieutils/accessors.go
thiagodeev: Yes.
About `copy`: " The number of elements copied is the minimum of the length of the source and destination slices". 
`dst[:pathSignificantBytes]` = `dst[:8]`
Only the first 8 bytes from activeBytes will be copied into dst. If activeBytes is shorter than 8 bytes, then only its full length will be copied (since dst is already filled with zeros, this is the expected output). 

Ex:
activeBytes =[10 20 30]
After the copy, dst becomes:
[10 20 30 0 0 0 0 0]

core/trie2/trieutils/accessors.go
```diff
@@ -136,14 +133,25 @@ func nodeKeyByPath(prefix db.Bucket, owner *felt.Address, path *Path, isLeaf boo
...
+	const (
+		prefixSize   = 1
+		nodeTypeSize = 1
+	)
+
+	key := make([]byte, prefixSize+
+		ownerLen+
+		nodeTypeSize+
+		len(pathBytes))
 	dst := key
```
rodrodros, line 142: Project style guidance:
1. Either all args on the same line (given that the line is smaller than 100 chars)
2. Each arg receiies their own line.

If the call to make used less than a 100 chars it would be
```golang
key := make(arg1, ..., argn)
```

If it goes beyond then:
```golang
key := make(
   arg1, ..., argn,
)
```

If it still goes beyond than 100 chars then
```golang
key := make(
   arg1,
   ...,
   argn,
)

File: core/trie2/trieutils/accessors.go [hunk 9]
```diff
@@ -198,57 +206,58 @@ func WriteNodeByHash(
...
+		nodeTypeSize         = 1
+		pathSignificantBytes = 8
+		hashSize             = 32
+	)
+
+	key := make([]byte, prefixSize+
+		ownerLen+
+		nodeTypeSize+
+		pathSignificantBytes+
+		hashSize)
```
rodrodros: Please apply similar style guidance as above comment 🙏

core/trie2/trieutils/accessors_test.go
```diff
@@ -0,0 +1,325 @@
...
+func prefixes() []prefixCase {
+	return []prefixCase{
+		{name: "ClassTrie", prefix: db.ClassTrie},
+		{name: "ContractTrieContract", prefix: db.ContractTrieContract},
+		{name: "ContractTrieStorage", prefix: db.ContractTrieStorage},
+	}
+}
+
+func owners() []ownerCase {
+	// they're all real Sepolia addresses
```
rodrodros, line 48: 👍

core/trie2/trieutils/accessors.go
thiagodeev: Thanks! Done

PR #3427, opened 2026-02-19
feat: add staleness check for docs and versioned docs

rodrodros:
Every time we make a new release, there is no check to verify that the latest docs are released as well. This adds a build action for that


brbrr (approved):
LGTM!

PR #3436, opened 2026-02-24
feat: update go dependencies + minor fix

rodrodros:
Additionally, fixed a small race condition in consensus package, shown by the change in behaviour of testify v1.11.0.

Before, there was a slight delay when using asserting with `require.EventuallyWithT` where it would wait a few miliseconds before triggering the condition (v1.10.0) vs triggering the condition immediately (v1.11.x). 

Description of the race condition in a comment in `proposer.New`.

### DIFF
--- a/consensus/proposer/proposer_test.go
+++ b/consensus/proposer/proposer_test.go
@@ -73,8 +73,9 @@ func TestProposer(t *testing.T) {
 
 	t.Run("Initial state", func(t *testing.T) {
 		requireEventually(t, 1, func(c *assert.CollectT) {
-			assert.NotNil(c, p.Preconfirmed())
-			assert.Empty(c, p.Preconfirmed().Block.Transactions)
+			preconfirmed := p.Preconfirmed()
+			require.NotNil(c, preconfirmed)
+			assert.Empty(c, preconfirmed.Block.Transactions)
 		})
 	})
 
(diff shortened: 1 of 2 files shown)


consensus/proposer/proposer_test.go
```diff
@@ -73,8 +73,11 @@ func TestProposer(t *testing.T) {
 
 	t.Run("Initial state", func(t *testing.T) {
 		requireEventually(t, 1, func(c *assert.CollectT) {
-			assert.NotNil(c, p.Preconfirmed())
-			assert.Empty(c, p.Preconfirmed().Block.Transactions)
+			preconfirmed := p.Preconfirmed()
+			if !assert.NotNil(c, preconfirmed) {
+				return
+			}
```
thiagodeev: Is there a reason not to use `require` here, and also in the other cases in this file?

```suggestion
			preconfirmed := p.Preconfirmed()
			require.NotNil(c, preconfirmed)
```

rodrodros: require makes the test fail immediately, while assert let's it continue. It's tied with the use of EventuallyWithT which collects fail assertions. 

Fail assertions only count in this case, if the condition was never met "eventually"

rodrodros: Although, on a second thought maybe, it is the same 🤔

thiagodeev: Yeah, I also think so. Both 'assert' and 'require' fail the test, but 'assert' allows us to continue the test. However, for this case, we are returning just after the 'assert' if it fails, so it seems 'require' will have essentially the same behavior

rodrodros: I believe it's a bit more complicated than that, `require` will call `t.FailNow` which then triggering `runtime.Goexit()` while returning also finishes the condition running. Both ways succesfully tell `EventuallyWithT` to retry in the next tick, although in two different ways.

PR #3439, opened 2026-02-27
feat(rpc): make v10 self contained

rodrodros:
As title says, make RPC v10 self contained so that it no longer depends on older versions such as v6, v7, v8 and v9.

Although this create some code duplication, it will allow us to have faster development on the latest version and improving the RPC little by little

### DIFF
--- a/rpc/v10/adapt_trace.go
+++ b/rpc/v10/adapt_trace.go
@@ -316,7 +316,7 @@ func AdaptFeederBlockTrace(
 		return nil, errors.New("mismatched number of txs and traces")
 	}
 
-	// Adapt every feeder block trace to rpc v9 trace
+	// Adapt every feeder block trace to rpc v10 trace
 	adaptedTraces := make([]TracedBlockTransaction, len(blockTrace.Traces))
 	for index := range blockTrace.Traces {
 		feederTrace := &blockTrace.Traces[index]
--- a/rpc/v10/block.go
+++ b/rpc/v10/block.go
@@ -132,7 +132,7 @@ func (h *Handler) BlockWithTxHashes(id *BlockID) (*BlockWithTxHashes, *jsonrpc.E
 	}, nil
 }
 
-// BlockWithTxHashes returns the block information with transaction receipts given a block ID.
+// BlockWithReceipts returns the block information with transaction receipts given a block ID.
 //
 // It follows the specification defined here:
 // https://github.com/starkware-libs/starknet-specs/blob/cce1563eff702c87590bad3a48382d2febf1f7d9/api/starknet_api_openrpc.json#L99
--- a/rpc/v10/simulation.go
+++ b/rpc/v10/simulation.go
@@ -383,7 +383,7 @@ func createSimulatedTransactions(
 
 	simulatedTransactions := make([]SimulatedTransaction, len(overallFees))
 	for i, overallFee := range overallFees {
-		// Adapt transaction trace to rpc v9 trace
+		// Adapt transaction trace to rpc v10 trace
 		trace := utils.HeapPtr(AdaptVMTransactionTrace(&traces[i]))
 
 		// Add root level execution resources
--- a/rpc/v10/trace.go
+++ b/rpc/v10/trace.go
@@ -210,7 +210,7 @@ func traceTransactionsWithState(
 	// Adapt traces
 	traces := make([]TracedBlockTransaction, len(executionResult.Traces))
 	for index := range executionResult.Traces {
-		// Adapt vm transaction trace to rpc v9 trace and add root level execution resources
+		// Adapt vm transaction trace to rpc v10 trace and add root level execution resources
 		trace := AdaptVMTransactionTrace(&executionResult.Traces[index])
 
 		trace.ExecutionResources = &ExecutionResources{
--- a/rpc/v9/block.go
+++ b/rpc/v9/block.go
@@ -274,7 +274,7 @@ func (h *Handler) BlockWithTxHashes(id *BlockID) (*BlockWithTxHashes, *jsonrpc.E
 	}, nil
 }
 
-// BlockWithTxHashes returns the block information with transaction receipts given a block ID.
+// BlockWithReceipts returns the block information with transaction receipts given a block ID.
 //
 // It follows the specification defined here:
 // https://github.com/starkware-libs/starknet-specs/blob/9377851884da5c81f757b6ae0ed47e84f9e7c058/api/starknet_api_openrpc.json#L99


rpc/v10/block_test.go
```diff
@@ -331,41 +328,42 @@ func assertTransactionsWithReceiptsEq(
...
+		adaptedReceipt := rpcv10.AdaptReceipt(
 			expectedReceipt,
 			expectedBlock.Transactions[i],
 			expectedTxnFinalityStatus,
 		)
 
-		require.Equal(t, adaptedTransaction.Transaction, actual[i].Transaction.Transaction)
-		require.Equal(t, adaptedTransaction.ProofFacts, actual[i].Transaction.ProofFacts)
+		actualTx := *actual[i].Transaction
+		actualTx.Hash = nil
```
thiagodeev: This line wasn't presented in the old version; the test already passes without this (it seems the hash is already nil). We could remove it

```suggestion
```

rpc/v10/transaction.go
```diff
@@ -13,23 +13,14 @@ import (
...
 	"github.com/NethermindEth/juno/utils"
 	"go.uber.org/zap"
 )
 
-type Transaction struct {
-	rpcv9.Transaction
-	ProofFacts []*felt.Felt `json:"proof_facts,omitempty"`
-}
-
 func AdaptTransaction(coreTx core.Transaction, includeProofFacts bool) Transaction {
```
thiagodeev, line 26: We can also remove this function. It's the same case as the `AdaptRPCTxToAddTxGatewayPayload` function (see my coment about it). It was created just to be a wrapper to the `v9.AdaptTransaction` function. Since now we are no longer importing `v9` code, we can remove it.

In order to do so, we need to:
1. Remove this function.
2. Rename the current `AdaptCoreTransaction` back to the original `AdaptTransaction` name. To make it work, we will need to:
    - inside the function, there's a call to the `adaptInvokeTransaction` function. We need to change this function to make it set the `ProofFacts` field to the returned `Transaction` pointer.

rpc/v10/adapt_transaction.go
```diff
@@ -0,0 +1,421 @@
...
+
+// AddTxGatewayPayload embeds a starknet.Transaction along with an optional ContractClass.
+type AddTxGatewayPayload struct {
+	starknet.Transaction
+	ContractClass json.RawMessage `json:"contract_class,omitempty"`
+}
+
+// AdaptCoreTransaction adapts a core.Transaction to a local *Transaction.
+// This is the v10-local equivalent of v9's AdaptTransaction.
+func AdaptCoreTransaction(t core.Transaction) *Transaction {
```
thiagodeev, line 30: We should rename this function back to the `AdaptTransaction` name (and remove the current one with this name) as I mentioned in my comment about the `AdaptTransaction` function. Also, let's update this description.

rpc/v10/state_update.go
```diff
@@ -22,14 +20,44 @@ type StateUpdate struct {
...
+	ContractAddress felt.Felt `json:"contract_address"`
+	Nonce           felt.Felt `json:"nonce"`
+}
+
+type DeployedContract struct {
+	Address   felt.Felt `json:"address"`
+	ClassHash felt.Felt `json:"class_hash"`
+}
+
+type DeclaredClassDiff struct {
```
thiagodeev: I think it would be good to keep the same 'DeclaredClass' name ([ref](https://github.com/NethermindEth/juno/blob/bab85294130976d1725543545d96c2ddb93acbb2/rpc/v6/state_update.go#L55))

rpc/v10/simulation.go
```diff
@@ -259,21 +258,33 @@ func (h *Handler) simulateTransactions(
...
+		576460752303421872,
+	},
+))
+
+// todo(rdr): does this check still make sense. Transaction previous to version 3 might not
+// be allowed anymore
 func isVersion3(version *felt.Felt) bool {
-	return version != nil && version.Equal(&rpcv6.RPCVersion3Value)
+	return version != nil && version.Equal(&RPCVersion3Value)
 }
```
thiagodeev: I know that this came from the old code, but what about creating the felt from the number? More clear. (I tested it locally, it's the same thing)

```suggestion
// A new Felt can be created with `felt.FromFelt` in this case
var RPCVersion3Value = felt.NewFromUint64[felt.Felt](3)

func isVersion3(version *felt.Felt) bool {
	return version != nil && version.Equal(RPCVersion3Value)
}
```

Also, I agree about the check. At least for the broadcast tx case, this check is not necessary since there's a 'validate' JSON tag rule that only allows v3 txs. But I think this should be done in a separate PR, which should also change other RPC pkgs if that's the case. I can create an issue for this if you want

rpc/v10/transaction_types.go
```diff
@@ -0,0 +1,449 @@
...
+	ProofFacts            []*felt.Felt          `json:"proof_facts,omitempty"`
+}
+
+// BroadcastedTransaction represents a transaction submitted via the RPC API.
+type BroadcastedTransaction struct {
+	Transaction
+	ContractClass json.RawMessage `json:"contract_class,omitempty" validate:"required_if=Transaction.Type DECLARE"`    //nolint:lll // validate tag
+	PaidFeeOnL1   *felt.Felt      `json:"paid_fee_on_l1,omitempty" validate:"required_if=Transaction.Type L1_HANDLER"` //nolint:lll // validate tag
+	Proof         utils.Base64    `json:"proof,omitempty" validate:"excluded_unless=Type INVOKE,omitempty,base64"`     //nolint:lll // validate tag
+	ProofFacts    []felt.Felt     `json:"proof_facts,omitempty" validate:"excluded_unless=Type INVOKE"`
```
thiagodeev: Duplicated field. The new `Transaction` type already contains the `ProofFacts` field

rpc/v10/adapt_transaction.go
```diff
@@ -0,0 +1,421 @@
...
+		CallData:              rpcTx.CallData,
+		EntryPointSelector:    rpcTx.EntryPointSelector,
+		Nonce:                 rpcTx.Nonce,
+		CompiledClassHash:     rpcTx.CompiledClassHash,
+		ResourceBounds:        resourceBoundsPtr,
+		Tip:                   rpcTx.Tip,
+		NonceDAMode:           nonceDAModePtr,
+		FeeDAMode:             feeDAModePtr,
+		AccountDeploymentData: rpcTx.AccountDeploymentData,
+		PaymasterData:         rpcTx.PaymasterData,
```
thiagodeev: TODO: include the ProofFacts here

rpc/v10/transaction.go
```diff
@@ -294,9 +316,8 @@ func adaptRPCTxToFeederPayload(
 	tx *BroadcastedTransaction,
 ) addTxGatewayPayload {
```
thiagodeev: We can remove this function. It was created in the RPCv10 PR only to support the new invoke fields. Since our old code was sharing the code with v9, we had to create this function to wrap the call to `v9.AdaptRPCTxToAddTxGatewayPayload` and then add the new v10 fields in the payload (also a custom type, wrapping a `v9` type).

Now, as we have a separate code for v10, we can just change the code to accomodate the new changes. In order to do so, AFAIK, we need to do the following:

1. Remove this function, use the `AdaptRPCTxToAddTxGatewayPayload` directly in the code. But we need to:
2. Inside the `AdaptRPCTxToAddTxGatewayPayload`, set the `Proof` and `ProofFacts` field to the `AddTxGatewayPayload`. But we need to:
    1. Change the `AdaptRPCTxToFeederTx` function to make it set the `ProofFacts` in the returned `starknet.Transaction` type.
    2. Add the `Proof` field to the `AddTxGatewayPayload` type (it already has the `ProofFacts` through the inner `starknet.Transaction` field)
3. Then, we could also remove the `addTxGatewayPayload` wrapper type, since we no longer would need it.

I'll leave a comment on each corresponding part of the code pointing to these steps.

rpc/v10/transaction.go
```diff
@@ -244,7 +266,7 @@ func (h *Handler) pushToFeederGateway(
 				fmt.Sprintf("marshal revised contract class: %v", err),
 			)
 		}
-		v9Tx.ContractClass = newContractClass
+		tx.ContractClass = newContractClass
 	}
 
 	payload := adaptRPCTxToFeederPayload(tx)
```
thiagodeev: Point 1 of my comment about the `AdaptRPCTxToAddTxGatewayPayload` function.
Here is where we should call the `AdaptRPCTxToAddTxGatewayPayload` directly.

rpc/v10/transaction.go
```diff
@@ -285,7 +307,7 @@ func (h *Handler) pushToFeederGateway(
 }
 
 type addTxGatewayPayload struct {
-	rpcv9.AddTxGatewayPayload
+	AddTxGatewayPayload
 	Proof      utils.Base64 `json:"proof,omitempty"`
 	ProofFacts []felt.Felt  `json:"proof_facts,omitempty"`
 }
```
thiagodeev: Point 3 of my comment about the `AdaptRPCTxToAddTxGatewayPayload` function.
We can remove this wrapper type, since we could implement the field directly in the `AddTxGatewayPayload` type.

rpc/v10/adapt_transaction.go
```diff
@@ -0,0 +1,421 @@
...
+	}
+}
+
+func AdaptRPCTxToAddTxGatewayPayload(rpcTx *BroadcastedTransaction) AddTxGatewayPayload {
+	return AddTxGatewayPayload{
+		Transaction:   AdaptRPCTxToFeederTx(&rpcTx.Transaction),
+		ContractClass: rpcTx.ContractClass,
+	}
```
thiagodeev, line 181: Point 2 of my comment about the `AdaptRPCTxToAddTxGatewayPayload` function.
We must set the `Proof` and `ProofFacts` field from the `BroadcastedTransaction` into the `AddTxGatewayPayload` here.

rpc/v10/adapt_transaction.go
```diff
@@ -0,0 +1,421 @@
...
+		EntryPointSelector:    rpcTx.EntryPointSelector,
+		Nonce:                 rpcTx.Nonce,
+		CompiledClassHash:     rpcTx.CompiledClassHash,
+		ResourceBounds:        resourceBoundsPtr,
+		Tip:                   rpcTx.Tip,
+		NonceDAMode:           nonceDAModePtr,
+		FeeDAMode:             feeDAModePtr,
+		AccountDeploymentData: rpcTx.AccountDeploymentData,
+		PaymasterData:         rpcTx.PaymasterData,
+	}
```
thiagodeev, line 173: Point 2.1 of my comment about the `AdaptRPCTxToAddTxGatewayPayload` function.
We need to change this function to set the new `rpcTx.ProofFacts` field into the returned `starknet.Transaction` type.

rpc/v10/adapt_transaction.go
```diff
@@ -0,0 +1,421 @@
...
+	"github.com/NethermindEth/juno/utils"
+)
+
+var ErrTransactionNotFound = errors.New("transaction not found")
+
+// AddTxGatewayPayload embeds a starknet.Transaction along with an optional ContractClass.
+type AddTxGatewayPayload struct {
+	starknet.Transaction
+	ContractClass json.RawMessage `json:"contract_class,omitempty"`
+}
```
thiagodeev, line 26: Point 2.2 of my comment about the `AdaptRPCTxToAddTxGatewayPayload` function.
We need to add the new `Proof` field here (we already have the `ProofFacts` in the `starknet.Transaction` type)

rpc/v10/adapt_transaction.go
```diff
@@ -0,0 +1,421 @@
...
+	}
+
+	if tx.Version.Uint64() == 3 {
+		tx.ResourceBounds = utils.HeapPtr(adaptResourceBounds(t.ResourceBounds))
+		tx.Tip = felt.NewFromUint64[felt.Felt](t.Tip)
+		tx.PaymasterData = &t.PaymasterData
+		tx.AccountDeploymentData = &t.AccountDeploymentData
+		tx.NonceDAMode = utils.HeapPtr(DataAvailabilityMode(t.NonceDAMode))
+		tx.FeeDAMode = utils.HeapPtr(DataAvailabilityMode(t.FeeDAMode))
+	}
```
thiagodeev, line 330: We should set the `ProofFacts` here, as I mentioned in my comment about the `AdaptTransaction` function.

rpc/v10/simulation.go
rodrodros: Yes, please create the issue 🙏

rpc/v10/transaction_types.go
```diff
@@ -398,37 +398,40 @@ func adaptExecutionResources(resources *core.ExecutionResources) *ExecutionResou
 }
 
 // Transaction represents a Starknet transaction in the RPC API.
+//
+//nolint:lll // We can't break the json tags lines
```
rodrodros, line 402: 👍

rpc/v10/transaction_types.go
```diff
@@ -398,37 +398,40 @@ func adaptExecutionResources(resources *core.ExecutionResources) *ExecutionResou
...
-	ProofFacts            []*felt.Felt          `json:"proof_facts,omitempty"`
+	AccountDeploymentData *[]*felt.Felt         `json:"account_deployment_data,omitempty" validate:"required_if=Type INVOKE,required_if=Type DECLARE"`
+	NonceDAMode           *DataAvailabilityMode `json:"nonce_data_availability_mode,omitempty" validate:"required"`
+	FeeDAMode             *DataAvailabilityMode `json:"fee_data_availability_mode,omitempty" validate:"required"`
+	ProofFacts            []felt.Felt           `json:"proof_facts,omitempty" validate:"excluded_unless=Type INVOKE"`
 }
 
 // BroadcastedTransaction represents a transaction submitted via the RPC API.
+//
+//nolint:lll // We can't break the json tags lines
```
rodrodros: 👍

rpc/v10/transaction_types.go
```diff
@@ -421,7 +421,7 @@ type Transaction struct {
 	AccountDeploymentData *[]*felt.Felt         `json:"account_deployment_data,omitempty" validate:"required_if=Type INVOKE,required_if=Type DECLARE"`
 	NonceDAMode           *DataAvailabilityMode `json:"nonce_data_availability_mode,omitempty" validate:"required"`
 	FeeDAMode             *DataAvailabilityMode `json:"fee_data_availability_mode,omitempty" validate:"required"`
-	ProofFacts            []felt.Felt           `json:"proof_facts,omitempty" validate:"excluded_unless=Type INVOKE"`
+	ProofFacts            *[]*felt.Felt         `json:"proof_facts,omitempty" validate:"excluded_unless=Type INVOKE"`
```
rodrodros: Why this change?

thiagodeev: A few reasons (to follow the current standard, to facilitate the work with this field since all other fields are designed in the same way...), but the most important is that this allows us to return an empty array in a final json (it'll be necessary in the next rpc rc-3 implementation)

By using []*felt.Felt with `omitempty`:
- ProofFacts: nil -> JSON `{}` (omitted)
- ProofFacts: []*felt.Felt{} (empty) -> JSON `{}` (omitted)
- ProofFacts: []*felt.Felt{"0x1", "0x2"} -> JSON `{"proof_facts: ["0x1", "0x2"]}`

By using *[]*felt.Felt with `omitempty`:
- ProofFacts: nil -> JSON `{}` (omitted) - same
- ProofFacts: &[]*felt.Felt{} (a pointer to an empty slice) -> JSON `{"proof_facts: []}` **an empty array!**
- ProofFacts:  &[]*felt.Felt{"0x1", "0x2"} -> JSON `{"proof_facts: ["0x1", "0x2"]}` - same

We could also use the 'omitzero', but by using the slice pointer, we can customize it better (like choose to omit it or not based on the 'includeProofFacts' flag

rodrodros: > to follow the current standard

Juno's coding standard has improved in the last year by not following some of it standards. Point is, standards in Juno shouldn't be followed blindly. Should be critiqued and kept/changed based on the discussion. 

From a JSON perspective, I agree with the `*[]felt.Felt`, but I don't agree with the `*[]*felt.Felt`. It doesn't affect the output and generates `n` extra pointers. 

Pointers should only be used when:
* passing big function params
* returning a type meant to be used by several go-routines or track a global state
* representing the optionality of a type

I don't see any of these applying.

Using your previous examples with `*[]felt.Felt` they have the same behaviour.

thiagodeev: Got it! I'll modify the code to use `*[]felt.Felt` then

rpc/v10/transaction.go
thiagodeev: I've changed my mind about it.
Indeed, since we are copying all v9 code into v10, we can update it to include the new fields instead of a wrapping function as we're doing before this PR; this is what I was talking about.
However, while I was coding, I realized there's another thing at play: whether we should show the `proof_facts` field or not in the final JSON based on the `includeProofFacts` flag. Therefore, I decided to keep things as they are, but change the purpose from being just a wrapper to add the new field to be a feature that will omit the field based on the `includeProofFacts` flag

rpc/v10/adapt_transaction.go
thiagodeev: I changed my mind. Ref https://github.com/NethermindEth/juno/pull/3439#discussion_r2904880260

rpc/v10/adapt_transaction.go
```diff
@@ -0,0 +1,428 @@
...
+
+	return tx
+}
+
+func adaptDeployAccountTrandaction(t *core.DeployAccountTransaction) *Transaction {
```
MaksymMalicki: nit: `adaptDeployAccountTransaction`

rpc/v10/adapt_trace.go
```diff
@@ -0,0 +1,436 @@
...
+) ([]TracedBlockTransaction, error) {
+	if blockTrace == nil {
+		return nil, nil
+	}
+
+	if len(block.Transactions) != len(blockTrace.Traces) {
+		return nil, errors.New("mismatched number of txs and traces")
+	}
+
+	// Adapt every feeder block trace to rpc v9 trace
```
MaksymMalicki: nit: remove/update the comment

rpc/v10/trace.go
```diff
@@ -214,8 +213,8 @@ func traceTransactionsWithState(
 		// Adapt vm transaction trace to rpc v9 trace and add root level execution resources
```
MaksymMalicki: nit: remove/adapt the comment

rpc/v10/simulation.go
MaksymMalicki: nit: update/remove the comment

rpc/v10/block.go
MaksymMalicki: nit: `BlockWithReceipts` instead of `BlockWithTxHashes`

rpc/v10/state_update.go
```diff
@@ -22,14 +20,44 @@ type StateUpdate struct {
...
-	Nonces                    []rpcv6.Nonce            `json:"nonces"`
-	DeployedContracts         []rpcv6.DeployedContract `json:"deployed_contracts"`
-	DeprecatedDeclaredClasses []*felt.Felt             `json:"deprecated_declared_classes"`
-	DeclaredClasses           []rpcv6.DeclaredClass    `json:"declared_classes"`
-	ReplacedClasses           []rpcv6.ReplacedClass    `json:"replaced_classes"`
-	MigratedCompiledClasses   []MigratedCompiledClass  `json:"migrated_compiled_classes"`
+	StorageDiffs              []StorageDiff           `json:"storage_diffs"`
+	Nonces                    []Nonce                 `json:"nonces"`
+	DeployedContracts         []DeployedContract      `json:"deployed_contracts"`
+	DeprecatedDeclaredClasses []*felt.Felt            `json:"deprecated_declared_classes"`
```
MaksymMalicki, line 57: Question: should we use a slice of `felt.Felt` pointers here? Does it make sense? I see that this struct directly maps to the `core` `StateUpdate`, maybe we can leave a `todo` here for the future change

thiagodeev: Discussed in the daily. We'll make the changes in the future

PR #3444, opened 2026-03-02
fix(state): revert migrated casm classes in old state

EgeCaner:
Issue: 

On reorg casm hash V1 was directly written to classes trie, however poseidon hash of the casm hash V1 should be stored in trie.

Optimization:

Rather than reading the class then fetching the state update of block that class has been declared to fetch the casm hash V1, now we read it from casm hash metadata bucket.

### DIFF
--- a/blockchain/blockchain_test.go
+++ b/blockchain/blockchain_test.go
@@ -885,6 +885,9 @@ func TestRevertHeadMigratedCasmClasses(t *testing.T) {
 	gotCasmHash, err := state.CompiledClassHash(&sierraHash)
 	require.NoError(t, err)
 	assert.Equal(t, v2CasmHash, gotCasmHash, "should return V2 after migrating class")
+	gotRoot, err := chain.StateCommitment()
+	require.NoError(t, err)
+	assert.Equal(t, stateUpdate1.NewRoot, &gotRoot)
 
 	// Revert head should revert the state to casm hash v1
 	require.NoError(t, chain.RevertHead())
@@ -897,7 +900,7 @@ func TestRevertHeadMigratedCasmClasses(t *testing.T) {
 	require.NoError(t, err)
 	assert.Equal(t, v1CasmHash, gotCasmHash, "should return V1 after reverting migrated class")
 
-	gotRoot, err := chain.StateCommitment()
+	gotRoot, err = chain.StateCommitment()
 	require.NoError(t, err)
 	assert.Equal(t, stateUpdate0.NewRoot, &gotRoot, "state root after revert should match block 0")
 }


blockchain/blockchain_test.go
```diff
@@ -794,6 +794,106 @@ func TestRevert(t *testing.T) {
 	})
 }
 
+// TestRevertHeadMigratedCasmClasses ensures that after storing a block with
+// MigratedClasses and then reverting, the classes trie and CASM hash metadata
+// are correctly reverted to V1.
+func TestRevertHeadMigratedCasmClasses(t *testing.T) {
```
rodrodros, line 800: Thank you 👍

blockchain/blockchain_test.go
```diff
@@ -794,6 +794,106 @@ func TestRevert(t *testing.T) {
...
+	// Revert head should revert the state to casm hash v1
+	require.NoError(t, chain.RevertHead())
+
+	state, closer, err := chain.HeadState()
+	require.NoError(t, err)
+	defer func() { _ = closer() }()
+
+	gotCasmHash, err := state.CompiledClassHash(&sierraHash)
+	require.NoError(t, err)
+	assert.Equal(t, v1CasmHash, gotCasmHash, "should return V1 after reverting migrated class")
```
rodrodros: Can we add a check before reverting verifying we get v2 and after reverting we get v1?

core/state_test.go
```diff
@@ -844,3 +844,69 @@ func TestCompiledClassHashAt(t *testing.T) {
...
 }
+
+func TestRevertMigratedCasmClasses(t *testing.T) {
+	testDB := memory.New()
+	txn := testDB.NewIndexedBatch()
+	state := core.NewDeprecatedState(txn)
+
+	sierraHash := felt.UnsafeFromString[felt.SierraClassHash]("0x1234")
+	v1CasmHash := felt.UnsafeFromString[felt.CasmClassHash]("0x1111")
+	v2CasmHash := felt.UnsafeFromString[felt.CasmClassHash]("0x2222")
```
rodrodros: This can be `felt.FromUint(0x1234)`

core/state_test.go
```diff
@@ -844,3 +844,69 @@ func TestCompiledClassHashAt(t *testing.T) {
...
+	require.NoError(t, err)
+	su1.NewRoot = &newRoot
+
+	// Revert block 1
+	require.NoError(t, state.Revert(1, su1))
+
+	// Classes trie and metadata should be back to V1
+	gotRoot, err := state.Commitment()
+	require.NoError(t, err)
+	assert.Equal(t, root0, gotRoot, "state root after revert should match pre-migration root")
```
rodrodros, line 911: Similar comment here as well as in the test before. Wdyt?

PR #3445, opened 2026-03-02
fix(state): make commitment formula version-aware for v0.14+

m-kus:
## Summary

  - `DeprecatedState.Commitment()` and `stateCommitment()` unconditionally return storageRoot when
  `classesRoot == 0`, which is incorrect for Starknet v0.14.0+ where the formula always applies
  `Poseidon([STARKNET_STATE_V0, storageRoot, classesRoot])`
  - This causes a state root mismatch on any v0.14+ network whose genesis block has no declared classes
  (zero class commitment)

## Fix

Thread protocolVersion string through the State interface and version-gate the `classesRoot.IsZero()` shortcut:

```go
  ver, _ := ParseBlockVersion(protocolVersion)
  if classesRoot.IsZero() && ver.LessThan(Ver0_14_0) {
      return storageRoot, nil
  }
```

Notes:
- `protocolVersion` is passed as a parameter rather than stored as struct state, keeping the formula explicitly version-aware at each call site.
- Empty string maps to version 0.0.0 via ParseBlockVersion, preserving pre-0.14 behavior for all existing tests.

### DIFF
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -328,7 +328,7 @@ func (b *Blockchain) Store(
 		}
 
 		state := core.NewDeprecatedState(txn)
-		err := state.Update(block.Number, stateUpdate, newClasses, false, block.ProtocolVersion)
+		err := state.Update(block.Header, stateUpdate, newClasses, false)
 		if err != nil {
 			return err
 		}
@@ -705,7 +705,7 @@ func (b *Blockchain) revertHead(txn db.IndexedBatch) error {
 
 	state := core.NewDeprecatedState(txn)
 	// revert state
-	if err = state.Revert(blockNumber, stateUpdate, header.ProtocolVersion); err != nil {
+	if err = state.Revert(header, stateUpdate); err != nil {
 		return err
 	}
 
@@ -850,7 +850,7 @@ func (b *Blockchain) updateStateRoots(
 	stateUpdate.OldRoot = &oldStateRoot
 
 	// Apply state update
-	err = state.Update(block.Number, stateUpdate, newClasses, true, block.ProtocolVersion)
+	err = state.Update(block.Header, stateUpdate, newClasses, true)
 	if err != nil {
 		return err
 	}
--- a/core/state.go
+++ b/core/state.go
@@ -213,12 +213,13 @@ func (s *DeprecatedState) verifyStateUpdateRoot(root *felt.Felt, protocolVersion
 // old or new root does not match the state's old or new roots,
 // [ErrMismatchedRoot] is returned.
 func (s *DeprecatedState) Update(
-	blockNumber uint64,
+	header *Header,
 	update *StateUpdate,
 	declaredClasses map[felt.Felt]ClassDefinition,
 	skipVerifyNewRoot bool,
-	protocolVersion string,
 ) error {
+	blockNumber := header.Number
+	protocolVersion := header.ProtocolVersion
 	err := s.verifyStateUpdateRoot(update.OldRoot, protocolVersion)
 	if err != nil {
 		return err
@@ -616,8 +617,10 @@ func (s *DeprecatedState) ContractDeployedAt(addr *felt.Felt, blockNumber uint64
 }
 
 func (s *DeprecatedState) Revert(
-	blockNumber uint64, update *StateUpdate, protocolVersion string,
+	header *Header, update *StateUpdate,
 ) error {
+	blockNumber := header.Number
+	protocolVersion := header.ProtocolVersion
 	err := s.verifyStateUpdateRoot(update.NewRoot, protocolVersion)
 	if err != nil {
 		return fmt.Errorf("verify state update root: %v", err)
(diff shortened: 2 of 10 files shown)


blockchain/blockchain.go
```diff
@@ -317,7 +328,7 @@ func (b *Blockchain) Store(
 		}
 
 		state := core.NewDeprecatedState(txn)
-		err := state.Update(block.Number, stateUpdate, newClasses, false)
+		err := state.Update(block.Number, stateUpdate, newClasses, false, block.ProtocolVersion)
 		if err != nil {
```
rodrodros: Could you update this methods which receives a `block` to only receive the block. Note that they are receiving both `block.number` and `block.ProtocolVersion`. It would be ideal if this method would pass the `block` variable and the function inside access the `Number` and `ProtocolVersion` fields

blockchain/blockchain.go
```diff
@@ -660,14 +671,14 @@ func (b *Blockchain) revertHead(txn db.IndexedBatch) error {
...
+	header, err := core.GetBlockHeaderByNumber(txn, blockNumber)
+	if err != nil {
 		return err
 	}
 
-	header, err := core.GetBlockHeaderByNumber(txn, blockNumber)
-	if err != nil {
+	state := core.NewDeprecatedState(txn)
+	// revert state
+	if err = state.Revert(blockNumber, stateUpdate, header.ProtocolVersion); err != nil {
```
rodrodros: Similarly here

core/state_interfaces.go
```diff
@@ -20,13 +20,15 @@ type State interface {
 		update *StateUpdate,
 		declaredClasses map[felt.Felt]ClassDefinition,
 		skipVerifyNewRoot bool,
+		protocolVersion string,
 	) error
 	// Revert rolls back the state update for blockNum, restoring the state to update.OldRoot.
 	// Requires the current root to match update.NewRoot.
-	Revert(blockNum uint64, update *StateUpdate) error
+	Revert(blockNum uint64, update *StateUpdate, protocolVersion string) error
```
rodrodros: This function and the method that implements it

core/state.go
```diff
@@ -607,8 +615,10 @@ func (s *DeprecatedState) ContractDeployedAt(addr *felt.Felt, blockNumber uint64
 	return deployedAt <= blockNumber, nil
 }
 
-func (s *DeprecatedState) Revert(blockNumber uint64, update *StateUpdate) error {
-	err := s.verifyStateUpdateRoot(update.NewRoot)
+func (s *DeprecatedState) Revert(
+	blockNumber uint64, update *StateUpdate, protocolVersion string,
+) error {
+	err := s.verifyStateUpdateRoot(update.NewRoot, protocolVersion)
```
rodrodros, line 624: here

core/state/state_test.go
```diff
@@ -931,18 +931,54 @@ func setupState(t *testing.T, stateUpdates []*core.StateUpdate, blocks uint64) *
...
+	require.NoError(t, err)
+	assert.False(t, pre014.IsZero())
+
+	// v0.14+: same state but Poseidon is always applied
+	v014, err := state.Commitment("0.14.2")
+	require.NoError(t, err)
+
+	assert.NotEqual(t, pre014, v014,
+		"pre-0.14 commitment (raw contractRoot) must differ from v0.14 (Poseidon)")
+}
```
rodrodros, line 1061: nitpick: could you change the style to 
(if line <= 100 chars)
All args in one separate line
```go
	assert.NotEqual(
            t, pre014, v014, "pre-0.14 commitment (raw contractRoot) must differ from v0.14 (Poseidon)",
    )
```
(if line > 100 chars)
One line per arg
```golang
	assert.NotEqual(
            t, 
            pre014, 
            v014, 
            "pre-0.14 commitment (raw contractRoot) must differ from v0.14 (Poseidon)",
    )
```

core/state/state_test.go
```diff
@@ -931,18 +931,54 @@ func setupState(t *testing.T, stateUpdates []*core.StateUpdate, blocks uint64) *
...
+func TestStateCommitmentPre014ReturnsContractRoot(t *testing.T) {
+	// Pre-0.14 with no declared classes (classRoot == 0): commitment must equal
+	// the raw contractRoot, not a Poseidon hash.
+	stateDB := newTestStateDB()
+	state, err := New(&felt.Zero, stateDB)
+	require.NoError(t, err)
+
+	storageKey := new(felt.Felt).SetUint64(0)
+	storageVal := new(felt.Felt).SetUint64(0x80)
+	contractAddr := new(felt.Felt).SetUint64(2)
```
rodrodros: Could you use either of the dedicated felt constructors

```golang
// create a felt in the heap and get a ptr (copies by ref)
var1 := felt.NewFromUint64[felt.Felt](123) // creates a *felt.Felt

// create the felt in the stack (copies by value)
var2 := felt.FromUint64[felt.Felt](123) // creates a felt.Felt
```

blockchain/blockchain.go
m-kus: Done

Using *Header (not *Block) because:
 - revertHead only has a *Header, not a full *Block
 - Header has both Number and ProtocolVersion - the only fields needed
 - Block embeds *Header, so block.Header works at call sites that have a block

PR #3447, opened 2026-03-03
fix(state): new state, 0.14.1 casm hash migration support

EgeCaner:
This PR adds casm hash migration support to new state.

Casm metadata writes were split: the blockchain layer wrote Declarations and Migrations, while reorgs caused the state layer to write and delete casm metadata. This PR consolidates all casm metadata writes in the blockchain layer.

### DIFF
--- a/blockchain/blockchain_test.go
+++ b/blockchain/blockchain_test.go
@@ -822,6 +822,7 @@ func TestRevertHeadMigratedCasmClasses(t *testing.T) {
 	chain := blockchain.New(testDB, &utils.Integration)
 
 	receipts0 := make([]*core.TransactionReceipt, 0)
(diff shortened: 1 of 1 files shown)


core/state/state.go
```diff
@@ -196,7 +196,12 @@ func (s *State) Update(
 	}
 
 	// Update the class trie
-	if err := s.updateClassTrie(update.StateDiff.DeclaredV1Classes, declaredClasses); err != nil {
+	err := s.updateClassTrie(
+		update.StateDiff.DeclaredV1Classes,
+		declaredClasses,
+		update.StateDiff.MigratedClasses,
+	)
+	if err != nil {
```
rodrodros, line 204: 👏

PR #3450, opened 2026-03-04
feat(rpc): make v9 self contained

thiagodeev:
Make RPC v9 self-contained, so that it no longer depends on code from v6 pkg.
Same as PR #3439, but for v9.

Note: this PR also removed the `"github.com/crate-crypto/go-ipa/bandersnatch/fp"` module from our code.
While I was copying the code to v9, I noticed there was this library being used only there, while the rest of the code uses the code from Consensys. This mod doesn't even have a stable release and has only a few stars on GitHub. So I decided to remove it.


rpc/v9/chain_test.go
```diff
@@ -0,0 +1,31 @@
+package rpcv9_test
+
+import (
+	"testing"
+
+	"github.com/NethermindEth/juno/mocks"
+	rpc "github.com/NethermindEth/juno/rpc/v6"
```
MaksymMalicki: is this intended?

rpc/v6/simulation.go
```diff
@@ -13,7 +13,7 @@ import (
 	rpccore "github.com/NethermindEth/juno/rpc/rpccore"
 	"github.com/NethermindEth/juno/utils"
 	"github.com/NethermindEth/juno/vm"
-	"github.com/crate-crypto/go-ipa/bandersnatch/fp"
+	"github.com/consensys/gnark-crypto/ecc/stark-curve/fp"
```
rodrodros, line 16: why making v9 self contained implies making changes on v6?

rpc/v9/chain_test.go
thiagodeev: For sure not, thanks!
I was periodically searching for the 'v6' keyword in the code filter, but it seems I forgot to run after copying this method from v6. I'll fix it

rpc/v6/simulation.go
thiagodeev: Sorry, I forgot to put in the PR description!
This came from a compiler error while I was copying the code to v9. I noticed there was this weird library being used only there, while the rest of the code uses the code from Consensys.
This mod doesn't even have a stable release and has only a few stars on GitHub. So I decided to remove it from our code.

Do you want me to revert this change?

rodrodros: Nop, this is good 👍

rpc/v9/transaction.go
rodrodros, line 314: Typo outside of this PR but also existent in v10 and v8 (can you please fix it as a follow up). It should be `Ecdsa`(Elliptic Curve Digital Signature Algorithm)

PR #3454, opened 2026-03-04
fix: fix flaky tests in sync and tracker packages

rodrodros:
## Summary

- **fix flaky `TestSubscribePending` broken pipe error**: prevents writing to a closed connection during test cleanup
- **fix flaky `TestWaitGroupFailure` race condition**: delayed goroutines could read `isShutdown` before `run()` set it to `true`, causing the `waitGroupWrapper` subtest to not panic as expected. Adds a `shutdownComplete` barrier so delayed goroutines deterministically wait for `run()` to finish setting `isShutdown`
- **refactor tracker tests**: minor readability improvements (reorder variable declarations, add section comments)

🤖 Generated with [Claude Code](https://claude.com/claude-code) and a bit of human spice

### DIFF
--- a/utils/tracker/tracker_test.go
+++ b/utils/tracker/tracker_test.go
@@ -71,12 +71,10 @@ func runTest(
 	t.Helper()
 
 	ctx, cancel := context.WithCancel(t.Context())
-	success := atomic.Uint32{}
 	service := NewService(tracker(ctx), nonDelayed, delayed)
 
-	clientWg := conc.NewWaitGroup()
+	// Start server go routines
 	serverWg := conc.NewWaitGroup()
-
 	serverWg.Go(func() {
 		service.run(ctx)
 	})
@@ -89,6 +87,10 @@ func runTest(
 		})
 	}
 
+	// Start client go routines
+	clientWg := conc.NewWaitGroup()
+
+	success := atomic.Uint32{}
 	for isDelayed, count := range map[bool]int{false: nonDelayed, true: delayed} {
 		for range count {
 			clientWg.Go(func() {


clients/feeder/test_feeder.go
```diff
@@ -76,8 +76,10 @@ func newTestServer(t testing.TB, network *utils.Network) *httptest.Server {
 			handleNotFound(dir, queryArg, w)
 			return
 		}
-		_, err = w.Write(read) //nolint:gosec // G705: no danger, test environment
-		require.NoError(t, err, "failed to write response")
+		// Write may fail with "broken pipe" if the client disconnects before the
+		// response is fully written (e.g. context cancellation during polling).
+		// This is expected, so we intentionally ignore the error.
+		w.Write(read) //nolint:errcheck,gosec // client may disconnect before response is fully written
```
brbrr: Assuming the call could return `BrokenPipe` or similar error (identifying that client dropped the connection), maybe we could still check for the other errors?

clients/feeder/test_feeder.go
```diff
@@ -76,8 +77,14 @@ func newTestServer(t testing.TB, network *utils.Network) *httptest.Server {
...
-		_, err = w.Write(read) //nolint:gosec // G705: no danger, test environment
-		require.NoError(t, err, "failed to write response")
+		//nolint:gosec // looks like a false positive from gosec.
+		_, err = w.Write(read)
+		// Write may fail with "broken pipe" if the client disconnects before the
+		// response is fully written (e.g. context cancellation during polling).
+		// This is expected, so we intentionally ignore the error.
+		if err != nil && !errors.Is(err, syscall.EPIPE) {
+			t.Errorf("failed to write response: %v", err)
+		}
```
thiagodeev, line 88: As I mentioned on Slack, I think the new solution will not solve the problem at all since it only checks for the `io.ErrClosedPipe`. Look this CI, it fails due to the same flakiness, but with a different error message
https://github.com/NethermindEth/juno/actions/runs/22680794304/job/65750390711

With the help of Claude, I propose this solution:

```suggestion

    _, err = w.Write(read) //nolint:gosec // G705: no danger, test environment
    if err != nil && r.Context().Err() == nil {
        t.Errorf("failed to write response: %v", err)
    }
```

I think it'll work. The docs about `r.Context()` say ([ref](https://pkg.go.dev/net/http#Request.Context)):

> "For incoming server requests, the context is canceled when the client's connection closes, the request is canceled (with HTTP/2), or when the ServeHTTP method returns."

clients/feeder/test_feeder.go
rodrodros: Yes, added an error for it, but if a lot of erros start to appear, then it is probably better to ignore them all

clients/feeder/test_feeder.go
rodrodros: Good catch!

PR #3534, opened 2026-04-08
refactor(db): reorganize bucket definitions with inner/outer type separation

thiagodeev:
### Summary

  - Introduces a private `innerBucket` enum that owns the iota-defined byte values, and a public `Bucket` wrapper type that references them. This separates the stable byte layout (which must never change) from the public API consumers interact with.
  - Reorganizes `Bucket` constants into domain groups (Block, Class, Contract, Mempool, Migration, State) with section headers.
  - Re-exports the enumer-generated methods (`String`, `BucketString`, `BucketValues`, `BucketStrings`, `IsABucket`) through the `Bucket` type so the public API remains unchanged.

  ### Why

  The previous flat `iota` list mixed unrelated buckets together, making it hard to find or add new entries. Because iota values are persisted in the database, reordering them for readability was never an option. The two-layer approach lets us organize the public constants freely while the `innerBucket` enum locks down the actual byte values.

  ### Validation

  Snapshot tests comparing every `Bucket` constant against its `innerBucket` counterpart were written to verify that all values, keys, string representations, and wrapper methods are correct (the file is not part of the PR code since it's not supposed to be merged, only to verify the accuracy; it is only here for reference):
https://github.com/NethermindEth/juno/blob/3698bbcd3c03eb2fd036403fb9838ec1137a8e7a/db/buckets_value_test.go


thiagodeev:
Codecov is complaining about code coverage, but there was almost no change in the code logic; we are only mirroring it now through new variables.
Tests were made in [this ](https://github.com/NethermindEth/juno/blob/3698bbcd3c03eb2fd036403fb9838ec1137a8e7a/db/buckets_value_test.go)file, but I thought it was not relevant to include it in the code to be merged since they're testing specifically the current variables, not the future ones.

cmd/juno/dbcmd.go
```diff
@@ -222,6 +222,7 @@ func dbSize(cmd *cobra.Command, args []string) error {
 	buckets := db.BucketValues()
 	items := make([][]string, 0, len(buckets)+3)
 	for _, b := range buckets {
+		b := db.Bucket(b)
```
rodrodros: I believe the change here is changing what `db.BucketValues()` return instead of doing the casting. It's not a very good approach to return an internal type and then cast it to the external one

db/buckets.go
```diff
@@ -2,64 +2,209 @@ package db
...
+
+// BucketStrings returns a slice of all String values of the enum
+func BucketStrings() []string {
+	return innerBucketStrings()
+}
+
+// IsABucket returns "true" if the value is listed in the enum definition. "false" otherwise
+func (b Bucket) IsABucket() bool {
+	return innerBucket(b).IsAinnerBucket()
+}
```
rodrodros: Why? Where is this function called?

db/buckets.go
```diff
@@ -2,64 +2,209 @@ package db
...
+// Throws an error if the param is not part of the enum.
+func BucketString(s string) (Bucket, error) {
+	ib, err := innerBucketString(s)
+	return Bucket(ib), err
+}
+
+// BucketValues returns all values of the enum
+func BucketValues() []innerBucket {
+	return innerBucketValues()
+}
```
rodrodros: This has to return the public buckets

db/buckets.go
thiagodeev: Nowhere, I just wanted to keep exporting all previous functions/methods generated by "enumer". Do you want me to keep only the ones that are used?

cmd/juno/dbcmd.go
thiagodeev: The code works normally without this casting, with the exception of lines 240 and 248 that call the `utils.AnyOf(...)` generic function  ('b' would be of type `innerBucket` while the other variables are `Bucket`, so we have a compiler error). I put the cast on the top just for the sake of simplicity; otherwise, I would put the cast on each one of the `utils.AnyOf(...)` calls. Using the internal type doesn't break anything in the code, since the variable value is the same, and the exported methods can be normally called as well.

Based on that, and the fact that this function is used only here, I decided to just return the inner value. However, the fact that it's only called here is also a good point to make the `db.BucketValues()` function return the exported type, since the overhead of creating a new slice only for that will not affect the performance of the node, only a few nanoseconds on this db tool call. Do you want me to make it return a `[]Bucket` slice instead? (related to your other [comment](https://github.com/NethermindEth/juno/pull/3534#discussion_r3052267044))

db/buckets.go
rodrodros: I think you can answer this yourself. Does this function have any use? Is this function going to be used in the future? 

If the answers are "No" and "I don't know", then those are strong indicators that this function shouldn't exists just because. This is not the first time we have a similar conversation, please let's stop having these issues.

cmd/juno/dbcmd.go
rodrodros: > The code works normally without this casting,...

That's not the point. That you can enter a house though a window doesn't mean that's what windows are for or it is not a well designed house. 

The point is that you are creating a private type hiding under a public type, and then you're completing ignoring it and using the inner type. 

Isn't the purpose of the outer type specifically to protect the rest of the packages against it's internal changes. Isn't that the purpose of having a public API? Your public types hide whatever the inner types hold are and will be – and you're breaking that assumption in this exact same PR. If the goal was just organizing buckets and grouping for the sake of readability an `.md` file would've done as well

db/buckets.go
```diff
@@ -2,64 +2,197 @@ package db
 
 import "slices"
 
-//go:generate go run github.com/dmarkham/enumer -type=Bucket -output=buckets_enumer.go
-type Bucket byte
-
 // Pebble does not support buckets to differentiate between groups of
 // keys like Bolt or MDBX does. We use a global prefix list as a poor
 // man's bucket alternative.
+type Bucket byte
```
EgeCaner: maybe
```go
type Bucket innerBucket
```

db/buckets.go
```diff
@@ -2,64 +2,197 @@ package db
...
 func (b Bucket) Key(key ...[]byte) []byte {
 	return append([]byte{byte(b)}, slices.Concat(key...)...)
 }
+
+func (b Bucket) String() string {
+	return innerBucket(b).String()
+}
+
+// BucketValues returns all values of the enum.
+func BucketValues() []Bucket {
```
EgeCaner, line 140: Should this function return all innerBucket values or buckets exposed by the public API?

thiagodeev: I think I understand what you mean.

It seems there's no automatic way to list only the exported buckets; returning all innerBuckets is the easy way.
We could implement a logic in the `BucketValues()` to filter out buckets starting with "unused" or "deprecated", then we would only need to make sure all used external and inner buckets share the same name. But, I believe it's relevant to leave it as is, since it's good to know whether we forgot to clean a deprecated bucket or not using the dbtool 😁

rodrodros (approved):

PR #3535, opened 2026-04-08
refactor(blockchain): replace method chaining with functional options

EgeCaner:
Method-chaining setters stay on the public API, so a fully constructed Blockchain can be reconfigured at any time. When options depend on each other, a setter must cascade updates to everything that was already configured based on a now-stale value. Each setter ends up needing to know about every other option. 

Functional options collect configuration before construction; the returned value has no setters, blockchain is immutable after construction.

### DIFF
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -81,7 +81,38 @@ type Blockchain struct {
 	transactionLayout core.TransactionLayout
 }
 
-func New(database db.KeyValueStore, network *utils.Network) *Blockchain {
+// options holds configuration for constructing a Blockchain.
+type options struct {
+	listener          EventListener
+	transactionLayout core.TransactionLayout
+}
+
+// Option is a functional option for configuring Blockchain options.
+type Option func(*options)
+
+// WithTransactionLayout sets the transaction storage layout.
+func WithTransactionLayout(layout core.TransactionLayout) Option {
+	return func(o *options) {
+		o.transactionLayout = layout
+	}
+}
+
+// WithListener sets the event listener for the blockchain.
+func WithListener(listener EventListener) Option {
+	return func(o *options) {
+		o.listener = listener
+	}
+}
+
+func New(database db.KeyValueStore, network *utils.Network, opts ...Option) *Blockchain {
+	o := options{
+		listener:          &SelectiveListener{},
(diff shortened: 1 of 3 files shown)


blockchain/blockchain.go
```diff
@@ -81,7 +81,38 @@ type Blockchain struct {
...
+}
+
+func New(database db.KeyValueStore, network *utils.Network, opts ...Option) *Blockchain {
+	o := options{
+		listener:          &SelectiveListener{},
+		transactionLayout: core.TransactionLayoutCombined,
+	}
+	for _, opt := range opts {
+		opt(&o)
+	}
```
rodrodros, line 114: 👍

rodrodros (approved):
Looks good

node/node.go
```diff
@@ -194,7 +194,11 @@ func New(cfg *Config, version string, logLevel *utils.LogLevel) (*Node, error) {
 	services := make([]service.Service, 0)
 	earlyServices := make([]service.Service, 0)
 
-	chain := blockchain.New(database, &cfg.Network)
+	var opts []blockchain.Option
```
rodrodros: You know beforehand the capacity of `opts` ;) 
Let's build it like that

node/node.go
```diff
@@ -194,7 +194,11 @@ func New(cfg *Config, version string, logLevel *utils.LogLevel) (*Node, error) {
...
+	if cfg.Metrics {
+		opts = append(opts, blockchain.WithListener(makeBlockchainMetrics()))
+	}
```
rodrodros, line 200: This complicates the worflow a bit, now there are 2 `if cfg.Metric`. Not sure if there is anything to do about it, it's also another drop in a function already overfilled with this.

The solution is probably refactoring this monster (in the future)

rodrodros (approved):
Looks good

node/node.go
rodrodros: The benefits it's that it get's allocated in a stack based slice and no heap allocation. Also, easier to read

node/node.go
EgeCaner: I agree, I don't like it either. Yet I don't see a straightforward way to avoid bigger refactor

thiagodeev (approved):

PR #3536, opened 2026-04-08
refactor: blockchain state access

EgeCaner:
Introduces StateBackend interface in new blockchain/statebackend package to abstract all state-mutating and state-reading operations. Currently ships deprecatedStateBackend (existing IndexedBatch-based state); the new trie2-based backend can be added as a second implementation behind the same interface.        
                                                                                                                                                                                   
Motivation      
                                                                                                                                                                                   
We are introducing a new trie2-based state backend alongside the existing deprecated one. Having all state logic inlined in blockchain.go with scattered if useNewState branches would be unmaintainable. The StateBackend interface gives each implementation its own file and lets us swap/test them independently, while Blockchain stays a thin coordinator.


blockchain/statebackend/block_ops.go
```diff
@@ -0,0 +1,200 @@
...
+	} else if !errors.Is(err, db.ErrKeyNotFound) {
+		return err
+	}
+
+	if expectedBlockNumber != block.Number {
+		return fmt.Errorf("expected block #%d, got block #%d", expectedBlockNumber, block.Number)
+	}
+	if !block.ParentHash.Equal(expectedParentHash) {
+		return ErrParentDoesNotMatchHead
+	}
```
rodrodros, line 45: Error key not found means we are syncing from the beginning, hence there is no information?

If that's the case, then block should also be number 0?

blockchain/statebackend/block_ops.go
```diff
@@ -0,0 +1,200 @@
...
+	block.Hash = &blockHash
+	stateUpdate.BlockHash = &blockHash
+	return commitments, nil
+}
+
+// signBlock applies the signature to the block if a signing function is provided.
+func signBlock(block *core.Block, stateUpdate *core.StateUpdate, sign utils.BlockSignFunc) error {
+	if sign == nil {
+		return nil
+	}
```
rodrodros: I don't agree with this defensive behaviour, if `sign` is nil what's the purpose of calling this function whatsoever. It makes the no op invisible to readers. I believe the guard, if any, should be explicitly written before this function gets called.

blockchain/statebackend/casm_metadata.go
```diff
@@ -0,0 +1,150 @@
...
+}
+
+// storeCasmHashMetadataV1 stores metadata for classes declared with V1 hash (protocol < 0.14.1).
+// It computes the V2 hash from the class definition.
+func storeCasmHashMetadataV1(
+	writer db.KeyValueWriter,
+	blockNumber uint64,
+	stateUpdate *core.StateUpdate,
+	newClasses map[felt.Felt]core.ClassDefinition,
+) error {
```
rodrodros, line 89: Do we still need this code? Assuming Juno only supports 0.14.2?

blockchain/statebackend/deprecated.go
```diff
@@ -0,0 +1,254 @@
...
+	"github.com/NethermindEth/juno/core"
+	"github.com/NethermindEth/juno/core/felt"
+	"github.com/NethermindEth/juno/db"
+	"github.com/NethermindEth/juno/db/memory"
+	"github.com/NethermindEth/juno/utils"
+)
+
+type deprecatedStateBackend struct {
+	baseState
+}
```
rodrodros, line 15:
```
type deprecatedStateBackend baseState
```
?

rodrodros (commented):
Is the PR just moving code around? Or is making any modifications?

blockchain/statebackend/casm_metadata.go
EgeCaner: Yes, we need it. It is called while we are syncing old blocks

blockchain/statebackend/block_ops.go
EgeCaner: Yes, it should be 0. `expectedBlockNumber` and `expectedParentHash` is initialized as 0 and felt.Zero respectively. If we get db.ErrKeyNotFound, they will remain with zero values

blockchain/statebackend/block_ops.go
EgeCaner: I agree, let me remove it and do the checks at call site

blockchain/statebackend/deprecated.go
EgeCaner: I'm not sure which one is better here. 

baseState does not have methods so embedding is not for method carry-over. 
IMO difference is, embedding says "deprecatedState has baseState" vs type definition says "deprecatedStateBackend is shaped like baseState". Which one would you recommend?

EgeCaner:
> Is the PR just moving code around? Or is making any modifications?

@rodrodros This PR shouldn't make any behavioral changes. Things were organized a bit differently to prepare the ground to support new state. Changes are mostly throwing an interface where old state and new state are incompatible :) Incompatibility comes from old state using indexed batch and new state using batch. Additionally, this PR extracts some generic logic into helpers which will soon be shared with the new state backend, such as `writeBlockContent` or `deleteBlockContent`. Also, some methods which were implemented by blockchain before have been turned into functions. One notable change is that indexed batch usage in `transactionLayout` has been replaced with batch and reader (this change originates from this PR #3187)

blockchain/statebackend/deprecated.go
rodrodros: I have no strong opinion on either, why not use baseState directly?

rodrodros (approved):
Looks good!

MaksymMalicki (approved):

blockchain/statebackend/deprecated.go
EgeCaner: In this case, I guess it is more of a styling choice, both ways should be identical. Also new state backend will look like following

```go
type newStateBackend struct {
	baseState
	stateDB *state.StateDB
}
```

it will make it consistent with each other.

rodrodros (approved):

PR #3538, opened 2026-04-09
fix(jsonrpc): Gracefully handle panics from handler calls

brbrr:
The server could panic when the underlying handler function panics. The original panic happened due to the type mismatch in the v10 JSON validator handler. We were using v9 handler (and v10 validator), and some of the underlying types are different. 

This PR adds graceful handling for panics in the request handling code on the server level. It does not feel ideal, as the actual issue is related to the lack of type-safety in our RPC handler definitions (meaning we could mix up the handlers, and it won't be detected unless we do a runtime check).

### DIFF
--- a/jsonrpc/server.go
+++ b/jsonrpc/server.go
@@ -457,19 +457,17 @@ func isNilOrEmpty(i any) (bool, error) {
 	}
 }
 
-// TODO: add recover() to catch panics from handlers/validators and return a JSON-RPC internal error
-// instead of crashing the HTTP connection
-func (s *Server) handleRequest(ctx context.Context, req *Request) (*response, http.Header, error) {
+func (s *Server) handleRequest(ctx context.Context, req *Request) (res *response, header http.Header, err error) {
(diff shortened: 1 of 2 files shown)


jsonrpc/server.go
```diff
@@ -457,41 +457,60 @@ func isNilOrEmpty(i any) (bool, error) {
...
+	res = &response{
 		Version: "2.0",
 		ID:      req.ID,
 	}
 
 	calledMethod, found := s.methods[req.Method]
 	if !found {
 		res.Error = Err(MethodNotFound, nil)
-		s.log.Trace("Method not found in request", zap.String("method", req.Method))
+		s.log.Trace("Method not found in request", zap.String("method", sanitizeForLog(req.Method)))
```
brbrr, line 485: Not sure if that's really necessary, but with sanitization, we could reduce log pollution. Potentially, we could also strip the string if it's too long, but I'm not sure if that's worthwhile.

PR #3539, opened 2026-04-09
refactor: remove old layout code (part 1)

thiagodeev:
This is part 1 of removing `TransactionLayout` from `core/`. This PR removes all imports/mentions of `TransactionLayout` outside of the `blockchain` type. After this, a follow-up PR will be made to delete  `core.TransactionLayout`.

## Summary
  - Extracts `core.TransactionLayout` into a new frozen package `migration/blocktransactions/txlayout` — a complete copy of the type, all methods, and tests — so migration code can read/write both per-tx and combined layouts independently of `core`
  - Removes the `WithTransactionLayout()` option available for the `blockchain.New()` constructor.
  - Updates all migration and test callers (`deprecatedmigration`, `blocktransactions`) to use `txlayout` instead of `core.TransactionLayout`          
  - Introduces `testStateBackend` in `deprecatedmigration/state_test.go` to replace `blockchain.New(...WithTransactionLayout(core.TransactionLayoutPerTx))` in tests, decoupling deprecated migration tests from the `blockchain` package (since it uses the new layout by default now, and we need the old layout to test the deprecated migrations)


migration/blocktransactions/txlayout/transaction_layout_test.go
thiagodeev, line 1: This is a full copy of the `core/transaction_layout_test.go` file.

migration/blocktransactions/txlayout/transaction_layout.go
thiagodeev, line 1: This is a full copy of the `core/transaction_layout.go` file. The only relevant difference is the new pkg level documentation.

deprecatedmigration/state_test.go
thiagodeev, line 1: As the code comments say, this is just a copy of the `statebackend.deprecatedStateBackend` type and its `Store(...)` method.

EgeCaner (approved):
Looks good to me, thanks!

PR #3541, opened 2026-04-10
feat(rpc): make v10 self contained (part 2)

thiagodeev:
Make RPC v10 self-contained, so that it no longer depends on code from v6/v9 pkgs.
It finishes the work started in PR https://github.com/NethermindEth/juno/pull/3439 by adding the missing methods on v10 (v10 handler was importing them from v6 and v9 since they didn't exist in v10)


rodrodros (approved):
LGTM!

PR #3551, opened 2026-04-14
refactor: remove old layout code (part 2)

thiagodeev:
This is part 2 of removing `TransactionLayout` from `core/`. It completely removes the `TransactionLayout` type, creates new db accessors for the corresponding methods, and uses them in the code instead of  `TransactionLayout`.

## Summary                                                                                                                                                               
                                                                            
  - Completely removes the `TransactionLayout` type and its associated methods (`TransactionByBlockAndIndex`, `TransactionByHash`, `BlockByNumber`, `WriteTransactionsAndReceipts`, `DeleteTxsAndReceipts`)
  - Reintroduces standalone accessor functions in `core/accessors.go` (e.g., `GetTransactionByBlockAndIndex`, `GetTransactionByHash`, `GetBlockByNumber`, `GetReceiptsByBlockNumber`, etc.) just like we had before ([ref](https://github.com/NethermindEth/juno/pull/3304/changes#diff-06febc522fadd3c796358ff24a417b074bec629e96d3ccd718bdf081ea9cd1bf)), but using the new combined layout
  - Removes the `TransactionLayout` logic from `blockchain.Blockchain`, `blockchain.EventFilter`, and `statebackend`, replacing all usages with the new accessors


core/accessors.go
rodrodros, line 1: Should there be tests for these methods as well?

core/accessors_test.go
```diff
@@ -0,0 +1,123 @@
...
+	"github.com/NethermindEth/juno/core/felt"
+	"github.com/NethermindEth/juno/db"
+	"github.com/NethermindEth/juno/db/memory"
+	adaptfeeder "github.com/NethermindEth/juno/starknetdata/feeder"
+	"github.com/NethermindEth/juno/utils"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestTransactionsAndReceipts(t *testing.T) {
```
EgeCaner: Can we divide tests into multiple tests or sub-tests rather than comments seperating sections?

rodrodros: Yes, tests should be modular. @thiagodeev please see other test files in Juno on how to approach this

EgeCaner (approved):

PR #3559, opened 2026-04-15
Reapply "refactor: deprecate HeapPtr use new instead"

rodrodros:
This reverts commit 3318620ef50b9c92dacf41073088a6b6c7238541 which was itself a revert from 03171e0e9e30fc1e497fe76d701335d36458d0b0

Meaning, we are re-applying 03171e0e9e30fc1e497fe76d701335d36458d0b0 because I accidentaly pushed to main. I've removed those permissions now


rodrodros:
No longer necessary

PR #3561, opened 2026-04-16
refactor(core): new `deprecatedstate` pkg

thiagodeev:
This PR moves the `DeprecatedState` code spread in the `core` pkg into a new `core/deprecatedstate` pkg.

Rationale behind this decision:
- state-specific code is spread under the `core` pkg, making it difficult to differentiate between it and `core` general code;
- an isolated pkg for the deprecated state code will help in the future removal of it;
- this follows the same pattern of the new state: a dedicated pkg for it.

Changes made:
- copy/pasted of the deprecated state code from `core` to `core/deprecatedState`;
- Renamed `DeprecatedState` and `deprecatedStateHistory` to `State` and `stateHistory` since now the pkg prefix gives the context (e.g: `deprecatedstate.NewState` instead of `deprecatedstate.NewDeprecatedState`;


vm/vm_test.go
```diff
@@ -108,7 +109,7 @@ func TestCallDeprecatedCairoMaxSteps(t *testing.T) {
 	simpleClass, err := gw.Class(t.Context(), classHash)
 	require.NoError(t, err)
 
-	testState := core.NewDeprecatedState(txn)
+	testState := deprecatedstate.NewState(txn)
```
rodrodros: Because `NewState` is now part of `deprecatedstate` package, it can be called just `deprecatedstate.New`

core/contract.go
```diff
@@ -84,112 +26,13 @@ func ContractAddress(
...
-			return err
-		}
-	}
-
-	return nil
-}
-
 // ContractNonce returns the amount transactions sent from this contract.
 // Only account contracts can have a non-zero nonce.
 func ContractNonce(addr *felt.Felt, txn db.KeyValueReader) (felt.Felt, error) {
```
EgeCaner: AFAIK ContractClassHash, ContractNonce and ContractStorage buckets are used by old state, new state stores them inside stateObject, If im not mistaken we can move their accessors to deprecated state as wel

core/contract.go
rodrodros, line 1: Does this files still makes sense under Core?

rodrodros (approved):
looks good

blockchain/statebackend/deprecated.go
```diff
@@ -61,8 +62,8 @@ func (b *deprecatedStateBackend) StateAtBlockNumber(
 		return nil, nil, err
 	}
 
-	return core.NewDeprecatedStateHistory(
-		core.NewDeprecatedState(txn),
+	return deprecatedstate.NewStateHistory(
```
EgeCaner: Nit: maybe `NewHistory` after `NewState` -> `New`

EgeCaner (approved):
Dropped one comment, rest looks good to me

core/contract.go
thiagodeev: Look at this comment from @EgeCaner.
Indeed, I can move `ContractClassHash` and `ContractNonce` functions to `deprecatedstate` as well.
We would have then only the `ContractAddress` function, which is used in multiple pkgs outside of the `core` pkg. Do you have a better idea of where to place it?

EgeCaner (approved):

core/contract.go
rodrodros: Definitely not a sole function for a single file, so let's find an adoptive file that is better for it. Let me know if you cannot find any good one

PR #3563, opened 2026-04-17
refactor(state): split state into `State` and `StateReader`

MaksymMalicki:
This PR finishes the already started split of the new `State`. We already have the proper constructors `New` and `NewStateReader`, however both produced the same `State` object with and without `Batch`. Even though, the caller was safe from wrongly utilising the constructors by the existing common state interfaces, technically if the interfaces were not used, the caller could call state mutation methods on the `StateReader`.

In this PR, the `State` is decoupled into `StateReader` - which only holds the read methods, and `State` which extends the `StateReader` with the state mutation methods. The diff is quite big due to methods reordering - the `State` declaration, constructor and methods remain in `state.go`, `StateReader` declaration, constructor and methods were moved to the `state_reader.go`. `stateHistory` uses the `StateReader` as the read layer now.

The only modification worth paying attention to is the `ContractStorage` method, which exists both in the `State` and `StateReader`. Previously, the `ContractStorage` was first reading from the dirty cache on the state object, later falling back to the DB. But this scenario was effectively a dead code - the only time the dirty buffer is filled is during the sync (`state.Update` called from the `blockchain.Store()`). We don't share the state objects, whenever we wanted read, we created a brand new `State` object using the `blockchain.HeadState()` or `blockchain.StateAtBlockNumber` - with an empty buffer. Now, that we split the `State` into the `State` and `StateReader`, we can drop this logic and make the `StateReader` read directly from the storage trie in the DB.


core/state/state.go
```diff
@@ -715,15 +542,255 @@ func (s *State) getStateObject(addr *felt.Felt) (*stateObject, error) {
...
+type StateReader struct {
+	initRoot     felt.Felt
+	db           *StateDB
+	contractTrie *trie2.Trie
+	classTrie    *trie2.Trie
+}
+
+// NewStateReader creates a read-only view of the state at the given root.
+// Should be used for read operations that don't require state mutations.
+func NewStateReader(stateRoot *felt.Felt, db *StateDB) (*StateReader, error) {
```
rodrodros: Define State Reader on it's own file.

It's contradictory that state_update which has its own file and are just 4 to 5 lines of code (and State Readers doesn't). I think state_update should go and be defined inside core/state/state.go

core/state/state.go
rodrodros, line 262: Can you explain why do we have this logic:
1. Call the function
2. Check if the obj is nil
3. Return nil

Instead of:
1. Check if the obj is not nil
2. Then call the function

Reducing the cost having another go routine working

core/state/state.go
```diff
@@ -55,152 +55,30 @@ func New(stateRoot *felt.Felt, db *StateDB, batch db.Batch) (*State, error) {
...
-	}
-	return contract.Nonce, nil
-}
-
+// ContractStorage on State checks dirty storage cache before falling back to the trie.
 func (s *State) ContractStorage(addr, key *felt.Felt) (felt.Felt, error) {
 	obj, err := s.getStateObject(addr)
 	if err != nil {
 		return felt.Felt{}, err
 	}
```
rodrodros: I am looking at this function call path but I don't see at which this fall back to the trie?

core/state/state_reader.go
rodrodros, line 1: remember to add as well the tests for the state reader

rodrodros (commented):
Why are the folders inside the state called "statesmth" instead of just "smth". Example:
1. `statefactory` instead of `factory`
2. `statetestutils` instead of `testutils`

core/state/state_reader.go
MaksymMalicki: yes yes, it was uncommited yet, I have a few changes comming, changed to draft

core/state/state.go
MaksymMalicki: It was split in the last commit

core/state/state.go
```diff
@@ -55,152 +52,20 @@ func New(stateRoot *felt.Felt, db *StateDB, batch db.Batch) (*State, error) {
...
-
-func (s *State) ContractNonce(addr *felt.Felt) (felt.Felt, error) {
-	contract, err := GetContract(s.db.disk, addr)
-	if err != nil {
-		return felt.Felt{}, err
-	}
-	return contract.Nonce, nil
-}
-
-func (s *State) ContractStorage(addr, key *felt.Felt) (felt.Felt, error) {
```
MaksymMalicki, line 116: This method was moved to the `StateReader` in a different form. Instead of reading the dirty storage and falling back to the trie, it immediately reaches the storage trie for this value.
REASON: The dirty cache is only populated during the `Update` - which is writing. State instances are not shared, previously for all the read operations a new `State` object instance was generated using `blockchain.HeadState` or `blockchain.StateAtBlockHash`. In Juno, we never read the state, that is currently being written to the DB

core/state/state.go
```diff
@@ -357,65 +222,11 @@ func (s *State) Revert(header *core.Header, update *core.StateUpdate) error {
 	if !newComm.Equal(update.OldRoot) {
 		return fmt.Errorf("state commitment mismatch: %v (expected) != %v (actual)", update.OldRoot, &newComm)
 	}
-	if s.batch != nil {
```
MaksymMalicki, line 360: Those non-nil guards are not needed anymore, `State` always enforces a non-nil batch

core/state/state.go
MaksymMalicki: No difference, good idea 👍

core/state/object.go
```diff
@@ -40,40 +38,15 @@ func (s *stateObject) setNonce(nonce *felt.Felt) {
 	s.contract.Nonce = *nonce
 }
 
-func (s *stateObject) getStorage(key *felt.Felt) (felt.Felt, error) {
```
MaksymMalicki, line 43: This method was only used in the `(s *State) ContractStorage`. Removal reason explained [here](https://github.com/NethermindEth/juno/pull/3563/changes#r3099590805)

EgeCaner (approved):

rodrodros (approved):

PR #3564, opened 2026-04-17
refactor(core): new `pending` pkg

thiagodeev:
This PR moves the pending code in the core pkg into a new core/pending pkg.

Changes made:

- copy/pasted of the pending code from `core` to `core/pending`;
- renamed `PendingState` and `PendingStateWriter` to `State` and `StateWriter` since now the pkg prefix gives the context (e.g: `pending.NewState` instead of `pending.NewPendingState`;


EgeCaner (approved):
Looks good

PR #3565, opened 2026-04-17
perf(core/felt): reduce unneeded heap allocations

brbrr:
note: the original implementation was failing `no quotes` test - it was successfully unmarshalling non-quoted hex strings, such as `0xdeadbeef`. I think it make sense to reject them as, JSON context, all of the felt values should be quoted strings

### DIFF
--- a/core/felt/felt.go
+++ b/core/felt/felt.go
@@ -70,7 +70,7 @@ func (z *Felt) UnmarshalJSON(data []byte) error {
 	}
 	copy(padded[maxHexDigits-len(src):], src)
 
-	var buf [32]byte
+	var buf [Bytes]byte
 	if _, err := hex.Decode(buf[:], padded[:]); err != nil {
 		return fmt.Errorf("felt: couldn't decode hex value: %w", err)
 	}
--- a/core/felt/felt.go
+++ b/core/felt/felt.go
@@ -82,16 +82,16 @@ func (z *Felt) UnmarshalJSON(data []byte) error {
 // unnecessary leading zeros. Uses a value receiver so encoding/json
 // can call it on non-addressable values (struct fields in `any`).
 func (z Felt) MarshalJSON() ([]byte, error) {
-	var raw [32]byte
+	var raw [Bytes]byte
 	fp.BigEndian.PutElement(&raw, fp.Element(z))
 
 	// Find first significant byte.
 	i := 0
-	for i < 31 && raw[i] == 0 {
+	for i < Bytes-1 && raw[i] == 0 {
 		i++
 	}
 
-	out := make([]byte, 3, 4+(32-i)*2)
+	out := make([]byte, 3, 4+(Bytes-i)*2)
 	out[0], out[1], out[2] = '"', '0', 'x'
 
 	// First byte may need a single hex digit (e.g. 0x3, not 0x03).
--- a/core/felt/felt.go
+++ b/core/felt/felt.go
@@ -14,9 +14,6 @@ import (
 const (
 	Base16 = 16
 	Base10 = 10
-
-	// maxHexDigits is the maximum number of hex digits in a felt value (32 bytes = 64 hex chars).
-	maxHexDigits = Bytes * 2
 )
 
 const (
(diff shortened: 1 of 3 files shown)


core/felt/felt.go
```diff
@@ -46,27 +49,59 @@ func (z *Felt) Impl() *fp.Element {
...
 
-	// remove leading and trailing quotes if any, normalise to lowercase
-	s := strings.ToLower(strings.Trim(string(data), `"`))
-	if !strings.HasPrefix(s, "0x") {
-		return errors.New("felt value must be a 0x-prefixed hex string")
+	// Left-pad with '0' to 64 hex chars so hex.Decode produces exactly 32 bytes.
+	var padded [64]byte
+	for i := range padded {
+		padded[i] = '0'
 	}
```
rodrodros, line 71: Does this makes sense? Aren't padded[i] `0` by default?

brbrr: yes, by default it's filled with zeros, but in this case, we fill it with `'0'` which is ASCII value of 0. So we end up with a string "000..", not. null bytes: https://www.ascii-code.com/48

rodrodros:
# `Felt` JSON marshaling benchmarks 

Before/after comparison of `Felt.MarshalJSON` and `Felt.UnmarshalJSON`.

- **Host**: Apple M2 Pro · `darwin/arm64` · Go 1.26.1
- **Command**: `go test -run=^$ -bench='BenchmarkMarshalJSON|BenchmarkUnmarshalJSON' -benchmem -count=10 ./core/felt/...`
- **old** = `main` at merge-base (`4924bc81`) 

## Latency (sec/op)

| Benchmark | old | new | Δ |
|---|---:|---:|---:|
| `MarshalJSON/zero` | 48.53 ns | 28.59 ns | **−41.09%** |
| `MarshalJSON/small` | 62.81 ns | 32.22 ns | **−48.71%** |
| `MarshalJSON/address` | 141.30 ns | 46.29 ns | **−67.24%** |
| `MarshalJSON/max_felt` | 141.80 ns | 46.89 ns | **−66.94%** |
| `MarshalJSON/padded_address` | 141.70 ns | 46.56 ns | **−67.15%** |
| `MarshalJSON/random` | 141.65 ns | 47.08 ns | **−66.76%** |
| `UnmarshalJSON/zero` | 85.66 ns | 63.04 ns | **−26.41%** |
| `UnmarshalJSON/small` | 131.90 ns | 62.21 ns | **−52.84%** |
| `UnmarshalJSON/address` | 438.65 ns | 62.65 ns | **−85.72%** |
| `UnmarshalJSON/max_felt` | 445.75 ns | 62.93 ns | **−85.88%** |
| `UnmarshalJSON/padded_address` | 438.10 ns | 62.86 ns | **−85.65%** |
| `UnmarshalJSON/random` | 442.90 ns | 62.95 ns | **−85.79%** |
| **geomean** | **168.5 ns** | **50.39 ns** | **−70.10%** |

All deltas are statistically significant (`p = 0.000`, `n = 10`).

## Memory (B/op)

| Benchmark | old | new | Δ |
|---|---:|---:|---:|
| `MarshalJSON/zero` | 8 B | 8 B | — |
| `MarshalJSON/small` | 24 B | 16 B | **−33.33%** |
| `MarshalJSON/address` | 288 B | 80 B | **−72.22%** |
| `MarshalJSON/max_felt` | 288 B | 80 B | **−72.22%** |
| `MarshalJSON/padded_address` | 288 B | 80 B | **−72.22%** |
| `MarshalJSON/random` | 288 B | 80 B | **−72.22%** |
| `UnmarshalJSON/zero` | 37 B | 0 B | **−100.00%** |
| `UnmarshalJSON/small` | 48 B | 0 B | **−100.00%** |
| `UnmarshalJSON/address` | 112 B | 0 B | **−100.00%** |
| `UnmarshalJSON/max_felt` | 112 B | 0 B | **−100.00%** |
| `UnmarshalJSON/padded_address` | 112 B | 0 B | **−100.00%** |
| `UnmarshalJSON/random` | 112 B | 0 B | **−100.00%** |

## Allocations (allocs/op)

| Benchmark | old | new | Δ |
|---|---:|---:|---:|
| `MarshalJSON/zero` | 1 | 1 | — |
| `MarshalJSON/small` | 2 | 1 | **−50%** |
| `MarshalJSON/address` | 4 | 1 | **−75%** |
| `MarshalJSON/max_felt` | 4 | 1 | **−75%** |
| `MarshalJSON/padded_address` | 4 | 1 | **−75%** |
| `MarshalJSON/random` | 4 | 1 | **−75%** |
| `UnmarshalJSON/zero` | 2 | 0 | **−100%** |
| `UnmarshalJSON/small` | 2 | 0 | **−100%** |
| `UnmarshalJSON/address` | 2 | 0 | **−100%** |
| `UnmarshalJSON/max_felt` | 2 | 0 | **−100%** |
| `UnmarshalJSON/padded_address` | 2 | 0 | **−100%** |
| `UnmarshalJSON/random` | 2 | 0 | **−100%** |

The remaining `1 alloc` on `MarshalJSON` is the returned `[]byte` — unavoidable since the method contract is to return a slice.

## Takeaways

- **`UnmarshalJSON` is now fully zero-alloc on every input shape.** The old implementation allocated twice per call — once for `strings.ToLower` / `strings.Trim`, once for the pooled `big.Int` used by `SetString`.
- **`MarshalJSON` drops from 4 → 1 alloc and ~66% faster on full-width inputs.** The single remaining alloc is the return slice; it's also 3.6× smaller (80 B vs 288 B) because the new code sizes it precisely instead of going through a `z.String()` intermediate.
- Geomean latency across all 12 sub-benchmarks improved by **−70.10%**; the UnmarshalJSON group is where the biggest wins are (−86% on any realistic address-sized input).

rodrodros (approved):
Looks good to me ;)

PR #3566, opened 2026-04-17
refactor: clean up `utils` pkg

thiagodeev:
Related to https://github.com/NethermindEth/juno/issues/1198
A refactor in the 'log' and 'network' files will be made in follow-up PRs.

  Cleanup of the `utils/` package: move domain types closer to their semantic homes, replace small wrappers with stdlib equivalents, and remove dead code.

  ### Relocated
  - `utils.DataSize` + byte-size constants (`Kilobyte`, `Megabyte`, …) → `db/`
  - `utils.Base64` → `core/transaction.go`. Dropped redundant `MarshalJSON`/`UnmarshalJSON` methods — they were functionally identical to the default encoding/json behavior for named string types. Also dropped the `RegisterCustomTypeFunc(Base64)` blocks in the v8/v9/v10 validators: the `validate:"base64"` tag applies natively to named string types (in v8/v9 the type wasn't even used — doubly dead code).
  - `BlockSignFunc` → `core/block.go`; `Sign` → `sequencer/` as unexported `newBlockSigner`. The sequencer is the only creator; other consumers (builder, blockchain, statebackend) only need the type.
  - `NonNilSlice` / `DerefSlice` consolidated into `utils/slices.go`.

  ### Replaced with stdlib / language builtins
  - `utils.RunAndWrapOnError(fn, err)` → `errors.Join(err, fn())` at 28 call sites. `errors.Join` is strictly better: it preserves both errors for `errors.Is`/`errors.As` (the old impl used `%v` on the run error, making it unreachable via `Unwrap`). Here are some tests comparing both: [ref](https://github.com/NethermindEth/juno/pull/3566/changes/5384bab0d624ccebc9a9abb94734676c0d410abe).
  - `utils.HeapPtr(x)` → Go 1.26's `new(x)` (now accepts expressions).
  - The mapping-style use `utils.Map(s, utils.HeapPtr[T])` → new `utils.ToPtrSlice(s)` helper (7 call sites in RPC trace adapters).
  - `utils.ToHex(*big.Int)` → `fmt.Sprintf("0x%x", …)`.
  - Dropped `github.com/pkg/errors` dependency in favor of stdlib `errors.Join`.

  ### Deleted
  - `utils/bigint.go`, `utils/const.go`, `utils/zero_value.go`,`utils/maps.go` (unused `ToMap`/`ToSlice`), and the now-empty `utils/wrap.go`, `utils/sign.go`, `utils/base64.go`.

  ### Incidental bug fix (sequencer block signer) - Me & my friend Claude Code
  While relocating `Sign`, it was noticed that the signature-chunking loop sliced `signatureBytes[start:step]` instead of `[start:start+step]`. For a 64-byte stark-curve ECDSA signature (`felt.Bytes = 32`), the second iteration read `signatureBytes[32:32]` — an empty slice — so the `S` component was silently dropped and replaced with `felt(0)`. Every block signature produced by the sequencer was malformed and would fail cryptographic verification.
The loop was replaced with a direct two-felt construction (signature size is fixed at the algorithm level) and the test was strengthened to reconstruct the signature and round-trip it through `PublicKey.Verify`.


utils/slices.go
```diff
@@ -78,6 +64,34 @@ func Set[T comparable](slice []T) []T {
...
+
+	return sl
+}
+
+func DerefSlice[T any](v *[]T) []T {
+	if v == nil {
+		return nil
+	}
+	return *v
+}
```
rodrodros, line 93: Maybe tests for this?

rodrodros (approved):
LGTM

Issue #2175, opened 2024-09-24
Introduce new `TransactionWithSignature` interface


kirugan:
Right now all transactions implement `Signature()` method, but we don't need it for L1Handler and Deploy transactions (right now they return empty slice). 
We already faced some issues with it, [here](https://github.com/NethermindEth/juno/pull/2153/files#) is an example:
```
if txSignature := transaction.Signature(); len(txSignature) > 0 {
   digest.Update(txSignature...)
} else {
   digest.Update(&felt.Zero)
}
```
it was rewritten as:
```
switch transaction.(type) {
  case *DeployTransaction, *L1HandlerTransaction:
     digest.Update(&felt.Zero)
  default:
    digest.Update(transaction.Signature()...)
}
```
but code above may lead to more questions: why we're specifying exactly these transactions? is it specific to this function only or applicable to entire codebase?

It can be solved by introducing new interface type `TransactionWithSignature` (name only for example, I'm pretty sure there is a better naming):
```
type TransactionWithSignature interface {
   Signature() []*felt.Felt
}
```
which can be used in example above as:
```
switch transaction.(type) {
case TransactionWithSignature:
     digest.Update(transaction.Signature()...)
default:
     digest.Update(&felt.Zero)
}
```

pri-3x:
@kirugan PR for this issue is opened: https://github.com/NethermindEth/juno/pull/2189
The workflows are awaiting approval to run. Could you please review and approve them?

PR #3571, opened 2026-04-22
refactor(core): model signed transactions separately from generic tra…

ongyimeng:
## Issue
https://github.com/NethermindEth/juno/issues/2175


ongyimeng:
outdated/ not relevant

PR #3572, opened 2026-04-22
feat(blockchain): add `stateBackend` for the new state

MaksymMalicki:


EgeCaner (approved):
Looks good to me, thank you!

blockchain/statebackend/statebackend.go
```diff
@@ -0,0 +1,237 @@
+package statebackend
```
MaksymMalicki, line 1: Test coverage for this will be fixed, when a flag to support the new state in the tests will be merged. For now this is fully isolated

blockchain/statebackend/statebackend.go
```diff
@@ -0,0 +1,237 @@
...
+func (b *stateBackend) HeadState() (core.StateReader, StateCloser, error) {
+	height, err := core.GetChainHeight(b.database)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	header, err := core.GetBlockHeaderByNumber(b.database, height)
+	if err != nil {
+		return nil, nil, err
+	}
```
rodrodros, line 24: Is it safe to access the DB directly here, shouldn't we use a snapshot?

rodrodros (commented):
Only one question

blockchain/statebackend/statebackend.go
MaksymMalicki: It makes sense to use a snapshot when we make for example 2 requests for a head state (nonces, tries etc.) and we need a consistent view to the DB. The problem is that the head values might change between the two reads if some other service updates the DB. Here, we make only the first read to the height (which is dynamic) and the rest of the reads is tied to the height that we read.

rodrodros (approved):

PR #3575, opened 2026-04-22
fix(state): unify the errors with the old state

MaksymMalicki:


core/state/accessors.go
```diff
@@ -19,21 +18,10 @@ func GetStateObject(r db.KeyValueReader, state *State, addr *felt.Felt) (*stateO
 	return &obj, nil
 }
 
-// Gets a contract instance from the database. If it doesn't exist returns a contract not deployed error
-func GetContract(r db.KeyValueReader, addr *felt.Felt) (stateContract, error) {
```
MaksymMalicki, line 23: because the old state returns the `db.KeyNotFound` in the contract data reads, we also want to return it in the new state. This wrapper makes no sense anymore

rodrodros (approved):

PR #3576, opened 2026-04-22
feat(core): checker to detect which state implementation is being used

MaksymMalicki:
This PR introduces a compatibility checker.
When the user tries to run Juno using the new state on the snapshot built using the old state and vice-versa, Juno will terminate on start-up an let the user know with the proper error message.
Currently, this is a standalone checker, which will be integrated into the pipeline in #3187

### DIFF
--- a/core/state_version.go
+++ b/core/state_version.go
@@ -7,6 +7,13 @@ import (
 	"github.com/NethermindEth/juno/db"
 )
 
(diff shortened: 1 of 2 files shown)


rodrodros:
We will use a Schema Version to check for the New State vs the Old State

PR #3579, opened 2026-04-24
chore: remove extra that in comment

purelualight:
remove extra that in comment


rodrodros:
PRs that fixes a single thing are not currently being accepted, please make it more general so that they fix many typos everywhere

PR #3582, opened 2026-04-24
refactor(rpc/v10): rework AddTransaction, decouple EstimateMessageFee, tighten validation

thiagodeev:
TODO:  apply the changes for other rpc pkgs as well

  ### Validation

  - Enabled struct-level validation via `validator.WithRequiredStructEnabled`. Lets us drop pointer indirection on `ResourceBoundsMap` / `ResourceBounds` (value types now) and rely on `validate` tags directly on the fields.
  - Removed redundant nil/shape checks in the adapters — JSON-schema + struct validation already cover them.
                                                                                                                                                                      
  ### `EstimateMessageFee` decoupled from `EstimateFee`

  - Previously, an L1 handler transaction was wrapped inside a `BroadcastedTransaction` so it could reuse the `EstimateFee` plumbing. That workaround is gone — `EstimateMessageFee` now has its own direct path with no `BroadcastedTransaction` round-trip.                  

  ### `AddTransaction` refactor

  - Multiple speed improvements (benchmarks below)
  - Split the monolithic `AdaptBroadcastedTransaction` into smaller, single-purpose functions.
  - Cleaner and faster feeder gateway payload generation.
  - Introduced a dedicated `ContractClass` type for the `contract_class` field instead of `json.RawMessage`. Two wins:
    - the field is now actually validated (previously it wasn't);
    - downstream code works with a typed value instead of raw JSON.
  - New direct adapter `BroadcastedTransaction → core.Transaction`, removing the intermediate hop through `starknet.Transaction`.

  #### Bug fix (v9; v10 unaffected)

  - For `DECLARE` transactions, `pushToFeederGateway` was mutating the original `tx.ContractClass`. By the time the tx reached `receivedTransactionFeed`, the mutated state caused the send to fail and the tx was never published.
  - Fixed in v9 to unblock benchmarks. v8 still carries the bug — left as-is intentionally: the full v10 changes will be ported to v8/v9 once the team approves the changes.                                                                               

### Benchmark result:
  
  Config: count=5, benchtime=3s, Intel i5-11400H @ 2.70GHz, Go 1.26.

  | Benchmark | old sec/op | new sec/op | Speedup | old B/op | new B/op | old allocs/op | new allocs/op |
  |---|---:|---:|---:|---:|---:|---:|---:|
  | Mempool/declare (stub) | 59.7 ms | 22.7 ms | 2.63x | 2.16 MiB | 235.6 KiB | 33,505 | 5,508 |
  | Mempool/invoke | 187.0 us | 92.6 us | 2.02x | 2.59 KiB | 1.06 KiB | 51 | 21 |
  | Mempool/deploy_account | 401.7 us | 241.5 us | 1.66x | 2.62 KiB | 1.08 KiB | 50 | 21 |
  | Gateway/declare (stub) | 43.8 ms | 28.4 ms | 1.54x | 3.95 MiB | 1.22 MiB | 28,142 | 11,029 |
  | Gateway/invoke | 102.1 us | 103.1 us | ~1x | 3.99 KiB | 3.72 KiB | 73 | 65 |
  | Gateway/deploy_account | 170.3 us | 166.1 us | ~1x | 3.75 KiB | 3.43 KiB | 69 | 60 |
  | Declare_RealCompiler_Mempool | 235.1 ms | 110.4 ms | 2.13x | 8.20 MiB | 3.17 MiB | 70,935 | 24,223 |
  | Declare_RealCompiler_Gateway | 134.6 ms | 113.4 ms | 1.19x | 6.98 MiB | 4.70 MiB | 46,857 | 29,757 |
     
  - Declare on the mempool path is 2.1-2.6x faster: new compiles Sierra once instead of twice and reuses the class hash for the feed broadcast.
  - Gateway/declare gains come from the streaming `ContractClassToGatewayPayload` (pooled `gzip.Writer`, `gzip.BestSpeed`, streamed json -> gzip -> base64 pipeline).
  - Non-declare gateway cases are unchanged: no `ContractClass` work involved.
  - `Declare_RealCompiler_Gateway` gain is modest because FFI Sierra->CASM compile dominates wall-clock time.
  - All new sub-benchmarks are equal or better than old. No regressions.


genesis/genesis.go
```diff
@@ -344,6 +344,7 @@ func executeTransactions(
 				DeployTransaction: core.DeployTransaction{
 					TransactionHash:     txn.Hash,
 					ContractAddressSalt: txn.ContractAddressSalt,
+					// bug: TxnDeployAccount does not contain SenderAddress field, so this is nil
```
rodrodros, line 347: add issue please 🙏

rpc/v10/adapt_transaction.go
```diff
@@ -68,28 +70,28 @@ func AdaptCoreTransaction(t core.Transaction) *Transaction {
...
 			MaxAmount:       felt.NewFromUint64[felt.Felt](rb[core.ResourceL1DataGas].MaxAmount),
 			MaxPricePerUnit: rb[core.ResourceL1DataGas].MaxPricePerUnit,
 		}
 	} else {
-		l1DataGasResourceBounds = &ResourceBounds{
+		l1DataGasResourceBounds = ResourceBounds{
 			MaxAmount:       &felt.Zero,
 			MaxPricePerUnit: &felt.Zero,
 		}
 	}
```
rodrodros: alternatively to avoid the `if` `else` block:

```go
    l1DataGasResourceBounds := ResourceBounds{
			MaxAmount:       &felt.Zero,
			MaxPricePerUnit: &felt.Zero,
	}
	if _, ok := rb[core.ResourceL1DataGas]; ok {
		l1DataGasResourceBounds = ResourceBounds{
			MaxAmount:       felt.NewFromUint64[felt.Felt](rb[core.ResourceL1DataGas].MaxAmount),
			MaxPricePerUnit: rb[core.ResourceL1DataGas].MaxPricePerUnit,
		}
	}
```

rpc/v10/adapt_transaction.go
```diff
@@ -425,3 +411,147 @@ func MakeJSONErrorFromGatewayError(err error) *jsonrpc.Error {
...
+		MaxAmount:       rb.L2Gas.MaxAmount.Uint64(),
+		MaxPricePerUnit: rb.L2Gas.MaxPricePerUnit,
+	}
+	coreResourceBounds[core.ResourceL1DataGas] = core.ResourceBounds{
+		MaxAmount:       rb.L1DataGas.MaxAmount.Uint64(),
+		MaxPricePerUnit: rb.L1DataGas.MaxPricePerUnit,
+	}
+
+	return coreResourceBounds
+}
```
rodrodros, line 431: Does this map has fixed keys, meaning, during it's entire lifetime has the same keys here defined?

If that's the case please add an issue to make it a struct

thiagodeev: Yes and no. AFAIK, currently, there are no more `core.Resource`s other than these three listed, but in the past, v3 txs contained only two. Since this is a core type, it was stored in the db when storing txs, so we need to be careful when changing it.

I'll create the issue

genesis/genesis.go
thiagodeev: Done
https://github.com/NethermindEth/juno/issues/3741

rpc/v10/adapt_transaction.go
thiagodeev: Done
https://github.com/NethermindEth/juno/issues/3742

thiagodeev:
@claude you're wrong. The gateway branch only calls `adaptAndCompileBroadcastedTxToCore` once, which is after sending the tx to the feeder. Before that, the code doesn't compile anything.
Read the code and check it

rpc/v10/adapt_transaction_test.go
```diff
@@ -0,0 +1,141 @@
...
+		t.Parallel()
+		broadcasted, _ := loadTx(t, sepoliaInvokeTxnPath)
+
+		first, err := rpc.AdaptBroadcastedTransactionToCore(t.Context(), broadcasted, nil, network)
+		require.NoError(t, err)
+		second, err := rpc.AdaptBroadcastedTransactionToCore(t.Context(), broadcasted, nil, network)
+		require.NoError(t, err)
+
+		assert.Equal(t, first.Hash(), second.Hash())
+	})
```
rodrodros: What's the purpose of this test?

Why this test should check for determinism in hash, it is calling two pure functions?

rpc/v10/addtxn_bench_test.go
rodrodros, line 1: rename file to `addtxn_bench_test`

rpc/v10/transaction.go
```diff
@@ -225,56 +173,29 @@ func (h *Handler) addToMempool(
...
-		}
-
-		gwSierraProg, errIn := utils.Gzip64Encode(sierraProgBytes)
-		if errIn != nil {
-			return AddTxResponse{}, jsonrpc.Err(jsonrpc.InternalError, errIn.Error())
-		}
+	classPayload, err := ContractClassToGatewayPayload(tx.ContractClass)
+	if err != nil {
+		return AddTxResponse{}, rpccore.ErrInternal.CloneWithData(
+			fmt.Sprintf("failed to get contract class payload: %v", err),
```
rodrodros: Use %w to pass the err

rodrodros (approved):

rpc/v10/adapt_transaction_test.go
thiagodeev: The `AdaptBroadcastedTransactionToCore` also generates the tx hash of a tx while adapting it to a core.Tx.
This test is just a safeguard that the hash calculation is consistent. IMO, there's no harm in having it.
Do you want me to remove it?

rpc/v10/transaction.go
thiagodeev: But this is not a Go error; the value goes directly to the `data` field in the json response. A plain string is the correct value here

rpc/v10/adapt_transaction_test.go
rodrodros: Why would the hash the first time vs the second time?

rpc/v10/transaction.go
rodrodros: Then it should be %s

rpc/v10/adapt_transaction_test.go
thiagodeev: The test is there exactly to prevent this 🙂. But I get it, I'll remove it

PR #3583, opened 2026-04-24
fix(rpcv10): recompute trace if initial_reads requested

brbrr:
There is an ordering issue with RPC trace cache for block transactions - we cache only by block hash, so we could end up with a situation where the cache was populated with trace without initial reads, and when the caller request same tries but with `initial_reads`, we'd just return cached response (with missing `initial_reads`)


EgeCaner (approved):
Looks good to me, thank you!

MaksymMalicki (approved):

PR #3584, opened 2026-04-27
refactor(core): abstract hash calculation engine

MaksymMalicki:
Both tries calculate the exact same hash. This PR ties the hash calculation engine on receipts, txs and blocks to the used state version - if Juno runs using legacy state, legacy trie is used for hash calculation. If Juno runs using new state, the new `trie2` implementation is used.

### DIFF
--- a/blockchain/statebackend/types.go
+++ b/blockchain/statebackend/types.go
@@ -36,7 +36,6 @@ type StateBackend interface {
 	) error
 	VerifyBlockHash(
 		b *core.Block,
-		network *networks.Network,
 		stateDiff *core.StateDiff,
 	) (*core.BlockCommitments, error)
 }
--- a/blockchain/statebackend/types.go
+++ b/blockchain/statebackend/types.go
@@ -34,6 +34,11 @@ type StateBackend interface {
 		newClasses map[felt.Felt]core.ClassDefinition,
 		sign core.BlockSignFunc,
 	) error
+	VerifyBlockHash(
+		b *core.Block,
+		network *networks.Network,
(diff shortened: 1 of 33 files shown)


core/receipt.go
```diff
@@ -115,7 +111,7 @@ func calculateCommitment[T any](
 
 		for i, res := range results {
 			key := felt.FromUint64[felt.Felt](uint64(i))
-			if _, err := trie.Put(&key, &res); err != nil {
+			if err := trie.Update(&key, &res); err != nil {
```
MaksymMalicki, line 114: safe, because `trie.Update` wraps the `trie.Put`.
[ref](https://github.com/NethermindEth/juno/blob/4310ba55d38028d708e8e986e5d8d4c5ac3b3851/core/trie/trie.go#L530)

blockchain/statebackend/types.go
```diff
@@ -34,6 +34,11 @@ type StateBackend interface {
 		newClasses map[felt.Felt]core.ClassDefinition,
 		sign core.BlockSignFunc,
 	) error
+	VerifyBlockHash(
+		b *core.Block,
+		network *networks.Network,
```
EgeCaner: Im not sure if we need to receive network as a param, this information is available in baseState, Finalize and Simulate already uses from there, maybe we can also remove it from the interface here, WDYT?

EgeCaner (approved):
Dropped one comment, rest look good to me!

blockchain/statebackend/types.go
MaksymMalicki: good catch

PR #3588, opened 2026-04-29
refactor: Use sonic for JSON un/marshaling

brbrr:
This PR replaces `encoding/json` with `sonic` for JSON un/marshaling.

Initially, I was planning to add it only for the RPCv10, but to reduce the complexity/conditioning, I decided to migrate all RPC versions. 

( ! ) Note: It introduces breaking changes - the error messages (that user may encounter) are differs from what `encoding/json` were producing, also, they are different between ARM64 and AMD64 architectures.

Changes:

  - Introduce utils/jsonx as a centralized marshal/unmarshal/decoder wrapper around sonic.
  - Migrate many call sites from encoding/json to jsonx across VM, RPC versions, clients, compiler, and core hashing.
  - Rework `LimitSlice` decoding to use sonic’s lazy AST iteration.

TODO: Pretouch logging needs to be removed before merging the PR.


brbrr:
here's the benchmarking data: 

**tl;dr: sonic uses less CPU (~−12%) at the same throughput, holds more memory (~+30%), and shows higher tail latency (p95 +9%). Median and p90 are flat to slightly better.**



| Metric | sonic  | main  | delta |
  |---|---|---|---|
  | http_req_duration avg | 63.1 ms | 62.8 ms | +0.5% |
  | http_req_duration median | 50.4 ms | 51.3 ms | −1.7% |
  | http_req_duration p90 | 90.2 ms | 92.4 ms | −2.4% |
  | http_req_duration p95 | **116.3 ms** | **106.6 ms** | **+9.1%** |
  | http_req_duration max | 2.11 s | 720 ms | +193% |
  | iteration_duration avg | 85.1 ms | 84.5 ms | +0.7% |
  | iteration_duration p95 | 169.6 ms | 159.8 ms | +6.1% |
  | **%CPU mean** | **116%** | **133%** | **−12.5%** |
  | %CPU p95 | 135% | 149% | −9.4% |
  | **RAM mean** | **910 MiB** | **696 MiB** | **+30.7%** |
  | RAM drift / run | +123 MiB | +73 MiB | +69% |

jsonrpc/server_test.go
```diff
@@ -313,7 +403,9 @@ func TestHandle(t *testing.T) {
 					"params" : { "num" : 5 }}],
 					[{"jsonrpc" : "2.0", "method" : "method",
 					"params" : { "num" : 44 }}]]`,
-			res: `[{"jsonrpc":"2.0","error":{"code":-32600,"message":"Invalid Request","data":"json: cannot unmarshal array into Go value of type jsonrpc.Request"},"id":null},{"jsonrpc":"2.0","error":{"code":-32600,"message":"Invalid Request","data":"json: cannot unmarshal array into Go value of type jsonrpc.Request"},"id":null}]`,
+			//nolint:lll // We can't break the json value
+			res:             `[{"jsonrpc":"2.0","error":{"code":-32600,"message":"Invalid Request","data":"placeholder"},"id":null},{"jsonrpc":"2.0","error":{"code":-32600,"message":"Invalid Request","data":"placeholder"},"id":null}]`,
```
rodrodros, line 408: Why not?

jsonrpc/server_test.go
```diff
@@ -520,7 +624,11 @@ func TestHandle(t *testing.T) {
 
 			// tests that have an empty response cannot be parsed into a json
 			if test.res != "" {
-				assert.JSONEq(t, test.res, string(res))
+				if test.ignoreErrorData {
+					assertResponseIgnoringErrorData(t, test.res, string(res))
```
rodrodros, line 629: This function seems unnecessary, why aren't the JSONs populated with the right error data?

jsonrpc/server_test.go
```diff
@@ -786,6 +894,55 @@ func TestRequest_MarshalLogObject(t *testing.T) {
...
+// TestPretouchProductionHandlers registers Juno's real RPC methods
+// (v8, v9, v10) on jsonrpc.Server instances and dumps the resulting
+// PretouchedTypes list. The list is the actual set of Go types that
+// sonic compiled encode/decode paths for at server startup, with no
+// placeholder/stub handlers.
+//
+// Network and storage interfaces (bcReader, syncReader, vm) are passed
+// nil — those are only touched when handlers run, not at registration.
+func TestPretouchProductionHandlers(t *testing.T) {
+	logger := log.NewNopZapLogger()
```
rodrodros, line 907: This tests looks out of place, it should be in the RPC package

rpc/v10/subscription_types.go
```diff
@@ -8,9 +8,14 @@ import (
...
-	Version string `json:"jsonrpc"`
-	Method  string `json:"method"`
-	Params  any    `json:"params"`
+	Version string             `json:"jsonrpc"`
+	Method  string             `json:"method"`
+	Params  SubscriptionParams `json:"params"`
+}
+
+type SubscriptionParams struct {
+	Result         any            `json:"result"`
```
rodrodros, line 17: Why Result can be Any?

Also define the sub type above the parent type

brbrr: That's how it's used currently - we pass various types as a `Result`. Potentially, we could use a generic type here, but I believe it's out of scope for this PR. Will move the type above the parent.

jsonrpc/server_test.go
brbrr: tl:dr; AMD and ARM builds produce different error outputs because they use different parsers under the hood, so we do exact quality checks in our tests, they will fail in CI on different archs. 

The idea is captured in the function doc comment:
https://github.com/NethermindEth/juno/pull/3588/changes#diff-a2e8c524803b22cc1bb50a1992a011374f1a29e432d04b54c984cb5ec9c7396dR24-R30

brbrr:
Some bench data before/after f1e08fe5ef78221db3c7350570925b3c64d5161d  **(local arm64 run)**
  
  | Metric | Δ |
  |---|---|
  | **ns/op** | **−11.02%** |
  | **allocs/op** | **−7.85%** |
  | **B/op** | **+9.60%** |

  The bytes regression is `ast.Node`'s fixed parse-tree overhead — paid once per request regardless of element count. Multi-param requests amortize it; single-param requests pay full price. Empirically tested two alternatives (`[]sonic.NoCopyRawMessage` and bigger `bufio` buffer) — both regressed. `ast.Node` turns out to be the cheap option.

  ### sec/op per axis

  | Axis                   | base (µ) | after (µ) |       Δ |
  | ---------------------- | -------: | --------: | ------: |
  | `no_params`            |    1.377 |     1.145 | −16.85% |
  | `no_params_ctx`        |    1.586 |     1.386 |       ~ |
  | `one_scalar_pos`       |    1.850 |     1.927 |       ~ |
  | `one_struct_pos`       |    2.376 |     2.012 | −15.32% |
  | `three_pos_mixed`      |    3.222 |     2.610 | −18.99% |
  | `three_named_required` |    3.530 |     3.005 | −14.87% |
  | `named_opt_missing`    |    2.137 |     2.272 |       ~ |
  | `named_opt_present`    |    3.372 |     2.978 | −11.68% |
  | `validator_pass`       |    2.383 |     2.107 | −11.58% |
  | `validator_fail`       |    2.926 |     2.643 |  −9.67% |
  | `type_error`           |    3.123 |     2.957 |       ~ |
  | `batch_small` (5)      |    15.04 |    10.66 | −29.13% |
  | `batch_large` (50)     |   105.40 |    95.00 |  −9.87% |
  | `response_100_structs` |    12.63 |    12.29 |       ~ |

  ### allocs/op per axis

  | Axis                   | base | after |       Δ |
  | ---------------------- | ---: | ----: | ------: |
  | `one_scalar_pos`       |   29 |    28 |  −3.45% |
  | `one_struct_pos`       |   32 |    28 | −12.50% |
  | `three_pos_mixed`      |   46 |    35 | −23.91% |
  | `three_named_required` |   46 |    36 | −21.74% |
  | `named_opt_missing`    |   30 |    29 |  −3.33% |
  | `named_opt_present`    |   46 |    36 | −21.74% |
  | `validator_pass`       |   33 |    29 | −12.12% |
  | `validator_fail`       |   45 |    41 |  −8.89% |
  | `type_error`           |   43 |    42 |  −2.33% |

  (no-params, batch, response axes unchanged at 18 / 117 / 888 / 18)



[cpu.out.zip](https://github.com/user-attachments/files/27528954/cpu.out.zip)
Here's the cpu profile data for the new bench tests added from **amd64 arch**. (had to archive the above file to upload it)

to access it, use: `go tool pprof -http=:8080 cpu.out`

Preview for one of the tests:

Screenshot 2026-05-08 at 17 24 18

brbrr:
Allocations

main:
Screenshot 2026-05-08 at 22 58 52

sonic:
Screenshot 2026-05-08 at 22 58 30

And here's the GC metrics data (as reported by node) before and after running the tests:

```
main     requests=96094  cpu=  gc=0.0113134s  alloc=1858906912 bytes
         per-req:  cpu=0  alloc=19344.7 bytes  gc-time=1.17733e-07
sonic    requests=96108  cpu=  gc=0.00684841s  alloc=1689997264 bytes
         per-req:  cpu=0  alloc=17584.4 bytes  gc-time=7.12574e-08
```

brbrr:
CPU data from AMD64 arch running end-to-end tests

tl;dr: Un/Marshaling took 11% vs 7% overal cpu time (main vs sonic) or 5.5s vs 3.4s CPU time, so ~70% decrease. 

main:
Screenshot 2026-05-08 at 23 08 42

sonic:
Screenshot 2026-05-08 at 23 08 57

rodrodros:
Clsoing since we won't use it for the time being

PR #3594, opened 2026-05-01
feat: catchup l1 head on start

EgeCaner:
Previously, the L1 client only subscribed to live `LogStateUpdate` events on startup. After downtime, this left Juno running with a stale L1 head until the next on-chain update was emitted, which can take hours on Starknet.                                                   

This PR catches up the L1 head on start by issuing a backward `eth_getLogs` scan over the core contract before subscribing. The catch-up is best-effort: free-tier RPC providers might cap range size or rate-limit, so on failure we log a warning and fall through to the live subscription rather than terminating.
                                                                                                                                                                                
  Changes         

  - Added `catchUpL1HeadUpdates` running once before `watchL1StateUpdates` in `Client.Run`
  - Added `eth_getLogs` query for the core contract's `LogStateUpdate`
  - Refactored Client construction to use functional options

### DIFF
--- a/l1/l1.go
+++ b/l1/l1.go
@@ -24,9 +24,13 @@ type Subscriber interface {
 	FinalisedHeight(ctx context.Context) (uint64, error)
 	LatestHeight(ctx context.Context) (uint64, error)
 	WatchLogStateUpdate(ctx context.Context, sink chan<- *contract.StarknetLogStateUpdate) (event.Subscription, error)
+	FilterLogStateUpdate(
+		ctx context.Context,
+		fromBlock,
+		toBlock uint64,
+	) ([]*contract.StarknetLogStateUpdate, error)
 	ChainID(ctx context.Context) (*big.Int, error)
 	TransactionReceipt(ctx context.Context, txHash common.Hash) (*types.Receipt, error)
-
 	Close()
 }
 
--- a/l1/l1.go
+++ b/l1/l1.go
@@ -2,6 +2,7 @@ package l1
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"math/big"
 	"time"
@@ -153,6 +154,21 @@ func (c *Client) Run(ctx context.Context) error {
 		return err
 	}
 
+	// catchUpL1HeadUpdates is best-effort: a backward eth_getLogs scan can fail
+	// on free-tier RPC providers that cap range size or rate-limit. On failure
+	// we skip catch-up and proceed straight to the live subscription — the L1
+	// head will lag until the next on-chain LogStateUpdate is observed, which
+	// is acceptable rather terminating the execution.
+	if err := c.catchUpL1HeadUpdates(ctx); err != nil {
+		c.logger.Warn("L1 head catch-up failed; resuming with live subscription only",
+			zap.Error(err),
+		)
+	}
+
+	return c.watchL1StateUpdates(ctx)
+}
+
+func (c *Client) watchL1StateUpdates(ctx context.Context) error {
 	buffer := 128
 
 	c.logger.Info("Subscribing to L1 updates...")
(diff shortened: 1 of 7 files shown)


l1/l1_test.go
```diff
@@ -174,6 +209,79 @@ func TestEventListener(t *testing.T) {
...
+	// Live subscription delivers nothing; the catch-up scan alone must
+	// populate nonFinalisedLogs so setL1Head fires the listener callback.
+	subscriber.
+		EXPECT().
+		WatchLogStateUpdate(gomock.Any(), gomock.Any()).
+		Return(newFakeSubscription(), nil).
+		AnyTimes()
+
+	// LatestHeight=10, FinalisedHeight=5, catchUpChunkSize=1000 →
+	// single chunk [0, 10] (from = max(0, 11-1000) = 0).
```
thiagodeev: Can we remove this `(from = max(0, 11-1000) = 0)` part? If I didn't misunderstand, this is not true. The `from` variable starts as 0, and it will only be set if `LatestHeight` + 1 is > `catchUpChunkSize` (1000), so it's not a `max(x, x)` operation. Just to avoid confusion when reading it (as I had myself)

l1/l1_pkg_test.go
```diff
@@ -476,3 +518,268 @@ func TestUnreliableSubscription(t *testing.T) {
...
+	got, err := chain.L1Head()
+	require.NoError(t, err)
+	assert.Equal(t, core.L1Head{
+		BlockNumber: 25,
+		BlockHash:   new(felt.Felt).SetUint64(25),
+		StateRoot:   new(felt.Felt).SetUint64(25),
+	}, got)
+}
+
+func TestCatchUpFilterError(t *testing.T) {
```
thiagodeev, line 638: For this test, WDYT about a logic to observe the logs and check whether the error is really being logged?

l1/l1.go
```diff
@@ -177,6 +218,92 @@ func (c *Client) Run(ctx context.Context) error {
...
+// subscription, so the partial state is an additive head-start, not a leak.
+func (c *Client) catchUpL1HeadUpdates(ctx context.Context) error {
+	latest, err := c.l1.LatestHeight(ctx)
+	if err != nil {
+		return errors.Join(err, errors.New("L1 catch-up: failed to get latest height"))
+	}
+	finalised, err := c.l1.FinalisedHeight(ctx)
+	if err != nil {
+		return errors.Join(err, errors.New("L1 catch-up: failed to get finalised height"))
+	}
```
thiagodeev, line 260: The rest of the code is using `fmt.Errorf(...)`, why not use it as well here?
Also, wouldn't it be better to switch the order here? Make "L1 catch-up: failed to get latest height" come first, and then add the other error?

l1/l1.go
```diff
@@ -177,6 +218,92 @@ func (c *Client) Run(ctx context.Context) error {
...
+		}
+		events, err := c.l1.FilterLogStateUpdate(ctx, from, to)
+		if err != nil {
+			return err
+		}
+		chunks++
+		total += len(events)
+		for _, ev := range events {
+			c.applyLogStateUpdate(ev)
+			if !ev.Raw.Removed && ev.Raw.BlockNumber <= finalised {
```
thiagodeev: Calling `if !ev.Raw.Removed` is useless in the `catchUpL1HeadUpdates` method; am I correct?
This check only makes sense in the subscription flow (in case of reorg, the node would return the same log, but populating this field as true). The `l1.FilterLogStateUpdate` call returns a stateless array of logs, this field value will always be false.

The case is the same for the `applyLogStateUpdate` method: the `if u.Raw.Removed` will never be true for the catchUp flow. However, it also doesn't cause any harm; it's just ignored. You can keep it as is, or maybe remove the check and put it directly in the `watchL1StateUpdates` method

thiagodeev (approved):
Nice work! LGTM.
Just a few nitpicks

l1/l1_test.go
EgeCaner: To be clear, I don't think the comment is wrong, read in signed arithmetic, `max(0, 11 - 1000) = max(0, -989) = 0` is a faithful summary of what the guarded subtraction computes. I added it as a concise way to show the from derivation. But the point of an inline comment is to reduce reader friction, and if it's making readers stop and re-derive whether it's literal Go or signed shorthand, it's failing at its job. Happy to drop it.

EgeCaner: Removed `(from = max(0, 11-1000) = 0)` part

l1/l1.go
rodrodros: I don't agree with this comment, fmt.Errorf(...) is when there is something to format, `errors.New` means there is nothing to format.

rodrodros: After seeing the change, I see that previously the error was not wrapped, so I agree with the current state

EgeCaner: Goal here is to wrap error message with context, I just switched to `fmt.Errorf` with `%w`. I might revert back if you disagree @rodrodros.

l1/l1_pkg_test.go
rodrodros: If done, this should be handled with care

EgeCaner: Im not sure if I should do it, it is not common practice in our repo to observe logs in test, and not sure if upside worths. Lets discuss in daily

rodrodros:
Bypassing coverage rule because missing tests will be add in a follow up

PR #3595, opened 2026-05-01
Introduce semantic wrappers around Execute

RafaelGranza:
## Summary

First of four incremental PRs refactoring `vm.Execute` (go) and `cairo_vm_execute` (rust) from a 7-positional-boolean signature into three semantic operations with typed options. **This PR is purely additive**. No call sites are migrated yet, and `Execute` remains fully functional.

## Why

`vm.Execute` takes 7 positional `bool` parameters and is invoked from 8 production sites and ~38 test setups. The shape has caused real documentation drift:
-  The comments next to call args in `rpc/v9/trace.go` and `rpc/v10/trace.go` describe parameters that don't exist in the signature (`// skipNonceCharge`, `// allowZeroMaxFee`, `// allowNoSignature` — none are real `Execute` params).

Every existing call site falls into one of three semantic buckets:

- **Trace**: replay an existing block (only `returnInitialReads` varies)
- **Simulate**: RPC `simulate_transactions` / `estimate_fee` (5 flags vary)
- **BuildBlock**: block production: builder + genesis (3 flags vary)

## Roadmap

| PR | Scope |
|---|---|
| **1 (this PR)** | Add `Simulate/Trace/BuildBlock` wrappers + options structs |
| 2 | Migrate 8 production call sites + ~38 test setups |
| 3 | Split Rust `cairo_vm_execute` into three dedicated entry points; wrappers call them directly |
| 4 | Remove `Execute` (Go interface, `*vm`, `ThrottledVM`, mock) and `cairo_vm_execute` (Rust + FFI header) |

## Changes in this PR

- `vm/vm.go`: adds `SimulateOptions`, `TraceOptions`, `BuildBlockOptions` and the three corresponding methods on the `VM` interface and `*vm` implementation. The methods delegate to `Execute` for now.
- `node/throttled_vm.go`: mirrors the three methods with throttle wrappers.
- `mocks/mock_vm.go`: adds gomock implementations.

## Out of scope (covered by follow-up PRs)

- No call sites are migrated. `Execute` is still the method actually being invoked everywhere in production and tests.
- No changes to Rust or the C FFI header.
- The implementation-detail comments on the new methods (e.g. "errStack and allowBinarySearch are always on") describe the current delegation to `Execute`. These will be replaced in PR 3 when each method gets its own Rust entry point and those flags cease to exist as configuration.


vm/vm.go
```diff
@@ -45,6 +45,32 @@ type CallResult struct {
 	ExecutionFailed bool
 }
 
+// SimulateOptions carries the per-request flags relevant to RPC simulate /
+// estimateFee. errStack and allowBinarySearch are constants for this path
+// and are set internally.
```
RafaelGranza, line 50: The documentation talking about flags that are constants or pre-defined will be removed after these methods stop depending on `Execute`

vm/vm.go
```diff
@@ -45,6 +45,32 @@ type CallResult struct {
...
+	ErrOnRevert        bool
+	IsEstimateFee      bool
+	ReturnInitialReads bool
+}
+
+// TraceOptions carries the per-request flags relevant to replaying an
+// existing block. All execution flags are off; only errStack is on.
+type TraceOptions struct {
+	ReturnInitialReads bool
+}
```
RafaelGranza, line 63: I know it could be a single bollean, but I decided to keep this as a struct, for 2 reasons:
- it keeps the same signature pattern as the other wrappers
- if we ever need extra flags, adding a field to this struct is less conflicting than adding and extra argument to a method.

rodrodros (commented):
Looks good 👍

RafaelGranza:
Move the changes to here: https://github.com/NethermindEth/juno/pull/3602

PR #3596, opened 2026-05-02
feat: history pruner service

EgeCaner:
## Summary

Adds a `pruner` service that bounds on-disk storage by deleting block data and state history below a configurable retention window. With `--retained-blocks=N`, the node keeps blocks in `(l1_head - N, l2_head]` and prunes everything below.

- **RPC stays fully functional inside the retention window.** Any block in `(l1_head - N, l2_head]` answers normally (events, traces, state-at-block, etc.). Below the floor: `event_filter` surfaces a typed `ErrBlockPruned`; `statebackend` returns `db.ErrKeyNotFound` (i.e. block-not-found), same as if the block never existed.
- **Floor is anchored on the L1-verified head, never on the local L2 head.** Pruned blocks are reorg-safe by construction — anything at or below the L1 head cannot be reorged out. New L2 heads only act as triggers; the floor itself is always recomputed from L1.
- **Retention can be changed across restarts.** Lowering `--retained-blocks` shrinks the window on the next sweep. Raising it grows the window gradually — the pruner pauses until the L1 head advances enough to reach the new floor; nothing is "un-pruned".
- **Disabled by default.** `--retained-blocks=0` keeps existing behaviour.
- **Irreversible.** Data deleted under a small window cannot be recovered without re-syncing.

## Open for discussion

`starknet_getEvents` with an unset `from_block` currently errors with `ErrBlockPruned` instead of silently clamping to the oldest retained block. Strict over silent — an unbounded query against a pruned node has clearly lost data, and we'd rather surface that than return a partial result the caller doesn't know is partial. Happy to flip to clamp-to-oldest if reviewers prefer.

## Details

See the package doc on `pruner/pruner.go` and the contract on `PruneUpto` in `pruner/accessors.go` for the exact deletion set, the two intentional carve-outs (BlockHashLag headers and the parent-hash mapping), and resume semantics on cancellation.

Touches outside `pruner/`:
- `blockchain/event_filter.go`, `blockchain/statebackend/*` — gate reads on retention via `pruner.RequireRetained` / `pruner.HeaderBy*IfStateRetained`.
- `core/block.go` — `BlockHashLag` and `BlockHashStorageContract` moved here from `sync` so the pruner carve-out and RPC `getRevealedBlockHash` share one source of truth.
- `node/node.go`, `cmd/juno/juno.go` — `--retained-blocks` flag and service wiring (sequencer and sync paths).
- `node/metrics.go` — Prometheus listener for prune events.

### DIFF
--- a/blockchain/event_filter.go
+++ b/blockchain/event_filter.go
@@ -10,6 +10,7 @@ import (
 	"github.com/NethermindEth/juno/core/felt"
 	"github.com/NethermindEth/juno/core/pending"
 	"github.com/NethermindEth/juno/db"
+	"github.com/NethermindEth/juno/pruner"
 )
 
 var errChunkSizeReached = errors.New("chunk size reached")
@@ -26,7 +27,7 @@ type EventFilterer interface {
 }
 
 type EventFilter struct {
-	txn            db.KeyValueStore
+	database       db.KeyValueStore
 	fromBlock      uint64
 	toBlock        uint64
 	matcher        EventMatcher
@@ -44,7 +45,7 @@ const (
 )
 
 func newEventFilter(
-	txn db.KeyValueStore,
+	database db.KeyValueStore,
 	contractAddresses []felt.Address,
 	keys [][]felt.Felt,
 	fromBlock, toBlock uint64,
@@ -53,7 +54,7 @@ func newEventFilter(
 	runningFilter *core.RunningEventFilter,
 ) *EventFilter {
 	return &EventFilter{
-		txn:            txn,
+		database:       database,
 		matcher:        NewEventMatcher(contractAddresses, keys),
 		fromBlock:      fromBlock,
 		toBlock:        toBlock,
@@ -91,7 +92,7 @@ func (e *EventFilter) SetRangeEndBlockByHash(
 	filterRange EventFilterRange,
 	blockHash *felt.Felt,
 ) error {
-	header, err := core.GetBlockHeaderByHash(e.txn, blockHash)
+	header, err := core.GetBlockHeaderByHash(e.database, blockHash)
 	if err != nil {
 		return err
 	}
@@ -100,7 +101,7 @@ func (e *EventFilter) SetRangeEndBlockByHash(
 
 // SetRangeEndBlockToL1Head sets an end of the block range to latest `l1_accepted` block
 func (e *EventFilter) SetRangeEndBlockToL1Head(filterRange EventFilterRange) error {
-	l1Head, err := core.GetL1Head(e.txn)
+	l1Head, err := core.GetL1Head(e.database)
 	if err != nil {
 		return err
 	}
@@ -151,7 +152,7 @@ func (e *EventFilter) Events(
 ) ([]FilteredEvent, ContinuationToken, error) {
 	var matchedEvents []FilteredEvent
 
-	latest, err := core.GetChainHeight(e.txn)
+	latest, err := core.GetChainHeight(e.database)
 	if err != nil {
 		return nil, ContinuationToken{}, err
 	}
@@ -162,6 +163,15 @@ func (e *EventFilter) Events(
 	if cToken != nil {
 		skippedEvents = cToken.processedEvents
 		startBlock = cToken.fromBlock
+	}
+
+	// Reject queries whose canonical start block has been pruned. Skipped
+	// when startBlock is in the pre-confirmed range (> latest), where
+	// retention semantics don't apply.
+	if startBlock <= latest {
+		if err := pruner.RequireRetained(e.database, startBlock); err != nil {
+			return nil, ContinuationToken{}, err
+		}
 	}
 
 	// Case [canonicalBlock, canonicalBlock]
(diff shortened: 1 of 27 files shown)


rodrodros:
> anything at or below the L1 head cannot be reorged out

What happens if there is a re-org on L1?

blockchain/event_filter.go
```diff
@@ -26,7 +27,7 @@ type EventFilterer interface {
 }
 
 type EventFilter struct {
-	txn            db.KeyValueStore
+	database       db.KeyValueStore
```
rodrodros, line 30: 👍

node/node.go
```diff
@@ -51,11 +52,12 @@ import (
...
-	mempoolLimit     = 1024
-	githubAPIUrl     = "https://api.github.com/repos/NethermindEth/juno/releases/latest"
-	latestReleaseURL = "https://github.com/NethermindEth/juno/releases/latest"
-	sequencerAddress = 1337
+	upgraderDelay      = 5 * time.Minute
+	mempoolLimit       = 1024
+	githubAPIUrl       = "https://api.github.com/repos/NethermindEth/juno/releases/latest"
+	latestReleaseURL   = "https://github.com/NethermindEth/juno/releases/latest"
+	sequencerAddress   = 1337
+	RetainedBlocksFlag = "retained-blocks"
```
rodrodros: Nobody is going to guess this flag name is about pruning. I believe it is better to change the external way we refer to this flag as `prune-blocks` 

In the description it is established that it should be used as `--prune-blocks=x` where `x != 0` for it to be on, but I think it is better UX if just by defining the flag the pruning mode gets activated. 

Suggested new usage:
```bash
--prune-mode    # triggers prunning from the L1 Head
--prune-mode=0  # same as above
--prune-mode=x  # triggers prunning from the L1 Head - x
```

What do you think?

I believe our CLI library can easily handle this

cmd/juno/juno.go
rodrodros, line 1: If we take the above comment idea, we should expand it to most of this file as well

pruner/testutils/helpers.go
```diff
@@ -0,0 +1,412 @@
...
+	"github.com/NethermindEth/juno/db/pebblev2"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// NewTestDB returns an isolated pebble-backed KeyValueStore for the test.
+// We use pebble (not the in-memory db) because pebble's batch DeleteRange
+// performs a real range delete, while the memory implementation only
+// scans keys with the start byte sequence as a literal prefix.
+func NewTestDB(t *testing.T) db.KeyValueStore {
```
rodrodros: nit: maybe we should put Pebble in the name to be more explicit

rodrodros (commented):
All looking good so far, some initial comments

core/block.go
```diff
@@ -15,6 +15,17 @@ import (
...
+// BlockHashLag is the Starknet protocol-defined lag for the BLOCK_HASH
+// syscall: when executing block N, contracts can read the hash of block
+// N - BlockHashLag.
+const BlockHashLag uint64 = 10
+
+// Block hash storage contract introduced with starknet v0.12.0.
+// Contract stores block number to block hash mapping,
+// and serve this information through [get_block_hash_syscall].
+// Queriable range is [first_v0_12_0_block, current_block - 10]
+var BlockHashStorageContract = &felt.One
```
rodrodros, line 27: From where are this constant extracted? If they are Starknet Constant, then they should be loaded from there, to avoid the breaking change when moving from Starknet versions.

Can be done in a follow up and we can discuss on the daily, because it can be a bit complicated to do

pruner/accessors.go
```diff
@@ -0,0 +1,318 @@
...
+//  1. Block headers in [endExclusive-BlockHashLag, endExclusive). The
+//     get_block_hash_syscall (see [core.BlockHashLag]), run during
+//     execution of any block in [endExclusive, endExclusive+BlockHashLag),
+//     reads a header by number from this window — pruning it would break
+//     the syscall.
+//
+//  2. The hash → number mapping for endExclusive-1. Resolving
+//     StateAtBlockHash(endExclusive.parentHash) needs this single mapping;
+//     it is cleaned up by the next PruneUpto call's oldestKept-1 sweep, so
+//     at rest exactly one extra mapping survives below endExclusive.
```
rodrodros: What do you mean by "at rest"?

pruner/pruner_test.go
```diff
@@ -0,0 +1,368 @@
+package pruner
```
rodrodros: I think we should restrict ourselves to `package pruner_test` here and avoid testing our impl and testing our functionality instead

pruner/accessors_test.go
```diff
@@ -0,0 +1,370 @@
+package pruner
```
rodrodros: Similar comment here. Let me know if there is a hard blocker here. Also, the change shouldn'tbe putting `pruner_test` and then setting the private function used here as public, at least without an explanation.

rodrodros (commented):
Two test files should be correctly set as `pruner_test`

pruner/accessors.go
EgeCaner: By "at rest" I was trying to mean "in steady state, when no prune call is running",  i.e. between calls. I've updated the wording to "between calls" I guess it's  more clear.

core/block.go
rodrodros: @EgeCaner can you add this issue to the issue list

rodrodros (approved):

PR #3597, opened 2026-05-02
feat: pruning migration

EgeCaner:
Adds optional migration for history-pruning (migration/historyprunner) that brings an existing node's database in line with the rolling pruner's retention window. Activated whenever `--prune-mode` is passed (with or without a value, including `--prune-mode=0`); no-op otherwise.
                                                                                                                                                                                        
The migration runs in two pipelined, resumable phases. The keeper window is anchored on the L1-verified head (matching the rolling pruner's reorg-safe floor): with `minHead = min(l1Head, chainHeight)` and `retainedBlocks` taken from the `--prune-mode=N` value (defaulting to 0 when the flag is passed bare), the window covers `[minHead - retainedBlocks + 1, chainHeight]`. If `minHead < retainedBlocks` the chain is shorter than the window and the migration no-ops, leaving the rolling pruner to take over once blocks arrive.
- Stager — range-deletes the cold number-keyed buckets and wipes reverse-lookup buckets (`TransactionBlockNumbersAndIndicesByHash`, `L1HandlerTxnHashByMsgHash`, `BlockHeaderNumbersByHash`), then copies in-window entries from the live history buckets (`ContractStorageHistory`, `ContractClassHashHistory`, `ContractNonceHistory`) into a scratch namespace.
- Restorer — wipes the live history buckets, copies keepers back from scratch, and rebuilds the reverse-lookup indices per block. Wipes the scratch namespace on completion.

Resume state (`stagerProgress` | `restorerProgress` | `numRetainedBlocks`, 24 bytes BE) is persisted between runs; the originally-configured retention pins the cutoff so a --prune-mode change mid-flight cannot retarget an in-progress migration (logged when ignored).

### DIFF
--- a/migration/historyprunner/committer.go
+++ b/migration/historyprunner/committer.go
@@ -8,6 +8,8 @@ import (
 	"go.uber.org/zap"
 )
 
+// committer is a pipeline stage that writes batches to the database and
+// releases the batch semaphore slot.
 type committer struct {
 	logger         log.StructuredLogger
 	batchSemaphore semaphore.ResourceSemaphore[db.Batch]
--- a/migration/historyprunner/committer.go
+++ b/migration/historyprunner/committer.go
@@ -1,0 +1,48 @@
+package historyprunner
+
+import (
+	"github.com/NethermindEth/juno/db"
+	"github.com/NethermindEth/juno/migration/pipeline"
+	"github.com/NethermindEth/juno/migration/semaphore"
+	"github.com/NethermindEth/juno/utils/log"
+	"go.uber.org/zap"
+)
+
+type committer struct {
+	logger         log.StructuredLogger
+	batchSemaphore semaphore.ResourceSemaphore[db.Batch]
+}
+
+var _ pipeline.State[db.Batch, struct{}] = (*committer)(nil)
+
+func newCommitter(
+	logger log.StructuredLogger,
+	batchSemaphore semaphore.ResourceSemaphore[db.Batch],
+) *committer {
+	return &committer{
+		logger:         logger,
+		batchSemaphore: batchSemaphore,
+	}
+}
+
+func (c *committer) Run(_ int, batch db.Batch, _ chan<- struct{}) error {
+	c.logger.Debug(
+		"writing batch",
+		zap.Int("batch_size", batch.Size()),
+	)
+	defer c.logger.Debug(
+		"wrote batch",
+		zap.Int("batch_size", batch.Size()),
+	)
+
+	if err := batch.Write(); err != nil {
+		return err
+	}
+
+	c.batchSemaphore.Put()
+	return nil
+}
+
+func (c *committer) Done(int, chan<- struct{}) error {
+	return nil
+}
--- a/migration/historyprunner/copy.go
+++ b/migration/historyprunner/copy.go
@@ -51,7 +51,10 @@ func copyStateHistory(
 				fillStorageHistoryKey(scratch.historyKey[:], &addr, &slot, blockBE),
 				fillStorageScratchKey(scratch.scratchKey[:], &addr, &slot, blockBE),
 			); err != nil {
-				return err
+				return fmt.Errorf(
+					"copying storage history at block %d, addr %x, slot %x: %w",
+					blockNum, addr, slot, err,
+				)
 			}
 		}
 	}
@@ -61,7 +64,10 @@ func copyStateHistory(
 			fillNonceHistoryKey(scratch.historyKey[:], &addr, blockBE),
 			fillNonceScratchKey(scratch.scratchKey[:], &addr, blockBE),
 		); err != nil {
-			return err
+			return fmt.Errorf(
+				"copying nonce history at block %d, addr %x: %w",
+				blockNum, addr, err,
+			)
 		}
 	}
 
@@ -70,7 +76,10 @@ func copyStateHistory(
 			fillClassHashHistoryKey(scratch.historyKey[:], &addr, blockBE),
 			fillClassHashScratchKey(scratch.scratchKey[:], &addr, blockBE),
 		); err != nil {
-			return err
+			return fmt.Errorf(
+				"copying class hash history at block %d, addr %x: %w",
+				blockNum, addr, err,
+			)
 		}
 	}
 
--- a/migration/historyprunner/copy.go
+++ b/migration/historyprunner/copy.go
@@ -1,0 +1,106 @@
+package historyprunner
+
+import (
+	"encoding/binary"
+	"fmt"
+
+	"github.com/NethermindEth/juno/core"
+	"github.com/NethermindEth/juno/db"
+)
+
+// copyScratch holds the reusable byte buffers for one worker's copy
+// operations. Each worker owns one copyScratch; buffers are overwritten on
+// every entry and consumed (via batch.Put) before the next iteration
+// touches them, so reuse is safe. Allocated once at stager/restorer
+// construction.
+type copyScratch struct {
+	historyKey [historyKeyBufLen]byte
+	scratchKey [scratchKeyBufLen]byte
+	value      [historyValueLen]byte
+}
+
+type copyDirection bool
+
+const (
+	historyToScratch copyDirection = false
+	scratchToHistory copyDirection = true
+)
+
+func copyStateHistory(
+	reader db.KeyValueReader,
+	batch db.Batch,
+	scratch *copyScratch,
+	diff *core.StateDiff,
+	blockNum uint64,
+	direction copyDirection,
+) error {
+	var blockBE [blockNumberSuffixLen]byte
+	binary.BigEndian.PutUint64(blockBE[:], blockNum)
+
+	move := func(historyKey, scratchKey []byte) error {
+		src, dst := historyKey, scratchKey
+		if direction == scratchToHistory {
+			src, dst = dst, src
+		}
+		return copyValue(reader, batch, scratch.value[:], src, dst)
+	}
+
+	for addr, slots := range diff.StorageDiffs {
+		for slot := range slots {
+			if err := move(
+				fillStorageHistoryKey(scratch.historyKey[:], &addr, &slot, blockBE),
+				fillStorageScratchKey(scratch.scratchKey[:], &addr, &slot, blockBE),
+			); err != nil {
+				return err
+			}
+		}
+	}
+
+	for addr := range diff.Nonces {
+		if err := move(
+			fillNonceHistoryKey(scratch.historyKey[:], &addr, blockBE),
+			fillNonceScratchKey(scratch.scratchKey[:], &addr, blockBE),
+		); err != nil {
+			return err
+		}
+	}
+
+	for addr := range diff.ReplacedClasses {
+		if err := move(
+			fillClassHashHistoryKey(scratch.historyKey[:], &addr, blockBE),
+			fillClassHashScratchKey(scratch.scratchKey[:], &addr, blockBE),
+		); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// copyValue reads source into buf, then writes buf to dest. buf is the
+// caller's reusable felt-sized scratch (32 bytes); after Put returns, the
+// underlying batch has copied buf, so the caller may overwrite it safely.
+func copyValue(
+	reader db.KeyValueReader,
+	writer db.KeyValueWriter,
+	buf,
+	source,
+	dest []byte,
+) error {
+	err := reader.Get(source, func(data []byte) error {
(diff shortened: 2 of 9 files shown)


migration/historyprunner/committer.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/NethermindEth/juno/migration/pipeline"
+	"github.com/NethermindEth/juno/migration/semaphore"
+	"github.com/NethermindEth/juno/utils/log"
+	"go.uber.org/zap"
+)
+
+type committer struct {
+	logger         log.StructuredLogger
+	batchSemaphore semaphore.ResourceSemaphore[db.Batch]
+}
```
rodrodros, line 16: Description please 🙏

migration/historyprunner/committer.go
```diff
@@ -0,0 +1,48 @@
...
+		return err
+	}
+
+	c.batchSemaphore.Put()
+	return nil
+}
+
+func (c *committer) Done(int, chan<- struct{}) error {
+	return nil
+}
```
rodrodros, line 50: This functions seem to be having arguments they don't use. Are they being restricted by an interface?

migration/historyprunner/copy.go
```diff
@@ -0,0 +1,106 @@
...
+	scratchKey [scratchKeyBufLen]byte
+	value      [historyValueLen]byte
+}
+
+type copyDirection bool
+
+const (
+	historyToScratch copyDirection = false
+	scratchToHistory copyDirection = true
+)
```
rodrodros, line 27: Lovely ❤️

migration/historyprunner/committer.go
EgeCaner: Yes we are constrained by the interface here 

```go
type State[I, O any] interface {
	Run(index int, input I, outputs chan<- O) error
	Done(index int, outputs chan<- O) error
}

```

migration/historyprunner/committer.go
EgeCaner: I added brief description for this, should I provide more details? like why it is useful etc...?

rodrodros: No need

migration/historyprunner/migrator.go
```diff
@@ -0,0 +1,412 @@
...
+		if errors.Is(err, db.ErrKeyNotFound) {
+			// No chain data yet; the rolling pruner takes over once blocks arrive.
+			return nil, nil
+		}
+		return nil, err
+	}
+	l1Head, err := core.GetL1Head(database)
+	if err != nil {
+		return nil, err
+	}
```
rodrodros, line 109: If we have blocks but no l1 head I agree we should stop, but does a freshly synced node with no L1 head yet will fail this right? Can not having an L1 Head but being on a fresh new head cause bad UX 🤔

rodrodros: I've tested it, it will crash, additionally with no extra info such as, why is it crashing:
```
16:13:19.401 09/05/2026 +01:00	ERROR	node/node.go:662	Error while running migrations	{"error": "key not found"}
```

migration/historyprunner/migrator.go
```diff
@@ -0,0 +1,412 @@
...
+	m.numRetainedBlocks = binary.BigEndian.Uint64(state[16:24])
+	return nil
+}
+
+func (m *Migrator) Migrate(
+	ctx context.Context,
+	database db.KeyValueStore,
+	network *networks.Network,
+	logger log.StructuredLogger,
+) ([]byte, error) {
```
rodrodros, line 90: Please add descriptive errors to each of the err returns for this function, otherwise error cause is obscure in many cases. Other functions are optional, but this is the main one

migration/historyprunner/copy.go
```diff
@@ -0,0 +1,106 @@
...
+			return err
+		}
+	}
+
+	for addr := range diff.ReplacedClasses {
+		if err := move(
+			fillClassHashHistoryKey(scratch.historyKey[:], &addr, blockBE),
+			fillClassHashScratchKey(scratch.scratchKey[:], &addr, blockBE),
+		); err != nil {
+			return err
```
rodrodros: Error wrapper for this as well please

migration/historyprunner/migrator.go
```diff
@@ -0,0 +1,412 @@
...
+	database db.KeyValueStore,
+	oldestBlockKept uint64,
+) error {
+	if m.restorerProgress != 0 {
+		return nil
+	}
+	batch := database.NewBatch()
+	if err := pruner.PruneBlockDataUpto(batch, oldestBlockKept); err != nil {
+		return err
+	}
```
rodrodros, line 189: This would benefit from error wrapping

migration/historyprunner/migrator.go
```diff
@@ -0,0 +1,412 @@
...
+	if err := wipeReverseLookupBuckets(batch); err != nil {
+		return err
+	}
+	if err := batch.Write(); err != nil {
+		return err
+	}
```
rodrodros, line 195: I am thinking that this batch can potentially be very big 🤔, you're pruning until the first block to keep, which the bigger mainnet, the lot more changes that the batch is going to hold. 

Add that to `wipeReversaeLookupBuckets(batch)` which I am not sure how much they add but that also compounds. 

I think for safety, even if now we are good, think of a huge mainnet (10Tb of data), would this still hold or would it be better to batch it through?

rodrodros (commented):
Overall looks like really good work. Just sharing some questions and nitpicks

I am curious about the cases when we start the migration close to the prune threshold
1. L1Head == L2Head, do we delete everything at this stage
2. L1Head - 1 == L2Head, we have only one block?

rodrodros:
I am thinking that instead of having prune mode with 0 as a default we can put a higher number (?). This can not make this edge cases the default?

migration/historyprunner/migrator.go
```diff
@@ -0,0 +1,412 @@
...
+
+	batch := database.NewBatch()
+	err = core.WriteBlockHeaderNumberByHash(batch, header.Hash, oldestBlockKept-uint64(1))
+	if err != nil {
+		return nil, false, err
+	}
+	if err := batch.Write(); err != nil {
+		return nil, false, err
+	}
+
```
rodrodros: Maybe this can be part of `setupRestorer`, what do you think?

RafaelGranza:
Mistakenly pushed some commits to this PR, my bad!
Quickly reverted them.

RafaelGranza:
While smoke-testing this PR I ran into a separate pruning issue. NOT RELATED to the `key not found` crash. 
Opening a follow-up PR to look into it; flagging here for visibility.

## Repro

1. Start with `--prune-mode=10` on a fresh DB (Sepolia)
2. Let L2 sync ~300 blocks (~30s)
3. Ctrl+C
4. Restart with the same flags

## Observed

The sync height keeps incrementing but every block fails to store with `key not found`, restarting the sync process each iteration:
```
WARN sync/sync.go:365 Failed storing Block {"number": 301, "error": "key not found"}
WARN sync/sync.go:498 Restarting sync process {"height": 302, "catchUpMode": false}
WARN sync/sync.go:365 Failed storing Block {"number": 302, "error": "key not found"}
WARN sync/sync.go:498 Restarting sync process {"height": 303, "catchUpMode": false}
```
So height advances (302 → 303 → 304 ...) but no block is actually persisted.
**The node is stuck in this loop indefinitely.**

## Other rough edges noticed during smoke-test (follow-ups)

- Running a previously`--prune-mode` DB **without** `--prune-mode` fails with a misleading `"database is from a newer, incompatible version of Juno; upgrade to use this database"` — the actual cause is the optional pruning migration being recorded in the schema metadata. Better error message needed.
- That same failure path then panics with `Couldn't initialised the running event filter` in the shutdown path — startup abort should gracefully skip the cleanup if blockchain isn't fully initialised.


**Will dig into it in a follow-up PR.**

migration/historyprunner/migrator.go
```diff
@@ -0,0 +1,412 @@
...
+		zap.Uint64("chain_height", chainHeight),
+		zap.Uint64("oldest_block_kept", oldestBlockKept),
+		zap.Uint64("l1_head", l1Head.BlockNumber),
+	)
+
+	if err := m.setupBeforeStager(database, oldestBlockKept); err != nil {
+		return nil, err
+	}
+
+	maxWorkers := runtime.GOMAXPROCS(0)
```
MaksymMalicki, line 128: doesn't this cause a big pressure on the committer? did you think about capping it in any way?

migration/historyprunner/migrator.go
EgeCaner: This section writes constant number of range delete thoımbstones regardless of pruned size

migration/historyprunner/migrator_test.go
```diff
@@ -0,0 +1,200 @@
...
+	// must either return a non-nil intermediate state (still going) or
+	// nil (done).
+	var (
+		state    []byte
+		attempts int
+	)
+	for {
+		attempts++
+		require.LessOrEqual(t, attempts, 10, "should converge in well under 10 attempts")
+		state = runCancellable(t, state, database, 2, retainedBlocks)
```
RafaelGranza, line 190: Funny, I already experienced this as a [flaky test](https://github.com/NethermindEth/juno/actions/runs/25817276189/job/75848965017?pr=3634).
[This one worked](https://github.com/NethermindEth/juno/actions/runs/25817276189/job/75848964969?pr=3634).

EgeCaner: Removed the attempt cap, now test just ensures multiple runs & cancels converges to same outcome

migration/historyprunner/migrator.go
EgeCaner: Moved to setup

EgeCaner:
> I am curious about the cases when we start the migration close to the prune threshold
> 
>     1. L1Head == L2Head, do we delete everything at this stage
> 
>     2. L1Head - 1 == L2Head, we have only one block?
@rodrodros 

With `--pruned-mode=0`:
- When `L1Head` ahead of `L2Head` - only latest block is kept 
- When `L2Head` == `L1Head` - only one block kept (latest and l1 head at the same time)
- When `L2Head` ahead of `L1Head`- [L1Head, L2Head] is kept

> I am thinking that instead of having prune mode with 0 as a default we can put a higher number (?). This can not make this edge cases the default?

Lets discuss this in daily, I will reflect the outcome of the discussion here later

EgeCaner:
@RafaelGranza problems you have reported [here](https://github.com/NethermindEth/juno/pull/3597#issuecomment-4424221604) should be solved by these PRs:
- > Running a previously--prune-mode DB without --prune-mode fails with a misleading "database is from a newer, incompatible version of Juno; upgrade to use this database" — the actual cause is the optional pruning migration being recorded in the schema metadata. Better error message needed -->  https://github.com/NethermindEth/juno/pull/3637 
- > That same failure path then panics with Couldn't initialised the running event filter in the shutdown path — startup abort should gracefully skip the cleanup if blockchain isn't fully initialised. "Sync getting stuck is also related to running filter initialisation"--> https://github.com/NethermindEth/juno/pull/3638

RafaelGranza:
> @RafaelGranza problems you have reported [here](https://github.com/NethermindEth/juno/pull/3597#issuecomment-4424221604) should be solved by these PRs:

@EgeCaner, I'm not sure if the changes you made cover the fixes in this [PR](https://github.com/NethermindEth/juno/pull/3634/changes).

It allows us to run `--prune-mode` on  a node that was initialized but had not fetched L1Head yet.
It also rejects running juno with `--prune-mode` + `--disable-l1-verification`

EgeCaner:
> > @RafaelGranza problems you have reported [here](https://github.com/NethermindEth/juno/pull/3597#issuecomment-4424221604) should be solved by these PRs:
> 
> @EgeCaner, I'm not sure if the changes you made cover the fixes in this [PR](https://github.com/NethermindEth/juno/pull/3634/changes).
> 
> It allows us to run `--prune-mode` on a node that was initialized but had not fetched L1Head yet. It also rejects running juno with `--prune-mode` + `--disable-l1-verification`

I have seen that you PR was covering those, haven't addressed them myself

migration/historyprunner/migrator.go
EgeCaner: Handled by @RafaelGranza  in this [PR](https://github.com/NethermindEth/juno/pull/3634)

rodrodros (approved):

migration/historyprunner/migrator.go
rodrodros: @MaksymMalicki why do you think it will create pressure in the committer?

rodrodros (approved):

PR #3600, opened 2026-05-04
fix: linter issues after updating to golangci-lint v2.12.1

rodrodros:


consensus/tendermint/timeout.go
```diff
@@ -6,13 +6,14 @@ import (
 	"github.com/NethermindEth/juno/consensus/types/wal"
 )
 
+// todo(rdr): Should we ignore calling this function outside the right step?
+
```
brbrr, line 10: nit: probably we don't need the newline here
```suggestion
// todo(rdr): Should we ignore calling this function outside the right step?
```

brbrr (approved):
I left a small nit, but otherwise LGTM.

(rant: I find the rule annoying, but I guess it makes sense, especially if the string's semantic meaning is the same)

consensus/tendermint/timeout.go
brbrr: note: I've tried to add a suggestion, but I'm not sure if that worked or not. Is it a diff rendering issue, or the actual diff is buggy?

EgeCaner (approved):

RafaelGranza (approved):
🙌

.golangci.yaml
RafaelGranza, line 63: While refactoring the `VM Execute` code I hit what seems to be a `dupSubExpr` false positive related to cgo:
https://github.com/go-critic/go-critic/issues/897#issuecomment-568892104

Have you run into this before?

If this has been a recurring annoyance for others too, maybe this would be a good opportunity to disable it while we're already updating the linter config.

rodrodros: I don't think disabling the linter just because they flag flase positive on a very specific part of the code is a good idea.

consensus/tendermint/timeout.go
rodrodros: The todo is separated intentionally since it is a script level todo and not a function specific

PR #3601, opened 2026-05-04
refactor(feeder): use the new `includeSignature` flag in the `get_state_update` feeder endpoint

thiagodeev:
Combine `get_state_update` + `get_signature` into a single feeder call by passing the new `includeSignature=true` parameter to `get_state_update`. Removes the redundant `f.client.Signature(ctx, blockID)` call in `starknetdata/feeder.stateUpdateWithBlock`.

  - Split `clients/feeder.Client.StateUpdateWithBlock` into two methods:
    - `StateUpdateWithBlock(ctx, blockID)` — unchanged behaviour, no signature.
    - `StateUpdateWithBlockAndSignature(ctx, blockID)` — single request, returns state update + block + signature.
  - Update `clients/feeder/testdata/**/state_update_with_block/*.json` fixtures to include the `signature` field (values pulled from the matching  `signature/*.json`)
  - New feature (through the `starknetdata/feeder.FeederAdapter` type) to automatically detect whether the feeder is updated or not, and use the appropriate endpoint accordingly.
  
 Reference: https://demerzelsolutions.slack.com/archives/C03090TS3TK/p1777375140952279
 
Extra:
- In `clients/feeder/feeder.go`, per-method `c.get` + `json.NewDecoder(body).Decode(&x)` boilerplate was replaced by a single generic `doRequest[T any](ctx, client, queryURL) (*T, error)` helper. Methods in the file are now sorted alphabetically. No behavioural change, just style.


starknetdata/feeder/feeder_test.go
RafaelGranza, line 261: I guess it is also a good time to also test both signatures are equal:
```golang
if includeSignature {
     assert.Equal(t, sig.Signature, response.Signature)
} else {
     assert.Nil(t, response.Signature)
}
```

clients/feeder/feeder.go
```diff
@@ -382,10 +383,15 @@ func (c *Client) Signature(ctx context.Context, blockID string) (*starknet.Signa
...
+	ctx context.Context,
+	blockID string,
+	includeSignature bool,
+) (*starknet.StateUpdateWithBlock, error) {
 	queryURL := c.buildQueryString("get_state_update", map[string]string{
-		blockNumberArg: blockID,
-		"includeBlock": "true",
+		blockNumberArg:     blockID,
+		"includeBlock":     "true",
+		"includeSignature": strconv.FormatBool(includeSignature),
```
RafaelGranza: Have you considered letting `includeSignature`as a constant `true`? Always requesting the signature.
It could return `nil` for proposed blocks.

I didn't thought about the trade-offs, just curious

RafaelGranza:
Not approving yet, only because I'd like to run this against my own Juno node first. Code-wise looks fine.

RafaelGranza (commented):
After testing locally, `sepolia-integration` works as expected, and I would have
approved based on the approach we'd originally agreed on.

But since you'll be adding the capability-detection on top of this so it stays
compatible with all Starknet versions, I'd rather wait and re-test once that
lands before approving.

starknetdata/feeder/feeder_test.go
thiagodeev: They are already being tested :) 

The `adaptedBlock` returned by the `sn2core.AdaptBlock` function contains the `sig.Signature` that was passed to it ([ref](https://github.com/NethermindEth/juno/blob/thiagodeev/feeder-optimizations/adapters/sn2core/sn2core.go#L38)), so `assert.Equal(t, block, adaptedBlock)` is testing whether the `block` (returned by our endpoint being tested) is equal to `adaptedBlock` (returned directly by the feeder + the necessary adaptations), and this check includes the signature contained inside both blocks

clients/feeder/feeder.go
thiagodeev: Solved by implementing a separate method for the new param

RafaelGranza:
Holding off on approval.

I know the capability detection is coming in a follow-up, but I'd rather see the optimization and the detection land as a single PR. Otherwise the node sits in a half-broken state in between, working on `sepolia-integration` but dropping `Block.Signatures` on mainnet/sepolia/integration.

Happy to approve as soon as the two are combined.

rodrodros (commented):
I see the change but I am unsure if this is the best way of doing it 🤔. Let's talk on Monday

starknetdata/feeder/migrationFeeder.go
```diff
@@ -0,0 +1,179 @@
...
+	// [service.Service] requires Run to block until ctx is cancelled; otherwise
+	// [node.Node.StartService] would treat the early return as a crash and shut
+	// down the whole node via its deferred cancel().
+	<-ctx.Done()
+	return nil
+}
+
+// startVerificationLoop runs verifyFeederUpdate immediately and then every
+// [verificationInterval] until the new endpoint is confirmed or ctx is done.
+func (f *MigrationFeeder) startVerificationLoop(ctx context.Context) {
```
RafaelGranza: Since this is not a goroutine, I guess a better name would be `runVerificationLoop`.

starknetdata/feeder/migrationFeeder.go
```diff
@@ -0,0 +1,179 @@
...
+		}
+	}
+}
+
+func (f *MigrationFeeder) verifyFeederUpdate(ctx context.Context) {
+	timeoutCtx, cancel := context.WithTimeout(ctx, verificationTimeout)
+	defer cancel()
+
+	resp, err := f.feeder.client.StateUpdateWithBlockAndSignature(timeoutCtx, "0")
+	if err == nil && resp != nil && len(resp.Signature) > 0 {
```
RafaelGranza: Small suggestion: could we probe `latestID` instead of "0"?

The check relies on `len(signature) > 0`, but `block 0` isn't guaranteed to have a signature on every network (genesis isn't always backfilled). If that happens, the migration silently never flips. 

`latest` always has a signature on an upgraded feeder, so the check answers the real question: _"does this feeder support the new flag?"_. 

WDYT?

```suggestion
	resp, err := f.feeder.client.StateUpdateWithBlockAndSignature(timeoutCtx, latestID)
	if err == nil && resp != nil && len(resp.Signature) > 0 {
```

starknetdata/feeder/migration_feeder.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+}
+
+func (f *MigrationFeeder) verifyFeederUpdate(ctx context.Context) {
+	timeoutCtx, cancel := context.WithTimeout(ctx, verificationTimeout)
+	defer cancel()
+
+	resp, err := f.feeder.client.StateUpdateWithBlockAndSignature(timeoutCtx, "0")
+	if err == nil && resp != nil && len(resp.Signature) > 0 {
+		f.isFeederUpdated.Store(true)
```
RafaelGranza: Nit: add an Info log after `isFeederUpdated.Store(true)`?

It helps see when the fast path kicks in. WDYT?

starknetdata/feeder/migrationFeeder.go
```diff
@@ -0,0 +1,179 @@
...
+//
+// A background verification loop ([Run]) periodically probes the new endpoint
+// and atomically flips an internal flag once it becomes available, so the
+// transition is automatic and requires no restart.
+//
+// When the upstream feeder gateway does support the new argument, [MigrationFeeder]
+// delegates directly to the wrapped [Feeder]. Until then, it falls back to
+// the old approach with two separate requests.
+type MigrationFeeder struct {
+	feeder *Feeder
```
RafaelGranza: Could we embed `*Feeder` instead of holding it as a field? That drops these 10 delegation methods:

- `BlockByNumber`
- `BlockHeaderLatest`
- `BlockLatest`
- `BlockPreLatest`
- `Class`
- `Transaction`
- `StateUpdate`
- `StateUpdatePending`
- `StateUpdatePendingWithBlock`
- `PreConfirmedBlockByNumber`

Only `StateUpdateWithBlock` and `Run` stay as explicit overrides.

It also future-proofs against new methods added to the `StarknetData` interface.

starknetdata/feeder/feeder_adapter.go
```diff
@@ -0,0 +1,179 @@
+package feeder
```
RafaelGranza, line 1: Rename `migrationFeeder.go` to `migration_feeder.go` to match the snake_case convention used elsewhere in the repo.

starknetdata/feeder/feeder.go
```diff
@@ -145,31 +146,37 @@ func (f *Feeder) StateUpdatePending(ctx context.Context) (*core.StateUpdate, err
...
-	var sig *starknet.Signature
-	if blockID != pendingID {
-		sig, err = f.client.Signature(ctx, blockID)
+	if blockID == pendingID {
+		resp, err := f.client.StateUpdateWithBlock(ctx, blockID)
+		if err != nil {
+			return nil, nil, err
+		}
+		stateUpBlock.Block = resp.Block
+		stateUpBlock.StateUpdate = resp.StateUpdate
```
RafaelGranza: Couldn't it be `stateUpBlock = *resp` since they are the same type?

RafaelGranza (commented):
Approach looks good. Left a few style nits.

One substantive ask: probe `latest` instead of `block 0` to verify the signature.
`block 0` isn't guaranteed to have a signature on every network.

Holding approval until `MigrationFeeder` (also, rename this file) tests land.

starknetdata/feeder/migrationFeeder.go
thiagodeev: Actually, I had myself manually tested the `get_signature?blockNumber=0` endpoint in all the current Starknet networks (Mainnet, Sepolia, and Sepolia Integration), so I knew the 0 block had a signature; but I'm okay to change it to 'latest'

starknetdata/feeder/migrationFeeder.go
thiagodeev: Nice suggestion! Thanks!

starknetdata/feeder/feeder.go
thiagodeev: Makes sense, thanks!

starknetdata/feeder/migration_feeder.go
thiagodeev: I also think it would be a good thing, but the "problem" is that we don't have access to a logger inside the starknetdata client, the logger from the `feeder.Client` is private. So I'm inclined to leave it as is

starknetdata/feeder/migration_feeder_test.go
```diff
@@ -0,0 +1,187 @@
+package feeder
```
RafaelGranza: Our convention is `package X_test` for tests, accessing the package only through its public API.  I suggest using `feeder_test`


To switch over here:

1.  import `starknetdata/feeder` (matching the `adaptfeeder` alias used in `feeder_test.go`)
2. replace `latestID`with the `"latest"` literal
3. declare `verificationInterval` as a local const inside `TestMigrationFeeder` (avoids polluting the test-package namespace)

starknetdata/feeder/migration_feeder_test.go
```diff
@@ -0,0 +1,187 @@
...
+		require.NotNil(t, blk)
+
+		// **************************************************************/
+		// ******* 1st verification tick, feeder is not updated yet *****/
+		// **************************************************************/
+		mockClient.EXPECT().
+			StateUpdateWithBlockAndSignature(gomock.Any(), latestID).
+			Return(nil, errors.New("mock error"))
+		time.Sleep(time.Second)
+		synctest.Wait()
```
RafaelGranza: We cover the error case and the success-with-sig case, but not the in-between: a non-error response with an empty signature, which should also leave the flag unset.

Could be worth adding one extra tick where `StateUpdateWithBlockAndSignature`returns a non-nil response with empty `Signature`, asserting the legacy path is still used afterwards.

RafaelGranza (approved):
Approving. Tested locally on `mainnet`, `sepolia` and `sepolia-integration`; sync works on all three.

Two caveats:
- Let's wait for CI to confirm nothing slipped through.
- Please only merge after addressing the test comments below 🙏

starknetdata/feeder/migration_feeder_test.go
thiagodeev: Yeah, but the main reason was around the "test behavior, not implementation" philosophy, right? The code is not importing any private helper or methods. I decided to leave it in the same pkg only to use the same private variables (avoid unnecessary repetition, and improve test consistency since we use the exact same values as production code). This change is not relevant IMO. WDYT?

starknetdata/feeder/migration_feeder_test.go
thiagodeev: I don't believe it's worthwhile because this case doesn't exist. In the feeder code, where we made the calls, in case of errors, the code explicitly returns `nil` for the response, and only returns non-nil if the decoder succeeds. ([ref](https://github.com/NethermindEth/juno/blob/thiagodeev/feeder-optimizations/clients/feeder/feeder.go#L394)).
The only case for this to happen that came to mind is the sequencer returning a different response from what's expected to be returned (like, omitting an expected field from the response, and our code will not return any error because `json.Decode` AFAIK ignores omitted fields, it just set it to zero), but this would be a bug on their side and break the entire network

RafaelGranza: Right, you've got the mechanism exactly. Where we disagree is whether that state is a sequencer bug or the intentional pre-migration behavior.

My read: when the upstream gateway hasn't been updated yet to honor `includeSignature=true`, it'll likely keep returning the regular response without the new field. That's not a bug, it's the migration window itself, which is precisely why `MigrationFeeder` exists. The `len(resp.Signature) > 0` guard in `verifyFeederUpdate` is the production code acknowledging this case (otherwise `err == nil && resp != nil` would suffice).

If we trust that guard to do the right thing, a test pinning the behavior is cheap insurance. If we don't, the guard itself is dead code and should go.

starknetdata/feeder/migration_feeder_test.go
RafaelGranza: Fair on the "no internal helpers or methods" point, but my worry is: once `package feeder` is OK for "just constants", it gets harder to push back when someone else uses it for "just a tiny helper" or "just one private method". The convention works as a barrier.

The fix here is still cheap: `"latest"` literal and a local const inside `TestMigrationFeeder`. The duplication is one line and any drift would cause the test to fail.

starknetdata/feeder/migration_feeder_test.go
thiagodeev: >...it'll likely keep returning the regular response without the new field.

Actually, no. You can test it yourself:
- This is a request for the Sepolia Integration feeder with the new parameter:
https://feeder.integration-sepolia.starknet.io/feeder_gateway/get_state_update?blockNumber=0&includeBlock=true&includeSignature=true
- This is the same request, but for Sepolia (which hasn't been upgraded yet). You'll see that there's an error.
https://feeder.alpha-sepolia.starknet.io/feeder_gateway/get_state_update?blockNumber=0&includeBlock=true&includeSignature=true

That's why I believe a test scenario for a nil signature with a non-nil error is not worthwhile.

>The len(resp.Signature) > 0 guard ... acknowledging this case. ... the guard itself is dead code and should go.

Yeah, makes sense. My intention was to add these simple guards "just in case", but since they're adding some confusion, I'll remove them

clients/feeder/feeder.go
```diff
@@ -373,109 +288,119 @@ func (c *Client) CasmClassDefinition(
...
+	transactionHash *felt.Felt,
+) (*starknet.TransactionStatus, error) {
+	queryURL := c.buildQueryString("get_transaction_status", map[string]string{
+		"transactionHash": transactionHash.String(),
+	})
+
+	return doRequest[starknet.TransactionStatus](ctx, c, queryURL)
+}
+
+func doRequest[T any](ctx context.Context, client *Client, queryURL string) (*T, error) {
```
rodrodros, line 394: I like the approach to new request, the only thing I disagree is the return type being a pointer and forcing the allocation

starknetdata/feeder/migration_feeder_test.go
rodrodros, line 1: Let's not use the `Migration` name since that is a concept heavily associated with DB migration. From the top of my head `FeederAdapter` but feel free to use another

starknetdata/feeder/migration_feeder.go
```diff
@@ -0,0 +1,123 @@
...
+
+func (f *MigrationFeeder) verifyFeederUpdate(ctx context.Context) {
+	timeoutCtx, cancel := context.WithTimeout(ctx, verificationTimeout)
+	defer cancel()
+
+	_, err := f.Feeder.client.StateUpdateWithBlockAndSignature(timeoutCtx, latestID)
+	if err == nil {
+		f.isFeederUpdated.Store(true)
+	}
+}
```
rodrodros: nit: make this function return a boolean instead of storing it

starknetdata/feeder/migration_feeder.go
```diff
@@ -0,0 +1,123 @@
...
+		select {
+		case <-ctx.Done():
+			return
+		case <-ticker.C:
+			f.verifyFeederUpdate(ctx)
+			if f.isFeederUpdated.Load() {
+				return
+			}
+		}
+	}
```
rodrodros: nit (continuation): Made this function store the value and exit. More readable

starknetdata/feeder/feeder_adapter.go
```diff
@@ -0,0 +1,123 @@
...
+
+// Complies with the Service interface
+func (f *MigrationFeeder) Run(ctx context.Context) error {
+	f.runVerificationLoop(ctx)
+
+	// [service.Service] requires Run to block until ctx is cancelled; otherwise
+	// [node.Node.StartService] would treat the early return as a crash and shut
+	// down the whole node via its deferred cancel().
+	<-ctx.Done()
+	return nil
```
rodrodros, line 56: A service closing without an error shouldn't stop the whole thing, for a follow up

starknetdata/feeder/feeder.go
```diff
@@ -145,31 +146,36 @@ func (f *Feeder) StateUpdatePending(ctx context.Context) (*core.StateUpdate, err
...
 
 func (f *Feeder) stateUpdateWithBlock(ctx context.Context, blockID string) (*core.StateUpdate, *core.Block, error) {
-	response, err := f.client.StateUpdateWithBlock(ctx, blockID)
-	if err != nil {
-		return nil, nil, err
-	}
-
-	if blockID == pendingID && response.Block.Status != "PENDING" {
-		return nil, nil, errors.New("no pending block")
-	}
```
rodrodros, line 155: Why did this dissapeared?

starknetdata/feeder/migration_feeder.go
```diff
@@ -0,0 +1,123 @@
...
+	return nil
+}
+
+// runVerificationLoop runs verifyFeederUpdate immediately and then every
+// [verificationInterval] until the new endpoint is confirmed or ctx is done.
+func (f *MigrationFeeder) runVerificationLoop(ctx context.Context) {
+	f.verifyFeederUpdate(ctx)
+	if f.isFeederUpdated.Load() {
+		return
+	}
```
rodrodros: Success on first call is not being tested

rodrodros (approved):
Non blocking approval, open comments should be solved first

starknetdata/feeder/feeder_adapter.go
rodrodros: I've already opened an issue

starknetdata/feeder/feeder.go
thiagodeev: This check was from 3 years ago. AFAIK, "PENDING" now corresponds to the PreLatest state (an already built block that will no longer be changed, differently from the mutable "pending" from three years ago), and based on multiple calls that I did for this endpoint, there's no longer a case where, when called with the 'pending' blockID, the block status is not "PENDING", or where there is no response at all (to give sense to the `"no pending block"` message); the endpoint always return the latest PreLatest block available

clients/feeder/feeder.go
thiagodeev: At first, I made it return by value, but then this ended up in a little more verbose code (90% of the methods return by pointer, so I had to assign the returned value to a local variable and return a pointer to it for all these methods), and with some bugs (e.g.: if the helper returns T, in case of early return due to an error, methods that return by pointer were returning a valid pointer to a zero-value struct instead of returning `nil`; to avoid this, I would have to create some checks inside the helper to differenciate between calls that return pointer vs calls that return by value, or a check after the helper call only to check whether the returned value points to an empty value or not, which would produce almost the same amout of code of the helper itself, producing almost no benefit at the end).

Since 90% of the methods already return by pointer (so they were already forcing an allocation), the real difference is small: only two methods are affected (BlockHeader and FeeTokenAddresses, which return by value) at the cost of having this unified logic without any kind of workaround. That's why I decided to leave it as is.

WDYT?

rodrodros: Ok, please create an issue to tackle later

thiagodeev: Done
https://github.com/NethermindEth/juno/issues/3686

PR #3602, opened 2026-05-05
Refactor `vm.Execute` 

RafaelGranza:
# Summary

Replaces the 7 positional boolean flags on `vm.VM.Execute` with typed, semantic wrappers, so callers describe what they're doing (`Simulate` / `Trace` / `BuildBlock`) instead of memorizing flag positions.
Also breaks down the method into smaller helpers. 

## Why

`vm.Execute` takes 7 positional `bool` parameters and is invoked from 8 production sites and ~38 test setups. The shape has caused real documentation drift:

Every existing call site falls into one of three semantic buckets:

- **Trace**: replay an existing block (only `returnInitialReads` varies)
- **Simulate**: RPC `simulate_transactions` / `estimate_fee` (5 flags vary)
- **BuildBlock**: block production: builder + genesis (3 flags vary)

# Notes for reviewers

* No Rust / FFI signature changes (it was introducing more code/complexity than needed).
* `Execute` still exists to bring more options if needed.
* `Execute` was broken down into small helpers
* Since I took a look at rust code, I removed  unnecessary cloning and memory allocation from `cair_vm_execute`.


vm/rust/src/entrypoint/execute/execute.rs
```diff
@@ -82,7 +82,7 @@ pub fn cairo_vm_execute(
     let mut writer_buffer = Vec::with_capacity(10_000);
 
     for (txn_index, txn_and_query_bit) in txns_and_query_bits.iter().enumerate() {
-        let class_info = match txn_and_query_bit.txn.clone() {
+        let class_info = match &txn_and_query_bit.txn {
             StarknetApiTransaction::Declare(_) => {
```
RafaelGranza, line 86: The transaction is only inspected for pattern matching, so cloning the entire value is unnecessary. Borrowing is enough here.

genesis/genesis.go
```diff
@@ -359,8 +359,12 @@ func executeTransactions(
 	}
 
 	blockInfo := vm.BlockInfo{Header: &genesisHeader}
-	executionResults, err := v.Execute(coreTxns, nil, []*felt.Felt{new(felt.Felt).SetUint64(1)},
-		&blockInfo, genesisState, true, true, true, true, false, false, false)
+	executionResults, err := v.BuildBlock(coreTxns, nil, []*felt.Felt{new(felt.Felt).SetUint64(1)},
```
rodrodros: I am aware of it is old code, can you change the old pattern of `new(felt.Felt).SetUint64(1)` to `new(felt.FromUint64(1))`

rodrodros: Also, when arguments are divided in multiple lines, we prefer one argument per line. We usually write function calls (and similarly method signatures) as:
```golang
// when total line char len < 100
func(a1 t1, a2 t2, ..., an tn)  { 

// when total line char len > 100 but arg line len < 100
func(
   a1 t1, a2 t2, ..., an tn,
) { 

// when arg line len > 100
func(
   a1 t1,
   a2 t2,
   ...
   an tn,
) {
```

I am aware this is old code but taking the opportunity to transmit Juno's style knowledge

rpc/v10/trace_test.go
```diff
@@ -733,19 +715,13 @@ func TestTraceBlockTransactions(t *testing.T) {
...
 			&vm.BlockInfo{Header: header},
 			gomock.Any(),
-			false,
-			false,
-			false,
-			true,
-			false,
-			false,
-			false).Return(vm.ExecutionResults{
+			vm.TraceOptions{}).Return(vm.ExecutionResults{
```
rodrodros: nit: split line

vm/vm.go
```diff
@@ -45,6 +45,40 @@ type CallResult struct {
 	ExecutionFailed bool
 }
 
+// ExecutionOptions carries every flag accepted by the underlying VM.
+// Used by the generic Execute method
+type ExecutionOptions struct {
```
rodrodros: There is a naming inconsistency where "Simluate" has "SimulateOptions" and "Trace" have "TraceOptions" but "Execute" has "Execut**ion**Options"

rodrodros: Could you explain to me why this type needs to be public, it seems like it could be private to me. No reason for the RPC to deal with the internals of `VM`

vm/vm.go
```diff
@@ -45,6 +45,40 @@ type CallResult struct {
...
+}
+
+// SimulateOptions carries the flags relevant to simulate / estimateFee.
+type SimulateOptions struct {
+	SkipChargeFee      bool
+	SkipValidate       bool
+	ErrOnRevert        bool
+	IsEstimateFee      bool
+	ReturnInitialReads bool
+}
```
rodrodros: I think Simulate and EstimateFee should be split but let me know what you think

vm/vm.go
rodrodros: If the aim is for this to be a type that includes all the other types, please use struct composition:
```golang
type T1 struct {
  T2
  T3
  T4
}
```
It will inherit all of the fields by default

vm/vm.go
```diff
@@ -444,6 +501,99 @@ func (v *vm) Execute(
 	}, nil
 }
 
+// Execute is the generic entry point exposing full RPC control.
+func (v *vm) Execute(
```
rodrodros: This can be private as well

rodrodros (commented):
Looks good, some nitpicks

vm/vm.go
RafaelGranza: I was keeping it public because I thought other use cases might come up, so having a more generic signature could be useful. But after our brief discussion, I believe the right approach is to create new wrappers as new use cases arise. I’ll make it private.

RafaelGranza: Fixing it

genesis/genesis.go
RafaelGranza: Sure! I'm in favor of the `boy scout rule`

vm/vm.go
RafaelGranza: Unfortunately, since the fields have the same name, it generates the `AmbiguousSelector` compilation error.

vm/vm.go
RafaelGranza: In our current implementation, the `starknet_simulateTransactions` RPC call accepts an optional `isEstimateFee` parameter, although this is not part of the spec.

We also have a separate implementation for `starknet_estimateFee`, which internally uses the `SimulationFlag` enum. Our previous design chose to keep both behaviors coupled.

I’m in favor of splitting the methods, but I would keep `isEstimateFee` as an option in `starknet_simulateTransactions` to avoid introducing a breaking change.

rodrodros: Makes sense 👍

rodrodros (approved):

PR #3607, opened 2026-05-06
fix(state): history writes only dirty nonce and class hash

MaksymMalicki:
In `State.flush`, the old code writes a nonce-history entry and a class-hash-history entry for every `stateObject` touched in a block — regardless of whether either field actually changed. So if a contract had only a storage change in a block, two redundant history entries were still emitted (same nonce, same classHash as the previous block).

To fix this, same pattern as in the `Revert` was applied - the history writing is detached from the `flush` method and utilises the state diff

### DIFF
--- a/core/state/state.go
+++ b/core/state/state.go
@@ -115,6 +115,7 @@ func (s *State) Update(
 
 		contract := newContractDeployed(*classHash, blockNum)
 		newObj := newStateObject(s, &addr, &contract)
+		newObj.classHashDirty = true
 		s.stateObjects[addr] = &newObj
 	}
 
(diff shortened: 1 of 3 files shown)


core/state/state.go
```diff
@@ -386,12 +387,21 @@ func (s *State) flush(
...
 
-				if err := WriteClassHashHistory(s.batch, &addr, blockNum, &obj.contract.ClassHash); err != nil {
-					return err
+				if obj.classHashDirty {
+					if err := WriteClassHashHistory(
+						s.batch,
+						&addr,
+						blockNum,
+						&obj.contract.ClassHash,
+					); err != nil {
```
rodrodros: split if statements please

core/state/state_test.go
```diff
@@ -699,6 +699,120 @@ func TestRevert(t *testing.T) {
...
 
+func TestStateUpdateWritesHistoryOnlyOnFieldChange(t *testing.T) {
+	addr := felt.NewUnsafeFromString[felt.Felt]("0xA")
+	classH0 := felt.NewUnsafeFromString[felt.Felt]("0xC0")
+	classH1 := felt.NewUnsafeFromString[felt.Felt]("0xC1")
+	storageKey := felt.NewUnsafeFromString[felt.Felt]("0x10")
+	val1 := felt.NewUnsafeFromString[felt.Felt]("0x42")
+	val2 := felt.NewUnsafeFromString[felt.Felt]("0x43")
+	nonce1 := felt.NewUnsafeFromString[felt.Felt]("0x1")
+	nonce2 := felt.NewUnsafeFromString[felt.Felt]("0x2")
```
rodrodros: Initialize from Uint64

rodrodros (commented):
This solution looks extremly hacky.  @MaksymMalicki can you explain the problem your trying to solve

rodrodros (approved):
Now this looks like the correct solution

rodrodros:
Codecov warning not too relevant, signaling error path

PR #3608, opened 2026-05-07
docs: fix links in staking-validator documentation

thiagodeev:


RafaelGranza (approved):
LGTM.

Since this endpoint point changed, I tested other endpoints starting with `https://docs.starknet.io/staking` and they're being migrated too.

While we're updating links from theses docs, I'd recommend changing `https://docs.starknet.io/staking/overview/#protocol_details` to `https://docs.starknet.io/learn/protocol/staking#protocol_details`. 

The old one still works, but it's already deprecated (redirected).

PR #3609, opened 2026-05-07
test(mempool): fix flaky TestRestoreMempool

RafaelGranza:
`TestRestoreMempool` relied on `time.Sleep(100ms)` to wait for the async `dbWriter` goroutine to persist transactions (**racy under load**).

Replaced with `require.EventuallyWithT` polling `LenDB()` until it reaches the expected count (timeout 2s, tick 10ms).
 
Went with polling instead of adding a sync method to prod code, since this need doesn't exist outside of tests.
 
Also removed a redundant sleep after `pool.Close()`. Given that `Close()` already drains the write channel via `wg.Wait()`.


brbrr (approved):
LGTM

PR #3612, opened 2026-05-07
fix(utils/log): logs signal properly where they were being called

### LINKED ISSUE #3610: Logs don't provide location information
Logs in Go give information of where they are being called, package, file name and line inside the file. This helps developer understand which actual part of the code is being log. 

The problem is that because we use a wrapper and we never do a call directly, all of our logs look like coming from the same place:

```bash
16:10:51.259 07/05/2026 +01:00	INFO	log/log.go:262	Stored Block	{"number": 928, "hash": "0x79f4...c515", "root": "0x66c7...5e37"}
16:10:51.270 07/05/2026 +01:00	INFO	log/log.go:262	Stored Block	{"number": 929, "hash": "0x5f99...c69a", "root": "0x67f2...9ae2"}
16:10:51.281 07/05/2026 +01:00	INFO	log/log.go:262	Stored Block	{"number": 930, "hash": "0x5ea4...3221", "root": "0x355f...469f"}
16:10:51.290 07/05/2026 +01:00	INFO	log/log.go:262	Stored Block	{"number": 931, "hash": "0x6855...c5e7", "root": "0x3eb6...9fac"}
```

Look that all of them are `log/log.go:262` which applies to all the info logs.

rodrodros:
Fix https://github.com/NethermindEth/juno/issues/3610

### DIFF
--- a/utils/log/log_test.go
+++ b/utils/log/log_test.go
@@ -388,6 +388,46 @@ func TestTrace(t *testing.T) {
 	})
 }
 
+func TestCallerInfo(t *testing.T) {
+	// Verifies that caller info points at the actual call site (this test
+	// file) rather than the ZapLogger wrapper in utils/log/log.go.
+	// Regression test for https://github.com/NethermindEth/juno/issues/3610.
+	logLevel := log.NewLevel(log.TRACE)
+
+	tests := map[string]func(l *log.ZapLogger){
+		"Debug": func(l *log.ZapLogger) { l.Debug("m") },
+		"Info":  func(l *log.ZapLogger) { l.Info("m") },
+		"Warn":  func(l *log.ZapLogger) { l.Warn("m") },
+		"Error": func(l *log.ZapLogger) { l.Error("m") },
+		"Trace": func(l *log.ZapLogger) { l.Trace("m") },
+	}
+
+	for name, emit := range tests {
+		t.Run(name, func(t *testing.T) {
+			var buf bytes.Buffer
+			logger, err := log.NewZapLogger(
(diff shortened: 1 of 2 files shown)


utils/log/log_test.go
```diff
@@ -388,6 +388,46 @@ func TestTrace(t *testing.T) {
...
+	// Regression test for https://github.com/NethermindEth/juno/issues/3610.
+	logLevel := log.NewLevel(log.TRACE)
+
+	tests := map[string]func(l *log.ZapLogger){
+		"Debug": func(l *log.ZapLogger) { l.Debug("m") },
+		"Info":  func(l *log.ZapLogger) { l.Info("m") },
+		"Warn":  func(l *log.ZapLogger) { l.Warn("m") },
+		"Error": func(l *log.ZapLogger) { l.Error("m") },
+		"Trace": func(l *log.ZapLogger) { l.Trace("m") },
+	}
```
RafaelGranza, line 403:
```suggestion
		"Trace": func(l *log.ZapLogger) { l.Trace("m") },
		"Tracew": func(l *log.ZapLogger) { l.Tracew("m") },
	}
```

Could we add `Tracew` to this map? 

The change in this PR was made directly inside `Tracew`. The `WithOptions(zap.AddCallerSkip(1))` line that got removed lived in its body.

It feels off to modify a function and not have a test exercising it.

RafaelGranza (approved):
LGTM

Only a small observation.

utils/log/log_test.go
rodrodros: Actually, TraceW is legacy code, I've removed it. Good catch!

PR #3614, opened 2026-05-07
docs: remove mentions of newdb snapshots

brbrr:
since 0.16.0, we dropped support for the old, non-cl db layout. With this, we can remove the `-newdb` variant, as it is not needed anymore.


docs/docs/snapshots.md
```diff
@@ -24,9 +23,6 @@ Thu Feb  5 15:35:59 CET 2026
 $curl -s -I -L https://juno-snapshots.nethermind.io/files/mainnet/latest | gawk -v IGNORECASE=1 '/^Content-Length/ { printf "%.2f GB\n", $2/1024/1024/1024 }'
 351.23 GB
```
rodrodros, line 25: It's a good time to update the sizes as well

rodrodros (approved):

rodrodros (approved):

PR #3616, opened 2026-05-08
fix(pruner): skip pruning when L1 head or chain height is missing

EgeCaner:
Skip pruning when the L1 head or chain height isn't yet in the DB. On a fresh node the sync and L1 services race during startup, so the first onNewL1Head / onNewBlock callback can fire before the counterpart has been persisted, treat that as a no-op instead of erroring out.

### DIFF
--- a/pruner/pruner.go
+++ b/pruner/pruner.go
@@ -16,6 +16,7 @@ package pruner
 
 import (
 	"context"
+	"errors"
 	"time"
 
 	"github.com/NethermindEth/juno/core"
@@ -162,6 +163,10 @@ func (p *Pruner) Run(ctx context.Context) error {
 func (p *Pruner) onNewBlock(ctx context.Context, block *core.Block) error {
 	l1Head, err := core.GetL1Head(p.database)
 	if err != nil {
+		if errors.Is(err, db.ErrKeyNotFound) {
+			// No L1 head yet — can't anchor the retention floor.
+			return nil
+		}
 		return err
 	}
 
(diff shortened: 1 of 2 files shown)


pruner/pruner.go
```diff
@@ -162,6 +163,10 @@ func (p *Pruner) Run(ctx context.Context) error {
 func (p *Pruner) onNewBlock(ctx context.Context, block *core.Block) error {
 	l1Head, err := core.GetL1Head(p.database)
 	if err != nil {
+		if errors.Is(err, db.ErrKeyNotFound) {
+			// No L1 head yet — can't anchor the retention floor.
+			return nil
+		}
 		return err
 	}
```
rodrodros, line 171: This makes sense when there is no L1 Head, but I don't think at the same time that they should be fully silenced

pruner/pruner.go
```diff
@@ -183,6 +188,10 @@ func (p *Pruner) onNewBlock(ctx context.Context, block *core.Block) error {
 func (p *Pruner) onNewL1Head(ctx context.Context, l1Head *core.L1Head) error {
 	chainHeight, err := core.GetChainHeight(p.database)
 	if err != nil {
+		if errors.Is(err, db.ErrKeyNotFound) {
+			// No chain height yet — nothing to prune against.
+			return nil
+		}
 		return err
```
rodrodros, line 195: Similarly here, why should the pruner service run if there is nothing to prune against. How can you receive an `L1head` while having no chain height?

rodrodros (commented):
I think this doesn't solve the issue completely. I believe there is a better way of implementing this

pruner/pruner.go
EgeCaner: We can log it with other level than error?

pruner/pruner.go
EgeCaner: L1 service and sync service is 2 different services, one syncs L2 blocks the other syncs L1 head, and they are not synchronized with each other,  any of them can arrive first.

EgeCaner: But logically if we don't have L1 head we cannot schedule prune and same for the empty chain

rodrodros (approved):

PR #3617, opened 2026-05-08
feat(sync): delta-aware pre_confirmed polling

thiagodeev:
A follow-up work of https://github.com/NethermindEth/juno/pull/3601, implementing another feeder improvement.

**Delta sync for `get_preconfirmed_block`** — the poller now echoes `knownBlockIdentifier` and `knownTransactionCount` on every tick so the feeder gateway can answer with a no-change marker, a delta of appended transactions/receipts/state diffs, or a full block when the round identifier no longer matches. The sync layer reconciles the three response shapes against the stored pre_confirmed instead of unconditionally re-decoding the whole block on every poll.
The PR also uses the `FeederAdapter` type from the `starknetdata` pkg to make the node work normally with both old and new versions of the feeder, switching to the new approach as soon as the new feeder updates are detected in the network.

Reference: https://demerzelsolutions.slack.com/archives/C03090TS3TK/p1777375140952279

Note: candidate txs are no longer supported with this new update, but since old sequencer still uses it, it can only be fully removed in the future.

TODO:
- address @todo's


starknetdata/feeder/feeder.go
```diff
@@ -73,16 +73,15 @@ func (f *Feeder) block(ctx context.Context, blockID string) (*core.Block, error)
...
-			return nil, fmt.Errorf("get signature for block %q: %v", blockID, sErr)
+		sig, err = f.client.Signature(ctx, blockID)
+		if err != nil {
+			return nil, fmt.Errorf("get signature for block %q: %v", blockID, err)
 		}
-		signature = sig.Signature
 	}
 
-	return sn2core.AdaptBlock(response, signature)
+	return sn2core.AdaptBlock(response, sig.Signature)
```
RafaelGranza: I guess this and other changes are due the rebase conflict, right?

core/pending/pending.go
```diff
@@ -96,6 +135,60 @@ func (p *PreConfirmed) Copy() *PreConfirmed {
...
+	nextHeader.EventsBloom = newBloomFilter
+
+	var eventCount uint64
+	for _, r := range receipts {
+		eventCount += uint64(len(r.Events))
+	}
+	nextHeader.EventCount += eventCount
+
+	nextHeader.TransactionCount += uint64(len(txs))
+	nextBlock.Header = &nextHeader
```
RafaelGranza, line 165: I'll never adapt with this, I'll always see a dangling pointer haha.

Now I know it works in Go due to the scape analysis.

adapters/sn2core/sn2core.go
```diff
@@ -545,27 +545,28 @@ func AdaptPreConfirmedBlock(
...
+		}
+		var stateDiff core.StateDiff
+		stateDiff, err = AdaptStateDiff(response.TransactionStateDiffs[i])
+		if err != nil {
+			return pending.PreConfirmed{}, err
 		}
+		txStateDiffs[preIdx] = &stateDiff
+		receipts[preIdx] = AdaptTransactionReceipt(response.Receipts[i])
+		eventCount += uint64(len(response.Receipts[i].Events))
+		preIdx++
```
RafaelGranza, line 569: Why changing the order?

utils/slices.go
```diff
@@ -41,6 +41,15 @@ func AnyOf[T comparable](e T, values ...T) bool {
...
 
+// Concat concatenates two slices into a new slice, preserving the order
+// of the elements (a first, then b).
+func Concat[T any](a, b []T) []T {
+	out := make([]T, len(a)+len(b))
+	copy(out, a)
+	copy(out[len(a):], b)
+	return out
+}
+
```
RafaelGranza: why not using `slices.Concat(...)`?

core/pending/pending.go
```diff
@@ -96,6 +135,60 @@ func (p *PreConfirmed) Copy() *PreConfirmed {
...
+	nextBlock.Transactions = utils.Concat(p.Block.Transactions, txs)
+	nextBlock.Receipts = utils.Concat(p.Block.Receipts, receipts)
```
RafaelGranza: Why you implemented a new `Concat` method only for this?
Why not using `slices.Concat()`

sync/pending_polling.go
```diff
@@ -319,21 +350,50 @@ func (s *Synchronizer) handlePreLatest(
...
+	update *pending.PreConfirmedUpdate,
 	stagedPreLatest *pending.PreLatest,
 ) {
-	pc.WithPreLatest(stagedPreLatest)
-	changed, err := s.StorePreConfirmed(pc)
+	var nextPreConfirmed *pending.PreConfirmed
+
+	switch update.Mode {
+	case pending.PreConfirmedNoChange:
+		nextPreConfirmed = s.preConfirmed.Load().Copy()
```
RafaelGranza: This is kind of a safe nit.

`s.preConfirmed.Load()` can be `nil` if `storeEmptyPreConfirmed` fails.

The server won't trigger this today, but a `nil` check before `.Copy()` would make it safer.

sync/pending_polling.go
```diff
@@ -319,21 +350,50 @@ func (s *Synchronizer) handlePreLatest(
...
+
+	case pending.PreConfirmedFull:
+		nextPreConfirmed = update.FullBlock
+
+	case pending.PreConfirmedDelta:
+		existing := s.preConfirmed.Load().Copy()
```
RafaelGranza: Same `nil` check before `.Copy()` here.

sync/pending_polling.go
```diff
@@ -45,8 +45,21 @@ func shouldPreservePreConfirmed(
...
+
+	if incomingB.Number == existingB.Number {
+		if incomingB.TransactionCount > existingB.TransactionCount {
+			return false
+		}
+		if incomingB.TransactionCount == existingB.TransactionCount &&
+			len(incomingPending.CandidateTxs) > len(existingPending.CandidateTxs) {
+			return false
+		}
+	}
```
RafaelGranza, line 60: When `Number`, `TransactionCount` and `len(CandidateTxs)` match but `BlockIdentifier` differs, we keep the old block.  On a round restart this usually self-heals once a new tx arrives.

But in the rare case where the new round has the same counts with different `txs`, we'd serve stale data. Adding `&& existingPending.BlockIdentifier == incomingPending.BlockIdentifier` would close that.

RafaelGranza: I would also add an test for this case if possible.

sync/pending_polling.go
RafaelGranza: Also, `ApplyDelta` does not mutate the receiver, so do we actually need a copy?

sync/pending_polling_test.go
```diff
@@ -349,141 +378,273 @@ func TestPollPendingData(t *testing.T) {
...
+							AppendCandidateTxs: []core.Transaction{newTx()},
+						}
+					case 5:
+						response = pending.PreConfirmedUpdate{
+							Mode:            pending.PreConfirmedNoChange,
+							BlockIdentifier: blockIdentifier,
+						}
+					}
+				default:
+					t.Fatal("unexpected number", number)
```
RafaelGranza: This runs on the poller goroutine, not the test one, so `t.Fatal` won't stop the test as expected.

`t.Errorf(...)` + `return` would be safer here.

RafaelGranza (commented):
LGTM. I only left a few nits.

I'll halt the approve because I'm waiting for the decision mechanism, which I think may be similar to the `includeSignature` one.

utils/slices.go
thiagodeev: I simply forgot to check whether it already existed haha. Thanks!

core/pending/pending.go
thiagodeev: Answered above

adapters/sn2core/sn2core.go
thiagodeev: Just a small simplification.
Instead of
```
if !X {
  ...
} else {
  ...
}
```
I changed to:

```
if X {
  ...
  continue
}
...
```

sync/pending_polling.go
thiagodeev: Correct, because before I didn't have the right idea about the blockIdentifier; I thought that the blockIdentifier would never change for a given block number. However, a few days ago, I talked with Ege and he explained to me that if the blockIdentifier changes for a given block number, we should replace our current block with the new one. This is noted in the PR description.
Thanks!

sync/pending_polling.go
thiagodeev: We have a case in the code (line 375) that assigned the `existing` var to the `nextPreConfirmed` var. So if we didn't copy, we would change the original PreConfirmed in the next steps

starknetdata/feeder/feeder_adapter.go
```diff
@@ -120,3 +121,37 @@ func (f *FeederAdapter) StateUpdateWithBlock(
...
+// the new "blockIdentifier" and "knownTransactionCount" arguments), it delegates
+// to [Feeder.PreConfirmedBlockByNumber]. Otherwise, it falls back to the old approach:
+// calling the client.PreConfirmedBlock endpoint with only the given block number.
+func (f *FeederAdapter) PreConfirmedBlockByNumber(
+	ctx context.Context,
+	blockNumber uint64,
+	blockIdentifier string,
+	knownTransactionCount uint64,
+) (pending.PreConfirmedUpdate, error) {
+	if f.isFeederUpdated.Load() {
```
RafaelGranza, line 136: Are we 100% starknet will ship both the updates together?
If not, we should not share the same atomic flag.

clients/feeder/feeder.go
```diff
@@ -358,14 +365,44 @@ func (c *Client) StateUpdateWithBlockAndSignature(
 	return doRequest[starknet.StateUpdateWithBlockAndSignature](ctx, c, queryURL)
 }
 
-func (c *Client) PreConfirmedBlock(ctx context.Context, blockNumber string) (*starknet.PreConfirmedBlock, error) {
+// PreConfirmedBlock fetches the pre_confirmed block at the given height.
```
RafaelGranza, line 370: This comment is duplicated in line 380. Is it correct?

core/pending/pending.go
```diff
@@ -96,6 +135,60 @@ func (p *PreConfirmed) Copy() *PreConfirmed {
...
+	}
+	nextHeader.EventsBloom = newBloomFilter
+
+	var eventCount uint64
+	for _, r := range receipts {
+		eventCount += uint64(len(r.Events))
+	}
+	nextHeader.EventCount += eventCount
+
+	nextHeader.TransactionCount += uint64(len(txs))
```
RafaelGranza, line 164: Are we sure we shouldn't count `candidateTxs`?

sync/pending_polling.go
```diff
@@ -254,14 +269,27 @@ func (s *Synchronizer) pollPreConfirmed(
...
-			preConfirmed, err := s.dataSource.PreConfirmedBlockByNumber(ctx, targetBlockNum)
+			var (
+				blockIdentifier              = "0x0"
+				knownTransactionCount uint64 = 0
+			)
+
+			currentPreConf := s.preConfirmed.Load()
+			if currentPreConf != nil {
+				blockIdentifier = currentPreConf.BlockIdentifier
+				knownTransactionCount = uint64(len(currentPreConf.Block.Transactions))
```
RafaelGranza, line 281: Are we sure we shouldn't count `candidateTxs`?

sync/pending_polling.go
```diff
@@ -319,21 +347,56 @@ func (s *Synchronizer) handlePreLatest(
...
+
+	currentPreConfirmed := s.preConfirmed.Load()
+	if currentPreConfirmed == nil {
+		s.logger.Debug("No pre_confirmed stored")
+		return
+	}
+
+	switch update.Mode {
+	case pending.PreConfirmedNoChange:
+		nextPreConfirmed = currentPreConfirmed.Copy()
```
RafaelGranza, line 370: If we have no changes? Would it be ok to early return? No changes nor copies required.


```suggestion
		return
```

sync/pending_polling.go
```diff
@@ -319,21 +347,56 @@ func (s *Synchronizer) handlePreLatest(
...
+	case pending.PreConfirmedFull:
+		nextPreConfirmed = update.FullBlock
+
+	case pending.PreConfirmedDelta:
+		existing := currentPreConfirmed.Copy()
+		if existing.BlockIdentifier != update.BlockIdentifier {
+			// Stored identifier drifted; drop. Next poll will return Full.
+			nextPreConfirmed = existing
+			break
+		}
```
RafaelGranza, line 381: Why copying de `currentPreConfirmed` if we don't need it?

```suggestion
		if currentPreConfirmed.BlockIdentifier != update.BlockIdentifier {
            return // Stored identifier drifted; drop. Next poll will return Full.
    }
    existing := currentPreConfirmed.Copy()
```

starknetdata/feeder/feeder_adapter.go
RafaelGranza, line 136: Could we add a fallback-path test here (`isFeederUpdated == false`), mirroring the `StateUpdateWithBlock` coverage?

clients/feeder/feeder.go
```diff
@@ -358,14 +365,44 @@ func (c *Client) StateUpdateWithBlockAndSignature(
...
+// the round identifier no longer matches. Set both to zero values to get a full block.
+func (c *Client) PreConfirmedBlockWithIdentifier(
+	ctx context.Context,
+	blockNumber string,
+	blockIdentifier string,
+	knownTransactionCount uint64,
+) (*starknet.PreConfirmedBlock, error) {
+	if blockIdentifier == "" {
+		blockIdentifier = "0x0"
+	}
```
RafaelGranza, line 398: Nit: the `"0x0"` sentinel also lives at `sync/pending_polling.go:273`, so the magic constant is duplicated. Extracting a named const (e.g. `PreConfirmedInitialIdentifier`) and referencing it from both sites could a smarter move.

starknetdata/feeder/feeder_adapter.go
thiagodeev: Yes, they guaranteed that. Ref: https://demerzelsolutions.slack.com/archives/C03090TS3TK/p1779736613873929?thread_ts=1777538734.003539&cid=C03090TS3TK

clients/feeder/feeder.go
thiagodeev: Yeah. Both are the same endpoints; the new one only has some new params. Since I was adding some docs to the new one, I decided to put a small doc in the old as well

core/pending/pending.go
thiagodeev: Yes. New feeder updates don't contain candidate txs. Ref: https://demerzelsolutions.slack.com/archives/C03090TS3TK/p1779725141246769?thread_ts=1777538734.003539&cid=C03090TS3TK

Also, I've pushed a new commit that completely removes 'candidate'  from Delta path. Other parts of the code we still need to support it, but since the Delta path is exclusively for the new feature, it's safe to remove them from it

sync/pending_polling.go
thiagodeev: Yes (https://github.com/NethermindEth/juno/pull/3617#discussion_r3334512511)

sync/pending_polling.go
thiagodeev: We have a small edge case of the case where `handlePreLatest` ([ref](https://github.com/NethermindEth/juno/blob/thiagodeev/feeder-preconfirmed-optimization/sync/pending_polling.go#L342)) receives a new PreLatest but it fails to store an empty PreConfirmed block containing it, but it will still return and assign it to `stagedPreLatest`. For this case, it's important that even though there are no changes in the PreConfirmed, we keep executing `handlePreConfirmed`, so that we can append the new `stagedPreLatest` to our current PreConfirmed ([ref](https://github.com/NethermindEth/juno/blob/thiagodeev/feeder-preconfirmed-optimization/sync/pending_polling.go#L390)).
Makes sense for you?

sync/pending_polling.go
thiagodeev: Related https://github.com/NethermindEth/juno/pull/3617#discussion_r3335424782

clients/feeder/feeder.go
thiagodeev: Agree, thanks!

starknetdata/feeder/feeder_adapter.go
thiagodeev: For sure. As I said on Slack, I'm not finished yet

adapters/sn2core/sn2core.go
```diff
@@ -609,7 +609,54 @@ func AdaptPreConfirmedBlock(
...
+	txStateDiffs = make([]*core.StateDiff, preConfirmedTxCount)
+
+	preIdx := 0
+	for i := range len(response.Transactions) {
+		txns[preIdx], err = AdaptTransaction(&response.Transactions[i])
+		if err != nil {
+			return nil, nil, nil, err
+		}
+		var stateDiff core.StateDiff
+		stateDiff, err = AdaptStateDiff(response.TransactionStateDiffs[i])
```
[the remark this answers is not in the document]
thiagodeev: IMO, not relevant.
Delta responses are by definition never empty. If the sequencer breaks its response contract, everything else breaks

sync/pending_polling.go
```diff
@@ -254,14 +270,27 @@ func (s *Synchronizer) pollPreConfirmed(
...
+			var (
+				blockIdentifier              = feeder.PreConfirmedBlankIdentifier
+				knownTransactionCount uint64 = 0
+			)
+
+			currentPreConf := s.preConfirmed.Load()
+			if currentPreConf != nil {
+				blockIdentifier = currentPreConf.BlockIdentifier
+				knownTransactionCount = uint64(len(currentPreConf.Block.Transactions))
+			}
```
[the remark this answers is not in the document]
thiagodeev: Not relevant IMO.
As said, if a new PreConfirmed block is stored between block number load and currentPreConfirmed load, and the node requests block X with a different identifier, the sequencer will return the full block, and the code will handle it properly

starknetdata/feeder/feeder_adapter.go
```diff
@@ -120,3 +121,37 @@ func (f *FeederAdapter) StateUpdateWithBlock(
...
+	response, err := f.client.PreConfirmedBlock(ctx, strconv.FormatUint(blockNumber, 10))
+	if err != nil {
+		return pending.PreConfirmedUpdate{}, err
+	}
+
+	adaptedPreConfirmed, err := sn2core.AdaptPreConfirmedBlock(response, blockNumber)
+	if err != nil {
+		return pending.PreConfirmedUpdate{}, err
+	}
+	adaptedPreConfirmed.BlockIdentifier = blockIdentifier
```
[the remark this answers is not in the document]
thiagodeev: Not relevant IMO.
This code (FeederAdapter) will only live for a month, no need to be that strict for "future maintenance"

RafaelGranza (approved):
LGTM

adapters/sn2core/sn2core.go
rodrodros: I don't follow the logic, wouldn't that have been the same as 

```
if !X {
 ...
 continue
}
...
```

The order change is unnecessary as far as I can see but no harm done. Function still legible 👍 (maybe a bit better)

adapters/sn2core/sn2core.go
```diff
@@ -609,7 +609,54 @@ func AdaptPreConfirmedBlock(
 		Transactions: txns,
 		Receipts:     receipts,
 	}
-	return pending.NewPreConfirmed(adaptedBlock, &stateUpdate, txStateDiffs, candidateTxs), nil
+	preConfirmed := pending.NewPreConfirmed(adaptedBlock, &stateUpdate, txStateDiffs, candidateTxs)
+	preConfirmed.BlockIdentifier = response.BlockIdentifier
```
rodrodros, line 613: This doesn't look good, calling the constructor for then setting a property.

core/pending/pending.go
```diff
@@ -65,6 +66,43 @@ type PreConfirmed struct {
...
+	PreConfirmedNoChange PreConfirmedUpdateMode = iota
+	// PreConfirmedDelta means new transactions/receipts/state diffs have been
+	// appended since the caller's known transaction count and should be merged
+	// onto the existing stored pre_confirmed.
+	PreConfirmedDelta
+	// PreConfirmedFull means the server's pre_confirmed is for a different
+	// round and the caller should discard existing data and replace it with
+	// Full.
+	PreConfirmedFull
+)
```
rodrodros, line 90: Wouldn't you say that every function called `PreConfirmedSmth` is kind of extra, why not call it just `Smth`.

core/pending/pending.go
```diff
@@ -65,6 +66,43 @@ type PreConfirmed struct {
 	CandidateTxs          []core.Transaction
 	// Optional field, exists if pre_confirmed is N+2 when latest is N
 	PreLatest *PreLatest
+	// BlockIdentifier is an identifier returned by the feeder gateway
+	// that uniquely identifies the current round of the pre_confirmed block.
+	// It is used to negotiate delta-sync responses on subsequent polls.
+	BlockIdentifier string
```
rodrodros, line 72: As far as I can see BlockIdentifier is not a string but always a smaller than a uint64. Why are you storing a string?

core/pending/pending.go
```diff
@@ -65,6 +66,43 @@ type PreConfirmed struct {
...
+
+// PreConfirmedUpdate is the result of a delta-aware pre_confirmed fetch. It
+// is the boundary type between data sources and the sync layer reconciler.
+// Only the fields appropriate for Mode are populated.
+type PreConfirmedUpdate struct {
+	Mode            PreConfirmedUpdateMode
+	BlockIdentifier string
+
+	// FullBlock is set when Mode == PreConfirmedFull.
+	FullBlock *PreConfirmed
```
rodrodros, line 100: Why BlockIdentifier is defined twice, once here and nother inside PreConfirmed?

core/pending/pending.go
```diff
@@ -65,6 +66,43 @@ type PreConfirmed struct {
...
+	// PreConfirmedFull means the server's pre_confirmed is for a different
+	// round and the caller should discard existing data and replace it with
+	// Full.
+	PreConfirmedFull
+)
+
+// PreConfirmedUpdate is the result of a delta-aware pre_confirmed fetch. It
+// is the boundary type between data sources and the sync layer reconciler.
+// Only the fields appropriate for Mode are populated.
+type PreConfirmedUpdate struct {
```
rodrodros, line 95: This description is non descriptive and explained in unnecessary complex vocabulary. What is a sync layer reconciler? Only the fileds appropriate for Mode are populated <- why does this sentece adds?

Update to:
```
// PreConfirmedUpdate represents the three types of responses we can have when fetching for 
// a pre-confirmed block:
//  1. NoChange the struct is empty
//  2. PreConfirmedDelta smth smth
//  3. smth smth

File: core/pending/pending.go [hunk 25]
```diff
@@ -65,6 +66,43 @@ type PreConfirmed struct {
...
+	BlockIdentifier string
+
+	// FullBlock is set when Mode == PreConfirmedFull.
+	FullBlock *PreConfirmed
+
+	// Append* are set when Mode == PreConfirmedDelta.
+	AppendTransactions []core.Transaction
+	AppendReceipts     []*core.TransactionReceipt
+	AppendStateDiffs   []*core.StateDiff
 }
```
rodrodros, line 106: I am sorry but this whole design doesn't makes any sense to me. Why would we have a type that is trying to be three types at the same time. This is in a clear need for polymorphism through interfaces. You need to define one interface which is `PreConfirmedUpdate` and have then have the three types defined instead that implements this interface.

adapters/sn2core/sn2core.go
```diff
@@ -609,7 +609,54 @@ func AdaptPreConfirmedBlock(
...
+		return nil, nil, nil, errors.New("nil preconfirmed block")
+	}
+
+	isInvalidPayloadSizes := len(response.Transactions) != len(response.TransactionStateDiffs) ||
+		len(response.Transactions) != len(response.Receipts)
+	if isInvalidPayloadSizes {
+		return nil, nil, nil, errors.New(
+			"invalid sizes of transactions, state diffs and receipts",
+		)
+	}
```
rodrodros, line 637: Why don't you add information to the error such as which size were expected, of what is the miss-match.

adapters/sn2core/sn2core.go
```diff
@@ -609,7 +609,54 @@ func AdaptPreConfirmedBlock(
...
+	response *starknet.PreConfirmedBlock,
+) (
+	txns []core.Transaction,
+	receipts []*core.TransactionReceipt,
+	txStateDiffs []*core.StateDiff,
+	err error,
+) {
+	if response == nil {
+		return nil, nil, nil, errors.New("nil preconfirmed block")
+	}
```
rodrodros, line 629: Remove, tired of mentioning the same thing

adapters/sn2core/sn2core.go
```diff
@@ -609,7 +609,54 @@ func AdaptPreConfirmedBlock(
...
+// AdaptPreConfirmedDelta extracts the per-transaction core types from a delta
+// 'get_preconfirmed_block' response.
+func AdaptPreConfirmedDelta(
+	response *starknet.PreConfirmedBlock,
+) (
+	txns []core.Transaction,
+	receipts []*core.TransactionReceipt,
+	txStateDiffs []*core.StateDiff,
+	err error,
+) {
```
rodrodros, line 626: I don't see any adaptation, just three types being returned? Shouldn't it return the DeltaUpdated type?

core/pending/pending.go
```diff
@@ -96,6 +134,57 @@ func (p *PreConfirmed) Copy() *PreConfirmed {
...
+// ApplyDelta returns a new PreConfirmed by appending the given transactions,
+// receipts, and merging the state diffs onto the
+// existing PreConfirmed. It does not modify the receiver.
+func (p *PreConfirmed) ApplyDelta(
+	txs []core.Transaction,
+	receipts []*core.TransactionReceipt,
+	txStateDiffs []*core.StateDiff,
+	blockIdentifier string,
+) *PreConfirmed {
+	next := *p
```
rodrodros, line 146: Question: Why does this function returns a new PreConfirmed instead of re-using the current one. And if it is necessary to create a new one, why does it returns a reference to it, instead of returning it by value?

core/pending/pending.go
```diff
@@ -96,6 +134,57 @@ func (p *PreConfirmed) Copy() *PreConfirmed {
...
+	nextBlock.Transactions = slices.Concat(p.Block.Transactions, txs)
+	nextBlock.Receipts = slices.Concat(p.Block.Receipts, receipts)
+
+	next.Block = &nextBlock
+
+	newStateDiff := core.EmptyStateDiff()
+	newStateDiff.Merge(p.StateUpdate.StateDiff)
+	for _, sd := range txStateDiffs {
+		newStateDiff.Merge(sd)
+	}
```
rodrodros, line 175: Re-mergeing all the state diffs again seems awfully inneficient. At least start from a copy of the other state diff

rodrodros (changes requested):
This PR is wrong on several levels. 

Architecture wise:
- One type containing three types instead of using Go's Polymorphism. 
- Constructors that are called and then fields being initialized separately
- Two overlapping types sharing the same field
- Adapter functions not building the right type instead 3 arrays. 

Style wise is also wrong with many style decisions that have been discussed numerous times and not being followed here.

adapters/sn2core/sn2core.go
thiagodeev: Since this is the only case in the code base where we need to set `BlockIdentifier` after the constructor initialization, I thought was better to manually set it only here instead of modifying the original constructor and changing all its occurences across the code base just to set blank identifiers.
WDYT?

adapters/sn2core/sn2core.go
thiagodeev: Sorry, but what are you tired of?
Most of the code of this function was taken from the existing `AdaptPreConfirmedBlock`, included in the PR #2954, which was approved and merged by you. I tried my best to mirror the existing behavior, so that it would be easier to review, and we could cover the same possible edge cases covered by the `AdaptPreConfirmedBlock` func.
That being said, I would prefer to leave this condition to avoid any behavior mismatch. WDYT?

adapters/sn2core/sn2core.go
thiagodeev: I agree about the name, I think I was just trying to follow the same 'adapt' patter since we are in the `adapters` pkg. WDYT about `ExtractDeltaFromPreConfirmed`?
Also, thinking about it now, WDYT of moving this logic to the `starknetdata` pkg since it's only used there so far?

adapters/sn2core/sn2core.go
thiagodeev: This part was just copied from https://github.com/NethermindEth/juno/blob/thiagodeev/feeder-preconfirmed-optimization/adapters/sn2core/sn2core.go#L522-L528. Ok, I'll add this

core/pending/pending.go
thiagodeev: The spec provided to us described this field as a string, so I just followed the spec
https://demerzelsolutions.slack.com/archives/C03090TS3TK/p1777375140952279

core/pending/pending.go
thiagodeev: I somehow agree, but it's a common Go pattern to name enum this way (to avoid name conflict, have the name fully self-describing it...etc), and we have some cases like that in our codebase as well. Do you want me to leave them just as `NoChange`, `Delta`, and `Full`?

core/pending/pending.go
thiagodeev: Indeed BlockIdentifier is now part of the PreConfirmed structure (and it's used for some specific PreConfirmed checks ([ref](https://github.com/NethermindEth/juno/blob/thiagodeev/feeder-preconfirmed-optimization/sync/pending_polling.go#L54)), but so that we don't need to have a full `PreConfirmed` block instantiated for 'delta' and 'nochange' responses just to store the BlockIdentifier, I created the same field for `PreConfirmedUpdate` as well.
Also, it's some sort of a way to separate both abstractions (`PreConfirmedUpdate` is an intermediate state just carrying the information)

core/pending/pending.go
thiagodeev: Because we still need to filter whether this new block containing the delta should be stored (look at the sync `StorePreConfirmed` method).
About returning by value: yeah, good point. I'll change it

core/pending/pending.go
thiagodeev: But that's exactly what we do here :)
`newStateDiff` starts as an empty StateDiff (line 171). Next, we merge the StateDiff of the current PreConfirmed block `p` into it (line 172), then we merge each one of the txs diffs into it (line 174)

core/pending/pending.go
thiagodeev: It's okay, it's another valid way to solve the problem :)
Do you want me to refactor the code and implement it this way?

adapters/sn2core/sn2core.go
rodrodros: I don't think it is good architecture, why don't you just returned the Delta response directly?

adapters/sn2core/sn2core.go
rodrodros: This function is called on feeder.go after a point that is is clear that `response` is not nil leaving this check unnecessary and code cluttering. 

If a year ago (on #2954) I've approve it, it was under a year ago circumstances and context and I can have very well missed it. 

Regardless of the time it was, a week or a year, unnecessary code, such an`if` condition that will never trigger is better to be deleted. The reason I've said I am tired of repeating myself because I've been very vocal about it in the last two weeks dailies and PR reviews.

core/pending/pending.go
rodrodros: This is a code smell, this should've pointed you that your design is not correct

core/pending/pending.go
rodrodros: Felts are also also strings, it doesn't me we don't wrap it on types. The block-identifier strings seems to be following small hex (lesser than u64) approach

core/pending/pending.go
rodrodros: After previous discussion this enum shouldn't exist

adapters/sn2core/sn2core.go
rodrodros: Yes, it would've been a trigger refactor but it is better to have the constructor include it some way.

thiagodeev:
Merged through https://github.com/NethermindEth/juno/pull/3705#discussion_r3367163082

PR #3618, opened 2026-05-08
refactor(state): use backward compatible class encoding

MaksymMalicki:
1. The new state's `Class` bucket encoding now matches the deprecated state's: WriteClass / GetClass use plain `encoder.Marshal`/`Unmarshal `of the whole `DeclaredClassDefinition` instead of the `[8B At][CBOR(Class)]` binary prefix — the fixed-offset `.At` had no consumer that actually benefited from it. Both implementations now write byte-identical Class entries, so the upcoming deprecated→new state migration can skip class re-encoding entirely.
2. As a small consistency follow-on, `core/deprecatedstate/state.go`'s `putClass` and `Class` are routed through the existing (until now unused) `core.WriteClass` / `core.GetClass` accessors. Direct batch access for the `Class` bucket are gone from both sides; the original control flow in `putClass` (check ErrKeyNotFound → write, else propagate) is preserved exactly.

### DIFF
--- a/core/deprecatedstate/state.go
+++ b/core/deprecatedstate/state.go
@@ -16,7 +16,6 @@ import (
 	"github.com/NethermindEth/juno/core/felt"
 	"github.com/NethermindEth/juno/core/trie"
 	"github.com/NethermindEth/juno/db"
-	"github.com/NethermindEth/juno/encoder"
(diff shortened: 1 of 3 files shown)


thiagodeev (approved):
LGTM

PR #3620, opened 2026-05-09
feat(cmd): separate all Juno flags into categories

rodrodros:
New usage looks:

```bash
➜  juno3 git:(rdr/try-better-cmd-help) ✗ ./build/juno --help
Starknet client implementation in Go.

Usage:
  juno [flags]
  juno [command]

Available Commands:
  compile-sierra Compile a Sierra class to CASM via stdin/stdout.
  completion  Generate the autocompletion script for the specified shell
  db          Database related operations
  genp2pkeypair Generate private key pair for p2p.
  help        Help about any command

HTTP RPC Flags:
      --disable-rpc-batch-requests     Disables handling of batched RPC requests.
      --http                           Enables the HTTP RPC server on the default port and interface.
      --http-host string               The interface on which the HTTP RPC server will listen for requests. (default "localhost")
      --http-port uint16               The port on which the HTTP server will listen for requests. (default 6060)
      --rpc-call-max-gas uint          Maximum number of Sierra gas to be executed in starknet_call requests (default 100000000)
      --rpc-call-max-steps uint        Maximum number of steps to be executed in starknet_call requests (default 4000000)
      --rpc-cors-enable                Enable CORS on RPC endpoints
      --rpc-max-block-scan uint        Maximum number of blocks scanned in single starknet_getEvents call (default 18446744073709551615)
      --rpc-request-timeout duration   Maximum time for an RPC request to complete. (default 1m0s)

WebSocket RPC Flags:
      --disable-received-txn-stream   The starknet_subscribeNewTransactions WebSocket API allows users to subscribe to new transactions. By default, it streams transactions that have been accepted on L2. Users can optionally provide a set of finality statuses to be notified about, including transactions from canonical blocks, blocks with softer finality guarantees such as pre-confirmed and pre-latest, as well as transactions not yet part of any block such as received and candidate. When subscribers select the RECEIVED status, they will be notified about transactions that have been submitted through this node — these transactions are local to the node and are not sourced from the network. When this flag is enabled, the node will no longer notify subscribers about transactions submitted through it.
      --ws                            Enables the WebSocket RPC server on the default port.
      --ws-host string                The interface on which the WebSocket RPC server will listen for requests. (default "localhost")
      --ws-port uint16                The port on which the WebSocket server will listen for requests. (default 6061)

Network & L1 Flags:
      --disable-l1-verification   Disables L1 verification since an Ethereum node is not provided.
      --eth-node string           WebSocket endpoint of the Ethereum node. To verify the correctness of the L2 chain, Juno must connect to an Ethereum node and parse events in the Starknet contract.
      --network Network           Options: mainnet, sepolia, sepolia-integration. (default mainnet)

Sync & Polling Flags:
      --preconfirmed-poll-interval duration   Sets how frequently pre_confirmed block will be updated(0s will disable fetching of pre_confirmed block). (default 500ms)
      --prelatest-poll-interval duration      Sets polling interval for pre-latest block updates. (0s will disable polling). (default 1s)
      --readiness-block-tolerance uint        Maximum blocks behind latest for /ready endpoints to return 200 OK (default 6)
      --remote-db string                      gRPC URL of a remote Juno node

Gateway Flags:
      --gw-api-key string    API key for gateway endpoints to avoid throttling
      --gw-timeouts string   Timeouts for requests made to the gateway. Can be specified in three ways:
                             - Single value (e.g. '5s'): After each failure, the timeout will increase dynamically.
                             - Comma-separated list (e.g. '5s,10s,20s'): Each value will be used in sequence after failures.
                             - Single value with trailing comma (e.g. '5s,'): Uses a fixed timeout without dynamic adjustment. (default "5s")

Pruning Flags:
      --prune-mode uint[=0]   Enables block-data and state-history pruning. Pruning is disabled by default; passing this flag (with or without a value) turns it on. The value is the size of the retention window in blocks, counted back from the latest L1-verified head:
                                --prune-mode      same as --prune-mode=0; prune up to the L1 head
                                --prune-mode=N    keep blocks in (l1_head - N, l2_head], prune below
                              Blocks at or above the L2 head are always kept. The floor is anchored on the L1-verified head — never on the local L2 head — so pruned blocks are reorg-safe. RPC remains fully functional for any block inside the retention window; requests targeting blocks below the floor fail because their data has been deleted. Pruning is irreversible: data deleted under a small window cannot be recovered without re-syncing. Changing this value across restarts is safe: the window grows or shrinks accordingly. Growth is gradual — pruning pauses until the L1 head advances enough to reach the new floor.

Logging Flags:
      --colour --colour=false   Use --colour=false command to disable colourized outputs (ANSI Escape Codes). (default true)
      --log-json                Use JSON encoding for log output.
      --log-level string        Options: trace, debug, info, warn, error. (default "info")

Logs HTTP Update Endpoint Flags:
      --http-update-host string   The interface on which the log level and gateway timeouts HTTP server will listen for requests. (default "localhost")
      --http-update-port uint16   The port on which the log level and gateway timeouts HTTP server will listen for requests.

Metrics & Profiling Flags:
      --metrics               Enables the Prometheus metrics endpoint on the default port.
      --metrics-host string   The interface on which the Prometheus endpoint will listen for requests. (default "localhost")
      --metrics-port uint16   The port on which the Prometheus endpoint will listen for requests. (default 9090)
      --pprof                 Enables the pprof endpoint on the default port.
      --pprof-host string     The interface on which the pprof HTTP server will listen for requests. (default "localhost")
      --pprof-port uint16     The port on which the pprof HTTP server will listen for requests. (default 6062)

Database Flags:
      --db-cache-size uint                 Determines the amount of memory (in megabytes) allocated for caching data in the database. (default 1024)
      --db-compaction-concurrency string   DB compaction concurrency range. Format: N (lower=1, upper=N) or M,N (lower=M, upper=N). Default: 1,GOMAXPROCS/2
      --db-compression string              Database compression profile. Options: zstd, snappy, minlz. Use zstd for low storage. (default "zstd")
      --db-max-handles int                 A soft limit on the number of open files that can be used by the DB (default 1024)
      --db-memtable-count uint             Determines the number of memtables the database can queue before stalling writes. (default 2)
      --db-memtable-size uint              Determines the amount of memory (in MBs) allocated for database memtables. (default 256)
      --db-path string                     Location of the database files. (default "/Users/rodro/Documents/Nethermind/starknet/juno3/juno")

Transaction Cache Flags:
      --submitted-transactions-cache-entry-ttl duration   Time-to-live for each entry in the submitted transactions cache (default 5m0s)
      --submitted-transactions-cache-size uint            Maximum number of entries in the submitted transactions cache (default 10000)

VM & Compilation Flags:
      --max-concurrent-compilations uint   Maximum concurrent Sierra compilations. (default 8)
      --max-vm-queue uint                  Maximum number for requests to queue after reaching max-vms before starting to reject incoming requests (default 72)
      --max-vms uint                       Maximum number for VM instances to be used for RPC calls concurrently (default 36)
      --versioned-constants-file string    Use custom versioned constants from provided file

Custom Network Flags:
      --cn-core-contract-address string   Custom network core contract address.
      --cn-feeder-url string              Custom network feeder URL.
      --cn-gateway-url string             Custom network gateway URL.
      --cn-l1-chain-id string             Custom network L1 chain id.
      --cn-l2-chain-id string             Custom network L2 chain id.
      --cn-name string                    Custom network name.
      --cn-unverifiable-range 0,100       Custom network range of blocks to skip hash verifications (e.g. 0,100).

P2P (experimental) Flags:
      --p2p                      EXPERIMENTAL: Enables p2p server.
      --p2p-addr string          EXPERIMENTAL: Specify p2p listening source address as multiaddr.  Example: /ip4/0.0.0.0/tcp/7777
      --p2p-feeder-node          EXPERIMENTAL: Run juno as a feeder node which will only sync from feeder gateway and gossip the new blocks to the network.
      --p2p-peers string         EXPERIMENTAL: Specify list of p2p peers split by a comma. These peers can be either Feeder or regular nodes.
      --p2p-private-key string   EXPERIMENTAL: Hexadecimal representation of a private key on the Ed25519 elliptic curve.
      --p2p-public-addr string   EXPERIMENTAL: Specify p2p public address as multiaddr.  Example: /ip4/35.243.XXX.XXX/tcp/7777

Sequencer Flags:
      --seq-block-time uint       Time to build a block, in seconds (default 60)
      --seq-disable-fees          Skip charge fee for sequencer execution
      --seq-enable                Enables sequencer mode of operation
      --seq-genesis-file string   Path to the genesis file

gRPC Flags:
      --grpc               Enable the HTTP gRPC server on the default port.
      --grpc-host string   The interface on which the gRPC server will listen for requests. (default "localhost")
      --grpc-port uint16   The port on which the gRPC server will listen for requests. (default 6064)

Plugins & Misc Flags:
      --config string        The YAML configuration file.
  -h, --help                 help for juno
      --new-state            EXPERIMENTAL: Use the new state package implementation
      --plugin-path string   Path to the plugin .so file

Other Flags:
  -v, --version   version for juno

Use "juno [command] --help" for more information about a command.

```

### DIFF
--- a/cmd/juno/juno.go
+++ b/cmd/juno/juno.go
@@ -433,7 +433,9 @@ func NewCmd(config *node.Config, run func(*cobra.Command, []string) error) *cobr
 	junoCmd.Flags().Uint(callMaxStepsF, defaultCallMaxSteps, callMaxStepsUsage)
 	junoCmd.Flags().Uint(callMaxGasF, defaultCallMaxGas, callMaxGasUsage)
 	junoCmd.Flags().Duration(rpcRequestTimeoutF, defaultRPCRequestTimeout, rpcRequestTimeoutUsage)
-	junoCmd.Flags().Bool(disableRPCBatchRequestsF, defaultDisableRPCBatchRequests, disableRPCBatchRequestsUsage)
+	junoCmd.Flags().Bool(
+		disableRPCBatchRequestsF, defaultDisableRPCBatchRequests, disableRPCBatchRequestsUsage,
+	)
 	setCategory(junoCmd, catHTTPRPC,
 		httpF, httpHostF, httpPortF, corsEnableF,
 		rpcMaxBlockScanF, callMaxStepsF, callMaxGasF,
@@ -444,7 +446,9 @@ func NewCmd(config *node.Config, run func(*cobra.Command, []string) error) *cobr
 	junoCmd.Flags().Bool(wsF, defaultWS, wsUsage)
 	junoCmd.Flags().String(wsHostF, defaultHost, wsHostUsage)
 	junoCmd.Flags().Uint16(wsPortF, defaultWSPort, wsPortUsage)
-	junoCmd.Flags().Bool(disableReceivedTxnStreamF, defaultDisableReceivedTxnStream, disableReceivedTxnStreamUsage)
+	junoCmd.Flags().Bool(
+		disableReceivedTxnStreamF, defaultDisableReceivedTxnStream, disableReceivedTxnStreamUsage,
+	)
 	setCategory(junoCmd, catWebSocket, wsF, wsHostF, wsPortF, disableReceivedTxnStreamF)
 
 	// --- Network & L1 ---
@@ -455,10 +459,16 @@ func NewCmd(config *node.Config, run func(*cobra.Command, []string) error) *cobr
 	setCategory(junoCmd, catNetwork, networkF, ethNodeF, disableL1VerificationF)
 
 	// --- Sync & Polling ---
-	junoCmd.Flags().Duration(preLatestPollIntervalF, defaultPreLatestPollInterval, preLatestPollIntervalUsage)
-	junoCmd.Flags().Duration(preConfirmedPollIntervalF, defaultPreConfirmedPollInterval, preConfirmedPollIntervalUsage)
+	junoCmd.Flags().Duration(
+		preLatestPollIntervalF, defaultPreLatestPollInterval, preLatestPollIntervalUsage,
+	)
+	junoCmd.Flags().Duration(
+		preConfirmedPollIntervalF, defaultPreConfirmedPollInterval, preConfirmedPollIntervalUsage,
+	)
 	junoCmd.Flags().String(remoteDBF, defaultRemoteDB, remoteDBUsage)
-	junoCmd.Flags().Uint(readinessBlockToleranceF, defaultReadinessBlockTolerance, readinessBlockToleranceUsage)
+	junoCmd.Flags().Uint(
+		readinessBlockToleranceF, defaultReadinessBlockTolerance, readinessBlockToleranceUsage,
+	)
 	setCategory(junoCmd, catSyncPolling,
 		preLatestPollIntervalF, preConfirmedPollIntervalF,
 		remoteDBF, readinessBlockToleranceF,
@@ -502,7 +512,9 @@ func NewCmd(config *node.Config, run func(*cobra.Command, []string) error) *cobr
 	junoCmd.Flags().String(dbPathF, defaultDBPath, dbPathUsage)
 	junoCmd.Flags().Uint(dbCacheSizeF, defaultCacheSizeMb, dbCacheSizeUsage)
 	junoCmd.Flags().Int(dbMaxHandlesF, defaultMaxHandles, dbMaxHandlesUsage)
-	junoCmd.Flags().String(dbCompactionConcurrencyF, defaultDBCompactionConcurrency, dbCompactionConcurrencyUsage)
+	junoCmd.Flags().String(
+		dbCompactionConcurrencyF, defaultDBCompactionConcurrency, dbCompactionConcurrencyUsage,
+	)
 	junoCmd.Flags().Uint(dbMemtableSizeF, defaultDBMemtableSize, dbMemtableSizeUsage)
 	junoCmd.Flags().Uint(dbMemtableCountF, defaultDBMemtableCount, dbMemtableCountUsage)
 	junoCmd.Flags().String(dbCompressionF, defaultDBCompression, dbCompressionUsage)
@@ -518,13 +530,19 @@ func NewCmd(config *node.Config, run func(*cobra.Command, []string) error) *cobr
 		defaultSubmittedTransactionsCacheEntryTTL,
 		submittedTransactionsCacheEntryTTL,
 	)
-	setCategory(junoCmd, catTxCache, submittedTransactionsCacheSizeF, submittedTransactionsCacheEntryTTLF)
+	setCategory(junoCmd, catTxCache,
+		submittedTransactionsCacheSizeF, submittedTransactionsCacheEntryTTLF,
+	)
 
 	// --- VM & Compilation ---
 	junoCmd.Flags().Uint(maxVMsF, uint(defaultMaxVMs), maxVMsUsage)
 	junoCmd.Flags().Uint(maxVMQueueF, 2*uint(defaultMaxVMs), maxVMQueueUsage)
-	junoCmd.Flags().Uint(maxConcurrentCompilationsF, defaultMaxConcurrentCompilations, maxConcurrentCompilationsUsage)
-	junoCmd.Flags().String(versionedConstantsFileF, defaultVersionedConstantsFile, versionedConstantsFileUsage)
+	junoCmd.Flags().Uint(
+		maxConcurrentCompilationsF, defaultMaxConcurrentCompilations, maxConcurrentCompilationsUsage,
+	)
+	junoCmd.Flags().String(
+		versionedConstantsFileF, defaultVersionedConstantsFile, versionedConstantsFileUsage,
+	)
 	setCategory(junoCmd, catVMCompile,
 		maxVMsF, maxVMQueueF, maxConcurrentCompilationsF, versionedConstantsFileF,
 	)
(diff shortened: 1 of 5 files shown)


cmd/juno/usage_test.go
```diff
@@ -0,0 +1,70 @@
...
+
+	juno "github.com/NethermindEth/juno/cmd/juno"
+	"github.com/NethermindEth/juno/node"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+const flagCategoryAnnotation = "juno_category"
```
brbrr: Probably it could be exported from usage.go?

docs/docs/_config-options.md
```diff
@@ -1,72 +1,76 @@
...
+| `network` | `mainnet` | Options: mainnet, sepolia, sepolia-integration |
+| `new-state` | `false` | EXPERIMENTAL: Use the new state package implementation |
+| `node.PruneModeFlag` | `uint64(0)` | Enables block-data and state-history pruning. Pruning is disabled by default; passing this flag (with or without a value) turns it on. The value is the size of the retention window in blocks, counted back from the latest L1-verified head:\n  --prune-mode      same as --prune-mode=0; prune up to the L1 head\n  --prune-mode=N    keep blocks in (l1_head - N, l2_head], prune below\nBlocks at or above the L2 head are always kept. The floor is anchored on the L1-verified head — never on the local L2 head — so pruned blocks are reorg-safe. RPC remains fully functional for any block inside the retention window; requests targeting blocks below the floor fail because their data has been deleted. Pruning is irreversible: data deleted under a small window cannot be recovered without re-syncing. Changing this value across restarts is safe: the window grows or shrinks accordingly. Growth is gradual — pruning pauses until the L1 head advances enough to reach the new floor |
```
brbrr: Something went wrong with parsing config name and default value.

cmd/juno/usage.go
```diff
@@ -0,0 +1,232 @@
...
+		// don't tag their flags. Render the standard flat list for them.
+		b.WriteString("\nFlags:\n")
+		b.WriteString(strings.TrimRight(cmd.LocalFlags().FlagUsages(), " \t\n"))
+		b.WriteByte('\n')
+		return
+	}
+
+	// orderedCategories controls render order. catOther is appended after these if
+	// non-empty, so a forgotten annotation is visible rather than silently dropped.
+	orderedCategories := []string{
```
brbrr: nit: we manually build the same slice in tests (expectedCategoryHeaders), and it feels like the order shouldn't change, so maybe we could define it a a global/dedicated generator function which tests could reuse?

docs/docs/_config-options.md
```diff
@@ -1,72 +1,76 @@
 <!-- This file is generated automatically. Any manual modifications will be overwritten. -->
 
+### Other
```
brbrr: Here, we are still using an old representation. Seems like config generation failed?

cmd/juno/juno.go
```diff
@@ -424,96 +424,183 @@ func NewCmd(config *node.Config, run func(*cobra.Command, []string) error) *cobr
...
-	junoCmd.Flags().String(grpcHostF, defaultHost, grpcHostUsage)
-	junoCmd.Flags().Uint16(grpcPortF, defaultGRPCPort, grpcPortUsage)
-	junoCmd.Flags().Uint(maxVMsF, uint(defaultMaxVMs), maxVMsUsage)
-	junoCmd.Flags().Uint(maxVMQueueF, 2*uint(defaultMaxVMs), maxVMQueueUsage)
-	junoCmd.Flags().String(remoteDBF, defaultRemoteDB, remoteDBUsage)
-	junoCmd.Flags().Uint(rpcMaxBlockScanF, defaultRPCMaxBlockScan, rpcMaxBlockScanUsage)
+	junoCmd.Flags().Bool(pprofF, defaultPprof, pprofUsage)
+	junoCmd.Flags().String(pprofHostF, defaultHost, pprofHostUsage)
+	junoCmd.Flags().Uint16(pprofPortF, defaultPprofPort, pprofPortUsage)
+	setCategory(junoCmd, catObservability,
```
brbrr: I'm not really sure if pprof goes together with metrics, as pprof instruments the binary, which affects node performance etc.

docs/docs/_config-options.md
brbrr: Ah, it seems that generate-config.js uses `juno.go` from main branch, not the local version. That's probably the reason.

brbrr: What is the plan for this file? I don't think it was regenerated with the fixes added in `generate-config.js`

brbrr (changes requested):

docs/docs/_config-options.md
rodrodros: It will re-generate once we merge to main, wdym what's the plan?

rodrodros: Either way, I've updated them which makes sense 👍 . Note that Juno docs don't use them as source. Let me see if I can make this a check

brbrr: My initial question was related to the fact that the changes to the file were commited, but contained some issues. 

IIRC, we use dedicated branch for docs, where we build/commit/deploy based on the state from main. 

Since _config-options.md is generated, maybe we could just remove it from main, to not create confusion? (Not in this pr though)

brbrr (approved):

docs/docs/_config-options.md
rodrodros: I think it is better for them to stay, because it is confortable to have a reproducible doc in your environment without to much hassle (no needing to push to main / other branch) or modifying your scripts to know what's on the docs. 

I lean towards the other of option of having a checks that verifies that if _config_option.md gets generated always matches the source repo

PR #3628, opened 2026-05-11
fix: pruner off by one

EgeCaner:
### Fix off-by-one in the pruner retention floor.

  - The floor was computed as `pivot - N + 1`, so we kept N-1 blocks below the pivot instead of N. With --prune-mode=0 this collapsed to deleting the pivot block itself — and during catch-up where L1 is ahead of L2, that meant wiping everything the node had synced.
  - Drop the +1 so the pivot is always retained: --prune-mode=0 keeps just the pivot, --prune-mode=N keeps N blocks below it.

### DIFF
--- a/cmd/juno/juno.go
+++ b/cmd/juno/juno.go
@@ -253,17 +253,18 @@ const (
 	pruneModeUsage                 = "Enables block-data and state-history pruning. Pruning is " +
 		"disabled by default; passing this flag (with or without a value) turns " +
 		"it on. The value is the size of the retention window in blocks, counted " +
-		"back from the latest L1-verified head:\n" +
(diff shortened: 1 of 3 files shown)


rodrodros (commented):
We are a missing a regression test case, i.e., a case that would fail if we were to undo this change.

Let me know if that is being tested, but so far, I only see the current tests being updated

EgeCaner:
> We are a missing a regression test case, i.e., a case that would fail if we were to undo this change.
> 
> Let me know if that is being tested, but so far, I only see the current tests being updated

The existing tests already fail under a revert of this change, the off-by-one shows up as a numeric mismatch in their asserted oldest/count values. Happy to add a dedicated boundary test if you'd still prefer it called out explicitly.

rodrodros (approved):

PR #3634, opened 2026-05-11
fix(node): catch up L1 head before history pruning migration

RafaelGranza:
Fixes the `key not found` crash when starting Juno with `--prune-mode` on a node where the L1 head hasn't been written yet.

Why it crashed: the pruning migration needs the L1 head to know which old blocks are safe to delete. The L1 head is written by the L1 client, which only starts AFTER migrations. So on a partially-synced DB the migration reads something nothing has written yet, and dies.

Fix: fetch the L1 head once from `--eth-node` before the migration runs. Same endpoint that's already required for L1 verification. No new flag.

Also rejects `--prune-mode` + `--disable-l1-verification` up front since pruning needs L1 finality (was a silent footgun before).

### DIFF
--- a/l1/l1.go
+++ b/l1/l1.go
@@ -167,6 +167,21 @@ func (c *Client) Run(ctx context.Context) error {
 	return c.watchL1StateUpdates(ctx)
 }
 
+// CatchUpL1Head runs the startup phase (chain ID check + backward
+// catch-up scan that writes the L1 head via setL1Head) without entering
+// the live subscription loop. Intended for callers that need an L1 head
+// present in the database before subsequent steps — notably the history
+// pruning migration, which runs before the regular L1 service starts.
+// The Client closes its Subscriber on return; do not reuse the instance
+// afterward.
+func (c *Client) CatchUpL1Head(ctx context.Context) error {
+	defer c.l1.Close()
+	if err := c.checkChainID(ctx); err != nil {
+		return err
+	}
+	return c.catchUpL1HeadUpdates(ctx)
+}
+
 func (c *Client) watchL1StateUpdates(ctx context.Context) error {
 	buffer := 128
 
--- a/l1/l1.go
+++ b/l1/l1.go
@@ -167,13 +167,9 @@ func (c *Client) Run(ctx context.Context) error {
 	return c.watchL1StateUpdates(ctx)
 }
 
-// CatchUpL1Head runs the startup phase (chain ID check + backward
-// catch-up scan that writes the L1 head via setL1Head) without entering
-// the live subscription loop. Intended for callers that need an L1 head
-// present in the database before subsequent steps — notably the history
-// pruning migration, which runs before the regular L1 service starts.
-// The Client closes its Subscriber on return; do not reuse the instance
-// afterward.
+// CatchUpL1Head verifies the chain ID then writes the L1 head to the
+// database, without entering the live subscription loop. Closes the
+// underlying Subscriber on return; the Client must not be reused.
 func (c *Client) CatchUpL1Head(ctx context.Context) error {
 	defer c.l1.Close()
 	if err := c.checkChainID(ctx); err != nil {
--- a/node/migration.go
+++ b/node/migration.go
@@ -51,15 +51,9 @@ func migrateIfNeeded(
 			return fmt.Errorf("deprecated migration failed: %w", err)
 		}
 
-		// The history pruning migration reads the L1 head to compute the
-		// reorg-safe cutoff. On a fresh sync the L1 client (which writes the
-		// head) only starts as a regular service AFTER this point, so a
-		// pruning-enabled node restarted mid-sync would fail with a cryptic
-		// "key not found". Bootstrap the head via a one-shot catch-up here,
-		// reusing the operator's existing --eth-node so no extra config is
-		// required. Skipped if the head is already on disk from a previous run.
+		// Make sure there is an available L1 head before starting pruning migration
 		if config.Prune {
-			if err := bootstrapL1HeadIfMissing(ctx, database, config, chain, logger); err != nil {
+			if err := fetchL1HeadIfMissing(ctx, database, config, chain, logger); err != nil {
 				return fmt.Errorf("bootstrap L1 head for pruning: %w", err)
 			}
 		}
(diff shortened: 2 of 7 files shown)


node/migration.go
```diff
@@ -32,26 +35,40 @@ func registerMigrations(cfg *Config) *migration.Registry {
...
+		// The history pruning migration reads the L1 head to compute the
+		// reorg-safe cutoff. On a fresh sync the L1 client (which writes the
+		// head) only starts as a regular service AFTER this point, so a
+		// pruning-enabled node restarted mid-sync would fail with a cryptic
+		// "key not found". Bootstrap the head via a one-shot catch-up here,
+		// reusing the operator's existing --eth-node so no extra config is
+		// required. Skipped if the head is already on disk from a previous run.
+		if config.Prune {
+			if err := bootstrapL1HeadIfMissing(ctx, database, config, chain, logger); err != nil {
+				return fmt.Errorf("bootstrap L1 head for pruning: %w", err)
```
rodrodros: Can we call it `fetchL1HeadIfMissing`?

node/migration.go
```diff
@@ -32,26 +35,40 @@ func registerMigrations(cfg *Config) *migration.Registry {
...
 			return fmt.Errorf("deprecated migration failed: %w", err)
 		}
 
+		// The history pruning migration reads the L1 head to compute the
+		// reorg-safe cutoff. On a fresh sync the L1 client (which writes the
+		// head) only starts as a regular service AFTER this point, so a
+		// pruning-enabled node restarted mid-sync would fail with a cryptic
+		// "key not found". Bootstrap the head via a one-shot catch-up here,
+		// reusing the operator's existing --eth-node so no extra config is
+		// required. Skipped if the head is already on disk from a previous run.
```
rodrodros: I think this comment can be reduced a bit to just: "Make sure there is an available L1 head before starting pruning migration since it is required by it". I believe the extra text is redundant

node/migration.go
```diff
@@ -73,3 +90,42 @@ func migrateIfNeeded(
...
+	database db.KeyValueStore,
+	config *Config,
+	chain *blockchain.Blockchain,
+	logger log.StructuredLogger,
+) error {
+	if _, err := core.GetL1Head(database); err == nil {
+		return nil
+	} else if !errors.Is(err, db.ErrKeyNotFound) {
+		return err
+	}
```
rodrodros: Separate `if` clauses please (nit)

node/migration.go
```diff
@@ -73,3 +90,42 @@ func migrateIfNeeded(
...
+
+	logger.Info("Bootstrapping L1 head for pruning migration via --eth-node catch-up")
```
rodrodros: Call it "fetch" and let's reduce the last part of the sentence which doesn't add value: "Fetching the L1 head before running the prune migration", or something similar. It is not necessary to annotate how it is being done

node/migration.go
```diff
@@ -73,3 +90,42 @@ func migrateIfNeeded(
...
+	client, err := newL1Client(config.EthNode, config.Metrics, chain, logger)
+	if err != nil {
+		return err
```
rodrodros: Wrap the error: "creating a new L1 client"

node/migration.go
```diff
@@ -73,3 +90,42 @@ func migrateIfNeeded(
...
+	}
+	if err := client.CatchUpL1Head(ctx); err != nil {
+		return err
+	}
+
+	if _, err := core.GetL1Head(database); err != nil {
+		if errors.Is(err, db.ErrKeyNotFound) {
+			return errors.New("catch-up completed without a finalised LogStateUpdate; " +
+				"retry once L1 has a finalised Starknet state update")
```
rodrodros: This is a very edge case right 🤔, good idea. Let's improve the wording, again you don't need to specify how you do it. Write something along the lines of "couldn't find a finalized Starknet state update, the specify L1 head doesn't exist".

node/node.go
```diff
@@ -170,6 +170,15 @@ func New(cfg *Config, version string, logLevel *log.Level) (*Node, error) {
...
 	}
 
+	// History pruning needs an L1-finalised cutoff to know which blocks are
+	// safe to drop. The cutoff is established by the L1 client, so disabling
+	// L1 verification makes pruning unsafe — reject the combination up front
+	// instead of crashing later when the migration cannot find an L1 head.
+	if cfg.Prune && cfg.DisableL1Verification {
+		return nil, errors.New("--prune-mode requires L1 verification; " +
+			"remove --disable-l1-verification or disable --prune-mode")
+	}
```
rodrodros, line 180: Can you please add tests?

node/migration.go
```diff
@@ -73,3 +90,42 @@ func migrateIfNeeded(
...
+// service runs on startup. Returns an explicit error if the catch-up
+// completes without finding any finalised LogStateUpdate, so the operator
+// sees a clear cause instead of a downstream "key not found".
+func bootstrapL1HeadIfMissing(
+	ctx context.Context,
+	database db.KeyValueStore,
+	config *Config,
+	chain *blockchain.Blockchain,
+	logger log.StructuredLogger,
+) error {
```
rodrodros, line 96: Can you please add tests?

node/migration_test.go
```diff
@@ -0,0 +1,42 @@
...
+func TestFetchL1HeadIfMissing_SkipsL1FetchWhenHeadPresent(t *testing.T) {
+	database := memory.New()
+	require.NoError(t, core.WriteL1Head(database, &core.L1Head{
+		BlockNumber: 1,
+		BlockHash:   felt.NewRandom[felt.Felt](),
+		StateRoot:   felt.NewRandom[felt.Felt](),
+	}))
+
+	cfg := &Config{EthNode: ""}
+	require.NoError(t, fetchL1HeadIfMissing(t.Context(), database, cfg, nil, log.NewNopZapLogger()))
```
rodrodros, line 27: How do you L1 Head was correctly fetched, other than there was no error?

node/migration_test.go
```diff
@@ -0,0 +1,42 @@
+package node
```
rodrodros, line 3: It has to be `package node_test`

node/migration_test.go
RafaelGranza: Sure, my bad

node/migration_test.go
RafaelGranza: `fetchL1HeadIfMissing` and `migrateIfNeeded` are private, so `node_test` can't reach them.
  Same happens in `metrics_test.go` in this package, isn't it OK?

rodrodros: No we are trying to move away from this approach. Past Juno way of writing codes doesn't necessarily reflects Juno's current way of writing code.

RafaelGranza: According to what I searched, using a file like`export_test.go` is the idiomatic way for exposing private symbols to external `_test` packages. We don't change the API, and the `_test.go` suffix keeps it out of production builds.
I implemented it, WDYT?

l1/l1.go
```diff
@@ -167,6 +167,21 @@ func (c *Client) Run(ctx context.Context) error {
...
 
+// CatchUpL1Head runs the startup phase (chain ID check + backward
+// catch-up scan that writes the L1 head via setL1Head) without entering
+// the live subscription loop. Intended for callers that need an L1 head
+// present in the database before subsequent steps — notably the history
+// pruning migration, which runs before the regular L1 service starts.
+// The Client closes its Subscriber on return; do not reuse the instance
+// afterward.
+func (c *Client) CatchUpL1Head(ctx context.Context) error {
+	defer c.l1.Close()
```
EgeCaner, line 174: Do we need to constraint this function to run once and close the client? Why not let caller to decide when to close?

node/migration_test.go
```diff
@@ -0,0 +1,49 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func TestFetchL1HeadIfMissing_SkipsL1FetchWhenHeadPresent(t *testing.T) {
+	database := memory.New()
+	want := &core.L1Head{
+		BlockNumber: 1,
+		BlockHash:   felt.NewRandom[felt.Felt](),
+		StateRoot:   felt.NewRandom[felt.Felt](),
+	}
```
rodrodros, line 23: nit: update `want` to `expected` is the most common go key word

node/migration_test.go
```diff
@@ -0,0 +1,49 @@
...
+	require.NoError(t, core.WriteL1Head(database, want))
+
+	logger := log.NewNopZapLogger()
+	cfg := &node.Config{EthNode: ""}
+	require.NoError(t, node.FetchL1HeadIfMissing(t.Context(), database, cfg, nil, logger))
+
+	got, err := core.GetL1Head(database)
```
rodrodros: `got` would be received. Not too important, but is nice to keep the naming consistent as much as possible across test files

EgeCaner (approved):
Dropped one comment, rest looks good to me!

l1/l1.go
RafaelGranza: Considered decoupling the close, but callers forgetting to close would leak the WebSocket connection, and there's no current use case that benefits from caller-managed lifecycle.

Kept the auto-close inside `CatchUpL1Head` and documented the contract in the doc comment.

Happy to revisit if a real use case for caller-managed lifecycle appears.

node/migration_test.go
```diff
@@ -0,0 +1,49 @@
+// Uses package node (not node_test) to test unexported helpers directly.
+// It avoids changing the API or production code just for testing
+package node
```
RafaelGranza, line 3: @rodrodros I reverted to `package node`. 

The `export_test.go` workaround was getting convoluted, and the other paths to test through public API would either change production signatures (exporting helpers, returning errors from `n.Run`) or add log capture infrastructure to production code just for tests. 

None felt worth it for two small private helpers. I added a short comment at explaining the choice.

PR #3638, opened 2026-05-13
fix(pruner):running filter rebuilt

EgeCaner:
## Problem

On a pruning node, the running event filter's lazy loader rebuilt from the most recent persisted anchor by walking back through aggregated filter windows and then filling forward block-by-block via `GetBlockHeaderByNumber`. Both of those reads can land below `OldestRetainedBlock`.

  ## Summary

  - Split running-event-filter initialization into a pluggable `RunningEventFilterInitializer`. `blockchain.New` defaults to `core.InitializeRunningEventFilter`; nodes started with `--prune` wire in `pruner.InitializeRunningEventFilter` via the new `WithRunningEventFilterInitializer` option.
  - Add a same-window fast path to both initializers: when the persisted snapshot's `next` is still inside the current aggregated window, resume it in place and fill `[next, latest]` instead of rebuilding from the prior anchor.
  - Pruner initializer clamps the fill range to `OldestRetainedBlock` and bounds the backward anchor scan by `floorAligned`, so we never read pruned headers or pruned aggregated filters. When no
  anchor is found, the rebuilt window is rooted at `floorAligned` with `[floorAligned, floor)` left as zero bitmap rows (those receipts are pruned, so any hit there would be unactionable anyway).
  - Replace `panic` paths in `BlocksForKeysInto`, `OnReorg`, and `Write` with wrapped error returns; wrap errors throughout `Insert` and the initializers with window/block context to make rebuild failures diagnosable.

  ## Tests
  - Extend `core/running_event_filter_test.go` for the same-window resume path.
  - Add `pruner/running_event_filter_test.go` covering floor-clamped fill, missing-anchor rebuild at floor, and below-floor snapshot handling.

### DIFF
--- a/blockchain/blockchain.go
+++ b/blockchain/blockchain.go
@@ -89,8 +89,9 @@ type Blockchain struct {
 
 // options holds configuration for constructing a Blockchain.
 type options struct {
-	listener     EventListener
-	stateVersion bool
+	listener                EventListener
+	stateVersion            bool
+	runningFilterInitialize core.RunningEventFilterInitializer
 }
 
 // Option is a functional option for configuring Blockchain options.
@@ -108,10 +109,21 @@ func WithNewState(enabled bool) Option {
 	return func(o *options) { o.stateVersion = enabled }
 }
 
+// WithRunningEventFilterInitializer overrides the initializer used to lazily
+// bring up the running event filter. Defaults to
+// [core.InitializeRunningEventFilter]; pass pruner.InitializeRunningEventFilter
+// on pruning nodes.
+func WithRunningEventFilterInitializer(initialize core.RunningEventFilterInitializer) Option {
+	return func(o *options) {
+		o.runningFilterInitialize = initialize
+	}
+}
+
 func New(database db.KeyValueStore, network *networks.Network, opts ...Option) *Blockchain {
 	o := options{
-		listener:     &SelectiveListener{},
-		stateVersion: false,
+		listener:                &SelectiveListener{},
+		stateVersion:            false,
+		runningFilterInitialize: core.InitializeRunningEventFilter,
 	}
 	for _, opt := range opts {
 		opt(&o)
@@ -123,7 +135,7 @@ func New(database db.KeyValueStore, network *networks.Network, opts ...Option) *
 	}
 	cachedFilters.WithFallback(fallback)
 
-	runningFilter := core.NewRunningEventFilterLazy(database)
+	runningFilter := core.NewRunningEventFilterLazy(database, o.runningFilterInitialize)
 
 	return &Blockchain{
 		database:      database,
--- a/core/running_event_filter.go
+++ b/core/running_event_filter.go
@@ -83,7 +83,7 @@ func (f *RunningEventFilter) Insert(
 	defer f.mu.Unlock()
 
 	if err := f.ensureInit(); err != nil {
-		return fmt.Errorf("couldn't initialize the running event filter: %w", err)
+		return err
 	}
 
 	if err := f.inner.Insert(bloom, blockNumber); err != nil {
@@ -115,7 +115,7 @@ func (f *RunningEventFilter) BlocksForKeys(keys [][]byte) *bitset.BitSet {
 	defer f.mu.RUnlock()
 
 	if err := f.ensureInit(); err != nil {
-		panic(fmt.Sprintf("Couldn't initialised the running event filter. Error: %v", err))
+		panic(err)
 	}
 
 	return f.inner.BlocksForKeys(keys)
@@ -127,7 +127,7 @@ func (f *RunningEventFilter) BlocksForKeysInto(keys [][]byte, out *bitset.BitSet
 	defer f.mu.RUnlock()
 
 	if err := f.ensureInit(); err != nil {
-		return fmt.Errorf("couldn't initialize the running event filter: %w", err)
+		return err
 	}
 
 	return f.inner.BlocksForKeysInto(keys, out)
@@ -139,7 +139,7 @@ func (f *RunningEventFilter) FromBlock() uint64 {
 	defer f.mu.RUnlock()
 
 	if err := f.ensureInit(); err != nil {
-		panic(fmt.Sprintf("Couldn't initialised the running event filter. Error: %v", err))
+		panic(err)
 	}
 
 	return f.inner.fromBlock
@@ -151,7 +151,7 @@ func (f *RunningEventFilter) ToBlock() uint64 {
 	defer f.mu.RUnlock()
 
 	if err := f.ensureInit(); err != nil {
-		panic(fmt.Sprintf("Couldn't initialised the running event filter. Error: %v", err))
+		panic(err)
 	}
 
 	return f.inner.toBlock
@@ -163,7 +163,7 @@ func (f *RunningEventFilter) NextBlock() uint64 {
 	defer f.mu.RUnlock()
 
 	if err := f.ensureInit(); err != nil {
-		panic(fmt.Sprintf("Couldn't initialised the running event filter. Error: %v", err))
+		panic(err)
 	}
 
 	return f.next
@@ -176,7 +176,7 @@ func (f *RunningEventFilter) Clone() *RunningEventFilter {
 	defer f.mu.RUnlock()
 
 	if err := f.ensureInit(); err != nil {
-		panic(fmt.Sprintf("Couldn't initialised the running event filter. Error: %v", err))
+		panic(err)
 	}
 
 	innerCopy := f.inner.Clone()
@@ -191,7 +191,7 @@ func (f *RunningEventFilter) InnerFilter() *AggregatedBloomFilter {
 	defer f.mu.RUnlock()
 
 	if err := f.ensureInit(); err != nil {
-		panic(fmt.Sprintf("Couldn't initialised the running event filter. Error: %v", err))
+		panic(err)
 	}
 
 	return f.inner
@@ -203,7 +203,7 @@ func (f *RunningEventFilter) OnReorg() error {
 	defer f.mu.Unlock()
 
 	if err := f.ensureInit(); err != nil {
-		return fmt.Errorf("couldn't initialize the running event filter: %w", err)
+		return err
 	}
 
 	currRangeStart := f.inner.FromBlock()
@@ -234,7 +234,7 @@ func (f *RunningEventFilter) Write() error {
 	defer f.mu.Unlock()
 
 	if err := f.ensureInit(); err != nil {
-		return fmt.Errorf("couldn't initialize the running event filter: %w", err)
+		return err
 	}
 
 	return WriteRunningEventFilter(f.database, f)
--- a/core/running_event_filter.go
+++ b/core/running_event_filter.go
@@ -127,7 +127,7 @@ func (f *RunningEventFilter) BlocksForKeysInto(keys [][]byte, out *bitset.BitSet
 	defer f.mu.RUnlock()
 
 	if err := f.ensureInit(); err != nil {
-		panic(fmt.Sprintf("Couldn't initialised the running event filter. Error: %v", err))
+		return fmt.Errorf("ensureInit: %w", err)
 	}
 
 	return f.inner.BlocksForKeysInto(keys, out)
@@ -201,7 +201,7 @@ func (f *RunningEventFilter) OnReorg() error {
 	defer f.mu.Unlock()
 
 	if err := f.ensureInit(); err != nil {
-		panic(fmt.Sprintf("Couldn't initialised the running event filter. Error: %v", err))
+		return fmt.Errorf("ensureInit: %w", err)
 	}
 
 	currRangeStart := f.inner.FromBlock()
@@ -232,7 +232,7 @@ func (f *RunningEventFilter) Write() error {
 	defer f.mu.Unlock()
 
 	if err := f.ensureInit(); err != nil {
-		panic(fmt.Sprintf("Couldn't initialised the running event filter. Error: %v", err))
+		return fmt.Errorf("ensureInit: %w", err)
 	}
 
 	return WriteRunningEventFilter(f.database, f)
--- a/core/running_event_filter.go
+++ b/core/running_event_filter.go
@@ -183,7 +183,9 @@ func (f *RunningEventFilter) Clone() *RunningEventFilter {
 	return NewRunningEventFilterHot(f.database, &innerCopy, f.next)
 }
 
-// InnerFilter returns a deep copy of the current AggregatedBloomFilter window.
+// InnerFilter returns the current AggregatedBloomFilter window.
+// The returned pointer aliases internal state; mutations are visible to f.
+// Use [RunningEventFilter.Clone] for an independent copy.
 func (f *RunningEventFilter) InnerFilter() *AggregatedBloomFilter {
 	f.mu.RLock()
 	defer f.mu.RUnlock()
--- a/core/running_event_filter.go
+++ b/core/running_event_filter.go
@@ -37,7 +37,7 @@ func (f *RunningEventFilter) ensureInit() error {
 		f.lazyOnce.Do(func() {
 			filter, err := f.initialize(f.database)
 			if err != nil {
-				f.initErr = fmt.Errorf("InitializeRunningEventFilter: %w", err)
+				f.initErr = fmt.Errorf("couldn't initialize the running event filter: %w", err)
 				return
 			}
 			f.inner = filter.inner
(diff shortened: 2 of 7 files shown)


core/running_event_filter.go
```diff
@@ -184,17 +201,21 @@ func (f *RunningEventFilter) OnReorg() error {
 	defer f.mu.Unlock()
 
 	if err := f.ensureInit(); err != nil {
-		panic(fmt.Sprintf("Couldn't initialised the running event filter. Error: %v", err))
+		return fmt.Errorf("ensureInit: %w", err)
```
rodrodros: write errors in natural language please. Suggestion: "couldn't initialize the running event filter: %w" (similar to what was before).

core/running_event_filter.go
```diff
@@ -211,96 +232,109 @@ func (f *RunningEventFilter) Write() error {
...
```
rodrodros: Similar

core/running_event_filter.go
```diff
@@ -211,96 +232,109 @@ func (f *RunningEventFilter) Write() error {
...
+	latest uint64,
+) error {
+	for blockNum := from; blockNum <= latest; blockNum++ {
+		header, err := GetBlockHeaderByNumber(database, blockNum)
+		if err != nil {
+			return fmt.Errorf("GetBlockHeaderByNumber %d: %w", blockNum, err)
+		}
+		if err := rf.Insert(header.EventsBloom, blockNum); err != nil {
+			return fmt.Errorf("Insert block %d: %w", blockNum, err)
+		}
```
rodrodros, line 303: Errors in natural language

core/running_event_filter.go
```diff
@@ -211,96 +232,109 @@ func (f *RunningEventFilter) Write() error {
...
+func fillRunningEventFilter(
+	database db.KeyValueStore,
+	rf *RunningEventFilter,
+	from,
+	latest uint64,
+) error {
+	for blockNum := from; blockNum <= latest; blockNum++ {
+		header, err := GetBlockHeaderByNumber(database, blockNum)
+		if err != nil {
+			return fmt.Errorf("GetBlockHeaderByNumber %d: %w", blockNum, err)
```
rodrodros: Suggestion: "getting block header by number %d: %w"

core/running_event_filter.go
```diff
@@ -211,96 +232,109 @@ func (f *RunningEventFilter) Write() error {
...
+		}
+		if err := rf.Insert(header.EventsBloom, blockNum); err != nil {
+			return fmt.Errorf("Insert block %d: %w", blockNum, err)
```
rodrodros: Suggestion: "inserting block %d in events bloom %w"

core/running_event_filter.go
```diff
@@ -211,96 +232,109 @@ func (f *RunningEventFilter) Write() error {
...
-
-		/// Check previous range
-		if errors.Is(err, db.ErrKeyNotFound) {
-			rangeStartAlligned -= int64(NumBlocksPerFilter)
-			lastStoredFilterRangeEnd -= NumBlocksPerFilter
-			continue
+		if !errors.Is(err, db.ErrKeyNotFound) {
+			return nil, fmt.Errorf("rebuild: scan for anchor at [%d,%d]: %w",
+				rangeStartAligned, lastStoredFilterRangeEnd, err)
 		}
```
rodrodros: When we wrap errors in Go we don't give context about the current function but about the function that it is being called on, it is for the caller of `rebuildRunningEventFilter` to give the context that the error was received while rebuilding. Suggestion: "scanning for a block at range [%d, %d]: %w". Substituting `anchor` because it makes the error confusing

core/running_event_filter.go
```diff
@@ -211,96 +232,109 @@ func (f *RunningEventFilter) Write() error {
...
-		}
-
-		err = runningFilter.Insert(header.EventsBloom, continueFrom)
-		if err != nil {
-			return nil, err
-		}
+	filter := NewAggregatedFilter(continueFrom)
+	runningFilter := NewRunningEventFilterHot(database, &filter, continueFrom)
+	if err := fillRunningEventFilter(database, runningFilter, continueFrom, latest); err != nil {
+		return nil, fmt.Errorf("rebuild (latest=%d, anchorFound=%t): %w", latest, found, err)
```
rodrodros: "filling running event filter (latest=%d, anchorFound=%t): %w"

core/running_event_filter.go
```diff
@@ -211,96 +232,109 @@ func (f *RunningEventFilter) Write() error {
...
+		next := stored.NextBlock()
+		// Caught up — return the snapshot as-is.
+		if next == latest+1 {
+			return NewRunningEventFilterHot(database, stored.InnerFilter(), next), nil
+		}
+		// Same-window gap — resume the snapshot and fill in place.
+		if next <= latest && latest <= stored.InnerFilter().ToBlock() {
+			rf := NewRunningEventFilterHot(database, stored.InnerFilter(), next)
+			if fillErr := fillRunningEventFilter(database, rf, next, latest); fillErr != nil {
+				return nil, fmt.Errorf("fill running filter gap: %w", fillErr)
```
rodrodros: "filling running event filter (latest=%d, anchorFound=%t): %w"

core/running_event_filter.go
```diff
@@ -70,18 +83,22 @@ func (f *RunningEventFilter) Insert(
 	defer f.mu.Unlock()
 
 	if err := f.ensureInit(); err != nil {
-		return err
+		return fmt.Errorf("ensureInit before block %d: %w", blockNumber, err)
```
rodrodros: similar

core/running_event_filter.go
```diff
@@ -13,27 +13,31 @@ import (
...
 
 func (f *RunningEventFilter) ensureInit() error {
-	if f.lazyInit {
+	if f.initialize != nil {
 		f.lazyOnce.Do(func() {
-			filter, err := loadRunningEventFilter(f.txn)
+			filter, err := f.initialize(f.database)
 			if err != nil {
-				f.initErr = err
+				f.initErr = fmt.Errorf("InitializeRunningEventFilter: %w", err)
```
rodrodros: Natural language please

core/running_event_filter.go
```diff
@@ -211,96 +232,109 @@ func (f *RunningEventFilter) Write() error {
...
+				return nil, fmt.Errorf("fill running filter gap: %w", fillErr)
+			}
+			return rf, nil
+		}
 	}
 
-	return rebuildRunningEventFilter(txn, latest)
+	// Multi-window gap, future-dated snapshot, or no snapshot — rebuild.
+	return rebuildRunningEventFilter(database, latest)
 }
```
rodrodros, line 286: In here you wrap the error if there is one as well:

"rebuilding running event filter: %w"

pruner/running_event_filter.go
```diff
@@ -0,0 +1,137 @@
...
+// possible and rebuilding otherwise. Use [core.InitializeRunningEventFilter]
+// on non-pruning nodes.
+func InitializeRunningEventFilter(database db.KeyValueStore) (*core.RunningEventFilter, error) {
+	latest, err := core.GetChainHeight(database)
+	if err != nil {
+		if errors.Is(err, db.ErrKeyNotFound) {
+			filter := core.NewAggregatedFilter(0)
+			return core.NewRunningEventFilterHot(database, &filter, 0), nil
+		}
+		return nil, err
```
rodrodros: error wrap

pruner/running_event_filter.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+
+	floor, err := OldestRetainedBlock(database)
+	if err != nil && !errors.Is(err, db.ErrKeyNotFound) {
+		return nil, fmt.Errorf("InitializeRunningEventFilter: OldestRetainedBlock: %w", err)
```
rodrodros: natural language

pruner/running_event_filter.go
```diff
@@ -0,0 +1,137 @@
...
+		// Same-window gap — resume the snapshot and fill in place.
+		if next <= latest && latest <= stored.InnerFilter().ToBlock() {
+			// Clamp to floor so we don't read pruned headers. Below-floor
+			// snapshot bits remain — harmless false positives, since the
+			// referenced receipts are pruned and no event query can act
+			// on a hit.
+			next = max(next, floor)
+			rf := core.NewRunningEventFilterHot(database, stored.InnerFilter(), next)
+			if fillErr := fillRunningEventFilter(database, rf, next, latest); fillErr != nil {
+				return nil, fmt.Errorf("fill running filter gap: %w", fillErr)
```
rodrodros: change it to follow same style as other like this

pruner/running_event_filter.go
```diff
@@ -0,0 +1,137 @@
...
+			}
+			return rf, nil
+		}
+	}
+
+	// Multi-window gap, future-dated snapshot, or no snapshot — rebuild.
+	return rebuildRunningEventFilter(database, latest, floor)
```
rodrodros: Wrap the error

pruner/running_event_filter.go
```diff
@@ -0,0 +1,137 @@
...
+) error {
+	for blockNum := from; blockNum <= end; blockNum++ {
+		header, err := core.GetBlockHeaderByNumber(txn, blockNum)
+		if err != nil {
+			return fmt.Errorf("GetBlockHeaderByNumber %d: %w", blockNum, err)
+		}
+		if err := rf.Insert(header.EventsBloom, blockNum); err != nil {
+			return fmt.Errorf("insert block %d: %w", blockNum, err)
+		}
+	}
```
rodrodros, line 94: Now you're using natural language 😱, I think the verbal form should "inserting" since you are erroring when "inserting..."

pruner/running_event_filter.go
```diff
@@ -0,0 +1,137 @@
...
+	txn db.KeyValueStore,
+	rf *core.RunningEventFilter,
+	from,
+	end uint64,
+) error {
+	for blockNum := from; blockNum <= end; blockNum++ {
+		header, err := core.GetBlockHeaderByNumber(txn, blockNum)
+		if err != nil {
+			return fmt.Errorf("GetBlockHeaderByNumber %d: %w", blockNum, err)
+		}
```
rodrodros, line 90: Natural language

pruner/running_event_filter.go
```diff
@@ -0,0 +1,137 @@
...
+	} else {
+		continueFrom = floor
+		windowStart = floorAligned
+	}
+
+	filter := core.NewAggregatedFilter(windowStart)
+	runningFilter := core.NewRunningEventFilterHot(database, &filter, continueFrom)
+	if err := fillRunningEventFilter(database, runningFilter, continueFrom, latest); err != nil {
+		return nil, fmt.Errorf(
+			"rebuild (latest=%d, anchorFound=%t, floor=%d): %w",
```
rodrodros: Similar content to before, a function shouldn't give context of itself, otherwise every function should wrap everytime it returns an error. with the function name. Better to handle it at the caller

pruner/running_event_filter.go
```diff
@@ -0,0 +1,137 @@
...
+
+	found := false
+	for {
+		_, err := core.GetAggregatedBloomFilter(database, rangeStartAligned, lastStoredFilterRangeEnd)
+		if err == nil {
+			found = true
+			break
+		}
+		if !errors.Is(err, db.ErrKeyNotFound) {
+			return nil, fmt.Errorf("rebuild: scan for anchor at [%d,%d]: %w",
```
rodrodros: similar, "scanning" instead of "scan" and avoid using "metaphores" such as "anchor". Use block, or first block, or other concept that is specific to blockchains and full nodes

core/running_event_filter.go
```diff
@@ -211,96 +232,109 @@ func (f *RunningEventFilter) Write() error {
...
-				txn,
-				&filter,
-				0,
-			)
-		} else {
-			return nil, err
-		}
+	stored, err := GetRunningEventFilter(database)
+	if err != nil && !errors.Is(err, db.ErrKeyNotFound) {
+		return nil, err
```
rodrodros: wrap the error

rodrodros (commented):
Looks good but there is a lot of inconsistencies with how errors are wrapped and return, diverging from Juno and Go style: 
* Errors not being written in natural language
* Errors not being wrapped
* Errors not using the right continuous present tense in the verb
* Errors starting with capital letters (the ones not written in natural language)
* Errors written differently, even if they are checking for the same thing. 

Otherwise seems positive but will need to give another review focusing on functionality

blockchain/blockchain.go
rodrodros, line 1: Some of the read functions (`FromBlock`, `ToBlock`, ...) still panic if `ensureInit` errors. 

Should we update their behaviour return an error instead of panicking? Or if `ensureInit` errors will be always because there is a bug in the code and not dependent on user input?

core/running_event_filter.go
```diff
@@ -211,96 +232,115 @@ func (f *RunningEventFilter) Write() error {
...
+	from,
+	latest uint64,
+) error {
+	for blockNum := from; blockNum <= latest; blockNum++ {
+		header, err := GetBlockHeaderByNumber(database, blockNum)
+		if err != nil {
+			return fmt.Errorf("getting block header by number %d: %w", blockNum, err)
+		}
+		if err := rf.Insert(header.EventsBloom, blockNum); err != nil {
+			return fmt.Errorf("inserting block %d in events bloom %w", blockNum, err)
```
rodrodros: (nit) missing `:` before `%w`

core/running_event_filter.go
```diff
@@ -211,96 +232,115 @@ func (f *RunningEventFilter) Write() error {
...
+		return nil, fmt.Errorf("getting stored running event filter: %w", err)
+	}
+	if err == nil {
+		next := stored.NextBlock()
+		// Caught up — return the snapshot as-is.
+		if next == latest+1 {
+			return NewRunningEventFilterHot(database, stored.InnerFilter(), next), nil
+		}
+		// Same-window gap — resume the snapshot and fill in place.
+		if next <= latest && latest <= stored.InnerFilter().ToBlock() {
```
rodrodros, line 268: I was reading a bit the code `InnerFilter` says that it returns a deep copy but it is jut a reference (we should updater `InnerFilter` description.

Also it is being called twice, once after the other, does it makes sense store the first result in a variable and then re-use it for the next call?

core/running_event_filter.go
```diff
@@ -70,18 +83,22 @@ func (f *RunningEventFilter) Insert(
...
 
 	if err := f.ensureInit(); err != nil {
-		return err
+		return fmt.Errorf("couldn't initialize the running event filter: %w", err)
 	}
 
-	err := f.inner.Insert(bloom, blockNumber)
-	if err != nil {
-		return err
+	if err := f.inner.Insert(bloom, blockNumber); err != nil {
```
rodrodros, line 89: I think I've found an issue at Bloom Filter window roll over. 

When we do `ensureInit` which calls `fillRunningEventFilter`, when `latest == innerFilter.ToBlock()`, it rebuilds the filter, stores it, and rolls to the next one. Specifically, this would happen with block `8191`. `ensureInit` would fill the block range `[0, 8191]` 

When `latest == innerFilter.ToBlock()`, `fillRuningEventFilter` will add the block and then swap the current inner at range [0, 8191] for a new one at range [8192, 16393]. 

Then we call `f.inner.Insert` with the same `blockNumber` which will error with a `8191` block number doesn't belong to range `[8192, 16393]`. 

This error will be determinstic every time we stop at `k * 8192 - 1`.  

----

Cheap fix is an `if` in `RunningEventFilter.Insert`, but I am happy if there is a better fix

rodrodros (approved):
Looks good, pre approving but please add a fix for the signaled comment (and a regression test)

blockchain/blockchain.go
EgeCaner: `ensureInit` doesn't depend on user input, but it can fail due to DB faults/corruption or an attempt to access pruned data (a bug). Once it errors, it will keep returning the same error and stucks syncing. The reason it's lazy in the first place is to keep blockchain.New() cheap by deferring initialization to runtime rather than doing it at construction time. One alternative is to just initialize the running filter eagerly at construction time, which sidesteps the error-vs-panic question entirely. The return error instead of panic change is trivial, I've opened a [PR](https://github.com/NethermindEth/juno/pull/3641) with it so we can merge that instead if you'd prefer.

core/running_event_filter.go
EgeCaner: Addressed the issue in this [PR](https://github.com/NethermindEth/juno/pull/3643), please take a look

Issue #3075, opened 2025-08-22
[Tendermint] Protobuf enum validation


infrmtcs:
The protobuf library will preserve the field on unknown enum. Hence, the adapter needs to check if the enum is defined and throw an error if not.

PR #3639, opened 2026-05-14
fix(p2p): reject unknown proto3 enum values in network adapters

RafaelGranza:
## [Issue](https://github.com/NethermindEth/juno/issues/3075)

Proto3 preserves unknown enum values on unmarshal instead of rejecting them. A peer can send an enum field with a value not defined, letting garbage flow through.

## Fix

Validate at the adapter boundary using the generated `_name` map. Unknown value gives an error.

Decided not to implement a generic `protoenum.Validate()` helper function, since the inline validation was short, easy to read, and only appears in 4 places.

### DIFF
--- a/adapters/p2p2consensus/p2p_consensus_adapters_test.go
+++ b/adapters/p2p2consensus/p2p_consensus_adapters_test.go
@@ -10,6 +10,7 @@ import (
 	transactiontestutils "github.com/NethermindEth/juno/adapters/testutils"
 	"github.com/NethermindEth/juno/blockchain/networks"
 	"github.com/NethermindEth/juno/starknet/compiler"
+	"github.com/starknet-io/starknet-p2p-specs/p2p/proto/common"
 	"github.com/starknet-io/starknet-p2p-specs/p2p/proto/consensus/consensus"
 	"github.com/stretchr/testify/require"
 )
@@ -92,4 +93,20 @@ func TestAdaptProposalTransaction(t *testing.T) {
 
 func TestAdaptProposalFin(t *testing.T) {
 	testP2PToConsensusToP2P(t, testutils.GetTestProposalFin, p2p2consensus.AdaptProposalFin, consensus2p2p.AdaptProposalFin)
+}
+
+func TestAdaptBlockInfo_UnknownL1DaMode(t *testing.T) {
+	_, p2pBlockInfo := testutils.GetTestBlockInfo(t)
+	p2pBlockInfo.L1DaMode = common.L1DataAvailabilityMode(99)
+
+	_, err := p2p2consensus.AdaptBlockInfo(p2pBlockInfo)
+	require.ErrorContains(t, err, "unknown l1_da_mode value: 99")
+}
+
+func TestAdaptProposalCommitment_UnknownL1DaMode(t *testing.T) {
+	_, p2pProposalCommitment := testutils.GetTestProposalCommitment(t)
+	p2pProposalCommitment.L1DaMode = common.L1DataAvailabilityMode(99)
+
+	_, err := p2p2consensus.AdaptProposalCommitment(p2pProposalCommitment)
+	require.ErrorContains(t, err, "unknown l1_da_mode value: 99")
 }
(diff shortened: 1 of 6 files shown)


adapters/p2p2consensus/p2p_consensus_adapters_test.go
```diff
@@ -93,3 +94,19 @@ func TestAdaptProposalTransaction(t *testing.T) {
...
+	require.ErrorContains(t, err, "unknown l1_da_mode value: 99")
+}
+
+func TestAdaptProposalCommitment_UnknownL1DaMode(t *testing.T) {
+	_, p2pProposalCommitment := testutils.GetTestProposalCommitment(t)
+	p2pProposalCommitment.L1DaMode = common.L1DataAvailabilityMode(99)
+
+	_, err := p2p2consensus.AdaptProposalCommitment(p2pProposalCommitment)
+	require.ErrorContains(t, err, "unknown l1_da_mode value: 99")
+}
```
RafaelGranza, line 112: The validators are tested indirectly through the public Adapt* functions.

I considered adding a dedicated `validation_test.go`, coded it, added tests for every validator branch, and ended up with a HUGE mess of tautological tests (one per nil-check, each mirroring the production code 1:1). 

Decided to keep the indirect tests instead.

consensus/p2p/vote/vote_test.go
```diff
@@ -302,4 +302,18 @@ func TestStarknetVoteAdapter_ErrorCases(t *testing.T) {
...
 	})
+
+	t.Run("ToVote rejects unknown VoteType", func(t *testing.T) {
+		voterBytes := createTestHashBytes(42)
+		result, err := StarknetVoteAdapter.ToVote(&consensus.Vote{
+			VoteType:    consensus.Vote_VoteType(99),
+			BlockNumber: 100,
+			Round:       5,
+			Voter:       &common.Address{Elements: voterBytes},
+		})
```
RafaelGranza, line 313: Happy path covered by `TestStarknetVoteAdapter_RoundTrip`

p2p/server/server_test.go
```diff
@@ -0,0 +1,41 @@
+package server
```
rodrodros: server_test?

RafaelGranza: I'm working on it. Never to make this error again

RafaelGranza: Fixed! `package server_test` now. 
I used the opportunity to rewrite as e2e through libp2p, so `Server.Run` and the full headers protocol path are covered as well, not only the direction validation.

p2p/server/server_test.go
```diff
@@ -0,0 +1,122 @@
...
+		var resp header.BlockHeadersResponse
+		require.NoError(t, protodelim.UnmarshalFrom(bufio.NewReader(stream), &resp))
+		require.IsType(t, &header.BlockHeadersResponse_Fin{}, resp.HeaderMessage)
+	})
+
+	t.Run("drops request for unknown direction", func(t *testing.T) {
+		_, serverHost, clientHost := setup(t)
+		// No reader EXPECTs: validation must fail before any DB call.
+
+		stream := openRequestStream(t, clientHost, serverHost, synccommon.Iteration_Direction(99))
```
brbrr, line 116: Not sure if that's reasonable here, but would it make sense to check if the stream was actually dropped/closed?

RafaelGranza: The `ReadAll` + `NoError` + `Empty` already check that. A clean EOF only happens if the server closed the stream without writing

adapters/p2p2consensus/p2p_consensus_adapters_test.go
ongyimeng: Nice insight. This reminds me of the question I asked about tests during daily. In general, behavior-focused tests are preferred over implementation-specific ones because they’re easier to maintain as the code evolves.

ongyimeng (approved):
Overall, this is a nice hardening change. Thanks for covering the gaps and adding test coverage.

PR #3641, opened 2026-05-14
running event filter: return errors from accessors instead of panicking

EgeCaner:


rodrodros (approved):

rodrodros:
Codecov is bypassable in this instance but other CI should pass before we merge

PR #3643, opened 2026-05-14
fix(running event filter): window-roll consistency on lazy load and reorg deletes stale filters

EgeCaner:
##  Summary

  - Atomic `Insert` / `OnReorg` with chain-state batch. `RunningEventFilter` gains `InsertWithBatch` / `OnReorgWithBatch`. `Store`, `Finalise`, and `RevertHead` now route the running-filter write through the same `db.Write` batch as `writeBlockContent` / `deleteBlockContent`, so the persisted aggregated filter and the chain head commit or roll back together. Addresses issue reported [here](https://github.com/NethermindEth/juno/pull/3638#discussion_r3241745246).
  - `OnReorg` deletes the stale persisted filter. On a backward window boundary cross, `onReorg` calls `DeleteAggregatedBloomFilter` for the current window before swapping back to the previous one. This drops the obsolete persisted filter; a future rollover repopulates the window. The delete is routed through the caller's batch via `OnReorgWithBatch`, so it commits atomically with the chain-state revert.

### DIFF
--- a/blockchain/statebackend/deprecated.go
+++ b/blockchain/statebackend/deprecated.go
@@ -79,6 +79,7 @@ func (b *deprecatedStateBackend) Store(
 		if err := verifyBlockSuccession(txn, block); err != nil {
 			return err
 		}
+
 		err := deprecatedstate.New(txn).Update(
 			block.Header,
 			stateUpdate,
@@ -89,14 +90,15 @@ func (b *deprecatedStateBackend) Store(
 			return err
 		}
 
-		if err := writeBlockContent(
+		err = writeBlockContent(
 			b.database,
 			txn,
 			block,
 			stateUpdate,
 			blockCommitments,
 			newClasses,
-		); err != nil {
+		)
+		if err != nil {
 			return err
 		}
 
@@ -222,14 +224,15 @@ func (b *deprecatedStateBackend) Finalise(
 			}
 		}
 
-		if err := writeBlockContent(
+		err = writeBlockContent(
 			txn,
 			txn,
 			block,
 			stateUpdate,
 			commitments,
 			newClasses,
-		); err != nil {
+		)
+		if err != nil {
 			return err
 		}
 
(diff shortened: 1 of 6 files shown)


blockchain/statebackend/deprecated.go
```diff
@@ -89,28 +89,24 @@ func (b *deprecatedStateBackend) Store(
 			return err
 		}
 
-		return writeBlockContent(
+		if err := writeBlockContent(
```
rodrodros: nit: separate if statement

blockchain/statebackend/deprecated.go
```diff
@@ -226,20 +222,19 @@ func (b *deprecatedStateBackend) Finalise(
 			}
 		}
 
-		return writeBlockContent(
+		if err := writeBlockContent(
```
rodrodros: same nit

blockchain/statebackend/statebackend.go
```diff
@@ -88,24 +88,23 @@ func (b *stateBackend) Store(
 			return err
 		}
 
-		return writeBlockContent(
+		if err := writeBlockContent(
```
rodrodros: same nit

blockchain/statebackend/statebackend.go
```diff
@@ -225,20 +224,19 @@ func (b *stateBackend) Finalise(
 			}
 		}
 
-		return writeBlockContent(
+		if err := writeBlockContent(
```
rodrodros: same nit

rodrodros (approved):
Looks good

blockchain/statebackend/deprecated.go
EgeCaner: updated

blockchain/statebackend/deprecated.go
EgeCaner: updated

blockchain/statebackend/statebackend.go
EgeCaner: updated

blockchain/statebackend/statebackend.go
EgeCaner: updated

PR #3644, opened 2026-05-15
fix(rpc): bound l1 head to chain height when l1 head is ahead of l2 head

EgeCaner:
### Problem

While the node is syncing, the L1 head can be ahead of the local chain height. Queries with the l1_accepted block tag then resolve to a block number the node doesn't have locally and return a not-found error.

The spec defines l1_accepted as:

> Tag l1_accepted refers to the latest Starknet block which was included in a state update on L1 and finalized by the consensus on L1.

Any local block at or below the L1 head already satisfies this definition, so falling back to the chain height is spec-compliant.

### Solution

Add an `l1AcceptedBlockNumber()` helper that returns `min(l1Head.BlockNumber, chainHeight)` and use it from every `IsL1Accepted()` branch in the v9 and v10 RPC handlers.

### DIFF
--- a/rpc/v10/block_test.go
+++ b/rpc/v10/block_test.go
@@ -99,6 +99,19 @@ func createBlockTestCases(
 			blockStatus: rpc.BlockAcceptedL1,
 		},
 		{
+			description: "blockID - l1_accepted bounded to chain height when L1 is ahead",
+			block:       block,
+			commitments: commitments,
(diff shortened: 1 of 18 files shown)


rodrodros (approved):

rodrodros (changes requested):
Approved unintentionally, still looking at it

rodrodros (approved):

PR #3645, opened 2026-05-15
refactor(state): make new state history write to its own buckets

MaksymMalicki:
This makes the new state utilise the new history buckets. This is necessary for the state history migration, as we don't want to overwrite the existing buckets, effectively corrupting them on crash. This PR follows the already existing pattern, where the new state has its own buckets for the trie and contracts.


EgeCaner (approved):
LGTM!

PR #3646, opened 2026-05-18
chore(ci): Request specific rust version

brbrr:
There were some CI failures where the worker failed to use the correct rust version. It was a fluke, potentially related to stale machine state or our CI cache. This PR attempts to fix the symptom, rather than the actual issue, as it remains obscure atm.


rodrodros:
@brbrr it stopped happening, I think we might no longer need this

rodrodros (approved):

Issue #2084, opened 2024-08-22
Investigate dependencies of Rust code in go packages


kirugan:
Context: the Starknet P2P explorer uses Juno as an external library for some of its features, but it's not as simple as doing `go get`, as it's required to compile the Rust dependencies. 

For example, here's a snippet of the build process if Juno is used as a library (without the additional Rust compilation):
```
...
26.31 /usr/bin/ld: cannot find -ljuno_starknet_compiler_rs: No such file or directory
26.31 /usr/bin/ld: cannot find -ljuno_starknet_core_rs: No such file or directory
26.31 /usr/bin/ld: /tmp/go-link-3421764336/000026.o: in function `mygetgrouplist':
26.31 /_/GOROOT/src/os/user/getgrouplist_unix.go:15: warning: Using 'getgrouplist' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
26.31 /usr/bin/ld: /tmp/go-link-3421764336/000025.o: in function `mygetgrgid_r':
26.31 /_/GOROOT/src/os/user/cgo_lookup_cgo.go:45: warning: Using 'getgrgid_r' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
...
```

List of packages that is used right now:
```
github.com/NethermindEth/juno/utils
github.com/NethermindEth/juno/clients/feeder
github.com/NethermindEth/juno/starknetdata/feeder
```

These packages shouldn't require rust code at all (especially `utils` package). We need to investigate what's going on

ShantelPeters:
Can I work on this ?

weiihann:
Did some investigations:

`github.com/NethermindEth/juno/clients/feeder`
- Depends on `github.com/NethermindEth/juno/starknet`
- Indirectly depends on `core, core/crypto (via starknet)`

That's why the following errors occur:
```
26.31 /usr/bin/ld: cannot find -ljuno_starknet_compiler_rs: No such file or directory
26.31 /usr/bin/ld: cannot find -ljuno_starknet_core_rs: No such file or directory
```

So compilation is needed for packages `core` and `starknet`, which require Rust dependencies.

Some solutions in mind:
1. Move the Rust code to its own packages for `core` and `starknet`
- `starknet` contains the `Compile` function, in which only `rpc` and `adapters/p2p2core` are using it, we can move it to a separate package (i.e. `compiler`) without much issues
[...]

PR #3648, opened 2026-05-18
Perf pure go cairo0 class hash

ongyimeng:
## Summary

Replace Cairo0 class hashing with a pure-Go implementation and optimize the path.

## Changes

- reimplement legacy Cairo0 hinted-hash canonicalization in Go
- remove the old Rust/FFI Cairo0 hashing path
- optimize Pedersen hashing and table initialization
- parallelize independent work in deprecated Cairo class hashing
- add stronger regression and parity tests

## Results

These results are from the benchmark, `BenchmarkClassV0Hash`.

Runtime improved compared to the latest upstream main Rust/FFI path:

- Smaller fixture: ~7.52 ms/op -> ~6.08 ms/op, about 19.2% faster.
- Larger fixture: ~79.67 ms/op -> ~59.46 ms/op, about 25.4% faster.

Go memory metrics use the initial pure-Go baseline since Go benchmarks do not count Rust-side FFI allocations.

Go memory metrics improved compared to the initial pure-Go baseline:

- Smaller fixture:
  - ~541,656 B/op -> ~458,824 B/op, about 15.3% lower.
  - 3,025 allocs/op -> 1,881 allocs/op, about 37.8% fewer.
- Larger fixture:
  - ~3.40 MB/op -> ~2.49 MB/op, about 26.8% lower.
  - ~19.39k allocs/op -> 12,042 allocs/op, about 37.9% fewer.

Memory tradeoff:

- Pedersen 8-bit precomputed tables retain about 1 MiB.
- Rough import-time process RSS delta is about 6.1 MiB. (measured by comparing an empty process against one that imports `core/crypto`).

Benchmarks were run on macOS, Intel i5, with `GOMAXPROCS=1` 

## Motivation
Issue #2084


core/crypto/pedersen_hash.go
```diff
@@ -31,13 +153,54 @@ type PedersenDigest struct {
...
+		d.digest = pedersen(&d.digest, elems[idx].Impl())
 	}
 	d.count += uint64(len(elems))
 	return d
 }
 
 func (d *PedersenDigest) Finish() felt.Felt {
-	d.digest = pedersenhash.Pedersen(&d.digest, new(fp.Element).SetUint64(d.count))
+	var count fp.Element
+	count.SetUint64(d.count)
```
rodrodros, line 164: You can use `count := felt.FromUint64(d.count)`

core/deprecated_cairo_program.go
```diff
@@ -0,0 +1,856 @@
...
+				for _, u16 := range utf16.Encode([]rune{r}) {
+					writeJSONHex16(buffer, u16)
+				}
+			}
+		}
+	}
+	buffer.WriteByte('"')
+}
+
+func writeJSONHex16(buffer *bytes.Buffer, value uint16) {
```
rodrodros: constant `hexChars` is only used here so it is a bit better if its defined in this scope

core/deprecated_cairo_program.go
```diff
@@ -0,0 +1,856 @@
...
+}
+
+func writeLegacyReferenceManager(buffer *bytes.Buffer, manager legacyReferenceManager) error {
+	buffer.WriteByte('{')
+	first := true
+	writeJSONFieldPrefix(buffer, "references", &first)
+	writeLegacyReferences(buffer, manager.References)
+	buffer.WriteByte('}')
+	return nil
+}
```
RafaelGranza: Nit: this returns error but never produces one. In Go that's usually a "can fail" signal.
Dropping the return matches the other infallible writers here (e.g. `writeLegacyAttributes`). WDYT?

core/deprecated_cairo_program_test.go
```diff
@@ -0,0 +1,110 @@
+package core
```
RafaelGranza, line 1: This file uses package `core` to reach internal helpers. We have a team **preference** for `package core_test` exercising only the public API. 

This may not be trivial for this PR, but could you give it a look?

core/crypto/pedersen_hash_test.go
```diff
@@ -101,23 +91,56 @@ func TestPedersenArray(t *testing.T) {
...
+}
+
+func mapFeltsToFPElems(felts []*felt.Felt) []*fp.Element {
+	fpElems := make([]*fp.Element, len(felts))
+	for i, elem := range felts {
+		fpElems[i] = elem.Impl()
+	}
+	return fpElems
+}
+
```
RafaelGranza, line 143: Nice touch. 
Differential tests against `gnark` give us a much stronger safety net than a fixed snapshot for future Pedersen tweaks.

core/class_hash.go
```diff
@@ -25,24 +15,80 @@ func deprecatedCairoClassHash(class *DeprecatedCairoClass) (felt.Felt, error) {
...
 
-	C.Cairo0ClassHash(classJSONCStr, (*C.char)(unsafe.Pointer(&hashBytes[0])))
-	hash.SetBytes(hashBytes[:])
-	C.free(unsafe.Pointer(classJSONCStr))
-	if hash.IsZero() {
-		return felt.Felt{}, errors.New("failed to calculate class hash")
+	wg.Wait()
+	if hintedClassHashErr != nil {
+		return felt.Felt{}, hintedClassHashErr
 	}
```
RafaelGranza, line 54:
```suggestion
    // Cheap ones run sequentially in the main goroutine.
    externalEntryPointsHash := hashDeprecatedEntryPoints(definition.EntryPoints.External)
    l1HandlerEntryPointsHash := hashDeprecatedEntryPoints(definition.EntryPoints.L1Handler)
    constructorEntryPointsHash := hashDeprecatedEntryPoints(definition.EntryPoints.Constructor)
    builtinsHash := hashBuiltinNames(program.Builtins)

    // Parallelize the two heavy ones: Pedersen-over-bytecode and JSON-canonical + Keccak.
    var (
        dataHash        felt.Felt
        hintedClassHash felt.Felt
    )
    var g errgroup.Group
    g.Go(func() error {
        dataHash = hashDeprecatedProgramData(program.Data)
        return nil
    })
    g.Go(func() error {
        var err error
        hintedClassHash, err = computeHintedClassHash(definition.Abi, program)
        return err
    })
    if err := g.Wait(); err != nil {
        return felt.Felt{}, err
    }
```

The split between the two goroutines is hard to follow at first since one has 1 statement and the other has 5. 

The actual intent is "parallelize the two heavy ops (`dataHash` and `hintedClassHash`)" with the small entry-point/builtin hashes piggybacking. Consider:

1. Running the small hashes sequentially in the main goroutine (they're trivially fast).

2. Using `errgroup.Group` (already a Juno convention, see `p2p/p2p.go`, `rpc/handlers.go`) for cleaner error handling and intent.

core/class_hash.go
```diff
@@ -1,19 +1,9 @@
 package core
 
-//#include <stdint.h>
```
RafaelGranza, line 3: Follow-up idea (Future PR): merge `core/class_hash.go` into `core/class.go`. The split made sense when this file hosted the CGO/FFI to Rust, but in pure Go the V0 hashing could live next to its type.

Similar to `SierraClass.Hash()` and `CasmClass.Hash()` already do in `class.go`

core/deprecated_cairo_program.go
```diff
@@ -0,0 +1,856 @@
...
+	if value.Destination != nil {
+		writeJSONFieldPrefix(buffer, "destination", &first)
+		writeJSONString(buffer, *value.Destination)
+	}
+	writeJSONFieldPrefix(buffer, "type", &first)
+	writeJSONString(buffer, value.Type)
+	if len(value.Value) > 0 {
+		writeJSONFieldPrefix(buffer, "value", &first)
+		writeJSONRaw(buffer, value.Value)
+	}
```
RafaelGranza, line 591: The `value` field on `legacyIdentifier` (line 733) is `json.RawMessage` passed through verbatim via `writeJSONRaw`. 

The reference behavior (from `starknet-core` in `starknet-rs`,  the source of the "truth") parses value as `Option<BigUint>`, normalizing it on serialization. So a source with non-canonical formatting (e.g. `042`, `1e5`) would survive into the hashed payload here, while `starknet-rs` would emit `42` and `100000`. 

It won't show up in the upstream test corpus since those fixtures are clean. 

Worth either parsing through `*big.Int` or documenting the canonical-input assumption.

RafaelGranza (approved):
Really solid work. The test setup with `starknet-rs` fixtures and the snapshot are great touches.

Could you walk through the comments before merging? Most are nits or future direction ideas, no blocking items.

RafaelGranza:
Sorry, misclicked and updated your branch with a merge instead of a rebase.
Heads up in case you want to clean it up.

core/crypto/pedersen_hash.go
ongyimeng: I don’t think this works as-is. `FromUint64` is generic and needs an explicit type, for example:

    felt.FromUint64[felt.Felt](d.count)

Even with that, `pedersen` works on `*fp.Element`, so wrapping as `felt.Felt` just to unwrap it again with `count.Impl()` isn’t better for performance.

rodrodros: Ah I see, makes sense, I didn't notice it was a `fp.Element`. Note though, that you can just do `fp.Element(felt.Felt)` casting and there shouldn't be any issues (no need to use Impl())

core/deprecated_cairo_program.go
ongyimeng: good point!

rodrodros (changes requested):
All looking good so far, blocking until I get a good grasp of the code myself

core/class_hash.go
ongyimeng: You understood my intent perfectly. I was curious about your suggestion so I benchmarked it, but it was actually slower.

I think it’s because the current code overlaps the work:

  max(A, B + C)

while the suggested version does:

  C + max(A, B)

where A and B are the heavier hashes, and C is the smaller hashes. So C becomes serialized instead of being hidden behind A.

That said, your suggestion did inspire me to try a 3-goroutine split, which led to marginal improvements. Thanks for the idea!

core/class_hash.go
rodrodros: `core/class.go` has already more than it should in my opinion. I believe a better direction would actually be splitting `core/class` or making it a package. 

Regarding this PR `core/class_hash.go` perhaps the best approach is to find it a better name like `core/deprecated_class_hash.go`. To be discussed and I agree it is better to do in a future PR to keep this diff readable

core/deprecated_cairo_program.go
rodrodros: @ongyimeng please also add a test reflecting this behaviour

core/deprecated_cairo_program_test.go
```diff
@@ -0,0 +1,110 @@
...
+	HintedClassHash string `json:"hinted_class_hash"`
+}
+
+type upstreamHashFixture struct {
+	name     string
+	artifact string
+	hashes   string
+}
+
+// These fixtures and expected hashes are copied from the upstream starknet-core legacy.rs tests.
```
rodrodros: upstream url please

core/deprecated_cairo_program_test.go
```diff
@@ -0,0 +1,110 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+const (
+	deprecatedFixturePath = "../clients/feeder/testdata/sepolia/class/" +
+		"0x5f18f9cdc05da87f04e8e7685bd346fc029f977167d5b1b2b59f69a7dacbfc8.json"
+	// This is a byte-for-byte canonical-output snapshot, not a normal JSON fixture.
+	// Keep the non-.json extension so editors do not autoformat it.
+	canonicalProgramPath  = "testdata/deprecated_cairo_program.txt"
```
rodrodros: These two constants could be scoped inside their respective functions

rodrodros (commented):
@ongyimeng looks good, can you add the benchmark tests to the diff as well 🙏

core/deprecated_cairo_program.go
ongyimeng: I don’t think this is an issue.

Let me explain step by step:

1. In Juno, `legacyIdentifier.value` is stored as raw JSON, not parsed into a numeric type:

   ```go
   type legacyIdentifier struct {
       // ...
       Value json.RawMessage `json:"value,omitempty"`
   }
   ```

   And in the hinted-hash writer we pass it through verbatim:

   ```go
   if len(value.Value) > 0 {
       writeJSONFieldPrefix(buffer, "value", &first)
       writeJSONRaw(buffer, value.Value)
   }

   func writeJSONRaw(buffer *bytes.Buffer, raw json.RawMessage) {
       if len(raw) == 0 {
           buffer.WriteString("null")
           return
       }
       buffer.Write(raw)
   }
   ```

2. Upstream `starknet-rs` models the same field as raw JSON too, not as `BigUint`:

   ```rust
   pub value: Option<Box<serde_json::value::RawValue>>,
   ```

   Reference: [legacy.rs#189](https://docs.rs/starknet-core/latest/src/starknet_core/types/contract/legacy.rs.html#189)

3. In the upstream hinted-class-hash path, the program serializer includes the original identifiers map directly.

   References:

   - [legacy.rs#752](https://docs.rs/starknet-core/latest/src/starknet_core/types/contract/legacy.rs.html#752)
   - [legacy.rs#761](https://docs.rs/starknet-core/latest/src/starknet_core/types/contract/legacy.rs.html#761)

4. That serialized structure is then turned into the final hash payload with `to_string_pythonic(...)`.

   Reference: [legacy.rs#608](https://docs.rs/starknet-core/latest/src/starknet_core/types/contract/legacy.rs.html#608)

So the upstream behavior here is “keep `value` as raw JSON and serialize it as such”, not “parse as `BigUint` and normalize before hashing”. That means Juno’s `json.RawMessage` + `writeJSONRaw(...)` is matching the upstream source of truth.

Small note: `042` is not valid JSON, so that example wouldn’t apply. `1e5` is valid JSON, but upstream would preserve that spelling too.

core/class_hash.go
```diff
@@ -25,24 +15,84 @@ func deprecatedCairoClassHash(class *DeprecatedCairoClass) (felt.Felt, error) {
...
+		defer wg.Done()
+		hintedClassHash, hintedClassHashErr = computeHintedClassHash(definition.Abi, program)
+	}()
 
-	C.Cairo0ClassHash(classJSONCStr, (*C.char)(unsafe.Pointer(&hashBytes[0])))
-	hash.SetBytes(hashBytes[:])
-	C.free(unsafe.Pointer(classJSONCStr))
-	if hash.IsZero() {
-		return felt.Felt{}, errors.New("failed to calculate class hash")
+	wg.Wait()
```
rodrodros, line 51: In this code you start from `main` thread and spawn go routines `A`, `B` and `C` while `main` waits idle for them to wait. You can improve this by giving tasks to `main` as well. 

Changing the flow to:
1. Spawn `A` 
2. Spawn `B`
3. Do the work of `C`
4. Wait for a `A` and `B` to finish

ongyimeng: I experimented further and benchmarked the variants against the 3-goroutine baseline. The best result was keeping `computeHintedClassHash(...)` on the main goroutine.

```text
1. main = small hashes, goroutines = dataHash + hintedClassHash (tried previously)
2. main = hintedClassHash, goroutines = dataHash + small hashes
3. main = dataHash, goroutines = hintedClassHash + small hashes
```

core/class_hash.go
ongyimeng: I intended `class_hash` to remain the coordinator for the overall hashing flow, while `deprecated_cairo_program` contains the program-specific logic needed to match the upstream Rust implementation.

Upstream calls this `LegacyProgram`, but I chose `deprecated_cairo_program` to better align with Juno’s existing naming conventions.

rodrodros (approved):
Good work!

core/deprecated_cairo_program_test.go
rodrodros: Let's give it a pass for now

PR #3650, opened 2026-05-18
fix(db): correct batch DeleteRange semantics

MaksymMalicki:
## Summary
- `db/memory/batch.go` `DeleteRange` was misusing `NewIterator`'s prefix arg as a range start and missing an iterator close. The prefix-filtered iterator silently dropped any key in `[start, end)` that didn't share `start` as a literal byte prefix — e.g. `DeleteRange("3", "7")` only deleted keys starting with `"3"`, missing `"4"`/`"5"`/`"6"` and their descendants. Now uses `NewIterator(nil, false)` + `Seek(start)` to match pebble's `DeleteRange` semantics; added `defer it.Close()`.
- Added a shared `BatchDeleteRange` regression test in `db/testutil.go` so memory, pebble, and pebblev2 are pinned to the same semantics. The existing `DeleteRange` subtest only covered `database.DeleteRange`, not the buggy `batch.DeleteRange` path — which is why this slipped.

### DIFF
--- a/db/memory/batch.go
+++ b/db/memory/batch.go
@@ -132,12 +132,18 @@ func (b *batch) DeleteRange(start, end []byte) error {
 		return errBatchClosed
 	}
 
-	it, err := b.NewIterator(start, false)
+	// Range-based, matching pebble's DeleteRange semantics: delete every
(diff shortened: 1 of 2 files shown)


EgeCaner (approved):
LGTM!

PR #3657, opened 2026-05-19
feat(migration): headstate migration

MaksymMalicki:
### Summary

Adds the `headstate` migration: a resume-safe transform that consolidates the deprecated per-field contract layout (`ContractClassHash`, `ContractNonce`, `ContractDeploymentHeight`) into the single `Contract` record introduced by the new-state work. Gated behind the existing `--new-state` flag. First of three new-state migrations.

### How it works

```
ContractClassHash[addr]
ContractNonce[addr]
ContractDeploymentHeight[addr]
          │
          ▼
Contract[addr] = { ClassHash, Nonce, DeployedHeight }
```

Iterates the `ContractClassHash` bucket to discover every contract address. Four worker goroutines (`ingestorCount`) read the three deprecated fields per address into shared `db.Batch`es; one committer drains batches at 96 MB. A semaphore caps in-flight batches at `ingestorCount + 1`. After all records are written, the three deprecated buckets are wiped via `DeleteRange`.

`StorageRoot` is left zero on each new record — the running node lazily backfills it on the contract's first storage write.

### What changes

- New `migration/headstate/` package (migrator, ingestor, committer, counter, tests).
- Registered in `node/migration.go` as an optional migration gated by `cfg.NewState`.
- `state.WriteContract` in `core/state/accessors.go` gains a public overload taking raw fields, used by the migration.

### Resume safety

- `state.HasContract` skips addresses whose Contract record already exists — no overwrite of newer running-node data.
- The trailing wipe re-issues `DeleteRange` over any residual ranges; already-empty ranges are no-ops.
- Ctx cancellation returns `(shouldRerun, ctx.Err())` so the runner re-invokes on next start.


migration/state/headstate/counter.go
```diff
@@ -0,0 +1,47 @@
...
+
+func (c *counter) log(byteSize uint64, completedAddrs int) {
+	c.size += byteSize
+	c.completedAddrs += uint64(completedAddrs)
+
+	now := time.Now()
+	elapsed := now.Sub(c.start).Seconds()
+	if elapsed > c.timeLogRate.Seconds() {
+		mbs := float64(c.size) / float64(db.Megabyte)
+		c.logger.Info(
```
brbrr, line 42: example log output: 
```
juno      | 11:22:12.007 26/05/2026 +00:00      INFO    headstate/counter.go:35 write speed     {"MB": 377.31350898742676, "MB/s": 3.1230496120267954, "completedContracts": 3768018, "completedContracts/s": 31188.141618862945, "time": 120.815726889}
```

ideally, we should round to ~last 2 digits after `.`

rodrodros (approved):
Good work.LGTM! I've done some nits and will merging if there are no further issues

PR #3659, opened 2026-05-19
feat(migration): trie migration

MaksymMalicki:
## Summary

One-shot migration that converts every deprecated Starknet trie on disk into the equivalent `trie2` layout. After this migration runs, the new state package can read directly from the new buckets and the deprecated buckets are wiped.

Bucket mapping:

| Deprecated         | New                       | Hash                 |
|--------------------|---------------------------|----------------------|
| `ClassesTrie`      | `ClassTrie`               | Poseidon             |
| `StateTrie`        | `ContractTrieContract`    | Pedersen             |
| `ContractStorage`  | `ContractTrieStorage`     | Pedersen (per-owner) |

## Format differences

Three things change between the two layouts: how nodes are keyed on disk, how each node's value is encoded, and how path compression is expressed.

### On-disk keying

Both layouts share a common prefix; only the suffix differs:

common (both)              suffix
─────────────              ─────────────────────────────────────────
bucket [|| owner]    →     path-length-byte || path-bytes              (deprecated)
→     nodeType-byte || path-length-byte || path-bytes  (new)

`owner` is present only for storage tries. The new layout's extra `nodeType` byte splits leaves from internal nodes into two index slices within the same bucket — `trie2` state lookups use this to short-circuit between leaf reads and internal-node traversals.

### Node encoding

Both layouts are raw byte streams (no length prefixes, no varints — field-element widths are fixed).

Deprecated: nodes are self-contained. Internal binary nodes embed the compressed paths to their children inline:

leaf       value
binary     value || left-child-path || right-child-path
[|| left-hash || right-hash, optional cache, ignored here]

`value` is the node's own Starknet trie hash, or the stored value when the node is a leaf. The trailing hash pair was a denormalised cache; the migrator does not read it (hashes are recomputed from scratch — see below).

New: every node has an explicit type tag and path compression lives in dedicated edge nodes:

value      value
binary     0x01 || left-edge-hash || right-edge-hash
edge       0x02 || child-hash || encoded-path-segment

### Path compression — the key structural change

The deprecated format compresses paths inside the parent binary node (via its embedded child-path fields). The new format moves compression into dedicated edge nodes sitting between binary nodes and their children:

deprecated:   binary ──────── child-path ────────► child
new:          binary ──► edge ──► child

One consequence: the deprecated root marker (a single entry at the bare bucket prefix recording the root's path) disappears. Whatever the deprecated root embedded becomes either a direct binary/leaf at the empty path or, when the deprecated root path is itself non-empty, an edge node at the empty path that points "down" to the real root.

## Traversal

DFS is a natural fit here. In the new layout a binary node's payload is `left-edge-hash || right-edge-hash` — so before we can write the parent, we need both children's hashes. A bottom-up walk reads each deprecated node exactly once and produces the child hash that the parent needs at the moment the parent is encoded; no separate hashing pass, no intermediate caches sized to the trie. Going top-down would force us to either re-read every child later or hold every visited node in memory until its subtree is hashed.

For each trie:

1. **Enumerate** — scan the deprecated bucket once. Class and state tries each occupy a whole bucket; storage tries share `ContractStorage` keyed by `bucket || owner`, so the enumerator splits them into per-owner descriptors. Each descriptor records the root path (from the bare-prefix marker entry) and the node count.
2. **DFS** — recurse from the root path. A deprecated leaf becomes a value node at the same path. An internal binary node, after both subtrees have been visited, becomes a binary node plus up to two edge nodes (one per non-empty child segment). If the trie's stored root path is itself non-empty, a single edge node at the empty path is written after the traversal completes, replacing the root marker.
3. **Resumability check** — before doing any work for a trie, the migrator looks up its new-format root key. If present, the trie is credited toward progress and skipped.

Pipeline: `IngestorCount` worker goroutines pull descriptors from the enumeration source; a single committer flushes filled batches to disk. A semaphore caps in-flight batches at `IngestorCount * 2`. Every flush and every channel send observes `ctx.Done`; on cancel `Migrate` returns the `shouldRerun` sentinel and the migration runner re-invokes on the next process start.

After the full pipeline finishes, the three deprecated buckets are wiped via `DeleteRange`. The wipe is gated on full success — a crashed mid-migration leaves the deprecated source intact, so partially migrated tries either have a new-format root (skipped on the next pass) or don't (re-migrated from scratch).

### Alternative considered: reverse-iteration BFS

An earlier attempt used reverse-iteration BFS over the deprecated bucket — iterating keys from longest-path to shortest so leaves are processed first, then their parents, then their parents, and so on. Each level's hashes are buffered until the next level up consumes them.

In practice it performed comparably to DFS on wall-clock time but with substantially higher peak memory — the buffer of "hashes waiting for their parent" grows roughly with the widest level of the trie, which for full or near-full subtrees is most of the leaves. DFS keeps only the current root-to-leaf path in flight, which is bounded by the trie depth (≤ 251), so memory stays flat regardless of trie size. Same correctness, worse memory profile, no speed win — dropped.

## Hashing

Starknet trie hashes for the new layout:

leaf       value
binary     hashFn(left-edge-hash, right-edge-hash)
edge       hashFn(child-hash, path-segment-as-felt) + segment-length

A zero-length edge short-circuits to the bare `child-hash` — the convention for absent edges. Class tries hash with Poseidon; contract and storage tries with Pedersen.

Performance: for small tries (below `SmallTrieThreshold` nodes) every edge hash is computed inline. Above the threshold, edge-hash jobs are batched (`parallelHashBatchSize`) and dispatched to a fixed-size worker pool for parallel computation. The scheduler preserves the original job order so the persisted bytes are byte-identical to a natively-built `trie2` — verified end-to-end in the tests by comparing the migrated DB against one built directly through `trie2.Trie.Update` (see `TestMigrationEndToEnd`).

___

### **PR Type**
Enhancement, Tests

___

### **Description**
- Introduce the trie migration phase to convert deprecated Starknet tries (`ClassesTrie`, `StateTrie`, `ContractStorage`) to the `trie2` layout

- Add a hash worker pool and scheduler to compute edge hashes in parallel for large tries

- Update the common pipeline counter to report progress percentages, total elapsed time, and generic units

- Integrate the trie phase into the main state migrator and wipe deprecated buckets upon completion

___

<details> <summary><h3> File Walkthrough</h3></summary>

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Enhancement</strong></td><td><details><summary>7 files</summary><table>
<tr>
  <td><strong>committer.go</strong><dd><code>Add SetProgress method for tracking pipeline progress</code>&nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/NethermindEth/juno/pull/3659/files#diff-4e33a2e7fc8513ba8aaf8a26640dd2a94362333c6b70f96f80d227f868029a81">+7/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>counter.go</strong><dd><code>Support generic unit names and progress percentages in logs</code></dd></td>
  <td><a href="https://github.com/NethermindEth/juno/pull/3659/files#diff-b7a93e2f9b5afe0b40a82a773632b7b13729d122ef98605a1c303ca213fa5526">+69/-21</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>codec.go</strong><dd><code>Add helpers to encode trie2 nodes and compressed paths</code>&nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/NethermindEth/juno/pull/3659/files#diff-3ef65adac2bcfd70da2b191ce2ad87ac6918e2d8e9a64558bbcfde268595efc9">+123/-0</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>hashpool.go</strong><dd><code>Implement worker pool for parallel edge hash computations</code></dd></td>
  <td><a href="https://github.com/NethermindEth/juno/pull/3659/files#diff-5c82803f0030a81b6b53cdcd179e8b1c78f2dd94a264881d3c30b4ce16656035">+62/-0</a>&nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>hashworker.go</strong><dd><code>Add scheduler to batch edge hash jobs for parallel execution</code></dd></td>
  <td><a href="https://github.com/NethermindEth/juno/pull/3659/files#diff-a0b9cb0ee0e64ba50cd42f42113467098a5f5a3c352267972cce66a00bd80658">+174/-0</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>ingestor.go</strong><dd><code>Implement ingestor to traverse and convert deprecated tries</code></dd></td>
  <td><a href="https://github.com/NethermindEth/juno/pull/3659/files#diff-eaa76f0f376b00b966627e072c9dae536e331456a9b786a25606427fcc4b7218">+330/-0</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>trie.go</strong><dd><code>Implement trie Migrator to enumerate and migrate tries</code>&nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/NethermindEth/juno/pull/3659/files#diff-a089426e96c0bf335a82294d75f0bf7dee3dab66870389a6bfdc844949a2c59b">+301/-0</a>&nbsp; </td>

</tr>
</table></details></td></tr><tr><td><strong>Documentation</strong></td><td><details><summary>3 files</summary><table>
<tr>
  <td><strong>constants.go</strong><dd><code>Add formatting and comment about phase memory load</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/NethermindEth/juno/pull/3659/files#diff-8a1c2ad8eef62c726dd2b8f6eca7f6dc279d4d0143edf486fcf0dfc999c3dd31">+5/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>migrator.go</strong><dd><code>Add TODO for future history pipeline progress reporting</code>&nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/NethermindEth/juno/pull/3659/files#diff-ba8fd8f14156264f756bc68d80de8ad726f2bc4ae30d423ab2e698e0c754f9d2">+3/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>storage_ingestor.go</strong><dd><code>Add TODO note regarding broken behavior with pruning</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/NethermindEth/juno/pull/3659/files#diff-14b72478147871e25026167dc1b35f95978daf96c7f18c2045a4a3eafd3d6e46">+1/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>
</table></details></td></tr><tr><td><strong>Tests</strong></td><td><details><summary>4 files</summary><table>
<tr>
  <td><strong>counter_test.go</strong><dd><code>Update counter test to match variable rename</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/NethermindEth/juno/pull/3659/files#diff-11f294764334ec45af59735427097e720f540b969d9a96747eb55f878c050939">+1/-1</a>&nbsp; &nbsp; &nbsp; </td>

</tr>

<tr>
  <td><strong>hashsched_internal_test.go</strong><dd><code>Add tests for in-flight hash batch error handling</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/NethermindEth/juno/pull/3659/files#diff-835f430ba28dfcbb38725e5c2466cfbbb4220933eec0317a35cf451b54bf6d7c">+136/-0</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>trie_test.go</strong><dd><code>Add end-to-end tests for trie migration capabilities</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/NethermindEth/juno/pull/3659/files#diff-594c40578b743798e11957f8a0088b673671fc1cba39ac384c8ebf79fc70b442">+441/-0</a>&nbsp; </td>

</tr>

<tr>
  <td><strong>migrator_test.go</strong><dd><code>Assert deprecated tries are wiped in main migrator tests</code>&nbsp; </dd></td>
  <td><a href="https://github.com/NethermindEth/juno/pull/3659/files#diff-bd328789a00d92ac17e5f1e640a6101c5496a7bcf574b14adcdf7c13a422173b">+10/-4</a>&nbsp; &nbsp; </td>

</tr>
</table></details></td></tr><tr><td><strong>Configuration changes</strong></td><td><details><summary>1 files</summary><table>
<tr>
  <td><strong>migrator.go</strong><dd><code>Integrate trie phase into the main state migrator</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></td>
  <td><a href="https://github.com/NethermindEth/juno/pull/3659/files#diff-c3aca4575cc6a79973e3a968248a412075d7f224987ec912a64168d09dd9dbc5">+2/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>
</table></details></td></tr></tr></tbody></table>

</details>

___


EgeCaner (approved):

EgeCaner (changes requested):
Accidentially approved, I am still reviewing

PR #3660, opened 2026-05-19
perf(core): stream cairo 0 classes

rodrodros:
Instead of storing in a buffer all the bytes of a Cairo 0 Class, chunk them in a local 4kb buffer that every time it gets filled, it streams it into the Starknet Keccak Digest. Avoiding memory spikes and reducing allocations.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -139,6 +139,14 @@ linters:
           - staticcheck
         path: vm/state.go
         text: "ST1003.*_cgo"
+      # Legacy Cairo0 canonical serializer streams ~100+ writes into a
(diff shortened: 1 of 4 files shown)


ongyimeng:
Thanks for suggesting this improvement. I reran `BenchmarkClassV0Hash` with `-count=15`, using both default `GOMAXPROCS` and `GOMAXPROCS=1`.

CPU time looks mostly neutral overall, but the memory improvement is clear:

- smaller fixture: ~459 KB/op -> ~333 KB/op (~27.5% less), 1880 -> 1871 allocs/op
- larger fixture: ~2.49 MB/op -> ~1.97 MB/op (~20.9% less), 12045 -> 12036 allocs/op

PR #3661, opened 2026-05-19
feat(pruner): wallclock min-age retention floor

EgeCaner:
## Summary
  Adds an optional **wallclock minimum-age retention floor** to the pruner. In addition to `--prune-mode`'s block-count window (keep the last N blocks), a block is now also protected if its on-chain timestamp is younger than `--prune-min-age` — i.e. "keep at least the last 1h of chain history" rather than "keep at least the last N blocks". The effective floor is `min(standardFloor, minAgeFloor)`, so whichever window reaches further back wins.

  - New flag `--prune-min-age` (default `1h`, `0` disables, requires `--prune-mode`).
  - Floor is refreshed every `floorTickInterval` (default `15m`) via binary search for the oldest block with `timestamp >= now - minAge`.
  - Suppressed in the L2 catch-up path when the incoming block's own timestamp is already older than `minAge` — during deep sync the cached sample reflects recently-*synced* blocks, not wallclock-recent ones, and would
  otherwise over-protect.

### DIFF
--- a/cmd/juno/juno.go
+++ b/cmd/juno/juno.go
@@ -109,6 +109,7 @@ const (
 	disableReceivedTxnStreamF           = "disable-received-txn-stream"
 	newStateF                           = "new-state"
 	pruneModeF                          = node.PruneModeFlag
+	pruneMinAgeF                        = node.PruneMinAgeFlag
 
 	defaultConfig                             = ""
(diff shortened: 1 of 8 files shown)


rodrodros (approved):
LGTM!

PR #3664, opened 2026-05-20
fix(consensus): clean ProposalStore after commit

RafaelGranza:
### [Issue 3115](https://github.com/NethermindEth/juno/issues/3115)

Adds `ProposalStore.DeleteByHeight`, called from `commitListener.OnCommit` after post-commit hooks finish. Before this, every proposal (winning and losing) at every height stayed in memory forever.

See inline comments for design decisions.

### DIFF
--- a/consensus/driver/commit_listener_test.go
+++ b/consensus/driver/commit_listener_test.go
@@ -1,0 +1,70 @@
+package driver_test
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/NethermindEth/juno/blockchain"
+	"github.com/NethermindEth/juno/builder"
+	"github.com/NethermindEth/juno/consensus/driver"
+	"github.com/NethermindEth/juno/consensus/proposal"
+	"github.com/NethermindEth/juno/consensus/starknet"
+	"github.com/NethermindEth/juno/consensus/types"
+	"github.com/NethermindEth/juno/core"
+	"github.com/NethermindEth/juno/core/felt"
+	"github.com/NethermindEth/juno/core/pending"
+	"github.com/NethermindEth/juno/utils/log"
+	"github.com/stretchr/testify/require"
+)
+
+func TestOnCommit_DropsProposalsAtCommittedHeight(t *testing.T) {
+	proposalStore := &proposal.ProposalStore[starknet.Hash]{}
+	listener := driver.NewCommitListener[starknet.Value, starknet.Hash](
+		log.NewNopZapLogger(),
+		proposalStore,
+	)
+
+	committedHeight := types.Height(42)
+
+	committedValue := felt.FromUint64[starknet.Value](1)
+	losingValue := felt.FromUint64[starknet.Value](2)
+	nextHeightValue := felt.FromUint64[starknet.Value](3)
+
+	proposalStore.Store(committedValue.Hash(), buildResultAtHeight(uint64(committedHeight)))
+	proposalStore.Store(losingValue.Hash(), buildResultAtHeight(uint64(committedHeight)))
+	proposalStore.Store(nextHeightValue.Hash(), buildResultAtHeight(uint64(committedHeight)+1))
+
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
+
+	done := make(chan struct{})
+	go func() {
+		listener.OnCommit(ctx, committedHeight, committedValue)
+		close(done)
+	}()
+
+	block := <-listener.Listen()
+	close(block.Persisted)
+	<-done
+
+	require.Nil(t, proposalStore.Get(committedValue.Hash()))
+	require.Nil(t, proposalStore.Get(losingValue.Hash()))
+	require.NotNil(t, proposalStore.Get(nextHeightValue.Hash()))
+}
+
+func buildResultAtHeight(blockNumber uint64) *builder.BuildResult {
+	return &builder.BuildResult{
+		PreConfirmed: &pending.PreConfirmed{
+			Block: &core.Block{
+				Header: &core.Header{
+					Number: blockNumber,
+				},
+			},
+			StateUpdate: &core.StateUpdate{},
+		},
+		SimulateResult: &blockchain.SimulateResult{
+			BlockCommitments: &core.BlockCommitments{},
+		},
+	}
+}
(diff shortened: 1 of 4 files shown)


consensus/proposal/proposal_store.go
```diff
@@ -31,3 +31,20 @@ func (p *ProposalStore[H]) Store(key H, value *builder.BuildResult) {
 	}
 	_, _ = p.underlying.LoadOrStore(key, value)
 }
+
+func (p *ProposalStore[H]) Delete(key H) {
+	p.underlying.Delete(key)
+}
+
+func (p *ProposalStore[H]) DeleteByHeight(height types.Height) {
```
RafaelGranza: **Design note**: Decided to split into two methods just to keep the API _honest_ without changing internals.

consensus/proposal/proposal_store.go
```diff
@@ -31,3 +31,20 @@ func (p *ProposalStore[H]) Store(key H, value *builder.BuildResult) {
...
+	p.underlying.Range(func(key, value any) bool {
+		buildResult, ok := value.(*builder.BuildResult)
+		if !ok {
+			return true
+		}
+		if types.Height(buildResult.PreConfirmed.Block.Number) == height {
+			p.underlying.Delete(key)
+		}
+		return true
+	})
```
RafaelGranza, line 49: This may not look safe, but:
  - `sync.Map.Range` explicitly allows calling `Delete` from inside the iteration callback (per godoc).
  - The tricky part is that `Range` is not guaranteed to see concurrent writes and deletes. That's fine here: this is only called after the block at `height` is committed, and the validator demux blocks any new proposal stream for heights below the current one (`proposal_stream_demux.go`), so no legitimate writer should be targeting this height anymore.

RafaelGranza: I considered a more thread-safe approach (for future-proofing).

I implemented `ProposalStore` as:
```golang
type ProposalStore[H types.Hash] struct {
	mu       sync.RWMutex
	byHash   map[H]*builder.BuildResult
	byHeight map[types.Height]map[H]struct{}
}
```

I added a new map and a mutex, so I could traverse a map without fearing it could be changed. Also I could `DeleteByHeight` in _O(k)_ instead of _O(n)_. 
_*k being the number of elements of a given height._

The main problem? I benchmarked it, and I was not very happy.

All methods got slower, `Get(key H)` became 17 times slower when trying 10 parallel accesses. It may not be that significant compared to the whole runtime, but introducing delay in a type made to be a fast cache, and also a more confusing implementation because of future-proofing? I didn't buy the trade-off.

consensus/proposal/proposal_store_test.go
```diff
@@ -1,12 +1,14 @@
-package proposal
+package proposal_test
```
RafaelGranza, line 1: 👆

consensus/driver/commit_listener_test.go
```diff
@@ -0,0 +1,70 @@
+package driver_test
```
RafaelGranza, line 1: 👆

consensus/proposal/proposal_store.go
ongyimeng: I think the understanding here may have been that once a height is committed, no proposal stream can still write proposals for that height, so `DeleteByHeight(height)` would be safe as soon as the post-commit hooks return.

I do not think that is guaranteed.

The p2p commit hook does not synchronously stop the proposal stream demux. It only notifies the demux by sending the committed height on `commitNotifier`, then returns: https://github.com/NethermindEth/juno/blob/main/consensus/p2p/p2p.go#L190-L195

The demux processes that notification later. Only when the demux receives the commit notification does it stop the current-height streams: https://github.com/NethermindEth/juno/blob/main/consensus/p2p/validator/proposal_stream_demux.go#L107-L118 https://github.com/NethermindEth/juno/blob/main/consensus/p2p/validator/proposal_stream_demux.go#L148-L164

Meanwhile, `commitListener` waits for the post-commit hooks to return and then proceeds with cleanup: https://github.com/NethermindEth/juno/blob/main/consensus/driver/commit_listener.go#L74-L82

That creates a timing window: the p2p hook may have returned, but the demux may not yet have processed the commit notification or stopped the same-height streams.

During that window, an already-running same-height stream can still finish and store into `ProposalStore`: https://github.com/NethermindEth/juno/blob/main/consensus/p2p/validator/proposal_stream.go#L119

So the issue is not mainly about brand-new lower- or same-height streams starting later. The issue is about already-existing same-height streams that were still in flight when cleanup ran.

In short: the hook notifies the demux to shut down those streams, it does not prove that the demux has completed that shutdown before cleanup runs.

cc: @rodrodros would appreciate your second opinion on this.

ongyimeng:
Apart from the one possible issue raised, this LGTM to me!

consensus/proposal/proposal_store.go
RafaelGranza: Good catch, you're right that the p2p hook only notifies the demux without waiting for it to drain. An in-flight same-height stream can still reach `proposalStore.Store(...)` between the hook returning and the demux processing `commitNotifier`.

Either cleanup approach I had on the table (`Delete(key)` or `DeleteByHeight(H)`) would hit this same race, since both run before the demux finishes its shutdown. So I'm widening the cleanup instead of trying to time it: `DeleteByHeight(H)` is renamed to `DeleteUpToHeight(H)` and now removes everything with _height <= H_.

A straggler from _height H_ can still slip in for a moment, but the next commit (_H+1_) sweeps it, so the leak window is bounded to one block instead of unbounded.

Added `TestProposalStore_DeleteUpToHeight_SweepsPriorHeights` to pin the new semantic.

ongyimeng (approved):

consensus/proposal/proposal_store.go
ongyimeng: Question: This delete method looks unused. Would it make sense to remove it if it’s no longer needed?

rodrodros (commented):
Give me a sec to look at this before merging. It feels unsafe to me that different write operations are done concurrently

rodrodros (approved):
There are more underlying issues with `ProposalStore` that this PR on itself doesn't completely fix. They will be solved at a later stage. 

Merging for now

PR #3665, opened 2026-05-20
chore(core): use  `package core_test` for tests

rodrodros:


core/deprecated_cairo_program_test.go
```diff
@@ -1,13 +1,14 @@
-package core
+package core_test
```
RafaelGranza, line 2: 👀

RafaelGranza (commented):
LGTM!

RafaelGranza (approved):

PR #3666, opened 2026-05-21
fix: preserve legacy serialization for class hashes

ongyimeng:
## Main Issues Resolved

1. Fixed legacy `cairo_type` spacing serialization

   Pre-0.10 Cairo0 artifacts can require `" : "` spacing in hinted-hash payloads.

   Juno now preserves this legacy formatting with an idempotent spacing patcher, so already-correct strings are not changed. Thanks @brbrr for the code suggestion.

2. Preserved `abi: null` vs `abi: []`

   Deprecated Cairo classes rely on the distinction between a null ABI and an empty ABI array for hashing.

   Juno now preserves that distinction during legacy ABI parsing.

3. Applied legacy spacing to nested references

   Legacy spacing compatibility is now also applied to `references[*].value` strings, not only top-level payloads.


core/deprecated_cairo_program.go
```diff
@@ -389,7 +389,7 @@ func writeDeprecatedCairoProgramCanonical(
 	writeJSONFieldPrefix(buffer, "prime", &first)
 	writeJSONString(buffer, program.Prime)
 	writeJSONFieldPrefix(buffer, "reference_manager", &first)
-	writeLegacyReferenceManager(buffer, program.ReferenceManager)
+	writeLegacyReferenceManager(buffer, program.ReferenceManager, program.CompilerVersion == nil)
```
brbrr: nit: extract the bool into a variable to make the intent clearer. something along these lines:

```
isLegacyFormat := program.CompilerVersion == nil
writeLegacyReferenceManager(buffer, rm, isLegacyFormat)
```

brbrr (approved):
The code looks good, and I've tested it against cairo0 classes, and it matches hashes stored in the DB

core/deprecated_cairo_program.go
rodrodros: Add a comment description on top explaining in-code what legacy format is

core/deprecated_cairo_program.go
```diff
@@ -747,14 +757,35 @@ type legacyReference struct {
...
+
+		builder.WriteString(" : ")
+		index++
+		changed = true
+	}
+	if !changed {
+		return "", false
+	}
+	return builder.String(), true
 }
```
rodrodros, line 801: Code is being scanned twice, the first one looking for `": "` and the second one re-building a new string which after full reconstruction might never be used again. 

I suggest the following approach only does one scan at worst and we only build the string if we are 100% there is a change:

1. first loop looking for " :" and ": ". 
2. nothing found, quick return (one scan)
3. If Found at index `k`
4. Move to a next loop that rebuilds the rest of the string correctly and start from `k`
5. Return the concatenation of what has been scanned + what has been rebuilt.

rodrodros (approved):
Looks good to me, please wait until @brbrr approval for merge

(I see that it has already been)

core/deprecated_cairo_program.go
ongyimeng: I chose slightly longer, explicit names like `isPre010LegacyFormat` and `patchPre010CairoTypeSpacing` to make the pre-0.10 cairo spacing quirk intent self-documenting.

PR #3674, opened 2026-05-22
refactor(lru): switch LRU implementation to golang-lru

brbrr:
Replaces geth LRU implementation with golang-lru library. Adds a wrapper package, which exposes only API that are currently used.


rodrodros (commented):
Looks good, looking forward to see some benchmarks, and why SimpleLRU was used in someplaces. What about using 2 queues LRUs as well?

brbrr:
@rodrodros I've assumed that geth was using golang-lru under the hood, so I thought it would be 1-1 migration. I've double checked, and they use their own impl. I'll run the tests and post it here as a comment.

utils/lru.go
rodrodros, line 1: I am thinking it is better to create a package inside utils/lru/lru.go so that imports are called `lru` by default. We would wrap whatever implementation we used from there

brbrr:
tl;dr: golang-lru is ~20–35% slower per-op + 1 alloc per Add. Absolute deltas are ns-scale. While the implementation is slower, I think it's fine, since the callsites are not bottlenecked by the LRU performance, but rather by IO/network. 
  
  ### Latency (ns/op)

  | Workload | Key / Size | old (ethereum) | new (hashicorp) | Δ | p |
  |---|---|---:|---:|---:|---:|
  | Add | int / 128 | 78.85 ± 3% | 104.35 ± 3% | **+32.3%** | 0.000 |
  | Add | int / 8192 | 102.2 ± 4% | 138.5 ± 4% | **+35.5%** | 0.000 |
  | Add | bloomKey / 16 | 66.45 ± 3% | 89.70 ± 4% | **+35.0%** | 0.000 |
  | Add | feltKey / 128 | 85.81 ± 3% | 110.00 ± 1% | **+28.2%** | 0.000 |
  | GetHit | int / 128 | 18.04 ± 2% | 24.27 ± 2% | +34.5% | 0.000 |
  | GetHit | int / 8192 | 23.89 ± 6% | 28.20 ± 1% | +18.0% | 0.000 |
  | GetHit | bloomKey / 16 | 21.43 ± 1% | 27.93 ± 4% | +30.3% | 0.000 |
  | GetHit | feltKey / 128 | 28.80 ± 2% | 32.22 ± 1% | +11.9% | 0.000 |
  | GetMiss | int / 128 | 18.29 ± 3% | 24.31 ± 2% | +32.9% | 0.000 |
  | GetMiss | bloomKey / 16 | 17.91 ± 0% | 24.86 ± 15% | +38.8% | 0.000 |
  | GetMiss | feltKey / 128 | 21.03 ± 6% | 26.07 ± 3% | +24.0% | 0.000 |
  | Mixed 80/20 | int / 128 | 52.16 ± 4% | 56.95 ± 1% | +9.2% | 0.001 |
  | Mixed 80/20 | bloomKey / 16 | 48.20 ± 12% | 53.02 ± 6% | +10.0% | 0.022 |
  | Mixed 80/20 | feltKey / 128 | 63.42 ± 10% | 70.09 ± 8% | +10.5% | 0.023 |
  | Parallel mixed | int / 128 | 233.0 ± 1% | 289.8 ± 2% | +24.4% | 0.000 |
  | Parallel mixed | bloomKey / 16 | 210.8 ± 8% | 232.7 ± 2% | +10.4% | 0.000 |
  | Parallel mixed | feltKey / 128 | 277.5 ± 5% | 319.5 ± 5% | +15.1% | 0.000 |
  | SimpleAdd (no lock) | feltKey / 10 | 69.34 ± 3% | 89.58 ± 6% | +29.2% | 0.000 |
  | SimpleGetHit (no lock) | feltKey / 10 | 24.06 ± 9% | 24.12 ± 5% | ~ | 0.469 |
  | SimpleMixed (no lock) | feltKey / 10 | 52.83 ± 10% | 56.30 ± 11% | ~ | 0.971 |
  | **geomean** | | **51.33** | **62.26** | **+21.3%** | |

  ### Memory (B/op, allocs/op)

  | Workload | Key / Size | old B/op | new B/op | old allocs | new allocs |
  |---|---|---:|---:|---:|---:|
  | Add | int / 128 | 0 | **80** | 0 | **1** |
  | Add | int / 8192 | 0 | **80** | 0 | **1** |
  | Add | bloomKey / 16 | 0 | **80** | 0 | **1** |
  | Add | feltKey / 128 | 0 | **96** | 0 | **1** |
  | GetHit (any) | — | 0 | 0 | 0 | 0 |
  | GetMiss (any) | — | 0 | 0 | 0 | 0 |
  | Mixed | int / 128 | 0 | 8 | 0 | 0 |
  | Mixed | bloomKey / 16 | 0 | 8 | 0 | 0 |
  | Mixed | feltKey / 128 | 0 | 9 | 0 | 0 |
  | Parallel | int / 128 | 0 | 8 | 0 | 0 |
  | Parallel | bloomKey / 16 | 0 | 8 | 0 | 0 |
  | Parallel | feltKey / 128 | 0 | 9 | 0 | 0 |
  | SimpleAdd | feltKey / 10 | 0 | **96** | 0 | **1** |
  | SimpleGetHit | feltKey / 10 | 0 | 0 | 0 | 0 |
  | SimpleMixed | feltKey / 10 | 0 | 9 | 0 | 0 |

  ### Summary

  | Path | Cost change |
  |---|---|
  | Reads (hit/miss) | +12% to +39% latency, 0 allocs |
  | Writes (Add) | +28% to +35% latency, **+1 heap alloc / 80–96 B per Add** |
  | Mixed 80/20 R/W | +9% to +11% |
  | Concurrent | +10% to +24% under contention |
  | Overall (geomean) | **+21.3%** |

blockchain/aggregated_bloom_filter_cache_test.go
```diff
@@ -147,7 +147,7 @@ func populateAggregatedBloomDeterministic(
 
 func TestMatchBlockIterator_InsertAndQueryRandomEvents(t *testing.T) {
 	numEvents := 64
-	numAggregatedBloomFilters := uint64(16)
+	var numAggregatedBloomFilters uint64 = blockchain.AggregatedBloomFilterCacheSize
```
rodrodros: Why?

blockchain/blockchain.go
```diff
@@ -129,7 +129,7 @@ func New(database db.KeyValueStore, network *networks.Network, opts ...Option) *
 		opt(&o)
 	}
 
-	cachedFilters := NewAggregatedBloomCache(AggregatedBloomFilterCacheSize)
+	cachedFilters := NewAggregatedBloomCache()
```
rodrodros: Why remove the input param?

rodrodros (commented):
I see that the API has been degraded in some cases, please restore it. I am not sure why you did it, can you explain?

utils/lru.go
```diff
@@ -0,0 +1,51 @@
...
+	if err != nil {
+		panic(fmt.Errorf("simplelru: %w (size=%d)", err, size))
+	}
+	return &SimpleLRU[K, V]{c: c}
+}
+
+func (l *SimpleLRU[K, V]) Add(key K, value V) (evicted bool) { return l.c.Add(key, value) }
+func (l *SimpleLRU[K, V]) Get(key K) (V, bool)               { return l.c.Get(key) }
+func (l *SimpleLRU[K, V]) Remove(key K) (present bool)       { return l.c.Remove(key) }
+func (l *SimpleLRU[K, V]) Purge()                            { l.c.Purge() }
```
rodrodros: Not tested

blockchain/blockchain.go
brbrr: This constructor is used _only_ with a comptime const:

```
// NOTE(Ege): consider making it configurable
const AggregatedBloomFilterCacheSize = 16
```

I'm not sure what the rationale was to make the cache size runtime configurable, but currently it's not being used anywhere. If we'd leave it as is, we'd make the constructor to panic on invalid size values (<= 0), or we'd have to do a size check, and return and propagate error across the stack i.e. changing return signature of `blockchain.New` across 40-ish files.

So I've decided to remove the configurationability until it is needed.

rodrodros: Add the configuration, panic on invalid size. There is no fear of panicking here, since it means it is a programmer error.

utils/lru.go
rodrodros: I see this comment was marked as resolved without being resolved, why?

brbrr: I think it was updated in https://github.com/NethermindEth/juno/pull/3674/changes/3ad6be2e0b81055b6bd89850ed33b840b9b6c72a. Let me know if that's not enough, or it's not what you meant.

blockchain/blockchain.go
brbrr: reverted in [1d8d578](https://github.com/NethermindEth/juno/pull/3674/commits/1d8d5783ec420627e24071b47702a5583c54acca)

utils/lru.go
brbrr: addressed in [4354dd1](https://github.com/NethermindEth/juno/pull/3674/commits/4354dd1d15b64bcabc01b1a8601e88bb1a583881)

sync/pending_polling.go
```diff
@@ -219,7 +219,7 @@ func (s *Synchronizer) pollPreLatest(ctx context.Context, out chan<- *pending.Pr
 			deliveredForHead = s.handleTickerPreLatest(
 				ctx,
 				currentHead,
-				&seenByParent,
+				seenByParent,
```
RafaelGranza, line 222: It is great that you made `NewSimpleLRU` to return a ptr.
This way it gets harder in the future to accidentally have two copies of the same `LRUcache`.

blockchain/aggregated_bloom_filter_cache.go
```diff
@@ -27,15 +27,18 @@ type EventFiltersCacheKey struct {
...
 type AggregatedBloomFilterCache struct {
-	cache        lru.Cache[EventFiltersCacheKey, *core.AggregatedBloomFilter]
+	cache        *utils.LRU[EventFiltersCacheKey, *core.AggregatedBloomFilter]
 	fallbackFunc func(EventFiltersCacheKey) (core.AggregatedBloomFilter, error)
 }
 
 // NewAggregatedBloomCache creates a new LRU cache for aggregated bloom filters
 // with the specified maximum size (number of ranges to cache).
 func NewAggregatedBloomCache(size int) AggregatedBloomFilterCache {
 	return AggregatedBloomFilterCache{
```
RafaelGranza, line 37: I know this is out of scope, but since we are touching it, could `NewAggregatedBloomCache` return `*AggregatedBloomFilterCache`? 

Its methods are all pointer-receivers, so a pointer is more consistent and avoids copies diverging on `fallbackFunc` in future uses.

Also, it's only called in `blockchain.go` and tests today, so the change stays small.

RafaelGranza (approved):
LGTM. 

About the performance deltas, they are **_ns-scale_**, small enough to sit near measurement noise, and these caches aren't on a hot path anyway (IO/network-bound), so I don't consider the "regression" a concern. 

I only left one nit below:

utils/lru.go
rodrodros: Maybe a git error, is not showing up in the diff

blockchain/aggregated_bloom_filter_cache.go
rodrodros: > Its methods are all pointer-receivers, so a pointer is more consistent 

This doesn't make sense to me, can you explain what you mean? How having pointer receivers affects what the constructor should return? 

> avoids copies diverging on fallbackFunc in future uses.

This is a valid one. 

>  could NewAggregatedBloomCache return *AggregatedBloomFilterCache?

Out of scope changes should be address in a different PR, let's try not squeeze them here

utils/lru.go
rodrodros: @brbrr to clarify, we should have a `utils/lru/lru.go`, but what I see is `utils/lru.go`? Is git or the code wrong

brbrr: ah, ok. I missed the `lru` folder in the middle. addressed in [f48a4c9](https://github.com/NethermindEth/juno/pull/3674/commits/f48a4c9e93fad62f36a816c8862a09fd64ed1522)

Issue #3670, opened 2026-05-22
`ProposalStore`: explore finalized-height guard and height-indexed storage


RafaelGranza:
`ProposalStore` works fine for the current use cases, but as new callers are added it risks turning into spaghetti: coordination rules (ordering, what's safe to write/delete and when) live across call sites instead of inside the type. Goal: push those rules into the store itself.

Some ideas:

- **Finalized-height guard.** `finalized atomic.Uint64` updated via CAS. `Finalized(h)` bumps the cursor and drops everything `<= h`. `Store` at a height `<= finalized` becomes a no-op.
- **Height-indexed storage.** `Height -> Hash -> *BuildResult` (per-height buckets) so deletion is O(heights removed) instead of a full scan, and the guard above becomes trivial.
- **API takes height.** `Store(height, hash, br)` so the invariant can actually be enforced.
- **Benchmarks.** Compare variants in [consensus/proposal/proposal_store_bench_test.go](consensus/proposal/proposal_store_bench_test.go), including `DeleteUpToHeight`/`Finalized` pressure.
- **Single-owner goroutine (later).** `atomic.Pointer[map]` snapshot for lock-free reads. Bigger refactor; revisit only if the simpler changes above aren't enough.

PR #3675, opened 2026-05-24
refactor(consensus/proposal): encapsulate ordering rules in store

RafaelGranza:
## Summary ([ISSUE](https://github.com/NethermindEth/juno/issues/3670))

`ProposalStore` was leaking coordination rules (ordering, what's safe to write/delete and when) across call sites. This refactor pushes those rules into the type itself so new callers can't violate them.

## What changed

- **Storage layout**: flat `sync.Map[Hash]*BR` to nested `sync.Map[Height]*sync.Map[Hash]*BR`.
- **Finalized cursor**: `atomic.Uint64` advanced monotonically via CAS in `DeleteUpToHeight`.
- **Store**: no-op if `height <= finalized`; recheck-then-self-clean after publish to handle races with a concurrent sweep.
- **Get**: late `height <= finalized` check before returning, so readers never observe entries past the cursor even during a race.
- **IsFinalized(height)**: utility for callers that want to distinguish cache miss from finalized-height query.
- **DeleteUpToHeight**: now O(num_heights). Drops outer entries; GC handles the inner sync.Maps.

The public API (`Get(hash)`, `Store(hash, br)`, `DeleteUpToHeight(h)`) is unchanged for callers; no signature changes leak into the `tendermint.Application` interface.

## Design note

There was a real trade-off in how to balance cost between Get and Delete:

| Layout | Get | Delete |
|---|---|---|
| Flat `sync.Map[Hash]*BR` | O(1) | O(N) where N = total entries |
| Nested `sync.Map[Height]*sync.Map[Hash]*BR` (this PR) | O(num_heights) | O(num_heights) |

Picked the nested layout because num_heights is small in practice (1 to 3 in steady state, up to ~5 with stragglers), so the O(num_heights) `Get` is sub-100ns and invisible end-to-end (network and VM latency dominate). The benchmarks below confirm the `Get` hit stays in the tens of nanoseconds.

The flip side is meaningful: `Delete` cost is decoupled from total entry count. That's where the big speedup comes from.

## Benchmark comparison

Apple M-series, `go test -bench -benchmem`. Spread across 5 live heights to reflect realistic steady-state consensus.

| Workload | New (bucketed) | Legacy (flat) | Delta |
|---|---:|---:|---|
| StoreThenGet (seq) | 586ns | 558ns | ~noise |
| ParallelGet | 24ns | 12ns | ~2x slower, sub-30ns |
| ParallelMixed (10% writes) | 51ns | 31ns | ~1.6x slower, sub-100ns |
| **DeleteUpToHeight (16 heights x 64 hashes)** | **1.6µs** | **101µs** | **~62x faster** |

The benchmark file (`proposal_store_bench_test.go`) is included as requested for future comparisons.

### DIFF
--- a/consensus/proposal/proposal_store.go
+++ b/consensus/proposal/proposal_store.go
@@ -9,68 +9,68 @@ import (
 )
 
 type ProposalStore[H types.Hash] struct {
-	byHeight  sync.Map // map[types.Height]*sync.Map[H]*builder.BuildResult
-	finalized atomic.Uint64
+	proposalsByHeight sync.Map // height -> hash -> BuildResult
+	finalizedHeight   atomic.Uint64
 }
 
 func (p *ProposalStore[H]) Get(key H) *builder.BuildResult {
 	var found *builder.BuildResult
-	p.byHeight.Range(func(h, bucket any) bool {
-		v, ok := bucket.(*sync.Map).Load(key)
+	p.proposalsByHeight.Range(func(height, proposals any) bool {
+		buildResult, ok := proposals.(*sync.Map).Load(key)
 		if !ok {
 			return true
 		}
-		if uint64(h.(types.Height)) <= p.finalized.Load() {
+		if p.IsFinalized(height.(types.Height)) {
 			return false
 		}
-		found = v.(*builder.BuildResult)
+		found = buildResult.(*builder.BuildResult)
 		return false
 	})
 	return found
 }
 
-func (p *ProposalStore[H]) Store(key H, value *builder.BuildResult) {
-	if value == nil {
+func (p *ProposalStore[H]) Store(key H, builtResult *builder.BuildResult) {
+	if builtResult == nil {
 		return
 	}
-	height := types.Height(value.PreConfirmed.Block.Number)
-	if uint64(height) <= p.finalized.Load() {
+	height := types.Height(builtResult.PreConfirmed.Block.Number)
+	if p.IsFinalized(height) {
 		return
 	}
-	bucket := p.bucketFor(height)
-	bucket.LoadOrStore(key, value)
+	bucket := p.proposalFor(height)
+	bucket.LoadOrStore(key, builtResult)
 
-	if uint64(height) <= p.finalized.Load() {
-		bucket.CompareAndDelete(key, value)
-		p.byHeight.CompareAndDelete(height, bucket)
+	if p.IsFinalized(height) {
+		bucket.CompareAndDelete(key, builtResult)
+		p.proposalsByHeight.CompareAndDelete(height, bucket)
 	}
 }
 
-func (p *ProposalStore[H]) bucketFor(height types.Height) *sync.Map {
-	if existing, ok := p.byHeight.Load(height); ok {
+func (p *ProposalStore[H]) proposalFor(height types.Height) *sync.Map {
+	if existing, ok := p.proposalsByHeight.Load(height); ok {
 		return existing.(*sync.Map)
 	}
-	actual, _ := p.byHeight.LoadOrStore(height, &sync.Map{})
+	actual, _ := p.proposalsByHeight.LoadOrStore(height, &sync.Map{})
 	return actual.(*sync.Map)
 }
 
 func (p *ProposalStore[H]) IsFinalized(height types.Height) bool {
-	return uint64(height) <= p.finalized.Load()
+	return uint64(height) <= p.finalizedHeight.Load()
 }
 
-func (p *ProposalStore[H]) DeleteUpToHeight(height types.Height) {
+func (p *ProposalStore[H]) FinalizeHeight(height types.Height) {
 	for {
-		cur := p.finalized.Load()
+		cur := p.finalizedHeight.Load()
 		if uint64(height) <= cur {
 			return
 		}
-		if p.finalized.CompareAndSwap(cur, uint64(height)) {
+		if p.finalizedHeight.CompareAndSwap(cur, uint64(height)) {
 			break
 		}
 	}
-	p.byHeight.Range(func(h, _ any) bool {
+	p.proposalsByHeight.Range(func(h, _ any) bool {
 		if h.(types.Height) <= height {
-			p.byHeight.Delete(h)
+			p.proposalsByHeight.Delete(h)
 		}
 		return true
 	})
--- a/consensus/proposal/proposal_store.go
+++ b/consensus/proposal/proposal_store.go
@@ -30,9 +30,6 @@ func (p *ProposalStore[H]) Get(key H) *builder.BuildResult {
 }
 
 func (p *ProposalStore[H]) Store(key H, builtResult *builder.BuildResult) {
-	if builtResult == nil {
-		return
-	}
 	height := types.Height(builtResult.PreConfirmed.Block.Number)
 	if p.IsFinalized(height) {
 		return
@@ -40,6 +37,8 @@ func (p *ProposalStore[H]) Store(key H, builtResult *builder.BuildResult) {
 	bucket := p.proposalFor(height)
 	bucket.LoadOrStore(key, builtResult)
 
+	// FinalizeHeight may have raced between the initial guard and LoadOrStore;
+	// if so, roll back so we never leave a proposal at a finalized height.
 	if p.IsFinalized(height) {
 		bucket.CompareAndDelete(key, builtResult)
 		p.proposalsByHeight.CompareAndDelete(height, bucket)
@@ -47,6 +46,8 @@ func (p *ProposalStore[H]) Store(key H, builtResult *builder.BuildResult) {
 }
 
 func (p *ProposalStore[H]) proposalFor(height types.Height) *sync.Map {
+	// Fast path avoids allocating a sync.Map on every call: LoadOrStore would
+	// construct one even when the bucket already exists, only to discard it.
 	if existing, ok := p.proposalsByHeight.Load(height); ok {
 		return existing.(*sync.Map)
 	}
--- a/consensus/proposal/proposal_store.go
+++ b/consensus/proposal/proposal_store.go
@@ -1,49 +1,76 @@
 package proposal
 
 import (
-	syncmap "sync"
+	"sync"
+	"sync/atomic"
 
 	"github.com/NethermindEth/juno/builder"
 	"github.com/NethermindEth/juno/consensus/types"
 )
 
 type ProposalStore[H types.Hash] struct {
-	underlying syncmap.Map
+	byHeight  sync.Map // map[types.Height]*sync.Map[H]*builder.BuildResult
+	finalized atomic.Uint64
 }
 
 func (p *ProposalStore[H]) Get(key H) *builder.BuildResult {
-	value, ok := p.underlying.Load(key)
-	if !ok {
-		return nil
-	}
-
-	buildResult, ok := value.(*builder.BuildResult)
-	if !ok {
-		return nil
-	}
-
-	return buildResult
+	var found *builder.BuildResult
+	p.byHeight.Range(func(h, bucket any) bool {
+		v, ok := bucket.(*sync.Map).Load(key)
+		if !ok {
+			return true
+		}
+		if uint64(h.(types.Height)) <= p.finalized.Load() {
+			return false
+		}
+		found = v.(*builder.BuildResult)
+		return false
+	})
+	return found
 }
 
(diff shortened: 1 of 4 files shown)


consensus/proposal/proposal_store.go
```diff
@@ -1,49 +1,76 @@
...
+	"sync"
+	"sync/atomic"
 
 	"github.com/NethermindEth/juno/builder"
 	"github.com/NethermindEth/juno/consensus/types"
 )
 
 type ProposalStore[H types.Hash] struct {
-	underlying syncmap.Map
+	byHeight  sync.Map // map[types.Height]*sync.Map[H]*builder.BuildResult
```
rodrodros: If it is going to be a comment, natural language is better. Just a suggestion, the goal is that it is clear what it is, and not use unnecessary symbols
```
// height -> hash -> BuildResult
```

Also, I am not sure the name `byHeight` is a good name, maybe call it just `data` or another name? If there is a `byHeight` then there should be also a `bySmthElse`. Another name suggestion is just `proposalsByHeight`.

consensus/proposal/proposal_store.go
```diff
@@ -1,49 +1,76 @@
...
+	finalized atomic.Uint64
```
rodrodros: Please call it `finalizedHeight` since it more clear name

consensus/proposal/proposal_store.go
```diff
@@ -1,49 +1,76 @@
...
-
-	return buildResult
+	var found *builder.BuildResult
+	p.byHeight.Range(func(h, bucket any) bool {
+		v, ok := bucket.(*sync.Map).Load(key)
+		if !ok {
+			return true
+		}
+		if uint64(h.(types.Height)) <= p.finalized.Load() {
+			return false
```
rodrodros, line 24: Please add a test so that code execution returns at line 24

consensus/proposal/proposal_store.go
```diff
@@ -1,49 +1,76 @@
...
+		return false
+	})
+	return found
 }
 
 func (p *ProposalStore[H]) Store(key H, value *builder.BuildResult) {
 	if value == nil {
 		return
 	}
-	_, _ = p.underlying.LoadOrStore(key, value)
```
rodrodros: Let's remove this guard in a follow up. If `value == nil` means this function shouldn't have been called in the first  place. I know this is old code, can be done in a follow up

consensus/proposal/proposal_store.go
```diff
@@ -1,49 +1,76 @@
...
+	height := types.Height(value.PreConfirmed.Block.Number)
+	if uint64(height) <= p.finalized.Load() {
+		return
+	}
+	bucket := p.bucketFor(height)
+	bucket.LoadOrStore(key, value)
+
+	if uint64(height) <= p.finalized.Load() {
+		bucket.CompareAndDelete(key, value)
+		p.byHeight.CompareAndDelete(height, bucket)
```
rodrodros: No tests that check this path

consensus/proposal/proposal_store.go
```diff
@@ -1,49 +1,76 @@
...
+	}
+}
+
+func (p *ProposalStore[H]) bucketFor(height types.Height) *sync.Map {
+	if existing, ok := p.byHeight.Load(height); ok {
+		return existing.(*sync.Map)
+	}
```
rodrodros, line 53: Isn't this `if` redundant?

consensus/proposal/proposal_store.go
```diff
@@ -1,49 +1,76 @@
...
 	}
-	_, _ = p.underlying.LoadOrStore(key, value)
+	height := types.Height(value.PreConfirmed.Block.Number)
+	if uint64(height) <= p.finalized.Load() {
+		return
+	}
+	bucket := p.bucketFor(height)
+	bucket.LoadOrStore(key, value)
+
+	if uint64(height) <= p.finalized.Load() {
```
rodrodros: Why not used `IsFinalized` here?

consensus/proposal/proposal_store.go
```diff
@@ -1,49 +1,76 @@
...
-	}
-
-	return buildResult
+	var found *builder.BuildResult
+	p.byHeight.Range(func(h, bucket any) bool {
+		v, ok := bucket.(*sync.Map).Load(key)
+		if !ok {
+			return true
+		}
+		if uint64(h.(types.Height)) <= p.finalized.Load() {
```
rodrodros: Why not used `IsFinalized` here?

consensus/proposal/proposal_store.go
rodrodros: Why not used `IsFinalized` here?

consensus/proposal/proposal_store_bench_test.go
```diff
@@ -0,0 +1,188 @@
...
+	"github.com/NethermindEth/juno/builder"
+	"github.com/NethermindEth/juno/consensus/proposal"
+	"github.com/NethermindEth/juno/consensus/starknet"
+	"github.com/NethermindEth/juno/consensus/types"
+	"github.com/NethermindEth/juno/core/felt"
+)
+
+type legacyStore[H types.Hash] struct {
+	underlying sync.Map
+}
```
rodrodros: Delete all this legacy code please. The way to benchmark is to go to a previous version and doing the benchmarks to compare, not keeping this code.

consensus/proposal/proposal_store.go
```diff
@@ -1,49 +1,76 @@
...
+	return actual.(*sync.Map)
 }
 
-func (p *ProposalStore[H]) Delete(key H) {
-	p.underlying.Delete(key)
+func (p *ProposalStore[H]) IsFinalized(height types.Height) bool {
+	return uint64(height) <= p.finalized.Load()
 }
 
 func (p *ProposalStore[H]) DeleteUpToHeight(height types.Height) {
```
rodrodros: Call it `FinalizeHeight`, since it is not just deleting, it is also no-oping any other operation below this `height`. It makes the function name carry more meaning

consensus/proposal/proposal_store.go
rodrodros: Shouldn't you check that it is finalized before even storing anything?

rodrodros:
@RafaelGranza the benchmark only shows performance data, but there is no memory usage benchmark, could you please add them as well

consensus/proposal/proposal_store.go
```diff
@@ -1,49 +1,76 @@
...
-	}
-
-	buildResult, ok := value.(*builder.BuildResult)
-	if !ok {
-		return nil
-	}
-
-	return buildResult
+	var found *builder.BuildResult
+	p.byHeight.Range(func(h, bucket any) bool {
```
rodrodros: `bucket any` is very ambiguous. Lets call this `proposals any` at least

consensus/proposal/proposal_store.go
```diff
@@ -1,49 +1,76 @@
...
+		v, ok := bucket.(*sync.Map).Load(key)
```
rodrodros: let's call `v` `buildResult` since that what we intend it to be

consensus/proposal/proposal_store.go
rodrodros: `h` is also ambiguous, as a general rule, single named vars are not good, please call it `height`.

consensus/proposal/proposal_store.go
```diff
@@ -1,49 +1,76 @@
...
+		if uint64(h.(types.Height)) <= p.finalized.Load() {
+			return false
+		}
+		found = v.(*builder.BuildResult)
+		return false
+	})
+	return found
 }
 
 func (p *ProposalStore[H]) Store(key H, value *builder.BuildResult) {
```
rodrodros: let's call `value` `builtResult` instead. I know this part is old code

consensus/proposal/proposal_store.go
```diff
@@ -1,49 +1,76 @@
...
+	bucket := p.bucketFor(height)
+	bucket.LoadOrStore(key, value)
+
+	if uint64(height) <= p.finalized.Load() {
+		bucket.CompareAndDelete(key, value)
+		p.byHeight.CompareAndDelete(height, bucket)
+	}
+}
+
+func (p *ProposalStore[H]) bucketFor(height types.Height) *sync.Map {
```
rodrodros: Let's call this `proposalFor`

consensus/proposal/proposal_store.go
RafaelGranza: These lines are only reachable during race conditions, we got a test `TestProposalStore_ConcurrentStoreAndDeleteUpToHeight` that tries to trigger this behavior.

consensus/proposal/proposal_store.go
RafaelGranza: This is here to handle a race condition.

We can't atomically combine the finalized check with the `LoadOrStore`: between the guard above and the insertion, another goroutine can advance `finalizedHeight` via `FinalizeHeight`. The re-check rolls back the insertion when that happens.

This way the function always leaves the store in a valid state, which is the key when designing lock-free data structures.

consensus/proposal/proposal_store.go
RafaelGranza: Also a race condition path

consensus/proposal/proposal_store.go
RafaelGranza: You're right that it's redundant for correctness, `LoadOrStore` alone would behave the same way. 

I kept it as a fast-path optimization: without the `Load`, every call allocates a fresh `&sync.Map{}` on the heap, even when the bucket already exists, and that allocation immediately becomes garbage because `LoadOrStore` returns the existing one. With the `Load`, the hot path (bucket already exists, which is the common case after the first `Store` at a given height) does a single read with zero allocations.

Since `Store` runs in the consensus loop, I think avoiding the per-call allocation is worth the extra line. Happy to drop it if you prefer the simpler version though.

consensus/proposal/proposal_store.go
RafaelGranza: I'd actually keep this guard. `*builder.BuildResult`is a pointer type, so `nil` is a valid value and callers can pass it (intentionally or by mistake). Without this check we'd panic on the next line dereferencing `builtResult.PreConfirmed.Block.Number`.

Storing `nil` would also be a no-op semantically: `Get(key)` already returns `nil` for missing keys, so there's no observable difference between "`Store(key, nil)`" and "never called `Store` at all". The guard just makes that explicit and cheap at the API boundary instead of forcing every caller to check before calling.

consensus/proposal/proposal_store_bench_test.go
RafaelGranza: Also adding memory usage benchmarks

consensus/proposal/proposal_store.go
ongyimeng: Interesting. did you happen to check whether `&sync.Map{}` escapes here, e.g. with `go build -gcflags="-m" ./consensus/proposal`? If it does, a short code comment would be helpful since the extra `Load` can look redundant at first glance. A tiny `benchmem` comparison could also be nice, but I wouldn’t block on that.

ongyimeng:
Nice refactor, the abstraction looks cleaner with these rules inside `ProposalStore`. I also like the atomic cursor + `sync.Map` approach, the concurrency design is quite elegant.

ongyimeng (approved):

consensus/proposal/proposal_store.go
rodrodros: @RafaelGranza leave a comment on the code explaining the rationale please

consensus/proposal/proposal_store.go
rodrodros: Leave a comment clarifying this

consensus/proposal/proposal_store.go
rodrodros: My question stands, why would you call *builder.BuildResult` with nil? You wouldn't so, it doesn't makes sense to guard against it. If we were going to guard against every instance of possible nil value code base would double in size.

The only time input parameters deserve to be sanitized is when there is external input, so far this function is only used by us.

rodrodros (approved):
LGTM! Merge at will

consensus/proposal/proposal_store.go
RafaelGranza: Agree, removing it

PR #3676, opened 2026-05-25
fix(l1): bound startup RPC calls with per-call timeouts

rodrodros:
Adding context deadlines to the interactions with the L1 to avoid node stalling for ever

### DIFF
--- a/l1/l1.go
+++ b/l1/l1.go
@@ -2,6 +2,7 @@ package l1
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"math/big"
 	"time"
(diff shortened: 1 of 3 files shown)


node/migration.go
```diff
@@ -109,8 +109,9 @@ func fetchL1HeadIfMissing(
 	if err != nil {
 		return fmt.Errorf("creating a new L1 client: %w", err)
 	}
+
 	if err := client.CatchUpL1Head(ctx); err != nil {
-		return err
+		return fmt.Errorf("catching up to the latest l1 head: %w", err)
```
RafaelGranza:
```suggestion
		return fmt.Errorf("catching up to the latest L1 head: %w", err)
```

l1/l1.go
```diff
@@ -131,20 +132,35 @@ func (c *Client) subscribeToUpdates(ctx context.Context, updateChan chan *contra
 	}
 }
 
+// checkChainID checks that the client is connected to the right L1 client
 func (c *Client) checkChainID(ctx context.Context) error {
-	gotChainID, err := c.l1.ChainID(ctx)
+	const chainIDCheckTimeout = 30 * time.Second
```
RafaelGranza, line 137: Would it make sense to let these Timeout variables defined together at the top of the file?
```golang
const (
    chainIDCheckTimeout    = 30 * time.Second
    heightCallTimeout      = 30 * time.Second
    filterCallTimeout      = 60 * time.Second
...
)
```

l1/l1.go
RafaelGranza, line 379: I know this isn't part of this PR, but looking at it now: `time.Sleep` here ignores `ctx`, so on Ctrl+C (or any cancellation from above) we'd wait up to `resubscribeDelay` before the function returns and the service shuts down. The fix is small:

```suggestion
select {
  case <-ctx.Done():
      return 0
  case <-time.After(c.resubscribeDelay):
  }
}
```

RafaelGranza (approved):
LGTM! 

I adde a small PR description for future clarity.
The title covers the what but not the why.

Also, I left a few nits.

l1/l1.go
rodrodros: I think they make more sense scoped to their respective function

l1/l1.go
rodrodros: #3684 will track this

PR #3683, opened 2026-05-25
feat(feeder): introduce Reader interface and update dependencies

thiagodeev:
- Added a new Reader interface in the feeder package, reflecting all Client public endpoints
- Updated the feeder client references across the RPC handlers to use the new Reader interface instead of the concrete Client type.
- Generated mock implementations for the Reader interface to facilitate testing.


clients/feeder/feeder.go
```diff
@@ -38,6 +38,32 @@ type Client struct {
...
+	FeeTokenAddresses(ctx context.Context) (starknet.FeeTokenAddresses, error)
+	PreConfirmedBlock(ctx context.Context, blockNumber string) (*starknet.PreConfirmedBlock, error)
+	PublicKey(ctx context.Context) (*felt.Felt, error)
+	Signature(ctx context.Context, blockID string) (*starknet.Signature, error)
+	StateUpdate(ctx context.Context, blockID string) (*starknet.StateUpdate, error)
+	StateUpdateWithBlock(ctx context.Context, blockID string) (*starknet.StateUpdateWithBlock, error)
+	Transaction(
+		ctx context.Context,
+		transactionHash *felt.Felt,
+	) (*starknet.DeprecatedTransactionStatus, error)
```
RafaelGranza, line 59: `Transaction` is already deprecated. 
Do you think it is worth a:
```golang
// TODO: remove once v8 migrates to TransactionStatus
```
So it's clear this is transitional?

mocks/mock_feeder.go
```diff
@@ -0,0 +1,238 @@
+// Code generated by MockGen. DO NOT EDIT.
+// Source: github.com/NethermindEth/juno/clients/feeder (interfaces: Reader)
+//
+// Generated by this command:
+//
+//	mockgen -destination=../mocks/mock_feeder.go -package=mocks github.com/NethermindEth/juno/clients/feeder Reader
+//
+
+// Package mocks is a generated GoMock package.
+package mocks
```
RafaelGranza, line 10: This isn't in `.github/codecov.yml`'s ignore list. 
Change it to `"**/mocks"`to cover this and any future nested mocks dir.

RafaelGranza (approved):
LGTM. 

Just mind the merge order with https://github.com/NethermindEth/juno/pull/3601.

The new `StateUpdateWithBlockAndSignature` method will need to land in `Reader` too.

clients/feeder/feeder.go
thiagodeev: Good point. Indeed, even though the original method contains the "deprecated" warning, when calling the interface, it doesn't show as deprecated.
I've marked it as such, thanks

mocks/mock_feeder.go
thiagodeev: Thanks! My bad, I just copied the 'generate' command from another file, but I haven't considered the path depth; this caused the command to generate the file inside the 'clients' folder.
Fixed

PR #3689, opened 2026-05-28
l1: stop retry waits on context cancellation

### LINKED ISSUE #3684: Avoid waiting time on context cancellation
I know this isn't part of this PR, but looking at it now: `time.Sleep` here ignores `ctx`, so on Ctrl+C (or any cancellation from above) we'd wait up to `resubscribeDelay` before the function returns and the service shuts down. The fix is small:

```suggestion
select {
  case <-ctx.Done():
      return 0
  case <-time.After(c.resubscribeDelay):
  }
}
```

_Originally posted by @RafaelGranza in https://github.com/NethermindEth/juno/pull/3676#discussion_r3300062040_

peter941221:
Fix l1 retry backoff to stop waiting once the parent context is cancelled.

The issue is that two retry loops in l1 still sleep for the full resubscribe delay even after upper-level shutdown has already started:
- subscribeToUpdates
- finalisedHeight

This patch replaces the blind sleep with a small helper that waits for either:
- the retry delay to elapse, or
- ctx.Done() to fire

That preserves the normal retry cadence, but removes the extra shutdown latency on Ctrl+C or any upstream cancellation.

Testing
- built the Rust VM dependency required by local Go test linking
- go test ./l1 -count=1 -timeout 120s
- focused cancellation regressions plus the existing finalised-height retry regression all passed

Issue
- closes #3684


l1/retry_wait_test.go
rodrodros, line 1: In Juno we are going for a test behaviour, don't test your implementation approach. 

This tests are using internal api, please rewrite this code using the `l1_test` name to differentiate the package. Then check that behaviour is maintained by executing the client.

l1/l1.go
```diff
@@ -127,11 +127,25 @@ func (c *Client) subscribeToUpdates(ctx context.Context, updateChan chan *contra
 				zap.Duration("tryAgainIn", c.resubscribeDelay),
 				zap.Error(err),
 			)
-			time.Sleep(c.resubscribeDelay)
+			if !waitForResubscribeOrCancel(ctx, c.resubscribeDelay) {
+				return nil, fmt.Errorf("context canceled before resubscribe was successful: %w", ctx.Err())
+			}
 		}
 	}
```
rodrodros, line 140: This approach is not very good, you don't need an extra function. 

What you have to do is:

```golang
timer :=time.NewTimer(d)
defer timer.Stop()

for {
   select { 
   case <- ctx.Done():
        // same thing
   case <- timer.C
       // same thing that happens at default
```

l1/l1.go
```diff
@@ -375,7 +389,9 @@ func (c *Client) finalisedHeight(ctx context.Context) uint64 {
 				return finalisedHeight
 			}
 			c.logger.Debug("Failed to retrieve L1 finalised height, retrying...", zap.Error(err))
-			time.Sleep(c.resubscribeDelay)
+			if !waitForResubscribeOrCancel(ctx, c.resubscribeDelay) {
+				return 0
+			}
```
rodrodros, line 394: Same here

l1/retry_wait_test.go
rodrodros: Of course, don't make the private API public to bypass this comment

peter941221: Moved the coverage to the public `l1_test` package and exercised the behavior through `Run` / `CatchUpL1Head` instead of internal helpers. Also kept the private API private. Targeted `l1` tests passed locally.

l1/l1.go
peter941221: Dropped the extra helper and inlined the timer/select wait in the retry path. Targeted `l1` tests passed locally.

l1/l1.go
```diff
@@ -127,7 +127,15 @@ func (c *Client) subscribeToUpdates(ctx context.Context, updateChan chan *contra
...
+			select {
+			case <-ctx.Done():
+				if !timer.Stop() {
+					<-timer.C
+				}
+				return nil, fmt.Errorf("context canceled before resubscribe was successful: %w", ctx.Err())
+			case <-timer.C:
+			}
 		}
 	}
```
rodrodros, line 140: This is what I meant. Why not implement it like this?
```golang
func (c *Client) subscribeToUpdates(ctx context.Context, updateChan chan *contract.StarknetLogStateUpdate) (event.Subscription, error) {
    timer := time.NewTimer(c.resubscribeDelay)
	for {
		select {
		case <-ctx.Done():
			return nil, fmt.Errorf("context canceled before resubscribe was successful: %w", ctx.Err())
		case <-timer.C:
			updateSub, err := c.l1.WatchLogStateUpdate(ctx, updateChan)
			if err == nil {
				return updateSub, nil
			}
			c.logger.Debug("Failed to subscribe to L1 state updates",
				zap.Duration("tryAgainIn", c.resubscribeDelay),
				zap.Error(err),
			)
			}
		}
	}
```

rodrodros:
Solution lacks quality and contributor doesn't follow suggestion. Feel free to re-open if willing to implement this the correct way

PR #3691, opened 2026-05-28
Add state accessor unit tests

### LINKED ISSUE #3688: The new state accessors have no dedicated unit tests
As the title describes, we need to write dedicated tests for "core/state/accessors.go".

1sgtpepper:
## Summary

Adds dedicated unit coverage for `core/state/accessors.go`.

The tests cover:
- contract accessor read/write/delete behavior
- nonce, class hash, and storage history accessor write/delete behavior
- declared class accessor read/write/delete behavior

Fixes #3688.

## Testing

```sh
go test ./core/state
```

### DIFF
--- a/core/state/accessors_test.go
+++ b/core/state/accessors_test.go
@@ -1,10 +1,11 @@
-package state
+package state_test
 
 import (
 	"testing"
 
 	"github.com/NethermindEth/juno/core"
 	"github.com/NethermindEth/juno/core/felt"
+	"github.com/NethermindEth/juno/core/state"
 	"github.com/NethermindEth/juno/db"
 	"github.com/NethermindEth/juno/db/memory"
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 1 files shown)


core/state/accessors_test.go
```diff
@@ -0,0 +1,133 @@
...
+	"github.com/NethermindEth/juno/db/memory"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestContractAccessors(t *testing.T) {
+	database := memory.New()
+	addr := felt.NewUnsafeFromString[felt.Felt]("0x123")
+	nonce := felt.UnsafeFromString[felt.Felt]("0x1")
+	classHash := felt.UnsafeFromString[felt.Felt]("0xabc")
```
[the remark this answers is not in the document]
1sgtpepper: Not applying this one: `WriteContract` and `stateContract` use `felt.Felt` values here, not `*felt.Felt`.

core/state/accessors_test.go
```diff
@@ -0,0 +1,133 @@
+package state
```
rodrodros: Test packages should be called `package state_test`

1sgtpepper: Updated in b93ca852 to use `package state_test` and qualify the accessor calls through `core/state`. Verified with `go test ./core/state`.

core/state/accessors_test.go
```diff
@@ -0,0 +1,143 @@
...
+		func(value []byte) error {
+			assert.Equal(t, *classHash, felt.FromBytes[felt.Felt](value))
+			return nil
+		},
+	))
+	require.NoError(t, database.Get(
+		db.ContractStorageHistoryAtBlockKey(addr, storageKey, blockNum),
+		func(value []byte) error {
+			assert.Equal(t, *storageValue, felt.FromBytes[felt.Felt](value))
+			return nil
```
RafaelGranza, line 81: Here we are testing the same function the accessor uses internally.
1. It is tautological
2. It enforces behavior of the accessor that is not enforced by the actual code.

core/state/accessors_test.go
```diff
@@ -0,0 +1,143 @@
...
+	"github.com/NethermindEth/juno/db/memory"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestContractAccessors(t *testing.T) {
+	database := memory.New()
+	addr := felt.NewUnsafeFromString[felt.Felt]("0x123")
+	nonce := felt.UnsafeFromString[felt.Felt]("0x1")
+	classHash := felt.UnsafeFromString[felt.Felt]("0xabc")
```
RafaelGranza, line 19: You should use `NewUnsafeFromString` as default.

RafaelGranza (commented):
Closing. 

This reads as unreviewed AI-generated output: three near-duplicate test bodies and a `TestHistoryAccessors` that asserts against the accessor's own key-derivation, pinning implementation instead of behavior. 

Juno expects contributors to curate what they submit, regardless of how it was produced.

PR #3694, opened 2026-06-02
feat(sync): request-driven pre_confirmed polling

RafaelGranza:
## Summary
- Each call to `PreConfirmed()` now triggers a fresh `pre_confirmed` fetch instead of waiting for the next ticker.
- If a fetch is already running, extra refresh requests are dropped. A burst of RPC calls produces a single fetch, not a storm.
- Each fetch is capped at 2s so a slow or hung feeder cannot block future fetches.
- The `--preconfirmed-poll-interval` ticker is now a fallback when there is no RPC traffic to drive fetches.

## Benchmark
Numbers come from `BenchmarkPreConfirmedUpdateFrequency`: a 5-second simulation where one RPC call hits `PreConfirmed()` every 50ms (mimicking a client, or more than one, dictating a higher request frequency).

Run: `go test -bench=PreConfirmedUpdateFrequency -run=^$ ./sync/`

| feeder latency | baseline (ticker only) | request-driven        | improvement |
|----------------|------------------------|-----------------------|-------------|
| 50ms           | 1.8 upd/s, mean 500ms  | 10.8 upd/s, mean 91ms | 6x          |
| 100ms          | 1.8 upd/s, mean 500ms  | 7.2 upd/s, mean 134ms | 4x          |
| 200ms          | 1.8 upd/s, mean 500ms  | 4.2 upd/s, mean 228ms | 2.3x        |

The running guard caught 44-78% of triggers as duplicates, so it was not a feeder hammer.

### DIFF
--- a/cmd/juno/juno.go
+++ b/cmd/juno/juno.go
@@ -199,8 +199,8 @@ const (
 	disableL1VerificationUsage = "Disables L1 verification since an Ethereum node is not provided."
 	preLatestPollIntervalUsage = "Sets polling interval for pre-latest block updates. " +
 		"(0s will disable polling)."
-	preConfirmedPollIntervalUsage = "Sets how frequently pre_confirmed block will be updated" +
-		"(0s will disable fetching of pre_confirmed block)."
+	preConfirmedPollIntervalUsage = "Fallback ticker interval for pre_confirmed polling when no " +
+		"RPC traffic is driving refreshes. (0s disables polling.)"
 	p2pUsage           = "EXPERIMENTAL: Enables p2p server."
 	p2pAddrUsage       = "EXPERIMENTAL: Specify p2p listening source address as multiaddr.  Example: /ip4/0.0.0.0/tcp/7777"
 	p2pPublicAddrUsage = "EXPERIMENTAL: Specify p2p public address as multiaddr.  Example: /ip4/35.243.XXX.XXX/tcp/7777"
--- a/sync/pending_polling.go
+++ b/sync/pending_polling.go
@@ -17,10 +17,15 @@ import (
 
 const (
 	preLatestCacheSize = 10
-	// Per-fetch cap so a hung feeder can't hold the running guard for the
-	// feeder client's full retry budget (~20s). TODO: consider exposing as flag.
-	preConfirmedFetchTimeout = 2 * time.Second
+	// Per-fetch cap scales with poll interval so slower configs get
+	// proportionally longer budgets. Floor avoids pathologically short caps.
+	preConfirmedFetchTimeoutMultiplier = 4
+	preConfirmedFetchTimeoutFloor      = 500 * time.Millisecond
 )
+
+func (s *Synchronizer) preConfirmedFetchTimeout() time.Duration {
+	return max(s.preConfirmedPollInterval*preConfirmedFetchTimeoutMultiplier, preConfirmedFetchTimeoutFloor)
+}
 
 // isGreaterThanTip reports whether the given number is at or beyond the highest known header.
 func (s *Synchronizer) isGreaterThanTip(blockNumber uint64) bool {
@@ -242,9 +247,12 @@ func (s *Synchronizer) pollPreLatest(ctx context.Context, out chan<- *pending.Pr
 	}
 }
 
-// requestPreConfirmedRefresh wakes the polling goroutine. Drops the request
-// if a fetch is already running: that fetch already serves it.
+// requestPreConfirmedRefresh wakes the polling goroutine; no-op if polling
+// is disabled or a fetch is already running (that fetch will serve it).
 func (s *Synchronizer) requestPreConfirmedRefresh() {
+	if s.preConfirmedPollInterval == 0 {
+		return
+	}
 	if s.preConfirmedFetching.Load() {
 		return
 	}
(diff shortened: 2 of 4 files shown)


sync/pending_polling.go
```diff
@@ -17,6 +17,9 @@ import (
 
 const (
 	preLatestCacheSize = 10
+	// Per-fetch cap so a hung feeder can't hold the running guard for the
+	// feeder client's full retry budget (~20s). TODO: consider exposing as flag.
+	preConfirmedFetchTimeout = 2 * time.Second
```
RafaelGranza: Hardcoded for now.  Worth exposing as a flag, or fine to keep at 2s?

sync/pending_polling_test.go
```diff
@@ -31,8 +31,8 @@ type MockDataSource struct {
 	DataSource
 	pendingErrorThreshold      uint
 	preConfirmedErrorThreshold uint
-	numCallsPreConfirmed       uint
-	numCallsPending            uint
+	numCallsPreConfirmed       atomic.Uint32
+	numCallsPending            atomic.Uint32
```
RafaelGranza, line 36: This is to avoid race conditions during tests

sync/pending_polling.go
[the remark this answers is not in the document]
RafaelGranza: IMO, this is NOT a problem, since the ticker will keep trying if the request failed.

sync/sync.go
```diff
@@ -600,6 +608,11 @@ func (s *Synchronizer) pollLatest(ctx context.Context) {
 }
 
 func (s *Synchronizer) PreConfirmed() (*pending.PreConfirmed, error) {
+	// Every read is a vote that the cached pre_confirmed should be refreshed.
+	// The polling goroutine picks this up on its next iteration and any subscribers
+	// of preConfirmedDataFeed (e.g. WebSocket clients) benefit from the same fetch.
+	s.requestPreConfirmedRefresh()
```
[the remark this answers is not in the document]
RafaelGranza: IMO, if the client is dictating the request frequency, this implies it requests frequently. 
So not returning the fresher info is not necessary, since the call will re-trigger again, from ticker or request.

sync/pending_polling.go
```diff
@@ -262,43 +275,63 @@ func (s *Synchronizer) pollPreConfirmed(
 		select {
 		case <-ctx.Done():
 			return
-
+		case <-s.preConfirmedTrigger:
 		case <-ticker.C:
```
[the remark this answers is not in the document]
RafaelGranza: I'm inclined of not caring about it.

sync/pending_polling.go
```diff
@@ -262,43 +275,63 @@ func (s *Synchronizer) pollPreConfirmed(
...
+	knownTransactionCount uint64,
+) (pending.PreConfirmedUpdate, error) {
+	s.preConfirmedFetching.Store(true)
+	defer func() {
+		select {
+		case <-s.preConfirmedTrigger:
+		default:
+		}
+		s.preConfirmedFetching.Store(false)
+	}()
```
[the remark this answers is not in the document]
RafaelGranza: I'll apply changes

sync/pending_polling.go
```diff
@@ -239,12 +242,22 @@ func (s *Synchronizer) pollPreLatest(ctx context.Context, out chan<- *pending.Pr
...
-// Each poll echoes the currently stored pre_confirmed's identifier and
-// transaction count so the server can return a no-change marker, a delta of
-// appended transactions, or a fresh full block when the round identifier no
-// longer matches. On success, the resulting update is forwarded to out.
+// requestPreConfirmedRefresh wakes the polling goroutine. Drops the request
+// if a fetch is already running: that fetch already serves it.
+func (s *Synchronizer) requestPreConfirmedRefresh() {
+	if s.preConfirmedFetching.Load() {
+		return
+	}
```
rodrodros, line 258: Mmm, I don't like this very much. You're trying to avoid multiple requests but I suspect there might be a better way.

sync/pending_polling_test.go
```diff
@@ -997,3 +1111,171 @@ func makeEmptyPreLatestForParent(parent *core.Header) pending.PreLatest {
...
+						sc.name, fetchLatency,
+						updates, float64(updates)/runDuration.Seconds(),
+						fetches, trigs, dropped,
+						meanDuration(intervals), p99Duration(intervals),
+					)
+				})
+			})
+		}
+	}
+}
```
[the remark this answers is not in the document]
RafaelGranza: this is a benchmark. Renaming to make the intent clear.

cmd/juno/juno.go
```diff
@@ -199,8 +199,8 @@ const (
 	disableL1VerificationUsage = "Disables L1 verification since an Ethereum node is not provided."
 	preLatestPollIntervalUsage = "Sets polling interval for pre-latest block updates. " +
 		"(0s will disable polling)."
-	preConfirmedPollIntervalUsage = "Sets how frequently pre_confirmed block will be updated" +
-		"(0s will disable fetching of pre_confirmed block)."
+	preConfirmedPollIntervalUsage = "Fallback ticker interval for pre_confirmed polling when no " +
+		"RPC traffic is driving refreshes. (0s disables polling.)"
```
rodrodros, line 203: The original description is better, please switch it back

sync/pending_polling.go
```diff
@@ -17,8 +17,16 @@ import (
 
 const (
 	preLatestCacheSize = 10
+	// Per-fetch cap scales with poll interval so slower configs get
+	// proportionally longer budgets. Floor avoids pathologically short caps.
+	preConfirmedFetchTimeoutMultiplier = 4
```
rodrodros, line 22: We have to work in your text, I have full context of this PR and this text seems unnecessarily complex to understand. 

Why not something as:
```
// Fetch timeout = max(pollInterval * multiplier, floor).
```

sync/pending_polling.go
```diff
@@ -17,8 +17,16 @@ import (
...
+	preConfirmedFetchTimeoutFloor      = 500 * time.Millisecond
 )
 
+func (s *Synchronizer) preConfirmedFetchTimeout() time.Duration {
+	return max(s.preConfirmedPollInterval*preConfirmedFetchTimeoutMultiplier, preConfirmedFetchTimeoutFloor)
+}
+
```
rodrodros, line 29: Additionally your constants are defined here, why not defined them just here and putting the comment here?

rodrodros:
Closing since it is still an issue but we won't move forward with this solution for now

PR #3700, opened 2026-06-03
feat(l1): Add L1 Eth types

brbrr:


l1/eth/address.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+	s = s[2:]
+	if len(s) != 2*AddressLength {
+		return fmt.Errorf("eth: invalid address length %d, want %d", len(s), 2*AddressLength)
+	}
+	if _, err := hex.Decode(a[:], s); err != nil {
+		return fmt.Errorf("eth: invalid address hex: %w", err)
+	}
+	return nil
+}
```
rodrodros: I don't think every error should be suffixed with the phrase "eth: ...", only the individual content. Is the role of the caller give the missing context.

l1/eth/address.go
```diff
@@ -0,0 +1,81 @@
...
+func (a *Address) UnmarshalText(input []byte) error {
+	return a.decodeHex(input)
+}
+
+// UnmarshalJSON parses an address from a JSON string. Matches go-ethereum's
+// common.Address.UnmarshalJSON: requires a quoted "0x"-prefixed hex string of
+// exactly AddressLength bytes; case-insensitive.
+func (a *Address) UnmarshalJSON(input []byte) error {
+	if len(input) < 2 || input[0] != '"' || input[len(input)-1] != '"' {
+		return fmt.Errorf("eth: address must be a JSON string")
```
rodrodros: `fmt.Errorf` but no argument is given? Maybe use errors.New(...) directly

l1/eth/address.go
```diff
@@ -0,0 +1,81 @@
...
+// SetBytes sets the address to the value of b.
+// If b is larger than len(a), b will be cropped from the left.
+func (a *Address) SetBytes(b []byte) {
+	if len(b) > len(a) {
+		b = b[len(b)-AddressLength:]
+	}
+	copy(a[AddressLength-len(b):], b)
+}
+
+// Bytes gets the byte representation of the underlying address.
```
rodrodros: // Bytes returns a copy of the underlying byte representation of the address

l1/eth/address.go
```diff
@@ -0,0 +1,81 @@
...
+func HexToAddress(s string) Address { return BytesToAddress(fromHex(s)) }
+
+// SetBytes sets the address to the value of b.
+// If b is larger than len(a), b will be cropped from the left.
+func (a *Address) SetBytes(b []byte) {
+	if len(b) > len(a) {
+		b = b[len(b)-AddressLength:]
+	}
+	copy(a[AddressLength-len(b):], b)
+}
```
rodrodros: I don't agree with this behaviour. I understand it is to match Go ethereum but unless there is a functionally depending on this, panic if it is an internal error, return an error if it is externally received

l1/eth/hash.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+	s = s[2:]
+	if len(s) != 2*HashLength {
+		return fmt.Errorf("eth: invalid hash length %d, want %d", len(s), 2*HashLength)
+	}
+	if _, err := hex.Decode(h[:], s); err != nil {
+		return fmt.Errorf("eth: invalid hash hex: %w", err)
+	}
+	return nil
+}
```
rodrodros: Similar issue with pre-pending error messages with "eth: "

l1/eth/hash.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+	return h.decodeHex(input[1 : len(input)-1])
+}
+
+func (h *Hash) decodeHex(s []byte) error {
+	if len(s) < 2 || s[0] != '0' || (s[1] != 'x' && s[1] != 'X') {
+		return fmt.Errorf("eth: hash missing 0x prefix")
+	}
+	s = s[2:]
+	if len(s) != 2*HashLength {
```
rodrodros: This function is so similar to the address one. Please refactor so that the code is not repeated.

l1/eth/address.go
```diff
@@ -0,0 +1,81 @@
...
+	hex.Encode(out[2:], a[:])
+	return out, nil
+}
+
+// UnmarshalText parses an address from its hex representation. Matches
+// go-ethereum's common.Address.UnmarshalText: requires "0x"-prefixed hex of
+// exactly AddressLength bytes; case-insensitive.
+func (a *Address) UnmarshalText(input []byte) error {
+	return a.decodeHex(input)
+}
```
rodrodros: I wonder about the inconsistency of pointer receives and value receivers. Can you explain them for me?

l1/eth/address_test.go
rodrodros, line 1: If this PR is here to avoid having geth dependencies, but we are keeping them for test, that in itself is making this effort null. Right? What is the follow up plan?

l1/eth/address_test.go
```diff
@@ -0,0 +1,218 @@
...
+	"github.com/NethermindEth/juno/l1/eth"
+	gethcommon "github.com/ethereum/go-ethereum/common"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// addressCorpus returns a deterministic, broad sample of 20-byte values for
+// parity tests against go-ethereum's common.Address. The geth package is still
+// in go.mod through PR1-PR4; these imports are removed in PR5.
+func addressCorpus(t *testing.T) [][20]byte {
```
rodrodros: Why return [20]byte and not eth.Address? 

Why making an identical function for the hash? Why not considering the use of generics?

An example using generics:

```golang
func corpus[T interface {eth.Address | eth.Hash}]() []T {
   var zero, allFF, allAA, ascending T
	size := len(zero)
	for i := range size {
		allFF[i] = 0xff
		allAA[i] = 0xaa
		ascending[i] = byte(i)
	}
    // and so on...
}

File: l1/eth/hash.go
Comment (rodrodros) [inline l1/eth/hash.go:1]: This file has so many common code with the Address type, is there no possibility of merging  them more overall?

File: l1/eth/hash.go [hunk 9]
```diff
@@ -0,0 +1,115 @@
...
+// Hash represents the 32 byte Keccak256 hash of arbitrary data.
+type Hash [HashLength]byte
+
+// BytesToHash returns Hash with value b.
+// If b is larger than HashLength, b will be cropped from the left.
+func BytesToHash(b []byte) Hash {
+	var h Hash
+	h.SetBytes(b)
+	return h
+}
```
rodrodros: We already have a standard way of using constructors in `felt`, is not better to call it `HashFromFelt` like we do there?

l1/eth/hash.go
```diff
@@ -0,0 +1,115 @@
...
+
+// HexToHash returns Hash with byte values of s.
+// If s is larger than 2*HashLength hex chars, it will be cropped from the left.
+func HexToHash(s string) Hash { return BytesToHash(fromHex(s)) }
```
rodrodros: I also find this function name misleading, it is not a `Hex` it is a string, so maybe it should be `HashFromString`

l1/eth/address.go
```diff
@@ -0,0 +1,81 @@
...
+import (
+	"encoding/hex"
+	"fmt"
+)
+
+// Lengths of hashes and addresses in bytes.
+const (
+	// AddressLength is the expected length of the address.
+	AddressLength = 20
+	addressHexLen = 2 + 2*AddressLength // "0x" + 40 hex chars
```
rodrodros: If this constant is only used inside **1** function, why not define it locally there?

l1/eth/hash.go
```diff
@@ -0,0 +1,115 @@
...
+func (h Hash) Bytes() []byte { return h[:] }
+
+// Hex converts a hash to a 0x-prefixed lowercase hex string.
+func (h Hash) Hex() string {
+	var buf [hashHexLen]byte
+	buf[0] = '0'
+	buf[1] = 'x'
+	hex.Encode(buf[2:], h[:])
+	return string(buf[:])
+}
```
rodrodros: `Hex` and `MarshallText` seem very similar, is there no opportunity for refactor there?

l1/eth/types.go
```diff
@@ -0,0 +1,103 @@
...
+	BlockNumber HexU64   `json:"blockNumber"`
+	Removed     bool     `json:"removed"`
+}
+
+// Receipt is the trimmed-down view of an Ethereum transaction receipt.
+// Only Logs is read in juno. All other fields in the
+// eth_getTransactionReceipt JSON response are ignored.
+type Receipt struct {
+	Logs []*Log `json:"logs"`
+}
```
rodrodros: Why a pointer to Log and not just the `[]Log`?

l1/eth/types.go
```diff
@@ -0,0 +1,103 @@
...
+		return fmt.Errorf("eth: invalid hex uint64: %w", err)
+	}
+	*h = HexU64(v)
+	return nil
+}
+
+// HexBytes is a []byte that JSON-decodes from a 0x-prefixed hex string,
+// matching the Ethereum JSON-RPC "data" wire format. An empty payload
+// ("0x") is allowed.
+type HexBytes []byte
```
rodrodros: Is in `HexBytes` possibly infinite in size, why is it not bounded?

l1/eth/types.go
rodrodros, line 1: I don't find it natural that `address` and `hash` get a type but the rest of them doesn't. Please group the types into their scripts. **Please** note that this doesn't mean that every type should get a script but a script can have types related with each other which is not the reality here.

Also, we should avoid having files being called "type" since they are usually not very descriptive.

l1/eth/types.go
```diff
@@ -0,0 +1,103 @@
...
+// type on structs that consume RPC responses (Header, Log).
+type HexU64 uint64
+
+func (h *HexU64) UnmarshalJSON(input []byte) error {
+	raw, err := unquoteHex(input)
+	if err != nil {
+		return fmt.Errorf("eth: hex uint64: %w", err)
+	}
+	if len(raw) == 0 {
+		return fmt.Errorf("eth: hex uint64 has no digits")
```
rodrodros: The `eth` prefix doesn't seem necessary to me.

rodrodros (commented):
This PR has many small problems which make them kind of a big problem. 

Plenty of opportunites for refactor and avoiding duplicated code as well as better placement of code.

This PR seems it was not self-reviewed. Please, self review your PRs (and I don't mean, self review them with an AI). Be auto-critical of your work.

l1/eth/address.go
brbrr: I was approaching it with the idea to have a matching behavior, so it would be easier to migrate without worrying of missing something. Also, I've added parity tests, if I'll panic here, they will fail/render themselves useless. 

I'll add a `// fixme` comment, and once the switch is done, I'll address it as a follow-up. If not, I'll add panic in this PR. Let me know what you think.

l1/eth/address_test.go
brbrr: This is a first step in that direction :) The plan is to add types, switch to them in non-l1 context (2 separate PRs), replace l1 client implementation, and only then remove geth dependency from the project entirely. 

It's present here for a parity tests, which I think is important.

l1/eth/address.go
brbrr: Hm, I thought that's desired pattern to use value receivers for methods that don't modify self. Also, it seems that it aligns with patterns used across the repo. Please correct me if I'm wrong/missing something.

l1/eth/types.go
brbrr: Yeah, I was on the fence of adding types.go, but having mostly empty files felt worse. I'll remove it, and group them meaningfully.

l1/eth/types.go
brbrr: `HexBytes` are only used for data field in Receipt logs. it's unbounded by protocol design. And since we are getting this data from a trusted RPC provider on Ethereum, I think that there is enough guarantees that this "issue" won't be exploited. 

We probably could add some arbitrary limit, but it feels like we still can end up hitting it 🤷

l1/eth/hash.go
brbrr: Good call. renamed some of the methods for clarity and alignment with Felt style.

l1/eth/address_test.go
brbrr: I'm not really good with generics in go, but your suggestion won't compile because the underlying types are different. claude suggests some workarounds, which isn't as elegant as your suggestion. I've applied the one that I find most suitable.

l1/eth/types.go
brbrr: Good point. I dragged it over from geth, but there is no reason for pointers in our case. updated.

l1/eth/address.go
rodrodros: I am ok, having it as a follow up 👍

l1/eth/address.go
rodrodros: The difference between a pointer receiver and value receiver is that for:
* pointer receiver, a method call, copies the pointer
* value receiver, a method call, copies the whole value

To make the methods cheaper and avoid the extra copy (specifically on types bigger than one word) we should use pointer receiver as a rule and value receiver exceptionally when they are required.

l1/eth/address_test.go
rodrodros: > I'm not really good with generics in go

Go generics are no different from any language, your comment translate as I am not good with generics, which is not a good explanation for not using them. If you don't use them how are you ever going to get good with them. 

For this to not repeat itself, the rule is that _if you're writing the same function for different types_ then you should use generics. Please apply this all of your future code. 

> your suggestion won't compile 

I tried locally before suggesting and LSP didn't give any warnings. Eitherway, there are other ways such as sealed interfaces.

l1/eth/types.go
rodrodros: I think the name might not be too good then, and yes we must bound them. We should write code for the worst case scenario, not for the best.

l1/eth/hex_codec.go
```diff
@@ -0,0 +1,114 @@
...
+		return errors.New("must be a JSON string")
+	}
+	return decodeHexText(dst, input[1:len(input)-1])
+}
+
+// unquoteHex strips the surrounding JSON quotes and the required 0x prefix,
+// returning the hex digits. Accepts "0x" or "0X" prefix.
+func unquoteHex(input []byte) ([]byte, error) {
+	if len(input) < 2 || input[0] != '"' || input[len(input)-1] != '"' {
+		return nil, errors.New("not a JSON string")
```
rodrodros, line 59: nit: let's add one test for this as well please, to assert the behaviour

rodrodros (approved):
Looks a lot better, thank you!

l1/eth/types.go
rodrodros: I saw the updated name 👍

l1/eth/address_test.go
rodrodros: After looking at the code, not that it is more organized, I don't think this is a big deal anymore. The current refactor is good 👍

l1/eth/address.go
brbrr: I've tested it with pointer receiver:

```
json.Marshal(addr) -> [17,17,...] (raw array)  
json.Marshal(&addr)   -> "0x1111..."  (correct)
```

So it silently ignores `MarshalText` breaks serialization. I'll keep it as-is.

Issue #3581, opened 2026-04-24
JSON Pretty Print


rodrodros:
Right now when a JSON is ill formatted we just return "JSON is not right" but give no explanation of which key was missing, if it was a missing `}` or anything like that. 

Since there is now an active stream of work that will add SONIC to our RPC package, we need to add this pretty printing which will be a positive UX change

rodrodros:
Solved!

PR #3702, opened 2026-06-04
feat(jsonrpc): clearer errors for malformed JSON and invalid params

RafaelGranza:
# Context: 
Related to Issue #3581

Before, a malformed JSON request returned only the raw decoder message.
Errors now point at what's wrong.

Also, long lines are truncated to 80 cols, so the marker stays aligned.

## Errors it Covers
- syntax errors
- unexpected EOF
- type mismatches
- trailing commas. 
- missing non-optional param field
- missing/unexpected params in list

## Examples:
Errors ate python styled.
```sh
{"id":@}
      ^
unexpected '@', expected a value [line 1, column 7]
```

```sh
...xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx" z}
                                                                           ^
unexpected 'z', expected ',' or '}' [line 1, column 230]
```

```sh
{"jsonrpc": 5, "method": "x", "id": 1}
            ^
field "jsonrpc" should be string, got number [line 1, column 13]
```

## Benchmark: 
We no longer copy the whole stream, we only keep the last 512 bytes read. So the overhead stays roughly constant instead of growing with the request size.

Large requests, `main` (no buffering) vs this PR:

| | main (no copy) | this PR | delta |
|--|--|--|--|
| time 1MB | ~14.9 ms/op | ~15.1 ms/op | noise |
| time 10MB | ~151 ms/op | ~153 ms/op | noise |
| mem 1MB | ~9.45 MB/op | ~9.42 MB/op | noise |
| mem 10MB | ~77.7 MB/op | ~79.3 MB/op | ~+1.6 MB |
| allocs 1MB | ~50 /op | ~53 /op | +3 /op |
| allocs 10MB | ~52 /op | ~57 /op | +5 /op |

# SONIC
When SONIC arrives it won't break. 
Its errors already carry position, so it will just work, only not printed as nicely until we add a case for it (see the `TODO` in `errorOffset`).

### DIFF
--- a/jsonrpc/pretty_error.go
+++ b/jsonrpc/pretty_error.go
@@ -6,24 +6,24 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"strings"
 	"unicode/utf8"
 )
 
-const unexpectedEOFMsg = "unexpected end of input (missing closing '}' or ']'?)"
-
-func clamp(v, lo, hi int) int {
-	return min(max(v, lo), hi)
-}
-
 func parseErrorOffset(input []byte, err error) (offset int, ok bool) {
-	var syntaxErr *json.SyntaxError
+	var (
+		syntaxErr *json.SyntaxError
+		typeErr   *json.UnmarshalTypeError
+	)
 	switch {
 	case errors.As(err, &syntaxErr):
-		return clamp(int(syntaxErr.Offset)-1, 0, len(input)), true
+		return min(int(syntaxErr.Offset)-1, len(input)), true
+	case errors.As(err, &typeErr):
+		return min(int(typeErr.Offset)-1, len(input)), true
 	case errors.Is(err, io.ErrUnexpectedEOF), errors.Is(err, io.EOF):
 		return len(input), true
 	default:
-		// TODO(granza): add a case for SONIC's *decoder.SyntaxError (.Pos) when it lands.
+		// TODO(granza): when we add SONIC, the errors will be already pretty.
 		return 0, false
 	}
 }
@@ -36,11 +36,86 @@ func lineAndColumn(input []byte, offset int) (line, col int) {
 	return line, col
 }
 
-func errorMessage(err error) string {
-	if errors.Is(err, io.ErrUnexpectedEOF) || errors.Is(err, io.EOF) {
-		return unexpectedEOFMsg
+func expectedToken(reason string) (string, bool) {
+	// The stdlib reason uses parser jargon
+	// This maps it to user-friendly messages
+	switch {
+	case strings.Contains(reason, "beginning of object key string"):
+		return "a string key or '}'", true
+	case strings.Contains(reason, "after object key:value pair"):
+		return "',' or '}'", true
+	case strings.Contains(reason, "after object key"):
+		return "':'", true
+	case strings.Contains(reason, "after array element"):
+		return "',' or ']'", true
+	case strings.Contains(reason, "beginning of value"):
+		return "a value", true
+	default:
+		return "", false
 	}
+}
+
+func precededByComma(input []byte, offset int) bool {
+	trimmed := bytes.TrimRight(input[:offset], " \t\r\n")
+	return len(trimmed) > 0 && trimmed[len(trimmed)-1] == ','
+}
+
+func describeError(input []byte, offset int, err error) string {
+	if offset < len(input) {
+		if expected, ok := expectedToken(err.Error()); ok {
+			symbol, _ := utf8.DecodeRune(input[offset:])
+			if (symbol == '}' || symbol == ']') && precededByComma(input, offset) {
+				return fmt.Sprintf("unexpected trailing comma before %q", symbol)
+			}
+			return fmt.Sprintf("unexpected %q, expected %s", symbol, expected)
+		}
+	}
+
+	if errors.Is(err, io.ErrUnexpectedEOF) || errors.Is(err, io.EOF) {
+		return "unexpected end of input"
+	}
+
 	return err.Error()
+}
+
+func offendingLine(input []byte, offset int) string {
+	start := bytes.LastIndexByte(input[:offset], '\n') + 1
+	if end := bytes.IndexByte(input[offset:], '\n'); end >= 0 {
+		return string(input[start : offset+end])
+	}
+	return string(input[start:])
+}
+
+// Returns the truncated string around a pivot and the new index of it
+func truncateAround(line string, pivot int, maxLineWidth int) (string, int) {
+	runes := []rune(line)
+	if len(runes) <= maxLineWidth {
+		return line, pivot
+	}
+
+	const ellipsis = "..."
+	maxSize := maxLineWidth - 2*len(ellipsis)
+	idx := min(pivot-1, len(runes))
+	start := max(0, idx-maxSize/2)
+	end := min(start+maxSize, len(runes))
+	start = max(0, end-maxSize)
+
+	left, right := "", ""
+	if start > 0 {
+		left = ellipsis
+	}
+	if end < len(runes) {
+		right = ellipsis
+	}
+	return left + string(runes[start:end]) + right, idx - start + len(left) + 1
+}
+
+func drawMarker(input []byte, offset, col int, msg string) string {
+	const mazSizeOfMessage = 80
+
+	line, markerCol := truncateAround(offendingLine(input, offset), col, mazSizeOfMessage)
+	gap := strings.Repeat(" ", markerCol-1)
+	return fmt.Sprintf("%s\n%s^\n%s|_ %s", line, gap, gap, msg)
 }
 
 func prettyParseError(input []byte, err error) string {
@@ -50,5 +125,6 @@ func prettyParseError(input []byte, err error) string {
 	}
 
 	line, col := lineAndColumn(input, offset)
-	return fmt.Sprintf("%s at line %d column %d", errorMessage(err), line, col)
+	msg := fmt.Sprintf("%s at line %d column %d", describeError(input, offset, err), line, col)
+	return drawMarker(input, offset, col, msg)
 }
--- a/jsonrpc/pretty_error.go
+++ b/jsonrpc/pretty_error.go
@@ -1,0 +1,60 @@
+package jsonrpc
+
+import (
+	"bytes"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"strings"
+)
+
+const unexpectedEOFMsg = "unexpected end of input (missing closing '}' or ']'?)"
+
+func clamp(v, lo, hi int) int {
+	return min(max(v, lo), hi)
+}
+
+func parseErrorOffset(input []byte, err error) (offset int, ok bool) {
+	var syntaxErr *json.SyntaxError
+	switch {
+	case errors.As(err, &syntaxErr):
+		return clamp(int(syntaxErr.Offset)-1, 0, len(input)), true
+	case errors.Is(err, io.ErrUnexpectedEOF), errors.Is(err, io.EOF):
+		return len(input), true
+	default:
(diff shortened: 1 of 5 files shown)


jsonrpc/server.go
```diff
@@ -331,13 +331,14 @@ func (s *Server) HandleReader(ctx context.Context, reader io.Reader) ([]byte, ht
 
 	header := http.Header{}
 
-	dec := json.NewDecoder(bufferedReader)
+	var captured bytes.Buffer
+	dec := json.NewDecoder(io.TeeReader(bufferedReader, &captured))
```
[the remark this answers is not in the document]
RafaelGranza: Kept the copy: it's bounded by the body-size limit (`MaxRequestBodySize` on HTTP, `ReadLimit` on WS) and only read on a parse error, so it can't grow unbounded. 

Added benchmarks for the success-path cost:

| bench         | ns/op | allocs/op            |
|---------------|-------|----------------------|
| single call   | ~3.2µs | 28 (25 without tee) |
| 20-call batch | ~67µs  | 426                  |

So the tee adds ~3 allocs / ~1.5% ns on success. Can move to a sync.Pool if it matters.

jsonrpc/server.go
```diff
@@ -323,7 +323,11 @@ func (s *Server) HandleReadWriter(ctx context.Context, rw io.ReadWriter) error {
 // It returns the response in a byte array, only returns an
 // error if it can not create the response byte array
 func (s *Server) HandleReader(ctx context.Context, reader io.Reader) ([]byte, http.Header, error) {
-	bufferedReader := bufio.NewReaderSize(reader, bufferSize)
+	// Tee the decoder's input into captured so a parse error can show its
+	// position. Every body is copied, not just bad ones, but the copy is small,
+	// capped by the upstream body-size limit, and only read on error.
+	var captured bytes.Buffer
+	bufferedReader := bufio.NewReaderSize(io.TeeReader(reader, &captured), bufferSize)
```
rodrodros: This is on the hot path, isn't it possible to perform the copy on error only? Why do we need to perform a copy? The output is never modified

jsonrpc/server_test.go
rodrodros, line 1: Good thing, the plumbing is there, bad thing, it is not very human readable. Showing the line and column number is really not ideal. People don't think in "columns", at any case it would be position x. Additionally, since we have the data, and we have the positions, the error formatting could be better in my opinion to close the issue the right way.

jsonrpc/server_test.go
```diff
@@ -209,11 +209,19 @@ func TestHandle(t *testing.T) {
 	}{
 		"invalid json": {
 			req: `{]`,
-			res: `{"jsonrpc":"2.0","error":{"code":-32700,"message":"Parse error","data":"invalid character ']' looking for beginning of object key string"},"id":null}`,
+			res: `{"jsonrpc":"2.0","error":{"code":-32700,"message":"Parse error","data":"invalid character ']' looking for beginning of object key string at line 1 column 2"},"id":null}`,
```
rodrodros: I believe the error should clearly identify to a human what's wrong
```
{]
 ^
 |_ Unexpected `]` character, expected `}`
```

rodrodros (commented):
Half done PR. Yes now more information is given to the user, but it is given in such way that is not human friendly. Let's improve upon that.

You can look into other RPCs error and get inspiration from those.

jsonrpc/server_test.go
RafaelGranza:
```bash
  {]
   ^
unexpected ']', expected a string key or '}' [line 3, column 4]
```

Changed to this format.

jsonrpc/pretty_error.go
```diff
@@ -0,0 +1,130 @@
...
+)
+
+func parseErrorOffset(input []byte, err error) (offset int, ok bool) {
+	var (
+		syntaxErr *json.SyntaxError
+		typeErr   *json.UnmarshalTypeError
+	)
+	switch {
+	case errors.As(err, &syntaxErr):
+		return min(int(syntaxErr.Offset)-1, len(input)), true
```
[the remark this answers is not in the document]
RafaelGranza: This is unreachable

jsonrpc/server.go
RafaelGranza: We cannot avoid a copy right now. We don't know where a json error could be
Once we hit one, we would need to rewind the reader, which streams do not allow

RafaelGranza:
Here are some of the new errors:
*There are no colors since the error can be used outside go and shell contexts.

```sh
{]
 ^
unexpected ']', expected a string key or '}' [line 1, column 2]
```

```sh
{"a" 1}
     ^
unexpected '1', expected ':' [line 1, column 6]
```

```sh
{"a":}
     ^
unexpected '}', expected a value [line 1, column 6]
```

```sh
{"a":1, }
        ^
unexpected trailing comma before '}' [line 1, column 9]
```

```sh
[1,2,]
     ^
unexpected trailing comma before ']' [line 1, column 6]
```

```sh
{"id":@}
      ^
unexpected '@', expected a value [line 1, column 7]
```

```
{"jsonrpc": "2.0", "method": "method", "id": 1
                                              ^
unexpected end of input [line 1, column 47]
```

```sh
{"a": "unterminated
                   ^
unexpected end of input [line 1, column 20]
```

```sh
{"jsonrpc": 5, "method": "x", "id": 1}
            ^
field "jsonrpc" should be string, got number [line 1, column 13]
```

```sh

^
unexpected end of input [line 1, column 1]
```

```sh
  {]
   ^
unexpected ']', expected a string key or '}' [line 3, column 4]
```

```sh
...xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx" z}
                                                                           ^
unexpected 'z', expected ',' or '}' [line 1, column 230]
```

```sh
{@yyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyy...
 ^
unexpected '@', expected a string key or '}' [line 1, column 2]
```

jsonrpc/server_test.go
RafaelGranza: Addressed

rodrodros (approved):
Looks good, thank you!

Before merging, could you please swap the `column` which makes sense for us programmers but might not be for non-programmers. My suggestion is `position` or similar. 

Finally, in the tests, can you please add at least one real Starknet request (multi line) with an error. The examples I saw in the unit tests are limited to 1 to 2 lines.

RafaelGranza:
@rodrodros 
I added both the vertical trim and the bounded buffer for the stream copy. 
They were closely related, so I did them together.

jsonrpc/pretty_error.go
```diff
@@ -0,0 +1,208 @@
...
+	newlinesSeen  int
+}
+
+func (c *windowBuffer) Write(p []byte) (int, error) {
+	c.consumedBytes += len(p)
+	c.newlinesSeen += bytes.Count(p, []byte{'\n'})
+
+	if len(p) >= maxWindowSize {
+		c.window = append(c.window[:0], p[len(p)-maxWindowSize:]...)
+		return len(p), nil
```
rodrodros, line 33: Test this execution path please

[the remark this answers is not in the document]
RafaelGranza: I will not be addressing this

jsonrpc/pretty_error.go
```diff
@@ -0,0 +1,208 @@
...
+	)
+	switch {
+	case errors.As(err, &typeErr):
+		return describeTypeError(typeErr)
+	case errors.Is(err, io.ErrUnexpectedEOF), errors.Is(err, io.EOF):
+		return "unexpected end of input"
+	case errors.As(err, &syntaxErr):
+		return describeSyntaxError(input, offset, err)
+	default:
+		return err.Error()
```
rodrodros, line 125: Why not apply the same switch case order as in `errorOffset`

jsonrpc/pretty_error.go
rodrodros, line 1: I appreciate how the functions are sorted 👍

jsonrpc/pretty_error.go
```diff
@@ -0,0 +1,208 @@
...
+	if len(p) >= maxWindowSize {
+		c.window = append(c.window[:0], p[len(p)-maxWindowSize:]...)
+		return len(p), nil
+	}
+
+	// temporarily increases in size
+	c.window = append(c.window, p...)
+	if len(c.window) > maxWindowSize {
+		c.window = c.window[:copy(c.window, c.window[len(c.window)-maxWindowSize:])]
+	}
```
rodrodros: This seems expensive to me, why copy something (potentially regrowing the array) to then overwrite it. 
I am curious if there is a better way by  avoid the call to append?

Nothing wrong with the current approach, I benchmarked one with a fixed array with size 512 and performance degraded for anything but the 10 megabyte case. 

Perhaps a low hanging fruit, if too complicated or not possible, let's leave as is

jsonrpc/server_test.go
```diff
@@ -209,11 +209,153 @@ func TestHandle(t *testing.T) {
...
 		"invalid json": {
 			req: `{]`,
-			res: `{"jsonrpc":"2.0","error":{"code":-32700,"message":"Parse error","data":"invalid character ']' looking for beginning of object key string"},"id":null}`,
+			res: `{"jsonrpc":"2.0","error":{"code":-32700,"message":"Parse error","data":"{]\n ^\nunexpected ']', expected a string key or '}' [line 1, position 2]"},"id":null}`,
 		},
 		"invalid json batch path": {
 			req: `[{]`,
-			res: `{"jsonrpc":"2.0","error":{"code":-32700,"message":"Parse error","data":"invalid character ']' looking for beginning of object key string"},"id":null}`,
+			res: `{"jsonrpc":"2.0","error":{"code":-32700,"message":"Parse error","data":"[{]\n  ^\nunexpected ']', expected a string key or '}' [line 1, position 3]"},"id":null}`,
+		},
```
rodrodros: The same function has grown so much that I believe you should make a separate test function directly targetting errors. It is not very readable having a hundred+ line test function doing more than one thing

jsonrpc/server_test.go
```diff
@@ -209,11 +209,153 @@ func TestHandle(t *testing.T) {
...
+		"trailing comma in object": {
+			req: `{
+  "jsonrpc": "2.0",
+  "method": "starknet_blockNumber",
+}`,
+			res: `{"jsonrpc":"2.0","error":{"code":-32700,"message":"Parse error","data":"{\n  \"jsonrpc\": \"2.0\",\n  \"method\": \"starknet_blockNumber\",\n}\n^\nunexpected trailing comma before '}' [line 4, position 1]"},"id":null}`,
+		},
+		"trailing comma with whitespace": {
+			req: `{"jsonrpc": "2.0", "method": "starknet_blockNumber", }`,
+			res: `{"jsonrpc":"2.0","error":{"code":-32700,"message":"Parse error","data":"{\"jsonrpc\": \"2.0\", \"method\": \"starknet_blockNumber\", }\n                                                     ^\nunexpected trailing comma before '}' [line 1, position 54]"},"id":null}`,
```
rodrodros: Also between every request and response together is hard to read, can you put a whiteline between pairs:
```
req:"..."
res:"..."
                                                                              <- white line
req:"..."
...

jsonrpc/pretty_error.go
RafaelGranza: C-style is cracked 🚀

jsonrpc/server_test.go
RafaelGranza: Sure

jsonrpc/pretty_error.go
RafaelGranza: I'll shift before appending.

PR #3703, opened 2026-06-04
chore(vm): update deps to v0.19.0-rc.0

rodrodros:


RafaelGranza (approved):
Reserialized the tests and they matched.
Also smoke tested locally, things worked.

LGTM

PR #3704, opened 2026-06-04
refactor: use l1/eth types in non-L1 callers

brbrr:
Switches the eth types usage from geth to our own defined types on non-l1 callers.

### DIFF
--- a/rpc/v10/wire_parity_test.go
+++ b/rpc/v10/wire_parity_test.go
@@ -10,7 +10,12 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-// See rpc/v8/wire_parity_test.go for rationale. Same JSON shape as v8/v9.
+// These tests pin the JSON-RPC wire shape of MsgToL1 and MsgFromL1 after
+// migrating their address fields from go-ethereum's common.Address to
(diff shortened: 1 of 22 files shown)


rpc/v10/wire_parity_test.go
```diff
@@ -0,0 +1,73 @@
...
+	"encoding/json"
+	"testing"
+
+	"github.com/NethermindEth/juno/core/felt"
+	"github.com/NethermindEth/juno/l1/eth"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// See rpc/v8/wire_parity_test.go for rationale. Same JSON shape as v8/v9.
```
rodrodros: if v8 is going to deprecated meaning v8 will eventually be deleted, shouldn't the rationale go in the latest rpc version?

rpc/v10/wire_parity_test.go
```diff
@@ -0,0 +1,78 @@
+package rpcv10
```
RafaelGranza:
```suggestion
package rpcv10_test
```

Same with the others

RafaelGranza (approved):
LGTM. 
One small nit

rodrodros (approved):
lgtm as well

PR #3707, opened 2026-06-06
fix(preconfirmed): race confiditions

EgeCaner:
Makes the pending poller the single writer to pre_confirmed storage (drops the eager writes in storeTask/revertTask), adds a guard so a blank-identifier empty can't overwrite a valid round at the same height via identifier drift, and derives poll parameters from a single ReadUnsafe() of storage instead of a separate blockNumberToPoll atomic.

### DIFF
--- a/sync/pending_polling.go
+++ b/sync/pending_polling.go
@@ -3,10 +3,8 @@ package sync
 import (
 	"context"
 	"errors"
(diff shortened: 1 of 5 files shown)


rodrodros (approved):
LGTM

PR #3708, opened 2026-06-08
fix(l1): juno shutdown doesn't hang

### LINKED ISSUE #3684: Avoid waiting time on context cancellation

rodrodros:
Fix https://github.com/NethermindEth/juno/issues/3684

### DIFF
--- a/l1/contract/starknet.go
+++ b/l1/contract/starknet.go
@@ -41,6 +41,7 @@ type StarknetFilterer struct {
 
 // StarknetLogStateUpdate represents a LogStateUpdate event raised by the Starknet contract.
 type StarknetLogStateUpdate struct {
+	// I am pretty sure this shouldn't be `big.Int` and more closesly
 	GlobalRoot  *big.Int
 	BlockNumber *big.Int
 	BlockHash   *big.Int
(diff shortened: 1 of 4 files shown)


l1/contract/starknet.go
```diff
@@ -57,7 +58,12 @@ func NewStarknetFilterer(address common.Address, filterer bind.ContractFilterer)
 }
 
 // bindStarknet binds a generic wrapper to an already deployed contract.
-func bindStarknet(address common.Address, caller bind.ContractCaller, transactor bind.ContractTransactor, filterer bind.ContractFilterer) (*bind.BoundContract, error) {
+func bindStarknet(
+	address common.Address,
+	caller bind.ContractCaller,
+	transactor bind.ContractTransactor,
+	filterer bind.ContractFilterer,
+) (*bind.BoundContract, error) {
```
RafaelGranza, line 66: Punished by the linter gods, eh?

l1/l1.go
```diff
@@ -361,27 +372,35 @@ func (c *Client) catchUpL1HeadUpdates(ctx context.Context) error {
 	}
 }
 
-func (c *Client) finalisedHeight(ctx context.Context) uint64 {
-	const finalisedHeightTimeout = 30 * time.Second
+// finalisedHeight blocks until the L1 finalised height is retrieved. If
+// context is cancelled, returns false
+func (c *Client) finalisedHeight(ctx context.Context) (uint64, bool) {
+	timer := time.NewTimer(0)
+	defer timer.Stop()
```
RafaelGranza, line 379: Nice way to solve it

RafaelGranza (approved):
Nicer than the `time.After` version I suggested earlier.

l1/contract/starknet.go
rodrodros: Nah, I actually find joy in cleaning Juno a bit, it was from own will 😅

PR #3712, opened 2026-06-08
chore(pruner): add information to log and errors

rodrodros:
And update prune-mode value to 128


EgeCaner (approved):
I agree with github actions comments, rest looks good to me!

PR #3719, opened 2026-06-11
perf(hash): speed up Poseidon hash by ~6%

RafaelGranza:
Small "implementation" Optimizations.
No behavior change, no signature change, same hashes, same test vectors.

I'm delivering this because it is a low hanging fruit. 
Further optimizations ("implementation" and  "algorithmic" wise) need to be explored.

### Changes
- `roundKeys`: slice-of-slices → fixed `[91][3]felt.Felt`, contiguous, one less pointer indirection per round
- state accessed as `*[3]felt.Felt` so the compiler drops bounds checks
- inlined the hash steps into the `HadesPermutation` loop to avoid per-round call indirections. 

### Benchmarks (benchstat, n=10, Apple M4)

Real-world:
| benchmark | base | new | Δ |
|---|---|---|---|
| SierraClassHash | 159.5µs | 150.2µs | -5.79% |
| TransactionV3Hash | 116.3µs | 109.7µs | -5.71% |

Micro (`core/crypto`):
| benchmark | Δ |
|---|---|
| Poseidon | -5.04% |
| PoseidonArray (3..40 felts) | -4.9% to -8.4% |
| geomean | -6.44% |

All `p=0.000`, allocations unchanged.

### DIFF
--- a/core/crypto/poseidon_hash.go
+++ b/core/crypto/poseidon_hash.go
@@ -14,9 +14,8 @@ const (
 
 // HadesPermutation applies the Hades permutation to state in place.
 // The hashing steps are intentionally inlined for performance reasons.
-func HadesPermutation(stateSlice []felt.Felt) {
+func HadesPermutation(state *[3]felt.Felt) {
 	initialiseRoundKeys.Do(setRoundKeys)
-	state := (*[3]felt.Felt)(stateSlice)
 
 	var squared, stateSum, triple felt.Felt
 	for i := range totalRounds {
@@ -49,8 +48,8 @@ var two = felt.FromUint64[felt.Felt](2)
 //
 // [Poseidon hash]: https://docs.starknet.io/learn/protocol/cryptography#poseidon-hash
 func Poseidon(x, y *felt.Felt) felt.Felt {
-	state := []felt.Felt{*x, *y, two}
-	HadesPermutation(state)
+	state := [3]felt.Felt{*x, *y, two}
+	HadesPermutation(&state)
 	return state[0]
 }
 
@@ -62,12 +61,12 @@ func Poseidon(x, y *felt.Felt) felt.Felt {
 //
 // [Poseidon array hashing]: https://docs.starknet.io/learn/protocol/cryptography#array-hashing
 func PoseidonArray(elems ...*felt.Felt) felt.Felt {
-	state := []felt.Felt{{}, {}, {}}
+	state := [3]felt.Felt{}
 
 	for i := range len(elems) / 2 {
 		state[0].Add(&state[0], elems[2*i])
 		state[1].Add(&state[1], elems[2*i+1])
-		HadesPermutation(state)
+		HadesPermutation(&state)
 	}
 
 	rem := len(elems) % 2
@@ -75,7 +74,7 @@ func PoseidonArray(elems ...*felt.Felt) felt.Felt {
 		state[0].Add(&state[0], elems[len(elems)-1])
 	}
 	state[rem].Add(&state[rem], &felt.One)
-	HadesPermutation(state)
+	HadesPermutation(&state)
 
 	return state[0]
 }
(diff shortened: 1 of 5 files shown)


core/crypto/poseidon_hash.go
```diff
@@ -9,45 +9,37 @@ import (
...
-	squared := felt.Felt{}
-	state[2].Mul(&state[2], squared.Mul(&state[2], &state[2]))
-	if full {
-		state[0].Mul(&state[0], squared.Mul(&state[0], &state[0]))
-		state[1].Mul(&state[1], squared.Mul(&state[1], &state[1]))
-	}
-}
+	var squared, stateSum, triple felt.Felt
+	for i := range totalRounds {
+		full := (i < fullRounds/2) || (totalRounds-i <= fullRounds/2)
```
RafaelGranza, line 22: Splitting the for loop into two (full true and full false) did not result in any speedup.

core/crypto/poseidon_hash.go
```diff
@@ -9,45 +9,37 @@ import (
...
-func addRoundKeys(state []felt.Felt, round int) {
-	state[0].Add(&state[0], &roundKeys[round][0])
-	state[1].Add(&state[1], &roundKeys[round][1])
-	state[2].Add(&state[2], &roundKeys[round][2])
-}
+// HadesPermutation applies the Hades permutation to state in place.
+// The hashing steps are intentionally inlined for performance reasons.
+func HadesPermutation(stateSlice []felt.Felt) {
+	initialiseRoundKeys.Do(setRoundKeys)
+	state := (*[3]felt.Felt)(stateSlice)
```
RafaelGranza: I decided to not change the signature of `HadesPermutation` since this method is exposed externally.

core/crypto/poseidon_hash.go
```diff
@@ -9,45 +9,37 @@ import (
...
 
-func subWords(state []felt.Felt, full bool) {
-	squared := felt.Felt{}
-	state[2].Mul(&state[2], squared.Mul(&state[2], &state[2]))
```
RafaelGranza, line 22: I basically inlined these helpers

core/crypto/poseidon_hash.go
```diff
@@ -90,23 +82,21 @@ func PoseidonArray(elems ...*felt.Felt) felt.Felt {
 
 var (
 	initialiseRoundKeys sync.Once
-	roundKeys           = [][]felt.Felt{}
+	roundKeys           [totalRounds][3]felt.Felt
```
RafaelGranza, line 84: fixed size

core/crypto/poseidon_hash.go
rodrodros: I agree in this particular case. We have to be careful with inlining function because even if they're being called once they are making the code more complex

core/crypto/poseidon_hash.go
rodrodros: What happens if the input doesn't match with the type cast? 

I don't think "it is being used externally" a valid reason for not updating the signature.

It's used inside of Juno as far as I am aware

RafaelGranza: **Type cast:** it doesn't add a failure mode. The old slice version already indexed `state[0..2]` directly, so an input shorter than 3 would have panicked with index-out-of-range anyway. The conversion panics on the exact same input (`len < 3`), just earlier and with a clearer message; longer inputs behave identically (first 3 used).

**Signature:** as far as I could find, Juno is the canonical Go implementation of the Starknet Poseidon hash, and `core/crypto` is imported externally (e.g. `starknet.go`), so someone *could* be depending on `HadesPermutation` through the exported API. That was my reasoning, not internal use alone. 
**Though I understand it's a weak point, since it's low-level stuff inside our exported API that nobody should really be calling directly.**

I can switch it to `*[3]felt.Felt` if it better matches our standards. And it's cleaner anyway.

rodrodros (approved):

PR #3723, opened 2026-06-12
feat(sync): multi-block pre_confirmed chain,  poller + storage

EgeCaner:
## Summary
This PR implements poller and storage in order to support multiple preconfirmed. Currently multi preconfirmed logic is not wired to production. Following PR will wire it to production and will remove prelatest, and candidate transactions.
  - Add `PreConfirmedBlockLatest` to the feeder client and propagate it through
    `starknetdata.StarknetData` and `sync.DataSource`. This hits the sequencer's
    "latest" variant of `get_preconfirmed_block`, whose response carries a
    top-level `block_number` so the caller can discover the server's
    pre_confirmed tip without tracking the height itself. The shared
    `PreConfirmedUpdateEnvelope` now decodes that field.
  - Introduce the `sync/preconfirmed` package with two pieces:
    - **`ChainStorage`** — an immutable, lock-free linked list of pre_confirmed
      blocks anchored above the canonical head. Single writer (the poller) /
      many concurrent readers via `atomic.Pointer`. Exposes `SnapshotForHead`
      (bounded by `BlockHashLag`) for consumers, `UnsafeSnapshot` for the
      poller, and `ApplyUpdate` / `AdvanceTo` for chain evolution. Handles
      bootstrap, extension, in-chain replacement (Delta or new round), and
      realignment when head moves or reverts.
    - **`Poller`** — single-goroutine tick loop: advance storage to head,
      poll the server's `latest`, backfill any gap below it (re-polling the
      previous tip with delta hints to capture its final view), then apply
      the latest update. NoChange / Delta / Full variants all dispatched.


starknet/pre_confirmed_update.go
```diff
@@ -68,21 +68,30 @@ var (
...
 }
 
 func (e *PreConfirmedUpdateEnvelope) UnmarshalJSON(data []byte) error {
 	var peek struct {
-		Changed   *bool   `json:"changed"`
-		Timestamp *uint64 `json:"timestamp"`
+		Changed     *bool   `json:"changed"`
+		Timestamp   *uint64 `json:"timestamp"`
+		BlockNumber *uint64 `json:"block_number"`
 	}
```
thiagodeev: Suggestion: we're unmarshalling the entire block twice here only to know the update type, but if we pay attention, all fields from Delta and all the fields from `peek` (except 'blockNumber') are present in the Full response. WDYT of changing `peek` to be `PreConfirmedBlock` + BlockNumber, so that we unmarshal once and populate eveything with it?


```suggestion
	var peek struct {
		PreConfirmedBlock
		BlockNumber *uint64 `json:"block_number"`
	}
```

clients/feeder/feeder.go
```diff
@@ -400,6 +405,35 @@ func (c *Client) PreConfirmedBlockWithIdentifier(
...
+
+// PreConfirmedBlockLatest fetches the highest pre_confirmed block the server
+// currently exposes. The response carries its block_number so the caller can
+// discover the pre_confirmed tip without tracking  the height itself.
+// Pass an empty identifier and zero txCount for a full reply.
+func (c *Client) PreConfirmedBlockLatest(
+	ctx context.Context,
+	blockIdentifier string,
+	knownTransactionCount uint64,
+) (starknet.PreConfirmedUpdate, uint64, error) {
```
rodrodros, line 411: I think it would be good to add a direct unit test to this function, even if it is tested indirectly, asserting this specific behaviour can help further on by pinpointing the issue

clients/feeder/feeder.go
```diff
@@ -400,6 +405,35 @@ func (c *Client) PreConfirmedBlockWithIdentifier(
...
+	env, err := c.fetchPreConfirmedUpdate(ctx, "latest", blockIdentifier, knownTransactionCount)
```
rodrodros: I am sure by `env` you mean `envelope` but by it being so much associated with `environment`, I believe it is better to use the full name or another name. I think just `envelope` can be as confusing. 

Suggestions: 
* `preConfirmedUpdate`
* `preConfirmedEnvelope`
* `preconfirmed`

sync/preconfirmed/chain_storage.go
```diff
@@ -0,0 +1,567 @@
...
+// wire send and the storage apply.
+var ErrBaseTxCountMismatch = errors.New("pre_confirmed base transaction count mismatch")
+
+// node is one entry in the chain's immutable linked list, pointing back
+// toward older blocks via parent. Nodes are never mutated in place — every
+// storage write produces fresh nodes for the affected slot and everything
+// newer than it, so concurrent readers walking a prior snapshot see a stable
+// graph. Popped nodes become unreferenced and GC-collectable.
+type node struct {
+	pc     *pending.PreConfirmed
```
rodrodros: Let's use descriptive name, `pc` can be many things. `preconfirmed` is a good one

sync/preconfirmed/chain_storage.go
```diff
@@ -0,0 +1,567 @@
...
+		return 0
+	}
+	return c.length
+}
+
+// Head returns the most recent pre-confirmed in the view, or nil if empty.
+func (c *ChainReader) Head() *pending.PreConfirmed {
+	if c == nil || c.length == 0 {
+		return nil
+	}
```
rodrodros: Shouldn't `c == nil` and `c.length == 0` be always true at the same time? Is there a case when one condition is met but the other isn't?

sync/preconfirmed/chain_storage.go
```diff
@@ -0,0 +1,567 @@
...
+	return c.head.pc
+}
+
+// NewestFirst yields entries from the most recent down to head+1, bounded by Length.
+func (c *ChainReader) NewestFirst() iter.Seq[*pending.PreConfirmed] {
+	return func(yield func(*pending.PreConfirmed) bool) {
+		if c == nil {
+			return
+		}
```
rodrodros: Why would `c` be nil, which scenario is this? Why would you call this on a `nil` pointer?

sync/preconfirmed/chain_storage.go
```diff
@@ -0,0 +1,567 @@
...
+			n = n.parent
+		}
+	}
+}
+
+// OldestFirst yields entries from head+1 up to the most recent, bounded by Length.
+func (c *ChainReader) OldestFirst() iter.Seq[*pending.PreConfirmed] {
+	return func(yield func(*pending.PreConfirmed) bool) {
+		if c == nil {
+			return
```
rodrodros: Same, it doesn't make sense to me that you need this check here.

sync/preconfirmed/chain_storage.go
```diff
@@ -0,0 +1,567 @@
...
+		walkOldestFirst(c.head, c.length, yield)
+	}
+}
+
+// TransactionByHash scans every chain entry's Block.Transactions and the
+// most recent entry's CandidateTxs. Returns ErrTransactionNotFound when missing.
+// Candidates live only at the most recent slot (they represent the next round's
+// speculation); lower entries are sealed from the sequencer's perspective.
+func (c *ChainReader) TransactionByHash(hash *felt.Felt) (core.Transaction, error) {
+	if c == nil || c.length == 0 {
```
rodrodros: I am really against this nil stuff check at the beginning. My point remains, why would you call on a nil `ChainReader`.

sync/preconfirmed/chain_storage.go
```diff
@@ -0,0 +1,567 @@
...
+		}
+		walkOldestFirst(c.head, c.length, yield)
+	}
+}
+
+// TransactionByHash scans every chain entry's Block.Transactions and the
+// most recent entry's CandidateTxs. Returns ErrTransactionNotFound when missing.
+// Candidates live only at the most recent slot (they represent the next round's
+// speculation); lower entries are sealed from the sequencer's perspective.
+func (c *ChainReader) TransactionByHash(hash *felt.Felt) (core.Transaction, error) {
```
rodrodros, line 102: This function is completely untested

sync/preconfirmed/chain_storage.go
```diff
@@ -0,0 +1,567 @@
...
+	for entry := range c.OldestFirst() {
+		if entry.Block.Number == blockNumber {
+			target = entry
+			break
+		}
+		stateDiff.Merge(entry.StateUpdate.StateDiff)
+	}
+	if target == nil {
+		return nil, pending.ErrPreConfirmedNotFound
+	}
```
thiagodeev, line 198: This check is unecessary, correct? We already know that `c.lenth` is > 0, and that the block number is between `bottom` and head block number; `target` should never be `nil`. The only case would be if we stored a nil preconfirmed in the chain, but that would be a bug, right?

sync/preconfirmed/chain_storage.go
```diff
@@ -0,0 +1,567 @@
...
+			viewHead = viewHead.parent
+		}
+		viewTip = upperBound
+	}
+
+	want := int(viewTip - wantBottom + 1)
+	if viewHead == c.head && want == c.length {
+		return *c
+	}
+	return ChainReader{head: viewHead, length: want}
```
thiagodeev: WDYT? We are returning by value anyway; same bytes size. Measuring both, I vote on remove this check to simplify a little bit the logic
```suggestion
	return ChainReader{head: viewHead, length: int(viewTip - wantBottom + 1)}
```

clients/feeder/feeder.go
EgeCaner: changed as `preConfirmedEnvelope`

sync/preconfirmed/chain_storage.go
EgeCaner: changed as `preconfirmed`

sync/preconfirmed/chain_storage.go
EgeCaner: removed the nil check on receiver

sync/preconfirmed/chain_storage.go
EgeCaner: removed the nil check on receiver

sync/preconfirmed/chain_storage.go
EgeCaner: removed the nil check on receiver

sync/preconfirmed/chain_storage.go
EgeCaner: removed the nil check on receiver

sync/preconfirmed/chain_storage.go
EgeCaner: Sorry seems like tests for this was in the next PR, I applied them to current one

starknet/pre_confirmed_update_test.go
```diff
@@ -22,146 +22,137 @@ func loadFeederTestdata(t *testing.T, relPath string) []byte {
...
+			var env starknet.PreConfirmedUpdateEnvelope
+			require.NoError(t, json.Unmarshal(raw, &env))
+			require.Equal(t, uint64(10936237), env.BlockNumber, "latest endpoint carries block_number")
+
+			full, ok := env.Update.(starknet.PreConfirmedBlock)
+			require.True(t, ok, "expected PreConfirmedBlock, got %T", env.Update)
+			require.Equal(t, "0x1cbe25d9", full.BlockIdentifier)
+			require.Equal(t, "PRE_CONFIRMED", full.Status)
+			require.NotZero(t, full.Timestamp)
+			require.NotNil(t, full.SequencerAddress)
```
thiagodeev: Nitpick: here and in the other parts, we could use `assert` instead of `required`

starknet/pre_confirmed_update_test.go
```diff
@@ -22,146 +22,137 @@ func loadFeederTestdata(t *testing.T, relPath string) []byte {
...
+
+			var env starknet.PreConfirmedUpdateEnvelope
+			require.NoError(t, json.Unmarshal(raw, &env))
+			require.Equal(t, uint64(10936237), env.BlockNumber, "latest endpoint carries block_number")
+
+			full, ok := env.Update.(starknet.PreConfirmedBlock)
+			require.True(t, ok, "expected PreConfirmedBlock, got %T", env.Update)
+			require.Equal(t, "0x1cbe25d9", full.BlockIdentifier)
+			require.Equal(t, "PRE_CONFIRMED", full.Status)
+			require.NotZero(t, full.Timestamp)
```
thiagodeev: Not related to your changes, but could you please add the timestamp validation for the PreConfirmedBlock.validate() method? I forgot to add in my PR.

Also, maybe we could use the PreConfirmedUpdateEnvelope.Validate() method here instead of all these manual assertions. Do as you prefer

starknet/pre_confirmed_update.go
```diff
@@ -140,21 +140,30 @@ var (
...
 //   - "changed": true + "timestamp"   → Full (new round)
 //   - "changed": true, no "timestamp" → Delta
+//
+// BlockNumber is set when the response carries a top-level "block_number"
+// (the "latest" endpoint includes it; the explicit-number endpoint does not,
+// since the caller already knows the requested number).
 type PreConfirmedUpdateEnvelope struct {
-	Update PreConfirmedUpdate
+	Update      PreConfirmedUpdate
+	BlockNumber uint64
```
thiagodeev, line 142: The new Validate() feat was implemented in main and included on this PR, but this PR introduces the new `BlockNumber` field, which is not being validated. However, it's hard to do so since the same type is used for both 'latest' and 'blockNumber' queries on the feeder 'get_preconfirmed_endpoint'. WDYT about adding a new type specific for the 'latest' call, use it in the feeder endpoint validation, and before returning it, cast it to the `PreConfirmedUpdateEnvelope` type?
This is just one way of doing it, but the main point is that we should validate this as well somehow

thiagodeev (commented):
Overall, LGTM.
I will not approve it since it's not supposed to be merged as it is (as you said, the changes must be integrated). Also, there are some points missing that I will only be able to properly review in the full PR integrating it

I left a few comments

sync/preconfirmed/chain_storage.go
EgeCaner: You're right, any block passing the [bottom, head] check is guaranteed present in a contiguous chain, so a nil target only means a gap, i.e. a bug. I kept the check and turned it into an invariant check and updated the error accordingly.

starknet/pre_confirmed_update.go
EgeCaner: Pre-confirmed decode benchmark
Summary: Decoding the stream directly via DecodePreConfirmedUpdate (no UnmarshalJSON) parses the payload once vs. one extra full re-scan for a single-pass UnmarshalJSON, and two for the original peek-then-re-decode. Each removed scan roughly halves decode time (full_big: 923µs → 656µs → 383µs; ~2.4× faster overall), at identical allocations.

Variants:
  - original — pre-refactor UnmarshalJSON that peeks for the discriminator, then re-decodes into the concrete variant (2 full byte-scans).
  - option1 — conventional UnmarshalJSON that decodes once into the flat wire struct and discriminates structurally (1 re-scan).
  - current — DecodePreConfirmedUpdate, decodes the stream directly with no UnmarshalJSON (0 re-scans).
```ml
  BenchmarkPreConfirmedDecodeVariants/full/original_unmarshaler-14           7168     165346 ns/op    44389 B/op     324 allocs/op
  BenchmarkPreConfirmedDecodeVariants/full/option1_unmarshaler-14            9933     116690 ns/op    44160 B/op     316 allocs/op
  BenchmarkPreConfirmedDecodeVariants/full/current_directDecode-14          17205      69790 ns/op    43865 B/op     310 allocs/op
  BenchmarkPreConfirmedDecodeVariants/full_big/original_unmarshaler-14       1292     923323 ns/op   318737 B/op    1453 allocs/op
  BenchmarkPreConfirmedDecodeVariants/full_big/option1_unmarshaler-14        1815     655790 ns/op   318513 B/op    1445 allocs/op
  BenchmarkPreConfirmedDecodeVariants/full_big/current_directDecode-14       3040     383155 ns/op   318232 B/op    1439 allocs/op
  BenchmarkPreConfirmedDecodeVariants/delta/original_unmarshaler-14         21176      56646 ns/op    19720 B/op     131 allocs/op
  BenchmarkPreConfirmedDecodeVariants/delta/option1_unmarshaler-14         29802      40317 ns/op    19600 B/op     124 allocs/op
  BenchmarkPreConfirmedDecodeVariants/delta/current_directDecode-14        48848      24343 ns/op    19304 B/op     118 allocs/op
  BenchmarkPreConfirmedDecodeVariants/no_change/original_unmarshaler-14   2178157       547.8 ns/op    1168 B/op      12 allocs/op
  BenchmarkPreConfirmedDecodeVariants/no_change/option1_unmarshaler-14    2086216       572.6 ns/op    1336 B/op      12 allocs/op
  BenchmarkPreConfirmedDecodeVariants/no_change/current_directDecode-14   3116310       384.7 ns/op    1160 B/op       9 allocs/op
```

starknet/pre_confirmed_update_test.go
EgeCaner: Added timestamp must be non-zero check to full preconfirmed `Validate`. I prefer to keep the manual checks here because they are asserting for exact value, Validate checks the structure

clients/feeder/feeder.go
```diff
@@ -382,6 +388,53 @@ func (c *Client) PreConfirmedBlockWithIdentifier(
...
+		"latest",
+		blockIdentifier,
+		knownTransactionCount,
+	)
+	if err != nil {
+		return nil, 0, err
+	}
+	// A full block on the latest query must carry its block_number; absent (or genesis 0)
+	// is invalid since the caller relies on it to discover the pre_confirmed tip.
+	if _, ok := preConfirmedEnvelope.Update.(starknet.PreConfirmedBlock); ok &&
```
rodrodros, line 423: Separate statement outside of `if`

Additionally, this seems to be a validators responsability when marshalling / unmarshalling or validating. WDYT?

starknet/pre_confirmed_update.go
```diff
@@ -135,44 +130,60 @@ var (
...
-//   - "changed": true + "timestamp"   → Full (new round)
+//   - "changed": true + "timestamp"   → Full block (new round)
 //   - "changed": true, no "timestamp" → Delta
-type PreConfirmedUpdateEnvelope struct {
-	Update PreConfirmedUpdate
+type preConfirmedWire struct {
+	Changed     *bool   `json:"changed"`
+	BlockNumber *uint64 `json:"block_number"`
+	PreConfirmedBlock
 }
```
rodrodros: The type is only used in the below function, why not use it inside it?

starknet/pre_confirmed_update.go
```diff
@@ -83,6 +75,9 @@ func (pb *PreConfirmedBlock) validate() error {
 	if pb.Version == "" {
 		return errors.New("version is required")
 	}
+	if pb.Timestamp == 0 {
+		return errors.New("timestamp is required")
```
rodrodros, line 79: We need a small validation test for this one

starknet/pre_confirmed_update.go
```diff
@@ -135,44 +130,60 @@ var (
...
+type PreConfirmedUpdateEnvelope struct {
+	Update      PreConfirmedUpdate
+	BlockNumber uint64
+}
+
+// preConfirmedWire is the flat shape the decoder fills. Discrimination is structural:
 //   - "changed": false                → NoChange
-//   - "changed": true + "timestamp"   → Full (new round)
+//   - "changed": true + "timestamp"   → Full block (new round)
 //   - "changed": true, no "timestamp" → Delta
```
rodrodros: Thanks for writing thoughtful comments.

starknet/pre_confirmed_update.go
```diff
@@ -135,44 +130,60 @@ var (
...
+func DecodePreConfirmedUpdate(r io.Reader) (PreConfirmedUpdateEnvelope, error) {
+	var raw preConfirmedWire
+	if err := json.NewDecoder(r).Decode(&raw); err != nil {
+		return PreConfirmedUpdateEnvelope{}, err
 	}
-	if err := json.Unmarshal(data, &peek); err != nil {
-		return err
+	if raw.Changed == nil {
+		return PreConfirmedUpdateEnvelope{}, errors.New(
+			"pre_confirmed update: missing required \"changed\" field",
```
rodrodros: I think the error can be just "missing required ..."

sync/preconfirmed/chain_storage.go
```diff
@@ -0,0 +1,581 @@
...
+// oldest-first order with contiguous block numbers. No args returns the
+// zero-value ChainReader.
+func NewChain(entries ...*pending.PreConfirmed) (ChainReader, error) {
+	if len(entries) == 0 {
+		return ChainReader{}, nil
+	}
+	var head *node
+	for i, pc := range entries {
+		if pc == nil {
+			return ChainReader{}, fmt.Errorf("building pre_confirmed chain: entry %d is nil", i)
```
rodrodros: This error can start directly from "entry..."

sync/preconfirmed/chain_storage.go
```diff
@@ -0,0 +1,581 @@
...
+		}
+		if i > 0 && pc.Block.Number != entries[i-1].Block.Number+1 {
+			return ChainReader{}, fmt.Errorf(
+				"building pre_confirmed chain: non-contiguous block numbers at index %d (%d after %d)",
```
rodrodros: same

sync/preconfirmed/chain_storage.go
```diff
@@ -0,0 +1,581 @@
...
+
+// NewestFirst yields entries from the most recent down to head+1, bounded by Length.
+func (c *ChainReader) NewestFirst() iter.Seq[*pending.PreConfirmed] {
+	return func(yield func(*pending.PreConfirmed) bool) {
+		n := c.head
+		for i := 0; i < c.length && n != nil; i++ {
+			if !yield(n.preconfirmed) {
+				return
+			}
+			n = n.parent
```
rodrodros: no single letter names, let's reserve those for method receivers

sync/preconfirmed/chain_storage.go
```diff
@@ -0,0 +1,581 @@
...
+		return nil, pending.ErrTransactionNotFound
+	}
+
+	for entry := range c.NewestFirst() {
+		for _, tx := range entry.Block.Transactions {
+			if tx.Hash().Equal(hash) {
+				return tx, nil
+			}
+		}
+	}
```
rodrodros, line 113: Idea for later, have some sort of cache that points you were to look a transaction hash 🤔 and transaction receipt but it might be unnecessary

rodrodros (approved):

rodrodros (changes requested):
Accidental approve

clients/feeder/feeder.go
EgeCaner: Ideally block number validation would live in `Validate`, but the envelope there can't tell whether the request used the `latest` tag or an explicit block number. To move it there we'd need either a separate envelope type for the latest response or an isLatest flag threaded into `starknet.DecodePreConfirmedUpdate(body)` to reject it during unmarshalling. We can also keep it as is. What do you recommend?

rodrodros: Ah I see, I think the check there is good then, not too many good solutions to choose from 👍

sync/preconfirmed/chain_storage.go
```diff
@@ -0,0 +1,569 @@
...
+// Implemented via recursion (no slice materialised); for a head-aligned view
+// the recursion depth is at most BlockHashLag, so the stack cost is bounded.
+func (c *ChainReader) OldestFirst() iter.Seq[*pending.PreConfirmed] {
+	return func(yield func(*pending.PreConfirmed) bool) {
+		if c == nil {
+			return
+		}
+		walkOldestFirst(c.head, c.length, yield)
+	}
+}
```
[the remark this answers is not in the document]
rodrodros: Irrelevant

thiagodeev (approved):

rodrodros (approved):

PR #3728, opened 2026-06-16
perf(hash): alloc-free hashes and empty-data cache

RafaelGranza:
## What
1. **Alloc-free `PoseidonDigest`** — the streaming digest no longer allocates.
2. **Short-circuit the most common Hades input `{0,1,0}`** (hashing a single zero felt): precomputed once at init and returned directly, skipping a full permutation.

## Allocs/op (backed by committed benchmarks, `b.ReportAllocs()`)
| benchmark | before | after |
|---|---|---|
| `BenchmarkSierraClassHash` | 15 | 1 |
| `BenchmarkTransactionV3Hash` | 14 | 3 |
| `BenchmarkPoseidonDigest` (40 felts) | 20 | 0 |

"before" = `origin/main`; "after" = this branch.

## Notes
~4% of Hades permutations on `SepoliaIntegration` blocks are exactly the empty-data input, now short-circuited. (~14% of permutations had repeated inputs overall. A broader memoization is a follow-up PR.)

### DIFF
--- a/core/class.go
+++ b/core/class.go
@@ -274,10 +274,10 @@ func SegmentedBytecodeHash(
 // Order matters (selector then index): it influences the class hash.
 func sierraEntryPointsHash(entryPoints []SierraEntryPoint) felt.Felt {
 	var digest crypto.PoseidonDigest
-	var index felt.Felt
-	for i := range entryPoints {
-		index.SetUint64(entryPoints[i].Index)
-		digest.Update(entryPoints[i].Selector, &index)
+	for _, ep := range entryPoints {
+		var index felt.Felt
+		index.SetUint64(ep.Index)
+		digest.Update(ep.Selector, &index)
 	}
 	return digest.Finish()
 }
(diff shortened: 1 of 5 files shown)


core/class.go
```diff
@@ -277,33 +271,37 @@ func SegmentedBytecodeHash(
...
-		result[2*i] = entryPoint.Selector
-		result[2*i+1] = felt.NewFromUint64[felt.Felt](entryPoint.Index)
+// Order matters (selector then index): it influences the class hash.
+func sierraEntryPointsHash(entryPoints []SierraEntryPoint) felt.Felt {
+	var digest crypto.PoseidonDigest
+	var index felt.Felt
+	for i := range entryPoints {
+		index.SetUint64(entryPoints[i].Index)
+		digest.Update(entryPoints[i].Selector, &index)
 	}
```
rodrodros: Why not create `index` inside the for loop?

core/class.go
```diff
@@ -277,33 +271,37 @@ func SegmentedBytecodeHash(
...
+	return digest.Finish()
+}
+
+// Order matters (selector, offset, builtins hash): it influences the class hash.
+func compiledEntryPointsHash(entryPoints []CasmEntryPoint, h Hasher) felt.Felt {
+	digest := h.NewDigest()
+	var offset, builtinsHash felt.Felt
+	for i := range entryPoints {
+		offset.SetUint64(entryPoints[i].Offset)
+		builtinsHash = compiledBuiltinsHash(entryPoints[i].Builtins, h)
```
rodrodros: Similar comment. Is it because of performance?

rodrodros (commented):
Looks good overall, but it would be nice to have benhcmarks

core/class.go
RafaelGranza: moved it inside

core/class.go
RafaelGranza: Passing `&offset` and` &builtinsHash` to `digest.Update` (an interface call) makes the compiler heap-allocate them. 
Declaring them once and reusing them keeps that to a single allocation instead of one per entry point.

RafaelGranza:
@rodrodros, I added `BenchmarkPoseidonDigest` with `b.ReportAllocs()` to lock in the digest's 0-alloc path.

`BenchmarkSierraClassHash` and `BenchmarkTransactionV3Hash` already exist with `ReportAllocs`, so those two counts are tracked too.

All three headline numbers now have a committed benchmark behind them.

core/class.go
rodrodros: Thank you, I see you added a comment 👍

rodrodros (approved):
LGTM!

PR #3730, opened 2026-06-17
feat(sync,rpc): integrate the multiple pre_confirmed mechanism

EgeCaner:
This PR replaces Juno's pre-confirmed handling with the new multiple pre_confirmed mechanism and removes pieces of the old design that are no longer needed such as the pre-latest block flow.


rpc/v10/events_test.go
```diff
@@ -641,132 +722,10 @@ func TestEvents_FilterWithLimit(t *testing.T) {
...
-		require.Equal(t, expected.TransactionHash, actual.TransactionHash)
-		require.Equal(t, expected.From, actual.From)
-		require.Equal(t, expected.Keys, actual.Keys)
-		require.Equal(t, expected.Data, actual.Data)
-	}
-}
+// TestEvents_ChainProgressesWhilePaginating was removed: it exercised the
+// pre_latest → pre_confirmed → canonical promotion path which no longer
+// exists with the multi-preconfirmed model. Equivalent multi-block coverage
+// lives in sync/preconfirmed/chain_storage_test.go.
```
thiagodeev: I think we can remove this, right? It's useful for those who are reviewing the PR, but not necessary to be kept in the code IMO

blockchain/blockchain_test.go
```diff
@@ -730,6 +735,221 @@ func TestEvents(t *testing.T) {
...
+	// (sepolia testdata) so we get coverage of slot traversal. Pagination
+	// subtests below compare against this set.
+	newFilter := func(t *testing.T) blockchain.EventFilterer {
+		t.Helper()
+		f, err := chain.EventFilter(nil, nil, preConfirmedFunc)
+		require.NoError(t, err)
+		require.NoError(t, f.SetRangeEndBlockByNumber(blockchain.EventFilterFrom, 0))
+		// ^uint64(0) is the "include all pre_confirmed" sentinel used by the
+		// RPC layer when the caller specifies the preConfirmed tag.
+		require.NoError(t, f.SetRangeEndBlockByNumber(blockchain.EventFilterTo, ^uint64(0)))
```
thiagodeev: Since our code mentions ` ^uint64(0)` so many times, WDYT about storing it with a variable name? We could also use `math.MaxUint64` instead, which is more meaningful IMO

thiagodeev (commented):
Partial review. Only nitpicks

sync/pending_polling.go
thiagodeev, line 1: WDYT of moving the `pollPendingData` method for the sync.go file, which already holds the `pollLatest` method ([ref](https://github.com/NethermindEth/juno/blob/feat/integ-multi-preconfirmed/sync/sync.go#L552-L571)) ? That way we could remove this file with only a single method

blockchain/event_filter.go
thiagodeev, line 1: Race condition: EventFilter.Events reads the chain tip twice for a getEvents query reaching up to pre_confirmed, which can be different from one call to another:

Let's say `latest` is 100, and we want to query from 90 to `pre_confirmed`.

  1. First read — `core.GetChainHeight` (event_filter.go:152), then canonicalEvents (event_filter.go:188) scans committed blocks up to that tip. Tip is block 100; it scans from 90 to 100.
  2. Second read — `pendingEvents` (event_filter.go:207) calls `preConfirmedFn` → `Synchronizer.PreConfirmedChain`, which re-reads the tip via HeadsHeader and builds the view with `ChainStorage.SnapshotForHead` (sync/sync.go:582). The view starts at head + 1.

  If the head advances between the two reads — e.g. blocks 101 and 102 get committed:

  - canonicalEvents already stopped at 100 (stale tip), so it never touched 101–102.
  - SnapshotForHead now starts the pre-confirmed view at 103 (new tip 102, +1), so it skips 101–102 too.

Blocks 101 and 102 are never scanned. Their events are dropped from the response, with no continuation token to recover them. Rare, but possible

thiagodeev (approved):
Overall, LGTM. But we need to address the race condition I mentioned here

blockchain/event_filter.go
rodrodros: Solved by Ege

rodrodros (approved):

PR #3731, opened 2026-06-17
perf(crypto): PoC Poseidon hash via Rust FFI (~1.7x)

RafaelGranza:
# PoC: Poseidon hash via Rust FFI (~1.7x faster)

> **Proof of concept.** Opening to discuss the approach, not to merge yet.

This swaps our Go Poseidon permutation for a call into Rust's version, behind a
build tag with the old Go code kept as a fallback. `Poseidon`, `PoseidonArray`
and `PoseidonDigest` all get faster for free, since they share one function.

### Apple M4, Arm

| benchmark | old Go | FFI | faster by |
|---|---|---|---|
| `Poseidon(x,y)` | ~8.7µs | ~4.5µs | ~1.9x |
| `PoseidonArray` (10k felts) | ~59ms | ~33ms | ~1.8x |
| `SierraClassHash` | ~151µs | ~84µs | ~1.8x |
| `TransactionV3Hash` | ~110µs | ~62µs | ~1.76x |
| AddTxn (verify hash + mempool push) | ~112µs | ~67µs | ~1.65x |

`ClassV0Hash` (uses Pedersen, not Poseidon) is unchanged, so nothing else was
affected.

### x86 (AMD EPYC 7642)
benchmark | gnark | FFI | faster by
-- | -- | -- | --
Poseidon(x,y) | ~7.42µs | ~4.92µs | 1.51x
PoseidonArray (10k) | ~56ms | ~36.6ms | 1.53x
SierraClassHash | ~203µs | ~125µs | 1.62x
TransactionV3Hash | ~140µs | ~80µs | 1.74x
AddTxn | ~152µs | ~90µs | 1.68x

## Why call Rust instead of just making the Go faster

Short version: in Go you can have a **fast** multiply or an **inlinable**
multiply, but not both. Rust gets both, and that is the whole gap.

The hash spends most of its time multiplying field elements. gnark's multiply is
written in assembly, so it is fast, but the Go compiler treats it as a black box:
it cannot mix the cheap additions (the "linear layer") into the gaps while the
multiply runs. Rust's compiler (LLVM) can, so the additions are basically free
there.

I tried every in-process option first. None worked:

| what I tried | result | why |
|---|---|---|
| Rewrite the additions inline in Go (returning values) | **2x slower** (~15µs) | lots of 32-byte copies on the stack |
| Write the mixing step in hand-written ARM64 assembly | **no change** | the mixing was never the slow part |
| Write the *whole* permutation in ARM64 assembly | **no change** | the multiply still spills to memory, so the additions still wait for it |
| Write the multiply in pure Go so the compiler can optimize it | **60% slower** (~12.5µs) | Go's compiler makes a multiply ~2x slower than the assembly |

So the only way to win is to use a compiler that produces a fast multiply AND
can inline it. That is Rust/LLVM. The cost of calling into Rust is tiny (~25ns,
vs a ~4.4µs hash), and Juno already calls Rust through cgo (the VM), so this adds
no new kind of dependency.

## How it works

- New small Rust crate `core/crypto/rust` that exports one function:
  `juno_hades_permutation`.
- We hand it the numbers exactly as they already sit in memory (no conversion).
  Go and Rust store field elements the same way, just in reversed limb order, so
  Rust only reverses them. I checked this matches our old output bit-for-bit over
  1,000,000 random inputs.
- A build tag (`no_rust_crypto`) picks the Rust path or the old Go path. With
  `CGO_ENABLED=0` it falls back to the old Go code, so it always builds.

## Is it correct?

`TestPoseidonCrossCheckRust` checks the output against known `starknet-crypto`
values (including 0, 1, and q-1) and passes on both the Rust and Go builds.
Existing Poseidon tests pass on both too.

## Things to decide (it is a PoC)

- This makes `core/crypto` need cgo (with the Go fallback). It is a widely used
  package, so that is a real decision.
- It adds 1 small allocation (96 bytes) per hash call. It does not grow with
  input size: a 10,000-felt array is still 1 allocation.
- The Rust crate must NOT use LTO. With LTO its Rust runtime symbols clash with
  the VM's Rust library when both end up in the same binary. The default build
  settings avoid that.

## Build

`make` builds the crate. Verified in all three modes: Rust (default), Go via
`-tags no_rust_crypto`, and `CGO_ENABLED=0` (Go).


rodrodros:
~~Why are we creating this PR, we will not add Rust to the codebase, specifically if the improvements are in the microseconds.~~

~~This was already discussed in the past daily. Why does this exist, there is nothing to decide.~~

My mistake, actually good work that shows that this specific part we won't be able to gain further performance from Go this way. Not even assembly shows anything. 

What we need to check is the differences on x64 and check if there are other ways of improving performance without resorting to another language.

PR #3737, opened 2026-06-18
feat(utils): Fix throttler behavior when max queue length is zero

haoshengzhen:
## Summary

This fixes `utils.Throttler` when `WithMaxQueueLen(0)` is used.

A zero queue length should mean that calls may run immediately when capacity is available, but no additional calls should wait once the resource is saturated. Previously, `Do` checked the queue length before trying to acquire an available slot, so even the first call could return `ErrResourceBusy`.

The updated logic first attempts a non-blocking acquire. Only when the throttler is already at capacity does it increment and check the queue length.

## Testing

- `go test ./utils -run 'TestThrottler' -count=1`
- `go test ./utils -count=1`
- `go test ./l1/eth ./jsonrpc ./utils ./utils/pipeline ./utils/broadcast ./utils/tracker -count=1`

### DIFF
--- a/utils/throttler.go
+++ b/utils/throttler.go
@@ -32,6 +32,15 @@ func (t *Throttler[T]) WithMaxQueueLen(maxQueueLen int32) *Throttler[T] {
 
 // Do lets caller acquire the resource within the context of a callback
 func (t *Throttler[T]) Do(doer func(resource *T) error) error {
+	select {
+	case t.sem <- struct{}{}:
+		defer func() {
+			<-t.sem
+		}()
+		return doer(t.resource)
+	default:
+	}
+
 	queueLen := t.queue.Add(1)
 	if queueLen > t.maxQueueLen {
 		t.queue.Add(-1)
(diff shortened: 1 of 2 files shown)


rodrodros:
Hi, thanks for the submission. Can you explain why would you need this?

haoshengzhen:
> Hi, thanks for the submission. Can you explain why would you need this?

Good question. The intent is to make `max-vm-queue=0` behave as a valid “no queue / fail fast” configuration.

`max-vm-queue` is exposed as the maximum number of requests allowed to wait after `max-vms` is reached. With a value of `0`, I would expect Juno to still execute requests immediately while a VM slot is available, and only reject requests when all VM slots are already busy.

The current implementation increments/checks the queue before trying to acquire an available slot, so `max-vm-queue=0` rejects even the first request while capacity is still available. That makes the zero value unusable and a bit inconsistent with the flag semantics.

This can be useful for latency-sensitive RPC setups where operators prefer fail-fast behavior under saturation instead of letting VM work queue up.

That said, if the intended behavior is that `max-vm-queue=0` should be invalid rather than meaning “no queue”, I can update the PR to add config validation instead.

rodrodros:
Hey, thanks for the submission, this is already fixed on #3750 so I will close it

PR #3727, opened 2026-06-15
refactor(l1): Add L1 Eth client

brbrr:
NOTE: This PR wouldn't be merged. The contents of this PR will split up in two. Keeping it as the reference.

Add a minimal Ethereum execution-layer client living under l1/eth/:

- l1/eth/client — JSON-RPC 2.0 over a single WebSocket, id-multiplexed so unary calls and eth_subscribe notifications share one conn. Surface: ChainID, BlockNumber, HeaderByNumber,
  TransactionReceipt, FilterLogs, SubscribeLogs.
- l1/eth/contract — Hand-decoded LogStateUpdate event (3×32-byte data, no indexed args). Replaces the abigen output.
- l1/eth_settlement.go — EthSettlement implements the new SettlementLayer interface used by l1.Client. All Ethereum-flavoured types (*big.Int, eth.Log, eth.Hash) stay inside this adapter; the L1 sync loop consumes a chain neutral StateUpdate.

Boundary changes:

- New l1.SettlementLayer interface replaces the old Subscriber. Methods are chain-neutral so a non-Ethereum settlement backend could slot in without touching l1.Client.
- The same EthSettlement instance also satisfies rpccore.L1Client (for starknet_getMessageStatus), so node.go constructs one client and hands it to both the sync loop and the RPC
  handlers.
- Deleted: l1/eth_subscriber.go, l1/contract/, rpc/rpccore/eth_receipt_adapter*.go, mocks/mock_subscriber.go.

Notes:

- `go-ethereum` dependencies are still present in tests. They will be removed in the follow-up PR.

PR #3738, opened 2026-06-18
feat(l1): introduce SettlementLayer and EthMessagingClient abstractions

brbrr:
This is the first part of #3727 (deprecated) that adds an abstraction layer between the Geth L1 client and Juno, separating the settlement layer and the actual RPC client.


rodrodros (commented):
This PR looks more complicated than it should, let's discuss it more on Monday

l1/l1_pkg_test.go
```diff
@@ -851,20 +861,20 @@ func TestSubscribeToUpdatesReturnsPromptlyOnCancel(t *testing.T) {
...
-			WatchLogStateUpdate(gomock.Any(), gomock.Any()).
+			WatchStateUpdate(gomock.Any(), gomock.Any()).
 			Return(nil, errors.New("boom")).
 			MinTimes(1)
 
 		client := NewClient(subscriber, chain, nopLog, WithResubscribeDelay(time.Hour))
 
 		ctx, cancel := context.WithCancel(t.Context())
-		done := make(chan event.Subscription, 1)
+		done := make(chan eth.Subscription, 1)
```
RafaelGranza: It reads very weird that `eth.Subscription` is a generic subscription, yet, it lives inside the `eth` package. I believe it would be a better fit living in `l1`.

for example, this file should not import `l1/eth`.

l1/l1_pkg_test.go
```diff
@@ -1,9 +1,8 @@
-package l1
+package l1_test
```
RafaelGranza, line 1: 🔥

l1/l1.go
```diff
@@ -101,23 +81,23 @@ func NewClient(
...
 	}
 }
 
 // subscribeToUpdates blocks until a subscription is established. If context is cancelled,
 // returns nil
 func (c *Client) subscribeToUpdates(
-	ctx context.Context, updateChan chan *contract.StarknetLogStateUpdate,
-) event.Subscription {
+	ctx context.Context, updateChan chan *StateUpdate,
+) eth.Subscription {
```
RafaelGranza: The same here, `Subscription` could be a "name neutral" interface.

node/migration.go
```diff
@@ -106,12 +107,13 @@ func fetchL1HeadIfMissing(
 
 	logger.Info("Fetching the L1 head before running the prune migration")
 	// Metrics are registered by the long-lived L1 client built in node.New; reusing
-	// them here would panic via prometheus.MustRegister.
-	client, err := newL1Client(config.EthNode, false, chain, logger)
+	// them here would panic via prometheus.MustRegister. Hence no listener.
+	settlement, err := newGethSettlement(ctx, config.EthNode, chain)
 	if err != nil {
 		return fmt.Errorf("creating a new L1 client: %w", err)
 	}
```
RafaelGranza, line 114: **Thoughts**: I guess the message is ok, since the `err` will explain it is a "dial" problem. Also, it is clearer to the user using the term `L1 client`.

l1/geth_settlement.go
```diff
@@ -0,0 +1,302 @@
...
+
+type gethSettlementOptions struct {
+	logger log.StructuredLogger
+}
+
+// WithSettlementLogger attaches a logger forwarded to the settlement.
+// Surfaces transport-level warnings at debug level.
+func WithSettlementLogger(l log.StructuredLogger) GethSettlementOption {
+	return func(o *gethSettlementOptions) { o.logger = l }
+}
```
RafaelGranza: Is this planning to be used in the next PR? 
Because the log is never used, and the comment `Surfaces transport-level warnings at debug level.` is not correct at the moment.

l1/geth_settlement.go
```diff
@@ -0,0 +1,302 @@
...
+		opt(&o)
+	}
+	logger := o.logger
+	if logger == nil {
+		logger = log.NewNopZapLogger()
+	}
+
+	rpcClient, err := rpc.DialContext(ctx, rawURL)
+	if err != nil {
+		return nil, fmt.Errorf("dial L1: %w", err)
```
RafaelGranza: Just as suggested by Rodrigo: "Errors should start 90% of time with `...ing`".

l1/geth_settlement.go
```diff
@@ -0,0 +1,302 @@
...
+	}
+	ethClient := ethclient.NewClient(rpcClient)
+	filterer, err := contract.NewStarknetFilterer(common.Address(contractAddress), ethClient)
+	if err != nil {
+		ethClient.Close()
+		return nil, fmt.Errorf("bind Starknet filterer: %w", err)
```
RafaelGranza: Same error format "binding Starknet filterer: %w"

l1/geth_settlement.go
```diff
@@ -0,0 +1,302 @@
...
+// off it), which means the listener can only be built AFTER the
+// settlement exists — hence a post-construction setter rather than a
+// constructor option.
+//
+// Concurrency contract: SetListener MUST be called before the
+// settlement is handed off to any goroutine (i.e. before l1.NewClient
+// in node.go). The field is read unlocked from every RPC method; a
+// concurrent SetListener would be a data race. The single-shot wiring
+// in node.go satisfies this — don't introduce a second caller.
+func (s *GethSettlement) SetListener(l EventListener) { s.listener = l }
```
RafaelGranza: I know you documented why this exists, but this is a VERY weird pattern IMO.

I'm not sure how to solve this, but is this really necessary?

l1/geth_settlement.go
```diff
@@ -0,0 +1,302 @@
...
+	t := time.Now()
+	return func() { s.listener.OnL1Call(method, time.Since(t)) }
+}
+
+// ChainID returns the Ethereum chain id (eth_chainId).
+func (s *GethSettlement) ChainID(ctx context.Context) (*big.Int, error) {
+	defer s.observe("eth_chainId")()
+	id, err := s.ethClient.ChainID(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("get chain id: %w", err)
```
RafaelGranza: Same here

l1/geth_l1_state_provider.go
```diff
@@ -0,0 +1,302 @@
...
+// FinalisedHeight returns the latest finalised L1 block number. A
+// missing finalised header is reported as eth.ErrNotFound so callers
+// can distinguish "node hasn't seen finality yet" from a transport
+// failure.
+func (s *GethSettlement) FinalisedHeight(ctx context.Context) (uint64, error) {
+	defer s.observe("eth_getBlockByNumber")()
+	head, err := s.ethClient.HeaderByNumber(ctx, gethFinalizedBlockNumber)
+	if err != nil {
+		if errors.Is(err, ethereum.NotFound) {
+			return 0, fmt.Errorf("finalised block not found: %w", eth.ErrNotFound)
```
RafaelGranza, line 122: This one seems fine

node/node.go
```diff
@@ -651,19 +678,18 @@ func newL1Client(
...
+		dialCtx, ethNode, chain.Network().CoreContractAddress, opts...,
+	)
 	if err != nil {
-		return nil, fmt.Errorf("set up ethSubscriber: %w", err)
-	}
-
-	opts := make([]l1.Option, 0, 1)
-	if includeMetrics {
-		opts = append(opts, l1.WithEventListener(makeL1Metrics(chain, ethSubscriber)))
+		return nil, fmt.Errorf("set up L1 settlement client: %w", err)
```
rodrodros: "setting up L1"

l1/l1_state_provider.go
```diff
@@ -0,0 +1,46 @@
...
+// SettlementLayer implementation does the felt.Felt conversion, so
+// l1.Client never touches go-ethereum types.
+type StateUpdate struct {
+	// L2BlockNumber is the Starknet block number being committed.
+	L2BlockNumber uint64
+	// L2BlockHash is the Starknet block hash for that block.
+	L2BlockHash *felt.Felt
+	// StateRoot is the Starknet global state root after the block
+	// (the "globalRoot" field in the on-chain event).
+	StateRoot *felt.Felt
```
rodrodros: Why are these pointers?

l1/geth_settlement.go
```diff
@@ -0,0 +1,277 @@
...
+// sync loop and the RPC handlers.
+//
+// WS connection drops are recovered below this layer: rpc.Client
+// reconnects unary calls, and subscription drops surface on Err() for
+// l1.Client.watchL1StateUpdates to resubscribe.
+type GethSettlement struct {
+	contractAddress eth.Address
+	url             string
+
+	rpcClient *rpc.Client
```
RafaelGranza: I assume these are going to be used in the next PR? since they are never read anywhere.

l1/export_test.go
```diff
@@ -0,0 +1,29 @@
+package l1
+
+import (
+	"context"
+)
+
+// Exposes package-private surface to the external l1_test package. The
+// tests live there to avoid a circular import: l1 -> mocks -> l1.
```
RafaelGranza: Our current pattern is not welcoming to this kind of test pattern (exposing private methods).
 I was trying to find ways to avoid this work-around, but IMO it is not worth the effort.

I think this is OK as it is.

l1/geth_forwarder_test.go
```diff
@@ -0,0 +1,155 @@
+package l1
+
+import (
+	"errors"
+	"math/big"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"github.com/NethermindEth/juno/core/felt"
```
RafaelGranza, line 10: This one also seems very hard to reach through public API.

Looks OK.

node/metrics.go
```diff
@@ -278,14 +282,33 @@ func makeL1Metrics(bcReader blockchain.Reader, l1Subscriber l1.Subscriber) l1.Ev
...
+		OnL1CallCb: func(method string, took time.Duration) {
+			requestLatencies.WithLabelValues(method).Observe(took.Seconds())
+		},
+	}
+}
+
+// registerL1SettlementMetrics registers the L1 height gauges that poll the
+// settlement layer. Kept separate from makeL1Metrics so the settlement can
+// be constructed with its listener already attached.
+func registerL1SettlementMetrics(settlement l1.SettlementLayer) {
```
RafaelGranza: This resolves my  previous `SetListener` concern. Now the listener is immutable after construction.
Much cleaner, thanks.

l1/l1_pkg_test.go
```diff
@@ -489,14 +499,14 @@ func TestUnreliableSubscription(t *testing.T) {
...
-			FilterLogStateUpdate(gomock.Any(), gomock.Any(), gomock.Any()).
+			FilterStateUpdate(gomock.Any(), gomock.Any(), gomock.Any()).
 			Return(nil, nil).
 			AnyTimes()
 
 		subscriber.EXPECT().Close().Times(1)
 
 		// Replace the subscriber.
-		client.l1 = subscriber
+		client.SetSettlement(subscriber)
```
rodrodros: This is very confusing, sometimes it is settlement, sometimes it is subscriber. Settlement in itself is a bit of an ambiguous name also. I find the subscriber name more straightforward, aren't we subscribed to an `L1` which give us state updates?

l1/l1_pkg_test.go
```diff
@@ -837,7 +852,7 @@ func TestFinalisedHeightReturnsPromptlyOnCancel(t *testing.T) {
 }
 
 // TestSubscribeToUpdatesReturnsPromptlyOnCancel is the same check for the
-// other retry loop: WatchLogStateUpdate fails repeatedly, the loop enters
+// other retry loop: WatchStateUpdate fails repeatedly, the loop enters
 // its inter-attempt wait, ctx is cancelled, and the function must return
 // without consuming resubscribeDelay.
 func TestSubscribeToUpdatesReturnsPromptlyOnCancel(t *testing.T) {
```
rodrodros, line 858: This is the classical function that tests implementation rather than functionality. It is my mistake for writing it this way, but if we are turning `L1` package into `L1_test` it sadly requires we rewrite tests.

We will not do that because it is not feasible, we will do smth else. See next comment

l1/export_test.go
```diff
@@ -0,0 +1,29 @@
...
+// SubscribeToUpdates exposes the resubscribe retry loop for testing
+// its interaction with context cancellation and resubscribeDelay.
+func (c *Client) SubscribeToUpdates(ctx context.Context, ch chan *StateUpdate) Subscription {
+	return c.subscribeToUpdates(ctx, ch)
+}
+
+// FinalisedHeight exposes the inner retry loop driven by setL1Head.
+func (c *Client) FinalisedHeight(ctx context.Context) (uint64, bool) {
+	return c.finalisedHeight(ctx)
+}
```
rodrodros: Add a deprecation tag for each of these function so that further use in tests is disincentivized, we will bypass the lint for this PR.

Additionally, write on top of the file:
```golang
// NOTE: This export file should stay as is and no other additions can be done to it.
// The goal is to eventually remove each of the method in this file by rewriting the tests
// to use the public API only.
```

node/node.go
```diff
@@ -619,30 +622,54 @@ func New(cfg *Config, version string, logLevel *log.Level) (*Node, error) {
...
 }
 
-func newL1Client(
-	ethNode string, includeMetrics bool, chain *blockchain.Blockchain, log log.StructuredLogger,
-) (*l1.Client, error) {
+// newGethSettlement validates the Ethereum endpoint URL and dials the L1
+// client. ws/wss is enforced at the URL level because subscribe-based
+// log delivery (eth_subscribe) requires a long-lived connection that
+// HTTP doesn't provide.
+func newGethSettlement(
```
rodrodros: `newGethSettlement`  doesn't say anything, someone reading the code out of the blue might be confused as for what a `gethSettlement` is. Please, let's find a better name.

l1/l1_state_provider.go
rodrodros, line 1: Let's call the file settlement_layer.go

l1/geth_settlement.go
```diff
@@ -0,0 +1,277 @@
...
+// GethSettlement is the go-ethereum-backed SettlementLayer, wrapping
+// ethclient plus the abigen StarknetFilterer to talk to the Starknet
+// core L1 bridge. It also satisfies rpccore.L1Client (via
+// TransactionReceipt), so node.go hands one instance to both the L1
+// sync loop and the RPC handlers.
+//
+// WS connection drops are recovered below this layer: rpc.Client
+// reconnects unary calls, and subscription drops surface on Err() for
+// l1.Client.watchL1StateUpdates to resubscribe.
+type GethSettlement struct {
```
rodrodros: I know this is the third time I comment on the same thing, but isn't this a client / subscription based mechanism? 

Maybe we can call the interface `SettlementLayerClient` or why not `L1Client`? 

L1 as in the general usage of it, it doesn't matter if it is an L3 and the L1 is an L2

rodrodros (commented):
First review, overall looks good and makes sense, will perform additional rounds

RafaelGranza (approved):
LGTM. Just [the one nit](https://github.com/NethermindEth/juno/pull/3738/changes#r3517487044) on the unused fields.

l1/geth_settlement.go
brbrr: I've removed them for now.

l1/l1_state_provider.go
brbrr: It was a mechanical change, driven by the old usage, and the fact that `L1Head` uses pointers. But also it seems consistent with `Felt` usage across the codebase 🤷. Do you think they should be values?

rodrodros: If the codebase use `**felt.Felt` you would write it here just like that because of what the codebase had?
I don't know if they should be values or pointers, I am asking why did you chose to define them as pointers. 

So, I repeat the question, why did you choose `*felt.Felt` over `felt.Felt`?

l1/geth_l1_state_provider.go
```diff
@@ -0,0 +1,262 @@
...
+			}
+		}
+	})
+}
+
+// GethL1StateProvider must satisfy both interfaces it serves.
+var (
+	_ L1StateProvider  = (*GethL1StateProvider)(nil)
+	_ rpccore.L1Client = (*GethL1StateProvider)(nil)
+)
```
rodrodros: Define on top of the file please

l1/geth_l1_state_provider.go
```diff
@@ -0,0 +1,262 @@
...
+
+// gethReceiptToEth copies the receipt fields juno reads. Only Logs is
+// consumed today; a consumer needing other fields must extend this.
+func gethReceiptToEth(r *types.Receipt) *eth.Receipt {
+	logs := make([]eth.Log, len(r.Logs))
+	for i, l := range r.Logs {
+		logs[i] = gethLogToEth(l)
+	}
+	return &eth.Receipt{Logs: logs}
+}
```
rodrodros: Return by value please

l1/l1_state_provider.go
rodrodros: You want pointers when the value is shared among many instances, or you want to stablish it as optional (meaning it can be defined or undefined (i.e. `nil`)). 

You want values when this doesn't apply. 

There is no performance relevant update, is about stablishing how do you think this values represent as part of the state.  I don't think `StateRoot` for example to something that could be nil in a state update 🤔

l1/l1.go
```diff
@@ -86,7 +65,7 @@ func WithCatchUpChunkSize(size uint64) Option {
 }
 
 func NewClient(
-	l1 Subscriber,
+	provider L1StateProvider,
 	chain *blockchain.Blockchain,
 	logger log.StructuredLogger,
```
rodrodros, line 70: Please add an issue that refactors the constructor into making the `logger` input part of the options

l1/geth_l1_state_provider_test.go
```diff
@@ -0,0 +1,343 @@
...
+// --- recording listener -----------------------------------------------
+
+type recordingListener struct {
+	mu    sync.Mutex
+	calls []recordedCall
+}
+
+type recordedCall struct {
+	method string
+	dur    time.Duration
```
rodrodros: full name for this field, duration

l1/geth_l1_state_provider_test.go
```diff
@@ -0,0 +1,343 @@
...
+// --- tests ------------------------------------------------------------
+
+func TestGethL1StateProvider_DialError(t *testing.T) {
+	// Closed server URL — dial succeeds (HTTP is connectionless) but the
+	// first RPC call should fail. We test failure via a real method
+	// instead of the constructor since http dialing doesn't actually
+	// open a connection eagerly.
+	srv := newGethJSONRPCServer(t)
+	srv.srv.Close()
+	ctx, cancel := context.WithTimeout(t.Context(), time.Second)
```
rodrodros, line 151: Except this test case, there is no point in any other test case in this file for having a timeout. Please remove them.

l1/geth_l1_state_provider_test.go
```diff
@@ -0,0 +1,343 @@
...
+	defer cancel()
+	s, err := l1.NewGethL1StateProvider(ctx, srv.URL(), eth.Address{})
+	require.NoError(t, err)
+	t.Cleanup(s.Close)
+	_, err = s.ChainID(ctx)
+	require.Error(t, err)
```
rodrodros: Additionally, all the test files check that there was a returned error, but they are not checking which error is returned. Please update for every test case, the expected error by using `require.ErrorContains(...)`

l1/geth_l1_state_provider.go
```diff
@@ -0,0 +1,262 @@
...
+//
+// Caller contract: sink MUST be drained promptly. A sink that stalls
+// back-pressures the abigen subscription channel and eventually the
+// underlying ws connection.
+func (s *GethL1StateProvider) WatchStateUpdate(
+	ctx context.Context,
+	sink chan<- *StateUpdate,
+) (Subscription, error) {
+	raw := make(chan *contract.StarknetLogStateUpdate, watchForwarderBuffer)
+	inner, err := s.filterer.WatchLogStateUpdate(&bind.WatchOpts{Context: ctx}, raw)
```
rodrodros: `raw`, `sink`, `inner`? Could we please use meaningful names.

Suggestions:
`raw` -> `rawEventsCh` or `gethEventsCh`
`inner` -> `sub` or `gethSub`

`sink` is optional, since it is geth nomenclature, but I would prefer `updatesSink` or `updatesCh` (since we are moving away from geth).

l1/geth_l1_state_provider.go
```diff
@@ -0,0 +1,262 @@
...
+// failure.
+func (s *GethL1StateProvider) FinalisedHeight(ctx context.Context) (uint64, error) {
+	defer s.observe("eth_getBlockByNumber")()
+	head, err := s.ethClient.HeaderByNumber(ctx, gethFinalizedBlockNumber)
+	if err != nil {
+		if errors.Is(err, ethereum.NotFound) {
+			return 0, fmt.Errorf("finalised block not found: %w", eth.ErrNotFound)
+		}
+		return 0, fmt.Errorf("getting finalised Ethereum block: %w", err)
+	}
```
rodrodros, line 125: Do we need to get the entire head just to get the block number, can't just we ask for the block number

l1/geth_l1_state_provider.go
```diff
@@ -0,0 +1,262 @@
...
+
+// forwardStateUpdates returns a subscription that decodes each raw event
+// into a StateUpdate and forwards it to sink, until inner errors or the
+// caller unsubscribes. Lifecycle (close, Unsubscribe, error delivery,
+// teardown) is owned by event.NewSubscription; this adds only the type
+// translation and the stall-vs-quit select.
+func forwardStateUpdates(
+	inner event.Subscription,
+	raw <-chan *contract.StarknetLogStateUpdate,
+	sink chan<- *StateUpdate,
```
rodrodros: Similar renames as the previous comment, applying the suggestions there mentioned

l1/l1_state_provider.go
brbrr: Thanks for the explanation. that a helpful insight.

l1/geth_l1_state_provider.go
brbrr: unfortunately there is no easy way to get finalized block number otherwise. There is a getBlockNumber, which returns only latest block, not what we need

rodrodros (approved):

PR #3740, opened 2026-06-18
fix(sn2core): handle candidate(??) transactions in AdaptPreConfirmedWithDelta

thiagodeev:
Fix panic in AdaptPreConfirmedWithDelta when the delta endpoint returns txs with nil state diff / receipt. Such txs are now stored as CandidateTxs instead of dereferenced


rodrodros (approved):

rodrodros:
Fast merge to avoid the current sequencer bug

PR #3749, opened 2026-06-19
feat: validate feeder responses on the client layer

thiagodeev:
This PR makes the feeder client validate gateway responses before handing them off, so malformed or incomplete data from the gateway is caught at the boundary and reported as an error instead of slipping through and causing failures later. 

Validation currently covers the pre_confirmed block and delta-update responses, checking that the fields the rest of the system relies on are present and consistent; other response types are left unvalidated for now and will be addressed in follow-up PRs. Because these guarantees are now enforced at the boundary, duplicate checks downstream in the sn2core adapter are no longer needed and have been removed.


clients/feeder/feeder.go
```diff
@@ -441,16 +448,28 @@ func (c *Client) TransactionStatus(
...
+// and returns itself with no changesin case of success, or an error if validation fails.
+type Validatable[T any] interface {
+	*T
+	Validate() (*T, error)
+}
+
+func doRequest[
+	T any,
+	V Validatable[T],
+](ctx context.Context, client *Client, queryURL string) (*T, error) {
```
rodrodros: Please update the formatting to

```golang
func doRequest[<generics T>](
   param1,
   param2,
) (returnParam) {
}

File: starknet/pre_confirmed_update.go [hunk 2]
```diff
@@ -33,6 +34,45 @@ type PreConfirmedDeltaUpdate struct {
...
+		if txs[i] == (Transaction{}) {
+			return fmt.Errorf("transaction at index %d is empty", i)
+		}
+		if receipts[i] == nil {
+			return fmt.Errorf("receipt at index %d is nil", i)
+		}
+		if stateDiffs[i] == nil {
+			return fmt.Errorf("transaction state diff at index %d is nil", i)
+		}
+	}
```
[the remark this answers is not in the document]
rodrodros: Candidate transactions are no longer part of the response, hence this is not a blocker. Candidate transactions will be deleted in a follow up.

adapters/sn2core/sn2core.go
```diff
@@ -513,22 +513,12 @@ func IsCandidateTx(response *starknet.PreConfirmedBlock, id int) bool {
 	return response.TransactionStateDiffs[id] == nil || response.Receipts[id] == nil
 }
 
+// AdaptPreConfirmedBlock adapts a pre_confirmed block into a pending.PreConfirmed.
+// It assumes that the response was already sanitized, not containing any nil/invalid values.
```
rodrodros: This comment ("It assumes...") is a bit redundant, please remove

adapters/sn2core/sn2core.go
```diff
@@ -513,22 +513,12 @@ func IsCandidateTx(response *starknet.PreConfirmedBlock, id int) bool {
...
-	}
-
-	isInvalidPayloadSizes := len(response.Transactions) != len(response.TransactionStateDiffs) ||
-		len(response.Transactions) != len(response.Receipts)
-	if isInvalidPayloadSizes {
-		return pending.PreConfirmed{}, errors.New(
-			"invalid sizes of transactions, state diffs and receipts",
-		)
-	}
-
```
rodrodros, line 531: Nice catch removing existing validation from the adapters 👍

adapters/sn2core/sn2core.go
```diff
@@ -622,26 +612,18 @@ func AdaptPreConfirmedBlock(
 }
 
 // AdaptPreConfirmedWithDelta returns a new pre_confirmed produced by applying
-// delta on top of current; current is not modified.
+// delta on top of current; current is not modified. It assumes that the delta
+// was already sanitized, not containing any nil/invalid values.
```
rodrodros: Yes, let's not add sanitasion comment everywhere, any one reading the code will find out.

clients/feeder/feeder.go
```diff
@@ -321,15 +321,22 @@ func (c *Client) FeeTokenAddresses(ctx context.Context) (starknet.FeeTokenAddres
 	return *addresses, err
 }
 
+// A string type that implements the Validatable constraint.
+type publicKey string
+
+func (val *publicKey) Validate() (*publicKey, error) {
+	return val, nil
+}
```
rodrodros: It looks like public key deserves his own type, given that it's a type defined in the middle of many functions, and also a big file.

clients/feeder/feeder.go
```diff
@@ -441,16 +448,29 @@ func (c *Client) TransactionStatus(
...
-func doRequest[T any](ctx context.Context, client *Client, queryURL string) (*T, error) {
-	result := new(T)
+// Validatable is a generic constraint satisfied by a pointer type *T whose
+// underlying value type T can validate itself. Implementers provide a
+// Validate method with a pointer receiver that checks the receiver's fields
+// and returns itself with no changes in case of success, or an error if validation fails.
+type Validatable[T any] interface {
+	*T
+	Validate() (*T, error)
+}
```
rodrodros: It seems to me like an interface defined in the middle of a type filled with functionality is counter intuitive for readers of `feeder.go`. Please move it to it's own script

clients/feeder/feeder.go
```diff
@@ -441,16 +448,29 @@ func (c *Client) TransactionStatus(
...
 	}
 	defer body.Close()
 
-	if err = json.NewDecoder(body).Decode(result); err != nil {
+	if err = json.NewDecoder(body).Decode(&result); err != nil {
 		return nil, err
 	}
-	return result, nil
+	return V(&result).Validate()
 }
```
rodrodros: It seems do request is a utility used only by client. In this circumstances with a lot of functions in a file. I think it is better if gets it's own script, maybe grouped together with the `Validated` interface

starknet/pre_confirmed_update.go
```diff
@@ -33,6 +34,45 @@ type PreConfirmedDeltaUpdate struct {
...
+	return validateTxsLength(
+		val.Transactions, val.Receipts, val.TransactionStateDiffs,
+	)
+}
+
+func validateTxsLength(
+	txs []Transaction,
+	receipts []*TransactionReceipt,
+	stateDiffs []*StateDiff,
+) error {
```
rodrodros: This function location seems a bit random, why defined here? It has two callers but it sits between both callers definition. Please, place it either on top or below both of them.

starknet/pre_confirmed_update.go
```diff
@@ -53,6 +93,33 @@ type PreConfirmedBlock struct {
...
+		return errors.New("l1_data_gas_price is required")
+	}
+	return validateTxsLength(
+		pb.Transactions, pb.Receipts, pb.TransactionStateDiffs,
+	)
+}
+
 func (PreConfirmedNoChange) isPreConfirmedUpdate()    {}
 func (PreConfirmedDeltaUpdate) isPreConfirmedUpdate() {}
 func (PreConfirmedBlock) isPreConfirmedUpdate()       {}
```
rodrodros: I think is better if all the methods of the same type to be grouped together

starknet/pre_confirmed_update.go
```diff
@@ -102,3 +169,71 @@ func (e *PreConfirmedUpdateEnvelope) UnmarshalJSON(data []byte) error {
 	}
 	return nil
 }
+
+func (e *PreConfirmedUpdateEnvelope) Validate() (*PreConfirmedUpdateEnvelope, error) {
+	switch update := e.Update.(type) {
+	case PreConfirmedNoChange:
```
rodrodros, line 180: return e, nil here

starknet/pre_confirmed_update.go
```diff
@@ -102,3 +169,71 @@ func (e *PreConfirmedUpdateEnvelope) UnmarshalJSON(data []byte) error {
...
+			return nil, err
+		}
+	case PreConfirmedBlock:
+		if err := update.validate(); err != nil {
+			return nil, err
+		}
+	default:
+		return nil, fmt.Errorf("invalid pre_confirmed update type %T", e.Update)
+	}
+	return e, nil
```
rodrodros: and remove this line

starknet/pre_confirmed_update.go
```diff
@@ -102,3 +169,71 @@ func (e *PreConfirmedUpdateEnvelope) UnmarshalJSON(data []byte) error {
...
+}
+
+// TODO: placeholder for now to avoid compiler errors. A proper validation
+// should be implemented in a follow-up PR.
+
+func (b *Block) Validate() (*Block, error) {
```
rodrodros: Define this type in the right location. Place holder functionality should be correctly located!

rodrodros (commented):
PR is still looks half baked..., I wouldn't consider it ready for review

starknet/pre_confirmed_update.go
thiagodeev: We can't. The compiler requires a return statement for this case.
Do you still want me to put a `return e, nil` in the `case PreConfirmedNoChange:` line?

starknet/pre_confirmed_update.go
```diff
@@ -33,6 +34,45 @@ type PreConfirmedDeltaUpdate struct {
...
+) error {
+	if len(txs) != len(receipts) ||
+		len(txs) != len(stateDiffs) {
+		return errors.New(
+			"transactions, receipts, and tx_state_diffs must have the same length",
+		)
+	}
+
+	for i := range txs {
+		if txs[i] == (Transaction{}) {
```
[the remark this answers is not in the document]
thiagodeev: The validation here is just to make sure the tx is not empty. Receipts and state diffs contain pointers, so we just need to check if it's nil, but since txs contains value structs, the only way to check if the original JSON was 'null' is by comparing against an empty Transaction struct

starknet/pre_confirmed_update.go
```diff
@@ -33,6 +36,21 @@ type PreConfirmedDeltaUpdate struct {
...
 }
 
+func (PreConfirmedDeltaUpdate) isPreConfirmedUpdate() {}
+
+func (val *PreConfirmedDeltaUpdate) validate() error {
+	if val.BlockIdentifier == "" {
+		return errors.New("block_identifier is required")
+	}
+	if len(val.Transactions) == 0 {
+		return errors.New("delta can not have zero transactions")
```
rodrodros: suggestion: "delta has no transaction"

I believe it is easier to read, or if keeping the "required" approach: "delta requires at least one transaction"

starknet/pre_confirmed_update.go
rodrodros: Why not `nil` directly here?

starknet/pre_confirmed_update.go
```diff
@@ -102,3 +171,20 @@ func (e *PreConfirmedUpdateEnvelope) UnmarshalJSON(data []byte) error {
...
+	case PreConfirmedNoChange:
+	case PreConfirmedDeltaUpdate:
+		if err := update.validate(); err != nil {
+			return err
+		}
+	case PreConfirmedBlock:
+		if err := update.validate(); err != nil {
+			return err
+		}
+	default:
```
rodrodros, line 189: Wrap this error, otherwise it will be hard to know if the missing block identifier  is delta update or pre-confirmed block. 

Update it to return fmt.Error("invalid pre-confirmed delta update: %w") 
the same for block

starknet/pre_confirmed_update_test.go
```diff
@@ -165,6 +165,229 @@ func TestPreConfirmedUpdateEnvelope_UnmarshalJSON(t *testing.T) {
...
+// tests can flip a single field to exercise one failure path at a time.
+func validBlock() starknet.PreConfirmedBlock {
+	return starknet.PreConfirmedBlock{
+		BlockIdentifier:       "abc123",
+		Transactions:          []starknet.Transaction{nonEmptyTx()},
+		Receipts:              []*starknet.TransactionReceipt{{}},
+		TransactionStateDiffs: []*starknet.StateDiff{{}},
+		Status:                "PRE_CONFIRMED",
+		Version:               "0.14.0",
+		SequencerAddress:      new(felt.Felt).SetUint64(0xaa),
```
rodrodros: Use the `felt.NewFromUint64` constructor here please. This is something been discussed for ages now

starknet/pre_confirmed_update_test.go
```diff
@@ -165,6 +165,229 @@ func TestPreConfirmedUpdateEnvelope_UnmarshalJSON(t *testing.T) {
...
+		err := env.Validate()
+		require.NoError(t, err)
+	})
+
+	t.Run("invalid Block propagates validate error", func(t *testing.T) {
+		b := validBlock()
+		b.Status = "ACCEPTED_ON_L2"
+		env := &starknet.PreConfirmedUpdateEnvelope{Update: b}
+		err := env.Validate()
+		require.Error(t, err)
```
rodrodros, line 223: when asserting error, check for ErrorContains(...) to know that you get what you're expecting for

starknet/pre_confirmed_update_test.go
```diff
@@ -165,6 +165,229 @@ func TestPreConfirmedUpdateEnvelope_UnmarshalJSON(t *testing.T) {
...
+	})
+
+	t.Run("nil Update hits the default branch", func(t *testing.T) {
+		// A zero-value envelope has a nil Update interface, which falls through
+		// the type switch to the default error branch.
+		env := &starknet.PreConfirmedUpdateEnvelope{}
+		err := env.Validate()
+		require.Error(t, err)
```
rodrodros: similar for these

starknet/pre_confirmed_update_test.go
```diff
@@ -165,6 +165,229 @@ func TestPreConfirmedUpdateEnvelope_UnmarshalJSON(t *testing.T) {
...
+	tests := []struct {
+		name    string
+		mutate  func(*starknet.PreConfirmedBlock)
+		wantErr bool
+	}{
+		{name: "valid", mutate: func(*starknet.PreConfirmedBlock) {}},
+		{
+			name:    "missing block_identifier",
+			mutate:  func(b *starknet.PreConfirmedBlock) { b.BlockIdentifier = "" },
+			wantErr: true,
```
rodrodros, line 256: If you wantErr (meaning it is always `true`) for every case, why even define it in the fixture?

starknet/pre_confirmed_update_test.go
```diff
@@ -165,6 +165,229 @@ func TestPreConfirmedUpdateEnvelope_UnmarshalJSON(t *testing.T) {
...
+		})
+	}
+}
+
+func TestPreConfirmedDeltaUpdate_validate(t *testing.T) {
+	tests := []struct {
+		name    string
+		mutate  func(*starknet.PreConfirmedDeltaUpdate)
+		wantErr bool
+	}{
```
rodrodros, line 338: Same

sync/pending_polling.go
```diff
@@ -211,7 +212,12 @@ func (s *Synchronizer) pollPreConfirmed(
 				ctx, current.Block.Number, current.BlockIdentifier, txCount,
 			)
 			if err != nil {
-				s.logger.Debug("Error while trying to poll pre_confirmed block", zap.Error(err))
+				msg := "Error while trying to poll pre_confirmed block"
```
rodrodros: the variable `msg` should be a constant since  it is not modified. 

Additionally, please don't write error messages such as "error while trying to ..." go directly to the core: "polling pre_confirmed block". I am aware this old text.

clients/feeder/request.go
```diff
@@ -0,0 +1,150 @@
...
+
+	if err = json.NewDecoder(body).Decode(&result); err != nil {
+		return nil, err
+	}
+
+	err = V(&result).Validate()
+	if err != nil {
+		return nil, errors.Join(
+			ErrInvalidFeederResponse,
+			fmt.Errorf("error while validating response from %s: %w", queryURL, err),
```
rodrodros: The word `error` is redundant when you're returning an error. This can be reduced to "validating response...",  a good string example: "querying \"%s\": %w"

sync/pending_polling.go
rodrodros: Also, please include the block number in the logs, it seems an oversight that it wasn't there alread

clients/feeder/request.go
```diff
@@ -0,0 +1,150 @@
...
+	base.Path += endpoint
+
+	params := url.Values{}
+	for k, v := range args {
+		params.Add(k, v)
+	}
+	base.RawQuery = params.Encode()
+
+	return base.String()
+}
```
rodrodros: Re-reading this method, this seems kinda of unnecessary that we parse the base (which never changes on every iteration). 

Refactor this to:
1. Make `buildQueryString` a function and receive the parsed sequencer path as an argument.
2. The `c.URl` should be stored already parsed at the client level

clients/feeder/request.go
```diff
@@ -0,0 +1,150 @@
...
+
+// get performs a "GET" http request with the given URL and returns the response body
+func (c *Client) get(ctx context.Context, queryURL string) (io.ReadCloser, error) {
+	var res *http.Response
+	var err error
+	wait := time.Duration(0)
+	for range c.maxRetries + 1 {
```
rodrodros: It is confusing that `Client` has it's method `get` define outside where the type itself was defined. Please move it back.

starknet/pre_confirmed_update.go
```diff
@@ -102,3 +171,20 @@ func (e *PreConfirmedUpdateEnvelope) UnmarshalJSON(data []byte) error {
...
+		return fmt.Errorf("invalid pre_confirmed update type %T", e.Update)
+	}
+	return nil
```
rodrodros, line 192: IF every branch of the case returns, then this line no longer becomes necessary, which also adds clarity that when it is `PreConfirmedNoChange` there is no validation to do

starknet/pre_confirmed_update.go
rodrodros: false, if you return on `case PreConfirmedNoChange` the compiler stops requiring anything

starknet/pre_confirmed_update_test.go
rodrodros: I made a mistake, there is one single case that is valid,

starknet/pre_confirmed_update_test.go
rodrodros: And here too

starknet/pre_confirmed_update.go
rodrodros: My mistake, current approach is good

starknet/pre_confirmed_update.go
rodrodros: It is my mistake

sync/pending_polling.go
thiagodeev: Done. I changed the normal branch to log block number.
The `if errors.Is(err, feeder.ErrInvalidFeederResponse) {` branch already includes the block number since the error chain contains the entire query URL on it, which includes the block number.

clients/feeder/feeder.go
```diff
@@ -29,7 +30,7 @@ var ErrDeprecatedCompiledClass = errors.New("deprecated compiled class")
 type Backoff func(wait time.Duration) time.Duration
 
 type Client struct {
-	url        string
+	url        *url.URL
```
rodrodros, line 33: Why a pointer instead of a value?

clients/feeder/feeder.go
```diff
@@ -145,8 +146,13 @@ func NopBackoff(d time.Duration) time.Duration {
 
 func NewClient(clientURL string) *Client {
 	defaultTimeouts := getDefaultFixedTimeouts()
+	base, err := url.Parse(clientURL)
+	if err != nil {
+		panic("malformed feeder base URL")
+	}
```
rodrodros: clientURL is suplied by the end user, you cannot panic here. It is bad UX.
You have to either change the constructor to return an error or change the input parameter to be an URL

clients/feeder/feeder.go
```diff
@@ -160,7 +166,7 @@ func NewClient(clientURL string) *Client {
 }
 
 // get performs a "GET" http request with the given URL and returns the response body
-func (c *Client) get(ctx context.Context, queryURL string) (io.ReadCloser, error) {
+func (c *Client) get(ctx context.Context, queryURL *url.URL) (io.ReadCloser, error) {
```
rodrodros, line 165: why pass the entire queryURL *url.URL when you only need a string?

clients/feeder/request.go
```diff
@@ -0,0 +1,60 @@
...
+type Validatable[T any] interface {
+	*T
+	Validate() error
+}
+
+func doRequest[T any, V Validatable[T]](
+	ctx context.Context,
+	client *Client,
+	fullURL *url.URL,
+) (*T, error) {
```
rodrodros, line 26: Why do we return `*T` instead of `T`?

starknet/pre_confirmed_update.go
```diff
@@ -53,9 +71,60 @@ type PreConfirmedBlock struct {
...
+	)
+}
+
+func validateTxsLength(
+	txs []Transaction,
+	receipts []*TransactionReceipt,
+	stateDiffs []*StateDiff,
+) error {
+	if len(txs) != len(receipts) ||
+		len(txs) != len(stateDiffs) {
```
rodrodros: why a multi line if? That goes against both Go and Juno styles

sync/pending_polling.go
```diff
@@ -211,7 +212,12 @@ func (s *Synchronizer) pollPreConfirmed(
 				ctx, current.Block.Number, current.BlockIdentifier, txCount,
 			)
 			if err != nil {
-				s.logger.Debug("Error while trying to poll pre_confirmed block", zap.Error(err))
+				const msg = "polling pre-confirmed block"
+				if errors.Is(err, feeder.ErrInvalidFeederResponse) {
+					s.logger.Error(msg, zap.Error(err))
```
rodrodros: Why block number is not logged here as well?

starknet/pre_confirmed_update.go
```diff
@@ -53,9 +71,60 @@ type PreConfirmedBlock struct {
...
+		return errors.New(
+			"transactions, receipts, and tx_state_diffs must have the same length",
+		)
```
rodrodros: Please add the length information, to know which of them was the different one

cmd/juno/juno.go
```diff
@@ -665,3 +676,18 @@ func NewCmd(config *node.Config, run func(*cobra.Command, []string) error) *cobr
...
 }
+
+func validateHTTPURL(rawURL string) error {
+	// rejects relative / scheme-less strings
+	u, err := url.ParseRequestURI(rawURL)
+	if err != nil {
+		return err
+	}
+	if u.Scheme != "http" && u.Scheme != "https" {
+		return fmt.Errorf("URL must use http or https scheme")
```
rodrodros: using `fmt.Error` without arguments. You should provide the Scheme that was passed:

cmd/juno/juno.go
```diff
@@ -665,3 +676,18 @@ func NewCmd(config *node.Config, run func(*cobra.Command, []string) error) *cobr
...
+	}
+	if u.Host == "" {
+		return fmt.Errorf("URL must have a host")
```
rodrodros: errors.New

cmd/juno/juno.go
```diff
@@ -430,10 +431,20 @@ func NewCmd(config *node.Config, run func(*cobra.Command, []string) error) *cobr
...
+				return fmt.Errorf("invalid gateway URL %s: %w", gatewayURL, err)
+			}
+
 			config.Network = networks.Network{
 				Name:                v.GetString(cnNameF),
-				FeederURL:           v.GetString(cnFeederURLF),
-				GatewayURL:          v.GetString(cnGatewayURLF),
+				FeederURL:           feederURL,
+				GatewayURL:          gatewayURL,
 				L1ChainID:           l1ChainID,
```
rodrodros, line 449: Should the configuration hold the strings or the actual URLs? Shouldn't it hold the urls?

node/node.go
```diff
@@ -358,7 +358,12 @@ func New(cfg *Config, version string, logLevel *log.Level) (*Node, error) {
 		if err != nil {
 			return nil, fmt.Errorf("invalid gateway timeouts: %w", err)
 		}
-		client = feeder.NewClient(cfg.Network.FeederURL).
+
+		feederURL, err := url.Parse(cfg.Network.FeederURL)
+		if err != nil {
+			return nil, fmt.Errorf("invalid feeder URL %s: %w", cfg.Network.FeederURL, err)
+		}
+		client = feeder.NewClient(feederURL).
```
rodrodros, line 366: Why do we have to parse them again, weren't the already parse at the cmd level.
The only issue is that they were parsed there but only stored in string form. 

If it is too big of a change, it can be an immediate follow up cleaning it. Otherwise, we will forget about it.

cmd/juno/juno.go
thiagodeev: Oh, sorry. I missed these ones

cmd/juno/juno.go
thiagodeev: New issue created to do it in a follow up PR
https://github.com/NethermindEth/juno/issues/3761

node/node.go
thiagodeev: New issue created to do it in a follow up PR
https://github.com/NethermindEth/juno/issues/3761

rodrodros (approved):
LGTM

PR #3750, opened 2026-06-19
feat(utils/throttler): throttler revamped

rodrodros:
- Throttler now has it's own sub package inside utils
- Swap setter for option pattern, forcing optional settings to throttler startup
- Max queue length flag type changed to uint64 (from int) to avoid negative values
- Add context object so call to resources and actions waiting on the queue could both be interrupted via `cancel()`
- Fix an issue where a queue with size 0 would cause the resource to not execute anything

### DIFF
--- a/node/throttled_vm.go
+++ b/node/throttled_vm.go
@@ -15,7 +15,7 @@ type ThrottledVM struct {
 
 func NewThrottledVM(res vm.VM, concurrenyBudget uint, maxQueueLen int32) *ThrottledVM {
 	return &ThrottledVM{
-		Throttler: throttler.NewThrottler(concurrenyBudget, &res).WithMaxQueueLen(maxQueueLen),
+		Throttler: throttler.NewThrottler(concurrenyBudget, &res, throttler.WithMaxQueueLen(maxQueueLen)),
 	}
 }
 
--- a/node/throttled_vm.go
+++ b/node/throttled_vm.go
@@ -1,6 +1,8 @@
 package node
 
 import (
+	"context"
+
 	"github.com/NethermindEth/juno/core"
 	"github.com/NethermindEth/juno/core/felt"
 	"github.com/NethermindEth/juno/utils/throttler"
@@ -28,7 +30,8 @@ func (tvm *ThrottledVM) Call(
 	errStack, returnStateDiff bool,
 ) (vm.CallResult, error) {
 	ret := vm.CallResult{}
-	return ret, tvm.Do(func(vm *vm.VM) error {
+	// vm.VM carries no ctx; queued VM calls aren't cancellable yet.
+	return ret, tvm.Do(context.Background(), func(vm *vm.VM) error {
 		var err error
 		ret, err = (*vm).Call(
 			callInfo,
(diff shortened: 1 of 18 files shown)


node/throttled_vm.go
```diff
@@ -28,7 +30,8 @@ func (tvm *ThrottledVM) Call(
 	errStack, returnStateDiff bool,
 ) (vm.CallResult, error) {
 	ret := vm.CallResult{}
-	return ret, tvm.Do(func(vm *vm.VM) error {
+	// vm.VM carries no ctx; queued VM calls aren't cancellable yet.
+	return ret, tvm.Do(context.Background(), func(vm *vm.VM) error {
```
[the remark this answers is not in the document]
rodrodros: This is part of longer spanning work, unless we spawn VMs as separate process (like we do with the compiler) it is impossible to stop VM execution hence we cannot do it now.

utils/throttler/throttler.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+// Do lets caller acquire the resource within the context of a callback
+func (t *Throttler[T]) Do(ctx context.Context, doer func(resource *T) error) error {
+	if err := ctx.Err(); err != nil {
+		return err // already cancelled, don't even enter the queue
+	}
+
+	activeReqs := t.currentRequests.Add(1)
+	defer t.currentRequests.Add(^uint64(0)) // decrement by 1
```
brbrr, line 69: that's impressively complicated way to decrement a value 😬, but I guess that's idiomatic.

utils/throttler/throttler.go
```diff
@@ -0,0 +1,94 @@
...
+// requests for resource `T`. See [throttler.Option] for other options.
+func NewThrottler[T any](maxConcurrentReqs uint, resource *T, opts ...Option) *Throttler[T] {
+	o := options{
+		maxQueueLen: 1024,
+	}
+	for _, opt := range opts {
+		opt(&o)
+	}
+
+	// guard against overlfow
```
brbrr:
```suggestion
	// guard against overflow
```

utils/throttler/throttler.go
```diff
@@ -0,0 +1,94 @@
...
+	maxRequests := o.maxQueueLen + uint64(maxConcurrentReqs)
+	if maxRequests < o.maxQueueLen {
+		maxRequests = math.MaxUint64
+	}
+
+	return &Throttler[T]{
+		resource: resource,
+		sem:      make(chan struct{}, maxConcurrentReqs),
```
brbrr, line 55: `maxConcurrentReqs == 0` will make the semaphore unbounded. Probably, we should have a meaningful limit here as well.

brbrr (approved):
LGTM. left a typo correction, and a note about unbounded concurrency case

utils/throttler/throttler.go
rodrodros: Yes, the other way is using integers but then you've to add overflow protections and protect against negative values. I think this is the right trade off since our types are never meant to be negative.

utils/throttler/throttler.go
rodrodros: I don't know what you mean by unbounded, it will block until it is received , which doesn't happens until a few lines below.

The real behaviour is that max vms=0 with a non zero queue (say size `n`) will cause `n` requests to be queued but none to execute, which seems kind like how it should behave. I don't know if we should guard against this

PR #3753, opened 2026-06-22
feat(jsonrpc): limit concurrent http requests

rodrodros:


jsonrpc/http.go
```diff
@@ -60,15 +68,25 @@ func (h *HTTP) ServeHTTP(writer http.ResponseWriter, req *http.Request) {
...
 		defer cancel()
 	}
+
+	if h.gate != nil {
+		if err := h.gate.Acquire(ctx); err != nil {
+			writer.Header().Set("Retry-After", "1")
+			http.Error(writer, "Too many requests", http.StatusServiceUnavailable)
+			return
+		}
+		defer h.gate.Release()
```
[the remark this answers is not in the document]
brbrr: seems like reasonable observation

jsonrpc/http.go
```diff
@@ -60,15 +68,25 @@ func (h *HTTP) ServeHTTP(writer http.ResponseWriter, req *http.Request) {
...
 	if h.requestTimeout > 0 {
 		var cancel context.CancelFunc
 		ctx, cancel = context.WithTimeout(ctx, h.requestTimeout)
 		defer cancel()
 	}
+
+	if h.gate != nil {
+		if err := h.gate.Acquire(ctx); err != nil {
+			writer.Header().Set("Retry-After", "1")
+			http.Error(writer, "Too many requests", http.StatusServiceUnavailable)
```
brbrr: Would it make sense to log an overflow? It would be useful to indicate that the node exceeds its limits.

jsonrpc/gate.go
```diff
@@ -0,0 +1,85 @@
...
+// queues up to maxQueue more before rejecting.
+func NewGate(maxConcurrent uint, maxQueue uint64) *Gate {
+	// Guard against overflow, mirroring throttler.NewThrottler.
+	maxRequests := maxQueue + uint64(maxConcurrent)
+	if maxRequests < maxQueue {
+		maxRequests = math.MaxUint64
+	}
+
+	return &Gate{
+		sem:         make(chan struct{}, maxConcurrent),
```
brbrr, line 42: `maxConcurrent == 0` will make the semaphore unbounded.

jsonrpc/http.go
```diff
@@ -60,15 +68,25 @@ func (h *HTTP) ServeHTTP(writer http.ResponseWriter, req *http.Request) {
...
+			return
+		}
+		defer h.gate.Release()
+	}
+
+	req.Body = http.MaxBytesReader(writer, req.Body, MaxRequestBodySize)
+	h.listener.OnNewRequest("any")
```
brbrr: With this line moved, we would count only successful requests. It should be placed back before the gate check.

jsonrpc/http.go
brbrr: "Too many requests" is the message of 429(client-related) HTTP error.  `StatusServiceUnavailable` is 503. 

I'd suggest changing the message, as it's clearly a server error, meaning the server can't handle the load, rather then client sends too many requests. Something along the lines of "Server overloaded"/"Processing queue is saturated".

jsonrpc/http.go
```diff
@@ -60,15 +68,25 @@ func (h *HTTP) ServeHTTP(writer http.ResponseWriter, req *http.Request) {
...
-
 	ctx := req.Context()
 	if h.requestTimeout > 0 {
 		var cancel context.CancelFunc
 		ctx, cancel = context.WithTimeout(ctx, h.requestTimeout)
 		defer cancel()
 	}
+
+	if h.gate != nil {
+		if err := h.gate.Acquire(ctx); err != nil {
```
brbrr, line 93: In case `Acquire` blocks, the context with a timeout that was created above will be affected. So the request will be terminated while waiting for a slot. Just want to confirm that this is expected behavior. 

A potentially valid use case would be when a user runs a node on a relatively slow machine and does a bunch of queries on a cron schedule, for data collection/processing etc.

jsonrpc/gate.go
```diff
@@ -0,0 +1,85 @@
...
+func (g *Gate) Running() int { return len(g.sem) }
+
+// Queued returns the number of requests waiting for a processing slot.
+func (g *Gate) Queued() int { return int(g.activeRequests.Load()) - len(g.sem) }
+
+func (g *Gate) increaseActiveReq() uint64 {
+	return g.activeRequests.Add(1)
+}
+
+func (g *Gate) decreaseActivReq() {
```
brbrr:
```suggestion
func (g *Gate) decreaseActiveReq() {
```

brbrr (approved):
Left a bunch of comments. They are not blocking IMO. Otherwise LGTM.

jsonrpc/http.go
rodrodros: Do you think it should be exposed as a metric too?

jsonrpc/http.go
rodrodros: Sorry, I am not following. What do you mean by the request will be terminated by waiting for a slot if context get cancelled? Isn't that how is it supposed to work on context cancellation? 

> potentially valid use case would be when a user runs a node on a relatively slow machine and does a bunch of queries on a cron schedule, for data collection/processing etc.

Not sure I follow, how does that affect them, they control their Juno's timeout queue and stuff 🤔

jsonrpc/gate.go
rodrodros: that's not how channel works, that will make it blocking

rodrodros:
@brbrr for your follow up reviews could you please group your comments by line, commenting over the same lines, in three different comments is kind of confusing.

jsonrpc/http.go
brbrr: > Isn't that how is it supposed to work on context cancellation?

Right. The above comment is an observation (that might be the design decision) that I made, and I wasn't sure about. 

The "problem" is that, conceptually, we are changing what `requestTimeout` means. Before, it was "how long the request is allowed to be processing", now, it's "how long it may wait for a gate semaphore + actual request processing". That's probably fine and expected, and yes, that's all could be configured and tweaked.

jsonrpc/http.go
brbrr: It seems like a good idea - will complement the added metrics nicely, and might be useful when tweaking the configuration.

jsonrpc/http.go
rodrodros: I see, valid one, let me think about it

rodrodros: After a bit of thinking, queue time will also count towards req timeout

jsonrpc/http.go
brbrr: You marked it as resolved, but I'm not sure what's your decision was on my comment. Can you elaborate?

jsonrpc/http.go
```diff
@@ -60,15 +80,30 @@ func (h *HTTP) ServeHTTP(writer http.ResponseWriter, req *http.Request) {
...
+			if errors.Is(err, ErrServerBusy) {
+				h.logServerBusy()
+				writer.Header().Set("Retry-After", "1")
+				http.Error(writer, "Server is busy", http.StatusServiceUnavailable)
+				return
+			}
+			return
```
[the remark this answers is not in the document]
brbrr: I think it might be useful to add a dedicated log for timeout exceeded case, since in this case user won't know why their request was dropped. Not a problem now with big limits, but might bite us when we decide to lower it.

brbrr (approved):
Left come comments, but they are not blocking. Feel free to merge if you think they don't worth to address

jsonrpc/http.go
rodrodros: Sorry, I agree with you. Moving it back before the gate 👍

PR #3762, opened 2026-06-23
docs: add pruned node to snapshots and missing pruning and compilation flags

brbrr:


docs/versioned_docs/version-0.16.0/snapshots.md
```diff
@@ -254,6 +300,25 @@ docker run -d \
   --eth-node <YOUR-ETH-NODE>
 ```
 
+</TabItem>
+<TabItem value="mainnet-pruned" label="Mainnet-Pruned">
+
+A pruned snapshot is a Mainnet database with historical state and block data removed. Run it with `--prune-mode` enabled so the database stays pruned as the node keeps syncing; without it the node would begin retaining full history again and the database would grow back to the full Mainnet size.
```
rodrodros: This is false, remove this text

rodrodros (approved):

Issue #3114, opened 2025-09-16
[Tendermint] Use proper WAL implementation instead of pebble


infrmtcs:
We can use the `wal` package from `pebble`.

PR #3763, opened 2026-06-24
feat(consensus): persist standalone WAL with crash-safe replay

ongyimeng:
## Summary

Replaces the consensus WAL, previously stored as generic CBOR-encoded values in the shared Pebble KV store, with a dedicated append-only WAL owned by consensus. The new WAL uses typed records, segment files, pruning, and crash-safe replay.

## Why

Consensus relies on WAL replay to recover after a crash mid-height. As generic KV values, that recovery path was harder to inspect and debug, and was coupled to the shared KV store. A standalone WAL gives consensus an explicit durability boundary with clearer replay and failure handling.

Issue #3114 

## What changed

- New consensus WAL:
  - typed record codec
  - append-only segments with rotation and tail repair
  - persisted prune watermarks
  - live-entry indexes
- Replaced the Pebble KV-backed WAL and replay path.
- Fixed `Start` entry replay. It already existed as an entry type, but the old generic CBOR decoding couldn't reconstruct it on replay.
- Improved error handling across WAL writes/flushes, commit persistence, and replay.

## Performance

Headline comparison is `BenchmarkTendermintWALStoreSustainedHeightLatency`, which is closest to the consensus write/delete flow: 1000 heights, real cleanup frequency, `count=5`.

Medians:

| metric | old WAL (KV) | new WAL | result |
|---|---:|---:|---|
| p50 | 155ms | 153ms | tie |
| p95 | 192ms | 177ms | ~7.8% faster |
| p99 | 239ms | 203ms | ~15.1% faster |
| max | 298ms | 299ms | tie |
| B/op | 1.77 MB | 1.47 MB | ~17% less |
| allocs/op | 41.1k | 28.2k | ~31% fewer |

Same latency at p50/max, better p95/p99, fewer allocations.

`B/op` and `allocs/op` are per 1000-height run, since Go counts the loop as one op under `-benchtime=1x`.

Note: This benchmark uses 1,000 heights, which fits within Pebble’s in-memory write path. It does not exercise the old KV-backed WAL’s eventual flush, compaction, or tombstone-reclamation costs. There is no steady-state latency regression at this scale, but it is not representative of recurring background costs in a long-running node which is opaque and state dependent.

## Testing

- Codec, replay, WAL store/writer, pruning, and failure-injection tests.
- Restart test: node recovers via WAL replay with peers stopped and no committed block.
- Benchmarks for entry durability, full-height writes, cleanup, replay, and sustained latency.

The remaining coverage gaps are mostly rare internal error paths. The main WAL recovery flows are already covered, so extra tests would add little value.


consensus/db/wal_writer.go
```diff
@@ -0,0 +1,251 @@
...
+
+func repairWALTail(walPath string, syncedOffset int64) error {
+	file, err := os.OpenFile(walPath, os.O_RDWR, 0)
+	if errors.Is(err, os.ErrNotExist) {
+		return nil
+	}
+	if err != nil {
+		return fmt.Errorf("Flush: open WAL for repair: %w", err)
+	}
+	defer file.Close()
```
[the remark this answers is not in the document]
ongyimeng: resolved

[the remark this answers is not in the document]
ongyimeng: resolved

db/memory/db.go
```diff
@@ -89,9 +93,32 @@ func (d *Database) Close() error {
...
+		return ""
+	}
+	if d.path != "" {
+		return d.path
+	}
+
+	path, err := os.MkdirTemp("", "juno-memory-db-*")
+	if err != nil {
+		panic(fmt.Errorf("create memory DB temp dir: %w", err))
+	}
```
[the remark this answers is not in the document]
ongyimeng: resolved

consensus/walstore/codec.go
```diff
@@ -0,0 +1,465 @@
...
+// Encode Pebble's batchrepr format directly so replay can parse WAL records
+// with ReadHeader and Read.
+func encodeBatch[V types.Hashable[H], H types.Hash, A types.Addr](
+	records []walRecordEnvelope[V, H, A],
+	seqNum uint64,
+	encodedBatch []byte,
+) ([]byte, error) {
+	if len(records) > math.MaxUint32 {
+		return nil, errors.New("encodeBatch: too many WAL records in single flush")
+	}
```
rodrodros, line 29: If this is a feasible limit? Can this happen, what happens when `len(records)` exceeds it? Does the program comes to a stop or does it try again with a smaller batch. 

All in all, I think since this is an internal function that has a public caller, the safety guard should be on the caller, making sure to split the records accordingly. 

Note though, that I don't think it is possible to ever get pow(2,32) records in memory, or am I wrong?

consensus/db/codec.go
```diff
@@ -0,0 +1,465 @@
...
+	}
+
+	const (
+		kindBytes             = 1
+		keyLenBytes           = 1
+		keyBytes              = 4
+		estimatedPayloadBytes = 128
+	)
+	estimatedRecordBytes := kindBytes + keyLenBytes + keyBytes +
+		valueLenBytes + estimatedPayloadBytes
```
rodrodros: `estimatedRecordBytes` looks like it could be a constant too

consensus/walstore/codec.go
```diff
@@ -0,0 +1,465 @@
...
+	// Match Pebble batchrepr's little-endian header encoding within this WAL record.
+	return binary.LittleEndian.AppendUint64(payload, value)
+}
+
+func appendInt64(payload []byte, value int64) []byte {
+	return appendUint64(payload, uint64(value))
+}
+
+func appendUint64Array[T ~[4]uint64](payload []byte, value T) []byte {
+	array := [4]uint64(value)
```
rodrodros: This performs a copy of the value, maybe you can make it a pointer, by receiving `*T`

consensus/walstore/codec.go
```diff
@@ -0,0 +1,465 @@
...
+			return types.Vote[H, A]{}, err
+		}
+		hash := H(id)
+		vote.ID = &hash
+	}
+	return vote, nil
+}
+
+// Timeout payload: step, height, round.
+func decodeTimeoutRecord(decoder *walRecordDecoder) (*wal.Timeout, error) {
```
rodrodros, line 276: Why do you return a timeout by reference?

consensus/db/codec.go
```diff
@@ -0,0 +1,465 @@
...
+		// Each batchrepr entry starts with the kind byte and the encoded key.
+		encodedBatch = append(encodedBatch, byte(pebble.InternalKeyKindSet), keyBytes)
+		encodedBatch = binary.BigEndian.AppendUint32(encodedBatch, uint32(i))
+
+		// Reserve space for the value length prefix and backfill it after writing
+		// the record payload.
+		valueLenIndex := len(encodedBatch)
+		encodedBatch = append(encodedBatch, 0, 0, 0, 0, 0)
+		valueIndex := len(encodedBatch)
+
```
rodrodros: Performance nitpick: Append is generally slower than indexing and adding values. Since you already know the max size then it is better to do index work and it will be slighter faster.

Appends, even if not exceeding max capacity, it adds an overhead to it

rodrodros (commented):
Some initial comments

consensus/db/codec.go
```diff
@@ -0,0 +1,465 @@
...
+		}
+		putFixedUvarint32(encodedBatch[valueLenIndex:valueIndex], uint32(valueLen))
+	}
+
+	return encodedBatch, nil
+}
+
+func appendWALRecordPayload[V types.Hashable[H], H types.Hash, A types.Addr](
+	payload []byte,
+	record walRecordEnvelope[V, H, A],
```
rodrodros: `walRecordEnvelope` seems like a big type, why not pass it as a reference to avoid the copy

consensus/db/codec.go
```diff
@@ -0,0 +1,465 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("encodeBatch: encode WAL envelope: %w", err)
+		}
+
+		// batchrepr stores value lengths as uvarints, so fill that prefix once
+		// the final payload size is known.
+		valueLen := len(encodedBatch) - valueIndex
+		if valueLen > math.MaxUint32 {
+			return nil, errors.New("encodeBatch: WAL record payload too large")
+		}
```
rodrodros: It seems to me you can know the maximum size `encodedBatch` can have, so to check the value on every loop seems unnecessary to me

consensus/walstore/codec.go
```diff
@@ -0,0 +1,465 @@
...
+	}
+
+	return encodedBatch, nil
+}
+
+func appendWALRecordPayload[V types.Hashable[H], H types.Hash, A types.Addr](
+	payload []byte,
+	record walRecordEnvelope[V, H, A],
+) ([]byte, error) {
+	payload = append(payload, byte(record.Kind))
```
rodrodros, line 88: Similar comment with the `append`. If you pass an index and treat that as your `0` index it will improve performance

consensus/db/codec.go
```diff
@@ -0,0 +1,465 @@
...
+	switch record.Kind {
+	case walRecordEntry:
+		payload = append(payload, byte(record.EntryKind))
+		switch record.EntryKind {
+		case walEntryStart:
+			payload = appendUint64(payload, uint64(record.StartHeight))
+		case walEntryProposal:
+			if record.ProposalEntry == nil {
+				return nil, errors.New("missing proposal WAL entry payload")
+			}
```
rodrodros: Why would it be nil. Isn't all the data sent through here in a controlled environment, that is, is it not impossible to `record.ProposalEntry` being `nil` here? 

If that's the case then this `if` is unnecessary

consensus/walstore/codec.go
```diff
@@ -0,0 +1,465 @@
...
+		switch record.EntryKind {
+		case walEntryStart:
+			payload = appendUint64(payload, uint64(record.StartHeight))
+		case walEntryProposal:
+			if record.ProposalEntry == nil {
+				return nil, errors.New("missing proposal WAL entry payload")
+			}
+			proposalMessage := (*types.Proposal[V, H, A])(record.ProposalEntry)
+			payload = appendMessageHeader(payload, proposalMessage.MessageHeader)
+			payload = appendInt64(payload, int64(proposalMessage.ValidRound))
```
rodrodros, line 98: append is so expensive 🫠, it is being used way more than I thought

consensus/db/codec.go
```diff
@@ -0,0 +1,465 @@
...
+			proposalMessage := (*types.Proposal[V, H, A])(record.ProposalEntry)
+			payload = appendMessageHeader(payload, proposalMessage.MessageHeader)
+			payload = appendInt64(payload, int64(proposalMessage.ValidRound))
+			if proposalMessage.Value == nil {
+				payload = append(payload, 0)
+			} else {
+				payload = append(payload, 1)
+				var err error
+				payload, err = appendValue(payload, *proposalMessage.Value)
```
rodrodros: Why not pass the Value by reference and avoid the copy?

consensus/db/codec.go
```diff
@@ -0,0 +1,465 @@
...
+				if err != nil {
+					return nil, err
+				}
+			}
+		case walEntryPrevote:
+			if record.PrevoteEntry == nil {
+				return nil, errors.New("missing prevote WAL entry payload")
```
rodrodros: Unnecessary nil guard defense, safety guards are established when they types are created, not when they are used.

consensus/db/codec.go
```diff
@@ -0,0 +1,465 @@
...
+			}
+			payload = appendVotePayload(payload, (*types.Vote[H, A])(record.PrevoteEntry))
+		case walEntryPrecommit:
+			if record.PrecommitEntry == nil {
+				return nil, errors.New("missing precommit WAL entry payload")
```
rodrodros: mark

consensus/db/codec.go
```diff
@@ -0,0 +1,465 @@
...
+			}
+			payload = appendVotePayload(payload, (*types.Vote[H, A])(record.PrecommitEntry))
+		case walEntryTimeout:
+			if record.TimeoutEntry == nil {
```
rodrodros: mark

consensus/walstore/codec.go
```diff
@@ -0,0 +1,465 @@
...
+
+func putFixedUvarint32(buf []byte, value uint32) {
+	const fixedUvarintShift = 7
+
+	buf[0] = byte(value) | 0x80
+	for i := 1; i < valueLenBytes-1; i++ {
+		buf[i] = byte(value>>(fixedUvarintShift*i)) | 0x80
+	}
+	buf[valueLenBytes-1] = byte(value >> (fixedUvarintShift * (valueLenBytes - 1)))
+}
```
rodrodros, line 137: Why fixed uvarint, instead of one with variable size?

consensus/db/codec.go
```diff
@@ -0,0 +1,465 @@
...
+		if err != nil {
+			return walRecordEnvelope[V, H, A]{}, err
+		}
+		record.Height = height
+	default:
+		return walRecordEnvelope[V, H, A]{}, fmt.Errorf("unknown WAL record kind %d", record.Kind)
+	}
+	if decoder.remaining() != 0 {
+		return walRecordEnvelope[V, H, A]{}, fmt.Errorf(
+			"trailing WAL record bytes: %d",
```
rodrodros: The error doesn't make clear what the problem is?  Perhaps you want to put something like "unexpected remaining WAL record bytes: %d" ?

consensus/walstore/codec.go
rodrodros, line 1: Codec holds both the coding and decoding logic wouldn't it makes sense having a `codec` package, with the coding and decoding logic properly separated, or having a script per coding / decoding step paired together?

rodrodros, line 1: Not relevant to this file but the whole folder.

Why is the package called `db` when it is clearly designed to be a WAL log, shouldn't it be renamed `wal` or `waldb`?

consensus/db/init_test.go
```diff
@@ -1,4 +1,4 @@
-package db
+package db_test
```
rodrodros: 👌

sync/data_source.go
rodrodros, line 1: What's the rationale behind this changes?

rodrodros (commented):
Initial review

consensus/walstore/codec.go
ongyimeng: The limit comes from using Pebble’s batchrepr format. The intent is to make that limit explicit as defensive documentation, not as an expected runtime path, since we only accumulate a handful of WAL records before hitting the flush boundary.

I don’t think this should move to the caller. It’s an encoding implementation detail owned by the codec, and moving it out would leak that detail.

consensus/walstore/codec.go
ongyimeng: Switching to `*T` makes the API less clear because it introduces `nil` handling, even though the helper doesn’t mutate or store the value.

If the concern is performance, I checked escape analysis with the compiler output and passing `value T` does not introduce a heap allocation here.

consensus/walstore/codec.go
ongyimeng: I returned `*wal.Timeout` because this helper decodes the final WAL payload type stored in `walRecordEnvelope.TimeoutEntry`.

This matches `decodeProposalRecord`, which also returns the final pointer payload. `decodeVotePayload` returns a value because it’s shared by prevote/precommit and the caller still needs to wrap it in the concrete WAL entry type.

consensus/db/codec.go
ongyimeng: I measured this with a microbenchmark using the common `encodeBatch` sizes from the WAL flow which is 1-2 records.

With `-count=10`, indexing was basically tied for the 2-record case and only ~0.7 ns faster for the 1-record case, with 0 allocs either way. That is negligible compared to the real WAL flush path, which is ~20 ms/op.

This feels mostly like a readability preference, and I find the append version clearer than the manual-offset version. Let me know if you still prefer indexing.

consensus/walstore/codec.go
rodrodros: Because this is internal API you can trust yourself and know that you will never call it knowing it is nil, by passing a reference, you saved from copying 4 words to just copying 1 word per call

consensus/db/codec.go
rodrodros: We can leave it as is. I personally don't think it is more readable since it is not clear enough what you want to do. For example:

```golang
	encodedBatch = append(encodedBatch, 0, 0, 0, 0, 0)
```

Looks very weird, whilst the intention is reserving extra 5 bytes. 

Additionally, if tomorrow the capacity requirements change we will end up with `append` silently allocating and not realizing it. 

None of this are big deal, I do think the indexing one is preferable and not leave it up to the compiler to optimize it for you. 

Leave if you think is good 👍

consensus/walstore/codec.go
ongyimeng: `batchrepr` uses a variable-size uvarint for the value length. I used a fixed 5-byte field to allow backfilling after encoding the payload, but I agree that we can switch to the variable-size form.

consensus/walstore/codec.go
ongyimeng: Codec is for the private on-disk WAL format and uses unexported details like `walRecordEnvelope` and the record/entry kind enums. Moving it to a separate package would force those details to be exported or leaked across package boundaries.

If readability is the concern, splitting encode/decode into separate files within the same package sounds fine, but I would avoid a separate codec package. Would you prefer that, or kept as it is?

ongyimeng: Agree, I’ll rename it to `walstore` to better match its intent now that it’s no longer PebbleDB-backed.

I’d avoid keeping `db` in the name since it no longer uses one, and avoid `wal` since the existing WAL types represent the domain entries like `Start` and `Proposal`. This package is focused on the WAL persistence store layer.

sync/data_source.go
ongyimeng: It was to make persistence failures visible to the consensus commit path. Previously, `Persisted` only signaled completion, so the listener couldn’t tell success from failure.

Now the sync side sends the error, or `nil` on success. That lets the commit listener avoid treating failed persistence as a successful commit, so it does not run post-commit hooks or finalize proposals when the block was not durably stored.

consensus/walstore/codec.go
rodrodros: Sorry, I don't follow, what do you mean by allowing backfilling?

consensus/walstore/wal_index.go
```diff
@@ -0,0 +1,101 @@
...
+	if s.first == 0 {
+		s.first = walNum
+		return true
+	}
+	if s.first == walNum {
+		return false
+	}
+	if slices.Contains(s.rest, walNum) {
+		return false
+	}
```
rodrodros: nit: these `if`s could be joined

consensus/walstore/wal_index.go
```diff
@@ -0,0 +1,101 @@
...
+func (s *tendermintWALStore[V, H, A]) updateIndexesFromCommittedRecords(
+	walNum pebblewal.NumWAL,
+	records []walRecordEnvelope[V, H, A],
+) {
+	for _, record := range records {
+		switch record.Kind {
+		case walRecordEntry:
+			entry, err := record.entry()
+			if err != nil {
+				panic(err)
```
rodrodros: Why should e panic here and handle the error in other places?

rodrodros (approved):
LGTM!

consensus/walstore/wal_index.go
ongyimeng: This was meant to document the invariant that records are already valid here, since they come from setEntry or decodeWALRecord. This path is impossible in practice, I’ll remove the extra error handling to stay consistent with the other impossible WAL checks we already removed.

ongyimeng:
Please hold on for abit, I’m still doing some final checks and cleaning up a few things.

consensus/walstore/codec.go
ongyimeng: The length prefix is written before the payload, but we only know the payload size after encoding it. So I reserved the worst-case 5 bytes up front and used this function to fill that fixed slot in afterward, without shifting the payload. A variable-size uvarint would need the payload encoded into a scratch buffer first, then the length prefix appended before it.

db/memory/db_test.go
```diff
@@ -1,15 +1,39 @@
-package memory
+package memory_test
```
RafaelGranza, line 1: 👀

db/memory/db_test.go
```diff
@@ -1,15 +1,39 @@
...
+}
+
+func TestMemoryDBCopyGetsIndependentPath(t *testing.T) {
+	memoryDB := memory.New()
+	originalPath := memoryDB.Path()
+
+	copyDB := memoryDB.Copy()
+	copyPath := copyDB.Path()
+	require.NotEmpty(t, copyPath)
+	require.NotEqual(t, originalPath, copyPath)
```
RafaelGranza: **Nit:** I would like you to test:
`require.NotEqual(t, memoryDB.Path(), copyPath)`

Since the path "could" have changed after the copy. It does not happen here, but covers any "weird" changes in the future.

db/database.go
```diff
@@ -46,6 +46,9 @@ type Helper interface {
 	// remove this once the metrics are refactored
 	// Returns the underlying database
 	Impl() any
+	// Returns a local filesystem path for consensus WAL files.
+	// Empty means the DB has no local path.
+	Path() string
```
RafaelGranza, line 51: *nit:* `Helper` seems like a mess of interface, everything that doesnt fit anywhere else ends here.
IMO `Path` could live inside `KeyValueStore`.

**You may consider not applying this change.**

ongyimeng: yea, I agree. This feels more like a smell in the existing `Helper` interface, which should probably be revisited separately.

PR #3764, opened 2026-06-25
feat(compiler): memory-aware concurrency limit

RafaelGranza:
# Auto-size Sierra compilation concurrency from available memory

## What

Today the number of concurrent Sierra→CASM compilations defaults to `GOMAXPROCS`, which ignores memory. A burst of large compilations can exhaust RAM.
This PR derives the limit from **available memory and CPU count**.

Formula (all values in MB):

```
limit = min( (availableMemory - nodeMemoryReserve) / maxMemoryPerCompilation , CPUs )
```

The limit is always at least 1: a memory-tight node compiles one at a time rather than refusing to start.

## Behavior change (heads-up for operators)

The default concurrency is no longer always `GOMAXPROCS`. On a many-core but memory-modest host it drops: e.g. 16 CPU / 16 GB with defaults goes from `16` to `(16-4)/4 = 3`. Intended, but nodes with heavy P2P or `getCompiledCasm` load will see less compile parallelism. Raise `--max-concurrent-compilations` (used as-is) or `--node-memory-reserve` to tune.

## Flags

| Flag | Default | Meaning |
|---|---|---|
| `--max-concurrent-compilations` | `unset` | `unset` (flag not provided) = derive from memory + CPUs. A provided value is used as-is (an explicit `0` disables compilations). |
| `--max-compilation-queue` | `unset` | `unset` = twice the resulting concurrency. A provided value is used as-is. |
| `--node-memory-reserve` | `4096` | RAM (MB) kept for the rest of the node, out of the compilation budget. **New.** |
| `--max-compilation-memory` | `4096` | RAM (MB) one compilation may use. Existing; it is the divisor above. |

Compilation-sizing flags are parsed as unsigned integers by the CLI, so a non-numeric or negative value is rejected at flag-parse time. Whether a flag was provided (across CLI, env, or YAML) is what selects derive-vs-explicit, so an explicit `0` is a valid, distinct value.

## Why these libraries

Two needs: read total RAM, and read the container memory limit. Choice:

| Need | Library | Why |
|---|---|---|
| Host/VM RAM | `pbnjay/memory` | One function, pure Go, no cgo, all OSes. |
| Container limit | `automemlimit` | Purpose-built for `GOMEMLIMIT`, pure Go, reads cgroup v1/v2. |

Both were already indirect dependencies; this only promotes them to direct.
`automemlimit`'s only dependency is `pbnjay/memory`, so the pair adds **no new transitive dependency**.

`gopsutil` was considered and dropped: it does not read the cgroup limit (same blind spot as `pbnjay`), and it would be an extra dependency on top of `pbnjay` (which `automemlimit` pulls anyway).

## Container awareness

A **cgroup** is the Linux kernel feature that caps how much memory a process group may use; it is how Docker and Kubernetes enforce memory limits. A process can sit on a 64 GiB host but be capped to 4 GiB by its cgroup, and the kernel kills it (OOM) if it goes over.

Reading total system memory (`sysinfo` / `/proc/meminfo`) returns the **host** RAM and ignores cgroups, so inside a memory-limited container it over-reports.

`AvailableMemoryMB()` instead returns `min(host RAM, cgroup limit)`, so it respects `docker run --memory=..` and Kubernetes limits.

| Environment | host-only reading | this PR |
|---|---|---|
| Bare metal / no limit | host RAM | host RAM |
| Container with memory limit | host RAM (wrong) | cgroup limit |
| macOS / Windows (no cgroup) | host RAM | host RAM |

## Margin default

`--node-memory-reserve` defaults to **4 GiB**, a conservative reserve for the rest of the node. No deeper analysis; it is configurable. Operators with a large `--db-cache-size` or many VMs can raise it.

## Testing

Verified by running the node and reading the startup log
(`Sierra compilation limits`).

| Environment | Config | `availableMemory` | concurrent compilations | Result |
|---|---|---|---|---|
| macOS native | default | 16384 (host) | 10 (CPUs) | boots |
| macOS native | `--max-compilation-memory 4096` | 16384 | 3 | boots |
| macOS native | `--node-memory-reserve 100000` | 16384 | 1 (floored) | boots |
| Docker `--memory=8g` | default | **8192 (cgroup)** | 1 | boots |
| Docker `--memory=6g` | default | **6144 (cgroup)** | 1 (floored) | boots |

The Docker rows confirm the value follows the cgroup limit, not the 16 GiB host.

When the budget fits no compilation the limit floors to 1, so the node boots and compiles one at a time instead of refusing to start.

### DIFF
--- a/cmd/juno/juno.go
+++ b/cmd/juno/juno.go
@@ -484,7 +484,6 @@ func NewCmd(config *node.Config, run func(*cobra.Command, []string) error) *cobr
 	// may mutate their values.
 	defaultNetwork := networks.Mainnet
 	defaultMaxVMs := 3 * runtime.GOMAXPROCS(0)
-	// Empty derives at startup; an explicit 0 disables compilations / the queue.
 	defaultCNUnverifiableRange := []int{} // Uint64Slice is not supported in Flags()
 
 	// --- HTTP RPC ---
--- a/cmd/juno/juno.go
+++ b/cmd/juno/juno.go
@@ -111,6 +111,7 @@ const (
 	maxConcurrentCompilationsF          = "max-concurrent-compilations"
 	maxCompilationQueueF                = "max-compilation-queue"
 	maxCompilationMemoryF               = "max-compilation-memory"
+	nodeMemoryReserveF                  = "node-memory-reserve"
 	maxCompilationCPUTimeF              = "max-compilation-cpu-time"
 	disableReceivedTxnStreamF           = "disable-received-txn-stream"
 	newStateF                           = "new-state"
@@ -175,6 +176,7 @@ const (
 	defaultRPCMaxConcurrentRequests           = 256000
 	defaultRPCMaxQueuedRequests               = 256000
 	defaultMaxCompilationMemory               = 4 * 1024 // MB (4 GB) per compilation process
+	defaultNodeMemoryReserve                  = 4 * 1024 // MB (4 GB) reserved for the rest of the node
 	defaultMaxCompilationCPUTime              = 10       // seconds of CPU time per compilation process
 	defaultDisableReceivedTxnStream           = false
 	defaultPruneMode                          = uint64(0)
@@ -260,11 +262,16 @@ const (
 	rpcMaxConcurrentRequestsUsage = "Maximum concurrent HTTP RPC requests; 0 disables the limit."
 	rpcMaxRequestQueueUsage       = "Maximum number of HTTP RPC requests to queue after " +
 		"reaching rpc-max-concurrent-requests limit."
-	maxConcurrentCompilationsUsage = "Maximum concurrent Sierra compilations."
-	maxCompilationQueueUsage       = "Maximum number of compilation requests to queue after " +
-		"reaching max-concurrent-compilations before starting to reject incoming requests."
+	maxConcurrentCompilationsUsage = "Maximum concurrent Sierra compilations. " +
+		"0 derives a safe value from available memory and CPU count."
+	maxCompilationQueueUsage = "Maximum number of compilation requests to queue after " +
+		"reaching max-concurrent-compilations before starting to reject incoming requests. " +
+		"0 uses twice the concurrency limit."
 	maxCompilationMemoryUsage = "Maximum memory (in MB) each Sierra compilation process may " +
 		"use; a compilation exceeding it is aborted. Enforced on Linux only. 0 disables the limit."
+	nodeMemoryReserveUsage = "Memory (in MB) reserved for the rest of the node and excluded from " +
+		"the compilation memory budget. Used only when max-concurrent-compilations is 0. " +
+		"Container memory limits (cgroups) are respected."
 	maxCompilationCPUTimeUsage = "Maximum CPU time (in seconds) each Sierra compilation process " +
 		"may consume; a compilation exceeding it is aborted. Enforced on Linux only. " +
 		"0 disables the limit."
@@ -477,8 +484,9 @@ func NewCmd(config *node.Config, run func(*cobra.Command, []string) error) *cobr
 	// may mutate their values.
 	defaultNetwork := networks.Mainnet
 	defaultMaxVMs := 3 * runtime.GOMAXPROCS(0)
-	defaultMaxConcurrentCompilations := runtime.GOMAXPROCS(0)
-	defaultCNUnverifiableRange := []int{} // Uint64Slice is not supported in Flags()
+	defaultMaxConcurrentCompilations := uint(0) // 0 = derive at startup (see node.New).
+	defaultMaxCompilationQueue := uint(0)       // 0 = derive at startup (see node.New).
+	defaultCNUnverifiableRange := []int{}       // Uint64Slice is not supported in Flags()
 
 	// --- HTTP RPC ---
 	junoCmd.Flags().Bool(httpF, defaultHTTP, httpUsage)
@@ -606,15 +614,18 @@ func NewCmd(config *node.Config, run func(*cobra.Command, []string) error) *cobr
 	junoCmd.Flags().Uint(maxVMQueueF, 2*uint(defaultMaxVMs), maxVMQueueUsage)
 	junoCmd.Flags().Uint(
 		maxConcurrentCompilationsF,
-		uint(defaultMaxConcurrentCompilations),
+		defaultMaxConcurrentCompilations,
 		maxConcurrentCompilationsUsage,
 	)
 	junoCmd.Flags().Uint(
 		maxCompilationQueueF,
-		2*uint(defaultMaxConcurrentCompilations),
+		defaultMaxCompilationQueue,
 		maxCompilationQueueUsage,
 	)
 	junoCmd.Flags().Uint(maxCompilationMemoryF, defaultMaxCompilationMemory, maxCompilationMemoryUsage)
+	junoCmd.Flags().Uint(
+		nodeMemoryReserveF, defaultNodeMemoryReserve, nodeMemoryReserveUsage,
+	)
 	junoCmd.Flags().Uint(
 		maxCompilationCPUTimeF, defaultMaxCompilationCPUTime, maxCompilationCPUTimeUsage,
 	)
@@ -626,6 +637,7 @@ func NewCmd(config *node.Config, run func(*cobra.Command, []string) error) *cobr
 		maxConcurrentCompilationsF,
 		maxCompilationQueueF,
 		maxCompilationMemoryF,
+		nodeMemoryReserveF,
 		maxCompilationCPUTimeF,
 		versionedConstantsFileF,
 	)
--- a/cmd/juno/juno.go
+++ b/cmd/juno/juno.go
@@ -175,6 +175,8 @@ const (
 	defaultRPCRequestTimeout                  = 1 * time.Minute
 	defaultRPCMaxConcurrentRequests           = 256000
 	defaultRPCMaxQueuedRequests               = 256000
+	defaultMaxConcurrentCompilations          = ""
+	defaultMaxCompilationQueue                = ""
 	defaultMaxCompilationMemory               = 4 * 1024 // MB (4 GB) per compilation process
 	defaultNodeMemoryReserve                  = 4 * 1024 // MB (4 GB) reserved for the rest of the node
 	defaultMaxCompilationCPUTime              = 10       // seconds of CPU time per compilation process
@@ -263,17 +265,16 @@ const (
 	rpcMaxRequestQueueUsage       = "Maximum number of HTTP RPC requests to queue after " +
 		"reaching rpc-max-concurrent-requests limit."
 	maxConcurrentCompilationsUsage = "Maximum concurrent Sierra compilations. " +
-		"Empty derives a safe value from available memory and CPU count; 0 disables compilations."
+		"Default is set based on available hardware resources."
 	maxCompilationQueueUsage = "Maximum number of compilation requests to queue after " +
-		"reaching max-concurrent-compilations before starting to reject incoming requests. " +
-		"Empty uses twice the concurrency limit."
-	maxCompilationMemoryUsage = "Maximum memory (in MB) each Sierra compilation process may " +
+		"reaching max-concurrent-compilations. Default sets the queue to twice the concurrency limit."
+	maxCompilationMemoryUsage = "Maximum virtual memory (in MB) a Sierra compilation process may " +
 		"use; a compilation exceeding it is aborted. Enforced on Linux only. 0 disables the limit."
-	nodeMemoryReserveUsage = "Memory (in MB) reserved for the rest of the node and excluded from " +
-		"the compilation memory budget. Used only when max-concurrent-compilations is empty."
 	maxCompilationCPUTimeUsage = "Maximum CPU time (in seconds) each Sierra compilation process " +
 		"may consume; a compilation exceeding it is aborted. Enforced on Linux only. " +
 		"0 disables the limit."
+	maxCompilationReserveMemory = "Memory (in MB) excluded from the compilations memory budget when " +
+		"calculating the default for `max-concurrent-compilations`"
 	pruneModeUsage = "Enables block-data and state-history pruning. Pruning is " +
 		"disabled by default; passing this flag (with or without a value) turns " +
 		"it on. The value is the size of the retention window in blocks, counted " +
@@ -484,8 +485,6 @@ func NewCmd(config *node.Config, run func(*cobra.Command, []string) error) *cobr
 	defaultNetwork := networks.Mainnet
 	defaultMaxVMs := 3 * runtime.GOMAXPROCS(0)
 	// Empty derives at startup; an explicit 0 disables compilations / the queue.
-	defaultMaxConcurrentCompilations := ""
-	defaultMaxCompilationQueue := ""
 	defaultCNUnverifiableRange := []int{} // Uint64Slice is not supported in Flags()
 
 	// --- HTTP RPC ---
@@ -624,7 +623,7 @@ func NewCmd(config *node.Config, run func(*cobra.Command, []string) error) *cobr
 	)
 	junoCmd.Flags().Uint(maxCompilationMemoryF, defaultMaxCompilationMemory, maxCompilationMemoryUsage)
 	junoCmd.Flags().Uint(
-		nodeMemoryReserveF, defaultNodeMemoryReserve, nodeMemoryReserveUsage,
+		nodeMemoryReserveF, defaultNodeMemoryReserve, maxCompilationReserveMemory,
 	)
 	junoCmd.Flags().Uint(
 		maxCompilationCPUTimeF, defaultMaxCompilationCPUTime, maxCompilationCPUTimeUsage,
--- a/cmd/juno/juno.go
+++ b/cmd/juno/juno.go
@@ -263,15 +263,14 @@ const (
 	rpcMaxRequestQueueUsage       = "Maximum number of HTTP RPC requests to queue after " +
 		"reaching rpc-max-concurrent-requests limit."
 	maxConcurrentCompilationsUsage = "Maximum concurrent Sierra compilations. " +
-		"0 derives a safe value from available memory and CPU count."
+		"Empty derives a safe value from available memory and CPU count; 0 disables compilations."
 	maxCompilationQueueUsage = "Maximum number of compilation requests to queue after " +
 		"reaching max-concurrent-compilations before starting to reject incoming requests. " +
-		"0 uses twice the concurrency limit."
+		"Empty uses twice the concurrency limit."
 	maxCompilationMemoryUsage = "Maximum memory (in MB) each Sierra compilation process may " +
 		"use; a compilation exceeding it is aborted. Enforced on Linux only. 0 disables the limit."
 	nodeMemoryReserveUsage = "Memory (in MB) reserved for the rest of the node and excluded from " +
-		"the compilation memory budget. Used only when max-concurrent-compilations is 0. " +
-		"Container memory limits (cgroups) are respected."
+		"the compilation memory budget. Used only when max-concurrent-compilations is empty."
 	maxCompilationCPUTimeUsage = "Maximum CPU time (in seconds) each Sierra compilation process " +
 		"may consume; a compilation exceeding it is aborted. Enforced on Linux only. " +
 		"0 disables the limit."
@@ -484,9 +483,10 @@ func NewCmd(config *node.Config, run func(*cobra.Command, []string) error) *cobr
 	// may mutate their values.
 	defaultNetwork := networks.Mainnet
 	defaultMaxVMs := 3 * runtime.GOMAXPROCS(0)
-	defaultMaxConcurrentCompilations := uint(0) // 0 = derive at startup (see node.New).
-	defaultMaxCompilationQueue := uint(0)       // 0 = derive at startup (see node.New).
-	defaultCNUnverifiableRange := []int{}       // Uint64Slice is not supported in Flags()
+	// Empty derives at startup; an explicit 0 disables compilations / the queue.
+	defaultMaxConcurrentCompilations := ""
+	defaultMaxCompilationQueue := ""
+	defaultCNUnverifiableRange := []int{} // Uint64Slice is not supported in Flags()
 
 	// --- HTTP RPC ---
 	junoCmd.Flags().Bool(httpF, defaultHTTP, httpUsage)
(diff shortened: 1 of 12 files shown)


node/node.go
```diff
@@ -291,6 +292,41 @@ func New(cfg *Config, version string, logLevel *log.Level) (*Node, error) {
...
+	// 0 only reaches here when the node does not compile, so it was not an error above.
+	// Keep one slot so the throttler is not degenerate.
+	concurrentCompilations = max(concurrentCompilations, 1)
+
+	logger.Info("setting Sierra compilation concurrency",
+		zap.Uint("limit", concurrentCompilations),
+		zap.Uint64("availableMemoryMB", availableMemoryMB),
+		zap.Uint("nodeMemoryReserveMB", cfg.NodeMemoryReserve),
+		zap.Uint("maxMemoryPerCompilationMB", cfg.MaxCompilationMemory),
+	)
```
RafaelGranza: I'm not 100% sure if I should leave the log here.

starknet/compiler/concurrency.go
```diff
@@ -0,0 +1,38 @@
...
+
+const megabyte = 1 << 20
+
+// AvailableMemoryMB returns the RAM this process can use, in MB.
+// Inside a container it uses the cgroup limit, otherwise the host RAM.
+func AvailableMemoryMB() uint64 {
+	hostMemory := memory.TotalMemory()
+	cgroupLimit, err := memlimit.FromCgroup()
+	if err == nil && cgroupLimit > 0 && cgroupLimit < hostMemory {
+		return cgroupLimit / megabyte
```
RafaelGranza, line 16: Won't be covering this line by a test

cmd/juno/juno.go
```diff
@@ -257,11 +259,16 @@ const (
...
+	maxConcurrentCompilationsUsage = "Maximum concurrent Sierra compilations. " +
+		"0 derives a safe value from available memory and CPU count."
+	maxCompilationQueueUsage = "Maximum number of compilation requests to queue after " +
+		"reaching max-concurrent-compilations before starting to reject incoming requests. " +
+		"0 uses twice the concurrency limit."
 	maxCompilationMemoryUsage = "Maximum memory (in MB) each Sierra compilation process may " +
 		"use; a compilation exceeding it is aborted. Enforced on Linux only. 0 disables the limit."
+	nodeMemoryReserveUsage = "Memory (in MB) reserved for the rest of the node and excluded from " +
+		"the compilation memory budget. Used only when max-concurrent-compilations is 0. " +
+		"Container memory limits (cgroups) are respected."
```
rodrodros: Unnecessary line. Remove the last whitespace from the line on top when updating.

cmd/juno/juno.go
```diff
@@ -472,8 +479,9 @@ func NewCmd(config *node.Config, run func(*cobra.Command, []string) error) *cobr
 	// may mutate their values.
 	defaultNetwork := networks.Mainnet
 	defaultMaxVMs := 3 * runtime.GOMAXPROCS(0)
-	defaultMaxConcurrentCompilations := runtime.GOMAXPROCS(0)
-	defaultCNUnverifiableRange := []int{} // Uint64Slice is not supported in Flags()
+	defaultMaxConcurrentCompilations := uint(0) // 0 = derive at startup (see node.New).
+	defaultMaxCompilationQueue := uint(0)       // 0 = derive at startup (see node.New).
```
rodrodros: Maybe they can be constants? Maybe they can be something else.

rodrodros (commented):
Rafael the 0 as a way to set is default is not a good solution because 0 can be a desired value in itself. A node set with `0,0` values can me a node that will reject any compilation requests. 

Since Cobra allow us to have some expression, can't the defaults be triggered at runtime and CLI print the default values? I don't know if this is a good approach. 

The other approach is to allow other values, such as "default" or "dynamic" which are non-numbers and if set, then the calculation is done

RafaelGranza:
@rodrodros
I Switched the default to `""` (empty string), so `0` stays valid as "disable". So the flags are now `Int`. It is the simplest option I found, but open to alternatives.

node/node.go
```diff
@@ -149,6 +152,19 @@ type Config struct {
 	PruneMinAge    time.Duration `mapstructure:"prune-min-age"`
 }
 
+// parseCompilationLimit reads a compilation-sizing flag. An empty value derives the
+// value at startup (derive is true); otherwise it must be a non-negative integer.
+func parseCompilationLimit(value string) (limit uint, derive bool, err error) {
+	if value == "" {
+		return 0, true, nil
+	}
+	n, err := strconv.ParseUint(value, 10, 64)
```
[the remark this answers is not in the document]
RafaelGranza: Makes sense

RafaelGranza:
# Explaining the chosen variables

| flag | value | role |
|---|---|---|
| `--max-compilation-memory` | 4 GB | per-compile address-space ceiling; also sets how many run at once |
| `--node-memory-reserve` | 4 GB | memory held back for the rest of the node |

Plus: the compile child process is pinned to `GOMAXPROCS=2`.

## Real vs virtual memory

Each compilation runs in its own child process, with two very different memory numbers:

| | a compilation | what it means |
|---|---|---|
| Real memory (RSS) | ~150-200 MB | what it actually occupies in RAM |
| Virtual memory (address space) | ~1.5-1.8 GB | what it maps, used or not |

They differ this much because **Go reserves far more virtual memory than it uses**.

## Compilation memory ceiling (4 GB)

`--max-compilation-memory` is an `RLIMIT_AS` limit, so it caps **virtual** memory. A compile maps ~1.6 GB of it, so the limit has to be around that, not the ~200 MB of real usage. Capping real memory is not reliable, only trhorugh `cgroup`, which needs special permision in containers.

Measured peak virtual memory of the heaviest honest mainnet contract:

| arch / cores | VmPeak |
|---|---|
| arm64 | ~1.6 GB |
| amd64, 16 cores | 1.77 GB |
| amd64, 128 cores (unpinned) | 1.93 GB |
| any, pinned to GOMAXPROCS=2 | ~1.58 GB |

4 GB sits conformably above that peak.

## Pinning the compile child to GOMAXPROCS=2

A compilation is one Rust FFI call, so extra Go threads do not make it faster; they only add per-thread stacks that grow the virtual memory peak. Inheriting a large host GOMAXPROCS pushes the memory peak up (1.93 GB at 128 cores, near the limit). Pinning keeps it host-independent (~1.58 GB) at no time cost.

## Node memory reserve (4 GB)

Measured on a real synced mainnet node via `process_resident_memory_bytes`, which counts the node's whole footprint together: Go heap, pebble (DB) cache, state, contract execution and RPC.

| | RSS |
|---|---|
| steady state | ~1.5 GB |
| peak over 24h | ~2.25 GB |
| worst over 7d | ~2.86 GB |

4 GB covers the peak with headroom.

## Validation

Checked derivation, fallback and virtual peak across Docker (arm64), native macOS (arm64) and a native amd64 host, from 1 to 128 cores. Build, tests and lint green.

RafaelGranza:
The bot complaining of "Stale PR description" does not hold.

All the changes were addressed.

node/node.go
```diff
@@ -148,6 +151,19 @@ type Config struct {
...
 
+// parseCompilationLimit reads a compilation-sizing flag. An empty value derives the
+// value at startup (derive is true); otherwise it must be a non-negative integer.
+func parseCompilationLimit(value string) (limit uint, derive bool, err error) {
+	if value == "" {
+		return 0, true, nil
+	}
+	n, err := strconv.ParseUint(value, 10, strconv.IntSize)
+	if err != nil {
+		return 0, false, fmt.Errorf("%q is not a non-negative integer", value)
```
rodrodros: A test please

node/node.go
```diff
@@ -292,6 +308,39 @@ func New(cfg *Config, version string, logLevel *log.Level) (*Node, error) {
...
+	}
+
+	queue, queueAuto, err := parseCompilationLimit(cfg.MaxCompilationQueue)
+	if err != nil {
+		return nil, fmt.Errorf("parsing max-compilation-queue: %w", err)
+	}
+	compilationQueue := queue
+	if queueAuto {
+		compilationQueue = 2 * concurrentCompilations
+	}
```
rodrodros: This whole thing kind of makes an already very big function, bigger. Can it be defined inside it's own function?

starknet/compiler/concurrency_test.go
rodrodros, line 1: Good 👍

rodrodros (approved):

node/node.go
```diff
@@ -164,6 +164,57 @@ func parseCompilationLimit(value string) (limit uint, derive bool, err error) {
...
 }
 
+func newThrottledCompilerFromConfig(
+	cfg *Config, logger *log.ZapLogger,
+) (*ThrottledCompiler, error) {
+	concurrent, concurrentAuto, err := parseCompilationLimit(cfg.MaxConcurrentCompilations)
+	if err != nil {
+		return nil, fmt.Errorf("parsing max-concurrent-compilations: %w", err)
+	}
+	var concurrentCompilations uint
```
rodrodros: Why isn't this a uint64?

node/node.go
```diff
@@ -164,6 +164,57 @@ func parseCompilationLimit(value string) (limit uint, derive bool, err error) {
 	return uint(n), false, nil
 }
 
+func newThrottledCompilerFromConfig(
```
rodrodros: There is a dedicated `throttled_compiler.go` file already existing, wouldn't you think this function should be defined directly there?

node/node.go
```diff
@@ -164,6 +164,57 @@ func parseCompilationLimit(value string) (limit uint, derive bool, err error) {
...
+	cfg *Config, logger *log.ZapLogger,
+) (*ThrottledCompiler, error) {
+	concurrent, concurrentAuto, err := parseCompilationLimit(cfg.MaxConcurrentCompilations)
```
rodrodros: What does `concurrentAuto` means, how does a someone reading from this code get what does this variable means. 

Think before naming your variables please. Suggestions:
* `dynamicallySet`
* `automaticallySet`

On other topic, what does `concurrent` means by itself? Perhaps you would like to name it `maxConcurrentCompilations` or `maxConcurrent`?

node/node.go
```diff
@@ -164,6 +164,57 @@ func parseCompilationLimit(value string) (limit uint, derive bool, err error) {
...
+			uint64(cfg.MaxCompilationMemory),
+		), 1)
+		logger.Info("deriving Sierra compilation concurrency",
+			zap.Uint("limit", concurrentCompilations),
+			zap.Uint64("availableMemoryMB", availableMemoryMB),
+			zap.Uint("nodeMemoryReserveMB", cfg.NodeMemoryReserve),
+			zap.Uint("maxMemoryPerCompilationMB", cfg.MaxCompilationMemory),
+		)
+	} else {
+		concurrentCompilations = concurrent
```
rodrodros: Kind of an unnecessary assign one might say?

node/node.go
```diff
@@ -164,6 +164,57 @@ func parseCompilationLimit(value string) (limit uint, derive bool, err error) {
...
+	cfg *Config, logger *log.ZapLogger,
+) (*ThrottledCompiler, error) {
+	concurrent, concurrentAuto, err := parseCompilationLimit(cfg.MaxConcurrentCompilations)
+	if err != nil {
+		return nil, fmt.Errorf("parsing max-concurrent-compilations: %w", err)
+	}
+	var concurrentCompilations uint
+	if concurrentAuto {
+		availableMemoryMB := compiler.AvailableMemoryMB()
+		concurrentCompilations = max(compiler.ConcurrencyLimit(
```
rodrodros: Here as well, why not re-using the variable already created?

node/node.go
```diff
@@ -164,6 +164,57 @@ func parseCompilationLimit(value string) (limit uint, derive bool, err error) {
...
+			zap.Uint("nodeMemoryReserveMB", cfg.NodeMemoryReserve),
+			zap.Uint("maxMemoryPerCompilationMB", cfg.MaxCompilationMemory),
+		)
+	} else {
+		concurrentCompilations = concurrent
+		logger.Info("using configured Sierra compilation concurrency",
+			zap.Uint("limit", concurrentCompilations))
+	}
+
+	queue, queueAuto, err := parseCompilationLimit(cfg.MaxCompilationQueue)
```
rodrodros: similar comment regarding variable

node/node.go
```diff
@@ -164,6 +164,57 @@ func parseCompilationLimit(value string) (limit uint, derive bool, err error) {
...
+	if err != nil {
+		return nil, fmt.Errorf("parsing max-compilation-queue: %w", err)
+	}
+	compilationQueue := queue
+	if queueAuto {
+		compilationQueue = 2 * concurrentCompilations
+	}
```
rodrodros: Maybe you want to name it `compilationQueueSize` or `queueSize`

node/node.go
```diff
@@ -308,51 +359,10 @@ func New(cfg *Config, version string, logLevel *log.Level) (*Node, error) {
...
-
-	queue, queueAuto, err := parseCompilationLimit(cfg.MaxCompilationQueue)
+	throttledCompiler, err := newThrottledCompilerFromConfig(cfg, logger)
 	if err != nil {
-		return nil, fmt.Errorf("parsing max-compilation-queue: %w", err)
-	}
-	compilationQueue := queue
-	if queueAuto {
-		compilationQueue = 2 * concurrentCompilations
+		return nil, err
```
rodrodros: Maybe you would like to wrap the error and specify that it happened "initializing the throttled compiler"?

rodrodros (commented):
I don't know how to transmit that while it is important to solve the issue. It is also important to do it in an elegant way that fits into the node style and makes the code easier to read. 

While of course some of there might be subjective, these so many style comments signals a carelessly written PR, with no after thought of code architecture, code cleanliness and code readability – aspects as important to Software Development as just solving the issue. 

I hope code **quality** improves for the next PRs.

starknet/compiler/concurrency.go
```diff
@@ -0,0 +1,38 @@
...
+	}
+	return hostMemory / megabyte
+}
+
+// ConcurrencyLimit returns how many compilations fit in memory, capped by maxParallel.
+// Memory is in MB. A zero maxMemoryPerCompilation means no memory limit.
+// Returns 0 when memory fits no compilation.
+func ConcurrencyLimit(
+	maxParallel uint,
```
rodrodros: In every aspect of the code this variable is referred to as "concurrency", why the disparity here calling it "parallel"?

rodrodros (changes requested):
Canceling approve until code quality increases. Please, write code consciously with quality – making the codebase is as readable as it was before (at the very least).

cmd/juno/juno.go
```diff
@@ -260,14 +264,17 @@ const (
...
+	maxConcurrentCompilationsUsage = "Maximum concurrent Sierra compilations. " +
+		"Default is set based on available hardware resources."
+	maxCompilationQueueUsage = "Maximum number of compilation requests to queue after " +
+		"reaching max-concurrent-compilations. Default sets the queue to twice the concurrency limit."
+	maxCompilationMemoryUsage = "Maximum virtual memory (in MB) a Sierra compilation process may " +
 		"use; a compilation exceeding it is aborted. Enforced on Linux only. 0 disables the limit."
 	maxCompilationCPUTimeUsage = "Maximum CPU time (in seconds) each Sierra compilation process " +
 		"may consume; a compilation exceeding it is aborted. Enforced on Linux only. " +
 		"0 disables the limit."
+	maxCompilationReserveMemory = "Memory (in MB) excluded from the compilations memory budget when " +
```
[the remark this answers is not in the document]
RafaelGranza: will leave it as is

node/throttled_compiler.go
```diff
@@ -14,6 +17,48 @@ type ThrottledCompiler struct {
...
+	}
+
+	logger.Info("Sierra compilation limits",
+		zap.Uint64("maxConcurrency", maxConcurrency),
+		zap.Bool("maxConcurrencyDerived", !cfg.MaxConcurrentCompilationsExplicit),
+		zap.Uint64("queueSize", queueSize),
+		zap.Bool("queueSizeDerived", !cfg.MaxCompilationQueueExplicit),
+		zap.Uint64("availableMemory", availableMemory),
+		zap.Uint("NodeMemoryReserve", cfg.NodeMemoryReserve),
+		zap.Uint("MaxCompilationMemory", cfg.MaxCompilationMemory),
```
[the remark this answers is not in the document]
RafaelGranza: fixed

node/node.go
RafaelGranza: I tried to keep this method at `throttled_compiler.go`,  but it was even messier there, logging, configs, parsing...
It was an actual code smell. The method felt off no matter where I put it.
Now I refactored the whole feature, and it fits much better.

node/node.go
RafaelGranza: fixed

node/node.go
RafaelGranza: The `throttledCompiler` expects an uint, so not much of a change.

node/node.go
RafaelGranza: Now the flow is error-free

RafaelGranza:
@rodrodros I refactored the code. 
The flags are now `uint64` and parsed at the cmd level: if you don't pass them, we derive the values. 
The throttled compiler setup lives entirely in its own file now, so `node.New` here is cleaner than the one in `main`.

I decided to parse the flags at the cmd level because otherwise the throttled compiler, which was error-free, would suddenly have to carry parsing errors. On top of that, parsing a CLI flag shouldn't live inside `throttled_compiler.go` or `node.New`.

docs/generate-config.js
```diff
@@ -106,18 +106,20 @@ const extractConfigs = (codebase) => {
     if (configName === "max-vm-queue") {
       defaultValue = "2 * max-vms";
     }
+    // The registered Cobra default is 0, but an absent flag derives the value at
+    // startup, so show the effective derived default rather than the raw 0.
     if (configName === "max-concurrent-compilations") {
       defaultValue = "CPU Cores";
```
rodrodros: I think we should update this default value to the formula, a bit tricky to do in a readable way

node/throttled_compiler.go
```diff
@@ -14,6 +17,48 @@ type ThrottledCompiler struct {
 	*throttler.Throttler[compiler.Compiler]
 }
 
+type ThrottledCompilerSettings struct {
+	maxConcurrency uint64
+	queueSize      uint64
+}
```
rodrodros: You are creating a struct (also public) to handle just two values, which is also used only one? Couldn't this function return the two appointed values directly?

node/throttled_compiler.go
```diff
@@ -14,6 +17,48 @@ type ThrottledCompiler struct {
...
+		zap.Uint("nodeMemoryReserveMB", cfg.NodeMemoryReserve),
+		zap.Uint("maxCompilationMemoryMB", cfg.MaxCompilationMemory),
+	)
+
+	return &ThrottledCompilerSettings{
+		maxConcurrency, queueSize,
+	}
+}
+
 func NewThrottledCompiler(
```
rodrodros: What's the point of `NewThrottledCompiler` now? It seems to me there is no longer a point for this function?

node/throttled_compiler.go
```diff
@@ -14,6 +17,48 @@ type ThrottledCompiler struct {
...
+	logger.Info("Sierra compilation limits",
+		zap.Uint64("maxConcurrency", maxConcurrency),
+		zap.Bool("maxConcurrencyExplicit", cfg.MaxConcurrentCompilationsExplicit),
+		zap.Uint64("queueSize", queueSize),
+		zap.Bool("queueSizeExplicit", cfg.MaxCompilationQueueExplicit),
+		zap.Uint64("availableMemoryMB", availableMemoryMB),
+		zap.Uint("nodeMemoryReserveMB", cfg.NodeMemoryReserve),
+		zap.Uint("maxCompilationMemoryMB", cfg.MaxCompilationMemory),
+	)
+
```
rodrodros: I think "explicit" information here is redundant. The user running it's node should know if he is being explicit about the values in his configs

node/throttled_compiler.go
```diff
@@ -14,6 +17,48 @@ type ThrottledCompiler struct {
...
+	return &ThrottledCompilerSettings{
+		maxConcurrency, queueSize,
+	}
+}
```
rodrodros: Also, returning a pointer. Why?

docs/generate-config.js
RafaelGranza: ooh, yeah, you are right. I guess I got confused here.
I'll fit the formula here somehow

node/throttled_compiler.go
RafaelGranza: It was to keep these two tied together, so that resolving the `setting` would actually return a `setting`.
I wasn't aware  structs could be lowercase (private), which it should be.

I'm not sure if it fits as over-engineering, but it is not needed.

node/throttled_compiler.go
RafaelGranza: I like the test that `NewThrottledCompiler` exercises (`node/throttled_compiler_test.go`).

It could also be exercised by the new functions, but we would require us to set them public and passing a `compiler.Compiler`, creating a messier signature, with the `compiler.Compiler` being create inside `Node.New()`.

This one works as a nice layering for the ThrottledCompiler.

node/throttled_compiler.go
RafaelGranza: fine

node/throttled_compiler.go
RafaelGranza: i´ll be removing it anyway

node/throttled_compiler.go
RafaelGranza: just to avoid extra copies. I'll be removing it anyway

starknet/compiler/concurrency.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+// ConcurrencyLimit returns how many compilations fit in memory, capped by maxConcurrency.
+// Memory is in MB. A zero maxMemoryPerCompilation means no memory limit.
+// Returns 1 when memory fits no compilation.
+func ConcurrencyLimit(
+	maxConcurrency uint64,
+	availableMemory uint64,
+	nodeMemoryReserve uint64,
+	maxMemoryPerCompilation uint64,
```
RafaelGranza, line 28: I considered computing `maxConcurrency` internally, but it was handy for testing its behavior.

node/throttled_compiler.go
```diff
@@ -24,6 +60,25 @@ func NewThrottledCompiler(
...
+		compiler.New(
+			&compiler.Config{
+				MaxMemory:  uint64(cfg.MaxCompilationMemory) * 1024 * 1024,
+				MaxCPUTime: uint64(cfg.MaxCompilationCPUTime),
+			},
+			"",
+			logger,
+		),
+		uint(maxConcurrency),
+		queueSize,
```
RafaelGranza: I wish I could always use `uint64` to maintain consistency all across the flow, but here the `uint` flows further into `NewThrottler`.

rodrodros (approved):

PR #3766, opened 2026-06-26
refactor(core): remove starknet import

rodrodros:
We were turning a deprecated cairo class from Core into a feeder response to then hash that. It didn't make sense, since we could hash it directly and remove the `starknet` import from `core` production code. 

No noticeable performance gains since hashing is very expensive already.


brbrr (approved):
LGTM

PR #3769, opened 2026-06-27
refactor(networks): store feeder and gateway URLs as *url.URL

### LINKED ISSUE #3761: Make Network type store URLs as a *url.URL type instead of a string
PR https://github.com/NethermindEth/juno/pull/3749 is adding URL validation to the feeder URL, and during the process, we parse the URL to the `url.URL` type. However, since Network type stores the feeder and gatway URLs as `string` we need to do some rework in the code multiple times (string > url.URL > string...)

Make Network type store URLs as a *url.URL type instead of a string

Ehsan-saradar:
## What

The `Network` type stored its feeder and gateway URLs as plain `string`s. Every consumer that needed an actual URL had to re-parse them, and since #3749 the feeder client takes a `*url.URL`, node startup parsed the string only to hand it straight back.

This changes `Network.FeederURL` and `Network.GatewayURL` to `*url.URL` and drops the redundant conversions:

- Predefined networks build their URLs up front (via a small `mustParseURL` helper).
- Custom-network URLs are parsed and validated once during config load (`validateHTTPURL` -> `parseHTTPURL`, returning the parsed URL).
- `node` no longer re-parses the feeder URL before constructing the client.

No behavioural change.

## Testing

- `go build ./cmd/juno/`
- `go test ./blockchain/networks/... ./cmd/juno/... ./node/... ./clients/...`

Closes #3761

### DIFF
--- a/blockchain/networks/network.go
+++ b/blockchain/networks/network.go
@@ -198,11 +198,13 @@ func (n *Network) L2ChainIDFelt() *felt.Felt {
 }
 
 // mustParseURL parses a raw URL and panics on error. It is only used for the
-// hardcoded predefined network URLs, which are always valid.
+// hardcoded predefined network URLs, which are always valid. ParseRequestURI
+// (unlike Parse) rejects relative/scheme-less strings, so a malformed constant
+// fails fast and diagnostically at startup.
 func mustParseURL(rawURL string) *url.URL {
-	u, err := url.Parse(rawURL)
+	u, err := url.ParseRequestURI(rawURL)
 	if err != nil {
-		panic(err)
+		panic(fmt.Sprintf("invalid network URL %q: %v", rawURL, err))
 	}
 	return u
 }
--- a/blockchain/networks/network.go
+++ b/blockchain/networks/network.go
@@ -4,6 +4,7 @@ import (
 	"encoding"
 	"fmt"
 	"math/big"
+	"net/url"
 	"strings"
 
 	"github.com/NethermindEth/juno/core/felt"
@@ -46,8 +47,8 @@ var errUnknownNetwork = fmt.Errorf("unknown network (known: %s)",
 
 type Network struct {
 	Name                string             `json:"name" validate:"required"`
-	FeederURL           string             `json:"feeder_url" validate:"required"`
-	GatewayURL          string             `json:"gateway_url" validate:"required"`
+	FeederURL           *url.URL           `json:"feeder_url" validate:"required"`
+	GatewayURL          *url.URL           `json:"gateway_url" validate:"required"`
 	L1ChainID           *big.Int           `json:"l1_chain_id" validate:"required"`
 	L2ChainID           string             `json:"l2_chain_id" validate:"required"`
 	CoreContractAddress eth.Address        `json:"core_contract_address" validate:"required"`
@@ -73,8 +74,8 @@ var (
 	// The docs states the addresses for each network: https://docs.starknet.io/learn/cheatsheets/chain-info#important-addresses
 	Mainnet = Network{
 		Name:                "mainnet",
-		FeederURL:           "https://feeder.alpha-mainnet.starknet.io/feeder_gateway/",
-		GatewayURL:          "https://alpha-mainnet.starknet.io/gateway/",
+		FeederURL:           mustParseURL("https://feeder.alpha-mainnet.starknet.io/feeder_gateway/"),
+		GatewayURL:          mustParseURL("https://alpha-mainnet.starknet.io/gateway/"),
 		L2ChainID:           "SN_MAIN",
 		L1ChainID:           big.NewInt(1),
 		CoreContractAddress: eth.AddressFromString("0xc662c410C0ECf747543f5bA90660f6ABeBD9C8c4"),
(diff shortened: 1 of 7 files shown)


blockchain/networks/network.go
```diff
@@ -195,3 +196,13 @@ func (n *Network) UnmarshalText(text []byte) error {
...
+
+// mustParseURL parses a raw URL and panics on error. It is only used for the
+// hardcoded predefined network URLs, which are always valid.
+func mustParseURL(rawURL string) *url.URL {
+	u, err := url.Parse(rawURL)
+	if err != nil {
+		panic(err)
+	}
+	return u
+}
```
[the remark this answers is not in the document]
Ehsan-saradar: Good catch — switched to `url.ParseRequestURI` so malformed/scheme-less constants now panic at startup, and the panic message includes the offending value. Done in cc6366072.

blockchain/networks/network_test.go
```diff
@@ -42,8 +42,8 @@ func TestNetwork(t *testing.T) {
 
 		for _, tc := range testCases {
 			t.Run(fmt.Sprintf("Network %v", tc.network), func(t *testing.T) {
-				assert.Equal(t, tc.feederURL, tc.network.FeederURL)
-				assert.Equal(t, tc.gatewayURL, tc.network.GatewayURL)
+				assert.Equal(t, tc.feederURL, tc.network.FeederURL.String())
+				assert.Equal(t, tc.gatewayURL, tc.network.GatewayURL.String())
```
ongyimeng: nit: Use `*url.URL` directly in test cases instead of converting back to string.

The production fields are now `*url.URL`, so the tests should compare URL values at the same abstraction level. Converting the actual value back to string keeps the assertion from the old string representation which weakens what this refactor is trying to verify.

```go
testCases := []struct {
    network    networks.Network
    feederURL  *url.URL
    gatewayURL *url.URL
}{
    {
        networks.Mainnet,
        parseURL(t, "https://feeder.alpha-mainnet.starknet.io/feeder_gateway/"),
        parseURL(t, "https://alpha-mainnet.starknet.io/gateway/"),
    },
}

func parseURL(t *testing.T, rawURL string) *url.URL {
    t.Helper()
    u, err := url.ParseRequestURI(rawURL)
    require.NoError(t, err)
    return u
}
```

clients/gateway/gateway.go
```diff
@@ -63,10 +64,9 @@ func (c *Client) WithListener(l EventListener) *Client {
 	return c
 }
 
-func NewClient(gatewayURL string, logger log.StructuredLogger) *Client {
-	gatewayURL = strings.TrimSuffix(gatewayURL, "/")
+func NewClient(gatewayURL *url.URL, logger log.StructuredLogger) *Client {
 	return &Client{
-		url: gatewayURL,
+		url: strings.TrimSuffix(gatewayURL.String(), "/"),
```
ongyimeng: nit: Client.url should probably be *url.URL too, since NewClient now accepts one.

The linked issue is about eliminating string to url.URL churn, but this still converts back to string at the boundary, so the refactor feels incomplete.

blockchain/networks/network.go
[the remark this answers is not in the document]
ongyimeng: this feedback was only addressed partially, predefined urls should also enforce scheme and host to be consistent.

cmd/juno/juno_test.go
```diff
@@ -980,8 +985,8 @@ func TestCustomNetworkURLValidation(t *testing.T) {
 				return
 			}
 			require.NoError(t, err)
-			assert.Equal(t, tc.feeder, config.Network.FeederURL)
-			assert.Equal(t, tc.gateway, config.Network.GatewayURL)
+			assert.Equal(t, tc.feeder, config.Network.FeederURL.String())
+			assert.Equal(t, tc.gateway, config.Network.GatewayURL.String())
```
ongyimeng: nit: same thing here, refer to earlier comment.

ongyimeng:
Thanks for your PR! 

I left a few comments to tighten the refactor so the tests, URL handling and validation all consistently follow the new `*url.URL` contract.

blockchain/networks/network_test.go
Ehsan-saradar: Done — test cases now hold `*url.URL` and assert on the value directly via the `parseURL` helper.

clients/gateway/gateway.go
Ehsan-saradar: Good point — `Client.url` is a `*url.URL` now, and the add_transaction path is built with `JoinPath` so there's no more string round-trip.

blockchain/networks/network.go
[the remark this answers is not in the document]
Ehsan-saradar: Done — `mustParseURL` now enforces the http/https scheme and a non-empty host too, so it's consistent with the custom-network checks.

cmd/juno/juno_test.go
Ehsan-saradar: Done here as well — same `parseURL` helper, comparing `*url.URL` directly.

blockchain/networks/network.go
```diff
@@ -195,3 +196,23 @@ func (n *Network) UnmarshalText(text []byte) error {
 func (n *Network) L2ChainIDFelt() *felt.Felt {
 	return new(felt.Felt).SetBytes([]byte(n.L2ChainID))
 }
+
+// mustParseURL parses a raw URL and panics on error. It is only used for the
+// hardcoded predefined network URLs, which are always valid. It enforces the
+// same constraints as custom-network URLs (http/https scheme, non-empty host)
+// so a malformed constant fails fast and diagnostically at startup.
+func mustParseURL(rawURL string) *url.URL {
```
rodrodros: Please keep the llm comments to the topic, they tend to overwrite and overexplain all the time.

```suggestion
// mustParseURL parses a raw URL and panics on error. 
```

ongyimeng (approved):
Thanks, LGTM!

PR #3770, opened 2026-06-28
fix(node): keep node running when a service exits without error

### LINKED ISSUE #3685: node: service closing without error shouldn't shut down the whole node
Follow-up from https://github.com/NethermindEth/juno/pull/3601#discussion_r3304634462.

## Problem

`node.Node.StartService` unconditionally `defer cancel()`s the node-level context as soon as a service's `Run` returns:

https://github.com/NethermindEth/juno/blob/main/node/node.go#L729-L744

```go
func (n *Node) StartService(
    wg *conc.WaitGroup, ctx context.Context, cancel context.CancelFunc, s service.Service,
) {
    wg.Go(func() {
        // Immediately acknowledge panicing services by shutting down the node
        // Without the deffered cancel(), we would have to wait for user to hit Ctrl+C
        defer cancel()
        if err := s.Run(ctx); err != nil {
            n.logger.Error(
                "Service error",
                zap.String("name", reflect.TypeOf(s).String()),
                zap.Error(err),
            )
        }
    })
}
```

The stated intent (per the inline comment) is to react to a panicking/failed service by tearing the whole node down. But the current implementation fires `cancel()` on **any** return, including a clean `nil`. That forces every `service.Service` implementation to keep `Run` blocking until `ctx.Done()` — even when it has nothing left to do — purely to avoid bringing the node down.

`MigrationFeeder.Run` (added in #3601) is an example of a service that has to artificially block on `<-ctx.Done()` for this reason:

https://github.com/NethermindEth/juno/blob/main/starknetdata/feeder/migration_feeder.go

## Proposal

Only trigger the node-wide cancel when `Run` returns an error (or panics). A service that completes its work and returns `nil` should be allowed to exit without taking the rest of the node with it.

Roughly:

```go
defer func() {
    if r := recover(); r != nil {
        // log + cancel
        cancel()
        panic(r) // or handle
    }
}()
if err := s.Run(ctx); err != nil {
    n.logger.Error("Service error", ...)
    cancel()
}
```

Care needs to be taken around services that are expected to run for the node's lifetime — for those, an early clean return is itself surprising and may still warrant logging, but not necessarily a full shutdown.

Ehsan-saradar:
Closes #3685.

`StartService` deferred `cancel()` on every service return, so a service that finished its work and returned `nil` would tear the whole node down. That forced short-lived services to artificially block on `<-ctx.Done()` just to stay alive.

Now the node is only shut down when a service returns an error or panics. A clean return lets the service exit on its own; if it happens before shutdown was requested (unexpected for a long-running service) we log a warning instead of stopping the node.

Added tests covering the three cases: error, panic, and clean exit.


node/node.go
```diff
@@ -736,15 +736,33 @@ func (n *Node) StartService(
...
-		// Without the deffered cancel(), we would have to wait for user to hit Ctrl+C
-		defer cancel()
+		name := reflect.TypeOf(s).String()
+
+		// A panicking service is critical: trigger a node-wide shutdown so the
+		// other services stop without waiting for the user to hit Ctrl+C, then
+		// re-panic so the WaitGroup can propagate it to the caller.
+		defer func() {
+			if r := recover(); r != nil {
+				n.logger.Error("Service panicked, shutting down node", zap.String("name", name), zap.Any("panic", r))
```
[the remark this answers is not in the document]
Ehsan-saradar: Good catch. The stack is now captured with `debug.Stack()` inside the recover, before the re-panic — at that point gopanic is still running our deferred func on top of the original panicking frames, so it points at the real panic site rather than the `wg.Wait()` re-panic. Added as a `stack` field on the log entry in 242f66528.

node/node.go
```diff
@@ -736,15 +736,33 @@ func (n *Node) StartService(
...
+			n.logger.Error("Service error, shutting down node", zap.String("name", name), zap.Error(err))
+			cancel()
+			return
+		}
+
+		// The service returned without an error. Let it exit on its own without
+		// taking the rest of the node down. A clean return before shutdown was
+		// requested is unexpected for a long-running service, so flag it.
+		if ctx.Err() == nil {
+			n.logger.Warn("Service stopped before node shutdown was requested", zap.String("name", name))
```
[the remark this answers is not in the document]
Ehsan-saradar: Agreed. Updated the `service.Service` doc to drop "Services should be blocking" and spell out the contract instead: a long-running service blocks until ctx is cancelled, a short-lived service may return nil once its work is done without shutting the node down, and an error/panic triggers a node-wide shutdown. Done in 242f66528.

node/start_service_test.go
```diff
@@ -0,0 +1,97 @@
+package node
```
rodrodros: Please use `package node_test`. It is important test node behaviour. The `_test` suffix forces you to only have acess to the public API when writing tests. 

Note that this doesn't mean you should change the public / private API and keep the tests the same but you need to update the tests so they are relevant but only through the use of public API. 

I believe after changing the package to `_test` the modifications to the test file should be minimal.

service/service.go
```diff
@@ -3,7 +3,12 @@ package service
...
-	// Run starts the service and returns an error. Services should not return non-critical errors,
-	// instead they should log and retry. Services should be blocking.
+	// Run starts the service. Services should not return non-critical errors,
+	// instead they should log and retry.
+	//
+	// A long-running service should block until ctx is cancelled and then return
+	// nil. A short-lived service may return nil once its work is done; a clean
+	// return does not shut the node down. Returning a (critical) error or
+	// panicking triggers a node-wide shutdown.
 	Run(ctx context.Context) error
```
rodrodros, line 10: Please reduce to:

```suggestion
	// Run starts the service. Services should not return non-critical errors,
	// instead they should log and retry. If a service returns an error the node
	// shutdowns. If a service returns a nil error, it is assumed its work has 
	// finished safely.
```

node/node.go
```diff
@@ -745,15 +737,37 @@ func (n *Node) StartService(
...
+		}()
+
 		if err := s.Run(ctx); err != nil {
-			n.logger.Error(
-				"Service error",
-				zap.String("name", reflect.TypeOf(s).String()),
-				zap.Error(err),
-			)
+			// A service that returns an error is critical: bring the node down.
+			n.logger.Error("Service error, shutting down node", zap.String("name", name), zap.Error(err))
```
rodrodros: "..., shutting down node" is unnecessary since that will be announced by the node already when it shut downs.

node/node.go
```diff
@@ -745,15 +737,37 @@ func (n *Node) StartService(
...
+
+		// A panicking service is critical: trigger a node-wide shutdown so the
+		// other services stop without waiting for the user to hit Ctrl+C, then
+		// re-panic so the WaitGroup can propagate it to the caller.
+		defer func() {
+			if r := recover(); r != nil {
+				// Capture the stack here, while it still points at the original
+				// panic site: the WaitGroup re-panics from wg.Wait(), so the
+				// stack surfaced there would point at the re-panic instead.
+				n.logger.Error("Service panicked, shutting down node",
```
rodrodros: We don't need to log when a service panicks since it shouldn't ever happen.

node/node.go
```diff
@@ -745,15 +737,37 @@ func (n *Node) StartService(
...
+					zap.String("name", name), zap.Any("panic", r), zap.ByteString("stack", debug.Stack()))
+				cancel()
+				panic(r)
+			}
```
rodrodros: Current node panics keep the stack trace for us to debug. I don't think this diff is necessary

node/node_service_test.go
rodrodros, line 1: Please call this file `node_service_test.go`

node/node_service_test.go
```diff
@@ -0,0 +1,97 @@
...
+		defer cancel()
+
+		failing := &fakeService{run: func(context.Context) error {
+			return errors.New("boom")
+		}}
+		n.StartService(wg, ctx, cancel, failing)
+		wg.Wait()
+
+		require.Error(t, ctx.Err(), "a service error should shut the node down")
+	})
```
rodrodros, line 56: This tests interact with the services but they don't check anything, just that they node stopped. 

Please use the zap.Observer utility and then check that the right final error logs where logged, otherwise you've no way of checking node behaviour.

node/start_service_test.go
```diff
@@ -0,0 +1,97 @@
...
+
+		// Its clean exit must not cancel the node-wide context...
+		require.NoError(t, ctx.Err(), "a clean service exit should not shut the node down")
+		// ...and the long-running service must still be running.
+		select {
+		case <-stopped:
+			t.Fatal("long-running service stopped: node was shut down unexpectedly")
+		default:
+		}
+
```
rodrodros: These two checks seem redundant, since both depend on the context to be triggered. Delete the first one (`require.NoERrror(...)`) and leave the second.

rodrodros (commented):
Thanks for your contribution

node/start_service_test.go
Ehsan-saradar: Done — moved to `package node_test`. Used a small `export_test.go` seam (`NewWithLogger`) so the external test can build a logger-only Node without widening the public API.

service/service.go
Ehsan-saradar: Done — reduced to your suggested wording.

node/node.go
Ehsan-saradar: Done — dropped the suffix; the log is now just `Service error`.

node/node.go
Ehsan-saradar: Done — removed the panic log entirely.

node/node.go
Ehsan-saradar: Agreed — removed the manual `debug.Stack()` capture. The WaitGroup re-panics with the original stack, so it's already preserved for debugging.

node/node_service_test.go
Ehsan-saradar: Done — renamed to `node_service_test.go`.

node/node_service_test.go
Ehsan-saradar: Done — now asserting on the emitted logs via `zaptest/observer`: `Service error` on the error path, and the early-exit warning on a clean return.

node/start_service_test.go
Ehsan-saradar: Done — removed the `require.NoError` check and kept the channel-based one.

node/export_test.go
rodrodros, line 1: I don't understand why do you need this?

node/node_service_test.go
```diff
@@ -0,0 +1,109 @@
...
+// expected effect on the node-wide context:
+//   - error  -> shut the node down
+//   - panic  -> shut the node down and propagate the panic
+//   - nil    -> let the service exit without taking the node down
+func TestStartService(t *testing.T) {
+	// newNode returns a node logging into an observer so tests can assert on the
+	// messages StartService emits.
+	newNode := func() (*node.Node, *observer.ObservedLogs) {
+		core, logs := observer.New(zapcore.DebugLevel)
+		return node.NewWithLogger(log.NewZapLoggerWithCore(core)), logs
```
rodrodros: Cannot you initialize the log here directly with the right logger?

node/node_service_test.go
```diff
@@ -0,0 +1,109 @@
...
+		<-exited // the short-lived service has returned cleanly
+
+		// Its clean exit must not take the long-running service down with it.
+		select {
+		case <-stopped:
+			t.Fatal("long-running service stopped: node was shut down unexpectedly")
+		default:
+		}
+		// And the early return should be flagged as unexpected.
+		const earlyExitMsg = "Service stopped before node shutdown was requested"
```
rodrodros: Why wrap this on a constant if it is used only once?

node/node_service_test.go
Ehsan-saradar: Yeah, the catch is `logger` is unexported, so from `package node_test` I can't set it on the struct directly. The only public way to get a Node with a logger is `New()`, but that spins up the DB, VM, mempool and all the services — way too heavy for a lifecycle test.

That's all `NewWithLogger` is for — it lives in `export_test.go`, so it's compiled only during tests and never becomes part of the package's actual API. It's the standard Go trick for testing from an external `_test` package while reaching one internal bit, and it's what keeps the test in `node_test` without changing the public/private API like you asked.

Happy to go the `New()` route instead if you'd rather, but it'd pull a lot of unrelated setup into the test. Let me know what you prefer.

node/node_service_test.go
Ehsan-saradar: Good point — inlined it.

node/node.go
```diff
@@ -736,15 +737,37 @@ func (n *Node) StartService(
...
+		// A panicking service is critical: trigger a node-wide shutdown so the
+		// other services stop without waiting for the user to hit Ctrl+C, then
+		// re-panic so the WaitGroup can propagate it to the caller.
+		defer func() {
+			if r := recover(); r != nil {
+				// Capture the stack here, while it still points at the original
+				// panic site: the WaitGroup re-panics from wg.Wait(), so the
+				// stack surfaced there would point at the re-panic instead.
+				n.logger.Error("Service panicked, shutting down node",
+					zap.String("name", name), zap.Any("panic", r), zap.ByteString("stack", debug.Stack()))
```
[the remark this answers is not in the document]
Ehsan-saradar: This is moot now — the panic log was dropped entirely in a later commit (we agreed it shouldn't ever fire), so there's no more `debug.Stack()` to encode. The recover path just cancels and re-panics, and the WaitGroup surfaces the original stack to the caller.

node/node_service_test.go
rodrodros: Yes no `NewWithLogger` please, use the available public API to do your tests

node/node.go
```diff
@@ -736,15 +736,33 @@ func (n *Node) StartService(
 	wg *conc.WaitGroup, ctx context.Context, cancel context.CancelFunc, s service.Service,
 ) {
 	wg.Go(func() {
-		// Immediately acknowledge panicing services by shutting down the node
-		// Without the deffered cancel(), we would have to wait for user to hit Ctrl+C
-		defer cancel()
+		name := reflect.TypeOf(s).String()
+
```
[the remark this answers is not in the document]
Ehsan-saradar: Good catch, thanks! Moved the recover defer to the top so it's registered first, and now the name is only computed when `s` isn't nil. A nil service gets caught by the recover, cancels the node and re-panics cleanly. Fixed.

node/export_test.go
Ehsan-saradar: Removed — no longer needed now that the tests build the node with `node.New()`.

node/node_service_test.go
Ehsan-saradar: Done — dropped `NewWithLogger` and `export_test.go`. The tests now build the node with `node.New()` (same minimal offline config as `TestNetworkVerificationOnNonEmptyDB`) and assert on behaviour: an error cancels the node context, a panic propagates, and a clean early exit leaves the long-running service untouched. One trade-off: `New()` builds its logger internally, so the observer-based log assertions had to go — happy to bring them back another way if you think they're worth it.

node/node_service_test.go
Ehsan-saradar: Small update: with the move to `node.New()` there's no public way to inject an observer, so these log assertions became behaviour checks — error cancels the node context, a clean exit leaves the other service running.

node/node_service_test.go
```diff
@@ -0,0 +1,111 @@
...
+			<-ctx.Done()
+			close(stopped)
+			return nil
+		}}
+
+		n.StartService(wg, ctx, cancel, longRunning)
+		n.StartService(wg, ctx, cancel, shortLived)
+
+		<-exited // the short-lived service has returned cleanly
+
```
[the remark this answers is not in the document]
Ehsan-saradar: Good point — added a bounded wait on both channel reads so a regression fails fast with a clear message instead of hanging the whole suite.

node/node.go
```diff
@@ -736,15 +736,36 @@ func (n *Node) StartService(
 	wg *conc.WaitGroup, ctx context.Context, cancel context.CancelFunc, s service.Service,
 ) {
 	wg.Go(func() {
-		// Immediately acknowledge panicing services by shutting down the node
-		// Without the deffered cancel(), we would have to wait for user to hit Ctrl+C
-		defer cancel()
+		// A panicking service is critical: trigger a node-wide shutdown so the
+		// other services stop without waiting for the user to hit Ctrl+C, then
+		// re-panic so the WaitGroup can propagate it (with its stack) to the
+		// caller. Registered first so it also covers a nil service.
```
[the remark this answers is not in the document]
Ehsan-saradar: The stack actually survives here: conc's Catcher grabs debug.Stack() at recover time inside the same goroutine, while the original panicking frames are still on the stack, and Wait() re-panics with that RecoveredPanic. So the trace still points at the real panic site — keeping the comment as is.

Ehsan-saradar:
Hey @rodrodros, ready for another pass whenever you get a chance — all threads are addressed (tests moved to `package node_test` building the node with `node.New()`, dropped `NewWithLogger`/`export_test.go`, and the behaviour assertions are in). Anything still open on your side, or good to merge?

node/node.go
```diff
@@ -752,15 +752,36 @@ func (n *Node) StartService(
...
+		}
+
+		// The service returned without an error. Let it exit on its own without
+		// taking the rest of the node down. A clean return before shutdown was
+		// requested is unexpected for a long-running service, so flag it.
+		if ctx.Err() == nil {
+			n.logger.Warn("Service stopped before node shutdown was requested", zap.String("name", name))
+		} else {
+			n.logger.Debug("Service stopped", zap.String("name", name))
 		}
```
rodrodros: Remove this warns and debugs. Replace with an Info: 

```golang
n.logger.Info("Service finished", zap.String("name", name) 
```

rodrodros (commented):
Looking good, please address my final comment and it is good to go 👍

node/node.go
Ehsan-saradar: Done — replaced with the single Info line in bceccc2d3.

PR #3771, opened 2026-06-28
fix(l1): retry transient chain ID errors instead of shutting down the node

### LINKED ISSUE #1385: L1 Verifier shouldn't shutdown the node when it encounters an error

Ehsan-saradar:
## What & why

A transient L1 RPC error during the startup chain-ID check would take the **whole node** down. Per #1385: `eth_chainId` returns `daily request count exceeded, request rate limited`, the L1 client's `Run` returns that error, and since `node.StartService` runs each service with `defer cancel()`, any service's `Run` returning shuts the node down. This also violated the `service.Service` contract, which asks services to log and retry non-critical errors rather than return them.

## What this changes

- **Retry transient failures.** A new `verifyChainID` (used only by `Run`) retries transient chain-ID failures — rate limits, timeouts, an unresponsive node — as warnings, reusing the client's existing resubscribe-delay loop (same pattern as `subscribeToUpdates` and `finalisedHeight`), until the check passes or `ctx` is cancelled. Retries are unbounded on purpose: a cap would just return an error and re-trigger the shutdown this fixes.
- **Keep genuine mismatches fatal.** A real L1/L2 network mismatch is a misconfiguration retrying can't fix, so it stays fatal (now a typed `chainIDMismatchError`, same user-facing message).
- **`checkChainID` stays single-attempt** — the one-shot `CatchUpL1Head` migration path relies on it failing fast.

After this change no transient L1 RPC error can shut the node down: `catchUpL1HeadUpdates` is already best-effort, `subscribeToUpdates`/`finalisedHeight` already retry forever, and the only remaining fatal paths are a genuine mismatch and a local DB-write failure in `setL1Head` — both legitimately critical.

## Relationship to #3685

Complementary, not overlapping: #3685 makes `StartService` tolerate a service returning `nil`, but an *error* return still stops the node. Transient chain-ID failures are errors, so the L1 client must stop emitting them regardless. No dependency on #3685.

## Tests

`go test ./l1/` passes. New `TestTransientChainIDErrorDoesNotShutDownNode` reproduces the issue's rate-limit scenario and asserts the node retries instead of shutting down (fails on the old code). `TestChainIDCheckTimeout` / `TestChainIDFetchError` were re-pointed at the fail-fast `CatchUpL1Head` path (same error-message assertions); the mismatch tests are unchanged.

Fixes #1385


l1/l1.go
```diff
@@ -143,7 +143,59 @@ func (c *Client) subscribeToUpdates(
...
+			}
+
+			// Transient: warn and retry, unless we're already shutting down.
+			if ctx.Err() != nil {
+				return nil
+			}
+			c.logger.Warn("Failed to verify L1 chain ID; retrying",
+				zap.Duration("tryAgainIn", c.resubscribeDelay),
+				zap.Error(err),
+			)
```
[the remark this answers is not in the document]
Ehsan-saradar: Good catch — fixed. The Warn now logs only the high-level message; the detailed error (which can include the L1 URL/API key) is moved to Debug, matching the existing convention in the resubscribe loop a few lines below.

l1/l1.go
```diff
@@ -165,18 +218,12 @@ func (c *Client) checkChainID(ctx context.Context) error {
...
-	)
+	return &chainIDMismatchError{network: c.network.String()}
 }
 
 func (c *Client) Run(ctx context.Context) error {
 	defer c.l1.Close()
-	if err := c.checkChainID(ctx); err != nil {
+	if err := c.verifyChainID(ctx); err != nil {
 		return err
 	}
```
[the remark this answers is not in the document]
Ehsan-saradar: Good catch — added an early `ctx.Err()` check after `verifyChainID` so we return cleanly on shutdown instead of starting catch-up.

brbrr (approved):
LGTM! Thanks for the contribution!

l1/l1.go
```diff
@@ -143,7 +143,60 @@ func (c *Client) subscribeToUpdates(
...
+
+			var mismatch *chainIDMismatchError
+			if errors.As(err, &mismatch) {
+				return err
+			}
+
+			// Transient: warn and retry, unless we're already shutting down.
+			if ctx.Err() != nil {
+				return nil
+			}
```
rodrodros, line 187: This `ctx.Err()` is not required since `ctx.Done()` is checked two instructions after

l1/l1.go
```diff
@@ -143,7 +143,60 @@ func (c *Client) subscribeToUpdates(
...
+			// err may contain the L1 URL (with API key), so keep it to Debug.
+			c.logger.Warn("Failed to verify L1 chain ID; retrying",
+				zap.Duration("tryAgainIn", c.resubscribeDelay),
+			)
+			c.logger.Debug("L1 chain ID verification failed", zap.Error(err))
```
rodrodros, line 192: Why do we need to logs for the same thing? Why doesn't the warn share the error as well?

l1/l1.go
```diff
@@ -165,20 +218,19 @@ func (c *Client) checkChainID(ctx context.Context) error {
...
 	defer c.l1.Close()
-	if err := c.checkChainID(ctx); err != nil {
+	if err := c.verifyChainID(ctx); err != nil {
 		return err
 	}
+	// verifyChainID returns nil on ctx cancellation; don't start any further
+	// RPC work (which would just fail with context canceled) if we're stopping.
+	if ctx.Err() != nil {
+		return nil
+	}
```
rodrodros, line 233: What is this patch here, why does `verifyChainID` returns nil on context cancellation rather than returning the context error?

l1/l1.go
```diff
@@ -143,7 +143,60 @@ func (c *Client) subscribeToUpdates(
...
+	network string
+}
+
+func (e *chainIDMismatchError) Error() string {
+	return fmt.Sprintf(
+		"mismatched network id between L1 and L2. L2 network is %s; "+
+			"is --eth-node pointing to the right network?",
+		e.network,
+	)
+}
```
rodrodros, line 159: There is no need for a custom error as I can see. It is being used only in one place

l1/l1.go
```diff
@@ -143,7 +143,60 @@ func (c *Client) subscribeToUpdates(
...
+			// Transient: warn and retry, unless we're already shutting down.
+			if ctx.Err() != nil {
+				return nil
+			}
+			// err may contain the L1 URL (with API key), so keep it to Debug.
+			c.logger.Warn("Failed to verify L1 chain ID; retrying",
+				zap.Duration("tryAgainIn", c.resubscribeDelay),
+			)
+			c.logger.Debug("L1 chain ID verification failed", zap.Error(err))
+			timer.Reset(c.resubscribeDelay)
```
rodrodros, line 193: Why resubscribe delay is used for separating between requests?

Ehsan-saradar:
@rodrodros thanks, these all make sense. Since the PR is already merged, do you want me to open a follow-up issue/PR for these, or are they minor enough to leave as-is?

PR #3774, opened 2026-06-30
refactor(feeder): return values instead of pointers from client methods

### LINKED ISSUE #3686: Make `clients/feeder/feeder.go` methods return by value instead of pointers
- Make methods return by value instead of pointers
- Change the `doRequest` utility accordingly

Ehsan-saradar:
## Description

Closes #3686.

Makes the `clients/feeder` `Client` methods (and the shared `doRequest` helper) return values instead of pointers, and updates the `Reader` interface and all call sites accordingly.

### What changed
- `doRequest[T, V]` now returns `(T, error)` instead of `(*T, error)`.
- All `Client` methods that previously returned `*starknet.X` now return `starknet.X`; the `Reader` interface signatures match.
- `BlockHeader` and `FeeTokenAddresses`, which previously dereferenced the pointer, are simplified to return the `doRequest` result directly.
- `CasmClassDefinition` keeps its deprecated-class semantics: it still signals "no compiled class" via `ErrDeprecatedCompiledClass`, and consumers pass `nil` to the adapter on that path instead of relying on a nil pointer return.
- Consumers updated to take the address locally where a pointer is still required: the `starknetdata/feeder` adapter, the `rpc/v8|v9|v10` trace/transaction handlers, and the affected tests.
- `mocks/mock_feeder.go` regenerated.

### Notes
- `PublicKey` now also returns `felt.Felt` by value (changed during review — it originally kept the pointer return, but there was no reason for the exception). `PreConfirmedBlockWithIdentifier` still returns the `starknet.PreConfirmedUpdate` interface, which is unchanged.

### Testing
- `golangci-lint` (diff + full): clean
- `make lint`: pass
- `go test ./...`: pass

### DIFF
--- a/clients/feeder/feeder.go
+++ b/clients/feeder/feeder.go
@@ -64,7 +64,7 @@ type Reader interface {
 		blockIdentifier string,
 		knownTransactionCount uint64,
 	) (starknet.PreConfirmedUpdate, uint64, error)
-	PublicKey(ctx context.Context) (*felt.Felt, error)
+	PublicKey(ctx context.Context) (felt.Felt, error)
 	Signature(ctx context.Context, blockID string) (starknet.Signature, error)
 	StateUpdate(ctx context.Context, blockID string) (starknet.StateUpdate, error)
 	StateUpdateWithBlockAndSignature(
@@ -277,24 +277,24 @@ func (c *Client) CasmClassDefinition(
 		blockNumberArg: "latest",
 	})
 
-	var class starknet.CasmClass
 	body, err := c.get(ctx, queryURL)
 	if err != nil {
-		return class, err
+		return starknet.CasmClass{}, err
 	}
 	defer body.Close()
 
 	definition, err := io.ReadAll(body)
 	if err != nil {
-		return class, err
+		return starknet.CasmClass{}, err
 	}
 
 	if deprecated, _ := starknet.IsDeprecatedCompiledClassDefinition(definition); deprecated {
-		return class, ErrDeprecatedCompiledClass
+		return starknet.CasmClass{}, ErrDeprecatedCompiledClass
 	}
 
+	var class starknet.CasmClass
 	if err = json.Unmarshal(definition, &class); err != nil {
-		return class, err
+		return starknet.CasmClass{}, err
 	}
 	return class, nil
 }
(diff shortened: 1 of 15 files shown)


clients/feeder/request.go
```diff
@@ -23,27 +23,27 @@ func doRequest[T any, V Validatable[T]](
...
 	}
 
 	err = V(&result).Validate()
 	if err != nil {
-		return nil, errors.Join(
+		return result, errors.Join(
 			ErrInvalidFeederResponse,
 			fmt.Errorf("querying %s: %w", fullURL, err),
 		)
 	}
```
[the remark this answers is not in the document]
Ehsan-saradar: Good catch, fixed. Now returning the zero value of T on both the decode and validation error paths so callers can never get partial data back.

clients/feeder/feeder.go
```diff
@@ -270,37 +266,37 @@ func (c *Client) BlockTrace(ctx context.Context, blockHash string) (*starknet.Bl
...
 	queryURL := buildQueryString(c.url, "get_compiled_class_by_class_hash", map[string]string{
 		classHashArg:   classHash.String(),
 		blockNumberArg: "latest",
 	})
 
+	var class starknet.CasmClass
 	body, err := c.get(ctx, queryURL)
 	if err != nil {
-		return nil, err
+		return class, err
```
rodrodros: return default values by returning `starknet.CasmClass{}`. Please apply this to all the methods that return by value

clients/feeder/feeder.go
```diff
@@ -312,11 +308,7 @@ func (c *Client) ClassDefinition(
...
 
-	addresses, err := doRequest[starknet.FeeTokenAddresses](ctx, c, queryURL)
-	if err != nil {
-		return starknet.FeeTokenAddresses{}, err
-	}
-	return *addresses, err
+	return doRequest[starknet.FeeTokenAddresses](ctx, c, queryURL)
 }
 
 func (c *Client) PublicKey(ctx context.Context) (*felt.Felt, error) {
```
rodrodros: This one should return by value too, is it possible?

rodrodros (commented):
LGTM, two small comments and good to merge

clients/feeder/feeder.go
Ehsan-saradar: Done — all error paths now return `starknet.CasmClass{}`. For the `doRequest`-based methods the helper itself returns the zero value of `T` on every error path, so this is consistent across all value-returning methods.

clients/feeder/feeder.go
Ehsan-saradar: Done — changed `PublicKey` to return `felt.Felt` by value (returns `felt.Felt{}` on the error path).

clients/feeder/feeder.go
```diff
@@ -46,35 +46,35 @@ type Client struct {
...
 		blockIdentifier string,
 		knownTransactionCount uint64,
 	) (starknet.PreConfirmedUpdate, error)
-	PublicKey(ctx context.Context) (*felt.Felt, error)
-	Signature(ctx context.Context, blockID string) (*starknet.Signature, error)
-	StateUpdate(ctx context.Context, blockID string) (*starknet.StateUpdate, error)
-	StateUpdateWithBlock(ctx context.Context, blockID string) (*starknet.StateUpdateWithBlock, error)
+	PublicKey(ctx context.Context) (felt.Felt, error)
+	Signature(ctx context.Context, blockID string) (starknet.Signature, error)
+	StateUpdate(ctx context.Context, blockID string) (starknet.StateUpdate, error)
```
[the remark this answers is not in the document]
Ehsan-saradar: Good catch — the description was stale. The change to `PublicKey` is intentional (requested during review), so I've updated the PR description to match the code.

rodrodros (approved):

PR #3776, opened 2026-06-30
fix(l1): address various issues in l1 chainID fetching logic

brbrr:
Addresses issues that I've missed when reviewing #3771

### DIFF
--- a/l1/l1.go
+++ b/l1/l1.go
@@ -148,14 +148,8 @@ func (c *Client) subscribeToUpdates(
 // forked Starknet networks would mean warning here instead of erroring.)
 var errChainIDMismatch = errors.New("mismatched network id between L1 and L2")
 
-// ensureChainID checks the L1 node is on the expected network, retrying transient
-// failures (rate limits, timeouts, an unresponsive node) until the check passes,
-// ctx is cancelled, or a network mismatch is found, so a flaky L1 never shuts the
-// node down (issue #1385). A mismatch is a misconfiguration and is returned fatally.
-//
-// The timer is reset after each failed probe, so a full resubscribeDelay always
-// elapses between the end of one attempt and the start of the next (gap is the
-// probe's response time + resubscribeDelay) — never back-to-back requests.
+// ensureChainID checks the L1 node is on the expected network, retrying on
+// transient failures until a valid response is received or context is canceled.
 func (c *Client) ensureChainID(ctx context.Context) error {
 	timer := time.NewTimer(0)
 	defer timer.Stop()
--- a/l1/l1.go
+++ b/l1/l1.go
@@ -143,54 +143,43 @@ func (c *Client) subscribeToUpdates(
 	}
 }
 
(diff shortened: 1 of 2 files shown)


l1/l1.go
```diff
@@ -143,54 +143,43 @@ func (c *Client) subscribeToUpdates(
...
 				return err
+			case ctx.Err() != nil:
+				// Cancelled mid-probe: return the real cause, don't warn.
+				return ctx.Err()
+			default:
+				c.logger.Warn("Failed to verify L1 chain ID, retrying",
+					zap.Duration("tryAgainIn", c.resubscribeDelay),
+					zap.Error(err),
+				)
+				timer.Reset(c.resubscribeDelay)
```
brbrr, line 175: We reset the timer after we get the response from L1 RPC. While it creates not even retry delays, it eliminates a possible back-to-back requests when L1 is super slow to respond. There is 30sec context on the RPC request, so chances are slim. But still, it feels better to reset the timer only after we get the response.

l1/l1.go
```diff
@@ -143,54 +143,43 @@ func (c *Client) subscribeToUpdates(
...
-			if errors.As(err, &mismatch) {
+			case errors.Is(err, errChainIDMismatch):
 				return err
+			case ctx.Err() != nil:
+				// Cancelled mid-probe: return the real cause, don't warn.
+				return ctx.Err()
+			default:
+				c.logger.Warn("Failed to verify L1 chain ID, retrying",
+					zap.Duration("tryAgainIn", c.resubscribeDelay),
+					zap.Error(err),
```
brbrr, line 173: Here, `err` might include RPC API key. Potentially, we could don't log it at all, detect/remove API key via regexp, or switch to Debug(which is non-default level, potentially safer). But honestly, I'm not sure that's justified here - if bad actor can access the logs, they probably can get the secret otherwise.

l1/l1.go
```diff
@@ -143,54 +143,43 @@ func (c *Client) subscribeToUpdates(
...
-func (c *Client) verifyChainID(ctx context.Context) error {
+// ensureChainID checks the L1 node is on the expected network, retrying transient
+// failures (rate limits, timeouts, an unresponsive node) until the check passes,
+// ctx is cancelled, or a network mismatch is found, so a flaky L1 never shuts the
+// node down (issue #1385). A mismatch is a misconfiguration and is returned fatally.
+//
+// The timer is reset after each failed probe, so a full resubscribeDelay always
+// elapses between the end of one attempt and the start of the next (gap is the
+// probe's response time + resubscribeDelay) — never back-to-back requests.
+func (c *Client) ensureChainID(ctx context.Context) error {
```
rodrodros: Brevity in comments please, "encsureChainID checks the L1 node is on the expected network, retrying on transient failures until a valid response is received or context is canceled". 

You don't have to explain how the function works, that's why the code is there.

rodrodros (approved):
Looks good

PR #3779, opened 2026-07-01
refactor(starknet): replace BlockStatus string with typed enum

0xsamalt:
- Introduces BlockStatus as a typed uint8 enum in starknet/block.go with constants BlockPending, BlockAcceptedOnL1, BlockAcceptedOnL2, BlockRejected, and BlockPreConfirmed
- Adds UnmarshalText so the feeder JSON response deserialises directly into the enum and invalid status strings now return an error 
- Updates all comparison sites and test assertions to use the typed constants

### DIFF
--- a/starknet/block.go
+++ b/starknet/block.go
@@ -2,6 +2,7 @@ package starknet
 
 import (
 	"errors"
+	"fmt"
 
 	"github.com/NethermindEth/juno/core/felt"
 )
@@ -12,6 +13,33 @@ import (
 type BlockHeader struct {
 	Hash   *felt.Felt `json:"block_hash"`
 	Number uint64     `json:"block_number"`
+}
+type BlockStatus uint8
+
+const (
+	BlockPending BlockStatus = iota + 1
+	BlockAcceptedOnL1
+	BlockAcceptedOnL2
+	BlockRejected
+	BlockPreConfirmed
+)
+
+func (bs *BlockStatus) UnmarshalText(data []byte) error {
+	switch str := string(data); str {
+	case "ACCEPTED_ON_L2":
+		*bs = BlockAcceptedOnL2
+	case "ACCEPTED_ON_L1":
+		*bs = BlockAcceptedOnL1
+	case "REJECTED":
+		*bs = BlockRejected
+	case "PENDING":
+		*bs = BlockPending
+	case "PRE_CONFIRMED":
+		*bs = BlockPreConfirmed
+	default:
+		return fmt.Errorf("unknown BlockStatus %q", str)
+	}
+	return nil
 }
 
 // TODO: placeholder for now to avoid compiler errors. A proper validation
@@ -31,7 +59,7 @@ type Block struct {
 	ReceiptCommitment     *felt.Felt            `json:"receipt_commitment"`
 	StateDiffCommitment   *felt.Felt            `json:"state_diff_commitment"`
 	StateDiffLength       uint64                `json:"state_diff_length"`
-	Status                string                `json:"status"`
+	Status                BlockStatus           `json:"status"`
 	Transactions          []*Transaction        `json:"transactions"`
 	Timestamp             uint64                `json:"timestamp"`
 	Version               string                `json:"starknet_version"`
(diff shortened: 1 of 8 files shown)


Ehsan-saradar:
Nice one, typed status is much better than a bare string 👍

One thing I'd double-check first though: `UnmarshalText` now errors on any status that isn't one of the five, and since blocks are decoded straight through `json.Unmarshal`, an unknown status fails the whole block decode — not just that field. With the old `string` we accepted anything and only really cared about `PENDING` / `PRE_CONFIRMED`.

So if the feeder ever sends something not in the list (pretty sure `ABORTED` / `REVERTED` can show up on blocks), we'd fail to read that block and sync would get stuck on it.

Wdyt about mapping unknown values to a `BlockUnknown` instead of erroring? Keeps it harmless like before, since we only check for `PENDING` / `PRE_CONFIRMED` anyway. Or if we'd rather keep it strict, probably worth confirming the list covers everything the gateway can actually send.

0xsamalt:
Checked this against both the spec and Juno's own fixtures. `BLOCK_STATUS` is a closed enum (`PENDING` / `ACCEPTED_ON_L2` / `ACCEPTED_ON_L1` / `REJECTED`, plus `PRE_CONFIRMED` since RPC 0.9) and `ABORTED`/`REVERTED` aren't block statuses, they're `TXN_EXECUTION_STATUS` values scoped to individual transaction receipts.

Grepped our own feeder testdata to be sure this isn't just spec theory: `REVERTED` shows up in exactly one fixture, and it's a transaction file, never a block file. Across every block and state-update fixture we've captured from the real feeder gateway, block-level status has only ever been `PENDING`/`ACCEPTED_ON_L2`/`ACCEPTED_ON_L1`/`REJECTED`/`PRE_CONFIRMED`, no evidence the feeder has ever sent anything else at the block level.

So I don't think there's a real path that hits the default error case. Given that, I'd rather keep it strict if the feeder ever does send something unexpected here, that's a signal of an actual protocol change we want to know about loudly, not something we want to silently coerce into a `BlockUnknown` fallback and risk masking. Happy to add the fallback anyway if you'd still prefer defensive handling, but wanted to share what I found first.

Ehsan-saradar:
Fair enough — the fixture check convinced me, let's keep it strict. Only nit: could the UnmarshalText error include the offending value (e.g. `unknown block status %q`)? Makes a future protocol change point straight at itself. 👍

0xsamalt:
Already included... the default case returns `fmt.Errorf("unknown BlockStatus %q", str)`, so the error message carries the offending value.

0xsamalt:
Rebased onto main to resolve conflicts. A few refactors landed on main since this opened (subscription-cache→deduper, feeder client value-return change, pending-endpoint removal) that touched the same code paths. 

No change to the BlockStatus type or its logic; the rebase also required updating a handful of test call sites (`clients/feeder/feeder_test.go`, `sync/preconfirmed/chain_storage,poller_test.go`, `rpc/v9|v10/transaction_test.go`) that still used the raw "PRE_CONFIRMED" string instead of the typed constant, since those tests were touched by other PRs after this one opened. 

Verified with go build/go vet/go test across the affected packages.

0xsamalt:
All feedback's been addressed and I rebased onto main a few days ago to clear conflicts, should apply clean now. Ready for review whenever you get a chance.

starknet/block.go
```diff
@@ -13,6 +14,33 @@ type BlockHeader struct {
 	Hash   *felt.Felt `json:"block_hash"`
 	Number uint64     `json:"block_number"`
 }
+type BlockStatus uint8
+
+const (
+	BlockPending BlockStatus = iota + 1
+	BlockAcceptedOnL1
+	BlockAcceptedOnL2
+	BlockRejected
```
rodrodros, line 23: Rejected is not an actual status 👀

rodrodros (commented):
This PRs addresses the problem at half there are many missing enums such as `ABORTED` and `REVERTED` and a wrong one called `REJECTED`. 

This PR seems low effort and drains time from the team. Thanks for your work but thoughtless contribution that harm Juno are not accepted.

PR #3782, opened 2026-07-02
perf(state): decode only state root when opening head state

ongyimeng:
## Summary

Optimizes latest-block state reads by avoiding full block-header decoding when opening head state.

`HeadState()` only needs `Header.GlobalStateRoot` to construct the state reader, but previously called `GetBlockHeaderByNumber`, which materialized the full `Header`, including heavier unused fields such as `EventsBloom`.

This adds `GetGlobalStateRootByBlockNumber`, which decodes only `GlobalStateRoot` from the stored CBOR header map, and uses it in `HeadState()`.

## Benchmarks
I ran local benchmarks that exercise the public RPC handler paths with Pebble DB.

This trades a modest increase in allocation count for a much larger reduction in latency and allocated bytes.

`allocs/op` increases due to extra small map/field bookkeeping during CBOR subset-struct decoding, visible in the allocation profile under `fxamacker/cbor.(*decoder).parseMapToStruct` at 300MB flat / 335.52MB cumulative. `B/op` still drops by about 50% because large unused fields are no longer materialized.

| Path                | Base     | New     | Latency  | B/op                          | allocs/op        |
|---------------------|---------:|--------:|---------:|-------------------------------:|-----------------:|
| `NonceLatest`       | ~12.75µs | ~7.64µs | **−40%** | ~4.0KiB → ~2.0KiB (**−50%**)  | 60 → 72 (**+20%**) |
| `ClassHashAtLatest` | ~12.3µs  | ~6.7µs  | **−46%** | ~4.0KiB → ~2.0KiB (**−50%**)  | 60 → 72 (**+20%**) |
| `StorageAtLatest`   | ~16.5µs  | ~8.9µs  | **−46%** | ~4.2KiB → ~2.0KiB (**−52%**)  | 64 → 76 (**+19%**) |

### DIFF
--- a/core/accessors.go
+++ b/core/accessors.go
@@ -2,6 +2,7 @@ package core
 
 import (
 	"encoding/binary"
+	"errors"
 	"iter"
 
 	"github.com/NethermindEth/juno/core/felt"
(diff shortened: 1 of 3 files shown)


rodrodros (approved):

rodrodros:
@ongyimeng please correct the first two nits of CC before merging

rodrodros:
Great work!

core/accessors.go
```diff
@@ -325,6 +326,24 @@ func GetBlockHeaderByNumber(r db.KeyValueReader, blockNum uint64) (*Header, erro
 	return header, err
 }
 
+// Only materialise GlobalStateRoot from the header,
```
brbrr: nit: to make the lsp a bit happier.

```suggestion
// GetGlobalStateRootByBlockNumber only materialise GlobalStateRoot from the header,
```

brbrr (approved):
Looks good!

PR #3790, opened 2026-07-03
refactor(v10): check storage existence only for zeros

brbrr:
Bench results:
(Generated using sepolia snapshot, 1 block temp db wasn't representative, as the data was hot, and results were marginal)

Note that only for the missing contract on the latest block we would face degradation. 

| Reader | Scenario | main sec/op | branch sec/op | delta time | main allocs | branch allocs | Δ allocs |
|---|---|---|---|---|---|---|---|
| latest | non_zero | 38.45µs | 24.74µs | -35.7%  | 81 | 70 | -13.6% |
| latest | zero_exists | 58.16µs | 57.08µs | ~  | 86 | 86 | ~ |
| latest | missing_contract | 21.71µs | 49.65µs | +128.7%  | 58 | 88 | +51.7% |
| historical | non_zero | 112.0µs | 59.85µs | -46.6% | 120 | 88 | -26.7% |
| historical | zero_exists | 129.5µs | 81.69µs | -36.9% | 124 | 93 | -25.0% |
| historical | missing_contract | 20.78µs | 21.96µs | ~  | 55 | 55 | ~ |

[Updated benchmark results](https://github.com/NethermindEth/juno/pull/3790#issuecomment-4982068896)

### DIFF
--- a/core/trie/trie.go
+++ b/core/trie/trie.go
@@ -47,14 +47,14 @@ func NewTrieReaderPoseidon(
 }
 
 // maxKeyByHeight[h] is 2^h - 1, the largest key a trie of height h can hold.
-var maxKeyByHeight = func() [256]*felt.Felt {
-	var table [256]*felt.Felt
+var maxKeyByHeight = func() [felt.Bits + 1]felt.Felt {
+	var table [felt.Bits + 1]felt.Felt
 	for h := range table {
 		x := felt.NewFromUint64[felt.Felt](2)
 		y := new(big.Int).SetUint64(uint64(h))
 		maxKey := x.Exp(x, y)
 		maxKey.Sub(maxKey, &felt.One)
-		table[h] = maxKey
+		table[h] = *maxKey
 	}
 	return table
 }()
@@ -66,29 +66,17 @@ func GetLeafPedersen(
 	height uint8,
 	key *felt.Felt,
 ) (felt.Felt, error) {
-	reader, err := newLeafReaderPedersen(r, prefix, height)
-	if err != nil {
-		return felt.Felt{}, err
-	}
-	return reader.Get(key)
-}
-
-// newLeafReaderPedersen builds a read-only reader that skips loading the root
-// key. Only Get is valid on it
-func newLeafReaderPedersen(
-	r db.KeyValueReader,
-	prefix []byte,
-	height uint8,
-) (TrieReader, error) {
 	if height > felt.Bits {
-		return TrieReader{}, fmt.Errorf("max trie height is %d, got: %d", felt.Bits, height)
+		return felt.Felt{}, fmt.Errorf("max trie height is %d, got: %d", felt.Bits, height)
 	}
-	return TrieReader{
+	reader := TrieReader{
 		readStorage: NewReadStorage(r, prefix),
 		height:      height,
-		maxKey:      maxKeyByHeight[height],
+		maxKey:      &maxKeyByHeight[height],
 		hash:        crypto.Pedersen,
-	}, nil
+	}
+
+	return reader.Get(key)
 }
 
 func newTrieReader(
@@ -107,12 +95,11 @@ func newTrieReader(
 		return TrieReader{}, err
 	}
 
-	maxKey := maxKeyByHeight[height]
 	return TrieReader{
 		readStorage: readStorage,
 		height:      height,
 		rootKey:     rootKey,
-		maxKey:      maxKey,
+		maxKey:      &maxKeyByHeight[height],
 		hash:        hash,
 	}, nil
 }
--- a/core/trie/trie.go
+++ b/core/trie/trie.go
@@ -46,6 +46,51 @@ func NewTrieReaderPoseidon(
 	return newTrieReader(r, prefix, height, crypto.Poseidon)
 }
 
+// maxKeyByHeight[h] is 2^h - 1, the largest key a trie of height h can hold.
+var maxKeyByHeight = func() [256]*felt.Felt {
+	var table [256]*felt.Felt
+	for h := range table {
+		x := felt.NewFromUint64[felt.Felt](2)
+		y := new(big.Int).SetUint64(uint64(h))
+		maxKey := x.Exp(x, y)
+		maxKey.Sub(maxKey, &felt.One)
+		table[h] = maxKey
+	}
+	return table
+}()
+
+// GetLeafPedersen reads a single leaf value by key from the flat Pedersen trie at prefix
+func GetLeafPedersen(
+	r db.KeyValueReader,
+	prefix []byte,
+	height uint8,
+	key *felt.Felt,
+) (felt.Felt, error) {
+	reader, err := newLeafReaderPedersen(r, prefix, height)
+	if err != nil {
+		return felt.Felt{}, err
+	}
+	return reader.Get(key)
+}
+
+// newLeafReaderPedersen builds a read-only reader that skips loading the root
+// key. Only Get is valid on it
+func newLeafReaderPedersen(
+	r db.KeyValueReader,
+	prefix []byte,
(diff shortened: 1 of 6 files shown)


rpc/v10/storage.go
```diff
@@ -105,22 +105,29 @@ func (h *Handler) StorageAt(
...
-	// This checks if the contract exists because if a key doesn't exist in contract storage,
-	// the returned value is always zero and error is nil.
-	_, err := stateReader.ContractClassHash(addressFelt)
+	value, err := stateReader.ContractStorage(addressFelt, key)
 	if err != nil {
 		if errors.Is(err, db.ErrKeyNotFound) {
 			return nil, rpccore.ErrContractNotFound
 		}
-		h.logger.Error("Failed to get contract class hash", zap.Error(err))
 		return nil, rpccore.ErrInternal.CloneWithData(err)
```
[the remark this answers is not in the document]
rodrodros: I think it is a good idea to log here as well

rodrodros (commented):
benchmark numbers please

rodrodros:
Looks good 👍 , you've good findings from Copilot and Claude

brbrr:
@rodrodros i've addressed the bot's feedback and added bench results to the description. please take a look.

RafaelGranza (commented):
I consider this PR to be in good shape and have all previous suggestions addressed.
Still, I'll take a second look tomorrow.

rpc/v10/storage.go
```diff
@@ -105,22 +105,29 @@ func (h *Handler) StorageAt(
...
+		h.logger.Error("Failed to get contract storage", zap.Error(err))
 		return nil, rpccore.ErrInternal.CloneWithData(err)
 	}
 
-	result.Value, err = stateReader.ContractStorage(addressFelt, key)
-	if err != nil {
-		return nil, rpccore.ErrInternal.CloneWithData(err)
+	// Only head readers return zero (not ErrKeyNotFound) for a missing contract,
+	// so a zero value needs an existence probe there; historical readers don't.
+	if value.IsZero() && (id.IsLatest() || id.IsPreConfirmed()) {
```
RafaelGranza: The `id.IsPreConfirmed()` is a little bit annoying here.

## Why I find it annoying 
You see, a `pre_confirmed` state is inconsistent to whether or not it uses a `historical` or `head` reader.

The CI Bot told us it always falls back to a `head` one, but that is not true. Actually, it almost always falls back to a `historical` one, the only way for it to use a `head` one is when the block is at genesis.

You can check that a `pre_confirmed` state is set by `baseState` on `sync/preconfirmed/chain_storage.go`, and the internal methods varies depending whether the block is near the 0th state (genesis).

## The problem it creates

It triggers unnecessary extra lookups.

We could fix this by making the genesis base in baseState return ErrKeyNotFound for a missing contract, like the historical base already does. then pre_confirmed is consistent everywhere and we don't touch the global reader.

But then we would need extra tests to ensure this behavior stays the same. Since someone could try to optimize `pre_confirmed` always using head readers.   

## Benchmarks to prove the problem exists

This is a **plausibility check**, I used a shallow chain (mainnet block 2, in-memory DB, so absolute latencies aren't production-representative), but the structural takeaway holds regardless of depth.

We can see that `preconfirmed_mature` matches the `historical`.

Reader | storage_only | storage + probe | delta
-- | -- | -- | --
historical | 60.3µs / 380 allocs | 120.6µs / 778 allocs | +100% / +398 allocs
preconfirmed_mature | 60.1µs / 380 allocs | 120.4µs / 778 allocs | +100% / +398 allocs
latest (head, hot) | 0.84µs / 13 allocs | 0.99µs / 16 allocs | +18% / +3 allocs

## What I expect before approving

I expect to hear your opinion and see one of these two:

1. Keep things as-is, and add a comment above the guard, to give more context about the `IsPreConfirmed` thing. Maybe the `pre_cofirmed` case is rare and the change does not pay off.

2. Actual fix:
- Remove `IsPreConfirmed`
- make the genesis base in `baseState` (`sync/preconfirmed/chain_storage.go`) return `ErrKeyNotFound` for a missing contract, like the historical base does.
- Benchmark the mature `pre_confirmed` to make sure it makes sense.
- create tests to lock that `pre_confirmed` with a missing contract returns `ErrContractNotFound`, for both the mature and the genesis cases. this is what stops someone from later optimizing `pre_confirmed` to always use head readers and breaking it silently.

RafaelGranza (commented):
It is a good optimization, but I noticed something strange about the `pre_confirmed` reader that the CI bot also let pass.

Please see my comment and give me feedback on the direction you will take.

Sorry for digging so deep, but I took my time to better understand these `readers`, `states` and block types.

rodrodros:
@brbrr why there is a case that worsens on performance, shouldn't everything be an improvement?

brbrr:
@rodrodros This PR flips the data fetching approach from defensive (check contract existence, and only then access the storage slot) to optimistic (get the storage slot, and if it fails, figure out why - contract or storage key missing). 

in the case of `missing_contract` case, we ending up doing 2 db operations vs 1 as before. 

This is PR builds up on the assumption that happy path requests are dominant.

brbrr:
here's bench results for the latest commit:

| Reader | Scenario | main sec/op | branch sec/op | Δ time | main allocs | branch allocs | Δ allocs |
|---|---|---|---|---|---|---|---|
| latest | non_zero | 31.72µs | 2.07µs | -93.5% | 47 | 19 | -59.6% |
| latest | zero_exists | 51.68µs | 33.59µs | -35.0% | 52 | 35 | -32.7% |
| latest | missing_contract | 14.68µs | 29.85µs | +103.4% | 24 | 37 | +54.2% |
| historical | non_zero | 105.81µs | 38.79µs | -63.3% | 123 | 74 | -39.8% |
| historical | zero_exists | 133.41µs | 61.11µs | -54.2% | 127 | 79 | -37.8% |
| historical | missing_contract | 16.36µs | 16.46µs | ~ | 58 | 58 | ~ |

core/trie/trie.go
```diff
@@ -46,6 +46,51 @@ func NewTrieReaderPoseidon(
 	return newTrieReader(r, prefix, height, crypto.Poseidon)
 }
 
+// maxKeyByHeight[h] is 2^h - 1, the largest key a trie of height h can hold.
+var maxKeyByHeight = func() [256]*felt.Felt {
```
rodrodros: Why 256, isn't `globalTrieHeight  = 251`?

core/trie/trie.go
```diff
@@ -46,6 +46,51 @@ func NewTrieReaderPoseidon(
...
+	var table [256]*felt.Felt
```
rodrodros: Why a pointer to felt rather than felt?

thiagodeev (approved):

core/trie/trie.go
```diff
@@ -46,6 +46,51 @@ func NewTrieReaderPoseidon(
...
+	reader, err := newLeafReaderPedersen(r, prefix, height)
+	if err != nil {
+		return felt.Felt{}, err
+	}
+	return reader.Get(key)
+}
+
+// newLeafReaderPedersen builds a read-only reader that skips loading the root
+// key. Only Get is valid on it
+func newLeafReaderPedersen(
```
rodrodros: Why are you adding a function that it is very short and only used once?

brbrr: i've been following the codebase patterns. I'll remove it.

core/trie/trie.go
brbrr: good call. I'll use the constant instead.

brbrr: actually, it should be `felt.Bits + 1`

EgeCaner (approved):
Looks good to me

PR #3792, opened 2026-07-03
perf(felt): add CBOR fast path for encode and decode

ongyimeng:
## Summary

`Felt.MarshalCBOR`/`UnmarshalCBOR` used to go through a generic reflection-based CBOR codec.
A `Felt` is always 4 `uint64` limbs, so this PR encodes/decodes that exact shape directly, and falls back to the generic codec for anything else.

## Benchmark

Single Felt, no DB:

  | before | after | Δ
-- | -- | -- | --
Marshal | ~100-110ns | ~27-36ns | -70%
Unmarshal | ~175-195ns | ~8-19ns | -93%

<br class="Apple-interchange-newline">
Real mainnet blocks through pebble:

Benchmark | before | after | Δ
-- | -- | -- | --
StateUpdate p50 | 51.4µs | 32.8µs | -36.2%
StateUpdate p95 | 415.1µs | 230.5µs | -44.5%
StateUpdate p99 | 630.9µs | 354.1µs | -43.9%
Block p50 | 516.3µs | 366.9µs | -28.9%
Block p99 | 1.10ms | 686.2µs | -37.6%
Block p95 | 2.10ms | 1.08ms | -48.6%

`n=8`, `p≤0.01` everywhere.

Allocation count unchanged (1 for Marshal, 0 for Unmarshal). Byte size unchanged except tiny Marshal inputs, which now allocate a fixed 48-byte buffer instead of one sized to the value.

## Testing

- Fuzz + 1B random values against the generic codec: 0 mismatches.
- Replayed from block 0 to ~2.1M real mainnet blocks (state diffs, receipts, classes) against a synced DB copy: 0 mismatches.


core/felt/cbor.go
```diff
@@ -0,0 +1,132 @@
...
+	cborUint32AdditionalInfo = 26
+	cborUint64AdditionalInfo = 0x1b
+
+	// cborArrayHeader4 is the CBOR array header byte for an array of Limbs items.
+	cborArrayHeader4 = 0x80 | Limbs
+	maxCBORUintLen   = 1 + 8
+)
+
+func encodeFeltLimbs(element *fp.Element) []byte {
+	buffer := make([]byte, 1, 1+Limbs*maxCBORUintLen)
```
thiagodeev: The original implementation has a buffer pool that caches unused bytes.Buffer objects for later reuse. Could you implement and benchmark it please?
[ref](https://github.com/fxamacker/cbor/blob/b9c41c670e13312be5cab17ccf9076bbdbaa9e89/encode.go#L1037), [ref](https://github.com/fxamacker/cbor/blob/b9c41c670e13312be5cab17ccf9076bbdbaa9e89/encode.go#L1070-L1086)

core/felt/cbor.go
```diff
@@ -0,0 +1,132 @@
...
+
+	case value <= math.MaxUint32:
+		buffer = append(buffer, cborUint32AdditionalInfo)
+		return binary.BigEndian.AppendUint32(buffer, uint32(value))
+
+	default:
+		buffer = append(buffer, cborUint64AdditionalInfo)
+		return binary.BigEndian.AppendUint64(buffer, value)
+	}
+}
```
thiagodeev: Most of the felts that we handle (tx and block hashes, addresses, etc) are full felts, which means they will need to go through all the cases until they finally got into the `default` case, which is the one that handles them. So, it's 4 Limbs x 4 dead cases, resulting in 16 dead conditions.
WDYT of reversing it? Starting with `case value > math.MaxUint32:`, then `case value > math.MaxUint16:` and so on, so we handle the full felt in the hot-path.
Could you please implement and benchmark it? 

Something similar can be done for the `decodeFeltLimbs` func as well

core/felt/cbor.go
```diff
@@ -0,0 +1,132 @@
...
+		buffer = append(buffer, cborUint16AdditionalInfo)
+		return binary.BigEndian.AppendUint16(buffer, uint16(value))
+
+	case value <= math.MaxUint32:
+		buffer = append(buffer, cborUint32AdditionalInfo)
+		return binary.BigEndian.AppendUint32(buffer, uint32(value))
+
+	default:
+		buffer = append(buffer, cborUint64AdditionalInfo)
+		return binary.BigEndian.AppendUint64(buffer, value)
```
thiagodeev: Still reviewing  the code, but could we replace `append()` with `copy()` without affecting the output? It would be slightly faster

thiagodeev (commented):
Nice work!
Partial review with some probable improvements

core/felt/cbor.go
RafaelGranza: We don't need this buffer anymore, because we already know the final size of the encoded output beforehand.

Benchmark-wise, I could implement it, but it has no realistic way of being faster, since we'd only be adding indirection.

RafaelGranza: Anyway, I implemented it and benchmarked:

| | direct alloc (current) | pooled scratch buffer | delta |
|---|---:|---:|---:|
| Marshal/zero | 22.4ns | 27.2ns | +22% |
| Marshal/nonzero | 32.8ns | 43.3ns | +32% |

Allocation count is identical in both.

I'll not push this measurement code, since I believe it makes no sense.

core/felt/cbor.go
RafaelGranza: Applied it. Real numbers (at this scale we're close to what a microbenchmark can even tell apart from noise):

| | before | after | delta |
|---|---:|---:|---:|
| Marshal/zero | 23.16ns | 32.70ns | +41% |
| Marshal/nonzero | 34.76ns | 25.43ns | -27% |
| Unmarshal/zero | 7.65ns | 11.88ns | +55% |
| Unmarshal/nonzero | 14.43ns | 8.41ns | -42% |

Real Felts (hashes, addresses) are basically never zero once in Montgomery form, so nonzero is the common case and this is a net win. Not shipping the benchmark file itself though, just the reordered switch.

core/felt/cbor.go
RafaelGranza: Tried it two ways, both slower:

| | append() (current) | PutUint* direct on slice | copy() from temp array | delta (vs append) |
|---|---:|---:|---:|---:|
| Marshal/zero | 30.99ns | 33.47ns | 34.88ns | +8% / +13% |
| Marshal/nonzero | 25.43ns | 25.50ns | 30.60ns | +0% / +20% |

Checked byte-identical output against 200k random values first, so it's not a correctness thing, just slower.

My guess:
`append()` with a small constant number of args is a pattern the compiler already knows well, while re-slicing manually plus calling `PutUint16/32/64` or a copy adds a second bounds check plus a function call that doesn't get optimized away as easily.

RafaelGranza: This reminds me of the **C++ Core Guidelines** on performance:


> ["Don't assume that complicated code is necessarily faster than simple code:"](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#per4-dont-assume-that-complicated-code-is-necessarily-faster-than-simple-code)
> Reason: Simple code can be very fast. Optimizers sometimes do marvels with simple code

Go's compiler isn't as aggressive as a modern C++/Rust one, but the idea holds up fine here too.

[Honestly the first ~7 items in that whole performance section](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#s-performance) are useful in any language, not just C++.

core/felt/cbor.go
```diff
@@ -0,0 +1,132 @@
...
+		return nil
+	}
+	return cbor.Unmarshal(data, (*fp.Element)(z))
+}
+
+const (
+	cborUint8AdditionalInfo  = 24
+	cborUint16AdditionalInfo = 25
+	cborUint32AdditionalInfo = 26
+	cborUint64AdditionalInfo = 0x1b
```
thiagodeev: Is there any reason not to set this as `27`?

core/felt/cbor_fastpath_test.go
```diff
@@ -0,0 +1,256 @@
...
+	}},
+}
+
+func genericUnmarshalCBOR(value *felt.Felt, data []byte) error {
+	return cbor.Unmarshal(data, (*fp.Element)(value))
+}
+
+func genericMarshalCBOR(value *felt.Felt) ([]byte, error) {
+	return cbor.Marshal((*fp.Element)(value))
+}
```
thiagodeev: WDYT of removing these helpers? They're used only once, so we could call `cbor.Unmarshal...` directly instead, just as you're doing `fast, err = value.MarshalCBOR()`, and not a separate method called `fastUnmarshalCBOR`

core/felt/cbor_fastpath_test.go
```diff
@@ -0,0 +1,256 @@
...
+		value := feltFromLimbs(rng.Uint64(), rng.Uint64(), rng.Uint64(), rng.Uint64())
+
+		fast, generic, err := encodeBoth(&value)
+		if err != nil {
+			t.Fatalf("iter %d: marshal error: %v", i, err)
+		}
+		if !bytes.Equal(fast, generic) {
+			t.Fatalf("iter %d: marshal mismatch for %v: fast=% x generic=% x",
+				i, value.Impl(), fast, generic)
+		}
```
thiagodeev: We can replace the if statements with require.NoError and assert/require.Equal

core/felt/cbor_fastpath_test.go
```diff
@@ -0,0 +1,256 @@
...
+
+		fastDecoded, genericDecoded, errFast, errGeneric := decodeBoth(fast)
+		if errFast != nil || errGeneric != nil {
+			t.Fatalf("iter %d: decode error: fast=%v generic=%v", i, errFast, errGeneric)
+		}
+		if !fastDecoded.Equal(&genericDecoded) {
+			t.Fatalf("iter %d: decode mismatch: fast=%s generic=%s",
+				i, fastDecoded.String(), genericDecoded.String())
+		}
```
thiagodeev: assert/require.Equal

core/felt/cbor.go
thiagodeev: Thanks.
On second thought, I realize we need to return the created buffer, so there's actually no need for a buffer pool

core/felt/cbor.go
thiagodeev: Thanks!
Now, thinking about it, I made a mistake recommending `copy`. I confused things as if we had two slices being merged, but actually we only have one slice + a new value for it, so `copy` doesn't apply here

core/felt/cbor.go
thiagodeev, line 1: Could you put more comments in this file? I'm missing descriptions about variable values, functions, and why these custom marshall functions exist

thiagodeev (approved):
LGTM.
Only a few comments with nitpicks

core/felt/cbor.go
rodrodros: Since this in the hot path we should

core/felt/cbor.go
RafaelGranza: fixing this one.
also updating the `cborArrayHeader4` for something more readable.

core/felt/cbor_fastpath_test.go
RafaelGranza: fair

core/felt/cbor.go
RafaelGranza: sure

core/felt/cbor.go
```diff
@@ -0,0 +1,145 @@
...
+)
+
+func encodeFeltLimbs(element *fp.Element) []byte {
+	// The buffer format is: [cborArrayHeader4] [limb 0] [limb 1] [limb 2] [limb 3]
+	buffer := make([]byte, 1, 1+Limbs*maxCBORUintLen)
+	buffer[0] = cborArrayHeader4
+
+	for limbIndex := range Limbs {
+		buffer = appendCBORUint(buffer, element[limbIndex])
+	}
```
rodrodros: Important since this is in the hot path, switch the `append` approach to an `indexing` approach. 

Every time there is an append there is an "if there is enough capacity" check, in the nano-seconds and questionable in probably any other case but this one. Please switch the approach.

Basically, instead of doing
```golang
x := make([]byte, 0, 10)
x = append(x, 1, 2, 3)
// and so on
```
Doing 
```
x := make([]byte, 10)
x[0] = 1
x[1] = 2
x[2] = 3
```

You will find that there is an actual performance increase (even if minimal)

core/felt/cbor.go
```diff
@@ -0,0 +1,145 @@
...
+	case value > math.MaxUint16:
+		buffer = append(buffer, cborUint32AdditionalInfo)
+		return binary.BigEndian.AppendUint32(buffer, uint32(value))
+
+	case value > math.MaxUint8:
+		buffer = append(buffer, cborUint16AdditionalInfo)
+		return binary.BigEndian.AppendUint16(buffer, uint16(value))
+
+	case value >= cborUint8AdditionalInfo:
+		return append(buffer, cborUint8AdditionalInfo, byte(value))
```
rodrodros: Here don't use `Append` but `Put`

core/felt/cbor.go
RafaelGranza: [I had investigated it before, and saw no optimization.](https://github.com/NethermindEth/juno/pull/3792#discussion_r3559633932)

TL;DR; Change applied

## Reinvestigating
For us to make things 100% clear, I tested again.
Here is the code I tested against:

```go
func encodeFeltLimbs(element *fp.Element) []byte {
	// The buffer format is: [cborArrayHeader4] [limb 0] [limb 1] [limb 2] [limb 3]
	buffer := make([]byte, 1+Limbs*maxCBORUintLen)
	buffer[0] = cborArrayHeader4

	bufferIndex := 1
	for limbIndex := range Limbs {
		bufferIndex = putCBORUint(buffer, bufferIndex, element[limbIndex])
	}

	return buffer[:bufferIndex]
}

func putCBORUint(buffer []byte, index int, value uint64) int {
	switch {
	case value > math.MaxUint32:
		buffer[index] = cborUint64AdditionalInfo
		binary.BigEndian.PutUint64(buffer[index+1:], value)
		return index + 1 + 8
                ...
```

And again I saw no optimization:

#### Table 1 - put vs append
case | append | put | put vs append
-- | -- | -- | --
zero | 31.28 ns | 35.15 ns | +12.4% (slower)
nonzero | 25.34 ns | 27.66 ns | +9.1% (slower)
geomean | 28.15 ns | 31.18 ns | +10.8% (slower)

## I checked the assembly

Well, I know that the append 'allocate more memory check' was supposed to be slower than an 'out of range' check. So I tested against an UNSAFE pointer just to see, and the unsafe one was still slower than append.

At this point I considered that the compiler was better optimizing append, since it has a more well defined  behavior than a 're-slice' + put. [Just like I previously assumed.](https://github.com/NethermindEth/juno/pull/3792#discussion_r3559653267).

Them I took a look at assembly 💀. And I was correct, it was an compiler optimization.
But it was not due to checks, but due to inlines!

So I inlined the Put, just like this:

```go
func encodeFeltLimbs(element *fp.Element) []byte {
	// The buffer format is: [cborArrayHeader4] [limb 0] [limb 1] [limb 2] [limb 3]
	buffer := make([]byte, 1+Limbs*maxCBORUintLen)
	buffer[0] = cborArrayHeader4

	index := 1
	for limbIndex := range Limbs {
		value := element[limbIndex]
		switch {
		case value > math.MaxUint32:
			buffer[index] = cborUint64AdditionalInfo
			binary.BigEndian.PutUint64(buffer[index+1:], value)
			index += 1 + 8
                ...
```
And the results matched what we were expecting!

#### Table 2 - put inlined vs append
case | append | put | put vs append
-- | -- | -- | --
zero | 31.28 ns | 29.88 ns | -4.5% (faster)
nonzero | 25.34 ns | 22.78 ns | -10.1% (faster)
geomean | 28.15 ns | 26.09 ns | -7.3% (faster)

core/felt/cbor.go
RafaelGranza: Tested against 22.8M fuzz felts, all equal the generic decoder.

core/felt/cbor.go
rodrodros: Nice investigation 👍

rodrodros:
@RafaelGranza, initially the PR only added Felt decoding because it showed a clear improvement on the benchmarks. Can you please add benchmarks showing Felt encoding benefits?

core/felt/cbor.go
```diff
@@ -0,0 +1,149 @@
...
+
+// Fast, felt-specialized CBOR marshaling.
+func (z *Felt) MarshalCBOR() ([]byte, error) {
+	return encodeFeltLimbs((*fp.Element)(z)), nil
+}
+
+// Fast, felt-specialized CBOR unmarshaling.
+// Falls back to the generic decoder if the felt is non-canonical or invalid
+func (z *Felt) UnmarshalCBOR(data []byte) error {
+	if decodeFeltLimbs((*fp.Element)(z), data) {
```
rodrodros: Why there is a cast to `*fp.Element` here, instead of using `*felt.Felt`. I am curious.

core/felt/cbor_fastpath_test.go
```diff
@@ -0,0 +1,239 @@
...
+	}
+	generic, err = cbor.Marshal(value)
+	return fast, generic, err
+}
+
+func decodeBoth(data []byte) (fast, generic felt.Felt, errFast, errGeneric error) {
+	errFast = fast.UnmarshalCBOR(data)
+	errGeneric = cbor.Unmarshal(data, &generic)
+	return fast, generic, errFast, errGeneric
+}
```
rodrodros, line 61: Sadly, it looks like you're using the same (un)marshaler for each test. 

* `value.MarshalCBOR()` calls the newly defined function
* `cbor.Unmarshal(data, &generic` also calls the new defined function, because it is generic over `felt.Felt` and uses it's implementation. 

If you want to bypass your own implemenation and use the truly generic one you should probably use  `fp.Element` as the generic type.

Same problem with decode.

core/felt/cbor.go
```diff
@@ -0,0 +1,149 @@
...
+	"github.com/fxamacker/cbor/v2"
+)
+
+// Fast, felt-specialized CBOR marshaling.
+func (z *Felt) MarshalCBOR() ([]byte, error) {
+	return encodeFeltLimbs((*fp.Element)(z)), nil
+}
+
+// Fast, felt-specialized CBOR unmarshaling.
+// Falls back to the generic decoder if the felt is non-canonical or invalid
```
rodrodros: What do you mean is non-canonical or invalid? Perhaps you want to say: "Falls back on shape missmatch". Let me know if I am wrong

rodrodros (approved):
LGTM!

core/felt/cbor.go
RafaelGranza: I just wanted a way to say that if the CBOR is weird in any sense, like wrong, with a different format or something that is not the default format.

i'll change my wordings

core/felt/cbor.go
RafaelGranza: The `*fp.Element` casts were there to match the nearby `cbor.Unmarshal(data, (*fp.Element)(z))` call.
I've changed the helper calls to use `*Felt` directly.

core/felt/cbor_fastpath_test.go
RafaelGranza: thank you very much for catching this.
fixed and checked with 38M Marshal and 38M Unmarshal + the tests

PR #3796, opened 2026-07-03
perf(core): Hand-rolled UnmarshalBinary for AggregatedBloomFilter

brbrr:
Reduces allocations and the GC pressure for a `getEvents` call

Bench results:
| Metric | main (old decode) | branch (new decode) | Change |
|---|--:|--:|--:|
| time/op | 6.278 ms | 1.340 ms | −78.65% |
| B/op | 25.75 MiB | 8.25 MiB | −67.96% |
| allocs/op | 49,157 | 2 | −100.00% |


rodrodros:
@brbrr how were the benchmark results achieved?

brbrr:
@rodrodros I decided not to commit the bench test, as it's quite simple. Basically I've tested direct unmarshaling and compared it against main branch. Let me know if you think I should commit the bench test. Next week I'll look into it's effects on rpc latency. 

```
func BenchmarkAggregatedBloomFilterUnmarshal(b *testing.B) {
	filter := core.NewAggregatedFilter(0)
	bf := bloom.New(core.EventsBloomLength, core.EventsBloomHashFuncs)
	bf.Add([]byte{0x01})
	require.NoError(b, filter.Insert(bf, 0))

	data, err := filter.MarshalBinary()
	require.NoError(b, err)

	b.ReportAllocs()
	for b.Loop() {
		var decoded core.AggregatedBloomFilter
		if err := decoded.UnmarshalBinary(data); err != nil {
			b.Fatal(err)
		}
	}
}
```

rodrodros:
@brbrr yes, please commit the benchmarks in its dedicated bench_test file, it is good to have them

core/aggregated_bloom_filter.go
```diff
@@ -252,35 +259,61 @@ func (f *AggregatedBloomFilter) MarshalBinary() ([]byte, error) {
...
-	if err := binary.Read(r, binary.BigEndian, &f.toBlock); err != nil {
-		return err
-	}
-
-	var count uint32
-	if err := binary.Read(r, binary.BigEndian, &count); err != nil {
-		return err
+	f.fromBlock = binary.BigEndian.Uint64(data[0:bytesUint64])
+	f.toBlock = binary.BigEndian.Uint64(data[bytesUint64 : 2*bytesUint64])
+	count := int(binary.BigEndian.Uint32(data[2*bytesUint64 : headerSize]))
```
Ehsan-saradar: Building on the earlier note about `count`: since `NewAggregatedFilter` always creates exactly `EventsBloomLength` rows, I'd validate the header invariants right here instead of trusting them —

- `count == EventsBloomLength`: a smaller (or zero) self-consistent `count` decodes fine, then `BlocksForKeys` panics `index out of range` on the next query. A zero-value `AggregatedBloomFilter{}` marshals to a 20-byte header with `count == 0` and hits this too.
- `toBlock == fromBlock + NumBlocksPerFilter - 1`: otherwise a corrupt `toBlock` widens the accepted range while the matrix still has 8192 columns — `Insert` then accepts an out-of-range block, grows a row past 8192 bits, and later intersections silently truncate to the shorter row → wrong results. Verified: `toBlock = 1<<40` decodes with no error and `Insert(b, 20000)` is accepted.

Two cheap guards here close all of it.

core/aggregated_bloom_filter.go
```diff
@@ -252,35 +259,61 @@ func (f *AggregatedBloomFilter) MarshalBinary() ([]byte, error) {
...
+		lo := i * wordsPerFilterRow
+		row := backing[lo : lo+wordsPerFilterRow : lo+wordsPerFilterRow]
+		for w := range wordsPerFilterRow {
+			row[w] = binary.BigEndian.Uint64(data[wordsAt+w*bytesUint64:])
 		}
+		f.bitmap[i] = *bitset.FromWithLength(uint(NumBlocksPerFilter), row)
+
+		off += blobLen
 	}
 	return nil
```
Ehsan-saradar, line 338: Worth asserting `off == len(data)` before returning. Right now trailing bytes after the last row are silently ignored, so a valid blob with junk appended still decodes (`err == nil`) — verified by appending 4 bytes. Makes decode canonical and catches framing corruption/truncation.

core/aggregated_bloom_filter.go
```diff
@@ -252,35 +259,61 @@ func (f *AggregatedBloomFilter) MarshalBinary() ([]byte, error) {
...
+		if blobLen < bitLenSize || off+blobLen > len(data) {
+			return io.ErrUnexpectedEOF
 		}
 
-		b := make([]byte, length)
-		if _, err := io.ReadFull(r, b); err != nil {
-			return err
+		bitLen := binary.BigEndian.Uint64(data[off:])
+		nWords := (blobLen - bitLenSize) / bytesUint64
+		if bitLen != NumBlocksPerFilter || nWords != wordsPerFilterRow ||
```
Ehsan-saradar: Minor/readability: a valid row is always exactly `rowSize - rowLenSize` bytes, so these three conditions collapse to a single `blobLen != <expected>` — which also makes reading `bitLen` provably in-bounds and drops `nWords` entirely.

Ehsan-saradar (commented):
Pulled the branch and reproduced the benchmark independently on a worst-case fully-populated filter (~8.1 MiB blob): allocs **49,157 → 2** and **25.75 MiB → 8.25 MiB** match your table exactly, and the ~78% speedup holds on a slower box too. @rodrodros — that should cover the reproducibility question; committing the benchmark would let CI guard it from here on.

The per-row validation is tight, nice. One theme across the inline notes: the decoder fully trusts the three header fields that `MarshalBinary` always writes consistently (`count`, `toBlock`, and the total length), so a corrupt/truncated DB blob can decode "successfully" and blow up on a later query. All corrupt-input hardening, not happy-path — so flagging as a comment, not blocking.

core/aggregated_bloom_filter.go
```diff
@@ -252,35 +259,61 @@ func (f *AggregatedBloomFilter) MarshalBinary() ([]byte, error) {
 }
 
 func (f *AggregatedBloomFilter) UnmarshalBinary(data []byte) error {
```
rodrodros, line 270: Description for function please, short and helpful please

core/aggregated_bloom_filter.go
RafaelGranza: Just a note for @brbrr . I believe it is a fair update, but note that only two of the three collapse, but not all three. 

`nWords != wordsPerFilterRow` and the modulo check do fold into a single `blobLen != blobLenWant`, and you're right that this also makes the `bitLen` read provably in-bounds.

But `bitLen != NumBlocksPerFilter` can't fold in. `bitLen` and `blobLen` are independent fields in the data.

core/aggregated_bloom_filter.go
```diff
@@ -252,35 +259,61 @@ func (f *AggregatedBloomFilter) MarshalBinary() ([]byte, error) {
...
+		nWords := (blobLen - bitLenSize) / bytesUint64
+		if bitLen != NumBlocksPerFilter || nWords != wordsPerFilterRow ||
+			(blobLen-bitLenSize)%bytesUint64 != 0 {
+			return ErrBloomFilterSizeMismatch
 		}
 
-		if err := f.bitmap[i].UnmarshalBinary(b); err != nil {
-			return err
+		wordsAt := off + bitLenSize
+		lo := i * wordsPerFilterRow
```
RafaelGranza: IMO the names are confusing.
I know this is usually the "marshal style", so it is just a nit.

core/aggregated_bloom_filter.go
RafaelGranza: fair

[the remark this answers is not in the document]
RafaelGranza: fair

core/aggregated_bloom_filter.go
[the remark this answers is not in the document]
RafaelGranza: fair to apply the checks

core/aggregated_bloom_filter_test.go
```diff
@@ -148,3 +150,147 @@ func TestAggregatedBloomFilter_Serialise(t *testing.T) {
...
+	require.NoError(t, err)
+
+	var decoded core.AggregatedBloomFilter
+	allocs := testing.AllocsPerRun(20, func() {
+		if err := decoded.UnmarshalBinary(data); err != nil {
+			t.Fatal(err)
+		}
+	})
+	// One backing []uint64 + one []bitset.BitSet header. Allow slack for
+	// interface boxing in AllocsPerRun itself; current code is ~30k.
```
RafaelGranza: Not anymore! only 2 (which matches the test)!

RafaelGranza (commented):
LGTM. But please address the changes I considered to be fair.
Just as a guarantee, I ran the fuzzer for 1M execs and saw no panics or crashes.

RafaelGranza (approved):

core/aggregated_bloom_filter_test.go
```diff
@@ -148,3 +150,147 @@ func TestAggregatedBloomFilter_Serialise(t *testing.T) {
...
+func TestAggregatedBloomFilter_MarshalBinary_HeaderLayout(t *testing.T) {
+	const from = 0x0102030405060708
+	filter := core.NewAggregatedFilter(from)
+
+	data, err := filter.MarshalBinary()
+	require.NoError(t, err)
+	require.GreaterOrEqual(t, len(data), 20)
+
+	require.Equal(t, uint64(from), binary.BigEndian.Uint64(data[0:8]))
+	require.Equal(t, uint64(from+core.NumBlocksPerFilter-1), binary.BigEndian.Uint64(data[8:16]))
```
RafaelGranza: Just for the linter, remove the unecessary conversion
```suggestion
	require.Equal(t, from+core.NumBlocksPerFilter-1, binary.BigEndian.Uint64(data[8:16]))
```

brbrr:
@rodrodros I've addressed all the feedback. please give it another look when you'll have some time.

core/aggregated_bloom_filter.go
```diff
@@ -251,36 +258,78 @@ func (f *AggregatedBloomFilter) MarshalBinary() ([]byte, error) {
 	return buf.Bytes(), nil
 }
 
+// UnmarshalBinary decodes MarshalBinary's output in place. It never panics on
+// corrupt DB bytes, returning an error instead. On error f may be partially
+// written and must be discarded.
```
thiagodeev:
```suggestion
// UnmarshalBinary decodes MarshalBinary's output in place. It never panics on
// corrupt DB bytes, returning an error instead. On error, f may be partially
// written and must be discarded.
```

core/aggregated_bloom_filter.go
```diff
@@ -251,36 +258,78 @@ func (f *AggregatedBloomFilter) MarshalBinary() ([]byte, error) {
...
+		bitLenSize = bytesUint64
+	)
+
+	if len(data) < headerSize {
+		return io.ErrUnexpectedEOF
+	}
+	f.fromBlock = binary.BigEndian.Uint64(data[0:bytesUint64])
+	f.toBlock = binary.BigEndian.Uint64(data[bytesUint64 : 2*bytesUint64])
+	count := int(binary.BigEndian.Uint32(data[2*bytesUint64 : headerSize]))
+	off := headerSize
```
rodrodros: I assume this variable represents, an offset, why not call it `offset`?

core/aggregated_bloom_filter.go
```diff
@@ -74,6 +74,13 @@ const (
 	NumBlocksPerFilter uint64 = 8192
 )
 
+// wordsPerFilterRow is the number of uint64 words bitset uses to store one
+// NumBlocksPerFilter-bit row.
+var wordsPerFilterRow = func() int {
+	row := makeBitset()
+	return len(row.Words())
+}()
```
rodrodros: Maybe we can make this a constant equal to `NumBlocksPerFilter - 1`. I believe it might be a bit more legible.

core/aggregated_bloom_filter.go
```diff
@@ -251,36 +258,78 @@ func (f *AggregatedBloomFilter) MarshalBinary() ([]byte, error) {
 	return buf.Bytes(), nil
 }
 
+// UnmarshalBinary decodes MarshalBinary's output in place. It never panics on
+// corrupt DB bytes, returning an error instead. On error f may be partially
+// written and must be discarded.
 func (f *AggregatedBloomFilter) UnmarshalBinary(data []byte) error {
```
rodrodros: Wording like "It never panics on corrup DB bytes, returning an error instead" may have a more direct wording of: "returns an error in case of DB corruption". There is no need to say what it doesn't do, if it doesn't do it. It is enough to say what the function does. 

```golang
// UnmarshalBinary decodes MarshalBinary's output in place. Returns an error
// in case of DB corruption. On error f may be partially written and must be 
// discarded. 
```

core/aggregated_bloom_filter.go
```diff
@@ -251,36 +258,78 @@ func (f *AggregatedBloomFilter) MarshalBinary() ([]byte, error) {
...
+		return ErrBloomFilterSizeMismatch
 	}
 
-	var count uint32
-	if err := binary.Read(r, binary.BigEndian, &count); err != nil {
-		return err
+	// Reject data too short for count rows before allocating, so truncated
+	// input fails fast rather than driving an oversized allocation.
+	blobLenWant := bitLenSize + wordsPerFilterRow*bytesUint64
+	rowSize := rowLenSize + blobLenWant
```
rodrodros: Both `blobLenWant` and `rowSize` could be constants, assuming that the previous comment is applied (i.e. making `wordsPerFilterRow` a constant)

core/aggregated_bloom_filter.go
rodrodros: Why is `off` created here when it is used 23 lines below? Why it is not defined right before being used?

core/aggregated_bloom_filter.go
```diff
@@ -251,36 +258,78 @@ func (f *AggregatedBloomFilter) MarshalBinary() ([]byte, error) {
...
+	if count > (len(data)-headerSize)/rowSize {
+		return io.ErrUnexpectedEOF
 	}
 
+	backing := make([]uint64, count*wordsPerFilterRow)
```
rodrodros, line 305: Why not use a matrix here? [][]uint64?

Have you consider using a static array, since `count` must be `EventsBloomLength` and `wordsFilterPerRow` is fixed as well: [a][b]uint64?

rodrodros, line 305: It also seems that you don't need all of that space at once. You only need one `row` that is re-used per loop and passed to the bitmap in `i`, or am I mistaken?

rodrodros (commented):
Solving the issue is not good enough if the code that accompanies is plagued by readability issues – causing issues for future developers. 

I would like for you to reinforce on this side of code writing and doing self reviews.

rodrodros:
@RafaelGranza I would've expect for you to catch some the things I catched on my review. As said previously on dailies, reviews are not only to check if an issue is solved, but to also check that is done in the best way possible – readability wise included. 

@brbrr please be more careful in follow up PRs and take your time to familiarize with what you wrote and try to give yourself AI-less self reviews.

core/aggregated_bloom_filter.go
```diff
@@ -251,36 +258,78 @@ func (f *AggregatedBloomFilter) MarshalBinary() ([]byte, error) {
...
-
-	if err := binary.Read(r, binary.BigEndian, &f.fromBlock); err != nil {
-		return err
+	const (
+		bytesUint32 = 4
+		bytesUint64 = 8
+		// Header layout (see MarshalBinary): fromBlock + toBlock + count.
+		headerSize = bytesUint64 + bytesUint64 + bytesUint32
+		rowLenSize = bytesUint32
+		bitLenSize = bytesUint64
```
thiagodeev: Can we rename `bitLenSize` to `bitsetLenSize`?

core/aggregated_bloom_filter.go
```diff
@@ -251,36 +258,78 @@ func (f *AggregatedBloomFilter) MarshalBinary() ([]byte, error) {
...
+		blobLen := int(binary.BigEndian.Uint32(data[off:]))
+		off += rowLenSize
+		if off+blobLen > len(data) {
+			return io.ErrUnexpectedEOF
+		}
+		// bitLen and blobLen are independent fields, so both are checked.
+		if blobLen != blobLenWant {
+			return ErrBloomFilterSizeMismatch
+		}
+		if bitLen := binary.BigEndian.Uint64(data[off:]); bitLen != NumBlocksPerFilter {
```
thiagodeev: Can we rename `bitLen` to `bitsetLen`?

core/aggregated_bloom_filter.go
```diff
@@ -74,6 +74,13 @@ const (
 	NumBlocksPerFilter uint64 = 8192
 )
 
+// wordsPerFilterRow is the number of uint64 words bitset uses to store one
+// NumBlocksPerFilter-bit row.
+var wordsPerFilterRow = func() int {
```
thiagodeev: We can make this a const instead of using `len(row.Words())`.

``` golang
const wordsPerFilterRow = int((NumBlocksPerFilter + 63) / 64) // 128
```

This is the formula actually being used by `BitSet.WriteTo`/ `BitSet.MarshalBinary` to calculate the used words - [ref](https://github.com/bits-and-blooms/bitset/blob/v1.24.5/bitset.go#L1284), [ref](https://github.com/bits-and-blooms/bitset/blob/v1.24.5/bitset.go#L787), [ref](https://github.com/bits-and-blooms/bitset/blob/v1.24.5/bitset.go#L152) (`(int(i) + wordMask) >> log2WordSize` == `(i + 63) / 64`).

By calling `len(row.Words())` we are getting the length of the internal storage slice, while `BitSet.WriteTo`/ `BitSet.MarshalBinary` only write the actual used words, obtained by calling `wordCount()`.
This works today because `makeBitset()` happens to allocate exactly that many words, but that isn’t really an API guarantee (thinking about a future change in the lib). Having it as a fixed const is better IMO since we are creating our own custom decode function with fixed values.

thiagodeev (commented):
Nice improvement!
Partial review

core/aggregated_bloom_filter.go
thiagodeev: AFAIU, the real improvement of the PR comes from exactly this. Before, there were thousands of allocations per filter, and now we allocate a big array once and use pieces of it to back all the thousands of slices, avoiding allocating them.
Reusing a single small array is not possible because we need individual backing arrays for all the slices/rows since each one of them stores different filters

EgeCaner: A matrix represented as a 1D array is faster than a 2D [][]uint64, it's a single contiguous allocation (better cache locality, no per-row allocations or extra pointer indirection).

A fixed [a][b]uint64 array still escapes to the heap because its sub-slices are stored in f.bitmap and outlive the function.

core/aggregated_bloom_filter_test.go
```diff
@@ -148,3 +150,172 @@ func TestAggregatedBloomFilter_Serialise(t *testing.T) {
...
+
+// UnmarshalBinary decodes untrusted DB bytes and must never panic on arbitrary
+// input, only return an error.
+func FuzzAggregatedBloomFilterUnmarshal(f *testing.F) {
+	filter := core.NewAggregatedFilter(0)
+	b := bloom.New(core.EventsBloomLength, core.EventsBloomHashFuncs)
+	b.Add([]byte{0x01})
+	require.NoError(f, filter.Insert(b, 0))
+	valid, err := filter.MarshalBinary()
+	require.NoError(f, err)
```
thiagodeev: I'm seeing this pattern multiple times (new filter > new bloom > b.Add > filter.Insert, and sometimes the filter.MarshalBinary). Could you create a small helper func to avoid all this boilerplate?

core/aggregated_bloom_filter.go
thiagodeev, line 1: One important thing related to the `bitset` pkg.
Inside the lib code, we can see that both MarshalBinary and UnmarshalBinary operations are performed by the binary order configured in the `binaryOrder` variable, which is `binary.BigEndian` by default ([ref](https://github.com/bits-and-blooms/bitset/blob/v1.24.5/bitset.go#L67)). Both MarshalBinary (through the `WriteTo` method - [ref](https://github.com/bits-and-blooms/bitset/blob/v1.24.5/bitset.go#L1352) and [ref](https://github.com/bits-and-blooms/bitset/blob/v1.24.5/bitset.go#L1278)) and UnmarshalBinary (through the `ReadFrom` method - [ref](https://github.com/bits-and-blooms/bitset/blob/v1.24.5/bitset.go#L1364) and [ref](https://github.com/bits-and-blooms/bitset/blob/v1.24.5/bitset.go#L1312)) use the binary order configured by the same `binaryOrder` variable. The lib provides public ways to flip this order (by calling `LittleEndian()` and `BigEndian()`), which would change the binary order being used for the entire lib, meaning that both Marshal and Unmarshal would be either LittleEndian or BigEndian at the same time.

This PR introduces a custom UnmarshalBinary method for the `AggregatedBloomFilter`, on which we directly call `binary.BigEndian...` for all the operations related to `bitset`; however, our MarshalBinary method is still the same, and it calls the `MarshalBinary` method of the `bitset` type itself ([ref](https://github.com/NethermindEth/juno/blob/perf/bloom-filter-decode-alloc/core/aggregated_bloom_filter.go#L243)).

So, if anywhere in the code the `LittleEndian()` public function of the `bitset` lib is called (even external dependencies, since it's a global variable of the `bitset` pkg), the `AggregatedBloomFilter.MarshalBinary` method would start to encode the value as LIttleEndian, and since we are now using our custom version with BigEndian hardcoded, when unmarshaling this same AggregatedBloomFilter that we just marshaled, it would be considered invalid.

At first, I thought the problem was only in the new changes by this PR, but it turns out that this is already a danger for our current code as well. In-memory AggregatedBloomFilter round-trip would fail instantly due to the new changes from this PR; however, since the filters were stored in the db using BigEndian, even though old code would succeed on in-memory round-trip by using the same new binary order (littleEndian), it would fail/corrupt when reading and storing from the db (stored as BigEndian).

Solutions? 
- customize our MarshalBinary method as well;
- a check like `bitset.BinaryOrder() == binary.BigEndian` somewhere in the code to fail fast and not corrupt;
- copy the `bitset.go` code and use our own implementation, not being affected by any change in the public global values from the lib...

So, this is at the same time something related and not related to this PR.
Right now, the `LittleEndian()` func is not called anywhere, but since it's something so subtle and hard to debug if it happens, I believe it's worth addressing it ASAP

thiagodeev (commented):
Another partial review

core/aggregated_bloom_filter.go
brbrr: While the endianness toggle seems like a valid concern. As it will cause a DB corruption (for existing DBs), because our de/serialization path after this PR is not symmetrical. But realistically, I don't think anyone would do that (both externally and internally), and if it indeed will happen, it all will blow up right away anyway.

core/aggregated_bloom_filter.go
```diff
@@ -251,43 +258,86 @@ func (f *AggregatedBloomFilter) MarshalBinary() ([]byte, error) {
...
-		if _, err := io.ReadFull(r, b); err != nil {
-			return err
+		blobLen := int(binary.BigEndian.Uint32(data[offset:]))
+		offset += rowLenSize
+		if offset+blobLen > len(data) {
+			return io.ErrUnexpectedEOF
+		}
+		// bitsetLen and blobLen are independent fields, so both are checked.
+		if blobLen != blobLenWant {
+			return ErrBloomFilterSizeMismatch
```
rodrodros: These three `ifs` protect the code against corrupted data, but there are no tests to assert this failsafe. This means that someone could erase this code in production and we would unknowingly loose our defenses.

Could you please add tests for this error cases

rodrodros (approved):
Please solve open comment and it is good to go 👍

PR #3807, opened 2026-07-09
feat(l1): implement hand-rolled ws-based l1 client

brbrr:


l1/eth/client/transport_ws.go
```diff
@@ -0,0 +1,527 @@
...
+				select {
+				case <-timer.C:
+				default:
+				}
+			}
+			timer.Reset(t.pingInterval)
+		case <-timer.C:
+			ctx, cancel := context.WithTimeout(context.Background(), t.pingTimeout)
+			err := t.conn.Ping(ctx)
+			cancel()
```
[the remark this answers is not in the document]
brbrr: From docs:

// Ping sends a ping to the peer and waits for a pong.
// Use this to measure latency or ensure the peer is responsive.
// Ping must be called concurrently with Reader as it does
// not read from the connection but instead waits for a Reader call
// to read the pong.

l1/eth_l1_state_provider.go
```diff
@@ -0,0 +1,331 @@
...
+// ErrClosed is returned by methods invoked after Close. Unlike client.ErrTransportClosed
+// (transient, recovered by redial), this is terminal.
+var ErrClosed = errors.New("L1 state provider closed")
+
+type EthL1StateProviderOption func(*ethL1StateProviderOptions)
+
+type ethL1StateProviderOptions struct {
+	logger   log.StructuredLogger
+	listener EventListener
+}
```
rodrodros: nit: invert these types declarations. First `ethL1` and then `EthL1`.

l1/eth_l1_state_provider.go
```diff
@@ -0,0 +1,331 @@
...
+// Must be called before the provider is handed off to any goroutine.
+func WithEthL1StateProviderListener(l EventListener) EthL1StateProviderOption {
+	return func(o *ethL1StateProviderOptions) { o.listener = l }
+}
+
+// Also satisfies rpccore.L1Client (via TransactionReceipt), so one instance serves both the sync
+// loop and the RPC handlers. Keeps the dial details to transparently redial a dropped conn: the
+// underlying client is one-shot, so on client.ErrTransportClosed we redial and retry once.
+// Subscriptions still need re-issuing by the caller, same as go-ethereum's rpc.Client.
+type EthL1StateProvider struct {
```
rodrodros, line 40: Starting a type description with "Also satisfies ..." can be confusing for readers.  Descripion should start with `EthL1StateProvider does this and that...."

It shouldn't also explain how it does what it does, that's why the code is there. 

Plrease rewrite the whole description since it looks like AI slop, lots of word, no content.

l1/eth_l1_state_provider.go
```diff
@@ -0,0 +1,331 @@
...
+	logger          log.StructuredLogger
+	listener        EventListener
+
+	mu     sync.Mutex // protects client and closed
+	client *client.Client
+	closed bool
+}
+
+// ws/wss is required if the caller intends to use WatchStateUpdate. The url and
+// any client options are remembered so dropped connections can be redialed transparently.
```
rodrodros: This description also starts wrong, it should start with:
"NewEthL1StateProvider creates a new EthL1StatePRovider and so on"

Additionally, the whole "ws/wss is required if..." seems unnecessary since the only possible options are websockets. 

And what's even the point of this comment?: "The url and any client options are remembered so dropped connections can be redialed transparently"

l1/eth_l1_state_provider.go
rodrodros, line 1: I will not mark any further comments on the descriptions of this files, so please review and fix them all.

I don't want to read AI slop.

Additionally, all function description should start with the function name. The function description is a _small_ description with _simple_ words of what the function does.

node/node.go
```diff
@@ -659,38 +659,67 @@ func New(cfg *Config, version string, logLevel *log.Level) (*Node, error) {
...
+// sync loop and rpccore.L1Client for the RPC handlers.
+type l1StateProviderFull interface {
+	l1.L1StateProvider
+	rpccore.L1Client
+}
+
 func newL1Client(
 	ctx context.Context,
-	ethNode string,
+	clientType, ethNode string,
```
rodrodros: arg per line

node/node.go
rodrodros, line 1: This script is already to big for node, adding more `l1` code hints that we should define a sibling `l1.go` script and move these definitions there (on the same `node` package)

cmd/juno/juno.go
rodrodros, line 1: I don't think people care if it is "geth" or "juno". I think the flag should be a boolean, `--use-new-l1-client` that uses this implementation when on, uses the old when off.

node/migration.go
```diff
@@ -108,19 +108,19 @@ func fetchL1HeadIfMissing(
 	logger.Info("Fetching the L1 head before running the prune migration")
 	// Metrics are registered by the long-lived L1 client built in node.New; reusing
 	// them here would panic via prometheus.MustRegister. Hence no listener.
-	provider, err := newGethL1StateProvider(ctx, config.EthNode, chain)
+	provider, err := newMigrationL1StateProvider(ctx, config.UseNewL1Client, config.EthNode, chain)
 	if err != nil {
 		return fmt.Errorf("creating L1 state provider: %w", err)
 	}
```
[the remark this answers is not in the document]
brbrr: not relevant.

cmd/juno/juno.go
```diff
@@ -208,7 +210,10 @@ const (
 	colourUsage                           = "Use `--colour=false` command to disable colourized outputs (ANSI Escape Codes)."
 	ethNodeUsage                          = "WebSocket endpoint of the Ethereum node. To verify the correctness of the L2 chain, " +
 		"Juno must connect to an Ethereum node and parse events in the Starknet contract."
-	disableL1VerificationUsage    = "Disables L1 verification since an Ethereum node is not provided."
+	disableL1VerificationUsage = "Disables L1 verification since an Ethereum node is not provided."
+
+	useNewL1ClientUsage = "Use the experimental L1 client implementation."
+
```
rodrodros: no spaces between usage lines, please remove the lines above and below

l1/eth/client/rpc.go
```diff
@@ -0,0 +1,76 @@
...
+		return fmt.Sprintf("json-rpc error %d: %s: %s", e.Code, e.Message, e.Data)
+	}
+	return fmt.Sprintf("json-rpc error %d: %s", e.Code, e.Message)
+}
+
+// parseResponseID rejects a null id: the server couldn't tell which request
+// it was answering, so there is no caller to route the reply to.
+func parseResponseID(raw json.RawMessage) (uint64, error) {
+	trimmed := bytes.TrimSpace(raw)
+	if len(trimmed) == 0 || bytes.Equal(trimmed, jsonNull) {
```
rodrodros: Why not using the `isJSonNull` utility here?

l1/eth/client/transport_ws.go
```diff
@@ -0,0 +1,493 @@
...
+			"ws: drop unparseable frame",
+			zap.Int("bytes", len(data)),
+			zap.Error(err),
+		)
+		return
+	}
+	switch {
+	case probe.Method == "eth_subscription":
+		t.dispatchNotification(data)
+	case len(probe.ID) > 0 && !bytes.Equal(probe.ID, jsonNull):
```
rodrodros: isJsonNull?

l1/eth/client/client.go
```diff
@@ -0,0 +1,162 @@
...
+	"github.com/NethermindEth/juno/utils/log"
+)
+
+type Client struct {
+	tr *wsTransport
+}
+
+const BlockFinalized = "finalized"
+
+var jsonNull = []byte("null")
```
rodrodros: This can be declared direcly inside the `isJsonNUll` helper

l1/eth/client/transport_ws.go
```diff
@@ -0,0 +1,493 @@
...
+	go t.readLoop() //nolint:gosec // G118: long-lived loop, not request-scoped
+	go t.pingLoop() //nolint:gosec // G118: long-lived loop, not request-scoped
+	return t, nil
+}
+
+// readLoop drops malformed frames rather than tearing the transport down —
+// a misbehaving remote manifests as a call timeout.
+func (t *wsTransport) readLoop() {
+	for {
+		_, data, err := t.conn.Read(context.Background())
```
rodrodros: Why is the parent context not being passed?

l1/eth/client/transport_ws.go
```diff
@@ -0,0 +1,493 @@
...
+
+// pingLoop pings only after pingInterval of write silence — regular traffic
+// already proves the conn is alive. A ping failure shuts the transport down
+// via the same path as a read error.
+func (t *wsTransport) pingLoop() {
+	timer := time.NewTimer(t.pingInterval)
+	defer timer.Stop()
+	for {
+		select {
+		case <-t.closed:
```
rodrodros: Same idea, this should receive a parent context

l1/eth/client/transport_ws.go
```diff
@@ -0,0 +1,493 @@
...
+		return
+	}
+	switch {
+	case probe.Method == "eth_subscription":
+		t.dispatchNotification(data)
+	case len(probe.ID) > 0 && !bytes.Equal(probe.ID, jsonNull):
+		t.dispatchResponse(data)
+	default:
+		t.logger.Trace(
+			"ws: drop frame with no id and no recognised method",
```
rodrodros: `ws` is unnecessary since loggers already inform which file is logging, please apply to all the loggings

l1/eth/client/client.go
```diff
@@ -0,0 +1,162 @@
...
+	}
+	return &Client{tr: ws}, nil
+}
+
+func (c *Client) Close() { c.tr.close() }
+
+func isJSONNull(raw json.RawMessage) bool {
+	return bytes.Equal(bytes.TrimSpace(raw), jsonNull)
+}
+
```
rodrodros: Doesn't it makes more sense for this helper method to be moved to some other place like the end of the file? It sits a bit randomly between methods

l1/eth/client/client.go
```diff
@@ -0,0 +1,162 @@
...
+func (c *Client) ChainID(ctx context.Context) (*big.Int, error) {
+	raw, err := c.tr.call(ctx, "eth_chainId")
+	if err != nil {
+		return nil, fmt.Errorf("getting chain ID: %w", err)
+	}
+	return decodeQuantityBig(raw)
+}
```
rodrodros, line 69: Why a big.Int instead of a uint256?

l1/eth/client/client.go
```diff
@@ -0,0 +1,162 @@
...
+	raw, err := c.tr.call(ctx, "eth_blockNumber")
+	if err != nil {
+		return 0, fmt.Errorf("getting block number: %w", err)
+	}
+	return decodeQuantityUint64(raw)
+}
+
+// HeaderByNumber maps a null result — the server's signal for "the named
+// block does not exist yet" — to eth.ErrNotFound.
+func (c *Client) HeaderByNumber(ctx context.Context, tag string) (*eth.Header, error) {
```
rodrodros: Why returning a pointer?

l1/eth/client/client.go
```diff
@@ -0,0 +1,165 @@
...
+	var h eth.Header
+	if err := json.Unmarshal(raw, &h); err != nil {
+		return nil, fmt.Errorf("decoding header: %w", err)
+	}
+	return &h, nil
+}
+
+// TransactionReceipt maps a null result (receipt unknown to the node) to
+// eth.ErrNotFound.
+func (c *Client) TransactionReceipt(ctx context.Context, txHash eth.Hash) (*eth.Receipt, error) {
```
rodrodros: Why returning a pointer?

l1/eth_l1_state_provider.go
```diff
@@ -0,0 +1,337 @@
...
+}
+
+func stateUpdateFromContract(ev *contract.LogStateUpdate) *StateUpdate {
+	return &StateUpdate{
+		L2BlockNumber: ev.BlockNumber,
+		L2BlockHash:   ev.BlockHash,
+		StateRoot:     ev.GlobalRoot,
+		L1RefHeight:   uint64(ev.Raw.BlockNumber),
+		Removed:       ev.Raw.Removed,
+	}
```
rodrodros: Why return a pointer?

brbrr: updated

l1/eth/client/client.go
brbrr: updated

l1/eth/client/client.go
brbrr: updated

l1/eth/client/client.go
brbrr: mostly because network.Network.L1ChainID is *bit.Int.

l1/eth/client/transport_ws.go
brbrr: The ctx passed to New is dial-scoped, it is bound it to a minute, so using it here would kill the connection  after 60sec. The loops now run under a transport-lifecycle ctx created in dialWS and cancelled in shutdown, which is the actual lifetime they need to follow.

l1/eth/client/transport_ws.go
brbrr: https://github.com/NethermindEth/juno/pull/3807#discussion_r3664088835

brbrr:
I've split this PR into 3 separate ones to make it easier to review. I'll keep it as draft for a time being as reference.

Issue #3785, opened 2026-07-03
perf(state): avoid full header decode in StateAtBlockHash


ongyimeng:
StateAtBlockHash currently calls pruner.HeaderByHashIfStateRetained, which resolves the block hash and fully decodes the corresponding block header. Opening historical state by hash only needs:

Block number from the hash → number index
GlobalStateRoot from the header

A partial-header decode could avoid materializing heavier unused fields such as EventsBloom.

Use https://github.com/NethermindEth/juno/pull/3782 as reference implementation

PR #3808, opened 2026-07-09
perf(state): avoid full header decode in StateAtBlockNumber

### LINKED ISSUE #3784: perf(state): avoid full header decode in StateAtBlockNumber
StateAtBlockNumber currently calls pruner.HeaderByNumberIfStateRetained, which fully decodes the block header even though opening historical state only needs:

GlobalStateRoot
Hash, for the state-retention hash → number check

A partial-header decode could avoid materializing heavier unused fields such as EventsBloom.

Use https://github.com/NethermindEth/juno/pull/3782 as reference implementation

0xsamalt:
## Summary

Closes #3784.

`StateAtBlockNumber` only needs a block's global state root to check retention and construct a `StateHistory`, but it was going through `pruner.HeaderByNumberIfStateRetained`, which fully decodes the block header including fields like `EventsBloom` that this path never uses.

## Background

#3782 already introduced partial-decode accessors for this exact purpose:
- `core.GetGlobalStateRootByBlockNumber`
- `core.GetBlockHeaderHashByNumber`

and `pruner.RequireStateRetainedByBlockNumber` already used the hash-based retention check pattern. The only remaining gap was that `StateAtBlockNumber` in `blockchain/statebackend/statebackend.go` hadn't been updated to use them.

## Changes

- **`pruner/retention.go`**: renamed `HeaderByNumberIfStateRetained` →`StateRootByNumberIfStateRetained`. It now returns `*felt.Felt` (the state root) instead of `*core.Header`, built from: `GetGlobalStateRootByBlockNumber` →`GetBlockHeaderHashByNumber` →`GetBlockHeaderNumberByHash` (retention check).
- **`blockchain/statebackend/statebackend.go`**: `StateAtBlockNumber` now calls `StateRootByNumberIfStateRetained` directly and passes the state root straight to `state.NewStateHistory`, avoiding the header decode.
- **`pruner/retention_test.go`**: renamed `TestHeaderByNumberIfStateRetained` →`TestStateRootByNumberIfStateRetained`; the "fully retained" case now asserts on the state root instead of a header field. The pruned/missing-block error cases only needed the function name updated no logic changes.

## Testing

- `go build ./...` — clean
- `go test ./pruner/... ./blockchain/statebackend/... ./core/...` —
  all pass
- No new tests added: the reused accessors already have full coverage
  from #3782's `TestPartialBlockHeaderAccessorsByNumber`, and no new decode logic was introduced here.

## Next

Following up with #3785 (`StateAtBlockHash`), same pattern.

### DIFF
--- a/pruner/retention.go
+++ b/pruner/retention.go
@@ -62,11 +62,7 @@ func RequireRetained(r db.KeyValueReader, blockNumber uint64) error {
 // See [PruneUpto] for the carve-out semantics. Returns db.ErrKeyNotFound
 // when state at blockNumber is not available.
 func StateRootByNumberIfStateRetained(r db.KeyValueReader, blockNumber uint64) (*felt.Felt, error) {
-	state, err := core.GetGlobalStateRootByBlockNumber(r, blockNumber)
-	if err != nil {
-		return nil, err
-	}
-	hash, err := core.GetBlockHeaderHashByNumber(r, blockNumber)
+	state, hash, err := core.GetStateRootAndHashByBlockNumber(r, blockNumber)
 	if err != nil {
 		return nil, err
 	}
--- a/pruner/retention.go
+++ b/pruner/retention.go
@@ -54,22 +54,26 @@ func RequireRetained(r db.KeyValueReader, blockNumber uint64) error {
 	return &BlockPrunedError{BlockNumber: blockNumber, OldestRetained: oldest}
 }
 
-// HeaderByNumberIfStateRetained returns the header for blockNumber only if
+// StateRootByNumberIfStateRetained returns the state root for blockNumber only if
 // state at that block is queryable. Use this for state access only; for
 // block-level data (transactions, receipts, etc.) use [RequireRetained] + the
 // plain core accessors instead — the two checks diverge at oldestKept-1,
 // where state is preserved by carve-out but block-level data is not.
 // See [PruneUpto] for the carve-out semantics. Returns db.ErrKeyNotFound
 // when state at blockNumber is not available.
-func HeaderByNumberIfStateRetained(r db.KeyValueReader, blockNumber uint64) (*core.Header, error) {
-	header, err := core.GetBlockHeaderByNumber(r, blockNumber)
+func StateRootByNumberIfStateRetained(r db.KeyValueReader, blockNumber uint64) (*felt.Felt, error) {
+	state, err := core.GetGlobalStateRootByBlockNumber(r, blockNumber)
 	if err != nil {
 		return nil, err
 	}
-	if _, err := core.GetBlockHeaderNumberByHash(r, header.Hash); err != nil {
+	hash, err := core.GetBlockHeaderHashByNumber(r, blockNumber)
+	if err != nil {
 		return nil, err
 	}
-	return header, nil
+	if _, err := core.GetBlockHeaderNumberByHash(r, hash); err != nil {
+		return nil, err
+	}
+	return state, nil
 }
 
 // RequireStateRetainedByBlockNumber checks state retention by number, avoiding the full
--- a/pruner/retention.go
+++ b/pruner/retention.go
@@ -62,7 +62,7 @@ func RequireRetained(r db.KeyValueReader, blockNumber uint64) error {
 // See [PruneUpto] for the carve-out semantics. Returns db.ErrKeyNotFound
 // when state at blockNumber is not available.
 func StateRootByNumberIfStateRetained(r db.KeyValueReader, blockNumber uint64) (*felt.Felt, error) {
-	state, hash, err := core.GetStateRootAndHashByBlockNumber(r, blockNumber)
+	hash, state, err := core.GetStateRootAndHashByBlockNumber(r, blockNumber)
 	if err != nil {
 		return nil, err
 	}
--- a/blockchain/statebackend/statebackend.go
+++ b/blockchain/statebackend/statebackend.go
@@ -34,12 +34,12 @@ func (b *stateBackend) HeadState() (core.StateReader, StateCloser, error) {
 func (b *stateBackend) StateAtBlockNumber(
 	blockNumber uint64,
 ) (core.StateReader, StateCloser, error) {
-	header, err := pruner.HeaderByNumberIfStateRetained(b.database, blockNumber)
+	stateRoot, err := pruner.StateRootByNumberIfStateRetained(b.database, blockNumber)
 	if err != nil {
 		return nil, nil, err
 	}
 
-	history, err := state.NewStateHistory(blockNumber, header.GlobalStateRoot, b.stateDB)
+	history, err := state.NewStateHistory(blockNumber, stateRoot, b.stateDB)
 	if err != nil {
 		return nil, nil, err
 	}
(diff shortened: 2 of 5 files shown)


pruner/retention.go
```diff
@@ -54,22 +54,26 @@ func RequireRetained(r db.KeyValueReader, blockNumber uint64) error {
...
 // when state at blockNumber is not available.
-func HeaderByNumberIfStateRetained(r db.KeyValueReader, blockNumber uint64) (*core.Header, error) {
-	header, err := core.GetBlockHeaderByNumber(r, blockNumber)
+func StateRootByNumberIfStateRetained(r db.KeyValueReader, blockNumber uint64) (*felt.Felt, error) {
+	state, err := core.GetGlobalStateRootByBlockNumber(r, blockNumber)
 	if err != nil {
 		return nil, err
 	}
-	if _, err := core.GetBlockHeaderNumberByHash(r, header.Hash); err != nil {
+	hash, err := core.GetBlockHeaderHashByNumber(r, blockNumber)
```
RafaelGranza: **This needs measurements and fixes**

Since this is a `perf` PR, so I'm being more picky related to performance stuff.

Both `GetBlockHeaderHashByNumber` and `GetGlobalStateRootByBlockNumber` internally call:

```go
err := r.Get(db.BlockHeaderByNumberKey(blockNum), func(data []byte) error {
		return encoder.Unmarshal(data, &header)
	})
```
so now we are querying the DB and marshaling twice.

One way to fix this is creating a new accessor.

RafaelGranza (commented):
This is a good **possible** optimization, but we still need to see the benchmarks (similar to the ones [here](https://github.com/NethermindEth/juno/pull/3782#issue-4793080382)) so we actually know this is an optimization.

 I know you are not implicitly claiming this to be a performance improvement, but this PR is of the `perf` type.
 And I cannot approve [performance without measurements](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#per6-dont-make-claims-about-performance-without-measurements).
 
I'm not sure if you are just starting your career and looking for "first PRs in open source", but if that is the case, and you are interested in performance, I recommend this very [small guide](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#s-performance). It is focused in C++, but the principles are good and any modern compiled or lower-level language should behave similarly. 
 
Also, there is another possible performance issue you need to address.

0xsamalt:
## Summary

Consolidates the header hash and global state root reads used by `StateAtBlockNumber` (and `StateRootByNumberIfStateRetained`) into a single accessor.

Previously this path called two separate reads — `GetGlobalStateRootByBlockNumber` and `GetBlockHeaderHashByNumber` — each doing its own `r.Get` and unmarshal.

This adds `GetStateRootAndHashByBlockNumber`, which does a single `r.Get` + unmarshal and returns both `GlobalStateRoot` and `Hash`, and uses it in `StateAtBlockNumber` / `StateRootByNumberIfStateRetained`.

## Benchmarks

I ran local benchmarks (`rpc/v10/state_read_bench_test.go`) exercising the public RPC handler paths with Pebble DB, following the format from #3782 / #3789.

Runs were interleaved (base/new/base/new) with CPU pinning (`taskset -c 0-3`, `GOMAXPROCS=4`) and `-count=20` to control for local hardware noise:

| Path | Base | New | Latency | B/op | allocs/op |
|---|---|---|---|---|---|
| `Nonce/ByNumber` | ~12.32µs | ~12.48µs | ~ (p=0.292, not significant) | 2.815Ki → 2.815Ki (+0%) | 60 → 60 (+0%) |
| `ClassHashAt/ByNumber` | ~13.03µs | ~12.15µs | −6.75% (p=0.000) | 2.815Ki → 2.815Ki (+0%) | 60 → 60 (+0%) |
| `StorageAt/ByNumber` | ~20.54µs | ~21.09µs | ~ (p=0.531, not significant) | 4.505Ki → 4.505Ki (+0%) | 89 → 89 (+0%) |

Only `ClassHashAt/ByNumber` shows a statistically significant improvement; `Nonce/ByNumber` and `StorageAt/ByNumber` which exercise the same consolidated accessor are flat. `B/op` and `allocs/op` show no measurable change on any path, which is inconsistent with removing a full extra DB read + unmarshal.

I also saw a significant delta on `ClassHashAt/Latest` (−6.22%, p=0.000), a path that routes through `HeadState()` rather than `StateAtBlockNumber()` and shouldn't be affected by this change at all. That, combined with the lack of any B/op or allocs/op movement, points to local hardware variance (thermal/turbo scaling) rather than a real effect from this PR.

I don't have a reliable perf number to report from local testing, despite several rounds of interleaved, CPU-pinned runs. Happy to have this re-benchmarked on other hardware or in CI if a hard number is needed before merge otherwise this PR should stand on correctness/code-clarity grounds (one DB read + unmarshal instead of two).

pruner/retention.go
0xsamalt: I have now included this in the following commit.

pruner/retention.go
```diff
@@ -54,22 +54,22 @@ func RequireRetained(r db.KeyValueReader, blockNumber uint64) error {
...
 // state at that block is queryable. Use this for state access only; for
 // block-level data (transactions, receipts, etc.) use [RequireRetained] + the
 // plain core accessors instead — the two checks diverge at oldestKept-1,
 // where state is preserved by carve-out but block-level data is not.
 // See [PruneUpto] for the carve-out semantics. Returns db.ErrKeyNotFound
 // when state at blockNumber is not available.
-func HeaderByNumberIfStateRetained(r db.KeyValueReader, blockNumber uint64) (*core.Header, error) {
-	header, err := core.GetBlockHeaderByNumber(r, blockNumber)
+func StateRootByNumberIfStateRetained(r db.KeyValueReader, blockNumber uint64) (*felt.Felt, error) {
+	state, hash, err := core.GetStateRootAndHashByBlockNumber(r, blockNumber)
```
Ehsan-saradar: The return values are swapped. `GetStateRootAndHashByBlockNumber` returns `(hash, stateRoot)`, but it's bound here as `state, hash`. So the retention check runs on the state root instead of the hash → `ErrKeyNotFound` for every block, and `StateAtBlockNumber` breaks. `TestStateRootByNumberIfStateRetained` fails on this branch because of it.

Fix: `hash, state, err := core.GetStateRootAndHashByBlockNumber(r, blockNumber)`

0xsamalt: Thanks for pointing it out! Fixing it to `hash, state, err := core.GetStateRootAndHashByBlockNumber(r, blockNumber)` and re-running `TestStateRootByNumberIfStateRetained` locally before pushing.

0xsamalt:
Fixed the swapped binding bug you flagged, Ehsan confirmed `TestStateRootByNumberIfStateRetained` passes now.

@RafaelGranza , re: the benchmarks I ran them properly interleaved/CPU-pinned but couldn't get a clean signal beyond noise, wrote up the details in the PR. I only have the one machine to test on, so I can't rule out local noise myself. If there's a way to get this benchmarked somewhere more controlled, happy to do that. Otherwise I'd lean on the correctness/clarity argument (one read+unmarshal instead of two) rather than force a number that isn't really there.

EgeCaner:
Thank you for your contribution! This has been addressed in #3835.

PR #3809, opened 2026-07-10
perf(rpc): update resolveBlockRange to use block numbers instead of headers

rodrodros:
The header is unnecessarily loaded when finding the block range to which iterate. Just by loading the block number we reduced the load from the micro-seconds to the nano-seconds. We also allocate and use much less memory.

| subscription `block_id` | read | time/op | B/op | allocs/op |
|---|---|---|---:|---:|
| `latest` / omitted | old — `HeadsHeader()` | 4.15 µs | 2,681 | 25 |
| | new — `Height()` | 385 ns (**~11× faster**) | 32 | 3 |
| `block_hash` | old — `BlockHeaderByHash()` | 4.13 µs | 2,729 | 25 |
| | new — `BlockNumberByHash()` | 413 ns (**~10× faster**) | 72 | 3 |
| `block_number` | old — `BlockHeaderByNumber()` | 3.68 µs | 2,657 | 22 |
| | new — *(no DB read)* | 0 | 0 | 0 |

### DIFF
--- a/rpc/v10/subscription_events.go
+++ b/rpc/v10/subscription_events.go
@@ -77,7 +77,7 @@ func newEventSubscriber(
 		}
 	}
 
-	requestedHeader, headHeader, rpcErr := handler.resolveBlockRange(blockID)
+	startBlock, latestBlock, rpcErr := handler.resolveBlockRange(blockID)
 	if rpcErr != nil {
 		return subscriber{}, rpcErr
 	}
@@ -103,14 +103,14 @@ func newEventSubscriber(
 			return state.processHistoricalEvents(
 				ctx,
 				id,
-				requestedHeader.Number,
-				headHeader.Number,
+				startBlock,
+				latestBlock,
 				fromAddrs,
 				keys,
 			)
 		},
-		onReorg:   state.onReorg,
 		onNewHead: state.onNewHead,
+		onReorg:   state.onReorg,
 	}
 	if finalityStatus != nil && *finalityStatus == TxnFinalityStatusWithoutL1(TxnPreConfirmed) {
 		s.onPreConfirmed = state.onPreConfirmed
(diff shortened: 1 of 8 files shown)


rpc/v10/subscription_events.go
```diff
@@ -103,14 +103,14 @@ func newEventSubscriber(
...
-				headHeader.Number,
+				startBlock,
+				latestBlock,
 				fromAddrs,
 				keys,
 			)
 		},
-		onReorg:   state.onReorg,
 		onNewHead: state.onNewHead,
+		onReorg:   state.onReorg,
```
RafaelGranza, line 113: I guess it's a Linter thing.

rpc/v10/subscriptions.go
```diff
@@ -166,31 +166,36 @@ func filterTxBySender(txn core.Transaction, senderAddr []felt.Address) bool {
...
-		return nil, nil, rpcErr
+	var startBlock uint64
+	if blockID.IsHash() {
+		startBlock, err = h.bcReader.BlockNumberByHash(blockID.Hash())
+		if err != nil {
+			return 0, 0, rpccore.ErrInternal.CloneWithData(err.Error())
+		}
+	} else {
+		startBlock = blockID.Number()
 	}
```
RafaelGranza, line 198: Fair points

rpc/v10/subscription_heads.go
```diff
@@ -80,35 +81,32 @@ func (s *headsSubscriberState) onNewHead(
...
 	ctx context.Context,
 	id string,
-	startHeader,
-	latestHeader *core.Header,
+	startBlock,
+	latestBlock uint64,
 ) error {
-	curHeader := startHeader
-	for {
+	for currentBlock := startBlock; currentBlock != latestBlock; currentBlock += 1 {
```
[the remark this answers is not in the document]
RafaelGranza: Also Fair.

RafaelGranza (commented):
These changes make sense, they are a good optimization.

Before I can approve, you need to address all the comments and two more things:

1. Fix the mocks in `subscriptions_test.go`.
2. Add tests for the hash path (`BlockNumberByHash` + the `ErrBlockNotFound` mapping). No coverage there today and a test can save us from regression.

RafaelGranza (approved):
LGTM

PR #3810, opened 2026-07-11
perf(sync): avoid getting the l2 head until necessary

rodrodros:
This improvements are for the fast failing path when we use the latest head for the block number only. 

This way, we don't load the head until they are necessary and just executing the execution and failing early goes from 7 nano-seconds to 1 mili-seconds. Bigger win is on memory where we can avoid loading 2.5 KiB of memory (give or take) and allocations are reduced from 36 to 4. 

Small win but a win.

### DIFF
--- a/sync/sync.go
+++ b/sync/sync.go
@@ -31,6 +31,13 @@ const (
 	OpVerify = "verify"
 	OpStore  = "store"
 	OpFetch  = "fetch"
+
+	// Reorg-check ops, one per exit level of isReverting: resolved from the
+	// local height alone, after fetching the remote head, or only after also
+	// reading the local header for a hash comparison.
+	OpReorgCheckFast   = "reorgCheckFast"
+	OpReorgCheckRemote = "reorgCheckRemote"
+	OpReorgCheckLocal  = "reorgCheckLocal"
 )
 
 // This is a work-around. mockgen chokes when the instantiated generic type is in the interface.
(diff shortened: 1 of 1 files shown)


RafaelGranza (approved):
Thanks for the PR! I agree this is a real performance improvement, but the description confused me a bit.:

> `failing early goes from 7 nano-seconds to 1 mili-seconds`

- I think it's the other way around, right?
- 1ms seems a lot! 
- 7ns seems to small hahaha.

Anyway, I was missing the benchmark table, so I built one:
 
Benchmark | ns/op | B/op | allocs/op
-- | -- | -- | --
isReverting fast-path — BEFORE (main) | 4747 | 2680 | 25
isReverting fast-path — AFTER (this PR) | 141 | 32 | 3
HeadsHeader() alone | 4754 | 2680 | 25
Height() alone | 70 | 32 | 3

The number of allocations depend on block size*

I also checked the rare cold-cache possibility ( a new fresh Pebble instance per read, so every read is a block-cache miss):

Scenario | Latency | B/op | allocs/op
-- | -- | -- | --
In-memory, warm | ~5.0 µs | 2680 | 25
On-disk Pebble, warm block cache | ~6.0 µs | 2683 | 25
On-disk Pebble, cold block cache (reopen each read) | median 328µs (max 709µs)| — | —

Approving since the change is a clear improvement.

PR #3811, opened 2026-07-13
refactor(rpc): remove deprecated CANDIDATE transaction status

### LINKED ISSUE #3693: Remove candidate txs from PreConfirmed logic after Starknet v0.14.3 upgrade
Candidate txs will be deprecated in v0.14.3, so we should remove them from our code as soon as the new upgrade arrives on Mainnet

### LINKED ISSUE #3768: Remove CANDIDATE support for subscriptions
rpc/v{9,10}/subscription_transactions.go / subscription_types.go — "CANDIDATE" is still accepted by the subscription UnmarshalJSON but no longer wired to a handler; a client subscribing to CANDIDATE alone silently receives nothing. Reasonable given candidates no longer exist; consider rejecting the value or dropping the enum in a follow-up for clearer client feedback.

Ehsan-saradar:
Candidate transactions are deprecated in Starknet v0.14.3, which went live on Mainnet on 2026-07-08. The network no longer emits `finality_status: "CANDIDATE"`, so its parsing and serialization is now dead code — this drops it.

## Changes
- Remove `CANDIDATE` from `starknet.FinalityStatus` (including its `UnmarshalText` case, i.e. the feeder-gateway parse path).
- Remove `TxnStatusCandidate` from the RPC v9/v10 layers and realign the `TxnStatus` / `TxnFinalityStatus` enums.
- Drop the leftover `CANDIDATE` value from the `starknet_subscribeNewTransactions` finality-status enum, so it is rejected on unmarshal instead of being silently accepted (the candidate handler itself was already removed earlier — this is the follow-up from #3768).
- Update the `disable-received-txn-stream` flag help and docs.

Closes #3693
Closes #3768


rpc/v9/subscriptions.go
```diff
@@ -297,9 +297,6 @@ func (s *TxnStatusWithoutL1) UnmarshalText(text []byte) error {
 	case "RECEIVED":
 		*s = TxnStatusWithoutL1(TxnStatusReceived)
 		return nil
-	case "CANDIDATE":
```
EgeCaner, line 300: JSON-RPC v9/v10 spec both supports `Candidate` status for transaction subscription filter and rejecting `Candidate` status as input violates the spec. Even we do not serve candidate transactions anymore, we should keep `Candidate` as viable input for spec compliance.

rpc/v10/subscription_types.go
```diff
@@ -98,9 +98,6 @@ func (s *TxnStatusWithoutL1) UnmarshalText(text []byte) error {
 	case "RECEIVED":
 		*s = TxnStatusWithoutL1(TxnStatusReceived)
 		return nil
-	case "CANDIDATE":
```
EgeCaner, line 101: JSON-RPC v9/v10 spec both supports Candidate status for transaction subscription filter and rejecting Candidate status as input violates the spec. Even we do not serve candidate transactions anymore, we should keep Candidate as viable input for spec compliance.

rpc/v9/transaction.go
```diff
@@ -1059,10 +1056,6 @@ func AdaptTransactionStatus(txStatus *starknet.TransactionStatus) (TransactionSt
 		status.Finality = TxnStatusReceived
 	case starknet.PreConfirmed:
 		status.Finality = TxnStatusPreConfirmed
-	case starknet.Candidate:
```
EgeCaner, line 1062: Not sure about this but if Feeder Gateway is able to respond with Candidate Status when we fallback to querying feeder for status, after cannot found tx locally. Then we might like to keep this 

cc: @rodrodros

rpc/v9/subscriptions.go
EgeCaner: [ref](https://github.com/starkware-libs/starknet-specs/blob/c2e93098b9c2ca0423b7f4d15b201f52f22d8c36/api/starknet_ws_api.json#L275) to spec

Ehsan-saradar:
Good catch, Ege — thanks. I double-checked the spec: TXN_STATUS still includes CANDIDATE, and the subscription filter (TXN_STATUS_WITHOUT_L1) is just TXN_STATUS minus ACCEPTED_ON_L1, so CANDIDATE is still a valid input. Rejecting it would break spec compliance, and on the feeder side it's safer to keep parsing it too in case the gateway still returns it on the status fallback.

Since candidate transactions were already removed in #3732 and the candidate status needs to stay for the spec, there's nothing safe left to remove here — so I'll close this PR. Thanks again for the review!

PR #3819, opened 2026-07-15
refactor(core): replace ResourceBounds map with a fixed-field struct

### LINKED ISSUE #3742: core: replace `ResourceBounds map[Resource]ResourceBounds` with a fixed-field struct
Both `core` and `starknet` resource bounds are modeled as `map[Resource]ResourceBounds`. Since these keys are effectively fixed, a `struct` with explicit `L1Gas`/`L2Gas`/`L1DataGas` fields would be clearer and avoid map allocations. The caveat (and the reason this needs care, not a blind swap): the resource set is not constant across versions — historically v3 transactions carried only two resources (L1Gas, L2Gas) before L1DataGas was added — and `core.ResourceBounds` is a persisted core type written to the DB when storing transactions, so any change must preserve backward compatibility with already-stored data (encoding/migration considerations).

Ehsan-saradar:
Closes #3742

Both `core` and `starknet` resource bounds were modeled as `map[Resource]ResourceBounds`. Since the keys are effectively fixed, this replaces them with a struct with explicit `L1Gas` / `L2Gas` / `L1DataGas` fields — clearer, and it drops a map allocation per transaction.

### Approach

- **New `core.ResourceBoundsMap` struct** (`L1Gas` / `L2Gas` / `L1DataGas`) replacing the `map[Resource]ResourceBounds` field on the v3 transaction types, and the same for the feeder `starknet` type.
- **No DB migration.** The struct implements `MarshalCBOR` / `UnmarshalCBOR` that keep the exact legacy `map[Resource]ResourceBounds` wire format. A golden test pins byte-for-byte equality across the three historical shapes — all three resources, the pre-0.13.4 two-resource form (no `l1_data_gas`), and the nil map on v0–v2 — so already-stored bytes still decode into the struct.
- **Hashes unchanged.** `l1_data_gas` only enters the tip/resources hash when it's actually set (non-nil `MaxPricePerUnit`), which reproduces the old map-key check, so pre-0.13.4 tx hashes stay identical.
- **Feeder JSON unchanged.** The `starknet` struct keeps the uppercase `L1_GAS` / `L2_GAS` / `L1_DATA_GAS` keys and uses `omitzero`, so an absent `l1_data_gas` round-trips both ways.
- Left the `vm` resource-bounds map as-is — it's the marshal-only Blockifier payload (with the short `L1_DATA` key), out of scope for this one.

### Testing

- `make lint` — pass
- Tests for all affected packages (`core`, `starknet`, `adapters/...`, `rpc/...`, `vm`, `genesis`, `consensus`, `clients/feeder`) — pass, including the new backward-compat golden tests in `core/resource_bounds_test.go` and `starknet/resource_bounds_test.go`.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -44,6 +44,13 @@ linters:
           deny:
             - pkg: github.com/sirupsen/logrus
               desc: use github.com/NethermindEth/juno/utils for logging
+    exhaustruct:
+      # These types have a meaningful zero value: a transaction with no resource
+      # bounds (v0-v2). Requiring every field to be spelled out at construction
+      # would only add noise, so exclude them from exhaustive-struct checking.
+      exclude:
+        - 'github.com/NethermindEth/juno/core\.ResourceBounds$'
+        - 'github.com/NethermindEth/juno/core\.ResourceBoundsMap$'
     funlen:
       lines: 120
       statements: 50
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -44,13 +44,6 @@ linters:
           deny:
             - pkg: github.com/sirupsen/logrus
               desc: use github.com/NethermindEth/juno/utils for logging
-    exhaustruct:
-      # These types have a meaningful zero value: a transaction with no resource
-      # bounds (v0-v2). Requiring every field to be spelled out at construction
-      # would only add noise, so exclude them from exhaustive-struct checking.
-      exclude:
-        - 'github.com/NethermindEth/juno/core\.ResourceBounds$'
-        - 'github.com/NethermindEth/juno/core\.ResourceBoundsMap$'
     funlen:
       lines: 120
       statements: 50
--- a/adapters/core2p2p/transaction.go
+++ b/adapters/core2p2p/transaction.go
@@ -131,11 +131,11 @@ func adaptResourceLimits(bounds core.ResourceBounds) *transaction.ResourceLimits
 	}
 }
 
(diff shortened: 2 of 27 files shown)


.golangci.yaml
```diff
@@ -44,6 +44,13 @@ linters:
           deny:
             - pkg: github.com/sirupsen/logrus
               desc: use github.com/NethermindEth/juno/utils for logging
+    exhaustruct:
+      # These types have a meaningful zero value: a transaction with no resource
+      # bounds (v0-v2). Requiring every field to be spelled out at construction
+      # would only add noise, so exclude them from exhaustive-struct checking.
+      exclude:
+        - 'github.com/NethermindEth/juno/core\.ResourceBounds$'
+        - 'github.com/NethermindEth/juno/core\.ResourceBoundsMap$'
```
rodrodros: Please remove the lint ignores

Ehsan-saradar: Done — the exclude is gone and the repo is back to zero lint ignores.

Rather than spelling the zero fields out at all 14 call sites, I gave the zero values a name, following the existing `core.EmptyStateDiff()` helper: `core.EmptyResourceBounds()` and `core.EmptyResourceBoundsMap()`. Both list every field explicitly, so a resource added later still surfaces there, and exhaustruct keeps checking every real construction site.

No behaviour change — the returned values are identical to the literals they replace. `gofmt`, `go build`, `golangci-lint run` and the full test suite are green locally.

Mind approving the workflows when you get a chance? They are sitting on `action_required`.

PR #3821, opened 2026-07-15
feat(sync): resolve declared classes for pre-confirmed blocks

EgeCaner:
Adds the ability to fetch and register class definitions declared in past pre-confirmed blocks (PreLatest blocks), which were previously left unresolved until finalization. The poller now resolves each backfilled slot's declared classes and stores them on the exact block that declares them, and pre-confirmed state reads merge classes across the whole chain instead of returning nil.

### DIFF
--- a/sync/preconfirmed/chain_storage.go
+++ b/sync/preconfirmed/chain_storage.go
@@ -167,19 +167,6 @@ func (c *ChainReader) PreConfirmedStateAt(
 	return pending.NewState(&stateDiff, newClasses, base, blockNumber), closer, nil
 }
 
-func mergeClassesInto(
-	dst, src map[felt.Felt]core.ClassDefinition,
-) map[felt.Felt]core.ClassDefinition {
-	if len(src) == 0 {
-		return dst
-	}
-	if dst == nil {
-		dst = make(map[felt.Felt]core.ClassDefinition, len(src))
-	}
-	maps.Copy(dst, src)
-	return dst
-}
-
 // PreConfirmedStateBeforeIndexAt returns the chain's view of state immediately
 // before transaction `index` at blockNumber. See PreConfirmedStateAt for the base-
 // resolution contract; here the chain additionally layers the target slot's
@@ -344,9 +331,9 @@ func (s *ChainStorage) ApplyUpdate(
 	return affected, nil
 }
 
-// withClasses returns base with the entries of extra added, without mutating base's map
+// mergeClassesCopying returns base with the entries of extra added, without mutating base's map
 // (copy-on-write). Returns base unchanged when extra is empty.
-func withClasses(
+func mergeClassesCopying(
 	base, extra map[felt.Felt]core.ClassDefinition,
 ) map[felt.Felt]core.ClassDefinition {
 	if len(extra) == 0 {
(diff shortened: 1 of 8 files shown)


sync/preconfirmed/chain_storage.go
```diff
@@ -155,13 +156,28 @@ func (c *ChainReader) PreConfirmedStateAt(
...
+) map[felt.Felt]core.ClassDefinition {
+	if len(src) == 0 {
+		return dst
+	}
+	if dst == nil {
+		dst = make(map[felt.Felt]core.ClassDefinition, len(src))
+	}
+	maps.Copy(dst, src)
+	return dst
 }
```
rodrodros: Only nit for this function is that seems to be placed a bit randomly. I would suggest to place it at the bottom (since it is not very important for the file).

sync/preconfirmed/chain_storage.go
```diff
@@ -155,13 +156,28 @@ func (c *ChainReader) PreConfirmedStateAt(
...
+}
+
+func mergeClassesInto(
+	dst, src map[felt.Felt]core.ClassDefinition,
+) map[felt.Felt]core.ClassDefinition {
+	if len(src) == 0 {
+		return dst
+	}
+	if dst == nil {
+		dst = make(map[felt.Felt]core.ClassDefinition, len(src))
```
rodrodros: performance nit: for this specific case calling `maps.Clone` on `dst` and returning it will be slightly faster.

Also helps readability a little bit. 

Feel free to ignore.

sync/preconfirmed/chain_storage.go
```diff
@@ -318,6 +344,22 @@ func (s *ChainStorage) ApplyUpdate(
 	return affected, nil
 }
 
+// withClasses returns base with the entries of extra added, without mutating base's map
+// (copy-on-write). Returns base unchanged when extra is empty.
+func withClasses(
```
rodrodros: nit: A slight readability improvement can be adding a verb to this function name. Suggestions: `cowNewClasses`?

sync/preconfirmed/chain_storage.go
```diff
@@ -318,6 +344,22 @@ func (s *ChainStorage) ApplyUpdate(
...
+	base, extra map[felt.Felt]core.ClassDefinition,
```
rodrodros, line 337: Aren't we using the `felt.SierraHash` type here because of legacy code?

sync/preconfirmed/poller.go
```diff
@@ -197,33 +206,141 @@ func (p *Poller) backfill(
...
+	}
+
+	declaredCount := declaredClassCount(storedStateDiff, updateStateDiffs)
+	if declaredCount == 0 {
+		return nil, nil
+	}
+
+	classes := make(map[felt.Felt]core.ClassDefinition, declaredCount)
+	for classHash := range declaredClassHashes(storedStateDiff, updateStateDiffs) {
+		class, err := p.dataSource.Class(ctx, classHash)
```
rodrodros, line 284: I don't think this approach is good. Ideally we would have an async/await mechanism (for Go) to make this requests for the feeder. This a behaviour repeats across the whole sync package that can be probably improved. Something to note on this PR but not to fix on this one

Sn improvement to be done next. Every I/O interactions to network and to disk should be done asynchronously. The hard thing is to design it in an efficient way (not generating too much garbage). 

cc @infrmtcs

rodrodros (approved):

PR #3833, opened 2026-07-17
perf(blockchain): read only hash from db (and not full header)

rodrodros:


brbrr (approved):
LGTM. But let's wait for a green CI.

EgeCaner (approved):

PR #3835, opened 2026-07-17
refactor(core): use lean header accessors

brbrr:
Note: pruner.HeaderByNumberIfStateRetained / HeaderByHashIfStateRetained are now prod-unused. Not sure if we will potentially need it in the future.

### DIFF
--- a/blockchain/statebackend/statebackend.go
+++ b/blockchain/statebackend/statebackend.go
@@ -34,12 +34,15 @@ func (b *stateBackend) HeadState() (core.StateReader, StateCloser, error) {
 func (b *stateBackend) StateAtBlockNumber(
 	blockNumber uint64,
 ) (core.StateReader, StateCloser, error) {
-	header, err := pruner.HeaderByNumberIfStateRetained(b.database, blockNumber)
+	if err := pruner.RequireStateRetainedByBlockNumber(b.database, blockNumber); err != nil {
+		return nil, nil, err
+	}
+	stateRoot, err := core.GetGlobalStateRootByBlockNumber(b.database, blockNumber)
 	if err != nil {
 		return nil, nil, err
 	}
 
-	history, err := state.NewStateHistory(blockNumber, header.GlobalStateRoot, b.stateDB)
+	history, err := state.NewStateHistory(blockNumber, stateRoot, b.stateDB)
 	if err != nil {
 		return nil, nil, err
 	}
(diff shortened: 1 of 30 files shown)


blockchain/statebackend/statebackend.go
```diff
@@ -34,12 +34,15 @@ func (b *stateBackend) HeadState() (core.StateReader, StateCloser, error) {
 func (b *stateBackend) StateAtBlockNumber(
 	blockNumber uint64,
 ) (core.StateReader, StateCloser, error) {
-	header, err := pruner.HeaderByNumberIfStateRetained(b.database, blockNumber)
```
brbrr, line 37: `HeaderByNumberIfStateRetained` become unused in prod. I've kept it for a time being but let me know if I should remove it.

blockchain/statebackend/statebackend.go
```diff
@@ -57,12 +60,16 @@ func (b *stateBackend) StateAtBlockHash(
 		return st, NoopStateCloser, nil
 	}
 
-	header, err := pruner.HeaderByHashIfStateRetained(b.database, blockHash)
```
brbrr, line 60: `HeaderByHashIfStateRetained` become unused in prod. I've kept it for a time being but let me know if I should remove it.

blockchain/statebackend/statebackend.go
```diff
@@ -34,12 +34,15 @@ func (b *stateBackend) HeadState() (core.StateReader, StateCloser, error) {
 func (b *stateBackend) StateAtBlockNumber(
 	blockNumber uint64,
 ) (core.StateReader, StateCloser, error) {
-	header, err := pruner.HeaderByNumberIfStateRetained(b.database, blockNumber)
+	if err := pruner.RequireStateRetainedByBlockNumber(b.database, blockNumber); err != nil {
+		return nil, nil, err
+	}
+	stateRoot, err := core.GetGlobalStateRootByBlockNumber(b.database, blockNumber)
```
EgeCaner: `pruner.RequireStateRetainedByBlockNumber` internally calls `GetBlockHeaderHashByNumber` and only parses hash, maybe we should also get the stateRoot in one read. Not sure how effective will be since I guess blob will also be cached

rpc/v10/sync_test.go
```diff
@@ -40,15 +42,15 @@ func TestSyncing(t *testing.T) {
 
 	synchronizer.EXPECT().HighestBlockHeader().Return(nil).Times(2)
 	t.Run("undefined highest block", func(t *testing.T) {
-		mockReader.EXPECT().BlockHeaderByNumber(startingBlock).Return(&core.Header{}, nil)
+		mockReader.EXPECT().BlockHeaderHashByNumber(startingBlock).Return(nil, nil)
```
EgeCaner: Not related with this PR but returning nil, nil is nor realistic for this API, I assume test uses it as don't care value due to shortcuit. Not sure worth changing rn but wanted to mention

core/accessors.go
```diff
@@ -362,6 +363,29 @@ func GetBlockHeaderHashByNumber(r db.KeyValueReader, blockNum uint64) (*felt.Fel
 	return header.Hash, nil
 }
 
+func GetBlockHeaderTimestampByNumber(r db.KeyValueReader, blockNum uint64) (uint64, error) {
```
EgeCaner, line 381: As I see these accessors are indirectly tested, it might worth having their own tests as well

rpc/v10/sync_test.go
rodrodros: Makes sense

blockchain/statebackend/statebackend.go
rodrodros: let's remove it! cc @EgeCaner

blockchain/statebackend/statebackend.go
rodrodros: Let's remove it as well. cc @EgeCaner

EgeCaner (approved):
Looks good to me, thank you!

PR #3855, opened 2026-07-23
feat(sync): measure total block processing time

### LINKED ISSUE #3849: logs: collect block processing metrics
It would benefit testing and development to have a clear view into block processing time. The idea is to measure block processing time from when it got attempted to get fetched, up until it gets written to db. As well as all the meaningful intermediate steps along the way.

Ehsan-saradar:
Adds a `total` sync step that measures a block's whole processing time: from the first fetch attempt until it is written to the database, including the time it spends queued between the fetch, verify and store stages.

Until now we only had per-stage timers (`fetch`, `verify`, `store`), so the queueing time between stages was invisible and the numbers never added up to the wall-clock time a block actually took.

- `sync_timers` picks the new step up automatically under `op="total"` (it is already a `HistogramVec` labelled by `op`, so no metric registration changes were needed).
- The duration is also added to the `Stored Block` log line as `took`.

Closes #3849

### DIFF
--- a/sync/sync.go
+++ b/sync/sync.go
@@ -32,6 +32,11 @@ const (
 	OpStore  = "store"
 	OpFetch  = "fetch"
 
+	// OpTotal spans a block's whole processing time: from the first fetch
+	// attempt until it is written to the database, including time spent
+	// queued between the fetch, verify and store stages.
+	OpTotal = "total"
+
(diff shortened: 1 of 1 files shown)


rodrodros:
Thanks for the PR, we will not go forward with this right now

PR #3856, opened 2026-07-23
refactor(cmd): use signal.NotifyContext for graceful shutdown

kirugan:
I found this function few months ago and always wanted to fix it in Juno. When we were writing `juno.go` long time ago we didn't know about it.


cmd/juno/juno.go
```diff
@@ -350,7 +344,7 @@ func main() {
 	})
 
 	if err := cmd.ExecuteContext(ctx); err != nil {
-		cancel()
+		fmt.Printf("error: %v\n", err)
 		os.Exit(1)
```
kirugan, line 350: `defer cancel()` won't be called once you reach `os.Exit(...)` so it doesn't actually matter. But I kept `defer cancel()` as it's good practice and later on you can modify "success" path and it will still work

cmd/juno/juno.go
```diff
@@ -314,18 +314,12 @@ var Version string
...
 
-	quit := make(chan os.Signal, 1)
-	signal.Notify(quit, os.Interrupt, syscall.SIGTERM)
-
-	ctx, cancel := context.WithCancel(context.Background())
-	go func() {
-		<-quit
-		cancel()
-	}()
+	ctx, cancel := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
```
kirugan, line 321: if you look at impl. it's essentially the same

rodrodros (commented):
Thank you!

cmd/juno/juno.go
```diff
@@ -350,7 +344,7 @@ func main() {
 	})
 
 	if err := cmd.ExecuteContext(ctx); err != nil {
-		cancel()
+		fmt.Printf("error: %v\n", err)
```
rodrodros, line 347: @kirugan could you please add a call to `cancel()` here to keep the linter happy 🙏

kirugan: sure, added

kirugan:
@rodrodros please take a look

PR #3858, opened 2026-07-24
refactor(core): use `[]felt.Felt` for transaction, event and message fields

RafaelGranza:
Replaces `[]*felt.Felt` for `[]felt.Felt` in transaction, event and message fields.

The main goal is to reduce GC pressure, improvecache locality, and tie the array of felts' lifespan to the objcted it lives in.

This PR introduces extra copies before hashing due to signature conflicts. These are [addressed in this PR.](https://github.com/NethermindEth/juno/pull/3859)


core/transaction.go
```diff
@@ -855,8 +864,8 @@ func EventsBloom(receipts []*TransactionReceipt) *bloom.BloomFilter {
 		for _, event := range receipt.Events {
 			fromBytes := event.From.Bytes()
 			filter.TestOrAdd(fromBytes[:])
-			for index, key := range event.Keys {
-				keyBytes := key.Bytes()
+			for index := range event.Keys {
+				keyBytes := event.Keys[index].Bytes()
```
RafaelGranza, line 868: This is a performance change.
Just to avoid creating an extra copy while iterating over the keys.
Does not happen in other hot places.

rodrodros (approved):

PR #3859, opened 2026-07-24
refactor(crypto): accept `[]felt.Felt` in Pedersen array and digest hashing

RafaelGranza:
Support for hashing `[]felt.Felt` on `Pedersen` and digest `Update`, so that hash calls can be zero-copy.

Also, `utils.ToPtrSlice` is removed since it is no longer used in the codebase.

### DIFF
--- a/core/crypto/pedersen_hash_test.go
+++ b/core/crypto/pedersen_hash_test.go
@@ -101,10 +101,31 @@ func TestPedersenArray(t *testing.T) {
 		}
 		digestWhole.Update(data...)
 		want := felt.UnsafeFromString[felt.Felt](tt.want)
-		got := crypto.PedersenArray(data...)
-		assert.Equal(t, want, got)
+		// PedersenArray (value slice) must match the pointer version exactly.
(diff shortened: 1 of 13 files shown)


core/crypto/pedersen_hash_test.go
```diff
@@ -149,52 +170,56 @@ var benchHashR felt.Felt
...
 	}
 }
 
-func BenchmarkPedersen(b *testing.B) {
-	randFelts := genRandomFeltPairs(b)
-	var f felt.Felt
-	b.ResetTimer()
-	for n := range b.N {
-		f = crypto.Pedersen(randFelts[n][0], randFelts[n][1])
+func BenchmarkPedersenElems(b *testing.B) {
```
rodrodros: Let's move the benchmarks to a dedicated pedersen_benchmark_test.go

Also, noticing that `_hash` suffix for the scripts in the `crypto` package can probably be stripped from all the files here (something to do after).

rodrodros (approved):
LGTM

PR #3861, opened 2026-07-24
chore(crypto): drop `_hash` suffix from `crypto` file names

### LINKED ISSUE #3860: Remove `_hash` suffix from files
`_hash` suffix for the scripts in the crypto package can probably be stripped from all the files

RafaelGranza:
Closes [Issue](https://github.com/NethermindEth/juno/issues/3860)

Also, there was a complete obsolete test file that I moved to the right place


core/crypto/poseidon_test.go
```diff
@@ -84,3 +84,18 @@ func TestPoseidonDigestUpdateArrayMatchesUpdate(t *testing.T) {
 		assert.Equal(t, byElems.Finish(), byArray.Finish())
 	}
 }
+
+// Test vectors from https://github.com/starkware-industries/poseidon
```
RafaelGranza, line 88: this test was sitting in its single file. no more need for that

brbrr (approved):

PR #3864, opened 2026-07-27
fix(rpc): report correct sync status in pruned mode

### LINKED ISSUE #3791: starknet_syncing incorrectly reports "not syncing" in pruned mode
# Summary

In pruned mode, Handler.Syncing (rpc/(v10/v9/v8)/sync.go) can return false (not syncing) while the node is in fact still syncing. This happens because it reads the starting block's header solely to obtain that block's hash, and in pruned mode that header may have already been pruned, so the read fails and the handler falls through to the default "not syncing" response.

### Details

The handler resolves the starting block hash like this:
```go
  startingBlockNumber, err := h.syncReader.StartingBlockNumber()
  if err != nil {
      return defaultSyncState, nil
  }
  startingBlockHeader, err := h.bcReader.BlockHeaderByNumber(startingBlockNumber)
  if err != nil {
      return defaultSyncState, nil   // <-- pruned starting header lands here
  }
```

 `BlockHeaderByNumber(startingBlockNumber)` is called only to read `startingBlockHeader.Hash` (and `.Number`, which we already have). When the node runs with history pruning enabled and has been syncing long enough that the starting block falls outside the retained window, that header no longer exists in the DB. The lookup returns an error, and every error path in this function returns `defaultSyncState`,  i.e. Syncing: false.

The result: a node that is actively syncing reports itself as fully synced, which is misleading to clients and breaks anything that polls `starknet_syncing` to gate readiness.

cicr99:
Fixes #3791 

`starknet_syncing` previously read the starting block hash from blockchain storage using the sync start block number. In pruned mode, that starting block header may have already been removed, causing the RPC method to return `false` even while the node is still syncing.

This PR stores the starting block header in the synchronizer once the starting block is successfully stored, and updates RPC v8/v9/v10 to read the starting block metadata from the synchronizer instead of prunable chain history.

## Changes

- Add `StartingBlockHeader()` to the sync reader interface.
- Cache the starting block header in `Synchronizer` when the sync start block is stored.
- Clear the cached starting header when the sync loop exits.
- Update `starknet_syncing` in RPC v8/v9/v10 to use the synchronizer’s cached start header.
- Update mocks, fake syncers, sequencer implementation, and RPC sync tests.

### DIFF
--- a/rpc/v10/sync.go
+++ b/rpc/v10/sync.go
@@ -38,10 +38,6 @@ func (s Sync) MarshalJSON() ([]byte, error) {
 func (h *Handler) Syncing() (*Sync, *jsonrpc.Error) {
 	defaultSyncState := &Sync{Syncing: new(bool)}
 
-	startingBlockHeader, err := h.syncReader.StartingBlockHeader()
-	if err != nil {
-		return defaultSyncState, nil
-	}
 	head, err := h.bcReader.HeadsHeader()
 	if err != nil {
 		return defaultSyncState, nil
@@ -51,6 +47,10 @@ func (h *Handler) Syncing() (*Sync, *jsonrpc.Error) {
 		return defaultSyncState, nil
 	}
 	if highestBlockHeader.Number <= head.Number {
+		return defaultSyncState, nil
+	}
+	startingBlockHeader, err := h.syncReader.StartingBlockHeader()
+	if err != nil || startingBlockHeader == nil {
 		return defaultSyncState, nil
 	}
 
--- a/rpc/v10/sync.go
+++ b/rpc/v10/sync.go
@@ -50,7 +50,7 @@ func (h *Handler) Syncing() (*Sync, *jsonrpc.Error) {
 		return defaultSyncState, nil
 	}
 	startingBlockHeader, err := h.syncReader.StartingBlockHeader()
-	if err != nil || startingBlockHeader == nil {
+	if err != nil {
 		return defaultSyncState, nil
 	}
 
--- a/rpc/v10/sync.go
+++ b/rpc/v10/sync.go
@@ -38,11 +38,7 @@ func (s Sync) MarshalJSON() ([]byte, error) {
 func (h *Handler) Syncing() (*Sync, *jsonrpc.Error) {
 	defaultSyncState := &Sync{Syncing: new(bool)}
 
-	startingBlockNumber, err := h.syncReader.StartingBlockNumber()
-	if err != nil {
-		return defaultSyncState, nil
-	}
-	startingBlockHash, err := h.bcReader.BlockHeaderHashByNumber(startingBlockNumber)
+	startingBlockHeader, err := h.syncReader.StartingBlockHeader()
 	if err != nil {
 		return defaultSyncState, nil
 	}
@@ -59,8 +55,8 @@ func (h *Handler) Syncing() (*Sync, *jsonrpc.Error) {
 	}
 
 	return &Sync{
-		StartingBlockHash:   startingBlockHash,
-		StartingBlockNumber: &startingBlockNumber,
+		StartingBlockHash:   startingBlockHeader.Hash,
+		StartingBlockNumber: &startingBlockHeader.Number,
 		CurrentBlockHash:    head.Hash,
 		CurrentBlockNumber:  &head.Number,
 		HighestBlockHash:    highestBlockHeader.Hash,
--- a/rpc/v10/sync_test.go
+++ b/rpc/v10/sync_test.go
@@ -34,18 +34,6 @@ func TestSyncing(t *testing.T) {
 		assert.Equal(t, &rpc.Sync{Syncing: &defaultSyncState}, syncing)
 	})
 
-	t.Run("nil starting block header", func(t *testing.T) {
-		mockReader.EXPECT().HeadsHeader().Return(&core.Header{Number: 1}, nil)
-		synchronizer.EXPECT().HighestBlockHeader().Return(
-			&core.Header{Number: 2, Hash: felt.NewFromUint64[felt.Felt](2)},
-		)
-		synchronizer.EXPECT().StartingBlockHeader().Return(nil, nil)
-
-		syncing, err := handler.Syncing()
-		assert.Nil(t, err)
-		assert.Equal(t, &rpc.Sync{Syncing: &defaultSyncState}, syncing)
-	})
-
 	t.Run("empty blockchain", func(t *testing.T) {
 		mockReader.EXPECT().HeadsHeader().Return(nil, errors.New("empty blockchain"))
 
--- a/rpc/v10/sync_test.go
+++ b/rpc/v10/sync_test.go
@@ -22,29 +22,30 @@ func TestSyncing(t *testing.T) {
 	defaultSyncState := false
 
 	startingBlockHeader := &core.Header{Number: 0, Hash: &felt.Zero}
-	mockReader.EXPECT().HeadsHeader().Return(&core.Header{Number: 1}, nil)
-	synchronizer.EXPECT().HighestBlockHeader().Return(
-		&core.Header{Number: 2, Hash: felt.NewFromUint64[felt.Felt](2)},
-	)
-	synchronizer.EXPECT().StartingBlockHeader().Return(nil, errors.New("nope"))
 	t.Run("undefined starting block header", func(t *testing.T) {
+		mockReader.EXPECT().HeadsHeader().Return(&core.Header{Number: 1}, nil)
+		synchronizer.EXPECT().HighestBlockHeader().Return(
+			&core.Header{Number: 2, Hash: felt.NewFromUint64[felt.Felt](2)},
+		)
+		synchronizer.EXPECT().StartingBlockHeader().Return(nil, errors.New("nope"))
+
 		syncing, err := handler.Syncing()
 		assert.Nil(t, err)
 		assert.Equal(t, &rpc.Sync{Syncing: &defaultSyncState}, syncing)
 	})
 
-	mockReader.EXPECT().HeadsHeader().Return(&core.Header{Number: 1}, nil)
-	synchronizer.EXPECT().HighestBlockHeader().Return(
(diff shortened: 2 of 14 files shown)


rpc/v8/sync.go
EgeCaner, line 41: Lets use partial decoder that only decodes block  num and hash from header, see https://github.com/NethermindEth/juno/blob/84de9c1f8eecb6abc7d9085dcebeecfbc8686ee9/core/accessors.go#L350

rpc/v8/sync.go
```diff
@@ -38,11 +38,7 @@ func (s Sync) MarshalJSON() ([]byte, error) {
 func (h *Handler) Syncing() (*Sync, *jsonrpc.Error) {
 	defaultSyncState := &Sync{Syncing: new(bool)}
 
-	startingBlockNumber, err := h.syncReader.StartingBlockNumber()
-	if err != nil {
-		return defaultSyncState, nil
-	}
-	startingBlockHeader, err := h.bcReader.BlockHeaderByNumber(startingBlockNumber)
+	startingBlockHeader, err := h.syncReader.StartingBlockHeader()
```
EgeCaner: nit: this can move below and only get read if need when not synced after `highestBlockHeader.Number <= head.Number` comparison. I won't expect much perf gains yet still reduces operations for synced path

sync/sync.go
```diff
@@ -115,6 +120,7 @@ type Synchronizer struct {
 	readOnlyBlockchain   bool
 	dataSource           DataSource
 	startingBlockNumber  *uint64
```
EgeCaner: Maybe retire `startingBlockNumber` since header contains that information and block number alone is not sufficient to reply `starknet_syncing`, thus I'm not sure publishing it before hash has any benefit.

rpc/v10/sync_test.go
```diff
@@ -21,19 +21,17 @@ func TestSyncing(t *testing.T) {
 	handler := rpc.New(mockReader, synchronizer, nil, nil)
 	defaultSyncState := false
 
-	startingBlock := uint64(0)
-	synchronizer.EXPECT().StartingBlockNumber().Return(startingBlock, errors.New("nope"))
-	t.Run("undefined starting block", func(t *testing.T) {
+	startingBlockHeader := &core.Header{Number: 0, Hash: &felt.Zero}
+	synchronizer.EXPECT().StartingBlockHeader().Return(nil, errors.New("nope"))
```
EgeCaner: Nit: Old test that carried over in time, I would prefer to move this mock setup inside the `t.Run` where it's used, expectations should be scoped to the subtest that relies on them, not the parent scope. It makes tests more readable.

rpc/v10/sync_test.go
```diff
@@ -21,19 +21,17 @@ func TestSyncing(t *testing.T) {
...
+	t.Run("undefined starting block header", func(t *testing.T) {
 		syncing, err := handler.Syncing()
 		assert.Nil(t, err)
 		assert.Equal(t, &rpc.Sync{Syncing: &defaultSyncState}, syncing)
 	})
 
-	synchronizer.EXPECT().StartingBlockNumber().Return(startingBlock, nil).AnyTimes()
+	synchronizer.EXPECT().StartingBlockHeader().Return(startingBlockHeader, nil).AnyTimes()
```
EgeCaner: Same here, I prefer to duplicate the mocking for each test

EgeCaner (changes requested):

sync/sync.go
cicr99: I've removed `StartingBlockNumber()` as it's no longer used after the new addition (only in tests, so removed those as well). However, removing `startingBlockNumber` altogether isn't as straight forward, as we discussed earlier, as this is the mechanism that is used for actually caching the starting header, or when a restart occurs

sync/sync.go
```diff
@@ -547,11 +556,37 @@ func (s *Synchronizer) revertHead(localHeader *core.Header) {
...
+	}
+	// Avoid overwriting a full header cached by storeTask while this fallback was reading from DB.
+	if !s.startingBlockHeader.CompareAndSwap(nil, header) {
+		header = s.startingBlockHeader.Load()
+		if header == nil {
+			return nil, errors.New("not running")
+		}
 	}
-	return *s.startingBlockNumber, nil
+	return header, nil
```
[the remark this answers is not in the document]
cicr99: I’d prefer to keep fallback caching because in `--remote-db` mode there is no `storeTask`, so the fallback is the only opportunity for the read-only node to retain the starting hash if the writer stores it and then pruning removes it.
I moved the active sync-run check before reading startingBlockHeader, so a cached fallback header cannot be returned once sync has stopped. I think that addresses the user-visible issue without needing a second post-CAS check. If sync finishes after the first check but before the method returns, returning the header is still acceptable because sync was active when the request was handled. Any cached value left behind after cleanup is inert, since later calls return not running before consulting the cache. So I’d prefer to keep the simpler version.

sync/sync_test.go
```diff
@@ -181,6 +182,165 @@ func TestSyncBlocks(t *testing.T) {
...
+		testDB,
+	)
+
+	storedStartingBlock := make(chan struct{}, 1)
+	synchronizer.WithListener(&sync.SelectiveListener{
+		OnSyncStepDoneCb: func(op string, blockNum uint64, took time.Duration) {
+			if op == sync.OpStore && blockNum == 0 {
+				storedStartingBlock <- struct{}{}
+			}
+		},
```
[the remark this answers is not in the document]
cicr99: I see the theoretical hazard, but I don’t think this test can realistically reach that path. The header is deleted only after the first OpStore for block 0 has been observed, and the test immediately reads StartingBlockHeader() and cancels the sync. It does not depend on, or attempt to exercise, any retry/reorg flow after that point. Since the channel is only used as a one-shot signal for the first store, I’d prefer to keep the callback simple unless we see this become flaky.

sync/sync.go
```diff
@@ -547,11 +556,38 @@ func (s *Synchronizer) revertHead(localHeader *core.Header) {
 	s.listener.OnReorg(localHeader.Number)
 }
 
-func (s *Synchronizer) StartingBlockNumber() (uint64, error) {
-	if s.startingBlockNumber == nil {
-		return 0, errors.New("not running")
+func (s *Synchronizer) StartingBlockHeader() (*core.Header, error) {
+	startingBlockNumber := s.startingBlockNumber.Load()
+	if startingBlockNumber == nil {
+		return nil, errors.New("not running")
```
EgeCaner: nit: better error message

sync/sync.go
```diff
@@ -547,11 +556,38 @@ func (s *Synchronizer) revertHead(localHeader *core.Header) {
...
+		return nil, err
+	}
+	header = &core.Header{
+		Number: *startingBlockNumber,
+		Hash:   hash,
+	}
+	// The sync loop may stop or restart while the fallback DB read is in flight.
+	// Only cache the fallback header if it still belongs to the same sync run.
+	if s.startingBlockNumber.Load() != startingBlockNumber {
+		return nil, errors.New("not running")
```
EgeCaner: nit: better error message

sync/sync.go
```diff
@@ -547,11 +556,38 @@ func (s *Synchronizer) revertHead(localHeader *core.Header) {
...
+	}
+	// Avoid overwriting a full header cached by storeTask while this fallback was reading from DB.
+	if !s.startingBlockHeader.CompareAndSwap(nil, header) {
+		header = s.startingBlockHeader.Load()
+		if header == nil {
+			return nil, errors.New("not running")
```
EgeCaner: nit: better error message

rpc/v10/sync.go
```diff
@@ -57,10 +49,14 @@ func (h *Handler) Syncing() (*Sync, *jsonrpc.Error) {
 	if highestBlockHeader.Number <= head.Number {
 		return defaultSyncState, nil
 	}
+	startingBlockHeader, err := h.syncReader.StartingBlockHeader()
+	if err != nil || startingBlockHeader == nil {
```
EgeCaner: nit: When err == nil we still nil-check the pointer, redundant for a normal (T, error) contract: non-nil error means don't touch the result, nil error should guarantee a non-nil result. If the callee can really return (nil, nil), that's the real smell, fix it at the source rather than guard at every call site. I guess this is not the case here, `StartingBlockHeader()` seems to be return err or valid ptr

sync/sync.go
```diff
@@ -547,11 +556,38 @@ func (s *Synchronizer) revertHead(localHeader *core.Header) {
...
+func (s *Synchronizer) StartingBlockHeader() (*core.Header, error) {
+	startingBlockNumber := s.startingBlockNumber.Load()
+	if startingBlockNumber == nil {
+		return nil, errors.New("not running")
+	}
+
+	header := s.startingBlockHeader.Load()
+	if header != nil {
+		return header, nil
+	}
```
EgeCaner, line 562: header read can be at top of this function, happy path does not require reading `startingBlockNumber`

sync/sync.go
```diff
@@ -547,11 +556,38 @@ func (s *Synchronizer) revertHead(localHeader *core.Header) {
...
+
+	hash, err := core.GetBlockHeaderHashByNumber(s.db, *startingBlockNumber)
+	if err != nil {
+		return nil, err
```
EgeCaner: nit: wrap err with context

sync/sync.go
```diff
@@ -547,11 +556,38 @@ func (s *Synchronizer) revertHead(localHeader *core.Header) {
...
+	}
+	header = &core.Header{
+		Number: *startingBlockNumber,
+		Hash:   hash,
+	}
```
EgeCaner: IMO we should either guarantee full header or make api return only number and hash. We can go with just reading full header above and using that, it will only run once I guess.

EgeCaner (changes requested):

sync/sync.go
EgeCaner: I made the change

EgeCaner (approved):
Looks good to me

PR #3872, opened 2026-07-28
feat(l1): [4] add hand-rolled ws-based JSON-RPC client

brbrr:
Depends on #3877

### DIFF
--- a/l1/eth/client/client.go
+++ b/l1/eth/client/client.go
@@ -1,0 +1,265 @@
+package client
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"math/big"
+	"net/url"
+	"sync"
+
+	"github.com/NethermindEth/juno/l1/eth"
+	"github.com/NethermindEth/juno/utils/log"
+	"go.uber.org/zap"
+	"golang.org/x/sync/singleflight"
+)
+
+// ErrClosed is terminal: the client was explicitly closed and will not redial.
+var ErrClosed = errors.New("client closed")
+
+type Client struct {
+	url  string
+	opts options
+
+	mu     sync.Mutex
+	tr     *wsTransport
+	closed bool
+
+	dials singleflight.Group
+}
+
+const BlockFinalized = "finalized"
+
+const maxQuantityNibbles = 64
+
+func New(ctx context.Context, rawURL string, opts ...Option) (*Client, error) {
+	u, err := url.Parse(rawURL)
+	if err != nil {
+		return nil, fmt.Errorf("parsing url: %w", err)
+	}
+	if u.Scheme != "ws" && u.Scheme != "wss" {
+		return nil, fmt.Errorf("unsupported url scheme %q (need ws/wss)", u.Scheme)
+	}
+	o := options{}
+	for _, opt := range opts {
+		opt(&o)
+	}
+	if o.logger == nil {
+		o.logger = log.NewNopZapLogger()
+	}
+	ws, err := dialWS(ctx, rawURL, o)
+	if err != nil {
+		return nil, err
+	}
+	return &Client{
+		url:  rawURL,
+		opts: o,
+		tr:   ws,
+	}, nil
+}
+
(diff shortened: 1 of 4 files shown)


l1/eth/client/rpc.go
```diff
@@ -0,0 +1,76 @@
...
+	Result  json.RawMessage `json:"result,omitempty"`
+	Error   *RPCError       `json:"error,omitempty"`
+}
+
+func (e *RPCError) Error() string {
+	if len(e.Data) > 0 {
+		return fmt.Sprintf("json-rpc error %d: %s: %s", e.Code, e.Message, e.Data)
+	}
+	return fmt.Sprintf("json-rpc error %d: %s", e.Code, e.Message)
+}
```
rodrodros: All these types are already implemented in the `jsonrpc` package, why re-implement them?

l1/eth/client/rpc.go
```diff
@@ -0,0 +1,76 @@
...
+	n, err := strconv.ParseUint(string(trimmed), 10, 64)
+	if err != nil {
+		return 0, fmt.Errorf("parsing id %q: %w", raw, err)
+	}
+	return n, nil
+}
+
+// decodeSubID rejects an empty id: it would register a sub under "" that no
+// notification could ever match.
+func decodeSubID(raw json.RawMessage) (string, error) {
```
rodrodros: Why are these defined here, if they are used only by `transport_ws`?

l1/eth/client/transport_ws.go
rodrodros, line 1: The PR has been split, but this one is till 2k lines of code with clear separation of components such as core and important ones like this

l1/eth/client/transport_ws.go
```diff
@@ -0,0 +1,505 @@
...
+		pingInterval: opts.pingInterval,
+		pingTimeout:  opts.pingTimeout,
+		closed:       make(chan struct{}),
+	}
+	loopCtx, cancel := context.WithCancel(context.Background())
+	t.cancelLoops = cancel
+	go t.readLoop(loopCtx) //nolint:gosec // G118: long-lived loop, not request-scoped
+	go t.pingLoop(loopCtx) //nolint:gosec // G118: long-lived loop, not request-scoped
+	return t, nil
+}
```
rodrodros, line 136: Why creating a context out of thin air and not re-using the other one?

I've read: https://github.com/NethermindEth/juno/pull/3807#discussion_r3664088835

Then don't pass a dial context right...., pass the dial context to the only necessary one.


This is clearly badly designed code

l1/eth/client/rpc.go
brbrr: My idea was that while these types look the same, they represent different RPCs - one for juno, anther for ethereum. So reusing them will create this additional coupling. So if at some point any of these will change, we'd need to update the second one, or split it. But I guess the chance is quite minimal. I'll reuse them.

l1/eth/client/transport_ws.go
brbrr: Note: Below is based on a recent commit, where I switched WS Client to become a long-lived, and handle re-dials internally. (Before that, it was a StateProvider's job, which was a bad placement). 

hm. let me explain. So the "parent ctx" is the one that is passed into `dialWS` (either on node startup (which is not needed, but provides a fail-fast resolution) and on _concurrent_ RPC/subscription requests). If two concurrent request experience a WS disconnection, they'll try to re-connect via `dialWS`, and will pass their (bounded) contexts. The "fastest" caller will trigger reconneciton, while others will wait (`singleflight`). 

This means that the context that we receive on reconnections is not long-lived, and can't be used to start the loops. We do, hovever, could receive a node-global context on startup, and store it in the client struct, this way we won't be creating the context "out of thin air", but I believe this considered a bad practice. 

FWIW, it's all boils down to the fact that underlaying WS library is context driven (while geth's one isn't). And since we want to have a long-lived WS connections and a ability to redial on disconnects, we need to create ctx from background.

PR #3889, opened 2026-07-30
perf(history): Check contract deployment only for slots with zero values

brbrr:
This PR prioritizes "happy" path over a missing storage slot. This means it speeds up reads for existing slots, but degrades the case of a missing contract - check `notfound` results below.

Non-zero values are unambiguously clear - we have the slot and the contract is deployed. For zero values, we need to check if that's an actual value or a missing contract.

I've run some tests against both pruned and non-pruned Sepolia snapshots, and got an exact match over ~12k lookups.

Bench results:

| case             | ns/op old | ns/op new | Δ ns/op | B/op old | B/op new | Δ B/op | allocs old | allocs new | Δ allocs |
|------------------|----------:|----------:|--------:|---------:|---------:|-------:|-----------:|-----------:|---------:|
| nonzero          |   279,511 |   189,680 |  −32.1% |    5,246 |    4,386 | −16.4% |         33 |         21 |   −36.4% |
| zero             |   288,086 |   286,786 |   −0.5% |    5,139 |    5,168 |  +0.6% |         32 |         32 |       0% |
| notfound         |    81,931 |   285,621 | +248.6% |      830 |    5,161 |  +522% |         12 |         32 |    +167% |

### DIFF
--- a/core/deprecatedstate/history.go
+++ b/core/deprecatedstate/history.go
@@ -2,6 +2,7 @@ package deprecatedstate
 
 import (
 	"errors"
+	"fmt"
 
 	"github.com/NethermindEth/juno/core"
 	"github.com/NethermindEth/juno/core/felt"
@@ -63,10 +64,10 @@ func (s *stateHistory) ContractStorage(addr, key *felt.Felt) (felt.Felt, error) 
 	val, err := s.state.ContractStorageAt(addr, key, s.blockNumber)
 	if err != nil {
 		if !errors.Is(err, ErrCheckHeadState) {
-			return felt.Felt{}, err
+			return felt.Felt{}, fmt.Errorf("reading storage history: %w", err)
 		}
 		if val, err = s.state.ContractStorage(addr, key); err != nil {
-			return felt.Felt{}, err
+			return felt.Felt{}, fmt.Errorf("reading head storage: %w", err)
 		}
 	}
 
--- a/core/deprecatedstate/history.go
+++ b/core/deprecatedstate/history.go
@@ -60,15 +60,23 @@ func (s *stateHistory) ContractNonce(addr *felt.Felt) (felt.Felt, error) {
 }
 
 func (s *stateHistory) ContractStorage(addr, key *felt.Felt) (felt.Felt, error) {
-	if err := s.checkDeployed(addr); err != nil {
-		return felt.Felt{}, err
-	}
-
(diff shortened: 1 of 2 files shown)


EgeCaner (approved):
Looks good to me! Thank you!

core/deprecatedstate/history.go
```diff
@@ -60,15 +60,23 @@ func (s *stateHistory) ContractNonce(addr *felt.Felt) (felt.Felt, error) {
...
-
 	val, err := s.state.ContractStorageAt(addr, key, s.blockNumber)
 	if err != nil {
-		if errors.Is(err, ErrCheckHeadState) {
-			return s.state.ContractStorage(addr, key)
+		if !errors.Is(err, ErrCheckHeadState) {
+			return felt.Felt{}, err
+		}
+		if val, err = s.state.ContractStorage(addr, key); err != nil {
+			return felt.Felt{}, err
```
EgeCaner: Nit: wrap with context

core/deprecatedstate/history.go
```diff
@@ -60,15 +60,23 @@ func (s *stateHistory) ContractNonce(addr *felt.Felt) (felt.Felt, error) {
...
-	if err := s.checkDeployed(addr); err != nil {
-		return felt.Felt{}, err
-	}
-
 	val, err := s.state.ContractStorageAt(addr, key, s.blockNumber)
 	if err != nil {
-		if errors.Is(err, ErrCheckHeadState) {
-			return s.state.ContractStorage(addr, key)
+		if !errors.Is(err, ErrCheckHeadState) {
+			return felt.Felt{}, err
```
EgeCaner: Nit: wrap with context

core/deprecatedstate/history.go
```diff
@@ -60,15 +60,23 @@ func (s *stateHistory) ContractNonce(addr *felt.Felt) (felt.Felt, error) {
...
+	}
+
+	// A non-zero value proves a write at or before blockNumber, and only
+	// deployed contracts are written to - skip the deployment probe.
+	if !val.IsZero() {
+		return val, nil
+	}
+
+	if err := s.checkDeployed(addr); err != nil {
 		return felt.Felt{}, err
```
EgeCaner, line 81: Nit: wrap with context

EgeCaner (commented):
Sorry just recognized these after approving. I guess wrapping error here would help

PR #3892, opened 2026-07-30
refactor(node): remove sequencer support (and other refactors)

rodrodros:


node/config.go
```diff
@@ -0,0 +1,101 @@
...
+	PprofPort                uint16           `mapstructure:"pprof-port"`
+	Colour                   bool             `mapstructure:"colour"`
+	PreConfirmedPollInterval time.Duration    `mapstructure:"preconfirmed-poll-interval"`
+	RemoteDB                 string           `mapstructure:"remote-db"`
+	VersionedConstantsFile   string           `mapstructure:"versioned-constants-file"`
+
+	// Sequencer support was removed and no CLI flag is registered for this
+	// key. It is still decoded from config files so that stale sequencer
+	// configs fail loudly in New instead of silently running a full node.
+	Sequencer bool `mapstructure:"seq-enable"`
```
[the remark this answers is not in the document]
rodrodros: Uff, not important

node/node.go
```diff
@@ -310,241 +171,140 @@ func New(cfg *Config, version string, logLevel *log.Level) (*Node, error) {
...
+		isRemoteDB,
+		database,
+	)
+	synchronizer.WithPlugin(junoPlugin)
 
-		feederGatewayDataSource := sync.NewFeederGatewayDataSource(chain, adaptfeeder.New(client))
-		synchronizer = sync.New(
+	if cfg.Network.GatewayURL == nil {
+		return nil, fmt.Errorf("network %q has no gateway URL configured", cfg.Network.Name)
+	}
```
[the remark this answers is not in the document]
rodrodros: I am not convinced of this one being a big deal. I prefer the tidiness of the checks being done when instatiating the relevant service

node/node.go
```diff
@@ -310,241 +171,140 @@ func New(cfg *Config, version string, logLevel *log.Level) (*Node, error) {
...
+	if cfg.P2P {
+		if cfg.Network == networks.Mainnet {
+			return nil, fmt.Errorf("P2P cannot be used on %v network", networks.Mainnet)
+		}
+		logger.Warn("P2P features enabled. Please note P2P is in experimental stage")
+
+		if !cfg.P2PFeederNode {
+			// Do not start the feeder synchronisation
+			synchronizer = nil
+		}
```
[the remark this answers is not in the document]
rodrodros: Valid, the solution is deleting the p2p from the cmd as well. Don't want to include that in this PR as well

node/sequencer.go
```diff
@@ -0,0 +1,148 @@
+package node
+
+// note(rdr): sequencer code that was cluttering the node.go initialization
+
+// func initalizeSequencer(
+// 	cfg *Config,
+// 	database db.KeyValueStore,
+// 	version string,
+// 	chain *blockchain.Blockchain,
+// 	compiler *ThrottledCompiler,
```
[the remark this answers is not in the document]
rodrodros: I prefer keeping the code, but I can delete it if someone thinks is not worth it

node/node.go
```diff
@@ -183,109 +85,71 @@ func New(cfg *Config, version string, logLevel *log.Level) (*Node, error) {
...
 		database, err = remote.New(
+			ctx,
 			cfg.RemoteDB,
-			context.TODO(),
 			logger,
 			grpc.WithTransportCredentials(insecure.NewCredentials()),
 		)
+		if err != nil {
+			return nil, fmt.Errorf("opening remote DB: %w", err)
+		}
```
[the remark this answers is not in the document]
rodrodros: This regression is not possible because `--remote-db` is for read only operations.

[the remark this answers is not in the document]
rodrodros: not an issue because `--remote-db` is read only

node/node.go
```diff
@@ -183,109 +85,71 @@ func New(cfg *Config, version string, logLevel *log.Level) (*Node, error) {
...
-	// L1 verification makes pruning unsafe — reject the combination up front
-	// instead of crashing later when the migration cannot find an L1 head.
-	if cfg.Prune && cfg.DisableL1Verification {
-		return nil, errors.New("--prune-mode requires L1 verification; " +
-			"remove --disable-l1-verification or disable --prune-mode")
-	}
-
-	dbIsRemote := cfg.RemoteDB != ""
+	isRemoteDB := cfg.RemoteDB != ""
 	var database db.KeyValueStore
```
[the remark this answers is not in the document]
rodrodros: answered already. `--remote-db` is read only

clients/feeder/feeder.go
```diff
@@ -182,7 +183,7 @@ func (c *Client) get(ctx context.Context, queryURL *url.URL) (io.ReadCloser, err
 			}
 
 			currentTimeout := timeouts.GetCurrentTimeout()
-			if currentTimeout >= mediumGrowThreshold {
+			if currentTimeout >= 2*time.Minute {
```
infrmtcs, line 186: I think keeping the constant make it easier to see and change it, and the name, while not the best name possible, can give us some ideas about what it is, while `2*time.Minute` doesn't auto translate to a meaningful reason.

clients/feeder/timeouts.go
```diff
@@ -70,15 +65,22 @@ func timeoutsListFromNumber(initial time.Duration) []time.Duration {
...
 
 func increaseDuration(prev time.Duration) time.Duration {
+	const (
+		fastGrowThreshold   = 1 * time.Minute
+		mediumGrowThreshold = 2 * time.Minute
+
+		growthFactorFast   = 2
+		growthFactorMedium = 1.5
+		growthFactorSlow   = 1.2
+	)
```
infrmtcs, line 82: I guess keeping this inside or outside of the function doesn't really change the perf, it just help to limit the usage to the function only.

node/node_test.go
```diff
@@ -150,14 +150,18 @@ func TestNetworkVerificationOnNonEmptyDB(t *testing.T) {
...
 			logLevel := log.NewLevel(log.INFO)
-			_, err = node.New(&node.Config{
-				DatabasePath:                       dbPath,
-				DBCompression:                      "zstd",
-				Network:                            test.network,
-				NewState:                           statetestutils.UseNewState(),
-				DisableL1Verification:              true,
-				SubmittedTransactionsCacheEntryTTL: time.Second,
-			}, "v0.1", logLevel)
+			_, err = node.New(t.Context(),
```
infrmtcs, line 153: nit: I think t.Context() should be in its own line.

node/config.go
infrmtcs, line 38: I think instead of keeping this in the config, we can use `MarkDeprecated` instead (see "Deprecating a flag or its shorthand" in https://pkg.go.dev/github.com/spf13/pflag#section-readme).
We can also use `MarkHidden` to hide it from user.

node/node.go
```diff
@@ -616,109 +378,51 @@ func New(cfg *Config, version string, logLevel *log.Level) (*Node, error) {
...
 
-		n.services = append(n.services, l1Client)
+		node.services = append(node.services, l1Client)
 		rpcHandler.WithL1Client(provider)
 	}
 
 	if semversion, err := semver.NewVersion(version); err == nil {
+		const upgraderDelay = 5 * time.Minute
+		const githubAPIUrl = "https://api.github.com/repos/NethermindEth/juno/releases/latest"
+		const latestReleaseURL = "https://github.com/NethermindEth/juno/releases/latest"
```
infrmtcs, line 413: Why do we move the constants here?

node/node.go
```diff
@@ -616,109 +378,51 @@ func New(cfg *Config, version string, logLevel *log.Level) (*Node, error) {
...
 
-	if !n.cfg.DisableL1Verification {
+	if !node.cfg.DisableL1Verification {
 		// Due to mutually exclusive flag we can do the following.
-		if n.cfg.EthNode == "" {
-			return nil, fmt.Errorf("ethereum node address not found; Use --disable-l1-verification flag if L1 verification is not required")
+		if node.cfg.EthNode == "" {
+			return nil, fmt.Errorf("ethereum node address not set; " +
+				"Use --disable-l1-verification flag if L1 verification is not required",
+			)
```
infrmtcs, line 396: Since we're touching this, I think we can use this instead:
https://pkg.go.dev/github.com/spf13/cobra#Command.MarkFlagsOneRequired

node/node.go
```diff
@@ -173,7 +59,23 @@ type Node struct {
...
+	version string,
+	logLevel *log.Level,
+) (*Node, error) {
+	if cfg.Sequencer {
+		return nil, errors.New("sequencer configuration is no longer supported")
+	}
+
+	// History pruning needs an L1-finalised cutoff to know which blocks are
+	// safe to drop. If no L1 client is given, new cutoffs cannot be set.
+	if cfg.Prune && cfg.DisableL1Verification {
```
infrmtcs, line 74: Since we're touching this, I think we should use this instead:
https://pkg.go.dev/github.com/spf13/cobra#Command.MarkFlagsMutuallyExclusive

node/sequencer.go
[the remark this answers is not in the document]
infrmtcs: Not sure if it's a better idea, but another way instead of commenting out is to add a build flag that we never enable.

infrmtcs (commented):
Please feel free to ignore the comments if any of them is too nit!

clients/feeder/timeouts.go
rodrodros: It's not meant as a performance change but as a legibiliy / scope one. Declare the constants where the are expected to be 🤔

node/node.go
rodrodros: Same idea of scoping them to where they are used. Global constants for for global usage. I believe is more tidy this way

PR #3901, opened 2026-08-03
ci(docs): scope docs-deploy trigger to relevant paths

### LINKED ISSUE #3897: docs-deploy.yml runs on every push to main regardless of changed files
Currently `.github/workflows/docs-deploy.yml` triggers on every push to `main` with no `paths` filter, so the docs site gets rebuilt and redeployed even when the change has nothing to do with `docs/` (e.g. a pure Go change in `core/` or `rpc/`).

The trigger can't simply be scoped to `docs/**` though — `docs/generate-config.js` fetches `cmd/juno/juno.go` from GitHub at build time to auto-generate the CLI flags table (`docs/_config-options.md`), so changes to that file also need to trigger a redeploy, otherwise the "Configuring Juno" page goes stale.

Proposed fix: add a `paths` filter covering both `docs/**` and `cmd/juno/juno.go`:

\`\`\`yaml
on:
  push:
    branches:
      - main
    paths:
      - 'docs/**'
      - 'cmd/juno/juno.go'
  workflow_dispatch:
\`\`\`

This avoids unnecessary docs builds/deploys on unrelated pushes while still keeping the generated config table in sync.

Ehsan-saradar:
## Summary
`docs-deploy.yml` rebuilds and redeploys the docs site on every push to `main`, even when the change has nothing to do with `docs/`. This scopes the trigger to only run when relevant files change.

`paths` can't simply be `docs/**` though — `docs/generate-config.js` fetches `cmd/juno/juno.go` from GitHub at build time to auto-generate the CLI flags table (`docs/_config-options.md`), so changes to that file also need to trigger a redeploy or the "Configuring Juno" page goes stale. The filter covers both.

Fixes #3897

## Test plan
- `actionlint` passes on the modified workflow with no errors/warnings
- Simulated GitHub's `paths` matching against real file lists from this repo's history:
  - docs-only PR (#3894 file list) → triggers
  - commit touching only `cmd/juno/juno.go` (`caefd43d9`) → triggers
  - Go-only commits with no docs/juno.go changes → does not trigger
  - mixed change (unrelated Go file + one docs file) → triggers


brbrr (changes requested):
Thanks for the contribution. To be able to merge it, we should also include the workflow itself

Ehsan-saradar:
Good catch @brbrr, thanks — done in a separate commit. `.github/workflows/docs-deploy.yml` is in the `paths` list now, so touching the deploy workflow actually redeploys and we get to see it work.

brbrr (approved):
Thanks!

PR #3904, opened 2026-08-04
perf(rpc): persist state diff length with block commitments

NazariiDenha:
`getBlockWithTxs`, `getBlockWithTxHashes` and `getBlockWithReceipts` each read and decoded the **entire `StateUpdate`** for a block in order to report one number - `state_diff_length`.
`getCommitmentsAndStateDiff` did two reads, and the only thing consumed from the second was `stateDiff.Length()`. 

### Change

Add `StateDiffLength *uint64` to `core.BlockCommitments`, which the same code path already reads

`writeBlockContent` derives it - the single point every writer passes through (sync, builder/consensus and p2p all reach it via `blockchain.Store`; it has 4 callers, all inside `statebackend`). A mandatory migration backfills existing blocks

## Results

Real Sepolia database (90 GB, 12.93M blocks), pebble cache at the node default of 1 GB

| method | before | after | delta |
|---|---|---|---|
| `getBlockWithTxHashes` | 51.0 µs | 41.6 µs | **-18.5%** |
| `getBlockWithTxs` | 51.5 µs | 42.5 µs | **-17.5%** |
| `getBlockWithReceipts` | 81.5 µs | 70.8 µs | **-13.1%** |

Including `json.Marshal` of the response:

| method | before | after | delta |
|---|---|---|---|
| `getBlockWithTxHashes` | 54.4 µs | 44.5 µs | **-18.2%** |
| `getBlockWithTxs` | 71.4 µs | 62.0 µs | **-13.1%** |
| `getBlockWithReceipts` | 113.0 µs | 104.2 µs | **-7.8%** |

All six drop 64 allocations per request. The absolute saving is ~9–10 µs regardless of method; the percentage differs only because the methods have different total cost.

CPU profiles agree independently: `StateUpdateByNumber` accounted for 13.25% / 18.10% / 8.53% of on-CPU time before, which per request works out to 9.2 / 9.3 / 9.2 µs — within 4% of the wall-clock delta. It is absent from the profiles afterwards.

## Migration

Registered last in the registry, after the history pruner, so pruned blocks are not backfilled. Verified on a real node: it ran automatically at startup as migration 4/4 and backfilled **12,928,078 blocks in 58 seconds**. Resumable via intermediate state, idempotent, and it skips blocks whose commitments were pruned.

## Why the field is a pointer

Zero is a legal length (a block with no state changes), so a plain `uint64` cannot distinguish "length is 0" from "never written" — a write path that forgot to set it would silently serve `state_diff_length: 0`. With a pointer the reader fails loudly instead. It costs nothing: `BlockHeader.StateDiffLength` in the RPC response is already `*uint64`, so a value type would have to be boxed anyway.

Verified on an un-migrated database that the guard fires as intended:

```
block 726743 has commitments without a state diff length
```


migration/statedifflength/migrator.go
```diff
@@ -0,0 +1,159 @@
...
+		return false, err
+	}
+	if commitments.StateDiffLength != nil {
+		return false, nil
+	}
+
+	stateUpdate, err := core.GetStateUpdateByBlockNum(r, blockNumber)
+	if err != nil {
+		return false, err
+	}
```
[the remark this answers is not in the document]
NazariiDenha: existing `commitments` and missing `stateupdate` already means that we have corrupted db after failed pruning. Skipping it will result in persistent RPC error.

migration/statedifflength/migrator.go
```diff
@@ -0,0 +1,159 @@
...
+	logger.Info("Backfilling state diff length",
+		zap.Uint64("fromBlock", m.nextBlock),
+		zap.Uint64("toBlock", height),
+	)
+
+	batch := database.NewBatchWithSize(batchByteSize)
+	lastLog := time.Now()
+	var updated uint64
+
+	for blockNumber := m.nextBlock; blockNumber <= height; blockNumber++ {
```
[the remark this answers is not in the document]
NazariiDenha: done

blockchain/statebackend/block_ops.go
```diff
@@ -140,7 +140,15 @@ func writeBlockContent(
 		return err
 	}
 
-	if err := core.WriteBlockCommitment(writer, block.Number, commitments); err != nil {
+	commitmentsWithLength := *commitments
+	if commitmentsWithLength.StateDiffLength == nil {
+		stateDiffLength := stateUpdate.StateDiff.Length()
+		commitmentsWithLength.StateDiffLength = &stateDiffLength
+	}
```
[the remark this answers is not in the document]
NazariiDenha: done for writeBlockContent

migration/statedifflength/migrator_test.go
```diff
@@ -0,0 +1,128 @@
...
+func run(t *testing.T, database db.KeyValueStore, m *statedifflength.Migrator) []byte {
+	t.Helper()
+	state, err := m.Migrate(t.Context(), database, &networks.Mainnet, log.NewNopZapLogger())
+	require.NoError(t, err)
+	return state
+}
+
+func TestMigrateBackfillsMissingLengths(t *testing.T) {
+	database := memory.New()
+	for blockNum := range uint64(3) {
```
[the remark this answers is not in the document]
NazariiDenha: added tests

migration/statedifflength/migrator.go
```diff
@@ -0,0 +1,159 @@
...
+
+const (
+	batchByteSize = 16 * db.Megabyte
+	timeLogRate   = 5 * time.Second
+)
+
+var (
+	shouldRerun    = []byte{}
+	shouldNotRerun = []byte(nil)
+)
```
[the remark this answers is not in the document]
NazariiDenha: done

rpc/v10/helpers.go
```diff
@@ -224,21 +224,16 @@ func transactionTypeFrom(t core.Transaction) TransactionType {
...
-		return nil, nil, err
+	if commitments.StateDiffLength == nil {
+		return nil, fmt.Errorf(
+			"block %d has commitments without a state diff length", blockNumber,
+		)
 	}
-
-	return commitments, stateUpdate.StateDiff, nil
+	return commitments, nil
 }
```
EgeCaner: I guess we can remove this helper as a whole and just call 
```go 
commitments, err := h.bcReader.BlockCommitmentsByNumber(blockNumber)
```
And after migration we can expect all blocks to have valid state diff length

blockchain/statebackend/block_ops.go
```diff
@@ -140,7 +140,13 @@ func writeBlockContent(
 		return err
 	}
 
-	if err := core.WriteBlockCommitment(writer, block.Number, commitments); err != nil {
+	commitmentsWithLength := *commitments
+	stateDiffLength := stateUpdate.StateDiff.Length()
+	commitmentsWithLength.StateDiffLength = &stateDiffLength
+
```
EgeCaner: IMO this part should reside in where block commitments were created and length should assigned at construction time.

blockchain/blockchain_test.go
```diff
@@ -303,6 +303,9 @@ func TestBlockCommitments(t *testing.T) {
 
 	commitments, err := chain.BlockCommitmentsByNumber(0)
 	require.NoError(t, err)
+
+	stateDiffLength := su.StateDiff.Length()
+	expectedCommitments.StateDiffLength = &stateDiffLength
```
EgeCaner: this can be assigned above where `expectedCommitments` created

core/block.go
```diff
@@ -105,6 +105,7 @@ type BlockCommitments struct {
 	EventCommitment       *felt.Felt
 	ReceiptCommitment     *felt.Felt
 	StateDiffCommitment   *felt.Felt
+	StateDiffLength       *uint64
```
EgeCaner: Can be value field. I get the rationale behind making it ptr field for existence check. But migration will run before everything else thus we can assume it is correctly set for all blocks, and not require `is_set` check

blockchain/statebackend/block_ops.go
EgeCaner: Seems like commitments are created in variations of `core.BlockHash(..)`

migration/statedifflength/migrator.go
```diff
@@ -0,0 +1,180 @@
...
+	}
+	return max(m.nextBlock, oldest), nil
+}
+
+// backfillBlock reports whether the block's commitments were rewritten.
+func backfillBlock(r db.KeyValueReader, w db.KeyValueWriter, blockNumber uint64) (bool, error) {
+	commitments, err := core.GetBlockCommitmentByBlockNum(r, blockNumber)
+	if err != nil {
+		if errors.Is(err, db.ErrKeyNotFound) {
+			return false, nil
```
EgeCaner: this should be an error missing commitments is sign of DB corruption, should fail loudly rather skipping silently

migration/statedifflength/migrator.go
```diff
@@ -0,0 +1,180 @@
+package statedifflength
```
EgeCaner, line 3: This migration can benefit from `migrations/pipeline.go`. See other migrations in `migration` package, and see how `pipeline.go` has been utilized to divide the work in to multiple workers.

rpc/v10/block_test.go
```diff
@@ -1448,3 +1441,57 @@ func TestBlockWithReceiptsWithResponseFlags(t *testing.T) {
...
+			handler := rpc.New(mockReader, nil, nil, log.NewNopZapLogger())
+			blockID := rpc.BlockIDFromNumber(block.Number)
+
+			rpcErr := tc.call(handler, &blockID)
+			require.NotNil(t, rpcErr, "expected an error, got a response")
+			assert.Equal(t, rpccore.ErrInternal.Code, rpcErr.Code)
+			assert.Contains(t, fmt.Sprintf("%v", rpcErr.Data), "state diff length")
+		})
+	}
+}
```
EgeCaner: If we are going to discard the is state diff length set check, we can remove this test as well

EgeCaner (changes requested):

blockchain/statebackend/block_ops.go
NazariiDenha: And we can't be sure that we have `StateDiff` on hand when we create commitments

migration/statedifflength/migrator.go
```diff
@@ -0,0 +1,153 @@
...
+		zap.Int("workers", concurrency),
+	)
+
+	batchSemaphore := semaphore.New(concurrency+1, func() db.Batch {
+		return database.NewBatchWithSize(batchByteSize)
+	})
+
+	source := pipeline.Source(blockRange(start, height))
+	readers := pipeline.New(
+		source, concurrency, newIngestor(database, batchSemaphore, concurrency, batchByteSize),
```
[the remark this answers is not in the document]
NazariiDenha: done

migration/statedifflength/migrator.go
```diff
@@ -0,0 +1,153 @@
...
+func backfillBlock(r db.KeyValueReader, batch db.KeyValueWriter, blockNumber uint64) error {
+	commitments, err := core.GetBlockCommitmentByBlockNum(r, blockNumber)
+	if err != nil {
+		return err
+	}
+
+	stateUpdate, err := core.GetStateUpdateByBlockNum(r, blockNumber)
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
NazariiDenha: done

migration/statedifflength/migrator.go
NazariiDenha: added pipeline

migration/statedifflength/migrator.go
```diff
@@ -0,0 +1,155 @@
...
+	"github.com/NethermindEth/juno/migration/semaphore"
+	"github.com/NethermindEth/juno/pruner"
+	"github.com/NethermindEth/juno/utils/log"
+	"go.uber.org/zap"
+)
+
+const (
+	maxConcurrency       = 8
+	defaultBatchByteSize = 24 * db.Megabyte
+	targetBatchByteSize  = 16 * db.Megabyte
```
EgeCaner: You can use bigger batch size similar migrations uses 

```go
	// batchByteSize is the initially allocated size of a batch.
	batchByteSize = 128 * db.Megabyte

	// targetBatchByteSize is the threshold at which a batch is written to the database.
	targetBatchByteSize = 96 * db.Megabyte
```

migration/statedifflength/migrator.go
```diff
@@ -0,0 +1,155 @@
...
+			}
+		}
+	}
+}
+
+// backfillBlock derives the state diff length from the block's state update and
+// writes the block's commitments with it into batch. A missing commitments or
+// state update record above the pruned prefix means the two buckets disagree, i.e.
+// the database is corrupt, so the read error is returned rather than skipped.
+func backfillBlock(r db.KeyValueReader, batch db.KeyValueWriter, blockNumber uint64) error {
```
EgeCaner: nit: move to `ingestor.go`

migration/statedifflength/ingestor.go
```diff
@@ -0,0 +1,62 @@
...
+	flushAt   int
+	batches   []task
+}
+
+var _ pipeline.State[uint64, task] = (*ingestor)(nil)
+
+func newIngestor(
+	database db.KeyValueReader,
+	batchSemaphore semaphore.ResourceSemaphore[db.Batch],
+	workers, flushAt int,
```
EgeCaner: nit: one line per param

migration/statedifflength/migrator.go
```diff
@@ -0,0 +1,155 @@
...
+	)
+
+	batchSemaphore := semaphore.New(concurrency+1, func() db.Batch {
+		return database.NewBatchWithSize(batchByteSize)
+	})
+
+	source := pipeline.Source(blockRange(start, height))
+	readers := pipeline.New(
+		source, concurrency, newIngestor(database, batchSemaphore,
+			concurrency, min(batchByteSize, targetBatchByteSize)),
```
EgeCaner: nit: one line per param

migration/statedifflength/migrator.go
```diff
@@ -0,0 +1,155 @@
...
+	)
+	committed := pipeline.New(
+		readers, 1, newCommitter(logger, batchSemaphore, height),
```
EgeCaner: nit: one line per param

migration/statedifflength/migrator.go
```diff
@@ -0,0 +1,155 @@
...
+
+// Migrator reads each block's state update, computes StateDiff.Length(), and
+// rewrites the block's commitments with it. Pre-existing blocks stored 0 for the
+// missing field, which this replaces with the real value.
+//
+// Work is fanned out across Concurrency reader goroutines that each fill their own
+// write batch; a single committer drains the batches to disk. There is no resume
+// cursor: an interrupted run restarts from the oldest retained block on the next
+// boot, which is safe because re-deriving a length yields the same value.
+type Migrator struct {
```
EgeCaner, line 47: Make it resumable on graceful shutdown, utilizing intermediate state to not repeat previous work on graceful shutdown. Crash should safely recover due to this migration not causing error on processing same data second time at run after crash.

blockchain/statebackend/block_ops.go
EgeCaner: I guess we can pass the state diff into `pre07Hash` and `post07Hash`, other ones seem already receiving the diff

EgeCaner (changes requested):

migration/statedifflength/migrator.go
NazariiDenha: Committer writes blocks out of order, so we need to add some logic of tracking contiguous range of written blocks. I think that extra logic doesn't worth 1 minute migration

NazariiDenha: added

blockchain/statebackend/block_ops.go
NazariiDenha: updated

core/block.go
```diff
@@ -213,6 +215,10 @@ func pre07Hash(
 	if err != nil {
 		return felt.Felt{}, nil, err
 	}
+	var sdLength uint64
+	if stateDiff != nil {
```
EgeCaner, line 219: nit: must not be called with nil state diff. Directly call state diff len without nil check, nil state diff here should be unexpected

core/block.go
```diff
@@ -384,19 +391,25 @@ func Post0132Hash(
...
+	stateDiff *StateDiff,
 	overrideSeqAddr *felt.Felt,
 	backend TempTrieBackend,
 ) (felt.Felt, *BlockCommitments, error) {
 	seqAddr := b.SequencerAddress
 	if overrideSeqAddr != nil {
 		seqAddr = overrideSeqAddr
 	}
+	var sdLength uint64
+	if stateDiff != nil {
```
EgeCaner, line 410: same here

migration/statedifflength/migrator.go
```diff
@@ -0,0 +1,166 @@
...
+	committed := pipeline.New(readers, 1, comm)
+
+	_, wait := committed.Run(ctx)
+	res := wait()
+
+	if res.Err != nil {
+		return nil, res.Err
+	}
+	if !res.IsDone {
+		resume := start
```
EgeCaner: I guess we can just use same resumability pattern [here](https://github.com/NethermindEth/juno/blob/b34346d0df8cf97b45a7939e629191e644587f91/migration/historyprunner/migrator.go#L358) rather than deriving from committer

migration/statedifflength/migrator.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+	if len(state) != intermediateStateSize {
+		return fmt.Errorf("statedifflength: invalid intermediate state size: got %d, want %d",
+			len(state), intermediateStateSize)
+	}
+	m.nextBlock = binary.BigEndian.Uint64(state)
+	return nil
+}
+
+func (m *Migrator) Migrate(
```
EgeCaner, line 66: Lets use block progress tracker to display progress [see](https://github.com/NethermindEth/juno/blob/b34346d0df8cf97b45a7939e629191e644587f91/migration/progresslogger/block_progress_tracker.go#L12) and utilization [example](https://github.com/NethermindEth/juno/blob/b34346d0df8cf97b45a7939e629191e644587f91/migration/historyprunner/migrator.go#L262)

migration/statedifflength/migrator_test.go
```diff
@@ -0,0 +1,216 @@
...
+
+// writeBlock stores a state update with storageDiffs entries plus a commitments
+// record whose StateDiffLength is left unset, mimicking a pre-migration block.
+func writeBlock(t *testing.T, database db.KeyValueStore, blockNum, storageDiffs uint64) {
+	t.Helper()
+
+	diffs := make(map[felt.Felt]map[felt.Felt]*felt.Felt)
+	inner := make(map[felt.Felt]*felt.Felt, storageDiffs)
+	for i := range storageDiffs {
+		inner[*new(felt.Felt).SetUint64(i)] = new(felt.Felt).SetUint64(i)
```
EgeCaner: use felt Ctors instead [see](https://github.com/NethermindEth/juno/blob/main/core/felt/ctors.go)

```suggestion
		inner[felt.FromUint64(i)] = felt.NewFromUint64(i)
```

migration/statedifflength/migrator_test.go
```diff
@@ -0,0 +1,216 @@
...
+	}
+	diffs[*new(felt.Felt).SetUint64(blockNum)] = inner
```
EgeCaner: same here

migration/statedifflength/migrator_test.go
```diff
@@ -0,0 +1,216 @@
...
+
+	require.NoError(t, core.WriteStateUpdateByBlockNum(database, blockNum, &core.StateUpdate{
+		StateDiff: &core.StateDiff{StorageDiffs: diffs},
+	}))
+	require.NoError(t, core.WriteBlockCommitment(database, blockNum, &core.BlockCommitments{
+		TransactionCommitment: new(felt.Felt).SetUint64(blockNum),
```
EgeCaner: same here

migration/statedifflength/migrator_test.go
```diff
@@ -0,0 +1,216 @@
...
+		require.Equal(t, blockNum+1, storedLength(t, database, blockNum), "block %d", blockNum)
+	}
+}
+
+func TestBeforeRejectsMalformedState(t *testing.T) {
+	m := &statedifflength.Migrator{}
+	require.Error(t, m.Before([]byte{1, 2, 3}))
+}
+
+func TestMigrateRotatesBatches(t *testing.T) {
```
EgeCaner: Maybe we can remove this test, and not make batch size and concurrency configurable. Seems like they are configurable only for tests, and Concurrency should not affect the result already, so I guess we can remove that and then only reason to keep batch size configurable is to force rotation here I guess we can just remove it and use constant directly and remove this test

migration/statedifflength/migrator_test.go
```diff
@@ -0,0 +1,216 @@
...
+) ([]byte, error) {
+	return m.Migrate(ctx, database, &networks.Mainnet, log.NewNopZapLogger())
+}
+
+func run(t *testing.T, database db.KeyValueStore, m *statedifflength.Migrator) []byte {
+	t.Helper()
+	state, err := migrate(t.Context(), database, m)
+	require.NoError(t, err)
+	return state
+}
```
EgeCaner: I guess we can remove this two helper, it is already short to call directly

```go
state, err := m.Migrate(ctx, database, &networks.Mainnet, log.NewNopZapLogger())
require.NoError(t,err)
```

migration/statedifflength/ingestor.go
```diff
@@ -0,0 +1,86 @@
...
+	"fmt"
+
+	"github.com/NethermindEth/juno/core"
+	"github.com/NethermindEth/juno/db"
+	"github.com/NethermindEth/juno/migration/pipeline"
+	"github.com/NethermindEth/juno/migration/semaphore"
+)
+
+// task is a filled batch handed from a reader to the committer
+type task struct {
```
EgeCaner: with using block progress tracker we can drop this type and use batch directly

migration/statedifflength/ingestor.go
```diff
@@ -0,0 +1,86 @@
...
+	blocks   int
+	maxBlock uint64
+}
+
+// ingestor reads blocks and fills per-worker batches. Each pipeline worker owns one
+// entry in batches, so no synchronisation is needed between them.
+type ingestor struct {
+	database  db.KeyValueReader
+	semaphore semaphore.ResourceSemaphore[db.Batch]
+	flushAt   int
```
EgeCaner: if this is not configurable after removing configurability from my other comments, lets remove this as well and direclty use constant below

migration/statedifflength/migrator.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+	return nil, nil
+}
+
+// startBlock is the higher of the resume checkpoint and the oldest retained block,
+// so a re-run skips finished work without touching the pruned prefix.
+func (m *Migrator) startBlock(r db.KeyValueReader) (uint64, error) {
+	oldest, err := pruner.OldestRetainedBlock(r)
+	if err != nil {
+		if errors.Is(err, db.ErrKeyNotFound) {
```
EgeCaner: We should be erroring here loudly, chain height shortcircuit should handle 0 block case, and in non zero block case we must have something here

migration/statedifflength/ingestor.go
```diff
@@ -0,0 +1,86 @@
...
+	outputs <- in.batches[index]
+	return nil
+}
+
+// backfillBlock derives the state diff length from the block's state update and
+// writes the block's commitments with it into batch. A missing commitments or
+// state update record above the pruned prefix means the two buckets disagree, i.e.
+// the database is corrupt, so the read error is returned rather than skipped.
+func backfillBlock(r db.KeyValueReader, batch db.KeyValueWriter, blockNumber uint64) error {
+	commitments, err := core.GetBlockCommitmentByBlockNum(r, blockNumber)
```
EgeCaner, line 67: One thing that I was doing in my custom CBOR walker experiment to optimize lenght extracting without fully decode is. Per ingestor having one commitment object they reuse to decode into `core.GetBlockCommitmentByBlockNum(r, blockNumber)` should be allocating one for each call it should reduce allocations of commitments from number of blocks to number of workers. 

```go
	err := r.Get(db.BlockCommitmentsKey(blockNumber), func(data []byte) error {
		return encoder.Unmarshal(data, &reusedObject)
	})
```

EgeCaner (changes requested):

migration/statedifflength/ingestor.go
EgeCaner: we should be able to reuse the object after batch.Write. Might have to zeroed before read in order to not carry missing fields from prev iter I guess

core/block.go
NazariiDenha: Some old migrations and also tests pass nil state diff here for blocks pre 0.13.2 (pre07Hash, post07Hash)

Actually, for blocks pre 0.13.2 we should not return this field at all, but this was before this pr and if needed can be fixed in future prs.

migration/statedifflength/ingestor.go
NazariiDenha: maybe can try and change this in next pr you told about improving speed of migration

migration/statedifflength/migrator.go
```diff
@@ -0,0 +1,157 @@
...
+const (
+	maxConcurrency      = 8
+	batchByteSize       = 128 * db.Megabyte
+	targetBatchByteSize = 96 * db.Megabyte
+	timeLogRate         = 5 * time.Second
+	// intermediateStateSize is the encoded resume checkpoint: one big-endian uint64.
+	intermediateStateSize = 8
+)
+
+var _ migration.Migration = (*Migrator)(nil)
```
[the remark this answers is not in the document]
NazariiDenha: skip this

migration/statedifflength/committer.go
```diff
@@ -0,0 +1,31 @@
...
+	return &committer{semaphore: batchSemaphore}
+}
+
+func (c *committer) Run(_ int, batch db.Batch, _ chan<- struct{}) error {
+	if err := batch.Write(); err != nil {
+		return err
+	}
+	c.semaphore.Put()
+	return nil
+}
```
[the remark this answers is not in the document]
NazariiDenha: skip

migration/statedifflength/migrator.go
```diff
@@ -0,0 +1,157 @@
...
+	"github.com/NethermindEth/juno/pruner"
+	"github.com/NethermindEth/juno/utils/log"
+	"go.uber.org/zap"
+)
+
+const (
+	maxConcurrency      = 8
+	batchByteSize       = 128 * db.Megabyte
+	targetBatchByteSize = 96 * db.Megabyte
+	timeLogRate         = 5 * time.Second
```
EgeCaner: Lets raise this to 30 sec, otherwise might be too noisy

migration/statedifflength/migrator_test.go
```diff
@@ -0,0 +1,197 @@
...
+	state, err := m.Migrate(t.Context(), database, &networks.Mainnet, log.NewNopZapLogger())
+	require.NoError(t, err)
+	require.Nil(t, state, "migration should not ask to re-run")
+
+	for blockNum := range uint64(3) {
+		require.Equal(t, blockNum+1, storedLength(t, database, blockNum), "block %d", blockNum)
+	}
+}
+
+func TestMigrateConcurrent(t *testing.T) {
```
EgeCaner: Seems like only difference between this and `TestMigrateBackfillsMissingLengths` seems to be number of blocks they use. We can rename and use this one since by default it is concurrent

EgeCaner (approved):
Dropped minor comments, rest looks good to me!

migration/statedifflength/migrator_test.go
NazariiDenha: done

migration/statedifflength/migrator.go
NazariiDenha: done

PR #3909, opened 2026-08-05
perf(blockchain): decode only the events subset of receipts in getEvents

brbrr:
Bench results for various test cases using sepolia snapshot:

| Scenario | Query shape | Latency | Allocs/op | Bytes/op |
|---|---|---|---|---|
| addr+key / 8k blocks | busiest contract + one event type (indexer paging Transfers) | 13.34 ms → 11.34 ms (**−15.0%**) | 23,720 → 17,900 (**−24.5%**) | 1,245 KiB → 972 KiB (**−21.9%**) |
| addr-only / 8k blocks | all events of the busiest contract | 13.33 ms → 11.40 ms (**−14.5%**) | 23,710 → 17,890 (**−24.5%**) | 1,243 KiB → 971 KiB (**−21.9%**) |
| match-all / 64 blocks | no filter, "all recent events" (explorer feed) | 1.45 ms → 1.10 ms (**−24.3%**) | 5,204 → 3,933 (**−24.4%**) | 264 KiB → 205 KiB (**−22.5%**) |
| sparse-addr / 8k blocks | rarely-active contract, 35 results | 2.11 ms → 1.82 ms (**−13.8%**) | 5,659 → 4,711 (**−16.8%**) | 281 KiB → 236 KiB (**−15.9%**) |
| rare-key / 8k blocks | key matching nothing, 0 results | 18.3 µs → 17.7 µs (−3.3%, p=0.009) | 39 → 39 (±0%) | 7.1 KiB → 7.1 KiB (±0%) |

### DIFF
--- a/blockchain/blockchain_test.go
+++ b/blockchain/blockchain_test.go
@@ -1,6 +1,7 @@
 package blockchain_test
 
 import (
+	"bytes"
 	"fmt"
 	"testing"
 
@@ -936,10 +937,10 @@ func TestEventsMultiPreConfirmed(t *testing.T) {
 		gotPerBlock := map[uint64]int{}
 		for i, e := range events {
 			require.NotNil(t, e.BlockNumber)
-			gotPerBlock[*e.BlockNumber]++
+			gotPerBlock[e.BlockNumber]++
 			// Events must be yielded oldest-first across canonical + pre_confirmed.
 			if i > 0 {
-				require.LessOrEqual(t, *events[i-1].BlockNumber, *e.BlockNumber,
+				require.LessOrEqual(t, events[i-1].BlockNumber, e.BlockNumber,
 					"events must be returned oldest-first by block number")
 			}
 		}
@@ -1013,7 +1014,7 @@ func TestEventsMultiPreConfirmed(t *testing.T) {
 		require.NotEmpty(t, events)
 		for _, e := range events {
 			require.NotNil(t, e.BlockNumber)
-			require.Equal(t, lastPreConfirmedBlockNum, *e.BlockNumber,
+			require.Equal(t, lastPreConfirmedBlockNum, e.BlockNumber,
 				"pre_confirmed tag should return only the tip block")
 		}
 	})
@@ -1041,7 +1042,7 @@ func TestEventsMultiPreConfirmed(t *testing.T) {
 			require.NotNil(t, e.From)
 			require.Equal(t, from[0], felt.Address(*e.From))
 			require.NotNil(t, e.BlockNumber)
-			require.Equal(t, lastPreConfirmedBlockNum, *e.BlockNumber)
+			require.Equal(t, lastPreConfirmedBlockNum, e.BlockNumber)
 		}
 	})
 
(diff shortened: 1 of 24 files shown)


blockchain/event_filter.go
```diff
@@ -146,23 +146,32 @@ func (e *EventFilter) Events(
...
+	// instead. Pinning a fixed head wouldn't help: once a block commits,
+	// SnapshotForBlock trims it, so a stale head yields an empty view. On error,
+	// drop it and serve canonical only.
+	var preConfirmed PreConfirmedReader
+	if e.toBlock > latest {
+		if fetched, err := e.preConfirmedFn(); err == nil {
+			preConfirmed = fetched
+		}
+
+		latest, err = core.GetChainHeight(e.database)
```
EgeCaner: Maybe we can re-derive the head from pre-confirmed bottom and not penalize preconfirmed path. `preConfirmedFn` uses `sync.PreconfirmedChain` which internally calls `ChainHeight`, this makes chain height calls on preconfirmed path 3 including this one. If we derive from preconfirmed chain, using `preconfirmedChain.Head().Number` and `preconfirmedChain.Length`, I guess we can keep preconfirmed path unaffected.

EgeCaner: This is for another PR 

We can expose `PreconfirmedChainForBlock(block_num)` etc... which avoids the second chain height read on preconfirmed path. 

Additionally `sync.PreconfirmedChain` creates empty pre-confirmed for parent block when there is no preconfirmed which requires DB reads to populate block hash contract which I guess should be only useful for simulation, estimation. Empty placeholder has no use here since it wont gonna contribute any events, thus such function would eliminate that as well.

blockchain/event_filter.go
```diff
@@ -306,6 +311,20 @@ func (e *EventFilter) canonicalEvents(
 	return matchedEvents, ContinuationToken{}, nil
 }
 
+// transactionEventsFromReceipts projects in-memory receipts down to the events
+// subset the matcher consumes, so both event sources feed it one shape. The call
+// inlines and the slice does not escape, making the projection stack-only.
+func transactionEventsFromReceipts(receipts []*core.TransactionReceipt) []core.TransactionEvents {
```
EgeCaner: This still should heap allocates the `[]core.TransactionEvents`, penalizes preconfirmed path. Not sure what is the best way to avoid it. Maybe we can pass down a function that returns events and their transaction hash, then preconfirmed can wire itself directly without intermediate object

EgeCaner (commented):
Dropped some comments, rest looks good to me!

blockchain/event_filter.go
brbrr: i've end up keeping the `[]core.TransactionEvents` in `EventFilter`. Please take a look when you have some time.

blockchain/event_matcher.go
```diff
@@ -183,10 +199,89 @@ func (e *EventMatcher) AppendBlockEvents(
...
+	return matchedEventsSofar, processedEvents, nil
+}
+
+// AppendBlockEventsFromReceipts appends the events of a pre-confirmed block that match
+// the filter to matchedEventsSofar. A pre-confirmed block holds full receipts in memory.
+// This function reads the events from the receipts, so the path needs no projection.
+// See AppendBlockEventsFromTransactionEvents for the reason the loops stay separate.
+//
+//nolint:dupl // The other source keeps its own loop on purpose. See above.
+func (e *EventMatcher) AppendBlockEventsFromReceipts(
```
EgeCaner, line 228: If we are explicitly gonna keep this for preconfirmed we can remove the block hash since preconfirmeds don't have hash. We can keep it is as and it will be generic for Receipt source, then I guess we might not need to imply it is for preconfirmed. It is FromReceipts, can be canonical too, lets keep doc generic

blockchain/event_matcher.go
```diff
@@ -183,10 +199,89 @@ func (e *EventMatcher) AppendBlockEvents(
...
+					var err error
+					if blockHash, err = blockHashFn(); err != nil {
+						return nil, 0, err
+					}
+					num := blockNum
+					blockNumPtr = &num
+				}
 				matchedEventsSofar = append(matchedEventsSofar, FilteredEvent{
-					BlockNumber:      &blockNum,
+					BlockNumber:      blockNumPtr,
```
EgeCaner: I guess this field doesn't have to be pointer anymore, I guess it might be there because of pending block, but now all blocks we have their number, preconfirmed and canonicals

blockchain/blockchain_test.go
```diff
@@ -1043,6 +1073,107 @@ func TestEventsMultiPreConfirmed(t *testing.T) {
 	})
 }
 
+// TestEventsPreConfirmedProjectionReuse tests the pre_confirmed receipt projection,
+// which uses one buffer again for each chain entry. The first entry has more
+// transactions than the second entry. If the buffer keeps the data of the first
+// entry, the second entry shows the transactions of the first entry. The test sets
+// the number of transactions directly, because the fixtures do not control it.
+func TestEventsPreConfirmedProjectionReuse(t *testing.T) {
```
EgeCaner: Is this relevant anymore

EgeCaner (approved):
Looks good to me

PR #3912, opened 2026-08-05
feat(node): introduce disabling synchronization

danielntmd:
Allow disabling L2 sync so a node can serve RPC from an existing DB snapshot. 

Gates synchronizer/P2P creation behind `!DisableSync,` defaults `syncReader` to `NoopSynchronizer` (also fixing a latent /ready nil-deref in sequencer mode), and adds a sync-independent `/ready/rpc` endpoint that reports 200 when the local DB head is readable. `NoopSynchronizer.PreConfirmedChain()` now returns the `ErrPreConfirmedNotFound` sentinel so RPC callers map it to `ErrBlockNotFound`.

### DIFF
--- a/cmd/juno/juno.go
+++ b/cmd/juno/juno.go
@@ -59,6 +59,7 @@ const (
 	pprofPortF                          = "pprof-port"
 	colourF                             = "colour"
 	preConfirmedPollIntervalF           = "preconfirmed-poll-interval"
+	disableSyncF                        = "disable-sync"
 	p2pF                                = "p2p"
 	p2pAddrF                            = "p2p-addr"
 	p2pPublicAddrF                      = "p2p-public-addr"
@@ -131,6 +132,7 @@ const (
 	defaultPprofPort                          = 6062
 	defaultColour                             = true
 	defaultPreConfirmedPollInterval           = 500 * time.Millisecond
+	defaultDisableSync                        = false
 	defaultP2p                                = false
 	defaultP2pAddr                            = ""
 	defaultP2pPublicAddr                      = ""
@@ -211,6 +213,7 @@ const (
 	disableL1VerificationUsage    = "Disables L1 verification since an Ethereum node is not provided."
 	preConfirmedPollIntervalUsage = "Sets how frequently pre_confirmed block will be updated" +
 		"(0s will disable fetching of pre_confirmed block)."
+	disableSyncUsage   = "Disables L2 synchronization."
 	p2pUsage           = "EXPERIMENTAL: Enables p2p server."
 	p2pAddrUsage       = "EXPERIMENTAL: Specify p2p listening source address as multiaddr.  Example: /ip4/0.0.0.0/tcp/7777"
 	p2pPublicAddrUsage = "EXPERIMENTAL: Specify p2p public address as multiaddr.  Example: /ip4/35.243.XXX.XXX/tcp/7777"
@@ -540,6 +543,7 @@ func NewCmd(config *node.Config, run func(*cobra.Command, []string) error) *cobr
 	setCategory(junoCmd, catNetwork, networkF, ethNodeF, disableL1VerificationF)
 
 	// --- Sync & Polling ---
+	junoCmd.Flags().Bool(disableSyncF, defaultDisableSync, disableSyncUsage)
 	junoCmd.Flags().Duration(
 		preConfirmedPollIntervalF, defaultPreConfirmedPollInterval, preConfirmedPollIntervalUsage,
 	)
@@ -548,7 +552,7 @@ func NewCmd(config *node.Config, run func(*cobra.Command, []string) error) *cobr
 		readinessBlockToleranceF, defaultReadinessBlockTolerance, readinessBlockToleranceUsage,
 	)
 	setCategory(junoCmd, catSyncPolling,
-		preConfirmedPollIntervalF, remoteDBF, readinessBlockToleranceF,
+		disableSyncF, preConfirmedPollIntervalF, remoteDBF, readinessBlockToleranceF,
 	)
 
 	// --- Gateway ---
--- a/cmd/juno/juno.go
+++ b/cmd/juno/juno.go
@@ -59,6 +59,7 @@ const (
 	pprofPortF                          = "pprof-port"
 	colourF                             = "colour"
 	preConfirmedPollIntervalF           = "preconfirmed-poll-interval"
+	syncF                               = "sync"
 	p2pF                                = "p2p"
 	p2pAddrF                            = "p2p-addr"
 	p2pPublicAddrF                      = "p2p-public-addr"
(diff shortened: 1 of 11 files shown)


node/http.go
```diff
@@ -311,3 +311,13 @@ func (h *readinessHandlers) isSynced() bool {
...
+
+func (h *readinessHandlers) HandleReadyRPC(w http.ResponseWriter, r *http.Request) {
+	if _, err := h.bcReader.HeadsHeader(); err != nil {
+		w.WriteHeader(http.StatusServiceUnavailable)
+		w.Write([]byte("RPC not ready: database not serving.")) //nolint:errcheck
+		return
+	}
+
+	w.WriteHeader(http.StatusOK)
+}
```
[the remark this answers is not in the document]
danielntmd: hardfail /ready is intended, as node is not intended to be run with sync disabled.

node/node.go
```diff
@@ -186,6 +187,27 @@ func New(cfg *Config, version string, logLevel *log.Level) (*Node, error) {
...
+			return nil, errors.New("--p2p requires synchronization; remove --p2p or --sync=false")
+		}
+		if cfg.Prune {
+			return nil, errors.New("--prune-mode requires synchronization; " +
+				"remove --prune-mode or --sync=false")
+		}
+		if cfg.RemoteDB != "" {
+			return nil, errors.New("--remote-db cannot be combined with --sync=false; " +
+				"remove --remote-db or --sync=false")
+		}
```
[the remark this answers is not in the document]
danielntmd: Intentionally out of scope, operators will not run with sync disabled and the gating is enough justification.

node/node.go
```diff
@@ -295,6 +302,7 @@ func New(cfg *Config, version string, logLevel *log.Level) (*Node, error) {
 	var client *feeder.Client
 	var gatewayClient *gateway.Client
 	var p2pService *p2p.Service
+	var syncReader sync.Reader = &sync.NoopSynchronizer{}
```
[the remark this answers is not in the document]
danielntmd: intended.

docs/docs/_config-options.md
```diff
@@ -40,6 +40,7 @@
 | `preconfirmed-poll-interval` | `500ms` | Sets how frequently pre_confirmed block will be updated(0s will disable fetching of pre_confirmed block) |
 | `readiness-block-tolerance` | `6` | Maximum blocks behind latest for /ready endpoints to return 200 OK |
 | `remote-db` |  | gRPC URL of a remote Juno node |
+| `sync` | `true` | Enables L2 blockchain synchronization. L1 verification is controlled independently by --disable-l1-verification. Disabling synchronization is intended for tests or benchmarks using a preloaded database. A remote database cannot be used when synchronization is disabled. Use /ready/rpc for readiness |
```
brbrr: seems like too wordy. especially since it's for development purposes only, I'm not event sure it's worth adding a doc record about it

docs/docs/monitoring.md
```diff
@@ -127,6 +127,7 @@ The `/ready` endpoint returns:
 - `503 Service Unavailable` during migrations or when the node is not yet synced
 
 This endpoint is intended for use with Kubernetes `readinessProbe` to control traffic routing.
+When synchronization is disabled with `--sync=false`, use `/ready/rpc` instead.
```
brbrr: Worth noting that /ready is a default readiness probe on k8s. And i'm not sure it's worth changing expected dafaults

node/http.go
```diff
@@ -311,3 +311,18 @@ func (h *readinessHandlers) isSynced() bool {
 func (h *readinessHandlers) HandleLive(w http.ResponseWriter, r *http.Request) {
 	w.WriteHeader(http.StatusOK)
 }
+
+func (h *readinessHandlers) HandleReadyRPC(w http.ResponseWriter, r *http.Request) {
+	if _, err := h.bcReader.HeadsHeader(); err != nil {
+		w.WriteHeader(http.StatusServiceUnavailable)
+		if errors.Is(err, db.ErrKeyNotFound) {
+			emptyDatabaseMessage := []byte("RPC not ready: database contains no blocks.")
```
brbrr: The variable probably could be a const?

From my understanding, `ErrKeyNotFound` on `HeadsHeader` could mean different things: `ChainHeight` is missing, or the block header at that height is missing. I don't think that the error message captures the context properly.

node/http.go
```diff
@@ -311,3 +311,18 @@ func (h *readinessHandlers) isSynced() bool {
...
+			w.Write(emptyDatabaseMessage) //nolint:errcheck // best-effort body
+		} else {
+			w.Write([]byte("RPC not ready: database not serving.")) //nolint:errcheck // best-effort body
```
brbrr: Can we improve the message? Maybe "database is not available" or something along these lines?

brbrr (commented):
Looks good overall. left a few comments for you to consider.

docs/docs/monitoring.md
danielntmd: K8s probe whatever path we configure, for existing operators that do use /ready, this would be an extra safeguard.  And for our internal use case, we explicitly use the new /ready/rpc to probe readiness. 

Also, the previous usage of `/ready` returned `200` only when:
1. The local canonical head was readable.
2. The synchronizer had observed a highest remote block.
3. The local head was no more than readiness-block-tolerance blocks behind that remote block.

We would be preserving the previous semantic meaning of "the node is synchronized closely enough" instead of also adding the meaning: "process can serve RPC".

wdyt?

danielntmd: I'll just remove this line instead so that it is not confusing to readers

node/node.go
```diff
@@ -186,6 +187,27 @@ func New(cfg *Config, version string, logLevel *log.Level) (*Node, error) {
...
+				"remove --prune-mode or --sync=false")
+		}
+		if cfg.RemoteDB != "" {
+			return nil, errors.New("--remote-db cannot be combined with --sync=false; " +
+				"remove --remote-db or --sync=false")
+		}
+
+		logger.Warn("L2 synchronization and plugin block events are disabled. " +
+			"Use /ready/rpc for readiness (/ready and /ready/sync will report 503).")
+	}
```
infrmtcs, line 209: I understand that this is the current pattern, but can we try this instead:
https://pkg.go.dev/github.com/spf13/cobra#Command.MarkFlagsMutuallyExclusive

infrmtcs (approved):

cmd/juno/juno.go
```diff
@@ -404,6 +407,10 @@ func NewCmd(config *node.Config, run func(*cobra.Command, []string) error) *cobr
 			return err
 		}
 
+		// Keep the CLI option positive while storing its inverse so existing Config
+		// literals that omit this new field continue to enable synchronization.
+		config.DisableSync = !v.GetBool(syncF)
```
infrmtcs: I'm actually wandering if we should replace this by a `--disable-sync` instead, which simplifies all these things

node/node.go
danielntmd: This pattern does not check for boolean state, so we can have the case where we --disable-sync=false and --p2p creates a false positive. Reverted.

PR #3917, opened 2026-08-06
perf(rpc): read block transaction hashes without decoding transactions

EgeCaner:
`starknet_getBlockWithTxHashes` decoded every transaction in a block only to
  call Hash() on each one. Add a partial CBOR projection that materialises
  just the TransactionHash field and skips the rest in place, exposed as
  `BlockTransactions.TransactionHashes` and reached through a new
  `TransactionHashesByBlockNumber` reader. `TxnHashes` becomes []felt.Felt so the
  result is one contiguous slice.

  Pin the pre-confirmed block to avoid a race: `BlockWithTxHashes` and
  `BlockWithTxs` read the header and the transactions through two separate
  `PreConfirmedChain` calls, so the poller could advance in between and pair a
  header with another block's contents. Both now handle pre-confirmed in
  their own branch, taking a single snapshot and returning early.


core/block_transaction.go
```diff
@@ -100,3 +103,33 @@ type executionStatusProjectionSlice = indexed.LazySlice[receiptExecutionStatusPr
...
+	for i := range offsets {
+		end := sectionEnd
+		if i < len(offsets)-1 {
+			end = offsets[i+1]
+		}
+		// Reset before decoding: CBOR null decodes into a value-typed felt as a no-op, so a
+		// transaction stored without a hash would otherwise yield the previous one's hash.
+		projection.TransactionHash = felt.Zero
+		if err := encoder.Unmarshal(b.Data[offsets[i]:end], &projection); err != nil {
+			return nil, err
```
brbrr:
```suggestion
			return nil, fmt.Errorf("decoding transaction %d: %w", i, err)
```

rpc/v10/block_test.go
```diff
@@ -435,11 +435,10 @@ func setupMockBlockTest(
 	l1Head *core.L1Head,
 	preConfirmedBase ...*pending.PreConfirmed,
 ) {
-	// mock L1 head
 	if l1Head != nil {
 		mockChain.EXPECT().L1Head().Return(*l1Head, nil).AnyTimes()
 	} else {
-		mockChain.EXPECT().L1Head().Return(core.L1Head{}, db.ErrKeyNotFound)
+		mockChain.EXPECT().L1Head().Return(core.L1Head{}, db.ErrKeyNotFound).AnyTimes()
```
brbrr: hm. This affects quite a lot of tests. Does all of them need this? Can we limit it to an expected number of calls?

core/block_transaction_serializer.go
```diff
@@ -122,4 +131,9 @@ var (
 		struct{},
 		[]*TransactionReceipt,
 	]{}
+	BlockTransactionsAllTransactionHashesPartialSerializer = blockTransactionsPartialSerializer[
+		extractAllTransactionHashes,
+		struct{},
+		[]felt.Felt,
```
RafaelGranza, line 201: I remember you mentioning that using felt.Slice here doesn't pay. Right?

I guess it makes sense, since its `Unmarshal` would never be called. It calls the `felt` `Unmarshal` directly.

core/block_transaction.go
```diff
@@ -100,3 +103,33 @@ type executionStatusProjectionSlice = indexed.LazySlice[receiptExecutionStatusPr
...
 }
+
+// TransactionHashes decodes the transaction-hash field of each transaction into one contiguous
+// slice, skipping the heavier transaction fields.
+func (b *BlockTransactions) TransactionHashes() ([]felt.Felt, error) {
+	offsets := b.Indexes.Transactions
+	sectionEnd := len(b.Data)
+	if len(b.Indexes.Receipts) > 0 {
+		sectionEnd = b.Indexes.Receipts[0]
+	}
```
RafaelGranza: (nit - optional): This is basically the same code as in lines 83-86. I guess we could use a helper here.

core/block_transaction_test.go
```diff
@@ -163,4 +164,19 @@ func TestBlockTransactionsSerializer(t *testing.T) {
 			)
 		}
 	})
+
+	t.Run("BlockTransactionsAllTransactionHashesPartialSerializer", func(t *testing.T) {
+		// The serializer reads each transaction's own hash from the transaction section.
+		expectedHashes := make([]felt.Felt, transactionCount)
+		for i := range transactionCount {
+			expectedHashes[i] = *transactions[i].Hash()
+		}
```
RafaelGranza: You could use the `expectedHashes` from `core/block_transaction_txhash_test.go`

rpc/v10/helpers.go
```diff
@@ -84,29 +84,6 @@ func (h *Handler) blockByID(blockID *BlockID) (*core.Block, *jsonrpc.Error) {
...
-		if err != nil {
-			if errors.Is(err, db.ErrKeyNotFound) {
-				return nil, rpccore.ErrBlockNotFound
-			}
-			return nil, rpccore.ErrInternal.CloneWithData(err)
-		}
-		return txns, nil
-	}
-}
-
```
RafaelGranza, line 109: Isn't it possible to have other helpers to replace this one?
`rpc/v10/block_test.go` seems to have some duplicated code.

rpc/v10/block_test.go
```diff
@@ -461,23 +462,39 @@ func setupMockBlockTest(
...
+			txnHashes, nil).AnyTimes()
+	}
+}
+
+func transactionHashesOf(transactions []core.Transaction) []felt.Felt {
+	hashes := make([]felt.Felt, len(transactions))
+	for i, transaction := range transactions {
+		hashes[i] = *transaction.Hash()
 	}
+	return hashes
```
RafaelGranza, line 503: we have basically this same function with 3 different names:

- `transactionHashesOf`
- `txnHashesOf`
- `expectedHashes`

Would it make sense to unify their names?

RafaelGranza:
Also, it's missing a benchmark

core/block_transaction_serializer.go
```diff
@@ -122,4 +131,9 @@ var (
...
+	]{}
```
infrmtcs, line 202: Why don't we do:
```
func (b *BlockTransactions) TransactionHashes() indexed.LazySlice[transactionHashProjection] {
	end := len(b.Data)
	if len(b.Indexes.Receipts) > 0 {
		end = b.Indexes.Receipts[0]
	}

	return indexed.NewLazySlice[transactionHashProjection](b.Indexes.Transactions, b.Data[:end])
}

type extractTransactionHash struct{}

func (extractTransactionHash) extract(b *BlockTransactions, subKey int) (transactionHashProjection, error) {
	return b.TransactionHashes().Get(subKey)
}
```

```suggestion
	BlockTransactionsAllTransactionHashesPartialSerializer = blockTransactionsPartialSerializer[
		extractTransactionHash,
		int,
		transactionHashProjection,
	]{}
```

EgeCaner: In the code below value escapes to heap, meaning per hash projection will get heap allocated. We end up 1 alloc per hash and 1 alloc for backing array where we collect `All`
```go
func (l LazySlice[T]) get(index int) (T, error) {
	start := l.indexes[index]
	end := len(l.data)
	if index < len(l.indexes)-1 {
		end = l.indexes[index+1]
	}

	var value T. <--- escapes to heap
	err := encoder.Unmarshal(l.data[start:end], &value)
	return value, err
}
```


In this implementation value we pass to encoder.Unmarshall, escapes once and gets reused. 1 alloc for reused object (allocs here does not scale with tx count) and 1 alloc for backing array.

```go
func (b *BlockTransactions) TransactionHashes() ([]felt.Felt, error) {
	offsets := b.Indexes.Transactions
	sectionEnd := len(b.Data)
	if len(b.Indexes.Receipts) > 0 {
		sectionEnd = b.Indexes.Receipts[0]
	}

	hashes := make([]felt.Felt, len(offsets))
	var projection transactionHashProjection  <--- Escapes once and gets reused
	for i := range offsets {
		end := sectionEnd
		if i < len(offsets)-1 {
			end = offsets[i+1]
		}
		// Reset before decoding: CBOR null decodes into a value-typed felt as a no-op, so a
		// transaction stored without a hash would otherwise yield the previous one's hash.
		projection.TransactionHash = felt.Zero
		if err := encoder.Unmarshal(b.Data[offsets[i]:end], &projection); err != nil {
			return nil, err
		}
		if projection.TransactionHash.IsZero() {
			return nil, fmt.Errorf("missing TransactionHash in transaction %d", i)
		}
		hashes[i] = projection.TransactionHash
	}
	return hashes, nil
}
```

core/block_transaction_serializer.go
EgeCaner: I guess what I mentioned what json.Marshall on []*felt.Felt vs []felt.Felt and if custom json marshaller on felt.Slice would help

EgeCaner:
Sequential closed model: `--vus 1 --iterations 100000`, corpus
`bench/rpc/corpus/v0_10/getBlockWithTxHashes.json`. All four runs replay the same
corpus and return the same response bytes (418,594,566), so the comparison is
like-for-like.

`cold` is the first run after node start; `warm` is an immediate repeat with the
page cache and pebble block cache populated.

| Row      | Source file                                       |
| -------- | ------------------------------------------------- |
| `before` | `before-{cold,warm}-getBlockWithTxHashes-seq.json` |
| `after`  | `after-{cold,warm}-getBlockWithTxHashes-seq.json`  |

## `http_req_duration` (ms)

| Run         |   min |   med |   avg |   p90 |   p99 |    max | req/s |
| ----------- | ----: | ----: | ----: | ----: | ----: | -----: | ----: |
| before-cold | 0.131 | 0.507 | 0.722 | 1.269 | 4.910 | 16.381 |  1280 |
| before-warm | 0.133 | 0.400 | 0.629 | 1.139 | 4.794 | 15.702 |  1442 |
| after-cold  | 0.123 | 0.519 | 0.643 | 1.004 | 3.201 | 10.147 |  1425 |
| after-warm  | 0.129 | 0.327 | 0.479 | 0.843 | 3.166 |  9.973 |  1844 |

## before → after

| Stat  |   cold |   warm |
| ----- | -----: | -----: |
| min   |  −6.1% |  −3.0% |
| med   |  +2.4% | −18.3% |
| avg   | −10.9% | −23.8% |
| p90   | −20.9% | −26.0% |
| p99   | −34.8% | −34.0% |
| max   | −38.1% | −36.5% |
| req/s | +11.3% | +27.9% |

## cold → warm (cache sensitivity)

| Stat  | before | after  |
| ----- | -----: | -----: |
| min   |  +1.5% |  +4.9% |
| med   | −21.1% | −37.0% |
| avg   | −12.8% | −25.5% |
| p90   | −10.2% | −16.0% |
| p99   |  −2.4% |  −1.1% |
| max   |  −4.1% |  −1.7% |
| req/s | +12.6% | +29.4% |

## Go micro-benchmark: before → after

`BenchmarkBlockWithTxHashes` over a temp pebble DB, `-count 10`, merge base
`6d423c72a` vs branch HEAD, compared with `benchstat`. Sources:
`before-BlockWithTxHashes.txt`, `after-BlockWithTxHashes.txt`.

| benchmark | before | after | change |
| --------- | -----: | ----: | -----: |
| **ns/op** | | | |
| `block=0_txs=18` | 38.31µ ± 1% | 26.17µ ± 1% | **−31.70%** (p=0.000) |
| `block=763497_txs=17` | 195.9µ ± 1% | 120.8µ ± 0% | **−38.32%** (p=0.000) |
| geomean | 86.63µ | 56.23µ | −35.09% |
| **B/op** | | | |
| `block=0_txs=18` | 12.073Ki | 4.059Ki | −66.38% (p=0.000) |
| `block=763497_txs=17` | 83.725Ki | 4.214Ki | −94.97% (p=0.000) |
| geomean | 31.79Ki | 4.136Ki | −86.99% |
| **allocs/op** | | | |
| `block=0_txs=18` | 237 | 43 | −81.86% (p=0.000) |
| `block=763497_txs=17` | 281 | 50 | −82.21% (p=0.000) |

core/block_transaction_serializer.go
```diff
@@ -78,6 +80,22 @@ func (extractExecutionStatus) extract(
 	}, nil
 }
 
+type extractAllTransactionHashes struct{}
+
+// extract reads each transaction's own TransactionHash field, decoded without the rest of the
+// transaction.
+func (extractAllTransactionHashes) extract(b *BlockTransactions, _ struct{}) ([]felt.Felt, error) {
+	return indexed.AllMapped(
```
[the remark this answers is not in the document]
RafaelGranza: I think its super minor.

I don't know the reasoning for the AllMapped, but its your call to change it or not.

core/block_transaction.go
```diff
@@ -85,32 +85,51 @@ func NewBlockTransactions(
...
+	transactionHashProjectionSlice   = indexed.LazySlice[transactionHashProjection]
+)
 
 // executionStatusProjections decodes receipts into the execution-status subset, skipping the
 // heavier receipt fields.
 func (b *BlockTransactions) executionStatusProjections() executionStatusProjectionSlice {
-	return indexed.NewLazySlice[receiptExecutionStatusProjection](b.Indexes.Receipts, b.Data)
+	return indexed.NewLazySlice[receiptExecutionStatusProjection](
+		b.Indexes.Receipts, b.receiptsSection(),
+	)
```
RafaelGranza, line 122: linter right?

rpc/v10/block.go
```diff
@@ -176,49 +176,65 @@ func (h *Handler) BlockTransactionCount(id *BlockID) (uint64, *jsonrpc.Error) {
...
+				return nil, rpccore.ErrBlockNotFound
+			}
+			return nil, rpccore.ErrInternal.CloneWithData(err)
+		}
+		preConfirmed := preConfirmedChain.Head()
+		if preConfirmed == nil {
+			return nil, rpccore.ErrBlockNotFound
+		}
+		return &BlockWithTxHashes{
+			Status:      BlockPreConfirmed,
```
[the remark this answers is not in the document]
RafaelGranza: not happy with it. I guess `chain.Head()` is not as optimized as the other checks.

Keep the versions without helpers.

rpc/v9/block_test.go
```diff
@@ -288,6 +288,14 @@ func TestBlockTransactionCount(t *testing.T) {
...
 }
 
+func transactionHashesOf(transactions []core.Transaction) []felt.Felt {
+	hashes := make([]felt.Felt, len(transactions))
+	for i, transaction := range transactions {
+		hashes[i] = *transaction.Hash()
+	}
+	return hashes
+}
+
```
RafaelGranza, line 298: Why not using the `transactionHashesOf` from `block.go`?

rpc/v10/block_test.go
```diff
@@ -467,23 +469,39 @@ func setupMockBlockTest(
 		mockChain.EXPECT().HeadsHeader().Return(block.Header, nil).AnyTimes()
 		mockChain.EXPECT().TransactionsByBlockNumber(block.Number).Return(
 			block.Transactions, nil).AnyTimes()
+		mockChain.EXPECT().TransactionHashesByBlockNumber(block.Number).Return(
+			transactionHashes, nil).AnyTimes()
```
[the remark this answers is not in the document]
RafaelGranza: I may be missing something, but moving to `.Times(1)` sounds fine.

rpc/v10/helpers.go
RafaelGranza: no longer interested in this

core/partial_cbor_test.go
```diff
@@ -429,3 +430,132 @@ func BenchmarkExecutionStatusProjection(b *testing.B) {
...
+func TestTransactionHashProjectionDecodesHash(t *testing.T) {
+	for _, transaction := range sampleTransactions() {
+		t.Run(fmt.Sprintf("%T", transaction), func(t *testing.T) {
+			var projection transactionHashProjection
+			require.NoError(t, encoder.Unmarshal(sampleTransactionBytes(t, transaction), &projection))
+			require.Equal(t, *transaction.Hash(), projection.TransactionHash,
+				"TransactionHash must receive the wire value, not discardedCBOR")
+		})
+	}
+}
```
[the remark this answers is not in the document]
RafaelGranza: I'm in favor of a benchmark, but measure CPU and memory usage, not only memory.

We don't know what kind of changes we can see in the future 😶‍🌫️

RafaelGranza (approved):
LGTM.

Only nits, address the comments you agree are relevant.

rpc/v10/block_test.go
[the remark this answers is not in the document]
EgeCaner: Times(1) should break it since same mock setup is shared by other BlockWith...  tests, it wont get called for others. I guess I can convert to MaxTime(1) etc...

core/partial_cbor_test.go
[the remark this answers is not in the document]
EgeCaner: Added minimal bench

| Transaction type | ns/op (full → projection) | B/op (full → projection) | allocs/op (full → projection) |
  |---|---|---|---|
  | Deploy | 1373 → 708 (1.9× faster) | 344 → 48 | 11 → 1 |
  | DeployAccount | 2732 → 1511 (1.8× faster) | 992 → 48 | 22 → 1 |
  | Invoke | 3318 → 1868 (1.8× faster) | 1216 → 48 | 26 → 1 |
  | Declare | 2829 → 1535 (1.8× faster) | 992 → 48 | 22 → 1 |
  | L1Handler | 1338 → 718 (1.9× faster) | 344 → 48 | 11 → 1 |

rpc/v10/block_test.go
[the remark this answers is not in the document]
EgeCaner: Constrained the mocks

core/block_transaction_serializer.go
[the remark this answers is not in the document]
EgeCaner: addressed

rpc/v9/block_test.go
EgeCaner: It has to be exported, it is minimal I prefer to keep it

brbrr (approved):
LGTM!

core/partial_cbor_test.go
```diff
@@ -480,3 +481,158 @@ func BenchmarkExecutionStatusProjection(b *testing.B) {
...
+			b.Run("hash_projection", func(b *testing.B) {
+				b.ReportAllocs()
+				for b.Loop() {
+					var projection transactionHashProjection
+					_ = encoder.Unmarshal(data, &projection)
+				}
+			})
+		})
+	}
+}
```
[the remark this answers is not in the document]
RafaelGranza: this benchmark fulfills my needs

core/indexed/lazy_slice.go
```diff
@@ -66,7 +67,7 @@ func AllMapped[T, R any](
 		// Reset: decoding merges into the existing value, so stale data would leak across elements.
 		value = *new(T)
 		if err := l.getInto(i, &value); err != nil {
-			return nil, err
+			return nil, fmt.Errorf("decoding element %d: %w", i, err)
```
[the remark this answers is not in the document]
RafaelGranza: not relevant

PR #3921, opened 2026-08-10
perf(blockchain): serve chain height and L1 head from memory

brbrr:
### **User description**
This caches the reads behind `Height()`, `Head()`, `HeadsHeader()` and the `EventFilter` upper bound. `HeadState()`, `EventFilter.Events` and `SetRangeEndBlockToL1Head` still read the database on each call - `Events` deliberately, so its range logic and retention check use one source of "latest".f

sepolia based benchamk results:

| Benchmark | Metric | Before (main) | After (head-cache) | Δ | p-value |
|---|---|---|---|---|---|
| `heads_header` | time/op | 7.186 µs | 5.398 µs | -24.89% | 0.002 |
| `heads_header` | bytes/op | 3.29 KiB | 3.04 KiB | -7.61% | 0.002 |
| `heads_header` | allocs/op | 38 | 33 | -13.16% | 0.002 |
| `height` | time/op | 1610.0 ns | 123.3 ns | -92.34% | 0.002 |
| `height` | bytes/op | 256 B | 0 B | -100.00% | 0.002 |
| `height` | allocs/op | 5 | 0 | -100.00% | 0.002 |
| `l1_head` | time/op | 2176.0 ns | 122.4 ns | -94.37% | 0.002 |
| `l1_head` | bytes/op | 329 B | 0 B | -100.00% | 0.002 |
| `l1_head` | allocs/op | 7 | 0 | -100.00% | 0.002 |

___

### **PR Type**
Enhancement, Tests

___

### **Description**
- Cache chain height and L1 head in memory via `atomic.Pointer`

- Add `WithRemoteDatabase` option to disable caching for remote-db followers

- Refresh/clear cached height on `Store`, `Finalise`, `RevertHead`, `StoreGenesis`

- Deep-copy L1 head on cache to isolate it from caller mutations

- Wire `--remote-db` flag to `blockchain.WithRemoteDatabase` in node setup

- Extensive new tests covering caching, isolation, and remote-db fallback

___

<details> <summary><h3> File Walkthrough</h3></summary>

<table><thead><tr><th></th><th align="left">Relevant files</th></tr></thead><tbody><tr><td><strong>Enhancement</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>blockchain.go</strong><dd><code>Cache chain height and L1 head in memory</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

blockchain/blockchain.go

<ul><li>Add <code>chainHeight</code> and <code>l1Head</code> atomic pointer fields to <code>Blockchain</code>, plus a <br><code>cacheHeads</code> flag<br> <li> Add <code>WithRemoteDatabase</code> option to disable caching for databases written <br>by another process<br> <li> Populate the cache at construction time from the database, and add <br><code>height()</code>/<code>cacheChainHeight()</code>/<code>cacheL1Head()</code> helpers<br> <li> Update <code>Height()</code>, <code>Head()</code>, <code>HeadsHeader()</code>, <code>L1Head()</code>, <code>SetL1Head()</code>, <br><code>Store()</code>, <code>Finalise()</code>, <code>RevertHead()</code> to read/refresh the in-memory cache <br>instead of always hitting the database<br> <li> Keep <code>EventFilter</code>'s height read against the database to share the <br>"latest" source with <code>Events</code></ul>

</details>

  </td>
  <td><a href="https://github.com/NethermindEth/juno/pull/3921/files#diff-37b6e7ed4f299b87b357b54376123393b34ee0699fc035f25bc0621627dcc6ba">+101/-4</a>&nbsp; </td>

</tr>

<tr>
  <td>
    <details>
      <summary><strong>node.go</strong><dd><code>Wire remote-db flag to blockchain option</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

node/node.go

<ul><li>Append <code>blockchain.WithRemoteDatabase()</code> option when <code>dbIsRemote</code> is true</ul>

</details>

  </td>
  <td><a href="https://github.com/NethermindEth/juno/pull/3921/files#diff-375d57e386f20eaa5f09f02bb9d28bfc48ac3dca18d0325f59492208219e5618">+3/-0</a>&nbsp; &nbsp; &nbsp; </td>

</tr>
</table></td></tr><tr><td><strong>Tests</strong></td><td><table>
<tr>
  <td>
    <details>
      <summary><strong>blockchain_test.go</strong><dd><code>Add tests for head caching behavior</code>&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; </dd></summary>
<hr>

blockchain/blockchain_test.go

<ul><li>Rename <code>chainHeightCounter</code> to <code>headReadCounter</code>, tracking both chain <br>height and L1 head reads<br> <li> Add tests verifying heights/L1 heads are served from memory without <br>hitting the database<br> <li> Add tests for L1 head cache isolation from caller mutations, clearing <br>cache on nil update, and failed writes not reaching cache/subscribers<br> <li> Add tests for cache population on restart, remote-database fallback <br>behavior, and cache invalidation/refresh across revert scenarios</ul>

</details>

  </td>
  <td><a href="https://github.com/NethermindEth/juno/pull/3921/files#diff-29f3c6e8d6136889adad56f05676a458cb547cd95bdc57ba71a6c2be0353123e">+285/-11</a></td>

</tr>
</table></td></tr></tr></tbody></table>

</details>

___


blockchain/blockchain.go
```diff
@@ -3,6 +3,7 @@ package blockchain
 import (
```
EgeCaner, line 3: Here I worry about the tight window where db is updated and cache is not, we might have to synchronize them, or invalidate cache before starting writing to DB, so we don't read stale cache value, at the moment nothing hints cache is stale. We will still have concurrent DB read/write case but it might still worth

blockchain/blockchain.go
```diff
@@ -358,14 +410,43 @@ func (b *Blockchain) SubscribeL1Head() L1HeadSubscription {
...
 func (b *Blockchain) L1Head() (core.L1Head, error) {
 	b.listener.OnRead("L1Head")
+	if l1Head := b.l1Head.Load(); l1Head != nil {
+		return *l1Head, nil
+	}
 	return core.GetL1Head(b.database)
 }
 
 func (b *Blockchain) SetL1Head(update *core.L1Head) error {
 	b.l1HeadFeed.Send(update)
```
EgeCaner: Not coming from this PR but I guess we should not publish L1 head before committing it.

EgeCaner (changes requested):
Here I worry about cache DB consistency, lets somehow make it more robust. Also I guess we can update the caches directly after writes without requiring to read the ChainHeight or L1Head back from DB in order to populate the cache

blockchain/blockchain.go
brbrr: A reader does a lock-free atomic load, so a reader is safe as it is. The pair {database commit, cache refresh} in a writer is not atomic, but the node has only one writer.

`Store` and `RevertHead` are stream callbacks on the same stream, and `storeTask` calls `revertTask` on that same goroutine. There is only one caller of `Store` in the tree. 

Thus a mutex would protect an interleave that no call path can cause, and it would tell a future reader that this package expects concurrent writers. I prefer to keep the rule in the field comment, where a grep can check it. It would make sense adding a mutex when we'll get a second writer.

You also asked to set the cache directly instead of a read of the database. `Store` and `Finalise` can do this, because it writes exactly `block.Number`. `RevertHead` cannot: `Blockchain` does not know the new height unless it reads the database or repeats the rule "delete at genesis, else N-1" in a second place. `RevertHead` also cannot calculate the new height when the cache is already nil. Thus the choice is one read-back rule for all three methods, or a mixed scheme with the height rule in two places. I decided to keep the read-back.

brbrr:
some K6 experiements:

### 1. Sequential latency — `--vus 1`, one call per request

User-visible per-call latency, no concurrency. Minimum over 20000 iterations per method:

| method | cached read | main µs | pr µs | Δ |
| --- | --- | ---: | ---: | ---: |
| `blockNumber` | `Height` | 47 | 44 | **-3.0** |
| `blockHashAndNumber` | `HeadsHeader` | 55 | 52 | **-3.0** |
| `getTransactionStatus` | `L1Head` | 64 | 61 | **-3.5** |
| `getBlockTransactionCount` (latest) | `Height` | 54 | 52 | **-2.0** |
| `getStorageAt` (by number) | none — reference | 62 | 62 | 0.0 |
| `chainId` | none — control | 44 | 44 | 0.0 |

Both no-op rows sit at zero; all five affected methods move -1.5 to -3.5 µs.

### 2. Under load — `--vus 200`, one call per request

| method | cached read | main req/s | pr req/s | Δ req/s | main p99 ms | pr p99 ms | Δ p99 |
| --- | --- | ---: | ---: | ---: | ---: | ---: | ---: |
| `blockNumber` | `Height` | 58,967 | 60,214 | +1.2% | 16.3 | 15.3 | -4.9% |
| `blockHashAndNumber` | `HeadsHeader` | 53,352 | 54,249 | -1.0% | 18.6 | 18.4 | -0.8% |
| `getBlockTransactionCount` (latest) | `Height` | 53,835 | 53,062 | -2.0% | 18.6 | 18.3 | +0.2% |
| `getTransactionStatus` | `L1Head` | 44,787 | 45,409 | -1.8% | 22.7 | 22.2 | -1.3% |
| `getStorageAt` (by number) | none — reference | 44,633 | 44,203 | -3.7% | 22.8 | 22.8 | +0.1% |

rodrodros:
I don't think this slight performance improvement is worth the extra complexity. We are bringing a DB optimization to an upper layer. I am worried about future inconsistencies of db height and memory height being different. I think this is something for our DB engine to sort it out (I believe it already does). 

Would like to hear other opinions: @EgeCaner @brbrr @infrmtcs

rodrodros:
I will close the PR

PR #3930, opened 2026-08-11
feat: add nightly RPC benchmark harness

danielntmd:
## Summary

  Adds the Juno-side foundation for reproducible nightly RPC performance benchmarks against a fixed mainnet snapshot.
 
## Changes

  - Adds a fixed, seeded mainnet corpus for starknet_getTransactionByHash.
  - Adds a pinned k6 runner covering sequential latency, concurrency, and throughput.
  - Validates node readiness, chain, block height, corpus checksum, and matching Juno/runner versions before measurement.
  - Records benchmark configuration, provenance, results, saturation, and failures in a manifest.
  - Adds integration tests for successful runs, invalid configuration, mismatched targets, RPC failures, saturation, and graceful termination.
  - Publishes matched Juno and runner images using immutable commit tags and moving nightly tags.
  - Supports exact commit versioning for dedicated benchmark images without changing the Makefile or normal Juno builds.
  
  See https://github.com/NethermindEth/argocd/pull/9780


bench/rpc/cmd/runner/manifest.go
```diff
@@ -0,0 +1,243 @@
...
+	}
+}
+
+func (r *runner) writeManifest() error {
+	data, err := json.MarshalIndent(r.currentManifest(), "", "  ")
+	if err != nil {
+		return fmt.Errorf("marshal manifest: %w", err)
+	}
+	data = append(data, '\n')
+	return writeAtomic(filepath.Join(r.config.resultsDir, "manifest.json"), data)
```
brbrr: Why is it named `atomic`? If it's referring to an atomic operation, I don't think it's actually atomic.

bench/rpc/cmd/runner/runner.go
```diff
@@ -0,0 +1,698 @@
+package main
```
brbrr, line 1:
```suggestion
package k6runner
```

Or something along these lines

Dockerfile
```diff
@@ -43,8 +44,16 @@ RUN go mod download
 # Copy the rest of the source
 COPY . .
 
-# Build with make juno
-RUN make juno
+# Dedicated benchmark images provide an exact source version. Existing image
+# builds retain the Makefile's git-describe version.
```
brbrr: Have you considered updating Makefile to something like this:

```
JUNO_VERSION ?= $(shell git describe --tags)
juno: rustdeps
      @go build $(GO_TAGS) -a -ldflags="-X main.Version=$(JUNO_VERSION)" -o build/juno ./cmd/juno/
```

danielntmd: I started with this approach, but failed the full suite CI (mandated by makefile change) on ARM and macOS.

I'll restore and try again.

bench/rpc/cmd/runner/runner.go
danielntmd: We should keep it as main for now as `bench/rpc/cmd/corpus-gen` and `cmd/juno`also follows this; renaming it would require exporting an entry point and more directory churn. 

If we have another consumer, then we can rename and create those internal file structure boundaries.

bench/rpc/cmd/runner/manifest.go
danielntmd: Ah, naming changed during migration, will rename since it does guarantee stronger guarantees implied by atomic.

Makefile
```diff
@@ -30,18 +30,19 @@ else
 endif
 
 PKG ?= ./...
+JUNO_VERSION ?= $(shell git describe --tags)
```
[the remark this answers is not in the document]
danielntmd: Worthwhile hardening, but current workflow is safe because it supplies non-empty sha

.github/workflows/benchmark-image.yaml
```diff
@@ -0,0 +1,136 @@
+name: Build RPC Benchmark Images
+
+on:
+  pull_request:
+    paths:
+      - "bench/rpc/**"
+      - ".github/workflows/benchmark-image.yaml"
+      # The Juno image is built from these, which carry the version plumbing.
+      - "Dockerfile"
+      - "Makefile"
```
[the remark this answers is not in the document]
danielntmd: sure

danielntmd:
superseded by #3944

PR #3942, opened 2026-08-13
ci: add path filter to docs-deploy workflow

Mgabal:
Prevent docs deployment from running on every push to main. Only trigger when files under docs/ or the workflow itself change. This avoids unnecessary CI usage and GitHub Pages deployments when unrelated code changes are pushed.

### DIFF
--- a/.github/workflows/docs-deploy.yml
+++ b/.github/workflows/docs-deploy.yml
@@ -4,6 +4,9 @@ on:
   push:
     branches:
       - main
+    paths:
+      - 'docs/**'
+      - '.github/workflows/docs-deploy.yml'
   workflow_dispatch:
 
 permissions:


.github/workflows/docs-deploy.yml
```diff
@@ -4,6 +4,9 @@ on:
   push:
     branches:
       - main
+    paths:
+      - 'docs/**'
+      - '.github/workflows/docs-deploy.yml'
```
brbrr, line 9: `npm run build` runs docs/generate-config.js which fetches juno.go from upstram. We probably should add it into the list of paths as well

brbrr (changes requested):
Thanks for the contribution. Our docs generation pipeline uses juno codebase to generate the docs, so we should include the relevant file(s) in the paths.

brbrr:
I just realized that this is a duplicate of https://github.com/NethermindEth/juno/pull/3901

PR #3943, opened 2026-08-13
refactor(l1): make logger a functional option in NewClient

### LINKED ISSUE #3820: l1: make logger a functional option in NewClient
### Context

Follow-up from review of #3738.

`l1.NewClient` currently takes `logger log.StructuredLogger` as a required positional parameter:

```go
func NewClient(
    provider L1StateProvider,
    chain *blockchain.Blockchain,
    logger log.StructuredLogger,
    opts ...Option,
) *Client
```

Mgabal:
## Summary
Closes #3820

## Changes
- Added `Logger` field to the internal `options` struct
- Added `WithLogger(log.StructuredLogger) Option` functional option
- Removed `logger` as a required positional parameter from `NewClient`
- Default logger is `log.NewNopZapLogger()` when not provided
- Updated all callers: `node/migration.go`, `node/node.go`
- Updated all test files to use `l1.WithLogger(nopLog)`

## Motivation
Makes `NewClient` consistent with the existing functional-options
pattern already used for `EventListener`, `ResubscribeDelay`,
`PollFinalisedInterval`, and `CatchUpChunkSize`. Logger becomes
optional with a safe default rather than a required argument.

### DIFF
--- a/.github/workflows/docs-deploy.yml
+++ b/.github/workflows/docs-deploy.yml
@@ -4,9 +4,6 @@ on:
   push:
     branches:
       - main
-    paths:
-      - 'docs/**'
-      - '.github/workflows/docs-deploy.yml'
   workflow_dispatch:
 
 permissions:
--- a/.github/workflows/docs-deploy.yml
+++ b/.github/workflows/docs-deploy.yml
@@ -4,6 +4,9 @@ on:
   push:
     branches:
       - main
+    paths:
+      - 'docs/**'
+      - '.github/workflows/docs-deploy.yml'
   workflow_dispatch:
 
 permissions:
--- a/l1/l1.go
+++ b/l1/l1.go
@@ -21,6 +21,8 @@ type Client struct {
 	network               *networks.Network
 	resubscribeDelay      time.Duration
 	pollFinalisedInterval time.Duration
+	// CatchUpChunkSize is the L1 block range per backward eth_getLogs request
+	// during the startup catch-up scan.
 	catchUpChunkSize      uint64
 	nonFinalisedLogs      map[uint64]*StateUpdate
 	listener              EventListener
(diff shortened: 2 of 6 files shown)


.github/workflows/docs-deploy.yml
```diff
@@ -4,6 +4,9 @@ on:
   push:
     branches:
       - main
+    paths:
+      - 'docs/**'
+      - '.github/workflows/docs-deploy.yml'
```
brbrr: These changes are irrelevant to this PR

l1/l1.go
```diff
@@ -36,9 +36,8 @@ type options struct {
 	EventListener         EventListener
 	ResubscribeDelay      time.Duration
 	PollFinalisedInterval time.Duration
-	// CatchUpChunkSize is the L1 block range per backward eth_getLogs request
-	// during the startup catch-up scan.
-	CatchUpChunkSize uint64
+	CatchUpChunkSize      uint64
```
brbrr, line 41: Why did you remove the comment above?

node/node.go
```diff
@@ -721,7 +721,7 @@ func newL1Client(
 		registerL1Metrics(provider)
 	}
 
-	return l1.NewClient(provider, chain, logger, l1Opts...), provider, nil
+	return l1.NewClient(provider, chain, append(l1Opts, l1.WithLogger(logger))...), provider, nil
```
brbrr: let's move the append from the function call.

brbrr (changes requested):
Please take a look at inline comments

l1/l1.go
brbrr: Please double-check what your LLM generates. The comment was moved to the wrong field.

brbrr (changes requested):
check comment

PR #3944, opened 2026-08-14
feat(bench): add managed nightly RPC benchmarks

danielntmd:
- Adds a nightly.sh run script that handles corpus generation/removal and warmup/measured k6 runs.
- Publish a dedicated benchmark image.
- Nightly.json catalog validation in CI.

### DIFF
--- a/.github/workflows/build-image.yaml
+++ b/.github/workflows/build-image.yaml
@@ -62,6 +62,6 @@ jobs:
           context: .
           push: true
           platforms: linux/amd64
-          tags: ${{ env.DOCKER_REGISTRY }}/${{ env.REPO_DEV }}/juno:${{ steps.set_tag.outputs.DOCKER_IMAGE_TAG  }}
+          tags: ${{ env.DOCKER_REGISTRY }}/${{ env.REPO_DEV }}/juno:${{ steps.set_tag.outputs.DOCKER_IMAGE_TAG }}${{ github.ref == 'refs/heads/main' && format(',{0}/{1}/juno:main', env.DOCKER_REGISTRY, env.REPO_DEV) || '' }}
           cache-from: type=gha,scope=${{ github.workflow }}
           cache-to: type=gha,mode=max,scope=${{ github.workflow }}
--- a/.github/workflows/rpc-benchmark.yaml
+++ b/.github/workflows/rpc-benchmark.yaml
@@ -1,0 +1,68 @@
+name: RPC Benchmark Image
+
+on:
+  pull_request:
+    paths:
+      - "bench/rpc/**"
+      - ".github/workflows/rpc-benchmark.yaml"
+  push:
+    branches:
+      - main
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
+
+env:
+  DOCKER_REGISTRY: nethermind.jfrog.io
+  BENCHMARK_IMAGE_REPOSITORY: nubia-oci-local-dev/juno-rpc-benchmark
+
+jobs:
+  build-image:
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+    permissions:
+      contents: read
+      id-token: write
+    steps:
+      - name: Checkout
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+
+      - name: Validate nightly catalog
+        run: bench/rpc/nightly/test/test-nightly.sh
+
+      - name: Setup Docker Buildx
+        uses: docker/setup-buildx-action@bb05f3f5519dd87d3ba754cc423b652a5edd6d2c # v4.2.0
+
+      - name: Install JFrog CLI
+        if: github.ref == 'refs/heads/main'
+        id: jfrog
+        uses: jfrog/setup-jfrog-cli@deda456d982fc5e9a7a020b63eb0d2968aedd33e # v5.1.0
+        env:
+          JF_URL: https://${{ env.DOCKER_REGISTRY }}
+        with:
+          oidc-provider-name: github-nethermindeth
+
+      - name: Login to registry with OIDC
+        if: github.ref == 'refs/heads/main'
+        uses: docker/login-action@dbcb813823bdd20940b903addbd779551569679f # v4.6.0
+        with:
+          registry: ${{ env.DOCKER_REGISTRY }}
+          username: ${{ steps.jfrog.outputs.oidc-user }}
+          password: ${{ steps.jfrog.outputs.oidc-token }}
+
+      - name: Build benchmark image
+        uses: docker/build-push-action@53b7df96c91f9c12dcc8a07bcb9ccacbed38856a # v7.0.0
+        with:
+          context: .
+          file: bench/rpc/nightly/Dockerfile
+          tags: |
+            ${{ env.DOCKER_REGISTRY }}/${{ env.BENCHMARK_IMAGE_REPOSITORY }}:latest
+            ${{ env.DOCKER_REGISTRY }}/${{ env.BENCHMARK_IMAGE_REPOSITORY }}:sha-${{ github.sha }}
+          cache-from: type=gha,scope=rpc-benchmark
+          cache-to: type=gha,mode=max,scope=rpc-benchmark
+          push: ${{ github.ref == 'refs/heads/main' }}
--- a/bench/rpc/corpus/nightly.json
+++ b/bench/rpc/corpus/nightly.json
@@ -1,0 +1,29 @@
+{
+  "blockHashAndNumber": "blockHashAndNumber --count=100000 --seed=1",
+  "blockNumber": "blockNumber --count=100000 --seed=1",
+  "chainId": "chainId --count=100000 --seed=1",
+  "syncing": "syncing --count=100000 --seed=1",
+
+  "getBlockTransactionCount": "getBlockTransactionCount --count=100000 --seed=1",
+  "getBlockWithReceipts": "getBlockWithReceipts --count=100000 --seed=1",
+  "getBlockWithTxHashes": "getBlockWithTxHashes --count=100000 --seed=1",
+  "getBlockWithTxs": "getBlockWithTxs --count=100000 --seed=1",
+  "getStateUpdate": "getStateUpdate --count=100000 --seed=1",
+  "traceBlockTransactions": "traceBlockTransactions --count=10000 --seed=1",
+
+  "getTransactionByBlockIdAndIndex": "getTransactionByBlockIdAndIndex --count=100000 --seed=1",
+  "getTransactionByHash": "getTransactionByHash --count=100000 --seed=1",
+  "getTransactionReceipt": "getTransactionReceipt --count=100000 --seed=1",
+  "getTransactionStatus": "getTransactionStatus --count=100000 --seed=1",
+  "traceTransaction": "traceTransaction --count=10000 --seed=1",
+
+  "getClass": "getClass --count=100000 --seed=1",
+  "getClassAt": "getClassAt --count=100000 --seed=1",
+  "getClassHashAt": "getClassHashAt --count=100000 --seed=1",
+  "getCompiledCasm": "getCompiledCasm --count=100000 --seed=1",
+  "getNonce": "getNonce --count=100000 --seed=1",
+  "getStorageAt": "getStorageAt --count=100000 --seed=1",
+
+  "getEvents": "getEvents --count=100000 --seed=1",
+  "getStorageProof": "getStorageProof --count=100000 --seed=1"
+}
--- a/bench/rpc/nightly/Dockerfile
+++ b/bench/rpc/nightly/Dockerfile
@@ -1,0 +1,29 @@
+# syntax=docker/dockerfile:1.7.0
+
+FROM golang:1.26-bookworm AS builder
+
+WORKDIR /src
+COPY go.mod go.sum ./
+COPY starknet-p2p-specs/go.mod starknet-p2p-specs/go.mod
+RUN go mod download
+
+COPY Makefile ./
+COPY bench/rpc/cmd/corpus-gen/ bench/rpc/cmd/corpus-gen/
+RUN CGO_ENABLED=0 make corpus-gen
+
+FROM docker.io/grafana/k6:1.7.1@sha256:4fd3a694926b064d3491d9b02b01cde886583c4931f1223816e3d9a7bdfa7e0f
+
+USER root
+RUN apk add --no-cache bash curl jq && \
+    mkdir -p /corpus /results && \
+    chown k6:k6 /corpus /results
+
+WORKDIR /app
+COPY --from=builder /src/build/corpus-gen build/corpus-gen
+COPY bench/rpc/nightly/nightly.sh bench/rpc/nightly/ci.json bench/rpc/nightly/
+COPY bench/rpc/nightly/internal/ bench/rpc/nightly/internal/
+COPY bench/rpc/run.js bench/rpc/run.js
+COPY bench/rpc/corpus/nightly.json bench/rpc/corpus/nightly.json
+
+USER k6
+ENTRYPOINT ["/app/bench/rpc/nightly/nightly.sh"]
--- a/bench/rpc/nightly/internal/corpus.sh
+++ b/bench/rpc/nightly/internal/corpus.sh
@@ -1,0 +1,73 @@
+generate_case() (
+  set -euo pipefail
+  local name=$1 command=$2 work case_dir
+  local -a sub
+
+  case_dir="$CORPUS_ROOT/$name"
+  work=$(mktemp -d "$CORPUS_ROOT/.staging/$name.XXXXXX")
+  trap 'rm -rf "$work"' EXIT INT TERM
+
+  read -ra sub <<<"$command"
+  "$CORPUS_GEN" --source-url "$NODE_URL" \
+    --concurrency="${CONCURRENCY:-32}" \
+    "${sub[@]}" >"$work/corpus.json" || return 1
+  printf '%s\n' "$command" >"$work/command"
+  [[ ! -e $case_dir ]] || {
+    echo "error: case appeared during generation: $name" >&2
+    return 1
+  }
+  mv -T "$work" "$case_dir"
+  work=
+  trap - EXIT INT TERM
+)
+
+corpus_is_current() {
+  local case_dir=$1 command=$2
+  [[ -f $case_dir/command ]] &&
+    [[ $(<"$case_dir/command") == "$command" ]]
+}
+
+prune_removed_cases() {
+  local case_dir name
+  for case_dir in "$CORPUS_ROOT"/*; do
+    [[ -d $case_dir ]] || continue
+    name=${case_dir##*/}
+    if ! jq -e --arg name "$name" 'has($name)' "$CATALOG" >/dev/null; then
+      echo "removing $name"
+      rm -rf -- "$case_dir"
+    fi
+  done
+}
+
+generate() {
+  [[ -x $CORPUS_GEN ]] || {
+    echo "error: $CORPUS_GEN not found; run 'make corpus-gen' first" >&2
+    return 1
+  }
+  mkdir -p "$CORPUS_ROOT/.staging"
+
+  local name command case_dir
+  local -a failed=()
+  while IFS=$'\t' read -r name command; do
+    echo "==> $name"
+    case_dir="$CORPUS_ROOT/$name"
+    if [[ -d $case_dir ]]; then
+      if corpus_is_current "$case_dir" "$command"; then
+        echo "reusing $name"
+        continue
+      fi
+      echo "removing outdated $name"
+      rm -rf -- "$case_dir"
+    fi
+    if ! generate_case "$name" "$command"; then
+      failed+=("$name")
+    fi
+  done < <(jq -r 'to_entries[] | [.key, .value] | @tsv' "$CATALOG")
+
+  if ((${#failed[@]} > 0)); then
+    echo "failed: ${failed[*]}" >&2
+    return 1
+  fi
+  prune_removed_cases
+  echo "corpora written to $CORPUS_ROOT"
+}
--- a/bench/rpc/nightly/nightly.sh
+++ b/bench/rpc/nightly/nightly.sh
@@ -1,0 +1,95 @@
+#!/usr/bin/env bash
+# Stable entrypoint for the managed nightly RPC benchmark.
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+RPC_DIR="$(cd "$SCRIPT_DIR/.." && pwd)"
+CATALOG="$RPC_DIR/corpus/nightly.json"
+CI_CONFIG="$SCRIPT_DIR/ci.json"
+CORPUS_GEN=${CORPUS_GEN:-"$RPC_DIR/../../build/corpus-gen"}
+RUN_SCRIPT="$RPC_DIR/run.js"
+
+: "${NODE_URL:?NODE_URL is required}"
+: "${SNAPSHOT_SHA256:?SNAPSHOT_SHA256 is required}"
+[[ $SNAPSHOT_SHA256 =~ ^[0-9a-f]{64}$ ]] || {
+  echo "error: SNAPSHOT_SHA256 must be a lowercase SHA-256 digest" >&2
+  exit 1
+}
+CORPUS_ROOT="${CORPUS_ROOT:-/corpus/cases}/$SNAPSHOT_SHA256"
+
+source "$SCRIPT_DIR/internal/preflight.sh"
+source "$SCRIPT_DIR/internal/corpus.sh"
+
+case_path() {
+  printf '%s/%s/corpus.json\n' "$CORPUS_ROOT" "$1"
+}
+
+benchmark() {
+  [[ -f $CI_CONFIG ]] || {
+    echo "error: k6 config not found: $CI_CONFIG" >&2
+    return 1
+  }
+  local out_dir=${RESULTS_DIR:-/results}
+  local warmup_iterations=${WARMUP_ITERATIONS:-200}
+  mkdir -p "$out_dir"
+
+  local name corpus summary case_failed
+  local -a failed=()
+  while IFS= read -r name; do
+    summary="$out_dir/$name.json"
+
+    echo
+    echo "==> $name"
+    corpus=$(case_path "$name")
+    if [[ ! -f $corpus ]]; then
+      echo "error: missing corpus for $name: $corpus" >&2
+      failed+=("$name")
+      continue
+    fi
+    corpus="$(cd "$(dirname "$corpus")" && pwd)/$(basename "$corpus")"
+
+    case_failed=0
+    if [[ $warmup_iterations != 0 ]]; then
+      echo "warming up"
+      (
+        K6_WEB_DASHBOARD=false k6 run \
+          --vus 1 \
+          --iterations "$warmup_iterations" \
+          --tag "case_id=$name" \
+          --tag phase=warmup \
+          -e NODE_URL="$NODE_URL" \
+          "$RUN_SCRIPT" <"$corpus" >/dev/null
+      ) || {
+        echo "error: warmup did not complete for $name" >&2
+        case_failed=1
+      }
+    fi
+
+    if ((case_failed == 0)); then
+      k6 run \
+        --config "$CI_CONFIG" \
+        --tag "case_id=$name" \
+        --tag phase=measure \
+        --summary-export "$summary" \
+        -e NODE_URL="$NODE_URL" \
+        "$RUN_SCRIPT" <"$corpus" || case_failed=1
+    fi
+    ((case_failed == 0)) || failed+=("$name")
+  done < <(jq -r 'keys_unsorted[]' "$CATALOG")
+
+  if ((${#failed[@]} > 0)); then
+    echo "failed: ${failed[*]}" >&2
+    return 1
+  fi
+}
+
+nightly_run() {
+  preflight
+
+  local failed=0
+  generate || failed=1
+  benchmark || failed=1
+  return "$failed"
+}
+
+nightly_run
--- a/.dockerignore
+++ b/.dockerignore
@@ -11,6 +11,9 @@ coverage/
 **/target/debug
 **/target/release
 
+bench/rpc/corpus/*
+!bench/rpc/corpus/*.json
+
 # GitHub metadata 
 .github/
 
--- a/bench/rpc/nightly/ci.json
+++ b/bench/rpc/nightly/ci.json
@@ -1,0 +1,9 @@
+{
+  "vus": 50,
+  "duration": "30s",
+  "summaryTrendStats": ["avg", "min", "med", "p(90)", "p(95)", "p(99)", "max"],
+  "thresholds": {
+    "checks": ["rate==1"],
+    "http_req_failed": ["rate==0"]
+  }
+}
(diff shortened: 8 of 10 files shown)


bench/rpc/nightly.sh
```diff
@@ -0,0 +1,101 @@
...
+  }
+}
+
+generate() {
+  export CORPUS_ROOT=${CORPUS_ROOT:-/corpus/cases}
+  exec "$SCRIPT_DIR/gen-all.sh" "$CATALOG" "$NODE_URL" \
+    --count="${COUNT:-10000}" \
+    --seed="${SEED:-1}" \
+    --concurrency="${CONCURRENCY:-32}"
+}
```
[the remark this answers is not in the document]
danielntmd: argo handles this

bench/rpc/gen-all.sh
```diff
@@ -30,29 +31,203 @@ if [[ ! -f "$CONFIG" ]]; then
...
-  echo "==> $name"
-  # Node URL first, then pass-through, then per-entry flags (pflag: last wins).
-  "$CORPUS_GEN" --source-url "$NODE_URL" "${GEN_ARGS[@]}" "$@" >"$OUT_DIR/$name.json" ||
-    { rm -f "$OUT_DIR/$name.json"; return 1; }
+generation_arguments() {
+  local command=$1
+  local -a sub
+  read -ra sub <<<"$command"
+  jq -cn --args '$ARGS.positional' -- "${GEN_ARGS[@]}" "${sub[@]}"
 }
```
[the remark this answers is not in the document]
danielntmd: Acknowledged, and defer semantic argument normalization.

bench/rpc/run-all.sh
```diff
@@ -17,45 +20,208 @@ source "$SCRIPT_DIR/resolve-node.sh"
...
+      --tag "case_id=$name" \
+      --tag phase=warmup \
+      --summary-mode=disabled \
+      --vus 1 \
+      --iterations "$WARMUP_ITERATIONS" \
+      --duration "$WARMUP_MAX_DURATION" || {
+        echo "error: warmup did not complete for $name" >&2
+        case_failed=1
+      }
+  fi
```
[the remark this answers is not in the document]
danielntmd: intentional, will tune if we need more time

bench/rpc/nightly.sh
```diff
@@ -0,0 +1,101 @@
...
+
+rpc_result() {
+  local method=$1
+  curl --fail --silent --show-error \
+    --header 'Content-Type: application/json' \
+    --data "$(jq -cn --arg method "$method" \
+      '{jsonrpc:"2.0",id:1,method:$method,params:[]}')" \
+    "$NODE_URL" |
+    jq -er 'if .error then error(.error.message // "JSON-RPC error") else .result end'
+}
```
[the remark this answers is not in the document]
danielntmd: addressed

bench/rpc/nightly/Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+
+COPY bench/rpc/cmd/corpus-gen/ bench/rpc/cmd/corpus-gen/
+RUN CGO_ENABLED=0 go build -o /out/corpus-gen ./bench/rpc/cmd/corpus-gen
+
+FROM docker.io/grafana/k6:1.7.1@sha256:4fd3a694926b064d3491d9b02b01cde886583c4931f1223816e3d9a7bdfa7e0f
+
+USER root
+RUN apk add --no-cache bash curl jq && \
+    mkdir -p /corpus /results && \
+    chown k6:k6 /corpus /results
```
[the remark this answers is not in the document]
danielntmd: handled by argo

docs/static/benchmark-results/rpc/index.html
```diff
@@ -0,0 +1,138 @@
...
+        options.push(latest);
+        for (const entry of history.runs) {
+          const option = document.createElement('option');
+          option.value = entry.summary;
+          option.textContent = `${entry.startedAt} · ${entry.status} · ${entry.junoVersion || entry.commit}`;
+          options.push(option);
+        }
+        runs.replaceChildren(...options);
+        runs.disabled = history.runs.length === 0;
+      } catch (_) { /* Latest remains usable before Argo creates history.json. */ }
```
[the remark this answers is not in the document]
danielntmd: deferred DiD.

bench/rpc/gen-all.sh
```diff
@@ -30,29 +31,203 @@ if [[ ! -f "$CONFIG" ]]; then
...
+      .corpus.method == $generated[0].meta.method and
+      .corpus.count == $generated[0].meta.count and
+      .corpus.seed == $generated[0].meta.seed and
+      .rpcVersion == $generated[0].meta.rpcVersion
+    ' manifest.json >/dev/null; then
+    echo "error: managed corpus metadata does not match current configuration for $name" >&2
+    return 1
+  fi
+  printf '%s  corpus.json\n' "$(jq -r '.corpus.sha256' manifest.json)" |
+    sha256sum -c - >/dev/null
```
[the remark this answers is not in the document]
danielntmd: deferred nit, both conditions fail safely

bench/rpc/run-all.sh
```diff
@@ -17,45 +20,208 @@ source "$SCRIPT_DIR/resolve-node.sh"
...
+
 failed=()
-for corpus in "${corpora[@]}"; do
-  name="$(basename "$corpus" .json)"
+while IFS= read -r name; do
+  command=$(jq -er --arg name "$name" '.[$name]' "$CONFIG")
+  summary="$OUT_DIR/$name.json"
+  dashboard="$OUT_DIR/$name.html"
+  result_manifest="$OUT_DIR/$name.manifest.json"
+  rm -f "$summary" "$dashboard" "$result_manifest"
```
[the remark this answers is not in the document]
danielntmd: N/A, handled by argo

bench/rpc/run.js
```diff
@@ -19,10 +19,21 @@ export function parseIntStrict(raw, label) {
...
+}
+export const options = { thresholds };
+
 const NODE_URL = requiredEnv('NODE_URL');
+// run-all passes a file; throughput.js retains its redirected-stdin interface.
+const CORPUS_FILE = __ENV.CORPUS_FILE || '/dev/stdin';
 
-// k6 open() cannot seek a pipe, so the corpus must be redirected in (`< file`), not piped.
-const corpus = new SharedArray('corpus', () => JSON.parse(open('/dev/stdin')).requests);
+const corpus = new SharedArray('corpus', () => JSON.parse(open(CORPUS_FILE)).requests);
```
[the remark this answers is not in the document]
danielntmd: deferred, corpus-gen required count >= 1 anyways

bench/rpc/run-all.sh
```diff
@@ -71,14 +237,88 @@ report="$OUT_DIR/report.md"
...
+        schemaVersion: 1,
+        run: {id: $run_id, status: $status, startedAt: $started_at, finishedAt: $finished_at},
+        benchmark: {commit: ($commit | optional), imageDigest: ($benchmark_image | optional)},
+        juno: {version: ($juno_version | optional), imageDigest: ($juno_image | optional), chainId: ($chain_id | optional),
+          blockNumber: $block_number, rpcVersion: ($rpc_version | optional)},
+        snapshot: {id: ($snapshot_id | optional), sha256: ($snapshot_sha | optional), blockNumber: $block_number},
+        warmupIterations: $warmup,
+        cases: $cases
+      }
+    ' >"$summary_tmp"
```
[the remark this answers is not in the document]
danielntmd: addressed

bench/rpc/gen-all.sh
```diff
@@ -30,29 +31,203 @@ if [[ ! -f "$CONFIG" ]]; then
...
+  if [[ -n ${CORPUS_ROOT:-} ]]; then
+    if [[ -d $CORPUS_ROOT/$name ]]; then
+      if ! verify_managed "$CORPUS_ROOT/$name" "$name" "$command"; then
+        failed+=("$name")
+        continue
+      fi
+      echo "reusing $name"
+    elif ! generate_managed "$name" "$command"; then
+      failed+=("$name")
+    fi
```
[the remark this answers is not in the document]
danielntmd: accepted design. snapshot rotation will invalidate.

.github/workflows/rpc-benchmark.yaml
```diff
@@ -0,0 +1,151 @@
+name: RPC Benchmark Image
+
+on:
+  pull_request:
+    paths:
+      - "bench/rpc/**"
+      - ".github/workflows/rpc-benchmark.yaml"
+      - "go.mod"
+      - "go.sum"
+      - "starknet-p2p-specs/go.mod"
```
infrmtcs: Shouldn't we also run it when main Go code changes?

bench/rpc/gen-all.sh
infrmtcs, line 1: I strongly believe that we're mixing too many logic here. The existing script is self-contained to generate any corpus set locally. If we want to extend to support CI runs like the added code, it should be in its own script, not this script.

bench/rpc/gen-all.sh
```diff
@@ -30,29 +31,203 @@ if [[ ! -f "$CONFIG" ]]; then
...
 fi
+if ! jq -e '
+  type == "object" and length > 0 and
+  all(to_entries[];
+    (.key | test("^[A-Za-z0-9][A-Za-z0-9._-]*$")) and
+    (.value | type == "string" and test("[^[:space:]]")))
+' "$CONFIG" >/dev/null 2>&1; then
+  echo "error: $CONFIG must map safe case IDs to generator commands" >&2
+  exit 1
+fi
```
infrmtcs: This doesn't help, because eventually we still have to pass the arg as string and rely on the corpus-gen command parser to validate them. I suggest to remove this.

bench/rpc/gen-all.sh
```diff
@@ -30,29 +31,203 @@ if [[ ! -f "$CONFIG" ]]; then
...
+        rpcVersion: $rpc_version,
+        corpus: {
+          file: "corpus.json",
+          sha256: $sha,
+          method: $method,
+          count: $count,
+          seed: $seed
+        }
+      }
+    ' >"$work/manifest.json"
```
infrmtcs: `corpus-gen` already had metadata. If we want to extend it, we should extend there, not here.

bench/rpc/gen-all.sh
infrmtcs: If we need to verify the corpus, we can write a separate `verify-all.sh` script, although I doubt we need it if we're running on the corpus we generated ourselves.
If what we want is just to avoid overwriting existing corpus, just throw an error if the folder exists should be already enough.

bench/rpc/README.md
```diff
@@ -61,13 +61,13 @@ mkdir -p bench/rpc/corpus/v0_10
 Shared iterations — sequential latency:
 
 ```
-k6 run bench/rpc/run.js -e NODE_URL=http://localhost:6060/v0_10 --vus 1 --iterations 200 < bench/rpc/corpus/v0_10/getTransactionByHash.json
+k6 run bench/rpc/run.js -e NODE_URL=http://localhost:6060/v0_10 -e CORPUS_FILE="$PWD/bench/rpc/corpus/v0_10/getTransactionByHash.json" --vus 1 --iterations 200
```
infrmtcs: Stdin pipe was intentional. By doing so, we don't have to worry about whether the path is relative to the script dir, or the shell path. Please revert this change.

bench/rpc/README.md
```diff
@@ -78,7 +78,8 @@ Ramping arrival rate — fixed offered load, req/s:
 k6 run bench/rpc/throughput.js -e NODE_URL=http://localhost:6060/v0_10 -e RATES=1000,2000,3000 -e DURATION=5s < bench/rpc/corpus/v0_10/getTransactionByHash.json
 ```
 
-Corpus must be redirected in (`<`), not piped. k6 prints its summary to stdout;
+`run.js` loads the corpus from `CORPUS_FILE`; `throughput.js` reads it from
+redirected stdin. k6 prints its summary to stdout;
```
infrmtcs: I think this illustrates the problem mentioned above. Requiring different flags for different scenarios is fine, but 2 different input methods don't really make sense.

bench/rpc/README.md
```diff
@@ -106,3 +107,101 @@ Corpus must be redirected in (`<`), not piped. k6 prints its summary to stdout;
...
+export SNAPSHOT_ID=juno_mainnet_v0.16.5_12945535.tar.zst
+export SNAPSHOT_SHA256=845c9960da60678a821a32a6c44e7b68fc6a3869743d751a0e5ff7ccf83384bd
+export SNAPSHOT_BLOCK_NUMBER=12945535
+export RPC_VERSION=0.10.3
+
+docker compose -f bench/rpc/compose.yaml build
+docker compose -f bench/rpc/compose.yaml --profile corpus run --rm corpus preflight
+docker compose -f bench/rpc/compose.yaml --profile corpus run --rm corpus
+docker compose -f bench/rpc/compose.yaml run --rm benchmark
+```
```
infrmtcs: In this era of AI, I believe less is more. More texts mean less chance of human review. Without human review anywhere, neither code nor docs, will only result in a single end which is hallucination.
Please tell the agent to be "brutally concise", following the existing style.

bench/rpc/run-all.sh
infrmtcs, line 1: Similar to gen-all, I strongly believe that we're mixing too many logic here. The existing script is self-contained to run any corpus set locally. If we want to extend to support CI runs like the added code, it should be in its own script, not this script.

bench/rpc/run.js
```diff
@@ -19,10 +19,21 @@ export function parseIntStrict(raw, label) {
...
 
+const thresholds = {};
+if (__ENV.WARMUP_TARGET) {
+  thresholds['iterations{phase:warmup}'] = [`count==${__ENV.WARMUP_TARGET}`];
+}
+if (__ENV.STRICT_CHECKS === '1') {
+  thresholds['checks{phase:measure}'] = ['rate==1'];
+  thresholds['http_req_failed{phase:measure}'] = ['rate==0'];
+}
+export const options = { thresholds };
```
infrmtcs: Why don't we use this: https://grafana.com/docs/k6/latest/using-k6/k6-options/how-to/#set-options-with-the-config-flag
ci config can live in ci.json.

docs/static/benchmark-results/rpc/index.html
infrmtcs, line 1: If I understand correctly, this folder is for the docs we publish on juno docs website. I'm not sure if we want to publish these info there.

.github/workflows/rpc-benchmark.yaml
```diff
@@ -0,0 +1,151 @@
...
+                RUN_ID=non-strict WARMUP_ITERATIONS=0 STRICT_CHECKS=0 K6_WEB_DASHBOARD=false \
+                /app/bench/rpc/run-all.sh /results/smoke.json http://127.0.0.1:1 \
+                  --vus 1 --iterations 1 >/dev/null 2>&1 && \
+              jq -e '\''.run.status == "passed" and .cases[0].status == "passed" and
+                .cases[0].checkFailureRate > 0'\'' \
+                /results/non-strict/summary.json >/dev/null && \
+              k6 archive /source/throughput.js \
+                -e NODE_URL=http://127.0.0.1:1 -e RATES=1 -e DURATION=1s \
+                -O /results/throughput.tar < /results/smoke/smoke.json >/dev/null && \
+              test -s /results/throughput.tar'
```
infrmtcs: Why?

infrmtcs (changes requested):

bench/rpc/run-all.sh
```diff
@@ -17,45 +20,208 @@ source "$SCRIPT_DIR/resolve-node.sh"
...
+trap 'forward_signal TERM 143' TERM
+
+run_k6() {
+  local status=0
+  k6 "$@" &
+  active_pid=$!
+  wait "$active_pid" || status=$?
+  active_pid=
+  return "$status"
+}
```
infrmtcs: Why do we need these lines?

bench/rpc/run-all.sh
infrmtcs: Why remove? If we want to improve the existing script, it should be throw an error if the folder already exists.

bench/rpc/run-all.sh
```diff
@@ -17,45 +20,208 @@ source "$SCRIPT_DIR/resolve-node.sh"
...
+      continue
+    fi
+    if ! verify_manifest "$corpus" "$manifest" "$name" "$command"; then
+      failed+=("$name")
+      continue
+    fi
+  fi
+
+  case_failed=0
+  if ((WARMUP_ITERATIONS > 0)); then
```
infrmtcs: If we want to do warmup, let's define the full scenario with warmup in a js file, not here.

bench/rpc/run-all.sh
```diff
@@ -17,45 +20,208 @@ source "$SCRIPT_DIR/resolve-node.sh"
...
+    ' "$manifest"); then
+    echo "error: invalid or mismatched corpus manifest: $manifest" >&2
+    return 1
+  fi
+  actual=$(sha256sum "$corpus" | awk '{print $1}')
+  if [[ $actual != "$expected" ]]; then
+    echo "error: corpus checksum does not match $manifest" >&2
+    return 1
+  fi
+}
```
infrmtcs: What is the purpose of this? By enforcing the schema in a non-typesafe language, we disable our ability to do things like these, because we don't have a good metadata to write: 
- Merge multiple corpora into one
- Shuffle a corpus

bench/rpc/README.md
```diff
@@ -106,3 +107,101 @@ Corpus must be redirected in (`<`), not piped. k6 prints its summary to stdout;
 | `--duration <time>`             | closed (`run.js`)      | stop after this run length (`30s`)                            |
 | `-e RATES=<r1,r2,...>`          | open (`throughput.js`) | ramp targets, req/s (`1000,2000,3000`)                        |
 | `-e DURATION=<time>`            | open (`throughput.js`) | per-stage ramp time (`5s`)                                    |
+
+## Managed corpus storage
+
+`corpus/all.json` is the sole active-case catalog. A case ID represents one
```
infrmtcs: This isn't the current behaviour. The current behaviour is that we can name it anything (maybe except nodes, because of the special `nodes.json` file)

So in order to create a new catalog, we can just add nightly.json containing all cases we want to run on CI.

Changing to this disable the ability to maintain multiple catalog locally.

bench/rpc/README.md
```diff
@@ -106,3 +107,101 @@ Corpus must be redirected in (`<`), not piped. k6 prints its summary to stdout;
...
+catalog command, complete generator argument vector, creation date, generating
+Juno commit, snapshot, RPC version, and corpus checksum.
+
+```sh
+CORPUS_ROOT=/corpus/cases \
+JUNO_COMMIT=$(git rev-parse HEAD) \
+SNAPSHOT_ID=juno_mainnet_v0.16.5_12945535.tar.zst \
+SNAPSHOT_SHA256=845c9960da60678a821a32a6c44e7b68fc6a3869743d751a0e5ff7ccf83384bd \
+SNAPSHOT_BLOCK_NUMBER=12945535 \
+RPC_VERSION=0.10.3 \
```
infrmtcs: Why provide RPC_VERSION? 
corpus-gen itself should query from the node to get the answer, instead of trusting an input.

bench/rpc/README.md
```diff
@@ -106,3 +107,101 @@ Corpus must be redirected in (`<`), not piped. k6 prints its summary to stdout;
...
+```
+
+Existing cases are checksum-verified and reused. Missing cases are generated
+in staging and atomically moved into place. The manifest records the ID,
+catalog command, complete generator argument vector, creation date, generating
+Juno commit, snapshot, RPC version, and corpus checksum.
+
+```sh
+CORPUS_ROOT=/corpus/cases \
+JUNO_COMMIT=$(git rev-parse HEAD) \
```
infrmtcs: If we need the version of the corpus-gen, corpus-gen should itself provide it instead of passing an env var here.

bench/rpc/nightly/internal/corpus.sh
```diff
@@ -0,0 +1,71 @@
...
+  work=
+  trap - EXIT INT TERM
+)
+
+prune_removed_cases() {
+  local case_dir name
+  for case_dir in "$CORPUS_ROOT"/*; do
+    [[ -d $case_dir ]] || continue
+    name=${case_dir##*/}
+    if ! jq -e --arg name "$name" 'has($name)' "$CATALOG" >/dev/null; then
```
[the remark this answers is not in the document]
danielntmd: validated by CI

bench/rpc/nightly/nightly.sh
```diff
@@ -0,0 +1,101 @@
...
+nightly_run() {
+  preflight
+
+  local failed=0
+  generate || failed=1
+  benchmark || failed=1
+  return "$failed"
+}
+
+nightly_run
```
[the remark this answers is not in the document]
danielntmd: handled by argo

bench/rpc/nightly/internal/corpus.sh
```diff
@@ -0,0 +1,71 @@
...
+    return 1
+  }
+  mv -T "$work" "$case_dir"
+  work=
+  trap - EXIT INT TERM
+)
+
+prune_removed_cases() {
+  local case_dir name
+  for case_dir in "$CORPUS_ROOT"/*; do
```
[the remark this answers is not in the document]
danielntmd: accepted behavior, snapshots will not change

.github/workflows/rpc-benchmark.yaml
```diff
@@ -0,0 +1,88 @@
...
+        uses: docker/setup-buildx-action@bb05f3f5519dd87d3ba754cc423b652a5edd6d2c # v4.2.0
+
+      - name: Build benchmark image
+        uses: docker/build-push-action@53b7df96c91f9c12dcc8a07bcb9ccacbed38856a # v7.0.0
+        with:
+          context: .
+          file: bench/rpc/nightly/Dockerfile
+          tags: ${{ env.DOCKER_REGISTRY }}/${{ env.BENCHMARK_IMAGE_REPOSITORY }}:test
+          cache-from: type=gha,scope=rpc-benchmark
+          cache-to: type=gha,mode=max,scope=rpc-benchmark
```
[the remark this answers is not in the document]
danielntmd: accepted CI tradeoff

bench/rpc/corpus/nightly.json
```diff
@@ -0,0 +1,29 @@
+{
+  "blockHashAndNumber": {"command": "blockHashAndNumber", "count": 100000, "seed": 1},
+  "blockNumber": {"command": "blockNumber", "count": 100000, "seed": 1},
+  "chainId": {"command": "chainId", "count": 100000, "seed": 1},
+  "syncing": {"command": "syncing", "count": 100000, "seed": 1},
```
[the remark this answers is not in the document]
danielntmd: deliberate, they will naturally deviate, and `traceBlockTransactions` and `traceTransactions` have 10,000 instead.

bench/rpc/nightly/internal/corpus.sh
```diff
@@ -0,0 +1,80 @@
...
+  work=$(mktemp -d "$CORPUS_ROOT/.staging/$name.XXXXXX")
+  trap 'rm -rf "$work"' EXIT INT TERM
+
+  read -ra sub <<<"$command"
+  "$CORPUS_GEN" --source-url "$NODE_URL" \
+    --count="$count" \
+    --seed="$seed" \
+    --concurrency="${CONCURRENCY:-32}" \
+    "${sub[@]}" >"$work/corpus.json" || return 1
+  printf '%s\n' "$command" >"$work/command"
```
danielntmd, line 14: Temporary sidecar to handle flags in `command` and future flags without duplicating parsing logic. Long-term solution should be to have `corpus-gen` expose a fingerprint in its `.meta`.

bench/rpc/nightly/nightly.sh
```diff
@@ -0,0 +1,101 @@
...
+  }
+  local out_dir=${RESULTS_DIR:-/results}
+  [[ -n ${K6_VUS:-} ]] || unset K6_VUS
+  [[ -n ${K6_DURATION:-} ]] || unset K6_DURATION
+  mkdir -p "$out_dir"
+
+  local name corpus summary case_failed
+  local -a failed=()
+  while IFS= read -r name; do
+    summary="$out_dir/$name.json"
```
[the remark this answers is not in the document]
danielntmd: summaries are not retained, this is a no-op.

bench/rpc/corpus/nightly.json
infrmtcs, line 1: I don't think we're currently supporting object as values like `{"command": "blockHashAndNumber", "count": 100000, "seed": 1}`

The reason why I decided to go with string was because as long as we still call `corpus-gen` via bash, it's simpler to just put what we want to call with corpus-gen rather than parsing the json and construct the call.

I think it makes sense to use this format if we finally remove `gen-all.sh` and replace by a subcommand in `corpus-gen`, but it takes more effort to do so.

bench/rpc/nightly/warmup.js
infrmtcs, line 1: The current state of this file can be run from default run.js, simply by setting `--vus 1 --iterations "
"$WARMUP_ITERATIONS"`

The scenario API is helpful if we want to define a more complicated scenarios with multiple stages and tags.

Therefore, I think we should either drop this file, or expand it to contains all steps you want to run in this benchmark

bench/rpc/nightly/Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+
+FROM golang:1.26-bookworm AS builder
+
+WORKDIR /src
+COPY go.mod go.sum ./
+COPY starknet-p2p-specs/go.mod starknet-p2p-specs/go.mod
+RUN go mod download
+
+COPY bench/rpc/cmd/corpus-gen/ bench/rpc/cmd/corpus-gen/
+RUN CGO_ENABLED=0 go build -o /out/corpus-gen ./bench/rpc/cmd/corpus-gen
```
infrmtcs: Can we also copy the Makefile and use it here?
The main idea is to reuse the same make command we define there, so any change in Makefile will be reflected here. 
It's especially useful because we don't run this build on CI on every PR.

bench/rpc/nightly/Dockerfile
infrmtcs, line 1: Building can be faster if we have a .dockerignore file in this format:
```
/bench/rpc
!/bench/rpc/file-you-want-to-use-in-dockerfile
```
I think this will be a problem when we put the generated corpus in `bench/rpc/corpus` which can contains GBs of data, making the docker build slower even though we don't need those files.
This also can help to simplify the COPY steps in Dockerfile.

bench/rpc/nightly/internal/corpus.sh
infrmtcs: Why don't we use `gen-all.sh` and the file format we're already using there?

bench/rpc/nightly/nightly.sh
```diff
@@ -0,0 +1,101 @@
...
+    echo "error: k6 config not found: $CI_CONFIG" >&2
+    return 1
+  }
+  [[ -f $WARMUP_SCRIPT ]] || {
+    echo "error: k6 warmup script not found: $WARMUP_SCRIPT" >&2
+    return 1
+  }
+  local out_dir=${RESULTS_DIR:-/results}
+  [[ -n ${K6_VUS:-} ]] || unset K6_VUS
+  [[ -n ${K6_DURATION:-} ]] || unset K6_DURATION
```
infrmtcs: Why do we need this? I think we can override via CLI flags, it will take highest precedence:
https://grafana.com/docs/k6/latest/using-k6/k6-options/how-to/#order-of-precedence

.github/workflows/rpc-benchmark.yaml
```diff
@@ -0,0 +1,88 @@
...
+
+  publish-image:
+    if: github.ref == 'refs/heads/main'
```
infrmtcs: Will this run on `workflow_dispatch`?

.github/workflows/rpc-benchmark.yaml
```diff
@@ -0,0 +1,88 @@
...
+        uses: docker/build-push-action@53b7df96c91f9c12dcc8a07bcb9ccacbed38856a # v7.0.0
+        with:
+          context: .
+          file: bench/rpc/nightly/Dockerfile
+          tags: |
+            ${{ env.DOCKER_REGISTRY }}/${{ env.BENCHMARK_IMAGE_REPOSITORY }}:latest
+            ${{ env.DOCKER_REGISTRY }}/${{ env.BENCHMARK_IMAGE_REPOSITORY }}:sha-${{ github.sha }}
+          cache-from: type=gha,scope=rpc-benchmark
+          cache-to: type=gha,mode=max,scope=rpc-benchmark
+          push: true
```
infrmtcs: I think having a separate job for publish will result in a lot of rebuilds, because the 2 jobs are running on 2 separate machines. Even though we're setting `cache-from` and `cache-to` here, I'm not sure about whether the image layers built from the `build-image` will be available here immediately.

If the analysis above is correct, I think we can add the `if` condition to Install and Login steps, while using the if condition as the value for `push:`

infrmtcs (commented):
This version is much better than the previous one. I've reviewed ~80% of this PR and left these comments, I think the remaining can be reviewed after we address these.

.github/workflows/rpc-benchmark.yaml
danielntmd: Yes, if the workflow dispatch targets main, it will publish.

bench/rpc/nightly/internal/corpus.sh
danielntmd: gen-all regenerates everything and we need to reconcile existing corpus data as we are trying to avoid having to generate the entire corpus every night. 

And since corpus-gen doesn't have a way to generate a fingerprint per corpora, there is no clean way to compare existing corpora against the new catalog without regenerating everything, duplicating the corpus-gen parsing logic, or using this sidecar.

bench/rpc/nightly/Dockerfile
danielntmd: Corpus will be stored under a PVC; this concern is not valid.

But I will still add the .dockerignore as a preventative guard in case the corpus is ever pushed to the repo.

bench/rpc/nightly/warmup.js
danielntmd: Will remove for now,  and if complex warmup scenarios are needed, we can implement it then.

bench/rpc/nightly/Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+FROM golang:1.26-bookworm AS builder
+
+WORKDIR /src
+COPY go.mod go.sum ./
+COPY starknet-p2p-specs/go.mod starknet-p2p-specs/go.mod
+RUN go mod download
+
+COPY Makefile ./
+COPY bench/rpc/cmd/corpus-gen/ bench/rpc/cmd/corpus-gen/
+RUN CGO_ENABLED=0 make corpus-gen
```
infrmtcs, line 12: Why do we need to disable CGO_ENABLED here?

infrmtcs (approved):

bench/rpc/nightly/Dockerfile
danielntmd: corpus-gen is compiled in a Debian-based Go builder and k6 image is Alpine-based. Without it, the binary links against Debian's glibc which Alpine does not have. We only disable CGO while compiling corpus-gen.

.github/workflows/build-image.yaml
```diff
@@ -62,6 +62,6 @@ jobs:
           context: .
           push: true
           platforms: linux/amd64
-          tags: ${{ env.DOCKER_REGISTRY }}/${{ env.REPO_DEV }}/juno:${{ steps.set_tag.outputs.DOCKER_IMAGE_TAG  }}
+          tags: ${{ env.DOCKER_REGISTRY }}/${{ env.REPO_DEV }}/juno:${{ steps.set_tag.outputs.DOCKER_IMAGE_TAG }}${{ github.ref == 'refs/heads/main' && format(',{0}/{1}/juno:main', env.DOCKER_REGISTRY, env.REPO_DEV) || '' }}
```
danielntmd, line 65: This adds a juno main tag which is pushed on main. This is due to the way the current infra is setup, we use latest tag for PR development pushes which is unsuitable for benchmark nightly.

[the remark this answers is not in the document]
danielntmd: no-op concern, argo handles it

PR #3953, opened 2026-08-18
perf(jsonrpc): drop discarded allocations and chunked response

NazariiDenha:
Three small fixes:
1. Don't allocate empty `Header` twice
2. Don't build response that discarded then
3. Set `Content-Length` header to remove extra terminating chunk

### DIFF
--- a/jsonrpc/http.go
+++ b/jsonrpc/http.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"maps"
 	"net/http"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
@@ -141,6 +142,8 @@ func (h *HTTP) ServeHTTP(writer http.ResponseWriter, req *http.Request) {
 				}
 			}()
 			ioWriter = gw
+		} else if err == nil {
+			writer.Header().Set("Content-Length", strconv.Itoa(len(resp)))
 		}
 		_, err = ioWriter.Write(resp)
 		if err != nil {
--- a/jsonrpc/server.go
+++ b/jsonrpc/server.go
@@ -70,8 +70,8 @@ type response struct {
 	ID      any    `json:"id"`
 }
 
-func errResponse(code int, data any) *response {
-	return &response{Version: "2.0", Error: Err(code, data)}
+func errResponse(code int, data any) response {
+	return response{Version: "2.0", Error: Err(code, data)}
 }
 
 type Error struct {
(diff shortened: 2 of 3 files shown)


jsonrpc/server.go
```diff
@@ -70,6 +70,10 @@ type response struct {
 	ID      any    `json:"id"`
 }
 
+func errResponse(code int, data any) *response {
+	return &response{Version: "2.0", Error: Err(code, data)}
+}
+
```
rodrodros, line 76: Why not return by value here?

jsonrpc/server.go
```diff
@@ -349,24 +353,22 @@ func (s *Server) HandleReader(ctx context.Context, reader io.Reader) ([]byte, ht
...
+	var header http.Header
 
 	dec := json.NewDecoder(bufferedReader)
 	dec.UseNumber()
 
 	if !requestIsBatch {
 		req := new(Request)
 		if jsonErr := dec.Decode(req); jsonErr != nil {
-			resp.Error = Err(InvalidJSON, prettyParseError(&errorRecoverBuffer, jsonErr))
+			resp = errResponse(InvalidJSON, prettyParseError(&errorRecoverBuffer, jsonErr))
```
rodrodros: then here you can do `resp = new(errResponse(...))` 🤔 and still get the pointer benefit

jsonrpc/server.go
```diff
@@ -349,24 +353,22 @@ func (s *Server) HandleReader(ctx context.Context, reader io.Reader) ([]byte, ht
 	var errorRecoverBuffer windowBuffer
 	bufferedReader := bufio.NewReaderSize(io.TeeReader(reader, &errorRecoverBuffer), bufferSize)
 	requestIsBatch := isBatch(bufferedReader)
-	resp := &response{
-		Version: "2.0",
-	}
 
-	header := http.Header{}
+	var resp *response
```
rodrodros, line 357: Although, as far as I can see, resp can be a value type as well?

rodrodros:
Please use commit lint on PR title

jsonrpc/http.go
```diff
@@ -141,6 +142,8 @@ func (h *HTTP) ServeHTTP(writer http.ResponseWriter, req *http.Request) {
 				}
 			}()
 			ioWriter = gw
+		} else if err == nil {
+			writer.Header().Set("Content-Length", strconv.Itoa(len(resp)))
```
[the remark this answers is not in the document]
rodrodros: Valid

rodrodros:
Also, please rebase from main on your branches, instead of merging. No big deal now, but it helps with tidiness with organizing PRs with manny commits or features implementation spanning more than one PR

jsonrpc/server.go
NazariiDenha: we check later if resp is nil, that means no response should be sent. Would need to have extra flag then if use value type

jsonrpc/server.go
NazariiDenha: we use `resp` as arg to `Marshal(v any)` later anyway

jsonrpc/server.go
NazariiDenha: we will need to do then
```
r := errResponse()
resp = &r
```

extra line in few places

rodrodros: You don't need to do two lines, because you can use `new`: https://go.dev/doc/go1.26#language

It's a nit though, only one small allocation served. Feel free apply or ignore

jsonrpc/http.go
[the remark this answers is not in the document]
rodrodros: Ready to approve and merge once we have a test asserting the new behaviour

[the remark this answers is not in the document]
NazariiDenha: added test

jsonrpc/server.go
NazariiDenha: return value now and use `new()`

rodrodros (approved):

PR #3959, opened 2026-08-20
perf(utils/compression): add buffer pool

rodrodros:


rodrodros:
Closing as there were no big performance improvements

PR #3962, opened 2026-08-21
fix(trie2): stop hashdb dirty cache storing class nodes twice

rodrodros:
`dirtyCache.putNode` dispatched on `isClass` without returning, so it fell through into the owner check below. Class nodes are always inserted with a zero owner (`Database.Update`), so every class node landed in both `classNodes` and `contractNodes`, doubling dirty-cache memory and causing `Commit` to persist each class node a second time under the `db.ContractTrieContract` prefix.

Reads were unaffected — `getNode` already early-returns on `isClass`, and disk reads are bucket-prefixed point lookups — so this was silent apart from the wasted memory, write IO, and polluted contract-trie keyspace. Adding the `return` matches the dispatch used by `getNode` and by pathdb's `nodeSet.node`.

Also fixes two accounting defects in the same struct: `len()` summed `len(contractStorageNodes)`, the owner count rather than the storage node count, and `size` was never incremented so `Commit` always logged `liveSize=0`. Drop the dead field and the log fields that only ever reported zero.

### DIFF
--- a/core/trie2/triedb/hashdb/database.go
+++ b/core/trie2/triedb/hashdb/database.go
@@ -178,10 +178,8 @@ func (d *Database) Commit(_ *felt.StateRootHash) error {
 	d.dirtyCache.reset()
 
 	d.logger.Debug("Flushed dirty cache to disk",
-		zap.Int("nodes", nodes-d.dirtyCache.len()),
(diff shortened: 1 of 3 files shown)


rodrodros:
Outside of the production path, will merge without review if CI goes green

